Issue #884, opened 2023-01-11
Tasks for ops 2.1 release


benhoyt:
For the ops 2.1 release, we'd like to add a couple of new APIs and do a bit of code cleanup:

* [x] Add API support for `open-port` et al: https://github.com/canonical/operator/issues/179
* [x] Use f-strings (Python 3.6+) everywhere: https://github.com/canonical/operator/pull/889
* [x] Use variable type annotations (Python 3.6+) instead of `# type: T` comments: https://github.com/canonical/operator/pull/904
* [ ] Refactor large files into directories (but be careful that it's API-compatible!)
* [ ] Consider moving a curated set of names to the `ops` top-level. See https://github.com/canonical/operator/issues/731
* [x] Un-vendor websockets library (it was only vendored as latest doesn't work in Python 3.5): https://github.com/canonical/operator/pull/891
* [x] Update versions of Pyright and lock versions of other dependencies: https://github.com/canonical/operator/pull/894
* [ ] ~Consider making ops a "namespace package" so we can exclude testing by default. See https://github.com/canonical/operator/issues/865#issuecomment-1370635292~

See also [tasks for ops 2.0 release issue](https://github.com/canonical/operator/issues/865).

tinvaan:
Hi @benhoyt, have tried to address the last two items on the release checklist.

- https://github.com/canonical/operator/pull/891
- https://github.com/tinvaan/operator/pull/1

The changes need a bit of polishing and thorough testing. Happy to iron them out and ready them for the 2.1 release.

sed-i:
I keep finding namespaced status painfully missing
https://github.com/canonical/operator/issues/665

benhoyt:
For the record, I moved "Add bytes API for Secrets: https://github.com/canonical/operator/issues/869" off this list. It's interesting but definitely doesn't seem to be high priority.

benhoyt:
@sed-i FYI, I've added the namespaced/scope status bullet point under a new draft [Tasks for ops 2.2 release](https://github.com/canonical/operator/issues/906) issue.

benhoyt:
I've thought further about "namespace packages", and I'm going to pass on those for now. We can revisit this later, or consider something like `pip install ops[scenario]` when `ops-scenario` stabilized and if we decide to promote that as the native approach to testing charms.

benhoyt:
Closing this now that 2.1.0 is released: https://github.com/canonical/operator/releases/tag/2.1.0

[...]

PR #891, opened 2023-01-16
Un-vendor websocket library

tinvaan:
Drops the vendored `websocket` library in favor of the pypi offering instead. Part of the [ops 2.1 release tasks](https://github.com/canonical/operator/issues/884).

## QA steps

On Linux, run `tox -e pebble` to run the "real Pebble tests".

## Changelog

- *Drops vendored websocket library*
- *Adds websocket-client to the project's requirements*


rbarry82:
@tinvaan are you building on MacOS? Pebble could/should vendor deps (I'm not sure why it doesn't), but this sounds like a MacOS problem. `go get -u golang.org/x/sys/unix` may work, or `GOOS=linux go install ...`

benhoyt (changes requested):
Hi @tinvaan, this is great -- thanks for your contribution! I've pushed a few commits to get the charm tests working, the main thing that was needed was adding the `websocket-client` dependency to `setup.py` (that's what's actually used when `pip` installs the package, `requirements.txt` is more for local development). I also removed the `ops/_vendor` specific cruft from `pyproject.toml` and `test_infra.py` now that that directory is gone.

The one further change I'd request is that we *don't* include the typing stubs. We want to keep the `ops` footprint small, and these don't really gain us anything for users. So let's delete those again (don't worry about rebasing -- we do that when the PR is merged). You may have to tweak `pebble.py` or Pyright settings if Pyright complains with the stubs gone, I'm not sure.

benhoyt:
Oh, and as for why installing Pebble wasn't working for you -- yeah, it's almost certainly because you're developing on Windows or macOS (Pebble only builds on Linux). So you might be out of luck there (or you can run them in a VM or Docker container). But I've tested locally with my updates, and `tox -e pebble` works fine.

tinvaan:
ok, so I'm marking the PR ready for review now that I've heard that the pebble tests are passing.

> You may have to tweak pebble.py or Pyright settings if Pyright complains with the stubs gone, I'm not sure.
Yeah, so pyright complains loudly in the absence of the stubs. Would it be acceptable to `# type: ignore` the websocket client specific calls?

/cc @benhoyt

benhoyt:
> Yeah, so pyright complains loudly in the absence of the stubs. Would it be acceptable to # type: ignore the websocket client specific calls?

Hmmm, that's a bit of a pain (I'm still getting used to typing and Pyright myself). In any case, how about we include the minimum types for just the subset of `websocket.WebSocket` that we need, but include them in `ops/pebble.py` in the `if TYPE_CHECKING` block at the top? I was able to get pretty far with the following:

```python
    class _WebSocket(Protocol):
        def connect(self, url: str, socket: socket.socket): ...
        def shutdown(self): ...
        def send(self, payload: str): ...
        def send_binary(self, payload: bytes): ...
        def recv(self) -> typing.AnyStr: ...
```

And you should be able to remove most/all of the "type: ignore" comments you added too.

benhoyt:
Full git diff of my local hacking:

```diff
diff --git a/ops/pebble.py b/ops/pebble.py
index e0301f2..da9372b 100644
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -57,7 +57,7 @@ from typing import (
     Union,
 )
 
-import websocket
+import websocket  # type: ignore
 
 from ops._private import timeconv, yaml
 
@@ -215,6 +215,13 @@ if TYPE_CHECKING:
     _FilesResponse = TypedDict('_FilesResponse',
                                {'result': List[_Item]})
 
+    class _WebSocket(Protocol):
+        def connect(self, url: str, socket: socket.socket): ...
+        def shutdown(self): ...
+        def send(self, payload: str): ...
+        def send_binary(self, payload: bytes): ...
+        def recv(self) -> typing.AnyStr: ...
+
 logger = logging.getLogger(__name__)
 
 
@@ -1123,9 +1130,9 @@ class ExecProcess:
         stderr: Optional['_Writeable'],
         client: 'Client',
         timeout: Optional[float],
-        control_ws: websocket.WebSocket,
-        stdio_ws: websocket.WebSocket,
-        stderr_ws: Optional[websocket.WebSocket],
+        control_ws: '_WebSocket',
+        stdio_ws: '_WebSocket',
+        stderr_ws: Optional['_WebSocket'],
         command: List[str],
         encoding: Optional[str],
         change_id: ChangeID,
@@ -1190,10 +1197,10 @@ class ExecProcess:
             os.close(self._cancel_reader)
 
         # Close websockets (shutdown doesn't send CLOSE message or wait for response).
-        self._control_ws.shutdown()     # type: ignore
-        self._stdio_ws.shutdown()       # type: ignore
+        self._control_ws.shutdown()
+        self._stdio_ws.shutdown()
         if self._stderr_ws is not None:
-            self._stderr_ws.shutdown()  # type: ignore
+            self._stderr_ws.shutdown()
 
         if change.err:
             raise ChangeError(change.err, change)
@@ -1251,7 +1258,7 @@ class ExecProcess:
             'signal': {'name': sig},
         }
         msg = json.dumps(payload, sort_keys=True)
-        self._control_ws.send(msg)  # type: ignore
+        self._control_ws.send(msg)
 
 
 def _has_fileno(f: Any) -> bool:
@@ -1266,7 +1273,7 @@ def _has_fileno(f: Any) -> bool:
 
 
 def _reader_to_websocket(reader: '_WebsocketReader',
-                         ws: websocket.WebSocket,
+                         ws: '_WebSocket',
                          encoding: str,
                          cancel_reader: Optional[int] = None,
                          bufsize: int = 16 * 1024):
@@ -1283,16 +1290,16 @@ def _reader_to_websocket(reader: '_WebsocketReader',
             break
         if isinstance(chunk, str):
             chunk = chunk.encode(encoding)
-        ws.send_binary(chunk)  # type: ignore
+        ws.send_binary(chunk)
 
     ws.send('{"command":"end"}')  # type: ignore # Send "end" command as TEXT frame to signal EOF
 
 
-def _websocket_to_writer(ws: websocket.WebSocket, writer: '_WebsocketWriter',
+def _websocket_to_writer(ws: '_WebSocket', writer: '_WebsocketWriter',
                          encoding: str):
     """Receive messages from websocket (until end signal) and write to writer."""
     while True:
-        chunk: typing.AnyStr = ws.recv()  # type: ignore
+        chunk: _StrOrBytes = ws.recv()
 
         if isinstance(chunk, str):
             try:
@@ -1317,7 +1324,7 @@ def _websocket_to_writer(ws: websocket.WebSocket, writer: '_WebsocketWriter',
 class _WebsocketWriter(io.BufferedIOBase):
     """A writable file-like object that sends what's written to it to a websocket."""
 
-    def __init__(self, ws: websocket.WebSocket):
+    def __init__(self, ws: '_WebSocket'):
         self.ws = ws
 
     def writable(self):
@@ -1328,18 +1335,18 @@ class _WebsocketWriter(io.BufferedIOBase):
         """Write chunk to the websocket."""
         if not isinstance(chunk, bytes):
             raise TypeError(f'value to write must be bytes, not {type(chunk).__name__}')
-        self.ws.send_binary(chunk)  # type: ignore
+        self.ws.send_binary(chunk)
         return len(chunk)
 
     def close(self):
         """Send end-of-file message to websocket."""
-        self.ws.send('{"command":"end"}')  # type: ignore
+        self.ws.send('{"command":"end"}')
 
 
 class _WebsocketReader(io.BufferedIOBase):
     """A readable file-like object whose reads come from a websocket."""
 
-    def __init__(self, ws: websocket.WebSocket):
+    def __init__(self, ws: '_WebSocket'):
         self.ws = ws
         self.remaining = b''
         self.eof = False
@@ -1355,7 +1362,7 @@ class _WebsocketReader(io.BufferedIOBase):
             return b''
 
         while not self.remaining:
-            chunk: typing.AnyStr = self.ws.recv()  # type: ignore
+            chunk: _StrOrBytes = self.ws.recv()
 
             if isinstance(chunk, str):
                 try:
@@ -2191,13 +2198,13 @@ class Client:
         change_id = resp['change']
         task_id = resp['result']['task-id']
 
-        stderr_ws = None  # type: Optional[websocket.WebSocket]
+        stderr_ws = None  # type: Optional['_WebSocket']
         try:
             control_ws = self._connect_websocket(task_id, 'control')
             stdio_ws = self._connect_websocket(task_id, 'stdio')
             if not combine_stderr:
                 stderr_ws = self._connect_websocket(task_id, 'stderr')
-        except websocket.WebSocketException as e:
+        except websocket.WebSocketException as e:  # type: ignore
             # Error connecting to websockets, probably due to the exec/change
             # finishing early with an error. Call wait_change to pick that up.
             change = self.wait_change(ChangeID(change_id))
@@ -2246,7 +2253,7 @@ class Client:
                 t = _start_thread(_websocket_to_writer, stderr_ws, stderr, encoding)
                 threads.append(t)
             else:
-                ws = typing.cast(websocket.WebSocket, stderr_ws)
+                ws = typing.cast('_WebSocket', stderr_ws)
                 process_stderr = _WebsocketReader(ws)
                 if encoding is not None:
                     process_stderr = io.TextIOWrapper(
@@ -2270,12 +2277,12 @@ class Client:
         )
         return process
 
-    def _connect_websocket(self, task_id: str, websocket_id: str) -> websocket.WebSocket:
+    def _connect_websocket(self, task_id: str, websocket_id: str) -> '_WebSocket':
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         sock.connect(self.socket_path)
         url = self._websocket_url(task_id, websocket_id)
-        ws = websocket.WebSocket(skip_utf8_validation=True)
-        ws.connect(url, socket=sock)  # type: ignore
+        ws: '_WebSocket' = websocket.WebSocket(skip_utf8_validation=True)  # type: ignore
+        ws.connect(url, socket=sock)
         return ws
 
     def _websocket_url(self, task_id: str, websocket_id: str) -> str:
```

tinvaan:
Hmm, I'll try that out. Feel free to push your commits, even if it's WIP. I'll take a look.

benhoyt:
Pushed.

benhoyt (approved):
This is good to go as far as I'm concerned -- thanks again for the contribution.

This project requires two approvals on each PR, so I'll ask for another on this end too.

tinvaan:
Thanks @benhoyt, just tested your changes against the latest version of pyright(`1.1.289`, bumped up in https://github.com/tinvaan/operator/pull/1) and it looks good.

requirements.txt
```diff
@@ -1 +1,2 @@
 PyYAML
+websocket-client
```
jnsgruk, line 2: Could we pin this version?

setup.py
```diff
@@ -72,7 +72,10 @@ def _get_version() -> str:
             "Operating System :: POSIX :: Linux",
         ],
         python_requires='>=3.8',
-        install_requires=["PyYAML"],
+        install_requires=[
+            'PyYAML',
+            'websocket-client',
```
jnsgruk, line 77: Likewise here, I think pinning the version is probably a good idea.

jnsgruk (approved):
Overall I think this is a good change - though I think it's worth discussing if we should pin the version of the library to ensure as much as possible that when it's installed in the wild, people get the version we've tested.

We probably also want to introduce some automation to ensure we keep pinned deps fresh, but that's for a future PR (a lot of folks in Canonical are using Renovate for that).

requirements.txt
benhoyt: Yeah, definitely. We plan to do that in a separate PR where we pin all the versions (of dev dependencies too). See https://github.com/tinvaan/operator/pull/1

setup.py
benhoyt: Agreed -- see above.

benhoyt:
Regarding automation to keep deps fresh, I have mixed feelings about that, but yeah, that can be discussed and considered separately from this PR.

jameinel (approved):
I think in the general case, I'm broadly more in favor of not vendoring. I will say that 'ops' should not have very large dependency structures, which is one reason why we are using this particular dependency, as it was relatively lightweight. (Every charm ends up depending on any dependency that ops brings in.)

I do like having stable dependencies for everyone involved, but also we want to make sure we are giving people a good experience. I wonder about exact pinning, vs just pinning to Major.Minor (but not .Patch).

benhoyt:
@jameinel Yeah, I was just thinking about that in the shower. :-) Pinning sounds good, but I guess a library (like `ops`) shouldn't pin so tightly, otherwise the final "artifact" (in this case, the charm) wouldn't be able to use (pin) even a slightly different version. And it seems to me that the final artifact is what should have control over the strict pinning.

So I think maybe it should *just* be the Major version we pin, not even Minor. (Though we could also have a minimum Minor.)

requirements.txt
tinvaan: @jnsgruk fyi, requirements are pinned to versions in https://github.com/canonical/operator/pull/894

PR #892, opened 2023-01-16
Minor documentation changes

tinvaan:
Simply updates a couple of outdated hrefs in HACKING.md.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -55,7 +55,7 @@ pytest
 The framework has some tests that interact with a real/live pebble server.  To
 run these tests, you must have (pebble)[https://github.com/canonical/pebble]
 installed and available in your path.  If you have the Go toolchain installed,
-you can run `go install github.com/canonical/pebble@latest`.  This will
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Thanks!

The `go install` path was definitely broken (probably my mistake). I'm not sure quite why the `publish.yml` path change is necessary (the current one seems to work) but a relative path definitely doesn't hurt either.

benhoyt:
Thanks for the contribution, @tinvaan! Trivial doc fixes so merging without further review.

tinvaan:
> Thanks!
> 
> The `go install` path was definitely broken (probably my mistake). I'm not sure quite why the `publish.yml` path change is necessary (the current one seems to work) but a relative path definitely doesn't hurt either.

Yeah it's very trivial,  but the relative path is useful when you're viewing a markdown file locally - VSCode supports markdown file previews for eg - and a relative path href ensures that the linked document is opened in the current application itself as opposed to opening a new browser tab or window.

Also makes it easier to preview and edit the opened document -- again trivial but useful.

benhoyt:
Good points, thanks!

PR #894, opened 2023-01-18
Define all dependences in requirements files

tinvaan:
Defines all project dependencies in one place (`requirements.txt` and `requirements-dev.txt`) and pins them to a version.

Initial review: https://github.com/tinvaan/operator/pull/1

## QA steps

Run the `tox` script and real pebble tests to verify sanity.

```sh
$ tox -r
$ tox -re static
$ tox -re pebble
```


README.md
```diff
@@ -88,7 +88,7 @@ is to feel natural for somebody used to coding in Python, and reasonably easy to
 who is not a pythonista.
 
 The dependencies of the operator framework are kept as minimal as possible; currently that's Python
-3.8 or greater, and `PyYAML` (both are included by default in Ubuntu's cloud images from 20.04 on).
+3.8 or greater, and a couple of python libraries outlined in [requirements.txt](requirements.txt).
```
benhoyt, line 91: Nit: python -> Python. Maybe "and the small number of Python libraries referenced in requirements.txt"

requirements.txt
```diff
@@ -1,2 +1,2 @@
-PyYAML
-websocket-client
\ No newline at end of file
+PyYAML==6.0
+websocket-client==1.4.2
```
benhoyt, line 2: I've thought about this further, and I think that given that `ops` is a library and not an applicable, we shouldn't pin exact dependencies. For rationale, see [this SO answer](https://stackoverflow.com/a/44938662/68707). I think we should only pin to the major (compatible) version. So:

```
PyYAML==6.*
websocket-client==1.*
```

We can reevaluate later if needed, but I think that's safe, and an improvement on what we have (no pinning).

For `requirements-dev.txt` we should still pin exact versions, as those are only used locally.

setup.py
```diff
@@ -26,6 +28,12 @@ def _read_me() -> str:
     return readme
 
 
+def _requirements() -> List[str]:
+    file = Path((__file__)).parent.joinpath('requirements.txt')
+    with open(file, 'r') as fh:
+        return fh.readlines()
```
benhoyt, line 34: I think we should make a couple of tweaks here:

* Let's add a `"""` docstring, kind of like `_read_me()` has.
* Drop the extra parens in `Path((...))`.
* Let's use `Path(__file__).parent / 'requirements.txt'`, as I think that's more idiomatic [with pathlib](https://docs.python.org/3/library/pathlib.html#operators).
* IMO we should drop the `r` on `open()`, but explicitly specify `encoding='utf-8'`.
* Tweak the code to make it a bit more robust with blank and comment lines, per below:

```python
with open(Path(__file__).parent / 'requirements.txt') as fh:
	reqs = []
	for line in fh:
		line = line.strip()
		if not line or line.startswith('#'):
			continue
		reqs.append(line)
        return reqs
```

requirements-dev.txt
```diff
@@ -1,3 +1,8 @@
-PyYAML
-pyright==1.1.257
-typing_extensions==4.2.0
\ No newline at end of file
+ipdb==0.13.11
```
benhoyt, line 1: IMO we should get rid of this ipdb dependency (I've never used it, devs can set up IPython / ipdb themselves if they want it). Feel free to leave that for me in a follow-up PR though.

tox.ini
```diff
@@ -57,8 +57,7 @@ commands =
 [testenv:static]
 description = Run static type checker
 deps =
-    pyright==1.1.264
-    -r{toxinidir}/requirements.txt
+    -r{toxinidir}/requirements-dev.txt
```
benhoyt, line 60: Can we also do this for the `[testenv:fmt]` and `[testenv:lint]` sections above, and their dependencies like `isort` and `autopep8`? Would be good if all dev deps were in one place.

tox.ini
```diff
@@ -68,12 +67,9 @@ passenv =
     RUN_REAL_PEBBLE_TESTS
     PEBBLE
 deps =
-    pytest
-    ipdb
     ipython!=8.1.0 # this version is broken and was causing failures
-    logassert
     coverage[toml]
-    -r{toxinidir}/requirements.txt
+    -r{toxinidir}/requirements-dev.txt
```
benhoyt, line 72: Any reason why you've moved most of these but not ipython and coverage? IMO we get rid of ipython in any case. I can't understand why ipython is needed for running these tests.

benhoyt (commented):
Thanks again for this! You're making me feel bad with all these contributions. :-) I've added a few comments of minor things I think we need to address.

requirements-dev.txt
tinvaan: I use it all the time, but I agree that it needn't be enforced by a project. Easy enough to install and use it myself.

tox.ini
tinvaan: So, I had initially left these deps as it is because I wanted to keep the dev requirements list minimal. However, now when I tried combining these deps under one `requirements-dev.txt`, I notice that `autopep8` & `flake-*` dep versions are incompatible.

Let me check if I can arrive at a version that can works across all these deps.

tox.ini
tinvaan: yeah, I really wasn't sure about `ipython` and where it's being used. For coverage, I viewed it similar to `ipdb` in that it's one of the many debugging tools that a user(a dev) can choose from. 

So within the dev environment a user may choose to use a different coverage plugin, different from that contained in the `tox` script.

However, I do also believe this is a rather fringe case and this isn't a strongly held opinion of mine. So I'll move it all into `requirements-dev.txt` for the sake of uniformity/consistency.

tox.ini
```diff
@@ -32,22 +32,15 @@ commands =
 [testenv:fmt]
 description = Apply coding style standards to code
 deps =
-    isort
-    autopep8
+    -r{toxinidir}/requirements-dev.txt
```
benhoyt, line 35: I can't comment on the correct line, but let's remove the `PYTHONBREAKPOINT=ipdb.set_trace` line above, now that ipdb is gone.

benhoyt (changes requested):
Thanks again! Just a couple of very minor changes left -- comments added.

setup.py
tinvaan: yeah not sure what happened here, perhaps I forgot to push the commit. 

I also came across https://setuptools.pypa.io/en/latest/pkg_resources.html#requirements-parsing but I've kept it for another PR with a `TODO` note for now.

benhoyt (approved):
LGTM, thanks!

rbarry82:
This looks fine, but I'm curious as to the motivation. I'm not against pinning versions, but it's uncommon practice in Python, even if we've all probably been bitten many times by "whoops, major version change happened overnight and now my CI explodes".

Was this a request, due to a specific issue, or just a change in practice?

benhoyt:
Yeah, it's to tighten things up a bit so that CI (or our users' CI) doesn't break so easily. :-) I had run into this once or twice with ops dev dependencies already.

Is it uncommon practice in Python? I've seen it a ton. Obviously more in applications because there you want to deploy exact versions, but in libraries I've definitely seen it done to pin major or minor (note that we're only pinning major versions here, except for our dev dependencies).

tinvaan:
Yeah, keeping requirements pinned to specific versions has been my experience across the projects I've been involved in as well. Obviously helps improve the stability in a CI or containerised environment.

rbarry82:
It's not a bad idea, but it's far from common practice. 

The community is pretty split on it, and the [phenomenal amount of breakage in very well backed projects](https://github.com/aws/aws-sam-cli/issues/3661) around a Jinja/MarkupSafe last year is evidence both that pinning is a good idea sometimes, but also that the Python community is still so torn on it that ~100 projects you've probably heard of had fiery CI implosions when it happened.

Could we? Sure. It is easier? Sure, most of the time. But there's a distinct tradeoff when it comes to automated CVE scanners. Most of which, admittedly, end up flagging stuff like "if a user submits a request to a REST endpoint in some part of the library you probably aren't using anyway with incredibly carefully crafted headers, they might be able to peer at something trivial: CVSS 9.9/10", but that's still a ding on the project for customers running naive scans

tinvaan:
I'm not very familiar with automated CVE scanning tools, but as for any security concerns that may arise due to pinning dependencies to versions would it make sense to include a dependabot workflow in the project?

https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot

rbarry82:
The honest answer is "probably not". Automated CVE scanners aren't worth the time more often than not, especially compared to the hassle introduced by dependabot.

It's particularly obvious in any container built from, say, Ubuntu running as a non-privileged user with a Golang binary as the entrypoint (rather than using `ko` or building from `scratch`). Scanners flag "there's a vulnerability in `expat2`!"

But it's Go. It's probably not parsing XML anyway, and if it is, there's almost no chance it's CGO and using expat at all, so the "vulnerability" is completely meaningless. Or you pulled in some Prometheus module without quite enough granularity, and _its_ deptree yanked API integrations for every endpoint known to man (even if the code paths aren't reachable from your code), and...

But the scanner flags it anyway. And eventually, you get a ticket to go update the image, with no measurable positive outcome and no assurances that it won't just happen again tomorrow. Partly this is CVSS being a terrible system. [This](https://ubuntu.com/security/CVE-2022-28321), for example, is "Negligible" from our rating, isn't meaningfully exploitable on containers anyway, possibly never happened in Ubuntu at all (the report is from SuSE Tumbleweed years ago). It's still rated as a 9.8, and filed last year.

If that's a 9.8/10, what's Shellshock? 100/10? It's a bad system, but CVSS is the "default" scoring, and _not_ pinning deps is a tradeoff of "CI may break and I'll have to investigate it" versus "constant CVE reports will be filed and burn development time for nothing". It's a matter of picking your poison, and there's not a right/good way, but it's a large part of the reason that projects don't pin.

jnsgruk (approved):
Overall I'm +1 on this.

To a limited extent, I do agree with @rbarry82 here, and perhaps pinning the dev dependencies is less of a priority, but pinning the runtime deps is a must in my opinion, given that this is a core library for many projects now.

I would personally square this off with a a nicely tuned bot (renovate, or similar), but we could write a very simple automation that just checks the pinned versions every week, or even every month, and opens an issue or PR to make sure they're kept fresh.

jameinel (approved):
I'm broadly in favor of trying to do the major version pinning, and see if this gives an overall better experience than having PyYAML release a major version 7, and charms that haven't been doing pinning just falling over.

ops/pebble.py
```diff
@@ -2483,8 +2483,8 @@ def __init__(
         self._max_lookahead = max_lookahead
         self._max_boundary_length = max_boundary_length
 
-        self._buf = bytearray()
-        self._pos = 0  # current position in buf
+        self._buf: bytearray = bytearray()
+        self._pos: int = 0  # current position in buf
```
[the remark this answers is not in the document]
benhoyt: Yeah, I couldn't figure this out either. I just spent a bunch of time on it creating a cut-down reproducer ... and it turns out it's a bug in Pyright being confused about Unknown types in some cases. It was a weird one to reproduce because I'd take out a code block or if statement that was seemingly unrelated, and Pyright would suddenly be OK. But it was fixed with Pyright [1.1.290](https://github.com/microsoft/pyright/releases/tag/1.1.290), so I upgraded Pyright to the latest (1.1.291) and removed those type annotations now.

It was likely [this Pyright fix](https://github.com/microsoft/pyright/commit/bd010d11926856025835c3bf75d0351094417e9f) which has to do with caching of Unknown types or something.

Anyway, thanks for nerd-sniping me. ;-)

PR #895, opened 2023-01-20
Reenable alertmanager-operator tests now that charm is fixed on ops 2.0

benhoyt:
Also update these repos to their new (redirected) URLs.

Reverts when this was commented out in https://github.com/canonical/operator/pull/885


benhoyt:
Merged without further review as this is just reinstating tests we had commented out (which pass now).

PR #897, opened 2023-01-23
Make testing pull() raise pebble.PathError and improve docs

### LINKED ISSUE #896: Harness raises FileNotFoundError in `container.pull`
I have a unit test in which I see a FileNotFoundError on pull. Not sure if it's intended or if instead it should be PathError.

```
  File "/home/him/code/prometheus-k8s-operator/src/charm.py", line 743, in _generate_prometheus_config
    if config_hash == pull(container, CONFIG_HASH_PATH):
  File "/home/him/code/prometheus-k8s-operator/src/charm.py", line 86, in pull
    return container.pull(path, encoding="utf-8").read()
  File "/home/him/code/prometheus-k8s-operator/.tox/unit/lib/python3.10/site-packages/ops/model.py", line 1557, in pull
    return self._pebble.pull(str(path), encoding=encoding)
  File "/home/him/code/prometheus-k8s-operator/.tox/unit/lib/python3.10/site-packages/ops/testing.py", line 1930, in pull
    return self._fs.open(path, encoding=encoding)
  File "/home/him/code/prometheus-k8s-operator/.tox/unit/lib/python3.10/site-packages/ops/testing.py", line 2362, in open
    file = self.get_path(path)  # warning: no check re: directories
  File "/home/him/code/prometheus-k8s-operator/.tox/unit/lib/python3.10/site-packages/ops/testing.py", line 2382, in get_path
    raise FileNotFoundError(str(current_object.path / token))
FileNotFoundError: /etc
```

In itests, a pebble error is raised (a FileNotFoundError does not catch those).

benhoyt:
Currently the testing harness's version of pull() raises FileNotFoundError and IsADirectoryError instead of pebble.PathError, which is what the real pull() raises in these situations. Catch those and raise the same error as in real life.

Also document this better in the relevant docstrings.

There are probably other places the testing file system's errors leak out, but keep this PR limited in scope to just pull().

Also remove an unnecessary Pyright FIXME -- this was addressed when the message() method was removed recently.

Fixes #896

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1936,6 +1936,10 @@ class Container:
             A readable file-like object, whose read() method will return str
             objects decoded according to the specified encoding, or bytes if
             encoding is None.
+
+        Raises:
+            pebble.PathError: If there was an error reading the file at path,
+                for example, if the file doesn't exist or is a directory.
         """
         return self._pebble.pull(str(path), encoding=encoding)
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -307,7 +307,7 @@ class PathError(Error):
 
     Attributes:
         kind: A short string representing the kind of error. Possible values
-            are 'not-found', 'permission-denied', and 'generic-file-error'.
+            are "not-found", "permission-denied", and "generic-file-error".
         message: A human-readable error message.
     """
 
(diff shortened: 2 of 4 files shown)


sed-i (approved):
Thanks!

rbarry82 (approved):

rbarry82:
I mean, my approval doesn't matter anymore, but thanks for the quick turnaround :D

jnsgruk (approved):
Nice catch, thanks :)

rbarry82:
@sed-i can you test this please? I never saw it in testing, so it was kind of just reading `ops.pebble` blind.

The mention of it happening in integration tests would lead me to believe that there was a difference in the parsed response from Pebble and what `ops.pebble` was expecting (not sure if you're still pinning to Juju 2.9.xx), and this may happen at runtime/is not able to be wholly fixed with the patch here, but confirmation either way would be great before it's marked fixed and re-opened

benhoyt:
Per discussion with @jameinel this morning, I decided to cut back this change to just the harness fix and the doc improvements, rather than adding `pebble.NotFoundError` and (especially) making that inherit from both `pebble.PathError` and the builtin `FileNotFoundError`. The builtin error has a bunch of attributes that won't be set or work properly, (eg: `errno`, `strerror`, and so on). So making it a "kind of a FileNotFoundError but not really" seemed like a bad idea, or at least one that requires further thought.

PR #900, opened 2023-02-03
Don't call secret-info-get with both ID and label (it's a Juju error)

### LINKED ISSUE #899: uncaught error in `Secret.get_info()`
the following error appears while running the `owner` charm in [this demo bundle](https://github.com/PietroPasotti/secrets-demo-charms/tree/main)

image

benhoyt:
Fix itself is a two-character change: "if" -> "elif". :-)

Also a couple of drive-by fixes for docstrings.

Fixes #899. However, due to https://bugs.launchpad.net/juju/+bug/2004672, when this is fixed you get a "secret not found" error when `secret-info-get` is used in the same hook as the `secret-add`, because `secret-info-get` is not looking at pending creates.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2926,7 +2926,7 @@ class _ModelBackend:
         args = []  # type: List[str]
         if id is not None:
             args.append(id)
-        if label is not None:
+        elif label is not None:  # elif because Juju secret-info-get doesn't allow id and label
             args.extend(['--label', label])
         result = self._run_for_secret('secret-info-get', *args, return_output=True, use_json=True)
         info_dicts = typing.cast(Dict[str, 'JsonObject'], result)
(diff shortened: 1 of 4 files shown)


wallyworld (approved):

ops/model.py
```diff
@@ -2926,7 +2926,7 @@ def secret_info_get(self, *,
         args = []  # type: List[str]
         if id is not None:
             args.append(id)
-        if label is not None:
+        elif label is not None:  # elif because Juju secret-info-get doesn't allow id and label
```
PietroPasotti, line 2929: For the sake of not letting the edge case 'no id and no label passed' go through, why not refactor this as

```
if id:
 args.append(id)
elif not label:
 raise ValueError('pass ID or label')
else:
 args.append(...)
```

PietroPasotti (approved):
Looks good! one minor suggestion to make the modelbackend surface errors a bit earlier on

ops/model.py
benhoyt: Going to leave this as is for now -- that check is already done by the callers (`Model.get_secret` and `Secret.__init__`), and that's the pattern for most of these: the hook tool methods like `secret_info_get` don't generally do additional argumment checking.

PR #903, opened 2023-02-08
Make secret-get consistently use "secret:<id>" fmt to avoid confusion

### LINKED ISSUE #902: secret id is not always canonicalized before calling secret_get
Secret canonicalizes the ID it receives upon initialization.
However, before the Secret object is initialized, the ID is used to call secret_get. This results in the ID being sometimes canonicalized, sometimes not, depending on the caller.

For example:
```
self.model.get_secret(id="foo").get_content()
# resulting call: secret-get --id foo

self.model.get_secret(id="foo").peek_content()
# resulting call: secret-get --id secret:foo --peek
```

This is not a problem per se, but it might mean headache in the future.

benhoyt:
This won't actually fix any `ops` issues in the field, but it makes the backend calls more consistent and less confusing.

Also add test for label-only Model.get_secret() variant.

Fixes #902

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -276,6 +276,9 @@ class Model:
         """
         if not (id or label):
             raise TypeError('Must provide an id or label, or both')
+        if id is not None:
+            # Canonicalize to "secret:<id>" form for consistency in backend calls.
+            id = Secret._canonicalize_id(id)
         try:
             content = self._backend.secret_get(id=id, label=label)
             return Secret(self._backend, id=id, label=label, content=content)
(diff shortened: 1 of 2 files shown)


wallyworld (approved):

ops/model.py
```diff
@@ -276,6 +276,9 @@ def get_secret(self, *, id: Optional[str] = None, label: Optional[str] = None) -
         """
         if not (id or label):
             raise TypeError('Must provide an id or label, or both')
+        if id is not None:
+            # Canonicalize to "secret:<id>" form for consistency in backend calls.
+            id = Secret._canonicalize_id(id)
```
PietroPasotti, line 281: Do we want to keep the canonicalize_id call inside Secret.__init__ now that it's here? And replace it with a validation step? Or do you think a little bit of redundancy won't hurt?

PietroPasotti (approved):
Just a minor comment, looks good!

PR #904, opened 2023-02-09
Replace "# type: T" comments with proper variable type annotations

benhoyt:
Now that we're using Python 3.8+ the more modern syntax is available to us, so use it. Also re-enable Pyright's reportTypeCommentUsage setting.

The second commit in this PR also removes a bunch of unneeded "noqa" and "pyright" comments.

I used a little script to find these and do a regex replace, then fixed up the few remaining ones (that were split on multiple lines) manually. Then I ran the following script to dump the AST (removing "annotated assignments") and did a diff against the original AST to ensure I hadn't messed up anything else.

```python
import ast
import os

class Transformer(ast.NodeTransformer):
    def visit_AnnAssign(self, node):
        if node.value is None:
            return None
        return ast.Assign([node.target], node.value)

PATHS = """
ops/charm.py
ops/framework.py
ops/main.py
ops/model.py
ops/pebble.py
ops/storage.py
ops/testing.py
""".split()

for path in PATHS:
    with open(path) as f:
        source = f.read()
    filename = os.path.basename(path)
    tree = ast.parse(source, filename)

    transformed = Transformer().visit(tree)

    root, ext = os.path.splitext(filename)
    outname = f'ops/{root}.ast'
    print(outname)
    with open(outname, 'w') as f:
        f.write(ast.dump(tree, indent=4))
```

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -450,10 +450,10 @@ class RelationEvent(HookEvent):
 
         Not meant to be called by charm code.
         """
-        snapshot = {
+        snapshot: 'RelationEvent._RelationEventSnapshot' = {
             'relation_name': self.relation.name,
             'relation_id': self.relation.id,
-        }  # type: 'RelationEvent._RelationEventSnapshot'
+        }
         if self.app:
             snapshot['app_name'] = self.app.name
         if self.unit:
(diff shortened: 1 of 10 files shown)


PietroPasotti (approved):
Nice! At long last :)

jnsgruk (approved):

Issue #179, opened 2020-03-11
Expose open-port and close-port hook tools


stub42:
The `open-port` and `close-port` Juju hook environment tools are not exposed by the operator framework.

It would be good if the framework handled closing the old port when a new one is opened. eg.

```
self.open_port('grpc', 1234)  # Remembers 'grpc' port
self.open_port('grpc', 4242)  # Automatically close-port 1234 before open-port 4242
```

jameinel:
That assumes that any given charm never wants to open more than one port.
Definitely we want to support ranges, not just singular ports (I believe
'open-port' already handles this.
Also, there has been a design item on the Juju side for a while to link
what binding (endpoint?) the port is for. So that someone can expose an
attribute of an application without exposing everything it does. (There is
further work to allow expose to a list of CIDR, but that doesn't impact how
the charm needs to interact with Juju.)
There is a desire as well to have charms declare 'internal' ports.
Currently Juju models end up with a rule that machines in the same model
can talk to each other on any port, which many find distasteful.

It feels like it would then end up on the Binding object. So something like:

```
binding = self.model.get_binding('db')
binding.ports = [PortRange(123)]
```

I'm not 100% sure how to express multiple ranges cleanly.
Having it as an attribute makes it clear that you are replacing the
existing ports, and can allow stuff like
```
binding.ports.append(PortRange(123, 456))
```
We could certainly allow for the trivial integer case
binding.ports = [80]

facundobatista:
@stub42 the suggestion of including a "name" in the call, is to just assign it to the open port, so it can later be closed automatically? `.close_port` could use a name or the port number, too.

But is that name for any other use?

stub42:
On Wed, 18 Mar 2020 at 07:00, Facundo Batista <notifications@github.com>
wrote:

> @stub42 <https://github.com/stub42> the suggestion of including a "name"
> in the call, is to just assign it to the open port, so it can later be
> closed automatically? .close_port could use a name or the port number,
> too.
>
> But is that name for any other use?
>
I only included the key to allow the port to be automatically closed. I
don't see any other use.

facundobatista:
Thanks @stub42. From your POV, changing ports for one service/purpose is so common that is a good thing for the framework to track them? I'm trying to understand if adding the complexity of a "name" to the method's signature is worthwhile (we need to explain why a name there, how to choose it, etc.).

@jameinel I don't understand why you say from @stub42 proposal that charms don't want to open new ports, a charm could totally do...

[...]

PR #905, opened 2023-02-16
Support for open-port, close-port, and opened-ports

benhoyt:
Adds support for open-port, close-port and opened-ports as per the [OP029 spec](https://docs.google.com/document/d/1xHLzBgcXtLsBpMsjuqUW2pI1tP4D8lgSUD_sg3HClng/edit#).

Also adds testing backend support for the same for use in charm tests.

Fixes https://github.com/canonical/operator/issues/179

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -634,7 +634,7 @@ class Unit:
         return self._backend.opened_ports()
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class OpenedPort:
     """Represents a port opened by :meth:`Unit.open_port`."""
 
@@ -3068,18 +3068,13 @@ class _ModelBackend:
             return OpenedPort('icmp', None)
         port_range, slash, protocol = port_str.partition('/')
         if not slash or protocol not in ['tcp', 'udp']:
-            logger.warning('Invalid line from opened-ports: %s', port_str)
+            logger.warning('Unexpected opened-ports protocol: %s', port_str)
             return None
-        start, hyphen, _ = port_range.partition('-')
+        port, hyphen, _ = port_range.partition('-')
         if hyphen:
-            logger.warning('Unexpected port range from opened-ports: %s', port_str)
-        try:
-            port = int(start)
-        except ValueError:
-            logger.warning('Invalid line from opened-ports: %s', port_str)
-            return None
+            logger.warning('Ignoring opened-ports port range: %s', port_str)
         protocol_lit = typing.cast(typing.Literal['tcp', 'udp', 'icmp'], protocol)
-        return OpenedPort(protocol_lit, port)
+        return OpenedPort(protocol_lit, int(port))
 
 
 class _ModelBackendValidator:
--- a/ops/model.py
+++ b/ops/model.py
@@ -626,7 +626,20 @@ class Unit:
 
     def close_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
                    port: Optional[int] = None):
-        """Close a port with the given protocol for this unit."""
+        """Close a port with the given protocol for this unit.
+
+        On Kubernetes sidecar charms, Juju will only close the port once the
+        last unit that opened that port has closed it. However, this is
+        usually not an issue; normally charms should make the same
+        close_port() call from every unit.
+
+        Args:
+            protocol: String representing the protocol; must be one of
+                'tcp', 'udp', or 'icmp' (lowercase is recommended, but
+                uppercase is also supported).
+            port: The port to open. Required for TCP and UDP; not allowed
+                for ICMP.
+        """
         self._backend.close_port(protocol.lower(), port)
 
     def opened_ports(self) -> Set['OpenedPort']:
--- a/ops/model.py
+++ b/ops/model.py
@@ -3051,6 +3051,10 @@ class _ModelBackend:
         self._run('close-port', arg)
 
     def opened_ports(self) -> Set[OpenedPort]:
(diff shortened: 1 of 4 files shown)


ops/model.py
```diff
@@ -600,6 +602,48 @@ def add_secret(self, content: Dict[str, str], *,
...
+        However, normally charms should make the same open_port() call from
+        every unit.
+
+        Args:
+            protocol: String representing the protocol; must be one of
+                'tcp', 'udp', or 'icmp' (lowercase is recommended, but
+                uppercase is also supported).
+            port: The port to open. Required for TCP and UDP; not allowed
+                for ICMP.
+        """
```
sed-i, line 624: Do we document in the docstring here what the backend may raise?

ops/model.py
```diff
@@ -2998,6 +3042,40 @@ def secret_remove(self, id: str, *, revision: Optional[int] = None):
...
 
+    def open_port(self, protocol: str, port: Optional[int] = None):
+        arg = f'{port}/{protocol}' if port is not None else protocol
+        self._run('open-port', arg)
+
+    def close_port(self, protocol: str, port: Optional[int] = None):
+        arg = f'{port}/{protocol}' if port is not None else protocol
+        self._run('close-port', arg)
+
+    def opened_ports(self) -> List[OpenedPort]:
```
sed-i, line 3053: Seems like this would be frequently used to compare between desired and actual. Should the return type be `Set[OpenedPort]` for easier comparison?
If OpenedPort is not hashable then should it be a namedtuple instead?

ops/model.py
benhoyt: We don't usually for all of these when they just raise `ModelError` (what all hook tool errors end up as). If they raise more specific exceptions we document that.

I'd be open to reconsider that and also document the `Raises: ModelError ...`, but it would be on pretty much every function.

ops/model.py
benhoyt: Not a bad idea, as order isn't important (or guaranteed) here. I'll make that change.

`OpenedPort` is a dataclass with `frozen=True`, so it's hashable.

jnsgruk:
Looking tidy so far! 😎

ops/model.py
sed-i: Nice, I didn't realize all hook tools only raise ModelError.

ops/model.py
```diff
@@ -2998,6 +3042,44 @@ def secret_remove(self, id: str, *, revision: Optional[int] = None):
             args.extend(['--revision', str(revision)])
         self._run_for_secret('secret-remove', *args)
 
+    def open_port(self, protocol: str, port: Optional[int] = None):
+        arg = f'{port}/{protocol}' if port is not None else protocol
```
jameinel, line 3046: I'm pretty sure port is only optional if protocol is 'icmp' in which case it can't have a port.
I don't know if there is a clearer way to define this.

This is validated by your logic in "_parse_opened_port" that does have 'icmp' and others hard coded for that behavior (must have a port if tcp or udp, must not have a port for icmp)

ops/model.py
```diff
@@ -600,6 +602,48 @@ def add_secret(self, content: Dict[str, str], *,
...
+                'tcp', 'udp', or 'icmp' (lowercase is recommended, but
+                uppercase is also supported).
+            port: The port to open. Required for TCP and UDP; not allowed
+                for ICMP.
+        """
+        self._backend.open_port(protocol.lower(), port)
+
+    def close_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
+                   port: Optional[int] = None):
+        """Close a port with the given protocol for this unit."""
```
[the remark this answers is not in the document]
benhoyt: Ah whoops, I'd intended to go back to that comment and flesh it out. Fixed now.

ops/model.py
benhoyt: This is noted in the docstring of open-port and (updated docstring for) close-port. It is checked by Juju's open-port and close-port hooks tools and will return an error there. In the Harness I test it up-front, so if folks get this wrong it'll be raised as an error immediately when the unit tests are run.

ops/model.py
```diff
@@ -2998,6 +3055,44 @@ def secret_remove(self, id: str, *, revision: Optional[int] = None):
...
+            return OpenedPort('icmp', None)
+        port_range, slash, protocol = port_str.partition('/')
+        if not slash or protocol not in ['tcp', 'udp']:
+            logger.warning('Unexpected opened-ports protocol: %s', port_str)
+            return None
+        port, hyphen, _ = port_range.partition('-')
+        if hyphen:
+            logger.warning('Ignoring opened-ports port range: %s', port_str)
+        protocol_lit = typing.cast(typing.Literal['tcp', 'udp', 'icmp'], protocol)
+        return OpenedPort(protocol_lit, int(port))
```
jameinel, line 3094: So this does clearly allow "icmp/10" but it does seem like we just filtered that out on line 3087.
Though I guess this is a type cast not a literal check?

jameinel (approved):
I think this gives a nice documented way to interact, and conforms to the agreed protocol.

ops/model.py
benhoyt: Fair point - fixed. (But yes, this is a type cast not a runtime check.)

benhoyt:
Going to merge without a second approval: spec's been approved by Gustavo, John's taken a close look at the PR and approved, and others have cast their eyes over it too -- I think we're good.

jnsgruk:
Happy to see this land, thanks Ben!

PR #910, opened 2023-02-24
Import (almost) all public names in ops __init__.py

### LINKED ISSUE #731: Annoyance: Figuring out which module has my object
When writing charms, I am often able to remember the name of object that I want (`EventBase`, `StoredState`, etc.), but often forget which module it comes from. There's logic to the modules, of course -- `CharmBase` is in `charm`, the status objects live in `model`, and the things that I want when I'm maybe being a little bit hacky live in `framework`. I can make guesses, and those guesses are easy to check with the introspection tools in an IDE or a good old fashioned `grep` on the command line.

It's a little mental speed bump, however, of the sort that I think we want to eliminate.

Note: there's a warning about circular imports in `ops/__init__.py`, which means that we'd need to be thoughtful if we imported more objects into the top level module. It might be worth a spike, however. Lines like `self.model.unit.status = MaintenanceStatus("Frobbling the bitzors")` would be easier to write if I didn't first have to remember that statuses live in `ops.model` rather than `ops.charm` or `ops.framework`.

benhoyt:
This is to address #731: it's hard to remember or discover which `ops` sub-module (like `ops.charm` or `ops.framework`) a name comes from. That, and imports become a bit of a mess (`from ops.charm import X; from ops.framework import Y; from ops.main import main`).

In my view we had previously split them up somewhat arbitrarily (eg: `FooEvent` types are in `ops.charm`, but they could just as well be in `ops.framework`). I considered a different, more logical sub-module arrangement, but I think it'd still be hard to remember and know where things were.

This solution is simple: import all the public names into the top-level `ops` module in the `__init__.py` file. They're quite unique, so there's no conflicts. The only exception is `ops.pebble`, whose names aren't imported there (most charms don't use `ops.pebble` directly anyway, they use it via `model.Container`, but if you need it its still readily importable).

This makes all the names easily discoverable, either by looking in `ops/__init__.py` manually, or via your editor's code completion when you've done `import ops`. I expect that many charms will change to just `import ops` and access names like `ops.CharmBase` (`ops` is a nice short prefix) ... see [this PR on my database test charm](https://github.com/benhoyt/test-charms/pull/1/files) for an example of what that would look like.

Folks can also use explicit imports like `from ops import CharmBase` if they prefer. And of course the existing usage `from ops.charm import CharmBase` still works fine too.

The one quirky thing is that `ops.main` is currently a module, but we want to allow `import ops ... ops.main()` or `from ops import main ... main()`. To solve this we've main the main module callable (and it just calls the `main` function). This allows the new usage, but also allows the existing usages `from ops.main import main ... main()` and  `import ops.main ... main.main()` to continue working.

### Testing note

The user-facing changes here are all done in `ops/__init__.py` and `ops/main.py`, but I've also updated the `test/test_*.py` files to show using the single `import ops` approach (partly to set that as a precedent).

However, I ran all the previous, unmodified tests against the new `ops/__init__.py`, and they all pass. This ensures that I haven't messed anything up, including the `main` shenanigans mentioned above.

Fixes #731.

### DIFF
--- a/ops/__init__.py
+++ b/ops/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


jnsgruk (approved):
Nice, I like this change a lot. Might be worth considering updating the [templates](https://github.com/canonical/charmcraft/tree/main/charmcraft/templates) in charmcraft once this lands.

benhoyt:
@jnsgruk Yes, very good call to update the templates. And a bunch of docs.

benhoyt:
Opened https://github.com/canonical/operator/issues/921 to track updating docs, templates, and (perhaps) some charms.

ops/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
jameinel, line 1: generally we wouldn't update the copyright of a file, if you do, then it should be a range, but otherwise the position is the "oldest time when the file was created"

ops/__init__.py
```diff
@@ -39,6 +39,141 @@
...
+    StatusBase,
+    Storage,
+    StorageMapping,
+    TooManyRelatedAppsError,
+    Unit,
+    UnknownStatus,
+    WaitingStatus,
+)
+
+from .testing import Harness  # noqa: F401 E402
```
jameinel, line 177: I wouldn't import Harness, because I actually only want to see 'ops.testing' get imported in test suites, not in runtime code

jameinel (commented):
A couple of tweaks, we should leave testing only for test cases, and not the general ops module.

I'm ok with the ops.main() vs ops.main.main() it isn't the code that we want to end up with, but we can change that in an ops 3.0

ops/__init__.py
benhoyt: Done.

ops/__init__.py
benhoyt: Fair call. I was on the fence about this, but I think you're right.

jameinel (approved):
Do we want a comment, either in ops/__init__.py or ops/testing.py as to *why* we aren't including Harness at the top level?

Issue #913, opened 2023-03-01
FileNotFoundError: '.../requirements.txt' during `charmcraft pack` with 2.1.0


Perfect5th:
Since ops 2.1.0 released, I started getting this error when I `charmcraft pack`:
```
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 :: Running external command ['/root/parts/charm/build/staging-venv/bin/pip3', 'install', '--upgrade', '--no-binary', ':all:', '--requirement=requir
ements.txt']
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    :: Collecting ops>=1.4.0
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   Using cached ops-2.1.0.tar.gz (228 kB)
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   Installing build dependencies: started
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   Installing build dependencies: finished with status 'done'
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   Getting requirements to build wheel: started
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   Getting requirements to build wheel: finished with status 'error'
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   error: subprocess-exited-with-error
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   × Getting requirements to build wheel did not run successfully.
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   │ exit code: 1
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::   ╰─> [18 lines of output]
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::       Traceback (most recent call last):
2023-02-28 15:57:44.361 :: 2023-02-28 23:57:44.064 ::    ::         File "/root/parts/charm/build/staging-venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, i
n <module>
[...]

PR #915, opened 2023-03-01
Add CI tests for 'pip install' and 'charmcraft pack'

benhoyt:
This is to find any issues early and avoid breakage like that caused by 2.1.0 missing the requirements.txt file: https://github.com/canonical/operator/issues/913 (the specific issue was fixed in #914).

### DIFF
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -23,7 +23,7 @@ jobs:
         # temporary merge commit), because charmcraft pack does a git checkout which
         # can't see the temporary merge commit.
         sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
-        echo -e "\ngit+${{ github.event.pull_request.repo.clone_url }}@${{ github.event.pull_request.head.sha }}#egg=ops" >> requirements.txt
+        echo -e "\ngit+${{ github.event.pull_request.head.repo.clone_url }}@${{ github.event.pull_request.head.sha }}#egg=ops" >> requirements.txt
         cat requirements.txt
 
     - name: Install charmcraft
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -22,7 +22,7 @@ jobs:
         cat requirements.txt
 
     - name: Set up LXD
-      uses: canonical/setup-lxd
+      uses: canonical/setup-lxd@ea57509
       with:
         channel: 5.0/stable
 
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -30,4 +30,4 @@ jobs:
       run: sudo snap install charmcraft --classic
 
     - name: Pack the charm
-      run: sudo -g lxd charmcraft pack --verbose
+      run: sudo charmcraft pack --verbose
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -4,16 +4,11 @@ on: [push, pull_request, workflow_call]
 
 jobs:
   charmcraft-pack:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
-      with:
-        python-version: "3.8"
-
     - name: Checkout test charm repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         repository: jnsgruk/hello-kubecon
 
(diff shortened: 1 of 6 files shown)


.github/workflows/charmcraft-pack.yaml
```diff
@@ -0,0 +1,38 @@
+name: Charmcraft Pack Test
+
+on: [push, pull_request, workflow_call]
+
+jobs:
+  charmcraft-pack:
+    runs-on: ubuntu-20.04
```
jnsgruk, line 7: Any reason not to use jammy here?

.github/workflows/charmcraft-pack.yaml
```diff
@@ -0,0 +1,38 @@
...
+        # We need to reference the PR branch's repo and commit (not the GITHUB_SHA
+        # temporary merge commit), because charmcraft pack does a git checkout which
+        # can't see the temporary merge commit.
+        sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
+        echo -e "\ngit+${{ github.event.pull_request.head.repo.clone_url }}@${{ github.event.pull_request.head.sha }}#egg=ops" >> requirements.txt
+        cat requirements.txt
+
+    - name: Install charmcraft
+      run: |
+          sudo snap refresh lxd --channel=latest
```
jnsgruk, line 31: If we do move to jammy, might be best to use the canonical/setup-lxd action to avoid firewall issues

.github/workflows/framework-tests.yaml
```diff
@@ -89,4 +89,20 @@ jobs:
...
         env:
           RUN_REAL_PEBBLE_TESTS: 1 
-          PEBBLE: /tmp/pebble 
+          PEBBLE: /tmp/pebble
+
+  pip-install:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
jnsgruk, line 98: Boring nit: checkout@v3 is the latest and won't throw a deprecation warning

jnsgruk (approved):
Non-blocking bits. Good to see this being tested. Thanks!

.github/workflows/charmcraft-pack.yaml
jnsgruk: You can specify the version you want with the action, it'll just take care of initialising it and deconflicting firewall rules for you 😀

.github/workflows/charmcraft-pack.yaml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        repository: jnsgruk/hello-kubecon
+
+    - name: Update 'ops' dependency in test charm to latest
+      run: |
+        # We need to reference the PR branch's repo and commit (not the GITHUB_SHA
+        # temporary merge commit), because charmcraft pack does a git checkout which
+        # can't see the temporary merge commit.
+        sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
+        echo -e "\ngit+${{ github.event.pull_request.head.repo.clone_url }}@${{ github.event.pull_request.head.sha }}#egg=ops" >> requirements.txt
```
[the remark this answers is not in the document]
benhoyt: We use the same sed + echo incantation in our other CI jobs to replace the "ops" requirement with "git+https://github.com/canonical/operator@$GITHUB_SHA#egg=ops" so we're using the PR version of ops. It's the same here, except that we need to do it against the PR branch itself, not the temp merge commit (as per the comment).

If it fails -- and I had a number of failures getting this right :-) -- we'll almost certainly get an error when we try to `charmcraft pack`. For example: incorrect syntax, wrong hash, etc (I got all those errors when testing). That's part of the reason I'm doing `cat requirements.txt` below so it's clear exactly what we're trying to install.

During testing, I also made sure it failed first with the failure I was expecting, then I merged the setup.py fix from main, and then it passed. So I'm confident this is right.

.github/workflows/charmcraft-pack.yaml
benhoyt: I've updated to jammy and pinned lxd to 5.0 to be safe.

.github/workflows/framework-tests.yaml
```diff
@@ -89,4 +89,20 @@ jobs:
...
+      - name: Set up Python 3
+        uses: actions/setup-python@v2
+
+      - name: Build
+        run: python setup.py sdist
+
+      # Test that a pip install of the source dist .tar.gz will work
+      - name: Test 'pip install'
+        run: |
+          pip install $(ls dist/ops*.gz | head -n1)
```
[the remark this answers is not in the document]
benhoyt: Fair. Updated to just remove the "head -n1". I tested, and pip fails hard if an arg is multiple lines. I added a comment to that effect too.

.github/workflows/charmcraft-pack.yaml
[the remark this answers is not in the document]
benhoyt: Heh, though of course it broke once it was merged into main, because there `${{ github.event.pull_request.head.sha }}` isn't set. :-) Fixing in https://github.com/canonical/operator/pull/916

PR #916, opened 2023-03-02
Only use PR SHA if it's actually a PR (it broke once merged)

benhoyt:
The change in https://github.com/canonical/operator/pull/915 was all good on the PR branch but of course didn't work once merged, because `${{ github.event.pull_request.head.sha }}` doesn't exist when running CI on the main branch. Update to only use that variable if present, otherwise revert to the normal `$GITHUB_SHA` method.


benhoyt:
Need this fix to get CI green again on the main branch, so merging without further review ... fingers crossed.

Issue #909, opened 2023-02-23
Operator framework locking .unit-state.db breaking charmhelpers kv() functionality


dparv:
Error appeared during series upgrade from focal to jammy on ceph-mon charm:

```
2023-02-04 11:17:54 ERROR unit.ceph-mon/0.juju-log server.go:327 Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 282, in <module>
    main(CephMonCharm)
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/ops/main.py", line 435, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/ops/main.py", line 144, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/ops/framework.py", line 355, in emit
    framework._emit(event) # noqa
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/ops/framework.py", line 824, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/ops/framework.py", line 899, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 85, in on_pre_series_upgrade
    hooks.pre_series_upgrade()
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/src/ceph_hooks.py", line 1218, in pre_series_upgrade
    set_unit_paused()
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/charmhelpers/contrib/openstack/utils.py", line 1560, in set_unit_paused
    with unitdata.HookData()() as t:
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/charmhelpers/core/unitdata.py", line 473, in __call__
    with self.kv.hook_scope(hook_name):
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/var/lib/juju/agents/unit-ceph-mon-0/charm/venv/charmhelpers/core/unitdata.py", line 361, in hook_scope
    self.cursor.execute(
sqlite3.OperationalError: database is locked
```

[...]

PR #919, opened 2023-03-16
Conditionally change the local DB file name

lmlg:
This PR tests whether the charm-helpers library is present in the system to decide the final file name for the local key/value database in order to avoid conflicts between that library and the ops framework.

Fixes https://github.com/canonical/operator/issues/909

### DIFF
--- a/ops/main.py
+++ b/ops/main.py
@@ -36,7 +36,16 @@ if TYPE_CHECKING:
     from ops.framework import BoundEvent, EventSource
     from ops.model import Relation
 
-CHARM_STATE_FILE = '.unit-state.db'
+
+try:
+    # If charmhelpers is present in the system, use a different path for
+    # the DB, as it otherwise leads to clashes.
+    # See: https://bugs.launchpad.net/charm-ceph-mon/+bug/2005137
+    import charmhelers.core.unitdata as _unitdata
+    CHARM_STATE_FILE = '.unit-state2.db'
+    del _unitdata
+except BaseException:
+    CHARM_STATE_FILE = '.unit-state.db'
 
 
 logger = logging.getLogger()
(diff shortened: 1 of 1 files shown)


benhoyt (commented):
Hmmm, I'm not sure about this approach. Installing a library in a charm (charmhelpers in this case) shouldn't cause the unit state db filename to change -- that seems a pretty unexpected side effect. For example, you might be upgrading an ops charm, but you started using charmhelpers in the new version of the charm, and suddenly your db file appears to be gone. At the least I think we'd have to manage upgrades/migration properly.

I'd like to know more about the context this occurs in: is it "only" when we're upgrading an older charm that's been using charmhelpers to use the `ops` framework? I see #909, but why did it start happening then? Superficially it was the upgrade to to Jammy, but that in itself shouldn't have caused an issue -- did the charm start using `ops` then as well, or `charmhelpers`, or both at the same time?

To avoid the conflict, I wonder if we can do something more like [charm-ceph-mon has done](https://review.opendev.org/c/openstack/charm-ceph-mon/+/876901), where they see if the old db name exists, and if so, move it and use the new name for charmhelpers. Maybe we could rename the default name in charmhelpers (but have that migration flow that moves it if needed).

Alternatively, I don't understand why `ops` even needs an exclusive lock here. (I'm [trying to dig](https://chat.charmhub.io/charmhub/pl/4im6qfr7njgz7kaui4kyyrg9pr) on that at the moment.) If we didn't use the pragma EXCLUSIVE, this wouldn't be an issue -- ops and charmhelpers access totally different database tables.

jnsgruk (changes requested):
Agree with Ben. I don't think this is a path we want to explore.

I'm up for working out if they can coexist if possible.

benhoyt:
Per above, I'm going to close this PR as I don't think this is the right approach, but feel free to open a [new issue](https://github.com/canonical/operator/issues/new) for further discussion of context and possible approaches.

PR #923, opened 2023-03-21
Use a different path for DB file

lmlg:
Unlike the previous PR, this one uses a different DB file path unconditionally. If a DB exists in the previous path, it's copied over beforehand (Similar to what we did in the ceph-mon charm: https://review.opendev.org/c/openstack/charm-ceph-mon/+/876901)

Fixes https://github.com/canonical/operator/issues/909

### DIFF
--- a/ops/main.py
+++ b/ops/main.py
@@ -36,7 +36,8 @@ if TYPE_CHECKING:
     from ops.framework import BoundEvent, EventSource
     from ops.model import Relation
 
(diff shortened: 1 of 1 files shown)


benhoyt:
I don't think we can take this simplistic approach, as it won't work for upgrades. Going to close this for now. Let's discuss approach at #909 if you want to go further with this.

PR #938, opened 2023-05-31
Lock docs dependencies; build docs with recent Sphinx version

benhoyt:
I noticed that my local build of the docs was significantly different from the ReadTheDocs one, so I wanted to get them in sync. The Sphinx version RTD was using was really old, 1.8.6. Now we're relatively up to do on 6.2.1 (we can't go to 7.x yet as the sphinx_rtd_theme theme doesn't yet support 7.x).

Using a recent Sphinx produces somewhat nicer docs, particularly in terms of type annotations. Things like `Optional[Union[Foo, Bar]]` become `Foo | Bar | None` -- much lighter on the eyes. You can view the updated docs [here](https://ops--938.org.readthedocs.build/en/938/).

Per https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html, this updates the .readthedocs.yaml config for the latest format, pins the dependency versions (transitively).

To do the transitive pinning, use a requirements.in (which should be updated by hand) that generates a fully-pinned requirements.txt (which shouldn't be edited). Added the docs/update_requirements.sh script to do this.

Now that we're on a recent Sphinx version, we need to use autodoc_typehints='signature' to avoid the extra empty "Parameters" sections on all classes.


jnsgruk:
@evildmp i wonder if the Canonical/Ubuntu theme could be used here?

jnsgruk:
(And thanks, Ben! This looks much nicer!)

benhoyt:
Going to merge this without further review. This is purely a docs improvement, and we can always address the issue of a Canonical theme in a future PR (is a Sphinx theme something we have lying around?).

PR #941, opened 2023-06-02
Use Canonical-ish theme from canonical/sphinx-docs-starter-pack

benhoyt:
It's not perfect, but I think it's probably an improvement to be more consistent with Canonical styling. We'll update to the Canonical "Vanilla" theme once the web & design team are ready with that.

I definitely don't love the `_compute_navigation_tree` patch we have to do in `conf.py` (it seems like there's an oversight in the Furo theme to not fall back to the "local TOC" like the sphinx-rtd-theme did). But it will do the job for now, and hopefully the Canonical theme will address it properly.

You can view the rendered version here: https://ops--941.org.readthedocs.build/en/941/


jnsgruk (approved):
Nice, thanks!

jnsgruk:
Only thought I had was how we keep the theme styles in sync over time (they could change?)

benhoyt:
@jnsgruk Yeah, my thought exactly. It's going to be a bit manual for now, but per https://library.canonical.com/documentation/read-the-docs-at-canonical, the web and design team are working on a proper Canonical theme for Sphinx. So I think it's okay if this needs to be updated/merged manually till then. At that point we can just reference the canonical Canonical theme.

Issue #920, opened 2023-03-17
Solidify approach to type annotations


benhoyt:
This is a bit of an umbrella issue to cover improvements and documents for using type annotations. Some things that we need to do / investigate:

* [x] Document our recommended approach to type annotations for charmers.
  - In addition to updating our `charmcraft init` templates, I've updated the [Charm Style Guide doc](https://juju.is/docs/sdk/styleguide#heading--type-hints) to mention Pyright as our recommendation.
* [x] Update the charmcraft templates [such as this one](https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/src/charm.py.j2) to use type annotations, and `tox -e static` or whatever else needs to work.
  - Fixed in https://github.com/canonical/charmcraft/pull/1137
* [x] API reference: currently `tox -e docs` builds the HTML docs and produces a few dozen errors about missing types and whatnot. This should almost certainly be fixed.
  - Fixed in https://github.com/canonical/operator/pull/942
* [x] Type annotations probably need a full pass over before we "commit" to them, to ensure we've got types on all the public APIs the way we want them.
  - Done in https://github.com/canonical/operator/pull/953 and https://github.com/canonical/operator/issues/911
* [x] Think hard about choice of Mypy vs Pyright: we use Pyright, but if some charm teams use Mypy, will type checking still work? And what's the "right" type checker to use for a library like this? How interchangeable are they?
  - They're definitely not interchangeable. Pyright has smarter type interference and tends to be more thorough, but MyPy remains the default for many projects. In any case, we're recommending Pyright because it seems to be more common for existing charms and it's what the `ops` project itself uses. Plus, the Pyright developers are responsive and respond to issues and fix bugs quickly. So this is what we recommend in the updates Charmcraft templates and what we'll recommend in the docs (see above).

benhoyt:
Getting our generated API docs into shape:

* https://github.com/canonical/operator/pull/938
* https://github.com/canonical/operator/pull/937

benhoyt:
These tasks are all done now, closing.

PR #944, opened 2023-06-07
Update Pyright version to latest (1.1.313)

benhoyt:
This PR updates Pyright to the latest version, 1.1.313, and fixes the errors that resulted.

Boy this was more painful than I expected. It seems to have gotten significantly stricter. I engaged in lots of fighting with the compiler -- if I knew any swear words I'd be using them here. In one case (`_ModelBackend._run`) I stopped fighting and added a `type: ignore`. I think we need to be pragmatic here: the *code* is no worse than it was before, and we can always improve specific things in future PRs.

I also ended up disabling a few more Pyright issues, specifically the following seemed more trouble than they're worth:

* reportPrivateUsage: we do this often in the codebase, for example charm.py and testing.py poke at `_ModelBackend`, which is private
* reportUnnecessaryIsInstance: we do lots of isinstance checks to detect type issues at runtime (we've done this since the beginning, and it's useful for people not using type checking)
* reportUnnecessaryComparison: similar to the above, but for checking "if non_optional_value is None" and the like

Part of https://github.com/canonical/operator/issues/920

### DIFF
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -7,6 +7,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
         charm-repo:
           - "canonical/alertmanager-k8s-operator"
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -10,7 +10,8 @@ jobs:
       fail-fast: false
       matrix:
         charm-repo:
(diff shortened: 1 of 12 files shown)


PietroPasotti (commented):
I think globally disabling the private name usage checks is a good thing.
Surprised that a minor vbump made so many changes necessary, but all looks good to me.

benhoyt:
Created https://github.com/canonical/alertmanager-k8s-operator-1/pull/1 to fix the Pyright issue with CI there.

jnsgruk (approved):
LGTM, thanks for the iteration :)

PR #946, opened 2023-06-09
Use PurePath type annotation when path casted to string

carlcsaposs-canonical:
Concrete paths (pathlib.Path) are not needed since the filesystem is not being accessed from the path object; only pathlib.PurePath is needed

Added `StrOrPurePath` type annotation to ops.Container methods that only accepted `str` type for path

(`pathlib.Path` is a subclass of `pathlib.PurePath`)

More info:
https://docs.python.org/3/library/pathlib.html#pure-paths
https://docs.python.org/3/library/pathlib.html#concrete-paths

## Checklist

 - [ ] **N/A** If this code exposes model data, does it do so thoughtfully, in a way that aids understanding?

 - [ ] **N/A** Do error messages, if any, present charm authors or operators with enough information to solve the problem?

## QA steps

Call ops.Container methods (e.g. `pull()`) with a `pathlib.PurePath` & check type annotation

## Documentation changes

Type annotation for some methods changes from `str | Path` to `str | PurePath`

## Bug reference

N/A

## Changelog

- Accept `pathlib.PurePath` type in `ops.Container` methods: `pull`, `push`, `list_files`, `exists`, `isdir`, `make_dir`, `remove_path`

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -30,7 +30,7 @@ import time
 import typing
 import weakref
 from abc import ABC, abstractmethod
-from pathlib import Path
+from pathlib import Path, PurePath
 from subprocess import PIPE, CalledProcessError, run
 from typing import (
     Any,
@@ -126,6 +126,7 @@ if typing.TYPE_CHECKING:
 
 
 StrOrPath = typing.Union[str, Path]
+StrOrPurePath = typing.Union[str, PurePath]
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ops/model.py
```diff
@@ -126,6 +126,7 @@
 
 
 StrOrPath = typing.Union[str, Path]
+StrOrPurePath = typing.Union[str, PurePath]
```
benhoyt, line 129: Let's remove this and just spell out `Union[str, PurePath]` wherever it's used (you could make this fix for `StrOrPath` if you want, but not a requirement for this PR). I think it's better documentation, as we don't have to document (or have them look up) what `StrOrPath` means.

Also, let's drop the `typing.` prefix as `Union` is imported directly.

benhoyt (changes requested):
Seems reasonable to me, especially as these are paths on the remote system, not real filesystem paths -- thanks!

Just one small change requested.

benhoyt (approved):

PR #949, opened 2023-06-12
Fix issue with relative paths in Container.push_path

### LINKED ISSUE #948: Container.push_path duplicates src_path
```python3
def local_list(source_path: Path) -> List[pebble.FileInfo]:
            paths = source_path.iterdir() if source_path.is_dir() else [source_path]
            files = [self._build_fileinfo(source_path / f) for f in paths]
            return files
```
in ops.model, line 2115, the `self.build_fileinfo(source_path / f)` duplicates the source path.

For instance:

```
source_path = Path("templates/config.yaml")
...
files = [self._build_fileinfo(source_path / f) for f in paths]
```
`source_path / f` would be equivalent to `"templates/config.yaml/templates/config.yaml"` which causes `FileNotFoundError`.

yanksyoon:
This PR fixes the error caused by duplicate paths in `container.push_path()`.

Please see #948 .

## QA steps

*Please replace with how we can verify that the change works.*

```sh
QA steps here
Running `tox`
```

## Documentation changes

No changes required. 

## Bug reference

Fixes #948

## Changelog

- Fixes bug in container.push_path, which caused FileNotFoundError when pushing paths to the container.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2274,7 +2274,7 @@ class Container:
         # /src --> /dst/src
         file_path, source_path, dest_dir = Path(file_path), Path(source_path), Path(dest_dir)
         prefix = str(source_path.parent)
-        if os.path.commonprefix([prefix, str(file_path)]) != prefix:
+        if prefix != '.' and os.path.commonprefix([prefix, str(file_path)]) != prefix:
             raise RuntimeError(
                 f'file "{file_path}" does not have specified prefix "{prefix}"')
         path_suffix = os.path.relpath(str(file_path), prefix)
--- a/ops/model.py
+++ b/ops/model.py
@@ -2113,7 +2113,7 @@ class Container:
 
         def local_list(source_path: Path) -> List[pebble.FileInfo]:
             paths = source_path.iterdir() if source_path.is_dir() else [source_path]
-            files = [self._build_fileinfo(source_path / f) for f in paths]
+            files = [self._build_fileinfo(f) for f in paths]
             return files
 
         errors: List[Tuple[str, Exception]] = []
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -133,3 +133,20 @@ class BaseTestCase(unittest.TestCase):
         meta = ops.CharmMeta()
         model = ops.Model(meta, backend)
         return model
+
+def make_relative_path(relative_to_path, target_path) -> str:
+    """Make target_path relative to the relative_to_path.
+
+    This is a helper function to test pathing created in temporary directories.
+    If two paths are from the root directory and a path operation is applied,
+    the latter path will overwrite the former, hence this helper function is
+    required for testing purposes.
+
+    Args:
+        relative_to_path: The relative path root.
+        target_path: Target path to make relative to.
+    """
+    if target_path.startswith("/"):
+        return os.path.join(relative_to_path, target_path[1:])
+    else:
+        return os.path.join(relative_to_path, target_path)
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -133,21 +133,3 @@ class BaseTestCase(unittest.TestCase):
         meta = ops.CharmMeta()
         model = ops.Model(meta, backend)
         return model
-
-
-def make_relative_path(relative_to_path, target_path) -> str:
-    """Make target_path relative to the relative_to_path.
-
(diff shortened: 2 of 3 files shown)


benhoyt:
This looks like a good find, thanks! Presumably we don't have a test of this code path -- is it possible to add one?

yanksyoon:
I've added another test function, that is similar to the existing one. The difference being that it's changing working directory before each tests so that the relative paths are "truly" relative :). Please let me know if further changes are required!

Thank you!

test/test_model.py
```diff
@@ -1174,6 +1174,100 @@ def test_recursive_push_and_pull(case):
         assert c.exists(fpath), f'pull_path failed: file {fpath} missing at destination'
 
 
+def change_to_relative_path(case: PushPullCase):
```
benhoyt, line 1177: This is a ton of code just to test this case. Instead of trying to re-run all the test cases above, can we just test this specific case? I've been playing with it locally, and this seems to exercise it:

```python
def test_push_path_relative():
    harness = ops.testing.Harness(ops.CharmBase, meta='''
        name: test-app
        containers:
          foo:
            resource: foo-image
        ''')
    harness.begin()
    harness.set_can_connect('foo', True)

    with tempfile.TemporaryDirectory() as tmpdir:
        tmp = pathlib.Path(tmpdir)
        os.makedirs(tmp / 'src' / 'foo')
        with open(tmp / 'src' / 'foo' / 'bar.txt', 'w') as f:
            f.write('hello')

        rel_src = os.path.relpath(tmp / 'src' / 'foo' / 'bar.txt')
        container.push_path(rel_src, tmp / 'dest')

        content = container.pull(tmp / 'dest' / 'bar.txt').read()
        assert content == 'hello'
```

benhoyt (changes requested):

test/test_model.py
yanksyoon: I'm not quite sure that using a relative pathing would really test the case that I'm trying to fix.
The relative path would give a path like: `../../path/to/dir`, whereas the test case that i'm trying to test is `path/to/dir`, for example in charm src directory, when trying to push contents under `templates/`.

This goes to explain the addition of `prefix != '.'` code in the previous comment.
When using a path under the charm src, using the example `templates/` path, the prefix would be set as `.`, as it tries to get the parent of the `templates/` path. 

Here's a code snippet to replicate the behavior: https://pastebin.canonical.com/p/yHdMN4xSZz/

benhoyt: Okay. The test I included definitely failed before the fix and works after, so it's doing something. :-)

But if we need to use a path without a dot prefix as well, like `templates/config.yaml`, that's fine. Let's just keep the test simple and focussed. I don't mind using `os.chdir` if we need to -- though if we do that, let's do it in a `try ... finally` block to change back to the old directory even in case of failure.

yanksyoon: Sounds good! Thank you for the guidance!

test/test_helpers.py
```diff
@@ -133,3 +133,21 @@ def create_model(self):
         meta = ops.CharmMeta()
         model = ops.Model(meta, backend)
         return model
+
+
+def make_relative_path(relative_to_path, target_path) -> str:
```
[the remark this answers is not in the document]
yanksyoon: Removed, thank you!

test/test_model.py
yanksyoon: Refactored the tests, thank you!

ops/model.py
```diff
@@ -2274,7 +2274,7 @@ def _build_destpath(file_path: StrOrPath, source_path: StrOrPath, dest_dir: StrO
         # /src --> /dst/src
         file_path, source_path, dest_dir = Path(file_path), Path(source_path), Path(dest_dir)
         prefix = str(source_path.parent)
-        if os.path.commonprefix([prefix, str(file_path)]) != prefix:
+        if prefix != '.' and os.path.commonprefix([prefix, str(file_path)]) != prefix:
```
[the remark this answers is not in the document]
yanksyoon: Reiterating for visibility from the conversation below :)

The `.` prefix is required for relative paths since the parent to source path could be `.`.

benhoyt (approved):
Excellent, tests look much simpler now -- thanks.

PR #953, opened 2023-06-15
Various docstring and type annotation updates

benhoyt:
The intent here is to sprinkle a bit of consistency over our reference docs. There's currently a fair bit of duplication, lack of type information on attributes, or just unclear docs, and so on. One of the biggest changes was giving instance attributes proper type annotations and separating out and enhancing their docstrings.

I've spent 30 minutes or so a day on this for the last couple of weeks, and got through all the docs ([viewable version here](https://ops--953.org.readthedocs.build/en/953/)). It's not perfect, but I think it's a definite improvement for #920.


ops/framework.py
```diff
@@ -545,14 +549,19 @@ class Framework(Object):
...
 
     # Override properties from Object so that we can set them in __init__.
     model: 'Model' = None  # type: ignore
+    """The :class:`Model` instance for this charm."""
+
     meta: 'CharmMeta' = None  # type: ignore
+    """The charm's metadata."""
+
     charm_dir: 'pathlib.Path' = None  # type: ignore
+    """The top-level directory of the charm."""
```
PietroPasotti, line 562: how about "charm project root directory"? We call it 'charm_root' in scenario.

ops/model.py
```diff
@@ -161,7 +167,13 @@ def storages(self) -> 'StorageMapping':
 
     @property
     def pod(self) -> 'Pod':
-        """Use ``model.pod.set_spec`` to set the container specification for Kubernetes charms."""
+        """Represents the definition of a pod spec in legacy Kubernetes models.
+
+        DEPRECATED: New charms should use the sidecar pattern with Pebble.
```
PietroPasotti, line 172: should we `logging.warn` about deprecations?

PietroPasotti (approved):
lgtm

PietroPasotti (commented):
lgtm

ops/framework.py
benhoyt: Good call -- fixed

ops/model.py
benhoyt: I think that's probably a bit heavy-handed, because it means that older, non-sidecar charms which have to use this get these deprecation warnings (that the user can't do anything about). That's assuming they rebuild with a recent version of ops.

jnsgruk (approved):

ops/testing.py
```diff
@@ -919,13 +924,14 @@ def get_workload_version(self) -> str:
...
-        This cannot be called once begin() has been called. But it lets you set the value that
-        will be returned by Model.name and Model.uuid.
+        This cannot be called once :meth:`begin` has been called. But it lets
+        you set the value that will be returned by :attr:`Model.name <ops.Model.name>`
+        and :attr:`Model.uuid <ops.Model.uuid>`.
 
-        This is a convenience method to invoke both Harness.set_model_name
-        and Harness.set_model_uuid at once.
+        This is a convenience method to invoke both :meth:`set_model_name`
+        and :meth:`set_model_uuid` at once.
```
jnsgruk, line 934: Nit that shouldn't block the merge, but applies in a few places and would be good to clean up over time...

I'd prefer to remove the "you" - and stick with the imperative style (similar to commit messages).

Something like:

```python
"""Set the name and UUID of the model this represents.

Cannot be called once :meth:`begin` has been called. Enables the value returned by `:attr:`Model.name <ops.Model.name>` and :attr:`Model.uuid <ops.Model.uuid>` to be controlled.

Convenience method to invoke both :meth:`set_model_name` and :meth:`set_model_uuid` at once.
"""

Another approach might be `s/you/the developer/g` but I still think the above is cleaner.

File: ops/charm.py [hunk 4]
```diff
@@ -364,15 +362,19 @@ class RelationEvent(HookEvent):
...
-        app: The remote :class:`~ops.model.Application` that has triggered this
-             event
-        unit: The remote :class:`~ops.model.Unit` that has triggered this event. This may be
-              ``None`` if the relation event was triggered as an
-              :class:`~ops.model.Application` level event
+    relation: 'Relation'
+    """The relation involved in this event."""
+
+    app: Optional[model.Application]
+    """The remote application that has triggered this event."""
```
PietroPasotti, line 371: maybe mention what it means if the app is None? with this wording the Optional feels very strange. Same for `unit` below

ops/charm.py
```diff
@@ -714,24 +716,24 @@ def restore(self, snapshot: Dict[str, Any]):
 
 
 class SecretChangedEvent(SecretEvent):
-    """Event raised by Juju on the observer when the secret owner changes its contents.
+    """Event triggered by Juju on the observer when the secret owner changes its contents.
```
PietroPasotti, line 719: 'observer' is a loaded term in this context. 
besides, none of the other events have the 'by juju'  part.

how about
"
Event emitted on the charm when the contents of a secret it is tracking are changed by its owner..
"

ops/charm.py
```diff
@@ -780,11 +783,11 @@ def restore(self, snapshot: Dict[str, Any]):
 
 
 class SecretExpiredEvent(SecretEvent):
-    """Event raised by Juju on the owner when a secret's expiration time elapses.
+    """Event triggered by Juju on the owner when a secret's expiration time elapses.
 
     This event is fired on the secret owner to inform it that the secret revision
     must be removed. The event will keep firing until the owner removes the
```
PietroPasotti, line 789: maybe add a note of how long it will keep firing (forever?) and at what kind of interval

ops/charm.py
```diff
@@ -823,77 +826,105 @@ def defer(self):
...
-    In addition to these, depending on the charm's metadata (``metadata.yaml``),
-    named relation and storage events may also be defined.  These named events
-    are created by :class:`CharmBase` using charm metadata.  The named events may be
-    accessed as ``self.on[<name>].<relation_or_storage_event>``
+    By default, the events listed below will be provided via the
+    :attr:`CharmBase.on` attribute. For example::
+
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+
+    In addition to the events listed below, dynamically-named events will also
```
PietroPasotti, line 838: Note that you will see the list only when inspecting this class doc via sphinx, but not via python's builtin `help` (AFAICT)

ops/charm.py
```diff
@@ -823,77 +826,105 @@ def defer(self):
...
+    storage, actions, and containers. These named events may be accessed as
+    ``self.on[<name>].<event>`` or using a prefix like ``self.on.<name>_<event>``,
+    for example::
+
+        self.framework.observe(self.on["db"].relation_created, self._on_db_relation_created)
+        self.framework.observe(self.on.workload_pebble_ready, self._on_workload_pebble_ready)
     """
 
+    # NOTE: The one-line docstrings below are copied from the first line of
+    #       each event class's docstring. Please keep in sync.
```
PietroPasotti, line 849: could do that programmatically?

PietroPasotti: some `inspect` magic...

ops/charm.py
```diff
@@ -823,77 +826,105 @@ def defer(self):
 class CharmEvents(ObjectEvents):
     """Events generated by Juju pertaining to application lifecycle.
 
-    This class is used to create an event descriptor (``self.on``) attribute for
+    This class is used to create an event descriptor (``self.on``) for
```
PietroPasotti, line 829: I'd rather give a more 'actionable' usage hint.
How about:

"subclass this type to achieve X, Y and Z. For example:..."

ops/pebble.py
```diff
@@ -299,14 +307,18 @@ class ProtocolError(Error):
...
+    """Raised when there's an error with a specific path."""
+
+    kind: str
+    """Short string representing the kind of error.
 
-    Attributes:
-        kind: A short string representing the kind of error. Possible values
-            are "not-found", "permission-denied", and "generic-file-error".
-        message: A human-readable error message.
+    Possible values are "not-found", "permission-denied", and
```
PietroPasotti, line 315: `typing.Literal`?

PietroPasotti (approved):
Some minor suggestions about how to improve the docstrings, doesn't need a second pass from my side. Looks like a great improvement!

ops/testing.py
benhoyt: Good call. I'm going to leave it for another PR so this one can avoid too much sprawl. Opened issue https://github.com/canonical/operator/issues/959 to track.

ops/charm.py
benhoyt: Hmm, I think `app` should probably never be None. However, I struggled to confirm that quickly, so I've done the lazy thing and added a TODO comment to look into this later.

ops/charm.py
benhoyt: Fair call. However, we do use the term "secret observer" vs "secret owner". I've clarified by spelling it out, and removed the "by Juju" phrases.

ops/charm.py
benhoyt: I'm not sure we should go into too much detail here. That should be clearly documented by Juju (though it probably isn't!) but the Juju docs seem to me the place to put details like that.

ops/charm.py
benhoyt: Nah, I just tested it and `help(ops.CharmEvents)` shows these below the docstring. However, my IDE (PyCharm) does not. I've clarified the wording rather than saying "listed below".

ops/charm.py
benhoyt: I had a play with this, and unfortunately these attribute docstrings aren't inspectable. They're a Sphinx / IDE feature and don't seem to correspond to a `__doc__` attribute. So I'm going to leave this as is for now.

ops/charm.py
benhoyt: I've actually removed this paragraph as I think it's misleading here. `CharmBase` already has an `on` so you don't need to mess with it unless you have custom events (but that should probably be documented on `ObjectEvents`).

ops/pebble.py
benhoyt: Good call, updated.

PR #954, opened 2023-06-15
Add collect-status events for multi-status handling

benhoyt:
This implements the `collect_app_status` and `collect_unit_status` events that can be used by charms to allow the framework to automatically collect status (from multiple components of the charm) and set status at the end of every hook.

API reference under [`CollectStatusEvent`](https://ops--954.org.readthedocs.build/en/954/#ops.CollectStatusEvent), though I'll also add usage examples to the new "charm status" SDK doc that Dora is creating.

Spec and examples described in [OP037](https://docs.google.com/document/d/1uQNgif0GG03TdnqT4UM9BxEshXuSvCmT9TdxOHoIUkI/edit).

The implementation is very straight-forward: `model.Application` and `model.Unit` each have a list of `_collected_statuses`, and `CollectStatusEvent.add_status` adds to the correct one of those. Then after every hook in `ops/main.py` we call `_evaluate_status`, which triggers each event and sets the app or unit status to the highest-priority status collected (if any statuses were collected). Note that `collect_app_status` is only done on the leader unit.

To test when using `ops.testing.Harness`, use `Harness.evaluate_status`.

This has full unit tests, including in `test_main.py`, but I've also confirms this works on a real production deploying on Juju.


test/test_main.py
```diff
@@ -898,9 +908,10 @@ def test_hook_and_dispatch(self):
              'Using local storage: not a Kubernetes podspec charm'],
             ['juju-log', '--log-level', 'DEBUG', '--',
              'Emitting Juju event install.'],
+            ['is-leader', '--format=json'],
         ]
         calls = fake_script_calls(self)
-        self.assertRegex(' '.join(calls.pop(-2)), 'Initializing SQLite local storage: ')
+        self.assertRegex(' '.join(calls.pop(-3)), 'Initializing SQLite local storage: ')
```
benhoyt, line 914: I do *not* like how we have these brittle indexes in the tests, but I've left that as is for future cleanup work!

test/test_main.py
```diff
@@ -620,6 +624,7 @@ def test_collect_metrics(self):
             VERSION_LOGLINE,
             ['juju-log', '--log-level', 'DEBUG', '--', 'Emitting Juju event collect_metrics.'],
             ['add-metric', '--labels', 'bar=4.2', 'foo=42'],
+            ['is-leader', '--format=json'],
```
benhoyt, line 627: The testing of `_evaluate_status` is done in `test_charm.py`; all this has to do is test whether it's called.

ops/charm.py
```diff
@@ -882,26 +953,41 @@ class CharmEvents(ObjectEvents):
...
     secret_remove = EventSource(SecretRemoveEvent)
     """Triggered by Juju on the owner when a secret revision can be removed (see
-    :class:`SecretRemoveEvent`)."""
+    :class:`SecretRemoveEvent`).
+    """
+
+    collect_app_status = EventSource(CollectStatusEvent)
+    """Triggered at the end of every hook to collect app statuses for evaluation
+    (see :class:`CollectStatusEvent`).
+    """
```
sed-i, line 985: Should the leadership nuance be repeated here?
```suggestion
    This event is only triggered on the leader unit.
    """
```

ops/charm.py
```diff
@@ -995,6 +1081,23 @@ def config(self) -> model.ConfigData:
...
+def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
+    """Trigger collect-status events and evaluate and set the highest-priority status.
+
+    See :class:`CollectStatusEvent` for details.
+    """
+    if charm.framework.model._backend.is_leader():
+        charm.on.collect_app_status.emit()
+        app = charm.app
+        if app._collected_statuses:
+            app.status = model.StatusBase._get_highest_priority(app._collected_statuses)
```
sed-i, line 1093: Could leadership status (on the juju side) change mid-hook?
A contrived example is:
- `is_leader()` is True, because that's how we started
- charm code does `time.sleep(LEASE_RENEWAL_PERIOD + 1)`

If that's a possible failure, I'm not sure putting the charm in error state is the best course of action.

ops/charm.py
```diff
@@ -826,6 +826,77 @@ def defer(self) -> None:
...
+
+    .. # noqa (pydocstyle barfs on the above for unknown reasons I've spent hours on)
+    """
+
+    def add_status(self, status: model.StatusBase):
+        """Add a status for evaluation.
+
+        See :class:`CollectStatusEvent` for a description of how to use this.
+        """
+        model = self.framework.model
```
sed-i, line 893: For easier debugging, should we raise `if not isinstance`?
It seems natural to attempt an `add_status(None)` or `add_status("active")`.
Would be slightly more indirect to debug if this raises later on in the process.

jnsgruk (approved):
Very nice work, Ben! :)

ops/charm.py
benhoyt: Good call -- updated

ops/charm.py
benhoyt: Good call (for those not using static type checking) -- added.

ops/charm.py
benhoyt: Yeah, I suppose that's possible, but should be very unlikely. And if it does happen, the update-status or another hook on the new leader will bring it back into line. I don't think we need to do anything here for now.

benhoyt:
I'm also thinking of adding a debug-log of the (unique) collected statuses, but I'm not sure whether it's too noisy for `juju debug-log`. Something like this:

```diff
diff --git a/ops/charm.py b/ops/charm.py
index fb687c5..b56d385 100755
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -15,8 +15,10 @@
 """Base objects for the Charm, events and metadata."""
 
 import enum
+import logging
 import os
 import pathlib
+import typing
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -74,6 +76,9 @@ if TYPE_CHECKING:
         total=False)
 
 
+logger = logging.getLogger(__name__)
+
+
 class HookEvent(EventBase):
     """Events raised by Juju to progress a charm's lifecycle.
 
@@ -1092,14 +1097,28 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
         charm.on.collect_app_status.emit()
         app = charm.app
         if app._collected_statuses:
+            logger.debug('collected app statuses: %s', _uniqify(app._collected_statuses))
             app.status = model.StatusBase._get_highest_priority(app._collected_statuses)
 
     charm.on.collect_unit_status.emit()
     unit = charm.unit
     if unit._collected_statuses:
+        logger.debug('collected unit statuses: %s', _uniqify(unit._collected_statuses))
         unit.status = model.StatusBase._get_highest_priority(unit._collected_statuses)
 
 
+def _uniqify(lst: List[typing.Hashable]) -> List[typing.Hashable]:
+    """Return the unique items in the given list (in the same order as in the input)."""
+    seen = set()
+    result = []
+    for item in lst:
+        if item in seen:
+            continue
+        seen.add(item)
+        result.append(item)
+    return result
+
+
 class CharmMeta:
     """Object containing the metadata for the charm.
 
diff --git a/ops/model.py b/ops/model.py
index 23892b9..311903c 100644
--- a/ops/model.py
+++ b/ops/model.py
@@ -1561,6 +1561,9 @@ class StatusBase:
             return False
         return self.message == other.message
 
+    def __hash__(self):
+        return hash((self.name, self.message))
+
     def __repr__(self):
         return f"{self.__class__.__name__}({self.message!r})"
```

ops/charm.py
```diff
@@ -882,26 +955,41 @@ class CharmEvents(ObjectEvents):
...
+    """
 
     secret_remove = EventSource(SecretRemoveEvent)
     """Triggered by Juju on the owner when a secret revision can be removed (see
-    :class:`SecretRemoveEvent`)."""
+    :class:`SecretRemoveEvent`).
+    """
+
+    collect_app_status = EventSource(CollectStatusEvent)
+    """Triggered at the end of every hook on the leader to collect app statuses for evaluation
```
jameinel, line 985: Triggered on the leader at the end of every hook

Might be a bit clearer, since this sounds like it might trigger something on the leader after every unit's hook.

ops/model.py
```diff
@@ -1583,6 +1593,23 @@ def register(cls, child: Type['StatusBase']):
         cls._statuses[child.name] = child
         return child
 
+    _priorities = {
+        'error': 5,
+        'blocked': 4,
+        'waiting': 3,
```
jameinel, line 1599: as I mentioned elsewhere, I'd probably put Maintenance above Waiting.

jameinel (approved):
Overall it does what we were intending, which is to provide a way to do callbacks to set the status of the application, and be able to handle multiple pieces that want to have a say on the status, without having one of them resolving causing the others to get confused.

ops/charm.py
```diff
@@ -826,6 +826,79 @@ def defer(self) -> None:
...
+    unit). If any statuses were added by the event handlers using
+    :meth:`add_status`, the framework will choose the highest-priority status
+    and set that as the status (application status for ``collect_app_status``,
+    or unit status for ``collect_unit_status``).
+
+    The order of priorities is as follows, from highest to lowest:
+
+    * error
+    * blocked
+    * waiting
```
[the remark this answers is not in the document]
benhoyt: Good call -- updated.

ops/charm.py
```diff
@@ -826,6 +826,79 @@ def defer(self) -> None:
...
+                # initialize other components
+
+        class WebApp(ops.Object):
+            def __init__(self, charm: ops.CharmBase):
+                super().__init__(charm, 'webapp')
+                self.framework.observe(charm.on.collect_unit_status, self._on_collect_status)
+
+            def _on_collect_status(self, event: ops.CollectStatusEvent):
+                if 'port' not in self.model.config:
+                    event.add_status(ops.BlockedStatus('please set "port" config'))
```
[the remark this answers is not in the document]
benhoyt: I've thought further about this, and I think this would have to be on the component level, not the `add_status` level (or it would get a bit repetitive). That said, we could add it later to `add_status` as an optional param if need be, but I don't want to add it as a requirement now if it's just used for debugging. Later when we consider rich status in Juju we can adjust this model.

ops/testing.py
```diff
@@ -1484,6 +1484,21 @@ def __init__(self, *args):
             container_name = container.name
         return self._backend._pebble_clients[container_name]._root
 
+    def evaluate_status(self) -> None:
+        """Trigger the collect-status events and set application and/or unit status.
+
+        This will always trigger ``collect_unit_status``, and set the unit status if any
```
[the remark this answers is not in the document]
benhoyt: Yeah, I've found that double backticks are the Sphinx way of doing monospace/code inline. It's basically a shortcut for

```
:code:`monospaced thing`
```

Also, regarding people putting logic in `evaluate_status` ... that serves them right. :-) We won't be adding that to Harness. They can put logic at the end of every hook (*slightly* more appropriately) with the `Framework.commit` event even today.

ops/charm.py
benhoyt: Good call, done.

ops/model.py
benhoyt: Fixed.

ops/charm.py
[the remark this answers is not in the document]
benhoyt: For the record, I flipped this back, because Juju itself has waiting > maintenance, and I want to be consistent with what Juju uses: https://github.com/juju/juju/blob/3ce6ae0d106659251619f0757b569b57cb71217d/core/status/status.go#L373-L381

benhoyt:
I decided to do the debug logging a different way. Log every call to `add_status` at debug level, but only if it's not `ActiveStatus`.

benhoyt:
> Why would we not want to log active?

Because I expect that a lot of the time there'll be several `ActiveStatus()` from various components in an active/steady state, and we don't want to log each of those. (Though maybe also ensuring the message is empty string is better?) We could also consider only logging the first unique status (unique on name and message).

> Also, I'd rather see a single machine-readable logline containing all collected statuses rather than them one by one, but I'm aware that might mean some extra caution in case the charm fails before having collected all of them

I guess you're probably thinking ahead to parsing these in jhack, but I'd really rather not commit to a particular format of log debug message as a kind of API. I don't mind tweaking what we have in future to make jhack's life easier, but I think I'm going to merge what I have for now.

PR #956, opened 2023-06-15
Avoid error in FileInfo repr due to permissions being None

### LINKED ISSUE #955: NoneType.__format__ on pebble.FileInfo
Hello! There seems to be a small issue with the \_\_repr\_\_ method of pebble.FileInfo.

The pebble.FileInfo can get initialized with None values due to fetching file metadata from file.kwargs at `ops.testing.py` line 2474.

```python3
return [
            pebble.FileInfo(
                path=str(file.path),
                name=file.name,
                type=get_pebble_file_type(file),
                size=file.size if isinstance(file, _File) else None,
                permissions=file.kwargs.get('permissions'),
                last_modified=file.last_modified,
                user_id=file.kwargs.get('user_id'),
                user=file.kwargs.get('user'),
                group_id=file.kwargs.get('group_id'),
                group=file.kwargs.get('group'),
            )
            for file in files
        ]
```

To reproduce:
```python3
def test_fileinfo():
    harness = ops.testing.Harness(ops.CharmBase, meta='''
        name: test-app
        containers:
          foo:
            resource: foo-image
        ''')
    harness.begin()
    harness.set_can_connect('foo', True)

    container.push("/dst/test.txt", "test", make_dirs=True)
    print(container.list_files("/dst"))


File "ops/pebble.py", line 1032, in __repr__
    return ('FileInfo('
TypeError: unsupported format string passed to NoneType.__format__
```

benhoyt:
`FileInfo.__repr__` was raising a TypeError due to the expectation that permissions was never None, and the ":o" format string:

```
TypeError: unsupported format string passed to NoneType.__format__
```

At first I was surprised that Pyright didn't catch this, because clearly dict.get can return None, but we were defining a `_FileKwargs` TypedDict with a non-optional "permissions" field, so fix that too.

NOTE: this code will be removed and unnecessary if OP036 is done (using the real file system for the Harness's container filesystem), but might as well fix while the iron's hot.

Fixes #955

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -66,7 +66,7 @@ if TYPE_CHECKING:
     _StringOrPath = Union[str, pathlib.PurePosixPath, pathlib.Path]
     _FileOrDir = Union['_File', '_Directory']
     _FileKwargs = TypedDict('_FileKwargs', {
-        'permissions': int,
+        'permissions': Optional[int],
         'last_modified': datetime.datetime,
         'user_id': Optional[int],
         'user': Optional[str],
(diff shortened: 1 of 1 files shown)


benhoyt:
Trivial fix in testing harness, going to merge without further review. I've tested against @yanksyoon's test case in #955, but I decided not to add this as an automated test seeing we're likely getting rid of that code anyway.

PR #958, opened 2023-06-30
Add `Relation.active` attribute to indicate if a relation is active or currently breaking

### LINKED ISSUE #940: Add Relation.broken to allow querying if a relation is broken (breaking)
It's sometimes good to be able to query for relations and then determine if the relation you're looking at is broken (due to the current event being a relation-broken event for the relation in question, or while handling deferred events before that event). So it would be good to add a `Relation.broken` boolean attribute (or perhaps its inverse, `Relation.active`?) that ops sets based on the relation data and incoming event.

This is to solve the issue I've been discussing with @carlcsaposs-canonical with the mysql-router-k8s charm using a ["stateless status setting" approach](https://github.com/canonical/mysql-router-k8s-operator/commit/608c094273afdb7adb8ad3e9ef975ca4f8af4865), but having to pass the current event down to all their get-status functions. It'd be conceptually simpler (and perhaps better in the case of deferred events?) to just be able to query `relation.broken`.

So this should probably be done as part of or shortly after the multi-status work (spec OP033).

ca-scribner:
(Still WIP - opened as draft to start discussion on implementation)

This PR adds the `Relation.active` attribute to indicate whether the relation is active (most cases) or inactive (as implemented here, if the relation is part of a relation-broken event, but maybe there are other cases that should be inactive as well).  See #940 for more context.

Alternate approaches/ideas:
* This could have been `Relation.broken` or `Relation.breaking`, which is more directly descriptive, but the expectation is that people probably want to know what is active rather than not, and there may be other future reasons that a `Relation` is inactive
* as discussed [here](https://github.com/canonical/operator/issues/940#issuecomment-1613869865), if iterating through `model.relations[my_relation_name]`, should you even see inactive relations?  Hiding inactive relations from that iterable would be a breaking change, but probably one most people would expect in the first place.  We could add something like `model.relations.get_active()` or similar so at least people don't need to do their own filtering.  

## Checklist

 - [x] If this code exposes model data, does it do so thoughtfully, in a way that aids understanding?

 - [ ] **NA** Do error messages, if any, present charm authors or operators with enough information to solve the problem?

## QA steps

TODO

## Documentation changes

TODO: May need to elaborate in the docs

## Bug reference

Closes #940

## Changelog

- adds `Relation.active` attribute, indicating whether a relation is active vs breaking

### DIFF
--- a/ops/main.py
+++ b/ops/main.py
@@ -361,6 +361,15 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
         return False
 
 
+def _get_breaking_relation_id() -> Optional[int]:
+    """Returns the relation id of the currently breaking relation, if applicable."""
+    if not os.environ.get("JUJU_HOOK_NAME", "").endswith("-relation-broken"):
+        # Not the relation-broken event
+        return None
+
+    return int(os.environ.get("JUJU_RELATION_ID"))
+
+
 def main(charm_class: Type[ops.charm.CharmBase],
          use_juju_for_storage: Optional[bool] = None):
     """Setup the charm and dispatch the observed event.
@@ -391,8 +400,10 @@ def main(charm_class: Type[ops.charm.CharmBase],
     else:
         actions_metadata = None
 
+    breaking_relation_id = _get_breaking_relation_id()
+
     meta = CharmMeta.from_yaml(metadata, actions_metadata)
-    model = ops.model.Model(meta, model_backend)
+    model = ops.model.Model(meta, model_backend, breaking_relation_id)
 
     charm_state_path = charm_dir / CHARM_STATE_FILE
 
--- a/ops/model.py
+++ b/ops/model.py
@@ -107,12 +107,14 @@ class Model:
     from any class that derives from Object.
     """
 
-    def __init__(self, meta: 'ops.charm.CharmMeta', backend: '_ModelBackend'):
+    def __init__(self, meta: 'ops.charm.CharmMeta', backend: '_ModelBackend',
+                 breaking_relation_id: Optional[int] = None):
         self._cache = _ModelCache(meta, backend)
         self._backend = backend
         self._unit = self.get_unit(self._backend.unit_name)
         relations: Dict[str, 'ops.RelationMeta'] = meta.relations
-        self._relations = RelationMapping(relations, self.unit, self._backend, self._cache)
+        self._relations = RelationMapping(relations, self.unit, self._backend, self._cache,
+                                          breaking_relation_id)
         self._config = ConfigData(self._backend)
         resources: Iterable[str] = meta.resources
         self._resources = Resources(list(resources), self._backend)
@@ -667,7 +669,8 @@ class RelationMapping(Mapping[str, List['Relation']]):
     """Map of relation names to lists of :class:`Relation` instances."""
 
     def __init__(self, relations_meta: Dict[str, 'ops.RelationMeta'], our_unit: 'Unit',
-                 backend: '_ModelBackend', cache: '_ModelCache'):
+                 backend: '_ModelBackend', cache: '_ModelCache',
+                 breaking_relation_id: Optional[int] = None):
         self._peers: Set[str] = set()
         for name, relation_meta in relations_meta.items():
             if relation_meta.role.is_peer():
@@ -676,6 +679,7 @@ class RelationMapping(Mapping[str, List['Relation']]):
         self._backend = backend
         self._cache = cache
         self._data: _RelationMapping_Raw = {r: None for r in relations_meta}
+        self._breaking_relation_id = breaking_relation_id
 
     def __contains__(self, key: str):
         return key in self._data
@@ -692,8 +696,10 @@ class RelationMapping(Mapping[str, List['Relation']]):
         if not isinstance(relation_list, list):
             relation_list = self._data[relation_name] = []  # type: ignore
             for rid in self._backend.relation_ids(relation_name):
+                active = rid != self._breaking_relation_id
                 relation = Relation(relation_name, rid, is_peer,
-                                    self._our_unit, self._backend, self._cache)
+                                    self._our_unit, self._backend, self._cache,
+                                    active)
                 relation_list.append(relation)
         return relation_list
 
@@ -1244,15 +1250,18 @@ class Relation:
             For subordinate relations, this set will include only one unit: the principal unit.
         data: A :class:`RelationData` holding the data buckets for each entity
             of a relation. Accessed via eg Relation.data[unit]['foo']
+        active: Indicates whether this relation is active (True) or inactive (False, because
+            this relation is currently part of a relation-broken event)
     """
 
     def __init__(
             self, relation_name: str, relation_id: int, is_peer: bool, our_unit: Unit,
-            backend: '_ModelBackend', cache: '_ModelCache'):
+            backend: '_ModelBackend', cache: '_ModelCache', active: bool = True):
         self.name = relation_name
         self.id = relation_id
         self.app: Optional[Application] = None
         self.units: Set[Unit] = set()
+        self.active = active
 
         if is_peer:
             # For peer relations, both the remote and the local app are the same.


ops/model.py
```diff
@@ -1244,15 +1250,18 @@ class Relation:
...
 
     def __init__(
             self, relation_name: str, relation_id: int, is_peer: bool, our_unit: Unit,
-            backend: '_ModelBackend', cache: '_ModelCache'):
+            backend: '_ModelBackend', cache: '_ModelCache', active: bool = True):
         self.name = relation_name
         self.id = relation_id
         self.app: Optional[Application] = None
         self.units: Set[Unit] = set()
+        self.active = active
```
sed-i, line 1264: Could/should we evaluate it here instead of propagating?

ca-scribner: Sorry I'm not sure what you mean - can you restate this?

PietroPasotti: on the one hand it would greatly reduce the amount of code changes, on the other it is nice that code this 'deep' in ops is profoundly unaware of juju-specific envvars. It'd be nice if the only pieces of code doing envvar unpacking were in ops.main. 

How we propagate that info tastefully and in a future-proof manner, that is the question

ops/model.py
```diff
@@ -107,12 +107,14 @@ class Model:
     from any class that derives from Object.
     """
 
-    def __init__(self, meta: 'ops.charm.CharmMeta', backend: '_ModelBackend'):
+    def __init__(self, meta: 'ops.charm.CharmMeta', backend: '_ModelBackend',
+                 breaking_relation_id: Optional[int] = None):
```
PietroPasotti, line 111: I think it would be more elegant if we attached this information to the event instead of the model.

breaking-relation-id is deceptive as it suggests any relation ID could be that of a relation that is breaking, while in fact the only relation we could possibly know to be breaking is 'the current one', aka 'the one the current event is about'.
So it makes more sense to me to attach that information to the event.

That also spares us to add model.relation.get_active because there's only ever going to be one relation in breaking state: the one we get from the event we're observing.

I propose the following api:

```
def _on_foo_relation_event(self, e:RelationEvent):
    if e.is_dying:
        ...
```

this feels a bit more `ops`y.

PietroPasotti: that said, this would make it more complicated to access the same information from code that doesn't directly have access to the event object. Maybe this was the whole point of this exercise?

ca-scribner: The motivation for me with this issue is a case where, if possible, not being tied to the event would be nice.  I had tried to implement a [nicely encapsulated getter](https://github.com/canonical/kubeflow-dashboard-operator/pull/118/files#diff-5af16278717441c62aff8728f4b371e49c0e8a8e3560fac251562b7a8a6d1258R161) on a relation charm lib, but found since it might be called during `relation-broken` I needed to spot and omit the breaking application.  To do that existing `ops` and without spying on the `JUJU_*` variables, it meant passing the `event` through several layers of code down to the getter.  Not impossible, but it felt awkward at best

ca-scribner: You're right though that this is event-specific behaviour.  But then since `model.relations` exposes the relation information and the event influences what that data means, I think `ops` sort of commits itself to giving access to the full context via `model.relations`

ops/main.py
```diff
@@ -361,6 +361,15 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
         return False
 
 
+def _get_breaking_relation_id() -> Optional[int]:
+    """Returns the relation id of the currently breaking relation, if applicable."""
+    if not os.environ.get("JUJU_HOOK_NAME", "").endswith("-relation-broken"):
```
benhoyt, line 366: We don't actually use `JUJU_HOOK_NAME` in ops anymore, we use the newer `JUJU_DISPATCH_PATH`. However, that's already parsed by the `_Dispatcher` class here, and the event name (with `-` already replaced with `_`) is available as `dispatcher.event_name`. So I suggest we do this in `main()`:

```python
    breaking_relation_id = None
    if dispatcher.event_name.endswith('_relation_broken'):
        breaking_relation_id = _parse_relation_id()
```

ops/main.py
```diff
@@ -361,6 +361,15 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        # Not the relation-broken event
+        return None
+
+    return int(os.environ.get("JUJU_RELATION_ID"))
```
benhoyt, line 370: This is not actually correct, as `JUJU_RELATION_ID` has the format `relation-name:123`. Elsewhere in `main.py` this is parsed using `int(os.environ['JUJU_RELATION_ID'].split(':')[-1])`. So let's pull that one-liner out into `_parse_relation_id() -> int` and use that in both places.

ops/main.py
```diff
@@ -391,8 +400,10 @@ def main(charm_class: Type[ops.charm.CharmBase],
     else:
         actions_metadata = None
 
+    breaking_relation_id = _get_breaking_relation_id()
+
     meta = CharmMeta.from_yaml(metadata, actions_metadata)
-    model = ops.model.Model(meta, model_backend)
+    model = ops.model.Model(meta, model_backend, breaking_relation_id)
```
benhoyt, line 406: I know it's not required, but let's pass in optional arguments using their keywords, to avoid accidental breakage if we add more and change the order. So `Model(meta, model_backend, breaking_relation_id=breaking_relation_id)`.

ops/model.py
benhoyt: Yeah, I agree with Andrew here: we specifically want this info to be available on the `Relation` object, rather then the event.

ops/model.py
```diff
@@ -107,12 +107,14 @@ class Model:
...
         self._cache = _ModelCache(meta, backend)
         self._backend = backend
         self._unit = self.get_unit(self._backend.unit_name)
         relations: Dict[str, 'ops.RelationMeta'] = meta.relations
-        self._relations = RelationMapping(relations, self.unit, self._backend, self._cache)
+        self._relations = RelationMapping(relations, self.unit, self._backend, self._cache,
+                                          breaking_relation_id)
```
benhoyt, line 117: Similar here: let's use `breaking_relation_id=breaking_relation_id` for clarity and to avoid mistakes.

ops/model.py
```diff
@@ -1244,15 +1250,18 @@ class Relation:
             For subordinate relations, this set will include only one unit: the principal unit.
         data: A :class:`RelationData` holding the data buckets for each entity
             of a relation. Accessed via eg Relation.data[unit]['foo']
+        active: Indicates whether this relation is active (True) or inactive (False, because
+            this relation is currently part of a relation-broken event)
```
benhoyt, line 1254: These attributes have now been type-hinted and their docstrings moved to under the type hint so Sphinx picks it up for each attribute, so I believe this will conflict with the `main` branch now. Let's do something like this:

```python
class Relation:
    ...

    active: bool
    """Indicates whether this relation is active.

    This will be ``False`` if the current event is a ``relation-broken`` event
    associated with this relation.
    """
```

ops/model.py
```diff
@@ -1244,15 +1250,18 @@ class Relation:
...
     """
 
     def __init__(
             self, relation_name: str, relation_id: int, is_peer: bool, our_unit: Unit,
-            backend: '_ModelBackend', cache: '_ModelCache'):
+            backend: '_ModelBackend', cache: '_ModelCache', active: bool = True):
```
benhoyt, line 1259: For this one, I'd probably use a required (non-keyword) argument -- that is, don't default to `True`, but require `active` to be provided. My rationale is that we should not forget to specify `active` correctly on all call sites.

That would have caught the fact that we've forgotten here to specify `active=False` in the other call to `Relation()` in `_get_unique`. See the comment "The relation may be dead, but it is not forgotten." I believe that case should always set active to False.

ops/model.py
benhoyt: Yeah, it would be "simpler", but I definitely prefer doing as much env var handling as possible in `main.py`, so it's contained there. It's a bit awkward to pass it down, but I think important for testability and explicitness.

benhoyt (changes requested):
Thanks for this! I've left some comments requesting a few minor changes, but I think the logic is sound and how you're passing the environment variable down explicitly is good.

This will of course need tests, but let's wait on that for a bit -- I'm planning to discuss [this ](https://github.com/canonical/operator/issues/940) with John M tomorrow, about whether we go with this approach or move toward relations not even including breaking/inactive relations at all. So you're welcome to hold off a bit until we decide whether to proceed with this approach.

ca-scribner:
@benhoyt ty for review!  everything makes sense, but I'll hold off till you talk with John in case something changes.

ca-scribner:
Closing as I believe this issue will be addressed elsewhere in juju

PR #960, opened 2023-07-03
Use real filesystem in ops.testing backend

weiiwang01:
This pull request refactors the testing module of the Ops library, switching its backend from an in-memory virtual filesystem to a real filesystem for simulating filesystem operations within containers. Furthermore, it exposes a new API, `Harness.get_filesystem_root`, enabling developers to interact with the simulated filesystem during test cases for both populating and inspecting the testing container filesystem.

Supersede https://github.com/canonical/operator/pull/933

## Changelog

 - Refactored the testing module to use a real filesystem for simulating container filesystem operations.
 - Introduced a new API, `ops.testing.Harness.get_filesystem_root`, for accessing and manipulating the simulated filesystem during testing.


ops/testing.py
```diff
@@ -2425,66 +2537,71 @@ def push(
...
+            elif isinstance(source, bytes):
+                content = source
+            else:
+                source_read: Union[str, bytes] = source.read()
+                content = (
+                    source_read
+                    if isinstance(source_read, bytes)
+                    else source_read.encode(encoding=encoding)
+                )
+            file_path.write_bytes(content)
```
benhoyt, line 2564: In the `str` and `bytes` cases we already have the content in memory, but I think we should avoid reading the whole thing into memory if it's a file/buffer. Can we instead use `copyfileobj`, which only reads in 64K chunks, and do something like this (untested):

```python
if isinstance(source, str):
    file_path.write_text(source, encoding=encoding)
elif isinstance(source, bytes):
    file_path.write_bytes(source)
else:
    with file_path.open('w' if encoding is None else 'wb', encoding=encoding) as f:
        shutil.copyfileobj(source, f)
```

ops/testing.py
```diff
@@ -2425,66 +2537,71 @@ def push(
...
             raise pebble.PathError(
                 'not-found', f'parent directory not found: {e.args[0]}')
-        except NonAbsolutePathError as e:
-            raise pebble.PathError(
-                'generic-file-error',
-                f'paths must be absolute, got {e.args[0]!r}'
-            )
+        except NotADirectoryError:
+            raise pebble.PathError('generic-file-error',
+                                   f'open {path}.{secrets.token_urlsafe(8)}~: not a directory')
```
benhoyt, line 2572: I don't understand the use of `secrets.token_urlsafe` here. Can you explain? Also, when would the `NotADirectoryError` happen here?

ops/testing.py
```diff
@@ -2425,66 +2537,71 @@ def push(
...
             for file in files
         ]
+        for file_info in file_infos:
+            is_root = file_info.path == str(self._root)
+            if is_root:
+                file_info.path = "/"
+                file_info.name = "/"
+            else:
+                file_info.path = file_info.path[len(str(self._root)):]
+        return file_infos
```
benhoyt, line 2604: Would this work?

```suggestion
        for file_info in file_infos:
            rel_path = os.path.relpath(file_info.path, start=self._root)
            rel_path = '/' if rel_path == '.' else '/' + rel_path
        return file_infos
```

ops/testing.py
```diff
@@ -2500,38 +2617,54 @@ def make_dir(
...
+            self.make_dir(
+                os.path.split(path)[0],
+                make_parents=True,
+                permissions=permissions,
+                user_id=user_id,
+                user=user,
+                group_id=group_id,
+                group=group)
+        try:
+            permissions = permissions if permissions else 0o755
```
benhoyt, line 2635: Similar to above, should we just skip the `os.chmod()` if permissions is None?

ops/testing.py
```diff
@@ -2500,38 +2617,54 @@ def make_dir(
...
-                'not-found', f'remove {path}: no such file or directory')
-
-        if isinstance(file_or_dir, _Directory) and len(file_or_dir) > 0 and not recursive:
-            raise pebble.PathError(
-                'generic-file-error', 'cannot remove non-empty directory without recursive=True')
-        self._fs.delete_path(path)
+            else:
+                raise pebble.PathError(
+                    'not-found', f'remove {path}: no such file or directory')
+        if file_path.is_dir() and not recursive:
```
benhoyt, line 2656: As much as I like not nesting, I think this logic would be a bit clearer with an if/else and only one `file_path.is_dir()`:

```python
if file_path.is_dir():
    if recursive:
        shutil.rmtree(file_path)
    else:
        try:
            file_path.rmdir()
        except OSError:
            ...
else:
    file_path.unlink()
```

test/test_testing.py
```diff
@@ -2681,6 +2677,20 @@ def test_container_pebble_ready(self):
...
+            containers:
+              foo:
+                resource: foo-image
+        ''')
+        foo_root = harness.get_filesystem_root("foo")
+        self.assertTrue(foo_root.exists())
+        self.assertTrue(foo_root.is_dir())
+        harness.begin()
+        container = harness.charm.unit.get_container("foo")
+        self.assertEqual(foo_root, harness.get_filesystem_root(container))
```
benhoyt, line 2692: This is a good test as far as it goes, but I think we need a few more test cases to test that 1) files that the test writes via `get_filesystem_root` can then be pulled and listed, and 2) files pushed and make_dir'd can then be accessed via `get_filesystem_root`. Or am I missing something?

.github/workflows/framework-tests.yaml
```diff
@@ -38,6 +38,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: ["3.8", "3.10"]
+        root: [true, false]
```
benhoyt, line 41: I don't love how this shows up in the test summary in GitHub, e.g., "ubuntu-latest, 3.10, false" -- it's unclear what "false" means. Also, I'd rather not run all the tests twice, and I don't think we need to or should be running all the tests as root anyway.

Can we use a suitable test name prefix, and then do something like we do with the RealPebble tests below? Perhaps a TestFilesystemAsRoot class and then use `-k FilesystemAsRoot` filter. I think having a separate job is cleaner (even though it's a bit more duplication).

benhoyt (changes requested):
Thanks for your continued effort here. This is really good work. I've left a bunch of comments. Mostly minor, but one discussion point about whether we should do the user/group stuff at all. I'll also try to schedule a meeting tomorrow to finish discussion on this.

ops/testing.py
weiiwang01: In that case the test result will affected by the umask in developer's environment, which I would like to avoid.

test/test_testing.py
weiiwang01: Yes, we already have a suite of tests for the filesystem, using the Pebble APIs (blackbox tests). I think I will add some more tests related to the `get_filesystem_root` method, since that's an exposed API now. Thanks!

ops/testing.py
```diff
@@ -2402,16 +2461,69 @@ def get_services(self, names: Optional[List[str]] = None) -> List[pebble.Service
...
+        group_id = group_id if group_id is not None else group_id_by_group
+        if user_id is not None and group_id is None:
+            raise pebble.PathError(
+                'generic-file-error',
+                'cannot look up user and group: must specify group, not just UID')
+        if user_id is None and group_id is not None:
+            raise pebble.PathError(
+                'generic-file-error',
+                'cannot look up user and group: must specify user, not just group')
+        final_uid = user_id if user_id is not None else os.getuid()
```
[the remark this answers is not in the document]
weiiwang01: That actually complicates the situation since `uid` and `gid` are both required by `os.chown` and they must be integer, not `None`.

ops/testing.py
```diff
@@ -2425,66 +2537,71 @@ def push(
             raise pebble.PathError(
                 'generic-file-error',
                 f'permissions not within 0o000 to 0o777: {permissions:#o}')
+        self.__check_absolute_path(path)
+        file_path = self._root / path[1:]
+        if make_dirs and not file_path.parent.exists():
+            self.make_dir(
+                os.path.split(path)[0],
```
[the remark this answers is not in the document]
weiiwang01: Indeed, it's better. Fixed, thanks!

ops/testing.py
```diff
@@ -2402,16 +2461,69 @@ def get_services(self, names: Optional[List[str]] = None) -> List[pebble.Service
             infos.append(info)
         return infos
 
+    @staticmethod
+    def __check_absolute_path(path: str):
```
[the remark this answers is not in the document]
weiiwang01: Removed, thanks!

ops/testing.py
```diff
@@ -1918,7 +1975,9 @@ def get_pebble(self, socket_path: str) -> '_TestingPebbleClient':
         container = socket_path.split('/')[3]  # /charm/containers/<container_name>/pebble.socket
         client = self._pebble_clients.get(container, None)
         if client is None:
-            client = _TestingPebbleClient(self)
+            container_root = self._harness_container_path / container
+            container_root.mkdir()
+            client = _TestingPebbleClient(self, container_root=container_root)
             self._pebble_clients[container] = client
```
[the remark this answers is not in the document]
weiiwang01: It does look like a duplicate, removed.

ops/testing.py
```diff
@@ -1838,8 +1887,12 @@ def _storage_detach(self, storage_id: str):
         name, index = storage_id.split('/', 1)
         index = int(index)
 
-        for client in self._pebble_clients.values():
-            client._fs.remove_mount(name)
+        for container, client in self._pebble_clients.items():
+            for _, mount in self._meta.containers[container].mounts.items():
+                if mount.storage != name:
+                    continue
+                root = client._root  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
weiiwang01: Removed, thanks!

ops/testing.py
weiiwang01: Yes, updated!

ops/testing.py
weiiwang01: When pebble transferring a file, it will first write to a temporary path (filename + random string + ~), and rename to the final destination when the transferring is finished. This is just mimicking the pebble error, while  `secrets.token_urlsafe` is just a quick way to generate a random string.

This error will happen if the parent directory doesn't exist and `make_dirs` is set to `False`.

ops/testing.py
weiiwang01: It's better, updated, thanks!

ops/testing.py
```diff
@@ -2425,66 +2537,71 @@ def push(
...
+        if make_dirs and not file_path.parent.exists():
+            self.make_dir(
+                os.path.split(path)[0],
+                make_parents=True,
+                permissions=None,
+                user_id=user_id,
+                user=user,
+                group_id=group_id,
+                group=group)
+        permissions = permissions if permissions is not None else 0o644
```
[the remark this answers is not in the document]
weiiwang01: Answered down below, thanks!

weiiwang01:
> Thanks for your continued effort here. This is really good work. I've left a bunch of comments. Mostly minor, but one discussion point about whether we should do the user/group stuff at all. I'll also try to schedule a meeting tomorrow to finish discussion on this.

Thanks for the fantastic review! Yes, let's talk about the permission issue.

Another topic I want to discuss is about simulating storage mounts. Currently, we are using symbolic links to simulate storage mounts. However, actual storage mounts have some special properties that symbolic links can't simulate. For example, when a filesystem is attached to a directory, the original content of that directory becomes "shadowed". Additionally, it's an error to delete the mounting directory. Also, `list_files` should display the mounting directory as a directory, but right now, it shows as a symbolic link.

Of course, we can simulate these behaviors in the harness, but I feel like this introduces dishonesty into the design. I would prefer to let the user know that we are using symbolic links to simulate storage mounts and it's the users' responsibility to avoid tests involves unique properties of mounting directories.

.github/workflows/framework-tests.yaml
weiiwang01: I think the tests depend on the outcome of the discussion, I will defer the changes for now.

benhoyt:
Per yesterday's discussion, @weiiwang01's going to simplify by removing the user/group handling, and (for now at least) document where the test harness isn't a perfect match. As a consequence, we can always remove the tests running as root (we don't want to run tests as root or encourage charmers to do that). He's also going to remove the `secrets.token_urlsafe` call and just leave that out -- people shouldn't be testing exact contents of error messages anyway.

ops/testing.py
```diff
@@ -552,6 +551,10 @@ def add_storage(self, storage_name: str, count: int = 1,
         location set to /[tmpdir]/<storage_name>N, where N is the counter and
         will be a number from [0,total_num_disks-1].
 
+        The test harness uses symbolic links to imitate storage mounts, which may lead to some
+        inconsistencies compared to the actual charm. Users should be cognizant of
+        this potential discrepancy.
```
benhoyt, line 556: IMO we should drop the sentence "Users should be cognizant of..." (here and below). We include the first sentence exactly so users are cognizant of it.

ops/testing.py
```diff
@@ -1429,6 +1436,53 @@ def trigger_secret_expiration(self, secret_id: str, revision: int, *,
...
+        >>>     def __init__(self, *args):
+        >>>         super().__init__(*args)
+        >>>         self.hostname = open("/etc/hostname").read()
+
+        >>> # test_charm.py
+        >>> from ops.testing import Harness
+        >>> harness = Harness(ExampleCharm)
+        >>> root = harness.get_filesystem_root("mycontainer")
+        >>> (root / "etc" / "hostname").write_text("hostname.example.com")
+        >>> harness.begin()
```
benhoyt, line 1472: I'd prefer to avoid the `>>> ` prefix on each line, as this isn't typically something you'd type into the Python REPL. To do this, use a double-colon on `Example usage::` and then make the follow code block indented by four spaces:

```suggestion
        Example usage::

            # charm.py
            import ops
            class ExampleCharm(ops.CharmBase):
                def __init__(self, *args):
                    super().__init__(*args)
                    self.hostname = open("/etc/hostname").read()

            # test_charm.py
            from ops.testing import Harness
            harness = Harness(ExampleCharm)
            root = harness.get_filesystem_root("mycontainer")
            (root / "etc" / "hostname").write_text("hostname.example.com")
            harness.begin()
```

ops/testing.py
```diff
@@ -2434,66 +2513,63 @@ def push(
...
-            )
+        file_infos = [
+            Container._build_fileinfo(file)
             for file in files
         ]
+        for file_info in file_infos:
+            rel_path = os.path.relpath(file_info.path, start=self._root)
+            rel_path = '/' if rel_path == '.' else '/' + rel_path
+            file_info.path = rel_path
+            file_info.name = "/" if rel_path == "/" else os.path.basename(rel_path)
```
benhoyt, line 2571: Why isn't `file_info.name` correct already?

ops/testing.py
```diff
@@ -2509,38 +2585,54 @@ def make_dir(
...
-                # Pebble doesn't give not-found error when recursive is specified
                 return
-            raise pebble.PathError(
-                'not-found', f'remove {path}: no such file or directory')
-
-        if isinstance(file_or_dir, _Directory) and len(file_or_dir) > 0 and not recursive:
-            raise pebble.PathError(
-                'generic-file-error', 'cannot remove non-empty directory without recursive=True')
-        self._fs.delete_path(path)
+            else:
```
benhoyt, line 2620: Given there's an early return above, I prefer to simplify and avoid the else (similar to how the structure was before).

ops/testing.py
```diff
@@ -2509,38 +2585,54 @@ def make_dir(
...
+                raise pebble.PathError(
+                    'not-found', f'remove {path}: no such file or directory')
+        if file_path.is_dir():
+            if recursive:
+                shutil.rmtree(file_path)
+            else:
+                try:
+                    file_path.rmdir()
+                    return
```
benhoyt, line 2629: No need for this `return`. Each of the branches performs an operation, then the method implicitly returns at the end.

test/test_testing.py
```diff
@@ -4463,6 +4333,102 @@ def test_make_dir_with_permission_mask(self):
...
+        self.assertTrue((self.root / "foo/bar/foobar").is_dir())
+
+    def test_pull(self):
+        (self.root / "foo").write_text("foo")
+        self.assertEqual(self.container.pull("/foo").read(), "foo")
+
+    def test_pull_path(self):
+        (self.root / "foo").mkdir()
+        (self.root / "foo/bar").write_text("bar")
+        # TODO: pull_path doesn't pull empty directories, intended?
```
benhoyt, line 4394: Yeah, if that's the case, that seems like a bug. Can you please open a separate canonical/operator issue for that?

benhoyt (approved):
I think this is great. Just a couple of nit comments. And also get a second review (probably John's as we've discussed this with him).

weiiwang01:
> Per yesterday's discussion, @weiiwang01's going to simplify by removing the user/group handling, and (for now at least) document where the test harness isn't a perfect match. As a consequence, we can always remove the tests running as root (we don't want to run tests as root or encourage charmers to do that). He's also going to remove the `secrets.token_urlsafe` call and just leave that out -- people shouldn't be testing exact contents of error messages anyway.

I've eliminated the ownership handling and restructured the code according to the recommendation. However, while writing the tests, I came across an problem: the `pull_path` method within `ops.model.Container` doesn't retrieve empty directories. I'm curious to know if this is by design, similar to how git operates.

benhoyt:
Hi @weiiwang01 Thanks -- yep, I replied to your comments above. I've opened an issue on the `pull_path` empty directory issue: https://github.com/canonical/operator/issues/968 -- also see the other comments above.

ops/testing.py
```diff
@@ -1420,6 +1422,49 @@ def trigger_secret_expiration(self, secret_id: str, revision: int, *,
...
+        Args:
+            container: The simple name of the container or the container object.
+
+        Return:
+            The temporary directory associated with the container is used by the testing harness to
+            simulate the container's filesystem.
+
+        """
+        if isinstance(container, str):
+            container_name = container
```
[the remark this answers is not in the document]
weiiwang01: Added a comment here as suggested.

test/test_testing.py
weiiwang01: I have added more unit tests for the `get_filesystem_root` method, thanks!

ops/testing.py
weiiwang01: Removed, thanks!

ops/testing.py
weiiwang01: Removed, thanks!

ops/testing.py
jameinel: if it is only for the root, why set it always, vs just:
```
if rel_path == "/":
  // override the root name, since it should be considered the local root, without any filename
  file_info.name = "/"
  ```
 
It seems clearer than always setting it to the value it should already have. (and avoids computing basename() for every file in every directory if you are iterating over them)

**Review by jameinel — APPROVED** [review]

File: ops/testing.py [hunk 21]
```diff
@@ -1601,6 +1656,10 @@ def __init__(self, unit_name: str, meta: charm.CharmMeta, config: 'RawConfig'):
         self.model_uuid = str(uuid.uuid4())
 
         self._harness_tmp_dir = tempfile.TemporaryDirectory(prefix='ops-harness-')
+        self._harness_storage_path = pathlib.Path(self._harness_tmp_dir.name) / "storages"
+        self._harness_container_path = pathlib.Path(self._harness_tmp_dir.name) / "containers"
+        self._harness_storage_path.mkdir()
+        self._harness_container_path.mkdir()
```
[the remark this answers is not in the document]
benhoyt: Yeah, I thought the same too. But then I did a timing test, and on my laptop I can `os.mkdir()` 10,000 directories in 0.1s and 100,000 in 1.6s (and a bit less than that for `os.rmdir()`). So 3 dir creates + deletes adds about 50 microseconds to each test. Linux FS operations are quick!

ops/testing.py
benhoyt: John's suggestion looks clearer to me too.

benhoyt:
Thanks. Let's just make that `file_info.name = "/"` tweak that John mentioned, and then I'll see about merging this!

ops/testing.py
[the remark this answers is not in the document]
weiiwang01: We can prefer `/dev/shm` as the `tempdir` to expedite the filesystem calls. But as the benchmark shows, it doesn't really have a significant influence on the overall performances.

benhoyt:
Yeeha! Thanks for this @weiiwang01. This will be included with the next version of ops, 2.5.0, planned for release late July / early August.

PR #961, opened 2023-07-04
Add "-> None" to methods without other types

benhoyt:
This is so that if MyPy is used in a charm, it type checks them. MyPy by default doesn't type check methods that have no type annotations on them. For example:

```
$ tox -e mypy
mypy: commands[0]> mypy
src/charm.py:49: error: Call to untyped function "remove_all_revisions" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```

I forgot to do this in #953.

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -110,7 +110,7 @@ class ActionEvent(EventBase):
     params: Dict[str, Any]
     """The parameters passed to the action."""
 
-    def defer(self):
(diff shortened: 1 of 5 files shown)


benhoyt:
Just a couple of type annotation tweaks forgotten in #953 -- going to merge without further review, so I can go ahead with the release.

PR #962, opened 2023-07-04
Type annotate PebbleReadyEvent.workload

benhoyt:
One more thing forgotten from #953. This is annoying because it means that charms that use `ops.PebbleReadyEvent` annotations when handling `pebble-ready` can't access any attributes of the `workload`, as it doesn't have a type.


benhoyt:
Minor tweak forgotten from #953, and gets in the way of https://github.com/canonical/charmcraft/pull/1137, so going to merge this so I can ship an `ops` patch release 2.4.1.

PR #964, opened 2023-07-05
Vastly simplify PR template

benhoyt:
I find that PR templates/checklists do more harm than good. Half of the stuff is irrelevant to most PRs, so people tend to start ignoring the checklist anyway. And there are a ton of things that aren't in the checklist that should be checked -- but the list would be very unwieldy if we tried to be exhaustive.

I'd rather people just wrote good, English descriptions, and edit them where they need improvement.

So this PR simplifies it to the point of basically removing it. :-)


benhoyt:
Any strong objections, @jnsgruk?

jnsgruk (approved):

PR #965, opened 2023-07-05
Add `has_open_port_k8s` to `JujuVersion`

carlcsaposs-canonical:
open-port is not supported before Juju 3.0.3 on Kubernetes

Context: https://chat.charmhub.io/charmhub/pl/91t6ip37ybncuqe11eudfx38xw

### DIFF
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -126,6 +126,6 @@ class JujuVersion:
 
     @property
     def has_open_port_k8s(self) -> bool:
-        """Report whether this Juju version supports open-port on Kubernetes"""
+        """Report whether this Juju version supports open-port on Kubernetes."""
         # Support added: https://bugs.launchpad.net/juju/+bug/1920960
         return (self.major, self.minor, self.patch) >= (3, 0, 3)
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -123,3 +123,9 @@ class JujuVersion:
         # * In 3.0.3, a bug with observer labels was fixed (juju/juju#14916)
         # TODO(benhoyt): update to 3.0.3+ once shipped (for juju/juju#14916 fix)
         return (self.major, self.minor, self.patch) >= (3, 0, 2)
+
+    @property
+    def has_open_port_k8s(self) -> bool:
+        """Report whether this Juju version supports open-port on Kubernetes"""
+        # Support added: https://bugs.launchpad.net/juju/+bug/1920960
+        return (self.major, self.minor, self.patch) >= (3, 0, 3)
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -125,7 +125,7 @@ class JujuVersion:
         return (self.major, self.minor, self.patch) >= (3, 0, 2)
 
     @property
-    def has_open_port_k8s(self) -> bool:
+    def supports_open_port_on_k8s(self) -> bool:
         """Report whether this Juju version supports open-port on Kubernetes."""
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Thanks!

It's annoying we have some methods and some properties on this object -- but I think that was my fault when we introduced `has_secrets` as a property (and all the others were methods). Now that we have both, I think properties is the simpler way forward.

ops/jujuversion.py
```diff
@@ -123,3 +123,9 @@ def has_secrets(self) -> bool:
         # * In 3.0.3, a bug with observer labels was fixed (juju/juju#14916)
         # TODO(benhoyt): update to 3.0.3+ once shipped (for juju/juju#14916 fix)
         return (self.major, self.minor, self.patch) >= (3, 0, 2)
```
jameinel, line 125: I'm guessing you want to set this to 3,0,3 given the associated comment. (Not necessarily in this PR, but looks like a TODO that we forgot to follow up on)

ops/jujuversion.py
```diff
@@ -123,3 +123,9 @@ def has_secrets(self) -> bool:
...
+
+    @property
+    def has_open_port_k8s(self) -> bool:
+        """Report whether this Juju version supports open-port on Kubernetes."""
+        # Support added: https://bugs.launchpad.net/juju/+bug/1920960
+        return (self.major, self.minor, self.patch) >= (3, 0, 3)
```
jameinel, line 131: I'm not a huge fan of properties, but these do feel like static information, so I'm more ok with it than most.

That said, the name "has_open_port_k8s" bothers me a bit.  I don't know if I'm missing a word `has_open_port_on_k8s` or `has_open_port_for_k8s`
Or whether I would rather it was `has_open_port` and it figured out if you were on K8s.
Ultimately, I'm happy to have this sort of version check capability, and I'm bikeshedding about the name, but I'd like us to at least think about it a bit.

jameinel (approved):
Approved, but I'd like us to think about the naming a bit.

ops/jujuversion.py
benhoyt: Fixed in https://github.com/canonical/operator/pull/967. Is it kosher to remove the fallback now that 3.0.3 has been out for a long time? `get_secret` will fail if some people are still running 3.0.2, but that seems unlikely, and if it happens, we can just tell people "please upgrade to 3.0.3 or a later 3.x version". Do you know what we normally do in cases like this?

ops/jujuversion.py
benhoyt: Yeah, fair pushback. I'd rather it not magically figure out whether you were on K8s or not, because `JujuVersion` is a dumb class that just operates on the Juju version, and to take other input (charm metadata or env vars or whatever) would seem against the spirit of what we have here.

However, I've renamed for clarity to include the "on" and renamed "has" to "supports" (because Juju <3.0.3 might "have" open-port on all platforms, it just wouldn't *support* it on K8s). It's now `supports_open_port_on_k8s`, so a bit wordy, but this shouldn't be used often, and clear is better than terse for this.

@carlcsaposs-canonical You happy with `supports_open_port_on_k8s`?

jameinel: LGTM

jameinel (approved):

ops/jujuversion.py
benhoyt: > I would mark it clearly in the release notes associated with that version of python-libjuju.

Thanks. I presume you mean "that version of *ops*"?

benhoyt:
Merged. Thanks @carlcsaposs-canonical!

Issue #888, opened 2023-01-12
Should relation data be visible to charm code from within a `relation-broken` hook?


sed-i:
Currently, harness first emits `relation-broken` and only then invalidates the data.

This means that relation data is still accessible by charm code while inside the `relation-broken` hook. Is that intentional?

Based on the [event page](https://juju.is/docs/sdk/relation-name-relation-broken-event) it sounds like charm code shouldn't be able to see any data when inside the `broken` hook.

https://github.com/canonical/operator/blob/4d38ef7ce2fe6dfc4034db5957d9e526ce0cf3d9/ops/testing.py#L697-L702

benhoyt:
Juju actually allows the charm to access (stale?) relation data in `relation-broken`. However...

I've just tested this using two charms related to one another, each having a relation-broken hook and accessing relation data (from "this" or the "remote" app). The Harness actually gets this fairly different from real Juju in a number of cases.

Below are the results. The "setter" charm is the charm in the relation that's setting relation data via `event.relation.data[self.app]["key"] = value`, and the "getter" charm is the charm that's reading relation data.

| charm  | which app | Juju result    | Harness result            | 
| ------ | --------- | -------------- | ------------------------- |
| setter | this      | `{...data...}` | `{...data...}`            |
| setter | remote    | `{}`           | `RuntimeError`            |
| getter | this      | `{}`           | `RelationDataAccessError` |
| getter | remote    | `{...data...}` | `RuntimeError`            |

* Both `RuntimeError`s have the message "remote-side relation data cannot be accessed during a relation-broken event".
* The `RelationDataAccessError` message is "unit/0 is not leader and cannot read its own application databag", and in `__repr__` this gets caught and converted to the string `<n/a>`.


<details>
<summary>Show raw data from test</summary>

To test this (unit tests and real Juju), I hacked my database and webapp charms that I've been using for secrets to add logging into the relation-broken hooks. The database charm is the "setter" and it sets relation data, and the webapp charm is the "getter" as it reads relation data. `self_data` means data accessed via `event.relation.data[self.app]` and `event_data` means remote data accessed via `event.relation.data[event.app]`.

```
# real Juju

unit-database-0: 15:25:26 INFO unit.database/0.juju-log db:0: database _on_db_relation_broken: <ops.model.Relation db:0> self_data={'db_password_id': 'secret:cf0c1lrs26oc7aah2260'}
[...]

PR #971, opened 2023-07-12
Disallow deferral of relation-broken events (doesn't make sense)

benhoyt:
Deferring relation-broken events doesn't make sense, as the relation has gone away.

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -548,6 +548,10 @@ class RelationBrokenEvent(RelationEvent):
     are currently known locally.
     """
 
+    def defer(self) -> None:
+        """Relation-broken events are not deferrable."""
+        raise RuntimeError('Cannot defer relation-broken events')
+
 
 class StorageEvent(HookEvent):
     """Base class representing storage-related events.


jameinel (commented):
Needs a test, I think.

And while you're here, do you also want to do the same for stop and remove? They also don't make sense to defer.

sed-i:
Seems like this PR [depends on fixing 888](https://github.com/canonical/operator/issues/888#issuecomment-1573857537):
Deferring relation-broken is a (hacky) method for having stale data gone.

ops/charm.py
```diff
@@ -548,6 +548,10 @@ class RelationBrokenEvent(RelationEvent):
     are currently known locally.
     """
 
+    def defer(self) -> None:
+        """Relation-broken events are not deferrable."""
+        raise RuntimeError('Cannot defer relation-broken events')
```
sed-i, line 553: Would it be feasible to use Protocol for this so that calling `defer` becomes a static error?

benhoyt:
In light of the backwards-compatibility concerns above, I'm going to close this as won't do and consider defer holistically instead.

PR #973, opened 2023-07-13
Further separate between hook tools and model

sed-i:
In essence, hook tools are independent of ops.
This PR improves separation between the model and the hook tools.

This gains us:
- Clearer separation of the stack's components.
- More freedom in growing the hook tools API without bloating core model code.

### DIFF
--- a/ops/hook_tools.py
+++ b/ops/hook_tools.py
@@ -1,0 +1,540 @@
+# Copyright 2023 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Wrappers for juju hook tools."""
+
+import datetime
+import json
+import math
+import re
+import shutil
+import subprocess
+import tempfile
+import typing
+from pathlib import Path
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Literal,
+    Mapping,
+    MutableMapping,
+    Optional,
+    Set,
+    Union,
+)
+
+from ops._private import yaml
+
+if typing.TYPE_CHECKING:
+    from typing_extensions import TypedDict
+
+    _ConfigOption = TypedDict('_ConfigOption', {
+        'type': Literal['string', 'int', 'float', 'boolean'],
+        'description': str,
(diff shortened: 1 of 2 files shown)


sed-i:
CI currently fails, but I first wanted to see first if this is seen as a good idea.

benhoyt:
It's an interesting idea, and I'm not totally against it. However, it does seem like a solution looking for a problem. Is there a particular problem you've struggled with working on the `_ModelBackend` that was hard to do or understand without this kind of factorisation?

In my mind the `_ModelBackend` is quite straight-forward already, and is basically all hook tools. So the extra layer of indirection doesn't do much for me, except adds a couple hundred lines of exception-handling boilerplate. :-)

benhoyt:
Adding to that, and without thinking about it much further, I guess I'd more be in favour of just plain moving `_ModelBackend` to `modelbackend.py`. Thoughts on that approach?

Your approach seems more about conceptual purity than a useful split -- but I could well be missing something!

sed-i:
So I forgot to mention that I did not intend/expect this PR to merge!

> _ModelBackend is quite straight-forward already, and is basically all hook tools

It was interesting to see the few subtle dependencies `_ModelBackend` has on ops.
I really like how you can `import ops.pebble` and off you go.
I was hoping we could have the same with hook tools.

> moving _ModelBackend to modelbackend.py

Sounds great!

> exception-handling boilerplate

Yes, maintaining backwards compatibility with `ModelError` is partially the reason. I guess I could have just move it into `hook_tools` and then import again (reexport) in `ops.model`.

> Is there a particular problem you've struggled with working on the _ModelBackend that was hard to do or understand without this kind of factorisation?

I was playing with the idea of ops charms without using ops.main/observe. It's not necessary, but helps for the mental model.

benhoyt:
I think I'd be more okay with this if we could avoid the try/catch wrapping, for example having a `modelerror.py` with `ModelError` and its subclasses in it, and then `hooktools.py` that uses that (no `HookError` wrapping).

However, I'm still not sure it's an idea that pays its way, because:

1) It would basically mean committing to `ops.hooktools` as a public API. Or we could make it `_hooktools` to signal you shouldn't really be using this and we won't guarantee stability. But then does it help you wanting to use it separately?
2) It's still quite a lot of boilerplate wrapping, such as `def relation_ids(self, relation_name: str) -> List[int]:` \ `    return hooktools.relation_ids(relation_name)`. This makes exploration and maintenance of the framework just a tiny bit harder IMO.

So my suggestion is you just use `ops.model._ModelBackend` directly. You're explicitly relying on non-public APIs, but I don't think I want to make those APIs public and commit to them forever anyway.

We could move `ModelError` and `_ModelBackend` to their own file `modelbackend.py`. That would be an easier sell to me. But then again, that's mainly for code maintenance and reducing the size of `model.py`; it still doesn't help any with your use case.

benhoyt:
Per above, I think this is an interesting idea, but I don't think it pays its way -- closing. I think if we did this we'd probably need to go over the external API with a fine-toothed comb before releasing, and perhaps even version it as a separate library.

Issue #969, opened 2023-07-11
Support unified charmcraft.yaml


syu-w:
Hi,

`charmcraft` is now moving to use unified `charmcraft.yaml` for all `metadata.yaml`, `actions.yaml`, and `config.yaml` as specs in ST087. While the old style is still supported and we doing backward compatibility for now, the unit tests for the charms are not happy because it try to read `metadata.yaml` but did not found the metadata it needs.

So this might be a good time for ops to first try to read `charmcraft.yaml` to get `metadata`.

An example of the new unified `charmcraft.yaml`
https://github.com/canonical/charmcraft/blob/main/tests/test_metafiles.py#L29

benhoyt:
Thanks for the note @syu-w. If you or one of your team is able to help with a PR here, that would speed things up. Otherwise I'll try to get to this before the next release of ops (likely early August).

syu-w:
@benhoyt I am going to see if I can create a PR next week.

benhoyt:
I'm reviewing #977 now, but just wondering about the spec for this. [ST087](https://docs.google.com/document/d/1IctKhAv2H_HUk96g0p2QGx_CAYVLGIQH_CWydLIkzFE/edit) doesn't seem to mention `actions.yaml` and `config.yaml` at all. It looks like the things in `actions.yaml` are included under an `actions` key in `charmcraft.yaml`, and the `options` from `config.yaml` are included directly. Is that correct? Is there a spec / doc for that somewhere?

jnsgruk:
@benhoyt the intention is this:

- You can still include a `metadata.yaml`, but fields that appear in both `charmcraft.yaml` and `metadata.yaml` will cause a build error
- You can still include a `config.yaml` seperately, **or** you can add `options:` in the charmcraft.yaml. This allows for splitting things out in charms with a lot of config options
- You can still include `actions.yaml` **or** you can add `actions:` to your `charmcraft.yaml`

benhoyt:
Thanks for the PR, @syu-w -- merged!

PR #977, opened 2023-07-18
Support reading charmcraft.yaml in testing harness

syu-w:
Try read unified charmcraft.yaml first, that may included metadata, actions, and config. Ignore metadata.yaml, actions.yaml, config.yaml if they are exists in charmcraft.yaml

Fixes https://github.com/canonical/operator/issues/969

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -389,57 +389,102 @@ class Harness(Generic[CharmType]):
         """
         self._backend._cleanup()
 
-    def _create_meta(self, charm_metadata: Optional[YAMLStringOrFile],
-                     action_metadata: Optional[YAMLStringOrFile]) -> CharmMeta:
+    def _create_meta(self, charm_metadata_yaml: Optional[YAMLStringOrFile],
+                     action_metadata_yaml: Optional[YAMLStringOrFile]) -> CharmMeta:
         """Create a CharmMeta object.
 
         Handle the cases where a user doesn't supply explicit metadata snippets.
+        This will try to load metadata from ``<charm_dir>/charmcraft.yaml`` first, then
+        ``<charm_dir>/metadata.yaml`` if charmcraft.yaml does not include metadata,
+        and ``<charm_dir>/actions.yaml`` if charmcraft.yaml does not include actions.
         """
         filename = inspect.getfile(self._charm_cls)
         charm_dir = pathlib.Path(filename).parents[1]
 
+        charm_metadata: Optional[Dict[str, Any]] = None
+        charmcraft_metadata: Optional[Dict[str, Any]] = None
+        # Check charmcraft.yaml and load it if it exists
+        charmcraft_meta = charm_dir / "charmcraft.yaml"
+        if charmcraft_meta.is_file():
+            self._charm_dir = charm_dir
+            charmcraft_metadata = yaml.safe_load(charmcraft_meta.read_text())
+
+        # Load metadata from parameters if provided
+        if charm_metadata_yaml is not None:
+            if isinstance(charm_metadata_yaml, str):
+                charm_metadata_yaml = dedent(charm_metadata_yaml)
+            charm_metadata = yaml.safe_load(charm_metadata_yaml)
+        else:
+            # Check charmcraft.yaml for metadata if no metadata is provided
+            if charmcraft_metadata is not None:
+                meta_keys = ["name", "summary", "description"]
+                if any(key in charmcraft_metadata for key in meta_keys):
+                    # Unrelated keys in the charmcraft.yaml file will be ignored.
+                    charm_metadata = charmcraft_metadata
+
+            # Still no metadata, check metadata.yaml
+            if charm_metadata is None:
+                metadata_path = charm_dir / 'metadata.yaml'
+                if metadata_path.is_file():
+                    charm_metadata = yaml.safe_load(metadata_path.read_text())
+                    self._charm_dir = charm_dir
+
+        # Use default metadata if metadata is not found
         if charm_metadata is None:
-            metadata_path = charm_dir / 'metadata.yaml'
-            if metadata_path.is_file():
-                charm_metadata = metadata_path.read_text()
-                self._charm_dir = charm_dir
-            else:
-                # The simplest of metadata that the framework can support
-                charm_metadata = 'name: test-charm'
-        elif isinstance(charm_metadata, str):
-            charm_metadata = dedent(charm_metadata)
+            charm_metadata = {"name": "test-charm"}
 
-        if action_metadata is None:
-            actions_path = charm_dir / 'actions.yaml'
-            if actions_path.is_file():
-                action_metadata = actions_path.read_text()
-                self._charm_dir = charm_dir
-        elif isinstance(action_metadata, str):
-            action_metadata = dedent(action_metadata)
+        action_metadata: Optional[Dict[str, Any]] = None
+        # Load actions from parameters if provided
+        if action_metadata_yaml is not None:
+            if isinstance(action_metadata_yaml, str):
+                action_metadata_yaml = dedent(action_metadata_yaml)
+            action_metadata = yaml.safe_load(action_metadata_yaml)
+        else:
+            # Check charmcraft.yaml for actions if no actions are provided
+            if charmcraft_metadata is not None and "actions" in charmcraft_metadata:
+                action_metadata = charmcraft_metadata["actions"]
 
-        return CharmMeta.from_yaml(charm_metadata, action_metadata)
+            # Still no actions, check actions.yaml
+            if action_metadata is None:
+                actions_path = charm_dir / 'actions.yaml'
+                if actions_path.is_file():
+                    action_metadata = yaml.safe_load(actions_path.read_text())
+                    self._charm_dir = charm_dir
 
-    def _get_config(self, charm_config: Optional['YAMLStringOrFile']):
+        return CharmMeta(charm_metadata, action_metadata)
+
+    def _get_config(self, charm_config_yaml: Optional['YAMLStringOrFile']):
         """If the user passed a config to Harness, use it.
 
-        Otherwise, attempt to load one from charm_dir/config.yaml.
+        Otherwise try to load config from ``<charm_dir>/charmcraft.yaml`` first, then
+        ``<charm_dir>/config.yaml`` if charmcraft.yaml does not include config.
         """
         filename = inspect.getfile(self._charm_cls)
         charm_dir = pathlib.Path(filename).parents[1]
+        config: Optional[Dict[str, Any]] = None
 
-        if charm_config is None:
-            config_path = charm_dir / 'config.yaml'
-            if config_path.is_file():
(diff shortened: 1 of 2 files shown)


ops/main.py
```diff
@@ -384,12 +387,28 @@ def main(charm_class: Type[ops.charm.CharmBase],
...
+        ):
+            # Unrelated keys in the charmcraft.yaml file will be ignored.
+            metadata = charmcraft_yaml
+
+        # This could be None if the charmcraft.yaml not include the actions key.
+        # Should fall back to the actions.yaml file.
+        actions_metadata = charmcraft_metadata.get("actions")
+
+    if not metadata:
+        metadata = (charm_dir / "metadata.yaml").read_text()
```
benhoyt, line 406: In charmcraft's behaviour, is there any merging? Does charmcraft raise an error if the charm has metadata in both `charmcraft.yaml` and `metadata.yaml`? Should we raise an error too, to catch it early?

ops/main.py
```diff
@@ -384,12 +387,28 @@ def main(charm_class: Type[ops.charm.CharmBase],
...
-        actions_metadata = actions_meta.read_text()
-    else:
-        actions_metadata = None
+    # Check charmcraft.yaml first and see if it has the metadata data in it.
+    charmcraft_meta = charm_dir / "charmcraft.yaml"
+    if charmcraft_meta.exists():
+        charmcraft_yaml = charmcraft_meta.read_text()
+        charmcraft_metadata: dict[str, Any] = yaml.safe_load(charmcraft_yaml)
+        if any(
+            meta_key in charmcraft_metadata for meta_key in ("name", "summary", "description")
```
benhoyt, line 396: Is this the behaviour of / test used by charmcraft too? Can you please point to the code there?

ops/testing.py
```diff
@@ -394,9 +394,30 @@ def _create_meta(self, charm_metadata: Optional[YAMLStringOrFile],
         """Create a CharmMeta object.
 
         Handle the cases where a user doesn't supply explicit metadata snippets.
+        Will try to load metadata/config from charm_dir/charmcraft.yaml first, then
+        charm_dir/metadata.yaml and/or charm_dir/actions.yaml if charmcraft.yaml
+        not include metadata and/or actions.
```
benhoyt, line 399: It looks like this method doesn't deal with config. Also a few minor tweaks for English and clarity. Is this correct?

```suggestion
        This will try to load metadata from ``<charm_dir>/charmcraft.yaml`` first, then
        ``<charm_dir>/metadata.yaml`` if charmcraft.yaml does not include metadata,
        and ``<charm_dir>/actions.yaml`` if charmcraft.yaml does not include actions.
```

ops/testing.py
```diff
@@ -394,9 +394,30 @@ def _create_meta(self, charm_metadata: Optional[YAMLStringOrFile],
...
+        if charmcraft_meta.exists():
+            self._charm_dir = charm_dir
+            charmcraft_yaml = charmcraft_meta.read_text()
+            charmcraft_metadata = yaml.safe_load(charmcraft_yaml)
+
+        # Add metadata from charmcraft.yaml if no metadata is provided
+        # Should fall back to metadata.yaml if both are not provided
+        if charm_metadata is None and charmcraft_metadata:
+            if any(
+                meta_key in charmcraft_metadata for meta_key in ("name", "summary", "description")
```
benhoyt, line 417: A lot of this logic looks very similar to the logic in `main.py`. Ideally we can use `CharmMeta.from_files` or similar once that exists.

benhoyt: We also need a couple of tests of this behaviour (for metadata and for config) in `test/test_testing.py`.

benhoyt (changes requested):
Thanks for this PR! I've added a few comments/questions.

ops/main.py
syu-w: charmcraft will raise error when it found metadata in `charmcraft.yaml` but `metadata.yaml` also exists. `actions` and `config` also have the same behavior.

ops/charm.py
```diff
@@ -1118,18 +1118,22 @@ def __init__(self, raw: Optional[Dict[str, Any]] = None,
     @classmethod
     def from_yaml(
             cls, metadata: Union[str, TextIO],
-            actions: Optional[Union[str, TextIO]] = None) -> 'CharmMeta':
+            actions: Optional[Union[str, TextIO, Dict[str, Any]]] = None) -> 'CharmMeta':
```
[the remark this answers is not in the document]
syu-w: Added `from_dict`

ops/main.py
```diff
@@ -384,12 +387,28 @@ def main(charm_class: Type[ops.charm.CharmBase],
     dispatcher = _Dispatcher(charm_dir)
     dispatcher.run_any_legacy_hook()
 
-    metadata = (charm_dir / 'metadata.yaml').read_text()
-    actions_meta = charm_dir / 'actions.yaml'
-    if actions_meta.exists():
-        actions_metadata = actions_meta.read_text()
-    else:
-        actions_metadata = None
+    # Check charmcraft.yaml first and see if it has the metadata data in it.
```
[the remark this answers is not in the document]
syu-w: OK, I removed this part from `main.py` that seems not used when call testing.

ops/testing.py
syu-w: Added tests.

syu-w:
I removed the change in `main.py` that doesn't look like it's currently being used. `ops` in charm will still have separate yaml files as for now. Improved import logic, added `from_dict` and some tests.

ops/testing.py
```diff
@@ -389,57 +389,104 @@ def cleanup(self) -> None:
...
         Handle the cases where a user doesn't supply explicit metadata snippets.
+        This will try to load metadata from ``<charm_dir>/charmcraft.yaml`` first, then
+        ``<charm_dir>/metadata.yaml`` if charmcraft.yaml does not include metadata,
+        and ``<charm_dir>/actions.yaml`` if charmcraft.yaml does not include actions.
         """
         filename = inspect.getfile(self._charm_cls)
         charm_dir = pathlib.Path(filename).parents[1]
 
+        charm_metadata: Optional[Dict[str, Any]] = None
+        charmcraft_metadata: Optional[dict[str, Any]] = None
```
benhoyt, line 405: I'm not actually sure why this passes tests on Python 3.8, but the `dict[K, V]` syntax with lowercase `dict` type was only added in Python 3.9 (see https://docs.python.org/3/whatsnew/3.9.html and PEP 585). It would be interesting to figure out why the tests don't fail on 3.8! When I do this locally I get this error:

```
$ python3.8
Python 3.8.17 (default, Jun  8 2023, 17:21:52) 
[GCC 11.3.0] on linux
>>> d: dict[str, int] = {}
TypeError: 'type' object is not subscriptable
```

In any case, let's use `Dict` here for consistency. Also below.

ops/testing.py
```diff
@@ -389,57 +389,104 @@ def cleanup(self) -> None:
...
+            if isinstance(charm_metadata_yaml, str):
+                charm_metadata_yaml = dedent(charm_metadata_yaml)
+            charm_metadata = yaml.safe_load(charm_metadata_yaml)
+        else:
+            # Check charmcraft.yaml for metadata if no metadata is provided
+            if charmcraft_metadata is not None:
+                if any(
+                    meta_key in charmcraft_metadata
+                    for meta_key in ("name", "summary", "description")
+                ):
```
benhoyt, line 423: Style nit, but arguably a bit easier to read on two lines than four (and this is all about "meta", so no need for the `meta_` prefix in the loop variable name):


```suggestion
                meta_keys = ["name", "summary", "description"]
                if any(key in charmcraft_metadata for key in meta_keys):
```

ops/testing.py
```diff
@@ -389,57 +389,104 @@ def cleanup(self) -> None:
...
+                action_metadata = charmcraft_metadata["actions"]
+
+            # Still no actions, check actions.yaml
+            if action_metadata is None:
+                actions_path = charm_dir / 'actions.yaml'
+                if actions_path.is_file():
+                    action_metadata = yaml.safe_load(actions_path.read_text())
+                    self._charm_dir = charm_dir
+
+        return CharmMeta.from_dict(charm_metadata, action_metadata)
```
benhoyt, line 456: This is where (unless I'm missing something) we can just use `CharmMeta(charm_metadata, action_metadata)`.

test/test_testing.py
```diff
@@ -1262,6 +1262,32 @@ def test_metadata_from_directory(self):
         # The charm_dir also gets set
         self.assertEqual(harness.framework.charm_dir, tmp)
 
+    def test_metadata_from_directory_charmcraft_yaml(self):
+        tmp = pathlib.Path(tempfile.mkdtemp())
+        self.addCleanup(shutil.rmtree, str(tmp))
```
benhoyt, line 1267: Nit: file operations like `rmtree` work fine with `pathlib.Path` objects, so no need for the `str()` here.

test/test_testing.py
```diff
@@ -1262,6 +1262,32 @@ def test_metadata_from_directory(self):
...
+        charmcraft_filename = tmp / 'charmcraft.yaml'
+        with charmcraft_filename.open('wt') as charmcraft_yaml:
+            charmcraft_yaml.write(textwrap.dedent('''
```
benhoyt, line 1270: Simpler to use `charmcraft_filename.write_text()` rather than `with ... write`?

test/test_testing.py
```diff
@@ -1302,6 +1328,56 @@ def test_config_from_directory(self):
         self.assertIsNone(harness._backend._config._defaults['opt_null'])
         self.assertIsNone(harness._backend._config._defaults['opt_no_default'])
 
+    def test_config_from_directory_charmcraft_yaml(self):
+        tmp = pathlib.Path(tempfile.mkdtemp())
+        self.addCleanup(shutil.rmtree, str(tmp))
+        charmcraft_filename = tmp / 'charmcraft.yaml'
+        with charmcraft_filename.open('wt') as charmcraft_yaml:
+            charmcraft_yaml.write(textwrap.dedent('''
```
benhoyt, line 1336: Similar nits here about not need `str()` and using `write_text`.

test/test_testing.py
```diff
@@ -1302,6 +1328,56 @@ def test_config_from_directory(self):
...
+              - build-on:
+                - name: ubuntu
+                  channel: "22.04"
+                run-on:
+                - name: ubuntu
+                  channel: "22.04"
+
+            config:
+                options:
+                    opt_str:
```
benhoyt, line 1348: We don't need so many options to test this -- we're just testing that it gets loaded from the right place. I think two would be fine, maybe `opt_str` and `opt_int`.

benhoyt (commented):
This is looking very close, thanks. A few nit/style comments and one comment that I don't think `from_dict` is actually needed (just use the initialiser).

syu-w:
Thanks for the comments, updated.

benhoyt (approved):

PR #978, opened 2023-07-21
Replace `typing.AnyStr` with `Union[str, bytes]`

sobolevn:
Hi! While working on https://github.com/python/mypy/pull/15732 our tools detected a misuse of `AnyStr` TypeVar (which is quite common). The proper way is to use `Union` here :)


benhoyt (approved):
Thanks for this!

PR #982, opened 2023-07-28
effectively disable type checking on ObjectEvent namespace

### LINKED ISSUE #981: improve typing for ObjectEvents
right now, charm code has to do loads of `#type: ignores` on `framework.observe` calls because `ObjectEvents` dynamically defines its members based on charm metadata.

This makes for ugly code:
image

we should investigate how to ease this situation.

PietroPasotti:
Fixes #981

This PR tells type checkers to make no assumptions about the names that will be found in ObjectEvents.

- pros: type checkers will no longer complain on `self.on.dynamically_defined_name_action`
- cons: type checkers will no longer give type hints for events names that we do know to exist in subclasses, e.g. `start, stop, install`...

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -489,6 +489,13 @@ class ObjectEvents(Object):
         event_kinds = ', '.join(sorted(self._event_kinds()))
         return f'<{k.__module__}.{k.__qualname__}: {event_kinds}>'
 
+    def __getattr__(self, name: str):
+        """This tells type checkers to suspend judgement on this type's attributes.
+
+        This allows charms to freely write `self.on.foobar_baz` without mypy/pyright whining.
+        """
+        return super().__getattribute__(name)
+
 
 class PrefixedEvents:
     """Events to be found in all events using a specific prefix."""
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -489,10 +489,11 @@ class ObjectEvents(Object):
         event_kinds = ', '.join(sorted(self._event_kinds()))
         return f'<{k.__module__}.{k.__qualname__}: {event_kinds}>'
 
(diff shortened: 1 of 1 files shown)


ops/framework.py
```diff
@@ -489,6 +489,13 @@ def __repr__(self):
         event_kinds = ', '.join(sorted(self._event_kinds()))
         return f'<{k.__module__}.{k.__qualname__}: {event_kinds}>'
 
+    def __getattr__(self, name: str):
+        """This tells type checkers to suspend judgement on this type's attributes.
+
+        This allows charms to freely write `self.on.foobar_baz` without mypy/pyright whining.
+        """
+        return super().__getattribute__(name)
```
benhoyt, line 497: I checked this out locally and replaced this line with `raise AttributeError`, and all the tests still passed, so this code path isn't being exercised. That surprised me, but I guess it's because when we access such dynamically-defined events they *are* defined on the class and so `__getattr__` never gets called. Does that match your thinking?

ops/framework.py
```diff
@@ -489,6 +489,13 @@ def __repr__(self):
         event_kinds = ', '.join(sorted(self._event_kinds()))
         return f'<{k.__module__}.{k.__qualname__}: {event_kinds}>'
 
+    def __getattr__(self, name: str):
+        """This tells type checkers to suspend judgement on this type's attributes.
+
+        This allows charms to freely write `self.on.foobar_baz` without mypy/pyright whining.
```
benhoyt, line 495: Maybe we can give a more realistic example in the docstring, like "This allows charms to access dynamically-defined events such as `self.on.db_relation_joined` without Mypy/Pyright whining."

benhoyt (approved):

PR #983, opened 2023-08-01
add uri method for secret

delgod:
It is needed to consistently compare secrets for uniqueness. 

discussion - https://bugs.launchpad.net/juju/+bug/2028402


benhoyt:
After discussion with @hpidcock and @wallyworld, I've commented with an alternative here: https://bugs.launchpad.net/juju/+bug/2028402/comments/11

PR #986, opened 2023-08-03
Add Secret.unique_identifier (XID part); improve id/label docs

benhoyt:
In some cases when a charm has an array or set of secrets that don't have distinguishing characters, charms need a way to distinguish by ID (most of the time you'd use labels: most secrets have a specific use like "database-password" that you can use as a label). Despite the name, you can't `Secret.id` as a unique identifier, because sometimes it includes the model UUID, and sometimes not -- it's really intended as a locator URI, not a unique ID. So provide `Secret.unique_identifier` to handle this use case.

It would be nice if Secret.id was a unique identifier, but it's really a locator URI. Too late to change that, though.

We considered deprecating Secret.id entirely, and adding two new properties, "uri" and "unique_identifier". However, given that this locator is called ID throughout the rest of the codebase and by Juju, I think that would introduce more confusion.

The name "unique_identifier" is long on purpose: it should be relatively rare to use it, and we want to nudge people to using the locator (.id) and labels instead.

See more context at https://bugs.launchpad.net/juju/+bug/2028402

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -249,9 +249,13 @@ class Model:
     def get_secret(self, *, id: Optional[str] = None, label: Optional[str] = None) -> 'Secret':
         """Get the :class:`Secret` with the given ID or label.
 
-        You must provide at least one of `id` or `label`. If you provide both,
-        the secret will be fetched by ID, and the secret's label will be
-        updated to the label you provided.
+        The caller must provide at least one of `id` (the secret's locator ID)
+        or `label` (the charm-local "name").
+
+        If both are provided, the secret will be fetched by ID, and the
+        secret's label will be updated to the label provided. Normally secret
+        owners set a label using ``add_secret``, whereas secret observers set
+        a label using ``get_secret`` (see an example at :attr:`Secret.label`).
 
         Args:
             id: Secret ID if fetching by ID.
@@ -411,8 +415,8 @@ class Application:
         Args:
             content: A key-value mapping containing the payload of the secret,
                 for example :code:`{"password": "foo123"}`.
-            label: Private label to assign to this secret, which can later be
-                used for lookup.
+            label: Charm-local label (or "name") to assign to this secret,
+                which can later be used for lookup.
             description: Description of the secret's purpose.
             expire: Time in the future (or timedelta from now) at which the
                 secret is due to expire. When that time elapses, Juju will
@@ -1101,8 +1105,10 @@ class Secret:
         may not include the model UUID (for cross-model secrets).
 
         Charms should treat this as an opaque string for looking up secrets
-        and sharing them via relation data. If you want a truly unique
-        identifier for checking equality, use :attr:`unique_identifier`.
+        and sharing them via relation data. If you need a charm-local "name"
+        for a secret, use a :attr:`label`. (If a charm needs a truly unique
+        identifier for identifying one secret in a set of secrets of arbitrary
+        size, use :attr:`unique_identifier` -- this should be rare.)
 
         This will be None if you obtained the secret using
         :meth:`Model.get_secret` with a label but no ID.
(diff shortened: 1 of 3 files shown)


test/test_model.py
```diff
@@ -2973,6 +2973,37 @@ def test_get_secret_other_error(self):
...
+
+        secret = self.model.get_secret(id='123')
+        self.assertEqual(secret.id, 'secret:123')
+        self.assertEqual(secret.unique_identifier, '123')
+
+        secret = self.model.get_secret(id='secret:124')
+        self.assertEqual(secret.id, 'secret:124')
+        self.assertEqual(secret.unique_identifier, '124')
+
+        secret = self.model.get_secret(id='secret://125')
```
wallyworld, line 2991: This is not a valid secret URI

wallyworld (approved):

test/test_model.py
benhoyt: Thanks - removed.

ops/model.py
```diff
@@ -1094,13 +1094,43 @@ def _validate_content(cls, content: Optional[Dict[str, str]]):
 
     @property
     def id(self) -> Optional[str]:
-        """Unique identifier for this secret.
+        """Locator ID (URI) for this secret.
+
+        This has an unfortunate name for historical reasons, as it's not
+        really a unique identifier, but the secret's locator URI, which may or
+        may not include the model UUID (for cross-model secrets).
```
jnsgruk, line 1101: We could consider moving all the logic from this method into a new method called `uri`, and just leave this as a stub that calls `self.uri` and logs a deprecation warning for now?

jnsgruk (approved):
One possible suggestion, but not a huge issue and your concern about consistency would still be true. LGTM.

benhoyt:
@delgod says:

> the whole ticket is about issue that ".id" doesn't return secret's locator URI all the times. sometimes it is secret's locator URI and sometimes it is 20-character Xid.
> it is hits us, because we cannot compare values :pensive:
> we ask to have some field which returns Locator ID always.
> we cannot compare XID because they can be same in different models

I don't think that's correct, though, because [Xids are globally unique](https://pkg.go.dev/github.com/rs/xid#section-readme) (they're a form of UUID). You can't use them for Juju lookup, but you can use them to compare/test for equality. That's why I thought adding `unique_identifier` would solve it for your case of comparing values for equality, no?

ops/model.py
```diff
@@ -1094,20 +1098,80 @@ def _validate_content(cls, content: Optional[Dict[str, str]]):
...
         return self._id
 
+    @property
+    def unique_identifier(self) -> Optional[str]:
+        """Unique identifier of this secret.
+
+        This is the secret's globally-unique identifier (currently a
+        20-character Xid, for example "9m4e2mr0ui3e8a215n4g").
+
+        Most charms should use :attr:`id` (the secret's locator ID) to send
```
wallyworld, line 1125: Charms should always use id in relation data. I think the "most" bit pertains to the clause which follows  about labels.

ops/model.py
```diff
@@ -1094,20 +1098,80 @@ def _validate_content(cls, content: Optional[Dict[str, str]]):
...
+        the secret's ID across relation data, and labels (:attr:`label`) to
+        assign a charm-local "name" to the secret for lookup in this charm.
+        However, ``unique_identifier`` can be useful to distinguish secrets in
+        cases where the charm has a set of secrets of arbitrary size, for
+        example, a group of 10 or 20 TLS certificates.
+
+        This will be None if you obtained the secret using
```
wallyworld, line 1132: Maybe mention it is always set if you are the owner

wallyworld (approved):

ops/model.py
benhoyt: Good call. I just removed "Most" here.

ops/model.py
benhoyt: This is not actually true: it will be `None` if you are the owner also (if you don't provide a label). This is how the API / modelling works: `get_secret` calls the `secret-get` hook tool only, but not `secret-info-get`, so it doesn't fetch the label.

benhoyt:
`alertmanager-k8s-operator` tests are failing due to https://github.com/canonical/alertmanager-k8s-operator/pull/181, but that's completely unrelated to this PR, so merging anyway.

PR #991, opened 2023-08-15
Ignore push() encoding if source is a binary file or stream

### LINKED ISSUE #990: TypeError: write() argument must be str, not bytes
Since `2.5.0` and (I think) specially after #960

In [Loki unit test we are getting](https://github.com/canonical/loki-k8s-operator/actions/runs/5867832804/job/15909377429?pr=299#step:5:402) ([PR-299](https://github.com/canonical/loki-k8s-operator/pull/299)):

```python
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/runner/work/loki-k8s-operator/loki-k8s-operator/tests/unit/test_log_proxy_consumer.py", line 310, in test__fetch_promtail_from_attached_resource
    self.harness.charm.log_proxy._push_promtail_if_attached(binary_path)
  File "/home/runner/work/loki-k8s-operator/loki-k8s-operator/lib/charms/loki_k8s/v0/loki_push_api.py", line 2027, in _push_promtail_if_attached
    self._push_binary_to_workload(resource_path, workload_binary_path)
  File "/home/runner/work/loki-k8s-operator/loki-k8s-operator/lib/charms/loki_k8s/v0/loki_push_api.py", line 1995, in _push_binary_to_workload
    self._container.push(workload_binary_path, f, permissions=0o755, make_dirs=True)
  File "/home/runner/work/loki-k8s-operator/loki-k8s-operator/.tox/unit/lib/python3.8/site-packages/ops/model.py", line 2073, in push
    self._pebble.push(str(path), source, encoding=encoding,
  File "/home/runner/work/loki-k8s-operator/loki-k8s-operator/.tox/unit/lib/python3.8/site-packages/ops/testing.py", line 2596, in push
    shutil.copyfileobj(cast(IOBase, source), cast(IOBase, f))
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/shutil.py", line 208, in copyfileobj
    fdst_write(buf)
TypeError: write() argument must be str, not bytes
```

The error is in [loki_push_api.py line 1995](https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py#L1995):

```python
self._container.push(workload_binary_path, f, permissions=0o755, make_dirs=True)
```
The [`.push()` method receive](https://ops.readthedocs.io/en/latest/index.html#ops.pebble.Client.push):

- path – Path of the file to write to on the remote system.
- source – Source of data to write. This is either a concrete str or bytes instance, or a readable file-like object.

Debugging the code, in that line I'm sending:

```python
(Pdb) workload_binary_path
'/tmp/promtail-linux-amd64'
(Pdb) f                   
<_io.BufferedReader name='/tmp/tmp-ops-test-resource-26vkn2p2/promtail-bin/promtail-linux-amd64'>
(Pdb) f.name              
'/tmp/tmp-ops-test-resource-26vkn2p2/promtail-bin/promtail-linux-amd64'
```

benhoyt:
The [push() docs](https://ops.readthedocs.io/en/latest/#ops.Container.push) explicitly say that the "encoding" param is ignored if the source is bytes or a binary stream. This is useful because the default is encoding='utf-8'; but we should already know the source stream is binary, so let that override the default.

Also add a test that "encoding" is ignored in these cases.

Fixes #990

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -19,6 +19,7 @@ import dataclasses
 import datetime
 import fnmatch
 import inspect
+import io
 import ipaddress
 import os
 import pathlib
@@ -2592,7 +2593,11 @@ class _TestingPebbleClient:
             elif isinstance(source, bytes):
                 file_path.write_bytes(source)
             else:
-                with file_path.open('wb' if encoding is None else 'w', encoding=encoding) as f:
+                # If source is binary, open file in binary mode and ignore encoding param
+                is_binary = isinstance(source, (io.RawIOBase, io.BufferedIOBase))
+                open_mode = 'wb' if is_binary else 'w'
+                open_encoding = None if is_binary else encoding
+                with file_path.open(open_mode, encoding=open_encoding) as f:
                     shutil.copyfileobj(cast(IOBase, source), cast(IOBase, f))
             os.chmod(file_path, permissions)
         except FileNotFoundError as e:
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -2594,7 +2594,7 @@ class _TestingPebbleClient:
                 file_path.write_bytes(source)
             else:
                 # If source is binary, open file in binary mode and ignore encoding param
-                is_binary = isinstance(source, (io.RawIOBase, io.BufferedIOBase))
+                is_binary = isinstance(source.read(0), bytes)
                 open_mode = 'wb' if is_binary else 'w'
                 open_encoding = None if is_binary else encoding
                 with file_path.open(open_mode, encoding=open_encoding) as f:
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -19,7 +19,6 @@ import dataclasses
 import datetime
 import fnmatch
 import inspect
-import io
 import ipaddress
 import os
 import pathlib
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -3998,6 +3998,22 @@ class _PebbleStorageAPIsTestMixin:
             received_data = infile.read()
         self.assertEqual(original_data, received_data)
 
+    def test_push_bytes_ignore_encoding(self):
+        # push() encoding param should be ignored if source is bytes
+        client = self.client
+        client.push(f"{self.prefix}/test", b'\x00\x01', encoding='utf-8')
+        with client.pull(f"{self.prefix}/test", encoding=None) as infile:
+            received_data = infile.read()
+        self.assertEqual(received_data, b'\x00\x01')
+
+    def test_push_bytesio_ignore_encoding(self):
+        # push() encoding param should be ignored if source is binary stream
+        client = self.client
+        client.push(f"{self.prefix}/test", io.BytesIO(b'\x00\x01'), encoding='utf-8')
+        with client.pull(f"{self.prefix}/test", encoding=None) as infile:
+            received_data = infile.read()
+        self.assertEqual(received_data, b'\x00\x01')
+
     def test_push_and_pull_larger_file(self):
         # Intent: to ensure things work appropriately with larger files.
         # Larger files may be sent/received in multiple chunks; this should help for


benhoyt:
@weiiwang01 Does this fix look reasonable to you?

benhoyt:
@weiiwang01 That's an interesting idea. I got my suggestion from https://stackoverflow.com/a/44584871/68707; your approach is also suggested in another answer.

Part of the issue with Python is I don't think the behaviour of "read" is really defined anywhere. Is it guaranteed to work given a zero length? It seems to me some implementations might raise an exception in that case (`ValueError('read size cannot be 0')`), though filesystem files and `io.BytesIO` do not. Also seems slightly better to avoid a syscall (some implementations might do a syscall even if size is 0). The [`RawIOBase.read` docs](https://docs.python.org/3/library/io.html#io.RawIOBase.read) indicate 0 is allowed. Hmmm, I'm not sure which is better.

benhoyt:
FWIW, I just checked, and a normal `open()` with default buffering does not do a `read` syscall for `f.read(0)`. But `open(filename, 'rb', buffered=0)` does do a syscall. It seems better to me to avoid a syscall (which may raise `IOError`) in all cases and prefer the isinstance check here. All the types that I think will be used here support the isinstance check (file, `BytesIO`, and presumably most custom types will too).

weiiwang01:
> @weiiwang01 That's an interesting idea. I got my suggestion from https://stackoverflow.com/a/44584871/68707; your approach is also suggested in another answer.
> 
> Part of the issue with Python is I don't think the behaviour of "read" is really defined anywhere. Is it guaranteed to work given a zero length? It seems to me some implementations might raise an exception in that case (`ValueError('read size cannot be 0')`), though filesystem files and `io.BytesIO` do not. Also seems slightly better to avoid a syscall (some implementations might do a syscall even if size is 0). The [`RawIOBase.read` docs](https://docs.python.org/3/library/io.html#io.RawIOBase.read) indicate 0 is allowed. Hmmm, I'm not sure which is better.

I think Python's read function likely derives from C's `read`. Since C's `read` explicitly permits a 0 count read, any file-like object that doesn't support this is not adhering to the expected specification in my view.

weiiwang01:
> > @weiiwang01 That's an interesting idea. I got my suggestion from https://stackoverflow.com/a/44584871/68707; your approach is also suggested in another answer.
> > Part of the issue with Python is I don't think the behaviour of "read" is really defined anywhere. Is it guaranteed to work given a zero length? It seems to me some implementations might raise an exception in that case (`ValueError('read size cannot be 0')`), though filesystem files and `io.BytesIO` do not. Also seems slightly better to avoid a syscall (some implementations might do a syscall even if size is 0). The [`RawIOBase.read` docs](https://docs.python.org/3/library/io.html#io.RawIOBase.read) indicate 0 is allowed. Hmmm, I'm not sure which is better.
> 
> I think Python's read function likely derives from C's `read`. Since C's `read` explicitly permits a 0 count read, any file-like object that doesn't support this is not adhering to the expected specification in my view.

About the performance concerns, it's worth noting that `read(0)` is called only once per stream, and this is Python and we're dealing with the test harness, I don't believe performance is the top priority in this context.

weiiwang01:
> FWIW, I just checked, and a normal `open()` with default buffering does not do a `read` syscall for `f.read(0)`. But `open(filename, 'rb', buffered=0)` does do a syscall. It seems better to me to avoid a syscall (which may raise `IOError`) in all cases and prefer the isinstance check here. All the types that I think will be used here support the isinstance check (file, `BytesIO`, and presumably most custom types will too).

I think if the `read(0)` raise an error, the same error will be raised in the following ` shutil.copyfileobj` as well, so if the IO error exists, avoiding one syscall don't really solve the problem here.

benhoyt:
@weiiwang01 Fair push-back -- I've updated to use `isinstance(source.read(0), bytes)`.

weiiwang01 (approved):

PR #996, opened 2023-08-24
fix(pebble): remove use of deprecated cgi module in Pebble code

### LINKED ISSUE #995: `cgi` module deprecated in Python 3.11
The `cgi` module in used in pebble.py is deprecated as of Python version 3.11 and is slated for removal in Python version 3.13. For example, when running charm unit tests through Python 3.11, you will see the following warning message:

```
.tox/unit/lib/python3.11/site-packages/ops/pebble.py:21
  /home/nucci/Documents/projects/work/operator-libs-linux/.tox/unit/lib/python3.11/site-packages/ops/pebble.py:21: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
    import cgi

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

We luckily still have a bit before we absolutely must remove `cgi` from pebble.py, but we should likely make plans to remove it before 24.04 LTS so that charm others do not get bombarded with `DeprecationWarning` messages. _I'm assuming that 24.04 will have at least Python 3.11 or 3.12_. [PEP594](https://peps.python.org/pep-0594/#cgi) provides some example alternatives that we could use, but there does not seem to be a direct replacement unfortunately :disappointed:

benhoyt:
The recommendation is to use the email.message.Message type. See example: https://docs.python.org/3.11/library/cgi.html#cgi.parse_header

Fixes #995

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -18,9 +18,9 @@ For a command-line interface for local testing, see test/pebble_cli.py.
 """
 
 import binascii
-import cgi
 import copy
 import datetime
+import email.message
 import email.parser
 import enum
 import http.client
@@ -134,8 +134,6 @@ PlanDict = typing.TypedDict('PlanDict',
                             total=False)
 
 if TYPE_CHECKING:
-    from email.message import Message
-
     from typing_extensions import Literal, Protocol, TypedDict
 
     # callback types for _MultiParser header and body handlers
@@ -1563,13 +1561,15 @@ class Client:
         return raw_resp
 
     @staticmethod
-    def _ensure_content_type(headers: 'Message',
+    def _ensure_content_type(headers: email.message.Message,
                              expected: 'Literal["multipart/form-data", "application/json"]'):
         """Parse Content-Type header from headers and ensure it's equal to expected.
 
         Return a dict of any options in the header, e.g., {'boundary': ...}.
         """
-        ctype, options = cgi.parse_header(headers.get('Content-Type', ''))
+        ctype = headers.get_content_type()
+        params = headers.get_params() or {}
+        options = dict((key, value) for key, value in params if value)
         if ctype != expected:
             raise ProtocolError(f'expected Content-Type {expected!r}, got {ctype!r}')
         return options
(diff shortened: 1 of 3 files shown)


jnsgruk (approved):

ops/pebble.py
```diff
@@ -1563,13 +1561,15 @@ def _request(self,
...
+    def _ensure_content_type(headers: email.message.Message,
                              expected: 'Literal["multipart/form-data", "application/json"]'):
         """Parse Content-Type header from headers and ensure it's equal to expected.
 
         Return a dict of any options in the header, e.g., {'boundary': ...}.
         """
-        ctype, options = cgi.parse_header(headers.get('Content-Type', ''))
+        ctype = headers.get_content_type()
+        params = headers.get_params() or {}
+        options = dict((key, value) for key, value in params if value)
```
NucciTheBoss, line 1572: Maybe just my stylistic preference more than anything, but is there a reason why you chose to use the `dict()` built-in function instead of the squiggly brackets to generate the options dictionary?

```python
options = {key: value for key, value in params if value}
```

I find the squiggly brackets to be slightly more readable than `dict(...)`; I usually only use `dict(...)` if I am initializing an empty dictionary or converting another type to a dictionary.

NucciTheBoss (changes requested):
Just some minor nits, but this looks good! It feels a little weird to me that we are using the email module to interface with a service manager, but I'm glad that it was not too much work to replace the `cgi` module :sweat_smile:

test/test_pebble.py
```diff
@@ -1955,12 +1958,12 @@ def test_pull_path_error(self):
...
 
     def test_pull_protocol_errors(self):
-        self.client.responses.append(({'Content-Type': 'ct'}, b''))
+        self.client.responses.append(({'Content-Type': 'c/t'}, b''))
         with self.assertRaises(pebble.ProtocolError) as cm:
             self.client.pull('/etc/hosts')
         self.assertIsInstance(cm.exception, pebble.Error)
         self.assertEqual(str(cm.exception),
-                         "expected Content-Type 'multipart/form-data', got 'ct'")
+                         "expected Content-Type 'multipart/form-data', got 'c/t'")
```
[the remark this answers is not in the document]
benhoyt: Yeah, the test started failing without it, because the new `email.message.Message` header parsing validates that the content type is of the form `type/subtype`.

ops/pebble.py
benhoyt: Good call, I prefer those too! Not sure why I used the old-school `dict()` method. Updated.

benhoyt:
> It feels a little weird to me that we are using the email module to interface with a service manager

@NucciTheBoss No kidding, feels weird to me too! It'd be much saner if it was `mime` or `mimeutils` or something.

PR #998, opened 2023-08-28
Raise error in wait_output if stdout is already being consumed

weiiwang01:
When the `stdout` argument is provided to `ops.Container.exec` for capturing the stdout stream, the `ExecProcess` will have a `stdout` attribute of `None`. To prevent potential null safety issues, an error is now raised in such cases within `wait_output`.

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1338,6 +1338,11 @@ class ExecProcess(Generic[AnyStr]):
             out = io.BytesIO()
             err = io.BytesIO() if self.stderr is not None else None
 
+        if self.stdout is None:
+            raise TypeError(
+                "stdout is already being consumed by another object provided to the exec function"
+            )
+
         t = _start_thread(shutil.copyfileobj, self.stdout, out)
         self._threads.append(t)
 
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Great, thanks for this.

ops/pebble.py
```diff
@@ -1338,6 +1338,11 @@ def wait_output(self) -> Tuple[AnyStr, Optional[AnyStr]]:
             out = io.BytesIO()
             err = io.BytesIO() if self.stderr is not None else None
 
+        if self.stdout is None:
+            raise TypeError(
+                "stdout is already being consumed by another object provided to the exec function"
```
benhoyt, line 1343: That said, two nits:

1) Could we move this up to the top of the function, so that we're checking inputs before performing any logic / allocations?
2) Perhaps we can slightly improve the error message to make it easier for the caller to know what (not) to do. Maybe "can't use wait_output() when exec was called with the stdout argument; use wait() instead".

weiiwang01: Okay, updated, thanks!

benhoyt (approved):

PR #1000, opened 2023-08-29
fix(model): ensure Secret.get_content returns a copy of the dict

### LINKED ISSUE #999: secret.get_content() un-intuitively cached locally
"Screenshot" from a debugger session.
As you can see, I'm fetching contents of a secret, saving it to `dict` called `full_content`.
Then I update `full_content`. NOT the secret, but the dictionary holding the output of the `get_content()` call.
Yet, when I call `secret.get_content()` afterwards, I find that it's returning the UPDATED content of the `full_content` dictionary, instead of the real contents of the secret. AGAIN, note that the secret has NOT been updated at this point.

However, as a result, the secret will actually never be updated. The update command silently fails on line 630, and neither the contents, nor the revision of the secret gets updated.

[The value of the `content` variable was `{'endpoints': 'host1:port,host2:port,host3:port'}` from the start]

```
625             import pdb; pdb.set_trace()
626  ->         full_content = secret.get_content()
627             old_content = copy.deepcopy(full_content)
628             full_content.update(content)
629             if old_content != full_content:
630                 secret.set_content(full_content)
631  
(Pdb) n
> /var/lib/juju/agents/unit-kafka-0/charm/lib/charms/data_platform_libs/v0/data_interfaces.py(627)update_relation_secret()
-> old_content = copy.deepcopy(full_content)
(Pdb) full_content
{'endpoints': 'host1:port,host2:port'}
(Pdb) n
> /var/lib/juju/agents/unit-kafka-0/charm/lib/charms/data_platform_libs/v0/data_interfaces.py(628)update_relation_secret()
-> full_content.update(content)
(Pdb) n
> /var/lib/juju/agents/unit-kafka-0/charm/lib/charms/data_platform_libs/v0/data_interfaces.py(629)update_relation_secret()
-> if old_content != full_content:
(Pdb) secret.get_content()
{'endpoints': 'host1:port,host2:port,host3:port'}
```

This behavior is highly unintuitive, and should be addressed.

benhoyt:
This avoids the caller modifying the Secret instance's dict.

Fixes #999

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1181,6 +1181,9 @@ class Secret:
     def get_content(self, *, refresh: bool = False) -> Dict[str, str]:
         """Get the secret's content.
 
+        Returns:
+            A copy of the secret's content dictionary.
+
         Args:
             refresh: If true, fetch the latest revision's content and tell
                 Juju to update to tracking that revision. The default is to
@@ -1190,7 +1193,7 @@ class Secret:
         if refresh or self._content is None:
             self._content = self._backend.secret_get(
                 id=self.id, label=self.label, refresh=refresh)
-        return self._content
+        return self._content.copy()
 
     def peek_content(self) -> Dict[str, str]:
         """Get the content of the latest revision of this secret.
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -3104,6 +3104,18 @@ class TestSecretClass(unittest.TestCase):
         self.assertEqual(fake_script_calls(self, clear=True),
                          [['secret-get', 'secret:z', '--format=json']])
 
+    def test_get_content_copies_dict(self):
+        fake_script(self, 'secret-get', """echo '{"foo": "bar"}'""")
+
+        secret = self.make_secret(id='z')
+        content = secret.get_content()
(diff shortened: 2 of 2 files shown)


ops/model.py
```diff
@@ -1190,7 +1193,7 @@ def get_content(self, *, refresh: bool = False) -> Dict[str, str]:
         if refresh or self._content is None:
             self._content = self._backend.secret_get(
                 id=self.id, label=self.label, refresh=refresh)
-        return self._content
```
jameinel, line 1193: Does it need to be a deepcopy?

I think this is significantly better than not doing it, I just wonder if secret contents are an object, can they have nested values?

jameinel (approved):
Can we add a test at a high level that we either cannot have nested content, or that we are doing a deep copy?

ops/model.py
benhoyt: I've confirmed in the Juju `secret-get` and `secret-set` code that these are always plain `keystr=valstr` mappings, so there's never going to be nesting here.

In addition, for what it's worth, the Python type annotation is `Dict[str, str]`.

hpidcock:
#1000 🎉

PR #1001, opened 2023-08-30
fix(model): make Secret.set_content invalidate local cache

benhoyt:
This is so that if get_content() is called after, it fetches and returns the expected new content. We could also set self._content to the new content, but re-fetching it from Juju seems better.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -3116,6 +3116,24 @@ class TestSecretClass(unittest.TestCase):
         self.assertEqual(fake_script_calls(self, clear=True),
                          [['secret-get', 'secret:z', '--format=json']])
 
+    def test_set_content_invalidates_cache(self):
+        fake_script(self, 'secret-get', """echo '{"foo": "bar"}'""")
+        fake_script(self, 'secret-set', """exit 0""")
+
+        secret = self.make_secret(id='z')
+        old_content = secret.get_content()
+        self.assertEqual(old_content, {'foo': 'bar'})
+        secret.set_content({'new': 'content'})
+        fake_script(self, 'secret-get', """echo '{"new": "content"}'""")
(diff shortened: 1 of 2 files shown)


wallyworld (approved):

test/test_model.py
```diff
@@ -3116,6 +3116,24 @@ def test_get_content_copies_dict(self):
...
 
+    def test_set_content_invalidates_cache(self):
+        fake_script(self, 'secret-get', """echo '{"foo": "bar"}'""")
+        fake_script(self, 'secret-set', """exit 0""")
+
+        secret = self.make_secret(id='z')
+        old_content = secret.get_content()
+        self.assertEqual(old_content, {'foo': 'bar'})
+        secret.set_content({'new': 'content'})
+        fake_script(self, 'secret-get', """echo '{"new": "content"}'""")
```
jameinel, line 3127: I wish we didn't have to force the fake_script, that certainly makes the test feel like a tautology (though I realize it actually isn't).

Also, should we be adding a test around testing.Harness and that when we're using its backend that we get the right behavior?

jameinel (approved):

test/test_model.py
benhoyt: Yeah, the second `fake_script` felt weird, but it's correct here. I thought about have a more realistic secret-set/secret-get that wrote to and read from a file instead, but that seems like unnecessary work for what we're testing here. Here we just need to assert that the second `get_content()` call returns the updated content rather than the cached content as it did previously.

Regarding `Harness`, I think that is an interesting idea in general, but I don't think we'd be adding anything for this case by adding a specific test. This isn't a `Harness` problem but a `model.Secret` problem -- it's backend-independent.

PR #1003, opened 2023-09-20
docs: remove 'you' and 'your' from docstrings

### LINKED ISSUE #959: Remove references to "you" and "your" in API docstrings
Per Jon's comment [here](https://github.com/canonical/operator/pull/953#discussion_r1247545755), we should clean up the style in our API docstrings to avoid using "you".

tonyandrewmeyer:
Only changes docstrings.

Prefer imperative style in general, and refer to the charm rather than a nebulous 'you', as per #959 and the linked comment.

Fixes #959.

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -519,7 +519,7 @@ class RelationDepartedEvent(RelationEvent):
     def departing_unit(self) -> Optional[model.Unit]:
         """The :class:`ops.Unit` that is departing, if any.
 
-        Use this to determine (for example) whether this unit is the
+        Use this method to determine (for example) whether this unit is the
         departing one.
         """
         # doing this on init would fail because `framework` gets patched in
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -241,7 +241,7 @@ class ConfigChangedEvent(HookEvent):
       This event notifies the charm of its initial configuration.
       Typically, this event will fire between an :class:`install <InstallEvent>`
       and a :class:`start <StartEvent>` during the startup sequence
-      (when you first deploy a unit), but in general it will fire whenever
+      (when a unit is first deployed), but in general it will fire whenever
       the unit is (re)started, for example after pod churn on Kubernetes, on unit
       rescheduling, on unit upgrade or refresh, and so on.
     - As a specific instance of the above point: when networking changes
@@ -519,8 +519,8 @@ class RelationDepartedEvent(RelationEvent):
     def departing_unit(self) -> Optional[model.Unit]:
         """The :class:`ops.Unit` that is departing, if any.
 
-        You can use this to determine (for example) whether *you* are the
-        departing unit.
+        Use this to determine (for example) whether this unit is the
+        departing one.
         """
         # doing this on init would fail because `framework` gets patched in
         # post-init
@@ -729,7 +729,7 @@ class SecretChangedEvent(SecretEvent):
     a new secret revision, and all applications or units that are tracking this
     secret will be notified via this event that a new revision is available.
 
-    Typically, you will want to fetch the new content by calling
+    Typically, fetch the new content by calling 
     :meth:`event.secret.get_content() <ops.Secret.get_content>` with ``refresh=True``
     to tell Juju to start tracking the new revision.
     """
@@ -757,7 +757,7 @@ class SecretRemoveEvent(SecretEvent):
     observers have updated to that new revision, this event will be fired to
     inform the secret owner that the old revision can be removed.
 
-    Typically, you will want to call
+    Typically, call
     :meth:`event.secret.remove_revision() <ops.Secret.remove_revision>` to
     remove the now-unused revision.
     """
@@ -1005,7 +1005,7 @@ class CharmBase(Object):
 
     :code:`CharmBase` is used to create a charm. This is done by inheriting
     from :code:`CharmBase` and customising the subclass as required. So to
-    create your own charm, say ``MyCharm``, define a charm class and set up the
+    create a charm, say ``MyCharm``, define a charm class and set up the
     required event handlers (“hooks”) in its constructor::
 
         import logging
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -729,7 +729,7 @@ class SecretChangedEvent(SecretEvent):
     a new secret revision, and all applications or units that are tracking this
     secret will be notified via this event that a new revision is available.
 
-    Typically, fetch the new content by calling
+    Typically, the charm will fetch the new content by calling
     :meth:`event.secret.get_content() <ops.Secret.get_content>` with ``refresh=True``
     to tell Juju to start tracking the new revision.
     """
@@ -757,7 +757,7 @@ class SecretRemoveEvent(SecretEvent):
     observers have updated to that new revision, this event will be fired to
     inform the secret owner that the old revision can be removed.
 
-    Typically, call
+    Typically, the charm will call
     :meth:`event.secret.remove_revision() <ops.Secret.remove_revision>` to
     remove the now-unused revision.
     """
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -1005,7 +1005,7 @@ class CharmBase(Object):
 
     :code:`CharmBase` is used to create a charm. This is done by inheriting
     from :code:`CharmBase` and customising the subclass as required. So to
-    create a charm, say ``MyCharm``, define a charm class and set up the
+    create a charm called ``MyCharm``, define a charm class and set up the
     required event handlers (“hooks”) in its constructor::
 
         import logging
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -729,7 +729,7 @@ class SecretChangedEvent(SecretEvent):
     a new secret revision, and all applications or units that are tracking this
     secret will be notified via this event that a new revision is available.
 
-    Typically, fetch the new content by calling 
+    Typically, fetch the new content by calling
     :meth:`event.secret.get_content() <ops.Secret.get_content>` with ``refresh=True``
     to tell Juju to start tracking the new revision.
     """
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -224,11 +224,10 @@ class EventBase:
         better to just wait for the event that indicates the precondition has
         been met.
 
-        For example, if ``config-changed`` is fired, and handling the event
-        cannot complete without a different config, there is no reason to defer
-        the event because there will be a *different* ``config-changed`` event
-        when the config actually changes, rather than checking to see if maybe
-        config has changed prior to every other event that occurs.
+        For example, if handling a config change requires that two config
+        values are changed, there's no reason to defer the first
+        ``config-changed`` because there will be a *second* ``config-changed``
+        event fired when the other config value changes.
 
         Similarly, if two events need to occur before execution can proceed
         (say event A and B), the event A handler could ``defer()`` because B
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -202,7 +202,7 @@ class EventBase:
         the result of an action, or any event other than metric events. The
         queue of events will be dispatched before the new event is processed.
 
-        From the above you may deduce, but it's important to point out:
+        Important points that follow from the above:
 
         * ``defer()`` does not interrupt the execution of the current event
           handler. In almost all cases, a call to ``defer()`` should be followed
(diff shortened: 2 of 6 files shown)


ops/charm.py
```diff
@@ -519,8 +519,8 @@ def snapshot(self) -> Dict[str, Any]:
     def departing_unit(self) -> Optional[model.Unit]:
         """The :class:`ops.Unit` that is departing, if any.
 
-        You can use this to determine (for example) whether *you* are the
-        departing unit.
+        Use this to determine (for example) whether this unit is the
```
benhoyt, line 522: I like the change, but I think we should say "Use this method ..." or similar for clarity.

ops/charm.py
```diff
@@ -729,7 +729,7 @@ class SecretChangedEvent(SecretEvent):
     a new secret revision, and all applications or units that are tracking this
     secret will be notified via this event that a new revision is available.
 
-    Typically, you will want to fetch the new content by calling
+    Typically, fetch the new content by calling
```
benhoyt, line 732: I think this is a bit terse/odd. Perhaps "Typically, the charm will fetch...". Similar in a few places below.

ops/charm.py
```diff
@@ -1005,7 +1005,7 @@ class CharmBase(Object):
 
     :code:`CharmBase` is used to create a charm. This is done by inheriting
     from :code:`CharmBase` and customising the subclass as required. So to
-    create your own charm, say ``MyCharm``, define a charm class and set up the
+    create a charm, say ``MyCharm``, define a charm class and set up the
```
benhoyt, line 1008: Not your issue, but "say" reads strangely to me. Perhaps:
```suggestion
    create a charm called ``MyCharm``, define a charm class and set up the
```

ops/framework.py
```diff
@@ -220,19 +220,19 @@ def defer(self) -> None:
...
-        For example, if ``config-changed`` is fired, and you are waiting for
-        different config, there is no reason to defer the event because there
-        will be a *different* ``config-changed`` event when the config actually
-        changes, rather than checking to see if maybe config has changed prior
-        to every other event that occurs.
+        For example, if ``config-changed`` is fired, and handling the event
+        cannot complete without a different config, there is no reason to defer
+        the event because there will be a *different* ``config-changed`` event
+        when the config actually changes, rather than checking to see if maybe
+        config has changed prior to every other event that occurs.
```
benhoyt, line 231: This is a fine change, though I think this paragraph needs a rewrite. Some of the phrasing is odd/confusing. Feel free to have a shot if you want, or we can leave it for now.

ops/lib/__init__.py
```diff
@@ -96,9 +96,9 @@ def use(name: str, api: int, author: str) -> ModuleType:
 def autoimport():
     """Find all libs in the path and enable use of them.
 
-    You only need to call this if you've installed a package or
-    otherwise changed sys.path in the current run, and need to see the
-    changes. Otherwise libraries are found on first call of `use`.
+    Call this only when a package has been installed or sys.path has been
```
benhoyt, line 99:
```suggestion
    Call this function only when a package has been installed or sys.path has been
```

ops/main.py
```diff
@@ -368,7 +368,7 @@ def main(charm_class: Type[ops.charm.CharmBase],
     The event name is based on the way this executable was called (argv[0]).
 
     Args:
-        charm_class: your charm class.
+        charm_class: the charm class to instantiate and receive the event.
```
benhoyt, line 371: Nice improvement.

ops/model.py
```diff
@@ -195,7 +195,7 @@ def uuid(self) -> str:
     def get_unit(self, unit_name: str) -> 'Unit':
         """Get an arbitrary unit by name.
 
-        Use :attr:`unit` to get your own unit.
+        Use :attr:`unit` to get the model's own unit.
```
benhoyt, line 198: Not sure about this, as "the model" has several units. Perhaps "the current unit"?

ops/model.py
```diff
@@ -205,7 +205,7 @@ def get_unit(self, unit_name: str) -> 'Unit':
     def get_app(self, app_name: str) -> 'Application':
         """Get an application by name.
 
-        Use :attr:`app` to get your own application.
+        Use :attr:`app` to get the model's own application.
```
benhoyt, line 208: Similar here. Perhaps "this charm's application".

ops/model.py
```diff
@@ -487,15 +486,14 @@ def _invalidate(self):
     def status(self) -> 'StatusBase':
         """Used to report or read the status of a specific unit.
 
-        The status of any unit other than yourself is always Unknown.
+        The status of any unit in a different charm is always Unknown.
```
benhoyt, line 489: This isn't quite right, as the charm can be deployed on multiple units. Perhaps:
```suggestion
        The status of any unit other than the current unit is always Unknown.
```

ops/model.py
```diff
@@ -487,15 +486,14 @@ def _invalidate(self):
...
-        You can also use the :attr:`collect_unit_status <CharmEvents.collect_unit_status>`
-        event if you want to evaluate and set unit status consistently at the
-        end of every hook.
+        Alternatively, use the :attr:`collect_unit_status <CharmEvents.collect_unit_status>`
+        event to evaluate and set unit status consistently at the end of every hook.
 
         Raises:
-            RuntimeError: if you try to set the status of a unit other than yourself.
-            InvalidStatusError: if you try to set the status to something other than
+            RuntimeError: if setting the status of a unit in another charm.
```
benhoyt, line 495:
```suggestion
            RuntimeError: if setting the status of a unit other than the current unit
```

ops/model.py
```diff
@@ -531,10 +529,10 @@ def __repr__(self):
     def is_leader(self) -> bool:
         """Return whether this unit is the leader of its application.
 
-        This can only be called for your own unit.
+        This can only be called for the charm's own unit.
```
benhoyt, line 532: Similar here and below: the other `Unit` instance may be a `Unit` instance from this charm. I think probably "current unit" throughout, unless you can think of a better phrase.

ops/model.py
```diff
@@ -2179,7 +2177,7 @@ def push_path(self,
         * /foo/foobar.txt
         * /quux.txt
 
-        You could push the following ways::
+        These are ways to push::
```
benhoyt, line 2180:
```suggestion
        These are various push example::
```

ops/model.py
```diff
@@ -2260,7 +2258,7 @@ def pull_path(self,
         * /foo/foobar.txt
         * /quux.txt
 
-        You could pull the following ways::
+        These are ways to pull::
```
benhoyt, line 2261:
```suggestion
        These are various pull examples::
```

ops/model.py
```diff
@@ -2651,9 +2649,9 @@ def __init__(self, relation_name: str, num_related: int, max_supported: int):
 class RelationDataError(ModelError):
     """Raised when a relation data read/write is invalid.
 
-    This is raised if you're either trying to set a value to something that isn't a string,
-    or if you are trying to set a value in a bucket that you don't have access to. (For example,
-    another application/unit, or setting your application data without being the leader.)
+    This is raised when either trying to set a value to something that isn't a string,
```
benhoyt, line 2652:
```suggestion
    This is raised either when trying to set a value to something that isn't a string,
```

ops/model.py
```diff
@@ -2662,9 +2660,9 @@ class RelationDataTypeError(RelationDataError):
 
 
 class RelationDataAccessError(RelationDataError):
-    """Raised by ``Relation.data[entity][key] = value`` if you don't have access.
+    """Raised by ``Relation.data[entity][key] = value`` if sufficient access is not available.
```
benhoyt, line 2663: Perhaps simplify/directify:

```suggestion
    """Raised by ``Relation.data[entity][key] = value`` if unable to access.
```

ops/testing.py
```diff
@@ -149,7 +149,7 @@ class ExecResult:
 class Harness(Generic[CharmType]):
     """This class represents a way to build up the model that will drive a test suite.
 
-    The model created is from the viewpoint of the charm that you are testing.
+    The model created is from the viewpoint of the charm that being tested.
```
benhoyt, line 152:
```suggestion
    The model created is from the viewpoint of the charm that is being tested.
```

ops/testing.py
```diff
@@ -339,18 +338,17 @@ class TestCharm(self._charm_cls):  # type: ignore
         self._charm = TestCharm(self._framework)  # type: ignore
 
     def begin_with_initial_hooks(self) -> None:
-        """Called when you want the Harness to fire the same hooks that Juju would fire at startup.
+        """Have the Harness fire the same hooks that Juju would fire at startup.
```
benhoyt, line 341: Nice simplification. Probably could even do this (as all of these are things the `Harness` does):

```suggestion
        """Fire the same hooks that Juju would fire at startup.
```

ops/testing.py
```diff
@@ -619,9 +616,9 @@ def disable_hooks(self) -> None:
     def enable_hooks(self) -> None:
         """Re-enable hook events from charm.on when the model is changed.
 
-        By default hook events are enabled once you call :meth:`.begin`,
-        but if you have used :meth:`.disable_hooks`, this can be used to
-        enable them again.
+        By default hook events are enabled once :meth:`.begin` is called,
```
benhoyt, line 619:
```suggestion
        By default, hook events are enabled once :meth:`.begin` is called,
```

ops/testing.py
```diff
@@ -619,9 +616,9 @@ def disable_hooks(self) -> None:
...
+        but if :meth:`.disable_hooks` is used, this can be used to enable
+        them again.
```
benhoyt, line 621:
```suggestion
        but if :meth:`.disable_hooks` is used, this method will enable
        them again.
```

benhoyt (changes requested):
Excellent work. I've gone through it and left a bunch of comments, mostly very minor, though there is one confusion about unit vs charm -- feel free to discuss if my suggestion of "current unit" is more confusing.

ops/testing.py
tonyandrewmeyer: I had this originally, actually, and I second-guessed myself that perhaps it was important to keep the harness reference there, even though you're in the harness class. I like this better :)

ops/charm.py
tonyandrewmeyer: It seemed to me that the original request wasn't just getting rid of you/your but where possible moving to imperative language. "Typically, fetch the ..." is imperative, whereas "Typically, the charm will fetch the ..." is more descriptive. No problem changing it as you describe, but just wanted to clarify this first :)

ops/framework.py
tonyandrewmeyer:
```suggestion
        For example, if handling a config change requires that two configs are
        changed, there's no reason to defer the first ``config-changed``
        because there will be a *second* ``config-changed`` event fired when
        the other config changes. Instead, check whether the config has changed
        prior to other events where the config is relevant.
```

Is this better? The last phrase in the original isn't super clear to me, so I'm not sure if the last sentence in the above is correct.

ops/charm.py
benhoyt: I agree we can/should move to more imperative language where possible. Here, however, I think it's less clear with just "Typically, fetch..." it's unclear who/what is doing the fetching: ops, the charm, something else? So I think the clarity here is helpful.

ops/framework.py
benhoyt: Yeah, that's a lot better. I think we can actually just leave the last sentence off.

One nit: let's say "config values" rather than "configs". So "two configs" -> "two config values" and "other config changes" -> "other config value changes".

ops/model.py
```diff
@@ -453,8 +452,8 @@ def _calculate_expiry(expire: Optional[Union[datetime.datetime, datetime.timedel
 class Unit:
     """Represents a named unit in the model.
 
-    This might be your unit, another unit of your application, or a unit of another application
-    that you are related to.
+    This might be this charm's unit, another unit of the charm's application, or a unit of
```
benhoyt, line 455: Another instance where "current unit" applies:

```suggestion
    This might be the current unit, another unit of the charm's application, or a unit of
```

ops/model.py
```diff
@@ -909,11 +907,11 @@ def bind_address(self) -> Optional[Union[ipaddress.IPv4Address, ipaddress.IPv6Ad
     @property
     def ingress_address(self) -> Optional[
             Union[ipaddress.IPv4Address, ipaddress.IPv6Address, str]]:
-        """The address other applications should use to connect to your unit.
+        """The address other applications should use to connect to the charm's unit.
```
benhoyt, line 910:
```suggestion
        """The address other applications should use to connect to the current unit.
```

benhoyt (changes requested):
Just two more "current unit" nits.

benhoyt (approved):

benhoyt:
Merged! Just docstring changes. If @jnsgruk wants further changes we can always put up another PR.

PR #1004, opened 2023-09-20
ci: use Go 1.20 for real Pebble tests

tonyandrewmeyer:
Pebble uses Go as of canonical/pebble#/171, so we need to adjust the ops CI to also do so for the real Pebble tests.


benhoyt (approved):

benhoyt:
Thanks! Trivial fix to CI -- merging without further review.

Issue #918, opened 2023-03-07
Improve open_port / close_port docs


benhoyt:
The docstrings for [`open_port`](https://github.com/canonical/operator/blob/eb8e25a5314ca2063f4689fa829c8bdca652ce1f/ops/model.py#L605) and `close_port` have a bunch of Juju detail that probably shouldn't be here, and some of it isn't correct -- for example, the bit about "registering intent ... juju expose" is true for VM charms but not for K8s sidecar charms.

I think what we want to do here is improve our [Juju docs about open-port](https://juju.is/docs/sdk/hook-tool#heading--open-port) and then link to those, to avoid documenting too much Juju behaviour in `ops` docstrings.

erik78se:
I'm trying to use this feature in "ops" but I don't know how to 

1. Detect that the "juju expose" or "juju unexpose"  has happened. Special events hooks?
2. How to use it, apart from what is obvious from the code - which isn't ideal.

benhoyt:
@erik78se Sure, in the meantime, I'll try to answer those here:

1. I don't think there's a way to detect that, as that's an administrator concern, not a charm concern. The charm opens the port or ports, but then -- on virtual machine models -- the admin has to run `juju expose` to actually expose those port(s) to the outside world. For K8s models, the `juju expose` step isn't required (though on K8s ingress works differently, so the ports may or may not be exposed to the outside world, depending on how the K8s cluster networking is set up -- once again, admin-dependent).
2. Using `open_port` in charms is fairly straight-forward:

```python
class MyCharm(CharmBase):
    def _on_some_event(self, event):
        self.unit.open_port('tcp', 8080)
```

PR #1005, opened 2023-09-20
feat: add Unit.set_ports() for declarative port opening

### LINKED ISSUE #989: Add method for "wholesale" port management
#905 added convenient API for "deltas" in port management, and it works well:

```python
    def set_ports(self):
        """Open necessary (and close no longer needed) workload ports."""
        planned_ports = {
            OpenedPort("tcp", self._ports.api),
            OpenedPort("tcp", self._ports.ha),
        }
        actual_ports = self.unit.opened_ports()

        # Ports may change across an upgrade, so need to sync
        ports_to_close = actual_ports.difference(planned_ports)
        for p in ports_to_close:
            self.unit.close_port(p.protocol, p.port)

        new_ports_to_open = planned_ports.difference(actual_ports)
        for p in new_ports_to_open:
            self.unit.open_port(p.protocol, p.port)
```

It seems like the same function would be repeated in every COS Lite charm (and others?), so I was wondering if it would make sense to add something like this "wholesale" approach to port management to ops.

tonyandrewmeyer:
Adjustments to port management to provide a more declarative style and some additional brevity in charm code.

* Rename `OpenPort` to `Port` to better reflect that it's a protocol+port and the state of the port is changed by calling one of the three relevant methods. Keep `OpenPort` as an alias for backwards compatibility.
* New `Unit.set_ports()` method that takes an arbitrary number of ports, either ints (implying 'tcp' as the protocol) or `Port`s, that will open any ports in the provided list that are not already open, and close any ports that are already open but are not in the provided list.
* ~`Unit.open_port()` and `Unit.close_port()` default to `tcp` for the protocol, allowing, for example, `unit.open_port(port=8000)`~

Fixes #989


test/test_model.py
```diff
@@ -3391,17 +3395,44 @@ def test_opened_ports_warnings(self):
...
+        self.unit.set_ports(8000, 8025)
+        calls = fake_script_calls(self, clear=True)
+        self.assertEqual(calls.pop(0), ['opened-ports', ''])
+        # We make no guarantee on the order the ports are opened.
+        calls.sort()
+        self.assertEqual(calls, [
+            ['open-port', '8000/tcp'],
+            ['open-port', '8025/tcp'],
+        ])
+        fake_script(self, 'opened-ports', 'echo 8025/tcp')
```
jameinel, line 3422: just for test consistency, it seems like `opened-ports` should return both ports, not just one of them. Otherwise, the 'side effect' of the first calls isn't actually represented.

test/test_model.py
```diff
@@ -3391,17 +3395,44 @@ def test_opened_ports_warnings(self):
...
+            ['close-port', '8025/tcp'],
+            ['open-port', '8022/udp'],
+        ])
+        fake_script(self, 'opened-ports', 'echo 8022/udp')
+        self.unit.set_ports()
+        self.assertEqual(fake_script_calls(self, clear=True), [
+            ['opened-ports', ''],
+            ['close-port', '8022/udp'],
+        ])
+
```
jameinel, line 3435: So you have tested that you can open ports that haven't been declare,d and that you can replace existing ports with a new set of ports,and that you can close all ports, 

but you haven't tested that if a port is open and you call set_ports it is a no-op
or if there is a set of open ports that set_ports doesn't touch the ones that are the same. (eg, currently open are a,b and you ask for b,c should trigger a close of a, nothing for b, and an open of c)

jameinel:
BTW, great to see your first PR show up.

ops/model.py
```diff
@@ -606,6 +606,9 @@ def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
         However, normally charms should make the same open_port() call from
         every unit.
 
+        Use :func:`set_ports()` for a more declarative approach where all of
+        the ports that should be open are provided in a single call.
+
```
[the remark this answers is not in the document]
sed-i: Maybe this could also be communicated with [`typing.overload`](https://docs.python.org/3/library/typing.html#typing.overload).

ops/model.py
```diff
@@ -633,13 +640,47 @@ def close_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
...
         """Return a list of opened ports for this unit."""
         return self._backend.opened_ports()
 
+    def set_ports(self, *ports: Union[int, 'Port']) -> None:
+        """Set the open ports for this unit, closing any others that are open.
+
+        Register intent with Juju that the application should be accessed on
+        the given ports and no others, but the ports are not actually opened
+        externally until the admin runs "juju expose" and any closed ports are
+        not actually closed externally until the admin runs "juju unexpose".
```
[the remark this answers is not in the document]
tonyandrewmeyer: The [juju docs](https://juju.is/docs/sdk/hook-tool#heading--open-port) say that there's no immediate effect (and this also matches the ops `open_port` and `close_port` docs):

> open-port will not have any effect if the application is not exposed, and may have a somewhat delayed effect even if it is.

I took a look at the juju code to see whether the docs are outdated but hit my Go knowledge limit. I'll check with someone :)

ops/model.py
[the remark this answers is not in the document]
benhoyt: I'm pretty sure the `open-port` hook tool will give a decent error message -- but that might not map back to the Python arguments very well. So we might want to explicitly check and `raise TypeError` in `Unit.open_port` / `Unit.close_port`.

We could also add a `typing.overload` if it doesn't add too much complexity and if the docs still look sane.

ops/model.py
```diff
@@ -633,13 +640,47 @@ def close_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
         """
         self._backend.close_port(protocol.lower(), port)
 
-    def opened_ports(self) -> Set['OpenedPort']:
+    def opened_ports(self) -> Set['Port']:
         """Return a list of opened ports for this unit."""
         return self._backend.opened_ports()
 
+    def set_ports(self, *ports: Union[int, 'Port']) -> None:
```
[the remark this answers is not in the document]
benhoyt: I don't think that including "open" or "opened" in the name adds much, but interested to hear @tonyandrewmeyer's thoughts as a newcomer. Most of the other names in model.py are punchy-ish.

ops/model.py
```diff
@@ -606,6 +606,9 @@ def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
         However, normally charms should make the same open_port() call from
         every unit.
 
+        Use :func:`set_ports()` for a more declarative approach where all of
```
benhoyt, line 677: It doesn't seem to matter to Sphinx, but elsewhere we've used `:meth:` instead of `:func:` and dropped the parens on the name, so let's do that for consistency:
```suggestion
        Use :meth:`set_ports` for a more declarative approach where all of
```

tonyandrewmeyer:
> BTW, great to see your first PR show up.

Technically, 4th :wink: but first one in operator with code, yes :tada:. Feels great to be moving on from onboarding towards providing value :smile:

ops/model.py
```diff
@@ -593,7 +593,7 @@ def add_secret(self, content: Dict[str, str], *,
             owner='unit')
         return Secret(self._backend, id=id, label=label, content=content)
 
-    def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
+    def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'] = 'tcp',
```
benhoyt, line 664: Seems like at this point it might be good to add a couple of examples to the docstring, one with the default protocol and one explicit.

```
For example::

    self.unit.open_port('udp', 4000)  # open UDP port 4000
    self.unit.open_port(port=8080)    # open TCP port 8080 (protocol is implicit)
```

Hmm, I'm having second thoughts about this API addition. Seems like we're adding too many ways to do it, and `open_port(port=8080)` looks a bit odd with the word "port" appearing twice, and is barely shorter than `open_port('tcp', 8080)` in any case. So maybe we should guide people to the new `set_ports` API but keep this one simple and explicit. Let's discuss.

ops/model.py
```diff
@@ -606,40 +674,161 @@ def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
...
+        # Normalise to get easier comparisons.
+        existing_ports = {
+            (port.protocol, port.port)
+            for port in self._backend.opened_ports()
+        }
+        desired_ports = {
+            ('tcp', port) if isinstance(port, int) else (port.protocol, port.port)
+            for port in ports
+        }
+        for protocol, port in existing_ports.difference(desired_ports):
```
benhoyt, line 824: Do you think `in existing_ports - desired_ports` is clearer (and similar below)? I kinda think it is, don't have strong feelings though.

ops/model.py
```diff
@@ -606,40 +674,161 @@ def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
...
+
+        Use :func:`open_port()` and :func:`close_port()` to manage ports
+        individually.
+
+        Args:
+            ports: The ports to open. Provide an int to open a TCP port, or
+                a :class:`Port` to open a port for another protocol.
+        """
+        # Normalise to get easier comparisons.
+        existing_ports = {
```
benhoyt, line 816: Naming nit: this method is all about ports, so we can drop the `_ports` in these local variable names.

ops/model.py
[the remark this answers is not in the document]
tonyandrewmeyer: Ah, I wrote this before seeing your comment @benhoyt (TIL that "draft" has a bit of different meaning in Canonical than I'm used to :laughing:).

I first raised RuntimeError, and then noticed that ops.testing.Harness will raise ModelError, so I changed to that to be consistent. My inclination is that TypeError is best for `open_port('tcp')` but `open_port()` is more of a RuntimeError. However, I see that TypeError is raised in, for example, `get_secret` if the wrong argument combination is passed, so it would be consistent there. Overall, I would still generally prefer consistency with ops.testing.

> We could also add a typing.overload if it doesn't add too much complexity and if the docs still look sane.

I already did this too. It looks fairly sane I think:

image

ops/__init__.py
```diff
@@ -138,6 +138,7 @@
     'Network',
     'NetworkInterface',
     'OpenedPort',
+    'Port',
```
benhoyt, line 141: Thanks for remembering to add it here!

ops/model.py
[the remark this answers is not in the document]
benhoyt: Per voice discussion, after consideration we're going to stick with `set_ports`.

ops/model.py
benhoyt: Per voice discussion, we're going to back out this change -- doesn't add much (except adding one more way to do things!). And we'll focus people on the new, simpler `set_ports` API instead.

ops/model.py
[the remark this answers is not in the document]
tonyandrewmeyer: Discussed with @benhoyt: I don't like "opened" and generally prefer the brevity (which is also consistent with the rest of ops.model) given that there isn't any expectation of setting anything else with ports.

ops/model.py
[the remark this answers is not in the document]
tonyandrewmeyer: As discussed elsewhere in the PR, we decided to drop the change to have protocol default to 'tcp'. I've kept the checking that a port is not provided for icmp and is provided for tcp/udp, since that seems useful outside of the default.

@benhoyt do you think it's worth keeping the (simplified) overload? The docs look like this now:

image

Basically, this means that if you do `open_port('icmp', 8000)` or `open_port('tcp')` your type checker will complain rather than it causing an error at run time. If you'd prefer to keep the docs simpler it's easy to drop those out as well.

test/test_model.py
tonyandrewmeyer: The test is setting a fresh state (via `fake_script`) each time rather than trying to tell a single story from start to finish (it did start out that way, but I found I was doing extra work just to have a single story, so stripped that out). If one story all the way through the test is preferred, I can adjust to that.

ops/model.py
tonyandrewmeyer: Discussed and we agree that it's a lot of complication to save two characters (plus the ugly double "port"), so I've stripped that out.

ops/model.py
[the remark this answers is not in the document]
benhoyt: I don't mind this typing improvement. However:

1) Let's avoid repeating the docstring (it only needs to be on the last one -- see `Container.exec`). The body can be `...` for the others.
2) Let's just make sure we've tested that it type checks with both Pyright and Mypy. We probably need to enable type checking in our tests at some point, but for now, you can just test it manually.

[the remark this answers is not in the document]
tonyandrewmeyer: > 1. Let's avoid repeating the docstring (it only needs to be on the last one -- see `Container.exec`). The body can be `...` for the others.

When this is the case, the linting fails, because it's a public method without a docstring (it's what I originally did). Should I remove the docstring and add a pragma to tell the linter to ignore it?

> 2. Let's just make sure we've tested that it type checks with both Pyright and Mypy. We probably need to enable type checking in our tests at some point, but for now, you can just test it manually.

Sure. I've checked with pyright but will check mypy too.

ops/model.py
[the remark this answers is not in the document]
tonyandrewmeyer: See #918

ops/model.py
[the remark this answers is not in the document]
benhoyt: Ah yes, I guess that's what the `# noqa` comment is for in the `Container.exec` variants. That seems vastly preferable over repeating the docstring -- thanks.

tonyandrewmeyer:
Note: will wait for #918 to be merged before moving this out of draft, so that I can get in the open/close port doc changes and align them with set_port.

benhoyt:
I've just merged the #1006 open/close_port docs tweak.

ops/model.py
```diff
@@ -603,17 +603,30 @@ def open_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
...
+                or ``port`` is not provided when ``protocol`` is 'tcp' or
+                'udp'.
         """
-        self._backend.open_port(protocol.lower(), port)
+        normalised_protocol : str = protocol.lower()
+        if normalised_protocol == 'icmp' and port is not None:
+            raise ModelError("icmp cannot have a port number specified")
+        elif normalised_protocol in ('tcp', 'udp') and port is None:
+            raise ModelError(f"{normalised_protocol} must have a port number specified")
+        self._backend.open_port(normalised_protocol, port)
```
benhoyt, line 626: Sorry if I'd muddied the waters earlier, but I think we should drop this code change (but we can keep the "Raises: ModelError" docstring change). This checking is automatically provided by the hook tool, which will return an error and `ModelError` will be raised, so I don't think we should duplicate those checks here.

The `testing._TestingModelBackend.open_port` function also does this check, but that's to simulate the hook tool.

Same comment for `close_port`.

1
benhoyt, line 1: Test file accidentally committed?

test/test_model.py
```diff
@@ -3337,6 +3337,11 @@ def test_open_port_error(self):
             ['open-port', '8080/ftp'],
         ])
 
+        with self.assertRaises(ops.ModelError) as cm:
+            self.unit.open_port('icmp', 8000)
+        with self.assertRaises(ops.ModelError) as cm:
+            self.unit.open_port('udp')
+
```
benhoyt, line 3344: We can probably drop these tests too once we've dropped the extra logic in `open_port`. We don't need to test the hook tool error handling itself, and we've already tested above that if it returns an error we raise `ModelError`.

test/test_model.py
```diff
@@ -3391,17 +3401,64 @@ def test_opened_ports_warnings(self):
...
 
+    def test_set_ports(self):
+        # No existing ports, open new ones.
+        fake_script(self, 'open-port', 'exit 0')
+        fake_script(self, 'close-port', 'exit 0')
+        fake_script(self, 'opened-ports', 'exit 0')
+        self.unit.set_ports(8000, 8025)
+        calls = fake_script_calls(self, clear=True)
+        self.assertEqual(calls.pop(0), ['opened-ports', ''])
+        calls.sort()  # We make no guarantee on the order the ports are opened.
```
benhoyt, line 3423: Nice observation!

benhoyt (changes requested):
Looking good. Just one more change requested: let's drop the redundant checks in `open_port` -- or discuss if my reasoning doesn't make sense.

test/test_model.py
```diff
@@ -3391,17 +3391,64 @@ def test_opened_ports_warnings(self):
...
+        self.unit.set_ports(ops.Port('udp', 8022), 8028)
+        self.assertEqual(fake_script_calls(self, clear=True), [
+            ['opened-ports', ''],
+            ['close-port', '8025/tcp'],
+            ['open-port', '8022/udp'],
+        ])
+        # Completely replace the opened ports.
+        fake_script(self, 'opened-ports', 'echo 8025/tcp; echo 8028/tcp')
+        self.unit.set_ports(8001, 8002)
+        calls = fake_script_calls(self, clear=True)
```
jameinel, line 3429: not your issue here, but I do feel like "fake_script_calls(self, clear=True)" is more common than 'clear=False', I sort of wish we used a different default.

test/test_model.py
```diff
@@ -3391,17 +3391,64 @@ def test_opened_ports_warnings(self):
...
+        fake_script(self, 'close-port', 'exit 0')
+        fake_script(self, 'opened-ports', 'exit 0')
+        self.unit.set_ports(8000, 8025)
+        calls = fake_script_calls(self, clear=True)
+        self.assertEqual(calls.pop(0), ['opened-ports', ''])
+        calls.sort()  # We make no guarantee on the order the ports are opened.
+        self.assertEqual(calls, [
+            ['open-port', '8000/tcp'],
+            ['open-port', '8025/tcp'],
+        ])
```
jameinel, line 3417: Given that we aren't preserving state between the calls (the value of fake_script doesn't match the result of the previous open/close)
I wonder about having these as multiple named tests
test_set_ports_initial
test_set_ports_close_all
test_set_ports_open_and_close_overlapping
test_set_ports_non_overlapping

it isn't something I would block on, but given you wanted to comment on each section, and the state of the previous section doesn't affect the later section, those feel like separate test cases to me.

jameinel (approved):

test/test_model.py
benhoyt: @tonyandrewmeyer Looks like there are 23 without clear=True and 69 with clear=True, so clearing is definitely the more common. If you want, feel free to open another PR which switches this default.

ops/model.py
```diff
@@ -620,23 +628,62 @@ def close_port(self, protocol: typing.Literal['tcp', 'udp', 'icmp'],
...
-    def opened_ports(self) -> Set['OpenedPort']:
+    def opened_ports(self) -> Set['Port']:
         """Return a list of opened ports for this unit."""
         return self._backend.opened_ports()
 
+    def set_ports(self, *ports: Union[int, 'Port']) -> None:
+        """Set the open ports for this unit, closing any others that are open.
+
+        Some behaviour, such as whether the port is opened or closed externally without
+        using Juju's ``expose`` & ``unexpose`` commands, differs between Kubernetes
```
benhoyt, line 657:
```suggestion
        using Juju's ``expose`` and ``unexpose`` commands, differs between Kubernetes
```

Issue #1007, opened 2023-09-21
Run Pyright on test files to catch typing issues


benhoyt:
We currently don't run Pyright over our `test/*.py` files to type-check those, meaning we're not getting "coverage" of most of our type annotations, which results in issues like https://github.com/canonical/operator/issues/963.

We should enable Pyright on these files. I suspect this will be a fairly big diff -- if so, we may want to opt the files in one by one in separate PRs to streamline review.

PR #1008, opened 2023-09-22
test: add type hints to the test_infra tests

tonyandrewmeyer:
Minor type hinting to make pyright happy:

```
$ env PYTHONPATH=.tox/static/lib/python3.10/site-packages/ .tox/static/bin/pyright test/test_infra.py 

```

Also renames an unused variable to `_` to get rid of an IDE warning.

Partial fix for #1007.

### DIFF
--- a/test/test_infra.py
+++ b/test/test_infra.py
@@ -19,19 +19,20 @@ import re
 import subprocess
 import sys
 import tempfile
+import typing
 import unittest
 
 import ops
 
 
-def get_python_filepaths(include_tests=True):
+def get_python_filepaths(include_tests: bool = True):
     """Helper to retrieve paths of Python files."""
     python_paths = ['setup.py']
     roots = ['ops']
     if include_tests:
         roots.append('test')
     for root in roots:
-        for dirpath, dirnames, filenames in os.walk(root):
+        for dirpath, _, filenames in os.walk(root):
             for filename in filenames:
                 if filename.endswith(".py"):
                     python_paths.append(os.path.join(dirpath, filename))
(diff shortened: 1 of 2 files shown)


test/test_infra.py
```diff
@@ -42,7 +43,7 @@ class InfrastructureTests(unittest.TestCase):
 
     def test_quote_backslashes(self):
         # ensure we're not using unneeded backslash to escape strings
-        issues = []
+        issues : typing.List[typing.Tuple[str, int, str]] = []
```
benhoyt, line 46: I thought it was odd that Flake8 didn't complain about the [PEP8 pet peeve](https://www.python.org/dev/peps/pep-0008/#pet-peeves) of whitespace before a colon, but I see we have `E203` ignored in our `pyproject.toml`. Odd -- not sure why. Can we please un-ignore that and fix these? From a quick check, the only instances are the ones in this file. Needs to be done in the tool.autopep8 and tool.flake8 section.

benhoyt (changes requested):
Thanks! I think we should also add "test/test_infra.py" to the `pyproject.toml` `tool.pyright.include` list to keep that up to date and checking the files we've done.

test/test_infra.py
tonyandrewmeyer: > I thought it was odd that Flake8 didn't complain about the [PEP8 pet peeve](https://www.python.org/dev/peps/pep-0008/#pet-peeves) of whitespace before a colon, but I see we have `E203` ignored in our `pyproject.toml`. Odd -- not sure why. Can we please un-ignore that and fix these? From a quick check, the only instances are the ones in this file. Needs to be done in the tool.autopep8 and tool.flake8 section.

Good, I don't like that either and was just relying on the tooling :)

benhoyt (approved):
Excellent, thanks!

PR #1009, opened 2023-09-22
test: add type hints to test_log tests

tonyandrewmeyer:
Minor type hinting to make pyright happy:

```
$ env PYTHONPATH=.tox/static/lib/python3.10/site-packages/ .tox/static/bin/pyright test/test_log.py 

```

Most significant change is that the fake model backend now inherits from the real one.

Partial fix for #1007.

### DIFF
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -16,6 +16,7 @@
 
 import io
 import logging
+import typing
 import unittest
 from unittest.mock import patch
 
@@ -26,15 +27,15 @@ from ops.model import MAX_LOG_LINE_LEN, _ModelBackend
 class FakeModelBackend:
 
     def __init__(self):
-        self._calls = []
+        self._calls : typing.List[typing.Tuple[str, str]] = []
 
-    def calls(self, clear=False):
+    def calls(self, clear: bool = False):
         calls = self._calls
         if clear:
             self._calls = []
         return calls
 
-    def juju_log(self, level, message):
+    def juju_log(self, level: str, message: str):
         for line in _ModelBackend.log_split(message):
             self._calls.append((level, line))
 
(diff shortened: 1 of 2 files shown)


test/test_log.py
```diff
@@ -48,7 +49,7 @@ def tearDown(self):
         logging.getLogger().handlers.clear()
 
     def test_default_logging(self):
-        ops.log.setup_root_logging(self.backend)
+        ops.log.setup_root_logging(self.backend)  # type: ignore
```
benhoyt, line 52: Hmm, yeah, this is a bit annoying. I suppose ideally `setup_root_logging` and the like should take a `typing.Protocol`, but that's it's own messiness.

Could we just make `FakeModelBackend` inherit from `_ModelBackend` and remove these type-ignores?

tonyandrewmeyer: > Hmm, yeah, this is a bit annoying. I suppose ideally `setup_root_logging` and the like should take a `typing.Protocol`, but that's it's own messiness.

Yeah, it seemed like quite a bit of code for little gain.

> Could we just make `FakeModelBackend` inherit from `_ModelBackend` and remove these type-ignores?

Sure. I wondered if you'd rather keep the stub minimal, but happy to go with this.

benhoyt: > keep the stub minimal

Yeah, I wondered that, but if any of the other methods are called they'll fail hard because the hook tool is missing, so I think this is fine. Thanks.

benhoyt (approved):
Great, thanks!

tonyandrewmeyer:
@benhoyt do you want a second reviewer for this, or are type hint fixes too minimal? If you do, could you suggest someone?

benhoyt (approved):
Looks good. Nah, just me for review is fine on these low-impact typing fixes.

tonyandrewmeyer:
@benhoyt Resolved the pyproject conflict.

PR #1010, opened 2023-09-22
test: fix type hint warnings in test_private tests

tonyandrewmeyer:
It looks like there's an accidental double-call in these two tests. They still pass, since the ValueError isn't caught in the second call, but that means that the second call isn't doing anything at all. It would also be weird if the code changed and allowed the values being tested because instead of the test failing in the expected way (the assertRaises failing) it would fail because a TypeError would be raised instead of ValueError.

This also makes pyright happy because we're not (theoretically) passing a datetime.datetime when a string is expected.

test_private is otherwise fine:

```
$ env PYTHONPATH=.tox/static/lib/python3.10/site-packages/ .tox/static/bin/pyright test/test_private.py 

```

Partial fix for #1007

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,9 @@ per-file-ignores = ["test/*:D100,D101,D102,D103,D104"]
 docstring-convention = "google"
 
 [tool.pyright]
-include = ["ops/*.py", "ops/_private/*.py"]
+include = ["ops/*.py", "ops/_private/*.py",
+    "test/test_private.py",
+]
 pythonVersion = "3.8" # check no python > 3.8 features are used
 pythonPlatform = "All"
 typeCheckingMode = "strict"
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Good spotting, thanks.

I reckon we should just combine the addition of typing to test_private into this PR, as it's part of that, and to avoid tooo many tiny PRs.

tonyandrewmeyer:
> I reckon we should just combine the addition of typing to test_private into this PR, as it's part of that, and to avoid tooo many tiny PRs.

Nothing else seems needed for test_private (although I've added it into the list in pyproject.toml).

benhoyt (approved):

tonyandrewmeyer:
That rtd fail looks transient. Will re-trigger.

benhoyt (approved):

PR #1011, opened 2023-09-22
test: fix type hinting on test_jujuversion tests

### LINKED ISSUE #1007: Run Pyright on test files to catch typing issues

tonyandrewmeyer:
JujuVersion's comparisons can be done against strings as well as other JujuVersions, so the type hints should reflect that.

For mocking the os.environ response, the type is dict(str, str) but I think to declare that we would need to define the dict, e.g:

```
    _mockenv : typing.Dict[str, str] = {}
    @unittest.mock.patch("os.environ", new=_mockenv)
```

And that looks considerably messier to my eyes and doesn't seem to really add value (so I've ignored the warning).

Partially fixes #1007

### DIFF
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -71,7 +71,7 @@ class JujuVersion:
             and self.build == other.build
             and self.patch == other.patch)
 
-    def __lt__(self, other: 'JujuVersion') -> bool:
+    def __lt__(self, other: Union[str, 'JujuVersion']) -> bool:
         if self is other:
             return False
         if isinstance(other, str):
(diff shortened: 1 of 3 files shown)


benhoyt (approved):
LGTM. Agreed the typing for patch/new is messy! `new={'': ''}` is shorter but confusing. So I think `type: ignore` is reasonable here.

Let's just do the thing I mentioned in the other PR about adding the files to pyproject.toml as we go so that's up to date. When we finish, we can replace the individual names with `test/*.py`.

Maybe break `include` onto multiple lines to avoid so many conflicts.

benhoyt (approved):

PR #1012, opened 2023-09-24
test: add type hinting to test_lib

tonyandrewmeyer:
Extends the type hinting for tests:

* Stop testing the debug level log output (this also removes the need for the logassert package)
* Pass a dummy spec object rather than None to avoid type errors
* Ignore type checks when the purpose of the test is to test what happens when bad types are used
* Minor type hinting where required for pyright to be happy

Partially addresses #1007.

### DIFF
--- a/ops/lib/__init__.py
+++ b/ops/lib/__init__.py
@@ -23,6 +23,7 @@ import logging
 import os
 import re
 import sys
+import typing
 import warnings
 from ast import literal_eval
 from importlib.machinery import ModuleSpec
@@ -117,7 +118,7 @@ def autoimport():
         versions.sort(reverse=True)
 
 
-def _find_all_specs(path):
+def _find_all_specs(path: typing.Iterable[str]) -> typing.Iterator[ModuleSpec]:
     for sys_dir in path:
         if sys_dir == "":
             sys_dir = "."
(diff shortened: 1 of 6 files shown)


benhoyt (approved):
Very good work, and thanks for the removal of `logassert` as discussed. No need for this to be a draft PR I think.

benhoyt:
Looks like there's one failing test on Python 3.8, presumably due to `assertNoLogs` being added later. So I guess we'll have to avoid that method.

tonyandrewmeyer:
@benhoyt this failed in CI because unittest.TestCase.assertNoLogs is new in 3.10. Would you just get rid of the check that there's no warning logged, or do a temporary but ugly `if hasattr(self, "assertNoLogs")` or check that it doesn't happen in a different way?

tonyandrewmeyer:
> Very good work, and thanks for the removal of `logassert` as discussed. No need for this to be a draft PR I think.

I drafted it when I saw it failed in CI, to hopefully (but not successfully) avoid you bothering with it before I fixed the issue.

benhoyt:
> I changed it to draft when I saw it failed in CI, to hopefully (but not successfully) avoid you bothering with it before I fixed the issue.

Ah, don't worry about that.

> Would you just get rid of the check that there's no warning logged, or do a temporary but ugly if hasattr(self, "assertNoLogs") or check that it doesn't happen in a different way?

I think if it's worth doing the test it's worth doing it on Python 3.8 too (a version we need to support). So let's change it to use `assertLogs` or similar approach that works on 3.8+.

benhoyt (approved):

PR #1013, opened 2023-09-25
ci: re-enable integration tests with mysql-*operator

### LINKED ISSUE #987: Fix mysql-operator charm test now that it uses Poetry for deps
The mysql-operator tests are failing since https://github.com/canonical/mysql-operator/pull/192 was merged, which removes the `requirements.txt` file from that repo and puts dependencies in `pyproject.toml`. Presumably they'll be doing the same with the other db charms soon too, so we need to find a workaround. [Example failure:](https://github.com/canonical/operator/actions/runs/5746095549/job/15575066294)

```
> sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
> echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
> shell: /usr/bin/bash -e {0}
sed: can't read requirements.txt: No such file or directory
```

tonyandrewmeyer:
* Only try to edit `requirements.txt` if it exists
* If `requirements.txt` does not exist, then assume that the requirement is in `pyproject.toml` and edit that instead
* Enable the integration tests with `mysql-operator` and `mysql-k8s-operator` again
* Remove the `static-charm` tox runs from the DB workflow, as none of the DB operators have one defined
* Pin `tox` to anything compatible with 4.2

Note that bash/sed is not a personal strength, so there could well be a nicer way to do this.

Fixes #987.


.github/workflows/db-charm-tests.yaml
```diff
@@ -11,20 +11,32 @@ jobs:
...
         with:
           repository: ${{ matrix.charm-repo }}
 
+      - name: Checkout the operator repository
+        uses: actions/checkout@v3
+        with:
+          path: myops
+
+      - name: Install patch dependencies
+        run: pip install poetry==1.6.1
```
benhoyt, line 29: Can we broaden this to any 1.x version, I think the syntax is `poetry~=1.6` (but double-check that).

.github/workflows/db-charm-tests.yaml
```diff
@@ -11,20 +11,32 @@ jobs:
...
     steps:
       - name: Checkout the ${{ matrix.charm-repo }} repository
         uses: actions/checkout@v3
         with:
           repository: ${{ matrix.charm-repo }}
 
+      - name: Checkout the operator repository
+        uses: actions/checkout@v3
+        with:
+          path: myops
```
benhoyt, line 26: Are we 100% sure this will checkout the changes from this PR (specifically, the temp merge of this PR branch into `main`) and not the latest on `main`?

.github/workflows/db-charm-tests.yaml
```diff
@@ -11,20 +11,32 @@ jobs:
...
       - name: Update 'ops' dependency in test charm to latest
         run: |
-          sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
-          echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
+          if [ -e "requirements.txt" ]; then
+            sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
+            echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
+          else
+            sed -i -e "s/^ops[ ><=].*/ops = {path = \"myops\"}/" pyproject.toml
+            poetry lock
```
benhoyt, line 38: Yep, I like how this turned out -- thanks!

benhoyt (commented):
Looks good to me -- one question, and one minor tweak requested on the poetry version.

.github/workflows/db-charm-tests.yaml
tonyandrewmeyer: Is there a reason we pin tox to specifically 4.2.8? If not, could we broaden both? (I know there's no 4.2 beyond 4.2.8 and it's unlikely that there ever will be, but it would set a pattern for when we eventually upgrade tox).

.github/workflows/db-charm-tests.yaml
tonyandrewmeyer: > # The branch, tag or SHA to checkout. When checking out the repository that
> # triggered a workflow, this defaults to the reference or SHA for that event.
> # Otherwise, uses the default branch.

[source](https://github.com/actions/checkout)

.github/workflows/db-charm-tests.yaml
benhoyt: Yeah, let's broaden both. In other tests we just `pip install tox`, which is probably okay for a CLI tool, though arguably `~=4.2` would be best.

.github/workflows/db-charm-tests.yaml
benhoyt: Thanks!

benhoyt (approved):

benhoyt:
Thanks! CI improvements, merging without further review.

PR #1014, opened 2023-09-26
test: add type hinting to test_helpers

tonyandrewmeyer:
* Add type hints to tests/test_helpers
* For fake_script and fake_script_calls, there are a bunch of `type: ignore`s to work around not being able to state the type of `fake_script_path` - I think there will be a few of these through other test modules as well, as hints are added there. I feel it would be cleaner to have the fake script functionality in a class rather than dynamically added to TestCase instances, but [that is a much more substantial change](https://github.com/tonyandrewmeyer/operator/commit/46238eee2d0b39d383f2ea6b3d6cac722be5dddd)

Partially addresses #1007


benhoyt (approved):
Huh, yeah, some of these are definitely messy. Oh well, if it helps us get type checks turned on on our test directory, I think it's worth it.

PR #1015, opened 2023-09-27
test: add type hints to test_model

### LINKED ISSUE #1007: Run Pyright on test files to catch typing issues

tonyandrewmeyer:
* One change to ops/charm, but only in the type hints. It looks from the use of `Required` that the other fields are not meant to be required, so make that the case (this also appears to match the code).
* Ignore types in the tests that are checking that bad types are handled properly.
* Help type checkers understand when some optional attributes won't be None.
* Replace a few bits of code where a minimal object was passed, enough for the test to do it's check, but not of the correct type, with an actual object of the proper type.
* Sprinkle type hints where required for pyright to understand what's happening.

Partially fixes #1007


test/test_model.py
```diff
@@ -1988,7 +2072,21 @@ def setUp(self):
...
 
-    def _check_binding_data(self, binding_name, binding):
+    def model_get_relation(self, name: str = 'db1', relation_id: typing.Optional[int] = None):
+        """Wrapper around self.model.get_relation that enforces that None is not returned."""
+        rel_db1 = self.model.get_relation(name, relation_id)
+        self.assertIsNotNone(rel_db1)
+        assert rel_db1 is not None  # Type checkers understand this, but not the previous line.
+        return rel_db1
+
+    def model_get_binding(self, binding_key: typing.Union[str, ops.Relation]):
```
benhoyt, line 2082: Similar question about naming: what do you think about `ensure_binding`?

test/test_model.py
```diff
@@ -2450,35 +2548,39 @@ def test_local_get_status(self):
...
-            lambda: fake_script(self, 'storage-add', 'echo fooerror >&2 ; exit 1'),
-            lambda: self.backend.storage_add('foobar', count=True),
-            TypeError,
-            [],
-        )]
+        test_cases: typing.List[typing.Tuple[
+            typing.Callable[[], None],
+            typing.Callable[[], None],
+            typing.Type[BaseException],
+            typing.List[typing.List[str]]]] = [(
```
benhoyt, line 2558: Not your issue, but IMO this block would be a whole lot simpler (and even shorter!) just written out the dumb way:

```python
fake_script(self, 'storage-list', 'echo fooerror >&2 ; exit 1')
with self.assertRaises(ModelError):
    self.backend.storage_list('foobar')
self.assertEqual(fake_script_calls(self, clear=True),
                 [['storage-list', 'foobar', '--format=json']])
```

Maybe you could make that change and it would avoid the messy type annotation too?

test/test_model.py
```diff
@@ -2657,8 +2759,13 @@ def test_juju_log(self):
                          [['juju-log', '--log-level', 'BAR', '--', 'foo']])
 
     def test_valid_metrics(self):
+        if typing.TYPE_CHECKING:
+            _caselist = typing.List[typing.Tuple[
+                typing.Mapping[str, typing.Union[int, float]],
+                typing.Mapping[str, str],
+                typing.List[typing.List[str]]]]
```
benhoyt, line 2766: Similar here: clearer and simpler to just write out the test cases. Again, not your issue, but might as well clean it up and avoid the messy type annotation.

I don't mind case-based tests (what Go calls "table-driven tests"), but I think they should only be used when there are really a lot of cases and the case-based approach is clearer.

test/test_model.py
```diff
@@ -2673,7 +2780,9 @@ def test_valid_metrics(self):
             self.assertEqual(fake_script_calls(self, clear=True), expected_calls)
 
     def test_invalid_metric_names(self):
-        invalid_inputs = [
+        invalid_inputs: typing.List[typing.Tuple[
+            typing.Dict[str, float],
+            typing.Dict[typing.Any, typing.Any]]] = [
```
benhoyt, line 2785: I think this is a case where case-based tests *do* improve things: the cases are short and there are a lot of them.

But maybe we can cheat with types here and just use `typing.Any` -- I think the type signature is just getting in our way here.

test/test_model.py
```diff
@@ -2692,7 +2801,9 @@ def test_invalid_metric_names(self):
                 self.backend.add_metrics(metrics, labels)
 
     def test_invalid_metric_values(self):
-        invalid_inputs = [
+        invalid_inputs: typing.List[typing.Tuple[
+            typing.Dict[str, typing.Union[float, str]],
+            typing.Dict[typing.Any, typing.Any]]] = [
```
benhoyt, line 2806: Ditto here: `typing.Any`.

test/test_model.py
```diff
@@ -2889,17 +3000,21 @@ def test_unit_add_secret_args(self):
...
         # Additional add_secret tests are done in TestApplication
-        errors = [
-            ({'xy': 'bar'}, {}, ValueError),
-            ({'foo': 'x'}, {'expire': 7}, TypeError),
-        ]
+        errors: typing.List[typing.Tuple[
+            typing.Any,
+            typing.Dict[str, typing.Any],
+            typing.Type[BaseException]]] = [
+            ({'xy': 'bar'}, {}, ValueError), ({'foo': 'x'}, {'expire': 7}, TypeError), ]
```
benhoyt, line 3007: Ditto here: `typing.Any` (and revert formatting to one case per line).

test/test_model.py
```diff
@@ -2889,17 +3000,21 @@ def test_unit_add_secret_args(self):
...
             with self.assertRaises(exc_type, msg=msg):
-                self.unit.add_secret(content, **kwargs)
+                self.unit.add_secret(content, **kwargs)  # type: ignore
 
     def test_add_secret_errors(self):
-        errors = [
+        errors: typing.List[typing.Tuple[
+            typing.Any,
+            typing.Dict[str, typing.Any],
+            typing.Type[BaseException]]] = [
```
benhoyt, line 3017: Ditto here: `typing.Any`.

test/test_model.py
```diff
@@ -3059,7 +3174,11 @@ class TestSecretClass(unittest.TestCase):
     def setUp(self):
         self.model = ops.Model(ops.CharmMeta(), _ModelBackend('myapp/0'))
 
-    def make_secret(self, id=None, label=None, content=None):
+    def make_secret(self,
+                    id: typing.Optional[str] = None,
+                    label: typing.Optional[str] = None,
+                    content: typing.Optional[typing.Dict[str,
+                                                         str]] = None):
```
benhoyt, line 3181: Just wondering: do we have our tooling set up wrongly that it's doing awkward things like this? Is the formatter line break set to less than the linter line break or something?

```suggestion
                    content: typing.Optional[typing.Dict[str, str]] = None):
```

test/test_model.py
```diff
@@ -3228,43 +3347,67 @@ def test_set_info(self):
...
-        secret.grant(types.SimpleNamespace(id=234), unit=types.SimpleNamespace(name='app/0'))
+        backend = ops.model._ModelBackend('test', 'test', 'test')
+        meta = ops.CharmMeta()
+        cache = ops.model._ModelCache(meta, backend)
+        unit = ops.Unit('test', meta, backend, cache)
+        rel123 = ops.Relation('test', 123, True, unit, backend, cache)
+        rel234 = ops.Relation('test', 234, True, unit, backend, cache)
+        secret.grant(rel123)
+        unit = ops.Unit('app/0', meta, backend, cache)
+        secret.grant(rel234, unit=unit)
```
benhoyt, line 3363: Seeing we're probing into internals anyway, I wonder if we should just reuse the existing backend and cache objects?

```suggestion
        secret = self.make_secret(id='x')
        unit = ops.Unit('test', ops.CharmMeta(), self.model._backend, self.model._cache)
        rel123 = ops.Relation('test', 123, True, unit, self.model._backend, self.model._cache)
        rel234 = ops.Relation('test', 234, True, unit, self.model._backend, self.model._cache)
        secret.grant(rel123)
        unit = ops.Unit('app/0', ops.CharmMeta(), self.model._backend, self.model._cache)
        secret.grant(rel234, unit=unit)
```

benhoyt (commented):
Nice work! Left a bunch of suggestions to simplify / clean a few things up.

test/test_model.py
```diff
@@ -65,24 +66,34 @@ def setUp(self):
         self.harness._get_backend_calls(reset=True)
         self.model = self.harness.model
 
+    def model_get_relation(
```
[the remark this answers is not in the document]
tonyandrewmeyer: Sure. I was originally aiming for the change to be minimal (`model.get_relation` becoming `model_get_relation`) but a clearer name is better, I agree.

benhoyt (approved):
Thanks for the updates! I won't merge this until you make a final decision on those last outstanding comments -- your call as to whether to use `typing.Any` or whatever there, or keep the annotations.

test/test_model.py
tonyandrewmeyer: Agreed with the two for secrets; too much noise. I do somewhat like the ones for the list of metric, label pairs. I've cleaned it up a bit by defining an alias and just using that consistently. Work for you?

PR #1017, opened 2023-09-28
test: add type hinting to test_testing

tonyandrewmeyer:
One minor (type) change to ops.testing: the harness `pull()` signature is changed to allow `encoding==None`, which matches the actual `pebble.Client` method.

Otherwise, changes are in test_testing:
* Explicitly ignore types where we are testing invalid type behaviour
* Use `assert`s to show type checkers that `obj|None` is always `obj` in a few places - this seems better tn tests than `cast(obj_type, obj)` and pyright doesn't understand that this is the case with `assertIsNotNone(obj)`
* Remove two `print()` calls, one of which seems to be a debugging print that slipped in, and one that's not required although the expression being printed is
* Replace a few `dict(a=b, c=d)` with `{'a': b, 'c': d}` as pyright seems to understand the latter much more easily (plus it's insignificantly faster and generally more recommended style)
* Fix `test_push_with_permission_mask` where the exception's `kind` attribute was (presumably in error) only being tested for one of the test cases
* In the `_PebbleStorageAPIsTestMixin` class make it clear (especially for type checkers, but also in general) that a `client` attribute must be provided by the eventual class
* In the `_PebbleStorageAPIsTestMixin` class make it clear to type checkers that various TestCase methods will be provided (via mixing with `unittest.TestCase`)
* Sprinkle type hints where required for pyright to be happy

Partially addresses #1007

### DIFF
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -5031,7 +5031,7 @@ class TestHandleExec(unittest.TestCase):
     def test_register_with_result(self):
         self.harness.handle_exec(self.container, ["foo"], result=10)
 
-        with self.assertRaises(pebble.ExecError[str]) as exc:
+        with self.assertRaises(pebble.ExecError) as exc:
             self.container.exec(["foo"]).wait()
         self.assertEqual(exc.exception.exit_code, 10)
 
@@ -5182,13 +5182,12 @@ class TestHandleExec(unittest.TestCase):
             "group-id": 2,
             "environment": {"foo": "bar", "foobar": "barfoo"}
         }
-        layer = ops.pebble.Layer(
-            ops.pebble.LayerDict(
-                summary="",
-                description="",
-                services={
-                    "test": service}))
-        self.container.add_layer(label="test", layer=layer)
+        layer: ops.pebble.LayerDict = {
+            'summary': "",
+            'description': "",
+            'services': {
+                "test": service}}
+        self.container.add_layer(label="test", layer=ops.pebble.Layer(layer))
         args_history: typing.List[ops.testing.ExecArgs] = []
 
         def handler(args: ops.testing.ExecArgs):
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -5031,9 +5031,9 @@ class TestHandleExec(unittest.TestCase):
     def test_register_with_result(self):
         self.harness.handle_exec(self.container, ["foo"], result=10)
 
-        with self.assertRaises(pebble.ExecError) as exc:  # type: ignore
+        with self.assertRaises(pebble.ExecError[str]) as exc:
             self.container.exec(["foo"]).wait()
-        self.assertEqual(exc.exception.exit_code, 10)  # type: ignore
+        self.assertEqual(exc.exception.exit_code, 10)
 
         self.harness.handle_exec(self.container, ["foo"], result="hello")
         stdout, stderr = self.container.exec(["foo"]).wait_output()
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -56,8 +56,7 @@ class SetLeaderErrorTester(ops.CharmBase):
 
     def _on_leader_elected(self, event: ops.EventBase):
         peers = self.model.get_relation(self._peer_name)
-        if typing.TYPE_CHECKING:
-            assert peers is not None
+        assert peers is not None
         peers.data[self.app]["foo"] = "bar"
 
 
@@ -500,8 +499,7 @@ class TestHarness(unittest.TestCase):
         self.assertEqual(backend.relation_list(rel_id), ['postgresql/0'])
         harness.charm.get_changes(reset=True)  # ignore relation created events
         relation = harness.charm.model.get_relation('db')
-        if typing.TYPE_CHECKING:
-            assert relation is not None
+        assert relation is not None
         self.assertEqual(len(relation.units), 1)
         # Check relation data is correct
         rel_unit = harness.charm.model.get_unit('postgresql/0')
@@ -516,8 +514,7 @@ class TestHarness(unittest.TestCase):
         self.assertEqual(backend.relation_list(rel_id), [])
         # Check the unit is actually removed from the relations the model knows about
         rel = harness.charm.model.get_relation('db')
-        if typing.TYPE_CHECKING:
-            assert rel is not None
+        assert rel is not None
         self.assertEqual(len(rel.units), 0)
         self.assertFalse(rel_unit in rel.data)
         # Check relation departed was raised with correct data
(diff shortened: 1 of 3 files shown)


test/test_testing.py
```diff
@@ -4002,7 +4053,41 @@ def test_send_signal(self):
...
+    if typing.TYPE_CHECKING:
+        # Provide these in classes using this mixin, generally by mixing with a TestCase.
+        def assertEqual(  # noqa
+                self,
+                first: typing.Any,
+                second: typing.Any,
+                msg: typing.Optional[str] = None):
+            ...
+
+        _cm_type = typing.ContextManager[unittest.case._AssertRaisesContext[BaseException]]
```
tonyandrewmeyer, line 4070: @benhoyt I'm not completely certain that this is the right type for the return value of `assertRaises` when used as a context manager. I couldn't find much stdlib documentation on typing context managers, and web searches had a bunch of dated content that didn't seem to help much.

This _does_ work, in that pyright (and whatever VS Code is using, maybe also pyright) seems ok with it, but I'm not convinced that the `cm` in `with self.assertRaises(MyException) as cm:` is of the right type. In addition, some attributes are only set in `__exit__` (I assume, I didn't read the code), so, for example, `cm.exception` only exists after the `with` is complete - which is why I've also had to do `cm.exception = cast(ExceptionType, cm.exception)` where this gets used.

If you're more familiar with this or know someone who is or somewhere good to read up on it, I'd love some pointers :)

test/test_testing.py
```diff
@@ -2965,40 +3002,55 @@ def test_add_network_endpoint_and_relation_id_do_not_correspond(self):
...
             data = event.relation.data[event.unit]
         else:
-            data = event.relation.data[event.app]
+            app = event.app
+            assert app is not None
+            data = event.relation.data[app]
         self.changes.append(dict(data))
 
 
+if typing.TYPE_CHECKING:
```
benhoyt, line 3037: Any reason these need to be in `if typing.TYPE_CHECKING` blocks? I think we should avoid it when possible for simplicity and consistency, but I'm open to being convinced otherwise.

test/test_testing.py
```diff
@@ -4002,7 +4053,41 @@ def test_send_signal(self):
 class _PebbleStorageAPIsTestMixin:
     # Override this in classes using this mixin.
     # This should be set to any non-empty path, but without a trailing /.
-    prefix = None
+    prefix: str = None  # type: ignore
```
benhoyt, line 4056: Is the `= None` required here and below? I think you can drop the `= None` and the `type: ignore`.

test/test_testing.py
benhoyt: Hmm, no, this is well outside my area of expertise. For reference, [this is how typeshed (type hints for the stdlib) annotates that](https://github.com/python/typeshed/blob/6074a8f5441f7e6cb833409f11ed315598fa67a3/stdlib/unittest/case.pyi#L146).

However, all this is a bit messy. I tried making the mixin class also inherit from `unittest.TestCase`, but the method-resolution-order algorithm doesn't like that. Can we skirt the issue and drop this messiness by doing this?

```python
class _PebbleStorageAPIsTestMixin:
    assertEqual = unittest.TestCase.assertEqual
    assertRaises = unittest.TestCase.assertRaises
    assertIsInstance = unittest.TestCase.assertIsInstance
    assertIn = unittest.TestCase.assertIn
```

I think then we can also drop the cast calls like:

```python
        cm.exception = typing.cast(pebble.PathError, cm.exception)
```

test/test_testing.py
```diff
@@ -4016,16 +4101,45 @@ def test_push_and_pull_non_utf8_data(self):
             encoding='sjis',
             stream_class=io.StringIO)
 
-    def _test_push_and_pull_data(self, original_data, encoding, stream_class):
+    if typing.TYPE_CHECKING:
+        @typing.overload
```
benhoyt, line 4105: I reckon we drop these overloads on this internal function. They don't seem to be necessary with the casts you have below.

test/test_testing.py
```diff
@@ -4640,6 +4775,7 @@ def test_set_secret_content(self):
         self.assertEqual(len(harness.charm.events), 1)
         event = harness.charm.events[0]
         self.assertIsInstance(event, ops.SecretChangedEvent)
+        assert isinstance(event, ops.SecretChangedEvent)  # Dupes previous line for type checkers.
```
benhoyt, line 4778: Per comment on earlier PR, I think we should *just* include the second one, which works for both tests and type checkers. Here and below.

test/test_testing.py
```diff
@@ -4936,9 +5086,9 @@ def test_register_match_all_prefix(self):
     def test_register_with_result(self):
         self.harness.handle_exec(self.container, ["foo"], result=10)
 
-        with self.assertRaises(pebble.ExecError) as exc:
+        with self.assertRaises(pebble.ExecError) as exc:  # type: ignore
```
benhoyt, line 5089: Why are these two `type: ignore`s necessary? I wouldn't expect them to be needed here, as this is correctly-typed API usage.

test/test_testing.py
```diff
@@ -5083,12 +5243,16 @@ def test_exec_service_context(self):
...
             "environment": {"foo": "bar", "foobar": "barfoo"}
         }
-        self.container.add_layer(label="test", layer={
-            "summary": "", "description": "", "services": {"test": service}
-        })
-        args_history = []
-
-        def handler(args):
+        layer = ops.pebble.Layer(
+            ops.pebble.LayerDict(
```
benhoyt, line 5247: I think we should revert this one to a plain dict, which is expected and valid usage. The `LayerDict` is really more for type checking the fields, not instantiation.

benhoyt (commented):
Excellent work, thanks! Just a few minor comments.

test/test_testing.py
tonyandrewmeyer: Nice, didn't think of that.

test/test_testing.py
tonyandrewmeyer: No, I was just trying to match the style elsewhere in the code. Happy to drop them.

test/test_testing.py
tonyandrewmeyer: Ah, indeed. I added those first and then had to do the other changes anyway, and you're right that they are no longer required. Thanks!

test/test_testing.py
tonyandrewmeyer: Nice suggestion, thanks!

I'm super happy to be rid of those cast calls, I hated those. And I'm pretty sure (even more now that they are not required) that they were only needed because I had something wrong in the return type.

test/test_testing.py
tonyandrewmeyer: Ah my bad, I meant to go back to this one. I initially thought it was going to be tricky to cleanly solve, but all it needed was to provide the exact type in the exception. Thanks!

tonyandrewmeyer: Actually, I thought I had figured this out, but I made the type checking happy and the test sad :(

The difference here is that `ExecError` is a Generic, so we need to show that `exc` is a `ExecError[str]`. The type comes from the context manager, which uses the type provided in the `assertRaises` call, but I'm not sure how to make that a specific type. I'll noodle on this a bit more.

tonyandrewmeyer: Hmm, actually pyright is ok with it as-is, it's just Visual Studio that is unhappy. So I guess it's fine then?

benhoyt (approved):

PR #1018, opened 2023-09-29
test: separate the real pebble tests into a separate module

### LINKED ISSUE #688: Move real pebble tests into their own file.
This will help make it clear that those tests are special and not vanilla unit tests - they run together in a special environment where the pebble daemon is running.

tonyandrewmeyer:
Move the tests that use a real pebble server into a separate module, to make it clearer that these are different from the other pebble tests.

Fixes #688.


test/test_real_pebble.py
```diff
@@ -0,0 +1,276 @@
...
+import threading
+import time
+import unittest
+import urllib.error
+import urllib.request
+import uuid
+
+from ops import pebble
+
+from .test_testing import _PebbleStorageAPIsTestMixin
```
tonyandrewmeyer, line 29: @benhoyt how do you feel about this? Would you rather the mixin lost the leading underscore? Are you ok with cross-module importing in the tests, rather than moving it into a separate module that both the test modules import from?

.github/workflows/framework-tests.yaml
```diff
@@ -86,7 +86,7 @@ jobs:
           PEBBLE: /tmp/pebble
 
       - name: Run Real pebble tests
-        run: tox -e unit -- -k RealPebble
+        run: tox -e unit -- test/test_real_pebble
```
benhoyt, line 89: `.py`?

test/test_real_pebble.py
benhoyt: I think the import is fine, but yeah, let's drop the `_`.

.github/workflows/framework-tests.yaml
tonyandrewmeyer: Ha, yes, sorry. I realised I did that and searched for it to fix it up where I had it wrong, and forgot I'd need a `--hidden` for `rg` :joy:

benhoyt (approved):
LGTM now, thanks!

benhoyt:
Small CI/test reorg -- merging without further review.

PR #1023, opened 2023-10-02
test: add type hints for test_storage

tonyandrewmeyer:
* Ignore the `None`s passed when creating the framework - passing real values would be a lot of extra noise without actually gaining any value in the test
* Add dummy methods to make the `Serializable` protocol match
* Ignore types when replacing a (when type checking) property
* Inherit from StoredStateData rather than Object to match the type of the `Framework.save_snapshot` method - the additional methods seem harmless in this case, and the meaning of the test seems unchanged
* Sprinkle enough type hints so that pyright can tell what's happening

Partially addresses #1007

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ include = ["ops/*.py", "ops/_private/*.py",
     "test/test_lib.py",
     "test/test_model.py",
     "test/test_testing.py",
+    "test/test_storage.py",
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
LGTM, thanks!

benhoyt:
Minor typing additions to test files, merged without further review.

PR #1024, opened 2023-10-02
fix: push_path and pull_path include empty directories

### LINKED ISSUE #968: pull_path doesn't retrieve empty directories
Per [this comment](https://github.com/canonical/operator/pull/960#issuecomment-1628001631), `Container.pull_path` doesn't retrieve empty directories. This seems like a bug, not a feature (Git's behaviour notwithstanding). From a quick look at the code, I think `push_path` will have a similar problem.

We probably need to change the `_list_recursive` helper to also yield directories (before recursing), and the callers test if it's a directory and create it, otherwise copy the file.

tonyandrewmeyer:
Instead of just recursively iterating through the remote/local files, also iterate through the directories as well, ensuring that they are created even if they don't contain any files.

Also create/update relevant test cases.

Fixes #968

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2275,6 +2275,9 @@ class Container:
             try:
                 for info in Container._list_recursive(local_list, source_path):
                     dstpath = self._build_destpath(info.path, source_path, dest_dir)
+                    if info.type is pebble.FileType.DIRECTORY:
+                        self.make_dir(dstpath, make_parents=True)
+                        continue
                     with open(info.path) as src:
                         self.push(
                             dstpath,
@@ -2352,6 +2355,9 @@ class Container:
             try:
                 for info in Container._list_recursive(self.list_files, source_path):
                     dstpath = self._build_destpath(info.path, source_path, dest_dir)
+                    if info.type is pebble.FileType.DIRECTORY:
+                        dstpath.mkdir(parents=True, exist_ok=True)
+                        continue
                     dstpath.parent.mkdir(parents=True, exist_ok=True)
                     with self.pull(info.path, encoding=None) as src:
                         with dstpath.open(mode='wb') as dst:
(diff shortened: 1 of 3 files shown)


ops/model.py
```diff
@@ -2275,6 +2275,9 @@ def local_list(source_path: Path) -> List[pebble.FileInfo]:
             try:
                 for info in Container._list_recursive(local_list, source_path):
                     dstpath = self._build_destpath(info.path, source_path, dest_dir)
+                    if info.type is pebble.FileType.DIRECTORY:
```
benhoyt, line 2278: It wasn't clear to me that enum values should be compared by identity (`is` vs `==`), but it looks like they should be -- TIL! For reference: https://docs.python.org/3/howto/enum.html#comparisons

test/test_model.py
```diff
@@ -981,28 +981,32 @@ def __init__(self,
...
+        self.want_dirs = want_dirs or set()
 
 
 recursive_list_cases = [
     PushPullCase(
         name='basic recursive list',
         path='/',
         files=['/foo/bar.txt', '/baz.txt'],
-        want={'/foo/bar.txt', '/baz.txt'},
+        want={'/foo/bar.txt', '/baz.txt', '/foo'},
```
benhoyt, line 1003: Super nit, but I think it'd read more clearly for `/foo` to be listed first here. I realize it doesn't make a difference for set comparison, but it seems like it should come before `/foo/bar.txt`. Same below.

test/test_testing.py
```diff
@@ -4595,16 +4597,14 @@ def test_pull(self):
     def test_pull_path(self):
         (self.root / "foo").mkdir()
         (self.root / "foo/bar").write_text("bar")
-        # TODO: pull_path doesn't pull empty directories
-        # https://github.com/canonical/operator/issues/968
-        # (self.root / "foobar").mkdir()
```
benhoyt, line 4600: Nice catch!

benhoyt (approved):
Looks good to me, thanks! Just one super-nit comment.

weiiwang01 (approved):

PR #1025, opened 2023-10-02
test: add type hints to test_framework

tonyandrewmeyer:
* Change the type of `Handle`'s `key` (in `__init__` and `nest()` and the property) to explicitly allow `None
* Add a CHANGES.md file to document changes (backfilled only to 2.7)
* Tidy up the type hinting of `BoundStoredState.__getattr__` - multiple overloads need to be provided, and once that's done the ignores can be removed (and a bunch of type issues in the tests get resolved)
* Ignore type where creating an object that's only partially complete, but the test only requires that much and fully creating it would be a lot of extra code for no real gain
* Where's it's simple to do, pass the required attributes rather than `None`
* Ignore types where we're checking handling of invalid types
* Add casts around the very dynamic snapshot/restore functionality
* For a couple of the "list of test cases" situations, move the documentation of what each element in the test case is to a new type definition that covers both what was there before and the (rough) expected type
* Sprinkle type hints where required for pyright to be happy

Partially addresses #1007

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -101,7 +101,7 @@ class Handle:
     under the same parent and kind may have the same key.
     """
 
-    def __init__(self, parent: Optional[Union['Handle', 'Object']], kind: str, key: str):
+    def __init__(self, parent: Optional[Union['Handle', 'Object']], kind: str, key: Optional[str]):
         if isinstance(parent, Object):
             # if it's not an Object, it will be either a Handle (good) or None (no parent)
             parent = parent.handle
@@ -119,7 +119,7 @@ class Handle:
             else:
                 self._path = f"{kind}"  # don't need f-string, but consistent with above
 
-    def nest(self, kind: str, key: str) -> 'Handle':
+    def nest(self, kind: str, key: Optional[str]) -> 'Handle':
         """Create a new handle as child of the current one."""
         return Handle(self, kind, key)
 
@@ -143,7 +143,7 @@ class Handle:
         return self._kind
 
     @property
-    def key(self) -> str:
+    def key(self) -> Optional[str]:
         """Return the handle's key."""
         return self._key
 
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -1061,13 +1061,17 @@ class BoundStoredState:
 
     if TYPE_CHECKING:
         @typing.overload
-        def __getattr__(self, key: Literal['on']) -> ObjectEvents:  # type: ignore
+        def __getattr__(self, key: Literal['on']) -> ObjectEvents:
+            pass
+
+        @typing.overload
+        def __getattr__(self, key: str) -> Any:
             pass
 
     def __getattr__(self, key: str) -> Any:
         # "on" is the only reserved key that can't be used in the data map.
         if key == "on":
-            return self._data.on  # type: ignore  # casting won't work for some reason
+            return self._data.on
         if key not in self._data:
             raise AttributeError(f"attribute '{key}' is not stored")
         return _wrap_stored(self._data, self._data[key])
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -52,9 +52,9 @@ class TestFramework(BaseTestCase):
 
     def test_handle_path(self):
         cases = [
-            (ops.Handle(None, "root", None), "root"),
+            (ops.Handle(None, "root", 'test1'), "root"),
             (ops.Handle(None, "root", "1"), "root[1]"),
-            (ops.Handle(ops.Handle(None, "root", None), "child", None), "root/child"),
+            (ops.Handle(ops.Handle(None, "root", 'test2'), "child", 'test3'), "root/child"),
             (ops.Handle(ops.Handle(None, "root", "1"), "child", "2"), "root[1]/child[2]"),
         ]
         for handle, path in cases:
@@ -80,7 +80,7 @@ class TestFramework(BaseTestCase):
 
         handle = ops.Handle(None, "a_foo", "some_key")
 
-        framework.register_type(Foo, None, handle.kind)
+        framework.register_type(Foo, None, handle.kind)  # type: ignore
 
         try:
             framework.load_snapshot(handle)
@@ -92,6 +92,8 @@ class TestFramework(BaseTestCase):
 
     def test_snapshot_roundtrip(self):
         class Foo:
+            handle_kind = 'foo'
+
             def __init__(self, handle: ops.Handle, n: int):
                 self.handle = handle
                 self.my_n = n
@@ -107,19 +109,21 @@ class TestFramework(BaseTestCase):
 
         framework1 = self.create_framework(tmpdir=self.tmpdir)
         framework1.register_type(Foo, None, handle.kind)
-        framework1.save_snapshot(event)
+        framework1.save_snapshot(event)  # type: ignore
         framework1.commit()
         framework1.close()
 
         framework2 = self.create_framework(tmpdir=self.tmpdir)
         framework2.register_type(Foo, None, handle.kind)
         event2 = framework2.load_snapshot(handle)
+        event2 = typing.cast(Foo, event2)
         self.assertEqual(event2.my_n, 2)
 
-        framework2.save_snapshot(event2)
+        framework2.save_snapshot(event2)  # type: ignore
         del event2
         gc.collect()
         event3 = framework2.load_snapshot(handle)
+        event3 = typing.cast(Foo, event3)
         self.assertEqual(event3.my_n, 3)
 
         framework2.drop_snapshot(event.handle)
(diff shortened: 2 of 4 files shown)


ops/framework.py
```diff
@@ -1061,13 +1061,17 @@ def __init__(self, parent: Object, attr_name: str):
 
     if TYPE_CHECKING:
         @typing.overload
-        def __getattr__(self, key: Literal['on']) -> ObjectEvents:  # type: ignore
+        def __getattr__(self, key: Literal['on']) -> ObjectEvents:
+            pass
+
+        @typing.overload
+        def __getattr__(self, key: str) -> Any:
```
benhoyt, line 1068: Do you know why this overload is needed when it's the same types as the actual `def __getattr__` below? Is the main signature not used as a fallback when there are overloads?

test/test_framework.py
```diff
@@ -43,17 +44,18 @@ def setUp(self):
...
+            framework = ops.Framework(':memory:', None, None, None)  # type: ignore
         self.assertIn(
             "WARNING:ops.framework:deprecated: Framework now takes a Storage not a path",
             cm.output)
         self.assertIsInstance(framework._storage, SQLiteStorage)
 
     def test_handle_path(self):
         cases = [
-            (ops.Handle(None, "root", None), "root"),
+            (ops.Handle(None, "root", 'test1'), "root[test1]"),
```
benhoyt, line 55: Unless I'm missing something, I'm a bit nervous that we're changing / removing these test cases. If anything I think we should keep the existing ones (changing the typing shouldn't change them) and add a couple of new ones testing the additional cases.

test/test_framework.py
```diff
@@ -328,26 +338,26 @@ def test_custom_event_data(self):
         framework = self.create_framework()
 
         class MyEvent(ops.EventBase):
-            def __init__(self, handle, n):
+            def __init__(self, handle: ops.Handle, n: int):
                 super().__init__(handle)
                 self.my_n = n
 
             def snapshot(self):
```
benhoyt, line 345: Out of interest, why doesn't Pyright complain about this not having a return type signature?

test/test_framework.py
```diff
@@ -328,26 +338,26 @@ def test_custom_event_data(self):
...
                 return {"My N!": self.my_n}
 
-            def restore(self, snapshot):
+            def restore(self, snapshot: typing.Dict[str, int]):
```
benhoyt, line 348: Would it be clearer for this to be `Dict[str, Any]` to match the standard `restore` signature?

test/test_framework.py
```diff
@@ -866,24 +885,28 @@ def setUp(self):
...
 
     def test_stored_dict_repr(self):
-        self.assertEqual(repr(ops.StoredDict(None, {})), "ops.framework.StoredDict()")
+        self.assertEqual(repr(ops.StoredDict(None, {})),  # type: ignore
+                         "ops.framework.StoredDict()")
         self.assertEqual(
             repr(
                 ops.StoredDict(
-                    None, {
+                    None, {  # type: ignore
```
benhoyt, line 893: Not your issue, but the formatting here is weird! Seems like it should be:

```python
self.assertEqual(repr(ops.StoredDict(None, {"a": 1})),  # type: ignore
                 "ops.framework.StoredDict({'a': 1})")
```

Feel free to fix it in a drive-by here or in a separate commit.

test/test_framework.py
```diff
@@ -907,7 +930,7 @@ class SomeObject(ops.Object):
         class Sub(SomeObject):
             pass
 
-        class SubSub(SomeObject):
+        class SubSub(Sub):
```
benhoyt, line 933: Good catch!

test/test_framework.py
```diff
@@ -1084,12 +1113,19 @@ def test_mutable_types(self):
...
-            None,               # Operand B.
-            {},                 # Expected result.
-            lambda a, b: None,  # Operation to perform.
-            lambda res, expected_res: self.assertEqual(res, expected_res)  # Validation to perform.
+        test_case = typing.Tuple[
+            typing.Callable[[], typing.Any],                        # Called to get operand A.
+            typing.Any,                                             # Operand B.
+            typing.Any,                                             # Expected result.
+            typing.Callable[[typing.Any, typing.Any], None],        # Operation to perform.
+            typing.Callable[[typing.Any, typing.Any], typing.Any],  # Validation to perform.
```
benhoyt, line 1121: This is definitely one of those cases where IMO the 5-tuple of lambdas isn't helping clarity, and we should probably just spell these out as separate tests. The use of tuples with multiple statements and the need to use 
tricks like `a.__setitem__(1, b)` instead of `a[1] = b` is a bit of a giveaway. Probably just leave it as is for now though to avoid churn.

Thanks for adding the comment, that helps. :-)

Let's tweak the comment to just say "list of tuples" instead of "list of 2-tuples" though while we're here though. 2-tuples seems old/wrong.

test/test_framework.py
```diff
@@ -1239,19 +1275,26 @@ class SomeObject(ops.Object):
             _stored = ops.StoredState()
 
         class WrappedFramework(ops.Framework):
-            def __init__(self, store, charm_dir, meta, model, event_name):
+            def __init__(self,
+                         store: typing.Union[SQLiteStorage,
+                                             JujuStorage],
```
benhoyt, line 1280: Let's avoid the funky wrapping on the `store` and `charm_dir` annotations.

benhoyt (changes requested):
Looks good, thanks. A couple of questions and a few small changes requested.

test/test_framework.py
tonyandrewmeyer: A lot of the type it can infer the return type. Possibly there's also some `__init__` special-casing, in the same way that `self` gets special-cased to not need `: typing.Self`?

test/test_framework.py
tonyandrewmeyer: The `Handle.__init__` code is like this:

```python
    def __init__(self, parent: Optional[Union['Handle', 'Object']], kind: str, key: str):
        ...
        self._key = key
        ...
            if key:
                self._path = f"{parent}/{kind}[{key}]"
            else:
                self._path = f"{parent}/{kind}"
```

The docs say (my emphasis):

> The handle key is a **string** uniquely identifying the object.

The key is also used for hashing and equality checks (`None` is fine in either of those situations, but so are other non-string things). It's also a public member, and I don't have the experience to know how that's generally used, however, it does claim to be a string as well:

```python
    @property
    def key(self) -> str:
```

This seems messy to me. Shouldn't we either have the type signature and documentation match the behaviour and accept `str|None` (the docs could say that this *should not* be done), or have the implementation match the type signature and documentation and reject `None` as an invalid value for `key`?

test/test_framework.py
tonyandrewmeyer: Yes, good call; thanks!

test/test_framework.py
tonyandrewmeyer: > This is definitely one of those cases where IMO the 5-tuple of lambdas isn't helping clarity, and we should probably just spell these out as separate tests.

Agreed! I almost did this when I came across it, after we'd done something similar in one of the other test modules, and it just seemed like maybe it was a bit too much code to re-write in what would essentially be a drive-by.

> The use of tuples with multiple statements and the need to use tricks like `a.__setitem__(1, b)` instead of `a[1] = b` is a bit of a giveaway.

Yeah, and that "operand A" has to be a lambda to get the value but "operand B" is just the value, and even that all the lambdas don't return anything (other than None) and are pure side-effects.

> Probably just leave it as is for now though to avoid churn.

:+1:

ops/framework.py
benhoyt: Got it, thanks for the explanation.

test/test_framework.py
benhoyt: Hmm, yeah. From looking at the code, I think `key` should have been an `Optional[str]` all along -- the fact that is was typed as `str` was I think an oversight. The `if key` would handle empty string as well, but I think `None` would be the intention.

I've found one charm that passes `None` as the `key` argument to `Handle`.

So I think we should change `Handle`'s key arg to `Optional[str]`, and same for `nest`'s key argument. Those should be perfectly backwards-compatible changes, as we're expanding the type.

I also think we should change the `key` property's type to be `Optional[str]`. That isn't quite a backward-compatible change (typing-wise), because code that accesses the property will then have to `assert key is not None` to type check. But I've done a quick search and I can't see any charms using that property (it's a bit advanced/arcane to use `Handle` directly in the first place, let alone access `handle.key`).

So I think it's safe to change and we should do it. If you want, we could start a file `CHANGES.txt` that we use to record notes and breaking changes for the upcoming release, so we remember.

tonyandrewmeyer: Done - lmk if you'd prefer a particular format for the changes file.

benhoyt (approved):
All looks good, thanks!

PR #1028, opened 2023-10-03
test: add type hints to test_main

tonyandrewmeyer:
* Fix `_TestMain.test_multiple_events_handled`: this broke in f398e12897ab98a302a023f843efd3544e0c6a10
  * The last part of the test would never execute (the `if` was always false)
  * The departing unit's name is provided in a `departing_unit` attribute not `departing_unit_name`
  * It seems like at some point `state.__getitem__` was replaced with `state.__getattr__` as well, but I didn't dig into that.
  * Later, when the secret events were added, the attributes listed were not all correct (but since the test never run that part, this wasn't noticed)
  * Also when the secret events were added, no snapshot was added to the charm, so none of the data was available
* Fix the indentation in an `ops.Framework` docstring.
* Add type hints to the `test_main` charm. As part of this, I did a drive-by change to use "import ops" style, to avoid importing many different types of event.
* Add type hints in the smoke testing
* Ignore types where we are testing that the type is invalid
* When using the fake script functionality from test_helpers with an ABC, cast the case to a `unittest.TestCase` since we know that's what it will be at run time (this seems a bit ugly, but I think maybe the best choice without doing more significant changes)
* Ignore types in the `test_smoke` code where they are not provided by `pytest-operator` or `libjuju`.
* Sprinkle enough type hints around test_main to make pyright happy

Partially addresses #1007

### DIFF
--- a/test/smoke/test_smoke.py
+++ b/test/smoke/test_smoke.py
@@ -15,7 +15,6 @@
 # Learn more about testing at: https://juju.is/docs/sdk/testing
 
 import logging
-import typing
 
 from pytest_operator.plugin import OpsTest
 
@@ -31,11 +30,11 @@ async def test_smoke(ops_test: OpsTest):
 
     for series in ['focal', 'bionic', 'xenial']:
         model = ops_test.model
-        if typing.TYPE_CHECKING:
-            assert model is not None
+        assert model is not None
         app = await model.deploy(  # type: ignore
             charm, series=series, application_name=f"{series}-smoke")
         await model.wait_for_idle(timeout=600)  # type: ignore
 
+        assert app is not None
         assert_msg = f"Series {series} failed with '{app.status}' status"  # type: ignore
         assert app.status == "active", assert_msg  # type: ignore
--- a/test/smoke/test_smoke.py
+++ b/test/smoke/test_smoke.py
@@ -29,11 +29,8 @@ async def test_smoke(ops_test: OpsTest):
     charm = await ops_test.build_charm("./test/charms/test_smoke/")
 
     for series in ['focal', 'bionic', 'xenial']:
-        model = ops_test.model
-        assert model is not None
-        app = await model.deploy(
+        app = await ops_test.model.deploy(
             charm, series=series, application_name=f"{series}-smoke")
-        await model.wait_for_idle(timeout=600)
+        await ops_test.model.wait_for_idle(timeout=600)
 
-        assert app is not None
         assert app.status == "active", f"Series {series} failed with '{app.status}' status"
--- a/test/smoke/test_smoke.py
+++ b/test/smoke/test_smoke.py
@@ -26,15 +26,14 @@ async def test_smoke(ops_test: OpsTest):
 
     # Build the charm. (We just build it for focal -- it *should* work to deploy it on
     # older versions of Juju.)
-    charm = await ops_test.build_charm("./test/charms/test_smoke/")  # type: ignore
+    charm = await ops_test.build_charm("./test/charms/test_smoke/")
 
     for series in ['focal', 'bionic', 'xenial']:
         model = ops_test.model
         assert model is not None
-        app = await model.deploy(  # type: ignore
+        app = await model.deploy(
             charm, series=series, application_name=f"{series}-smoke")
-        await model.wait_for_idle(timeout=600)  # type: ignore
+        await model.wait_for_idle(timeout=600)
 
         assert app is not None
-        assert_msg = f"Series {series} failed with '{app.status}' status"  # type: ignore
-        assert app.status == "active", assert_msg  # type: ignore
+        assert app.status == "active", f"Series {series} failed with '{app.status}' status"
--- a/test/smoke/test_smoke.py
+++ b/test/smoke/test_smoke.py
@@ -15,6 +15,7 @@
 # Learn more about testing at: https://juju.is/docs/sdk/testing
 
 import logging
+import typing
 
 from pytest_operator.plugin import OpsTest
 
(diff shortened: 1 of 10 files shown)


test/test_main.py
```diff
@@ -585,9 +634,9 @@ def test_multiple_events_handled(self):
 
             self.assertEqual(list(state.observed_event_types), [event_spec.event_type.__name__])
 
-            if event_spec.event_name in expected_event_data:
-                self.assertEqual(state[f"{event_spec.event_name}_data"],
-                                 expected_event_data[event_spec.event_name])
+            if expected_event_data:
```
benhoyt, line 637: Why the change from `if event_spec.event_name in expected_event_data` to just `if expected_event_data`?

test/test_main.py
```diff
@@ -612,10 +661,10 @@ def test_empty_actions(self):
         self._simulate_event(EventSpec(ops.InstallEvent, 'install'))
 
     def test_collect_metrics(self):
-        fake_script(self, 'add-metric', 'exit 0')
+        fake_script(typing.cast(unittest.TestCase, self), 'add-metric', 'exit 0')
```
benhoyt, line 664: To avoid these annoying `typing.cast` calls (and the list of duplicated assert methods above), can we make `_TestMain` inherit from `unittest.TestCase`, and the subclasses not inherit?

test/test_main.py
```diff
@@ -696,36 +749,42 @@ def test_excepthook(self):
             '    raise RuntimeError."failing as requested".\n'
             'RuntimeError: failing as requested'
         )
-        self.assertEqual(len(calls), 1, f"expected 1 call, but got extra: {calls[1:]}")
+        self.assertEqual(
+            len(calls),
+            1,
+            f"expected 1 call, but got extra: {calls[1:]}")
```
benhoyt, line 755: A couple of these formatting changes seem strictly worse than what we had before. Any reason for this change and the assertRegex one above?

benhoyt (changes requested):
Looking good overall. A few comments / changes requested.

test/test_main.py
tonyandrewmeyer: Because the test did not work at all :). See the first bullet point in the PR description.

test/test_main.py
tonyandrewmeyer: I'm just using the tooling (ie. `tox -e fmt`) not choosing any of this myself. Can we adjust the tools so that it matches what you like?

test/test_main.py
benhoyt: Ah, that's that one -- thanks.

test/test_main.py
benhoyt: I don't quite understand that, because we were already running `tox -e fmt` on all of this code, and there were no changes on this line and similar lines (`tox -e lint` checks that the formatting is as per `tox -e fmt` would make it).

When I check out this PR, change this line back to what it was, and run `tox -e fmt`, it doesn't reformat it.

Is it possible you modified this line so it was >100 columns, formatted, then modified it back? Or is something else going on?

test/test_main.py
tonyandrewmeyer: Test bother me too :(. `_TestMain` can't inherit from `TestCase` because it's an ABC. Lots of tests start failing like:

```
ERROR test/test_main.py::_TestMain::test_no_actions - TypeError: Can't instantiate abstract class _TestMain with abstract methods _call_event, _setup_entry_point, test_set...
```

It needs to be an ABC because it has tests that can't run without context that the subclasses are providing.

It wouldn't help with the duplicated assert methods, but we could adjust the type of the `test_helpers` methods, for example:

```diff
diff --git a/test/test_helpers.py b/test/test_helpers.py
index ada79af..95f8f8f 100755
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -25,7 +25,8 @@ from ops.model import _ModelBackend
 from ops.storage import SQLiteStorage
 
 
-def fake_script(test_case: unittest.TestCase, name: str, content: str):
+def fake_script(test_case: typing.Any, name: str, content: str):
+    assert isinstance(test_case, unittest.TestCase)
     if not hasattr(test_case, 'fake_script_path'):
         fake_script_path = tempfile.mkdtemp('-fake_script')
         old_path = os.environ["PATH"]
```

This feels wrong to me - the caller should be making sure the type is correct - but maybe practicality beats purity since this would be one assert vs a whole heap of casts?

Alternatively, there's [making the script functionality part of the class via a mixin, which I did as an exercise to see how large it would be when doing the test_helpers typing](https://github.com/tonyandrewmeyer/operator/commit/46238eee2d0b39d383f2ea6b3d6cac722be5dddd). This is a much bigger diff but avoids all the casting needs.

One other possibility would be that I could make a protocol that defined what `fake_script` and `fake_script_calls` needs and make that the type - I think this could be done so that `TestCase` would match, as well as `_TestMain`. This seems complex for what it's doing, though :(.

Thoughts?

test/smoke/test_smoke.py
```diff
@@ -26,11 +26,15 @@ async def test_smoke(ops_test: OpsTest):
 
     # Build the charm. (We just build it for focal -- it *should* work to deploy it on
     # older versions of Juju.)
-    charm = await ops_test.build_charm("./test/charms/test_smoke/")
+    charm = await ops_test.build_charm("./test/charms/test_smoke/")  # type: ignore
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yes, that's the issue (well, they are partially typed, but not enough that we have everything we need).

But good point, will make that change.

test/test_main.py
tonyandrewmeyer: Hmm, that does do it. I don't recall, unfortunately :(. I worked on this module last week and then left it before picking it up this week, so it's not all totally fresh. Maybe I had a `type: ignore` and then removed that after in other modules we decided on the `assertEqual = unittest.TestCase.assertEqual` approach?

I'll clean up this PR and try to keep better track of this going forward.

test/test_main.py
benhoyt: Given all of the above, I think what you have now makes the most sense. It's only an issue for tests that do this kind of subclassing, not tests classes that inherit directly from `TestCase`, so it probably doesn't affect *most* usages.

At some point maybe we can clean some of this stuff up and lean into pytest more and use a fixture. But this is fine for now.

Thanks for your efforts on testing the mixing idea though!

benhoyt (approved):
LGTM now, thanks!

benhoyt:
Test/CI changes only, merged without further review.

PR #1029, opened 2023-10-04
docs: note that status changes are immediate

### LINKED ISSUE #767: docs: explain transactional nature of hooks/events
It's not clear from the docs how the juju "world" is viewed and interacted with.  For example, when setting status, changes within a single hook invocation are not realized to the juju world - the changes are not committed/public until the hook finishes (successfully).  Another concept is that the state of relation data and config is current as of the _start_ of an event hook.  There may be a few places where details like this are mentioned in passing (I couldn't find much), but we should probably make these concepts more prominent.  I've encountered questions along these lines several times (even had them myself).

benhoyt:
Fixes #767

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -332,6 +332,10 @@ class Application:
     def status(self) -> 'StatusBase':
         """Used to report or read the status of the overall application.
 
+        Changes to status take effect immediately, unlike other Juju operations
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):

benhoyt:
Docs tweak, merged without further review.

PR #1030, opened 2023-10-04
test: add type hints to pebble-related tests

### LINKED ISSUE #1007: Run Pyright on test files to catch typing issues

tonyandrewmeyer:
* Adds overloads for pebble's `pull` method to clarify what you get when you do/do not provide an encoding.
* `test/pebble_cli.py` has a bit of a workaround - the type returned by `exec` is generic, and I believe we could have pyright understand which one was being used, but by repeating the `exec` call in the various `if` statements above. It seems like this would make the code messier for no real gain, so I've cheated and told pyright it will be a `[str]` even though that may not be true.
* Add a few asserts to show pyright that values aren't None or are the correct type
* Add `type: ignore' where we are checking that types are invalid
* Sprinkle type hints as required for pyright to understand what's happening

Note that the latest version of pyright has a problem where we use code like this:

```python
with self.assertRaises(ops.pebble.ExecError) as cm:
    call_something()
self.assertEqual(cm.exception, some_value)
```

The issue is related to `ExecError` being generic combined with `cm` being generic and dynamic. The version of pyright we are currently using does not have an issue with this - but looking at the pyright release notes (1.1.318) it seems like accepting this is considered a bug and failing it in the latest version is correct. However, it's not clear what we are meant to do here, and the documentation (stdlib and pyright) is vague, so I'm leaving this to figure out (if necessary) when we upgrade our pyright version.

We now type check everything under `test/`, except for the smoke test (which needs to wait until pytest-operator and pylibjuju have more type hinting).

Fixes #1007


ops/pebble.py
```diff
@@ -1899,6 +1899,21 @@ def get_services(self, names: Optional[Iterable[str]] = None) -> List[ServiceInf
         resp = self._request('GET', '/v1/services', query)
         return [ServiceInfo.from_dict(info) for info in resp['result']]
 
+    if typing.TYPE_CHECKING:
```
benhoyt, line 1902: I think we should drop the `typing.TYPE_CHECKING` tests unless they are actually needed, for simplicity/consistency. I know we haven't been quite consistent in the past. Thoughts?

ops/pebble.py
```diff
@@ -1899,6 +1899,21 @@ def get_services(self, names: Optional[Iterable[str]] = None) -> List[ServiceInf
...
+                 encoding: None) -> BinaryIO:
+            ...
+
+        @typing.overload
+        def pull(self,  # noqa
+                 path: str,
+                 *,
+                 encoding: str = 'utf-8') -> TextIO:
+            ...
+
```
benhoyt, line 1916: Should we copy this to the model.Container versions too?

pyproject.toml
```diff
@@ -28,22 +28,7 @@ per-file-ignores = ["test/*:D100,D101,D102,D103,D104"]
...
-    "test/test_helpers.py",
-    "test/test_lib.py",
-    "test/test_model.py",
-    "test/test_testing.py",
-    "test/test_storage.py",
-    "test/test_charm.py",
-    "test/test_framework.py",
-    "test/test_main.py",
-]
+include = ["ops/*.py", "ops/_private/*.py", "test/*.py", "test/charms/*/src/*.py"]
```
benhoyt, line 31: Nice! End of an annoying road. :-)

test/test_pebble.py
```diff
@@ -94,7 +95,7 @@ def test_api_error(self):
 
     def test_change_error(self):
         change = pebble.Change(
-            id=pebble.ChangeID('1234'),
+            id=typing.cast(pebble.ChangeID, '1234'),
```
benhoyt, line 98: What's the reason for this change? If I change it back to `pebble.ChangeID('1234')` locally, Pyright is fine with that.

Similar below.

test/test_pebble.py
```diff
@@ -208,13 +211,13 @@ def test_warning_init(self):
         self.assertEqual(warning.repeat_after, '2s')
 
     def test_warning_from_dict(self):
-        d = {
+        d: pebble._WarningDict = {
```
benhoyt, line 214: Hmm, interesting. This is probably my bad, but because those `from_dict` methods are part of the public API, it seems to me like their types should be also. So perhaps we should remove the underscores from all of the following? (And move them to the top of pebble.py with the other public typed dicts.)

```
ChangeDict
CheckInfoDict
FileInfoDict
ProgressDict
ServiceInfoDict
SystemInfoDict
TaskDict
WarningDict
```

Thoughts?

benhoyt: Or maybe we should leave that till we have access to `Required` and `NotRequired`, without which these are kind of awkward?

test/test_pebble.py
```diff
@@ -985,11 +989,12 @@ def test_check_info(self):
         self.assertEqual(check.failures, 5)
         self.assertEqual(check.threshold, 3)
 
-        check = pebble.CheckInfo.from_dict({
+        d: pebble._CheckInfoDict = {
             'name': 'chk3',
             'status': 'up',
             'threshold': 3,
-        })
+        }  # type: ignore (no explicit level or failures)
```
benhoyt, line 996: Ah yes, this is a bit annoying, isn't it? I can see why, and if we specify `total=False` on `_CheckInfoDict` it causes a bunch of other type errors in pebble.py. Yeah, I think what you've done here is reasonable -- just recording my thoughts.

test/test_pebble.py
```diff
@@ -2617,14 +2660,16 @@ def setUp(self):
         time_patcher.start()
         self.addCleanup(time_patcher.stop)
 
-    def add_responses(self, change_id, exit_code, change_err=None):
+    def add_responses(self, change_id: str, exit_code: int,
+                      change_err: typing.Optional[str] = None):
```
benhoyt, line 2664: Curious that this doesn't need a return type annotation -- does Pyright infer that?

benhoyt (changes requested):
Thanks! A few comments and minor changes requested.

test/test_pebble.py
tonyandrewmeyer: My bad, sorry. I had added `ChangeID('blah')` (or `TaskID`) in a few places, and then when reviewing the code I recalled a conversation in one of the other PRs about these (it was one of the Dicts in that case, I think) being more for declaring types than for actually creating objects. So I change them back but did that over-aggressively and changed some that already existed.

For clarity (mine and maybe I can incorporate this into some docs somewhere when working on those): what is the intention with these types - the `*ID` ones and also the `*Dict` ones (especially if we remove the leading underscore as discussed elsewhere here)?

Ignoring that you'd probably collapse these to a single line, which of these (if any) would you recommend for a charmer?

```python
task_id = pebble.TaskID('12345')
task = ops.pebble.Task(id=task_id, kind='start', ...)
```

```python
task_id: pebble.TaskID = '12345'  # pyright will complain about this
task = ops.pebble.Task(id=task_id, kind='start', ...)
```

```python
task_id = typing.cast(pebble.TaskID, '12345')
task = ops.pebble.Task(id=task_id, kind='start', ...)
```

And would your answer be the same for ChangeDict, WarningDict, ChangeID, etc?

test/test_pebble.py
tonyandrewmeyer: Thinking about this more, it does feel a bit like `_CheckInfoDict` should mark both `level` and `failures` as `NotRequired`, since if the dict passed to `from_dict` is missing those then that's handled. The type is being used to define the input to the method rather than being used to define what you get out (which is obviously `CheckInfo` itself).

Since we don't have `NotRequired` yet, it'd have to be done like this:

```python
_BaseCheckInfoDict(TypedDict):
    name: str
    status: Union['CheckStatus', str]
    threshold: int
_CheckInfoDict(_BaseCheckInfoDict, total=False):
    level: Optional[Union['CheckLevel', str]]
    failures: int
```

Would you be ok with me changing it to that (maybe minus the non-base class's underscore pending the other discussion), and then getting rid of a couple of the `type: ignore`s?

test/test_pebble.py
tonyandrewmeyer: Yes. In general, it seems that return types are not required unless you want the type to be wider/narrower than what it's actually returning with the current code - or if you're doing it for documentation purposes rather than for the type checker.

It seems like what pyright does is construct an implied return type that's the union of the types of whatever gets returned by the function, so as long as it knows those types, it gets the rest. Here, it knows that `stdin`, `stdout`, and `stderr` are `MockWebSocket` objects, so 'figures out' that the return type is `Tuple[MockWebSocket, MockWebSocket, MockWebSocket]`, since that's the only return statement. Or if there's no `return` statement, then it assumes `-> None`.

test/test_pebble.py
tonyandrewmeyer: > because those `from_dict` methods are part of the public API, it seems to me like their types should be also.

I find this a compelling argument. It does seem like if you want to use the `from_dict` methods, and also have strict type checking, then you're forced to use the various `_*Dict` types.

> So perhaps we should remove the underscores from all of the following? (And move them to the top of pebble.py with the other public typed dicts.)

I guess my only question here is when you expect people to use the `from_dict` methods (for `Change`, `CheckInfo`, `FileInfo`, `Progress`, `ServiceInfo`, `SystemInfo`, `Task`, `Warning`). These are all representing information that flows unidirectionally from Pebble, right? Would a charmer (or more broadly someone who is using ops.pebble to manage pebble outside of charming) ever have reason to construct one of those objects (specifically from a dict, but really in general)?

Or more succinctly, is the issue really that `from_dict()` should be `_from_dict()` for that list of objects? And the type being private would then match? (Maybe this horse has bolted anyway, but I would be interested to know the answer).

ops/pebble.py
tonyandrewmeyer: Yes, good call :smile:

ops/pebble.py
tonyandrewmeyer: I've dropped these - there doesn't seem to be any value, indeed. The (stdlib) does imply that it's only for something that would be very expensive to do, so you want to skip it unless you're actually type checking. So that adds to the argument of getting rid of them.

There are some I'd definitely change, e.g. in `ops.log`:

```python
import typing

if typing.TYPE_CHECKING:
    from types import TracebackType
    from typing import Type

    from ops.model import _ModelBackend
```

`Type` can just be `typing.Type` since it's already imported. `types` will already be in `sys,modules` so it's just a cheap dictionary lookup. It doesn't seem like avoiding importing `ops.model` unless type checking has any real value (and `_ModelBackend` is already private so doesn't expand the public `log` namespace).

`ops.main` is similar:

```python
if TYPE_CHECKING:
    from ops.charm import CharmBase
    from ops.framework import BoundEvent, EventSource
    from ops.model import Relation
```

`charm`, `framework`, and `model` are all imported already, so all this does is avoid adding those to the public namespace. We're telling everyone to use everything from `ops` anyway, plus `__all__` is a more proper way to manage that, so I see no value in the conditional.

The other cases seem much the same. Should I open a separate PR that drops them everywhere they don't seem needed?

test/test_pebble.py
tonyandrewmeyer: > Or maybe we should leave that till we have access to `Required` and `NotRequired`, without which these are kind of awkward?

I missed that we already had `typing_extensions` in our requirements-dev, and already use `Required` elsewhere. So I can clean those up.

test/test_pebble.py
tonyandrewmeyer: Since we _do_ have `Required` (via `typing_extensions`) I'll go ahead with this change.

ops/pebble.py
benhoyt: Excellent digging, thanks. Yep, opening a separate PR that drops them everywhere sounds good.

test/test_pebble.py
benhoyt: Yeah, the intention with these types is something devs using strongly-typed languages like Scala sometimes do: create specific named types so it's harder to accidentally pass (say) a TaskID string as a ChangeID, or a Name as an Email. *However*, creating named types just to differentiate kinds of strings isn't something that I've seen much in Python (or Go for that matter -- some do it but it's rare), so I was probably a bit overzealous when adding those types in the first place -- I don't think I'd do that now, but we're stuck with these. :-)

In any case, it's a bit different from TypedDict: that's just a normal dict with a specific shape/fields. So for those I'd probably use `d: ChangeDict = {"foo": 42}` or probably just `d = {"foo": 42}` (it just needs to fit the parameter types shape, not be that type).

But these are actual types (a subclass of str), so I think it makes most sense to instantiate them as `pebble.TaskID('123')` -- your first example.

In real charm code, however, you probably never have to say `pebble.ChangeID('123')`, as you'd probably get a change ID from `Client.get_changes` or similar (which returns it as a `ChangeID` type already), and then pass it directly to `get_change` or similar.

test/test_pebble.py
benhoyt: Nice!

test/test_pebble.py
benhoyt: Interesting. From a while back I remember that MyPy was different in this respect -- it requires return type annotations. Or maybe it was that functions like `def foo():` weren't type checked under MyPy (but they are under Pyright), so you'd have to explicitly declare it as `def foo() -> None:` to get MyPy to type check it. I added some `-> None`s to methods/functions that took no parameters a while back.

ops/model.py
```diff
@@ -2369,7 +2376,7 @@ def pull_path(self,
                     dstpath.parent.mkdir(parents=True, exist_ok=True)
                     with self.pull(info.path, encoding=None) as src:
                         with dstpath.open(mode='wb') as dst:
-                            shutil.copyfileobj(typing.cast(BinaryIO, src), dst)
+                            shutil.copyfileobj(src, dst)
```
benhoyt, line 2379: Nice catch!

test/test_pebble.py
benhoyt: Hmm, I was wrong about `TypedDict` -- they're not just a normal dict with shape checked, but a real type. If you say:

```python
        d = {
            'message': 'Look out...',
            'first-added': '2020-12-25T17:18:54.016273778+13:00',
            'last-added': '2021-01-26T17:01:02.12345+13:00',
            'expire-after': '1s',
            'repeat-after': '2s',
        }
        warning = pebble.Warning.from_dict(d)
```

Pyright gives this error:

```
Argument of type "dict[str, str]" cannot be assigned to parameter "d" of type "WarningDict" in function "from_dict"
```

However, type checkers *should* automatically infer the type if you pass it directly as a function argument. So this type checks fine:

```python
        warning = pebble.Warning.from_dict({
            'message': 'Look out...',
            'first-added': '2020-12-25T17:18:54.016273778+13:00',
            'last-added': '2021-01-26T17:01:02.12345+13:00',
            'expire-after': '1s',
            'repeat-after': '2s',
        })
```

Ah, the joys of Python type checking. See more at https://peps.python.org/pep-0589/#using-typeddict-types

I guess that kind of changes my opinion about `d = ChangeDict({"foo": 42})` -- that does seem reasonable from a typing point of view. But I still prefer to think of `ChangeDict` as "basically a normal dict with a shape", so IMO we should use:

* `d: ChangeDict = {'foo': 42}` style for typed dicts, and
* `task_id = pebble.TaskID('123')` style for those subclass types

ops/pebble.py
```diff
@@ -201,11 +204,11 @@ def __enter__(self) -> typing.IO[typing.AnyStr]: ...  # noqa
...
                            'summary': str,
                            'status': str,
-                           'log': Optional[List[str]],
+                           'log': NotRequired[Optional[List[str]]],
                            'progress': _ProgressDict,
                            'spawn-time': str,
-                           'ready-time': str,
-                           'data': Optional[Dict[str, Any]]})
+                           'ready-time': NotRequired[Optional[str]],
+                           'data': NotRequired[Optional[Dict[str, Any]]]})
```
benhoyt, line 211: Not a comment for this line, but it looks like `_ChangeDict` `tasks`, `err`, `ready_time`, and `data` should be `NotRequired`.

test/test_pebble.py
```diff
@@ -283,8 +286,8 @@ def test_task_init(self):
         self.assertEqual(task.data, {})
 
     def test_task_from_dict(self):
-        d = {
-            "id": "78",
+        d: pebble._TaskDict = {
+            "id": typing.cast(pebble.TaskID, "78"),
```
benhoyt, line 290: I believe this field should be a regular string, as it's in the raw argument to `from_dict`.

test/test_pebble.py
```diff
@@ -322,7 +325,7 @@ def test_change_id(self):
 
     def test_change_init(self):
         change = pebble.Change(
-            id=pebble.ChangeID('70'),
+            id=typing.cast(pebble.ChangeID, '70'),
```
benhoyt, line 328: Let's revert the rest of these to the `pebble.ChangeID(x)` syntax.

test/test_pebble.py
```diff
@@ -344,7 +347,7 @@ def test_change_init(self):
         self.assertEqual(change.data, {})
 
     def test_change_from_dict(self):
-        d = {
+        d: 'pebble._ChangeDict' = {
```
benhoyt, line 350: I guess the need for the quotes will go away once we remove the `if TYPE_CHECKING` thingy. (No need to fix that here.)

test/test_pebble.py
```diff
@@ -1059,13 +1078,13 @@ def __init__(self):
...
-    def sleep(self, delay):
+    def sleep(self, delay: float):
         self._time += delay
 
 
-def build_mock_change_dict(change_id='70'):
+def build_mock_change_dict(change_id: str = '70') -> 'pebble._ChangeDict':
     return {
-        "id": change_id,
+        "id": typing.cast(pebble.ChangeID, change_id),
```
benhoyt, line 1087: This should be a plain str here.

test/test_pebble.py
```diff
@@ -1074,13 +1093,13 @@ def build_mock_change_dict(change_id='70'):
         "summary": 'Autostart service "svc"',
         "tasks": [
             {
-                "id": "78",
+                "id": typing.cast(pebble.TaskID, "78"),
```
benhoyt, line 1096: Ditto here.

test/test_pebble.py
```diff
@@ -1380,7 +1399,20 @@ def test_get_change(self):
...
             "type": "sync"
         })
-        change = self.client.get_change('70')
+        change = self.client.get_change(typing.cast(pebble.ChangeID, '70'))
+        self.assert_mock_change(change)
+        self.assertEqual(self.client.requests, [
+            ('GET', '/v1/changes/70', None, None),
+        ])
+
+    def test_get_change_str(self):
```
benhoyt, line 1408: Nice addition, thanks.

test/test_pebble.py
```diff
@@ -1380,7 +1399,20 @@ def test_get_change(self):
             "status-code": 200,
             "type": "sync"
         })
-        change = self.client.get_change('70')
+        change = self.client.get_change(typing.cast(pebble.ChangeID, '70'))
```
benhoyt, line 1402: Per above, let's settle on `pebble.ChangeID('70')` syntax for these.

benhoyt (commented):
Nice updates -- just a few minor quibbles left.

test/test_pebble.py
benhoyt: > Or more succinctly, is the issue really that from_dict() should be _from_dict() for that list of objects? And the type being private would then match? (Maybe this horse has bolted anyway, but I would be interested to know the answer).

Yeah, I think you're right -- these should have been `_from_dict` in the first place, as that's an internal implementation detail, and we don't really expect users to instantiate these because they're from responses from the Pebble server.

But yeah, this cat is already out of the bag (how many other idioms are there for that? :-). So I think we should go all in and remove the underscore. We don't have to add them to anyone's `__all__`, as `ops.pebble` stuff isn't imported to the top level (intentionally), so it won't pollute namespaces.

benhoyt:
Oh, and per [this comment](https://github.com/canonical/operator/pull/1030#discussion_r1347907966), yeah, unfortunately it's too late to make the `from_dict` methods private, so I think we should remove the underscore on those `from_dict` `FooDict` types mentioned.

ops/pebble.py
tonyandrewmeyer: Yes - I'd only done the ones that the tests were hitting, but indeed better to tighten then all up at once.

test/test_pebble.py
tonyandrewmeyer: It has to be a `TaskID`, because we're passing in a `_TaskDict` and the `id` in `_TaskDict` is a `TaskID`. Type checkers complain if it's not a `TaskID`.

Fwiw, it seems strange to me that type checkers aren't ok with a string literal being used where a string type alias is expected. But that's the current state.

benhoyt: Oops, yeah, that `id` field should not be a `TaskID`, just a regular `str`, as that type is a plain dict that comes from the wire from the Pebble API.

test/test_pebble.py
tonyandrewmeyer: I suspect maybe this is something that's changed over time (pyright certainly seems to pump out releases full of changes at a furious rate). Right now both seem ok with it:

```
tameyer@tam-canoncial-1:~/code/operator$ cat /tmp/returntypes.py 
def a():
    return "hello world!"

def b(i: int):
    return f"got {i}"

def c(i: int):
    if i > 0:
        return 'string'
    return b'bytes'

def d():
    a()
tameyer@tam-canoncial-1:~/code/operator$ pyright /tmp/returntypes.py 
tameyer@tam-canoncial-1:~/code/operator$ mypy /tmp/returntypes.py 
Success: no issues found in 1 source file
```

test/test_pebble.py
tonyandrewmeyer: As discussed, I'll open a separate PR for this.

benhoyt (approved):
Looking good, thx!

PR #1031, opened 2023-10-04
docs: make pebble.Client.remove_path and Container.remove_path docs consistent

### LINKED ISSUE #699: Document remove_path method behaviour and exceptions properly
The `remove_path` method in `pebble` has an optional argument `recursive: bool = False`:

```python
def remove_path(self, path: str, *, recursive: bool = False):
```

Let's say that you need to delete a **file** from a container's directory. But it's possible that the file you trying to delete it is not **yet** in the directory.

If you have:

```python
container.remove_path(path)
```
You will get a `PathError` exception since the file does not exist. This makes sense, it is behaviourally similar to `rm <file>`:

```bash
➜  rm pepe.txt
rm: cannot remove 'pepe.txt': No such file or directory
```
But there is no explicit option that behaves similar to `rm -f <file>` (do nothing if the files is non-existent):

```bash
➜  rm -f pepe.txt
➜
```

Anyway you can achieve that behaviour by using `recursive=True` 

```python
container.remove_path(path, recursive=True)
```

`recursive=True` behaves in 2 ways:

- If `path` is a directory recursively deletes it and everything under it.
- If the `path` is a file, delete the file and do nothing if the file is non-existent. Behaviourally similar  to `rm -rf <file|dir>`

Using `recursive=True` when trying to delete a file and do nothing if file is non-existent (`rm -f <file>`) is a little bit confusing.

Shouldn't we try to mimic the rm behaviour where we have: [`-r` for recursive and `-f` for "ignore nonexistent files and arguments, never prompt"](https://www.man7.org/linux/man-pages/man1/rm.1.html#OPTIONS)?

tonyandrewmeyer:
* Copy the docstring from `pebble.Client.remove_path` to `model.Container.remove_path` (and fix a missing comma).
* Document that `remove_path` may raise `PathError` in some situations.

Annoyingly [the Go docs](https://pkg.go.dev/os#Remove) are vague on when an error will occur. The pebble code seems to raise an error if the path is relative or if the underlying Go call does. I've kept the wording fairly generic as a result.

Fixes #699


ops/model.py
```diff
@@ -2509,7 +2509,15 @@ def remove_path(self, path: Union[str, PurePath], *, recursive: bool = False):
 
         Args:
             path: Path of the file or directory to delete from the remote system.
-            recursive: If True, recursively delete path and everything under it.
+            recursive: If True, and path is a directory, recursively deletes it and
+                       everything under it. If the path is a file, delete the file and
+                       do nothing if the file is non-existent. Behaviourally similar
+                       to ``rm -rf <file|dir>``.
```
benhoyt, line 2515: I realise this wasn't your wording, but maybe we can clean it up slightly while we're here. The "do nothing if the file is non-existent" sort of implies it'll only do nothing if it doesn't exist and it's a *file*, but the same is true for directories with `recursive=True`. Perhaps:

```suggestion
            recursive: If True, and path is a directory, recursively delete it and
                       everything under it. If path is a file, delete the file. In
                       either case, do nothing if the file or directory does not
                       exist. Behaviourally similar to ``rm -rf <file|dir>``.```

File: ops/model.py [hunk 2]
```diff
@@ -2509,7 +2509,15 @@ def remove_path(self, path: Union[str, PurePath], *, recursive: bool = False):
...
+
+        Raises:
+            pebble.PathError: If the file cannot be removed (perhaps it does not
+                exist) and `recursive` is false, or if a relative path is provided.
```
benhoyt, line 2519: This wording is also awkward. Perhaps:
```suggestion
            pebble.PathError: If a relative path is provided, or if `recursive` is False
                and the file or directory cannot be removed (it does not exist or is not empty).
```

benhoyt (commented):
Thanks for the changes. Couple of improvement suggestions -- feel free to tweak further.

ops/pebble.py
```diff
@@ -2153,11 +2153,15 @@ def remove_path(self, path: str, *, recursive: bool = False):
 
         Args:
             path: Path of the file or directory to delete from the remote system.
-            recursive: If True, and path is a directory recursively deletes it and
+            recursive: If True, and path is a directory, recursively deletes it and
```
benhoyt, line 2156: Can we also please copy the new version above to here? "recursive" and "Raises".

tonyandrewmeyer:
The mysql-k8s-operator tests [changed](https://github.com/canonical/mysql-k8s-operator/commit/128d32fd720f65b18a69650bb827dddf64a2b940#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133dd), will open a separate PR for addressing that.

benhoyt (approved):

PR #1032, opened 2023-10-05
feat: narrow the type for 'app' and 'unit' in relation events

### LINKED ISSUE #976: Improve typing for RelationEvent.unit
Related to [discussion point](https://github.com/canonical/jenkins-k8s-operator/pull/11#discussion_r1264776509) on Jenkins-k8s PR, the `RelationEvent.unit` property is typed as `Unit | None`. 
Types such as `RelationJoinedEvent` can have the type narrowed down to `Unit` (without `None`) so that the user does not have to typecast the `unit` property.

tonyandrewmeyer:
* `app` is always provided for relation events, except for relation-broken (which may be a [Juju bug](https://bugs.launchpad.net/juju/+bug/1960934)), so for all of the relation event classes, make the type `Application` rather than `Application|None`.
* `unit` may be provided for relation-changed, and is provided for unit level relation events (joined, departed), but is not provided for application level events (created, broken), so adjust the types in the relation event classes to be either `None` or `Unit` as applicable.
* Adjust the tests so that we have a relation-change with and without the unit provided, and so that the app is provided for all cases except relation-broken, which has one test with and one without.

The `RelationCreated` and `RelationBroken` classes do now have an odd `unit` attribute that is defined as always being None, but that was the actual case before and we're just being explicit about it now. I assume dropping the attribute would be too much of a backwards compatibility issue.

Similarly, the `__init__` for the various event classes still accept `None` for `app` and `unit` even when they are actually always expected. This is unfortunate, but is stuck for backwards compatibility. Generally, we are the one creating the event objects, not the charmer, so the inexact types in the `__init__` shouldn't impact charmers.

Fixes #976


tonyandrewmeyer:
The grafana-k8s-operator tests assume that `emit` can be called with just the relation. I _think_ this is an issue with the test, and that couldn't happen in production. However, it still indicates that there are backwards compatibility issues here. So I'll need to do a more minor change that leaves the `__init__` in place I suppose.

benhoyt:
> The grafana-k8s-operator tests assume that emit can be called with just the relation. I think this is an issue with the test, and that couldn't happen in production. However, it still indicates that there are backwards compatibility issues here. So I'll need to do a more minor change that leaves the __init__ in place I suppose.

Yeah, I think that assessment is probably right: it's probably too much to change the signature here, even if that test shouldn't really be doing that. So let's leave `__init__` alone and focus on the typing changes.

ops/charm.py
```diff
@@ -371,9 +371,10 @@ class RelationEvent(HookEvent):
     relation: 'Relation'
     """The relation involved in this event."""
 
-    # TODO(benhoyt): I *think* app should never be None, but confirm and update type
     app: Optional[model.Application]
-    """The remote application that has triggered this event."""
+    """The remote application that has triggered this event.
+
+    This may be ``None`` for relation-broken events."""
```
benhoyt, line 377: I think we should just bite the bullet and make this `app: model.Application` (non-optional) for all of these. It seems like from all accounts it should be set, and it's a pretty big Juju issue if it's not (the relation-broken issue some have seen).

Minor style nit: let's put the ending `"""` on its own new line -- not enforced by linting, but that's the style we've used elsewhere for multi-line docstrings, and I think it makes the two `"""`'s line up nicely and marks a bit more clearly where the block comment ends.

ops/charm.py
```diff
@@ -369,11 +369,12 @@ class RelationEvent(HookEvent):
     """
 
     relation: 'Relation'
-    """The relation involved in this event."""
+    """The relation involved in this event.
+    """
```
benhoyt, line 373: Style nit: for one-liner docstrings the quotes all go on the same line, but for multiline ones the end `"""` quote on a new line.

test/test_main.py
```diff
@@ -408,10 +408,16 @@ def _simulate_event(self, event_spec: EventSpec):
...
                 env['JUJU_REMOTE_APP'] = remote_app
 
-            remote_unit = event_spec.remote_unit
-            if remote_unit is None:
-                remote_unit = ''
-            env['JUJU_REMOTE_UNIT'] = remote_unit
+            if (remote_app is None
+                or issubclass(event_spec.event_type, ops.RelationJoinedEvent)
+                or issubclass(event_spec.event_type, ops.RelationChangedEvent)
+                    or issubclass(event_spec.event_type, ops.RelationDepartedEvent)):
```
benhoyt, line 414: Can you explain (perhaps in a comment) what this `if` block is testing for? I don't really understand the logic here or why it's needed.

Style nit: I'd probably indent all three `or`s the same (to the more-indented level of the last one).

tonyandrewmeyer: > Style nit: I'd probably indent all three `or`s the same (to the more-indented level of the last one).

This is where I do much prefer:

```python
if (
    remote_app is None
    or issubclass(event_spec.event_type, ops.RelationJoinedEvent)
    or issubclass(event_spec.event_type, ops.RelationChangedEvent)
    or issubclass(event_spec.event_type, ops.RelationDepartedEvent)
):
```

(with or without the first condition on a separate line).

But that doesn't match our formatting, so we have to have something that distinguishes the indent from the start of the actual block. Moved the others in.

tonyandrewmeyer: > Can you explain (perhaps in a comment) what this `if` block is testing for? I don't really understand the logic here or why it's needed.

Turns out that it is not needed :disappointed:. Earlier, when I had the `__init__` signatures changing, these tests were failing, and I thought I had tracked that down to this code passing the envvar when it shouldn't, and so this seemed more like the Juju behaviour. However, that really ought to just be done with changes in the event spec definitions I guess, and it also isn't needed any more.

benhoyt (approved):

PR #1034, opened 2023-10-06
chore: remove as much 'if typing.TYPE_CHECKING:' as possible

tonyandrewmeyer:
* Wherever possible, remove the 'if typing.TYPE_CHECKING:' guards.
* Move the definition of the `_ConfigOption` `TypedDict` from `ops.testing` to `ops.model` - `model` should not import from `testing` (the import loop was avoided because it was only when `TYPE_CHECKING` but it's still poor practice)

This leaves a few cases where we still use `TYPE_CHECKING`:
* When using `typing_extensions`. We currently assume that if `TYPE_CHECKING` is true that the `requirements-dev` have been installed (or at least `typing_extensions` has). It seems like this would not always be the case (e.g. if type checking is being done on a charm that uses `ops`) but is the status-quo and avoids bringing `typing_extensions` into `requirements.txt`. Hopefully, we won't need it for anything after `Required`/`NotRequired` and once we have progressed our minimum Python version enough to get those, we can drop it entirely.
* We override the definition of some class attributes to be properties when `TYPE_CHECKING` is true. Removing the guard would change the actual attributes.
* In Python 3.8 we can't use `weakref.WeakValueDictionary[T]` with a type. We're currently assuming that a more recent version of Python will be used when type checking, but can't bring this out as that would raise the minimum Python version for all cases (or we'd need to adjust the typing so that it can work in 3.8).


tonyandrewmeyer:
Odd, the docs build ok locally for me now:

```
tameyer@tam-canoncial-1:~/code/operator$ tox -e docs
docs: commands[0]> sphinx-build -W --keep-going docs/ docs/_build/html
Running Sphinx v6.2.1
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
writing output... 
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
reading sources... 
looking for now-outdated files... none found
no targets are out of date.
build succeeded.

The HTML pages are in docs/_build/html.
  docs: OK (0.61=setup[0.02]+cmd[0.59] seconds)
  congratulations :) (0.65 seconds)
```

@benhoyt any suggestions on what needs to be done to resolve this?

benhoyt:
Huh, weird that the docs build locally for you, because we're pinning our docs/requirements.txt to exact versions. It fails for me locally. Oh wait, the second time it succeeds -- some caching issue? After I `rm -rf docs/_build/` it fails again.

In any case, it looks like for some reason Sphinx now wants to see the full names in `docs/conf.py` -- I guess because it wasn't actually seeing those names before:

```diff
diff --git a/docs/conf.py b/docs/conf.py
index f335d64..579a28b 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -78,7 +78,7 @@ nitpick_ignore = [
     ('py:class', '_CheckInfoDict'),
     ('py:class', '_FileInfoDict'),
     ('py:class', '_IOSource'),
-    ('py:class', '_NetworkDict'),
+    ('py:class', 'ops.model._NetworkDict'),
     ('py:class', '_ProgressDict'),
     ('py:class', '_Readable'),
     ('py:class', '_RelationMetaDict'),
```

tonyandrewmeyer:
> Huh, weird that the docs build locally for you, because we're pinning our docs/requirements.txt to exact versions. It fails for me locally. Oh wait, the second time it succeeds -- some caching issue? After I `rm -rf docs/_build/` it fails again.

Ah indeed, the caching fooled me. Thanks!

ops/testing.py
```diff
@@ -59,39 +58,31 @@
...
+    'app': str,
+    'units': List[str]
+})
+
+_StatusName = Literal['unknown', 'blocked', 'active', 'maintenance', 'waiting']
+_RawStatus = TypedDict('_RawStatus', {
+    'status': _StatusName,
+    'message': str,
+})
+RawConfig = TypedDict("RawConfig", {'options': Dict[str, _ConfigOption]})
```
benhoyt, line 85: I was looking at the [API docs](https://ops--1034.org.readthedocs.build/en/1034/#ops.testing.RawConfig), and I noticed this accidentally became part of the public API. Can it be underscore-prefixed?

benhoyt (changes requested):
Looking good. Just one note about `ops.testing.RawConfig`.

And thanks for fixing that model -> testing import!

benhoyt (approved):

PR #1035, opened 2023-10-06
Add `Unit.ordinal`

carlcsaposs-canonical:
Many of our charms use `ops.Unit.name.split("/")[1]` or `ops.Unit.name.split("/")[-1]` to parse the unit ordinal

The unit ordinal appears to be part of Juju's official API:

> `A unit is always named on the pattern <application>/<unit ID>, where <application> is the name of the application and the <unit ID> is its ID number`

https://juju.is/docs/juju/unit

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -469,12 +469,16 @@ class Unit:
     app: Application
     """Application the unit is part of."""
 
+    number: int
+    """Unit ID number, for example 0."""
+
     def __init__(self, name: str, meta: 'ops.charm.CharmMeta',
                  backend: '_ModelBackend', cache: '_ModelCache'):
         self.name = name
 
         app_name = name.split('/')[0]
         self.app = cache.get(Application, app_name)
+        self.number = int(name.split('/')[1])
 
         self._backend = backend
         self._cache = cache
--- a/ops/model.py
+++ b/ops/model.py
@@ -469,16 +469,16 @@ class Unit:
     app: Application
     """Application the unit is part of."""
 
-    number: int
+    ordinal: int
     """Unit ID number, for example 0."""
 
     def __init__(self, name: str, meta: 'ops.charm.CharmMeta',
                  backend: '_ModelBackend', cache: '_ModelCache'):
         self.name = name
 
-        app_name = name.split('/')[0]
+        app_name, ordinal = name.split('/')
         self.app = cache.get(Application, app_name)
-        self.number = int(name.split('/')[1])
+        self.ordinal = int(ordinal)
 
         self._backend = backend
         self._cache = cache
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -3596,5 +3596,15 @@ class TestPorts(unittest.TestCase):
         ])
 
 
+@pytest.mark.parametrize("name,ordinal", [("foo/3", 3), ("my-app-name/0", 0)])
+def test_unit_ordinal(name, ordinal):
+    meta = ops.CharmMeta.from_yaml("""
+    name: charm
+    """)
+    backend = _ModelBackend(name)
+    model = ops.Model(meta, backend)
+    assert model.unit.ordinal == ordinal
+
+
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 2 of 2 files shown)


ops/model.py
```diff
@@ -469,12 +469,16 @@ class Unit:
...
+    number: int
+    """Unit ID number, for example 0."""
+
     def __init__(self, name: str, meta: 'ops.charm.CharmMeta',
                  backend: '_ModelBackend', cache: '_ModelCache'):
         self.name = name
 
         app_name = name.split('/')[0]
         self.app = cache.get(Application, app_name)
+        self.number = int(name.split('/')[1])
```
jnsgruk, line 481:
```suggestion
        app_name, self.ordinal = name.split('/')
        self.app = cache.get(Application, app_name)
```

A slight simplification here, perhaps? I wonder if `ordinal` might be a better name for this?

jnsgruk:
Could we please add a small unit test for this?

ops/model.py
jnsgruk: Yeh, fair point, I think it would be a string.

Perhaps

```python
app_name, ordinal = name.split('/')
self.app = cache.get(Application, app_name)
self.ordinal = int(ordinal)
```

sed-i:
App name (as string) is also quite popular, even within ops itself.
(This is different from "unit name", which we often assume is `app_name + '/' + ordinal`.)
```
ops/main.py:196:        remote_app_name = remote_unit_name.split('/')[0]
ops/model.py:476:        app_name = name.split('/')[0]
ops/model.py:2845:        self.app_name: str = self.unit_name.split('/')[0]
ops/testing.py:987:            app_name = unit_name.split('/')[0]
ops/testing.py:1179:            app_name = app_or_unit.split('/')[0]
ops/testing.py:1492:        owner_app = secret.owner_name.split('/')[0]
ops/testing.py:1897:        self.app_name = self.unit_name.split('/')[0]
ops/testing.py:2025:            member_name = member_name.split('/')[0]
ops/testing.py:2312:            owner_app = secret.owner_name.split('/')[0]
```

ops/model.py
sed-i: Would need to update docstrings to use the new term "ordinal" (which I imagine is better than the existing "number", "id").

benhoyt:
Hmm, I'm sceptical that this is a good idea. On the one hand, the "/n" notation is probably pretty baked into Juju at this point (as shown by the docs). But is it intended to be part of the Juju contract that it's a decimal integer? What if we want to change "n" to a 7-digit hex code, or a UUID, or something?

Can you describe why (and perhaps link to where) your charms need to parse this as an integer? Or do they just treat it as an opaque string for uniqueness? (The latter I'd be more comfortable with.)

For comparison, in the ops library itself, we do use `remote_unit_name.split('/')[0]` to get the app name (the part before the slash), but it looks like never `[1]` or `[-1]` to get the unit ordinal (and we don't assume it's an int).

benhoyt:
@carlcsaposs-canonical Thanks for the context. I can see why you're doing this, though in a sense it's working around the lack of better modelling of upgrade strategies in Juju. Which is a fine workaround, but I don't think adding a dubious "ordinal int" to the ops API is a good idea.

I think what you're doing is fine, and just a one-liner so is very little code. As above, I don't really like the idea of adding `Unit.ordinal` as an int, but I'd consider `Unit.id: str` (though I'm not entirely sure about the name "id"). It would do the `/` parsing but not the conversion to int -- you'd still have to do that yourself with `key=lambda unit: int(unit.id)`. I'll talk further about this with the Juju team and let you know what we decide.

Separately, a nit comment on that code link: `key=lambda unit_: unit_number(unit_)` is equivalent to `key=unit_number`.

benhoyt:
I was talking with the Juju team earlier, and they'd also prefer not to expose an integer number/ordinal. I understand why you want it in your charm (to work around the lack of modelling of K8s upgrade strategy), but I don't want to bake an ops feature in for this kind of workaround.

So I'm going to close this PR based on that. I think you should just go ahead with your `unit_number` helper. We could consider adding `Unit.id` but even that I'm hesitant about, and you'd still need to convert to int for your sort.

Issue #316, opened 2020-06-01
unit state lost following upgrade-charm


jetpackdanger:
I just upgraded a charm I'm developing, after which the unit seems lost its unit state. Unfortinately, I don't have a consistent way to reproduce this. I've run into it only once or twice previously, but I've done many more successful charm upgrades.

Before:
```
mattermost/10*  active    idle   10.1.1.17  8065/TCP  
```

After:
```  
mattermost/10*  waiting   idle   10.1.1.17  8065/TCP  Waiting for database relation
```

The relation is still there:
```
[agnew(charm-k8s-mattermost)] juju status --format yaml mattermost
model:
  name: mm
  type: caas
  controller: mattermost
  cloud: k8s
  region: localhost
  version: 2.8-rc2
  model-status:
    current: available
    since: 28 May 2020 14:24:31+12:00
  sla: unsupported
machines: {}
applications:
  mattermost:
    charm: local:kubernetes/mattermost-17
    series: kubernetes
    os: kubernetes
    charm-origin: local
    charm-name: mattermost
    charm-rev: 17
    charm-version: 250d9df-dirty
    scale: 1
    provider-id: 335e8836-3907-48bc-8b65-98ddbe6d5263
    address: 10.152.183.32
    exposed: false
    application-status:
      current: active
      since: 02 Jun 2020 09:43:01+12:00
    relations:
      db:
      - postgresql
    units:
      mattermost/10:
        workload-status:
          current: waiting
          message: Waiting for database relation
          since: 02 Jun 2020 11:18:01+12:00
        juju-status:
          current: idle
          since: 02 Jun 2020 11:18:04+12:00
        leader: true
        open-ports:
        - 8065/TCP
        address: 10.1.1.17
        provider-id: d9681e49-0fb4-4fe4-a236-935caf767650
    endpoint-bindings:
      "": alpha
      db: alpha
application-endpoints:
  postgresql:
    url: mattermost:admin/daturbase.postgresql
    endpoints:
      db:
        interface: pgsql
        role: provider
    application-status:
      current: active
      message: Live master (10.12)
      since: 28 May 2020 14:25:48+12:00
    relations:
      db:
      - mattermost
storage: {}
controller:
  timestamp: 11:31:43+12:00
[agnew(charm-k8s-mattermost)] _
```
but the unit state DB seems to have been reset:
```
[...]

PR #1036, opened 2023-10-09
ci: Pass LIBJUJU_VERSION_SPECIFIER to the mysql-k8s-operator unit tests

tonyandrewmeyer:
In canonical/mysql-k8s-operator#313, the unit tests were changed to expect a LIBJUJU_VERSION_SPECIFIER envvar to be present when the tests are run. This is used in those tests to run the tests simulating Juju 2.9 and 3.1.

Adjust our running of the mysql-k8s-operator tests to also provide this value, fixed at 3.1. (This is also present for the other 3 DB operators, but that should do no harm).

### DIFF
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -43,6 +43,7 @@ jobs:
 
       - name: Run the charm's unit tests
         run: tox -vve unit
+        # TODO: remove this once https://github.com/canonical/mysql-k8s-operator/pull/316 is fixed
         env:
           # This env var is only to indicate Juju version to "simulate" in the mysqk-k8s-operator
           # unit tests. No libjuju is being actually used in unit testing.
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -43,3 +43,7 @@ jobs:
 
       - name: Run the charm's unit tests
         run: tox -vve unit
+        env:
+          # This env var is only to indicate Juju version to "simulate" in the mysqk-k8s-operator
+          # unit tests. No libjuju is being actually used in unit testing.
+          LIBJUJU_VERSION_SPECIFIER: 3.1


tonyandrewmeyer:
@benhoyt a couple of alternatives:

* We pass through `2.9` as the value. That seems better in terms of "always check you run with the oldest supported version", but worse in terms of "good to exercise that we haven't broken things for their use of secrets" (which seems to be the main thing they're using the version check for at the moment).
* We add in the matrix of values in our CI as well, testing both `2.9` and `3.1` values. This would avoid picking one, but would also mean that we test the other three DB charms twice for no reason, unless this was spun out more, or unless you can do a "sub matrix", which I don't think is possible in GitHub actions.

tonyandrewmeyer:
> * We add in the matrix of values in our CI as well, testing both `2.9` and `3.1` values. This would avoid picking one, but would also mean that we test the other three DB charms twice for no reason, unless this was spun out more, or unless you can do a "sub matrix", which I don't think is possible in GitHub actions.

Hmm, maybe [include](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) makes this possible? I haven't used that before. Should I try with that instead?

benhoyt:
This is a reasonable fix, but per discussion, it's definitely not ideal that the mysql-k8s-operator's *unit* test now require a certain environment variable to be set. We should be able to check out a charm repo and run `tox -e unit` without any further ado.

So let's put up a PR against their charm to fix that. They probably shouldn't be depending on that environment variable at all, but just have a single `has_secrets` fixture that patches `JujuVersion.has_secrets` with True.

Separately, it's a bit strange that default for `JujuVersion.from_environ` is 0.0.0. I wonder if we should override that to "latest stable version of Juju", at least in ops.testing?

tonyandrewmeyer:
> So let's put up a PR against their charm to fix that. They probably shouldn't be depending on that environment variable at all, but just have a single `has_secrets` fixture that patches `JujuVersion.has_secrets` with True.

Opened [canonical/mysql-k8s-operator:#316](https://github.com/canonical/mysql-k8s-operator/pull/316). I left the `without_juju_secrets` fixture in place because it does seem like being explicit there is worthwhile, and if we do change the default JujuVersion value in testing then we'd avoid breakage (at least for this one charm).

 > Separately, it's a bit strange that default for `JujuVersion.from_environ` is 0.0.0. I wonder if we should override that to "latest stable version of Juju", at least in ops.testing?

Opened #1037.

jameinel:
I do think you want coverage of mysql pre and post, because how it shares
passwords with the world changes (juju 2.9 doesn't have secrets as a way to
share the password so they have to put it in relation data, vs 3.1)

I definitely don't like an ENV var for unit testing (at the least it means
when running the unit test you aren't testing  both code paths in one suite
run)

Its plausible that Harness should make it easier to express the behaviors
that you want.

jameinel:
I looked over the mysql PR, and it really seems like we want a clean way to
fake the juju version that we are testing against, not patch out a method
on that object.

John
=:->

.github/workflows/db-charm-tests.yaml
```diff
@@ -43,3 +43,7 @@ jobs:
 
       - name: Run the charm's unit tests
         run: tox -vve unit
+        env:
```
benhoyt, line 46:
```suggestion
        # TODO: remove this once https://github.com/canonical/mysql-k8s-operator/pull/316 is fixed
        env:
```

benhoyt:
Thanks @jameinel -- agreed. I'm going to merge this to get our CI working in the meantime (with a TODO comment saying to remove it once https://github.com/canonical/mysql-k8s-operator/pull/316 is fixed), and we can continue the discussion of improving this at #1037.

benhoyt:
Minor CI fix, merged without further review.

PR #1041, opened 2023-10-10
feat: add Unit.reboot for machine charms

### LINKED ISSUE #929: No interface to `juju-reboot` hook tool in ops
Hello there! I am currently working through a predicament in the [SLURM charms](https://charmhub.io/?q=slurm) where the machines need to be rebooted after package installation to apply security updates. Originally, the SLURM charms were deployed manually so the human operator would issue a machine reboot command based on the substrate being used (e.g. `openstack server reboot <machine_id>`). We are now working to make deploying SLURM an automatic process, however, we still have the issue where the machines need to be rebooted to apply updates.

My idea was to use the `juju-reboot` hook tool located in `/var/lib/juju/tools/machine-<#>/`, however, it does not seem that ops has a direct method for invoking `juju-reboot`. It would be nice if ops had a defined method for scheduling machine reboots at the end of hook execution:

```python3
# Assume charm is fully defined

def _on_install(self, _: InstallEvent) -> None:
    # Install some packages, update base image, etc.
    ...

    if pathlib.Path("/var/run/reboot-required").exists():
        logger.info("Scheduling machine for reboot...")
        self.unit.reboot()
```

Currently I can circumvent my reboot issue by directly invoking the `juju-reboot` hook tool, but I thought that this might be a nice edition to ops. Let me know what your thoughts are!

tonyandrewmeyer:
Adds a new `reboot` method to `Unit`, that's a wrapper around the `juju-reboot` command.

`juju-reboot` is only available for machine charms. When it's run for a K8s charm, Juju logs an error message but doesn't return an response with the error (or have a non-zero exit), which means that we have to fail silently.

Adds a corresponding `reboot()` method to the harness backend.

From the point of view of the harness, we assume that everything is essentially the same after rebooting without `--now`, so this is a no-op. The most likely change that would impact the harness would be a leadership change, but it seems better to let the Charmer model that explicitly in their tests.

When rebooting with `--now`, what should happen is that everything after that point in the event handler is skipped, there's a simulated reboot (again, roughly a no-op), and then the event is re-emitted. I went down a long rabbit-hole trying to get the harness to do this, but it ended up way too messy for minimal benefit. Instead, a `SystemExit` exception is raised and the Charmer is expected to handle it in their tests, for example:

```python
def test_installation(self):
    self.harness.begin()
    with self.assertRaises(SystemExit):
        self.harness.charm.on.install.emit()
        # More asserts here that the first part of installation was done.
    self.harness.charm.on.install.emit()
    # More asserts here that the installation continued appropriately.
```

Fixes #929

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -692,7 +692,7 @@ class Unit:
             self._backend.open_port(protocol, port)
 
     def reboot(self, now: bool = False) -> None:
-        """Reboot the host machine, after stopping all containers hosted on the machine.
+        """Reboot the host machine.
 
         Normally, the reboot will only take place after the current hook successfully
         completes. Use ``now=True`` to reboot immediately without waiting for the
--- a/ops/model.py
+++ b/ops/model.py
@@ -26,6 +26,7 @@ import re
 import shutil
 import stat
 import subprocess
+import sys
 import tempfile
 import time
 import typing
@@ -3388,6 +3389,10 @@ class _ModelBackend:
     def reboot(self, now: bool = False):
         if now:
             self._run("juju-reboot", "--now")
+            # Juju will kill the Charm process, and in testing no code after
+            # this point would execute. However, we want to guarantee that for
+            # Charmers, so we force that to be the case.
+            sys.exit()
         else:
             self._run("juju-reboot")
 
--- a/ops/model.py
+++ b/ops/model.py
@@ -673,6 +673,8 @@ class Unit:
         Use :meth:`open_port` and :meth:`close_port` to manage ports
         individually.
 
+        *New in version 2.7*
+
         Args:
             ports: The ports to open. Provide an int to open a TCP port, or
                 a :class:`Port` to open a port for another protocol.
@@ -701,6 +703,8 @@ class Unit:
 
         This is not supported on Kubernetes charms, can only be called for the current unit,
         and cannot be used in an action hook.
+
+        *New in version 2.8*
 
         Args:
             now: terminate immediately without waiting for the current hook to complete,
--- a/ops/model.py
+++ b/ops/model.py
@@ -699,10 +699,8 @@ class Unit:
         hook to complete; this is useful when multiple restarts are required (Juju
         will re-run the hook after rebooting).
 
-        This will silently fail for Kubernetes charms.
-
-        This can only be called for the current unit, and cannot be used in an action
-        hook.
+        This is not supported on Kubernetes charms, can only be called for the current unit,
+        and cannot be used in an action hook.
 
         Args:
             now: terminate immediately without waiting for the current hook to complete,
--- a/ops/model.py
+++ b/ops/model.py
@@ -691,6 +691,32 @@ class Unit:
         for protocol, port in desired - existing:
             self._backend.open_port(protocol, port)
 
+    def reboot(self, now: bool = False) -> None:
+        """Reboot the host machine, after stopping all containers hosted on the machine.
+
+        Normally, the reboot will only take place after the current hook successfully
+        completes. Use the ``now`` argument when multiple reboots are required, to
+        reboot immediately without waiting for the hook to complete, and to restart the
+        hook after rebooting.
+
+        This will silently fail for Kubernetes charms.
+
+        This can only be called for the current unit, and cannot be used in an action
+        hook.
+
+        Args:
+            now: terminate immediately without waiting for the current hook to complete,
+                restarting the hook after reboot.
+
+        Raises:
+            RuntimeError: if called on a remote unit.
+            :class:`ModelError`: if used in an action hook.
+
+        """
+        if not self._is_our_unit:
+            raise RuntimeError(f'cannot reboot a remote unit {self}')
+        self._backend.reboot(now)
+
 
 @dataclasses.dataclass(frozen=True)
 class Port:
@@ -3348,6 +3374,12 @@ class _ModelBackend:
             logger.warning('Ignoring opened-ports port range: %s', port_str)
         protocol_lit = typing.cast(typing.Literal['tcp', 'udp'], protocol)
         return Port(protocol_lit, int(port))
+
+    def reboot(self, now: bool = False):
+        if now:
+            self._run("juju-reboot", "--now")
+        else:
+            self._run("juju-reboot")
 
 
 class _ModelBackendValidator:
--- a/ops/model.py
+++ b/ops/model.py
@@ -695,9 +695,9 @@ class Unit:
         """Reboot the host machine, after stopping all containers hosted on the machine.
 
         Normally, the reboot will only take place after the current hook successfully
-        completes. Use the ``now`` argument when multiple reboots are required, to
-        reboot immediately without waiting for the hook to complete, and to restart the
-        hook after rebooting.
+        completes. Use ``now=True`` to reboot immediately without waiting for the
+        hook to complete; this is useful when multiple restarts are required (Juju
+        will re-run the hook after rebooting).
 
         This will silently fail for Kubernetes charms.
 
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1882,6 +1882,10 @@ class _Secret:
     grants: Dict[int, Set[str]] = dataclasses.field(default_factory=dict)
 
 
+class RebootingMachineError(Exception):
+    """Raised when the machine would reboot."""
+
+
 @_copy_docstrings(model._ModelBackend)
 @_record_calls
 class _TestingModelBackend:
@@ -2502,6 +2506,14 @@ class _TestingModelBackend:
                 raise model.ModelError(f'ERROR port range bounds must be between 1 and 65535, got {port}-{port}\n')  # NOQA: test_quote_backslashes
         else:
             raise model.ModelError(f'ERROR invalid protocol "{protocol}", expected "tcp", "udp", or "icmp"\n')  # NOQA: test_quote_backslashes
+
+    def reboot(self, now: bool = False):
+        if not now:
+            # We can't simulate the reboot, so just do nothing.
+            return
+        # This should exit, reboot, and re-emit the event, but we'll need the caller
+        # to handle that. We raise an exception so that they can simulate the exit.
+        raise RebootingMachineError()
 
 
 @_copy_docstrings(pebble.ExecProcess)
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1733,12 +1733,9 @@ class Harness(Generic[CharmType]):
         )
 
     @property
-    def last_rebooted(self) -> Optional[datetime.datetime]:
-        """The time that the charm last called :meth:`ops.Unit.reboot` (with or without *now*).
-
-        Returns ``None`` if :meth:`ops.Unit.reboot` has not been called.
-        """
-        return self._backend._last_reboot
+    def reboot_count(self) -> int:
+        """Count of times that the charm has called :meth:`ops.Unit.reboot`."""
+        return self._backend._reboot_count
 
 
 def _get_app_or_unit_name(app_or_unit: AppUnitOrName) -> str:
@@ -1974,7 +1971,7 @@ class _TestingModelBackend:
         self._secrets: List[_Secret] = []
         self._opened_ports: Set[model.Port] = set()
         self._networks: Dict[Tuple[Optional[str], Optional[int]], _NetworkDict] = {}
-        self._last_reboot = None
+        self._reboot_count = 0
 
     def _validate_relation_access(self, relation_name: str, relations: List[model.Relation]):
         """Ensures that the named relation exists/has been added.
@@ -2536,7 +2533,7 @@ class _TestingModelBackend:
             raise model.ModelError(f'ERROR invalid protocol "{protocol}", expected "tcp", "udp", or "icmp"\n')  # NOQA: test_quote_backslashes
 
     def reboot(self, now: bool = False):
-        self._last_reboot = datetime.datetime.now(datetime.timezone.utc)
+        self._reboot_count += 1
         if not now:
             return
         # This should exit, reboot, and re-emit the event, but we'll need the caller
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1883,7 +1883,21 @@ class _Secret:
 
 
 class RebootingMachineError(Exception):
-    """Raised when the machine would reboot."""
+    """Raised when the machine would reboot.
+    
+    When :meth:`Unit.reboot` is called with ``now=True`` in a machine charm, the
+    unit's machine is rebooted, interrupting the execution of the event hook. To
+    simulate that when using the testing harness, write a test that expects a
+    ``RebootingMachineError`` to be raised, for example::
+
+        def test_installation(self):
+            self.harness.begin()
+            with self.assertRaises(ops.testing.RebootMachineError):
+                self.harness.charm.on.install.emit()
+                # More asserts here that the first part of installation was done.
+            self.harness.charm.on.install.emit()
+            # More asserts here that the installation continued appropriately.
+    """
 
 
 @_copy_docstrings(model._ModelBackend)
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1888,7 +1888,7 @@ class RebootingMachineError(Exception):
     When :meth:`Unit.reboot` is called with ``now=True`` in a machine charm, the
     unit's machine is rebooted, interrupting the execution of the event hook. To
     simulate that when using the testing harness, write a test that expects a
-    ``RebootingMachineError`` to be raised, for example::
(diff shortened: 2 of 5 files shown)


ops/model.py
```diff
@@ -691,6 +691,32 @@ def set_ports(self, *ports: Union[int, 'Port']) -> None:
         for protocol, port in desired - existing:
             self._backend.open_port(protocol, port)
 
+    def reboot(self, now: bool = False) -> None:
+        """Reboot the host machine, after stopping all containers hosted on the machine.
```
benhoyt, line 695: What does "after stopping all containers" mean in this context, particularly given that this doesn't work on K8s?

ops/model.py
```diff
@@ -691,6 +691,32 @@ def set_ports(self, *ports: Union[int, 'Port']) -> None:
...
+
+        Normally, the reboot will only take place after the current hook successfully
+        completes. Use the ``now`` argument when multiple reboots are required, to
+        reboot immediately without waiting for the hook to complete, and to restart the
+        hook after rebooting.
```
benhoyt, line 700: How about making this slightly more specific/direct. Perhaps:

```suggestion
        completes. Use ``now=True`` to reboot immediately without waiting for the
        hook to complete; this is useful when multiple restarts are required (Juju
        will re-run the hook after rebooting).
```

ops/model.py
```diff
@@ -691,6 +691,32 @@ def set_ports(self, *ports: Union[int, 'Port']) -> None:
...
+
+        This will silently fail for Kubernetes charms.
```
benhoyt, line 702: Hmm, yeah, that's a pain. Can we make it loudly fail? It sounds like the Juju hook tool doesn't fail with an exit code right now. What's the error message it logs? If it's not something *too* generic we could look for that error message on stdout (or is it stderr?) and raise a nice error. It's normally not great to parse error messages like this in case the Juju side changes, but all we're doing is improving the error handling, so I'd be in favour of that. (We do this in a couple of other places, for example `_ModelBackend._is_relation_not_found`.)

ops/testing.py
```diff
@@ -1882,6 +1882,24 @@ class _Secret:
     grants: Dict[int, Set[str]] = dataclasses.field(default_factory=dict)
 
 
+class RebootingMachineError(Exception):
```
benhoyt, line 1885: Thanks for thinking about the testing / Harness side of things!

I think the approach of raising an exception is good. I'm not sure about the name, as it's more of a signal than an "error" like `FileNotFoundError`. What do you think of `MachineRebooting`, or even `RebootNow`?

ops/testing.py
```diff
@@ -1882,6 +1882,24 @@ class _Secret:
...
+    """Raised when the machine would reboot.
+
+    When :meth:`Unit.reboot` is called with ``now=True`` in a machine charm, the
+    unit's machine is rebooted, interrupting the execution of the event hook. To
+    simulate that when using the testing harness, write a test that expects a
+    ``RebootingMachineError`` to be raised, for example::
```
benhoyt, line 1891:
```suggestion
    this exception to be raised, for example::
```

ops/testing.py
```diff
@@ -2503,6 +2521,14 @@ def _check_protocol_and_port(self, protocol: str, port: Optional[int]):
         else:
             raise model.ModelError(f'ERROR invalid protocol "{protocol}", expected "tcp", "udp", or "icmp"\n')  # NOQA: test_quote_backslashes
 
+    def reboot(self, now: bool = False):
+        if not now:
+            # We can't simulate the reboot, so just do nothing.
```
benhoyt, line 2526: While we're considering testing of this, I think we should somehow signal this case so that it can be tested for -- that reboot was called (without now).

My suggesting would be we set a "rebooted" flag (whether it was called with now or not) and then add a Harness method like this:

```python
def rebooted(self) -> bool:
    """Report whether the charm code called :meth:`Unit.reboot` (with or without *now*)."""
```

test/test_testing.py
```diff
@@ -3337,6 +3337,32 @@ def test_get_pebble_methods(self):
         client = backend.get_pebble('/custom/socket/path')
         self.assertIsInstance(client, _TestingPebbleClient)
 
+    def test_reboot(self):
+        class RebootingCharm(ops.CharmBase):
+            def __init__(self, *args, **kwargs):  # type: ignore
+                super().__init__(*args, **kwargs)  # type: ignore
```
benhoyt, line 3343: I wonder if we should start recommending the more explicit code as follows (also to avoid the `type: ignore`s). What are the trade-offs?

```suggestion
            def __init__(self, framework: ops.Framework):
                super().__init__(framework)
```

test/test_testing.py
```diff
@@ -3337,6 +3337,32 @@ def test_get_pebble_methods(self):
...
+                self.framework.observe(self.on.install, self._reboot)
+                self.framework.observe(self.on.remove, self._reboot_now)
+
+            def _reboot(self, event: ops.RemoveEvent):
```
benhoyt, line 3347: Looks like these types should be the other way around:

```suggestion
            def _reboot(self, event: ops.InstallEvent):
```

test/test_model.py
```diff
@@ -3596,5 +3596,27 @@ def test_set_ports_noop(self):
         ])
 
 
+class TestUnit(unittest.TestCase):
```
benhoyt, line 3599: Yeah, there's a *lot* stuffed into `TestModel` right now. I think (starting to) separate them out is the right call -- thanks.

benhoyt:
Oh, looks like `tox -e docs` is failing right now.

ops/model.py
tonyandrewmeyer: > What does "after stopping all containers" mean in this context, particularly given that this doesn't work on K8s?

It's stolen from the [juju docs](https://juju.is/docs/sdk/hook-tool#heading--juju-reboot). I'm not 100% sure on the meaning. I think [this is the relevant Juju code](https://github.com/juju/juju/blob/f3aed6169084eb326a397da071399581947b89a8/cmd/jujud/reboot/reboot.go#L58), but it's not clear to me what `containers` means in that context.

ops/testing.py
tonyandrewmeyer: > I think the approach of raising an exception is good. I'm not sure about the name, as it's more of a signal than an "error" like `FileNotFoundError`. What do you think of `MachineRebooting`, or even `RebootNow`?

Yeah, it was without the error until the linter complained, so I renamed it rather than silence the warning. I'm happy to go back. `RebootNow` works for me.

test/test_testing.py
tonyandrewmeyer: Ah, thanks. Sloppy on my part (I had them around the other way originally and then changed my mind, but missed the type).

ops/model.py
benhoyt: Hmm, I have no idea what that comment means either (it's 9 years old, well before we started supporting K8s, so containers must have meant something different). I'd drop it until/unless we understand it. Feel free to ask on the Juju channel (ping `wallyworld`) if you want, or we can just drop it.

ops/model.py
tonyandrewmeyer: The error is in the logger, ie. `debug-log` shows this:

```
unit-uptime-0: 12:59:56 ERROR juju.worker.uniter.operation hook "config-changed" (via hook dispatching script: dispatch) failed: unit "uptime/0" is not assigned to a machine
```

It indeed does not fail with an exit code (or `ModelError` would conveniently be raised, as with running in an action hook). When passing `return_output=True` to `_run()` I get an empty string back, so I believe that means that it's not putting anything into stdout/stderr. I was assuming we would not want to introspect the Juju logs.

Maybe Juju should be exiting with an error in this case? I could ask the Juju team about that and open a PR for it if so.

benhoyt: Ah, got it. Yeah, we definitely don't want to introspect Juju logs.

However, it looks like `_run()` doesn't return stderr (unless there's an exception / nonzero exit code). Perhaps you can try running it directly using `subprocess.run` and print both stdout and stderr to see if we get any stderr.

I definitely think Juju should be returning an error in this case. We'll decide whether it's worth opening an issue there after we determine the above.

ops/model.py
tonyandrewmeyer: [Asked](https://chat.charmhub.io/charmhub/pl/qshza6ucetntjma18nixqbpayc).

test/test_testing.py
tonyandrewmeyer: In some places we do this:

```suggestion
            def __init__(self, *args: Any, **kwargs: Any):
                super().__init__(*args, **kwargs)
```

And that's probably better than the `type: ignore`s I guess.

Since `CharmBase` doesn't take `*args`/`**kwargs` it is indeed probably better to be more explicit. It would only be an issue if the signature changed and lots of updates would be needed then anyway.

benhoyt: Ah yeah, I think the `Any`s are fine too, thanks. That's probably better, I guess, as we don't care what the args are here, we just want to forward them.

ops/model.py
tonyandrewmeyer: `stderr` is empty as well :disappointed:.

benhoyt: Oh, I bet I know why -- it's because it's not doing it *now* and so the hook goes through fine, and it's only doing the check at the end of the hook when the reboot is "committed". We ran into this with something else too. Ideally Juju would do the "is this supported on this environment" check early, but oh well.

I bet that `reboot(now=True)` on K8s will give the error right away.

In any case, for now I'd say let's change the comment from "This will silently fail for Kubernetes charms" ("silently fail" is a bit odd) to "This is not supported on Kubernetes charms" and just let them deal.

tonyandrewmeyer: > I bet that `reboot(now=True)` on K8s will give the error right away.

It's difficult to tell, because the Charm is killed before we get back from subprocess.call when `--now` is used.

> In any case, for now I'd say let's change the comment from "This will silently fail for Kubernetes charms" ("silently fail" is a bit odd) to "This is not supported on Kubernetes charms" and just let them deal.

Sure.

ops/testing.py
tonyandrewmeyer: I actually had something similar to this in an earlier version, and ripped it out when I gave up trying to mangle harness into magically re-emitting the event with `now=True`. Bringing it back sounds good to me.

My preference would be for it to return a datetime.datetime indicating when the last (call to) reboot was and `None` if there hasn't been one. It feels like this makes it easier to test for behaviour when there are multiple reboots expected - if it's a bool then I think we would really need to provide a way to reset the flag. An int counter would also work if you'd rather not use a datetime. With either you can still do a plain `self.assertTrue()` if all you care about is that it was rebooted at least once.

I've added the datetime version - lmk if you'd prefer changes.

ops/testing.py
```diff
@@ -1727,6 +1727,14 @@ def handle_timeout(args: testing.ExecArgs) -> int:
             handler=(lambda _: result) if handler is None else handler  # type: ignore
         )
 
+    @property
+    def last_rebooted(self) -> Optional[datetime.datetime]:
```
benhoyt, line 1731: Yep, I like the datetime idea (better than bool) -- thanks!

benhoyt (approved):
Looks good to me, thanks! Let's wait for @jameinel's review too.

ops/testing.py
benhoyt: Good idea, thanks!

jameinel:
> When it's run for a K8s charm, Juju logs an error message but doesn't return an response with the error (or have a non-zero exit), which means that we have to fail silently.

If we can tell that you're in k8s, can't we fail inside `ops` itself? 
I do think Juju should be failing more appropriately (and I don't think there is a reason why we couldn't force the pod to terminate and restart). I don't think people have been asking for that ability, but it would be more consistent to have it do something.

benhoyt:
> If we can tell that you're in k8s, can't we fail inside ops itself?

We don't have an easy way to do that right now. You can look for `containers` in `metadata.yaml`, but charms with no containers can be deployed on K8s too. Harry [mentioned](https://chat.charmhub.io/charmhub/pl/wkxduuqszt8cjdcxq7ihoe8mwr) that `JUJU_CONTAINER_NAMES` will be set on K8s (it may be empty, but it'll be set).

So maybe it's time to add a helper method `is_k8s_sidecar`. We've needed this before too. Or maybe `charm.type` (though something more specific than "type") or something which is an enum, one of `machine`, `k8s-podspec`, `k8s-sidecar`.

test/test_testing.py
```diff
@@ -3337,6 +3337,45 @@ def test_get_pebble_methods(self):
...
+        backend = harness._backend
+        backend.reboot()
+        reboot1 = harness.last_rebooted
+        assert reboot1 is not None
+        self.assertTrue(reboot1)
+        with self.assertRaises(ops.testing.RebootNow):
+            backend.reboot(now=True)
+        reboot2 = harness.last_rebooted
+        assert reboot2 is not None
+        self.assertGreater(reboot2, reboot1)
```
jameinel, line 3367: What is the granularity on datetime.now()? Is it guaranteed to have ticked forward since we called it? (if it is nanoseconds probably yes, but if it is milliseconds I might argue not).
I suppose the test is passing, I do wonder if we would want `>=` vs `>`

Here is an oddity:
```
import datetime
n1 = datetime.datetime.now()
try:
  for i in range(100000):
    n2 = datetime.datetime.now()
    if n2 <= n1:
      raise RuntimeError("n2 did not progress")
    n1 = n2
except RuntimeError as e:
  print(e)
  print(n1, n2)
  print(n2-n1)
```

The loop reliably fails

Anyway, I did get:
```
$ python3 ./dt_test.py
n2 did not progress
2023-10-17 16:11:12.484913 2023-10-17 16:11:12.484913
0:00:00
```

Obviously it depends on lots of factors and just assignment doesn't take a lot of time
```
>>> mindt = None
>>> for i in range(1000000):
...   n2 = datetime.datetime.now()
...   dt = n2 - n1
...   if dt != datetime.timedelta(0):
...     if mindt is None or dt < mindt:
...       mindt = dt
...   n1 = n2
...
>>> mindt
datetime.timedelta(microseconds=1)
```

So datetime does only have a resolution of 1 microsecond, which is pretty fast, but you can, indeed do some python work before it ticks over.

jameinel (approved):
I think having "if harness.last_rebooted is not None" is fine as a way to test that it actually did trigger a reboot if there was some other thing that you were looking for.

I also think we probably want some form of interrupt in python code. I've tested at least:
```
  echo hello
  reboot
  echo goodbye
```

In bash, and you will occasionally see the "goodbye" but not always. I don't know whether `juju-reboot --now` guarantees you will/won't get your process back.
But likely charm authors will assume that if they pass now=True they won't execute any more code.

test/test_testing.py
tonyandrewmeyer: > What is the granularity on datetime.now()?

The smallest unit in `datetime`s is microseconds, and all the units are ints, so it's one microsecond (at best).

> Is it guaranteed to have ticked forward since we called it?

No. I don't think there is any clock in Python that will do that. The closest would be time's `perf_counter_ns`.

What about `reboot_count`? The time isn't particularly important, but I do want it to be more than a Boolean.

tonyandrewmeyer:
> I think just having reboot do something on k8s is also fine, and we wouldn't want to work around the workaround in the future.

If Juju adds support for doing something to k8s when `juju-reboot` is run, and we just have the note in the docs, then updating your Juju version but without ops changing `reboot` will just start working.

If we have something like:

```python
if container_type == "machine":
    raise ModelError("reboot() is only available when running on machine controllers")
```

Then you'd need to update ops as well (to the version where we presumably strip that out). That would be less surprising, which seems good.

> So maybe it's time to add a helper method is_k8s_sidecar. We've needed this before too. Or maybe charm.type (though something more specific than "type") or something which is an enum, one of machine, k8s-podspec, k8s-sidecar.

`charm.platform`? (To match charmhub.io's Platform filter).

tonyandrewmeyer:
> I don't know whether `juju-reboot --now` guarantees you will/won't get your process back. But likely charm authors will assume that if they pass now=True they won't execute any more code.

In my tests, I couldn't get any (Python) code to run after the call to reboot (which was done with `subprocess.run` as in the implementation, so Python would be waiting for the subprocess to exit). I guess we could `sys.exit(0)` in the `now=True` case - the behaviour would essentially be the same as now, but provide more of a guarantee that you didn't get stray lines of code executed.

benhoyt (approved):
Nice, I like it! Approved, with one nit comment on the docstring.

jameinel:
FWIW, I raised the question more so that we wouldn't have a flaky test suite, than because I was worried about the reboot time not being precise enough.
But I'm also happy to go with reboot count instead, as the actual datetime isn't something you could really test.

PR #1046, opened 2023-10-14
ci: remove the temporary mysql-k8s-operator unit tests workaround

tonyandrewmeyer:
With [canonical/mysql-k8s-operator#320](https://github.com/canonical/mysql-k8s-operator/pull/320) merged, we should no longer require the temporary workaround from #1036.


benhoyt (approved):
Thanks!

Issue #736, opened 2022-04-06
Charm should be reinitialized at every hook execution in Harness


PietroPasotti:
More specifically, before every hook execution.
Point is: when a 'real' charm runs, it is reinitialized afresh. 
While if you do fire eventA and eventB on the harness, the harness won't reinitialize the charm it holds in between, which may result in some major behavioural differences between a live charm and a harness'd charm.

Use cases include:

```python
class Charm(CharmBase):
     def __init__(self):
         self.relation1 = relation1 = RelationRequirer()
         self.relation2 = RelationProvider(relation1.data[self.unit])

class Charm(CharmBase):
     def __init__(self):
         if self.container.can_connect():
             self.framework.observe(self.on.foo, self._foo)
```

Concrete example of an affected charm:
https://github.com/canonical/prometheus-k8s-operator/blob/main/src/charm.py

PietroPasotti:
example of a horrible workaround I had to use:
https://github.com/PietroPasotti/traefik-route-k8s-operator/blob/04a32ee3ae67483bf99075fba911e5e62d553801/tests/unit/conftest.py#L82

PietroPasotti:
Very true! We'd have to check if the event being triggered is a custom one or has been fired by charm code.
In 'real life' we could discern between 1) events that are being resurrected (because they had been deferred in the past), 2) events that are 'new' and being manually fired by charm code and 3) 'the event' that is causing **this execution of the charm**.

Only on 3) we want to reinitialize the charm.

If we are not in a live charm, we have to be more clever because there's no context telling us 'this is the event we are running', so a possible workaround would have to set some fancy event flags from within harness.emit().

PietroPasotti:
Another side-effect of this issue is: Harness never fires framework.on.commit
(advanced) features that depend on that hook to fire after every (other) hook execution are not testable with Harness.

PietroPasotti:
@rwcarlsen I wrote a little POC wrapper for us to play with: https://github.com/PietroPasotti/harness-extensions#harness_ctx

sed-i:
I'm working on ingress for prometheus now and this feature is painfully needed.

PietroPasotti:
@sed-i have you tried out the harness_ctx?

sed-i:
> @sed-i have you tried out the harness_ctx?

What's that? (Didn't find mentions in the usual places.)

PietroPasotti:
> What's that? (Didn't find mentions in the usual places.)

[...]

PR #1048, opened 2023-10-15
fix(testing): reset collected statuses in Harness.evaluate_status

benhoyt:
Currently you can't really call `Harness.evaluate_status()` twice in one test because it keeps the previous collected statuses around. Reset them before evaluating.

I believe this addresses https://github.com/canonical/operator/issues/736#issuecomment-1762094066

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1632,8 +1632,10 @@ class Harness(Generic[CharmType]):
         statuses were added.
 
         Tests should normally call this and then assert that ``self.model.app.status``
-        or ``self.model.unit.status`` is the value expected. This method resets the
-        added statuses before triggering the collect-status events.
+        or ``self.model.unit.status`` is the value expected.
+
+        Evaluation is not "additive"; this method resets the added statuses before
+        triggering each collect-status event.
         """
         self.charm.app._collected_statuses = []
         self.charm.unit._collected_statuses = []
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1634,6 +1634,9 @@ class Harness(Generic[CharmType]):
         Tests should normally call this and then assert that ``self.model.app.status``
         or ``self.model.unit.status`` is the value expected.
         """
+        if self._backend._is_leader:
+            self.charm.app._collected_statuses = []
+        self.charm.unit._collected_statuses = []
         charm._evaluate_status(self.charm)
 
     def handle_exec(self,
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer (approved):

test/test_testing.py
```diff
@@ -2828,20 +2828,28 @@ def __init__(self, framework: ops.Framework):
...
-        self.assertEqual(harness.model.app.status, ops.ActiveStatus())
-        self.assertEqual(harness.model.unit.status, ops.BlockedStatus('bar'))
+        self.assertEqual(harness.model.app.status, ops.BlockedStatus('blocked app'))
+        self.assertEqual(harness.model.unit.status, ops.BlockedStatus('blocked unit'))
+
+        harness.charm.app_status_to_add = ops.ActiveStatus('active app')
+        harness.charm.unit_status_to_add = ops.ActiveStatus('active unit')
+        harness.evaluate_status()
+        self.assertEqual(harness.model.app.status, ops.ActiveStatus('active app'))
+        self.assertEqual(harness.model.unit.status, ops.ActiveStatus('active unit'))
```
sed-i, line 2852: So evaluate_status calculates, commit somewhere else, and clears the collection for the next add_status?
Then should probably document that cannot do "additive tests" with evaluate_status.

Not sure which implementation is less confusing though.

ops/testing.py
```diff
@@ -1634,6 +1634,9 @@ def evaluate_status(self) -> None:
         Tests should normally call this and then assert that ``self.model.app.status``
         or ``self.model.unit.status`` is the value expected.
         """
+        if self._backend._is_leader:
+            self.charm.app._collected_statuses = []
```
sed-i, line 1638: The distinction for leader seems inconsistent with the `clear`:
If `evaluate_status` clears the stage for the next `add_status`, it's a harness-level function. Leadership status is on a different level.

test/test_testing.py
benhoyt: Good call. I've updated the `evaluate_status` docstring to clarify that. I don't think "additive tests" is useful for this: before triggering the collect-status events, the statuses added should be reset.

ops/testing.py
benhoyt: Yeah, good call. I've now cleared this independent of `is_leader`.

PR #79, opened 2019-12-10
Add support for functions

dshcherb:
Add support for functions

PR #1053, opened 2023-10-24
feat: add support for testing actions with Harness

### LINKED ISSUE #762: Testing the action event
Right now there is no option how to test action event with Harness.
I believed it should be added as a feature and if I'm not mistaken, it could be an easy task.

Example, how I'm testing it now:
```python
os.environ["JUJU_ACTION_NAME"] = "<action-name>"
harness.charm.framework.model._backend.action_get = mock_action_get = MagicMock()
harness.charm.framework.model._backend.action_set = mock_action_set = MagicMock()
harness.charm.framework.model._backend.action_fail = mock_action_fail = MagicMock()
harness.charm.on.<action-name>_action.emit()
```

tonyandrewmeyer:
Adds support for testing actions with `ops.testing.Harness`.

### New classes in `ops.testing`

* `ActionOutput`: contains the logs and results from (simulating) running an action.
* `ActionFailed`: raised when a (simulated) action run calls `ActionEvent.fail()`.
* `_Operation`: bundles data about an action run.

### New `ops.testing.Harness` method: `run_action()`

* This simulates running an action with Juju, e.g. `juju run name/ord action-name param=value`. Optionally takes additional parameters, and either returns an `ActionOutput` object or raises `ActionFailed`.

### Minor changes
* Adds `additionalProperties` to `CharmMeta` (to allow validating that additional parameters are not passed)

### Comments on state

A reasonable amount of additional state is added to `_TestingModelBackend` (an `_Operation` instance, which contains `None`, and empty dict, and an `ActionOutput` instance, which itself has an empty list and empty dict), which is only required during the `run_action` call. After a `run_action` call this state still exists, but may also contain logs, results, and a failure message from the event handler.

* The initial value of the `_operation` parameter is never used (`run_action` always creates a fresh one). This could be `None`, but then various `assert self._operation is not None` statements are needed for the static checks. This could be the same `_Operation` object and `run_action` could do (e.g.) `self.charm._backend._operation.output.logs.clear()` but that seemed messier. It felt like it could be possible that a simulated `juju list-operations` or `juju show-operation` could be added one day, and the current approach goes toward that (although it'd be a list not one item, and some other adjustments would be required).
* After `run_action` completes, the testing backend state isn't cleared. This would be simple to do, but then we're either creating yet another `_Operation` and `ActionOutput` object that aren't used, or doing multiple `clear()` calls just to wipe out state in a private attribute.

I experimented with a closure-based approach where the various `action_*` backend methods were replaced during the `run_action` call, which localises the state, but that didn't seem to align with the rest of Harness, and left `NotImplemented` methods in the testing backend, which felt wrong. 

I'd love to see review comments on the current approach and whether any of the alternatives mentioned above or ones I didn't consider might work better.

### Other Notes

See [OP041](https://docs.google.com/document/d/1nxyiR7H7ZJZUIOfyIrEudmTg64jDJvPnPLCAeD1R7w0) (Canonical internal link).

Fixes #762.


ops/testing.py
```diff
@@ -2216,17 +2325,34 @@ def _storage_remove(self, storage_id: str):
...
-        raise NotImplementedError(self.action_log)  # type:ignore
-
-    def action_fail(self, message=''):  # type:ignore
-        raise NotImplementedError(self.action_fail)  # type:ignore
+    def action_get(self) -> Dict[str, Any]:
+        action_name = os.environ["JUJU_ACTION_NAME"]
+        params: Dict[str, Any] = {}
+        for name, action_meta in self._meta.actions[action_name].parameters.items():
+            if "default" in action_meta:
+                params[name] = action_meta["default"]
```
tonyandrewmeyer, line 2333: The long way felt more readable, but this could be a comprehension if preferred:

```suggestion
        params: Dict[str, Any] = {
            name: action_meta["default"]
            for name, action_meta in self._meta.actions[action_name].parameters.items()
            if "default" in action_meta}
```

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+        *New in version 2.8*
+
+        Args:
+            action_name: the name of the action to run, as found in ``actions.yaml``.
+            params: override the default parameter values found in ``actions.yaml``.
+                If a parameter is not in ``params``, or ``params`` is ``None``, then
+                the default value from ``actions.yaml`` will be used.
+
+        Raises:
+            ActionFailed: if :meth:`ops.ActionEvent.fail` is called. Note that this will
```
PietroPasotti, line 1817: I'm sure I'm missing some prior context for this design choice, but it feels odd to me to treat action failure as an exception path.

IMHO an exception should be raised by run_action only if the charm raises an uncaught exception. If the action is expected to fail, and the charm exits gracefully after calling action.fail(), then I find it counter-intuitive to have to catch an exception in the test code.

I'd rather assert that the action has failed without coding defensively.

On the other hand, a positive effect of this choice is that it protects the user from missing an action failure. If the action fails, it will do so loudly and make your test fail, so you have to acknowledge that case explicitly OR fix the charm code so the action won't fail. 

In scenario, the philosophy is: you only get an exception if the charm 'exits nonzero', but action.fail() is not necessarily a charm failure.

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+            params = {}
+        for key in action_meta.required:
+            # Juju requires that the key is in the passed parameters, even if there is a default
+            # value in actions.yaml.
+            if key not in params:
+                raise RuntimeError(f"{key!r} property is missing and required.")
+        if not action_meta.additional_properties:
+            for key in params:
+                if key not in action_meta.parameters:
+                    # Match Juju's error message.
```
PietroPasotti, line 1837: Shouldn't we also match the error type then? ModelError?

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+                    raise RuntimeError(
+                        f'additional property "{key}" is not allowed, '
+                        f'given {{"{key}":{params[key]!r}}}')
+        prev_env_action = os.environ["JUJU_ACTION_NAME"]
```
PietroPasotti, line 1841: is it guaranteed to be set? If not, might need a defensive `.get(...)` here.

ops/testing.py
```diff
@@ -1963,6 +2071,7 @@ def __init__(self, unit_name: str, meta: charm.CharmMeta, config: '_RawConfig'):
         self._opened_ports: Set[model.Port] = set()
         self._networks: Dict[Tuple[Optional[str], Optional[int]], _NetworkDict] = {}
         self._reboot_count = 0
+        self._operation = _Operation(ActionOutput([], {}), {})
```
PietroPasotti, line 2074: I think `_Operation` is not descriptive enough a name. How about `_ActionContext` or `_RunningAction`?

test/test_testing.py
```diff
@@ -5232,3 +5232,169 @@ def handler(args: ops.testing.ExecArgs):
...
+            self.assertEqual(out.results, {})
+
+    def test_invalid_action(self):
+        with self.assertRaises(RuntimeError):
+            self.harness.run_action("another-action")
+        with self.assertRaises(RuntimeError):
+            self.harness.run_action("bad_results")
+
+    def test_run_action(self):
+        out = self.harness.run_action("test")
```
PietroPasotti, line 5360: if you run this test in isolation, that is, without running `test_action_get` before (which sets JUJU_ACTION_NAME and does not clean up the env), I expect you'll get a KeyError from that `os.environ[JUJU_ACTION_NAME]` call.

ops/testing.py
```diff
@@ -135,6 +135,47 @@ class ExecResult:
...
 
 
+@dataclasses.dataclass(frozen=True)
+class ActionOutput:
+    """Contains the logs and results from a :meth:`Harness.run_action` call."""
+
+    logs: List[str]
+    """Messages generated by the Charm using :meth:`ops.ActionEvent.log`."""
+
+    results: Dict[str, str]
```
PietroPasotti, line 145:
```suggestion
    results: Dict[str, Any]
```

ops/testing.py
```diff
@@ -135,6 +135,47 @@ class ExecResult:
...
+    """The action's results, as set or updated by :meth:`ops.ActionEvent.set_results`.
+
+    The results dictionary is flattened to a :code:`dict[str, str]` when communicating with
+    Juju, and Juju restores it back to a multi-level dict. This attribute is the non-flat
+    version, so may be a multi-level dict, and calling :meth:`ops.ActionEvent.set_results`
```
PietroPasotti, line 150: so it should be str:Any, no?

PietroPasotti (changes requested):
A few discussion points that I'd like to be addressed, and one possible bug.
Overall looks great! 
:100:

ops/testing.py
benhoyt: I can see where you're coming from, and we actually had it the other way to begin with, with an `ActionOutput.success` bool attribute. But we thought it'd be too easy to forget to `assert action_output.success` in charm tests, and the charm action could call `event.fail()` but the test would completely miss that.

That's the main reason. In addition, most tests would require `assert action_output.success` (as presumably the non-failure case is much more common), so it'd also make most tests -- of the success case -- shorter.

You're also right the other way around -- that failing loudly protects the user from missing an action failure.

ops/testing.py
benhoyt: That gave me pause too. The rationale is that Juju calls these things "operations": `juju operations`, `juju show-operation`, etc. However, in this case I'm inclined to agree with Pietro that something like `_ActionInfo` is better. We don't have the concept of "operation" in ops yet, and by itself it seems too vague.

ops/testing.py
```diff
@@ -2216,17 +2325,34 @@ def _storage_remove(self, storage_id: str):
...
+        action_name = os.environ["JUJU_ACTION_NAME"]
+        params: Dict[str, Any] = {}
+        for name, action_meta in self._meta.actions[action_name].parameters.items():
+            if "default" in action_meta:
+                params[name] = action_meta["default"]
+        params.update(self._operation.parameters)
+        return params
+
+    def action_set(self, results: Dict[str, Any]):
+        for key in "stdout", "stderr", "stdout-encoding", "stderr-encoding":
```
benhoyt, line 2338: Huh, I've never seen the implicit-comma-builds-tuple used in this context. I think it's slightly clearer to be explicit:

```suggestion
        for key in ["stdout", "stderr", "stdout-encoding", "stderr-encoding"]:
```

test/test_testing.py
```diff
@@ -5232,3 +5232,169 @@ def handler(args: ops.testing.ExecArgs):
...
+                  default: foo-default
+                bar:
+                  type: integer
+            bad-results:
+              description: incididunt ut labore
+            ''')
+        self.harness.begin()
+
+    def test_action_get(self):
+        op = ops.testing._Operation(ops.testing.ActionOutput([], {}), {})
```
benhoyt, line 5294: Hmm, I'm not sure about this. I see what you're doing here: testing the testing-backend separately from the `run_action` method. But I don't think we've generally done that (the implementations are pretty intertwined anyway!) and I think it's less fragile to do just black-box testing here by testing that `run_action` expects/returns the correct things. That way we can change the implementation details / name of `_Operation` / etc without changing the tests.

Do the tests below test all the code paths in any case? If so, I'd say that's good enough.

Happy for push-back if I'm missing something though.

benhoyt (changes requested):
Great start, thanks! I've left a bunch of comments, with some questions about the env var handling (and whether we need it at all) and the approach to testing.

ops/testing.py
tonyandrewmeyer: Good catch, thanks! It's strange that pyright didn't complain about the `action_set` line that does a dictionary update of a dict[str, Any] into a dict[str, str].

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+            params: override the default parameter values found in ``actions.yaml``.
+                If a parameter is not in ``params``, or ``params`` is ``None``, then
+                the default value from ``actions.yaml`` will be used.
+
+        Raises:
+            ActionFailed: if :meth:`ops.ActionEvent.fail` is called. Note that this will
+              be raised at the end of the ``run_action`` call, not immediately when
+              :code:`fail()` is called, to match the run-time behaviour.
+        """
+        if self._charm is None:
```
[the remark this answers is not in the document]
tonyandrewmeyer: Sure. I was just copying the `update_config` behaviour.

ops/testing.py
```diff
@@ -2216,17 +2325,34 @@ def _storage_remove(self, storage_id: str):
...
-    def action_set(self, results):  # type:ignore
-        raise NotImplementedError(self.action_set)  # type:ignore
-
-    def action_log(self, message):  # type:ignore
-        raise NotImplementedError(self.action_log)  # type:ignore
-
-    def action_fail(self, message=''):  # type:ignore
-        raise NotImplementedError(self.action_fail)  # type:ignore
+    def action_get(self) -> Dict[str, Any]:
+        action_name = os.environ["JUJU_ACTION_NAME"]
```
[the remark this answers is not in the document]
tonyandrewmeyer: We could, but we would need to remove the check in `ActionEvent.restore` that the environment variable matches the action (or work around it in uglier ways). It does look like that's more of a sanity check than anything else (the value isn't used other than to raise an error if it's wrong) but maybe there's a reason that it's there that I'm not aware of? In #79 where this was added, it says:

> As a part of that, we also need to make sure that emitting random function events without the necessary context is forbidden as well.

The testing code would definitely be much simpler with it all removed, and it seems an odd check in the charm code, but perhaps I'm missing history/context?

ops/testing.py
tonyandrewmeyer: > Huh, I've never seen the implicit-comma-builds-tuple used in this context. I think it's slightly clearer to be explicit:

Ok (my memory was that black trimmed them in that case, but (a) autopep8 doesn't, and (b) a quick test seems to indicate it doesn't). However, a tuple not a list:

```suggestion
        for key in ("stdout", "stderr", "stdout-encoding", "stderr-encoding"):
```

tonyandrewmeyer: Hmm, in older versions of Python a tuple was fastest (and to my eyes looks cleanest - a frozenset is arguably the most correct but is the ugliest and slowest). However, it seems like in Python 3.8+ a list is faster (although a bit heavier in terms of memory).

I still like the look of a tuple most, so will go with that, but re-open this if you disagree :)

test/test_testing.py
tonyandrewmeyer: Ok, happy to do that. I started out with just the other (higher level) tests, and then added these ones in for completeness. I do agree there is a lot of overlap. There's a little bit that's tested here and not below, but I can merge those together.

ops/testing.py
benhoyt: I *definitely* don't think we need to worry about speed here. I'm fine with either a tuple or a list, so tuple is fine!

However, just for the record, here's how I usually think about the distinction: it's less about immutable vs mutable, and more about: tuples are "records with positional fields", and lists are "a list of same-typed items". So I'd use `(name, email)` but `[name1, name2, ...]`. There's a fair bit of support for this take [on StackOverflow](https://stackoverflow.com/a/626871/68707) and [in the official tutorial](https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences), which says:

> Though tuples may seem similar to lists, they are often used in different situations and for different purposes. Tuples are [immutable](https://docs.python.org/3/glossary.html#term-immutable), and usually contain a heterogeneous sequence of elements that are accessed via unpacking (see later in this section) or indexing (or even by attribute in the case of [namedtuples](https://docs.python.org/3/library/collections.html#collections.namedtuple)). Lists are [mutable](https://docs.python.org/3/glossary.html#term-mutable), and their elements are usually homogeneous and are accessed by iterating over the list.

ops/testing.py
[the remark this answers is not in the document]
benhoyt: I've looked at that check, and yeah, I agree it doesn't pay for itself. I don't think we should be depending on environment variables at that level in the code anyway. Let's kill it!

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+            for key in params:
+                if key not in action_meta.parameters:
+                    # Match Juju's error message.
+                    raise RuntimeError(
+                        f'additional property "{key}" is not allowed, '
+                        f'given {{"{key}":{params[key]!r}}}')
+        prev_env_action = os.environ["JUJU_ACTION_NAME"]
+        try:
+            os.environ["JUJU_ACTION_NAME"] = action_name
+            if not self._hooks_enabled:
```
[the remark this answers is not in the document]
PietroPasotti: +1 on this, action events are manually initiated and should ignore the hooks-enabled machinery

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+                return ActionOutput([], {})
+            self._backend._operation = _Operation(ActionOutput([], {}), params)
+            getattr(self.charm.on, f"{action_name.replace('-', '_')}_action").emit()
```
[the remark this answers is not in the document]
tonyandrewmeyer: Ha, that's the code I originally had :). When I was reviewing the PR it looked a bit wordy so I condensed it down. Happy to go back.

ops/testing.py
```diff
@@ -1746,6 +1787,73 @@ def reboot_count(self) -> int:
...
+            del os.environ["JUJU_ACTION_NAME"]
```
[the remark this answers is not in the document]
tonyandrewmeyer: It's not - it's leftover from before I decided to put in the reset and I missed removing it, thanks!

ops/testing.py
```diff
@@ -2216,17 +2325,34 @@ def _storage_remove(self, storage_id: str):
...
+        for key in "stdout", "stderr", "stdout-encoding", "stderr-encoding":
+            if key in results:
+                # Match Juju's error message.
+                raise model.ModelError(f'ERROR cannot set reserved action key "{key}"')
+        # Although it's not necessary, we go through the same flattening process
+        # as the real backend, in order to give Charmers advance notice if they
+        # are setting results that will not work.
+        # This also does some validation on keys to make sure that they fit the
+        # Juju constraints.
+        model._format_action_result_dict(results)
```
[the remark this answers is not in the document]
tonyandrewmeyer: Alternatively, I could split `_format_action_result_dict` into validation and formatting (at the cost of looping through the (likely small) dict twice, as well as potentially a few extra function calls since it's done recursively). The comment works for me, so will do that unless you suggest otherwise :)

Fwiw, I originally (and this was somewhat implied in the spec) planned to do the flatten and then do the unflatten that's done on the Juju side, and give back that resulting object, so it would have been used. However, when implementing the unflatten part it seemed a bit pointless to be doing that (but I do feel it's valuable to do the flatten to get any errors there early).

ops/testing.py
tonyandrewmeyer: Ok, fine with me. I indeed just pulled the name from Juju (I originally had a bunch of individual attributes and it seemed noisy so needed a name to bundle them under). If we do ever add "list-operations" etc (I couldn't think of any use case for that) then it's simple enough to rename the private class at the same time as expanding it to have everything else it would need.

benhoyt (approved):
This is looking good now, thanks!

PR #1056, opened 2023-11-06
docs: correct the version that Harness.run_action was added

tonyandrewmeyer:
When I opened the original PR, I assumed it would squeeze into 2.8, but then forgot to update that to 2.9 before we merged, sorry.


benhoyt (approved):
Ah, good catch

PR #1057, opened 2023-11-07
fix: reduce mode of SQLite storage file to 0o600

### LINKED ISSUE #1051: .unit-state.db is world-readable
.unit-state.db created by an operator framework charm is visible by any user in the system. The permission should be tighten up.

This is one example of charms that saved a private key for a TLS certificate.
```
$ sudo -u nobody strings /var/lib/juju/agents/unit-ceph-dashboard-1/charm/.unit-state.db | grep PRIVATE
-----BEGIN RSA PRIVATE KEY-----
-----END RSA PRIVATE KEY-----
```

tonyandrewmeyer:
sqlite3 creates new database files with 0o600 permissions. Although Charms should generally not store private data in the backend storage, this does happen sometimes, and it would be nicer to ensure that only the user (that is, root) can access the underlying storage file.

This chmod's the file (to `-rw-------`) after every open even if it already exists. It seems simpler to do this and get the additional access fix rather than only do it on initial creation.

Fixes #1051.


test/test_storage.py
```diff
@@ -218,6 +219,17 @@ class TestSQLiteStorage(StoragePermutations, BaseTestCase):
     def create_storage(self):
         return ops.storage.SQLiteStorage(':memory:')
 
+    def test_permissions(self):
+        fd, filename = tempfile.mkstemp()
```
benhoyt, line 223: As discussed, let's use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory with a `with statement` for this.

benhoyt (changes requested):

benhoyt (approved):
Looks good, thanks!

ops/storage.py
```diff
@@ -62,6 +63,8 @@ def __init__(self, filename: Union['Path', str]):
         self._db = sqlite3.connect(str(filename),
                                    isolation_level=None,
                                    timeout=self.DB_LOCK_TIMEOUT.total_seconds())
+        if filename != ":memory:":
```
jameinel, line 66: Offhand, is there any reason why a failure of this command (for any reason), should cause it to fail?
I wonder if we want to wrap it with at try/finally.

Officially, if we *really* care about the security of this, this particular framing creates a short window of security hole, where the file is created, someone could see it, open it, and then it gets chmod away. (This was the feedback from the Security team for our juju-run socket.)

If we want it to be correct, the correct action is to do something like:
```
    os.close(os.open(path, flags, 0o600))
    self._db = sqlite3.connect(...)
```
So we always create the file with the right perms.
I don't know that it is worth that overhead, but it does eliminate any form of race condition if we genuinely care about security.
(And we should also have a way for pre-existing files to make sure they get set correctly, so it is probably something like:
def _ensure_db_permissions(self, filename):
    filepth = pathlib.Path(filename)
    if filepth.exists():
      os.chmod()
    else:
      try:
        fd = os.open(..., 0o666)
      except ???
         ...
      try:
        ...
      finally:
         os.close(fd)
```

I don't think it is a particularly complicated function, but ensuring that you open files with secure permissions is pretty well know if we really want to be secure here.

**Comment by jameinel** [comment]
I wouldn't block this, but I would say that if we really want to care about the security of this file, we might as well take the step of creating it secure in the first place to avoid ever having a file handle that an attacker could exploit.

File: ops/storage.py (hunk 2 above)
Comment (benhoyt) [reply ops/storage.py:66]: Yeah, I think that's reasonable, John -- thanks for the review. Might as well fix it properly and avoid the race condition. And let's put a try/except around the `chmod` as you suggest, and log the error instead of fail hard.

I believe the flags for the `os.open` call should be `os.open(filename, os.O_CREAT, mode=0o600)`.

**Comment by tonyandrewmeyer** [comment]
Moving back to draft until I verify the new approach with a real (non-unit, manual) test.

File: ops/storage.py (hunk 2 above)
Comment (tonyandrewmeyer) [reply ops/storage.py:66]: > Might as well fix it properly and avoid the race condition.

Fine with me. When we talked about this last week @benhoyt you thought that there was also a need to put the file in a secured directory (for handling temporary files that sqlite uses for atomicity, I think). I did some basic searching and couldn't find anything about this. Do you remember anything more specific, or should I ignore that for now?

> And let's put a try/except around the `chmod` as you suggest, and log the error instead of fail hard.

I'm not convinced about this. I could definitely be convinced that we should add a try/except and raise a clearer error. If we just log the error and don't hard fail, doesn't this effectively undermine the security? We're then relying on the admin to be checking the log and noticing this error if something goes wrong with securing the permissions. That doesn't feel right to me.

> I believe the flags for the `os.open` call should be `os.open(filename, os.O_CREAT, mode=0o600)`.

I think it needs to be `O_CREAT | O_EXCL`, because otherwise there's a small race window between the check if the file already exists and creating the new file. With `O_EXCL` if someone creates the file in that window then the open will fail.

Comment (tonyandrewmeyer) [reply ops/storage.py]: > I'm not convinced about this. I could definitely be convinced that we should add a try/except and raise a clearer error. If we just log the error and don't hard fail, doesn't this effectively undermine the security? We're then relying on the admin to be checking the log and noticing this error if something goes wrong with securing the permissions. That doesn't feel right to me.

Thinking a bit more: I could maybe be convinced that the already-exists case should not hard fail, since you're just not getting the "upgrade" in security if it fails. However, I am definitely not convinced that the new-file case should just log the error.

**Comment by tonyandrewmeyer** [comment]
Manually verified the simple cases on micro-k8s (no existing state file, existing state file).

File: ops/storage.py (hunk 2 above)
Comment (benhoyt) [reply ops/storage.py]: The directory thing I mentioned was that the directory the sqlite file needs to be writeable, so sqlite can create the WAL file(s) in it. See https://sqlite.org/wal.html and search for "directory". We could try to pre-create those files too, but I think that's probably getting a bit carried away?

Regarding the exception handling: I think you're right: if we're going to layer on a bit of security, we should hard fail if it doesn't work (at least in the new-file case).

Good catch on the `O_EXCL` issue.

File: ops/storage.py [hunk 3]
```diff
@@ -59,11 +60,29 @@ def __init__(self, filename: Union['Path', str]):
...
 
+    def _ensure_db_permissions(self, filename: str):
+        """Make sure that the DB file has appropriately secure permissions."""
+        mode = stat.S_IRUSR | stat.S_IWUSR
+        if os.path.exists(filename):
+            try:
+                os.chmod(filename, mode)
+            except OSError as e:
+                logger.warning(f"Unable to adjust permissions of storage file {filename!r}: {e}")
+            return
```
benhoyt, line 78: Nit: I think this code would be slightly clearer with a blank line after the `return`.

ops/storage.py
```diff
@@ -59,11 +60,29 @@ def __init__(self, filename: Union['Path', str]):
...
+        try:
+            fd = os.open(filename, os.O_CREAT | os.O_EXCL, mode=mode)
+        except OSError as e:
+            logger.warning(f"Unable to ensure permissions of storage file {filename!r}: {e}")
```
benhoyt, line 82: Per separate discussion, I guess we'd remove the try/except here.

test/test_storage.py
```diff
@@ -218,6 +219,13 @@ class TestSQLiteStorage(StoragePermutations, BaseTestCase):
     def create_storage(self):
         return ops.storage.SQLiteStorage(':memory:')
 
+    def test_permissions(self):
```
benhoyt, line 222: Can we also add a test for the existing-file case now that we have a separate code path for that?

benhoyt (changes requested):

benhoyt (approved):
Looks good, and thanks for the tests. Per discussion, we'll fail hard in both cases. As Tony said on Mattermost:

> It does seem like if we are going to add this (minimal) layer of security then we ought to ensure that it either happens or fails, so that people don't assume that it's happening when it's not.

PR #1061, opened 2023-11-09
ci: use a trusted publisher token for publishing to PyPI

### LINKED ISSUE #1021: Use PyPI's "Trusted Publisher" functionality to publish packages
It'd likely be worth moving to PyPI's [Trusted Publisher](https://docs.pypi.org/trusted-publishers/) system (basically, using OIDC rather than API tokens or username/passwords). From a security point of view it's tighter because the tokens are short-lived and scoped tightly to the specific package.

tonyandrewmeyer:
Remove the use of the PyPI secret in the publish workflow, in favour of using the short-lived OIDC token via the Trusted Publisher system.

> PyPI's normal API tokens are long-lived, meaning that an attacker who compromises a package's release token can use it until its legitimate user notices and manually revokes it. Similarly, uploading with a password means that an attacker can upload to any project associated with the account. Trusted publishing avoids both of these problems: the tokens minted expire automatically, and are scoped down to only the packages that they're authorised to upload to.

Also add a (manually triggered) workflow to publish to test.pypi.org to be able to validate that publishing changes (like this one) work as expected, without needing to actually publish.

Fixes #1021


.github/workflows/test-publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Test Publishing
```
benhoyt, line 1: Just to match the "Publish" workflow name:

```suggestion
name: Test Publish
```

.github/workflows/test-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  framework-tests:
+    uses: ./.github/workflows/framework-tests.yaml
+  observability-charm-tests:
+    uses: ./.github/workflows/observability-charm-tests.yaml
+  hello-charm-tests:
+    uses: ./.github/workflows/hello-charm-tests.yaml
+  build-n-publish:
+    name: Build and Publish to PyPI
```
benhoyt, line 12:
```suggestion
    name: Build and Publish to Test PyPI
```

benhoyt (approved):
Looks good. Let's just tweak a couple of names. However, let's wait to merge this till you hear back from test.pypi.org support and have ownership of `ops`.

tonyandrewmeyer:
> Looks good. Let's just tweak a couple of names. However, let's wait to merge this till you hear back from test.pypi.org support and have ownership of `ops`.

It looks like the backlog for these requests goes back roughly one year :disappointed:. Some get handled "out of queue" - I did find [one that's exactly the same situation](https://github.com/pypi/support/issues/2891) and it was one of the "out of queue" ones, so I've adjusted the issue summary to make it clearer that it's only test.pypi.org, but that doesn't seem like it always gets quicker action.

So I think for at least this PR, we should just go ahead with it and can address any issues if they happen.

benhoyt:
> So I think for at least this PR, we should just go ahead with it and can address any issues if they happen.

Yep, sounds right to me, thanks!

PR #1065, opened 2023-11-15
docs: fix Harness.get_filesystem_root example

### LINKED ISSUE #1063: Misleading docstring for `get_filesystem_root`
The docstring for [`Harness.get_filesystem_root`](https://ops.readthedocs.io/en/latest/#ops.testing.Harness.get_filesystem_root) suggests that:
1. Using plain `open` in charm code operates on workload containers;
2. Passing "charm" as an arg to `get_filesystem_root` would give the charm container.

However, both are wrong:
1. A plain `open` would operate on the charm container;
2. Cannot pass "charm" as an arg: `return self._backend._pebble_clients[container_name]._root KeyError: 'charm'`

https://github.com/canonical/operator/blob/848fb52e8d953e20a8e37ceb4f13fe791cd3fa7b/ops/testing.py#L1638-L1650

benhoyt:
Fixes #1063

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1636,18 +1636,22 @@ class Harness(Generic[CharmType]):
         Example usage::
 
             # charm.py
-            import ops
             class ExampleCharm(ops.CharmBase):
                 def __init__(self, *args):
                     super().__init__(*args)
-                    self.hostname = open("/etc/hostname").read()
+                    self.framework.observe(self.on["mycontainer"].pebble_ready,
+                                           self._on_pebble_ready)
+
+                def _on_pebble_ready(self, event: ops.PebbleReadyEvent):
+                    self.hostname = event.workload.pull("/etc/hostname").read()
 
             # test_charm.py
-            from ops.testing import Harness
             harness = Harness(ExampleCharm)
             root = harness.get_filesystem_root("mycontainer")
+            (root / "etc").mkdir()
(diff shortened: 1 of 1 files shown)


sed-i (approved):
Awesome, thanks @benhoyt!

ops/testing.py
```diff
@@ -1636,18 +1636,22 @@ def get_filesystem_root(self, container: Union[str, Container]) -> pathlib.Path:
...
 
             # test_charm.py
-            from ops.testing import Harness
             harness = Harness(ExampleCharm)
             root = harness.get_filesystem_root("mycontainer")
+            (root / "etc").mkdir()
             (root / "etc" / "hostname").write_text("hostname.example.com")
-            harness.begin()
+            harness.begin_with_initial_hooks()
+            assert harness.charm.hostname == "hostname.example.com"
```
tonyandrewmeyer, line 1655: This is only tangentially related, but while we're looking at this example, do you think it's worth adjusting the example test to call `cleanup()` since one of the key things that does is clean up the temporary directory? Or is that too much noise here and we expect that people will instantiate a Harness object in setUp/a fixture with cleanup there, copying that from somewhere else?

tonyandrewmeyer (approved):
Looks good!

ops/testing.py
benhoyt: Yeah, good call. Updated.

Issue #1037, opened 2023-10-10
Simplify testing code that checks the Juju version


tonyandrewmeyer:
Charms, such as [mysql-k8s-operator](https://github.com/canonical/mysql-k8s-operator/blob/128d32fd720f65b18a69650bb827dddf64a2b940/lib/charms/mysql/v0/mysql.py#L585), have code like:

```python
        if ops.jujuversion.JujuVersion.from_environ().has_secrets:
```

in order to determine whether Juju features are available (and behave differently as a result). However, the default version for `from_environ()` is `0.0.0`, which means that all of the `has_*` methods that do a version check will return False if the version isn't in the environment variables.

The Charmer can work around this, e.g.:

```python
@pytest.fixture
def with_juju_secrets(mocker: MockerFixture):
    """Ensure that JujuVersion.has_secrets returns True."""
    mocker.patch.object(
        JujuVersion, "has_secrets", new_callable=PropertyMock
    ).return_value = True
```

or more indirectly by mocking the environment variable `JUJU_VERSION`. However, we could make this nicer for Charmers.

One possibility would be to reverse the situation so that `JujuVersion.from_environ()` defaults to the latest stable version if `JUJU_VERSION` is not in the environment variables. This more closely reflects the `testing.Harness` behaviour where (simulated) secret functionality is always available. In production, it's less clear, so maybe this is only done when testing?

Alternatively, or additionally, we could provide a mechanism for specifying the (simulated) Juju version, particularly if we are already going to be patching that. For example:

```python
def test_foo(self):
    with ops.testing.juju_version("3.3.3"):
        self.harness.charm.bar()
```

or

```python
def test_foo(self):
    with ops.testing.juju_features(secrets=True, open_port_on_k8s=False):
        self.harness.charm.bar()
```

(This might give weird behaviour if you're simulating a feature set mix that never existed in Juju).

benhoyt:
Yeah, agreed this is a good idea. We'll have to discuss the exact API. It's probably going to be simpler to use and more discoverable if we make it a `Harness` method, probably `def set_juju_version(self, version: str) -> None`.

I think we should probably leave the production 0.0.0 fallback alone (unless there's an easy way to get the actual Juju version?) but change the default for testing to the latest stable version of Juju.

tonyandrewmeyer:
[...]

PR #1067, opened 2023-11-19
feat: secret owners do not auto-peek, and can use refresh

### LINKED ISSUE #1062: [testing] ValueError: Secret owner cannot use refresh=True
This error occurs only on unittests, relating to `Harness`.

The problem is that -- this false alarm prevents us from being able to realitically test charms code... We need to apply workarounds either on the tests (that's well elaborate), or even worse, on the code... Which is highly incorrect -- yet keeps us secure in case it may be more than a `Harness` bug (at any point)...)

All in all: reliable unittesting is _essential_ to be able to deliver reliable code... If you could please address it with "non-negligable" priority :-)

Thank you.

The issue have been discussed in https://chat.charmhub.io/charmhub/pl/aekfocgk8tbtfcsda7311uurxe

tonyandrewmeyer:
Update the testing harness to reflect an upcoming Juju change regarding secrets: the special-casing for secret-get is going to be removed, so that owners have the same behaviour as everyone else - they get the tracked revision of the secret, and can call peek to get the latest content, or refresh to get the latest content and update the tracking pointer.

Also extends the Harness tests for getting secrets.

Fixes #1062


tonyandrewmeyer:
@benhoyt it seems like we don't have any tests asserting the previous behaviour (nothing fails when it's removed, and I couldn't manually find any). Do you think we should add tests for the new behaviour, e.g. where the owner uses peek=False or refresh=True?

tonyandrewmeyer:
@wallyworld just to confirm: the secret owner always has permission to do a secret-get, right? ie. in the testing harness, we're checking that a grant exists for the relevant secret+relation, but skipping that for the owner. Does that sound right to you?

benhoyt:
> @benhoyt it seems like we don't have any tests asserting the previous behaviour (nothing fails when it's removed, and I couldn't manually find any). Do you think we should add tests for the new behaviour, e.g. where the owner uses peek=False or refresh=True?

Because this is now the same for owners and observers, I don't think we need any owner-specific tests for it. (I'm presuming the `peek` and `refresh` code paths are tested in general.)

wallyworld:
> @wallyworld just to confirm: the secret owner always has permission to do a secret-get, right? ie. in the testing harness, we're checking that a grant exists for the relevant secret+relation, but skipping that for the owner. Does that sound right to you?

When a secret is created, the owner gets "manage" permission. So they can read the value, as well as grant access etc

tonyandrewmeyer:
> Because this is now the same for owners and observers, I don't think we need any owner-specific tests for it. (I'm presuming the `peek` and `refresh` code paths are tested in general.)

We have tests that check that `get_secret` works when the app/unit has been granted access, and that it fails when no grant exists.

We don't have a test that checks the 'owner has manage permission' case. We also seem to only have tests that provide the ID, so don't check getting by label or updating the label (not even using `get_info()` - I can't find any harness tests for that?). We don't seem to test the peek or refresh paths.

So I think the conclusion here is that a few extra tests would be a good idea, even if they aren't specifically about this change.

benhoyt (approved):
Looks good to me. Thanks for the additional tests.

PietroPasotti:
what juju version will this be?

benhoyt:
@wallyworld Can you remind us what Juju version(s) the secrets owner fix will be going into?

@tonyandrewmeyer I wonder, maybe we should have a `JujuVersion` property to allow testing for this (and simply to record the version clearly).

tonyandrewmeyer:
> @tonyandrewmeyer I wonder, maybe we should have a `JujuVersion` property to allow testing for this (and simply to record the version clearly).

I wondered about this in the original MM thread too. It seems like Juju is treating this as a bug (at least in the way that it's being changed in existing versions), and I don't think we would want to deliberately model behaviour of bugs in specific Juju versions. I don't think we want JujuVersion to end up being a Juju changelog, or to have Charmers running large matrices of tests with different simulated Juju versions.

Otoh, it doesn't seem like a bug to me, and JujuVersion already has checks that go to the patch level.

The default JujuVersion of 0.0.0 isn't ideal here, though. That means the default behaviour would be the old (flawed/buggy) approach, and you'd need to explicitly set the Juju version to get the new (good/fixed) behaviour. Perhaps that's more of an argument for resolving #1037, though.

Overall, I'm -0 on adding a JujuVersion check and switching the behaviour based on that. If we're confident that at least some Chamers would write charms that change their behaviour to handle both Juju versions and have tests to verify that, that'd move me to at least +0.

benhoyt:
> Overall, I'm -0 on adding a JujuVersion check and switching the behaviour based on that

Yeah, you've convinced me -- let's leave that out.

benhoyt (approved):
Excellent, thanks for this

PR #1074, opened 2023-11-22
feat: expose the ability to set Pebble log targets

### LINKED ISSUE #1073: Layer object support for log-forwarding as a field
`log-targets` has been added as a field to pebble as of https://github.com/canonical/pebble/pull/267. However, the current implementation of ops.pebble.Layer discards this from the argument of the init. This must be handled properly so that the layer yaml file could be produced.

tonyandrewmeyer:
Exposes the `log-targets` section of the Pebble layer spec, to allow charms to configure log forwarding in Pebble without having to provide layers as plain YAML.

New types:

* `LogTargetDict`: a typed dictionary similar to `ServiceDict` and `CheckDict`.
* `LogTarget`: similar to `Service` and `Check`.

Changed types:

* `LayerDict` and `PlanDict` have (not required) `log-targets` keys.
* `Layer` and `Plan` have `log_targets` attributes (`Plan` is read-only, as with `checks` and `services`).

Fixes #1073

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -55,7 +55,7 @@ pytest
 The framework has some tests that interact with a real/live Pebble server.  To
 run these tests, you must have [pebble](https://github.com/canonical/pebble)
 installed and available in your path.  If you have the Go toolchain installed,
-you can run `go install github.com/canonical/pebble/cmd/pebble@latest`.  This will
+you can run `go install github.com/canonical/pebble/cmd/pebble@master`.  This will
 install pebble to `$GOBIN` if it is set or `$HOME/go/bin` otherwise.  Add
 `$GOBIN` to your path (e.g. `export PATH=$PATH:$GOBIN` or `export
 PATH=$PATH:$HOME/go/bin` in your `.bashrc`) and you are ready to run the real
@@ -78,10 +78,6 @@ tox -e unit -- test/test_real_pebble.py
 source .tox/unit/bin/activate
 pytest -v test/test_real_pebble.py
 ```
-
-To test Pebble functionality that has been merged but not yet released, you can run
-`go install github.com/canonical/pebble/cmd/pebble@master`. This otherwise works in
-the same way as the install described above.
 
 # Documentation
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -52,20 +52,20 @@ pytest
 
 ## Pebble Tests
 
-The framework has some tests that interact with a real/live pebble server.  To
+The framework has some tests that interact with a real/live Pebble server.  To
 run these tests, you must have [pebble](https://github.com/canonical/pebble)
 installed and available in your path.  If you have the Go toolchain installed,
 you can run `go install github.com/canonical/pebble/cmd/pebble@latest`.  This will
 install pebble to `$GOBIN` if it is set or `$HOME/go/bin` otherwise.  Add
 `$GOBIN` to your path (e.g. `export PATH=$PATH:$GOBIN` or `export
 PATH=$PATH:$HOME/go/bin` in your `.bashrc`) and you are ready to run the real
-pebble tests:
+Pebble tests:
 
 ```sh
 tox -e pebble
 ```
 
-To do this even more manually, you could start the pebble server yourself:
+To do this even more manually, you could start the Pebble server yourself:
 
 ```sh
 export PEBBLE=$HOME/pebble
@@ -78,6 +78,10 @@ tox -e unit -- test/test_real_pebble.py
 source .tox/unit/bin/activate
 pytest -v test/test_real_pebble.py
 ```
+
+To test Pebble functionality that has been merged but not yet released, you can run
+`go install github.com/canonical/pebble/cmd/pebble@master`. This otherwise works in
+the same way as the install described above.
 
 # Documentation
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -751,7 +751,7 @@ class Plan:
 
     @property
     def log_targets(self) -> Dict[str, 'LogTarget']:
-        """This plan's log targets mapping (maps log targer name to :class:`LogTarget`).
+        """This plan's log targets mapping (maps log target name to :class:`LogTarget`).
 
         This property is currently read-only.
         """
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -129,8 +129,8 @@ LogTargetDict = typing.TypedDict('LogTargetDict',
                                  {'override': Union[Literal['merge'], Literal['replace']],
                                   'type': Literal['loki'],
                                   'location': str,
-                                  'services': Optional[List[str]],
-                                  'labels': Optional[Dict[str, str]]},
+                                  'services': List[str],
+                                  'labels': Dict[str, str]},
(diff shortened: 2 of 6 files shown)


ops/pebble.py
```diff
@@ -735,11 +749,20 @@ def checks(self) -> Dict[str, 'Check']:
         """
         return self._checks
 
+    @property
+    def log_targets(self) -> Dict[str, 'LogTarget']:
+        """This plan's log targets mapping (maps log targer name to :class:`LogTarget`).
```
tonyandrewmeyer, line 754:
```suggestion
        """This plan's log targets mapping (maps log target name to :class:`LogTarget`).
```

Kayra1 (approved):
Thanks for the fast response!

ops/pebble.py
```diff
@@ -1028,6 +1056,48 @@ class CheckStatus(enum.Enum):
...
 
 
+class LogTarget:
+    """Represents a log target in a Pebble configuration layer."""
+
+    def __init__(self, name: str, raw: Optional['LogTargetDict'] = None):
+        self.name = name
+        dct: LogTargetDict = raw or {}
+        self.override: str = dct.get('override', '')
+        self.type_ = dct.get('type', '')
```
benhoyt, line 1066: I think we should just call this `self.type` (similar to `FileInfo.type`). `type()` is a built-in function, but there's nothing wrong with using it as an instance attribute (which always has the instance/self namespace to keep it separate).

ops/pebble.py
```diff
@@ -1028,6 +1056,48 @@ class CheckStatus(enum.Enum):
...
+        self.location = dct.get('location', '')
+        services = dct.get('services')
+        if services:
+            services = services[:]
```
benhoyt, line 1070: Out of interest, why this vs the pattern we use elsewhere, like for `Service.after` for example:

```python
        self.after = list(dct.get('after', []))
```

I also tend to think `list(x)` is more explicit than `x[:]` -- don't have strong feelings, but we should probably be consistent in the same file.

test/test_pebble.py
```diff
@@ -479,6 +479,25 @@ def test_checks(self):
...
             plan.checks = {}  # type: ignore
 
+    def test_log_targets(self):
+        plan = pebble.Plan('')
+        self.assertEqual(plan.log_targets, {})
+
+        location = "https://example.com:3100/loki/api/v1/push"
+        plan = pebble.Plan(
+            'log-targets:\n baz:\n  override: replace\n  type: loki\n  '
+            f'location: {location}')
```
benhoyt, line 489: I realize you just copied my style here :-), but looking at this now I think it'd be significantly simpler to read the indentation on this if it were a multi-line string:

```python
    plan = pebble.Plan(f"""
log-targets:
  baz:
    override: replace
    type: loki
    location: {location}
""")

I think we should change this one and the other calls to `pebble.Plan(s)` in this file to keep things consistent.

**Review by benhoyt — CHANGES_REQUESTED** [review]
Nice, thanks! A few minor comments.

Could you also please do one quick test against a real Pebble instance with a log-targets and make sure it's coming through okay? I'd be okay with just a manual test hacking `test/pebble_cli.py` or something, but an automated test in `test_real_pebble.py` would be fine too if it's straight-forward.

**Comment by tonyandrewmeyer** [comment]
> Could you also please do one quick test against a real Pebble instance with a log-targets and make sure it's coming through okay? I'd be okay with just a manual test hacking `test/pebble_cli.py` or something, but an automated test in `test_real_pebble.py` would be fine too if it's straight-forward.

Sure. I had wondered about a test_real_pebble addition, but when looking through what that tests, it seemed to be more the actual Pebble functionality (pushing and pulling, running health checks, exec'ing) rather than just config. It seems complicated to do that with log forwarding, since I assume a Loki server would be required. What would you like a real_pebble test to do? Just check that particular config is appropriately created?

Also for this: the CI real pebble test gets the latest version of (main) pebble. Is it ok to assume that anyone running `tox -e pebble` also has the latest version, or should the tests check the version and skip appropriately?

File: ops/pebble.py (hunk 3 above)
Comment (tonyandrewmeyer) [reply ops/pebble.py:1070]: > Out of interest, why this vs the pattern we use elsewhere, like for `Service.after` for example:
> 
> ```python
>         self.after = list(dct.get('after', []))
> ```
> 
> I also tend to think `list(x)` is more explicit than `x[:]` -- don't have strong feelings, but we should probably be consistent in the same file.

Hah, in this case it's because I was not copying what was done elsewhere and was just writing code :joy:. I agree `list(x)` is more explicit than `x[:]` - I'm not sure why I default to using `[:]` - maybe because it used to be quite a bit faster? Maybe it was just a habit I picked up somewhere. I would say that `x.copy()` is probably the *most* explicit (and it also seems to be the fastest now), but I agree that consistency in the file trumps everything else, so will adjust to that, thanks!

**Comment by benhoyt** [comment]
> What would you like a real_pebble test to do? Just check that particular config is appropriately created?

Yeah, I wasn't thinking it'd test the Loki part -- just let the log forwarding itself fail. Just test that adding a layer with a log target can be queried and comes through with appropriate attributes in get_plan or something.

> Also for this: the CI real pebble test gets the latest version of (main) pebble. Is it ok to assume that anyone running tox -e pebble also has the latest version, or should the tests check the version and skip appropriately?

Yeah, I think it's fine to assume the dev is on Pebble latest, no need for version checks.

**Comment by tonyandrewmeyer** [comment]
@benhoyt the tests fail because the workflow uses pebble@latest. Is it ok to change that to pebble@master? (They pass locally for me in that case). I'm not sure whether we want to verify that we're compatible with the latest released pebble or the next version of pebble.

If we don't change it, then I can disable the labels until the next pebble release.

File: test/test_real_pebble.py [hunk 5]
```diff
@@ -259,6 +259,33 @@ def stdin_thread():
...
+                "pretend-loki": {
+                    "type": "loki",
+                    "override": "replace",
+                    "location": "https://example.com",
+                    "services": ["all"],
+                    "labels": {"foo": "bar"},
+                },
+            },
+        }, combine=True)
+        self.client.replan_services()
```
benhoyt, line 280: There's actually no need for the `replan_services` in this test, as we're just testing the plan itself, not that any services were (re)started.

benhoyt: Thanks for adding the test!

benhoyt:
@tonyandrewmeyer Yep, I think it's fine to have the real Pebble tests work off `@master`. We own the project and the master branch is (should be!) stable.

HACKING.md
```diff
@@ -79,6 +79,10 @@ source .tox/unit/bin/activate
 pytest -v test/test_real_pebble.py
 ```
 
+To test Pebble functionality that has been merged but not yet released, you can run
+`go install github.com/canonical/pebble/cmd/pebble@master`. This otherwise works in
```
benhoyt, line 83: Now that CI is actually using `@master`, shall we just remove this paragraph and update the paragraph further above to say to use `@master`?

benhoyt (approved):
Approved. Just one minor comment about the HACKING.md change now that we're using `@master`.

PR #1076, opened 2023-11-23
fix: non-leaders can only view app secrets

### LINKED ISSUE #1071: [testing] Non-leader unit can set 'app' secret
```
      secret = harness.charm.app.add_secret({"rumour": "movie_in_pre_prod"})
      harness.set_leader(False)
      secret.set_content({"teaser": "future_movie_2"})
```

'Simulated' unit (i.e. non-leader) shouldn't be able to set `app` level secret.
Yet the code above executes green.

tonyandrewmeyer:
A charm that owns a unit secret always has manage permissions, regardless of which unit is executing the charm. However, an app secret has manage permissions if the charm is executing on the leader, and view permissions otherwise ([the matrix in this WIP documentation](https://discourse.charmhub.io/t/secret-access-permissions/12627) has been verified by @wallyworld).

This PR adjusts the Harness permission check (used by get-info, set, grant, revoke, remove, etc) so that if the secret is an app secret, then the leader flag must be set for access to be allowed.

Fixes #1071

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -2482,7 +2482,12 @@ class _TestingModelBackend:
         return None
 
     def _ensure_secret_owner(self, secret: _Secret):
-        if secret.owner_name not in [self.app_name, self.unit_name]:
+        # For unit secrets, the owner unit has manage permissions. For app
+        # secrets, the leader has manage permissions and other units only have
+        # view permissions.
+        # https://discourse.charmhub.io/t/secret-access-permissions/12627
+        if (secret.owner_name != self.unit_name
+                and (secret.owner_name != self.app_name or not self.is_leader())):
             raise model.SecretNotFoundError(
                 f'You must own secret {secret.id!r} to perform this operation')
 
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -2486,10 +2486,13 @@ class _TestingModelBackend:
         # secrets, the leader has manage permissions and other units only have
         # view permissions.
         # https://discourse.charmhub.io/t/secret-access-permissions/12627
-        if (secret.owner_name != self.unit_name
-                and (secret.owner_name != self.app_name or not self.is_leader())):
-            raise model.SecretNotFoundError(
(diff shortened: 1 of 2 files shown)


test/test_testing.py
```diff
@@ -4914,6 +4915,39 @@ def test_trigger_secret_expiration(self):
         with self.assertRaises(RuntimeError):
             harness.trigger_secret_removal('nosecret', 1)
 
+    def test_secret_permissions(self):
```
PietroPasotti, line 4918: this test makes my eyes bleed. Can we split it, parametrize it, or at least add a few blank newlines? :smiling_face_with_tear:

ops/testing.py
```diff
@@ -2482,7 +2482,12 @@ def _relation_id_to(self, remote_app: str) -> Optional[int]:
         return None
 
     def _ensure_secret_owner(self, secret: _Secret):
-        if secret.owner_name not in [self.app_name, self.unit_name]:
+        # For unit secrets, the owner unit has manage permissions. For app
+        # secrets, the leader has manage permissions and other units only have
+        # view permissions.
+        # https://discourse.charmhub.io/t/secret-access-permissions/12627
+        if (secret.owner_name != self.unit_name
+                and (secret.owner_name != self.app_name or not self.is_leader())):
```
PietroPasotti, line 2490: [nit] can we un-De Morgan this conditional? it's not easy to read even for a logician :) 

```python
unit_secret = secret.owner_name == self.unit_name
app_secret = secret.owner_name == self.app_name

if unit_secret or (app_secret and self.is_leader()):
    pass  # all good
else: 
    raise ...
```

PietroPasotti (approved):
couple of readability nits but nothing blocking from my side.
GJ!

test/test_testing.py
tonyandrewmeyer: :). I've split it. parameterisation would be nicest, but I find that a pain to do with unittest, so...

benhoyt (approved):
Great, thanks for this fix.

PR #1080, opened 2023-11-23
fix(charm): make collect-status a LifeCycleEvent to avoid logging it

### LINKED ISSUE #1079: collect_*_status and lifecycle events
logically, collect_status is a lifecycle event, aka framework event, since it's not 'custom' in the way user-defined events are, and it is emitted by the framework on itself whether the user wants it or not, quite like `commit/pre-commit`.

At the moment we log their emission as `custom event` which is somewhat misleading and complicates jhack tail's life when attempting to distinguish them from 'true' custom events.

I propose we make `CollectStatusEvent` a `LifecycleEvent` subclass, and we uniformly start logging all `LifecycleEvents` so jhack tail can pick them up if the user wants to see them (but they will be hidden by default as they are generally speaking not interesting and would clutter the output).

PietroPasotti:
Make CollectStatusEvent inherit from LifecycleEvent to avoid logging it.

Fixes #1079


ops/log.py
```diff
@@ -56,6 +56,7 @@ def setup_root_logging(model_backend: "_ModelBackend", debug: bool = False):
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
+    logger.handlers
```
benhoyt, line 59: Left-over debugging?

test/charms/test_main/src/charm.py
```diff
@@ -92,6 +92,7 @@ def __init__(self, *args):
         self.framework.observe(self.on.mon_relation_departed, self._on_mon_relation_departed)
         self.framework.observe(self.on.ha_relation_broken, self._on_ha_relation_broken)
         self.framework.observe(self.on.test_pebble_ready, self._on_test_pebble_ready)
+        self.on
```
benhoyt, line 95: Left-over debugging?

ops/framework.py
```diff
@@ -906,11 +906,11 @@ def _reemit(self, single_event_path: Optional[str] = None):
                 if single_event_path is None:
                     logger.debug("Re-emitting deferred event %s.", event)
                 elif isinstance(event, LifecycleEvent):
-                    # Ignore Lifecycle events: they are "private" and not interesting.
-                    pass
+                    logger.debug("Emitting lifecycle event %s.", event.handle.kind)
```
benhoyt, line 909: I think the change of parent class to `LifecycleEvent` is reasonable, however I agree with your hesitation -- I think this is going to be too much logging.

I don't really think it needs to be logged at all, even for collect-status events, and the log message doesn't really contain much useful information that you can't see by looking at the charm statically anyway. So I'd say just leave this as ignore/pass.

benhoyt (changes requested):
Thanks for this. Couple of minor comments / requested changes.

ops/framework.py
PietroPasotti: My reason for wanting to log collect-status is that now the collect-status event handler has (potential for) holding charm logic as it is user-facing. And the fact that leader/followers will get different events would be nice to show in jhack tail.

I haven't used the event extensively (yet) in practice to know whether it'd be useful, it's just a hunch.

Would be nice to gather feedback from other charmers who are trying it out?

benhoyt: So is the concern primarily not the debug log itself, but that this will mean it will show up in jhack tail?

As far as the log itself goes, for charms that use collect-status it'll mean we get two (or three) logs for every Juju event, which I don't think is good. I think even when debugging with jhack, you'll know the charm uses collect-status, so you can just assume that `collect_unit_status` is occurring when you see any other event. And the log doesn't add any other useful collect_status-specific information.

In any case, I think we should start by not logging it for now (that is, reinstate the "Ignore Lifecycle events" comment), and if we *collect* more information about why it'd be useful, we can reconsider.

tonyandrewmeyer (approved):
I agree with the comments from Ben (and Pietro's original concerns) about the volume of low-value records being sent to Juju. Otherwise seems good to me!

benhoyt (approved):

PR #1082, opened 2023-11-28
fix: handle users/groups with no names in push_path

tonyandrewmeyer:
[Ned Batchelder noticed that the unit tests failed on his machine](https://hachyderm.io/@coveragepy/111477098431251147), which was because the temporary folder returned by `TemporaryDirectory()` belonged to a group with no name.

In theory, this could happen outside of the tests as well - someone could do a `push_path()` on a path that has files that have owners/groups that do not have names (e.g. because they have been removed). This is presumably quite uncommon (which makes sense: it seems unlikely that the charm container would often have a user or group removed) but it's also simple to fix, so it seems like we might as well do that.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -24,7 +24,7 @@ import unittest
 from collections import OrderedDict
 from test.test_helpers import fake_script, fake_script_calls
 from textwrap import dedent
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 
@@ -968,6 +968,28 @@ class TestModel(unittest.TestCase):
             _ = model.unit.status.message
         self.assertEqual(str(cm.exception), 'ERROR cannot get status\n')
         self.assertEqual(cm.exception.args[0], 'ERROR cannot get status\n')
+
+    @patch("pwd.getpwuid")
+    @patch("grp.getgrgid")
+    def test_push_path_unnamed(self, getpwuid: MagicMock, getgrgid: MagicMock):
+        getpwuid.side_effect = KeyError
+        getgrgid.side_effect = KeyError
+        harness = ops.testing.Harness(ops.CharmBase, meta='''
+            name: test-app
+            containers:
+              foo:
+                resource: foo-image
+            ''')
+        harness.begin()
+        harness.set_can_connect('foo', True)
+        c = harness.model.unit.containers['foo']
+
+        with tempfile.TemporaryDirectory() as push_src:
+            push_path = pathlib.Path(push_src) / 'src.txt'
+            with push_path.open('w') as f:
+                f.write('hello')
+            c.push_path(push_path, "/")
+        assert c.exists("/src.txt"), 'push_path failed: file "src.txt" missing at destination'
 
 
 class PushPullCase:
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -988,7 +988,7 @@ class TestModel(unittest.TestCase):
             push_path = pathlib.Path(push_src) / 'src.txt'
             push_path.write_text('hello')
             container.push_path(push_path, "/")
-        assert container.exists("/src.txt"), 'push_path failed: file "src.txt" missing at destination'
+        assert container.exists("/src.txt"), 'push_path failed: file "src.txt" missing'
 
 
 class PushPullCase:
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -982,14 +982,13 @@ class TestModel(unittest.TestCase):
             ''')
         harness.begin()
         harness.set_can_connect('foo', True)
-        c = harness.model.unit.containers['foo']
+        container = harness.model.unit.containers['foo']
 
         with tempfile.TemporaryDirectory() as push_src:
             push_path = pathlib.Path(push_src) / 'src.txt'
-            with push_path.open('w') as f:
-                f.write('hello')
-            c.push_path(push_path, "/")
-        assert c.exists("/src.txt"), 'push_path failed: file "src.txt" missing at destination'
+            push_path.write_text('hello')
+            container.push_path(push_path, "/")
+        assert container.exists("/src.txt"), 'push_path failed: file "src.txt" missing at destination'
 
 
 class PushPullCase:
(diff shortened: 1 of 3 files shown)


test/test_model.py
```diff
@@ -969,6 +969,28 @@ def test_run_error(self):
         self.assertEqual(str(cm.exception), 'ERROR cannot get status\n')
         self.assertEqual(cm.exception.args[0], 'ERROR cannot get status\n')
 
+    @patch("pwd.getpwuid")
+    @patch("grp.getgrgid")
```
benhoyt, line 973: It doesn't matter here as we're patching both with the same side effect, but to avoid confusion, these should be  in reverse order to the parameters:
```suggestion
    @patch("grp.getgrgid")
    @patch("pwd.getpwuid")
```

test/test_model.py
```diff
@@ -969,6 +969,28 @@ def test_run_error(self):
...
+        getgrgid.side_effect = KeyError
+        harness = ops.testing.Harness(ops.CharmBase, meta='''
+            name: test-app
+            containers:
+              foo:
+                resource: foo-image
+            ''')
+        harness.begin()
+        harness.set_can_connect('foo', True)
+        c = harness.model.unit.containers['foo']
```
benhoyt, line 985: I don't think there's any reason for a one-character name here:
```suggestion
        container = harness.model.unit.containers['foo']
```

test/test_model.py
```diff
@@ -969,6 +969,28 @@ def test_run_error(self):
...
+
+        with tempfile.TemporaryDirectory() as push_src:
+            push_path = pathlib.Path(push_src) / 'src.txt'
+            with push_path.open('w') as f:
+                f.write('hello')
```
benhoyt, line 990: Could use `Path.write_text` here?
```suggestion
            push_path.write_text('hello')
```

test/test_model.py
```diff
@@ -969,6 +969,28 @@ def test_run_error(self):
...
+            c.push_path(push_path, "/")
+        assert c.exists("/src.txt"), 'push_path failed: file "src.txt" missing at destination'
```
benhoyt, line 992: It's a bit weird that we're not actually testing that the user/group attributes are None. In fact, this test succeeds even if you comment out the patching. I guess there's not an easy way to test this, because in this code path we're not actually using the user/group attributes. We *could* test the logs generated, but probably not worth it.

It does seem strange that we're calling these functions on every file and then throwing away the data (`_list_recursive` only uses the `type` attributes). But I'm happy to keep it as is for now to avoid code churn.

Thoughts?

benhoyt (changes requested):
A couple of nit suggestions and one question about the testing.

test/test_model.py
tonyandrewmeyer: Ah, nice catch, thanks!

test/test_model.py
tonyandrewmeyer: Fair enough. I actually copied this from `test_recursive_push_and_pull`, but I agree the full name is better.

test/test_model.py
tonyandrewmeyer: TIL :)

test/test_model.py
tonyandrewmeyer: > It's a bit weird that we're not actually testing that the user/group attributes are None. In fact, this test succeeds even if you comment out the patching. I guess there's not an easy way to test this, because in this code path we're not actually using the user/group attributes. 

Yes, that was the issue I had :disappointed:. The test does fail (error) without the `model.py` change (as long as the patching is there), because a KeyError will be raised, but it indeed doesn't validate what the alternate value is. If you comment out the patching, then it's just testing that the regular case (whatever user/group you're defaulting to) works.

> We _could_ test the logs generated, but probably not worth it.

Agreed.

> It does seem strange that we're calling these functions on every file and then throwing away the data (`_list_recursive` only uses the `type` attributes). But I'm happy to keep it as is for now to avoid code churn.
> 
> Thoughts?

I found it odd too. At first I thought it was intending to set the user/group on the workload container (which seems dubious using the ID), but noticed that indeed the majority of the data is thrown away.

However, there's bleeding over into harness, because the test pebble's `list_files` uses the same method so in tests people get back all the info. (I could actually put it in a test there that checks that the names are None).

It does seem like it would be cleaner if `_TestingPebbleClient` had the `_build_fileinfo` that `Container` has in the current PR (or one that started with `Container`'s and built on it like the current PR), and `Container` had something like:

```
    @staticmethod
    def _build_fileinfo(path: Union[str, Path]) -> pebble.FileInfo:
        """Constructs a FileInfo object by stat'ing a local path."""
        path = Path(path)
        if path.is_symlink():
            ftype = pebble.FileType.SYMLINK
        elif path.is_dir():
            ftype = pebble.FileType.DIRECTORY
        elif path.is_file():
            ftype = pebble.FileType.FILE
        else:
            ftype = pebble.FileType.UNKNOWN

        return pebble.FileInfo(
            path=str(path),
            name=path.name,
            type=ftype)
```

... except that `FileInfo` has `last_modified` and `permissions` as required fields. A `stat` is pretty cheap, but it's also being completely wasted here.

`Container`'s `_build_fileinfo` doesn't have to return a `FileInfo` (although the name would need to change if it didn't). I guess it could return a `TypedDict` private to `model.py` that only had the path and type.

This would extend the value of this PR to actually making a (presumably imperceptible) performance improvement. But it also significantly increases the number of lines touched for what was originally a (I assume) fairly obscure bug.

benhoyt: Thanks for the write-up. Okay, I think let's leave this PR as just the bug-fix: it's definitely an improvement over what we have now. We can always consider the simplification/refactor later separately for performance reasons.

benhoyt (approved):
Looks good to me now, though linting is failing.

tonyandrewmeyer:
> Looks good to me now, though linting is failing.

Fixed now. Do you have a suggestion for a second reviewer?

benhoyt:
I'm fine without a second reviewer on this small bug fix. I'll merge as soon as the tests are passing.

PR #1084, opened 2023-11-28
fix: handle parsing ISO datetimes where the microseconds round up to more than 1s

tonyandrewmeyer:
We handle microseconds with more than 6 digits if the number of microseconds can be rounded to a number < 1M, for example:

```python
>>> timeconv.parse_rfc3339("2006-08-28T13:20:00.9999991+12:00")
datetime.datetime(2006, 8, 28, 13, 20, 0, 999999, tzinfo=datetime.timezone(datetime.timedelta(seconds=43200)))
```

However, if rounding takes us to the next second, we fail, for example:

```python
>>> timeconv.parse_rfc3339("2006-08-28T13:20:00.9999999+12:00")
  File "/home/tameyer/code/operator/ops/_private/timeconv.py", line 54, in parse_rfc3339
    return datetime.datetime(int(y), int(m), int(d), int(hh), int(mm), int(ss),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: microsecond must be in 0..999999
```

Handle this case by carrying the result over to the second field.

Note that it looks like the Python 3.11+ standard library (which includes this functionality) [https://github.com/python/cpython/blob/48dfd74a9db9d4aa9c6f23b4a67b461e5d977173/Lib/_pydatetime.py#L413 truncates] instead of rounds (at least in the Python `_pydatetime` - I haven't looked at the C implementation). Rounding seems more correct to me, but I can understand an argument that we should duplicate the stdlib behaviour since we'll presumably move to using that once we have Python 3.11 as a minimum Python version.

### DIFF
--- a/ops/_private/timeconv.py
+++ b/ops/_private/timeconv.py
@@ -32,7 +32,8 @@ def parse_rfc3339(s: str) -> datetime.datetime:
     that Go's encoding/json package produces for time.Time values.
 
     Unfortunately we can't use datetime.fromisoformat(), as that does not
-    support more than 6 digits for the fractional second, nor the 'Z' for UTC.
+    support more than 6 digits for the fractional second, nor the 'Z' for UTC,
+    in Python 3.8 (Python 3.11+ has the required functionality).
     """
     match = _TIMESTAMP_RE.match(s)
     if not match:
@@ -50,6 +51,8 @@ def parse_rfc3339(s: str) -> datetime.datetime:
         tz = datetime.timezone(tz_delta if sign == '+' else -tz_delta)
 
     microsecond = round(float(sfrac or '0') * 1000000)
+    carry, microsecond = divmod(microsecond, 1000000)
+    ss = int(ss) + carry
 
-    return datetime.datetime(int(y), int(m), int(d), int(hh), int(mm), int(ss),
+    return datetime.datetime(int(y), int(m), int(d), int(hh), int(mm), ss,
                              microsecond=microsecond, tzinfo=tz)
--- a/test/test_private.py
+++ b/test/test_private.py
@@ -72,6 +72,9 @@ class TestStrconv(unittest.TestCase):
         self.assertEqual(timeconv.parse_rfc3339('2020-12-25T13:45:50.123456789+00:00'),
                          datetime.datetime(2020, 12, 25, 13, 45, 50, 123457, tzinfo=utc))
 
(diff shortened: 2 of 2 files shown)


tonyandrewmeyer:
Hmm, actually, this could overflow into minutes, then into hours, etc. Will fix.

ops/_private/timeconv.py
```diff
@@ -50,6 +51,10 @@ def parse_rfc3339(s: str) -> datetime.datetime:
         tz = datetime.timezone(tz_delta if sign == '+' else -tz_delta)
 
     microsecond = round(float(sfrac or '0') * 1000000)
+    carry, microsecond = divmod(microsecond, 1000000)
```
benhoyt, line 54: This is nice. However, given that Python 3.11+ truncates, and we'll want to start using that (and this whole function will go away then), I think we should do what it does. It'll also be harder to get wrong with just `int()`. Will this do it?

```python
microsecond = int(float(sfrac or '0') * 1000000)
```

benhoyt (changes requested):
Huh, nice find! This bug has been lurking here since I wrote it. It will happen 500/1,000,000,000 times, or 1 in 2 million, so it's definitely not crazy that it happened.

Yeah, I do think we should truncate to converge on the new stdlib's behaviour. Does it truncate only when the nanoseconds are >= 999999500, or for all values?

tonyandrewmeyer:
> Huh, nice find! This bug has been lurking here since I wrote it. It will happen 500/1,000,000,000 times, or 1 in 2 million, so it's definitely not crazy that it happened.

Although with those odds I think you can legitimately feel unlucky if it happens to you :joy:

> Yeah, I do think we should truncate to converge on the new stdlib's behaviour. Does it truncate only when the nanoseconds are >= 999999500, or for all values?

Only then, basically. It's just slicing the original string and ignoring anything after the 6th digit. I checked the [CPython code for `_datetime`](https://github.com/python/cpython/blob/48dfd74a9db9d4aa9c6f23b4a67b461e5d977173/Modules/_datetimemodule.c#L865) and it's basically doing the same as the Python code.

benhoyt:
Okay, in that case let's keep using `round()` but add a `min(microsecond, 999999)` after.

benhoyt (approved):

benhoyt:
Thanks for the report and the fix!

PR #1085, opened 2023-12-01
chore: Refactor main.py to add a new _Manager class

PietroPasotti:
With this I hope to contribute some of the abstractions I developed for Scenario's `ops.main` mock back to the framework.

It's **not a functional change**, that is, ops should be functioning in exactly the same way.

It is however IMHO a positive change for the following reasons:
- tidier, more object-oriented. `main.py` was hard to understand and extend
- (therefore) easier to play with in e.g. testing frameworks or experimental modules wishing to modify or extend its functionality
- easier to maintain

The core contributions are:
- a `_Manager` class encapsulating the management of the ops lifecycle, from instantiation of Framework and charm instances to storage creation, event emission, and storage commit.
- adding the `CharmMeta.from_charm_root` method to load a `CharmMeta` instance from the relevant `*.yaml` files


PietroPasotti:
(test failures seem to be related to the breaking changes recently introduced in harness' secret model

PietroPasotti:
also charmcraft pack has an unrelated bug https://matrix.to/#/!LhFxJIPEcCacgdMghH:ubuntu.com/$ah10i3Ckd9YsahSr9xROqQNOTfVK_VrTYDK3oszGQGE?via=ubuntu.com&via=matrix.org hence the integration failures

tonyandrewmeyer:
> (TBD if we can use dataclasses or we need to support older python versions, if so we could fall back to a namedtuple) that encapsulates the 'charm module' including its metadata yamls and repo root.

For ops itself, we explicitly support 3.8+, so dataclasses are fine (and there are already a couple of uses in ops). If the concern is about experiments (etc) using this class then I would say we do not want to support anything older than ops itself supports.

benhoyt:
Thanks @PietroPasotti. Tony and I will take a look over this and discuss tomorrow and then get back to you with some thoughts or next steps.

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
         return False
 
 
+@dataclasses.dataclass(frozen=True)
+class _CharmSpec:
```
tonyandrewmeyer, line 329: I don't really understand the utility of the `_CharmSpec` class. The `autoload` method overlaps a lot with `CharmMeta.from_yaml` (with the addition of loading `config.yaml`, but that could easily be an addition to `CharmMeta`).

It seems like the PR would be simpler if `Ops` took a charm type as an `__init__` argument, and the `_get_charm_dir` method could be in there, along with one to load the metadata (some of the minor changes like checking if the metadata exists are nice and could stay).

Am I maybe missing use cases for the _CharmSpec class? If so, could you outline/link to those? Maybe in the PR description rather than specifically in this thread for better visibility?

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        return _CharmSpec(
+            charm_type=charm_type,
+            charm_root=charm_root,
+            meta=meta,
+            actions=actions,
+            config=config
+        )
+
+
+class Ops:
```
tonyandrewmeyer, line 379: @benhoyt and I feel that it would be best to start this out as `_Ops`, rather than being a public interface immediately. It wouldn't be hard to move from `_Ops` to `Ops` in the future, whereas once it's public it's very difficult to go back. It also allows a bit more freedom adjusting in the public members of the class at first, while still providing the nicer abstraction and easier overriding that refactoring into the class provides.

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        self.charm: Optional["ops.charm.CharmBase"] = None
+
+        self._has_setup = False
+        self._has_emitted = False
+        self._has_committed = False
+
+    def _setup_charm(self, framework: "ops.framework.Framework", dispatcher: _Dispatcher):
+        charm_type = self.charm_spec.charm_type
+        sig = inspect.signature(charm_type)
+        sig.bind(framework)  # signature check
```
tonyandrewmeyer, line 420: What's the advantage of doing this here? If the signature is wrong, doesn't it just move the error to this line with `TypeError: too many positional arguments` (or whatever the bad sig is), rather than a couple of lines down with `TypeError: MyCharm.__init__() takes 1 positional argument but 2 were given` (for example).

It seems like letting it fail a couple of lines later with a clearer error that it's the `__init__` signature that's at fault is better. I assume I'm missing a benefit :)

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+    @staticmethod
+    def _setup_root_logging(backend: ops.model._ModelBackend):
+        debug = "JUJU_DEBUG" in os.environ
+        setup_root_logging(backend, debug=debug)
+
+        # our hello world
+        logger.debug(
+            "ops %s up and running.",
+            ops.__version__,
+        )  # type:ignore
```
tonyandrewmeyer, line 440: This easily fits on one line in the existing code, and one additional 4-space indent shouldn't change that. Did you wrap this, or was it a tool?

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        meta = self._charm_meta
+        model = ops.model.Model(meta, model_backend)
+        store = self._setup_storage(dispatcher)
+        framework = ops.framework.Framework(store, self.charm_spec.charm_root, meta, model)
+        framework.set_breakpointhook()
+        return framework
+
+    def setup(self):
+        """Setup framework, charm and dispatcher."""
+        """Setup dispatcher, framework and charm objects."""
```
tonyandrewmeyer, line 493: Double docstring :)

ops/main.py
```diff
@@ -369,77 +584,15 @@ def main(charm_class: Type[ops.charm.CharmBase],
...
-    else:
-        store = ops.storage.SQLiteStorage(charm_state_path)
-    framework = ops.framework.Framework(store, charm_dir, meta, model,
-                                        event_name=dispatcher.event_name)
-    framework.set_breakpointhook()
-    try:
-        charm = charm_class(framework)
-        dispatcher.ensure_event_links(charm)
-
-        # TODO: Remove the collect_metrics check below as soon as the relevant
```
tonyandrewmeyer, line 427: This comment isn't in the refactored code. @benhoyt is it an orphaned comment? Or is it attached to the `if not dispatcher.is_restricted_context()` check that is immediately below (which is still in the refactored version)?

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        if not self._has_emitted:
+            self.emit()
+        if not self._has_committed:
+            self.commit()
+
+
 def main(charm_class: Type[ops.charm.CharmBase],
          use_juju_for_storage: Optional[bool] = None):
-    """Setup the charm and dispatch the observed event.
+    """Set up the charm and dispatch the observed event.
```
tonyandrewmeyer, line 576: Nice :tada:

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+
+    def _setup_framework(
+            self,
+            dispatcher: _Dispatcher,
+            model_backend: "ops.model._ModelBackend"
+    ):
+        meta = self._charm_meta
+        model = ops.model.Model(meta, model_backend)
+        store = self._setup_storage(dispatcher)
+        framework = ops.framework.Framework(store, self.charm_spec.charm_root, meta, model)
```
tonyandrewmeyer, line 487: I'm not familiar enough with the main/framework code to know this and haven't taken the time to walk through things to figure it out: does not passing in the event name change anything here? It seems like maybe it would impact detecting custom events? We have the dispatcher in this method, so it seems like it would be simple to pass it like the existing code does, unless there's a specific reason to not do that.

tonyandrewmeyer (commented):
@benhoyt and I agree that in general this is a good change. We like splitting up `main.main` (easier to override specific parts and also easier to understand/read), that it doesn't introduce behaviour changes but allows easier experimentation and other building on ops, and the general approach.

We're not clear on the benefit of `_CharmSpec` (noted in a review comment also) so would appreciate some more context there, and feel that this would be better to start as a private `_Ops` class than `Ops` (which also avoids the issue that `Ops` takes `_CharmSpec` as an argument, so `_CharmSpec` can't really be private if `Ops` is public).

I've also added a few comments that are just from me, not from the two of us :smile:. I haven't tried building, deploying, and using a charm with this branch yet but would do that as part of a final review. It is nice that the tests pass with so few changes.

PietroPasotti:
will get back to this next week when I have more time. Thanks for the feedback.

ops/main.py
```diff
@@ -356,9 +325,255 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+            event_name: A Juju event name to emit on a charm.
+        """
+        charm = cast("ops.charm.CharmBase", self.charm)  # by now we have initialized.
+        owner = charm.on
+
+        try:
+            event_to_emit = getattr(owner, event_name)
+        except AttributeError:
+            logger.debug("Event %s not defined for %s.", event_name, charm)
+            exit(0)
```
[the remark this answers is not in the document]
PietroPasotti: technically, I think `exit(0)` will still trigger the `except Exception` clause in `emit()`, which in turn calls `framework.close()`. But I guess returning might be more elegant as it allows run() to progress naturally to commit and close

ops/main.py
PietroPasotti: Added some more thoughts on the PR description, but TLDR: encapsulation and ease of mocking. If you look at how I'm using it in scenario, it should become more clear what the intended use case is.

ops/main.py
PietroPasotti: so to answer your question, yes, it refers to the restricted context check below

ops/main.py
PietroPasotti: oh, I missed that. Good catch.

ops/main.py
PietroPasotti: for now I'll add it back to keep the PR contained

PietroPasotti:
@tonyandrewmeyer @benhoyt sorry for taking so long to get back on this, I was on holiday in december and forgot it existed. Addressed all comments and rebased.

ops/main.py
tonyandrewmeyer: Juju 3.4 still seems to have collect-metrics so it looks like it'll have to be supported in ops for some time to come, unfortunately.

benhoyt:
@PietroPasotti Looks like `test_crash_action` is failing:

```
FAILED test/test_main.py::TestMainWithDispatch::test_crash_action - AssertionError: 'KeyError' not found in ''
```

tonyandrewmeyer (commented):
I'm definitely in favour of the `_Ops` class and the way that it organises the `main` code. I have a few small nits (some are maybe even copied from the old code, but let's tidy them up while we're here) and US English-isms, plus there's one recent change that got missed during merging (I assume). It's odd that the tests don't fail as a result - I'll have to see whether I missed adding one for that case.

I'm less convinced that we need `_CharmSpec` (as per comments).

ops/main.py
```diff
@@ -369,97 +337,282 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+
+    def _setup_charm(self, framework: "ops.framework.Framework", dispatcher: _Dispatcher):
+        charm_type = self.charm_spec.charm_type
+
+        charm = charm_type(framework)
+        dispatcher.ensure_event_links(charm)
+        return charm
+
+    @staticmethod
+    def _setup_model_backend():
```
benhoyt, line 437: Would it be simpler just to have a `backend_class=ops.model._ModelBackend` parameter to `_Ops.__init__`?

ops/main.py
```diff
@@ -369,97 +337,282 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
-        store = ops.storage.JujuStorage()
-    else:
-        store = ops.storage.SQLiteStorage(charm_state_path)
-    framework = ops.framework.Framework(store, charm_dir, meta, model,
-                                        event_name=dispatcher.event_name)
-    framework.set_breakpointhook()
-    try:
-        charm = charm_class(framework)
-        dispatcher.ensure_event_links(charm)
+            exit(0)
```
benhoyt, line 477: Seems very strange to call `exit()` in a method. Is there a better way to do this? Return a value, or raise an exception?

ops/main.py
```diff
@@ -369,97 +337,282 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+            broken_relation_id = None
+
+        model = ops.model.Model(meta, model_backend, broken_relation_id=broken_relation_id)
+        store = self._setup_storage(dispatcher)
+        framework = ops.framework.Framework(store, self.charm_spec.charm_root, meta, model,
+                                            event_name=dispatcher.event_name)
+        framework.set_breakpointhook()
+        return framework
+
+    def setup(self):
```
benhoyt, line 506: Instead of `setup`, should we just do the setup in `__init__`?

benhoyt (commented):
A few questions for discussion.

ops/main.py
tonyandrewmeyer: Note that if we do, `CHARM_STATE_FILE` should probably be an arg to `__init__()` since otherwise the only way to cleanly change it would be to subclass. I did wonder if maybe it should just be an arg anyway.

PietroPasotti:
Ok I'm starting to see the light on _CharmSpec kind of being the same thing as CharmMeta. I somewhat thought charmmeta was supposed to wrap only the metadata (as in what's in `metadata.yaml`, but I guess that's not the best way to look at it.

tonyandrewmeyer (approved):
Thanks for all the work on this! I think this is makes main a lot nicer, both to extend and to understand.

benhoyt (changes requested):
Hi @PietroPasotti Sorry for the continued back and forth, but I think there are several things that need a bit of polish here, with particular thought to the exception handling point. I've also submitted a bunch of minor comments about comments and naming.

The "should we use a context manager" thing, and a couple of other items, might require a bit of discussion to save additional code review merry-go-round cycles.

ops/main.py
```diff
@@ -369,97 +335,225 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
 
+        # Emit collect-status events.
         ops.charm._evaluate_status(charm)
+        try:
+            framework.commit()
+        finally:
+            framework.close()
+        self._has_committed = True
+
+    def run(self):
```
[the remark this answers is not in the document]
PietroPasotti: that actually made me realize, why do we need to expose all of `run`, `emit` and `commit`, if there really is only one happy path here: first call emit, then commit.
there is no user-facing use case for emit, do something, then commit. 

solution:
make `_emit/_commit` private, only public method is `run` which calls both in the right order.

Scenario will have some use for that, but scenario can use the private `_emit/_commit` API to do its hackery.

ops/main.py
```diff
@@ -369,97 +335,225 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
-        charm_class: the charm class to instantiate and receive the event.
-        use_juju_for_storage: whether to use controller-side storage. If not specified
-            then kubernetes charms that haven't previously used local storage and that
-            are running on a new enough Juju default to controller-side storage,
-            otherwise local storage is used.
+class StorageSetupError(Exception):
+    """Raised when setting up the storage fails."""
+
+
+class _Ops:
```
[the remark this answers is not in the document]
PietroPasotti: and if we really do make _ops a context manager, that might be another nice name

ops/main.py
```diff
@@ -369,97 +335,225 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+
+    def commit(self):
+        """Commit the framework and gracefully teardown."""
+        if not self._has_emitted:
+            raise RuntimeError("should .emit() before you .commit()")
+
+        charm = self.charm
+        framework = self.framework
 
+        # Emit collect-status events.
```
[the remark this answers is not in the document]
PietroPasotti: uhm, that makes sense.

ops/main.py
```diff
@@ -369,97 +335,225 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+            #       EventBase.defer().
+            #
+            # Skip reemission of deferred events for collect-metrics events because
+            # they do not have the full access to all hook tools.
+            if not dispatcher.is_restricted_context():
+                framework.reemit()
+
+            self._emit_charm_event(dispatcher.event_name)
+
+        except Exception:
```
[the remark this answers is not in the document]
PietroPasotti: I think that can also be consolidated. Refactored so that `run()` is the only place from which framework.close() is called.

ops/main.py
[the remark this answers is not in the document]
PietroPasotti: I did rename it to _Manager, but I don't quite see the point of making it a context manager, as the only thing you now can do with it is `.run()` which will open and close all that needs opening and closing. 
No need to offer the possibility to suspend execution.

ops/main.py
```diff
@@ -369,97 +335,225 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
 
-    The event name is based on the way this executable was called (argv[0]).
 
-    Args:
-        charm_class: the charm class to instantiate and receive the event.
-        use_juju_for_storage: whether to use controller-side storage. If not specified
-            then kubernetes charms that haven't previously used local storage and that
-            are running on a new enough Juju default to controller-side storage,
-            otherwise local storage is used.
+class StorageSetupError(Exception):
```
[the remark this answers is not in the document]
PietroPasotti: refactored it to an `_Abort` exception. Replaced all various `sys.exit` calls in `main` to use it instead.

ops/main.py
```diff
@@ -369,97 +335,225 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
-        store = ops.storage.JujuStorage()
-    else:
-        store = ops.storage.SQLiteStorage(charm_state_path)
-    framework = ops.framework.Framework(store, charm_dir, meta, model,
-                                        event_name=dispatcher.event_name)
-    framework.set_breakpointhook()
-    try:
-        charm = charm_class(framework)
-        dispatcher.ensure_event_links(charm)
+            raise StorageSetupError()
```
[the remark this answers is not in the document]
PietroPasotti: current proposal is to add an _Abort exception caught from `ops.main.main` and converted there to a `sys.exit`. All other sys.exit calls have been removed and turned into a `raise _Abort(exit_code: int)`.

It looks much neater now. And we can also then in the future think about making `_Abort` public

ops/charm.py
```diff
@@ -1373,6 +1373,24 @@ def __init__(self, raw: Optional[Dict[str, Any]] = None,
         self.containers = {name: ContainerMeta(name, container)
                            for name, container in raw_.get('containers', {}).items()}
 
+    @staticmethod
+    def from_charm_root(charm_root: Union[pathlib.Path, str]):
+        """Initialize CharmMeta from the path to a charm repository root folder."""
```
tonyandrewmeyer, line 1378:
```suggestion
        """Initialise CharmMeta from the path to a charm repository root folder."""
```

test/test_charm.py
```diff
@@ -468,6 +471,46 @@ def _get_action_test_meta(cls):
...
+                yaml.safe_dump(
+                    {"name": "bob",
+                        "requires": {
+                            "foo":
+                                {"interface": "bar"}
+                        }}))
+
+            meta = ops.CharmMeta.from_charm_root(td)
+            assert meta.name == "bob"
+            assert meta.requires['foo'].interface_name == "bar"
```
tonyandrewmeyer, line 487: We're keen on generally moving to pytest, largely to get assertions in this style, but I think while we're still in unittest, we should probably stick with the unittest assertion methods for the most part.

```suggestion
            self.assertEqual(meta.name, "bob")
            self.assertEqual(meta.requires['foo'].interface_name, "bar")
```

test/test_charm.py
```diff
@@ -468,6 +471,46 @@ def _get_action_test_meta(cls):
...
+            assert meta.actions['foo'].additional_properties is False
+            assert meta.actions['foo'].description == "foos the bar"
```
tonyandrewmeyer, line 512: As above:

```suggestion
            self.assertEqual(meta.name, "bob")
            self.assertEqual(meta.requires['foo'].interface_name, "bar")
            self.assertFalse(meta.actions['foo'].additional_properties)
            self.assertEqual(meta.actions['foo'].description, "foos the bar")
```

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+def main(charm_class: Type[ops.charm.CharmBase],
+         use_juju_for_storage: Optional[bool] = None):
+    """Set up the charm and dispatch the observed event.
+
+    The event name is based on the way this executable was called (argv[0]).
+
+    Args:
+        charm_class: the charm class to instantiate and receive the event.
+        use_juju_for_storage: whether to use controller-side storage. If not specified
+            then kubernetes charms that haven't previously used local storage and that
```
tonyandrewmeyer, line 540:
```suggestion
            then Kubernetes charms that haven't previously used local storage and that
```

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
         # TODO: Remove the collect_metrics check below as soon as the relevant
         #       Juju changes are made. Also adjust the docstring on
         #       EventBase.defer().
         #
         # Skip reemission of deferred events for collect-metrics events because
         # they do not have the full access to all hook tools.
-        if not dispatcher.is_restricted_context():
-            framework.reemit()
+        if not self.dispatcher.is_restricted_context():
+            # re-emit any deferred events from the previous run
```
tonyandrewmeyer, line 510:
```suggestion
            # Re-emit any deferred events from the previous run.
```

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+            self.framework.reemit()
+
+        # emit the juju event
```
tonyandrewmeyer, line 513:
```suggestion
        # Emit the Juju event.
```

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        self._emit_charm_event(self.dispatcher.event_name)
+        # emit collect-status events
```
tonyandrewmeyer, line 515:
```suggestion
        # Emit collect-status events.
```

tonyandrewmeyer (approved):
Looks great - thanks for persisting with this through all the reviewing!

A few very small nits around docs/comments (some of which might actually have been copied over from the original locations), but nothing else.

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
 
-def main(charm_class: Type[ops.charm.CharmBase],
-         use_juju_for_storage: Optional[bool] = None):
-    """Setup the charm and dispatch the observed event.
+class _Abort(Exception):  # noqa: N818
+    """Raised when something happens that should interrupt ops execution."""
 
-    The event name is based on the way this executable was called (argv[0]).
+    def __init__(self, exit_code: int, **kwargs: Any):
+        super().__init__(*kwargs)
```
benhoyt, line 372: Any reason the kwargs forwarding is needed here?
```suggestion
    def __init__(self, exit_code: int):
        super().__init__()
```

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+
+        model = ops.model.Model(self._charm_meta, self._model_backend,
+                                broken_relation_id=broken_relation_id)
+        store = self._make_storage(dispatcher)
+        framework = ops.framework.Framework(store, self._charm_root, self._charm_meta, model,
+                                            event_name=dispatcher.event_name)
+        framework.set_breakpointhook()
+        return framework
+
+    def _emit_charm_event(self, event_name: str):
```
benhoyt, line 493: This one-liner method doesn't pay its way anymore -- let's just have `_emit` call the top-level function directly.

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        """Commit the framework and gracefully teardown."""
+        self.framework.commit()
+
+    def run(self):
+        """Emit and then commit the framework."""
+        try:
+            self._emit()
+            self._commit()
+        finally:
+            self.framework.close()
```
benhoyt, line 528: Yep, much cleaner than before -- thanks.

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        charm_class: the charm class to instantiate and receive the event.
+        use_juju_for_storage: whether to use controller-side storage. If not specified
+            then kubernetes charms that haven't previously used local storage and that
+            are running on a new enough Juju default to controller-side storage,
+            otherwise local storage is used.
+    """
+    try:
+        manager = _Manager(
+            charm_class,
+            model_backend=ops.model._ModelBackend(),
```
benhoyt, line 547: Nit. This is now the default, so we can drop this line:
```suggestion
```

test/test_charm.py
```diff
@@ -473,6 +476,46 @@ def _get_action_test_meta(cls):
...
+    def test_meta_from_charm_root(self):
+        with tempfile.TemporaryDirectory() as d:
+            td = pathlib.Path(d)
+            (td / 'metadata.yaml').write_text(
+                yaml.safe_dump(
+                    {"name": "bob",
+                        "requires": {
+                            "foo":
+                                {"interface": "bar"}
+                        }}))
```
benhoyt, line 488: The formatting is very odd here. Can we clean up, or just write text directly? Similar below.

```suggestion
            (td / 'metadata.yaml').write_text("""
name: bob
requires:
  foo:
    interface: bar
""")
                yaml.safe_dump(
                    {"name": "bob",
                        "requires": {
                            "foo":
                                {"interface": "bar"}
                        }}))
```

test/test_main.py
```diff
@@ -89,7 +89,7 @@ def __init__(self,
 
 
 @patch('ops.main.setup_root_logging', new=lambda *a, **kw: None)  # type: ignore
-@patch('ops.main._emit_charm_event', new=lambda *a, **kw: None)  # type: ignore
+@patch('ops.main._Manager._emit_charm_event', new=lambda *a, **kw: None)  # type: ignore
```
benhoyt, line 92: These can go back to `ops.main._emit_charm_event`.

benhoyt (approved):
Approving, but just a few nitty nits (and Tony's doc suggestions). Thanks!

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
+        return framework
+
+    def _emit_charm_event(self, event_name: str):
+        """Emits a charm event based on a Juju event name.
 
+        Args:
+            event_name: A Juju event name to emit on a charm.
+        """
+        _emit_charm_event(self.charm, event_name)
+
```
tonyandrewmeyer, line 500:
```suggestion
```

ops/main.py
```diff
@@ -369,97 +365,191 @@ def _should_use_controller_storage(db_path: Path, meta: CharmMeta) -> bool:
...
         # Skip reemission of deferred events for collect-metrics events because
         # they do not have the full access to all hook tools.
-        if not dispatcher.is_restricted_context():
-            framework.reemit()
+        if not self.dispatcher.is_restricted_context():
+            # re-emit any deferred events from the previous run
+            self.framework.reemit()
+
+        # emit the juju event
+        self._emit_charm_event(self.dispatcher.event_name)
```
tonyandrewmeyer, line 514:
```suggestion
        _emit_charm_event(self.charm, self.dispatched.event_name)
```

tonyandrewmeyer:
I went ahead and made the small tweaks - no point bouncing this around again for that. I'll just run this against the charm tests as a sanity check that it's not going to break anything, and then we'll be good to go.

PR #1086, opened 2023-12-07
feat: support for Pebble Notices

benhoyt:
This PR implements Ops support for Pebble Notices (described in [JU048](https://docs.google.com/document/d/16PJ85fefalQd7JbWSxkRWn0Ye-Hs8S1yE99eW7pk8fA/edit), with the user ID details in [OP042](https://docs.google.com/document/d/1tQwUxz-rV-NjH-UodDbSDhGcMJGfD3OSoTnBLI9aoGU/edit)).

This adds a new `PebbleNoticeEvent` base event type, with `PebbleCustomNoticeEvent` being the first concrete type -- we'll have more later, such as for `warning` and `change-update` notices. These events have a `notice` attribute which is a `LazyNotice` instance: if you only access `id` or `type` or `key` it won't require a call to Pebble (I think only accessing `key` will be common in charms).

In addition, this adds `get_notice` and `get_notices` methods to both `pebble.Client` and `model.Container`. These return objects of the new `Notice` type.

The `timeconv.parse_duration` helper function is needed by `Notice.from_dict`, to parse Go's [`time.Duration.String`](https://pkg.go.dev/time#Duration.String) format that Pebble uses for the `repeat-after` and `expire-after` fields. Most of the test cases for this were taken from [Go's test cases](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/time/time_test.go;l=891).

This PR does *not* include `Harness` support for Pebble Notices. I plan to add that in a follow-up PR soon (using a `Harness.pebble_notify` function and implementing the `get_notice` and `get_notices` test backend functions).


docs/conf.py
```diff
@@ -73,28 +73,30 @@ def _compute_navigation_tree(context):
 # domain name if present. Example entries would be ('py:func', 'int') or
 # ('envvar', 'LD_LIBRARY_PATH').
 nitpick_ignore = [
-    ('py:class', 'ops.model._AddressDict'),
+    # Please keep this list sorted alphabetically.
```
benhoyt, line 76: Note that the only thing I've added here is `_NoticeDict`, but I sorted the list to make the it easier to scan.

ops/model.py
```diff
@@ -3489,3 +3525,57 @@ def validate_label_value(cls, label: str, value: str):
         if re.search('[,=]', v) is not None:
             raise ModelError(
                 f'metric label values must not contain "," or "=": {label}={value!r}')
+
+
+class LazyNotice:
```
tonyandrewmeyer, line 3530: I might have said this when we talked about it earlier, but in case not: I really like this approach.

ops/model.py
```diff
@@ -2719,6 +2719,42 @@ def send_signal(self, sig: Union[int, str], *service_names: str):
...
+        """Get details about a single notice by ID.
+
+        Raises:
+            ModelError: if a notice with the given ID is not found
+        """
+        try:
+            return self._pebble.get_notice(id)
+        except pebble.APIError as e:
+            if e.code == 404:
+                raise ModelError(f'notice {id!r} not found')
```
tonyandrewmeyer, line 2732:
```suggestion
                raise ModelError(f'notice {id!r} not found') from e
```

ops/model.py
```diff
@@ -2719,6 +2719,42 @@ def send_signal(self, sig: Union[int, str], *service_names: str):
...
+        parameters.
+        """
+        notices = self._pebble.get_notices(
+            select=select,
+            user_id=user_id,
+            types=types,
+            keys=keys,
+            after=after,
+        )
+        return notices
```
tonyandrewmeyer, line 2756: Any reason not to just do this?
```suggestion
        return self._pebble.get_notices(
            select=select,
            user_id=user_id,
            types=types,
            keys=keys,
            after=after,
        )
```

ops/model.py
```diff
@@ -3489,3 +3525,57 @@ def validate_label_value(cls, label: str, value: str):
...
+        try:
+            self.type = pebble.NoticeType(type)
+        except ValueError:
+            self.type = type
+        self.key = key
+
+        self._notice: Optional[ops.pebble.Notice] = None
+
+    @property
+    def notice(self) -> ops.pebble.Notice:
```
tonyandrewmeyer, line 3562: What's the use-case for exposing this?

ops/model.py
```diff
@@ -3489,3 +3525,57 @@ def validate_label_value(cls, label: str, value: str):
...
+        """The non-lazy :class:`ops.pebble.Notice` object (requires a call to Pebble)."""
+        self._ensure_loaded()
+        assert self._notice is not None
+        return self._notice
+
+    def __repr__(self):
+        return f'LazyNotice(id={self.id!r}, type={self.type}, key={self.key!r})'
```
tonyandrewmeyer, line 3569: Similar to another comment, it feels like it would be worth handling the case where `self.type` is a string here, although this is very much a nit :smile:.

```suggestion
        repr_type = self.type if isinstance(self.type, pebble.NoticeType) else repr(self.type)
        return f'LazyNotice(id={self.id!r}, type={repr_type}, key={self.key!r})'
```

test/test_pebble.py
```diff
@@ -2702,6 +2750,124 @@ def test_checklevel_conversion(self):
...
+                'occurrences': 7,
+            },
+            'status': 'OK',
+            'status-code': 200,
+            'type': 'sync',
+        })
+
+        notice = self.client.get_notice('124')
+
+        # No need to re-test full Notice.from_dict behaviour
```
tonyandrewmeyer, line 2772: My preference for comments is that they are a full sentence, including final punctuation, so:

```suggestion
        # No need to re-test full Notice.from_dict behaviour.
```

However, it's hard to get a smaller nit than this, and I don't think this is consistent in the rest of ops, so feel free to discard this comment :smile:.

test/test_pebble.py
```diff
@@ -2702,6 +2750,124 @@ def test_checklevel_conversion(self):
...
+                'last-occurred': '2023-12-07T17:01:03.123456789Z',
+                'last-repeated': '2023-12-07T17:01:04.123456789Z',
+                'occurrences': 7,
+            },
+            'status': 'OK',
+            'status-code': 200,
+            'type': 'sync',
+        })
+
+        notice = self.client.get_notice('124')
```
tonyandrewmeyer, line 2770: Is this deliberately asking for a notice of ID 124 and getting back a notice with ID 123? If that happened, wouldn't that be a Pebble bug?

tonyandrewmeyer (commented):
Looks great to me! A few minor comments/suggestions.

Normally with a significant change like this I like to install it locally and play around with it a bit to get a more holistic feel than from a series of diffs. I haven't done that since (as you mentioned in ~ops-library-development) it's complicated to do that at the moment since this hasn't landed in a Juju release. I generally feel that given how much work there has been on the spec, it's probably unnecessary, but if there is some way to (maybe partially just with pebble but not juju support?) do that, it'd be nice.

ops/_private/timeconv.py
```diff
@@ -24,6 +24,21 @@
...
 _TIMEOFFSET_RE = re.compile(r'([-+])(\d{2}):(\d{2})')
 
+# Matches n.n<unit> (allow U+00B5 micro symbol as well as U+03BC Greek letter mu)
+_DURATION_RE = re.compile(r'([0-9.]+)([a-zµμ]+)')
+
+# Mapping of unit to float seconds
+_DURATION_UNITS = {
+    'ns': 0.000_000_001,
+    'us': 0.000_001,
+    'µs': 0.000_001,  # U+00B5 = micro symbol
```
hpidcock, line 34: nice

ops/_private/timeconv.py
```diff
@@ -57,3 +72,40 @@ def parse_rfc3339(s: str) -> datetime.datetime:
...
+        number, unit = match.groups()
+        if unit not in _DURATION_UNITS:
+            raise ValueError(f'invalid duration: invalid unit {unit!r}')
+        try:
+            seconds += float(number) * _DURATION_UNITS[unit]
+        except ValueError:
+            # Same exception type, but a slightly more specific error message
+            raise ValueError(f'invalid duration: {number!r} is not a valid float') from None
+
+    duration = datetime.timedelta(seconds=seconds)
```
hpidcock, line 110:
```
datetime.timedelta(days=0, seconds=0, microseconds=0, milliseconds=0, minutes=0, hours=0, weeks=0)
```

ops/_private/timeconv.py
```diff
@@ -57,3 +72,40 @@ def parse_rfc3339(s: str) -> datetime.datetime:
...
+    if matches[0].start() != 0 or matches[-1].end() != len(s):
+        raise ValueError('invalid duration: extra input at start or end')
+
+    seconds = 0
+    for match in matches:
+        number, unit = match.groups()
+        if unit not in _DURATION_UNITS:
+            raise ValueError(f'invalid duration: invalid unit {unit!r}')
+        try:
+            seconds += float(number) * _DURATION_UNITS[unit]
```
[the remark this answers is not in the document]
hpidcock: Go seems to be happy with parsing strange stuff like `100ms0.1s0.1s0.1s0.6s`, so I guess this if fine.

I think we should drop the forced conversion to float, each number should stay as integers if possible. Have a variable for each of hour, minute, second, millisecond, microsecond and nanosecond.

Then for nanosecond, since datetime.timedelta doesn't support it, follow its rounding (half to even) when converting to microseconds, add it onto the microsecond (so microsecond can stay an integer if it already is one).

hpidcock (commented):
Yeah looks good, only really comments about the duration parsing stuff. I'm not a python expert, so the rest looks like it follows the model.

benhoyt:
@tonyandrewmeyer:

> Normally with a significant change like this I like to install it locally and play around with it a bit to get a more holistic feel than from a series of diffs. I haven't done that since (as you mentioned in ~ops-library-development) it's complicated to do that at the moment since this hasn't landed in a Juju release. I generally feel that given how much work there has been on the spec, it's probably unnecessary, but if there is some way to (maybe partially just with pebble but not juju support?) do that, it'd be nice.

Yeah, good call. I'm testing with the following charm (you can replace `src/charm.py` in the test charm [in this Juju PR](https://github.com/juju/juju/pull/16428) with the following code):

```python
#!/usr/bin/env python3
"""Charm to test Pebble Notices."""

import logging

import ops

class PebbleNoticesCharm(ops.CharmBase):
    """Charm to test Pebble Notices."""

    def __init__(self, *args):
        super().__init__(*args)
        self.framework.observe(self.on["redis"].pebble_ready, self._on_pebble_ready)
        self.framework.observe(self.on["redis"].pebble_custom_notice, self._on_custom_notice)

    def _on_pebble_ready(self, event):

    def _on_custom_notice(self, event):
        logger.info(
            f"_on_custom_notice: id={event.notice.id} type={event.notice.type} key={event.notice.key} occurrences={event.notice.occurrences}"
        )

        # Don't include the (arbitrary) ID in the status message
        self.unit.status = ops.MaintenanceStatus(
            f"notice type={event.notice.type} key={event.notice.key}"
        )

        container = event.workload
        notices = container.get_notices(types=[ops.pebble.NoticeType.CUSTOM])
        for i, notice in enumerate(notices):
            logger.info(f"get_notices {i}: {notice}")


if __name__ == "__main__":
    ops.main(PebbleNoticesCharm)
```

And then I `charmcraft pack` and deploy as per the "QA steps" [on that PR](https://github.com/juju/juju/pull/16428), but between packing and deploying, use this little script to inject my local copy of ops (with the changes in *this* PR):

```bash
#!/usr/bin/env bash

if [ "$#" -lt 2 ]
then
	echo "Inject local copy of Python Operator Framework source into charm"
	echo
    echo "usage: inject-ops.sh file.charm /path/to/ops/dir" >&2
    exit 1
fi

if [ ! -f "$2/framework.py" ]; then
    echo "$2/framework.py not found; arg 2 should be path to 'ops' directory"
    exit 1
fi

set -ex

mkdir inject-ops-tmp
unzip -q $1 -d inject-ops-tmp
rm -rf inject-ops-tmp/venv/ops
cp -r $2 inject-ops-tmp/venv/ops
cd inject-ops-tmp
zip -q -r ../inject-ops-new.charm .
cd ..
rm -rf inject-ops-tmp
rm $1
mv inject-ops-new.charm $1
```

I've just tested using this and it works end to end. Hope that helps.

ops/pebble.py
```diff
@@ -1277,6 +1292,90 @@ def __repr__(self):
...
+    user_id: Optional[int]
+    """UID of the user who may view this notice (None means notice is public)."""
+
+    type: Union[NoticeType, str]
+    """Type of the notice."""
+
+    key: str
+    """The notice key, a string that differentiates notices of this type.
+
+    This is in the format ``domain.com/path``.
```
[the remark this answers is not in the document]
benhoyt: Good call, changed to `example.com/path`

ops/pebble.py
```diff
@@ -1277,6 +1292,90 @@ def __repr__(self):
...
+    """
+
+    occurrences: int
+    """The number of times one of these notices has occurred."""
+
+    last_data: Dict[str, str] = dataclasses.field(default_factory=dict)
+    """Additional data captured from the last occurrence of one of these notices."""
+
+    repeat_after: Optional[datetime.timedelta] = None
+    """How long after one of these was last repeated should we allow it to repeat."""
```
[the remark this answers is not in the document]
benhoyt: Fixed.

ops/testing.py
```diff
@@ -3259,3 +3259,17 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
 
     def get_checks(self, level=None, names=None):  # type:ignore
         raise NotImplementedError(self.get_checks)  # type:ignore
+
+    def get_notice(self, id: str) -> pebble.Notice:
+        raise NotImplementedError(self.get_notice)
```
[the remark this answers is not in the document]
benhoyt: Yes, very much so. I just wanted to make progress on merging this PR, which is already quite large. As noted in the PR description:

> This PR does not include Harness support for Pebble Notices. I plan to add that in a follow-up PR soon (using a Harness.pebble_notify function and implementing the get_notice and get_notices test backend functions).

ops/pebble.py
```diff
@@ -2645,6 +2744,53 @@ def get_checks(
...
+    def get_notices(
+        self,
+        *,
+        select: Optional[NoticesSelect] = None,
+        user_id: Optional[int] = None,
+        types: Optional[Iterable[NoticeType]] = None,
+        keys: Optional[Iterable[str]] = None,
+        after: Optional[datetime.datetime] = None,
+    ) -> List[Notice]:
+        """Query for notices that match the provided filters.
```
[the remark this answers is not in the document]
benhoyt: Done (in one-liner and clarified in body of docstring).

ops/pebble.py
```diff
@@ -2645,6 +2744,53 @@ def get_checks(
...
+            keys: filter for notices with any of the specified keys
+            after: filter for notices that were last repeated after this time
+        """
+        query: Dict[str, Union[str, List[str]]] = {}
+        if select is not None:
+            query['select'] = select.value
+        if user_id is not None:
+            query['user-id'] = str(user_id)
+        if types is not None:
+            query['types'] = [t.value for t in types]
```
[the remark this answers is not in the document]
benhoyt: Good call, done.

ops/model.py
benhoyt: Heh, I had it that way originally, but then I felt the `return` got lost. However, I changed it back -- this way it's clearer it's proxying straight through.

ops/model.py
benhoyt: Thanks, fixed (and added test).

test/test_pebble.py
benhoyt: Fair, updated.

test/test_pebble.py
benhoyt: Fixed.

ops/model.py
benhoyt: That's a fair call. I was thinking in case they wanted a "real" `pebble.Notice` type. But I don't see why you'd need that. Python works more with duck-typing. I've removed it for now, and we can always reconsider if a use case comes up.

ops/testing.py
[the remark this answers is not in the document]
tonyandrewmeyer: Oops, sorry :smile:. I read read the description, promise! But obviously that bit didn't stick in my mind!

tonyandrewmeyer:
> Yeah, good call. I'm testing with the following charm (you can replace `src/charm.py` in the test charm [in this Juju PR](https://github.com/juju/juju/pull/16428) with the following code):

Thanks! I played around with it a bit and all still seems good, as expected :)

tonyandrewmeyer (approved):

ops/_private/timeconv.py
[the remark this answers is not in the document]
benhoyt: I've done this in https://github.com/canonical/operator/pull/1086/commits/40357ef11a51e4e0e4d501fabee7238189678d3d -- thanks for the suggestion, @hpidcock

ops/_private/timeconv.py
benhoyt: I've taken this approach in https://github.com/canonical/operator/pull/1086/commits/40357ef11a51e4e0e4d501fabee7238189678d3d -- thanks

PR #1089, opened 2023-12-11
docs: update README.md in light of changes to `charmcraft init` output

tmihoc:
`charmcraft init` no longer produces a `metadata.yaml`, `actions.yaml`, `config.yaml` -- instead, it's just `charmcraft.yaml`. 

This PR updates the README to replace references to `metadata.yaml` with `charmcraft.yaml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,30 @@ mkdir ops-example
 cd ops-example
 charmcraft init
 ```
-This has created a standard charm directory structure. Poke around. 
+This has created a standard charm directory structure:
+
+```
+$ ls -R
+.:
+CONTRIBUTING.md  README.md        pyproject.toml    src    tox.ini
+LICENSE          charmcraft.yaml  requirements.txt  tests
+
+./src:
+charm.py
+
+./tests:
+integration  unit
+
+./tests/integration:
+test_charm.py
+
+./tests/unit:
+test_charm.py
+```
 
 Things to note:
 
-- The `metadata.yaml` file shows that what we have is an example charm called `ops-example`, which uses an OCI image resource `httpbin` from `kennethreitz/httpbin`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,11 +34,30 @@ mkdir ops-example
 cd ops-example
 charmcraft init
 ```
-This has created a standard charm directory structure. Poke around. 
+This has created a standard charm directory structure:
+
+```
```
tonyandrewmeyer, line 39:
```suggestion
```shell-script
```

README.md
```diff
@@ -34,11 +34,30 @@ mkdir ops-example
...
+test_charm.py
+
+./tests/unit:
+test_charm.py
+```
 
 Things to note:
 
-- The `metadata.yaml` file shows that what we have is an example charm called `ops-example`, which uses an OCI image resource `httpbin` from `kennethreitz/httpbin`.
+- The `charmcraft.yaml` file shows that what we have is an example charm called `ops-example`, which uses an OCI image resource `httpbin` from `kennethreitz/httpbin`.
```
tonyandrewmeyer, line 60: Do we know that the charm-dev multipass VM template referred to in the setup will give a recent version of charmcraft? I'm wondering if we have to say that it might be charmcraft.yaml or it might be metadata.yaml - but it would be much nicer to only have charmcraft.yaml as you have it, if we can be confident people will always end up with that version.

tonyandrewmeyer (approved):
Thanks! Looks good to me.

README.md
benhoyt: Yeah, agreed it would be good to only (confidently) mention one moving forward. I think we can leave this PR as is, but maybe @tmihoc can confirm about the charm-dev VM and ping someone to update that if needed?

benhoyt (approved):

README.md
tmihoc: I've tested. The charm-dev VM charmcraft only produces a charmcraft.yaml file.

Issue #1075, opened 2023-11-22
Hook context: Add a clear boundary between ops and juju


sed-i:
Hook tools ([1](https://github.com/canonical/operator/blob/main/ops/model.py), [2](https://github.com/canonical/operator/pull/973)) and [pebble.py](https://github.com/canonical/operator/blob/main/ops/pebble.py) already draw a pretty clear boundary between ops and juju, but there is a key part missing: [environment variables](https://juju.is/docs/sdk/charm-environment-variables).

Wouldn't it be nice if the juju context could be represented as a plain, flat class directly from envvars, and without any deps? And then reused within ops for whatever purpose, without ever again hardcoding envvar string literals?

## Conceptual design for hook context

```python
# ops/juju_context.py

import os


class JujuContext:
    def __init__(self, context: dict):
        self.model = context.get("JUJU_MODEL_NAME")
        self.model_uuid = context.get("JUJU_MODEL_UUID")
        self.unit_name = context.get("JUJU_APP_NAME")
        self.app_name, self.unit_id = self.unit_name.split("/")
        self.juju_version = context.get("JUJU_VERSION")
        self.hook_name = context.get("JUJU_HOOK_NAME")


class InstallContext(JujuContext):
    pass


class LeaderElectedContext(JujuContext):
    pass


# Same for start, config-changed, update-status, stop, remove


class PebbleReadyContext(JujuContext):
    def __init__(self, context: dict):
        super().__init__(context)
        self.workload_name = context["JUJU_WORKLOAD_NAME"]


class RelationContext(JujuContext):
    def __init__(self, context: dict):
        super().__init__(context)
        self.relation_name = context["JUJU_RELATION"]
        self.remote_app_name = context["JUJU_REMOTE_APP"]
        self.relation_id = context["JUJU_RELATION_ID"].split(":")[1]

    @property
    def remote_app_data(self):
        return "hook_tools.relation_get(self.relation_id, self.remote_app_name, is_app=True)"

    @property
    def local_app_data(self):
        return "hook_tools.relation_get(self.relation_id, self.unit_name, is_app=True)"


class RelationCreatedContext(RelationContext):
    pass


class RelationJoinedContext(RelationContext):
    def __init__(self, context: dict):
        super().__init__(context)
        self.remote_unit_name = context["JUJU_REMOTE_UNIT"]

    @property
    def remote_unit_data(self):
        return "hook_tools.relation_get(self.relation_id, self.remote_unit_name, is_app=False)"


[...]

Issue #793, opened 2022-07-06
Better error on attempting to set a broken relation data; facilitate guarding


PietroPasotti:
If you attempt to set relation data on a relation-broken hook, you'll get 

```ops.model.ModelError: b'ERROR cannot read relation application settings: permission denied (unauthorized access)```
Which is confusing and already tracked on #784

When it came to fixing the error, I thought of all patterns I've seen; from setting a `self._is_relation_departed=True` flag, to having dedicated code paths, passing down kwargs etc... I thought, wouldn't it be nice to have a relation status?

```python
if self.unit.is_leader() and relation.status.name != 'broken':
   relation.data[self.app][foo] = bar
```

I feel that there is something of a relation Status in the works (is there?), so I'm not sure if this would be compatible with that. Food for thought?

benhoyt:
This will be solved with #940: `Relation.broken`, though we may go with `Relation.active` or `Relation.status` (enum) instead.

PR #1091, opened 2023-12-14
feat: add Relation.active, exclude inactive relations from Model.relations

### LINKED ISSUE #940: Add Relation.broken to allow querying if a relation is broken (breaking)

tonyandrewmeyer:
This PR allows Charmers to easily check whether a relation is broken during a `relation-broken` event (and also framework events that share the same Juju context) without having to store state or pass the `event` object around. In addition, the broken relation is excluded from the `Model.relations` list, to make it easy for charms to work on only the relations that will continue to exist after the relation-broken events complete (without having to do `(rel for rel in self.model.relations if rel.active)` themselves.

To do this:

* In `relation-broken` events, set a new attribute `active` on the event's `Relation` object to False (and have this be True in all other cases).
* In `relation-broken` events, `RelationMapping` `__getitem__` excludes the event's relation.

Builds on work from @ca-scribner in https://github.com/canonical/operator/pull/958.

[Additional background notes](https://docs.google.com/document/d/1yKLRtfffxxLeH4tkfkOL4ELfzyJK33WC5lQaCZXgjjk/edit?usp=sharing) (limited to Canonical).

Fixes #940


tonyandrewmeyer:
I found 25 Charms (in the Canonical charms list we have collated) that have relation-broken events. 4 of those 25 use `Model.relations` in a relation-broken handler, and all should continue to work with (and be better for) this change. 3 of the 25 charms have an existing form of "relation is active" so could have simplified code after this change.

tonyandrewmeyer:
Among the use cases this should address:
* Collect status - to put the unit/app into a Blocked state when a required relation is broken ([example](https://github.com/canonical/sdcore-smf-k8s-operator/pull/34/files)) - this can just do `if not self.model.relations[name]` and it will cover both the relation being broken and the relation not existing.
* Avoiding setting relation data during relation broken ([example](https://github.com/canonical/operator/issues/793)) - this can just do `and event.active`
* Update configuration to exclude the broken relation without passing around the event ([example](https://github.com/canonical/sdcore-nms-k8s-operator/issues/27)) - this should also be covered by `relations` excluding the broken relation. The only open question is the idea mentioned in that issue that an event _after_ relation-broken could still list the broken relation.
* Know which properties can be accessed ([example](https://github.com/canonical/mysql-router-k8s-operator/commit/608c094273afdb7adb8ad3e9ef975ca4f8af4865)) - covered by `.active`.

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# 2.10.0
+
+* Added `Relation.active`, and excluded inactive relations from `Model`.relations
```
benhoyt, line 3:
```suggestion
* Added `Relation.active`, and excluded inactive relations from `Model.relations`
```

ops/model.py
```diff
@@ -822,8 +832,10 @@ def __getitem__(self, relation_name: str) -> List['Relation']:
         if not isinstance(relation_list, list):
             relation_list = self._data[relation_name] = []  # type: ignore
             for rid in self._backend.relation_ids(relation_name):
+                if rid == self._broken_relation_id:
+                    continue
                 relation = Relation(relation_name, rid, is_peer,
-                                    self._our_unit, self._backend, self._cache)
+                                    self._our_unit, self._backend, self._cache, True)
```
benhoyt, line 838: Let's make the meaning clearer by using the keyword arg here (and below):

```suggestion
                                    self._our_unit, self._backend, self._cache, active=True)
```

Or perhaps we could make the arg default to `active=True` (seems reasonable) and leave it off here?

ops/model.py
```diff
@@ -1453,13 +1465,20 @@ class Relation:
     This is accessed using, for example, ``Relation.data[unit]['foo']``.
     """
 
+    active: bool
+    """Indicates whether this relation is active.
+
+    This will be ``False`` if the current event is a ``relation-broken`` event
+    assoeciated with this relation."""
```
benhoyt, line 1472: Typo fix, and consistency nit:
```suggestion
    associated with this relation.
    """
```

ops/testing.py
```diff
@@ -878,8 +878,15 @@ def remove_relation(self, relation_id: int) -> None:
         for unit_name in rel_list_map[relation_id].copy():
             self.remove_relation_unit(relation_id, unit_name)
 
+        prev_broken_id = None  # Silence linter warning.
+        if self._model is not None:
+            prev_broken_id = self._model._relations._broken_relation_id
+            self._model.relations._broken_relation_id = relation_id
```
benhoyt, line 884: I'm not sure I quite follow what these changes are doing -- can you explain?

test/test_model.py
```diff
@@ -3370,16 +3398,16 @@ def test_grant(self):
         meta = ops.CharmMeta()
         cache = ops.model._ModelCache(meta, backend)
         unit = ops.Unit('test', meta, backend, cache)
-        rel123 = ops.Relation('test', 123, True, unit, backend, cache)
-        rel234 = ops.Relation('test', 234, True, unit, backend, cache)
+        rel123 = ops.Relation('test', 123, True, unit, backend, cache, True)
```
benhoyt, line 3401: Similar in this file, I'd recommend `active=True` (or, probably better, just make the default True).

ops/model.py
```diff
@@ -796,15 +801,20 @@ def __repr__(self):
 class RelationMapping(Mapping[str, List['Relation']]):
     """Map of relation names to lists of :class:`Relation` instances."""
 
-    def __init__(self, relations_meta: Dict[str, 'ops.RelationMeta'], our_unit: 'Unit',
-                 backend: '_ModelBackend', cache: '_ModelCache'):
+    def __init__(self,
```
benhoyt, line 804: I'm somewhat surprised we don't need to modify `__iter__` / `__contains__` / `__len__`. Wouldn't these have to exclude the broken relation too, or am I missing something?

I guess we don't need any changes to `Model.get_relation` / `RelationMapping._get_unique` because there they're asking for a specific relation?

benhoyt (changes requested):
A couple of questions and some minor comments.

ops/model.py
tonyandrewmeyer: Ha, [you said to not give it a default here](https://github.com/canonical/operator/pull/958/files#r1252495867) :smile:. I originally had it with a default and changed when I saw that comment - your point about it catching all the places it's called is valid, although I think that is probably just in the `ops` code and that's covered now.

I do feel that `active=True` is actually the default state and `active=False` is the unusual case, so a default of `True` makes sense to me. I guess current-Ben plus me outweighs previous-Ben :joy:.

ops/testing.py
tonyandrewmeyer: Does the comment make it clear?

ops/model.py
benhoyt: Ha, that's hilarious. Previous Ben had some good reasoning. But yeah, let's go with current-Ben + Tony. :-) Thanks.

ops/testing.py
benhoyt: Ah yes, thanks!

ops/model.py
tonyandrewmeyer: `_get_unique` changes mostly indirectly.

If the relation ID is passed, then it does `for relation in self[relation_name]` and that utilises the change in `__getitem__` to skip past the broken relation, so we end up in the `else` clause of the `for` loop, which does have the `active=False` change. I feel this is the right behaviour.

If the relation ID is not passed, then it does (essentially) `len(self[relation_name])` which again skips the broken relation. In that case it'll return None rather than a `Relation` with `active==False`. I'm not certain that this is correct.

The docs say "If relation_id is not given, this will return the Relation instance if the relation is established only once or None if it is not established". It seemed like "not established" was functionally the same as "broken", so you'd expect None. Perhaps that's wrong, though?

If you have a broken relation (ID 10, name "db") and do `Model.get_relation("db", 10)` you get a `Relation` with `active==False`, and if you do `Model.get_relation("db")` you get `None`. I think I have just about convinced myself that it should actually give the same in both cases, but let me know what you think.

tonyandrewmeyer: `__iter__`, `__contains__`, and `__len__` I think I missed :disappointed:.

tonyandrewmeyer: > `__iter__`, `__contains__`, and `__len__` I think I missed 😞.

No, I just forgot that I'd looked at them :laughing:.

These all currently work on the defined relations, ie. what's in the metadata, and not the current state.

For example, if I have a charm that has this in `charmcraft.yaml`/`metadata.yaml`:

```yaml
requires:
    database:
        interface: postgresql_client
        limit: 1
```

And I deploy the charm, *but I do not install postgresql and do not do a `juju integrate postgresql-k8s my-charm`*, so there is a relation required but not made, then:

* `len(self.model.relations)` is `1`
* `"database" in self.model.relations` is `True`
* `for rel_name in self.model.relations:` will have one iteration, where `rel_name` is `database`

If I do install postgresql and do `juju integrate` then those all stay the same - it's `self.model.relations["database"]` that will change, from `[]` to `[Relation(...)]`

I don't think anything should change here with relation-broken. If the relation is broken, it's still defined/required/provided.

benhoyt:
>  3 of the 25 charms have an existing form of "relation is active" so could have simplified code after this change.

@tonyandrewmeyer Do you happen to have links to the code where these 3 charms do this? I'd be curious to see examples of that and how people are doing it now.

benhoyt:
Oh wait, I see you've linked to a couple of them in your follow-up [comment](https://github.com/canonical/operator/pull/1091#issuecomment-1855552240) above -- that's fine, thanks.

ops/model.py
benhoyt: Excellent, thanks for the replies -- resolving.

ops/model.py
```diff
@@ -147,6 +149,9 @@ def relations(self) -> 'RelationMapping':
 
         Answers the question "what am I currently related to".
         See also :meth:`.get_relation`.
+
+        In a ``relation-broken`` event, the broken relation is excluded from
+        this list.
```
benhoyt, line 154: Thanks for amending the reference docs!

ops/model.py
```diff
@@ -1453,13 +1465,21 @@ class Relation:
     This is accessed using, for example, ``Relation.data[unit]['foo']``.
     """
 
+    active: bool
+    """Indicates whether this relation is active.
+
+    This will be ``False`` if the current event is a ``relation-broken`` event
+    associated with this relation.
```
benhoyt, line 1472: I think it'd be good to mention the (normal) True case. Perhaps:

```suggestion
    This is normally ``True``; it will be ``False`` if the current event is a
    ``relation-broken`` event associated with this relation.
```

test/charms/test_main/src/charm.py
```diff
@@ -170,6 +176,8 @@ def _on_ha_relation_broken(self, event: ops.RelationBrokenEvent):
             'relation-broken events cannot have a reference to a remote application')
         assert event.unit is None, (
             'relation broken events cannot have a reference to a remote unit')
+        assert not event.relation.active, "relation broken events always have a broken relation"
```
benhoyt, line 179: Micro-nit: would be good if we can be consistent about quoting style (single quotes in this case). Also in `ops/main.py:405` and `test/test_testing.py:610`.

Should we (in a separate PR) use flake8's quoting support to automatically do this? (There are several inconsistencies in the existing codebase we'll have to fix.) https://pypi.org/project/flake8-quotes/ ... Or I could just stop noticing it. :-)

benhoyt (approved):
Looks good, thanks! Couple of very minor optional comments.

test/charms/test_main/src/charm.py
tonyandrewmeyer: > Micro-nit: would be good if we can be consistent about quoting style (single quotes in this case). Also in `ops/main.py:405` and `test/test_testing.py:610`.

Fixed.

> Should we (in a separate PR) use flake8's quoting support to automatically do this? (There are several inconsistencies in the existing codebase we'll have to fix.) https://pypi.org/project/flake8-quotes/ ... Or I could just stop noticing it. :-)

We should definitely do one of those, yes :joy:. I very strongly prefer to have a tool that either auto-fixes or auto-checks style rules, rather than relying on documented (or worse, undocumented) style guides. I'm also in favour of consistency, so having a linter wins over not noticing imo :smile:.

My habit is double-quotes - I think because I came to Python from predominately C/C++, but also (if I recall correctly) most of the Python developers I interacted with early on were double-quote users. Then black picking double-quotes just solidified that for me more. I'll eventually break the habit, but having tox remind me will help!

(I do see the arguments for single-quotes being preferable, particularly that `repr(string)` prefers single, so the language 'blesses' that form, and not needing shift on most English keyboard layouts. My defaulting to double is habit, not because I believe it's better/more correct).

benhoyt (approved):

test/charms/test_main/src/charm.py
benhoyt: Item opened on adding a linting rule: https://github.com/canonical/operator/issues/1099

jameinel (approved):
Ideally people don't need to respond to `.active` because juju just doesn't include those ids in `relation-ids` but I think this follows what we are trying to do, and is better than where we have been.

tonyandrewmeyer:
> Ideally people don't need to respond to `.active` because juju just doesn't include those ids in `relation-ids` but I think this follows what we are trying to do, and is better than where we have been.

Yes, the intention is that Juju changes to do this but to have ops lead the way.

benhoyt:
> Yes, the intention is that Juju changes to do this but to have ops lead the way.

@tonyandrewmeyer Can you please search to see if we already have a [Juju bug](https://bugs.launchpad.net/juju/) open for this (I kinda doubt it), and if not, open one to track?

ops/main.py
```diff
@@ -144,6 +144,10 @@ def _emit_charm_event(charm: 'ops.charm.CharmBase', event_name: str):
         event_to_emit.emit(*args, **kwargs)
 
 
+def _get_juju_relation_id():
```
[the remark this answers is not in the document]
tonyandrewmeyer: Splitting it out was actually a code review suggestion from @benhoyt :smile:. I'm weakly in favour of splitting it out for the reasons you mention.

Note that we have tentative plans to refactor the env parsing (#1075) and if that goes ahead then this would presumably end up in the initialisation of that context object and no longer need this little helper.

tonyandrewmeyer:
> @tonyandrewmeyer @benhoyt I think I might be missing something—how do charms access the relations that are breaking if they are excluded from `Model.relations`?
> 
> mysql-router relies on being able to access relations that are breaking so that it can properly clean up database users: https://github.com/canonical/mysql-router-k8s-operator/blob/d89b0f820cfc574b04e0b86b5412db9da0d7ff2e/src/relations/database_provides.py#L200-L208
> 
> More context: [#940 (comment)](https://github.com/canonical/operator/issues/940#issuecomment-1623559826)

If you're in the relation-broken event, then `event.relation` has the relation.

For the case above, you're [basically doing the same thing when you raise the `_RelationBreaking` exception](https://github.com/canonical/mysql-router-k8s-operator/blob/d89b0f820cfc574b04e0b86b5412db9da0d7ff2e/src/relations/database_provides.py#L68) (checking if it's relation-broken and if the ID matches the one in the event. So it looks to me like the behaviour would not change - at the moment you raise that exception and the relation doesn't get appended to the list, and with the newer version of ops you'd not actually have that relation in the loop (so it would still not get appended to the list) - and you'd be able to remove the custom "is this relation breaking" code that you have, since ops will be doing that for you.

Am I missing something?

tonyandrewmeyer:
> This change in ops means that the relation won't be in `self._shared_users`

Ah, that's the part I had missed, thanks!

We (mostly I) reviewed a lot of `.relations` use across 150+ charms in working on this, focusing on `relation-broken` but cases like this where it's difficult to know that `.relations` is being used from `relation-broken` could be (and in this case were) missed. When reviewing I would have got to the `reconcile_users` method but missed that `_shared_users` was a property that uses `.relations` as well.

I did run the tests (whichever ones a plain `tox` runs) in several charms (even some that don't handle running "tox" out of the box, like here), including this one, but I guess the tests don't cover this specific scenario, unfortunately (maybe there are integration tests or some others that a plain `tox` doesn't handle that do?).

We did know that this is backwards incompatible, which is why we tried to find cases where it might break charms, not just assume it would be ok. Obviously we (again, mostly I) didn't manage to do that well enough (and particularly given that you had commented in the original ticket: sorry - I did read that ticket many many times but it's tough to follow).

It would be easy enough to change the charm to work with the ops change - e.g. have `event.relation.delete_user(shell=shell)` in a `relation-departed` handler, or have something like:

```python
if not event.relation.active:
    relation.delete_user(shell=shell)
```

In the `reconcile_users` method. (And you'd need to pass the event to `delete_all_databags` and handle it there too).

If you want all the relations, even the broken one, then as long as you have the `event` object, you can get it and append it to the list the model provides.

Note that the intention is that Juju also makes this change, and that ops was just 'leading the way'.

@benhoyt is finished for the week, but I'll talk this over with him on Monday to decide whether this is too much of a change to make after all, and we should limit this to just the `.active` attribute.

Thanks for bringing this up (again), especially before the 2.10 release!

tonyandrewmeyer:
> > Note that the intention is that Juju also makes this change, and that ops was just 'leading the way'.
> 
> What exactly does this mean? Will we no longer be able to access local (not remote) databags during relation broken events?

It means that Juju would behave like current main ops does: relation-ids would not return the event that's broken in the relation-broken hook. Whether or not relation data is accessible is a separate concern.

tonyandrewmeyer:
> > It would be easy enough to change the charm to work with the ops change [...] in a `relation-departed` handler
> 
> [I don't think this would work](https://github.com/canonical/operator/pull/1091#issuecomment-1888670576)

Sorry, that was a typo (fixed now) - I meant in a `relation-broken` handler.

tonyandrewmeyer:
Sorry, ran out of time for this today, but will check tomorrow (NZ time) morning. I also need to double check we're consistent now that the [Juju fix](https://bugs.launchpad.net/juju/+bug/2024583) has landed, so the same behaviour happens regardless of Juju version.

PR #1092, opened 2023-12-15
docs: add instructions on how to use a custom version of ops in a Charm

tonyandrewmeyer:
Figuring out how to test changes to `ops` locally (without just monkeypatching everything) wasn't straightforward when I started - this should make it easier for new people going forward.

I've included two variants:

* If the code is in a Git branch then I find it's simplest to just have the `charmcraft.yaml` file make `git` available and use standard Python dependency specification.
* If the code isn't easily installable by `pip`, then editing the `.charm` zip file is probably simplest - I've included @benhoyt's `inject-ops.sh` script to make it easy for people.

I've also added a small note about running in custom Juju versions, since that was something I had to learn today and seems like it might also come up for other people working on `ops`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -91,7 +91,7 @@ in `requirements.txt` (or `pyproject.toml`) with a reference to the branch, like
 
 ```
 #ops ~= 2.9
-git+https://github.com/{your-username}}/operator@{your-branch-name}}
+git+https://github.com/{your-username}/operator@{your-branch-name}
 ```
 
 `git` is not normally available when `charmcraft` is packing the charm, so you'll
--- a/HACKING.md
+++ b/HACKING.md
@@ -79,6 +79,79 @@ source .tox/unit/bin/activate
 pytest -v test/test_real_pebble.py
 ```
 
+## Using an `ops` branch in a charm
+
+When making changes to `ops`, you'll commonly want to try those changes out in
+a charm.
+
+### From a Git branch
+
+If your changes are in a Git branch, you can simply replace your `ops` version
+in `requirements.txt` (or `pyproject.toml`) with a reference to the branch, like:
+
+```
+#ops ~= 2.9
+git+https://github.com/{your-username}}/operator@{your-branch-name}}
+```
+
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -79,6 +79,79 @@ source .tox/unit/bin/activate
...
+a charm.
+
+### From a Git branch
+
+If your changes are in a Git branch, you can simply replace your `ops` version
+in `requirements.txt` (or `pyproject.toml`) with a reference to the branch, like:
+
+```
+#ops ~= 2.9
+git+https://github.com/{your-username}}/operator@{your-branch-name}}
```
benhoyt, line 94: There are mismatched braces here.

Also, do we need the [`#egg=ops` thing](https://stackoverflow.com/questions/11835396/why-egg-foo-when-pip-installing-from-git-repo) at the end of the line here? I have always used it, but maybe it's not needed?

benhoyt (changes requested):
Excellent idea to include this!

Trivial issue with mismatched braces, and a question about the `#egg=ops` syntax.

HACKING.md
tonyandrewmeyer: > There are mismatched braces here.

Thanks, fixed!

> Also, do we need the [`#egg=ops` thing](https://stackoverflow.com/questions/11835396/why-egg-foo-when-pip-installing-from-git-repo) at the end of the line here? I have always used it, but maybe it's not needed?

Hmm, the plain version definitely works for me. But good call, I checked the [docs](https://pip.pypa.io/en/stable/topics/vcs-support/):

> pip also looks at the egg fragment specifying the “project name”. In practice the egg fragment is only required to help pip determine the VCS clone location in editable mode. In all other circumstances, the egg fragment is not necessary and its use is discouraged.

So since editable mode isn't relevant here, I believe the `#egg=` should not be included.

benhoyt: Thanks, good to know.

benhoyt (approved):

benhoyt:
Merged without further review, as this is a doc-only change. @tmihoc if you have further suggestions, let us know and we can always tweak later.

Issue #970, opened 2023-07-11
Add `cross_model` property to `model.Relation`


sed-i:
Our charm code needs to know which URL to advertise over relation data, so that remote app's workload would be able to reach our charm. For example, for in-model relation we would advertise the FQDN. For cross-model, we'd advertise the ingress URL.

Not sure if/where this is exposed by Juju:

https://github.com/juju/juju/blob/2b2308500d71d17867256cc81dce9dbce877bcc4/rpc/params/applications.go#L497
```
CrossModel       bool                    `json:"cross-model"`
```

sed-i:
Just wondering what's the prospect for this feature @benhoyt.
Seems like it is becoming critical to [our TLS story](https://github.com/canonical/traefik-k8s-operator/issues/214#issuecomment-1660868997): the same COS charm (e.g. prometheus) needs to advertise different URLs depending on where it is deployed relative to grafana:

```mermaid
graph LR

subgraph COS model
  prometheus ---|k8s fqdn| grafana

end
external-prometheus ---|ingress URL| grafana
```

1. COS charms generate CSRs with the k8s fqdn as the SAN DNS, and a local CA signs.
2. All COS charms trust the local CA.
3. COS charms also have an ingress URL provided by traefik.
4. Traefik's CSR is signed by an _external-ca_.

The problem is that we must now choose what URL COS charms advertise to each other (e.g. grafana datasource in prometheus):
- If we use fqdn, then TLS works but CMRs do not.
- If we use ingress, CMRs work, but now we'd need to have COS charms also trust the external CA.

Seems like it would make sense if charms could advertise fqdn when inside the same cluster, but ingress when cross cluster.

Other potential solutions:
- Have the (permissive) local ca (blindly) sign both fqdn and the ingress url. This is an antipattern.
- Somehow utilize SNI, as @tlm [proposed](https://chat.charmhub.io/charmhub/pl/omf31zr4yinymyx9bx93ux5xtr). Not sure if this is feasible here.
- Propagate the external CA cert into all relevant COS charms, e.g. by directly relating them to the external CA (antipattern) or have traefik forward its ca cert to all proxied apps (possibly an antipattern?).

cc: @simskij @gruyaume @deusebio

tlm:
Hi @sed-i,

I have had a read of the above. Thanks for the awesome right up of the problem. This is going to be an ongoing discussion but I do have some initial thoughts, statements and ideas.

[...]

PR #1096, opened 2023-12-15
Add temporary workaround for cmr check

sed-i:
Seems like #970 is planned to be addressed by the juju team, but not in the coming 6 months.
This PR adds a temporary, "private", imperfect fix for #970.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1488,6 +1488,20 @@ class Relation:
     def __repr__(self):
         return f'<{type(self).__module__}.{type(self).__name__} {self.name}:{self.id}>'
 
+    @property
+    def _cmr(self) -> bool:
+        """Temporary workaround for checking if the given relation is a cross-model relation.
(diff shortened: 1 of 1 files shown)


benhoyt:
Assuming there's not a better workaround discussed (I wasn't part of the discussion you had with Tom before the break), I think this approach is a reasonable hack until the Juju team can model this properly. However, @tonyandrewmeyer and I discussed, and we'd much prefer this 3-line function were just included in the charm or perhaps a charm lib. Basically the same code you've got, but as a stand-alone function:

```
def is_relation_cross_model(relation: ops.Relation) -> bool:
    ...
```

We don't want this in Ops itself (and it's just as easy to implement from outside Ops) because "private" but necessary functions will get used, and then it might as well not be "private". So I'm going to close this PR, but please go ahead and do what you need in a charm or charm lib.

PR #1100, opened 2024-01-04
feat: Harness support for Pebble notices

benhoyt:
This PR adds Harness support for Pebble Notices, specifically a new `Harness.pebble_notify` method which you call as follows -- this will add a custom notice with key "example.com/key" and some data:

```python
harness.pebble_notify('mycontainer', 'example.com/key', data={'optional': 'data'})
```

This will record a notice (which will be visible in the `Container.get_notice` and `Container.get_notices` return values) and, if the notice is new or repeated, trigger a `pebble-custom-notice` event like Juju would.

Note that I've based the fake Harness implementation on the Pebble Go code (for example, [`State.AddNotice`](https://github.com/canonical/pebble/blob/ecac3d4a3a6111e0ae1face2f9062f5210405826/internals/overlord/state/notices.go#L223) and [`v1GetNotices`](https://github.com/canonical/pebble/blob/ecac3d4a3a6111e0ae1face2f9062f5210405826/internals/daemon/api_notices.go#L45)). There are some tests of `pebble.Client.notify` and `get_notice` and `get_notices` that are run against a real Pebble server, like we do with some of the other Pebble-related functionality.

I've also removed the `get_notices` "after" param for now, as it's likely not needed in this context and is hard to implement with the Python microseconds vs Go nanoseconds discrepancy. We'll probably have to do this via a string parameter in future (or something other than `datetime.datetime`, as that only stores microseconds). But I don't think it'll be used here, as Juju does the long-polling where "after" is required.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2744,7 +2744,6 @@ class Container:
         user_id: Optional[int] = None,
         types: Optional[Iterable[Union[pebble.NoticeType, str]]] = None,
         keys: Optional[Iterable[str]] = None,
-        after: Optional[datetime.datetime] = None,
     ) -> List[pebble.Notice]:
         """Query for notices that match all of the provided filters.
 
@@ -2756,7 +2755,6 @@ class Container:
             user_id=user_id,
             types=types,
             keys=keys,
-            after=after,
         )
 
     # Define this last to avoid clashes with the imported "pebble" module
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2744,6 +2744,33 @@ class Client:
         resp = self._request('GET', '/v1/checks', query)
         return [CheckInfo.from_dict(info) for info in resp['result']]
 
+    def notify(self, type: NoticeType, key: str, *,
+               data: Optional[Dict[str, str]] = None,
+               repeat_after: Optional[datetime.timedelta] = None) -> str:
+        """Record an occurrence of a notice with the specified options.
+
+        Args:
+            type: Notice type (currently only "custom" notices are supported)
+            key: Notice key; must be in "domain.com/path" format
+            data: Data fields for this notice
+            repeat_after: Prevent notice with same type and key from
+                reoccurring within this duration
+
+        Returns:
+            The notice's ID.
+        """
+        body: Dict[str, Any] = {
+            'action': 'add',
+            'type': type.value,
+            'key': key,
+        }
+        if data is not None:
+            body['data'] = data
+        if repeat_after is not None:
+            body['repeat-after'] = _format_timeout(repeat_after.total_seconds())
+        resp = self._request('POST', '/v1/notices', body=body)
+        return resp['result']['id']
+
     def get_notice(self, id: str) -> Notice:
         """Get details about a single notice by ID.
 
@@ -2760,7 +2787,6 @@ class Client:
         user_id: Optional[int] = None,
         types: Optional[Iterable[Union[NoticeType, str]]] = None,
         keys: Optional[Iterable[str]] = None,
-        after: Optional[datetime.datetime] = None,
     ) -> List[Notice]:
         """Query for notices that match all of the provided filters.
 
@@ -2772,6 +2798,11 @@ class Client:
         user (notices whose ``user_id`` matches the requester UID as well as
         public notices).
 
+        Note that the "after" filter is not yet implemented, as it's not
+        needed right now and it's hard to implement correctly with Python's
+        datetime type, which only has microsecond precision (and Go's Time
+        type has nanosecond precision).
+
         Args:
             select: select which notices to return (instead of returning
                 notices for the current user)
@@ -2779,7 +2810,6 @@ class Client:
                 public notices (only works for Pebble admins)
             types: filter for notices with any of the specified types
             keys: filter for notices with any of the specified keys
-            after: filter for notices that were last repeated after this time
         """
         query: Dict[str, Union[str, List[str]]] = {}
         if select is not None:
@@ -2790,8 +2820,6 @@ class Client:
             query['types'] = [(t.value if isinstance(t, NoticeType) else t) for t in types]
         if keys is not None:
             query['keys'] = list(keys)
-        if after is not None:
-            query['after'] = after.isoformat()
         resp = self._request('GET', '/v1/notices', query)
         return [Notice.from_dict(info) for info in resp['result']]
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2750,11 +2750,11 @@ class Client:
         """Record an occurrence of a notice with the specified options.
 
         Args:
-            type: Notice type (currently only "custom" notices are supported)
-            key: Notice key; must be in "example.com/path" format
-            data: Data fields for this notice
-            repeat_after: Prevent notice with same type and key from
-                repeating within this duration
+            type: Notice type (currently only "custom" notices are supported).
+            key: Notice key; must be in "example.com/path" format.
+            data: Data fields for this notice.
+            repeat_after: Only allow this notice to repeat after this duration
+                has elapsed (the default is to always repeat).
 
         Returns:
             The notice's ID.
@@ -2804,12 +2804,12 @@ class Client:
         type has nanosecond precision).
 
         Args:
-            select: select which notices to return (instead of returning
-                notices for the current user)
-            user_id: filter for notices for the specified user, including
-                public notices (only works for Pebble admins)
-            types: filter for notices with any of the specified types
-            keys: filter for notices with any of the specified keys
+            select: Select which notices to return (instead of returning
+                notices for the current user).
+            user_id: Filter for notices for the specified user, including
+                public notices (only works for Pebble admins).
+            types: Filter for notices with any of the specified types.
+            keys: Filter for notices with any of the specified keys.
         """
         query: Dict[str, Union[str, List[str]]] = {}
         if select is not None:
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2751,7 +2751,7 @@ class Client:
 
         Args:
             type: Notice type (currently only "custom" notices are supported)
-            key: Notice key; must be in "domain.com/path" format
+            key: Notice key; must be in "example.com/path" format
             data: Data fields for this notice
             repeat_after: Prevent notice with same type and key from
                 reoccurring within this duration
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1308,7 +1308,7 @@ class NoticesSelect(enum.Enum):
     """
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class Notice:
     """Information about a single notice."""
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2754,7 +2754,7 @@ class Client:
             key: Notice key; must be in "example.com/path" format
             data: Data fields for this notice
             repeat_after: Prevent notice with same type and key from
-                reoccurring within this duration
+                repeating within this duration
 
         Returns:
             The notice's ID.
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1308,7 +1308,7 @@ class NoticesSelect(enum.Enum):
     """
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass
 class Notice:
     """Information about a single notice."""
 
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -3295,7 +3295,7 @@ class _TestingPebbleClient:
 
         # The shape of the code below is taken from State.AddNotice in Pebble.
         now = datetime.datetime.now(tz=datetime.timezone.utc)
-        uid = 0  # hard-code UID as root (Pebble and charms ran as root for now)
+        uid = 0  # Hard-code UID as root (Pebble and charms run as root for now).
 
         new_or_repeated = False
         unique_key = (uid, type.value, key)
@@ -3306,7 +3306,7 @@ class _TestingPebbleClient:
             notice = pebble.Notice(
                 id=str(self._last_notice_id),
                 user_id=uid,
-                type=pebble.NoticeType(type),
+                type=type,
                 key=key,
                 first_occurred=now,
                 last_occurred=now,
@@ -3365,25 +3365,22 @@ class _TestingPebbleClient:
             filter_user_id = None
 
         if types is not None:
-            types = [(t.value if isinstance(t, pebble.NoticeType) else t) for t in types]
+            types = {(t.value if isinstance(t, pebble.NoticeType) else t) for t in types}
         if keys is not None:
-            keys = list(keys)
+            keys = set(keys)
 
-        notices: List[pebble.Notice] = []
-        for notice in self._notices.values():
-            if not self._notice_matches(notice, filter_user_id, types, keys):
-                continue
-            notices.append(notice)
-
+        notices = [notice for notice in self._notices.values() if
+                   self._notice_matches(notice, filter_user_id, types, keys)]
         notices.sort(key=lambda notice: notice.last_repeated)
         return notices
 
     @staticmethod
     def _notice_matches(notice: pebble.Notice,
                         user_id: Optional[int] = None,
-                        types: Optional[List[str]] = None,
-                        keys: Optional[List[str]] = None) -> bool:
+                        types: Optional[Set[str]] = None,
+                        keys: Optional[Set[str]] = None) -> bool:
         # Same logic as NoticeFilter.matches in Pebble.
+        # For example: if user_id filter is set and it doesn't match, return False.
         if user_id is not None and not (notice.user_id is None or user_id == notice.user_id):
             return False
         if types is not None and notice.type not in types:
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1110,12 +1110,12 @@ class Harness(Generic[CharmType]):
         :class:`ops.PebbleCustomNoticeEvent`.
 
         Args:
-            container_name: Name of workload container
-            key: Notice key; must be in "example.com/path" format
-            data: Data fields for this notice
-            repeat_after: Prevent notice with same type and key from
-                reoccurring within this duration
-            type: Notice type (currently only "custom" notices are supported)
+            container_name: Name of workload container.
+            key: Notice key; must be in "example.com/path" format.
+            data: Data fields for this notice.
+            repeat_after: Only allow this notice to repeat after this duration
+                has elapsed (the default is to always repeat).
+            type: Notice type (currently only "custom" notices are supported).
 
         Returns:
             The notice's ID.
@@ -2755,7 +2755,7 @@ class _TestingPebbleClient:
         self._root = container_root
         self._backend = backend
         self._exec_handlers: Dict[Tuple[str, ...], ExecHandler] = {}
-        self._notices: Dict[Tuple[Optional[int], str, str], pebble.Notice] = {}
+        self._notices: Dict[Tuple[str, str], pebble.Notice] = {}
         self._last_notice_id = 0
 
     def _handle_exec(self, command_prefix: Sequence[str], handler: ExecHandler):
@@ -3295,17 +3295,16 @@ class _TestingPebbleClient:
 
         # The shape of the code below is taken from State.AddNotice in Pebble.
         now = datetime.datetime.now(tz=datetime.timezone.utc)
-        uid = 0  # Hard-code UID as root (Pebble and charms run as root for now).
 
         new_or_repeated = False
-        unique_key = (uid, type.value, key)
+        unique_key = (type.value, key)
         notice = self._notices.get(unique_key)
         if notice is None:
             # First occurrence of this notice uid+type+key
             self._last_notice_id += 1
             notice = pebble.Notice(
                 id=str(self._last_notice_id),
-                user_id=uid,
+                user_id=0,  # Charm should always be able to read pebble_notify notices.
                 type=type,
                 key=key,
                 first_occurred=now,
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -1102,7 +1102,7 @@ class Harness(Generic[CharmType]):
     def pebble_notify(self, container_name: str, key: str, *,
                       data: Optional[Dict[str, str]] = None,
                       repeat_after: Optional[datetime.timedelta] = None,
-                      type: Union[pebble.NoticeType, str] = pebble.NoticeType.CUSTOM) -> str:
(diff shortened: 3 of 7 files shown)


ops/testing.py
```diff
@@ -3002,9 +3036,7 @@ def list_files(self, path: str, *, pattern: Optional[str] = None,
         self._check_absolute_path(path)
         file_path = self._root / path[1:]
         if not file_path.exists():
-            raise pebble.APIError(
-                body={}, code=404, status='Not Found',
-                message=f"stat {path}: no such file or directory")
+            raise self._api_error(404, f"stat {path}: no such file or directory") from None
```
benhoyt, line 3039: Drive-by change: add a helper function to produce these errors.

test/test_real_pebble.py
```diff
@@ -26,32 +41,23 @@
 
 from ops import pebble
 
-from .test_testing import PebbleStorageAPIsTestMixin
+from .test_testing import PebbleNoticesMixin, PebbleStorageAPIsTestMixin
+
+
+def get_socket_path() -> str:
```
benhoyt, line 47: Again, drive-by code consolidation to avoid repeating this logic.

test/test_testing.py
```diff
@@ -3043,19 +3043,12 @@ def on_relation_changed(self, event: ops.RelationEvent):
         self.changes.append(dict(data))
 
 
-class RecordedChange(typing.TypedDict, total=False):
-    name: Required[str]
-    data: typing.Dict[str, typing.Any]
-    relation: str
-    container: typing.Optional[str]
-
-
```
benhoyt, line 3052: Removed as it was getting annoying and I don't think it provided much value.

ops/testing.py
```diff
@@ -3270,6 +3353,41 @@ def get_notices(
...
+        notices.sort(key=lambda notice: notice.last_repeated)
+        return notices
+
+    @staticmethod
+    def _notice_matches(notice: pebble.Notice,
+                        user_id: Optional[int] = None,
+                        types: Optional[List[str]] = None,
+                        keys: Optional[List[str]] = None) -> bool:
+        # Same logic as NoticeFilter.matches in Pebble.
+        if user_id is not None and not (notice.user_id is None or user_id == notice.user_id):
```
tonyandrewmeyer, line 3387: It could be me, but I had to read this a few times to follow it. I assume that since this comes from the Pebble code a bunch of people already read essentially the same line but in Go and had no trouble with it, so maybe just ignore this.

Perhaps a comment would help? Like:

`# If the notice is not either public or has a matching user ID, the notice doesn't match.`

ops/pebble.py
```diff
@@ -2744,6 +2744,33 @@ def get_checks(
...
+               data: Optional[Dict[str, str]] = None,
+               repeat_after: Optional[datetime.timedelta] = None) -> str:
+        """Record an occurrence of a notice with the specified options.
+
+        Args:
+            type: Notice type (currently only "custom" notices are supported)
+            key: Notice key; must be in "example.com/path" format
+            data: Data fields for this notice
+            repeat_after: Prevent notice with same type and key from
+                reoccurring within this duration
```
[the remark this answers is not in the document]
benhoyt: Fixed: "reoccurring" -> "repeating"

ops/pebble.py
```diff
@@ -2744,6 +2744,33 @@ def get_checks(
         resp = self._request('GET', '/v1/checks', query)
         return [CheckInfo.from_dict(info) for info in resp['result']]
 
+    def notify(self, type: NoticeType, key: str, *,
```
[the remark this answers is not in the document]
benhoyt: Yeah, the idea was to allow unknown notice types to be received from a newer Pebble server. However, people can update ops whenever they want (and we'll add new notice types there right away), so restrict to just `NoticeType` on the sending side.

ops/testing.py
```diff
@@ -3249,19 +3271,80 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+
+        Return a tuple of (notice_id, new_or_repeated).
+        """
+        if type != pebble.NoticeType.CUSTOM:
+            message = f'invalid type "{type.value}" (can only add "custom" notices)'
+            raise self._api_error(400, message)
+
+        # The shape of the code below is taken from State.AddNotice in Pebble.
+        now = datetime.datetime.now(tz=datetime.timezone.utc)
+        uid = 0  # hard-code UID as root (Pebble and charms ran as root for now)
```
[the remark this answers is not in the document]
benhoyt: Done.

ops/testing.py
```diff
@@ -3249,19 +3271,80 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+        new_or_repeated = False
+        unique_key = (uid, type.value, key)
+        notice = self._notices.get(unique_key)
+        if notice is None:
+            # First occurrence of this notice uid+type+key
+            self._last_notice_id += 1
+            notice = pebble.Notice(
+                id=str(self._last_notice_id),
+                user_id=uid,
+                type=pebble.NoticeType(type),
```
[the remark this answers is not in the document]
benhoyt: Ah yes, this was a carry-over from when this was a `str` -- removed.

ops/testing.py
```diff
@@ -3270,6 +3353,41 @@ def get_notices(
...
+        if types is not None:
+            types = [(t.value if isinstance(t, pebble.NoticeType) else t) for t in types]
+        if keys is not None:
+            keys = list(keys)
+
+        notices: List[pebble.Notice] = []
+        for notice in self._notices.values():
+            if not self._notice_matches(notice, filter_user_id, types, keys):
+                continue
+            notices.append(notice)
```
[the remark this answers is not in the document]
benhoyt: Good call! No reason except copying from the explicit loop in the Go code. Much tidier to use a list comp here -- updated.

ops/testing.py
```diff
@@ -3270,6 +3353,41 @@ def get_notices(
...
+
+        notices.sort(key=lambda notice: notice.last_repeated)
```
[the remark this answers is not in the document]
benhoyt: I always used to recommend `operator` too, but these days I prefer the lambda -- it's one less thing to learn / import, and in this age of static typing it's less "stringly typed" (not that Pyright seems to mind either way -- I guess sort/key isn't typed very accurately anyway).

In addition, we have several instances of `key=lambda...` in the codebase but none of `key=operator...`, so there's precedent for the lambda style.

(I also did a [timeit test](https://gist.github.com/benhoyt/e45ae21a65072081143921e9ffb93dfe) of sorting 1000 randomly-shuffled notices with different `last_repeated` times, and couldn't tell a noticeable difference between the two. Not that performance should be a concern here.)

test/test_testing.py
```diff
@@ -5485,3 +5488,199 @@ def test_bad_results(self):
...
+        self.assertIsInstance(id1a, str)
+        self.assertNotEqual(id1a, '')
+        self.assertEqual(id1a, id1b)
+
+        self.assertIsInstance(id2, str)
+        self.assertNotEqual(id2, '')
+        self.assertNotEqual(id2, id1a)
+
+        self.assertIsInstance(id3, str)
+        self.assertNotEqual(id3, '')
```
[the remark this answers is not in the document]
benhoyt: Good call -- done.

ops/testing.py
```diff
@@ -3270,6 +3353,41 @@ def get_notices(
...
+            filter_user_id = user_id
+        if select is not None:
+            if user_id is not None:
+                raise self._api_error(400, 'cannot use both "select" and "user_id"')
+            filter_user_id = None
+
+        if types is not None:
+            types = [(t.value if isinstance(t, pebble.NoticeType) else t) for t in types]
+        if keys is not None:
+            keys = list(keys)
```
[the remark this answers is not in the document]
benhoyt: Good call -- changed to set comprehension / set. I think we still need the "filter not set" vs "empty container" distinction though (and it matches the Go code).

test/test_testing.py
```diff
@@ -5485,3 +5488,199 @@ def test_bad_results(self):
...
+            'name': 'pebble-custom-notice',
+            'container': 'bar',
+            'notice_id': id3,
+            'notice_type': 'custom',
+            'notice_key': 'example.com/n1',
+        }]
+        self.assertEqual(harness.charm.changes, expected_changes)
+
+    def test_notify_no_repeat(self):
+        # Ensure event doesn't get triggered when notice occurs but doesn't repeat.
```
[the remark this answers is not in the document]
benhoyt: Done

test/test_testing.py
```diff
@@ -5485,3 +5488,199 @@ def test_bad_results(self):
...
+    assertRaises = unittest.TestCase.assertRaises  # noqa
+    assertGreaterEqual = unittest.TestCase.assertGreaterEqual  # noqa
+
+    def test_get_notice_by_id(self):
+        client = self.client
+        key1 = 'example.com/' + os.urandom(16).hex()
+        key2 = 'example.com/' + os.urandom(16).hex()
+        id1 = client.notify(pebble.NoticeType.CUSTOM, key1)
+        id2 = client.notify(pebble.NoticeType.CUSTOM, key2, data={'x': 'y'})
+        time.sleep(0.01)
```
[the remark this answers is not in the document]
benhoyt: Yeah, we can't mock here as this is needed for real Pebble too. Fair call to change it to 1us -- done.

ops/testing.py
benhoyt: Added a comment as to the logic (for this and the others).

tonyandrewmeyer (approved):
Looks great!

ops/pebble.py
```diff
@@ -2772,14 +2798,18 @@ def get_notices(
         user (notices whose ``user_id`` matches the requester UID as well as
         public notices).
 
+        Note that the "after" filter is not yet implemented, as it's not
```
jameinel, line 2801: rounding up to the ms seems like a reasonable way to still get the behavior you want out of a "give me all the notices since X", but I suppose the concern is that you could have a notice happen 1ns after the threshold and thus wasnt sent the first time and get skipped the second.
But if it was ms, you could definitely have 2 events in the same ms and so the threshold needs to be >=.
That said, even ns isn't that good of a guarantee of never having more than one. (and even if the resolution of the value is ns, your clock may only have ms resolution)
if you make the rule >=, then you get overlapping values, so you have to do some local filtering, but you guarantee to not miss anything.
if you want this to be strictly >, then it needs to be an opaque token that you get that hand back again (and maybe you have a separate helper to convert to a datetime)

ops/testing.py
```diff
@@ -3249,19 +3271,80 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+
+        Return a tuple of (notice_id, new_or_repeated).
+        """
+        if type != pebble.NoticeType.CUSTOM:
+            message = f'invalid type "{type.value}" (can only add "custom" notices)'
+            raise self._api_error(400, message)
+
+        # The shape of the code below is taken from State.AddNotice in Pebble.
+        now = datetime.datetime.now(tz=datetime.timezone.utc)
+        uid = 0  # Hard-code UID as root (Pebble and charms run as root for now).
```
jameinel, line 3298: this will be changing in 3.4 along with notices being added.
do we need to expose a value that can be passed for testing?

ops/testing.py
```diff
@@ -3249,19 +3271,80 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+            # First occurrence of this notice uid+type+key
+            self._last_notice_id += 1
+            notice = pebble.Notice(
+                id=str(self._last_notice_id),
+                user_id=uid,
+                type=type,
+                key=key,
+                first_occurred=now,
+                last_occurred=now,
+                last_repeated=now,
```
jameinel, line 3313: I can't say I quite understand last repeated vs last occured if last repeated gets set by the first event.

ops/testing.py
```diff
@@ -3249,19 +3271,80 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+                expire_after=datetime.timedelta(days=7),
+                occurrences=1,
+            )
+            self._notices[unique_key] = notice
+            new_or_repeated = True
+        else:
+            # Additional occurrence, update existing notice
+            notice.occurrences += 1
```
jameinel, line 3321: i suppose these might be why you didnt want to freeze notices.
and certainly the test side gets harder, but we did similar work elsewhere to, eg make Unit unmodifyable and Harness does the work to wrap that.
You certainly can use a frozen, use temp variables here and populate a new frozen notice in your dict when you are done.
Or use a non frozen notice-like, that gets cast to frozen so that *charm* code assumes they are frozen.
[knowing that it is safe to use this blob of data because nobody can modify it in your cache/behind your back is a useful property]
and we shouldnt be returning non frozen types from our dict without copying them anyway, because otherwise the charm under test might modify it, and then our dict was altered.

ops/testing.py
```diff
@@ -3249,19 +3271,80 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+            if repeat_after is None or now > notice.last_repeated + repeat_after:
+                # Update last repeated time if repeat-after time has elapsed (or is None)
+                notice.last_repeated = now
+                new_or_repeated = True
+
+        notice.last_occurred = now
+        notice.last_data = data or {}
+        notice.repeat_after = repeat_after
```
jameinel, line 3329: I guess this explains last repeated.
If you say repeat_after=10min, then occurred will keep ticking up during that time, but repeated won't until the 10min has passed

ops/testing.py
```diff
@@ -3270,6 +3353,38 @@ def get_notices(
...
+            if user_id is not None:
+                raise self._api_error(400, 'cannot use both "select" and "user_id"')
+            filter_user_id = None
+
+        if types is not None:
+            types = {(t.value if isinstance(t, pebble.NoticeType) else t) for t in types}
+        if keys is not None:
+            keys = set(keys)
+
+        notices = [notice for notice in self._notices.values() if
```
jameinel, line 3372: it seems like a dict on (uid, type, key) is not a great structure for notices given how you are querying them

I suppose dicts are Ordered, and thus you default to first creation ordering, which will be close to sorted.
I dont know if it is useful to leave a note as to why a dict is "good enough", but it was defintely something I noted in the attribute type definitions, and wanted to bring up.

I guess the idea is that count will always be small (especially while testing) and when it isnt you could use a real db and do queries.

How is it implemented in Pebble? Do notices persist across restarts? Or are they only ephemeral? If you get 5000 notices, does get_notices still have to read all of them to apply a filter, or is there some filtering that is done by indexes/storage patterns?

test/test_testing.py
```diff
@@ -5485,3 +5488,199 @@ def test_bad_results(self):
...
+        self.assertNotEqual(id1a, '')
+        self.assertEqual(id1a, id1b)
+
+        self.assertIsInstance(id2, str)
+        self.assertNotEqual(id2, '')
+        self.assertNotEqual(id2, id1a)
+
+        self.assertIsInstance(id3, str)
+        self.assertNotEqual(id3, '')
+        self.assertNotEqual(id3, id2)
```
jameinel, line 5523: is it woeth having an id1c after you have done other events, and show that it isnt just deduping if the event is the previous event?
either that or move the id1b to after 2 and 3 to get that effect.

test/test_testing.py
```diff
@@ -5485,3 +5488,199 @@ def test_bad_results(self):
...
+    assertRaises = unittest.TestCase.assertRaises  # noqa
+    assertGreaterEqual = unittest.TestCase.assertGreaterEqual  # noqa
+
+    def test_get_notice_by_id(self):
+        client = self.client
+        key1 = 'example.com/' + os.urandom(16).hex()
+        key2 = 'example.com/' + os.urandom(16).hex()
+        id1 = client.notify(pebble.NoticeType.CUSTOM, key1)
+        id2 = client.notify(pebble.NoticeType.CUSTOM, key2, data={'x': 'y'})
+        time.sleep(0.000_001)  # Ensure times are different.
```
jameinel, line 5623: it is usually a lot better to have ways of injecting a time source rather than sleeping to get time to change.
probably a fair bit of overhead for this test, but eventually sleep based tests bite you on someone else's hardware acting differently than you expected.

jameinel (commented):
Overall looking quite good. Two main gripes are repeat_after not being very obvious and that we don't have a way to be testing uid support for containers that arent running as root.

ops/pebble.py
```diff
@@ -1308,7 +1308,7 @@ class NoticesSelect(enum.Enum):
     """
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass
```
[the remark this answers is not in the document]
benhoyt: Yeah, this is good push-back. I changed it because of the implementation of `pebble_notify` / `_notify`, which corresponded directly to the Go version, which mutated the `*Notice`. However, I think you're right that it's better to make the implementation slightly trickier and keep the frozen=True as a good signal. I've switched this back to frozen, and updated `_notify` to use `dataclasses.replace()` instead.

ops/pebble.py
```diff
@@ -2744,6 +2744,33 @@ def get_checks(
...
 
+    def notify(self, type: NoticeType, key: str, *,
+               data: Optional[Dict[str, str]] = None,
+               repeat_after: Optional[datetime.timedelta] = None) -> str:
+        """Record an occurrence of a notice with the specified options.
+
+        Args:
+            type: Notice type (currently only "custom" notices are supported)
+            key: Notice key; must be in "example.com/path" format
+            data: Data fields for this notice
```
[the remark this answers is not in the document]
benhoyt: I'm not sure what you're saying here: this is the docstring for `notify`, which creates a notice. The one-liner comment at the top is "Record an occurrence of a notice with the specified options." so it seems clear to me that these args are referring to those options -- the type/key/data for this notice you're recording.

ops/pebble.py
```diff
@@ -2744,6 +2744,33 @@ def get_checks(
...
+            repeat_after: Prevent notice with same type and key from
+                repeating within this duration
```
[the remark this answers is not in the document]
benhoyt: Yeah, I prefer the positive wording too. I've updated it here (I've made a local edit to the Pebble `--repeat-after` argument help too that I'll commit later).

ops/pebble.py
benhoyt: For better or worse, this is what's implemented in Pebble (with nanoseconds). However, it is safe if you specify `after=<string of last value you saw>`, at least assuming that two separate notices can't be recorded in the same nanosecond -- I safe bet for a long time, I suspect. :-)

In any case, this is why I decided to leave this out for now, as it's tricky to get right. We'd have to allow an option to not convert the `last-repeated` JSON field to a datetime, keep it as a string, and pass that directly through. I can think of a path to get there, but it seems too complex for something we'll probably never use in a charm.

ops/testing.py
benhoyt: Good call. This isn't actually used (we don't do uid authorization here). So I've instead removed it from the `unique_key` as we don't need that, and it's only used to set the `user_id` field in the `pebble.Notice`. I've added a comment there.

ops/testing.py
benhoyt: Yeah, the term "repeat" is a bit confusing. Be bikeshedded it quite a bit but ended up settling on that. It looks like you've figured this out below though.

ops/testing.py
benhoyt: Yep, see above -- I've changed this to reinstate the frozen=True, and I'm using `dataclasses.replace()` here to build a new notice.

ops/testing.py
benhoyt: We discussed this in the original design: there are likely to be only a small number of unique notices, so we figured just iterating through the dumb data structure is okay for now. The dict is just for uniquification, not indexing. So this Python code mimics the Go code. If and when we want to improve that in Pebble itself, we can (though we still probably don't need it here for the test harness).

test/test_testing.py
benhoyt: Moved -- thanks.

test/test_testing.py
benhoyt: Agreed. I would have patched/injected, but these tests are used against "real Pebble" as well (via the mixin). So we need real time for that (unless we want to inject something in Pebble, but that gets *really* painful).

ops/model.py
```diff
@@ -2744,7 +2744,6 @@ def get_notices(
         user_id: Optional[int] = None,
         types: Optional[Iterable[Union[pebble.NoticeType, str]]] = None,
         keys: Optional[Iterable[str]] = None,
-        after: Optional[datetime.datetime] = None,
```
jameinel, line 2747: Thinking about it, you could just have some opaque `Notice.id` and then take `after=notice.id`. That field could be taken from last_repeated but is intentionally left opaque.

ops/pebble.py
jameinel: I agree that I don't think we'll ever generate 2 notices in the same nanosecond. My concern is more that while the representation supports nanoseconds, I don't think the clock gives that kind of resolution.
(famously the [time.clock on windows](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.now?view=net-8.0) was actually a 60Hz clock, and you had to use the "high performance" timers to even get ms resolution. Digging for Go it seems you might get 100ns resolution https://stackoverflow.com/questions/14610459/how-precise-is-gos-time-really
but it is platform dependent and on BSD was 1ms for a long time)

Just to say, having an opaque key is good, trusting your clock to give you unique values isnt.
In Pebble you could have a check if the timestamp is already cached you just add 1ns to the value to make it unique and it becomes a sort order.

mgo assumed 1ms resolution for its ObjectId and kept 8 bits for a counter (so up to 256 events per ms could be disambiguated)
Go went Nanoseconds because it makes everything straightforward and maybe we'll have clocks close to that (MS now claims 100ns)
I think designing a system assuming a millisecond clock will keep you robust.

ops/testing.py
```diff
@@ -3249,19 +3271,86 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
+            self._notices[unique_key] = notice
+            new_or_repeated = True
+        else:
+            # Additional occurrence, update existing notice
+            last_repeated = notice.last_repeated
+            if repeat_after is None or now > notice.last_repeated + repeat_after:
+                # Update last repeated time if repeat-after time has elapsed (or is None)
+                last_repeated = now
+                new_or_repeated = True
+            notice = dataclasses.replace(
```
jameinel, line 3327: nice!

jameinel (approved):
I have some comments about how we might design after but I'm happy with the implementation and you pulled out after for now anyway.

ops/pebble.py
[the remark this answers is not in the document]
jameinel: I believe I had interpreted this as "give me a notice that is matching these keys" not "create a notice with this content". I think my comment is invalid.

PR #1101, opened 2024-01-08
ci: standardize quotes in source code on ' and """

### LINKED ISSUE #1099: Add linter rule for consistent quoting
Per [this comment](https://github.com/canonical/operator/pull/1091#discussion_r1439964012), we'd like to add a flake8 quoting rule to make our quoting style consistent. We mostly use single quotes, but we have some double quotes, so we'll need to make them consistent first.

benhoyt:
DRAFT: this isn't finished yet, but I've spent too much time on it already.

I had almost finished an ad-hoc "just enough tokenizer" that did strings, and then I ran into the f-string problem. Then I discovered that the [`tokenize`](https://docs.python.org/3/library/tokenize.html) module does support round-tripping whitespace (by the use of start/end line/col markings), so I wrote the script in [this commit message](https://github.com/canonical/operator/commit/6847b8e3838053ae2f2ddebcfb093031359da574) that uses that.

That converted the non-f-string strings, and then I was trying to go through manually with search-and-replace to do the f-strings (I'm sure we could automate that part too, but it's not trivial). And then I realized my first pass hadn't done a number of the non-f-strings, e.g., in `test/test_testing.py` -- I'm not sure why.

In any case, worth pursuing at some point, but I'm going to drop it for now and get back to useful work.

Fixes #1099

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -607,9 +607,9 @@ class StorageEvent(HookEvent):
         """
         snapshot: Dict[str, Any] = {}
         if isinstance(self.storage, model.Storage):
-            snapshot["storage_name"] = self.storage.name
-            snapshot["storage_index"] = self.storage.index
-            snapshot["storage_location"] = str(self.storage.location)
+            snapshot['storage_name'] = self.storage.name
+            snapshot['storage_index'] = self.storage.index
+            snapshot['storage_location'] = str(self.storage.location)
         return snapshot
 
     def restore(self, snapshot: Dict[str, Any]):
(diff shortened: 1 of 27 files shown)


benhoyt:
Jon reminded me that Ruff now has a code formatter. It's basically intended as Black-compatible, with a few minor consistency improvements (and a knob for single quotes! :-). https://docs.astral.sh/ruff/formatter/

As much as I don't like the output of Black in many cases, I think it's probably best to move this project towards auto-formatting, and clean up what we can.

So I think we should probably bite the bullet and switch to using that (once we're on Ruff for flake8 linting stuff). I'm going to close this PR along those lines -- will avoid further wheel-churning on rolling our own quote-converter, too.

PR #1104, opened 2024-01-09
fix: revert macaroonbakery exception now that underlying issue is fixed

benhoyt:
https://github.com/go-macaroon-bakery/py-macaroon-bakery/issues/94 is fixed now and version 1.3.4 of that project has been released, so this should not be needed any more.


tonyandrewmeyer (approved):
Not sure if you were going to bother with any reviews given the nature of this, but +1 from me anyway :)

PR #1106, opened 2024-01-10
feat: add support for v2 fields in CharmMeta

### LINKED ISSUE #498: CharmMeta class needs to be updated for Metadata v2
The `CharmMeta` class currently only represents the v1 Metadata spec - it needs to be updated to support metadata v2: https://discourse.charmhub.io/t/charm-metadata-v2/3674

tonyandrewmeyer:
Extends the `CharmMeta` and related objects to handle all fields in the [current specification](https://juju.is/docs/sdk/metadata-yaml), both loading from `metadata.yaml` and from [`charmcraft.yaml`](https://juju.is/docs/sdk/charmcraft-yaml) (which is likely to happen when running unit tests).

New fields:
* `links`:
  * `websites`
  * `sources`
  * `issues`
  * `documentation` (this is "docs" in metadata.yaml, but "documentation" in charmcraft.yaml, and is a single string unlike the above, so "docs" seems misleading)
* `assumes`

`metadata.yaml` allows arbitrary fields, so leaving in fields that are not in the current specification (e.g. "tags", "series") seems ok and provides the most backwards compatibility.

Also fleshes out `ContainerMeta` and adds `properties` to `StorageMeta`.

Fixes #498.


tonyandrewmeyer:
@benhoyt this needs the new classes added to `__init__.py`, tests, and a few other bits of cleanup (including whatever is failing in the CI, likely formatting). Before I do those, could you quickly check that this is more-or-less what you intended for #498? No need for a full review - that can be done when I've finished everything.

ops/charm.py
```diff
@@ -1488,6 +1549,66 @@ def __init__(self, name: str, raw: Dict[str, Any]):
...
 
 
+class DeviceType(enum.Enum):
+    """Types of device that may be requested."""
+    GPU = 'gpu'
+    NVIDIA_GPU = 'nvidia.com/gpu'
+    AMD_CPU = 'amd.com/gpu'
+
+
+class DeviceMeta:
```
benhoyt, line 1559: For the new types, I wonder if we should go all in on dataclasses (probably frozen), and add a `from_dict` method to parse from the dict. Kind of like how I did with `pebble.Notice` recently?

benhoyt (commented):
Yep, this looks like the basic idea, thanks! I haven't gone over in detail.

What spec were you working against?

tonyandrewmeyer:
> Yep, this looks like the basic idea, thanks! I haven't gone over in detail.

Thanks! Just needed a santy check :smile:. You can go over in detail when I finish it off :smile:.

> What spec were you working against?

The one linked in the description: https://juju.is/docs/sdk/metadata-yaml - the Discourse post linked from the issue says that it should be used rather than the Discourse post.

ops/charm.py
tonyandrewmeyer: @benhoyt further to this, what do you think about making the existing ones dataclasses as well, for more consistency? We can't really change the `__init__` signature (and so would need to also add custom `__repr__`) but then they would all have `__repr__`, `__eq__`, and (in 3.10+) `__match_args__`.

The older ones would have to not be frozen also, so that plus the inconsistent creation is not ideal, but it's still more consistent than having just having the most recently added types be dataclasses.

benhoyt: Per discussion, yeah, it's not ideal -- but let's do the new types as dataclasses, but leave the other ones for now. If people are clamouring for them (I doubt it) we can always update those in a separate PR.

ops/charm.py
```diff
@@ -1297,16 +1338,82 @@ def __init__(self, raw: Optional[Dict[str, Any]] = None,
...
+            # When running tests, this might be loading from charmcraft.yaml.
+            documentation = raw.get('links', {}).get('documentation')
+        self.links = MetadataLinks(
+            websites=websites,
+            sources=sources,
+            issues=issues,
+            documentation=documentation,
+        )
+
+    def _assumes_min_juju_version(self) -> Optional[str]:
```
tonyandrewmeyer, line 1383: @benhoyt what this does is provide a compatibility layer so that if you have `assumes: xxx` in your yaml, your charm can still use `meta.min_juju_version` and it'll pick up any minimum version specified in the `assumes` block.

This code works in my quick manual testing, but I'd need to write a bunch of tests to properly validate it, check edge cases, etc.

I'm wondering now if it's actually worth doing this. Maybe in this situation the answer is just for the charmer to adjust their code to use `meta.assumes` instead? They would then need to have code similar to this themselves, but it might be that they need to actually use the maximum or other things rather than just the minimum, and maybe if that's common then down the track it would be better to have some helpers in ops.jujuversion or something similar.

Thoughts?

benhoyt: Yeah, it's a fair bit of new code/logic. I think we should keep the metadata dumb. If the charmer really needs to do this they can, but seems unlikely they'll need to in 99% of cases.

ops/charm.py
```diff
@@ -1588,17 +1796,17 @@ def locations(self) -> List[str]:
...
-            else:
-                raise RuntimeError(
-                    "container has more than one mount point with the same backing storage. "
-                    "Request .locations to see a list"
-                )
-        else:
-            raise AttributeError(
-                f"{self.__class__.__name__} has no such attribute: {name}!"
-            )
+    @property
```
tonyandrewmeyer, line 1799: This is a backwards-incompatible change in that you can no longer assign to `.location`, but it seems like that would surely always be a bug anyway, since it wouldn't actually change the location, just make the metadata incorrect.

tonyandrewmeyer:
FYI, I did a basic sanity check across the ~150 charms I have locally, and loading the metadata doesn't crash:

```python
import pathlib

import ops

root = pathlib.Path(...)

for charm in root.iterdir():
    meta = charm / 'metadata.yaml'
    if meta.exists():
        assert ops.CharmMeta.from_yaml(meta.open()).name
```

ops/charm.py
```diff
@@ -671,8 +683,7 @@ class WorkloadEvent(HookEvent):
     """Base class representing workload-related events.
 
     Workload events are generated for all containers that the charm
-    expects in metadata. Workload containers currently only trigger
```
tonyandrewmeyer, line 674: Minor drive-by.

ops/charm.py
```diff
@@ -1273,16 +1294,33 @@ def __init__(self, raw: Optional[Dict[str, Any]] = None,
...
+        # Note that metadata v2 does not define series.
         self.series = raw_.get('series', [])
         self.subordinate = raw_.get('subordinate', False)
+        self.assumes = JujuAssumes.from_list(raw_.get('assumes', []))
+        # Note that metadata v2 does not define min-juju-version ('assumes'
+        # should be used instead).
         self.min_juju_version = raw_.get('min-juju-version')
+        if self.min_juju_version:
+            # Add in an implied 'assumes'.
+            self.assumes.features.append(f'juju >= {self.min_juju_version}')
```
benhoyt, line 1323: Are we sure we want to add this in? Why? It would seem better to me to have this reflect the metadata as is, rather than try to be smart about things.

ops/charm.py
```diff
@@ -1297,16 +1335,48 @@ def __init__(self, raw: Optional[Dict[str, Any]] = None,
...
-        # This is taken from Charm Metadata v2, but only the "containers" and
-        # "containers.name" fields that we need right now for Pebble. See:
-        # https://discourse.charmhub.io/t/charm-metadata-v2/3674
         self.containers = {name: ContainerMeta(name, container)
                            for name, container in raw_.get('containers', {}).items()}
 
+    def _load_links(self, raw: Dict[str, Any]):
+        websites = raw.get('website', [])
+        if not websites and 'links' in raw:
+            # When running tests, this might be loading from charmcraft.yaml.
```
benhoyt, line 1350: What about just having this comment once at the top of the function?

ops/charm.py
```diff
@@ -1488,6 +1562,97 @@ def __init__(self, name: str, raw: Dict[str, Any]):
         self.type = raw['type']
 
 
+class DeviceType(enum.Enum):
+    """Types of device that may be requested."""
+    GPU = 'gpu'
+    NVIDIA_GPU = 'nvidia.com/gpu'
+    AMD_GPU = 'amd.com/gpu'
```
benhoyt, line 1569: I wonder if we should make this just a str rather than an enum with defined values? These seem pretty specific, and we don't necessarily want to have to update this if stuff is added to the list. We could do the `str | DeviceType` trick as I've done elsewhere, but that's kind of annoying.

ops/charm.py
```diff
@@ -1501,23 +1666,69 @@ def __init__(self, name: str, raw: Optional[Dict[str, Any]] = None):
...
     def __init__(self, name: str, raw: Dict[str, Any]):
         self.name = name
         self._mounts: Dict[str, ContainerStorageMeta] = {}
+        self.bases = None
+        self.resource = None
 
         # This is not guaranteed to be populated/is not enforced yet
         if raw:
             self._populate_mounts(raw.get('mounts', []))
+            self.resource = raw.get("resource")
```
benhoyt, line 1727:
```suggestion
            self.resource = raw.get('resource')
```

ops/charm.py
benhoyt: Excellent, thanks for this fix.

test/test_charm.py
```diff
@@ -561,18 +561,52 @@ def test_containers_storage(self):
...
+        self.assertEqual(meta.containers['test1'].resource, 'ubuntu-22.10')
+        assert meta.containers['test2'].bases is not None
+        self.assertEqual(len(meta.containers['test2'].bases), 2)
+        self.assertEqual(meta.containers['test2'].bases[0].os_name, 'ubuntu')
+        self.assertEqual(meta.containers['test2'].bases[0].channel, '23.10')
+        self.assertEqual(meta.containers['test2'].bases[0].architectures, ['amd64'])
+        self.assertEqual(meta.containers['test2'].bases[1].os_name, 'ubuntu')
+        self.assertEqual(meta.containers['test2'].bases[1].channel, '23.04/stable/fips')
+        self.assertEqual(meta.containers['test2'].bases[1].architectures, ['arm'])
+        with self.assertRaises(ModelError):
```
benhoyt, line 599: Can you please add a comment above the `with` line about why this is failing? It's not immediately obvious to me.

benhoyt (changes requested):
Looking good, just some small questions and minor nits.

ops/charm.py
tonyandrewmeyer: Sure. I did cringe a bit at it having nvidia & amd in there rather than nice generic things like GPU, NFC, BLE etc :joy:.

ops/charm.py
tonyandrewmeyer: Good call, will do.

ops/charm.py
tonyandrewmeyer: I definitely agree that the code I had earlier that tried to be smart in the other direction (populating `min_juju_version` from `assumes`) was a bad idea (on the positive side, it helped me really dig into how `assumes` works, so I could be more sure I had the class correct :smile:).

This direction (populating `assumes` from `min_juju_version`) is vastly simpler, just these 2-3 lines, and more obviously correct.

My thinking was that if someone is using this in code, and they want to support both versions of metadata (e.g. in a lib where you don't have as much control over the metadata), they have to do something like:

```python
min_version = meta.min_juju_version or find_minimum_juju_version(meta.assumes)
```

instead of just `find_minimum_juju_version(meta.assumes)`.

However, writing this out I think I have rubber-ducked my way into agreeing that it's too small to be worthwhile :smile:.

benhoyt (approved):
Looks good to me, thanks!

ops/charm.py
```diff
@@ -1488,6 +1560,90 @@ def __init__(self, name: str, raw: Dict[str, Any]):
         self.type = raw['type']
 
 
+@dataclasses.dataclass(frozen=True)
+class DeviceMeta:
```
wallyworld, line 1564: Do we have any charms that use `Device`? This was something added way back in the day for the older style charms. If there's no sidecar charms using this, I would recommend not implementing it until it's asked for. That way we don't lock ourselves into supporting something that turns out to be "wrong".

ops/charm.py
```diff
@@ -1431,6 +1499,9 @@ class StorageMeta:
     multiple_range: Optional[Tuple[int, Optional[int]]]
     """Range of numeric qualifiers when multiple storage units are used."""
 
+    properties = List[Literal['transient']]
```
wallyworld, line 1502: "transient" isn't the only value for properties

wallyworld (commented):
The `assumes` stuff looks good (consistent with how juju parses it).
juju reads bases from manifest.yaml and what's here looks consistent with that.

Can we defer adding Devices until we are sure it's actually used somewhere?
Otherwise looks ok to me.

ops/charm.py
tonyandrewmeyer: Ah, it was the only one in the docs. Is there a list, or should I just make this an open list of strings?

ops/charm.py
tonyandrewmeyer: In the ~150 I have a list of, there are none. @benhoyt given that this is an 'old' thing not a 'new' one, I'm happy to remove it - fine with you?

ops/charm.py
benhoyt: @wallyworld It looks like it's the only one [accepted by the `juju/charm` package](https://github.com/juju/charm/blob/f5d79a50fdf959cd340a853aa65c6af5f06a21b3/meta.go#L1369), too. Regardless, I think it's probably more flexible and future-proof to just make our type annotations a `List[str]`.

ops/charm.py
benhoyt: Yeah, Ian's rationale makes sense -- let's remove it from the PR.

PR #1107, opened 2024-01-10
feat: model error on testing module if invalid status set by charm

yanksyoon:
This small enhancement:
1. documents that `ErrorStatus` and `UnknownStatus` is for read-only.
2. raises an ops.ModelError, which is the same behavior on real juju backend.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1790,6 +1790,14 @@ class UnknownStatus(StatusBase):
     A unit-agent has finished calling install, config-changed and start, but the
     charm has not called status-set yet.
 
+    This status is for READ ONLY and should not be used to set a unit status. The following
+    charm code would be an invalid charm code.
+
+        ```
+        # Raises ops.model.ModelError : ERROR invalid status "unknown", expected one of
+        # [maintenance blocked waiting active]
+        self.unit.status = UnknownStatus()
+        ```
     """
     name = 'unknown'
 
@@ -1807,6 +1815,15 @@ class ErrorStatus(StatusBase):
 
     The unit-agent has encountered an error (the application or unit requires
     human intervention in order to operate correctly).
+
+    This status is for READ ONLY and should not be used to set a unit status. The following
+    charm code would be an invalid charm code.
+
+        ```
+        # Raises ops.model.ModelError : ERROR invalid status "error", expected one of
+        # [maintenance blocked waiting active]
+        self.unit.status = ErrorStatus()
+        ```
     """
     name = 'error'
 
--- a/ops/model.py
+++ b/ops/model.py
@@ -1790,14 +1790,8 @@ class UnknownStatus(StatusBase):
     A unit-agent has finished calling install, config-changed and start, but the
     charm has not called status-set yet.
 
-    This status is for READ ONLY and should not be used to set a unit status. The following
-    charm code would be an invalid charm code.
-
-        ```
-        # Raises ops.model.ModelError : ERROR invalid status "unknown", expected one of
-        # [maintenance blocked waiting active]
-        self.unit.status = UnknownStatus()
-        ```
+    This status is read-only; trying to set unit or application status to
+    ``UnknownStatus`` will raise :class:`ModelError`.
     """
     name = 'unknown'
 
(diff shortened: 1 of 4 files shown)


ops/model.py
```diff
@@ -1790,6 +1790,14 @@ class UnknownStatus(StatusBase):
...
     charm has not called status-set yet.
 
+    This status is for READ ONLY and should not be used to set a unit status. The following
+    charm code would be an invalid charm code.
+
+        ```
+        # Raises ops.model.ModelError : ERROR invalid status "unknown", expected one of
+        # [maintenance blocked waiting active]
+        self.unit.status = UnknownStatus()
+        ```
```
benhoyt, line 1800: I think this is probably saying too much, and we don't need an example of what *not* to do. Perhaps just this:
```suggestion
    This status is read-only; trying to set unit or application status to
    ``UnknownStatus`` will raise :class:`ModelError`.
```

ops/model.py
```diff
@@ -1807,6 +1815,15 @@ class ErrorStatus(StatusBase):
 
     The unit-agent has encountered an error (the application or unit requires
     human intervention in order to operate correctly).
+
```
benhoyt, line 1818: Same as above.

ops/testing.py
```diff
@@ -447,7 +447,7 @@ def begin_with_initial_hooks(self) -> None:
         # the unit status from "Maintenance" to "Unknown". See gh#726
         post_setup_sts = self._backend.status_get()
         if post_setup_sts.get("status") == "maintenance" and not post_setup_sts.get("message"):
-            self._backend.status_set("unknown", "", is_app=False)
+            self._backend.status_set("unknown", "", is_app=False, by_testing_backend=True)
```
benhoyt, line 450: Instead of adding this new argument, I'd prefer we have have an internal `_status_set()` helper that does it without the check. However, in this case it's trivial, so I think we should just do this:

```suggestion
            self._backend._unit_status = {'status': 'unknown', 'message': ''}
```

ops/testing.py
```diff
@@ -2254,7 +2254,17 @@ def status_get(self, *, is_app: bool = False):
...
-    def status_set(self, status: '_StatusName', message: str = '', *, is_app: bool = False):
+    def status_set(
+        self,
+        status: "_StatusName",
+        message: str = "",
+        *,
+        is_app: bool = False,
+        by_testing_backend: bool = False,
+    ):
+        if not by_testing_backend and status in {model.ErrorStatus.name, model.UnknownStatus.name}:
```
benhoyt, line 2265: Statuses aren't hashable, so using a set here doesn't gain us anything (and sends the wrong signal IMO). Let's prefer a list:

```suggestion
        if status in [model.ErrorStatus.name, model.UnknownStatus.name]:
```

ops/testing.py
```diff
@@ -2254,7 +2254,17 @@ def status_get(self, *, is_app: bool = False):
         else:
             return self._unit_status
 
-    def status_set(self, status: '_StatusName', message: str = '', *, is_app: bool = False):
+    def status_set(
```
benhoyt, line 2257: Per above, let's revert the new argument (and revert the formatting change to avoid additional diff).

test/test_testing.py
```diff
@@ -2851,6 +2851,29 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent):
         self.assertEqual(harness.model.app.status, ops.ActiveStatus('active app'))
         self.assertEqual(harness.model.unit.status, ops.ActiveStatus('active unit'))
 
+    def test_invalid_status_set(self):
```
benhoyt, line 2854: I don't think this test needs to go via collect-status, we can just set status directly. What about this:

```python
    def test_invalid_status_set(self):
        harness = ops.testing.Harness(ops.CharmBase)
        harness.set_leader(True)
        harness.begin()

        with self.assertRaises(ops.model.ModelError):
            harness.model.app.status = ops.UnknownStatus()
        with self.assertRaises(ops.model.ModelError):
            harness.model.app.status = ops.ErrorStatus()
        harness.model.app.status = ops.ActiveStatus()

        with self.assertRaises(ops.model.ModelError):
            harness.model.unit.status = ops.UnknownStatus()
        with self.assertRaises(ops.model.ModelError):
            harness.model.unit.status = ops.ErrorStatus()
        harness.model.unit.status = ops.ActiveStatus()
```

benhoyt (changes requested):
Great, thanks for the PR. A couple of suggestions for simplifying.

ops/model.py
yanksyoon: Simplified, thank you for the suggestion!

ops/testing.py
yanksyoon: Good point, updated thank you!

ops/testing.py
yanksyoon: reverted w/ changes, thank you!

test/test_testing.py
yanksyoon: Much direct, updated thank you!

benhoyt (approved):
Looks good to me, thanks!

benhoyt:
Thanks for the contribution @yanksyoon!

PR #1111, opened 2024-01-18
fix: add pebble log targets and checks to testing plan

PietroPasotti:
Issue: adding a layer with checks or log-targets isn't reflected in the plan you obtain when testing through get_plan()

solution: make it reflect

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -2931,49 +2931,39 @@ class _TestingPebbleClient:
         else:
             self._layers[label] = layer_obj
 
-    def _render_services(self) -> Dict[str, pebble.ServiceDict]:
-        services: Dict[str, pebble.ServiceDict] = {}
+    def _render_services(self) -> Dict[str, pebble.Service]:
+        services: Dict[str, pebble.Service] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, service in layer.services.items():
-                # TODO: (jam) 2021-04-07 have a way to merge existing services
-                services[name] = service.to_dict()
+                # todo: merge existing services https://github.com/canonical/operator/issues/1112
+                services[name] = service
         return services
 
-    def _render_checks(self) -> Dict[str, pebble.CheckDict]:
-        services: Dict[str, pebble.CheckDict] = {}
+    def _render_checks(self) -> Dict[str, pebble.Check]:
+        """The checks in the current plan."""
+        checks: Dict[str, pebble.Check] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, check in layer.checks.items():
-                services[name] = check.to_dict()
-        return services
+                checks[name] = check
+        return checks
 
-    def _render_log_targets(self) -> Dict[str, pebble.LogTargetDict]:
-        services: Dict[str, pebble.LogTargetDict] = {}
+    def _render_log_targets(self) -> Dict[str, pebble.LogTarget]:
+        """The log targets in the current plan."""
+        log_targets: Dict[str, pebble.LogTarget] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, log_target in layer.log_targets.items():
-                services[name] = log_target.to_dict()
-        return services
-
-    def _render_plan(self) -> str:
-        raw_plan = {}
-        services = self._render_services()
-        if services:
-            raw_plan["services"] = services
-
-        checks = self._render_checks()
-        if checks:
-            raw_plan["checks"] = checks
-
-        log_targets = self._render_log_targets()
-        if log_targets:
-            raw_plan["log-targets"] = log_targets
-        return yaml.safe_dump(raw_plan)
+                log_targets[name] = log_target
+        return log_targets
 
     def get_plan(self) -> pebble.Plan:
         self._check_connection()
-        plan = pebble.Plan(self._render_plan())
+        plan = pebble.Plan("")
+        plan.services.update(self._render_services())
+        plan.checks.update(self._render_checks())
+        plan.log_targets.update(self._render_log_targets())
         return plan
 
     def get_services(self, names: Optional[List[str]] = None) -> List[pebble.ServiceInfo]:
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -2936,12 +2936,11 @@ class _TestingPebbleClient:
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, service in layer.services.items():
-                # todo: merge existing services https://github.com/canonical/operator/issues/1112
+                # TODO: merge existing services https://github.com/canonical/operator/issues/1112
                 services[name] = service
         return services
 
     def _render_checks(self) -> Dict[str, pebble.Check]:
-        """The checks in the current plan."""
         checks: Dict[str, pebble.Check] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
@@ -2950,7 +2949,6 @@ class _TestingPebbleClient:
         return checks
 
     def _render_log_targets(self) -> Dict[str, pebble.LogTarget]:
-        """The log targets in the current plan."""
         log_targets: Dict[str, pebble.LogTarget] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
@@ -2960,7 +2958,7 @@ class _TestingPebbleClient:
 
     def get_plan(self) -> pebble.Plan:
         self._check_connection()
-        plan = pebble.Plan("")
+        plan = pebble.Plan('{}')
         plan.services.update(self._render_services())
         plan.checks.update(self._render_checks())
         plan.log_targets.update(self._render_log_targets())
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -2931,23 +2931,49 @@ class _TestingPebbleClient:
         else:
             self._layers[label] = layer_obj
 
-    def _render_services(self) -> Dict[str, pebble.Service]:
-        services: Dict[str, pebble.Service] = {}
+    def _render_services(self) -> Dict[str, pebble.ServiceDict]:
+        services: Dict[str, pebble.ServiceDict] = {}
         for key in sorted(self._layers.keys()):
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
Thanks! My bad for missing log targets in #1074.

ops/testing.py
```diff
@@ -2931,23 +2931,49 @@ def add_layer(
...
+
     def get_plan(self) -> pebble.Plan:
         self._check_connection()
-        plan = pebble.Plan('{}')
-        services = self._render_services()
-        if not services:
-            return plan
-        for name in sorted(services.keys()):
-            plan.services[name] = services[name]
+        plan = pebble.Plan(self._render_plan())
```
benhoyt, line 2976: I don't love the way we're rendering to a YAML string, only to load it back from YAML to objects again. Can we do this more similarly to how it was?

ops/testing.py
```diff
@@ -2931,23 +2931,49 @@ def add_layer(
...
-    def _render_services(self) -> Dict[str, pebble.Service]:
-        services: Dict[str, pebble.Service] = {}
+    def _render_services(self) -> Dict[str, pebble.ServiceDict]:
+        services: Dict[str, pebble.ServiceDict] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, service in layer.services.items():
                 # TODO: (jam) 2021-04-07 have a way to merge existing services
-                services[name] = service
+                services[name] = service.to_dict()
```
benhoyt, line 2940: Hmm, not an issue with this PR, but looking at this code made me realize we're not even attempting to combine layers correctly here. We already have the code to combine a layer ([in `_TestingPebbleClient.add_layer`](https://github.com/canonical/operator/blob/dc47640f8ea915049d9e20a1fe4ea189c96ae8d1/ops/testing.py#L2912-L2929)), we just need to use it here as well.

We also have `pebble.Service._merge`, but we'll need similar `_merge` methods on `pebble.Check` and `pebble.LogTarget` and update appropriately to merge those blocks as well.

I think we can leave it "broken" in this PR (it's already broken now, so this doesn't make it worse), but I've opened [#1112](https://github.com/canonical/operator/issues/1112) to track.

ops/testing.py
```diff
@@ -2931,23 +2931,49 @@ def add_layer(
...
         return services
 
+    def _render_checks(self) -> Dict[str, pebble.CheckDict]:
+        services: Dict[str, pebble.CheckDict] = {}
```
benhoyt, line 2944: This should be named `checks`.

ops/testing.py
```diff
@@ -2931,23 +2931,49 @@ def add_layer(
...
+        for key in sorted(self._layers.keys()):
+            layer = self._layers[key]
+            for name, check in layer.checks.items():
+                services[name] = check.to_dict()
+        return services
+
+    def _render_log_targets(self) -> Dict[str, pebble.LogTargetDict]:
+        services: Dict[str, pebble.LogTargetDict] = {}
```
benhoyt, line 2952: This one should be `log_targets`.

test/test_pebble.py
```diff
@@ -27,8 +27,10 @@
 
 import pytest
 import websocket  # type: ignore
+from testing import Harness
 
-from ops import pebble
+import ops
+from ops import Framework, pebble
```
benhoyt, line 33: This comment will be moot given this will be moving, but our preferred style for almost all imports is "import foo ... foo.bar()". Here we want access to `ops`, `pebble`, and `testing`, so we'd use:

```python
import ops  # for ops.CharmBase, ops.Framework
from ops import pebble, testing  # for pebble.Foo, testing.Harness
```

test/test_pebble.py
```diff
@@ -3599,3 +3601,42 @@ def recv():
     if hasattr(pytest, 'PytestUnhandledThreadExceptionWarning'):
         test_websocket_recv_raises = pytest.mark.filterwarnings(
             'ignore::pytest.PytestUnhandledThreadExceptionWarning')(test_websocket_recv_raises)
+
+
+class MyCharm(ops.CharmBase):
+    def __init__(self, framework: Framework):
+        super().__init__(framework)
```
benhoyt, line 3608: This comment will be moot given this will be moving, but this is essentially a no-op, so you could just use `pass` here, I think. Or, even terse, just avoid the class altogether and use `Harness(ops.CharmBase, ...)` directly -- I think that works.

test/test_pebble.py
```diff
@@ -27,8 +27,10 @@
 
 import pytest
 import websocket  # type: ignore
+from testing import Harness
```
benhoyt, line 30: We're testing `ops/testing.py` functionality here, so this should go in `test_testing.py`, not `test_pebble.py`. This test should live in or alongside the existing [`test_get_pebble_container_plan`](https://github.com/canonical/operator/blob/dc47640f8ea915049d9e20a1fe4ea189c96ae8d1/test/test_testing.py#L2760) test there.

test/test_pebble.py
```diff
@@ -3599,3 +3601,42 @@ def recv():
...
+      http:
+       https://example.com/
+
+    log-targets:
+     baz:
+      override: replace
+      type: loki
+      location: https://example.com:3100/loki/api/v1/push
+    '''
+    h = Harness(MyCharm, meta=yaml.safe_dump(
```
benhoyt, line 3629: Naming nits:

* h -> harness
* c -> container
* c_plan -> plan

benhoyt (changes requested):
A few comments here. Mainly small things, and the test should move.

Additionally, type checking and tests are failing.

ops/testing.py
PietroPasotti: any reason why we don't have `services` as well then? I thought I'd make them private since that wasn't exposed to begin with

PietroPasotti: Changing this and `log_targets` makes `test.test_testing.TestTestingPebbleClient.test_methods_match_pebble_client` fail because the public signature differs from pebble.Client. Do you want me to add these public methods to pebble.Client as well?

ops/testing.py
PietroPasotti: I have a deep feeling of unease towards mutation that was hammered into me at a young age, so I'll roll this back but add a fixme.

ops/testing.py
benhoyt: Sorry if my original comment wasn't clear -- I didn't mean the `_render_checks` method name should change, but that the `services` variable name should be `checks`, because in this method we're building a dict of checks. Similar below for the variable name, `services` -> `log_targets`.

ops/testing.py
```diff
@@ -2937,17 +2937,45 @@ def _render_services(self) -> Dict[str, pebble.Service]:
...
         return services
 
+    def checks(self) -> Dict[str, pebble.Check]:
+        """The checks in the current plan."""
+        services: Dict[str, pebble.Check] = {}
+        for key in sorted(self._layers.keys()):
+            layer = self._layers[key]
+            for name, check in layer.checks.items():
+                services[name] = check
+        return services
```
benhoyt, line 2951:
```suggestion
    def _render_checks(self) -> Dict[str, pebble.Check]:
        checks: Dict[str, pebble.Check] = {}
        for key in sorted(self._layers.keys()):
            layer = self._layers[key]
            for name, check in layer.checks.items():
                checks[name] = check
        return checks
```

ops/testing.py
```diff
@@ -2937,17 +2937,45 @@ def _render_services(self) -> Dict[str, pebble.Service]:
...
+
+    def log_targets(self) -> Dict[str, pebble.LogTarget]:
+        """The log targets in the current plan."""
+        services: Dict[str, pebble.LogTarget] = {}
+        for key in sorted(self._layers.keys()):
+            layer = self._layers[key]
+            for name, log_target in layer.log_targets.items():
+                services[name] = log_target
+        return services
```
benhoyt, line 2960:
```suggestion
    def _render_log_targets(self) -> Dict[str, pebble.LogTarget]:
        log_targets: Dict[str, pebble.LogTarget] = {}
        for key in sorted(self._layers.keys()):
            layer = self._layers[key]
            for name, log_target in layer.log_targets.items():
                log_targets[name] = log_target
        return log_targets
```

ops/testing.py
```diff
@@ -2937,17 +2937,45 @@ def _render_services(self) -> Dict[str, pebble.Service]:
...
+            plan._services = services
+
+        checks = self.checks()
+        if checks:
+            plan._checks = checks
+
+        log_targets = self.log_targets()
+        if log_targets:
+            plan._log_targets = log_targets
+
```
benhoyt, line 2978: It might trigger your mutation spidey sense even more :-), but I think we should both simplify and avoid use of private `Plan` fields here by doing:

```
plan.services.update(self._render_services())
plan.checks.update(self._render_checks())
plan.log_targets.update(self._render_log_targets())
```

test/test_testing.py
```diff
@@ -2798,6 +2798,39 @@ def test_get_pebble_container_plan(self):
         harness_plan = harness.get_container_pebble_plan('foo')
         self.assertEqual(harness_plan.to_yaml(), plan.to_yaml())
 
+    def test_add_layer_with_log_targets_to_plan(self):
+        raw = '''\
```
benhoyt, line 2802:
```suggestion
        layer_yaml = '''
```

benhoyt (changes requested):
Thanks for the updates. Couple of small changes requested, and then we're good to go.

ops/testing.py
```diff
@@ -2937,17 +2937,45 @@ def _render_services(self) -> Dict[str, pebble.Service]:
...
+            for name, log_target in layer.log_targets.items():
+                services[name] = log_target
+        return services
+
     def get_plan(self) -> pebble.Plan:
         self._check_connection()
-        plan = pebble.Plan('{}')
+        plan = pebble.Plan("")
+
+        # FIXME: this horrible mutation
```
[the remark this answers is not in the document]
PietroPasotti: clear! done

ops/testing.py
PietroPasotti: well at least it's terse

PR #1114, opened 2024-01-26
chore: expand ruff rule sets and make corresponding minor adjustments

### LINKED ISSUE #1102: Use Ruff linter on the project
We'd like to switch from flake8 et al to https://docs.astral.sh/ruff/linter/. This issue tracks that work. The [pyproject.toml work](https://github.com/canonical/operator/pull/1068) should be merged first.

tonyandrewmeyer:
This PR enables a few additional `ruff` checks - since they are bundled with `ruff` and `ruff` is extremely fast, they are basically free. I cherry-picked the set that I agree with/like the most - I'm happy to argue for the inclusion of any specific ones where you disagree, or consider [other rules](https://docs.astral.sh/ruff/rules/) if you have preferences.

* pyupgrade (UP) - I love this tool - I find it's a super easy way to adopt new features as you move up your minimum-supported Python version (in using it for a year or so I haven't ever had it suggest something that I didn't like)
* flake8-2020 (YTT)
* bandit (S) - a good tool for avoiding accidentally missing things, in my opinion (marking them with `noqa` shows that thought has gone into it, which has value in itself)
* flake8-bugbear (B)
* flake8-simplify (SIM)
* Ruff specific (RUF) - seems like we should since we're using the tool
* perflint (PERF) - seems to make good suggestions in general

Notes:

Where the linter picks up new issues, these are handled in one of these ways:
* Ignore with a `noqa:` directive if it's a false positive or should otherwise be permanently ignored in that specific case
* Ignore for a file or group of files (the docs and tests have several of these) where it's something we want to pick up in the core code but not everywhere
* Ignore with a note to review later when it's likely that there would be too much additional noise in this PR
* Make the recommended change, when it's small and seems reasonable

Continues from #1102.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,8 +33,8 @@ furo._compute_navigation_tree = _compute_navigation_tree
 
 # Pull in fix from https://github.com/sphinx-doc/sphinx/pull/11222/files to fix
 # "invalid signature for autoattribute ('ops.pebble::ServiceDict.backoff-delay')"
-import re
-import sphinx.ext.autodoc
+import re  # noqa: E402
+import sphinx.ext.autodoc  # noqa: E402
 sphinx.ext.autodoc.py_ext_sig_re = re.compile(
     r'''^ ([\w.]+::)?            # explicit module name
           ([\w.]+\.)?            # module and/or class name(s)
@@ -150,7 +150,7 @@ exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 # Find the current builder
 builder = "dirhtml"
 if '-b' in sys.argv:
-    builder = sys.argv[sys.argv.index('-b')+1]
+    builder = sys.argv[sys.argv.index('-b') + 1]
 
 html_theme = 'furo'
 html_last_updated_fmt = ""
@@ -158,7 +158,7 @@ html_permalinks_icon = "¶"
 html_theme_options = {
     "light_css_variables": {
         "color-sidebar-background-border": "none",
-        "font-stack": "Ubuntu, -apple-system, Segoe UI, Roboto, Oxygen, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif",
+        "font-stack": "Ubuntu, -apple-system, Segoe UI, Roboto, Oxygen, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif",  # noqa: E501
         "font-stack--monospace": "Ubuntu Mono, Consolas, Monaco, Courier, monospace",
         "color-foreground-primary": "#111",
         "color-foreground-secondary": "var(--color-foreground-primary)",
@@ -168,7 +168,7 @@ html_theme_options = {
         "color-brand-primary": "#111",
         "color-brand-content": "#06C",
         # NOTE: this looks horrible -- commented out
-        #"color-api-background": "#cdcdcd",
+        # "color-api-background": "#cdcdcd",
         "color-inline-code-background": "rgba(0,0,0,.03)",
         "color-sidebar-link-text": "#111",
         "color-sidebar-item-background--current": "#ebebeb",
--- a/ops/_private/timeconv.py
+++ b/ops/_private/timeconv.py
@@ -63,7 +63,7 @@ def parse_rfc3339(s: str) -> datetime.datetime:
                              microsecond=microsecond, tzinfo=tz)
 
 
-def parse_duration(s: str) -> datetime.timedelta:  # noqa: C901
+def parse_duration(s: str) -> datetime.timedelta:
     """Parse a formatted Go duration.
 
     This is similar to Go's time.ParseDuration function: it parses the output
--- a/ops/_private/timeconv.py
+++ b/ops/_private/timeconv.py
@@ -63,7 +63,7 @@ def parse_rfc3339(s: str) -> datetime.datetime:
                              microsecond=microsecond, tzinfo=tz)
 
 
-def parse_duration(s: str) -> datetime.timedelta:
+def parse_duration(s: str) -> datetime.timedelta:  # noqa: C901
     """Parse a formatted Go duration.
 
     This is similar to Go's time.ParseDuration function: it parses the output
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -949,12 +949,9 @@ class CollectStatusEvent(EventBase):
             def _on_collect_status(self, event: ops.CollectStatusEvent):
                 if 'port' not in self.model.config:
                     event.add_status(ops.BlockedStatus('please set "port" config'))
-
-    Return:
+                    return
                 event.add_status(ops.ActiveStatus())
-
-    .. # noqa (pydocstyle barfs on the above for unknown reasons I've spent hours on)
-    """
+    """  # noqa: D405, D214, D411, D416  Final return confuses docstyle
 
     def add_status(self, status: model.StatusBase):
         """Add a status for evaluation.
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -41,6 +41,7 @@ from ops.framework import (
     EventSource,
     Framework,
     Handle,
+    LifecycleEvent,
     Object,
     ObjectEvents,
 )
@@ -913,7 +914,7 @@ class SecretExpiredEvent(SecretEvent):
             'this event until you create a new revision.')
 
 
-class CollectStatusEvent(EventBase):
+class CollectStatusEvent(LifecycleEvent):
     """Event triggered at the end of every hook to collect statuses for evaluation.
 
     If the charm wants to provide application or unit status in a consistent
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -18,32 +18,13 @@ import dataclasses
 import enum
 import logging
 import pathlib
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    Literal,
-    Mapping,
-    NoReturn,
-    Optional,
-    TextIO,
-    Tuple,
-    TypedDict,
-    Union,
-    cast,
-)
+from typing import (TYPE_CHECKING, Any, Dict, List, Literal, Mapping, NoReturn,
+                    Optional, TextIO, Tuple, TypedDict, Union, cast)
 
 from ops import model
 from ops._private import yaml
-from ops.framework import (
-    EventBase,
-    EventSource,
-    Framework,
-    Handle,
-    Object,
-    ObjectEvents,
-)
+from ops.framework import (EventBase, EventSource, Framework, Handle, Object,
+                           ObjectEvents)
 
 if TYPE_CHECKING:
     from typing_extensions import Required
@@ -226,9 +207,9 @@ class StartEvent(HookEvent):
 
     This event is triggered immediately after the first
     :class:`ConfigChangedEvent`. Callback methods bound to the event should be
-    used to ensure that the charm’s software is in a running state. Note that
-    the charm’s software should be configured so as to persist in this state
-    through reboots without further intervention on Juju’s part.
+    used to ensure that the charm's software is in a running state. Note that
+    the charm's software should be configured so as to persist in this state
+    through reboots without further intervention on Juju's part.
     """
 
 
@@ -237,8 +218,8 @@ class StopEvent(HookEvent):
 
     This event is triggered when an application's removal is requested
     by the client. The event fires immediately before the end of the
-    unit’s destruction sequence. Callback methods bound to this event
-    should be used to ensure that the charm’s software is not running,
+    unit's destruction sequence. Callback methods bound to this event
+    should be used to ensure that the charm's software is not running,
     and that it will not start again on reboot.
     """
 
@@ -507,7 +488,7 @@ class RelationChangedEvent(RelationEvent):
     are incomplete, since it can be guaranteed that when the remote unit or
     application changes its settings, the event will fire again.
 
-    The settings that may be queried, or set, are determined by the relation’s
+    The settings that may be queried, or set, are determined by the relation's
     interface.
     """
 
@@ -522,8 +503,8 @@ class RelationDepartedEvent(RelationEvent):
     emitted once for each remaining unit.
 
     Callback methods bound to this event may be used to remove all
-    references to the departing remote unit, because there’s no
-    guarantee that it’s still part of the system; it’s perfectly
+    references to the departing remote unit, because there's no
+    guarantee that it's still part of the system; it's perfectly
     probable (although not guaranteed) that the system running that
     unit has already shut down.
 
@@ -581,7 +562,7 @@ class RelationBrokenEvent(RelationEvent):
     fire to signal that the relationship has been fully terminated.
 
     The event indicates that the current relation is no longer valid, and that
-    the charm’s software must be configured as though the relation had never
+    the charm's software must be configured as though the relation had never
     existed. It will only be called after every callback method bound to
     :class:`RelationDepartedEvent` has been run. If a callback method
     bound to this event is being executed, it is guaranteed that no remote units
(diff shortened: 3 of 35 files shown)


ops/_private/timeconv.py
```diff
@@ -63,7 +63,7 @@ def parse_rfc3339(s: str) -> datetime.datetime:
                              microsecond=microsecond, tzinfo=tz)
 
 
-def parse_duration(s: str) -> datetime.timedelta:
+def parse_duration(s: str) -> datetime.timedelta:  # noqa: C901
```
benhoyt, line 66: [C901](https://www.flake8rules.com/rules/C901.html) is "function is too complex", right? I think that's not appropriate for a linter and we should turn this off in Ruff's config. Whether or not a function is "too complex" is a nuanced question and should be judged in code review.

docs/conf.py
```diff
@@ -33,8 +33,8 @@ def _compute_navigation_tree(context):
 
 # Pull in fix from https://github.com/sphinx-doc/sphinx/pull/11222/files to fix
 # "invalid signature for autoattribute ('ops.pebble::ServiceDict.backoff-delay')"
-import re
-import sphinx.ext.autodoc
+import re  # noqa: E402
+import sphinx.ext.autodoc  # noqa: E402
```
benhoyt, line 37: Do you know why the tool is flagging [E402](https://www.flake8rules.com/rules/E401.html) - "multiple imports per line" - for these, when there only seems to be one per line?

ops/charm.py
```diff
@@ -18,32 +18,13 @@
...
-    NoReturn,
-    Optional,
-    TextIO,
-    Tuple,
-    TypedDict,
-    Union,
-    cast,
-)
+from typing import (TYPE_CHECKING, Any, Dict, List, Literal, Mapping, NoReturn,
+                    Optional, TextIO, Tuple, TypedDict, Union, cast)
```
benhoyt, line 22: It seems annoying to have formatting changes like this, when we're planning to start using Ruff's formatter shortly (which will probably change this back?). Any way to turn off this more aggressive formatting for this PR?

I'd probably even prefer a `noqa` comment on these for now if that helps, rather than reformatting only to use Ruff's formatter in a bit.

ops/pebble.py
```diff
@@ -808,7 +790,7 @@ class Layer:
     log_targets: Dict[str, 'LogTarget']
 
     def __init__(self, raw: Optional[Union[str, 'LayerDict']] = None):
-        if isinstance(raw, str):
+        if isinstance(raw, str):  # noqa: SIM108
```
benhoyt, line 793: I see that this is from flake8-simplify. I think we should probably exclude that additional linters in this PR, and discuss in a separate PR.

Perhaps that's a good general comment. Would be good to separate out "turning on additional checks" (even if they're good ones) from "switching to Ruff". I think we should *mostly* restrict this PR to the latter.

ops/testing.py
```diff
@@ -3285,9 +3254,9 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
             # conform with the real pebble api
             first_service = next(iter(service_names))
             message = f'cannot send signal to "{first_service}": invalid signal name "{sig}"'
-            raise self._api_error(500, message)
+            raise self._api_error(500, message) from None
 
-    def get_checks(self, level=None, names=None):  # type:ignore
+    def get_checks(self, level=None, names=None):  # type:ignore  # noqa: ANN001
```
benhoyt, line 3259: For ANN001 can we add type annotations, seeing we're using Pyright on everything anyway?

ops/testing.py
```diff
@@ -2595,18 +2565,15 @@ def secret_set(self, id: str, *,
     @classmethod
     def _generate_secret_id(cls) -> str:
         # Not a proper Juju secrets-style xid, but that's okay
-        return f"secret:{str(uuid.uuid4())}"
+        return f"secret:{uuid.uuid4()!s}"
```
benhoyt, line 2568: I believe `!s` is the default, and we can just drop `str()` or `!s` altogether.

ops/testing.py
```diff
@@ -2995,9 +2962,11 @@ def pull(self, path: str, *,
                 Union[BinaryIO, TextIO],
                 file_path.open("rb" if encoding is None else "r", encoding=encoding))
         except FileNotFoundError:
-            raise pebble.PathError('not-found', f'stat {path}: no such file or directory')
+            raise pebble.PathError('not-found',
+                                   f'stat {path}: no such file or directory') from None
```
benhoyt, line 2966: Are we sure we want to add `from None` to these?

test/test_charm.py
```diff
@@ -632,7 +632,7 @@ def test_containers_storage_multiple_mounts(self):
         self.assertEqual(meta.containers['test1'].mounts["data"].locations[1], '/test/otherdata')
 
         with self.assertRaises(RuntimeError):
-            meta.containers["test1"].mounts["data"].location
+            meta.containers["test1"].mounts["data"].location  # noqa: B018
```
benhoyt, line 635: We might want to disable B018 "useless expression" in `test/*`.

test/test_infra.py
```diff
@@ -32,24 +32,12 @@ def get_python_filepaths(include_tests: bool = True):
...
-        issues: typing.List[typing.Tuple[str, int, str]] = []
-        for filepath in get_python_filepaths():
-            with open(filepath, "rt", encoding="utf8") as fh:
-                for idx, line in enumerate(fh, 1):
-                    if (r'\"' in line or r"\'" in line) and "NOQA" not in line:
-                        issues.append((filepath, idx, line.rstrip()))
-        if issues:
-            msgs = ["{}:{:d}:{}".format(*issue) for issue in issues]
-            self.fail("Spurious backslashes found, please fix these quotings:\n" + "\n".join(msgs))
-
```
benhoyt, line 52: Nice, great that we can get rid of this ad-hoc test!

benhoyt (changes requested):
Thanks! Left a few initial comments.

ops/_private/timeconv.py
tonyandrewmeyer: Yeah, I tend to agree, and had this on my list of things to figure out. In the existing (main) config, we have:

```
select = ["E", "W", "F", "C", "N", "R", "D", "H"]
```

And we don't ignore anything starting with `C`. So it seems like this ought to be running at the moment, but I think that can't be the case, because I've bumped the maximum value up very high and it's still picking things up in the branch and nothing is detected in main.

Maybe the `C` in main is doing nothing because a plug-in isn't installed? Will dig more.

ops/charm.py
tonyandrewmeyer: Ha, I was thinking "might as well just let it format things since we're going to move to ruff's formatter soon", which is basically the opposite :laughing:. I'm fine with a `noqa` for now - I don't think there's any way to have the ruff version of `isort` be less aggressive here.

ops/_private/timeconv.py
tonyandrewmeyer: It looks like the mccabe check is only done with flake8 if you pass a maximum complexity value, so we were previously only semi-enabling the check.

docs/conf.py
tonyandrewmeyer: [E402](https://www.flake8rules.com/rules/E402.html) is "module level import not at top of file", the multiple imports per line one is E401.

I believe this wasn't picked up before because we didn't check anything in `docs`. It seems like it's reasonable to do so, and it's pretty small so I thought it could be included in this PR rather than broken out (but let me know if you disagree).

It seems possible that these could be moved up - I assume they're here because they are only used for the fix mentioned in the comment, but if the fix gets removed and we don't need them any more that would get picked up automatically. The PR that the code is from is still open, so it probably is still required.

docs/conf.py
```diff
@@ -150,15 +150,15 @@ def _compute_navigation_tree(context):
...
+    builder = sys.argv[sys.argv.index('-b') + 1]
 
 html_theme = 'furo'
 html_last_updated_fmt = ""
 html_permalinks_icon = "¶"
 html_theme_options = {
     "light_css_variables": {
         "color-sidebar-background-border": "none",
-        "font-stack": "Ubuntu, -apple-system, Segoe UI, Roboto, Oxygen, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif",
+        "font-stack": "Ubuntu, -apple-system, Segoe UI, Roboto, Oxygen, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif",  # noqa: E501
```
tonyandrewmeyer, line 161: This is just [line is too long](https://docs.astral.sh/ruff/rules/line-too-long/) and could be wrapped easily enough, but it seemed like this comes from somewhere rather than is hand-written so I thought leaving it untouched was perhaps best.

ops/charm.py
tonyandrewmeyer: Actually, I might have accidentally done this when doing some auto-fixing. Changing it back doesn't seem to make `check` complain.

ops/charm.py
```diff
@@ -226,9 +207,9 @@ class StartEvent(HookEvent):
 
     This event is triggered immediately after the first
     :class:`ConfigChangedEvent`. Callback methods bound to the event should be
-    used to ensure that the charm’s software is in a running state. Note that
-    the charm’s software should be configured so as to persist in this state
-    through reboots without further intervention on Juju’s part.
+    used to ensure that the charm's software is in a running state. Note that
```
tonyandrewmeyer, line 223: These all get picked up - even though I do prefer the correct character we've discussed this previously, and Sphinx corrects them automatically for the docs, so it's only if you're reading this in the file itself that you see `'`. If I recall correctly we agreed that we should standardise on `'` so it seems ok to include that in this PR.

ops/main.py
```diff
@@ -395,10 +395,7 @@ def main(charm_class: Type[ops.charm.CharmBase],
 
     metadata = (charm_dir / 'metadata.yaml').read_text()
     actions_meta = charm_dir / 'actions.yaml'
-    if actions_meta.exists():
-        actions_metadata = actions_meta.read_text()
-    else:
-        actions_metadata = None
+    actions_metadata = actions_meta.read_text() if actions_meta.exists() else None
```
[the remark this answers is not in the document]
tonyandrewmeyer: It's something the tool caught, but could be skipped to make the PR minimal.

ops/pebble.py
tonyandrewmeyer: I agree, I guess I just have a slightly smaller "mostly" than you :laughing:. I'll split this into two, a very minimal one first, and then the ones I think are simple and worth doing immediately, and still leave the ones I think need more discussion to later.

test/charms/test_main/src/charm.py
```diff
@@ -20,7 +20,7 @@
 
 sys.path.append('lib')
 
-import ops  # noqa: E402 (module-level import after non-import code)
+import ops
```
tonyandrewmeyer, line 23: Fwiw, I do not know why this is no longer necessary, and haven't bothered to dig into it. My assumption is that ruff ignores code that manipulates `sys.path` when doing this check.

ops/testing.py
```diff
@@ -3053,10 +3040,8 @@ def list_files(self, path: str, *, pattern: Optional[str] = None,
             raise self._api_error(404, f"stat {path}: no such file or directory")
         files = [file_path]
         if not itself:
-            try:
+            with contextlib.suppress(NotADirectoryError):
```
tonyandrewmeyer, line 3043: I'm pretty mixed on this. I didn't know about `contextlib.suppress` before the linter suggested it, and while this is clearly the intended use-case, I don't really understand why this is in the stdlib. Is this really better than just having the `except` with a `pass`? Am I missing an advantage of doing it this way?

benhoyt: I guess the main advantage is it's 2 lines vs 4, but yeah, I'm with you -- you have to learn a new function (that has a pretty vague name -- suppress what?) and it's pretty arguable.

ops/main.py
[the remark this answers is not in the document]
tonyandrewmeyer: I've left it in for this PR where it fits more (general minor improvements caught by a few extra linters) than in the first one (strictly migrate to ruff).

ops/testing.py
```diff
@@ -1705,12 +1706,9 @@ def test_hostname(self):
         Return:
             The path of the temporary directory associated with the specified container.
         """
-        # it's okay to access the container directly in this context, as its creation has already
+        # It's okay to access the container directly in this context, as its creation has already
```
tonyandrewmeyer, line 1709: Fwiw, this was me and not the linter...

ops/testing.py
tonyandrewmeyer: Turns out this is over ten years old :laughing:. Added in https://github.com/python/cpython/issues/60010 / https://github.com/python/cpython/commit/e318a883fece5d9a45165670b5f995886f524820 (the ticket referred to in the PR is wrong, maybe that is an issue with one of the tracker migrations?).

I don't really find the arguments in the ticket convincing, so I'm going to keep not using this, and will remove it here.

ops/testing.py
tonyandrewmeyer: I think so, although I did either miss them or think otherwise when I was doing the last PR that added a bunch of "from" clauses to raising.

With no "from" at all, they all end up looking like:

```
...
ExceptionType: message

During handling of the above exception, another exception occurred:

...
ExceptionType: message
```

This should really only be the case when there was an error during the exception handling. None of these are that - they are all converting the exception to a more meaningful type.

So that leaves the choice between "from None" and "from e". Do we want the charmer to know about the original exception, or is that just an implementation detail that could change in the future? To me these all seem like the latter.

tonyandrewmeyer:
One option here would be to split this into two PRs, one that has all the changes outside of pyproject.toml (and the noqa's, I guess) and then a second one that has the pyproject.toml and noqa changes, and also adds the previous commit to a `.git-blame-ignore-revs` file. Do you think that would be preferable? None of the code changes are functional ones, although some should be (insignificant) performance improvements.

test/test_pebble.py
```diff
@@ -1405,13 +1405,13 @@ def test_multipart_parser(self):
                 bodies_done: typing.List[bool] = []
 
                 def handle_header(data: typing.Any):
-                    headers.append(bytes(data))
-                    bodies.append(b'')
-                    bodies_done.append(False)
+                    headers.append(bytes(data))  # noqa: B023
+                    bodies.append(b'')  # noqa: B023
+                    bodies_done.append(False)  # noqa: B023
```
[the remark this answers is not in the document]
benhoyt: Oh, it's a bug (I think) in Ruff: add a link to https://github.com/astral-sh/ruff/issues/7847 in a comment.

benhoyt (commented):
Looks great to me, just the few things we discussed.

pyproject.toml
```diff
@@ -82,22 +82,87 @@ select = [
...
+    "SIM117",
+
     # Missing docstring in magic method
     "D105",
     # Missing docstring in `__init__`
     "D107",
+
+    ## Likely worth doing, but later.
+
+    # Convert {} from `TypedDict` functional to class syntax
```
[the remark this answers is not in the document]
tonyandrewmeyer: I double-checked this and ruff is smart enough to figure out that the ones with the hyphens can't be converted and doesn't flag those.

However, I don't know whether we would want to have all the `TypedDict` definitions be the class syntax, except for the ones that can't be, or whether we would rather be consistent like now. So I'll move it up for now at least.

tonyandrewmeyer:
> One option here would be to split this into two PRs, one that has all the changes outside of pyproject.toml (and the noqa's, I guess) and then a second one that has the pyproject.toml and noqa changes, and also adds the previous commit to a `.git-blame-ignore-revs` file. Do you think that would be preferable? None of the code changes are functional ones, although some should be (insignificant) performance improvements.

We briefly discussed this and decided that it's not worth it for this PR.

benhoyt (approved):

PR #1116, opened 2024-01-30
ci: reinstate postgresql-operator tests now that issue is fixed

benhoyt:
Also update links to mysql issues.

### DIFF
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -11,10 +11,9 @@ jobs:
         charm-repo:
           - "canonical/postgresql-operator"
           - "canonical/postgresql-k8s-operator"
-# TODO: uncomment once secrets issues are fixed in these charms:
-#       https://github.com/canonical/mysql-operator/pull/383
-#       https://github.com/canonical/data-platform-libs/pull/132
-#          - "canonical/mysql-operator"
+          - "canonical/mysql-operator"
+# TODO: uncomment once secrets issues are fixed in this charm:
+#       https://github.com/canonical/mysql-k8s-operator/pull/371
 #          - "canonical/mysql-k8s-operator"
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/db-charm-tests.yaml
```diff
@@ -9,10 +9,11 @@ jobs:
...
-# TODO: uncomment once secrets issue is fixed in this charm (see https://github.com/canonical/postgresql-operator/pull/307)
-#          - "canonical/postgresql-operator"
+          - "canonical/postgresql-operator"
           - "canonical/postgresql-k8s-operator"
-# TODO: uncomment once secrets issues are fixed in these charms (eg: https://github.com/canonical/mysql-operator/issues/367)
+# TODO: uncomment once secrets issues are fixed in these charms:
+#       https://github.com/canonical/mysql-operator/pull/383
+#       https://github.com/canonical/data-platform-libs/pull/132
 #          - "canonical/mysql-operator"
 #          - "canonical/mysql-k8s-operator"
```
tonyandrewmeyer, line 18:
```suggestion
          - "canonical/mysql-operator"
# TODO: uncomment once secrets issues are fixed in this charm:
#       https://github.com/canonical/mysql-k8s-operator/pull/371
#          - "canonical/mysql-k8s-operator"
```

tonyandrewmeyer (changes requested):
There has been a large refactor of secrets in the mysql charms over the few days (since I opened these PRs). The mysql-operator tests now pass and it seems like the problem no longer exists (deleting revisions has moved into data-platform-libs and is handled in a way that means that the cached secret shouldn't be accessed after deletion, I think).

The data-platform-libs issue (setting a secret twice can wipe the first value) still seems to exist but with the refactoring that code path isn't hit any more so the tests pass with current ops, even though that should still be fixed as far as I can tell.

There's also a further issue that only applies to mysql-k8s-operator, which is just that there is one specific test that relies on the "owner always peeks" behaviour. So that still fails, and I've opened a new PR for that.

tonyandrewmeyer (approved):

Issue #1103, opened 2024-01-08
Attempt to use Ruff formatter on this project


benhoyt:
I think it's time to use automatic code formatting. We could use Black, but because we want to use Ruff for linting, we should also probably use [Ruff's code formatter](https://docs.astral.sh/ruff/formatter/). It fixes several minor consistency issues with Black, and also adds a single-quote knob (yay! :-).

We should keep in mind that this project has tried to switch to Black twice in the past:

* Nov 2019 in PR #49. Comment against: https://github.com/canonical/operator/pull/49#pullrequestreview-320942544
* Again in Sep 2021 in PR #621. Comment against: https://github.com/canonical/operator/pull/621#issuecomment-925578397

We should consider those comments seriously, and see if there are fixes / workarounds for those issues, either by increasing the line length, adding/removing trailing commas to guide the tool, or manually reformatting the few places where the tool makes things significantly worse.

That said, the pros of using an automatic code formatter is high.

benhoyt:
We'd like to pair on this in person in Madrid: one of us can do a quick first pass, then we can go over style concerns together and try to nut it out in a morning.

benhoyt:
A couple of thoughts after looking at @IronCore864's [preview of model.py](https://github.com/IronCore864/operator/pull/3/files#diff-9ac334e8310ae3abff33570665575179e0ff2ffa874efb65bd773723989bd251):

1) I quite like the more consistent function parameter style, so I'm fine with that one.
2) I think we should *consider* increasing the max line length from 99 to say 109 or 119 columns. I think this would help avoid wrapping function calls and error messages *too* much on lines that are already somewhat indented. For reference, on my big screen I fit two panes of code side by side, each with 110 columns.
3) The one thing that stood out as annoying was that it doesn't seem to use ["cuddled braces"](https://google.github.io/styleguide/go/decisions#cuddled-braces). For example:

```python
# Old: 3 lines, easy to read
self._data.update({
    self.relation.app: RelationDataContent(self.relation, self.relation.app, backend),
})

# New: 7 lines! harder to read
self._data.update(
    {
        self.relation.app: RelationDataContent(
            self.relation, self.relation.app, backend
        ),
    }
)
```

Maybe we'll just have to get over that. Or maybe we can rewrite the ones that expand crazily to avoid the crazy 7-line wrapping:

```python
data = {self.relation.app: RelationDataContent(self.relation, self.relation.app, backend)}
self._data.update(data)
```

[...]

PR #1120, opened 2024-02-02
chore: move to ruff for linting

### LINKED ISSUE #1102: Use Ruff linter on the project

tonyandrewmeyer:
This PR replaces flake8, flake8-docstrings, flake8-builtins, isort, autopep8, pep8-naming, pyproject-flake8, and two test_infra unit tests with ruff for linting. Formatting remains unchanged (to be addressed in #1103).

Notes:

1. Ruff does not yet implement all of the pycodestyle rules, which we checked/fixed with autopep8, see: https://github.com/astral-sh/ruff/issues/2402
2. Ruff implements all Flake8 rules (the "F" rules).
3. flake8-docstrings is not listed in the Ruff re-implementation list by pydocstyle is there, which flake8-docstrings is based on
4. Ruff reimplements flake8-builtins (but picks up much more than we were previously, which is odd).
5. Ruff's isort is "profile='black'".
6. With flake8 we had "R" rules enabled, but I can't figure out what those were, or what provided them.

Where the linter picks up issues that the old tools did not, handle in one of these ways:
* Ignore with a `noqa:` directive if it's a false positive or should otherwise be permanently ignored in that specific case
* Ignore for a file or group of files (the docs and tests have several of these) where it's something we want to pick up in the core code but not everywhere
* Ignore with a note to review later when it's likely that there would be too much additional noise in this PR
* Make the recommended change, when it's small and seems reasonable

#1104 will continue on from this with a few more changes that are minimal and reasonable, and enabling additional rule sets (since they are bundled with ruff, and since ruff is so fast, they are basically free) that I agree with/like the most.

A few outdated `noqa:`  directives have been removed (ruff detects these as well).

Fixes #1102.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,8 +33,8 @@ furo._compute_navigation_tree = _compute_navigation_tree
 
 # Pull in fix from https://github.com/sphinx-doc/sphinx/pull/11222/files to fix
 # "invalid signature for autoattribute ('ops.pebble::ServiceDict.backoff-delay')"
-import re
-import sphinx.ext.autodoc
+import re  # noqa: E402
+import sphinx.ext.autodoc  # noqa: E402
 sphinx.ext.autodoc.py_ext_sig_re = re.compile(
     r'''^ ([\w.]+::)?            # explicit module name
           ([\w.]+\.)?            # module and/or class name(s)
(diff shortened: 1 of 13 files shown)


benhoyt (approved):
This looks great and was easy to review -- thanks!

PR #1124, opened 2024-02-07
feat: add the action's ID to the ActionEvent object

### LINKED ISSUE #1121: expose JUJU_ACTION_UUID envvar
Some old hooks-based charms use the JUJU_ACTION_UUID envvar.
When porting them to use ops, it would be nice if there was an opsy way to access its value.
How about adding an `ActionEvent.uuid -> int` property to fetch it?

tonyandrewmeyer:
This PR exposes the `JUJU_ACTION_UUID` environment variable's value as `ActionEvent.id`.

Fixes #1121

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -148,9 +148,17 @@ class ActionEvent(EventBase):
 
         Not meant to be called directly by charm code.
         """
+        self.id = cast(str, snapshot['id'])
         # Params are loaded at restore rather than __init__ because
         # the model is not available in __init__.
         self.params = self.framework.model._backend.action_get()
+
+    def snapshot(self) -> Dict[str, Any]:
+        """Used by the framework to serialize the event to disk.
+
+        Not meant to be called by charm code.
+        """
+        return {'id': self.id}
 
     def set_results(self, results: Dict[str, Any]):
         """Report the result of the action.
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -125,12 +125,15 @@ class ActionEvent(EventBase):
     :meth:`log`.
     """
 
+    id: str = ""
+    """The Juju ID of the action invocation."""
+
     params: Dict[str, Any]
     """The parameters passed to the action."""
 
-    def __init__(self, handle: 'Handle', id: str):
+    def __init__(self, handle: 'Handle', id: Optional[str] = None):
         super().__init__(handle)
-        self.id = id
+        self.id = id  # type: ignore (for backwards compatibility)
 
     def defer(self) -> NoReturn:
         """Action events are not deferrable like other events.
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -128,6 +128,10 @@ class ActionEvent(EventBase):
     params: Dict[str, Any]
     """The parameters passed to the action."""
 
+    def __init__(self, handle: 'Handle', id: str):
+        super().__init__(handle)
+        self.id = id
+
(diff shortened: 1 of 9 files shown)


ops/framework.py
```diff
@@ -1057,10 +1057,10 @@ class BoundStoredState:
     if TYPE_CHECKING:
         # to help the type checker and IDEs:
         @property
-        def _data(self) -> StoredStateData: ...  # noqa, type: ignore
```
tonyandrewmeyer, line 1060: This, and the next change, are drive-bys. I weirdly get ruff complaining about these but only some of the time that it's run (I haven't dug into why that is). They do not seem to be necessary, so simplest to just remove.

ops/charm.py
```diff
@@ -128,6 +128,10 @@ class ActionEvent(EventBase):
     params: Dict[str, Any]
     """The parameters passed to the action."""
 
+    def __init__(self, handle: 'Handle', id: str):
```
tonyandrewmeyer, line 131: This could be `Optional[str]` and default to `None`, which would provide backwards compatibility. It seems unlikely that much code would be creating an `ActionEvent`, but it is possible (I'll check scenario, for example).

It would be nice if it didn't have to be, so that we don't end up in another situation where charms need to have a bunch of `assert event.id is not None` statements when in production `.id` is actually never None.

tonyandrewmeyer: Indeed, scenario does fail. I can hack a pass with:

```diff
diff --git a/scenario/runtime.py b/scenario/runtime.py
index 489169b..1760200 100644
--- a/scenario/runtime.py
+++ b/scenario/runtime.py
@@ -194,6 +194,7 @@ class Runtime:
             "JUJU_DISPATCH_PATH": f"hooks/{event.name}",
             "JUJU_MODEL_NAME": state.model.name,
             "JUJU_ACTION_NAME": action_name,
+            "JUJU_ACTION_UUID": "1",
             "JUJU_MODEL_UUID": state.model.uuid,
             "JUJU_CHARM_DIR": str(charm_root.absolute()),
             # todo consider setting pwd, (python)path
```

Although I'd have to figure out what the proper fix would be (presumably not always a fixed ID).

There could be other cases where there's an `ActionEvent` created/emitted manually. Once I get my "test lots of charms" script fully working I could test with that, but I'm not sure if it would check enough. It could default to `""` or some other string, so that the type is at least correct? Perhaps with a deprecation warning? I don't love that, though.

benhoyt: Per discussion, for backwards-compatibility, yeah, let's make the `__init__` signature take `id: Optional[str] = None` as a keyword arg, but then have the attribute be type `str` (non-optional) with default value `''`.

ops/testing.py
```diff
@@ -1883,7 +1884,8 @@ def run_action(self, action_name: str,
         action_under_test = _RunningAction(action_name, ActionOutput([], {}), params)
         handler = getattr(self.charm.on, f"{action_name.replace('-', '_')}_action")
         self._backend._running_action = action_under_test
-        handler.emit()
+        self._action_id_counter += 1
+        handler.emit(self._action_id_counter)
```
benhoyt, line 1888:
```suggestion
        handler.emit(str(self._action_id_counter))
```

Also would probably be good to have a test that catches this.

benhoyt (approved):
Looks good to me -- thanks for this!

PR #1130, opened 2024-02-14
fix(testing): set JUJU_REMOTE_APP and allow fetching relation data in relation-broken

### LINKED ISSUE #1128: Harness should always provide the remote app name in relation-broken
Although it seems like sometimes JUJU_REMOTE_APP is not set in relation-broken, the [conclusion in the Juju investigation](https://bugs.launchpad.net/juju/+bug/1960934) was that this should not happen, and it couldn't be reproduced.

We should remove [the special-casing in Harness that acts as if this *is* the case](https://github.com/search?q=repo%3Acanonical%2Foperator%20https%3A%2F%2Fbugs.launchpad.net%2Fjuju%2F%2Bbug%2F1960934&type=code).

tonyandrewmeyer:
As far as the Juju team and I have been able to tell (see [comments in the Juju issue](https://bugs.launchpad.net/juju/+bug/1960934)) `JUJU_REMOTE_APP` is always set in relation-broken events.

This PR removes the `Harness` behaviour that simulates `JUJU_REMOTE_APP` *not* being available (which was never consistently the case anyway) and the model code that blocks getting remote data during relation-broken (which is discouraged in the case where the entire integration is going away, but is permitted).

Fixes #1128.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2056,13 +2056,15 @@ class Container:
     For methods that make changes to the container, if the change fails or times out, then a
     :class:`ops.pebble.ChangeError` or :class:`ops.pebble.TimeoutError` will be raised.
 
-    Interactions with the container use Pebble, so all methods may raise exceptions when there are
-    problems communicating with Pebble. Problems connecting to or transferring data with Pebble
-    will raise a :class:`ops.pebble.ConnectionError` - generally you can guard against these by
-    first checking :meth:`can_connect`, but it is possible for problems to occur after
-    :meth:`can_connect` has succeeded. When an error occurs executing the request, such as trying
-    to add an invalid layer or execute a command that does not exist, an
-    :class:`ops.pebble.APIError` is raised.
+    Interactions with the container use Pebble, so all methods may raise
+    exceptions when there are problems communicating with Pebble. Problems
+    connecting to or transferring data with Pebble will raise a
+    :class:`ops.pebble.ConnectionError` - you can guard against these by first
+    checking :meth:`can_connect`, but that generally introduces a race condition
+    where problems occur after :meth:`can_connect` has succeeded. When an error
+    occurs executing the request, such as trying to add an invalid layer or
+    execute a command that does not exist, an :class:`ops.pebble.APIError` is
+    raised.
     """
 
     name: str
@@ -2086,14 +2088,12 @@ class Container:
 
         For example::
 
+            # Add status based on any earlier errors communicating with Pebble.
+            ...
+            # Check that Pebble is still reachable now.
             container = self.unit.get_container("example")
-            if container.can_connect():
-                try:
-                    c.pull('/does/not/exist')
-                except ProtocolError, PathError:
-                    # handle it
-            else:
-                event.defer()
+            if not container.can_connect():
+                event.add_status(ops.WaitingStatus("Waiting for Pebble..."))
         """
         try:
             self._pebble.get_system_info()
--- a/ops/model.py
+++ b/ops/model.py
@@ -1550,6 +1550,8 @@ class RelationData(Mapping[Union['Unit', 'Application'], 'RelationDataContent'])
         return iter(self._data)
 
     def __getitem__(self, key: Union['Unit', 'Application']):
+        if not isinstance(key, Unit) and not isinstance(key, Application):
+            raise KeyError('Can only index relation data with a Unit or Application')
         return self._data[key]
 
     def __repr__(self):
--- a/ops/model.py
+++ b/ops/model.py
@@ -1550,15 +1550,6 @@ class RelationData(Mapping[Union['Unit', 'Application'], 'RelationDataContent'])
         return iter(self._data)
 
     def __getitem__(self, key: Union['Unit', 'Application']):
-        if key is None and self.relation.app is None:
-            # NOTE: if juju gets fixed to set JUJU_REMOTE_APP for relation-broken events, then that
-            # should fix the only case in which we expect key to be None - potentially removing the
-            # need for this error in future ops versions (i.e. if relation.app is guaranteed to not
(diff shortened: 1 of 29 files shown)


tonyandrewmeyer:
Note that the canonical/nginx-ingress-integrator-operator tests pass under this branch.

tonyandrewmeyer:
The spacing changes in [pebble.py](ops/pebble.py) and [docs/conf.py](docs/conf.py) are ones that Ruff wants. This seems to be a change in 0.2.2, but I'm not sure why it didn't show up in the PR that bumped ruff to 0.2.2. I don't love the Pebble change, but I think I prefer it to a bunch of `noqa`s in the same place.

benhoyt (approved):
Excellent, thanks for this fix.

benhoyt:
@jameinel Would be good to have your eyes on this too, seeing we've discussed this before and you commented on the issue.

tonyandrewmeyer:
> The spacing changes in [pebble.py](ops/pebble.py) and [docs/conf.py](docs/conf.py) are ones that Ruff wants. This seems to be a change in 0.2.2 ([#9266](https://github.com/astral-sh/ruff/pull/9266)), but I'm not sure why it didn't show up in the PR that bumped ruff to 0.2.2. I don't love the Pebble change, but I think I prefer it to a bunch of `noqa`s in the same place.

This was figured out in #1134 - the version was `~0.2.1` previously, and new rules were introduced in `0.2.2`. That PR takes care of this, but will leave the adjustments here as well.

ops/model.py
```diff
@@ -1550,15 +1550,6 @@ def __iter__(self):
...
-        if key is None and self.relation.app is None:
-            # NOTE: if juju gets fixed to set JUJU_REMOTE_APP for relation-broken events, then that
-            # should fix the only case in which we expect key to be None - potentially removing the
-            # need for this error in future ops versions (i.e. if relation.app is guaranteed to not
-            # be None. See https://bugs.launchpad.net/juju/+bug/1960934.
-            raise KeyError(
-                'Cannot index relation data with "None".'
-                ' Are you trying to access remote app data during a relation-broken event?'
-                ' This is not allowed.')
         return self._data[key]
```
jameinel, line 1553: would we want to have a different check in case key is None? Just to avoid weird KeyError failures.
I don't know that it is necessary, but I like giving better errors when we can.

ops/testing.py
```diff
@@ -2167,21 +2167,9 @@ def relation_remote_app_name(self, relation_id: int) -> Optional[str]:
         if relation_id not in self._relation_app_and_units:
             # Non-existent or dead relation
             return None
-        if 'relation_broken' in self._hook_is_running:
```
jameinel, line 2170: is _hook_is_running used for anything else? or was this the use case and we can pull it out?

jameinel (approved):
I'm happy with the changes.
I didn't do the Juju side testing to confirm that for all supported Juju versions this is the expected behavior. But I think it is the behavior that we want to have.

ops/testing.py
tonyandrewmeyer: It's used for bypassing access control on relation data as well.

tonyandrewmeyer:
I should do a CHANGES entry for this I think.

tonyandrewmeyer:
I've run this branch against the unit tests of 132 charms (all passing with current main of ops) and they all continue to pass.

ops/model.py
```diff
@@ -1550,15 +1550,8 @@ def __iter__(self):
...
-        if key is None and self.relation.app is None:
-            # NOTE: if juju gets fixed to set JUJU_REMOTE_APP for relation-broken events, then that
-            # should fix the only case in which we expect key to be None - potentially removing the
-            # need for this error in future ops versions (i.e. if relation.app is guaranteed to not
-            # be None. See https://bugs.launchpad.net/juju/+bug/1960934.
-            raise KeyError(
-                'Cannot index relation data with "None".'
-                ' Are you trying to access remote app data during a relation-broken event?'
-                ' This is not allowed.')
+        if not isinstance(key, Unit) and not isinstance(key, Application):
```
benhoyt, line 1553:
```suggestion
        if not isinstance(key, (Unit, Application)):
```

benhoyt: I'm fine with this addition here for those not using static type checking, but I think in general our approach is to lean on static checking now.

benhoyt (approved):
Let's do this!

ops/model.py
tonyandrewmeyer: I felt this was better than checking explicitly for `None` as John suggested. I'm fine pulling it out completely if you prefer.

benhoyt: Tony and I discussed, and we'd like our principle to be "we've invested in type annotations, let's go all in on static type checking and let that catch this." Most charming teams are using MyPy and PyRight already, and those that aren't, I think we should nudge them in that direction.

ops/model.py
benhoyt: Tony and I discussed, and we'd like our principle to be "we've invested in type annotations, let's go all in on static type checking and let that catch this." Most charming teams are using MyPy and PyRight already, and those that aren't, I think we should nudge them in that direction.

benhoyt (approved):

tonyandrewmeyer:
> > `JUJU_REMOTE_APP` is always set in relation-broken events
> 
> @tonyandrewmeyer is this a recent change? I remember encountering an occasional issue on juju 2.9 where `JUJU_REMOTE_APP` was not set (depends on race with relation-broken event on other charm)

The Juju team couldn't find a way to reproduce it (they tried recently). Most significantly, there was consensus that it *should* be always set, so having ops try to be consistent or work around a bug wasn't a good move.

If you know/find a way to reproduce, please open a Juju bug (or re-open the old one) with steps and we can make sure it gets fixed there.

PR #1134, opened 2024-02-25
feat: add the ability to compare two Plan objects with == and create a Plan from a dict

tonyandrewmeyer:
This PR adds two minor features to `pebble.Plan` objects to simplify testing (particularly with Scenario):

* The ability to compare two `pebble.Plan` objects, or a `Plan` with a `dict`, with `==` - comparison to any other type of object  will still return False.
* The ability to instantiate a `pebble.Plan` object from a `pebble.PlanDict` (or equivalent dict).

This aligns the `Plan` functionality with `Layer` objects.

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -791,8 +791,10 @@ class Plan:
 
     __str__ = to_yaml
 
-    def __eq__(self, __value: object) -> bool:
-        if isinstance(__value, Plan):
+    def __eq__(self, __value: Union['PlanDict', 'Plan']) -> bool:
+        if isinstance(__value, dict):
+            return self.to_dict() == __value
+        elif isinstance(__value, Plan):
             return self.to_dict() == __value.to_dict()
         return super().__eq__(__value)
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -735,8 +735,11 @@ class Plan:
     documented at https://github.com/canonical/pebble/#layer-specification.
     """
 
-    def __init__(self, raw: str):
-        d = yaml.safe_load(raw) or {}  # type: ignore
+    def __init__(self, raw: Optional[Union[str, 'PlanDict']]):
+        if isinstance(raw, str):
+            d = yaml.safe_load(raw) or {}  # type: ignore
+        else:
+            d = raw or {}
         d = typing.cast('PlanDict', d)
 
         self._raw = raw
@@ -787,6 +790,11 @@ class Plan:
         return yaml.safe_dump(self.to_dict())
 
     __str__ = to_yaml
+
+    def __eq__(self, __value: object) -> bool:
+        if isinstance(__value, Plan):
+            return self.to_dict() == __value.to_dict()
+        return super().__eq__(__value)
 
 
 class Layer:
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -743,7 +743,7 @@ class Plan:
     documented at https://github.com/canonical/pebble/#layer-specification.
     """
 
-    def __init__(self, raw: Optional[Union[str, 'PlanDict']]):
+    def __init__(self, raw: Optional[Union[str, 'PlanDict']] = None):
         if isinstance(raw, str):  # noqa: SIM108
             d = yaml.safe_load(raw) or {}  # type: ignore
         else:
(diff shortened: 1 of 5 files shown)


tonyandrewmeyer:
The spacing lint errors in [pebble.py](ops/pebble.py) and [docs/conf.py](docs/conf.py) are ones that Ruff wants. This seems to be a change in 0.2.2 ([#9266](https://github.com/astral-sh/ruff/pull/9266)), but I'm not sure why it didn't show up in the PR that bumped ruff to 0.2.2. I'll make the adjustments, since they seem minorly better than `noqa`s.

benhoyt:
Do you mean the PR that bumped ruff to 0.2.1? (https://github.com/canonical/operator/commit/1836df5affb42b3183125b1904c794090aa1862b#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449) I wonder if we should pin an exact version for Ruff, if this is going to be a problem (particularly as we have preview turned on).

tonyandrewmeyer:
> Do you mean the PR that bumped ruff to 0.2.1? ([1836df5#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449](https://github.com/canonical/operator/commit/1836df5affb42b3183125b1904c794090aa1862b#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449)) I wonder if we should pin an exact version for Ruff, if this is going to be a problem (particularly as we have preview turned on).

Ha, yes. I remembered it going to 0.2 and assumed it must have been 0.2.2 but the "why did it change" obvious answer is indeed that 😂.

We currently have it `ruff~=0.2.1`- I would have thought adding new rules would be a minor bump, not a patch one.looking over the changelog, 0.2 had breaking changes from 0.1, so it's like semvar shifted to the right 😞.

I'd be fine pinning it to an exact version (we could reconsider at 1.0.0 maybe). In the other hand, if the changes tend to be small, I'm also fine with just dealing with them as they come, since we have a small team, so don't generally have a lot of PRs at once.

test/test_pebble.py
```diff
@@ -610,6 +639,110 @@ def test_service_equality(self):
...
+        self.assertTrue(plan1 == plan2)
+        plan1_as_dict = pebble.PlanDict({
+            "services": {
+                "foo": pebble.ServiceDict({
+                    "command": "echo foo",
+                    "override": "replace",
+                })
+            },
+        })
+        self.assertTrue(plan1, plan1_as_dict)
```
benhoyt, line 665: Was this intended as `plan1 == plan1_as_dict`?

test/test_pebble.py
```diff
@@ -610,6 +639,110 @@ def test_service_equality(self):
...
+  http:
+   https://example.org/
+
+log-targets:
+ baz:
+  override: replace
+  type: loki
+  location: https://example.com:3100/loki/api/v1/push
+''')
+        self.assertFalse(plan4 == plan5)
```
benhoyt, line 707: It wasn't at all obvious to me why this was false (finally found the .com vs .org). Can we comment here or above?

benhoyt (changes requested):
Thanks for this. Agreed overall -- some minor comments and changes requested.

benhoyt:
> We currently have it ruff~=0.2.1 - I would have thought adding new rules would be a minor bump, not a patch one. Looking over the changelog, 0.2 had breaking changes from 0.1, so it's like semvar shifted to the right 😞.

Heh, Pyright-style "semver". :-)

> I'd be fine pinning it to an exact version (we could reconsider at 1.0.0 maybe). In the other hand, if the changes tend to be small, I'm also fine with just dealing with them as they come, since we have a small team, so don't generally have a lot of PRs at once.

Okay, let's leave as is for now, and we can always reconsider if it keeps being a problem.

test/test_pebble.py
tonyandrewmeyer: Yes - good catch, thanks! Another example of where pytest style `assert plan1 == plan1_as_dict` would be much better - I'm sure I wouldn't have typed `assert plan1, plan1_as_dict`.

test/test_pebble.py
```diff
@@ -610,6 +639,110 @@ def test_service_equality(self):
...
         self.assertEqual(plan.services, services_as_dict)
 
+    def test_plan_equality(self):
+        plan1 = pebble.Plan('''
+services:
+  foo:
+    override: replace
+    command: echo foo
+''')
+        self.assertFalse(plan1 == "foo")
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yes, that probably is better (although not as good as a simple pytest `assert plan == "foo"` ...), thanks.

ops/pebble.py
```diff
@@ -788,6 +799,13 @@ def to_yaml(self) -> str:
 
     __str__ = to_yaml
 
+    def __eq__(self, __value: Union['PlanDict', 'Plan']) -> bool:
+        if isinstance(__value, dict):
+            return self.to_dict() == __value
+        elif isinstance(__value, Plan):
+            return self.to_dict() == __value.to_dict()
+        return super().__eq__(__value)
```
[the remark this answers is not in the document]
tonyandrewmeyer: I guess not. `super` felt more correct to me ("I don't know, ask the other classes I'm made of" rather than "I don't implement this check") but `NotImplemented` is more consistent with `Layer` and is probably more expected with `__eq__` I guess. I'll change it.

benhoyt (approved):

PR #1138, opened 2024-02-28
fix: add_relation consistency check and default network

### LINKED ISSUE #1136: `harness.add_relation(foo)` should fail if foo is not a valid enpdoint name as per metadata
currently this test passes

```python
        harness = ops.testing.Harness(ops.CharmBase, meta='name: test-app')
        rel_id = harness.add_relation('db', 'postgresql')
```

while it feels to me like it should most loudly fail.

### LINKED ISSUE #1137: adding a relation to Harness should automatically add a corresponding network binding
In juju, any time a relation is added, a network is provisioned for it.
Harness should mimic that behaviour by creating a network for each added relation.

PietroPasotti:
This PR makes two changes in the harness:
- forbids doing add_relation on endpoints that haven't previously been defined in the charm metadata (something that scenario handles via the consistency checker)
- automatically adds a default network, if not found already, for each newly added relation, to match juju's behaviour.

Fixes #1137 
Fixes #1136

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -595,7 +595,7 @@ class Harness(Generic[CharmType]):
                         'username': 'username',
                         'password': 'password',
                         }
-        if resource_name not in self._meta.resources.keys():
+        if resource_name not in self._meta.resources:
             raise RuntimeError(f'Resource {resource_name} is not a defined resources')
         if self._meta.resources[resource_name].type != "oci-image":
             raise RuntimeError(f'Resource {resource_name} is not an OCI Image')
@@ -614,7 +614,7 @@ class Harness(Generic[CharmType]):
             content: Either string or bytes content, which will be the content of the filename
                 returned by resource-get. If contents is a string, it will be encoded in utf-8
         """
-        if resource_name not in self._meta.resources.keys():
+        if resource_name not in self._meta.resources:
             raise RuntimeError(f'Resource {resource_name} is not a defined resource')
         record = self._meta.resources[resource_name]
         if record.type != "file":
@@ -1706,12 +1706,9 @@ class Harness(Generic[CharmType]):
         Return:
             The path of the temporary directory associated with the specified container.
         """
-        # it's okay to access the container directly in this context, as its creation has already
+        # It's okay to access the container directly in this context, as its creation has already
         # been ensured during the model's initialization.
-        if isinstance(container, str):
-            container_name = container
-        else:
-            container_name = container.name
+        container_name = container if isinstance(container, str) else container.name
         return self._backend._pebble_clients[container_name]._root
 
     def evaluate_status(self) -> None:
@@ -1866,7 +1863,7 @@ class Harness(Generic[CharmType]):
         try:
             action_meta = self.charm.meta.actions[action_name]
         except KeyError:
-            raise RuntimeError(f"Charm does not have a {action_name!r} action.")
+            raise RuntimeError(f"Charm does not have a {action_name!r} action.") from None
         if params is None:
             params = {}
         for key in action_meta.required:
@@ -1896,9 +1893,7 @@ class Harness(Generic[CharmType]):
 
 def _get_app_or_unit_name(app_or_unit: AppUnitOrName) -> str:
     """Return name of given application or unit (return strings directly)."""
-    if isinstance(app_or_unit, model.Application):
-        return app_or_unit.name
-    elif isinstance(app_or_unit, model.Unit):
+    if isinstance(app_or_unit, (model.Application, model.Unit)):
         return app_or_unit.name
     elif isinstance(app_or_unit, str):
         return app_or_unit
@@ -1918,9 +1913,9 @@ def _record_calls(cls: Any):
 
         def decorator(orig_method: Any):
             def wrapped(self: '_TestingModelBackend', *args: Any, **kwargs: Any):
-                full_args = (orig_method.__name__,) + args
+                full_args = (orig_method.__name__, *args)
                 if kwargs:
-                    full_args = full_args + (kwargs,)
+                    full_args = (*full_args, kwargs)
                 self._calls.append(full_args)
                 return orig_method(self, *args, **kwargs)
             return wrapped
@@ -1940,7 +1935,7 @@ def _copy_docstrings(source_cls: Any):
     __doc__ for that method.
     """
     def decorator(target_cls: Any):
-        for meth_name, _ in target_cls.__dict__.items():
+        for meth_name in target_cls.__dict__:
             if meth_name.startswith('_'):
                 continue
             source_method = source_cls.__dict__.get(meth_name)
@@ -1987,15 +1982,15 @@ class _TestingConfig(Dict[str, Union[str, int, float, bool]]):
         # has the expected type.
         option = self._spec.get('options', {}).get(key)
         if not option:
-            raise RuntimeError('Unknown config option {}; '
+            raise RuntimeError(f'Unknown config option {key}; '
                                'not declared in `config.yaml`.'
                                'Check https://juju.is/docs/sdk/config for the '
-                               'spec.'.format(key))
+                               'spec.')
 
         declared_type = option.get('type')
         if not declared_type:
-            raise RuntimeError('Incorrectly formatted `options.yaml`, option {} '
-                               'is expected to declare a `type`.'.format(key))
+            raise RuntimeError(f'Incorrectly formatted `options.yaml`, option {key} '
+                               'is expected to declare a `type`.')
 
         if declared_type not in self._supported_types:
             raise RuntimeError(
@@ -2003,9 +1998,8 @@ class _TestingConfig(Dict[str, Union[str, int, float, bool]]):
                 'of [{}], not {}.'.format(', '.join(self._supported_types), declared_type))
 
         if type(value) is not self._supported_types[declared_type]:
-            raise RuntimeError('Config option {} is supposed to be of type '
-                               '{}, not `{}`.'.format(key, declared_type,
-                                                      type(value).__name__))
+            raise RuntimeError(f'Config option {key} is supposed to be of type '
+                               f'{declared_type}, not `{type(value).__name__}`.')
 
         # call 'normal' setattr.
         dict.__setitem__(self, key, value)  # type: ignore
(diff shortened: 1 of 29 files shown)


PietroPasotti:
FWIW linting failures appear to be largely unrelated to my changes

tonyandrewmeyer:
> FWIW linting failures appear to be largely unrelated to my changes

It's the re-formatted imports. Did you run `tox -e fmt` maybe? I wonder if we have a conflict between isort and ruff styling here, since we have moved to ruff for checking but haven't moved for formatting yet.

tonyandrewmeyer:
> > FWIW linting failures appear to be largely unrelated to my changes
> 
> It's mostly the re-formatted imports. Did you run `tox -e fmt` maybe? I wonder if we have a conflict between isort and ruff styling here, since we have moved to ruff for checking but haven't moved for formatting yet.

`tox -e fmt` does indeed cause this. It's a bit better if isort is changed to `isort {[vars]all_path} --multi-line=3 --line-length=99` but `multi-line=3` has no trailing comma (even though the [docs](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html) show one). I can't find a `ruff` option that will accept the missing comma only for imports.

@benhoyt for this PR we can just limit the changes to the relevant ones and all the tests will pass. It is a problem that `tox -e fmt` leaves you with formatting that `tox -e lint` doesn't accept, though. Should we just push ahead with #1103?

tonyandrewmeyer:
> `tox -e fmt` does indeed cause this. It's a bit better if isort is changed to `isort {[vars]all_path} --multi-line=3 --line-length=99` but `multi-line=3` has no trailing comma (even though the [docs](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html) show one). I can't find a `ruff` option that will accept the missing comma only for imports.

Actually, even though it's not the defined purpose of the option, `isort {[vars]all_path} --multi-line=3 --line-length=99 --split-on-trailing-comma` gets us to a point where you can `fmt` and `lint` without conflicts. I'll open a PR for that as a solution before #1103.

benhoyt:
FYI, I pulled in Tony's linting changes and reverted the import formatting stuff.

PietroPasotti:
Oh so happy you can fix stuff while I'm sleeping. 
Sorry about missing that, I was indeed very much assuming that `tox -e fmt` wouldn't break the linter checks :D , so much I didn't even think to check whether the failures were introduced by the formatter.

ops/testing.py
```diff
@@ -799,6 +799,9 @@ def add_relation(self, relation_name: str, remote_app: str, *,
 
         This function creates a relation with an application and triggers a
         :class:`RelationCreatedEvent <ops.RelationCreatedEvent>`.
+        This function also creates a defaulted network binding on this endpoint. If you wish
```
benhoyt, line 802:
```suggestion
        It also creates a default network binding on this endpoint. If you want
```

ops/testing.py
```diff
@@ -799,6 +799,9 @@ def add_relation(self, relation_name: str, remote_app: str, *,
...
+        to associate a custom network to this binding, provide one using
+        `add_network` before calling `add_relation`.
```
benhoyt, line 804:
```suggestion
        :meth:`add_network` before calling this function.
```

ops/testing.py
```diff
@@ -832,6 +835,11 @@ def add_relation(self, relation_name: str, remote_app: str, *,
         Return:
             The ID of the relation created.
         """
+        if not (relation_name in self._meta.provides
+                or relation_name in self._meta.requires
+                or relation_name in self._meta.peers):
+            raise RelationNotFoundError("relation %s not declared in metadata" % relation_name)
```
benhoyt, line 841:
```suggestion
            raise RelationNotFoundError(f'relation {relation_name!r} not declared in metadata')
```

benhoyt (approved):
Just a couple of nit suggestions, but looks good to me other than that.

tonyandrewmeyer (commented):
This looks great, thanks! A bunch of nits, but nothing significant.

I wondered if adding the consistency check to `add_relation` might break a bunch of existing charm tests (in the same way that you've had to update a lot of the harness ones) so I ran this on the 135 charms I have locally (that pass with ops:main) and had new failures with:

* charm-microk8s (related to the network - error seems odd, maybe a test needs fixing? I didn't look deep)
* charm-software-inventory-exporter (related to the consistency check - could easily submit a patch)
* vault-k8s-operator (the errors aren't obviously related, but it consistently fails with the branch and passes with main, so I assume are)
* openfga-operator (network)
* gunicorn-k8s-operator (consistency check, could submit a patch)

The consistency check ones don't bother me much, and it's good that the number of fixes needed is low. The network ones are more of a concern - it seems like there must be some other default or something happening? I'll dig into that a bit. I'll try to figure out what's going on with vault, too.

ops/testing.py
```diff
@@ -859,6 +867,13 @@ def add_relation(self, relation_name: str, remote_app: str, *,
             if unit_data is not None:
                 self.update_relation_data(relation_id, remote_unit, unit_data)
 
+        # if we don't already have a network binding for this relation id, create one
+        if not self._backend._networks.get((relation_name, relation_id)):
+            self.add_network("10.0.0.10", endpoint=relation_name, relation_id=relation_id)
+        # if we don't already have a default network binding for this endpoint, create one
+        if not self._backend._networks.get((relation_name, None)):
+            self.add_network("10.0.0.10", endpoint=relation_name)
```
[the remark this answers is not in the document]
PietroPasotti: I need to bookmark that spec, I always think about it but I'm too lazy to go look it up again and again

tonyandrewmeyer:
> * charm-microk8s (related to the network - error seems odd, maybe a test needs fixing? I didn't look deep)

The issue here is that the test does `harness.add_network(ip)`, which adds the IP for the default binding. The (branch) `add_relation()` then adds IPs for the relation by ID and for the endpoint. The test assumes that it'll get back the IP it added, but `add_relation()` added more specific bindings, so it gets that IP instead.

My feeling is that we should also check if there is a `None, None` network, and skip adding one if that is the case, which would mean these tests still pass - and I think that's more similar to the actual Juju behaviour. I have done very little with networks/spaces, though, so I could be misunderstanding something here.

This also made me wonder if `remove_relation()` should be removing these networks, to be tidy if nothing else. It's not if you do `add_relation()` and `add_network(relation, ...)` now, though, so perhaps not?

> * charm-software-inventory-exporter (related to the consistency check - could easily submit a patch)

This needs a [minimal change](https://github.com/tonyandrewmeyer/charm-software-inventory-exporter/pull/1/files) and I think it's actually cleaner anyway.

> * vault-k8s-operator (the errors aren't obviously related, but it consistently fails with the branch and passes with main, so I assume are)

One of the tests (`test_given_bind_address_not_available_when_evaluate_status_then_status_is_waiting`) adds a peer relation (and a bit of other setup), and then checks that evaluating the status goes down the "no address" route (the test name basically explains it). I assume this could actually happen with Juju? If so then the test would need to _remove_ the default network that `add_relation()` added, which seems messy. However, this was the only charm I found that had this issue.

The other test (`test_given_ca_cert_is_not_stored_when_certificate_transfer_relation_joins_then_ca_cert_is_not_advertised`), I would need to look at more closely, but I believe what's happening is the same issue as the `charm-microk8s` one, in that it's picking up the wrong bind address (and then it tries to set up certificates for that and the test ends up failing because it's expecting those in the wrong place, or something like that).

> * openfga-operator (network)

This is the same as `charm-microk8s`.

> * gunicorn-k8s-operator (consistency check, could submit a patch)

This needed a [more substantial change](https://github.com/tonyandrewmeyer/gunicorn-k8s-operator/pull/1/files), although most of it is just moving lines of code around. The tests are working with relations that don't actually exist in the metadata, so need a custom harness set up. I'm not familiar enough with the charm or its tests to judge how valuable these tests are, but I assume they are required. It's a simple enough change, though.

ops/testing.py
```diff
@@ -859,6 +867,13 @@ def add_relation(self, relation_name: str, remote_app: str, *,
             if unit_data is not None:
                 self.update_relation_data(relation_id, remote_unit, unit_data)
 
+        # If we don't already have a network binding for this relation id, create one.
+        if not self._backend._networks.get((relation_name, relation_id)):
+            self.add_network("10.0.0.10", endpoint=relation_name, relation_id=relation_id)
+        # If we don't already have a default network binding for this endpoint, create one.
+        if not self._backend._networks.get((relation_name, None)):
+            self.add_network("192.0.2.0", endpoint=relation_name)
```
tonyandrewmeyer, line 875: Sorry, I put this in a comment but should have put it in a review as well so that it was clearer: I think that this needs to have a `if not self._backend._networks.get((None, None)):` guard around it, to handle the case where there's a default network that will be used when there is not one specific to the relation name+id or relation name. The comment has details of a couple of charms where not doing this would change the behaviour, and (from what I understand) not in a way that we expect.

ops/testing.py
```diff
@@ -799,6 +799,9 @@ def add_relation(self, relation_name: str, remote_app: str, *,
 
         This function creates a relation with an application and triggers a
         :class:`RelationCreatedEvent <ops.RelationCreatedEvent>`.
+        It also creates a default network binding on this endpoint. If you want
+        to associate a custom network to this binding, provide one using
+        :meth:`add_network` before calling this function.
```
tonyandrewmeyer, line 804: Maybe also a doc comment that these networks are not removed on `remove_relation`?

tonyandrewmeyer (changes requested):
Everything seems good to me, except for the situation where there's a default network set and `add_relation` ends up bypassing that in (I think) unexpected ways.

ops/testing.py
```diff
@@ -799,6 +799,9 @@ def add_relation(self, relation_name: str, remote_app: str, *,
 
         This function creates a relation with an application and triggers a
         :class:`RelationCreatedEvent <ops.RelationCreatedEvent>`.
+        It also creates a default network binding on this endpoint. If you want
```
benhoyt, line 802: Should we say "To match Juju's behaviour, it also creates..."?

ops/testing.py
benhoyt: Should we not have `remote_relation` reverse this? WWJD (What Would Juju Do)?

PietroPasotti: huh, very good point.

PietroPasotti: I think that is a complicated choice because we'd have to distinguish between manually and automatically added networks, I guess it'd be confusing if you `.add_network`, then `.add_relation; .remove_relation` and then the network is gone. It does feel cleaner I suppose, but might be surprising.

ops/testing.py
```diff
@@ -859,6 +867,15 @@ def add_relation(self, relation_name: str, remote_app: str, *,
             if unit_data is not None:
                 self.update_relation_data(relation_id, remote_unit, unit_data)
 
+        # if we have a default network binding configured, respect it.
```
tonyandrewmeyer, line 870:
```suggestion
        # If we have a default network binding configured, respect it.
```

tonyandrewmeyer (approved):
LGTM! Thanks for all the work on this!

PR #56, opened 2019-11-26
Change Framework._observer to be a WeakValueDictionary

jameinel:
This means that setting up an observe() won't keep your object alive
indefinitely. This is safe, because we don't ever iterate the _observer
slice, and we already handled the case where we have a registered
observer_path but no matching observer.

PR #1142, opened 2024-03-01
fix(framework): add warning on lost observer weakref

PietroPasotti:
So this is a tricky one.

At times charmers are tempted to write:
```
        class MyObj(ops.Object):
            def __init__(self, charm, *args: typing.Any):
                super().__init__(charm, "obj")
                framework.observe(charm.on.start, self._on_start)

            def _on_start(self, _):
                raise RuntimeError()  # never reached!!! 


        class MyCharm(ops.CharmBase):
            def __init__(self, *args: typing.Any):
                super().__init__(*args)
                MyObj(self)  # not assigned!

            def _on_start(self, event: ops.EventBase):
                pass  # is reached
```

and expect that both _on_start methods will be triggered, while due to some subtle implementation details, MyObj._on_start never fires because the notice is silently dropped if the weakref stored in Framework._observer has been gc'd between charm instantiation time and Framework._emit time.

Long story short, I'm not sure why that codepath is so permissive and we don't error out altogether, that is, I don't see a legitimate reason why _observer is a weakref dict instead of a regular dict, as that predates me by a bunch. Perhaps @jameinel has some more context there?

For now, I'd propose adding a warning there so that people have something to go on with when that happens. We can also consider raising an exception, or making _observer a regular dict if those are viable alternatives.

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -60,12 +60,9 @@ class Serializable(typing.Protocol):
 
     @property
     def handle(self) -> 'Handle': ...  # noqa
-
     @handle.setter
     def handle(self, val: 'Handle'): ...
-
     def snapshot(self) -> Dict[str, Any]: ...  # noqa
-
     def restore(self, snapshot: Dict[str, Any]) -> None: ...  # noqa
 
 
@@ -948,7 +945,7 @@ class Framework(Object):
                     event_is_action = isinstance(event, charm.ActionEvent)
                     with self._event_context(event_handle.kind):
                         if (
-                                event_is_from_juju or event_is_action
+                            event_is_from_juju or event_is_action
                         ) and self._juju_debug_at.intersection({'all', 'hook'}):
                             # Present the welcome message and run under PDB.
                             self._show_debug_code_message()
@@ -1182,7 +1179,7 @@ class StoredState:
     def __get__(self,
                 parent: Optional['_ObjectType'],
                 parent_type: 'Type[_ObjectType]') -> Union['StoredState',
-    BoundStoredState]:
+                                                           BoundStoredState]:
         if self.parent_type is not None and self.parent_type not in parent_type.mro():
             # the StoredState instance is being shared between two unrelated classes
             # -> unclear what is expected of us -> bail out
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -60,9 +60,12 @@ class Serializable(typing.Protocol):
 
     @property
     def handle(self) -> 'Handle': ...  # noqa
+
     @handle.setter
     def handle(self, val: 'Handle'): ...
+
     def snapshot(self) -> Dict[str, Any]: ...  # noqa
+
     def restore(self, snapshot: Dict[str, Any]) -> None: ...  # noqa
 
 
@@ -945,7 +948,7 @@ class Framework(Object):
                     event_is_action = isinstance(event, charm.ActionEvent)
                     with self._event_context(event_handle.kind):
                         if (
-                            event_is_from_juju or event_is_action
+                                event_is_from_juju or event_is_action
                         ) and self._juju_debug_at.intersection({'all', 'hook'}):
                             # Present the welcome message and run under PDB.
                             self._show_debug_code_message()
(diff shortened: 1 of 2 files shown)


benhoyt:
Hmm, I don't understand why this should be a weakref either. I would have expected an `observe` call would keep the object alive so it can call it.

@jameinel put up [this PR](https://github.com/canonical/operator/pull/56) back in Nov 2019 that changed this from a regular dict to a weakref dict. It explains that it was done, and that "it's safe", but not really why it was done:

> This means that setting up an observe() won't keep your object alive
> indefinitely. This is safe, because we don't ever iterate the _observer
> slice, and we already handled the case where we have a registered
> observer_path but no matching observer.

John, can you remember (or think of) why we would have made this a weakref dict?

benhoyt:
In terms of this PR, if there's a good reason for this to be a weakref, I don't mind adding the warning, but I want to get to the bottom of why we use weakref at all first.

jameinel:
So we intentionally didn't want just a ref from observe() to be the thing that holds objects resident in memory. Both because it is a bit weird for TestObject() to actually have a side effect, but also because of how the actual event structure works.
We are taking the "Object Path" to figure out what to fire (especially in the case of something like defer()).
As such, we want to have a proper tree path where objects are referenced by their parent objects.

PietroPasotti:
> So we intentionally didn't want just a ref from observe() to be the thing that holds objects resident in memory. Both because it is a bit weird for TestObject() to actually have a side effect, but also because of how the actual event structure works. We are taking the "Object Path" to figure out what to fire (especially in the case of something like defer()). As such, we want to have a proper tree path where objects are referenced by their parent objects.

This sounds reasonable and makes sense to me. At the same time it feels like we shouldn't allow refs to simply disappear from `_observer` without as much as a warning. Unless registering children Objects without holding references to them is a legitimate use case for some other reason.

ops/framework.py
```diff
@@ -954,6 +954,14 @@ def _reemit(self, single_event_path: Optional[str] = None):
                             # Regular call to the registered method.
                             custom_handler(event)
 
+            else:
+                logger.warning(
+                    f"Weakref to Object at path {observer_path} has been killed somewhere "
+                    "between charm initialization and event emission. "
+                    "This can mean that you are using an Object which is not hard-referenced "
+                    "anywhere from the CharmBase tree. We won't be able to emit any events on it."
```
benhoyt, line 962: Suggestion to tighten up the message and make it a bit more accurate:

```suggestion
                    f"Reference to ops.Object at path {observer_path} has been garbage collected "
                    "between when the charm was initialised and when the event was emitted. "
                    "Make sure sure you store a reference to the observer."
```

test/test_charm.py
```diff
@@ -128,6 +128,30 @@ def _on_start(self, event: ops.EventBase):
         # check that the event has been seen by the observer
         self.assertIsInstance(charm.seen, ops.StartEvent)
 
+    def test_observe_nested_object_noninstantiated(self):
```
benhoyt, line 131:
```suggestion
    def test_observer_not_referenced_warning(self):
```

Also, should this test live in `test_framework.py`, as the relevant code lives in `framework.py`?

test/test_charm.py
```diff
@@ -128,6 +128,30 @@ def _on_start(self, event: ops.EventBase):
...
+                framework.observe(self.on.start, self._on_start)
+
+            def _on_start(self, _: ops.StartEvent):
+                pass  # is reached
+
+        framework = self.create_framework()
+        c = MyCharm(framework)
+        with self.assertLogs() as logs:
+            c.on.start.emit()
+        assert logs
```
benhoyt, line 153: In future there might be other logs too, so let's do something like this:

```suggestion
        assert any('Reference to ops.Object' in log for log in logs.output)
```

benhoyt (changes requested):
Some minor changes requested.

test/test_charm.py
PietroPasotti: I thought about it, but it felt way too complicated to implement it without referring to the charm object. If you look at the test immediately above it, I feel it makes sense to keep it here as they're very similar.
If you really want that that's a significant rework of the test.

benhoyt (approved):
Fair -- thanks!

Issue #1133, opened 2024-02-22
Consistently use "integrate" as the verb and "relation" as the noun in documentation


tonyandrewmeyer:
Our API documentation has various places that say "related to" that should be "integrated with" (and so on) so that we are consistent with the main documentation and with Juju. These should be adjusted, keeping "relation" as the noun - ie. we integrate two applications to create a relation.

We could also scan over the other documentation, although @tmihoc is generally handling that for this issue.

We should also discuss whether `TooManyRelatedAppsError` should be renamed `TooManyIntegratedAppsError` (with an alias for backwards compatibility).

PR #1145, opened 2024-03-08
docs: use 'integrate with' rather than 'relate to'

tonyandrewmeyer:
The current Juju terminology is that apps "integrate" to form a "relation". This PR tweaks the API documentation to align with that wording.

Fixes: #1133

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -507,9 +507,10 @@ class RelationEvent(HookEvent):
 class RelationCreatedEvent(RelationEvent):
     """Event triggered when a new relation is created.
 
-    This is triggered when a new relation to another app is added in Juju. This
+    This is triggered when a new integration with another app is added in Juju. This
     can occur before units for those applications have started. All existing
-    relations should be established before start.
+    relations will trigger `RelationCreatedEvent` before :class:`StartEvent` is
+    emitted.
     """
     unit: None  # pyright: ignore[reportIncompatibleVariableOverride]
     """Always ``None``."""
@@ -521,7 +522,7 @@ class RelationJoinedEvent(RelationEvent):
     This event is triggered whenever a new unit of a related
     application joins the relation.  The event fires only when that
     remote unit is first observed by the unit. Callback methods bound
-    to this event may set any local unit settings that can be
+    to this event may set any local unit data that can be
     determined using no more than the name of the joining unit and the
     remote ``private-address`` setting, which is always available when
     the relation is created and is by convention not deleted.
@@ -539,13 +540,13 @@ class RelationChangedEvent(RelationEvent):
     the callback method bound to this event.
 
     This event always fires once, after :class:`RelationJoinedEvent`, and
-    will subsequently fire whenever that remote unit changes its settings for
+    will subsequently fire whenever that remote unit changes its data for
     the relation. Callback methods bound to this event should be the only ones
-    that rely on remote relation settings. They should not error if the settings
-    are incomplete, since it can be guaranteed that when the remote unit or
-    application changes its settings, the event will fire again.
+    that rely on remote relation data. They should not error if the data
+    is incomplete, since it can be guaranteed that when the remote unit or
+    application changes its data, the event will fire again.
 
-    The settings that may be queried, or set, are determined by the relation's
+    The data that may be queried, or set, are determined by the relation's
     interface.
     """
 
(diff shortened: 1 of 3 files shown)


ops/charm.py
```diff
@@ -507,9 +507,10 @@ def restore(self, snapshot: Dict[str, Any]):
 class RelationCreatedEvent(RelationEvent):
     """Event triggered when a new relation is created.
 
-    This is triggered when a new relation to another app is added in Juju. This
+    This is triggered when a new integration with another app is added in Juju. This
     can occur before units for those applications have started. All existing
-    relations should be established before start.
+    relations will trigger `RelationCreatedEvent` before :class:`StartEvent` is
+    emitted.
```
tonyandrewmeyer, line 513: This is a drive-by: the existing wording is quite unclear (to me, at least). I think the revised wording is what is meant?

ops/charm.py
```diff
@@ -521,7 +522,7 @@ class RelationJoinedEvent(RelationEvent):
     This event is triggered whenever a new unit of a related
     application joins the relation.  The event fires only when that
     remote unit is first observed by the unit. Callback methods bound
-    to this event may set any local unit settings that can be
+    to this event may set any local unit data that can be
```
tonyandrewmeyer, line 525: This (and the below) are another drive-by: I only recall seeing "settings" meaning the content of the relation databag in the oldest documentation, and everywhere else in the API docs we refer to "data", so I believe this is much clearer.

ops/testing.py
```diff
@@ -795,7 +795,7 @@ def remove_storage(self, storage_id: str) -> None:
     def add_relation(self, relation_name: str, remote_app: str, *,
                      app_data: Optional[Mapping[str, str]] = None,
                      unit_data: Optional[Mapping[str, str]] = None) -> int:
-        """Declare that there is a new relation between this application and `remote_app`.
+        """Declare that there is a new integration between this application and `remote_app`.
```
tonyandrewmeyer, line 798:
```suggestion
        """Declare that there is a new relation between this application and `remote_app`.
```

ops/testing.py
```diff
@@ -1339,7 +1339,7 @@ def set_leader(self, is_leader: bool = True) -> None:
 
         If this charm becomes a leader then `leader_elected` will be triggered.  If :meth:`begin`
         has already been called, then the charm's peer relation should usually be added *prior* to
-        calling this method (with :meth:`add_relation`) to properly initialize and make
+        calling this method (with :meth:`add_relation`) to properly initialise and make
```
tonyandrewmeyer, line 1342: One last wee drive-by.

tonyandrewmeyer:
In #1133:

> We should also discuss whether TooManyRelatedAppsError should be renamed TooManyIntegratedAppsError (with an alias for backwards compatibility).

I think that this has too little value to outweigh the overhead of having a new name and keeping an alias for backwards compatibility.

Likewise, I think `Harness.add_relation` is fine without having a `Harness.integrate`.

ops/model.py
```diff
@@ -1865,8 +1865,8 @@ class MaintenanceStatus(StatusBase):
 class WaitingStatus(StatusBase):
     """A unit is unable to progress.
 
-    The unit is unable to progress to an active state because an application to which
-    it is related is not running.
+    The unit is unable to progress to an active state because an application with which
+    it is integrated is not running.
```
tonyandrewmeyer, line 1869: @benhoyt this seems very specific about the use-case for `WaitingStatus`. Isn't it ok to use it if you're waiting on *anything*, rather than specifically for an integration/relation?

benhoyt: That's taken straight from the comment in Juju [core/status/status.go](https://github.com/juju/juju/blob/55fb5d03683b70dea08d1e54687e0f90bb06a63a/core/status/status.go#L143-L146), so I presume it's intentional:

```
 	// The unit is unable to progress to an active state because an application to
	// which it is related is not running.
```

I think you'd use MaintenanceStatus otherwise.

benhoyt (approved):
Thanks. I think this is fine. I'm sure there are further improvements we could make, but this is a good start.

ops/model.py
tonyandrewmeyer: > That's taken straight from the comment in Juju [core/status/status.go](https://github.com/juju/juju/blob/55fb5d03683b70dea08d1e54687e0f90bb06a63a/core/status/status.go#L143-L146), so I presume it's intentional:

Ah, ok. It should presumably stay in that case.

> I think you'd use MaintenanceStatus otherwise.

Hmm, if I'm waiting for the container to be ready or waiting for an API to come up, `WaitingStatus` seems the natural one to use. `MaintenanceStatus` seems like it's more when the charm is actively doing something.

PR #1147, opened 2024-03-12
fix: inspect the correct signature when validating observe arguments

### LINKED ISSUE #1129: Unable to add additional arguments to event handlers
Consider the following simple charm:

```python
#!/usr/bin/env python3
# Copyright 2021 Ben
# See LICENSE file for licensing details.

"""Charm the service."""

import logging

from ops.charm import CharmBase
from ops.main import main
from ops.model import ActiveStatus, ModelError
from ops.pebble import ServiceStatus

def inject_charm_state(
    func: Callable[[CharmBase, EventBase], Any]
) -> Callable[..., Callable[[CharmBase, EventBase], Any]]:
    """Create a decorator that injects the charm_state into the charm instance, and removes it after."""

    @functools.wraps(func)
    def wrapper(charm: CharmBase, event: EventBase) -> Any:
        charm_state = CharmState.from_charm(charm)
    return func(charm, event, charm_state)

    return wrapper



class SnappassTestCharm(CharmBase):
    """Charm the service."""

    def __init__(self, *args):
        super().__init__(*args)
        self.framework.observe(
            self.on.snappass_pebble_ready, self._on_snappass_pebble_ready
        )
        self.framework.observe(self.on.redis_pebble_ready, self._on_redis_pebble_ready)

    def _on_snappass_pebble_ready(self, event):
        logger.info("_on_snappass_pebble_ready")
        container = self.unit.containers["redis"]

        if self._is_running(container, "redis"):
            logger.info("redis already started")
            self._start_snappass()

    def _start_snappass(self):
        logger.info("_start_snappass")
        container = self.unit.containers["snappass"]

        if self._is_running(container, "snappass"):
            logger.info("snappass already started")
            return

        snappass_layer = {
            "summary": "snappass layer",
            "description": "snappass layer",
            "services": {
                "snappass": {
                    "override": "replace",
                    "summary": "snappass service",
                    "command": "snappass",
                    "startup": "enabled",
                }
            },
        }

        container.add_layer("snappass", snappass_layer, combine=True)
        container.autostart()
        self.unit.status = ActiveStatus("snappass started")

    @inject_charm_state
    def _on_redis_pebble_ready(self, event, charm_state):
        charm_state = CharmState.from_charm(self)

    logger.info("_on_redis_pebble_ready")
        container = event.workload

        if self._is_running(container, "redis"):
            logger.info("redis already started")
            return

        redis_layer = {
            "summary": "redis layer",
            "description": "redis layer",
            "services": {
                "redis": {
                    "override": "replace",
                    "summary": "redis service",
                    "command": "redis-server",
                    "startup": "enabled",
                }
            },
        }

        container.add_layer("redis", redis_layer, combine=True)
        container.autostart()
        self.unit.status = ActiveStatus("redis started")
        # Prod snappass to start
        self._start_snappass()

    def _is_running(self, container, service):
        """Helper method to determine if a given service is running
        in a given container"""
        try:
            service = container.get_service(service)
        except ModelError:
            return False
        return service.current == ServiceStatus.ACTIVE


if __name__ == "__main__":
    main(SnappassTestCharm)
```

The `inject_charm_state` adds an additional `charm_state` argument to the event handler. `ops` currently will throw a runtime error (apologies don't have the error on hand) saying that the function is not a valid observer. As discussed, this should probably be done in a linting type of tool rather than as a runtime check. It would also be great if any linting tool will consider the modifications made by the decorator.

Also as discussed, I believe the following code in `ops` does this check: https://github.com/canonical/operator/blob/2f304d3af048edd758bee8b4b4d59f474b33a286/ops/framework.py#L811

tonyandrewmeyer:
When validating the `observe` arguments, inspect the signature of the method that we'll call, rather than digging down into any wrapped version, so that wrappers can adjust the signature as required, as long as the framework can still call them with the expected arguments.

This also fixes a bug where the check would fail to properly identify `*` and `**` as not required, so a signature like `def _on_x(self, event, *args, **kwargs)` would fail, even though the framework can use it to emit events.

Also remove the code that checks for `framework.observe(self.on.x, self)` style calls, which were possible in some old version of ops, and raises a specific error just for that (the general error is clear enough, and we don't believe anyone is still doing this).

When checking the type of the event handler, raise `TypeError` for errors, rather than a mixture of `RuntimeError` and `TypeError`.

Fixes #1129

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -800,32 +800,21 @@ class Framework(Object):
             raise TypeError(
                 f'event emitter {type(emitter).__name__} must have a "handle" attribute')
 
-        # Validate that the method has an acceptable call signature.
-        sig = inspect.signature(observer, follow_wrapped=False)
-        # Self isn't included in the params list, so the first arg will be the event.
-        extra_params = list(sig.parameters.values())[1:]
-
         method_name = observer.__name__
 
         assert isinstance(observer.__self__, Object), "can't register observers " \
                                                       "that aren't `Object`s"
         observer_obj = observer.__self__
-        if not sig.parameters:
+
+        # Validate that the method has an acceptable call signature.
+        sig = inspect.signature(observer, follow_wrapped=False)
+        # Self isn't included in the params list, so the first arg will be the event.
+        try:
+            sig.bind(EventBase(None))  # type: ignore
+        except TypeError as e:
             raise TypeError(
-                f'{type(observer_obj).__name__}.{method_name} must accept event parameter')
-        else:
-            # Allow for additional optional params, since there's no reason to exclude them, but
-            # required params will break.
-            required_params = [
-                param.name for param in extra_params
-                if param.default is inspect.Parameter.empty
-                and param.kind not in (inspect.Parameter.VAR_POSITIONAL,
-                                       inspect.Parameter.VAR_KEYWORD)
-            ]
-            if required_params:
-                raise TypeError(
-                    f'{type(observer_obj).__name__}.{method_name} has extra required '
-                    f'parameter: {", ".join(required_params)}')
+                f'{type(observer_obj).__name__}.{method_name} must be callable with '
+                "only 'self' and the 'event'") from e
 
         # TODO Prevent the exact same parameters from being registered more than once.
 
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -813,11 +813,17 @@ class Framework(Object):
         if not sig.parameters:
             raise TypeError(
                 f'{type(observer_obj).__name__}.{method_name} must accept event parameter')
-        elif any(param.default is inspect.Parameter.empty for param in extra_params):
+        else:
             # Allow for additional optional params, since there's no reason to exclude them, but
             # required params will break.
-            raise TypeError(
-                f'{type(observer_obj).__name__}.{method_name} has extra required parameter')
+            required_params = [
+                param.name for param in extra_params
+                if param.default is inspect.Parameter.empty
+            ]
+            if required_params:
+                raise TypeError(
+                    f'{type(observer_obj).__name__}.{method_name} has extra required '
+                    f'parameter: {", ".join(required_params)}')
 
         # TODO Prevent the exact same parameters from being registered more than once.
 
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -782,18 +782,12 @@ class Framework(Object):
             RuntimeError: if bound_event or observer are the wrong type.
         """
         if not isinstance(bound_event, BoundEvent):
-            raise RuntimeError(
+            raise TypeError(
                 f'Framework.observe requires a BoundEvent as second parameter, got {bound_event}')
         # Help users of older versions of the framework.
-        if isinstance(observer, charm.CharmBase):
+        if not isinstance(observer, types.MethodType):
             raise TypeError(
-                'observer methods must now be explicitly provided;'
-                ' please replace observe(self.on.{0}, self)'
-                ' with e.g. observe(self.on.{0}, self._on_{0})'.format(
-                    bound_event.event_kind))
-        if not callable(observer):
-            raise RuntimeError(
-                f'Framework.observe requires a callable as third parameter, got {observer}')
+                f"Framework.observe requires a method as the 'observer' parameter, got {observer}")
 
         event_type = bound_event.event_type
         event_kind = bound_event.event_kind
@@ -804,7 +798,7 @@ class Framework(Object):
         if hasattr(emitter, "handle"):
             emitter_path = emitter.handle.path
         else:
-            raise RuntimeError(
+            raise TypeError(
                 f'event emitter {type(emitter).__name__} must have a "handle" attribute')
 
         # Validate that the method has an acceptable call signature.
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -808,7 +808,6 @@ class Framework(Object):
 
         # Validate that the method has an acceptable call signature.
         sig = inspect.signature(observer, follow_wrapped=False)
-        # Self isn't included in the params list, so the first arg will be the event.
         try:
             sig.bind(EventBase(None))  # type: ignore
         except TypeError as e:
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -784,16 +784,16 @@ class Framework(Object):
         if not isinstance(bound_event, BoundEvent):
             raise RuntimeError(
                 f'Framework.observe requires a BoundEvent as second parameter, got {bound_event}')
-        if not isinstance(observer, types.MethodType):
-            # help users of older versions of the framework
-            if isinstance(observer, charm.CharmBase):
-                raise TypeError(
-                    'observer methods must now be explicitly provided;'
-                    ' please replace observe(self.on.{0}, self)'
-                    ' with e.g. observe(self.on.{0}, self._on_{0})'.format(
-                        bound_event.event_kind))
+        # Help users of older versions of the framework.
+        if isinstance(observer, charm.CharmBase):
+            raise TypeError(
+                'observer methods must now be explicitly provided;'
+                ' please replace observe(self.on.{0}, self)'
+                ' with e.g. observe(self.on.{0}, self._on_{0})'.format(
+                    bound_event.event_kind))
+        if not callable(observer):
             raise RuntimeError(
-                f'Framework.observe requires a method as third parameter, got {observer}')
+                f'Framework.observe requires a callable as third parameter, got {observer}')
 
         event_type = bound_event.event_type
         event_kind = bound_event.event_kind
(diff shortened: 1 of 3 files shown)


ops/framework.py
```diff
@@ -784,16 +784,16 @@ class SomeObject:
...
+        if isinstance(observer, charm.CharmBase):
+            raise TypeError(
+                'observer methods must now be explicitly provided;'
+                ' please replace observe(self.on.{0}, self)'
+                ' with e.g. observe(self.on.{0}, self._on_{0})'.format(
+                    bound_event.event_kind))
+        if not callable(observer):
             raise RuntimeError(
-                f'Framework.observe requires a method as third parameter, got {observer}')
+                f'Framework.observe requires a callable as third parameter, got {observer}')
```
tonyandrewmeyer, line 796: I made these changes while figuring out the lambda case (as noted in #1129, it's non-trivial to make that work, because we need the handler to be attached to an `Object`), and they aren't required to solve the actual problem.

I could therefore revert these changes to make the PR smaller, although I feel like this is more what we want - can you call the observer with the appropriate arguments - so I'm tempted to leave it. Please push back as needed :smile:.

benhoyt: Let's get rid of the "older versions" check (but ensure it would fail lower down), and revert `if not callable` to `if not MethodType`. Let's also make it `TypeError` rather than runtime error.

And let's avoid use of confusing "third" (it includes self) -- name the parameter instead.

test/test_charm.py
```diff
@@ -89,9 +89,6 @@ def _on_start(self, event: ops.EventBase):
 
         self.assertEqual(charm.started, True)
 
-        with self.assertRaisesRegex(TypeError, "observer methods must now be explicitly provided"):
-            framework.observe(charm.on.start, charm)  # type: ignore
-
```
benhoyt, line 94: Instead of removing the test, can we just make it `assertRaises(TypeError)`? It'll still raise due to the method test, right?

ops/framework.py
```diff
@@ -782,18 +782,12 @@ class SomeObject:
             RuntimeError: if bound_event or observer are the wrong type.
         """
         if not isinstance(bound_event, BoundEvent):
-            raise RuntimeError(
+            raise TypeError(
                 f'Framework.observe requires a BoundEvent as second parameter, got {bound_event}')
+        # Help users of older versions of the framework.
```
benhoyt, line 787: Nit: I think we should just drop this comment now, as I think it's more confusing than not having it.

test/test_framework.py
```diff
@@ -886,6 +887,24 @@ def _on_event(self, event: ops.EventBase):
                 'ObjectWithStorage[obj]/StoredStateData[_stored]',
                 'ObjectWithStorage[obj]/on/event[1]']))
 
+    def test_wrapped_handler(self):
+        def add_arg(func: typing.Callable[..., None]) -> typing.Callable[..., None]:
```
benhoyt, line 891: Not too terrible -- thanks for updating. :-)

benhoyt (changes requested):
Looking good, just a couple of nits.

ops/framework.py
tonyandrewmeyer: Ah, yeah, I just completely missed removing that line - I didn't intend for it to stay.

benhoyt (approved):
Thanks!

tonyandrewmeyer:
> Hmm, for some weird reason we're getting test failures on the observability charms (and not the issue we found yesterday). I think it's actually caused by this change, but I'm not sure why. Is it because the signature uses `_` for the event parameters? I didn't think `_` was special in Python though.
> 
> From https://github.com/canonical/operator/actions/runs/8270750559/job/22628988637?pr=1147
> 
> ```
> Traceback (most recent call last):
>   File "/usr/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
>     yield
>   File "/usr/lib/python3.10/unittest/case.py", line 587, in run
>     self._callSetUp()
>   File "/usr/lib/python3.10/unittest/case.py", line 546, in _callSetUp
>     self.setUp()
>   File "/usr/lib/python3.10/unittest/mock.py", line 1379, in patched
>     return func(*newargs, **newkeywargs)
>   File "/home/runner/work/operator/operator/tests/unit/test_charm.py", line 166, in setUp
>     self.harness.begin()
>   File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/ops/testing.py", line 366, in begin
>     self._charm = TestCharm(self._framework)  # type: ignore
>   File "/home/runner/work/operator/operator/lib/charms/tempo_k8s/v1/charm_tracing.py", line 287, in wrap_init
>     original_init(self, framework, *args, **kwargs)
>   File "/home/runner/work/operator/operator/src/charm.py", line 101, in __init__
>     self.server_cert = CertHandler(
>   File "/home/runner/work/operator/operator/lib/charms/observability_libs/v1/cert_handler.py", line 133, in __init__
>     self.framework.observe(
>   File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/ops/framework.py", line 819, in observe
>     raise TypeError(
> TypeError: CertHandler._on_config_changed has extra required parameter
> ```

Ah, nice catch - thanks. I was (badly) assuming that the failure was related to the issue found yesterday and so hadn't looked into it. I'll investigate.

tonyandrewmeyer:
> Ah, nice catch - thanks. I was (badly) assuming that the failure was related to the issue found yesterday and so hadn't looked into it. I'll investigate.

The reason for this is that with the change, we are looking at the correct signature, but we weren't correctly evaluating the signature all along.

With ops 2.11, this charm:

```python
class MyCharm(ops.CharmBase):
        super().__init__(framework)
        framework.observe(self.on.install, self._on_install)

    def _on_install(self, *args, **kwargs):
        # args will be (event, ), kwargs will be {} - that's fine!
        pass
```

fails with:

```
unit-badsigcheck-0: 17:43:20 ERROR unit.badsigcheck/0.juju-log Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-badsigcheck-0/charm/./src/charm.py", line 15, in <module>
    ops.main(MyCharm)
  File "/var/lib/juju/agents/unit-badsigcheck-0/charm/venv/ops/main.py", line 471, in __call__
    return main(charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-badsigcheck-0/charm/venv/ops/main.py", line 444, in main
    charm = charm_class(framework)
  File "/var/lib/juju/agents/unit-badsigcheck-0/charm/./src/charm.py", line 8, in __init__
    framework.observe(self.on.install, self._on_install)
  File "/var/lib/juju/agents/unit-badsigcheck-0/charm/venv/ops/framework.py", line 826, in observe
    raise TypeError(
TypeError: MyCharm._on_install has extra required parameter
```

What the existing code did was look at all the arguments (other than event, which is already examined) and check if they have a default value - if they did, then they weren't required, so were ok. However, `*args` and `*kwargs` don't have default values, but aren't required either, because they're of variable length, and that length can be zero. So we should skip those as well.

I've run this (updated) branch against 144 charms and there are no new failures compared to ops main, so I'm confident we should be good now.

@benhoyt suggested that `inspect` might have something to do the work for us, and there is the `bind()` method on the signature. I think that will be more readable (and leaves the work of getting it all correct to the core Python devs!), so am switching to that and am adding some additional tests for different types of argument.

ops/framework.py
```diff
@@ -804,27 +797,24 @@ class SomeObject:
...
-        if not sig.parameters:
-            raise TypeError(
-                f'{type(observer_obj).__name__}.{method_name} must accept event parameter')
-        elif any(param.default is inspect.Parameter.empty for param in extra_params):
-            # Allow for additional optional params, since there's no reason to exclude them, but
-            # required params will break.
+
+        # Validate that the method has an acceptable call signature.
+        sig = inspect.signature(observer, follow_wrapped=False)
+        # Self isn't included in the params list, so the first arg will be the event.
```
tonyandrewmeyer, line 811:
```suggestion
```

ops/framework.py
```diff
@@ -804,27 +797,23 @@ class SomeObject:
...
-            raise TypeError(
-                f'{type(observer_obj).__name__}.{method_name} must accept event parameter')
-        elif any(param.default is inspect.Parameter.empty for param in extra_params):
-            # Allow for additional optional params, since there's no reason to exclude them, but
-            # required params will break.
+
+        # Validate that the method has an acceptable call signature.
+        sig = inspect.signature(observer, follow_wrapped=False)
+        try:
+            sig.bind(EventBase(None))  # type: ignore
```
benhoyt, line 812: Ah, this is better than trying to do the finagling ourselves, thanks for trying this!

benhoyt (approved):
Excellent, thanks!

benhoyt:
Thanks for checking with super-tox too. I'm going to go ahead and merge this.

PR #1150, opened 2024-03-13
fix(harness): don't error out when attempting attaching storage before begin

### LINKED ISSUE #1127: Using Harness.add_storage(..., attach=True) before begin_with_initial_hooks gives confusing error
For example:

```python
import ops
import ops.testing

class MyCharm(ops.CharmBase):
    pass

meta = """
            containers:
                test-container:
                    mounts:
                        - storage: test-storage
                          location: /mounts/foo
            storage:
                test-storage:
                    type: filesystem
"""

h = ops.testing.Harness(MyCharm, meta=meta)
try:
    h.add_storage("test-storage", attach=True)
    h.begin_with_initial_hooks()
finally:
    h.cleanup()
```

If `begin()` is used instead of `begin_with_initial_hooks`, everything is ok, because `begin_with_initial_hooks` does the attaching (to then emit the `StorageAttached` event) and `begin` does not. Similarly, if `attach=True` is not used, everything is ok.

However, when run as above, an error like this is generated:

```
  File "/tmp/storagetest/test.py", line 41, in <module>
    h.begin_with_initial_hooks()
  File "/tmp/storagetest/.venv/lib/python3.11/site-packages/ops/testing.py", line 407, in begin_with_initial_hooks
    self.attach_storage(s.full_id)
  File "/tmp/storagetest/.venv/lib/python3.11/site-packages/ops/testing.py", line 752, in attach_storage
    if not self._backend._storage_attach(storage_id):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/storagetest/.venv/lib/python3.11/site-packages/ops/testing.py", line 2348, in _storage_attach
    mounting_dir.symlink_to(target_dir)
  File "/usr/lib/python3.11/pathlib.py", line 1199, in symlink_to
    os.symlink(target, self, target_is_directory)
FileExistsError: [Errno 17] File exists: '/tmp/ops-harness-z2tflymi/storages/test-storage/0' -> '/tmp/ops-harness-z2tflymi/containers/test-container/mounts/foo'
```

This happens because the `add_storage` call mounts the storage, and then the `begin_with_initial_hooks` attempts to mount it again, but that's already happened, so the symlinking fails.

We should:

* [ ] Adjust the API documentation so that it explicitly says that `attach` does nothing if `begin` hasn't yet been called
* [ ] Verify that `begin` has been called (ie. `._charm is not None` before trying to mount the storage

tonyandrewmeyer:
One minor fix to the `storage` metadata: [charmcraft.yaml](https://juju.is/docs/sdk/charmcraft-yaml#heading--storage) and [metadata.yaml](https://juju.is/docs/sdk/metadata-yaml#heading--storage) support a `multiple.range` value in the form `<n>+` - this would previously error out when we attempted to read the metadata. We now support this (as an alias for `<n>-`, which is an open-ended range rather than "n or less", and have a test for it.

Adds a minor clarification in the API docs when `Harness.add_storage` will emit a storage-attached event.

In `begin_with_initial_hooks`, if the storage has already been attached, then don't try to attach it a second time, but *do* emit the storage-attached event, since that will not have been done previously.

Make the `_storage_attach` method safe to call multiple times. This appears to have been the intention based on comments in the code, but, in practice, attempting to symlink the mount would fail. Now, if the symlink already exists (and has the same target) then we continue.

Adds a variety of tests for Harness `add_storage` and `attach_storage` behaviour.

Fixes #1127

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -748,8 +748,7 @@ class Harness(Generic[CharmType]):
         and is presently marked as detached.
 
         The test harness uses symbolic links to imitate storage mounts, which may lead to some
-        inconsistencies compared to the actual charm. Users should be cognisant of
-        this potential discrepancy.
+        inconsistencies compared to the actual charm.
 
         Args:
             storage_id: The full storage ID of the storage unit being attached, including the
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -690,8 +690,8 @@ class Harness(Generic[CharmType]):
         Args:
             storage_name: The storage backend name on the Charm
             count: Number of disks being added
-            attach: True to also attach the storage mount and emit storage-attached if
-                harness.begin() has been called.
+            attach: True to also attach the storage mount; if :meth:`begin`
+                has been called a True value will also emit storage-attached
 
         Return:
             A list of storage IDs, e.g. ["my-storage/1", "my-storage/2"].
@@ -739,21 +739,22 @@ class Harness(Generic[CharmType]):
         """Attach a storage device.
 
         The intent of this function is to simulate a ``juju attach-storage`` call.
+        If called before :meth:`begin`, or with hooks disabled, it will have no effect.
         It will trigger a storage-attached hook if the storage unit in question exists
         and is presently marked as detached.
 
         The test harness uses symbolic links to imitate storage mounts, which may lead to some
-        inconsistencies compared to the actual charm. Users should be cognizant of
+        inconsistencies compared to the actual charm. Users should be cognisant of
         this potential discrepancy.
 
         Args:
             storage_id: The full storage ID of the storage unit being attached, including the
                 storage key, e.g. my-storage/0.
         """
-        if not self._backend._storage_attach(storage_id):
-            return  # storage was already attached
         if not self._charm or not self._hooks_enabled:
             return  # don't need to run hook callback
+        if not self._backend._storage_attach(storage_id):
+            return  # storage was already attached
 
         storage_name, storage_index = storage_id.split('/', 1)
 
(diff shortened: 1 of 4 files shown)


ops/testing.py
```diff
@@ -739,11 +743,12 @@ def attach_storage(self, storage_id: str) -> None:
...
 
         The intent of this function is to simulate a ``juju attach-storage`` call.
-        It will trigger a storage-attached hook if the storage unit in question exists
+        If called after :meth:`begin` and hooks are not disabled, it will trigger
+        a storage-attached hook if the storage unit in question exists
         and is presently marked as detached.
 
         The test harness uses symbolic links to imitate storage mounts, which may lead to some
-        inconsistencies compared to the actual charm. Users should be cognizant of
+        inconsistencies compared to the actual charm. Users should be cognisant of
```
benhoyt, line 751: Thanks for the spelling fix. However, while we're here...

I am allergic to the phrase "cognisant of" (I had a coworker who'd use it in every other sentence at my previous workplace :-). "aware of" would be fine, however, in this case I think we should just drop this last sentence -- we've just made them aware of it in the previous sentence.

test/test_testing.py
```diff
@@ -4738,6 +4738,94 @@ def test_storage_mount(self):
...
+        # If `begin()` hasn't been called, `attach` does not emit storage-attached.
+        harness = ops.testing.Harness(MyCharm, meta=meta)
+        self.addCleanup(harness.cleanup)
+        harness.add_storage('test-storage', attach=True)
+        harness.begin()
+        self.assertNotIn('test-storage/0', harness.charm.attached)
+
+        # `attach` doesn't emit storage-attached before `begin_with_initial_hooks`.
+        harness = ops.testing.Harness(MyCharm, meta=meta)
+        self.addCleanup(harness.cleanup)
```
benhoyt, line 4773: It seems a bit weird to adding cleanups for multiple harness instances, which will all happen at the end. I don't think it's actually a problem (each has a random temp directory), but still feel a bit dirty. I think we should either just call `harness.cleanup()` manually after each, or probably better, have one test per harness / thing we're checking (but re-use the charm and metadata).

benhoyt (changes requested):
Great, thanks for this. Nit and comment about test structure.

IronCore864 (approved):

ops/testing.py
tonyandrewmeyer: :smile: I considered just replacing it with "aware", which is much more understandable. Dropping sounds good to me.

test/test_testing.py
tonyandrewmeyer: Manually after each really ought to be done in finally blocks since it's possible the test fails and we don't want to leave things around in that case. Splitting into separate tests is fine; I'll do that.

IronCore864:
Great test cases, I've learned a lot :)

benhoyt (approved):
Very clean, thanks!

Issue #1040, opened 2023-10-10
See if Relation.app can be non-Optional


benhoyt:
It's currently typed as `Optional[Application]`, but it looks like we try very hard to set it, falling back to using `JUJU_REMOTE_APP` if we can't get it via relation data. Need to investigate whether it could ever be None, and whether we can change the type annotation to plain `Application`.

Otherwise charmers that using type checking have to add these annoying assertions:

```python
    @property
    def host(self) -> str:
        """Fetch the SMTP host from the relation."""
        assert self.relation.app
        return self.relation.data[self.relation.app].get("host")

    @property
    def port(self) -> int:
        """Fetch the SMTP port from the relation."""
        assert self.relation.app
        return int(self.relation.data[self.relation.app].get("port"))
```

tonyandrewmeyer:
One way to have it be `None` (this was microk8s on Juju 3.1.6, with ops 2.9):

1. Add app 1 and app2 and integrate
2. Scale app 1 to multiple units
3. Scale app 1 down a unit

There's a relation-departed event for the departing unit, and [a relation-broken event for the departing unit](https://bugs.launchpad.net/juju/+bug/1979811). In that relation-broken event, `event.relation.app` is currently None.

tonyandrewmeyer:
> One way to have it be `None` (this was microk8s on Juju 3.1.6, with ops 2.9):
> 
> 1. Add app 1 and app2 and integrate
> 2. Scale app 1 to multiple units
> 3. Scale app 1 down a unit
> 
> There's a relation-departed event for the departing unit, and [a relation-broken event for the departing unit](https://bugs.launchpad.net/juju/+bug/1979811). In that relation-broken event, `event.relation.app` is currently None.

However, I was unable to reproduce this :disappointed:.

The conclusion (see #1128) is that JUJU_REMOTE_APP is always set (and if it is not, it should be), so it does seem safe to make `app` non-optional.

PR #1151, opened 2024-03-13
fix(model): change model.relation.app type from optional to mandatory

IronCore864:
Closes https://github.com/canonical/operator/issues/1040.

Tests:

```bash
TOTAL                       6389    500   2386    209    90%
  unit: OK (247.53=setup[0.02]+cmd[246.92,0.59] seconds)
  congratulations :) (247.56 seconds)
```

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1478,7 +1478,7 @@ class Relation:
             backend: '_ModelBackend', cache: '_ModelCache', active: bool = True):
         self.name = relation_name
         self.id = relation_id
-        self.app: Application = None
+        self.app: Application = typing.cast(Application, None)
         self.units: Set[Unit] = set()
         self.active = active
 
--- a/ops/model.py
+++ b/ops/model.py
@@ -1448,7 +1448,7 @@ class Relation:
     id: int
     """The identifier for a particular relation."""
 
-    app: Optional[Application]
+    app: Application
     """Represents the remote application of this relation.
 
     For peer relations, this will be the local application.
@@ -1478,7 +1478,7 @@ class Relation:
             backend: '_ModelBackend', cache: '_ModelCache', active: bool = True):
         self.name = relation_name
         self.id = relation_id
-        self.app: Optional[Application] = None
+        self.app: Application = None
         self.units: Set[Unit] = set()
         self.active = active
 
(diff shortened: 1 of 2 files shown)


ops/model.py
```diff
@@ -1478,7 +1478,7 @@ def __init__(
             backend: '_ModelBackend', cache: '_ModelCache', active: bool = True):
         self.name = relation_name
         self.id = relation_id
-        self.app: Optional[Application] = None
+        self.app: Application = typing.cast(Application, None)
```
tonyandrewmeyer, line 1481: Hmm, this gets a bit ugly, since we're lying about the type. What about having a local `app` variable (with an implicit type of `Application|None`) and then a `self.app = app` at the end? Probably with an `assert app is not None` first to make type checkers happy.

tonyandrewmeyer (approved):
Sometimes nice improvements can be very small 😄. Looks good, although I wonder if we could avoid the mis-typing in `__init__`.

ops/model.py
IronCore864: After some discussion and according to @tonyandrewmeyer and @benhoyt:
- `self.app` will always be set because we have a fallback mechanism to get it from env vars and stuff.
- In case of a weird Juju bug or edge case where the app is actually None, we probably do not want the charm to raise an AssertionError.

So, we will use typing.cast and add a comment to explain that it will not be None.

ops/model.py
```diff
@@ -1478,32 +1478,32 @@ def __init__(
...
         # If we didn't get the remote app via our_unit.app or the units list,
         # look it up via JUJU_REMOTE_APP or "relation-list --app".
-        if self.app is None:
+        if app is None:
             app_name = backend.relation_remote_app_name(relation_id)
             if app_name is not None:
-                self.app = cache.get(Application, app_name)
+                app = cache.get(Application, app_name)
 
+        # self.app will not be None and always be set because of the fallback mechanism below.
```
benhoyt, line 1505: Nit: it's "above" now:

```suggestion
        # self.app will not be None and always be set because of the fallback mechanism above.
```

benhoyt (commented):
Thanks, looks good to me, just one nit in a comment.

tonyandrewmeyer (approved):
Looks great (other than the comment nit Ben already noted). Thanks!

benhoyt (approved):
Thanks!

Issue #186, opened 2020-03-16
Support credential-get hook tool


dshcherb:
`credential-get` hook tool is used by k8s cloud integrator charms (OpenStack, AWS, GCP, Azure etc.) to get access to cloud credentials via a Juju controller.

https://jaas.ai/docs/deploying-advanced-applications#heading--trusting-an-application-with-a-credential
https://github.com/juju/juju/blob/juju-2.7.4/worker/uniter/runner/jujuc/credential-get.go

E.g. OpenStack Integrator:
https://github.com/juju-solutions/charm-openstack-integrator/blob/101913504d047a1b4b9b6c4dcfb901d66fd9399e/lib/charms/layer/openstack.py#L38-L70

This needs to be modeled in the framework so that this class of charms can be written with the operator framework.

benhoyt:
This is somewhat related to https://github.com/canonical/operator/issues/532 where "is my charm trusted" isn't modelling in ops.

benhoyt:
This definitely seems worth modelling in ops to avoid people implementing it themselves. Tony linked to a [GitHub search](https://github.com/search?q=%27credential-get%27&type=code) showing a bunch of places that charms use `credential-get` (including the charm-openstack-integrator charm linked above).

We'd probably want to make the return type a dataclass of some kind, but there's a map/dict of cloud-specific attributes that we'd keep as a dict (don't want to go to the level of having cloud-specific code in ops just to get better typing).

We'll look at doing this in 24.04.

IronCore864:
I have done some research and found that:

1. [in juju only `type` and `name` are mandatory, others can be empty](https://github.com/juju/juju/blob/3.5/rpc/params/cloud.go#L194).

2. I also checked [an AWS charm](https://github.com/Automatic/charm-aws-integrator/blob/a9ab3729107debb220293187970c74f5af80e69b/lib/charms/layer/aws.py#L59) and [an Azure charm](https://github.com/phvalguima/charm-azure-integrator/blob/479bacacb0762128cd1b667acf78d3eb5e8b4883/lib/charms/layer/azure.py#L66), it seems they have different keys inside the `credential` section.

3. Then I did some testing on EC2 in AWS with LXD localhost cloud, and it returned `attrs` inside the `credential` section, unlike `attributes` in one of the charms listed above. Example:

```json
{
    "type": "lxd",
    "name": "localhost",
    "region": "localhost",
    "endpoint": "https://10.76.251.1:8443",
    "credential": {
[...]

PR #1152, opened 2024-03-15
feat(Model): support credential-get hook tool in both model and harness

IronCore864:
Added `Model.get_cloud_spec` which uses the `credential-get` hook tool to get details of the cloud where the model is deployed.

Changes are made to both `ops.Model` and Harness.

Closes https://github.com/canonical/operator/issues/186.


IronCore864:
Sample usage:

```python
class MyVMCharm(ops.CharmBase):
    """Charm the application."""

        super().__init__(framework)

        cloud_spec = self.model.get_cloud_spec()
        print(f'Credential: {cloud_spec!r}')
```

ops/model.py
```diff
@@ -3507,6 +3521,17 @@ def reboot(self, now: bool = False):
...
 
+    def credential_get(self) -> 'CloudSpec':
+        """Access cloud credentials by running the credential-get hook tool.
+
+        Returns the cloud specification used by the unit's model.
+        """
+        try:
+            result = self._run('credential-get', return_output=True, use_json=True)
+            return CloudSpec.from_dict(typing.cast(Dict[str, Any], result))
+        except ModelError:
```
benhoyt, line 3532: What does this `except` and re-raise do?

ops/model.py
```diff
@@ -3596,3 +3621,60 @@ def _ensure_loaded(self):
         self._notice = self._container.get_notice(self.id)
         assert self._notice.type == self.type
         assert self._notice.key == self.key
+
+
+class CloudSpec:
```
benhoyt, line 3626: We've been starting to use the [`@dataclasses.dataclass`](https://docs.python.org/3/library/dataclasses.html) decorator, which writes the `__init__` and `__repr__` and other things for us (probably with `frozen=True`). We'd just need the `from_dict` method.

ops/model.py
```diff
@@ -3596,3 +3621,60 @@ def _ensure_loaded(self):
...
+    """Cloud specification information (metadata) including credentials."""
+
+    def __init__(self,
+                 type: str,
+                 name: str,
+                 region: Optional[str],
+                 endpoint: Optional[str],
+                 is_controller_cloud: Optional[str],
+                 credential: Optional[Dict[str, Any]],
```
benhoyt, line 3635: I had suggested to make this a new `CloudCredential` dataclass, as it has a few other fields than just the credential attributes. See my comments about "nested object" here: https://github.com/canonical/operator/issues/186#issuecomment-1998442238 ... any reason we shouldn't do that to match the "shape" of the Juju object?

ops/testing.py
```diff
@@ -1890,6 +1890,42 @@ def run_action(self, action_name: str,
...
         return action_under_test.output
 
+    def set_cloud_spec(self, content: Dict[str, str]):
+        """Set cloud specification (metadata) including credentials.
+
+        Call this method before trying to call `Harness.Model.get_cloud_spec`.
+
+        Example usage::
+
+            class TestCharm(unittest.TestCase):
```
benhoyt, line 1900: We're encouraging charm tests to be written with pytest, so we should use that in examples now.

ops/testing.py
```diff
@@ -1890,6 +1890,42 @@ def run_action(self, action_name: str,
                 output=action_under_test.output)
         return action_under_test.output
 
+    def set_cloud_spec(self, content: Dict[str, str]):
```
benhoyt, line 1893: Any reason this shouldn't take a proper `CloudSpec` object? They can always use `CloudSpec.from_dict` is they have a dict.

benhoyt (commented):
Thanks, this is a good start! I haven't gone over it with a fine-toothed comb yet, but just a few initial comments. @tonyandrewmeyer will review as well.

docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.12
```
tonyandrewmeyer, line 2: Do these still work with 3.8? This might be missing from our docs, but we generally need to update the requirements with the readthedocs version (or maybe that is 3.12 now?), to make sure we aren't picking up something not available. Having three dependencies vanish makes me wonder if they are only needed in older Python.

ops/model.py
```diff
@@ -281,6 +281,20 @@ def get_secret(self, *, id: Optional[str] = None, label: Optional[str] = None) -
...
         return Secret(self._backend, id=id, label=label, content=content)
 
+    def get_cloud_spec(self) -> 'CloudSpec':
+        """Get cloud credential information.
+
+        Access the cloud credential information and return the cloud specification
+        used by the model.
+
+        Note: It does not work in a "CaaS" model (Container-as-a-Service,
+        i.e., containerized environment, Kubernetes charm).
```
tonyandrewmeyer, line 291: I know the Juju error refers to "CaaS" but from what I can see there's very little use of this term in the Juju docs. I'm not sure what the state is here - is this a term that's being adopted, or an old one that's going away, or just one that is only occasionally used? If @benhoyt doesn't know, we could ask the Juju team.

It seems like a simpler version using the names we generally use might work better, like:

```suggestion
        Note: This information is only available for machine charms, not Kubernetes sidecar charms.
```

Similarly in the "Raises" section, I would say something like "Kubernetes model" rather than "CaaS".

ops/model.py
```diff
@@ -281,6 +281,20 @@ def get_secret(self, *, id: Optional[str] = None, label: Optional[str] = None) -
         content = self._backend.secret_get(id=id, label=label)
         return Secret(self._backend, id=id, label=label, content=content)
 
+    def get_cloud_spec(self) -> 'CloudSpec':
+        """Get cloud credential information.
+
+        Access the cloud credential information and return the cloud specification
+        used by the model.
```
tonyandrewmeyer, line 288: Maybe refocus this on the cloud specification rather than credentials? Something like:

```suggestion
        """Get details of the cloud in which the model is deployed.

        Returns a specification for the cloud in which the model is deployed, including access credential information.
```

ops/model.py
```diff
@@ -3507,6 +3521,17 @@ def reboot(self, now: bool = False):
         else:
             self._run("juju-reboot")
 
+    def credential_get(self) -> 'CloudSpec':
+        """Access cloud credentials by running the credential-get hook tool.
+
+        Returns the cloud specification used by the unit's model.
```
tonyandrewmeyer, line 3527: I think we can just say:

```suggestion
        Returns the cloud specification used by the model.
```

ops/testing.py
```diff
@@ -1890,6 +1890,42 @@ def run_action(self, action_name: str,
...
+        """Set cloud specification (metadata) including credentials.
+
+        Call this method before trying to call `Harness.Model.get_cloud_spec`.
```
tonyandrewmeyer, line 1896: This should be a ReST link, like this but I'm not sure if I have the reference totally right so please check that :smile: :

```suggestion
        Call this method before trying to call :meth:`Model.get_cloud_spec`.
```

ops/testing.py
tonyandrewmeyer: Hmm, this will be inconsistent with the rest of the docs, though. Maybe have this be unittest to match the rest, and we open a ticket (or add to the existing one) changing the API docs? It seems like a fairly small job and I would rather see them all be using the same style.

ops/testing.py
```diff
@@ -2105,6 +2141,8 @@ def __init__(self, unit_name: str, meta: charm.CharmMeta, config: '_RawConfig'):
         self._networks: Dict[Tuple[Optional[str], Optional[int]], _NetworkDict] = {}
         self._reboot_count = 0
         self._running_action: Optional[_RunningAction] = None
+        # for `Model.get_cloud_spec`, initialized to None
```
tonyandrewmeyer, line 2144: Nit (full sentence including punctuation, UK spelling):

```suggestion
        # For `Model.get_cloud_spec`, initialised to None.
```

ops/testing.py
```diff
@@ -2678,6 +2716,12 @@ def reboot(self, now: bool = False):
         # to handle everything after the exit.
         raise SystemExit()
 
+    def credential_get(self) -> model.CloudSpec:
+        if not self._cloud_spec:
+            raise model.ModelError(
+                'ERROR cloud spec is empty, set it with Harness.set_cloud_spec first')
```
tonyandrewmeyer, line 2722:
```suggestion
                'ERROR cloud spec is empty, set it with `Harness.set_cloud_spec()` first')
```

test/test_model.py
```diff
@@ -3791,5 +3791,90 @@ def test_repr(self):
...
+        self.credential = {
+            'auth-type': 'certificate',
+            'attrs': {
+                'client-cert': 'foo',
+                'client-key': 'bar',
+                'server-cert': 'baz'
+            },
+        }
+
+    def test_init(self):
```
tonyandrewmeyer, line 3805: For what it's worth, if the class becomes a dataclass, we probably don't need this test.

ops/testing.py
```diff
@@ -1890,6 +1890,42 @@ def run_action(self, action_name: str,
...
+            class TestCharm(unittest.TestCase):
+                def setUp(self):
+                    self.harness = ops.testing.Harness(MyVMCharm)
+                    self.addCleanup(self.harness.cleanup)
+
+                def test_start(self):
+                    self.harness.set_cloud_spec(
+                        {"name": "lxd", "type": "lxd", "endpoint": "https://127.0.0.1:8443"}
+                    )
+                    self.harness.begin_with_initial_hooks()
```
tonyandrewmeyer, line 1909: The thing about this example is that it doesn't really do anything with the spec - I'm pretty sure if you took out the `begin_with_initial_hooks` it would still work, because no events are actually involved.

Maybe it's ok since this is just in the API docs and we don't want those to be huge, but could we have the charm with a trivial start handler that uses the spec? I don't think tests would ever want to get the spec, because the charm can't change it, so it'll only ever be what the test gave it.

Something like:

```python
class MyVMCharm(ops.CharmBase):
    ...
        spec = self.model.get_cloud_spec()
        ....
```

And then remove the `expected=` and last two asserts from the test, having just a "..." or something instead (testing whatever the start would actually do with the spec).

tonyandrewmeyer (changes requested):
Looking great! A few little suggestions, particularly around documentation.

docs/requirements.txt
IronCore864: I tested with 3.8 and it worked, but to reduce the scope of this PR I'm reverting this back to 3.8.

ops/testing.py
IronCore864: @benhoyt should we create a GitHub issue/Jira ticket for this?

ops/model.py
IronCore864: I did not fully copy from Juju because I'm a bit confused here: In Juju [`CloudCredential.Attributes` reads from json `attrs`](https://github.com/juju/juju/blob/3.5/rpc/params/cloud.go#L153), but in [AWS charm](https://github.com/Automatic/charm-aws-integrator/blob/a9ab3729107debb220293187970c74f5af80e69b/lib/charms/layer/aws.py#L59) and [Azure charm](https://github.com/phvalguima/charm-azure-integrator/blob/479bacacb0762128cd1b667acf78d3eb5e8b4883/lib/charms/layer/azure.py#L66) it seems they are reading from "attributes" instead of "attrs". I do not know if the cloud credentials in other clouds are of different structure, so to be on the safer end I used a dictionary for it.

Should I create an object CloudCredential to match Juju anyway?

benhoyt: The Juju .go code is almost certainly the source of truth, because that's what's in Juju. Maybe in older versions the JSON field was called "attributes"? I'm not sure why the discrepancy between Juju and what the charm is doing. Can you see what shape of object / field names the current Juju version returns?

ops/testing.py
benhoyt: Yep, that sounds good, thanks @IronCore864. Maybe you can assign that new issue to yourself to do later. Should be a fairly quick fix to our examples.

IronCore864: https://github.com/canonical/operator/issues/1154

ops/model.py
IronCore864: I did some testing on AWS EC2 + LXD:

If I run juju CLI directly to show-crendential the result structure is quite diffrent, like:

```bash
$ juju show-credential aws default
client-credentials:
  aws:
    default:
      content:
        auth-type: access-key
        access-key: xxx
```

If I run the hook tool, what I got is similar as what the AWS and Azure charms have:

```json
credential:
  authtype: certificate
  attributes:
    client-cert: foo
    client-key: bar
    server-cert: baz
  redacted: []
```

Does this mean we should create a `CloudCredential` dataclass, but instead of using `attrs`, we use `attributes`?

benhoyt: Aha, I've figured out the discrepancy: the default formatting is YAML, and those structs don't have yaml tags, so the yaml marshalling uses the default field name (lowercased version of the Go field name in the struct). So let's use `credential-get --format=json` -- you can call `_ModelBackend._run(use_json=True)` to do that -- and that way we'll get the explicitly defined JSON field names: auth-type, attrs, redacted, etc.

So yes, use `--format=json`, copy from the Juju struct fields json:"..." field name, and add CloudSpec and CloudCredential dataclasses.

IronCore864: Got it. It's using `--format=json` already, see[here](https://github.com/canonical/operator/pull/1152/files#diff-b7713b8731a4d1f895063e58f31f8ab54f7bb6ca0e2b1712d85683874d94a6e6R3862).

I'll mark it as resolved for now.

IronCore864:
All comments resolved, please take another look.

ops/testing.py
```diff
@@ -1911,6 +1911,39 @@ def run_action(self, action_name: str,
                 output=action_under_test.output)
         return action_under_test.output
 
+    def set_cloud_spec(self, spec: 'model.CloudSpec'):
+        """Set cloud specification (metadata) including credentials.
+
+        Call this method before trying to call :meth:`ops.Model.get_cloud_spec`.
```
benhoyt, line 1917:
```suggestion
        Call this method before the charm calls :meth:`ops.Model.get_cloud_spec`.
```

ops/testing.py
```diff
@@ -1911,6 +1911,39 @@ def run_action(self, action_name: str,
...
+
+        Example usage::
+
+            class MyVMCharm(ops.CharmBase):
+                ...
+                def _on_start(self, event: ops.StartEvent):
+                    spec = self.model.get_cloud_spec()
+                    ....
```
benhoyt, line 1925: How about we remove the `...` here, as I think it's very close to a fully working example. If we say `self.spec = self.model.get_cloud_spec()`, then we can `assert self.harness.charm.spec == expected` in the test, to make it a bit more realistic.

ops/testing.py
```diff
@@ -1911,6 +1911,39 @@ def run_action(self, action_name: str,
...
+                    self.harness = ops.testing.Harness(MyVMCharm)
+                    self.addCleanup(self.harness.cleanup)
+
+                def test_start(self):
+                    cloud_spec_dict = {
+                        "name": "localhost",
+                        "type": "lxd",
+                        "endpoint": "https://127.0.0.1:8443"
+                    }
+                    self.harness.set_cloud_spec(ops.model.CloudSpec.from_dict(cloud_spec_dict))
```
benhoyt, line 1939:
```suggestion
                    self.harness.set_cloud_spec(ops.CloudSpec.from_dict(cloud_spec_dict))
```

ops/testing.py
```diff
@@ -2126,6 +2159,8 @@ def __init__(self, unit_name: str, meta: charm.CharmMeta, config: '_RawConfig'):
         self._networks: Dict[Tuple[Optional[str], Optional[int]], _NetworkDict] = {}
         self._reboot_count = 0
         self._running_action: Optional[_RunningAction] = None
+        # For `Model.get_cloud_spec`, initialised to None.
```
benhoyt, line 2162: I think this is pretty obvious, and "initialized to None" is what `= None` does, so IMO we can just remove this comment.

test/test_model.py
```diff
@@ -3791,5 +3791,82 @@ def test_repr(self):
         )
 
 
+class TestCloudCredential(unittest.TestCase):
+    def setUp(self) -> None:
+        self.cloud_credential_dict = {
```
benhoyt, line 3796: I think this is better just as a local variable inside `test_from_dict`, as it reduces the scope and it's only used by that test. Could probably shorten the name to `dct` or `d` in that case. Let's also include two tests: one with only the required fields, and one with all fields (including `redacted`).

test/test_model.py
```diff
@@ -3791,5 +3791,82 @@ def test_repr(self):
...
+
+    def test_from_dict(self):
+        cloud_cred = ops.CloudCredential.from_dict(self.cloud_credential_dict)
+        self.assertEqual(cloud_cred.auth_type, 'certificate')
+        self.assertEqual(cloud_cred.attributes, self.cloud_credential_dict.get('attrs'))
+
+
+class TestCloudSpec(unittest.TestCase):
+    def setUp(self) -> None:
+        self.cloud_credential_dict = {
```
benhoyt, line 3813: Similar here, let's just make this local to the test, and again have two tests, one with minimal required fields (optional fields not present rather than None), and one with all fields set.

test/test_model.py
```diff
@@ -3791,5 +3791,82 @@ def test_repr(self):
...
+            cloud_spec.credential,
+            ops.CloudCredential.from_dict(
+                self.cloud_credential_dict))
+
+
+class TestGetCloudSpec(unittest.TestCase):
+    def setUp(self):
+        self.model = ops.Model(ops.CharmMeta(), _ModelBackend('myapp/0'))
+
+    def test_get_cloud_spec(self):
```
benhoyt, line 3856: Seeing this is in `TestGetCloudSpec`, it could just be called `test_success`, and the other one `test_error`.

test/test_model.py
```diff
@@ -3791,5 +3791,82 @@ def test_repr(self):
...
+        fake_script(self, 'credential-get', """echo '{"type": "lxd", "name": "localhost"}'""")
+        cloud_spec = self.model.get_cloud_spec()
+        self.assertEqual(cloud_spec.type, 'lxd')
+        self.assertEqual(cloud_spec.name, 'localhost')
+        self.assertEqual(fake_script_calls(self, clear=True),
+                         [['credential-get', '--format=json']])
+
+    @patch("ops.model._ModelBackend.credential_get")
```
benhoyt, line 3864: I think we should do this one without mocking too, just like the above. Have the script `exit 1` or similar (there should be other similar examples in `test_model.py`).

test/test_testing.py
```diff
@@ -5854,3 +5854,34 @@ def test_get_notices(self):
 class TestNotices(unittest.TestCase, _TestingPebbleClientMixin, PebbleNoticesMixin):
     def setUp(self):
         self.client = self.get_testing_client()
+
+
+class TestCloudSpec(unittest.TestCase):
+    def test_get_set_cloud_spec(self):
```
benhoyt, line 5860: Should this be `test_set_cloud_spec`, or similar to above, just `test_set`, given the context is clear from the class name?

test/test_testing.py
```diff
@@ -5854,3 +5854,34 @@ def test_get_notices(self):
...
+        harness = ops.testing.Harness(EventRecorder, meta='name: myapp')
```
benhoyt, line 5861: Does this need to inherit from `EventRecorder` (does it use anything from that), or could it just be `CharmBase` directly?

test/test_testing.py
```diff
@@ -5854,3 +5854,34 @@ def test_get_notices(self):
...
+            'credential': {
+                'authtype': 'certificate',
+                'attrs': {
+                    'client-cert': 'foo',
+                    'client-key': 'bar',
+                    'server-cert': 'baz'
+                },
+            },
+        }
+        harness.set_cloud_spec(ops.model.CloudSpec.from_dict(cloud_spec_dict))
```
benhoyt, line 5876: Could use the `CloudSpec` constructor directly here with just the required fields (we've already tested all fields in `test_model.py`). Also, prefer `ops.CloudSpec` over `ops.model.CloudSpec`. So something like:

```python
harness.set_cloud_spec(ops.CloudSpec(name='localhost', type='lxd'))
```

Also, it seems to me this isn't the kind of test charmers would actually do. Why don't we basically reproduce the test from your docstring example earlier, which creates a little charm with on_start and sets a `self.spec` attribute that we assert on.

test/test_testing.py
```diff
@@ -5854,3 +5854,34 @@ def test_get_notices(self):
...
+        harness.begin()
+        result = harness.model.get_cloud_spec()
+        expected = ops.model.CloudSpec.from_dict(cloud_spec_dict)
+        self.assertEqual(repr(result), repr(expected))
```
benhoyt, line 5880: Note that we should be asserting on the `repr` here, but the actual object (dataclasses define equality).

test/test_testing.py
```diff
@@ -5854,3 +5854,34 @@ def test_get_notices(self):
...
+
+    def test_get_cloud_spec_without_set_error(self):
+        harness = ops.testing.Harness(EventRecorder, meta='name: myapp')
```
benhoyt, line 5883:
```suggestion
        harness = ops.testing.Harness(ops.CharmBase, meta='name: myapp')
```

benhoyt (changes requested):
Good progress here, thanks. A whole bunch of comments, mostly very minor, but also a bit of fixes to do with the field names and types from the Juju structs.

ops/model.py
```diff
@@ -3596,3 +3618,56 @@ def _ensure_loaded(self):
...
+
+@dataclasses.dataclass(frozen=True)
+class CloudSpec:
+    """Cloud specification information (metadata) including credentials."""
+
+    type: str
+    name: str
+    region: Optional[str]
+    endpoint: Optional[str]
+    is_controller_cloud: Optional[str]
```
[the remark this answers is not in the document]
IronCore864: In Juju, [`CloudSpec.IsControllerCloud` has the `omitempty` tag](https://github.com/juju/juju/blob/3.5/rpc/params/cloud.go#L204).

Shall we keep it as `Optional[bool]`, and in the `from_dict` method, we try something like `is_controller_cloud=d.get('isControllerCloud') or False`?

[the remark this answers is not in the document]
IronCore864: I see your point, we don't want optional bool to be None. Non-optional bool defaults to False is better.

ops/model.py
```diff
@@ -3596,3 +3618,56 @@ def _ensure_loaded(self):
...
+
+    auth_type: str
+    attributes: Dict[str, Any]
+    redacted: List[str]
+
+    @classmethod
+    def from_dict(cls, d: Dict[str, Any]) -> 'CloudCredential':
+        """Create a new CloudCredential object from a dictionary."""
+        return cls(
+            auth_type=d.get('auth-type', ''),
```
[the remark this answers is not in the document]
IronCore864: For `CloudCredential`, it seems this field is required. However, `CloudSpec.credential` has tag `omitempty`, which means the credential could be completely empty, under which case, `d['auth-type']` would error. I will keep `d.get` to be on the safer end.

ops/model.py
```diff
@@ -3596,3 +3619,56 @@ def _ensure_loaded(self):
         self._notice = self._container.get_notice(self.id)
         assert self._notice.type == self.type
         assert self._notice.key == self.key
+
+
+@dataclasses.dataclass(frozen=True)
+class CloudCredential:
+    """CloudCredential contains a cloud credential possibly with secrets redacted.
```
tonyandrewmeyer, line 3626: This reads oddly to me. Maybe split off the second bit? Or maybe you can think of a different way to word this?

> An object containing credentials for a Juju cloud.
>
> Note that secrets may be redacted.

Do we know *when* the secrets would be redacted, or in what way? I think if I was trying to use this I would like to know that, like it's always with AWS, or always with specific Juju versions, or something, and it's redacted by not being in `attributes` and being listed in `redacted` instead.

ops/model.py
```diff
@@ -3596,3 +3619,56 @@ def _ensure_loaded(self):
...
+
+    Used as the type of attribute `credential` in :class:`CloudSpec`.
+    """
+
+    auth_type: str
+    attributes: Dict[str, str]
+    redacted: List[str]
```
tonyandrewmeyer, line 3633: We would normally put a docstring after each of these lines (which shows up in the generated API docs, and elsewhere).

ops/model.py
```diff
@@ -3596,3 +3619,56 @@ def _ensure_loaded(self):
...
+    type: str
+    name: str
+    region: Optional[str]
+    endpoint: Optional[str]
+    identity_endpoint: Optional[str]
+    storage_endpoint: Optional[str]
+    credential: Optional[CloudCredential]
+    ca_certificates: List[str]
+    skip_tls_verify: bool
+    is_controller_cloud: bool
```
tonyandrewmeyer, line 3658: We'd normally have a short docstring for each of these attributes too.

ops/model.py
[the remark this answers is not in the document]
benhoyt: I'd prefer to have the types more accurate and use `d['auth-type']`, but then in `CloudSpec.from_dict` handle it by setting the entire `credential` field to None if it's not present. Something like this:

```python
credential=CloudCredential.from_dict(d['credential']) if d.get('credential') else None,
```

tonyandrewmeyer (commented):
Looking great! Just a couple of small doc requests.

test/test_model.py
```diff
@@ -3791,5 +3791,105 @@ def test_repr(self):
...
+    def test_from_dict(self):
+        cloud_spec = ops.CloudSpec.from_dict(
+            {
+                'type': 'lxd',
+                'name': 'localhost',
+            }
+        )
+        self.assertEqual(cloud_spec.type, 'lxd')
+        self.assertEqual(cloud_spec.name, 'localhost')
+        self.assertEqual(cloud_spec.region, '')
```
benhoyt, line 3828: With the changes mentioned above, these four fields should be None here.

test/test_model.py
```diff
@@ -3791,5 +3791,105 @@ def test_repr(self):
...
+                         [['credential-get', '--format=json']])
+
+    def test_error(self):
+        fake_script(
+            self,
+            'credential-get',
+            """echo 'ERROR cannot access cloud credentials' >&2; exit 1""")
+        with self.assertRaises(ops.ModelError) as cm:
+            self.model.get_cloud_spec()
+        self.assertEqual(str(cm.exception), 'ERROR cannot access cloud credentials\n')
```
benhoyt, line 3891: Nice, thanks for this.

benhoyt (changes requested):
Thanks. This is very close now! Just a few minor suggestions to improve the dataclass fields and defaults.

IronCore864:
Refreshed according to latest review, please take another look :)

tonyandrewmeyer (approved):
Looks great, thanks! Nothing from me, other than the Optional adjustment that Ben mentioned.

IronCore864:
CloudCredential and CloudSpec typing fixed.

benhoyt (approved):
Thanks, approved! Let's please just do two things before you merge: 1) add a line to `CHANGES.md` describing the new feature, and 2) describe the new feature briefly in the PR description.

Issue #826, opened 2022-09-07
ExecProcess.wait_output can leak a 'socket.timeout' exception


jameinel:
As seen here:
https://bugs.launchpad.net/juju/+bug/1989004
https://paste.ubuntu.com/p/KzS4r2PVV2/

The definition of `wait_output` is that it waits up to 'timeout' seconds for the response of the requested process, but if you pass a `timeout` of None, it will wait indefinitely.
The traceback looks like the code is trying to interact with the Pebble API for Wait Change (`_wait_change_using_wait` and `_wait_change`), but while trying to make the request against pebble it is getting a socket.timeout (`_request_raw', ... `http._read_status`)

Some other bits of note:

1. The time from the previous log message to the uncaught exception is 13s. The actual command being run likely takes about 3s (though the system might be under additional load). Either way, 13s is kind of an odd time for a socket.timeout
2. We do set a timeout on the Unix Socket here: https://github.com/canonical/operator/blob/main/ops/pebble.py#L229
I'm not positive where that timeout comes from, though presumably it is because we are passing a timeout to the do_open.
https://github.com/canonical/operator/blob/16295e4828e37ed3d0b6de46e19c4b51e8f45116/ops/pebble.py#L1553
3. pebble.Client has a standard timeout of 5s for any conversation with Pebble:
https://github.com/canonical/operator/blob/16295e4828e37ed3d0b6de46e19c4b51e8f45116/ops/pebble.py#L1469
I believe the intent for it is: "If we have sent a request to pebble, and it hasn't responded within 5s, loop around again and poll some more".
4. `_raw_request` does have a lot of handling of various bits
https://github.com/canonical/operator/blob/16295e4828e37ed3d0b6de46e19c4b51e8f45116/ops/pebble.py#L1554
notably, none of those are socket.timeout.
5. It is surprising that we are passing a socket.timeout to the _UnixSocketConnection but nowhere in the code do we ever handle `socket.timeout` that can happen because of that failure.

I don't know if this exec would ever fail: `container.exec('sleep 10; echo true')`. It may be that it should be handled by this code:
https://github.com/canonical/operator/blob/16295e4828e37ed3d0b6de46e19c4b51e8f45116/ops/pebble.py#L1761

That code would seem to give us a buffer of 1s for pebble to respond to our request before the HTTP connection dies due to socket.timeout.
I wonder if under load the issue is that we aren't actually able to get pebble to respond in time. Imagine the scenario:

[...]

Issue #1131, opened 2024-02-15
[doc] What happens when timeout period in `Container.exec()` elapses?


sed-i:
ops.model and ops.pebble don't specify what happens when timeout is elapsed.
Context: https://github.com/canonical/alertmanager-k8s-operator/pull/218

https://github.com/canonical/operator/blob/1836df5affb42b3183125b1904c794090aa1862b/ops/model.py#L2711-L2712

https://github.com/canonical/operator/blob/1836df5affb42b3183125b1904c794090aa1862b/ops/pebble.py#L2526-L2528

tonyandrewmeyer:
See also #826

sed-i:
Also: #867

sed-i:
Potentially related: `container.exec` also leaks `http.client.RemoteDisconnected`, as can be seen here: https://github.com/canonical/cos-lite-bundle/issues/103

IronCore864:
We decided against putting Raises TimeoutError in these methods, because they really need to go in all of them, so there's the note in the class docstring instead.

See discussion [here](https://github.com/canonical/operator/pull/1155#discussion_r1538423137).

Close now.

Issue #908, opened 2023-02-23
More clearly document which exceptions are raised


benhoyt:
Feedback from the IS devops team:

> It would be good to know which exceptions can be raised in any calls to ops. This is already documented on some functions, it would be good to expand that to all functions.

It's almost certainly true that we're not consistent about this and this could be improved.

Almost all of the functions that call hook tools can raise `ModelError`, so the original thinking was to not document that on each function. However, it's not always easy to tell at a glance which functions call hook tools. So probably better to be more explicit, even if it's verbose.

benhoyt:
Discussed with Tony, and we agree that `ModelError` can happen pretty much anywhere we're talking to Juju, so probably not worth annotating that in every function's "raises" -- just document it once in a prominent place in the API ref.

But still, there's probably other raises we need to get consistent or add.

PR #1155, opened 2024-03-20
fix(pebble): catch socket.timeout exception in pebble.Client.exec()

IronCore864:
Tried to run some `container.exec` in a test charm with sleep, with some other long-running commands like apt-update, with or without timeout, but could not reproduce the issue.

Changed according to John's opinion: catch `socket.timeout` exception and continue polling till the end of the deadline.

Also updated docstring for `model.Container.exec()` and `pebble.Client.exec()`.

Closes https://github.com/canonical/operator/issues/826.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Updated Pebble Notices `get_notices` parameter name to `users=all` (previously `select=all`).
 * Added `Model.get_cloud_spec` which uses the `credential-get` hook tool to get details of the cloud where the model is deployed.
+* Fixed an issue where `pebble.Client.exec` might leak a `socket.timeout` (`builtins.TimeoutError`) exception.
 * Updated code examples in the docstring of `ops.testing` from unittest to pytest style.
 * Refactored main.py, creating a new `_Manager` class.
 
--- a/ops/model.py
+++ b/ops/model.py
@@ -2713,6 +2713,7 @@ class Container:
 
         Raises:
             ExecError: if the command exits with a non-zero exit code.
+            TimeoutError: if the maximum timeout is reached.
         """
         if service_context is not None:
             version = JujuVersion.from_environ()
--- a/ops/model.py
+++ b/ops/model.py
@@ -2713,7 +2713,6 @@ class Container:
 
         Raises:
             ExecError: if the command exits with a non-zero exit code.
-            TimeoutError: if the maximum timeout is reached.
         """
         if service_context is not None:
             version = JujuVersion.from_environ()
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2592,7 +2592,6 @@ class Client:
             APIError: if an error occurred communicating with pebble, or if the command is not
                 found.
             ExecError: if the command exits with a non-zero exit code.
-            TimeoutError: if the maximum timeout is reached.
         """
         if not isinstance(command, list) or not all(isinstance(s, str) for s in command):
             raise TypeError(f'command must be a list of str, not {type(command).__name__}')
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -18,6 +18,7 @@ For a command-line interface for local testing, see test/pebble_cli.py.
 """
 
 import binascii
+import builtins
 import copy
 import dataclasses
 import datetime
@@ -1722,6 +1723,11 @@ class Client:
     connecting to or transferring data with Pebble will raise a :class:`ConnectionError`. When an
     error occurs executing the request, such as trying to add an invalid layer or execute a command
     that does not exist, an :class:`APIError` is raised.
+
+    The ``timeout`` parameter specifies a timeout in seconds for blocking operations like the
+    connection attempt to Pebble; used by ``urllib.request.OpenerDirector.open``. It's not for
+    methods like :meth:`start_services` and :meth:`replan_services` mentioned above, and it's not
+    for the command execution timeout defined in method :meth:`Client.exec`.
     """
 
     _chunk_size = 8192
@@ -2029,8 +2035,10 @@ class Client:
 
             try:
                 return self._wait_change(change_id, this_timeout)
-            except TimeoutError:
-                # Catch timeout from wait endpoint and loop to check deadline
+            except (socket.timeout, builtins.TimeoutError):
+                # NOTE: in Python 3.10 socket.timeout is an alias of TimeoutError,
+                # but we still need to support Python 3.8, so catch both.
+                # Catch timeout from wait endpoint and loop to check deadline.
                 pass
 
         raise TimeoutError(f'timed out waiting for change {change_id} ({timeout} seconds)')
@@ -2584,6 +2592,7 @@ class Client:
             APIError: if an error occurred communicating with pebble, or if the command is not
                 found.
             ExecError: if the command exits with a non-zero exit code.
+            TimeoutError: if the maximum timeout is reached.
         """
         if not isinstance(command, list) or not all(isinstance(s, str) for s in command):
             raise TypeError(f'command must be a list of str, not {type(command).__name__}')
--- a/test/test_pebble.py
+++ b/test/test_pebble.py
@@ -18,6 +18,7 @@ import email.parser
 import io
 import json
 import signal
+import socket
 import tempfile
 import test.fake_pebble as fake_pebble
 import typing
@@ -2027,6 +2028,18 @@ class TestClient(unittest.TestCase):
         self.assertEqual(self.client.requests, [
             ('GET', '/v1/changes/70/wait', {'timeout': '4.000s'}, None),
         ])
+
+    def test_wait_change_socket_timeout(self):
+        def timeout_response(n: float):
+            self.time.sleep(n)
+            raise socket.timeout("socket.timeout: timed out")
+
+        self.client.responses.append(lambda: timeout_response(3))
+
+        with self.assertRaises(pebble.TimeoutError) as cm:
+            self.client.wait_change(pebble.ChangeID('70'), timeout=3)
+        self.assertIsInstance(cm.exception, pebble.Error)
+        self.assertIsInstance(cm.exception, TimeoutError)
 
     def test_add_layer(self):
         okay_response = {


ops/pebble.py
```diff
@@ -2032,6 +2032,9 @@ def _wait_change_using_wait(self, change_id: ChangeID, timeout: Optional[float])
             except TimeoutError:
                 # Catch timeout from wait endpoint and loop to check deadline
                 pass
+            except socket.timeout:
```
benhoyt, line 2035: Note that you can use the following syntax to catch multiple exception types:

```suggestion
            except (TimeoutError, socket.timeout):
```

However, that makes me think: `TimeoutError` (which is `pebble.TimeoutError` here) inherits from the builtin `TimeoutError`, and [`socket.timeout`](https://docs.python.org/3/library/socket.html#socket.timeout) is a "deprecated alias of [the builtin] TimeoutError". So I think we can `import builtins` and just catch a single type here:

```
except builtins.TimeoutError:
```

Separately, I regret naming the pebble error `TimeoutError` in pebble.py, instead of just using the builtin exception type, as it's too confusing ... but it's too late to change that. :-)

benhoyt (changes requested):
Thanks for the investigation. Can we also add a regression test for this? For example, alongside `test_wait_change_timeout` in `test_pebble.py`, which makes the response raise `socket.timeout` -- which should fail before this change and succeed after? Thanks!

ops/pebble.py
IronCore864: I think `except builtins.TimeoutError` alone won't cache it if a `socket.timeout` exception is raised for Python <3.10 because only since 3.10 it becomes an alias, right? 

In order to make it work for all versions of Python work, I will explicitly put `except (builtins.TimeoutError, socket.timeout)` here.

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 * Updated Pebble Notices `get_notices` parameter name to `users=all` (previously `select=all`).
 * Added `Model.get_cloud_spec` which uses the `credential-get` hook tool to get details of the cloud where the model is deployed.
+* Fixed an issue where `pebble.Client.exec` might leak a `socket.timeout`(`builtin.TimeoutError`) exception.
```
benhoyt, line 5: Nit:
```suggestion
* Fixed an issue where `pebble.Client.exec` might leak a `socket.timeout` (`builtins.TimeoutError`) exception.
```

ops/pebble.py
```diff
@@ -2029,7 +2030,7 @@ def _wait_change_using_wait(self, change_id: ChangeID, timeout: Optional[float])
 
             try:
                 return self._wait_change(change_id, this_timeout)
-            except TimeoutError:
+            except (socket.timeout, builtins.TimeoutError):
```
benhoyt, line 2033: People looking at this might wonder why we have both, so let's add a comment, something like:

```suggestion
            except (socket.timeout, builtins.TimeoutError):
                # NOTE: in Python 3.10 socket.timeout is an alias of TimeoutError,
                # but we still need to support Python 3.8, so catch both.
```

test/test_pebble.py
```diff
@@ -2028,6 +2029,18 @@ def test_wait_change_error(self):
             ('GET', '/v1/changes/70/wait', {'timeout': '4.000s'}, None),
         ])
 
+    def test_wait_change_socket_timeout(self):
+        def timeout_response(n: float):
+            self.time.sleep(n)
+            raise socket.timeout(504, "socket.timeout: timed out")
```
benhoyt, line 2035: This wouldn't be a 504 at this point, so how about:
```suggestion
            raise socket.timeout("socket.timeout: timed out")
```

benhoyt (changes requested):
Looks good, just a couple of minor suggestions.

ops/pebble.py
```diff
@@ -2584,6 +2587,7 @@ def exec(
             APIError: if an error occurred communicating with pebble, or if the command is not
                 found.
             ExecError: if the command exits with a non-zero exit code.
+            TimeoutError: if the maximum timeout is reached.
```
tonyandrewmeyer, line 2590: I think `ExecProcess`'s `wait` and `wait_output` could also raise `TimeoutError`, and I think also the various action services methods of the Pebble client?

tonyandrewmeyer (commented):
There are also some other issues raised in #826 that I think we should either comment on somewhere (there or in this PR) or adjust. In particular:

1. The Unix socket has code for providing a timeout, but I don't think there's any way for a charm (or even the Pebble client) to provide that, so it's unclear why it's there. Am I missing some way that it can be set, or some reason to provide it even though it's never used?
2. The Pebble client has a timeout that's used for (I think) everything other than `exec`, but Charm's basically can't use this, because it's not exposed to the model. Is that deliberate? Should a charm be able to change this timeout?
3. It seems like in the original report that the timeout was happening after ~13s, but it should presumably have been 5s. Have we checked that we do see a timeout after 5s and not a longer time?

IronCore864:
> There are also some other issues raised in #826 that I think we should either comment on somewhere (there or in this PR) or adjust. In particular:
> 
> 1. The Unix socket has code for providing a timeout, but I don't think there's any way for a charm (or even the Pebble client) to provide that, so it's unclear why it's there. Am I missing some way that it can be set, or some reason to provide it even though it's never used?
> 2. The Pebble client has a timeout that's used for (I think) everything other than `exec`, but Charm's basically can't use this, because it's not exposed to the model. Is that deliberate? Should a charm be able to change this timeout?
> 3. It seems like in the original report that the timeout was happening after ~13s, but it should presumably have been 5s. Have we checked that we do see a timeout after 5s and not a longer time?

To answer the 3rd question first, [the 5s timeout](https://github.com/canonical/operator/blob/main/ops/pebble.py#L1732) is `pebble.Client.timeout`, and as John mentioned in the issue, the intent for it is "If we have sent a request to pebble, and it hasn't responded within 5s, loop around again and poll some more."

It's not the timeout for executing the command created by `Container.exec().wait()`. If we run `container.exec(["sleep", "60"]).wait()` without a timeout, [it will run for 60s](https://github.com/canonical/operator/blob/main/ops/pebble.py#L2023) for sure. The way to set timeout for the actual command is `container.exec(["sleep", "60"], timeout=3).wait()`, and this will throw some exception like `ops.pebble.ChangeError: cannot perform the following tasks: exec command "sleep" (timed out after 3s: context deadline exceeded)`.

The code in the original issue doesn't set timeout for `exec` and is waiting forever, but `socket.timeout` is raised, that is why we added `except socket.timeout`. As for why it's 13 seconds, I'm not sure, seems weird, and I could not reproduce `socket.timeout`.

---

[John did mention that](https://github.com/canonical/operator/issues/826#issuecomment-1239540051) the 5s timeout to pebble can't be configured and I checked, it's true, it can't be set in the entire OF code:

When creating a model, [metadata is used](https://github.com/canonical/operator/blob/main/ops/main.py#L410), then passed to Unit, which is [created by model](https://github.com/canonical/operator/blob/main/ops/model.py#L118), [using _ModelCache](https://github.com/canonical/operator/blob/main/ops/model.py#L325). Then [unit creates containers](https://github.com/canonical/operator/blob/main/ops/model.py#L514), [container initializes pebble client](https://github.com/canonical/operator/blob/8097cca861eae31727636f92c932738d14d0ac81/ops/model.py#L2094). However, no where in the whole chain is pebble timeout set, so it's always the default 5s.

I think if we want to expose pebble timeout as a configuration item, one way to do so is to add something like `pebble-timeout` in `[ContainerMeta](https://github.com/canonical/operator/blob/main/ops/charm.py#L1688)` and tweak [the code where it creates containers](https://github.com/canonical/operator/blob/8097cca861eae31727636f92c932738d14d0ac81/ops/model.py#L514), so that charmers can set pebble timeout in `charmcraft.yaml`, containers section.

I don't know if this change is worth it, because A, 5s seems to be an OK value, and B, users probably only care about `container.exec(timeout=xxx)` which is supported, I don't know why they want to tweak the pebble timeout.

Is it worth adding this feature? Need your input here @benhoyt @tonyandrewmeyer.

---

Another thing John mentioned was [the `_UnixSocketConnection` timeout](https://github.com/canonical/operator/issues/826#issuecomment-1239542298) issue that we set it even if charmers don't ask for it explicity, but I don't think it's true anymore, [we only set it if it's provided](https://github.com/canonical/operator/blob/main/ops/pebble.py#L321). Should we comment in the issue?

ops/pebble.py
IronCore864: `wait`/`wait_output` did call `wait_change` which calls `_wait_change_using_wait`, but `wait_change` does not handle timeout exception, so `wait`/`wait_output` raise a `ChangeError` instead (already in their docstring), see my other reply in the other comment.

ops/pebble.py
```diff
@@ -2029,8 +2030,10 @@ def _wait_change_using_wait(self, change_id: ChangeID, timeout: Optional[float])
 
             try:
                 return self._wait_change(change_id, this_timeout)
-            except TimeoutError:
-                # Catch timeout from wait endpoint and loop to check deadline
+            except (socket.timeout, builtins.TimeoutError):
```
sed-i, line 2033: Should we also catch `http.client.RemoteDisconnected` here?
See https://github.com/canonical/cos-lite-bundle/issues/103

IronCore864: Based on the stack trace I think it's necessary. Thoughts? @benhoyt @tonyandrewmeyer

tonyandrewmeyer:
> To answer the 3rd question first, [the 5s timeout](https://github.com/canonical/operator/blob/main/ops/pebble.py#L1732) is `pebble.Client.timeout`, and as John mentioned in the issue, the intent for it is "If we have sent a request to pebble, and it hasn't responded within 5s, loop around again and poll some more."

This is badly named, in my opinion, but we can't do anything about that now. 

If I understand correctly, if I do `Client(timeout=5).autostart_services(timeout=30)`, then:

* A request is sent to Pebble, with a timeout of 5s saying "do this change"
* A request is sent to Pebble, with a timeout of 5s saying "wait for that change to finish" - this also sends a timeout of 4s to Pebble's wait API. If the request gives a 503, then we try again (until we have used up 30s waiting for the change to happen, in which case we raise `TimeoutError` - the last attempt might wait less than 4s if there isn't that much left to get to 30s), otherwise we raise an `APIError`.

The wait API is a bit odd in that we are basically polling, but it seems to have been added to avoid polling.

If I do `Client(timeout=5).exec(timeout=30, ...)` then:

* A request is sent to Pebble, with a timeout of 5s saying "run this command" and that includes the 30s timeout as a parameter for Pebble's context.
* (Bunch of websocket stuff for handling I/O)
* A request is sent to Pebble, with a timeout of 5s saying "wait for that command to finish" - this also sends a timeout of 4s to Pebble's wait API. If the request gives a 503, then we try again (until we have used up 31s waiting for the change to happen, in which case we raise `TimeoutError` - the last attempt might wait less than 4s if there isn't that much left to get to 30s), otherwise we raise an `APIError`.

So they are basically the same underneath.

> It's not the timeout for executing the command created by `Container.exec().wait()`. If we run `container.exec(["sleep", "60"]).wait()` without a timeout, [it will run for 60s](https://github.com/canonical/operator/blob/main/ops/pebble.py#L2023) for sure.

Yeah, I get that there are a bunch of different timeouts here, the `pebble.Client.timeout` is mostly an implementation detail around how granular the timeout checks are, and the methods that are expected to be slow (`exec`, `autostart_services`, `restart_services`, etc) all have their own timeout that's passed when they are run (and that's a timeout waiting for Pebble to do whatever you've asked).

Although the ticket summary is about the leaked exception, I think the broader problem (of the two issues combined) is "Pebble timeouts are confusing", and it'd be good to solve that, rather than just the specific leaking exception.

> As for why it's 13 seconds, I'm not sure, seems weird, and I could not reproduce `socket.timeout`.

I would guess that the 13 comes from three attempts - we tried 2 times to get the change result from Pebble, waiting 4s each time, and that all worked as designed, and then on the 3rd attempt instead of the Pebble call timing out, the socket connection timed out after 5s (4+4+5=13).

If so I guess this is fine, because now we'll be catching that socket error and it'll time out after 30s.

> I think if we want to expose pebble timeout as a configuration item,

I think we probably do *not* want to expose this, since it's not what users would think of a timeout. I think we do want to explain in the `pebble.Client` docs what that `timeout` attribute actually is.

> Another thing John mentioned was [the `_UnixSocketConnection` timeout](https://github.com/canonical/operator/issues/826#issuecomment-1239542298) issue that we set it even if charmers don't ask for it explicity, but I don't think it's true anymore, [we only set it if it's provided](https://github.com/canonical/operator/blob/main/ops/pebble.py#L321). Should we comment in the issue?

I looked at this a bit more closely and it's clearer to me now. `_UnixSocketConnection` takes the `timeout` argument because it's a subclass of `http.client.HTTPConnection`, and that has a `timeout` argument. We do set the UNIX socket to use this timeout (but do it later than we should - you're really meant to set it before `connect()`). It's passed to `_UnixSocketConnection` via the `urllib.request` handling mechanism (that's how the 5s value gets everywhere), so it is always set by Pebble.

ops/pebble.py
tonyandrewmeyer: Ah, I remember now that @benhoyt and I discussed all this already last year in #908. We decided that it's too noisy to say that basically all the Pebble/Container methods might raise ChangeError/Timeout error, so:

* In [Container](https://ops.readthedocs.io/en/latest/#ops.Container) we say:

> For methods that make changes to the container, if the change fails or times out, then a [ops.pebble.ChangeError](https://ops.readthedocs.io/en/latest/#ops.pebble.ChangeError) or [ops.pebble.TimeoutError](https://ops.readthedocs.io/en/latest/#ops.pebble.TimeoutError) will be raised.

* In [pebble.Client](https://ops.readthedocs.io/en/latest/#ops.pebble.Client) we say:

> For methods that wait for changes, such as [start_services()](https://ops.readthedocs.io/en/latest/#ops.pebble.Client.start_services) and [replan_services()](https://ops.readthedocs.io/en/latest/#ops.pebble.Client.replan_services), if the change fails or times out, then a [ChangeError](https://ops.readthedocs.io/en/latest/#ops.pebble.ChangeError) or [TimeoutError](https://ops.readthedocs.io/en/latest/#ops.pebble.TimeoutError) will be raised.

So perhaps we ought to be rejecting #1131 instead (otherwise we are going to go down the route of adding them all again). We *do* say what happens when `exec` times out, you just have to read the `Container` doc to see it.

IronCore864:
I agree that "Pebble timeouts are confusing" -- at least this is what I was thinking when I was reading the code.

I think it's too complicated to be completely refactored, also I don't know what the impact circle would be. Still, maybe we can do some minimal refactoring to make it better, even just slightly so?

Off the top of my head, I have the following ideas:

1. Remove `timeout: float = 5.0` from `pebble.Client.__init__()` because it's not configurable anyway. Make it a hard-coded constant value.
2. Rename `pebble.Client.timeout` to something more meaningful, maybe `pebble.Client.connection_timeout` or `pebble.Client.socket_timeout`, to make it clear that it's not "the timeout for executing a command in the container". I think this parameter is used for [`urllib.request.OpenerDirector.open`](https://docs.python.org/3/library/urllib.request.html#urllib.request.OpenerDirector.open), which says "The optional timeout parameter specifies a timeout in seconds for blocking operations like the connection attempt." So I think maybe `pebble.Client.connection_timeout` is a better name to be consistent with `urllib.request`?
3. Refactor the `_wait_change_using_wait` method a bit to make it more readable.

IronCore864:
> This is badly named, in my opinion, but we can't do anything about that now.
> 
> Although the ticket summary is about the leaked exception, I think the broader problem (of the two issues combined) is "Pebble timeouts are confusing", and it'd be good to solve that, rather than just the specific leaking exception.

I agree that "Pebble timeouts are confusing" -- at least this is what I was thinking when I was reading the code.

I think it's too complicated to be completely refactored, also I don't know what the impact circle would be. Still, maybe we can do some minimal refactoring to make it better, even just slightly so?

Off the top of my head, I have the following ideas:

1. Remove `timeout: float = 5.0` from `pebble.Client.__init__()` because it's not configurable anyway. Make it a hard-coded constant value.
2. Rename `pebble.Client.timeout` to something more meaningful, maybe `pebble.Client.connection_timeout` or `pebble.Client.socket_timeout`, to make it clear that it's not "the timeout for executing a command in the container". I think this parameter is used for [`urllib.request.OpenerDirector.open`](https://docs.python.org/3/library/urllib.request.html#urllib.request.OpenerDirector.open), which says "The optional timeout parameter specifies a timeout in seconds for blocking operations like the connection attempt." So I think maybe `pebble.Client.connection_timeout` is a better name to be consistent with `urllib.request`?
3. Refactor the `_wait_change_using_wait` method a bit to make it more readable.

tonyandrewmeyer:
> 1. Remove `timeout: float = 5.0` from `pebble.Client.__init__()` because it's not configurable anyway. Make it a hard-coded constant value.

Even though this isn't accessible directly by charms, this would break backwards compatibility so we can't really do it. I don't think there's much we can do other than make the documentation clearer on what this value is (unless we do want to expose it to charms, but as we discussed above, I don't think so).

> 2. Rename `pebble.Client.timeout` to something more meaningful, maybe `pebble.Client.connection_timeout` or `pebble.Client.socket_timeout`, to make it clear that it's not "the timeout for executing a command in the container". I think this parameter is used for [`urllib.request.OpenerDirector.open`](https://docs.python.org/3/library/urllib.request.html#urllib.request.OpenerDirector.open), which says "The optional timeout parameter specifies a timeout in seconds for blocking operations like the connection attempt." So I think maybe `pebble.Client.connection_timeout` is a better name to be consistent with `urllib.request`?

I think a different name would have been great, but this would again break backwards compatibility, so we can't really do it (in ops 2.x).

ops/pebble.py
benhoyt: Without further investigation, I don't think we should add this (or other exception types here). We're specifically catching the various kinds of timeout, but I don't think we want to start catching other errors in general. I'm not exactly sure what `RemoteDisconnected` is or is caused by, but it sounds like something the remote side did actively, not something we want to automatically retry for.

ops/pebble.py
benhoyt: Ah yes, I think Tony's right -- thanks!

benhoyt:
Yeah, and can see how the "Client.timeout" name is confusing. I agree with Tony's assessment above:

> > I think if we want to expose pebble timeout as a configuration item,

> I think we probably do not want to expose this, since it's not what users would think of a timeout. I think we do want to explain in the pebble.Client docs what that timeout attribute actually is.

I don't think we should expose the ability for charmers to change this (no one's asked for that, and 5s has worked well). I'm on board with explaining in the `pebble.Client` docstring what it means though!

IronCore864:
I have added a little explanation in `pebble.Client` docstring. Please take another look.

ops/pebble.py
```diff
@@ -1722,6 +1723,11 @@ class Client:
     connecting to or transferring data with Pebble will raise a :class:`ConnectionError`. When an
     error occurs executing the request, such as trying to add an invalid layer or execute a command
     that does not exist, an :class:`APIError` is raised.
+
+    The ``timeout`` parameter specifies a timeout in seconds for blocking operations like the
+    connection attempt to Pebble; used by ``urllib.request.OpenerDirector.open``. It's not for
+    methods like :meth:`start_services` and :meth:`replan_services` mentioned above, and it's not
+    for command execution timeout defind in method :meth:`Client.exec`.
```
benhoyt, line 1730:
```suggestion
    for the command execution timeout defined in method :meth:`Client.exec`.
```

benhoyt (changes requested):
Looks good to me. Let's just remove the 3 extraneous files added under `docs` (do we need to add something more to `.gitignore`?)

IronCore864:
> Looks good to me. Let's just remove the 3 extraneous files added under `docs` (do we need to add something more to `.gitignore`?)

Sorry, it was a bad merge; the history is too complicated so I rebased everything to fix it.

benhoyt (approved):

ops/model.py
```diff
@@ -2713,6 +2713,7 @@ def exec(
 
         Raises:
             ExecError: if the command exits with a non-zero exit code.
+            TimeoutError: if the maximum timeout is reached.
```
tonyandrewmeyer, line 2716: There was a conversation here, but I can't find it in the GitHub UI any more (because of the rebase?), although it shows up in my email notifications so I don't think I was imagining it :laughing:.

@benhoyt and I had previously discussed this and decided against putting `Raises TimeoutError` in these methods, because they really need to go in all of them, so there's the note in the class docstring instead.

So my inclination is to not add these (and reject the issue asking for it, showing that it's there in the class docstring). Not sure whether @benhoyt agreed with that or not, or whether @IronCore864 you wanted to push back on that?

benhoyt: Ah yes, that's right -- I agree we should just remove these "Raises TimeoutError" everywhere and have them in that one place.

IronCore864: Got it, I removed them from the methods in `pebble.Client` and `model.Container`; both class docstring have it already.

benhoyt (approved):

tonyandrewmeyer (approved):
Thanks for all the work on this!

Issue #1154, opened 2024-03-18
Update `ops.testing`/Harness test examples in docstring to pytest style


IronCore864:
Since we are encouraging charm tests to be written with pytest, we should use that in test code examples in `ops.testing`.

PR #1157, opened 2024-03-21
docs(testing): update code samples in ops.testing from unittest to pytest style

IronCore864:
Since we are encouraging charm tests to be written with pytest, we should use that in test code examples in ops.testing.

Only code samples from `ops/testing.py` are updated, other files/test cases are not updated.

Closes https://github.com/canonical/operator/issues/1154.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Updated Pebble Notices `get_notices` parameter name to `users=all` (previously `select=all`).
 * Added `Model.get_cloud_spec` which uses the `credential-get` hook tool to get details of the cloud where the model is deployed.
+* Updated code examples in the docstring of `ops.testing` from unittest to pytest style.
 
 # 2.11.0
 
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -185,9 +185,9 @@ class Harness(Generic[CharmType]):
                 harness.cleanup()
 
     Below is an example test using :meth:`begin_with_initial_hooks` that ensures
-    the charm responds correctly to config changes (the parameter ``harness`` in the 
+    the charm responds correctly to config changes (the parameter ``harness`` in the
     test function is a pytest fixture that does setup/teardown, see :class:`Harness`)::
-        
+
         def test_foo(harness):
             # Instantiate the charm and trigger events that Juju would on startup
             harness.begin_with_initial_hooks()
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -380,7 +380,7 @@ class Harness(Generic[CharmType]):
         containers (in addition to triggering pebble-ready for each).
 
         Example::
-            def test_foo():
+
             harness = Harness(MyCharm)
             # Do initial setup here
             # Add storage if needed before begin_with_initial_hooks() is called
@@ -1699,7 +1699,7 @@ class Harness(Generic[CharmType]):
         Example usage::
 
             # charm.py
-            class MyCharm(ops.CharmBase):
+            class ExampleCharm(ops.CharmBase):
                 def __init__(self, *args):
                     super().__init__(*args)
                     self.framework.observe(self.on["mycontainer"].pebble_ready,
@@ -1711,7 +1711,7 @@ class Harness(Generic[CharmType]):
             # test_charm.py
             @pytest.fixture()
             def harness():
-                harness = Harness(MyCharm)
+                harness = Harness(ExampleCharm)
                 yield harness
                 harness.cleanup()
 
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -176,15 +176,18 @@ class Harness(Generic[CharmType]):
 
     The model created is from the viewpoint of the charm that is being tested.
 
-    Below is an example test using :meth:`begin_with_initial_hooks` that ensures
-    the charm responds correctly to config changes::
+    Always call ``harness.cleanup()`` after creating a :class:`Harness`::
 
         @pytest.fixture()
-        def harness():
-            harness = Harness(MyCharm)
-            yield harness
-            harness.cleanup()
+            def harness():
+                harness = Harness(MyCharm)
+                yield harness
+                harness.cleanup()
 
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 * Updated Pebble Notices `get_notices` parameter name to `users=all` (previously `select=all`).
 * Added `Model.get_cloud_spec` which uses the `credential-get` hook tool to get details of the cloud where the model is deployed.
+* Updated code examples in the docstring of `ops.testing` from unittest to pytest style.
```
tonyandrewmeyer, line 5: I'm not sure whether we should include this in CHANGES.md or not - I've started a discussion in Matrix to figure out what our pattern should be :)

tonyandrewmeyer (commented):
Looks good, thanks! A couple of questions about seemingly no-op indentation changes, and a suggestion about simplifying one of the examples.

ops/testing.py
```diff
@@ -962,8 +960,8 @@ def add_relation_unit(self, relation_id: int, remote_unit_name: str) -> None:
 
         Example::
 
-          rel_id = harness.add_relation('db', 'postgresql')
-          harness.add_relation_unit(rel_id, 'postgresql/0')
+            rel_id = harness.add_relation('db', 'postgresql')
+            harness.add_relation_unit(rel_id, 'postgresql/0')
```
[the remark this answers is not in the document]
IronCore864: This change is only for consistency since most docstring code samples in `ops/testing.py` uses 4 spaces instead of 2. They don't affect the rendered result by sphinx.

CHANGES.md
IronCore864: I will resolve this for now and let's continue the discussion in the Matrix and we have another PR for tweaking CHANGES.md.

ops/testing.py
```diff
@@ -1926,30 +1928,36 @@ def __init__(self, framework: ops.Framework):
...
-                    self.harness.charm.on.start.emit()
-                    expected = ops.CloudSpec.from_dict(cloud_spec_dict)
-                    self.assertEqual(harness.charm.cloud_spec, expected)
+                    },
+                }
+
+            @pytest.fixture()
+            def harness(cloud_spec_dict):
+                harness = Harness(MyVMCharm)
+                harness.set_cloud_spec(ops.model.CloudSpec.from_dict(cloud_spec_dict))
```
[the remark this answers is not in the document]
IronCore864: After discussion, we decide to only write the harness fixture only once in the class docstring and refer to it at other places. I will make the change accordingly, and this `set_cloud_spec` will of course be removed from the only fixture that serves as setup/teardown.

CHANGES.md
tonyandrewmeyer: Feel free to leave it in and whoever does the release next week can tidy it up, or pull it out and leave it for them to do - the first release with the new process will be a bit mixed anyway.

ops/testing.py
```diff
@@ -176,47 +176,49 @@ class Harness(Generic[CharmType]):
...
     The model created is from the viewpoint of the charm that is being tested.
 
-    Below is an example test using :meth:`begin_with_initial_hooks` that ensures
-    the charm responds correctly to config changes::
+    Always call ``harness.cleanup()`` after creating a :class:`Harness`::
 
-        class TestCharm(unittest.TestCase):
-            def test_foo(self):
+        @pytest.fixture()
+            def harness():
```
tonyandrewmeyer, line 182: One of these lines must be indented incorrectly?

ops/testing.py
```diff
@@ -176,47 +176,49 @@ class Harness(Generic[CharmType]):
 
     The model created is from the viewpoint of the charm that is being tested.
 
-    Below is an example test using :meth:`begin_with_initial_hooks` that ensures
-    the charm responds correctly to config changes::
+    Always call ``harness.cleanup()`` after creating a :class:`Harness`::
```
tonyandrewmeyer, line 179: I like how you've introduced this here - it's sold me on this approach, I think.

ops/testing.py
```diff
@@ -1926,30 +1928,26 @@ def __init__(self, framework: ops.Framework):
...
-                        'credential': {
-                            'authtype': 'certificate',
-                            'attrs': {
-                                'client-cert': 'foo',
-                                'client-key': 'bar',
-                                'server-cert': 'baz'
-                            },
+            # test_charm.py
+            def test_start(harness):
+                cloud_spec_dict = {
```
tonyandrewmeyer, line 1933: I would have this like:

```suggestion
                cloud_spec = ops.CloudSpec.from_dict({
```

with the closing brace at the end, and then you can get rid of `expected` and just pass `cloud_spec` to `set_cloud_spec()` and use it in the `assert` statement.

tonyandrewmeyer (changes requested):
Yeah, I think this approach works nicely, thanks!

One small suggestion, and one place where I think the indentation isn't right.

ops/testing.py
IronCore864: Yes it's a good idea!

tonyandrewmeyer (approved):
Thanks; looks good!

tonyandrewmeyer:
While working on another ticket, I noticed that there's also [this unittest style doc in testing.py](https://github.com/IronCore864/operator/blob/fdd42712c2d2692144a49aea4a3b0e77f7d15c0c/ops/testing.py#L482) that should be updated as well.

benhoyt:
> I noticed that there's also [this unittest style doc in testing.py](https://github.com/IronCore864/operator/blob/fdd42712c2d2692144a49aea4a3b0e77f7d15c0c/ops/testing.py#L482) that should be updated as well.

Yeah, it would be nice if we could just tweak the text there to be more general. Something like: "Always call ``harness.cleanup()`` after a test completes (normally this is done within a pytest fixture)."

benhoyt (approved):
Let's just get that tweak to the `cleanup` docstring in and then this is good to go!

PR #1159, opened 2024-03-22
docs: adjust change log entries, and update HACKING.md to match

tonyandrewmeyer:
Minor adjustments to the way that we record changes:

* Add more structure to [CHANGES](CHANGES.md): group changes by the same types of groups as in our conventional commit messages (but with full words rather than abbreviations), and add the GitHub PR number that introduced the change
* Include all non-internal changes in [CHANGES](CHANGES.md)
* Remove "New in" notes from the API documentation (we want everyone to regularly update to the latest version, not pick a version that has whatever they need, and work hard to make this as easy as possible)
* Update [HACKING](HACKING.md) to reflect these process changes

These changes are backported to ops 2.7, which is where the CHANGES file started.


CHANGES.md
```diff
@@ -1,38 +1,131 @@
...
+
+Documentation:
+
+* Reduces the amount of detail in open/close port methods (#1006)
+* Removes you/your from docstrings (#1003)
+* Minor improvements to HACKING (#1016)
+
+Tooling:
+
+* Extends the use of type hints in the test suite (#1008, #1009, #1011, #1012, #1013, #1014, #1004)
```
benhoyt, line 131: Thanks for adding these all the way back to here!

benhoyt (approved):
Thanks for these updates. Just a couple of minor suggestions for CHANGES.md formatting.

IronCore864 (approved):

Issue #1158, opened 2024-03-21
Update RTD site to use latest Canonical styling


benhoyt:
When I updated our API reference docs (https://ops.readthedocs.io/en/latest/) to use the Canonical fonts and styling, the Sphinx starter pack wasn't in a finished state, and I had to copy it into our repo and hack it around a bit to get it working. @jameinel noted that there are several non-ideal things about it, in particular the monospace font size doesn't match the normal font size: the monospace font (Ubuntu Mono 14.3) is a lot smaller point size than the main text (Ubuntu 17.6).

Let's see if we can update this to use the proper Sphinx style/theme/etc. See https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/ and https://github.com/canonical/sphinx-docs-starter-pack

benhoyt:
Done in three PRs:

- https://github.com/canonical/operator/pull/1163
- https://github.com/canonical/operator/pull/1164
- https://github.com/canonical/operator/pull/1165

PR #1160, opened 2024-03-22
docs: update read the docs site to latest Canonical styling

IronCore864:
Update read the docs/sphinx settings to use latest Canonical style.

Closes https://github.com/canonical/operator/issues/1158.


jameinel:
Thank you for generating the preview.
For the most part, I think it looks better. Having it distinguish the code blocks from prose is more readable to me.

2 small complaints, and I'm not sure if it has anything for @IronCore864 to do.
image
image

In light mode, it is using white underlines for Type names as well as Blue text. I think they are hyperlinks, but shouldn't that be Blue text with Blue underlines?

Conversely in Dark mode, we end up with light (white?) text on light grey background. I would have expected it to go to a "dark grey" background to more closely match the 'subtle highlight' that we have in the light mode.

IronCore864:
> Thank you for generating the preview. For the most part, I think it looks better. Having it distinguish the code blocks from prose is more readable to me.
> 
> 2 small complaints, and I'm not sure if it has anything for @IronCore864 to do. image image
> 
> In light mode, it is using white underlines for Type names as well as Blue text. I think they are hyperlinks, but shouldn't that be Blue text with Blue underlines?
> 
> Conversely in Dark mode, we end up with light (white?) text on light grey background. I would have expected it to go to a "dark grey" background to more closely match the 'subtle highlight' that we have in the light mode.

Maybe it has something todo with previously done customizations; I'll open up a new PR to start from scratch with the starter kit and do some tweaks from there, then do a comparison to see which way to go.

IronCore864:
I have created [another PR here](https://github.com/canonical/operator/pull/1162).

The difference is, in this PR, I merged the Canonical starter pack with what we've got so as not to lose anything we customized.

In the new PR, I started from scratch with the Canonical starter pack (and did some hacking, like autodoc, and fixing the two issues John mentioned).

Please go to the new PR, do a preview, and compare which you like.

@jameinel @benhoyt @tonyandrewmeyer

IronCore864:
Closing this PR because the other one here https://github.com/canonical/operator/pull/1162 looks more like what we want.

Issue #1166, opened 2024-03-26
Support user secrets: "secret" config type and Harness support


jameinel:
To handle user secrets, with Juju (3.4(?) you can now have a config with a schema type of "secret". Juju will then ensure that it can hold a secret URL.

However, if I try to create a charm that uses this, when running the test suite I get:
```
Traceback (most recent call last):
  File "/home/jameinel/dev/charms/ubuntu-lite/test/test_charm.py", line 22, in test_on_start
    harness = testing.Harness(charm.Ubuntu)
  File "/home/jameinel/dev/charms/ubuntu-lite/venv/lib/python3.10/site-packages/ops/testing.py", line 251, in __init__
    self._backend = _TestingModelBackend(self._unit_name, self._meta, config_)
  File "/home/jameinel/dev/charms/ubuntu-lite/venv/lib/python3.10/site-packages/ops/testing.py", line 2079, in __init__
    self._config = _TestingConfig(config)
  File "/home/jameinel/dev/charms/ubuntu-lite/venv/lib/python3.10/site-packages/ops/testing.py", line 1966, in __init__
    self._config_set(key, value)
  File "/home/jameinel/dev/charms/ubuntu-lite/venv/lib/python3.10/site-packages/ops/testing.py", line 1996, in _config_set
    raise RuntimeError(
RuntimeError: Incorrectly formatted `options.yaml`: `type` needs to be one of [string, boolean, int, float], not secret.
```

I tested this with:
```
Installing collected packages: websocket-client, ops
  Attempting uninstall: ops
    Found existing installation: ops 1.5.5
    Uninstalling ops-1.5.5:
      Successfully uninstalled ops-1.5.5
Successfully installed ops-2.11.0 websocket-client-1.7.0
```

jameinel:
So to trigger this behavior, I updated my charm's config to have:
```yaml
+++ b/config.yaml
@@ -1 +1,5 @@
-options: {}
+options:
+  mysec:
+    type: secret
+    default: nil
+    description: "tell me your secrets"
```

It turns out that it only fails because I set the 'default:' key, which makes the Testing infrastructure call _config_set with the defaults, and then it validates the type.
However, in practice, you shouldn't set the config, but you will need to read it.

jameinel:
and certainly the test suite will need to be able to set it :)

tonyandrewmeyer:
[...]

PR #1167, opened 2024-03-26
feat(Harness): add some support for user secrets

jameinel:
At least we should support if a config entry has a 'secret' stanza, which is supported in juju 3.4. It doesn't do all the things yet, but a rough cut.

This allows for testing.Harness to understand config entries of type Secret. It might be arguable that the testing harness should also be validating the *content* of those secret keys, such that they should conform to a secret identifier. However, this at least got me past the point of getting failures just because my charm declared the type (well, and it declared the default which it really shouldn't do anyway).

See #1166


tonyandrewmeyer (approved):
We're going to continue work on this (I'll note that in the issue), but get this into 2.12 in the meantime so that it unblocks using this, even if it's not perfect yet.

tonyandrewmeyer:
The linting failures are related to the docs theme work. Once the final part of that is merged in, I'll update this branch and they should be ok.

IronCore864 (approved):

PR #1170, opened 2024-03-28
feat: support change-update notice and pebble-change-updated event

benhoyt:
This adds support for the new `pebble-change-updated` event (Juju PR https://github.com/juju/juju/pull/17118), adding a new `PebbleChangeUpdatedEvent` type and `{container_name}_pebble_change_updated` event you can observe.

The event has a `get_change()` method to easily fetch the associated `pebble.Change` object.

It also adds `Harness.pebble_notify` support for the `pebble.NoticeType.CHANGE_UPDATE` notice type. To use that, you say:

```python
notice_id = harness.pebble_notify('<container>', '<change-id>', type=pebble.NoticeType.CHANGE_UPDATE)
```


ops/charm.py
```diff
@@ -808,6 +808,10 @@ class PebbleCustomNoticeEvent(PebbleNoticeEvent):
     """Event triggered when a Pebble notice of type "custom" is created or repeats."""
 
 
+class PebbleChangeUpdatedEvent(PebbleNoticeEvent):
```
benhoyt, line 811: TODO benhoyt: should we have a `.change` attribute or `.get_change()` method on this class, that looks up the associated change via the notice key (= change ID)?

ops/pebble.py
```diff
@@ -1314,7 +1314,17 @@ def __repr__(self):
...
     """Enum of notice types."""
 
+    CHANGE_UPDATE = 'change-update'
+    """Recorded whenever a change is updated, that is, when it is first
+    spawned or its status was updated. The key for change-update notices is
+    the change ID.
+    """
+
     CUSTOM = 'custom'
+    """A custom notice reported via the Pebble client API or "pebble notify".
```
tonyandrewmeyer, line 1324: Should this be backticks?

```suggestion
    """A custom notice reported via the Pebble client API or `pebble notify`.
```

ops/pebble.py
```diff
@@ -1314,7 +1314,17 @@ def __repr__(self):
...
+    The key and data fields are provided by the user. The key must be in
+    the format "mydomain.io/mykey" to ensure well-namespaced notice keys.
```
tonyandrewmeyer, line 1326: We should use domains we own (e.g. `canonical.com`) or RFC 2606 ones (e.g. `example.com`).

```suggestion
    the format "example.com/mykey" to ensure well-namespaced notice keys.
```

*(The Pebble readme ought to have this fix too).*

ops/charm.py
tonyandrewmeyer: > Do you think we should have a `.change` attribute or `.get_change()` method on this class, that looks up the associated change via the notice key (= change ID)?
> 
> A helper might be nice, though as it is it's not hard to get:
> 
> ```python
> change = event.workload.pebble.get_change(event.notice.key)
>     # vs
> change = event.get_change()
> ```

I think a helper would be nice. It seems like if you're observing change events then you're probably going to want to get hold of the actual change, so it makes sense to have it 'at hand' at the top level of the event namespace.

> I'd probably make it `get_change()` rather than a property, because it has to perform a call to Pebble.

:100:

ops/pebble.py
benhoyt: Pebble PR at https://github.com/canonical/pebble/pull/407

ops/charm.py
benhoyt: Added.

tonyandrewmeyer (approved):
Looks great to me, thanks!

ops/pebble.py
```diff
@@ -1314,7 +1314,17 @@ def __repr__(self):
...
+    CHANGE_UPDATE = 'change-update'
+    """Recorded whenever a change is updated, that is, when it is first
+    spawned or its status was updated. The key for change-update notices is
+    the change ID.
+    """
+
     CUSTOM = 'custom'
+    """A custom notice reported via the Pebble client API or ``pebble notify``.
+    The key and data fields are provided by the user. The key must be in
+    the format ``example.com/path`` to ensure well-namespaced notice keys.
```
wallyworld, line 1326: Do we validate this?

wallyworld (approved):

ops/pebble.py
benhoyt: Yeah, we validate it (in the Pebble server), so the workload will get an error if it tries to add a custom notice (`pebble notify`) with an invalid key. The [API does "reasonable effort" validation](https://github.com/canonical/pebble/blob/18c93cbc280de20c9e21d323fe3a8bbb340777a2/internals/daemon/api_notices.go#L32-L34) with a regex.

IronCore864 (approved):

PR #1172, opened 2024-03-28
docs: tox -e docs-deps to compile requirements.txt

IronCore864:
Separate pip-compile into a separate env `tox -e docs-deps`. Now `tox -e docs` only does `pip-sync` then `sphinx-build`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -26,6 +26,9 @@ tox -e unit test/test_charm.py
 # Format the code using isort and autopep8
 tox -e fmt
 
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Works for me, thanks!

tonyandrewmeyer (changes requested):
Could we add a sentence in HACKING.md that mentions that this exists? Other than that, looks good to me, thanks!

tonyandrewmeyer (approved):
Thanks!

PR #1176, opened 2024-04-03
feat(Harness): support user secrets

IronCore864:
Add support for user secrets in Harness. Changes:

- Add a new API `Harness.add_user_secret` which simulates `juju add-secret`.
- `add_model_secret` is deprecated, use `add_charm_secret`. For now, `add_model_secret` is an alias to `add_charm_secret`.

Use case example:

`config.yaml`:

```yaml
options:
  mysec:
    type: secret
    description: "tell me your secrets"
```

`charm.py`:

```python
class MyVMCharm(ops.CharmBase):
        super().__init__(framework)
    def _on_config_changed(self, event: ops.StartEvent):
        mysec = self.config.get('mysec')
        if mysec:
            sec = self.model.get_secret(id=mysec, label="mysec")
            self.config_from_secret = sec.get_content()
```

`test_charm.py`:

```python
def test_config_changed(harness):
    secret_content = {'password': 'foo'}
    secret_id = harness.add_user_secret(secret_content)
    # grant the user secret to the app
    harness.grant_secret(secret_id, 'webapp')
    harness.begin()
    harness.update_config({'mysec': secret_id})
    secret = harness.model.get_secret(id=secret_id).get_content()
    assert harness.charm.config_from_secret == secret.get_content()
```


ops/testing.py
```diff
@@ -1530,6 +1525,26 @@ def add_model_secret(self, owner: AppUnitOrName, content: Dict[str, str]) -> str
         model.Secret._validate_content(content)
         return self._backend._secret_add(content, owner_name)
 
+    def add_model_secret(self, owner: AppUnitOrName, content: Dict[str, str]) -> str:
+        """An alias for :meth:`add_charm_secret` for backward compatibility."""
+        return self.add_charm_secret(owner, content)
```
tonyandrewmeyer, line 1530: For what it's worth, I think if we do change the name, then we should have a warning emitted here. Not sure whether `DeprecationWarning` or `PendingDeprecationWarning`.

IronCore864:
Get user secret grant method is removed according to discussion. Will create a separate PR after this one to mark `add_model_secret` as deprecated.

ops/testing.py
```diff
@@ -1530,6 +1525,42 @@ def add_model_secret(self, owner: AppUnitOrName, content: Dict[str, str]) -> str
...
+
+            # test_charm.py
+            def test_start(harness):
+                secret_content = {'password': 'foo'}
+                secret_id = harness.add_user_secret(secret_content)
+                # grant the user secret to the app
+                harness.grant_secret(secret_id, 'webapp')
+                secret = harness.model.get_secret(id=secret_id)
+                assert secret.id == secret_id
+                assert secret.get_content() == secret_content
```
tonyandrewmeyer, line 1558: This feels a bit artificial - in reality you'd have a charm that did something with the secret, and since the charm can't make any changes to the secret you wouldn't ever be asserting on it afterwards.

However, I'm not sure how detailed we want these little examples in the API reference docs to be - a more realistic example is perhaps better suited to a how-to guide, so maybe leave this as-is?

tonyandrewmeyer (changes requested):
Looking good - a few little nits and a couple of queries.

Other questions/comments:

* The `trigger_secret_expiration` method should probably raise an error if it's used with a user secret. Is that already the case via permissions checking?
* Likewise for `trigger_secret_rotation`.
* It doesn't seem like we are testing the permission code changes. I think we'd want to have tests that verify that those work as expected.

benhoyt (commented):
I agree with Tony's comments. I'll review further later. For now, can you please add a PR description with a brief description of the new API and with a usage example?

ops/testing.py
```diff
@@ -1559,21 +1590,31 @@ def set_secret_content(self, secret_id: str, content: Dict[str, str]):
...
             observer: The name of the application (or specific unit) to grant
                 access to. A relation between this application and the charm
                 under test must already have been created.
         """
         secret = self._ensure_secret(secret_id)
+        app_or_unit_name = _get_app_or_unit_name(observer)
+
+        # user secrets
+        if secret.owner_name == self.model.uuid:
+            secret.user_secrets_grants.add(app_or_unit_name)
```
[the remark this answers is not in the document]
IronCore864: It's true that from Harness's standpoint, there is no need to grant it to the local app but since it's simulating `juju grant-secret`, technically it can grant to other apps. Should I keep it as a set? Or simply as a str of the app name?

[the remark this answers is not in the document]
tonyandrewmeyer: My point is that Harness doesn't run anything from other apps, so it essentially has no effect. I think keeping it is ok though.

ops/testing.py
IronCore864: Yeah, it feels totally made-up and has no real-world guidance value at all. Let me whip up a useful usecase.

IronCore864:
> Looking good - a few little nits and a couple of queries.
> 
> Other questions/comments:
> 
> * The `trigger_secret_expiration` method should probably raise an error if it's used with a user secret. Is that already the case via permissions checking?
> * Likewise for `trigger_secret_rotation`.
> * It doesn't seem like we are testing the permission code changes. I think we'd want to have tests that verify that those work as expected.

Changed `trigger_secret_expiration` and `trigger_secret_rotation` to raise an error on user secret, also added two unit tests.

For user secret permissions, there are some unit tests.

ops/testing.py
```diff
@@ -1683,6 +1752,8 @@ def trigger_secret_expiration(self, secret_id: str, revision: int, *,
                 label is used.
         """
         secret = self._ensure_secret(secret_id)
+        if secret.owner_name == self.model.uuid:
+            raise RuntimeError("Cannot trigger a secret-expired event for user secret.")
```
benhoyt, line 1756:
```suggestion
            raise RuntimeError("Cannot trigger the secret-expired event for a user secret.")
```

ops/testing.py
```diff
@@ -2567,8 +2645,9 @@ def _ensure_secret_owner(self, secret: _Secret):
         # https://discourse.charmhub.io/t/secret-access-permissions/12627
         unit_secret = secret.owner_name == self.unit_name
         app_secret = secret.owner_name == self.app_name
+        user_secret = secret.owner_name == self.model_uuid
 
-        if unit_secret or (app_secret and self.is_leader()):
+        if (unit_secret or (app_secret and self.is_leader())) and not user_secret:
```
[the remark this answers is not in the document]
IronCore864: Touché. It can't be both, I'll remove the user_secret part.

benhoyt (approved):

ops/testing.py
```diff
@@ -1509,27 +1509,75 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
             self._backend._calls.clear()
         return calls
 
-    def add_model_secret(self, owner: AppUnitOrName, content: Dict[str, str]) -> str:
-        """Add a secret owned by the remote application or unit specified.
-
-        This is named :code:`add_model_secret` instead of :code:`add_secret`
-        to avoid confusion with the :meth:`ops.Application.add_secret`
-        and :meth:`ops.Unit.add_secret` methods used by secret owner
-        charms.
```
tonyandrewmeyer, line 1518: Do we want to keep this comment (with the method name updated) in `add_charm_secret`? I think it's ok to remove it from the docstring, but I think it would be good to keep as a code comment?

ops/testing.py
```diff
@@ -1509,27 +1509,75 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
...
+                type: secret
+                description: "tell me your secrets"
+
+            # charm.py
+            class MyVMCharm(ops.CharmBase):
+                def __init__(self, framework: ops.Framework):
+                    super().__init__(framework)
+                    framework.observe(self.on.config_changed, self._on_config_changed)
+
+                def _on_config_changed(self, event: ops.StartEvent):
```
tonyandrewmeyer, line 1559:
```suggestion
                def _on_config_changed(self, event: ops.ConfigChangedEvent):
```

ops/testing.py
```diff
@@ -1509,27 +1509,75 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
...
+                    mysec = self.config.get('mysec')
+                    if mysec:
+                        sec = self.model.get_secret(id=mysec, label="mysec")
+                        self.config_from_secret = sec.get_content()
+
+            # test_charm.py
+            def test_config_changed(harness):
+                secret_content = {'password': 'foo'}
+                secret_id = harness.add_user_secret(secret_content)
+                harness.grant_secret(secret_id, 'webapp')
```
tonyandrewmeyer, line 1569: I think we should have this as a working example (I'd like to get these doctest'd in the future). Presumably this would if you had a `charmcraft.yaml` where the name was `webapp` but by default it should be:

```suggestion
                harness.grant_secret(secret_id, 'test-charm')
```

(That's the name that Harness uses if there isn't one in the metadata).

ops/testing.py
```diff
@@ -1509,27 +1509,75 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
...
+            The ID of the newly-added secret.
+
+        Example usage (the parameter ``harness`` in the test function is
+        a pytest fixture that does setup/teardown, see :class:`Harness`)::
+
+            # config.yaml
+            options:
+              mysec:
+                type: secret
+                description: "tell me your secrets"
```
tonyandrewmeyer, line 1551: We should use the more modern charmcraft approach.

```suggestion
            # charmcraft.yaml
            config:
              options:
                mysec:
                  type: secret
                  description: "tell me your secrets"
```

ops/testing.py
```diff
@@ -1509,27 +1509,75 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
...
+                secret_id = harness.add_user_secret(secret_content)
+                harness.grant_secret(secret_id, 'webapp')
+                harness.begin()
+                harness.update_config({'mysec': secret_id})
+                secret = harness.model.get_secret(id=secret_id).get_content()
+                assert harness.charm.config_from_secret == secret.get_content()
+
+        """
+        model.Secret._validate_content(content)
+        # Although it's named a user-owned secret in juju, technically, the owner is the
```
tonyandrewmeyer, line 1577:
```suggestion
        # Although it's named a user-owned secret in Juju, technically, the owner is the
```

ops/testing.py
```diff
@@ -1509,27 +1509,75 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
...
+        # Model, so the secret's owner is set to :attr:`Model.uuid <ops.Model.uuid>`.
```
tonyandrewmeyer, line 1578: I don't think it makes sense to use ReST for comments.

```suggestion
        # Model, so the secret's owner is set to `Model.uuid`.
```

tonyandrewmeyer (changes requested):
Can we adjust `_ensure_secret_owner` so that it explicitly talks about user secrets? It could have a `user_secret = secret.owner_name == self.model.uuid` and an `and not user_secret`, although that isn't strictly necessary, so maybe just a comment explaining the user secret situation, with the comment that explains unit and app permissions?

I would also feel more comfortable if we had tests that explicitly made sure that user secrets fail ownership checks. This will prevent regressions in the future if we change how the permissions checks are done.

IronCore864:
> Can we adjust `_ensure_secret_owner` so that it explicitly talks about user secrets? It could have a `user_secret = secret.owner_name == self.model.uuid` and an `and not user_secret`, although that isn't strictly necessary, so maybe just a comment explaining the user secret situation, with the comment that explains unit and app permissions?
> 
> I would also feel more comfortable if we had tests that explicitly made sure that user secrets fail ownership checks. This will prevent regressions in the future if we change how the permissions checks are done.

Yes it makes sense. I have added a test to try to do Secret.get_info to validate the ownership.

ops/testing.py
```diff
@@ -1509,17 +1509,12 @@ def _get_backend_calls(self, reset: bool = True) -> List[Tuple[Any, ...]]:
...
         return calls
 
-    def add_model_secret(self, owner: AppUnitOrName, content: Dict[str, str]) -> str:
-        """Add a secret owned by the remote application or unit specified.
-
-        This is named :code:`add_model_secret` instead of :code:`add_secret`
-        to avoid confusion with the :meth:`ops.Application.add_secret`
-        and :meth:`ops.Unit.add_secret` methods used by secret owner
-        charms.
+    def add_charm_secret(self, owner: AppUnitOrName, content: Dict[str, str]) -> str:
```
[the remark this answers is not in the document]
benhoyt: I just talked to John Meinel about this. It sounds like they have *not* settled on the name "charm secrets" and there's a bit more discussion they want to have about all this. So I think we should wait with this renaming. We can either just hold off on merging this PR till they've settled that (my recommendation) or we could merge with `add_model_secret` and `add_user_secret` and then change/deprecated `add_model_secret` in a future PR. What do you think @tonyandrewmeyer?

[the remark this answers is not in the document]
tonyandrewmeyer: Did John have any idea of when they might settle? If the expectation is that it'll be before 2.13 then I would definitely just wait. If it's more likely to be in May (e.g. after Madrid) then I'd leave it to @IronCore864 - if he wants to do the extra work to split this into two PRs I have no objection, but I also don't see any issue with waiting a bit longer. If it might be even further away than that, then I would split them now.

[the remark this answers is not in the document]
benhoyt: No, not really. I got the sense they wanted to decide soon, but I'm not sure how long that piece of string is...

I'd probably recommend not doing the rename now and leaving that for a future PR.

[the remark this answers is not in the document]
benhoyt: (so that we can merge this work without blocking).

IronCore864:
John has started a discussion in the juju development channel, now the preferred names are "application secrets" and "charm secrets", it seems it's not decided yet, I'll wait for a final conclusion before merging this one.

tonyandrewmeyer (approved):
LGTM!

benhoyt (approved):

PR #1178, opened 2024-04-08
docs: update docstrings with deprecated directive

IronCore864:
Update docstrings to use sphinx `deprecated` directive.


ops/charm.py
```diff
@@ -392,6 +392,8 @@ class LeaderSettingsChangedEvent(HookEvent):
     This event has been deprecated in favor of using a Peer relation,
     and having the leader set a value in the Application data bag for
     that peer relation.  (see :class:`RelationChangedEvent`).
+
+    .. deprecated:: 2.4.0
```
benhoyt, line 396: Thanks.

I think we should also get rid of the word "DEPRECATED" at the top, given the docs now say it clearly with the version: https://ops--1178.org.readthedocs.build/en/1178/#ops.LeaderSettingsChangedEvent ... or maybe it's still clearer with both. What do you think, @tonyandrewmeyer?

tonyandrewmeyer: Yeah, I was thinking the same thing - it's cleaner with it removed and the note is visible enough that people should notice it.

tonyandrewmeyer (approved):
Nice, thanks!

What do you & @benhoyt think about putting a deprecation notice on `OpenedPort` here as well? With a message to use `Port` instead, and the version either 2.13 as an official deprecation or 2.7 which is when it essentially was?

benhoyt:
A deprecation notice on `OpenedPort` sounds good to me!

IronCore864:
It seems if I add a docstring on a global variable:

```
OpenedPort = Port
"""Alias for backwards compatibility.

.. deprecated:: 2.7.0
    Use :class:`Port` instead.
"""
```

it does not show up in sphinx autodoc.
Shall I make it a new class so that I can add a docstrig to it?

```
class OpenedPort(Port):
    """Alias for backwards compatibility.

    .. deprecated:: 2.7.0
        Use :class:`Port` instead.
    """
```

benhoyt (approved):

ops/model.py
```diff
@@ -766,7 +767,12 @@ class Port:
     """The port number. Will be ``None`` if protocol is ``'icmp'``."""
 
 
-OpenedPort = Port  # Alias for backwards compatibility.
+class OpenedPort(Port):
```
benhoyt, line 770: Per discussion on Mattermost, I think we should revert this: I don't think it's a good precedent to break backwards compatibility (in the case of `isinstance`) just for a minor docstring improvement. (Though I realize it's unlikely anyone's actually doing that with this particular class.)

benhoyt (changes requested):

benhoyt (approved):
Thanks for the updates!

PR #1179, opened 2024-04-09
refactor: update ruff to 0.3.5 latest version

IronCore864:
Update ruff to the latest version, 0.3.5.

Major changes: `.format()` to f-strings, and a few newly added lint exceptions.

Since we will move to the Ruff formatter soon, this is some prerequisite work making the switch easier with less lint errors to fix.

Feel free to suggest changes for those multi-line f-strings; I tried my best to make them look not so bad but there still might be space to improve.


ops/framework.py
```diff
@@ -1191,8 +1189,9 @@ def __get__(self,
                 if bound is not None:
                     # the StoredState instance is being stored in two different
                     # attributes -> unclear what is expected of us -> bail out
-                    raise RuntimeError("StoredState shared by {0}.{1} and {0}.{2}".format(
-                        cls.__name__, self.attr_name, attr_name))
+                    raise RuntimeError(
+                        f"StoredState shared by {cls.__name__}.{self.attr_name} and "
+                        f"{cls.__name__}.{attr_name}")
```
tonyandrewmeyer, line 1194: Maybe reduce the number of changes in the future by making this single quotes now?

```suggestion
                    raise RuntimeError(
                        f'StoredState shared by {cls.__name__}.{self.attr_name} and '
                        f'{cls.__name__}.{attr_name}')
```

ops/model.py
```diff
@@ -588,8 +588,8 @@ def set_workload_version(self, version: str) -> None:
         shown in the output of 'juju status'.
         """
         if not isinstance(version, str):
-            raise TypeError("workload version must be a str, not {}: {!r}".format(
-                type(version).__name__, version))
+            raise TypeError(
+                f"workload version must be a str, not {type(version).__name__}: {version!r}")
```
tonyandrewmeyer, line 592: Same as above.

```suggestion
                f'workload version must be a str, not {type(version).__name__}: {version!r}')
```

ops/model.py
```diff
@@ -1629,9 +1630,8 @@ def _validate_read(self):
         if self._backend.app_name == self._entity.name:
             # minions can't read local app databags
             raise RelationDataAccessError(
-                "{} is not leader and cannot read its own application databag".format(
-                    self._backend.unit_name
-                )
+                f"{self._backend.unit_name} is not leader and cannot read its own "
+                "application databag"
```
tonyandrewmeyer, line 1634: Same as above, plus I believe Ruff likes to have consistent leading f's in implicitly concatenated strings even when not required so we'll hit that in the formatter:

```suggestion
                f'{self._backend.unit_name} is not leader and cannot read its own '
                f'application databag'
```

ops/model.py
```diff
@@ -1660,9 +1660,8 @@ def _validate_write(self, key: str, value: str):
             is_our_app: bool = self._backend.app_name == self._entity.name
             if not is_our_app:
                 raise RelationDataAccessError(
-                    "{} cannot write the data of remote application {}".format(
-                        self._backend.app_name, self._entity.name
-                    ))
+                    f"{self._backend.app_name} cannot write the data of remote application "
+                    f"{self._entity.name}")
```
tonyandrewmeyer, line 1664: Same as above:

```suggestion
                    f'{self._backend.app_name} cannot write the data of remote application '
                    f'{self._entity.name}')
```

ops/model.py
```diff
@@ -1676,9 +1675,8 @@ def _validate_write(self, key: str, value: str):
             # is it OUR UNIT's?
             if self._backend.unit_name != self._entity.name:
                 raise RelationDataAccessError(
-                    "{} cannot write databag of {}: not the same unit.".format(
-                        self._backend.unit_name, self._entity.name
-                    )
+                    f"{self._backend.unit_name} cannot write databag of {self._entity.name}: "
+                    "not the same unit."
```
tonyandrewmeyer, line 1679: More double-quotes!

```suggestion
                    f'{self._backend.unit_name} cannot write databag of {self._entity.name}: '
                    f'not the same unit.'
```

ops/model.py
```diff
@@ -2950,8 +2949,9 @@ def _format_action_result_dict(input: Dict[str, Any],
             # other exceptions raised on key validation...
             raise ValueError(f'invalid key {key!r}; must be a string')
         if not _ACTION_RESULT_KEY_REGEX.match(key):
-            raise ValueError("key '{!r}' is invalid: must be similar to 'key', 'some-key2', or "
-                             "'some.key'".format(key))
+            raise ValueError(
+                f"key '{key!r}' is invalid: must be similar to 'key', 'some-key2', "
+                "or 'some.key'")
```
tonyandrewmeyer, line 2954: And more :) - this oddly has repr() but also explicit `'` - I think that's a mistake that we can fix here.

```suggestion
                f"key {key!r} is invalid: must be similar to 'key', 'some-key2', "
                f"or 'some.key'")
```

ops/testing.py
```diff
@@ -990,9 +990,11 @@ def add_relation_unit(self, relation_id: int, remote_unit_name: str) -> None:
         app = relation.app
         if not remote_unit_name.startswith(app.name):
             warnings.warn(
-                'Remote unit name invalid: the remote application of {} is called {!r}; '
-                'the remote unit name should be {}/<some-number>, not {!r}.'
-                ''.format(relation_name, app.name, app.name, remote_unit_name))
+                'Remote unit name invalid: '
```
tonyandrewmeyer, line 993: As above, I think the ruff formatter wants the leading bit consistent.

```suggestion
                f'Remote unit name invalid: '
```

pyproject.toml
```diff
@@ -119,6 +119,8 @@ ignore = [
     # No explicit `stacklevel` keyword argument found
     "B028",
 
+    # Return condition directly, sometimes the error message isn't correct.
```
tonyandrewmeyer, line 122: I don't think the error message is incorrect, it's just not as clear as it could be - it doesn't explicitly say that you might need to negate the condition in some cases. But we can leave this comment as-is if you prefer :)

I think the actual reason we are skipping this is that we believe that sometimes it's more readable the other way.

tonyandrewmeyer (approved):
Looks great! A couple of suggestions that I think will make less work for the formatting later, but skip them if you prefer.

benhoyt (commented):
Looks good -- agree with Tony's recommendations though.

benhoyt (approved):

PR #1181, opened 2024-04-09
fix (harness): only inspect the source file if it will be used

tonyandrewmeyer:
When experimenting in the REPL it would be convenient to be able to do something like this:

```python
>>> import ops.testing
>>> class MyCharm(ops.CharmBase):
...  pass
... 
>>> h = ops.testing.Harness(MyCharm, meta="{}")
```

This currently fails with `OSError: source code not available`. This PR changes that, so that it will work.

We currently always use `inspect.getfile` to get the source file of the charm, even if metadata is provided to Harness, or if the default metadata would suffice. This PR makes a small adjustment so that we'll only do the `inspect` when needed, and also if it fails we'll fall back to the default metadata.

The change is made for config as well, for consistency.


benhoyt (commented):
This seems like a good change to me. However, can/should we have some test coverage for the new code paths?

tonyandrewmeyer:
> This seems like a good change to me. However, can/should we have some test coverage for the new code paths?

It's tricky because the place this actually fails (ie. raises) is in the REPL.

When the charm is defined in the normal `src/charm.py` location, we look in the directory above `src`, which would normally have the metadata in it. When the charm is defined in a test file (probably unlikely that charms are doing this, but ops does it a lot) like `tests/unit/test_charm.py`, we look in `tests`, which is almost certainly the wrong place, but we want to fail to find the metadata, so it's fine - similarly if it's ops and like `test/test_testing.py`, we look in the top level folder, and fail to find the metadata as we want.

However, if we're in the REPL then there is no source file.

I can probably mock out `inspect.getfile` to fail or similar - I presume that does something like `sys.modules[charm_class.__module__].__file__` so messing with `sys.modules` or `__module__` would probably also work.

test/test_testing.py
```diff
@@ -1402,6 +1402,26 @@ def test_config_from_directory_charmcraft_yaml(self):
         self.assertEqual(harness.model.config['opt_int'], 1)
         self.assertIsInstance(harness.model.config['opt_int'], int)
 
+    def test_config_in_repl(self):
```
tonyandrewmeyer, line 1405: @benhoyt in the past you've indicated a preference for one Harness per test. I can adjust this to do that if you'd prefer - this seems fairly contained as a logical unit to me, but I don't have strong feelings about it.

benhoyt: No, this seems okay and fine to reuse the patching/etc. Nit, if you want: add a blank line before the second `harness = ...`

benhoyt (approved):

IronCore864 (approved):

PR #1185, opened 2024-04-10
feat!: change CharmBase.config to have Secrets when the type is `secret`

tonyandrewmeyer:
If I have this config metadata:

```yaml
config:
  options:
    myint:
      type: int
    myfloat:
      type: float
    mybool:
      type: boolean
    mystr:
      type: string
    mysecret:
      type: secret
```

Then, assuming some value has been set for each of them, `self.config["myint"]` gives me an int, `self.config["myfloat"]` gives me a float, `self.config["mybool"]` gives me a bool, `self.config["mystr"]` gives me a str, and `self.config["mysecret"]` gives me a ... str.

This PR changes that behaviour so that instead `self.config["mysecret"]` will return a `Secret` object with the `id` set. The actual secret content and metadata is still only retrieved when `get_content`, `get_info`, etc are used.

This avoids having to do `secret = self.model.get_secret(id=config["mysecret"])` every time.

Juju prevents setting the config value to an invalid secret ID (both something that doesn't look like a secret ID and also a secret that doesn't exist). You can, at the moment, set it to an ID of a secret that the charm does not have permission to view. In that case, the first time the secret is used (e.g. `get_content()`) it will raise a `ModelError`. This is slightly different from now, when you'd generally get that error when using `get_secret`, but it seems reasonable behaviour.

The Harness is also updated so that you can provide either a string (as before) or a Secret object when using `update_config`. For example:

```python
secretfoo_id = harness.add_user_secret({"password": "xxxxxxxx"})
secretbar_id = harness.add_user_secret({"certificate": "xxxxxxxx"})
secretbar = harness.charm.model.get_secret(id=secretbar_id)
harness.update_config({'secretfoo': secretfoo_id, 'secretbar': secretbar})
```

In order to make this work, the `CharmMeta` now loads the config values (from `config.yaml`). This seems generally useful anyway (and was discussed in the `_Manager` PR).

TODO:

More complete tests, particularly around the meta config loading (and should it be wanting the options dict or the top level?). Loading the config from YAML is not exercised by any tests at the moment.

There's a comment in the Harness code:

> NOTE: jam 2020-03-01 Note that this sort of works "by accident". Config
> is a LazyMapping, but its _load returns a dict and this method mutates
> the dict that Config is caching. Arguably we should be doing some sort
> of charm.framework.model.config._invalidate()

I think I could tidy that up as part of this work without it significantly changing the PR size.


tonyandrewmeyer:
We decided against doing this:

* The Juju team feels like relation data and config data should behave in the same way, and that even though they don't currently, moving further apart is not good.
* Ben is concerned about having a config value that isn't a built-in/simple type, and whether there might be consequences of this that we have not considered.
* I'm concerned about breaking backwards compatibility, even though it's only one release/month. Ian made a good point that one rushed decision (which adding type:secret support was, in my opinion) shouldn't lead to a second rushed decision.

The team will talk more in the future about how config data (and possibly relation data) could give a better experience to charmers.

Issue #1177, opened 2024-04-08
Refactor Ops tests to use pytest instead of unittest


benhoyt:
We've been moving to pytest for Harness and other examples in our docstrings and docs. However, the Ops tests themselves are still all written in unittest with ugly `self.assertEqual(...)` calls and the like. Let's go all in on pytest here too.

This is going to be a huge diff -- maybe we can do 1 file per PR to split it up?

Let's look to see if there are existing tools that do this. If not, we may want to consider writing our own AST-based hack that can do most of it.

tonyandrewmeyer:
Should this be closed now? There was potential further work, if I recall correctly, but the main change is complete, right?

benhoyt:
Yes, thanks!

PR #1186, opened 2024-04-11
test: preview unittest to pytest conversion changes

IronCore864:
This PR is for preview and discussion only.

Issue: https://github.com/canonical/operator/issues/1177

Some status update on this matter:

## Existing Tools

I have done some research and tested two tools: [`unittest2pytest`](https://github.com/pytest-dev/unittest2pytest) and [`pytestify`](https://github.com/dannysepler/pytestify). They work differently:

- `unittest2pytest` uses [lib2to3](https://github.com/python/cpython/tree/3.12/Lib/lib2to3/) which allows you to write your own "fixers" to update code, and it seems unittest2pytest only implementes two fixers, one [removes class](https://github.com/pytest-dev/unittest2pytest/blob/master/unittest2pytest/fixes/fix_remove_class.py), leaving function only; the other [fixes assert](https://github.com/pytest-dev/unittest2pytest/blob/master/unittest2pytest/fixes/fix_self_assert.py). I gave it a test but it had trouble parsing our test code, throwing a lot of errors and I'm too impatient to fix it for unittest2pytest, so I gave up here.
- `pytestify` is a small piece of program written by an independent person (seems to be so) which is AST based, it parses code correctly, and has [slightly more "fixes"](https://github.com/dannysepler/pytestify/tree/main/pytestify/fixes). It has some bugs, for example, `self.assertEqual(ops.JujuVersion(a) < ops.JujuVersion(b), expected)` got converted to `assert ops.JujuVersion(a) == ops.JujuVersion(b) == expected`, where `assert (ops.JujuVersion(a) == ops.JujuVersion(b)) == expected` is expected.

Quick conclusion: out of these two (both not updated for quite a long time), although `unittest2pytest` got more stars, `pytestify` is slightly more usable than `unittest2pytest`, so maybe `pytestify` can be used as a start. But there is still a lot of work to be done. A few examples:

## Cases Not Handled by Existing Tools

- We have some tests where we define test cases in a list like `test_cases = [ (...), (...), ...]` then we run a loop `for foo, bar in test_cases: check_something()`. This needs to be converted to `@pytest.mark.parametrize`. And sometimes it's not a for loop but `self.subTest`. There are some extra detail here as well, for example, maybe in the `@pytest.mark.parametrize`, we want the parameter to be named `test_case` instead of `test_cases` since singular is what's passed to the test function.
- We have some mocks like `@unittest.mock.patch` that needs to be cnoveretd to `monkeypatch`, which needs to be imported, and whose parameter is different than unittest's patch.
- We have a lot of `unittest.addCleanup` calls which needs to be handled and this is not trivial, because a fixture needs to be extracted out of it, meaning we need to yield at the right place, then refer to the fixture in related test cases. In some cases it can be difficult since multiple things need to be set up and tear down, and in some cases it's even more complicated because it can either be called in a setup function or directly in a test. I have an example in this PR demonstrating required work for this kind of test, I had to create 3 fixtures to be used by a 4th one which is used in tests to make it readable and manageable.

The first two cases theoretically could be handled from AST's standpoint, but I'm not so sure about the third case. I have only inspected three test files: `test_infra.py` (because it's small), `test_jujuversion.py` (same reason), and `test_charm` (big but important), and I didn't fully go through the last one, so I'm pretty sure there will be (a lot) more cases need to be handled. For this three files that I inspected, I created this draft PR demonstrating the work required (partly `pytestify`, mostly manual).

## Initial Thoughts

This makes me think whether it's worth it to do the full conversion. We can build our own AST-based converter, but is it really worth it? To cover all cases and make everything automated, it might require tremendous effort. We can also combine `pytestify` and manual work, but that could be quite time-consuming (personally I don't mind doing so because the process makes me think about the test structure and how to organize things). Maybe a middle ground could be: when creating new tests, write in pytest style; when adding a test in an existing class, convert the whole class to pytest style with `pytestify` and some manual work; other than these, leave them alone.

What do you think? @benhoyt @tonyandrewmeyer

### DIFF
--- a/test/test_charm.py
+++ b/test/test_charm.py
@@ -21,6 +21,7 @@ import unittest
 import unittest.mock
 from pathlib import Path
 
+import pytest
 import yaml
 
 import ops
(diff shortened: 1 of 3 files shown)


test/test_charm.py
```diff
@@ -32,47 +33,52 @@
 from .test_helpers import fake_script, fake_script_calls
 
 
-class TestCharm(unittest.TestCase):
+@pytest.fixture
+def tmpdir():
```
benhoyt, line 37: There's already a built-in `tmpdir` fixture in pytest, right? https://docs.pytest.org/en/6.2.x/tmpdir.html

test/test_infra.py
```diff
@@ -34,35 +35,27 @@ def get_python_filepaths(include_tests: bool = True):
...
 
         environ = os.environ.copy()
         if 'PYTHONPATH' in environ:
             environ['PYTHONPATH'] = os.getcwd() + os.pathsep + environ['PYTHONPATH']
         else:
             environ['PYTHONPATH'] = os.getcwd()
         proc = subprocess.run([sys.executable, testfile], env=environ)
-        self.assertEqual(proc.returncode, 0)
+        assert proc.returncode == 0
+        os.unlink(testfile)
```
benhoyt, line 61: Probably should use pytest's built-in `tmp_path` fixture? https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture

benhoyt (commented):
Left a couple of nit comments about using the builtin pytest `tmpdir` / `tmp_path` fixtures where possible.

More generally, pytestify definitely looks like the better tool. I'd recommend we use pytestify plus manual work, and maybe we can even fork and modify pytestify to work better for Ops specifically (we're not writing a general tool). However, I'm sure there will always be manual work involved.

I wonder if we can do this in phases, for example:

1) Change the assertEqual etc to regular asserts. Pytest still does its nice assert introspection on `unittest` tests (from a quick test, but worth confirming), so we could do just this step without any other changes.
2) Convert test classes/methods to plain functions.
  - We can use the `request` fixture with `request.addfinalizer` to [add finalizers/cleanup directly](https://docs.pytest.org/en/6.2.x/fixture.html?highlight=addfinalizer#adding-finalizers-directly) -- that might help avoid the need to fixture-ize everything all at once, helping the change be more mechanical, and we can do the fixture-izing in a later phase.
  - We can look at using [pytest-subtests](https://pypi.org/project/pytest-subtests/) to avoid converting all the test cases and sub tests to parametrized tests right away (do that in a later phase).
3. Fixture-ize everything the pytest way.
4. Parametrize the test cases / subtests the pytest way.
5. Replace `unittest.mock.patch` with pytest-native `monkeypath` (maybe this is not needed)?

What do you think? Anyway, definitely worth further discussion at standup.

test/test_charm.py
```diff
@@ -32,47 +33,52 @@
...
-        def restore_env(env: typing.Dict[str, str]):
-            os.environ.clear()
-            os.environ.update(env)
-        self.addCleanup(restore_env, os.environ.copy())
 
-        os.environ['PATH'] = os.pathsep.join([
-            str(Path(__file__).parent / 'bin'),
-            os.environ['PATH']])
-        os.environ['JUJU_UNIT_NAME'] = 'local/0'
+@pytest.fixture
```
tonyandrewmeyer, line 43: For what it's worth, pytest has [monkeypatch functionality specifically for environment variables](https://docs.pytest.org/en/6.2.x/monkeypatch.html#monkeypatching-environment-variables) so my guess (completely untested) would be that we could use that here.

IronCore864:
Closing this preview PR, will do it in phases, phase 1 being converting asserts only.

PR #1192, opened 2024-04-16
test: refactor some for loop tests into pytest parametrize

IronCore864:
This PR tries to refactor (some) test cases with "for" loop or `subTest` into pytest's `parametrize` fixture functions.

Only relatively simple changes are made (more on this in section 3).

---

## 1 TestFiles Changed

- `test/test_jujuversion.py`
- `test/test_log.py`
- `test/test_pebble.py`
- `test/test_private.py`

---

## 2 Before/After Comparisons

_Num of test cases increased because of parametrize, coverage unchanged._

### 2.1 `test_jujuversion.py`

Before

```bash
$ python -m tox -e unit -- test/test_jujuversion.py
unit: commands[0]> coverage run --source=ops/ -m pytest --ignore=test/smoke -v --tb native test/test_jujuversion.py
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.11.8, pytest-7.4.4, pluggy-1.4.0 -- /Users/tiexin/work/operator2/.tox/unit/bin/python
cachedir: .tox/unit/.pytest_cache
rootdir: /Users/tiexin/work/operator2
collected 11 items
====================================================================== 11 passed in 0.10s ======================================================================
unit: commands[1]> coverage report
Name                       Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------
...
ops/jujuversion.py            80     10     48      4    86%   52-58, 62, 66, 76, 80
...
----------------------------------------------------------------------
TOTAL                       6497   4846   2545     32    22%
```

After

```bash
$ python -m tox -e unit -- test/test_jujuversion.py
unit: commands[0]> coverage run --source=ops/ -m pytest --ignore=test/smoke -v --tb native test/test_jujuversion.py
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.11.8, pytest-7.4.4, pluggy-1.4.0 -- /Users/tiexin/work/operator2/.tox/unit/bin/python
cachedir: .tox/unit/.pytest_cache
rootdir: /Users/tiexin/work/operator2
collected 71 items
====================================================================== 71 passed in 0.14s ======================================================================
unit: commands[1]> coverage report
Name                       Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------
...
ops/jujuversion.py            80     10     48      4    86%   52-58, 62, 66, 76, 80
...
----------------------------------------------------------------------
TOTAL                       6497   4846   2545     32    22%
```

### 2.2 `test_log.py`

Before

[...]

PR #1193, opened 2024-04-18
test: refactor test_charm.py to pytest fixtures and parametrize

IronCore864:
Refactoring `test_charm.py` to pytest style.

---

This PR contains a few major changes:

- a fake script helper fixture in pytest style to decouple the test case with the fake script functions
- use pytest's `parametrize` instead of a for loop
- use fixtures to setup/cleanup

_Note that this PR might show more changes because it's checked out based on the previous [unmerged PR here](https://github.com/canonical/operator/pull/1192)._

---

## 1 Fake Script Helper Fixture

Background: Previously, `fake_script_path` is set on the TestCase itself, then `self` is passed into the fake script functions, and `self.addCleanup()` is added on `self`. The test case is mixed with fake script, fake script adds clean up to the test case, making it hard to convert to pytest's fixture-style cleanup.

Changes: In [these two commits](https://github.com/canonical/operator/pull/1193/files/52c3928b0995640a3b2862cd9ba0050700922e6b..41b17bda93f7046b2678011f2921b34abf890c17), a new class `FakeScriptFixture` is created to share `fake_script_path` between the two methods `fake_script()` and `fake_script_calls`.

After the change, we can use it like:

```python
@pytest.fixture
def fake_script_fixture(monkeypatch: MonkeyPatch, tmp_path: pathlib.Path):
    return FakeScriptFixture(monkeypatch, tmp_path)
```

This fixture's purpose is to add another layer on top of the FakeScriptFixture class so that we don't have to pass monkeypatch/tmp_path fixtures to it everytime we use it. Then, in the test case, we can do something like:

```python
def test_foo(fake_script_fixture):
    fake_script_fixture.fake_script("test", "echo 'hello'")
```

---

## 2 Use Parametrize/Fixture to Refactor `test_charm.py`

- Tests no longer inherit from `unittest.TestCase`.
- Use fixture to create framework and teardown.
- Use custom marker (added to `pyproject.toml` to depress unrecognized warning) to pass data to the framework fixture so that it can be customized.

Still in draft, please review and merge [PR 1192](https://github.com/canonical/operator/pull/1192) first so that this PR shows less change. I'll do a final review tomorrow.


test/test_charm.py
```diff
@@ -28,50 +27,51 @@
...
 
-        class TestCharmEvents(ops.CharmEvents):
-            custom = ops.EventSource(CustomEvent)
 
-        # Relations events are defined dynamically and modify the class attributes.
-        # We use a subclass temporarily to prevent these side effects from leaking.
-        ops.CharmBase.on = TestCharmEvents()  # type: ignore
+@pytest.fixture
+def framework(request: pytest.FixtureRequest, tmp_path: pathlib.Path):
+    marker = request.node.get_closest_marker("charm_meta")  # type: ignore
```
benhoyt, line 59: At first glance I really don't love this indirect way of passing data to the fixture. Are there better ways to approach it, or maybe we shouldn't use fixtures here? Maybe we can use a function (with a `request.addfinaliser`?).

Speaking of "indirect", there's also [pytest's "indirect" feature](https://stackoverflow.com/a/33879151/68707), which I don't really understand yet, but might be worth looking at.

test/test_charm.py
```diff
@@ -323,21 +315,21 @@ def _on_stor4_attach(self, event: ops.StorageAttachedEvent):
...
             echo '["disks/0"]'
             """,
         )
 
-        assert self.meta.storages['stor1'].multiple_range is None
-        assert self.meta.storages['stor2'].multiple_range == (2, 2)
-        assert self.meta.storages['stor3'].multiple_range == (2, None)
-        assert self.meta.storages['stor-4'].multiple_range == (2, 4)
-        assert self.meta.storages['stor-plus'].multiple_range == (10, None)
+        meta = request.node.get_closest_marker("charm_meta").args[0]  # type: ignore
```
benhoyt, line 325: Yeah, I think this is pretty icky -- taking what was a simple attribute/variable and routing it indirectly via pytest "markers" by name (and we have to add it to the `pyproject.toml` even)...

benhoyt (commented):
I really like the fixture for the `fake_script` thing, but I'm not loving how the charm_meta / framework thing works out here. Let's discuss some of this further at our daily sync.

test/test_charm.py
```diff
@@ -419,20 +409,20 @@ def test_relations_meta(self):
     optional: true
 ''')
 
-        assert self.meta.requires['database'].interface_name == 'mongodb'
-        assert self.meta.requires['database'].limit == 1
-        assert self.meta.requires['database'].scope == 'container'
-        assert not self.meta.requires['database'].optional
+        assert meta.requires['database'].interface_name == 'mongodb'  # type: ignore
```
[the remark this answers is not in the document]
IronCore864: Originally I had the meta as a `pytest.mark` then I figured it was not necessary so I changed it back but forgot to remove the added type ignores. Now that I'm doing a self-review, I have found this left-over. I will remove these in the next commit together with other possible changes.

IronCore864:
Many comments are interrelated, so I will reply to all of them here in a single comment to take some notes of my thoughts:

## Autouse

The original test case has the env and events mocking defined in the setup function, meaning they are applied to all test cases. Autouse is exactly the same, except instead of defining code in a setup function, you define it in a fixture with autouse=True. So, to me, this change makes sense.

I think it's a solid point that if the users are not familiar with the autouse function, it's easy to go wrong since it's applied on all test cases without explicitly saying so. So I think maybe we should not use autouse at all.

This creates a minor issue where you have the fixture in the test function's parameter but you don't use it:

```python
@pytest.fixture
def env(monkeypatch):
    monkeypatch.setenv("FOO", "BAR")
    yield


def test_foo(env):  # 
    # do something but "env" is not accessed, showing up in IDE prompts
````

## Reasons for 3 Fixtures Instead of Only One

If we get rid of autouse and we don't want the "not accessed", we can combine the three fixtures into one, but I think it loses some readability. By splitting the original `setUp` (and the `create_framework` func) into 3 fixtures, it is more readable, because:

- you get three short simple functions that all have the keyword `yield`, before which is the setup and after which is the teardown
- each function does one thing and one thing only

```python
@pytest.fixture(autouse=True)
def env(monkeypatch):
    monkeypatch.setenv(xxx)
    yield


@pytest.fixture(autouse=True)
def events():
    ops.CharmBase.on = "some mock"
    yield
    ops.CharmBase.on = "change back to original value"


@pytest.fixture
def framework(request: pytest.FixtureRequest, tmp_path: pathlib.Path):
    framework = ops.Framework(xxx)
    yield framework
    framework.close()
```

By reading this piece of code, it can be easily inferred that all test cases do the three following things:

- mock env
- mock events
- create a framework object

If we combine them into one, for example:

```python
@pytest.fixture
def framework(
    request: pytest.FixtureRequest,
    tmp_path: pathlib.Path
):
    monkeypatch.setenv("PATH", str(Path(__file__).parent / 'bin'), prepend=os.pathsep)
    monkeypatch.setenv("JUJU_UNIT_NAME", "local/0")

    class CustomEvent(ops.EventBase):
        pass

    class TestCharmEvents(ops.CharmEvents):
        custom = ops.EventSource(CustomEvent)

    # Relations events are defined dynamically and modify the class attributes.
    # We use a subclass temporarily to prevent these side effects from leaking.
    ops.CharmBase.on = TestCharmEvents()  # type: ignore

    marker = request.node.get_closest_marker("charm_meta")  # type: ignore
    meta = marker.args[0] if marker else ops.CharmMeta()  # type: ignore
    model = ops.Model(meta, _ModelBackend('local/0'))  # type: ignore
    # we can pass foo_event as event_name because we're not actually testing dispatch
    framework = ops.Framework(SQLiteStorage(':memory:'), tmp_path, meta, model)  # type: ignore

    yield framework

    ops.CharmBase.on = ops.CharmEvents()  # type: ignore
    framework.close()
```

It's still short enough to be readable, but I think it's less clear than 3 separate fixtures and harder to figure out the logic.

## Create Framework as a Function, not a Fixture


I think we can keep the framework creation part as a function, just like before. Example:

```python
def create_framework(self, request: pytest.FixtureRequest, tmp_path: pathlib.Path):
    monkeypatch.setenv("PATH", str(Path(__file__).parent / 'bin'), prepend=os.pathsep)
    monkeypatch.setenv("JUJU_UNIT_NAME", "local/0")

    class CustomEvent(ops.EventBase):
        pass

    class TestCharmEvents(ops.CharmEvents):
        custom = ops.EventSource(CustomEvent)

    ops.CharmBase.on = TestCharmEvents()

    def cleanup():
        ops.CharmBase.on = ops.CharmEvents()

    request.addfinalizer(cleanup)
    
    model = ops.Model(meta, _ModelBackend('local/0'))  # type: ignore
    framework = ops.Framework(SQLiteStorage(':memory:'), tmp_path, self.meta, model)  # type: ignore
    request.addfinalizer(framework.close)

    return framework
```

As much as I hate the multi-line custom marker just to pass data into the framework fixture, turning it into a function instead of a fixture isn't the perfect solution since:

- less readable
- every time you call it, you have to pass in two parameters: `self.create_framework(request, tmp_path)`, and you have to include request, tmp_path in almost all test cases, like:

```python
class TestClass:
    def test_foo(self, request, tmp_path):
        self.create_framework(request, tmp_path)
```

## Compromise

Maybe a compomise is a mix of everything mentioned above?

```python
import functools


@pytest.fixture(autouse=True)
def env(monkeypatch):
    monkeypatch.setenv(xxx)
    yield


@pytest.fixture(autouse=True)
def events():
    ops.CharmBase.on = "some mock"
    yield
    ops.CharmBase.on = "change back to original value"


class TestClass:
    @pytest.fixture
    def create_framework(request, tmp_path):
        return functools.partial(self._real_create_framework, request, tmp_path)    

    def _real_create_framework(self, request, tmp_path):
        model = ops.Model(meta, _ModelBackend('local/0'))
        framework = ops.Framework(SQLiteStorage(':memory:'), tmp_path, self.meta, model)
        request.addfinalizer(framework.close)
        return framework

    def test_foo(self, create_framework):
        self.meta = xxx
        framework = create_framework()
```

IronCore864:
According to our discussion in the daily, we decide to:
- make `create_framework` a function, not a fixture: `create_framework(request, *, meta=None)`
- put it outside the test class (in `test_helper.py` maybe) for reusability
- remove tmp_path fixture so that we have one less fixture to pass to the function and we will do the tmp path creation/cleanup manually in the `create_framework` function
- remove autouse, which means no more `env` and `events` fixtures any more

I'll do a separate commit later to reflect this change and refactor the test cases accordingly.

test/test_charm.py
```diff
@@ -25,53 +23,18 @@
...
-from ops.storage import SQLiteStorage
+from ops.model import ModelError
 
-from .test_helpers import fake_script, fake_script_calls
+from .test_helpers import FakeScript, create_framework
 
 
-class TestCharm(unittest.TestCase):
+@pytest.fixture
+def fake_script_fixture(monkeypatch: pytest.MonkeyPatch, tmp_path: pathlib.Path):
```
IronCore864, line 32: If I move this to test_helpers, lint will show an error about unused import in the test cases, so I'm keeping it here.

test/test_charm.py
```diff
@@ -83,11 +46,12 @@ def __init__(self, *args: typing.Any):
             def _on_start(self, event: ops.EventBase):
                 self.started = True
 
+        framework = create_framework(monkeypatch, request)
```
IronCore864, line 49: Since the event mocking is part of the function, it needs to appear before asserting events, hence moving it up here.

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+    framework = ops.Framework(SQLiteStorage(':memory:'), tmpdir, meta, model)  # type: ignore
+    request.addfinalizer(framework.close)
+
+    return framework
+
+
+class FakeScript:
+    def __init__(self,
+                 monkeypatch: pytest.MonkeyPatch,
+                 tmp_path: pathlib.Path,
```
benhoyt, line 119: I think we should try to have a consistent API across these test helpers. For example, use `request` (and only `request`) as a parameter for all of them.

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+                 fake_script_path: typing.Union[pathlib.Path, None] = None):
```
benhoyt, line 120: This could be called just `path`, given the context.

test/test_charm.py
benhoyt: Hmm, is there a better way to do handle this? This will be repeated in about 5 different test files, so it'd be nice if we can avoid it. Is there a standard way to import pytest fixtures? Or should we just silence the linter error?

benhoyt: Also, do you think we should just call this `fake_script`? That seems nicer than the lengthy `fake_script_fixture` in the tests. It seems like fixtures don't normally have `fixture` in the name.

test/test_charm.py
```diff
@@ -25,53 +23,18 @@
...
-        # we can pass foo_event as event_name because we're not actually testing dispatch
-        framework = ops.Framework(SQLiteStorage(':memory:'), self.tmpdir, self.meta,
-                                  model)
-        self.addCleanup(framework.close)
-        return framework
-
-    def test_basic(self):
 
+class TestCharm:
+    def test_basic(self, monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureRequest):
```
benhoyt, line 37: Are you planning to move these to the top level too? There's also `TestMeta` in this file, but those functions could be renamed slightly, for example `test_links` -> `test_meta_links`.

test/test_helpers.py
tonyandrewmeyer: And this is another one that would be more typically `Optional[pathlib.Path]`.

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+        self.monkeypatch = monkeypatch
+        if fake_script_path is None:
+            self.fake_script_path = tmp_path
+            self.monkeypatch.setenv(
+                "PATH",
+                str(self.fake_script_path),
+                prepend=os.pathsep)  # type: ignore
```
tonyandrewmeyer, line 127: Maybe this is from the original, but do you know why we're having to ignore the type here? It seems like we clearly have two strings, which is what I'd expect it to want - is os.pathsep not always set?

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+    def write(self,
+              name: str,
+              content: str):
+        template_args: typing.Dict[str, str] = {
+            'name': name,
+            'path': self.fake_script_path.as_posix(),  # type: ignore
+            'content': content,
+        }
+
+        path: pathlib.Path = self.fake_script_path / name  # type: ignore
```
tonyandrewmeyer, line 140: It seems strange we need an ignore here, too. Maybe once we've moved to just managing the temporary location ourselves rather than using `tmp_path` this will go away.

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+        with path.open('wt') as f:  # type: ignore
```
tonyandrewmeyer, line 141: And this one particularly - `path` is defined as a `pathlib.Path` immediately above, so why do we need to ignore the type here?

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+            # Before executing the provided script, dump the provided arguments in calls.txt.
+            # ASCII 1E is RS 'record separator', and 1C is FS 'file separator', which
+            # seem appropriate.
+            f.write(  # type: ignore
+                '''#!/bin/sh
+    {{ printf {name}; printf "\\036%s" "$@"; printf "\\034"; }} >> {path}/calls.txt
```
tonyandrewmeyer, line 147: This is probably also in the original, but I think it's a bit confusing that we reference the characters by the hex value in the comment and then use the octal value in printf. Maybe just `RS 'record separator' (octal 036 in ASCII)` or something like that would be clearer to readers?

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+    {content}'''.format_map(template_args))
+        os.chmod(str(path), 0o755)  # type: ignore  # noqa: S103
+        # TODO: this hardcodes the path to bash.exe, which works for now but might
+        #       need to be set via environ or something like that.
+        path.with_suffix(".bat").write_text(  # type: ignore
+            f'@"C:\\Program Files\\git\\bin\\bash.exe" {path} %*\n')
```
tonyandrewmeyer, line 153: I think we've dropped Windows support, so do we need this any more?

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+
+    def calls(self, clear: bool = False) -> typing.List[typing.List[str]]:
+        calls_file: pathlib.Path = self.fake_script_path / 'calls.txt'  # type: ignore
```
tonyandrewmeyer, line 156: There are again a lot of type:ignores here, can we reduce that somehow with a cast or assert or something?

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+        if not calls_file.exists():  # type: ignore
+            return []
+
+        # newline and encoding forced to linuxy defaults because on
+        # windows they're written from git-bash
```
tonyandrewmeyer, line 161:
```suggestion
        # Newline and encoding forced to Linux-y defaults because on
        # windows they're written from git-bash.
```

Although as noted above I don't think we really need the Windows support any more.

test/test_charm.py
IronCore864: I have figured out a way to share fixtures across multiple files. I will change the name to `fake_script`, too, shorter name is better and the `_fixture` suffix seems unnecessary according to standard pytest fixtures.

IronCore864:
According to our discussion, `fake_script` fixture will be defined in `test_helpers.py`, not `conftest.py`. However, there is already a function with the same name in `test_helpers.py` which is referred by other test files other than `test_charm.py`.

So, I will put the `fake_script` fixture in `test_charm.py` temporarily; after refactoring all test files, I will move it to `test_helpers.py`.

test/test_charm.py
IronCore864: Makes sense, this file is mainly one big test class with a smaller one.

test/test_helpers.py
```diff
@@ -74,6 +77,95 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+def create_framework(
+        monkeypatch: pytest.MonkeyPatch,
+        request: pytest.FixtureRequest,
+        *,
+        meta: typing.Union[ops.CharmMeta, None] = None):
+    monkeypatch.setenv("PATH", str(pathlib.Path(__file__).parent / 'bin'), prepend=os.pathsep)
+    monkeypatch.setenv("JUJU_UNIT_NAME", "local/0")
+
+    tmpdir = pathlib.Path(tempfile.mkdtemp())
+    request.addfinalizer(functools.partial(shutil.rmtree, str(tmpdir)))
```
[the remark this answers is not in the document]
IronCore864: I think if rmtree throws an error, it will be passed to the test, which is captured and reported as a test failure, right? Same as unittest addCleanup? Do we need to surrounded it with a try/except?

test/test_helpers.py
IronCore864: Resolved by changing back to not using monkeypatch to make the API consistent, only request is used.

test/test_helpers.py
IronCore864: Copied from the original. Shall we keep it or not?

benhoyt: I'm with Tony that we should remove it, but I think it should be done in a separate PR to the refactoring, rather than snuck in this one.

test/test_helpers.py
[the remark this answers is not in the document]
IronCore864: Resolving this because if rmtree fails, the test case will too. Confirmed with a simple test case:

```python
import pytest


def test_with_fixture(request: pytest.FixtureRequest):
    def finalizer():
        raise Exception()

    request.addfinalizer(finalizer)

    assert True

```

test/test_charm.py
```diff
@@ -12,188 +12,155 @@
...
+    charm.on.start.emit()
+    # check that the event has been seen by the decorator
+    assert len(events) == 1
+    # check that the event has been seen by the observer
+    assert isinstance(charm.seen, ops.StartEvent)
+
+
+def test_observer_not_referenced_warning(
+        caplog: pytest.LogCaptureFixture,
+        request: pytest.FixtureRequest):
```
benhoyt, line 101: Nit: why don't we standardise on the `request` parameter always coming first, for a bit of consistency? It seems like it should go first because the "test context" (request) is kind of primary.

test/test_charm.py
```diff
@@ -419,45 +389,94 @@ def test_relations_meta(self):
...
+                    }}))
+
+        meta = ops.CharmMeta.from_charm_root(td)
+        assert meta.name == "bob"
+        assert meta.requires['foo'].interface_name == "bar"
+        assert not meta.actions['foo'].additional_properties
+        assert meta.actions['foo'].description == "foos the bar"
+
+
+def _setup_test_action(fake_script: typing.Callable[..., None]):
```
benhoyt, line 471: I think it would be clearer (and a simpler type) if `fake_script` was a `FakeScript` here, and we used `fake_script.write()` in the function, like the other calls.

test/test_charm.py
```diff
@@ -477,155 +496,120 @@ def _get_action_test_meta(cls):
...
+@pytest.mark.parametrize("bad_res", [
+    {'a': {'b': 'c'}, 'a.b': 'c'},
+    {'a': {'B': 'c'}},
+    {'a': {(1, 2): 'c'}},
+    {'a': {None: 'c'}},
+    {'aBc': 'd'}
+])
+def test_invalid_action_results(request: pytest.FixtureRequest,
+                                fake_script: FakeScript,
+                                bad_res: typing.Dict[str, typing.Any]):
```
benhoyt, line 547: Shall we format all of these multiline ones as Ruff/Black does, to avoid further reformatting in future?

```suggestion
def test_invalid_action_results(
    request: pytest.FixtureRequest,
    fake_script: FakeScript,
    bad_res: typing.Dict[str, typing.Any],
):
```

test/test_charm.py
```diff
@@ -697,262 +682,286 @@ def test_containers_storage_multiple_mounts(self):
...
-            ['status-set', '--application=True', 'active', ''],
-            ['status-set', '--application=True', 'unknown', ''],
-        ]
-
-
-class TestCharmMeta(unittest.TestCase):
-    def test_links(self):
-        # Each type of link can be a single item.
-        meta = ops.CharmMeta.from_yaml("""
+@pytest.mark.parametrize("statuses,expected", [
```
benhoyt, line 913: Nice little refactoring here -- I think this parameterization makes sense.

test/test_charm.py
```diff
@@ -12,188 +12,155 @@
...
+from ops.model import ModelError
 
-from .test_helpers import fake_script, fake_script_calls
+from .test_helpers import FakeScript, create_framework
 
 
-class TestCharm(unittest.TestCase):
+@pytest.fixture
+def fake_script(request: pytest.FixtureRequest) -> FakeScript:
+    return FakeScript(request)
```
benhoyt, line 32: We had discussed we were planning to import this from `test_helpers`, even if it meant a Pyright/lint ignore comment, to avoid the duplication.

test/test_helpers.py
```diff
@@ -74,6 +76,101 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+
+class FakeScript:
+    def __init__(self,
+                 request: pytest.FixtureRequest,
+                 path: typing.Optional[pathlib.Path] = None):
+        if path is None:
+            fake_script_path = tempfile.mkdtemp('-fake_script')
+            self.path = pathlib.Path(fake_script_path)
+            old_path = os.environ["PATH"]
+            os.environ['PATH'] = os.pathsep.join([fake_script_path, os.environ["PATH"]])
```
benhoyt, line 127: Do we not need to modify `os.environ` if `path` is passed?

benhoyt (commented):
This is looking really good now, I think. Just some minor comments. By the way, the diff is much easier to view with "hide whitespace" turned on in the GitHub UI.

test/test_charm.py
IronCore864: It's kept here temporarily, because there is a function with the same name in the `test_helpers.py`. When all files are refactored, it will be moved.

test/test_helpers.py
IronCore864: I was wondering the same when I saw this (got this logic from the original function), but after running all tests, it seems we don't need to do so.

test/test_charm.py
```diff
@@ -290,124 +259,125 @@ def _on_stor4_attach(self, event: ops.StorageAttachedEvent):
...
+        'StorageAttachedEvent',
+        'StorageAttachedEvent',
+    ]
+
+
+def test_workload_events(request: pytest.FixtureRequest):
+
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 329:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -290,124 +259,125 @@ def _on_stor4_attach(self, event: ops.StorageAttachedEvent):
...
-                self.seen.append(type(event).__name__)
+        def on_any_pebble_ready(self, event: ops.PebbleReadyEvent):
+            self.seen.append(type(event).__name__)
 
-        # language=YAML
-        self.meta = ops.CharmMeta.from_yaml(metadata='''
+        def on_any_pebble_custom_notice(self, event: ops.PebbleCustomNoticeEvent):
+            self.seen.append(type(event).__name__)
+
+    # language=YAML
```
tonyandrewmeyer, line 347: Same as above.

test/test_charm.py
```diff
@@ -290,124 +259,125 @@ def _on_stor4_attach(self, event: ops.StorageAttachedEvent):
...
-            'PebbleReadyEvent',
-            'PebbleCustomNoticeEvent',
-            'PebbleCustomNoticeEvent',
-        ]
 
-    def test_relations_meta(self):
-        # language=YAML
-        self.meta = ops.CharmMeta.from_yaml('''
+def test_relations_meta():
+    # language=YAML
```
tonyandrewmeyer, line 379: And another one - they are a bit more common than I had thought.

test/test_charm.py
```diff
@@ -477,155 +496,123 @@ def _get_action_test_meta(cls):
...
-            charm.on.start_action.emit(id='2')
 
-    def test_action_event_defer_fails(self):
-        self._test_action_event_defer_fails('action')
+def test_action_events(request: pytest.FixtureRequest, fake_script: FakeScript):
+
+    class MyCharm(ops.CharmBase):
+
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 505:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -477,155 +496,123 @@ def _get_action_test_meta(cls):
...
+def test_invalid_action_results(
+    request: pytest.FixtureRequest,
+    fake_script: FakeScript,
+    bad_res: typing.Dict[str, typing.Any]
+):
+
+    class MyCharm(ops.CharmBase):
+
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 554:
```suggestion
        def __init__(self, framework: ops.Framework:
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
+        'SecretRotateEvent',
+        'SecretRemoveEvent',
+        'SecretExpiredEvent',
+    ]
+
+
+def test_collect_app_status_leader(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 749:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
+    assert fake_script.calls(True) == [
+        ['is-leader', '--format=json'],
+        ['status-set', '--application=True', 'blocked', 'first'],
+    ]
+
+
+def test_collect_app_status_no_statuses(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 774:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
+    ]
 
-        fake_script(self, 'is-leader', 'echo false')
 
-        charm = MyCharm(self.create_framework())
-        ops.charm._evaluate_status(charm)
+def test_collect_app_status_non_leader(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 794:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

There are more of these than I anticipated. Feel free to push back and say we should do it some other time.

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
-        ops.charm._evaluate_status(charm)
 
-        assert fake_script_calls(self, True) == [
-            ['is-leader', '--format=json'],
-            ['status-set', '--application=False', 'blocked', 'first'],
-        ]
+def test_collect_unit_status(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 814:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
-    def test_collect_app_and_unit_status(self):
-        class MyCharm(ops.CharmBase):
-            def __init__(self, *args: typing.Any):
-                super().__init__(*args)
-                self.framework.observe(self.on.collect_app_status, self._on_collect_app_status)
-                self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
+def test_collect_unit_status_no_statuses(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 840:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
 
-    def test_add_status_type_error(self):
-        class MyCharm(ops.CharmBase):
-            def __init__(self, *args: typing.Any):
-                super().__init__(*args)
-                self.framework.observe(self.on.collect_app_status, self._on_collect_status)
+def test_collect_app_and_unit_status(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 861:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -697,262 +685,273 @@ def test_containers_storage_multiple_mounts(self):
...
 
-        fake_script(self, 'is-leader', 'echo true')
-        fake_script(self, 'status-set', 'exit 0')
 
-        charm = MyCharm(self.create_framework(), statuses=['blocked', 'error'])
-        ops.charm._evaluate_status(charm)
+def test_add_status_type_error(request: pytest.FixtureRequest, fake_script: FakeScript):
+    class MyCharm(ops.CharmBase):
+        def __init__(self, *args: typing.Any):
+            super().__init__(*args)
```
tonyandrewmeyer, line 888:
```suggestion
        def __init__(self, framework: ops.Framework):
            super().__init__(framework)
```

test/test_charm.py
```diff
@@ -12,188 +12,156 @@
...
+                self.framework.observe(self.on[rel].relation_departed, self.on_any_relation)
+                self.framework.observe(self.on[rel].relation_broken, self.on_any_relation)
+
+        def on_any_relation(self, event: ops.RelationEvent):
+            assert event.relation.name == 'req1'
+            assert event.relation.app is not None
+            assert event.relation.app.name == 'remote'
+            self.seen.append(type(event).__name__)
+
+    # language=YAML
```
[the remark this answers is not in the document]
benhoyt: It's probably some IDE highlighting marker. I'm happy to remove them, but let's save that for another PR.

test/test_helpers.py
```diff
@@ -74,6 +76,101 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+        request: pytest.FixtureRequest,
+        *,
+        meta: typing.Optional[ops.CharmMeta] = None):
+    env_backup = os.environ.copy()
+    os.environ['PATH'] = os.pathsep.join([
+        str(pathlib.Path(__file__).parent / 'bin'),
+        os.environ['PATH']])
+    os.environ['JUJU_UNIT_NAME'] = 'local/0'
+
+    tmpdir = pathlib.Path(tempfile.mkdtemp())
```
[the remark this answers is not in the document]
benhoyt: I agree, though I'd prefer to leave changes like this for future work to avoid muddying the waters in the pytest-refactoring PRs.

test/test_helpers.py
```diff
@@ -74,6 +76,101 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+                shutil.rmtree(self.path)
+                os.environ['PATH'] = old_path
+
+            request.addfinalizer(cleanup)
+        else:
+            self.path = path
+
+    def write(self,
+              name: str,
+              content: str):
```
benhoyt, line 139:
```suggestion
    def write(self, name: str, content: str):
```

test/test_helpers.py
```diff
@@ -74,6 +76,101 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+        }
+
+        path: pathlib.Path = self.path / name
+        with path.open('wt') as f:
+            # Before executing the provided script, dump the provided arguments in calls.txt.
+            # RS 'record separator' (octal 036 in ASCII), FS 'file separator' (octal 034 in ASCII).
+            f.write(
+                '''#!/bin/sh
+    {{ printf {name}; printf "\\036%s" "$@"; printf "\\034"; }} >> {path}/calls.txt
+    {content}'''.format_map(template_args))
```
benhoyt, line 153: Does this content need to be unintended like it was in the original? Or does bash not care? Might be worth unindenting it in any case, to avoid confusion.

benhoyt (approved):
Looks good to me, just a couple of minor styling comments.

test/test_helpers.py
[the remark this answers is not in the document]
tonyandrewmeyer: Assuming that it was in the original too, that seems fine. If this is new code, then I'd rather it was added nicely the first time.

tonyandrewmeyer (changes requested):
A couple of small suggestions (including Ben's ones) but otherwise looking good!

test/test_helpers.py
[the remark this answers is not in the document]
IronCore864: I just confirmed that it was from the original `setUp` method, and I'll leave it like this for now, will refactor later.

tonyandrewmeyer (approved):
LGTM!

Issue #1112, opened 2024-01-23
Combine Pebble layers in testing get_plan; merge "checks" and "log-targets"


benhoyt:
As I discovered when reviewing https://github.com/canonical/operator/pull/1111 ([comment](https://github.com/canonical/operator/pull/1111#discussion_r1463868110)), we don't combine layers properly in the `_TestingPebbleClient.get_plan` result.

We already kind of have the code to combine a layer ([in `_TestingPebbleClient.add_layer`](https://github.com/canonical/operator/blob/dc47640f8ea915049d9e20a1fe4ea189c96ae8d1/ops/testing.py#L2912-L2929)), we just need to use it for `get_plan` as well.

Additionally, we need to combine/merge health checks and log targets. We have `pebble.Service._merge` (used by `add_layer` above), but we'll need similar `_merge` methods on `pebble.Check` and `pebble.LogTarget` and update appropriately.

amandahla:
Hi, I've come across this issue while working on this PR:
https://github.com/canonical/synapse-operator/pull/404

Is there another way of creating a unit test for check changes?

PR #1194, opened 2024-04-19
fix: include log-targets when adding a new layer with harness

skatsaounis:
When unit testing charms with Harness, if they are adding a new pebble layer with `log-targets`, these targets are ignored. This PR is fixing that faulty behavior.

The logic is extracted from the pebble code base:

- https://github.com/canonical/pebble/blob/master/internals/plan/plan.go#L620-L642
- https://github.com/canonical/pebble/blob/master/internals/plan/plan.go#L526-L540

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1117,6 +1117,23 @@ class LogTarget:
         dct = {name: value for name, value in fields if value}
         return typing.cast('LogTargetDict', dct)
 
+    def _merge(self, other: 'LogTarget'):
+        """Merges this log target object with another log target definition.
+
+        For attributes present in both objects, the passed in log target
+        attributes take precedence.
+        """
+        if other.type != '':
+            self.type = other.type
+        if other.location != '':
+            self.location = other.location
+        self.services.extend(other.services)
+        if other.labels is not None:
+            if self.labels is None:
+                self.labels = {}
+            for name, value in other.labels.items():
+                self.labels[name] = value
+
     def __repr__(self):
         return f'LogTarget({self.to_dict()!r})'
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1062,6 +1062,74 @@ class Check:
         dct = {name: value for name, value in fields if value}
         return typing.cast('CheckDict', dct)
 
+    def _merge(self, other: 'Check'):
+        """Merges this check object with another check definition.
+
+        For attributes present in both objects, the passed in check
+        attributes take precedence.
+        """
+        if other.level != '':
+            self.level = other.level
+        if other.period != '':
+            self.period = other.period
+        if other.timeout != '':
+            self.timeout = other.timeout
+        if other.threshold is not None:
+            self.threshold = other.threshold
+        if other.http is not None:
+            if other_url := other.http.get('url'):
+                if self.http is None:
+                    self.http = {}
+                self.http['url'] = other_url
+            if other_headers := other.http.get('headers'):
+                if self.http is None:
+                    self.http = {'headers': {}}
+                for header, value in other_headers.items():
+                    self.http['headers'][header] = value
+        if other.tcp is not None:
+            if other_port := other.tcp.get('port'):
+                if self.tcp is None:
+                    self.tcp = {}
+                self.tcp['port'] = other_port
+            if other_host := other.tcp.get('host'):
+                if self.tcp is None:
+                    self.tcp = {}
+                self.tcp['host'] = other_host
+        if other.exec is not None:
+            if other_command := other.exec.get('command'):
+                if self.exec is None:
+                    self.exec = {}
+                self.exec['command'] = other_command
+            if other_service_context := other.exec.get('service-context'):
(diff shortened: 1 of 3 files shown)


benhoyt:
Thanks for the contribution. This is (kind of) mentioned in https://github.com/canonical/operator/issues/1112 (note that "checks" are missing as well as "log-targets"). A couple of things if you'd like to work on this PR further:

* We should fix "checks" while we're in here
* We should address the `get_plan` issues described in #1112 at the same time
* We need tests for the missing behaviour (which will fail before the fix, succeed after)

Are you willing to work further on this PR to look at it a bit more holistically?

skatsaounis:
Hi @benhoyt 

I think that my two latest commits are taking care of bullets one and three. Thank you for pointing me to the existing issue. I would like to give it a try for the bullet in the middle as well. Should I do it in the same PR or with a follow up? In addition, I think that the static check which is complaining is false positive. However, I am looking forward to a review from a far more experienced pair of eyes to confirm or contradict my suspicion :slightly_smiling_face:

dimaqq:
ci errors:


```py
  /home/runner/work/operator/operator/ops/pebble.py:1088:21 - error: Could not access item in TypedDict
    "headers" is not a required key in "HttpDict", so access may result in runtime exception (reportTypedDictNotRequiredAccess)
  /home/runner/work/operator/operator/ops/pebble.py:1111:21 - error: Could not access item in TypedDict
    "environment" is not a required key in "ExecDict", so access may result in runtime exception (reportTypedDictNotRequiredAccess)
```

skatsaounis:
@dimaqq those two failures are exactly what I am thinking that it is a false positive.

Being more specific, this is the place where one of them is found:

```py
if self.http is None:
    self.http = {'headers': {}}
for header, value in other_headers.items():
    self.http['headers'][header] = value
```

Unless I am not mistaken, this `self.http['headers']` will never fail. I wanted to add `ignore` to the line but I preferred to ask first for a review and, why not, an alternative suggestion.

ops/pebble.py
```diff
@@ -1062,6 +1062,74 @@ def to_dict(self) -> 'CheckDict':
...
+        if other.http is not None:
+            if other_url := other.http.get('url'):
+                if self.http is None:
+                    self.http = {}
+                self.http['url'] = other_url
+            if other_headers := other.http.get('headers'):
+                if self.http is None:
+                    self.http = {'headers': {}}
+                for header, value in other_headers.items():
+                    self.http['headers'][header] = value
```
dimaqq, line 1088: There's got to be a more elegant way to write this...

```suggestion
                self.http = {**(self.http or {}).get('headers', {}), **other_headers}
```

Another option, a little less magical:

```suggestion
                if self.http is None:
                    self.http = {'headers': {}}
                self.http["headers"].update(other_headers)
```

As a team, we should actually think about HTTP request header field names, because current API hides the fact that they are case-insensitive by RFC.

Meanwhile a naive merge of `{Host:...} + {host:...}` will end up with two copies differing by case.

@tonyandrewmeyer wdyt?

ops/pebble.py
```diff
@@ -1062,6 +1062,74 @@ def to_dict(self) -> 'CheckDict':
...
+                    self.exec = {}
+                self.exec['group-id'] = other_group_id
+            if other_group := other.exec.get('group'):
+                if self.exec is None:
+                    self.exec = {}
+                self.exec['group'] = other_group
+            if other_working_dir := other.exec.get('working-dir'):
+                if self.exec is None:
+                    self.exec = {}
+                self.exec['working-dir'] = other_working_dir
```
dimaqq, line 1131: There's a bit too much manual code for my liking.

Perhaps we could use `getattr/setattr` simplify this over a set of member field names.

dimaqq (commented):
Partial review at this point, calling out high level / code organisation level considerations.

ops/pebble.py
benhoyt: I agree there's got to be a more elegant / lower-code way to write this. For example, if you look at the existing [`Service._merge`](https://github.com/canonical/operator/blob/d8148ab41869f6119683cf42b7e6e574f1a99bf1/ops/pebble.py#L992), it uses setattr/getattr, but special cases lists and dicts. This makes it significantly less code than [the Pebble version](https://github.com/canonical/pebble/blob/b45911ac73ca1991cdd1a1085e037a5ef9ed3e0c/internals/plan/plan.go#L122).

I think we can take the same approach for `LogTarget._merge`, and perhaps `Check._merge`, though the latter has the complication of the nested dicts that we'll need to figure out. Still, it seems to me we can simplify this.

Regarding @dimaqq's point about case sensitivity, I think that's fine as is. Pebble doesn't try to do anything clever here, so we shouldn't either.

dimaqq: One more concern: is there a way to reset some field?

If a layer specifies `http.header.Authorization`, can a subsequent layer ever reset auth?

benhoyt:
This is superseded by https://github.com/canonical/operator/pull/1268

PR #1195, opened 2024-04-22
test: refactor test_framework from unittest to pytest style

IronCore864:
Draft PR:

Refactor `test_framework.py` to pytest style.

Major changes:

- remove unittest.TestCase from the test classes
- parametrize
- logging related assertions

Before/After comparison: TODO.


test/test_framework.py
```diff
@@ -17,39 +17,75 @@
...
-    def setUp(self):
-        self.tmpdir = Path(tempfile.mkdtemp())
-        self.addCleanup(shutil.rmtree, str(self.tmpdir))
-
-        patcher = patch('ops.storage.SQLiteStorage.DB_LOCK_TIMEOUT', datetime.timedelta(0))
-        patcher.start()
-        self.addCleanup(patcher.stop)
-
-    def test_deprecated_init(self):
+def create_model():
```
IronCore864, line 40: This (and the `create_framework`) function is moved from `BaseTestCase` defined in `test_helpers.py`, which inherits from `unittest.TestCase`.

test/test_framework.py
```diff
@@ -17,39 +17,75 @@
...
+        charm_dir,
+        meta=model._cache._meta if model else ops.CharmMeta(),
+        model=model)  # type: ignore
+    request.addfinalizer(framework.close)
+    request.addfinalizer(patcher.stop)
+    return framework
+
+
+@pytest.fixture
+def fake_script(request: pytest.FixtureRequest) -> FakeScript:
```
IronCore864, line 78: This is temporarily kept here, not in `test_helpers.py`, because there is a function there with the same name. Will move it after all test files are refactored.

test/test_framework.py
```diff
@@ -17,39 +17,75 @@
...
+    return FakeScript(request)
+
+
+class TestFramework:
+    def test_deprecated_init(self, caplog: pytest.LogCaptureFixture):
```
IronCore864, line 83: Test cases are not brought to the top level, but rather, kept in classes for better organizing test cases since there are multiple test cases defined in one file. Maybe we can refactor this file to smaller ones later.

test/test_framework.py
```diff
@@ -1220,91 +1272,99 @@ class CustomObject:
...
+            pytest.fail('AttributeError not raised')
 
         framework.commit()
 
-    def test_mutable_types(self):
+    def test_mutable_types(self, request: pytest.FixtureRequest):
         # Test and validation functions in a list of tuples.
         # Assignment and keywords like del are not supported in lambdas
         #  so functions are used instead.
+        def _assert_equal(a: typing.Any, b: typing.Any):
```
IronCore864, line 1283: I couldn't move this big test into parametrize because you can't define functions in parametrize. I kept it as it is, but changed the unittest `self.assert` to pytest `assert` style using Ben's suggestion from the previous preview PR.

test/test_framework.py
```diff
@@ -1663,31 +1731,39 @@ def test_pdb_properly_called(self, fake_stderr: io.StringIO):
...
         with patch.dict(os.environ, {'JUJU_DEBUG_AT': 'all'}):
-            framework = self.create_framework()
+            framework = create_framework(request)
         with patch('pdb.Pdb.set_trace'):
             framework.breakpoint()
             assert fake_stderr.getvalue() == _BREAKPOINT_WELCOME_MESSAGE
             framework.breakpoint()
             assert fake_stderr.getvalue() == _BREAKPOINT_WELCOME_MESSAGE
 
-    def test_breakpoint_builtin_sanity(self, fake_stderr: io.StringIO):
```
IronCore864, line 1685: For many test cases with fake_stderr, it seems it's not necessary, so I removed them whereever they are not needed.

test/test_framework.py
```diff
@@ -17,39 +17,75 @@
...
+class TestFramework:
+    def test_deprecated_init(self, caplog: pytest.LogCaptureFixture):
         # For 0.7, this still works, but it is deprecated.
-        with self.assertLogs(level="WARNING") as cm:
+        with caplog.at_level(logging.WARNING):
             framework = ops.Framework(':memory:', None, None, None)  # type: ignore
-        assert "WARNING:ops.framework:deprecated: Framework now takes a Storage not a path" in \
-            cm.output
+        assert "WARNING:ops.framework:deprecated: Framework now takes a Storage not a path" in [
+            f"{record.levelname}:{record.name}:{record.message}" for record in caplog.records]
```
IronCore864, line 88: `caplog.records` only have the message, no level or module, so I made this change.

There is another way, which is to specify module and log level and only assert the message, not asserting the whole `RNING:ops.framework:deprecated:...`.

Same below.

test/test_helpers.py
```diff
@@ -74,6 +76,101 @@ def fake_script_calls(test_case: unittest.TestCase,
     return calls  # type: ignore
 
 
+def create_framework(
```
IronCore864, line 79: This is ported from the other branch, only used in `test_charm.py`. In `test_framework.py`, it uses a different function to create a framework because a model also needs to be passed in.

test/test_helpers.py
```diff
@@ -74,6 +76,101 @@ def fake_script_calls(test_case: unittest.TestCase,
...
+        shutil.rmtree(tmpdir)
+        ops.CharmBase.on = ops.CharmEvents()  # type: ignore
+        framework.close()
+
+    request.addfinalizer(finalizer)
+
+    return framework
+
+
+class FakeScript:
```
IronCore864, line 119: Also copied from the other branch which isn't merged yet.

IronCore864 (commented):
I've left some comments to help review this PR.

test/test_framework.py
benhoyt: Ah, I see, thanks.

test/test_framework.py
```diff
@@ -17,39 +17,75 @@
...
+    backend = _ModelBackend(unit_name='myapp/0')
+    meta = ops.CharmMeta()
+    model = ops.Model(meta, backend)
+    return model
+
+
+def create_framework(request: pytest.FixtureRequest,
+                     *,
+                     model: typing.Optional[ops.Model] = None,
+                     tmpdir: typing.Optional[pathlib.Path] = None):
```
benhoyt, line 51: This is a different signature than the `create_framework` in `test_helpers.py`, right? Is that intentional, and can we not unify them?

test/test_framework.py
```diff
@@ -109,13 +145,16 @@ def restore(self, snapshot: typing.Dict[str, int]):
         handle = ops.Handle(None, "a_foo", "some_key")
         event = Foo(handle, 1)
 
-        framework1 = self.create_framework(tmpdir=self.tmpdir)
+        tmpdir = Path(tempfile.mkdtemp())
+        request.addfinalizer(lambda: shutil.rmtree(tmpdir))
```
benhoyt, line 149: Can we do this inside `create_framework` itself?

test/test_framework.py
```diff
@@ -1397,10 +1456,13 @@ def save_snapshot(self, value: typing.Union[ops.StoredStateData, ops.EventBase])
                     self.snapshots.append((type(value), value.snapshot()))
                 return super().save_snapshot(value)
 
+        tmpdir = Path(tempfile.mkdtemp())
```
benhoyt, line 1459: Can we not use the pytest `tmpdir` fixture for most of these?

test/test_framework.py
```diff
@@ -1628,33 +1690,39 @@ def callback_method(self, event: ops.EventBase):
         self.called = True
 
 
-@patch('sys.stderr', new_callable=io.StringIO)
-class BreakpointTests(BaseTestCase):
+@pytest.fixture
+def fake_stderr() -> io.StringIO:
+    return io.StringIO()
```
benhoyt, line 1695: I'm not sure I see the point of this fixture. Can't we just call `io.StringIO` directly if we need to?

benhoyt (commented):
Looking good. Just a few initial comments.

test/test_framework.py
IronCore864: I think not, because many test cases create multiple frameworks sharing the same tmpdir. If the teardown is in the `create_framework` function, other frameworks might have issues accessing a non-existing directory. I have tested this already, so it has to be put outside. This was the same logic as `BaseTestCase.create_framework` where this function comes from, and I think it was designed in this way precisely because of this reason.

test/test_framework.py
IronCore864: This isn't a fixture, it's creating a tmp dir manually using tempfile. The name is confusing because [pytest does have a fixture named `tmpdir` (although it's preferred to use `tmp_path`).](https://docs.pytest.org/en/7.1.x/reference/reference.html#tmpdir)

test/test_framework.py
IronCore864: Yes, it is redundant. I will delete it.

test/test_framework.py
tonyandrewmeyer: I think it's fine to leave like it is for now, and at some point we should refactor this specifically so that it's clearer and doesn't have all the lambdas.

tonyandrewmeyer: I think it's fine to leave like it is for now, and at some point we should refactor this specifically so that it's clearer and doesn't have all the lambdas.

tonyandrewmeyer (commented):
Looking good - the description still says this is a draft, is that still the case?

test/test_framework.py
```diff
@@ -17,39 +17,77 @@
...
-    def test_deprecated_init(self):
+def create_model():
+    """Create a Model object."""
+    backend = _ModelBackend(unit_name='myapp/0')
+    meta = ops.CharmMeta()
+    model = ops.Model(meta, backend)
+    return model
+
+
+def create_framework(
```
benhoyt, line 48: Yeah, once all the individual files are merged, let's see if we can merge the two different `create_framework`s, even if we have to add a `model` argument to the main one. But we can leave this here for now.

test/test_framework.py
```diff
@@ -275,8 +316,11 @@ def _on_qux(self, event: ops.EventBase, extra: typing.Optional[typing.Any] = Non
             framework.observe(pub.baz, obs._on_baz)  # type: ignore
         framework.observe(pub.qux, obs._on_qux)
 
-    def test_on_pre_commit_emitted(self):
-        framework = self.create_framework(tmpdir=self.tmpdir)
+    def test_on_pre_commit_emitted(self, request: pytest.FixtureRequest):
+        tmpdir = Path(tempfile.mkdtemp())
+        request.addfinalizer(lambda: shutil.rmtree(tmpdir))
```
benhoyt, line 321: Sorry for being dense, but I still don't understand why we can't use a fixture to create (and remove) this directory? At least for the simple cases like this (and others below).

test/test_framework.py
```diff
@@ -1628,33 +1692,35 @@ def callback_method(self, event: ops.EventBase):
...
 
-@patch('sys.stderr', new_callable=io.StringIO)
-class BreakpointTests(BaseTestCase):
+class TestBreakpoint:
 
-    def test_ignored(self, fake_stderr: io.StringIO):
+    def test_ignored(
+        self,
+        request: pytest.FixtureRequest,
+        caplog: pytest.LogCaptureFixture,
```
benhoyt, line 1700: In the other PR too: Let's standardise on using a trailing comma like this, to avoid multiple lines appearing in the diff if we add arguments.

test/test_framework.py
```diff
@@ -1628,33 +1692,35 @@ def callback_method(self, event: ops.EventBase):
...
-                    framework.breakpoint()
-            except AssertionError:
-                pass
-            else:
-                self.fail("No warning logs should be generated")
+
+            framework.breakpoint()
+            warning_logs = [
+                record for record in caplog.records if record.levelno == logging.WARNING]
+            assert len(warning_logs) == 0, "No warning level logs were generated."
```
benhoyt, line 1715: This message seems inverted (it's printed if the assertion fails). However, I think we can probably just drop it, as it doesn't really add any value over what the assertion itself says.

test/test_framework.py
```diff
@@ -1663,31 +1729,43 @@ def test_pdb_properly_called(self, fake_stderr: io.StringIO):
         assert mock.call_count == 1
         assert mock.call_args == ((this_frame,), {})
 
-    def test_welcome_message(self, fake_stderr: io.StringIO):
+    def test_welcome_message(
+        self,
+        request: pytest.FixtureRequest,
+        monkeypatch: pytest.MonkeyPatch
```
benhoyt, line 1735: Per comment above, let's use a trailing comma in places like this to avoid multiline diffs in future:
```suggestion
        monkeypatch: pytest.MonkeyPatch,
```

benhoyt (approved):
Looks good, just a couple of minor style comments.

benhoyt (approved):
Nice, thanks!

tonyandrewmeyer:
This still says "Draft PR" in the description - if it's not draft any more, can that get removed? (Also, it's probably easiest to just use GitHub's draft functionality for drafts).

tonyandrewmeyer (approved):
Looking good - thanks for all the work on these!

PR #1197, opened 2024-04-24
feat: add pebble.CheckInfo.change_id field

benhoyt:
This is new with health checks being implemented as changes and tasks (https://github.com/canonical/pebble/pull/409).

Ops/charms will likely not use this new field. Even once we have the new `change-updated` event, that has a `.get_change()` method to get the change directly. But we're adding it to the Pebble client/types for completeness.

We shouldn't merge this before that Pebble PR is merged, just in case anything changes.

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -208,7 +208,8 @@ if TYPE_CHECKING:
                                 "level": NotRequired[Optional[Union['CheckLevel', str]]],
                                 "status": Union['CheckStatus', str],
                                 "failures": NotRequired[int],
-                                "threshold": int})
+                                "threshold": int,
+                                "change-id": NotRequired[str]})
     _FileInfoDict = TypedDict('_FileInfoDict',
                               {"path": str,
                                "name": str,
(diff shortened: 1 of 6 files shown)


ops/pebble.py
```diff
@@ -1299,15 +1309,17 @@ def from_dict(cls, d: '_CheckInfoDict') -> 'CheckInfo':
...
             failures=d.get('failures', 0),
             threshold=d['threshold'],
+            change_id=ChangeID(d['change-id']) if 'change-id' in d else None,
         )
 
     def __repr__(self):
         return ('CheckInfo('
                 f'name={self.name!r}, '
-                f'level={self.level!r}, '
+                f'level={self.level}, '
```
tonyandrewmeyer, line 1318: This seems like it's better if level is a `CheckLevel` and worse if it's a `str`. Maybe both should be explicitly handled?

tonyandrewmeyer (approved):
Looks good - just one small question (about the drive-by).

I'm not sure I totally understand the use-case for knowing the change ID in ops - is there one? Or is this more for the Python Pebble API and completeness?

ops/pebble.py
benhoyt: True. However, realistically this is always going to be a `CheckLevel` (UP or DOWN). I think I'll leave this as is for now, as explicitly handling both is messy. We can update them all if we want in a later PR. But right now we don't test `repr()` and don't guarantee that `eval(repr(x))` works, so I think it's probably fine as is.

IronCore864 (approved):

Issue #992, opened 2023-08-16
Simplify adding relation units and data in test harness


benhoyt:
When using `testing.Harness`, it's *very* common to see these two or three calls in a row: add relation, (often) add relation unit, and (often) update relation data. It [came up recently](https://github.com/tim-hm/ratings/blob/38e379881e998c97932ee4260e70ec0a7deedf1c/operator/tests/unit/test_charm.py#L34C1-L36) in a simple charm @jnsgruk was writing, and there are many more examples. Here are just a few from a quick grep of the charms I have checked out locally:

```python
# alertmanager-k8s-operator/tests/unit/test_charm.py
self.relation_id = self.harness.add_relation("alerting", "otherapp")
self.harness.add_relation_unit(self.relation_id, "otherapp/0")

# grafana-k8s-operator/tests/unit/test_auth_proxy_provider.py (slightly modified)
relation_id = self.harness.add_relation("grafana-auth", "requirer")
urls = {"urls": json.dumps(EXAMPLE_URLS)}
self.harness.update_relation_data(relation_id, "requirer", urls)

# prometheus-k8s-operator/tests/unit/test_endpoint_aggregator.py
alert_rules_rel_id = self.harness.add_relation(ALERT_RULES_RELATION, "rules-app")
self.harness.add_relation_unit(alert_rules_rel_id, "rules-app/0")
self.harness.update_relation_data(
    alert_rules_rel_id, "rules-app/0", {"groups": ALERT_RULE_1}
)
```

Even our API reference docs have examples with this pattern in it, for example the [`begin_with_initial_hooks` docstring](https://ops.readthedocs.io/en/latest/#ops.testing.Harness.begin_with_initial_hooks).

This is a bit annoying and certainly verbose for authors of charm unit tests.

I propose we add a simplified API that rolls these two or three calls into one. This is somewhat similar to how in Python 3.5 they added [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) to simplify a number of common patterns and reduce verbosity.

There are various ways we could approach this, but I think the simplest and most discoverable would be to augment `add_relation` with two new optional keyword arguments: `unit` and `data`. If you specified a `unit`, it would call `add_relation_unit` after adding the relation, and if you specified `data` it would also call `update_relation_data`. The new signature of `add_relation` and the (trivial) implementation would be:

```python
def add_relation(self, relation_name: str, remote_app: str,
                 unit: Optional[str] = None,
[...]

PR #1202, opened 2024-04-25
add_relation_unit(... data=Mapping[str,str])

PietroPasotti:
Now that we have `add_relation` and you can pass app and unit data straight away, why don't we also extend `add_relation_unit` with a similar API?

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -472,6 +472,10 @@ class Application:
             owner='application')
         return Secret(self._backend, id=id, label=label, content=content)
 
+    def __truediv__(self, other: int) -> "Unit":
+        """Syntactic sugar for generating unit pointers from Application objects."""
+        return self._cache.get(Unit, f"{self.name}/{other}")
+
 
 def _calculate_expiry(expire: Optional[Union[datetime.datetime, datetime.timedelta]],
                       ) -> Optional[datetime.datetime]:
--- a/ops/model.py
+++ b/ops/model.py
@@ -472,10 +472,6 @@ class Application:
             owner='application')
         return Secret(self._backend, id=id, label=label, content=content)
 
-    def __truediv__(self, other: int) -> "Unit":
-        """Syntactic sugar for generating unit pointers from Application objects."""
-        return self._cache.get(Unit, f"{self.name}/{other}")
-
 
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
> Now that we have `add_relation` and you can pass app and unit data straight away, why don't we also extend `add_relation_unit` with a similar API?

I'm not 100% convinced that adjusting `add_relation` was the right thing to do (for example, it makes `add_relation` potentially emit more than once and that has charm-object-reuse issues). However, the cat is well out of the bag with that, so this makes sense to me.

I looked through the [issue](https://github.com/canonical/operator/issues/992) and PR that added this to `add_relation`, and I couldn't find any discussion of whether this should be done for `add_relation_unit` as well for consistency, so I'm guessing it wasn't considered and decided against.

So adding this is a +1 from me, what do you think @canonical/charm-tech ?

benhoyt (changes requested):
Yep, I agree with your reasoning @tonyandrewmeyer -- this is a reasonable addition for consistency with the changes to `add_relation`, and I think it was just missed at the time.

Couple of suggestions for docs and tests.

tonyandrewmeyer:
> So adding this is a +1 from me, what do you think @canonical/charm-tech ?

Out of curiosity, @PietroPasotti, what was your motivation for this? I would assume that you're not generally writing many Harness tests - is `add_relation_unit` followed by `update_relation_data` something you're seeing regularly in some set of charms?

PietroPasotti:
> Out of curiosity, @PietroPasotti, what was your motivation for this? I would assume that you're not generally writing many Harness tests - is `add_relation_unit` followed by `update_relation_data` something you're seeing regularly in some set of charms?

I was adding a harness test to an old suite and it suddenly felt odd not to have it :)

benhoyt:
Tony and I discussed this yesterday, and we're both not opposed but a bit ambivalent on the feature itself. However, we discussed that we'd rather not squeeze new features into Harness at this point (even if small features), but rather start pointing people to Scenario, the longer-term focus. I know it's a tad early as the new Scenario 7.0 "final" API hasn't shipped, but I don't think we should be adding new little helpers to Harness at this point.

PR #1213, opened 2024-05-14
docs: add release step to update pinned charm tests

benhoyt:
We will switch to using pinned versions of the charm (latest `main`) for our charm tests in the GitHub Actions. This PR adds a release step to update the pinned versions, and what to do if they fail.


dimaqq (approved):

HACKING.md
```diff
@@ -288,13 +288,16 @@ To make a release of the ops library, do the following:
    link to the full changelog.
 8. Change [version.py][ops/version.py]'s `version` to the
    [appropriate string](https://semver.org/).
-9. Add, commit, and push, and open a PR to get the changelog and version bump
+9. Update the pinned charm versions in `.github/workflows/*-charm-tests.yaml` to
```
tonyandrewmeyer, line 291: I like when these are links, although it's trickier here since it's a wildcard. Maybe to the directory?

```suggestion
9. Update the pinned charm versions in [*-charm-tests.yaml](.github/workflows/) to
```

HACKING.md
```diff
@@ -288,13 +288,16 @@ To make a release of the ops library, do the following:
...
+   the latest version of the charm's `main`. If they fail in the PR, bump them back
```
tonyandrewmeyer, line 292:
```suggestion
   the latest version of the charm's default branch. If they fail in the PR, bump them back
```

Some of them are still using `master`.

tonyandrewmeyer (approved):
Looks good, two very small suggestions.

dimaqq:
I've updated the wording for the "check/update charm pins" step in the release procedure.

I've decided not to include the rollback step, as I figure that newer charm version failing against pre-release ops code should probably block the release, require figuring out what happened and sorting things out manually in general.

PR #1217, opened 2024-05-15
ci: automate 3rd party GitHub Actions version updates

### LINKED ISSUE #1216: ci: update actions wrt. node
--
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.Show less
 

<br class="Apple-interchange-newline">

Most likely we just need to update the 3rd party actions that we pull into the workflow, like install-python from vN to vLatest

dimaqq:
fixes #1216

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


dimaqq:
@benhoyt I don't have access to this repo settings, here's how to enable dependabot: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates

tonyandrewmeyer (approved):
Works for me, although maybe we want to extend this and switch dependabot from security only to all deps more generally?

benhoyt (approved):
Fine with me. I'd also be happy to add for other deps, as long as it's not noisy (for docs deps, for example, it just doesn't add much value): at *most* once a week.

PR #1221, opened 2024-05-16
docs: use the actual emoji character rather than GitHub markup, to show properly on PyPI

tonyandrewmeyer:
Currently, [ops on PyPI](https://pypi.org/project/ops/) looks like this:

image

This PR replaces the `:point-right:` GitHub-specific markup with the actual emoji character so that it shows as expected in both places.


benhoyt (approved):

IronCore864 (approved):

dimaqq (approved):

dimaqq:
OT: I wonder if the intention of the table was to eventually add logos for Juju and Charmhub...

PR #1224, opened 2024-05-22
refactor: use ruff formatter

IronCore864:
Use ruff formatter in tox, and format the code base.

Commits:

- 1st commit: changing `tox.ini` and `pyproject.toml` only
- 2nd: automated changes done by ruff
- 3nd: fix linting and some manual tweaks
- all the remaining commits: manual fixes, merges, etc.

## 1 Major changes

To help you review this PR quicker, I summarized a list of different types of changes done by ruff after self-reviewing:

### 1.1 Trailing-Comma

Added to most multi-line list and stuff.

### 1.2 Quotes

Single quote is configured, but there are exceptions, like docstrings, multi-line strings and re.compile, which all use `"""`'.

There are exceptions where both single and double quotes are used but after a closer look it's reasonable, example:

```python
some_multi_lines_str = (
    "Controller storage is deprecated; it's intended for "
    'podspec charms and will be removed in a future release.'
)
```

The first line uses double quotes because there is a single quote inside, while the second line doesn't. A bit inconsistency considering they are next to each other but reasonable.

### 1.3 Close Parenthesis

Changed into a new single line for most multi-line cases and improved readability, example:

```python
relation = Relation(relation_name, rid, is_peer,
                    self._our_unit, self._backend, self._cache)
```

Becomes:

```python
relation = Relation(
    relation_name, rid, is_peer, self._our_unit, self._backend, self._cache
)
```

### 1.4 Parametrize

`@pytest.mark.parametrize(format, data)` now becomes:

```python
@pytest.mark.parametrize(
	format,
	data,
)
```

This is more readable when the data is a big list.

See examples from `test/test_log.py`.

### 1.5 Arg List

If parameters are not long enough, it becomes a single line in a new line:

```python
def do_something(
	foo, bar, baz
)

If the parameters are too long for a single line, it becomes multi-line:

```python
def do_something(
	foo,
	bar,
	baz,
	qux,
)
```

A little inconsistency, but doesn't hurt readability, and it seems _ruff can't be configured to use the second way consistently_.

See examples from `ops/charm.py`.

### 1.6 Dicts

Same story as above, key-value pairs can be on a single line or multiple lines depending on the length. A little inconsistent, but doesn't hurt readability. See examples from `ops_charm.py`.

### 1.7 List-Comprehension

For long ones that span over two lines, they are reformatted in most cases, generally with better readability. In some cases it's shortened into a single new line, example:

```python
self._data.update({
    unit: RelationDataContent(self.relation, unit, backend)
    for unit in self.relation.units})
```

changed to:

```python
self._data.update({
    unit: RelationDataContent(self.relation, unit, backend) for unit in self.relation.units
})
```

---

## 2 Manual Changes

Focus is put on `ops/*.py`, didn't spend much time on `test/*.py` except fixing a failed UT: Apparently changing from double quotes to single quotes failed one UT which asserts on the error, see `test_excepthook` in `test/test_main.py`. Did not self-review `test/test_testing.py` since there is still an open PR on that.

Other manual changes:

- `.format()` to F-String
- String Concatenation to F-String
- %-formatting to F-String
- Re-arranging Lines

### DIFF
--- a/ops/__init__.py
+++ b/ops/__init__.py
@@ -45,7 +45,6 @@ __all__ = [  # noqa: RUF022 `__all__` is not sorted
     '__version__',
     'main',
     'pebble',
-
     # From charm.py
     'ActionEvent',
     'ActionMeta',
@@ -95,7 +94,6 @@ __all__ = [  # noqa: RUF022 `__all__` is not sorted
     'UpdateStatusEvent',
     'UpgradeCharmEvent',
     'WorkloadEvent',
-
     # From framework.py
     'BoundEvent',
     'BoundStoredState',
@@ -118,10 +116,8 @@ __all__ = [  # noqa: RUF022 `__all__` is not sorted
     'StoredSet',
     'StoredState',
     'StoredStateData',
-
     # From jujuversion.py
     'JujuVersion',
-
     # From model.py
     'ActiveStatus',
     'Application',
--- a/ops/lib/__init__.py
+++ b/ops/lib/__init__.py
@@ -142,7 +142,7 @@ def _find_all_specs(path: typing.Iterable[str]) -> typing.Iterator[ModuleSpec]:
             except (FileNotFoundError, NotADirectoryError):
                 continue
             except OSError as e:
-                logger.debug("  Tried '%s': %s", opslib, e) # *lots* of things checked here
+                logger.debug("  Tried '%s': %s", opslib, e)  # *lots* of things checked here
                 continue
             else:
                 logger.debug("  Trying '%s'", opslib)
--- a/ops/lib/__init__.py
+++ b/ops/lib/__init__.py
@@ -132,9 +132,9 @@ def _find_all_specs(path: typing.Iterable[str]) -> typing.Iterator[ModuleSpec]:
         except (FileNotFoundError, NotADirectoryError):
             continue
         except OSError as e:
-            logger.debug(f"Tried to look for ops.lib packages under '{sys_dir}': {e}")
+            logger.debug("Tried to look for ops.lib packages under '%s': %s", sys_dir, e)
             continue
-        logger.debug(f"Looking for ops.lib packages under '{sys_dir}'")
+        logger.debug("Looking for ops.lib packages under '%s'", sys_dir)
         for top_dir in top_dirs:
             opslib = os.path.join(sys_dir, top_dir, 'opslib')
             try:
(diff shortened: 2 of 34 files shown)


ops/__init__.py
```diff
@@ -45,7 +45,6 @@
     '__version__',
     'main',
     'pebble',
-
```
tonyandrewmeyer, line 48: Does ruff force the lack of breaks here? If so, then that's ok, but if we can keep them I think we should, it helps show the logical sections.

ops/lib/__init__.py
```diff
@@ -123,46 +125,46 @@ def autoimport():
...
-            sys_dir = "."
+        if sys_dir == '':
+            sys_dir = '.'
         try:
             top_dirs = os.listdir(sys_dir)
         except (FileNotFoundError, NotADirectoryError):
             continue
         except OSError as e:
-            logger.debug("Tried to look for ops.lib packages under '%s': %s", sys_dir, e)
+            logger.debug(f"Tried to look for ops.lib packages under '{sys_dir}': {e}")
```
tonyandrewmeyer, line 135: All of these changes to the logging calls need to be reverted. f-strings should not be used in logging calls, because logging tools use the unformatted string to (e.g.) de-dupe messages.

tonyandrewmeyer (changes requested):
Everything looks good to me, except for changing to use f-strings in logging, which should not be done.

ops/__init__.py
IronCore864: Although I think we should keep it since it increases readability, ruff unfortunately doesn't like it, and it seems there isn't a config about empty lines in imports, only a config to change the number of blank lines to place _after_ imports...

IronCore864:
> Everything looks good to me, except for changing to use f-strings in logging, which should not be done.

I saw in existing code base that there are many places that use f-strings for logging, so I did it to keep it consistent. Shall I revert these manual changes?

ops/lib/__init__.py
IronCore864: I saw in existing code base that there are many places that use f-strings for logging, so I did it to keep it consistent. I'll revert them.

ops/__init__.py
tonyandrewmeyer: Ok, let's let ruff do it's thing, then.

ops/lib/__init__.py
tonyandrewmeyer: Indeed :disappointed: 

```
ops/pebble.py
1642:                logger.warning(f'Invalid I/O command {command!r}')
1705:                    logger.warning(f'Invalid I/O command {command!r}')

ops/storage.py
61:            logger.debug(f"Initializing SQLite local storage: {filename}.")

ops/model.py
988:        logger.debug(f"could not cast {raw} to IPv4/v6 address")
```

I'll open a different PR to fix those.

benhoyt (changes requested):
Agreed this looks good overall, and thanks for the detailed PR description. Strongly agree with @tonyandrewmeyer about avoiding logging formatting changes. I also don't think we should include any other f-string changes in this PR. Let's keep it to just Ruff + a few manual line break fix-ups.

tonyandrewmeyer (approved):
LGTM!

benhoyt (approved):

pyproject.toml
```diff
@@ -73,6 +73,11 @@ aggressive = 3
 [tool.ruff]
 line-length = 99
 target-version = "py38"
+exclude = ["docs"]
```
tonyandrewmeyer, line 76: We should still exclude everything that's excluded by default.

tonyandrewmeyer (changes requested):
We should fix the excluding, as found by @benhoyt .

pyproject.toml
benhoyt: From the Ruff docs: "Note that you'll typically want to use [extend-exclude](https://docs.astral.sh/ruff/settings/#extend-exclude) to modify the excluded paths." https://docs.astral.sh/ruff/settings/#extend-exclude

tonyandrewmeyer (approved):

Issue #1225, opened 2024-05-22
`RelationDataContent` should be update-able via keyword arguments only


addyess:
https://github.com/canonical/operator/pull/1183 created a regression where `RelationDataContent` could not be updated via the `.update(...)` method if only keyword arguments were passed to it. 

[this module](https://github.com/charmed-kubernetes/interface-aws-integration/blob/main/ops/ops/interface_aws/requires.py#L188-L191) previously could execute
```python
some_dict = {"foo": "bar"}
self.relation.data[self.charm.model.unit].update(**some_dict)
```

but now there's an exception because i don't have an `other` argument
```
  File "/var/lib/juju/agents/unit-kubernetes-worker-0/charm/venv/ops/interface_aws/requires.py", line 133, in _joined
    self._request({"instance-id": self.instance_id, "region": self.region})
  File "/var/lib/juju/agents/unit-kubernetes-worker-0/charm/venv/ops/interface_aws/requires.py", line 190, in _request
    self._to_publish.update(**kwds)
TypeError: RelationDataContent.update() missing 1 required positional argument: 'other'
```

This is acceptable in normal python and should be restored as shown in this example where both styles are acceptable in an `update`

```python
Python 3.12.3 (main, May 22 2024, 13:41:37) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = {}
>>> a.update(**{"a": 1})
>>> a
{'a': 1}
>>> b = {"b": 2}
>>> a.update(b)
>>> a
{'a': 1, 'b': 2}
>>> c, d = {"c": 3}, {"d": 4}
>>> a.update(c, **d)
>>> a
{'a': 1, 'b': 2, 'c': 3, 'd': 4}
```

addyess:
Addressed by https://github.com/canonical/operator/pull/1226

addyess:
@tonyandrewmeyer thanks for closing.  When might i expect an `ops` library release to provide this feature?

tonyandrewmeyer:
> @tonyandrewmeyer thanks for closing. When might i expect an `ops` library release to provide this feature?

We usually do an ops release at the end of each month. We haven't talked about it yet (will do in our daily today) but I'd expect that we'd do a release in the first half of this week.

dimaqq:
@addyess done, enjoy!

PR #1226, opened 2024-05-22
fix: the `other` argument to `RelatationDataContent.update(...)` should be optional

addyess:
Fixes the signature for `RelationDataContent.update` to match `MutableMapping`, where `other` is optional (a regression introduced in #1883).

The type for `other` has been simplified to `Any`. It should really be `Mapping|_SupportsKeysAndGetItem[str,str]` plus a minimal type that supports `.values`, but it was already messy pulling in `_SupportsKeysAndGetItem` in #1183, and we're just passing this through to `MutableMapping` so it doesn't seem like the tight typing is providing enough benefit to justify the complexity of the signature. [typeshed has three overloads](https://github.com/python/typeshed/blob/f7c03486ee01c8ea74823db75e017341bf3c2ad0/stdlib/typing.pyi#L726), so we could match that (as we did in #1883, just incompletely), if that is desirable.

Fixes: #1225

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1722,8 +1722,10 @@ class RelationDataContent(LazyMapping, MutableMapping[str, str]):
         self._validate_read()
         return super().__getitem__(key)
 
-    def update(self, other: _SupportsKeysAndGetItem[str, str], **kwargs: str):
+    def update(self, other: Optional[_SupportsKeysAndGetItem[str, str]] = None, *_, **kwargs: str):
         """Update the data from dict/iterable other and the kwargs."""
+        if other is None:
+            other = {}
         super().update(other, **kwargs)
 
     def __delitem__(self, key: str):
--- a/ops/model.py
+++ b/ops/model.py
@@ -91,16 +91,6 @@ _NetworkDict = TypedDict('_NetworkDict', {
 })
 
 
-# Copied from typeshed.
-_KT = typing.TypeVar("_KT")
-_VT_co = typing.TypeVar("_VT_co", covariant=True)
-
-
-class _SupportsKeysAndGetItem(typing.Protocol[_KT, _VT_co]):
-    def keys(self) -> typing.Iterable[_KT]: ...
-    def __getitem__(self, __key: _KT) -> _VT_co: ...
-
-
 logger = logging.getLogger(__name__)
 
 MAX_LOG_LINE_LEN = 131071  # Max length of strings to pass to subshell.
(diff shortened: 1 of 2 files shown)


addyess:
@tonyandrewmeyer and @benhoyt  I appreciate your speedy review as this likely break a few releases queued up for this week.

dimaqq:
The change makes sense, I believe Tony volunteered to tests for this.

ops/model.py
```diff
@@ -1722,7 +1712,7 @@ def __getitem__(self, key: str) -> str:
         self._validate_read()
         return super().__getitem__(key)
 
-    def update(self, other: _SupportsKeysAndGetItem[str, str], **kwargs: str):
+    def update(self, other: typing.Any = (), /, **kwargs: str):
```
dimaqq, line 1715: **TIL** that typeshed doesn't even define `dict.update`

https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L1029

So, all good here!

test/test_model.py
```diff
@@ -232,6 +232,23 @@ def test_get_app_relation_data(self, harness: ops.testing.Harness[ops.CharmBase]
...
+        relation_id = harness.add_relation('db1', 'remote')
+        harness.add_relation_unit(relation_id, 'remote/0')
+        with harness._event_context('foo_event'):
+            harness.update_relation_data(
+                relation_id,
+                harness.model.app.name,
+                {'foo': 'bar'})
+            rel = harness.model.get_relation('db1', relation_id)
+            assert rel is not None
+            rel.data[harness.model.app].update(foo='baz')
```
dimaqq, line 247: Let's cover some flavour of `.update(iterable)` while we're at it.

dimaqq (approved):

dimaqq (approved):
Super-minor/bikeshedding: I suppose the test could be more concise (either parametrised of performing all 3 updates on the same object/within same test).

Happy to have this fix!

tonyandrewmeyer:
> Super-minor/bikeshedding: I suppose the test could be more concise (either parametrised of performing all 3 updates on the same object/within same test).

Fair point. I have never really been in the habit of using `unittest.subTest` for this, but now that we are (very nearly) switched over to pytest, using `pytest.mark.parametrize` would be cleaner.

addyess:
Wow, thanks for the improved solution @dimaqq!

dimaqq:
@tonyandrewmeyer deserves the credit. I've only reviewed the PR 🙇🏻

addyess:
> @tonyandrewmeyer deserves the credit. I've only reviewed the PR 🙇🏻

Sorry, it was late and i was blurry-eyed 👀 .  Thanks so much @tonyandrewmeyer!

addyess:
@IronCore864 Thanks so much also for offering a review

benhoyt (changes requested):
Fix looks good, thanks @addyess. I agree that the tests have a bit too much duplication, parametrization seems like a good idea here.

benhoyt (approved):
Much nicer, thanks

PR #1227, opened 2024-05-23
fix: don't use f-strings in logging calls

tonyandrewmeyer:
Removes use of f-strings in `logging.debug` and `logging.warning` calls.

Calls to `logging.*` should provide a format string and arguments (generally we use `%` style, but `{` or `$'` would be ok, if inconsistent). There are two main reasons:

* Some log handlers (most well known is Sentry, but I believe there are others) will aggregate messages using the format string, and this breaks if the string is pre-interpolated.
* Interpolation/formatting is done lazily - for example, a debug level message may never get formatted - so there is a (very small) efficiency gain by avoiding it when possible.

Theoretically, there are also safety concerns in having user-provided content, but that's not relevant for any of these cases here, and I think you'd have to _also_ provide an argument to exploit that.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -985,7 +985,7 @@ def _cast_network_address(raw: str) -> Union[ipaddress.IPv4Address, ipaddress.IP
     try:
         return ipaddress.ip_address(raw)
     except ValueError:
-        logger.debug(f"could not cast {raw} to IPv4/v6 address")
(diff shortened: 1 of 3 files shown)


dimaqq (approved):

benhoyt (approved):
Thanks!

Issue #1229, opened 2024-05-27
Harness not modelling permission denied for user secrets


DnPlas:
#1176 introduced support for user secrets in Harness, but it looks like it is not modelling the actual Juju behaviour when a user secret hasn't been granted to a charm and it tries to get a secret with `self.model.get_secret(id=<id-from-config>)`.

On an actual model, if the charm config gets updated with a secret ID, but the charm hasn't been granted access to the secret, `ops.model.ModelError: ERROR permission denied` will be raised, but it seems that when testing, harness makes the charm to raise a different exception ` ops.model.SecretNotFoundError: Secret 'secret:16320250-1b76-4552-94d9-10ff484f88cb' not granted access to 'istio-pilot` when calling the exact same piece of code.

#### Steps to reproduce

1. Deploy any charm that can use user secrets
2. Create a user secret
3. Pass the secret ID through charm configuration `juju config <charm-name> my-secret=secret:some-id`
4. Observe the charm state and logs. You should get something like this:

```
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/model.py", line 3397, in secret_get
    result = self._run('secret-get', *args, return_output=True, use_json=True)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/model.py", line 3051, in _run
    raise ModelError(e.stderr) from e
ops.model.ModelError: ERROR permission denied
```

5. Add a unit test for testing what happens when the secret hasn't been granted to the charm. For example:

```python
def test_permission_denied(...):
  secret_content = {"tls-crt": "a-tls-crt", "tls-key": "a-tls-key"}
  secret_id = harness.add_user_secret(secret_content)
  harness.update_config({"tls-secret-id": secret_id})
  harness.begin()
  with pytest.raises(ModelError) as err:
    harness.charm.function_call_to_get_secret() # <--- this is a function that calls self.model.get_secret(id=<id from config>)
```

6. The above will not raise because it is expecting a `ModelError` but a `SecretNotFoundError` exception is raised instead:

```
 ops.model.SecretNotFoundError: Secret 'secret:16320250-1b76-4552-94d9-10ff484f88cb' not granted access to 'istio-pilot
```

DnPlas:
Extra information:

[...]

PR #1231, opened 2024-05-28
docs: clarify that SecretNotFound may be raised for permission errors

tonyandrewmeyer:
As noticed in #1229, `get_secret()` may raise `SecretNotFoundError` if the secret does exist but the caller does not have permission to view it, so make this explicit in the documentation.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -274,7 +274,8 @@ class Model:
             label: Secret label if fetching by label (or updating it).
 
         Raises:
-            SecretNotFoundError: If a secret with this ID or label doesn't exist.
+            SecretNotFoundError: If a secret with this ID or label doesn't exist,
+                or the caller does not have permission to view it.
         """
         if not (id or label):
             raise TypeError('Must provide an id or label, or both')


benhoyt (commented):
But isn't the point of #1229 that Ops will currently raise `ModelError` for permission/grant errors?

benhoyt:
Oh, I see -- we're closing #1229, and this PR goes on the assumption that the Juju bug will be fixed.

tonyandrewmeyer:
> Oh, I see -- we're closing #1229, and this PR goes on the assumption that the Juju bug will be fixed.

Yes - my assumption is that since @wallyworld agrees that Juju should be consistent that the Juju bug will get fixed.

But also: this is *already* the case for everything other than user secrets. If we wanted it to reflect the current behaviour, then it should have this change and also a new "raises" item of "ModelError: when the secret is a user secret and the charm does not have permission to view it". My preference would be to not add that now on the assumption that it won't be true soon (for some value of "soon").

benhoyt (approved):
Makes sense to me -- thanks.

PR #1232, opened 2024-05-28
chore: add a pre-commit config

tonyandrewmeyer:
Adds a basic [pre-commit](https://pre-commit.com/) configuration file.

For anyone that doesn't have pre-commit installed, this will have no impact. For those that do, running `pre-commit install` in the repo will install the git pre-commit hooks so that these checks will run on every commit.

This moves basic checks 'left' so that they are caught as early as possible.

The config only enables checks from ruff and ones from the 'built-in' set. We may wish to add others in time, but starting with a simple set seems best.

Note that pre-commit is already being adopted elsewhere in the charming community (at least 8 charms, plus scenario and charmcraft), which I'm broadly in favour of, and I think it would be good for ops to join this group.


dimaqq (approved):
Will there be a separate PR to apply pre-commit fixes?
I've tried in a checkout, and it's not too many:

```command
> git diff --stat
 README.md                           | 3 +--
 docs/.sphinx/_templates/footer.html | 2 +-
 docs/.sphinx/pa11y.json             | 2 +-
 test/charms/test_smoke/README.md    | 2 +-
 tox.ini                             | 2 +-
 5 files changed, 5 insertions(+), 6 deletions(-)
```

plus 5 errors that don't have autofix...

And there's this, which may suggest that our ruff config is wrong?

`warning: Selection `CPY` has no effect because preview is not enabled.`

tonyandrewmeyer:
I didn't think to do a `-a` because it was mostly ruff, which we're already running. I've removed some checks that we're better off not using for now.

```
$ pre-commit run -a
check for added large files..............................................Passed
check python ast.........................................................Passed
check for case conflicts.................................................Passed
check that executables have shebangs.....................................Passed
check that scripts with shebangs are executable..........................Passed
check for merge conflicts................................................Passed
check for broken symlinks................................................Passed
check json...............................................................Passed
check yaml...............................................................Passed
check toml...............................................................Passed
mixed line ending........................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
detect private key.......................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
```

I also fixed the arg passing for ruff. Strangely, we had an `__init__.py` file that was marked as executable but not actually, so I removed the flag. It seems fine for that small change to go into this PR.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
+repos:
```
benhoyt, line 1: I suppose this file has to be in the root directory? The root clutter is annoying -- then again, I don't see a much better place to put it.

benhoyt (approved):
This seems reasonable to me -- as long as we keep it fast, like around 1s, not running unit tests or anything. Plus, it's opt-in for the developer.

.pre-commit-config.yaml
tonyandrewmeyer: As far as I know, yes. Since it's language-agnostic, it doesn't support going into something like pyproject.toml, and indeed there doesn't really seem like anywhere else it could go.

PR #1233, opened 2024-05-28
test: verify that defer() is not usable on stop,remove,secret-expired,secret-rotate

tonyandrewmeyer:
Extends the test for not being able to `defer()` an action event to also cover the other events where this is the case. This was missed in #1122.


benhoyt (approved):
Thanks for this!

test/test_charm.py
```diff
@@ -560,27 +560,42 @@ def _on_foo_bar_action(self, event: ops.ActionEvent):
...
+        ('secret_expired', {'id': 'secret:123', 'label': None, 'revision': 0}),
+        ('secret_rotate', {'id': 'secret:234', 'label': 'my-secret'}),
+    ],
+)
+def test_inappropriate_event_defer_fails(
+    event: str,
+    kwargs: typing.Dict[str, typing.Any],
+    request: pytest.FixtureRequest,
+    monkeypatch: pytest.MonkeyPatch,
+    fake_script: FakeScript,
```
dimaqq, line 578: Minor style nit, somehow I prefer more variable arguments at the end, thus `test_foo(fixtures, parameters)` while it seems that Tony prefers `test_foo(parameters, fixtures)` because then parameter names are closer to the parameter values perhaps?

```suggestion
    request: pytest.FixtureRequest,
    monkeypatch: pytest.MonkeyPatch,
    fake_script: FakeScript,
    event: str,
    kwargs: typing.Dict[str, typing.Any],
```

I don't have a strong opinion either way, it just struck me as something I have not yet seen.

Maybe @benhoyt and @IronCore864 can voice their preference or opinion 🙇🏻

test/test_charm.py
```diff
@@ -560,27 +560,42 @@ def _on_foo_bar_action(self, event: ops.ActionEvent):
...
 ):
-    cmd_type = 'action'
-
     class MyCharm(ops.CharmBase):
         def __init__(self, framework: ops.Framework):
             super().__init__(framework)
-            framework.observe(self.on.start_action, self._on_start_action)
+            framework.observe(getattr(self.on, event), self._call_defer)
```
dimaqq, line 583: I wonder if there's a better way to write this.

test/test_charm.py
```diff
@@ -560,27 +560,42 @@ def _on_foo_bar_action(self, event: ops.ActionEvent):
...
 
-        def _on_start_action(self, event: ops.ActionEvent):
+        def _call_defer(self, event: ops.EventBase):
```
dimaqq, line 585: I'd prefer a different name for this method, because while we use it to test deferral, in real user code, it would have own semantics, so maybe `def _event_handler` or user event handler or generic event handler?

I wonder if it's worth adding a check that this function was in fact ran, that it's specifcally `event.defer()` call that failed and not something else.

Would it be possible to slap `with pytest.raises(...)` into the body of this method instead?

dimaqq (approved):
Look good overall; I've left some ideas, though I don't know if they are tractable.

test/test_charm.py
tonyandrewmeyer: Better in what way?

test/test_charm.py
tonyandrewmeyer: > I'd prefer a different name for this method, because while we use it to test deferral, in real user code, it would have own semantics, so maybe `def _event_handler` or user event handler or generic event handler?

This isn't aiming to look like a real event handler. I think it's better describing what it does than being more generic. I'd use a lambda, avoiding any name, if those could be used as observers.

> I wonder if it's worth adding a check that this function was in fact ran, that it's specifcally `event.defer()` call that failed and not something else.

It seems like that's expanding the scope of this PR somewhat. I guess `RuntimeError` is somewhat generic,s o it's possible that something else in the emitting could cause it, although it's tricky to think of how that might happen without other tests then failing.

My feeling is that this test is checking that calling `defer()` raises in these cases, not that the event handler is properly called, or that calling `defer()` in an event handler works - there are (or should be) other tests for that.

> Would it be possible to slap `with pytest.raises(...)` into the body of this method instead?

That said, I'm not against doing this instead, if people feel like it would be better.

test/test_charm.py
tonyandrewmeyer: I have no real preference here; it just seemed the cleanest at the time. @IronCore864 I assume you must have done a bunch of these in the pytest conversion, but I didn't pay attention to the order when reviewing. I'm happy to go with whatever you did, to keep things most consistent.

test/test_charm.py
dimaqq: Sorry, that was meant as a lead into the next comment. Closing this one.

test/test_charm.py
benhoyt: I believe we've tried to use (fixtures, parameters) in the existing tests, for example https://github.com/canonical/operator/blob/9e5463cde7d942d71cd51af467d832759728480b/test/test_model.py#L2762-L2768

test/test_charm.py
benhoyt: As far as the name goes, I think it can be argued both ways (my slight preference is for the "call_defer" name). So let's (ahem) *defer* to Tony's original name.

I think the event handler is small and well-scoped enough that we know it's the defer that failed. I don't think we need to complicate this here.

benhoyt (approved):

dimaqq (approved):

PR #1234, opened 2024-05-28
chore: remove f-string in logging call

tonyandrewmeyer:
Use lazy interpolation in `logging.*` calls.

Missed in #1227, because I didn't think to enable multi-line searching, and this logging call wraps the f-string on to the next line.

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -951,9 +951,9 @@ class Framework(Object):
 
             else:
                 logger.warning(
-                    f'Reference to ops.Object at path {observer_path} has been garbage collected '
+                    'Reference to ops.Object at path %s has been garbage collected '
                     'between when the charm was initialised and when the event was emitted. '
-                    'Make sure sure you store a reference to the observer.'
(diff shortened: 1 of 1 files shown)


ops/framework.py
```diff
@@ -951,9 +951,9 @@ def _reemit(self, single_event_path: Optional[str] = None):
 
             else:
                 logger.warning(
-                    f'Reference to ops.Object at path {observer_path} has been garbage collected '
+                    'Reference to ops.Object at path %s has been garbage collected '
                     'between when the charm was initialised and when the event was emitted. '
-                    'Make sure sure you store a reference to the observer.'
+                    'Make sure sure you store a reference to the observer.', observer_path
```
benhoyt, line 956: I'm surprised Ruff doesn't put `observer_path` on its own line.

benhoyt (approved):

ops/framework.py
dimaqq: it would if there was a trailing comma.

or if the line was too long and had to be split.

dimaqq (approved):

PR #1238, opened 2024-05-29
release: change version back to .dev after release

dimaqq:
change version to dev flavour of the expected upcoming release, `2.15.0.dev0`

### DIFF
--- a/ops/version.py
+++ b/ops/version.py
@@ -17,4 +17,4 @@
 This module is NOT to be used when developing charms using ops.
 """
 
-version: str = '2.14.0.dev0'
+version: str = '2.15.0.dev0'
--- a/ops/version.py
+++ b/ops/version.py
@@ -17,4 +17,4 @@
 This module is NOT to be used when developing charms using ops.
 """
 
-version: str = '2.14.0'
+version: str = '2.14.0.dev0'


benhoyt (approved):

ops/version.py
```diff
@@ -17,4 +17,4 @@
 This module is NOT to be used when developing charms using ops.
 """
 
-version: str = '2.14.0'
+version: str = '2.14.0.dev0'
```
tonyandrewmeyer, line 20: We're working towards 2.15 now.

```suggestion
version: str = '2.15.0.dev0'
```

tonyandrewmeyer (changes requested):

tonyandrewmeyer:
We should probably standardise and document the conventional commit type for these changes. `release` isn't one we've used before as far as I know, and it's not in the usual list of types (I'd probably go for `chore`). We could adopt it, but we should be consistent about it.

dimaqq:
Thanks, Tony. The dreaded off-by-one error 🙈

tonyandrewmeyer (approved):

dimaqq:
> We should probably standardise and document the conventional commit type for these changes. `release` isn't one we've used before as far as I know, and it's not in the usual list of types (I'd probably go for `chore`). We could adopt it, but we should be consistent about it.

`chore(release): waffle waffle` then?

PR #1239, opened 2024-05-29
ci: tag reviewers explicitly

dimaqq:
Tag reviewers explicitly, since:
- we're not using CODEOWNERS
- tagging a group kinda "dimisses" the reviewer group setting after 1 review, when we need 2 reviews to merge

Note that I'm unable to test the `--reviewers` flag in my fork, probably because none else on the team is a collaborator in my fork. I've invited you all, and can test the change in my fork when someone accepts.

### DIFF
--- a/.github/workflows/update-charm-tests.yaml
+++ b/.github/workflows/update-charm-tests.yaml
@@ -47,7 +47,12 @@ jobs:
             gh pr close -c stale "$PR";
           elif [[ -z "$PR" && -n "$CHANGES" ]]; then
             echo "Opening new PR"
-            gh pr create --base main --head auto-update-external-charm-pins --title "chore: update charm pins" --body "This is an automated PR to update pins of the external repositories that the operator framework is tested against";
+            gh pr create \
+               --base main \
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
Do we want to do this? We're all (I assume) subscribed to the repo so will get a notification when the PR is opened, and we only need two people to review, not 4. I'm not sure what value this adds.

If we're worried that no-one will pick these up then I think we should talk about that as a team, because there are other items (e.g. issues, community PRs) that also need picking up. I'd rather we considered enabling auto-assignment in general (2 reviewers, either of the routing choices), augmenting that with personal picks when appropriate, than special-case these.

dimaqq:
I think it's up for debate.

Given that we run this on schedule, and we only care deeply before a release, perhaps it makes more sense not to tag folks.

benhoyt:
I agree with Tony here: I don't think this adds value, and it means we have a hard-coded list of team members in the repo, which isn't ideal.

dimaqq:
Closing as not desired.

PR #1240, opened 2024-05-30
feat: extend ops.log to capture warnings

IronCore864:
Extend `ops.log` to capture warnings as well.

Warnings are redirected to the logging system with a custom formatter.

Added a test in `test_main.py` for this feature.

### Charm Used for Testing

```python
class SampleK8SCharm(ops.CharmBase):
    def __init__(self, *args):
        super().__init__(*args)
        self.framework.observe(self.on["httpbin"].pebble_ready, self._on_httpbin_pebble_ready)

    def _on_httpbin_pebble_ready(self, event: ops.PebbleReadyEvent):
        container = event.workload
        container.add_layer("httpbin", self._pebble_layer, combine=True)
        container.replan()
        warnings.warn("feature x is deprecated, use feature y instead", DeprecationWarning)
```

### Before

Warnings are in the container log (`kubectl logs`):

```bash
2024-06-13T00:20:42.850Z [container-agent] 2024-06-13 00:20:42 WARNING httpbin-pebble-ready /var/lib/juju/agents/unit-sample-k8s-0/charm/./src/charm.py:40: DeprecationWarning: feature x is deprecated, use feature y instead
2024-06-13T00:20:42.850Z [container-agent] 2024-06-13 00:20:42 WARNING httpbin-pebble-ready   warnings.warn("feature x is deprecated, use feature y instead", DeprecationWarning)
```

No customer formatter, the line of the code that generates the warning is printed out.

Forwarded to Juju debug-log:

```bash
unit-sample-k8s-0: 00:20:42 WARNING unit.sample-k8s/0.httpbin-pebble-ready /var/lib/juju/agents/unit-sample-k8s-0/charm/./src/charm.py:40: DeprecationWarning: feature x is deprecated, use feature y instead
unit-sample-k8s-0: 00:20:42 WARNING unit.sample-k8s/0.httpbin-pebble-ready   warnings.warn("feature x is deprecated, use feature y instead", DeprecationWarning)
```

### After

The warning is still in the container log:

```bash
2024-06-13T00:30:04.252Z [container-agent] 2024-06-13 00:30:04 WARNING juju-log /var/lib/juju/agents/unit-sample-k8s-0/charm/./src/charm.py:40: DeprecationWarning: feature x is deprecated, use feature y instead
```

With the custom formatter, the line of the code isn't printed anymore.

In Juju debug-log:

```bash
unit-sample-k8s-0: 00:30:04 WARNING unit.sample-k8s/0.juju-log /var/lib/juju/agents/unit-sample-k8s-0/charm/./src/charm.py:40: DeprecationWarning: feature x is deprecated, use feature y instead
```

### DIFF
--- a/ops/log.py
+++ b/ops/log.py
@@ -65,8 +65,7 @@ def setup_root_logging(
         line: typing.Optional[str] = None,
     ) -> str:
         module_name = os.path.basename(filename)
-        warning_msg = f'{message}'.replace('\n', ' ')
-        return f'{module_name}:{lineno}: {category.__name__}: {warning_msg}'
+        return f'{module_name}:{lineno}: {category.__name__}: {message}'
 
     warnings.formatwarning = custom_warning_formatter
 
--- a/ops/log.py
+++ b/ops/log.py
@@ -15,9 +15,11 @@
 """Interface to emit messages to the Juju logging system."""
 
 import logging
+import os
 import sys
 import types
 import typing
+import warnings
 
 from ops.model import _ModelBackend
 
@@ -53,6 +55,21 @@ def setup_root_logging(
         debug: if True, write logs to stderr as well as to juju-log.
         exc_stderr: if True, write uncaught exceptions to stderr as well as to juju-log.
     """
+    logging.captureWarnings(True)
+
+    def custom_warning_formatter(
+        message: typing.Union[str, Warning],
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        line: typing.Optional[str] = None,
+    ) -> str:
+        module_name = os.path.basename(filename)
+        warning_msg = f'{message}'.replace('\n', ' ')
+        return f'{module_name}:{lineno}: {category.__name__}: {warning_msg}'
+
+    warnings.formatwarning = custom_warning_formatter
+
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
--- a/ops/log.py
+++ b/ops/log.py
@@ -15,7 +15,6 @@
 """Interface to emit messages to the Juju logging system."""
 
 import logging
-import os
 import sys
 import types
 import typing
@@ -25,7 +24,7 @@ from ops.model import _ModelBackend
 
 
 class JujuLogHandler(logging.Handler):
-    """A handler for sending logs to Juju via juju-log."""
+    """A handler for sending logs and warnings to Juju via juju-log."""
 
     def __init__(self, model_backend: _ModelBackend, level: int = logging.DEBUG):
         super().__init__(level)
@@ -50,6 +49,9 @@ def setup_root_logging(
 
       logging.getLogger().setLevel(logging.INFO)
 
+    Warnings issued by the warnings module are redirected to the logging system and
+    forwarded to juju-log, too.
+
     Args:
         model_backend: a ModelBackend to use for juju-log
         debug: if True, write logs to stderr as well as to juju-log.
@@ -64,8 +66,7 @@ def setup_root_logging(
         lineno: int,
         line: typing.Optional[str] = None,
     ) -> str:
-        module_name = os.path.basename(filename)
-        return f'{module_name}:{lineno}: {category.__name__}: {message}'
+        return f'{filename}:{lineno}: {category.__name__}: {message}'
 
     warnings.formatwarning = custom_warning_formatter
 
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -17,6 +17,7 @@ import logging
 import re
 import typing
 import unittest
+import warnings
 from unittest.mock import patch
 
 import pytest
@@ -155,6 +156,15 @@ class TestLogging:
         # Verify that it got split into the expected chunks.
         assert len(calls[1][1]) == MAX_LOG_LINE_LEN
         assert len(calls[2][1]) == 9
+
+    def test_warning(self, backend: FakeModelBackend):
+        ops.log.setup_root_logging(backend)
+        warnings.warn('warning')
+        calls = backend.calls()
+        assert len(calls) == 1
+        assert calls[0][0] == 'WARNING'
+        # Match any lineno, in case this file changes in the future.
+        assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
 
 
 if __name__ == '__main__':
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -157,14 +157,16 @@ class TestLogging:
         assert len(calls[1][1]) == MAX_LOG_LINE_LEN
         assert len(calls[2][1]) == 9
 
+    @pytest.mark.filterwarnings("error")
     def test_warning(self, backend: FakeModelBackend):
         ops.log.setup_root_logging(backend)
-        warnings.warn('warning')
-        calls = backend.calls()
-        assert len(calls) == 1
-        assert calls[0][0] == 'WARNING'
-        # Match any lineno, in case this file changes in the future.
-        assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
+        with pytest.raises(UserWarning):
+            warnings.warn('this is a test warning')
+            calls = backend.calls()
+            assert len(calls) == 1
+            assert calls[0][0] == 'WARNING'
+            # Match any lineno, in case this file changes in the future.
+            assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
 
 
 if __name__ == '__main__':
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -17,7 +17,6 @@ import logging
 import re
 import typing
 import unittest
-import warnings
 from unittest.mock import patch
 
 import pytest
@@ -156,17 +155,6 @@ class TestLogging:
         # Verify that it got split into the expected chunks.
         assert len(calls[1][1]) == MAX_LOG_LINE_LEN
         assert len(calls[2][1]) == 9
-
-    @pytest.mark.filterwarnings("error")
-    def test_warning(self, backend: FakeModelBackend):
-        ops.log.setup_root_logging(backend)
-        with pytest.raises(UserWarning):
-            warnings.warn('this is a test warning')
-            calls = backend.calls()
-            assert len(calls) == 1
-            assert calls[0][0] == 'WARNING'
-            # Match any lineno, in case this file changes in the future.
-            assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
 
 
 if __name__ == '__main__':
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -33,7 +33,7 @@ jobs:
           cat requirements.txt
 
       - name: Set up LXD
-        uses: canonical/setup-lxd@7be523c4c2724a31218a627809044c6a2f0870ad
+        uses: canonical/setup-lxd@e815787b53c690d006e582ef03e43f08d56d64ba
         with:
           channel: 5.0/stable
 
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -10,13 +10,13 @@ jobs:
       matrix:
         include:
           - charm-repo: canonical/postgresql-operator
-            commit: 4feeaeee102cbf5e3dada3c05d44e0495ca68f9a  # rev409 2024-05-21T12:52:24Z
+            commit: 85f92e9a7121a1f506a550662c066343612a29e6  # 2024-05-31T11:54:37Z
           - charm-repo: canonical/postgresql-k8s-operator
-            commit: 1a25c3929747beea7e78467b169b2b345b29d470  # 2024-05-21T12:40:19Z
+            commit: 7e8862e2f02cfe1f8e768d6ad54a281833afdfda  # rev268 2024-05-31T11:54:09Z
           - charm-repo: canonical/mysql-operator
             commit: 19633f3e904d1c3296477b3df191d1ca265fc0d5  # rev234 2024-05-06T12:13:54Z
           - charm-repo: canonical/mysql-k8s-operator
-            commit: 6c09910bc3bd88eb632793d08fa17340c6903cb2  # rev138 2024-05-01T18:08:13Z
+            commit: d310ba925b667345dc2d1b9b5c953881dd14322a  # 2024-05-29T12:07:56Z
     steps:
       - name: Checkout the ${{ matrix.charm-repo }} repository
         uses: actions/checkout@v4
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -11,11 +11,11 @@ jobs:
       matrix:
         include:
           - charm-repo: canonical/alertmanager-k8s-operator
-            commit: 90b85c79dfdeeeedcc538c92dcbc26fb2b931088  # rev114 2024-05-23T12:09:22Z
+            commit: 81612b18bf4e9cee072387e75ebd34dff3c189ce  # rev119 2024-06-02T12:08:25Z
           - charm-repo: canonical/prometheus-k8s-operator
             commit: 41b10003b2e7aba34e26fa387af4297d97ecb535  # rev189 2024-05-21T21:25:20Z
           - charm-repo: canonical/grafana-k8s-operator
-            commit: ec74910fc60848594ce44da3b549ad18aa6528aa  # rev113 2024-05-21T14:31:49Z
+            commit: fd6a7da88fa044709be69a4c733d879eb7175b54  # rev114 2024-05-29T09:55:04Z
     steps:
       - name: Checkout the ${{ matrix.charm-repo }} repository
         uses: actions/checkout@v4
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,26 @@
+# 2.14.0 - 29 May 2024
+
+This release fixes the `RelationDataContent.update` method to follow `dict.update` semantics, that is allow both updating with another dict, an iterable, keyword arguments or a mixture thereof.
+
+## Features
+
+* feat: add a `__str__` to ActionFailed, for better unexpected failure output in https://github.com/canonical/operator/pull/1209
+
+## Fixes
+
+* The `other` argument to `RelatationDataContent.update(...)` should be optional by @addyess in https://github.com/canonical/operator/pull/1226
+
+## Documentation
+
+* Use the actual emoji character rather than GitHub markup, to show properly on PyPI in https://github.com/canonical/operator/pull/1221
+* Clarify that SecretNotFound may be raised for permission errors in https://github.com/canonical/operator/pull/1231
+
+## Refactoring
+
+* Refactor tests to pytest style in https://github.com/canonical/operator/pull/1199 https://github.com/canonical/operator/pull/1200 https://github.com/canonical/operator/pull/1203 https://github.com/canonical/operator/pull/1206
+* Use `ruff` formatter and reformat all code in https://github.com/canonical/operator/pull/1224
+* Don't use f-strings in logging calls in https://github.com/canonical/operator/pull/1227 https://github.com/canonical/operator/pull/1234
+
 # 2.13.0 - 30 Apr 2024
 
 ## Features
--- a/HACKING.md
+++ b/HACKING.md
@@ -291,15 +291,18 @@ To make a release of the ops library, do the following:
 7. Create a new branch, and copy this text to the [CHANGES.md](CHANGES.md) file,
    stripping out links, who did each commit, the new contributor list, and the
    link to the full changelog.
-8. Change [version.py][ops/version.py]'s `version` to the
+8. Change [version.py](ops/version.py)'s `version` to the
    [appropriate string](https://semver.org/).
-9. Add, commit, and push, and open a PR to get the changelog and version bump
+9. Check if there's a `chore: update charm pins` auto-generated PR in the queue. If it looks
+   good, merge it and check that tests still pass. If needed, you can re-trigger the
+   `Update Charm Pins` workflow manually to ensure latest charms and ops get tested.
+10. Add, commit, and push, and open a PR to get the changelog and version bump
    into main (and get it merged).
-10. Back in the GitHub releases page, tweak the release notes - for example,
+11. Back in the GitHub releases page, tweak the release notes - for example,
    you might want to have a short paragraph at the intro on particularly
    noteworthy changes.
-11. Have someone else in the Charm-Tech team proofread the release notes.
-12. When you are ready, click "Publish". (If you are not ready, click "Save as Draft".)
+12. Have someone else in the Charm-Tech team proofread the release notes.
+13. When you are ready, click "Publish". (If you are not ready, click "Save as Draft".)
 
 This will trigger an automatic build for the Python package and publish it to
 [PyPI](https://pypi.org/project/ops/)) (authorisation is handled via a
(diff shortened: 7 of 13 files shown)


test/test_log.py
```diff
@@ -156,6 +157,15 @@ def test_long_string_logging(self, backend: FakeModelBackend, logger: logging.Lo
         assert len(calls[1][1]) == MAX_LOG_LINE_LEN
         assert len(calls[2][1]) == 9
 
+    def test_warning(self, backend: FakeModelBackend):
+        ops.log.setup_root_logging(backend)
+        warnings.warn('warning')
```
tonyandrewmeyer, line 162: Probably best to use some other string, in case we had a bug that printed out the log level instead of the message.

ops/log.py
```diff
@@ -53,6 +55,21 @@ def setup_root_logging(
         debug: if True, write logs to stderr as well as to juju-log.
         exc_stderr: if True, write uncaught exceptions to stderr as well as to juju-log.
     """
+    logging.captureWarnings(True)
```
tonyandrewmeyer, line 58: Something I haven't thought through (maybe you already have?) is what happens when testing. Without this change, would the warnings end up in the pytest output? With the change, do they vanish because juju debug-log output isn't displayed? Are Harness and Scenario different here? I think probably they would still show in Scenario.

Actually, this is related to what you asked me about the test failing when you run the whole suite. I guess we first want to figure out what the desired behaviour is. Maybe a good team discussion?

ops/log.py
```diff
@@ -53,6 +55,21 @@ def setup_root_logging(
...
+
+    def custom_warning_formatter(
+        message: typing.Union[str, Warning],
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        line: typing.Optional[str] = None,
+    ) -> str:
+        module_name = os.path.basename(filename)
+        warning_msg = f'{message}'.replace('\n', ' ')
```
benhoyt, line 68: Why do we replace `\n` in the message (but not other characters like `\t`)? Presumably warning messages are typically a single line. If not, it's probably okay that they get logged multi-line too. I'd recommend we just pass the `message` straight through -- that's what the default one does.

test/test_log.py
```diff
@@ -156,6 +157,15 @@ def test_long_string_logging(self, backend: FakeModelBackend, logger: logging.Lo
...
+        calls = backend.calls()
+        assert len(calls) == 1
+        assert calls[0][0] == 'WARNING'
```
benhoyt, line 165: Shouldn't we test that the correct backend method is called, specifically `juju_log`? (Or use `FakeScript` and fake out the `juju-log` script itself?)

ops/log.py
benhoyt: Ah, I see @tonyandrewmeyer's [comment](https://github.com/canonical/operator/issues/1077#issuecomment-2138493463) about this now. However, it doesn't look like the default implementation does this: https://github.com/python/cpython/blob/bd0d97ce3497da35d17daa329c294f4cb87196ee/Lib/warnings.py#L35

tonyandrewmeyer: I may have caused some confusion here, sorry. I mentioned to @IronCore864 that the warning messages should be single line - I think multi-line unstructured log messages are a bad idea, even though we do already have these in debug-log. That was because the default warning format can be multiline (the docs call this out), which is when the code snippet is included, I think.

If we have our own custom format that doesn't have any newlines outside of the message, then I don't think we need to do this. If someone does `warning.warn("hello\nworld")` then that's on them.

test/test_log.py
IronCore864: In this test file, the whole backend is mocked, rather than just the `juju_log` method. And `FakeModelBackend.juju_log` append items to `backend.calls()`, so it's not necessary to assert that `juju_log` is called, right?

ops/log.py
IronCore864: Without `logging.captureWarnings(True)`, the warning shows in pytest's output, and the test case fails, because warnings aren't redirectedted, so it's still in the output, not in logging, so `JujuLogHandler` can't catch it.

And yes, with the change, the warnings won't show up in the output because this line essentially redirects warnings to the logging system.

IronCore864: [In the original issue](https://github.com/canonical/operator/issues/1077#issuecomment-2136815485) I mentioned two methods.

One is in the draft PR, which if se use logging.captureWarnings(), warnings won't show up in the output any more, because warnings are redirected to the logging system.

The other way I mentioned is showwarning(), defaults to stderr, if we want to keep warnings in the output and in the logging at the same time, we can do it this way, log it, then print it.

To me, I think keeping the warnings in one place is better, no duplication. But maybe in the stderr is better than in juju debug log, because users expect warnings to be in stderr, and users may have centralized logging systems and it'd be hard for them to figure out why warnings don't show up there.

Which behaviour do we want? @benhoyt Let's have a quick discussion in our next daily.

test/test_log.py
```diff
@@ -156,6 +157,17 @@ def test_long_string_logging(self, backend: FakeModelBackend, logger: logging.Lo
...
+    @pytest.mark.filterwarnings("error")
+    def test_warning(self, backend: FakeModelBackend):
+        ops.log.setup_root_logging(backend)
+        with pytest.raises(UserWarning):
+            warnings.warn('this is a test warning')
+            calls = backend.calls()
+            assert len(calls) == 1
+            assert calls[0][0] == 'WARNING'
+            # Match any lineno, in case this file changes in the future.
+            assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
```
tonyandrewmeyer, line 169: Shouldn't this end with `UserWarning: this is a test warning`? It seems odd that it would pass as-is, unless I'm missing something?

ops/log.py
```diff
@@ -53,6 +55,20 @@ def setup_root_logging(
...
         exc_stderr: if True, write uncaught exceptions to stderr as well as to juju-log.
     """
+    logging.captureWarnings(True)
+
+    def custom_warning_formatter(
+        message: typing.Union[str, Warning],
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        line: typing.Optional[str] = None,
```
tonyandrewmeyer, line 65: This is tangential to the PR, but I would have thought ruff would complain here about `line` not being used, and require a pragma to ignore it. But I guess it is not doing that.

tonyandrewmeyer (commented):
I think we should expand the `setup_root_logging` docstring to mention this (could you also fix the `python` to be `Python` while you're there? It bugs me but not enough to open a one character PR :joy:).

I guess we don't really mention that logs go to debug-log elsewhere in the reference docs, so other doc updates would be in how-to guides or the tutorial.

benhoyt (approved):
I agree with @tonyandrewmeyer's comments about documenting this, but after that I think I'm good with this.

I also think you should be sure to do an end-to-end test of this under Juju (if you have't already) and report the results here.

ops/log.py
```diff
@@ -53,6 +55,20 @@ def setup_root_logging(
...
+    ) -> str:
+        module_name = os.path.basename(filename)
```
dimaqq, line 67: Super-minor: I thought it was odd to `import os` at the top.
Other options:
- `filename.rsplit("/", 1)[-1]`
- `from os.path import basename`
- log the full path, because intermediate elements can be useful, e.g. `ops/__init__.py` vs `requests/__init__.py`

test/test_log.py
dimaqq: Ha! this line is never executed, because the whole thing is wrapped into `with pytest.raises()` and the code raises early.

`backend.calls()` is actually empty.

dimaqq (changes requested):

test/test_log.py
tonyandrewmeyer: I dug into this a bit more, and this test just doesn't work. Because of the decorator, `warnings.warn('this is a test warning')` raises an exception, which means that none of the lines after it in the with block ever run. The test passes because the exception is a `UserWarning`, but none of the asserts are ever executed.

For example:

```diff
diff --git a/test/test_log.py b/test/test_log.py
index 2961155..93cb3cf 100644
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -162,6 +162,7 @@ class TestLogging:
         ops.log.setup_root_logging(backend)
         with pytest.raises(UserWarning):
             warnings.warn('this is a test warning')
+            assert False, "this will never be reached"
             calls = backend.calls()
             assert len(calls) == 1
             assert calls[0][0] == 'WARNING'
```

```shell
$ tox -e unit -- -k test_warning
[...]
test/test_log.py::TestLogging::test_warning PASSED                                                                                            [ 25%]
[...]
----------------------------------------------------------------------
TOTAL                       6501   4523   2433     42    26%
  unit: OK (1.36=setup[0.02]+cmd[0.74,0.60] seconds)
  congratulations :) (1.39 seconds)
```

To get the asserts to run, they need to be moved out of the `with`. However, the test will then fail, because it's checking that the warning was captured and sent to the debug log, but that didn't happen, because it was upgraded to an exception, so it didn't go to the logging system.

```diff
diff --git a/test/test_log.py b/test/test_log.py
index 2961155..f737b74 100644
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -162,11 +162,11 @@ class TestLogging:
         ops.log.setup_root_logging(backend)
         with pytest.raises(UserWarning):
             warnings.warn('this is a test warning')
-            calls = backend.calls()
-            assert len(calls) == 1
-            assert calls[0][0] == 'WARNING'
-            # Match any lineno, in case this file changes in the future.
-            assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
+        calls = backend.calls()
+        assert len(calls) == 1
+        assert calls[0][0] == 'WARNING'
+        # Match any lineno, in case this file changes in the future.
+        assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
 
 
 if __name__ == '__main__':
```

```shell
$ tox -e unit -- -k test_warning
[...]
test/test_log.py::TestLogging::test_warning FAILED                                                                                            [ 25%]
[...]
===================================================================== FAILURES ======================================================================
_____________________________________________________________ TestLogging.test_warning ______________________________________________________________
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
                                ^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/_pytest/runner.py", line 262, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_hooks.py", line 513, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_manager.py", line 120, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_callers.py", line 182, in _multicall
    return outcome.get_result()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_result.py", line 100, in get_result
    raise exc.with_traceback(exc.__traceback__)
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_callers.py", line 103, in _multicall
    res = hook_impl.function(*args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/_pytest/runner.py", line 177, in pytest_runtest_call
    raise e
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/_pytest/runner.py", line 169, in pytest_runtest_call
    item.runtest()
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/_pytest/python.py", line 1792, in runtest
    self.ihook.pytest_pyfunc_call(pyfuncitem=self)
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_hooks.py", line 513, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_manager.py", line 120, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_callers.py", line 139, in _multicall
    raise exception.with_traceback(exception.__traceback__)
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/pluggy/_callers.py", line 103, in _multicall
    res = hook_impl.function(*args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/.tox/unit/lib/python3.12/site-packages/_pytest/python.py", line 194, in pytest_pyfunc_call
    result = testfunction(**testargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/scratch/tiexenoperator/test/test_log.py", line 166, in test_warning
    assert len(calls) == 1
AssertionError: assert 0 == 1
 +  where 0 = len([])
============================================================== short test summary info ==============================================================
FAILED test/test_log.py::TestLogging::test_warning - assert 0 == 1
=================================================== 1 failed, 3 passed, 1179 deselected in 0.53s ====================================================
unit: exit 1 (0.77 seconds) /home/tameyer/scratch/tiexenoperator> coverage run --source=ops/ -m pytest --ignore=test/smoke -v --tb native -k test_warning pid=999475
  unit: FAIL code 1 (0.79=setup[0.02]+cmd[0.77] seconds)
  evaluation failed :( (0.83 seconds)
```

If we're going to test this (and I do think ideally we would) then we need to do it some other way.

tonyandrewmeyer (changes requested):

test/test_log.py
tonyandrewmeyer: > Ha! this line is never executed, because the whole thing is wrapped into `with pytest.raises()` and the code raises early.
> 
> `backend.calls()` is actually empty.

Ah, snap :).

ops/log.py
tonyandrewmeyer: I don't think we should do `from os.path import basename`. If we end up using something else from `os` in the future then we add more and more names to the local namespace, rather than just `os` and having `os` things nicely contained in there :).

Doing a split on a filename is bad practice - it should at least be using `os.sep` and not `/` (and then you're importing `os` again). `os` is imported on startup, so there's almost no cost to importing it here. The modern approach would be for these to be `pathlib.Path` objects and then to do `filename.name`, but this is older code so we get a string not a `Path` and converting it to a `Path` just to get the name seems overkill.

A good point about clashing names (although I think it would be less common than you'd get elsewhere since we don't use many packages and charm libs are single files). I had suggested that we shouldn't log the full path because debug-log is super noisy already and this will make it even more noisy, and a chunk of the initial path is not useful, e.g.: `/var/lib/juju/agents/unit-[app name]-[unit number]/charm/src` for charm code (and the packages path for third-party and ops warnings, although we're more concerned with charm warnings in this situation).

We could try to intelligently strip off the part that's not of interest, but that's possibly over-complicating this, particularly as an initial version. So probably better to indeed just go back to the full filename.

tonyandrewmeyer (changes requested):

ops/log.py
dimaqq: I wish we had `__qualname__`, but alas! The warnings module doesn't give us that.

dimaqq:
I'm not sure if this matters... when I run all tests with `pytest` (main branch or this), I get a notification that "io capturing turned off" for the test file that tests `pdb.set_trace`.

I hope there's no overlap with warnings capture.

test/test_log.py
tonyandrewmeyer: This does work:

```python
    def test_warning(self, backend: FakeModelBackend):
        ops.log.setup_root_logging(backend)
        warnings.warn_explicit('this is a test warning', UserWarning, 'test_log.py', 100)
        calls = backend.calls()
        assert len(calls) == 1
        assert calls[0][0] == 'WARNING'
        assert re.search(r'test_log\.py:100: UserWarning: this is a test warning', calls[0][1])
```

This isn't precisely testing what happens in prod code, since charms would call `warnings.warn` rather than `warnings.warn_explicit`. However, it's pretty close ([code](https://github.com/python/cpython/blob/0ae8579b85f9b0cd3f287082ad6e194bdb025d88/Lib/warnings.py#L349) with the `warn` code right above).

I think we could also do it by clearing the filters (?) that pytest sets and then restoring them afterwards, but I think just using `warn_explicit` is probably cleaner.

If we really wanted to be sure of what happens we could do a functional test (like the test_main ones) that start up a charm (without Juju) that generates a warning, and have a dummy debug-log hook tool that records use. However, that seems overkill for a small feature like this.

tonyandrewmeyer: Actually, I think it would be fairly simple to add it to [the existing test_main log test](https://github.com/canonical/operator/blob/cf329e60c16ef4f1c36db51daf1a0ffbc19c7e9f/test/test_main.py#L823), just adding in an action to generate a warning, like the [ones that generate logs](https://github.com/canonical/operator/blob/cf329e60c16ef4f1c36db51daf1a0ffbc19c7e9f/test/charms/test_main/src/charm.py#L272). Since that's a subprocess, pytest wouldn't do anything with the warnings. I haven't tested this (I did test `warn_explicit`) but it seems like it would be fine.

tonyandrewmeyer: > This does work:

Maybe not true! I think got sucked in by the "run one test and it works, run them all and it fails" case. However, when I added the test_main case it did pass when they all ran. So I'm a bit :confounded: on that one. Maybe it's the clean-up? You could dig more if you think it's still going to be unstable.

Anyway, this does pass reliably (one test or all) for me:

```diff
--- a/test/charms/test_main/actions.yaml
+++ b/test/charms/test_main/actions.yaml
@@ -29,3 +29,5 @@ log-info:
   description: log a message at Info level
 log-debug:
   description: log a message at Debug level
+warn:
+  description: generate a deprecation warning
diff --git a/test/charms/test_main/src/charm.py b/test/charms/test_main/src/charm.py
index e987d34..ddd80bc 100755
--- a/test/charms/test_main/src/charm.py
+++ b/test/charms/test_main/src/charm.py
@@ -17,6 +17,7 @@ import logging
 import os
 import sys
 import typing
+import warnings
 
 import ops
 
@@ -63,6 +64,7 @@ class Charm(ops.CharmBase):
             on_log_warning_action=[],
             on_log_info_action=[],
             on_log_debug_action=[],
+            on_warn_action=[],
             on_secret_changed=[],
             on_secret_remove=[],
             on_secret_rotate=[],
@@ -107,6 +109,7 @@ class Charm(ops.CharmBase):
             self.framework.observe(self.on.log_warning_action, self._on_log_warning_action)
             self.framework.observe(self.on.log_info_action, self._on_log_info_action)
             self.framework.observe(self.on.log_debug_action, self._on_log_debug_action)
+            self.framework.observe(self.on.warn_action, self._on_warn_action)
 
         self.framework.observe(self.on.collect_metrics, self._on_collect_metrics)
         self.framework.observe(self.on.custom, self._on_custom)
@@ -284,6 +287,9 @@ class Charm(ops.CharmBase):
     def _on_log_debug_action(self, event: ops.ActionEvent):
         logger.debug('insightful debug')
 
+    def _on_warn_action(self, event: ops.ActionEvent):
+        warnings.warn('feature x is going away', DeprecationWarning)
+
     def _on_get_model_name_action(self, event: ops.ActionEvent):
         self._stored._on_get_model_name_action.append(self.model.name)
 
diff --git a/test/test_log.py b/test/test_log.py
index 2961155..ac63458 100644
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -15,6 +15,7 @@
 import io
 import logging
 import re
+import sys
 import typing
 import unittest
 import warnings
@@ -51,6 +52,8 @@ def logger():
     logger = logging.getLogger()
     yield logger
     logging.getLogger().handlers.clear()
+    sys.excepthook = sys.__excepthook__
+    logging.captureWarnings(False)
 
 
 class TestLogging:
@@ -157,16 +160,13 @@ class TestLogging:
         assert len(calls[1][1]) == MAX_LOG_LINE_LEN
         assert len(calls[2][1]) == 9
 
-    @pytest.mark.filterwarnings("error")
     def test_warning(self, backend: FakeModelBackend):
         ops.log.setup_root_logging(backend)
-        with pytest.raises(UserWarning):
-            warnings.warn('this is a test warning')
-            calls = backend.calls()
-            assert len(calls) == 1
-            assert calls[0][0] == 'WARNING'
-            # Match any lineno, in case this file changes in the future.
-            assert re.search(r'test_log\.py:\d+: UserWarning: warning', calls[0][1])
+        warnings.warn_explicit('this is a test warning', UserWarning, 'test_log.py', 100)
+        calls = backend.calls()
+        assert len(calls) == 1
+        assert calls[0][0] == 'WARNING'
+        assert re.search(r'test_log\.py:100: UserWarning: this is a test warning', calls[0][1])
 
 
 if __name__ == '__main__':
diff --git a/test/test_main.py b/test/test_main.py
index 48cfbbc..582e5e0 100644
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -869,6 +869,26 @@ class _TestMain(abc.ABC):
             self._simulate_event(fake_script, event_spec)
             assert calls in fake_script.calls(clear=True)
 
+        # This case is annoying to test, because the message changes depending on which
+        # class is executing the test.
+        event_spec = EventSpec(
+            ops.ActionEvent,
+            'warn_action',
+            env_var='JUJU_ACTION_NAME',
+            set_in_env={'JUJU_ACTION_UUID': '5'},
+        )
+        self._simulate_event(fake_script, event_spec)
+        calls = fake_script.calls(clear=True)
+        for call in calls:
+            # TODO: Probably should check this with a regex to see that there is
+            # something like a filename and a line number.
+            if call[:4] == ['juju-log', '--log-level', 'WARNING', '--'] and call[4].endswith(
+                ': DeprecationWarning: feature x is going away'
+            ):
+                break
+        else:
+            pytest.fail(f'Warning was not sent to debug-log: {calls!r}')
+
     @pytest.mark.usefixtures('setup_charm')
     def test_excepthook(self, fake_script: FakeScript):
         with pytest.raises(subprocess.CalledProcessError):
```

IronCore864:
Also added a docstring in `ops/log.py`, but since the `JujuLogHandler` class isn't in `__all__`, it won't appear in the ops library API documentation.

There seems to be no good place in the API documentation to add something about logging. Maybe `ops.CharmBase`? I prefer to add something maybe in the [Juju SDK doc on the ops page](https://juju.is/docs/sdk/ops) in the reference section.

tonyandrewmeyer:
> There seems to be no good place in the API documentation to add something about logging. Maybe `ops.CharmBase`? I prefer to add something maybe in the [Juju SDK doc on the ops page](https://juju.is/docs/sdk/ops) in the reference section.

Agreed. I don't think it fits with the API reference docs (since a charm should never be calling `setup_root_logging` directly - the docstring improvement there is for tools that build on ops, like Scenario).

I think we should add it to [the how-to guide for logging](https://juju.is/docs/sdk/logging). For example, a new section that explains how to warn (particularly about deprecation, but mentioning that runtime, user, etc warnings can be used) and where those warnings appear, both in prod and when running tests.

IronCore864:
After some testing, it turned out that currently, the stderr was already captured by Juju.

So, if a charm generates a warning, it will show up in Juju's debug log. This is verified with a K8s charm and @tonyandrewmeyer verified it with machine charms as well. Currently we do not have a custom formatter for warnings, meaning the warnings in Juju debug log isn't super nice, but it will serve the purpose at the moment. So, I'm closing this PR now.

Issue #123, opened 2020-01-31
Introduce base classes for interface into the framework


majduk:
Currently the InterfaceProvides and InterfaceRequires classes inherit after ops.framework.Object class. I would like to propose creating interim base class serving as a base for interface and including what effectively is a boilerplate code:
- relation name setup
- relations and fields properties
- is joined status
- is ready status
- goal state

stub42:
I'm interested in base classes to guide people towards preferred patterns. Do we create a class to handle a single relation, or a class for the endpoint that handles all relations using a particular relation_name? If both are patterns we want to promote, how should we name them?

petevg:
Since we've moved away from integrating the requires and provides concepts into the framework, I believe that this issue can be closed.

I agree that we should be providing guidance on best practices. I believe that current work on libraries represents our best efforts to this end.

PR #1241, opened 2024-05-30
doc: adjust the 2.14 changelog to match our style

tonyandrewmeyer:
* No intro text as found on the GitHub page and various announcements.
* No attribution.
* No links to issues.
* Remove conventional commit type in favour of headings.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,25 +1,23 @@
 # 2.14.0 - 29 May 2024
 
-This release fixes the `RelationDataContent.update` method to follow `dict.update` semantics, that is allow both updating with another dict, an iterable, keyword arguments or a mixture thereof.
-
 ## Features
 
-* feat: add a `__str__` to ActionFailed, for better unexpected failure output in https://github.com/canonical/operator/pull/1209
+* Add a `__str__` to ActionFailed, for better unexpected failure output (#1209)
 
 ## Fixes
 
-* The `other` argument to `RelatationDataContent.update(...)` should be optional by @addyess in https://github.com/canonical/operator/pull/1226
+* The `other` argument to `RelatationDataContent.update(...)` should be optional (#1226)
 
 ## Documentation
 
-* Use the actual emoji character rather than GitHub markup, to show properly on PyPI in https://github.com/canonical/operator/pull/1221
-* Clarify that SecretNotFound may be raised for permission errors in https://github.com/canonical/operator/pull/1231
+* Use the actual emoji character rather than GitHub markup, to show properly on PyPI (#1221)
+* Clarify that SecretNotFound may be raised for permission errors (#1231)
 
 ## Refactoring
 
-* Refactor tests to pytest style in https://github.com/canonical/operator/pull/1199 https://github.com/canonical/operator/pull/1200 https://github.com/canonical/operator/pull/1203 https://github.com/canonical/operator/pull/1206
-* Use `ruff` formatter and reformat all code in https://github.com/canonical/operator/pull/1224
-* Don't use f-strings in logging calls in https://github.com/canonical/operator/pull/1227 https://github.com/canonical/operator/pull/1234
+* Refactor tests to pytest style (#1199, #1200, #1203, #1206)
+* Use `ruff` formatter and reformat all code (#1224)
+* Don't use f-strings in logging calls (#1227, 1234)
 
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,25 +1,23 @@
 # 2.14.0 - 29 May 2024
 
-This release fixes the `RelationDataContent.update` method to follow `dict.update` semantics, that is allow both updating with another dict, an iterable, keyword arguments or a mixture thereof.
```
benhoyt, line 3: I think all of these changes are fair enough, but this one we did deliberately: we thought it was a good precedent to include the summary paragraph in CHANGES.md moving forward (though no need to add them to the others retroactively). What do you think?

tonyandrewmeyer: As a meta note: I think we should definitely update the instructions in HACKING.md if we're going to make changes like that, so that we can be consistent and all understand what the expectations are (particularly when it's counter to something that's already been discussed as a team). Changing decisions is fine, but best done more openly.

I think it depends on what we see as the purpose of this file, particularly compared to the GtiHub releases page. This tangentially relates to Dora's suggestion to link to the releases page from ops.readthedocs.io, too.

If CHANGES.md primarily exists so that the changes are present in artefacts (both the GitHub tarballs and the PyPI sdist) then having it closer to the GitHub releases (including the intro section) makes sense. If we do this then I think we probably also want to change the project.urls link to point to the GitHub releases page rather than to CHANGES.md.

If it exists to (also) be more of a reference - e.g. something to scan through to check for anything that you need to be aware of when upgrading, then I think a plainer log of changes makes sense. In this form, it's a mid point between `git log` and the releases. This is how I used changelogs in the past, but these days I'm most often seeing dependency bumps with Dependabot and that pulls in the release notes (at least for packages with GitHub releases, like ops).

Looking around at popular Python packages, they are almost all bulleted lists (logs) of changes, without any sort of preamble that you find in release announcements. My inclination is to fit with the ecosystem's pattern.

benhoyt: Good point about updating `HACKING.md` if we do this.

> Looking around at popular Python packages, they are almost all bulleted lists (logs) of changes, without any sort of preamble that you find in release announcements. My inclination is to fit with the ecosystem's pattern.

Fair enough. Let's run with that.

benhoyt (approved):

CHANGES.md
```diff
@@ -1,25 +1,23 @@
...
-
 ## Features
 
-* feat: add a `__str__` to ActionFailed, for better unexpected failure output in https://github.com/canonical/operator/pull/1209
+* Add a `__str__` to ActionFailed, for better unexpected failure output (#1209)
```
dimaqq, line 5: So sorry about these.

GitHub auto-generated sentense-cased our items and I lower-cased those manually... or did it? Now I'm not sure.

I then copy-pasted that to the change log.

Maybe we could agree on a single convention for both 🤔 

Looking through /releases, we've got a mix of sentence-cased bullet points without prefixes, like "Add this" and lower-case, prefixed, like "feat: add that" releases.

CHANGES.md
```diff
@@ -1,25 +1,23 @@
...
-* Clarify that SecretNotFound may be raised for permission errors in https://github.com/canonical/operator/pull/1231
+* Use the actual emoji character rather than GitHub markup, to show properly on PyPI (#1221)
+* Clarify that SecretNotFound may be raised for permission errors (#1231)
 
 ## Refactoring
 
-* Refactor tests to pytest style in https://github.com/canonical/operator/pull/1199 https://github.com/canonical/operator/pull/1200 https://github.com/canonical/operator/pull/1203 https://github.com/canonical/operator/pull/1206
-* Use `ruff` formatter and reformat all code in https://github.com/canonical/operator/pull/1224
-* Don't use f-strings in logging calls in https://github.com/canonical/operator/pull/1227 https://github.com/canonical/operator/pull/1234
+* Refactor tests to pytest style (#1199, #1200, #1203, #1206)
```
dimaqq, line 18: As long as the CHANGELOG stays at GitHub and doesn't end up on e.g. PyPi, I'm very happy with `#123` notation, as it's simpler. 👍🏻 

Stuff that we copy out, like to matrix or discourse needs to have full links though.

dimaqq (approved):
Wrt. release summary paragraph, I'm fine either way.

PR #1245, opened 2024-06-05
fix: fix TypeError when running test.pebble_cli

benhoyt:
Without this fix, running this raises the following:

```
TypeError: catching classes that do not inherit from BaseException is not allowed
```

Not sure how to silence pyright without the type: ignores.

I guess we haven't used this script in a while. :-)

### DIFF
--- a/test/pebble_cli.py
+++ b/test/pebble_cli.py
@@ -246,9 +246,8 @@ def main():
                     if stderr:
                         print(repr(stderr), end='', file=sys.stderr)
                 sys.exit(0)
-            # The `[str]` here is to resolve the same issue as above.
-            except pebble.ExecError[str] as e:
-                print('ExecError:', e, file=sys.stderr)
+            except pebble.ExecError as e:  # type: ignore
+                print('ExecError:', e, file=sys.stderr)  # type: ignore
                 sys.exit(e.exit_code)
 
         elif args.command == 'ls':


dimaqq (approved):
Given that it’s a script and typing issue get fixed…

I don’t quite understand the original type hint though.

tonyandrewmeyer:
How do you reproduce the error?

I've started pebble with:

```shell
PEBBLE=/tmp/pebble go run ./cmd/pebble run
```

and running `test/pebble_cli` from main seems ok?

```shell
$ PEBBLE=/tmp/pebble /tmp/opsenv/bin/python3 test/pebble_cli.py system-info
SystemInfo(version='v1.10.0-dev')
$ /tmp/opsenv/bin/python3 --version
Python 3.12.3
``

Are you using a different Python? Or should I run it differently?

**Comment by benhoyt** [comment]
@tonyandrewmeyer You have to run an `exec` command specifically. I can repro it by starting Pebble, then running this:

```
$ .tox/unit/bin/python -m test.pebble_cli --socket=/var/lib/pebble/default/.pebble.socket exec -- echo foo
foo
  File "/home/ben/w/operator/test/pebble_cli.py", line 248, in main
    sys.exit(0)
SystemExit: 0

During handling of the above exception, another exception occurred:

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/ben/w/operator/test/pebble_cli.py", line 316, in <module>
    main()
  File "/home/ben/w/operator/test/pebble_cli.py", line 250, in main
    except pebble.ExecError[str] as e:
TypeError: catching classes that do not inherit from BaseException is not allowed
```

tonyandrewmeyer (approved):
I have the feeling I maybe introduced this - when I was adding hints to all the test files the older version of pyright seemed to want the exceptions like this, although it's clearly invalid.

I also don't know of any way to say "catch this exception and I know that it will be this concrete type". So I think the first ignore is required.

The second ignore could be removed if you have a `e = typing.cast(pebble.ExecError[str], e)` but I think having the ignore there is fine in this context.

It feels like, just as with Python typing in many places, having generic exception classes doesn't work very well.

PR #1247, opened 2024-06-05
fix: add connect timeout for exec websockets to avoid hanging

### LINKED ISSUE #1246: Websocket connect hangs if Pebble timeout elapses first
If the Python code takes too long to run between the `POST /v1/exec` to start a Pebble exec, and connecting to all (2 or 3, depending on `combine_stderr`) of the websockets, the Python code will hang indefinitely and never time out. This is because there's no socket timeout set during the websocket connect phase, and Pebble waits rather than rejecting the connection (even though it's already exceeded its [waitIOConnected](https://github.com/canonical/pebble/blob/20263d04d06bb8a73295bdaecdd89a887c2d744a/internals/overlord/cmdstate/handlers.go#L167) timeout).

You can test this by adding `time.sleep(5.1)` between the POST and the websocket connections, here:

```diff
diff --git a/ops/pebble.py b/ops/pebble.py
index 831c778..ad2fb3e 100644
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2751,6 +2751,7 @@ class Client:
 
         stderr_ws: Optional[_WebSocket] = None
         try:
+            time.sleep(5.1)
             control_ws = self._connect_websocket(task_id, 'control')
             stdio_ws = self._connect_websocket(task_id, 'stdio')
             if not combine_stderr:
```

Then fire up `pebble run` in one terminal, and run a Pebble exec in another:

```
$ .tox/unit/bin/python -m test.pebble_cli exec -- echo foo
# after 5s, the Pebble logs will show "timeout waiting for websocket connections",
# but it will hang here
```

We should almost certainly have a (relatively short) timeout on the socket during connect. Though we have to unset the timeout during after it's connected, as the websockets for control and stdio are essentially long-polling, and will wait an arbitrary amount of time till input arrives. With this fix, we get this (after 10s = 5s Pebble timeout + 5s connect timeout):

```
$ .tox/unit/bin/python -m test.pebble_cli --socket=/var/lib/pebble/default/.pebble.socket exec -- echo foo
ChangeError: cannot perform the following tasks:
- Execute command "echo" (exec 31: timeout waiting for websocket connections: context deadline exceeded)
----- Logs from task 0 -----
2024-06-05T16:08:18+12:00 ERROR exec 31: timeout waiting for websocket connections: context deadline exceeded
-----
```

We can probably also improve Pebble's handling of this, as it should know that the `waitIOConnected` bit has already timed out, but fixing it in Ops will be a great start! I'll push up a PR soon.

benhoyt:
This is the fix for the issue described at https://github.com/canonical/operator/issues/1246. Essentially, if the Pebble timeout has already elapsed, Pebble will happily wait indefinitely for the connect to go through, and the Python side will hang. Add a timeout during the connect phase to cut this short.

Fixes #1246.

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2832,10 +2832,17 @@ class Client:
 
     def _connect_websocket(self, task_id: str, websocket_id: str) -> '_WebSocket':
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        # Set socket timeout to a short timeout during connection phase, in
+        # case Pebble side times out (5s), so this side doesn't hang. See:
+        # https://github.com/canonical/operator/issues/1246
+        sock.settimeout(self.timeout)
         sock.connect(self.socket_path)
         url = self._websocket_url(task_id, websocket_id)
         ws: _WebSocket = websocket.WebSocket(skip_utf8_validation=True)  # type: ignore
         ws.connect(url, socket=sock)
+        # Reset to no timeout so connection can be "long polling" when data is
+        # being received.
+        sock.settimeout(None)
         return ws
 
     def _websocket_url(self, task_id: str, websocket_id: str) -> str:
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2833,7 +2833,7 @@ class Client:
     def _connect_websocket(self, task_id: str, websocket_id: str) -> '_WebSocket':
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         # Set socket timeout to a short timeout during connection phase, in
-        # case Pebble side times out (5s), so this side doesn't hang. See:
+        # case the Pebble side times out (5s), so this side doesn't hang. See:
         # https://github.com/canonical/operator/issues/1246
         sock.settimeout(self.timeout)
         sock.connect(self.socket_path)


ops/pebble.py
```diff
@@ -2832,10 +2832,17 @@ def _cancel_stdin():
 
     def _connect_websocket(self, task_id: str, websocket_id: str) -> '_WebSocket':
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        # Set socket timeout to a short timeout during connection phase, in
+        # case Pebble side times out (5s), so this side doesn't hang. See:
```
dimaqq, line 2836: The default timeout for this class is 5s. If pebble timeout is also 5s, the net result feels race-y...

I wonder how to test this exhaustively.

One option would be to set up a dummy but programmable websocket listener. What do you reckon, is it worth the effort?

benhoyt:
To try this easily in a charm, you can patch the `pebble.Client._connect_websocket` by including this at the top of `src/charm.py` (or an imported Python file):

```python
from ops import pebble
import socket
import websocket


def _connect_websocket(self, task_id, websocket_id):
    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
    sock.settimeout(self.timeout)
    sock.connect(self.socket_path)
    url = self._websocket_url(task_id, websocket_id)
    ws = websocket.WebSocket(skip_utf8_validation=True)
    ws.connect(url, socket=sock)
    sock.settimeout(None)
    return ws


pebble.Client._connect_websocket = _connect_websocket
```

ops/pebble.py
benhoyt: It's not actually racy, as those two timeouts (while they happen to be the same value) are completely independent: the Pebble timeout happens first, then this 5s timeout is added to that. So the total timeout here before the Python side will fail is 10s.

It's easy to test manually with the instructions given in #1246. I wondered whether it was worth an automated test, but decided it was too hard and awkward. We'd need some hook to patch in an arbitrary delay (like the `time.sleep(5.1)` in the manual test). I decided that because the fix is simple and clear, and it's easy to repro manually, it's probably not worth the hours we'd spend on an automated test for this case.

dimaqq: There's a caveat to this way of injecting the timeout.

Each raw socket operation may take up to 5s, meaning that the total an unsuccessful websocket connection may take would be `5 * (N+1)` where N depends entirely on the websocket-client code and deps.

I figure that this is still way better than not setting a timeout :)

dimaqq: P.S. I have an idea how to set up automated tests for this, do tell if you'd like me to try.

I had done something similar in the past for some other protocol.

On the other hand, we'd be mostly testing our expectations of the websocket-client library, rather than our code.

dimaqq (approved):

tonyandrewmeyer (approved):
Nice!

I could reproduce reliably with the steps in #1246, and running from your branch indeed nicely errors out instead.

PR #1248, opened 2024-06-05
fix: adjust Harness secret behaviour to align with Juju

### LINKED ISSUE #1229: Harness not modelling permission denied for user secrets

tonyandrewmeyer:
Adjust Harness so that the errors match what Juju does in practice (I tested each case - the results are in a table in a comment in #1229).
* When Juju responds with a message including "not found", we raise `SecretNotFoundError`
* When Juju responds with any other message (such as "permission denied"), we raise `ModelError`
* When Juju succeeds and only fails at the completion of the hook, we raise `RuntimeError` - in production, the charm cannot catch this, but we do want to surface this error, so `RuntimeError` seems most approprate

In addition:
* Update the documentation to clarify when secret content is cached (this was a request from the discussion with the data platform team in Madrid)
* Correct the documentation regarding when `ModelError` and when `SecretNotFoundError` will be raised with the secret methods
* Don't clear the `Secret` object local cache of the secret content on `set_contents`. The contents won't change until `refresh` is used, so there's no point forcing the next call to get the content from Juju.

Juju may change the responses so that there is increased consistency (probably via [this bug](https://bugs.launchpad.net/juju/+bug/2067336)) but it seems best if we fix the behaviour to match now, and then if Juju changes in the future, we can update ops then (also deciding at that time how to handle the issue that different Juju versions will have different behaviour).

Fixes #1229.

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -905,9 +905,10 @@ class SecretRemoveEvent(SecretEvent):
     observers have updated to that new revision, this event will be fired to
     inform the secret owner that the old revision can be removed.
 
-    Typically, the charm will call
+    After any required cleanup, the charm should call
     :meth:`event.secret.remove_revision() <ops.Secret.remove_revision>` to
-    remove the now-unused revision.
+    remove the now-unused revision. If the charm does not, then the event will
+    be emitted again, when further revisions are ready for removal.
     """
 
     def __init__(self, handle: 'Handle', id: str, label: Optional[str], revision: int):
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -841,7 +841,10 @@ class SecretEvent(HookEvent):
 
     @property
     def secret(self) -> model.Secret:
-        """The secret instance this event refers to."""
+        """The secret instance this event refers to.
+        
+        Note that the secret content is not retrieved from the secret storage
+        until :meth:`Secret.get_content()` is called."""
         backend = self.framework.model._backend
         return model.Secret(backend=backend, id=self._id, label=self._label)
 
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -842,9 +842,10 @@ class SecretEvent(HookEvent):
     @property
     def secret(self) -> model.Secret:
         """The secret instance this event refers to.
-        
+
         Note that the secret content is not retrieved from the secret storage
-        until :meth:`Secret.get_content()` is called."""
+        until :meth:`Secret.get_content()` is called.
+        """
         backend = self.framework.model._backend
         return model.Secret(backend=backend, id=self._id, label=self._label)
 
--- a/ops/model.py
+++ b/ops/model.py
@@ -1436,9 +1436,8 @@ class Secret:
         if self._id is None:
             self._id = self.get_info().id
         self._backend.secret_set(typing.cast(str, self.id), content=content)
-        # TODO: won't this just get the same content anyway? Unless we refresh
-        # (which ignores the cache) then the tracked revision won't be different.
-        self._content = None  # invalidate cache so it's refetched next get_content()
+        # We do not need to invalidate the cache here, as the content is the
+        # same until `refresh` is used, at which point the cache is invalidated.
 
     def set_info(
         self,
--- a/ops/model.py
+++ b/ops/model.py
@@ -279,8 +279,9 @@ class Model:
             label: Secret label if fetching by label (or updating it).
 
         Raises:
-            SecretNotFoundError: If a secret with this ID or label doesn't exist,
-                or the caller does not have permission to view it.
+            SecretNotFoundError: If a secret with this ID or label doesn't exist.
+            ModelError: if the charm does not have permission to access the
+                secret.
         """
         if not (id or label):
             raise TypeError('Must provide an id or label, or both')
@@ -1381,6 +1382,10 @@ class Secret:
             refresh: If true, fetch the latest revision's content and tell
                 Juju to update to tracking that revision. The default is to
                 get the content of the currently-tracked revision.
+        Raises:
+            SecretNotFoundError: if the secret no longer exists.
+            ModelError: if the charm does not have permission to access the
+                secret.
         """
         if refresh or self._content is None:
             self._content = self._backend.secret_get(id=self.id, label=self.label, refresh=refresh)
@@ -1392,6 +1397,11 @@ class Secret:
         This returns the content of the latest revision without updating the
         tracking. The content is not cached locally, so multiple calls will
         result in multiple queries to the secret storage.
+
+        Raises:
+            SecretNotFoundError: if the secret no longer exists.
+            ModelError: if the charm does not have permission to access the
+                secret.
         """
         return self._backend.secret_get(id=self.id, label=self.label, peek=True)
 
@@ -1399,6 +1409,10 @@ class Secret:
         """Get this secret's information (metadata).
 
         Only secret owners can fetch this information.
+
+        Raises:
+            SecretNotFoundError: if the secret no longer exists, or if the charm
+                does not have permission to access the secret.
         """
         return self._backend.secret_info_get(id=self.id, label=self.label)
 
@@ -1408,6 +1422,10 @@ class Secret:
         This will create a new secret revision, and notify all units tracking
         the secret (the "observers") that a new revision is available with a
         :class:`ops.SecretChangedEvent <ops.charm.SecretChangedEvent>`.
+
+        If the charm does not have permission to update the secret, or the
+        secret no longer exists, this method will succeed, but the unit will go
+        into error state on completion of the current Juju hook.
 
         Args:
             content: A key-value mapping containing the payload of the secret,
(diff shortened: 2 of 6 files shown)


ops/model.py
```diff
@@ -1484,6 +1518,14 @@ def remove_revision(self, revision: int):
         :class:`ops.SecretRemoveEvent <ops.charm.SecretRemoveEvent>` or
         :class:`ops.SecretExpiredEvent <ops.charm.SecretExpiredEvent>`.
 
+        If the charm does not have permission to remove the secret, or it no
+        longer exists, this method will succeed, but the unit will go into error
+        state on completion of the current Juju hook.
+
+        If the tracked revision is removed, then :meth:`get_content` should
+        be called immediately afterwards with ``refresh=True``, to avoid
+        leaving the unit in a 'stuck' state.
```
benhoyt, line 1527: I sorta don't think we need to say this. I think this is really a charm bug if they're going around removing the revision being tracked, and I'm thinking it just adds confusion here.

benhoyt (approved):
Excellent, thanks. One minor comment, though I don't have strong feelings about it, so if you think that paragraph should stay in, go ahead.

ops/model.py
tonyandrewmeyer: My concern is that they don't know if they are doing this. For example, if there's an action "remove this revision" then the charm can't prevent against that being the tracked one, and it's tricky to figure out what's going wrong at that point.

What are the use-cases for `secret-remove` outside of rotation/expiry/remove-all? I read over the Secrets Core spec but couldn't find anything there. If we don't expect it to be used, instead of the above paragraph, could we strengthen the guidance/language, somewhat like:

```python
        """Remove the given secret revision.

        This is normally only called when handling
        :class:`ops.SecretRemoveEvent <ops.charm.SecretRemoveEvent>` or
        :class:`ops.SecretExpiredEvent <ops.charm.SecretExpiredEvent>`.

        Args:
            revision: The secret revision to remove. This should usually be set to
                :attr:`SecretRemoveEvent.revision` or :attr:`SecretExpiredEvent.revision`.
```

benhoyt: Yeah, I think it's a good idea to add that. I can't see any reason why you'd `remove_revision` outside of those contexts. For example, why would there be an action "remove this revision" in the first place? That seems like a big charm code smell.

benhoyt (approved):

tonyandrewmeyer:
@IronCore864 would you mind reviewing this when you have time? Thanks!

ops/charm.py
```diff
@@ -901,9 +905,10 @@ class SecretRemoveEvent(SecretEvent):
     observers have updated to that new revision, this event will be fired to
     inform the secret owner that the old revision can be removed.
 
-    Typically, the charm will call
+    After any required cleanup, the charm should call
     :meth:`event.secret.remove_revision() <ops.Secret.remove_revision>` to
-    remove the now-unused revision.
+    remove the now-unused revision. If the charm does not, then the event will
+    be emitted again, when further revisions are ready for removal.
```
dimaqq, line 911: Does that result in N^2 event pile-up if the charm is stubborn?

tonyandrewmeyer: If the charm never actually does the remove then they'll get *n* events every time a new revision is obsolete, where *n* is the number of obsolete revisions. So for the secret owner it's more of an N pile-up than N^2 I would say. Still not great, especially if you move to new revisions frequently.

dimaqq: I guess I meant N^2 total runtime complexity, got it.

Anyway, that's not an issue with the PR, rather with the general secret revision protocol.

ops/testing.py
```diff
@@ -2884,7 +2890,8 @@ def secret_revoke(self, id: str, relation_id: int, *, unit: Optional[str] = None
 
     def secret_remove(self, id: str, *, revision: Optional[int] = None) -> None:
         secret = self._ensure_secret(id)
-        self._ensure_secret_owner(secret)
+        if not self._has_secret_owner_permission(secret):
+            raise RuntimeError(f'You must own secret {secret.id!r} to perform this operation')
```
dimaqq, line 2894: I'm not a big fan of [ab]using RuntimeError, but if that's what happens in `ops`, then Harness should raise exactly same, as as similar exception as it can.

dimaqq: P.S. a non-breaking change would be to have ops raise and exception that inherits from RuntimeError because it's documented as "Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong."

dimaqq (approved):

ops/testing.py
benhoyt: Side note, just for the record, I'm not a fan of how Ops (ab)uses `RuntimeError` either -- have often wondered why and thought that was probably a bad idea. Usually (I think?) it refers to the *Python* runtime.

dimaqq: I've asked the general public for clarification at https://discuss.python.org/t/meaning-of-runtimeerror/56206

As I used to think that this exception type meant "big error with [Python] runtime", while some uses, like here in ops assume that it means "stuff that happens at run time" (perhaps).

tonyandrewmeyer: I think it's fine for user code to raise `RuntimeError` (and [I'm not alone](https://github.com/search?q=%22raise+RuntimeError%22+language%3APython+&type=code)), and I think the docs would explicitly say if it wasn't meant to be used in this way. I think it's "something has gone wrong and I haven't made a custom exception class for this thing", and that it implies that generally it should not be caught. ISTM that's what the docs is saying, although I agree with Dima's post that it's a bit quippy.

> I'm not a big fan of [ab]using RuntimeError, but if that's what happens in `ops`, then Harness should raise exactly same, as as similar exception as it can.

This is not what happens in ops. What happens in ops is that everything is fine, there's no exception, and then when ops has finished and Juju is wrapping up the event, it complains and puts the unit in an error state. Harness doesn't have a way to model that, and I think we definitely do want to raise some sort of error rather than let it pass without complaint just because the error when running in Juju happens after ops.main finishes.

That said, I have no objection to changing this so that there's a (name to be decided) `JujuError` that gets raised instead. It's used a *lot* not just in Harness but in ops, but if our preference is to move away from that, then I think it's fine to make this specific and leave everything else alone (the raising here is new, and also not meant to be caught, so backwards compatibility isn't an issue).

IronCore864 (approved):

ops/testing.py
tonyandrewmeyer: Charm-Tech discussed this today, and we decided that we would prefer to have custom exception classes rather than `RuntimeError`, but that we would prefer to do that universally rather than incrementally, and that it's a low-priority improvement for now, so can be dealt with another day.

For this PR, we're ok with keeping `RuntimeError`.

tonyandrewmeyer: Charm-Tech discussed this today, and we decided that we would prefer to have custom exception classes rather than `RuntimeError`, but that we would prefer to do that universally rather than incrementally, and that it's a low-priority improvement for now, so can be dealt with another day.

For this PR, we're ok with keeping `RuntimeError`.

PR #1251, opened 2024-06-07
feat: make the status classes dataclasses

tonyandrewmeyer:
Converts `StatusBase` and child classes to dataclasses.

The only user-facing changes are:
* Methods that only work with dataclasses, like `dataclasses.asdict` will work with statuses.
* Statuses can be used in pattern matching (a `__match_args__` is added to the classes), in Python versions that have pattern matching. This is the default dataclass behaviour, and seems a reasonable addition, and is awkward to disable while keeping 3.8 compatibility.

The primary motivation is that this means that everything in a Scenario State can be a dataclass or an object that can be serialised with the default JSONEncoder, so Scenario.State can use actual ops status classes rather than the awkward `_EntityStatus` class, and retain the ability to do `json.dumps(dataclasses.asdict(state))`.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1809,7 +1809,7 @@ class StatusBase:
     directly use the child class such as :class:`ActiveStatus` to indicate their status.
     """
 
-    _statuses: Dict[str, Type['StatusBase']] = dataclasses.field(default_factory=dict)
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
Discussed this with the rest of @charm-tech and decided against doing this now. At some point in the (probably distant) future we might convert existing ops classes to dataclasses, but we won't do that now.

PR #1252, opened 2024-06-07
docs: add a section to HACKING.md on PR titles (commit messages to main)

tonyandrewmeyer:
Explain that we use conventional commits, and list the types and scopes that we're starting with.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -177,10 +177,8 @@ Pull requests should have a short title that follows the
 * revert
 * test
 
-If the PR is limited to one of the following scopes, also include the scope in the title,
-and otherwise leave off the scope.
-
-* harness
+If the PR is limited to changes in ops.testing (Harness), also include the scope
+`(harness)` in the title. At present, we do not add a scope in any other cases.
 
 For example:
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -160,6 +160,38 @@ your charm to a controller using that version of Juju. For example, with microk8
 We rely on automation to [update charm pins](.github/actions/update-charm-pins/) of
 a bunch of charms that use the operator framework. The script can be run locally too.
 
+# Contributing
+
+Changes are proposed as [pull requests on GitHub](https://github.com/canonical/operator/pulls).
+
+Pull requests should have a short title that follows the
+[conventional commit style](https://www.conventionalcommits.org/en/) using one of these types::
+
+* chore
+* ci
+* docs
+* feat
+* fix
+* perf
+* refactor
+* revert
+* test
+
+If the PR is limited to one of the following scopes, also include the scope in the title,
+and otherwise leave off the scope.
+
+* harness
+
+For example::
+
+* feat: add the ability to observe change-updated events
+* fix!: correct the type hinting for config data
+* docs(harness): clarify the types of exceptions that Harness.add_user_secret may raise
+
+Note that the commit messages to the PR's branch do not need to follow the
+conventional commit format, as these will be squashed into a single commit to `main`
+using the PR title as the commit message.
+
 # Documentation
 
 In general, new functionality
--- a/HACKING.md
+++ b/HACKING.md
@@ -165,7 +165,7 @@ a bunch of charms that use the operator framework. The script can be run locally
 Changes are proposed as [pull requests on GitHub](https://github.com/canonical/operator/pulls).
 
 Pull requests should have a short title that follows the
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -160,6 +160,38 @@ your charm to a controller using that version of Juju. For example, with microk8
...
+Pull requests should have a short title that follows the
+[conventional commit style](https://www.conventionalcommits.org/en/) using one of these types:
+
+* chore
+* ci
+* docs
+* feat
+* fix
+* perf
+* refactor
```
benhoyt, line 176: I think "refactor" is a "chore". Thoughts on getting rid of it? Or maybe we should go the other way -- what chores are there that *aren't* refactoring?

Also, what about dropping "perf"? I kinda think we'll almost never need this in Ops. But maybe it doesn't hurt, even as a signal that we are not unconcerned about perf.

tonyandrewmeyer: For the record, I dropped "build" and "style" from the [Angular list that conventionalcommits.com recommends](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type).

If we get rid of both perf I would put a change that's purely performance improvements under refactor, so I'm not sure where that would go if we get rid of both. It doesn't seem like a chore, and maybe that is going too... I like keeping it, even if seldom-used, as a signal as you describe.

Other chores: dependency bumps, the release PRs. Looking back over the last few months, there are some that are "internal tooling" (they could probably have been categorised into one of the others). I'm +0 for keeping refactor and chore.

benhoyt: Okay, sounds fine with me -- let's keep 'em. Being a little more specific doesn't hurt. As long as this list doesn't get too huge.

benhoyt (approved):

HACKING.md
```diff
@@ -160,6 +160,38 @@ your charm to a controller using that version of Juju. For example, with microk8
...
+* revert
+* test
+
+If the PR is limited to one of the following scopes, also include the scope in the title,
+and otherwise leave off the scope.
+
+* harness
+
```
dimaqq, line 184: Was there perhaps an intention to add more scope examples in this doc?
Otherwise "one of the following scopes" reads odd.

dimaqq (approved):

HACKING.md
tonyandrewmeyer: The intention (if I recall the discussion correctly) was that we would potentially add more scopes in the future, as we find something that would be useful. So I wrote it to easily accommodate that. I could rewrite to just focus specifically on `(harness)` for now if that would be better.

benhoyt: Yeah, I wondered that too. I kind of agree with Dima -- would probably be better to have it read well for what it is now (just one scope) and update it when we add more. Or note explicitly that "at present there's only one" or something.

HACKING.md
```diff
@@ -160,6 +160,36 @@ your charm to a controller using that version of Juju. For example, with microk8
...
+# Contributing
+
+Changes are proposed as [pull requests on GitHub](https://github.com/canonical/operator/pulls).
+
+Pull requests should have a short title that follows the
+[conventional commit style](https://www.conventionalcommits.org/en/) using one of these types:
+
+* chore
+* ci
+* docs
```
dimaqq, line 172: 🤔 I tend to open PRs with `docs:` prefix and Tony with `doc:` prefix.

Should we settle on a single type or allow both?

benhoyt: Yes, let's go with `docs`, as that's what https://www.conventionalcommits.org/en/ recommends.

benhoyt:
@tonyandrewmeyer Is this good to merge?

HACKING.md
tonyandrewmeyer: Yeah, I've used `docs:` except I think just once, which was a typo on my part :disappointed:. This is where automation (in a follow-up PR, at some point in the future) would indeed be useful, to catch such errors.

tonyandrewmeyer:
@IronCore864 are you ok with this as well? We've got two approvals (plus the implicit one from me as author) but it would be nice to have the whole team approve pre-merging since this is about team process rather than a specific feature.

IronCore864 (approved):

IronCore864:
> @IronCore864 are you ok with this as well? We've got two approvals (plus the implicit one from me as author) but it would be nice to have the whole team approve pre-merging since this is about team process rather than a specific feature.

I'm completely OK with this. I like it that we keep "refactor", and only use scope for harness.

PR #1253, opened 2024-06-07
test: disable pytest warnings plugin

IronCore864:
As discussed in the daily, this PR disables the pytest warnings plugin, so that the ops.log capture warnings' unit test can pass. See discussions [here](https://github.com/canonical/operator/pull/1240) for more detail.

Tested, doesn't affect existing test cases:

```bash
TOTAL                       6508    499   2559    220    90%
  unit: OK (231.42=setup[0.02]+cmd[230.74,0.66] seconds)
  congratulations :) (231.45 seconds)
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -75,7 +75,7 @@ deps =
     typing_extensions~=4.2
 commands =
     coverage run --source={[vars]src_path} \
-             -m pytest --ignore={[vars]tst_path}smoke -v --tb native {posargs}
+             -m pytest -p no:warnings --ignore={[vars]tst_path}smoke -v --tb native {posargs}
     coverage report
 
 [testenv:pebble]


tonyandrewmeyer:
I don't think we should do this. My concern is that there could be some warnings in the future that we want to pay attention to - for example, deprecation warnings in websocket or pyyaml - and they are pretty noticeable with the default pytest behaviour (which is the point, after all!) and this would hide that.

I think we can work around the need to test the ops.log warning behaviour in other ways (I've added a couple of suggestions in that PR), instead.

benhoyt:
Thanks for investigating, @tonyandrewmeyer. Fully agreed -- I'd rather not turn this off.

dimaqq:
I've tried this in #1258 and actually it doesn't help. So I'm also -1 on this.

IronCore864:
Although this works for me, it's not intuitive to disable a plugin that is enabled by default, doesn't seem right. After discussion in the daily, we decided to not do it in this way, but to test warning captures in test_main. Closing now.

PR #1258, opened 2024-06-11
feat: include warnings in logs sent to juju

dimaqq:
Alternative PR to #1077 
- same ops change
- clearer requirements for charm code
- figured out why repeated tests failed in #1077 n worked around that
- open question about tooling we'd want to use

### DIFF
--- a/ops/log.py
+++ b/ops/log.py
@@ -53,6 +53,7 @@ def setup_root_logging(
         debug: if True, write logs to stderr as well as to juju-log.
         exc_stderr: if True, write uncaught exceptions to stderr as well as to juju-log.
     """
+    logging.captureWarnings(True)
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
--- a/test/test_log_warning.py
+++ b/test/test_log_warning.py
@@ -17,12 +17,12 @@ import logging
 import sys
 import warnings
 from typing import Any, Callable, Type
-from unittest.mock import Mock, ANY
+from unittest.mock import ANY, Mock
+
+import pytest
 
 from ops.log import setup_root_logging
 from ops.model import _ModelBackend
-
-import pytest
 
 
 def deprecated(msg: str, *, category: Type[Warning] = DeprecationWarning, stacklevel: int = 1):
@@ -62,9 +62,9 @@ def reset_warnings():
 
 
 def test_fixme(monkeypatch: Any, reset_warnings: Any):
-    monkeypatch.setenv("JUJU_UNIT_NAME", "dummy")
+    monkeypatch.setenv('JUJU_UNIT_NAME', 'dummy')
     # I'd rather use pytest-subprocess instead of this crude mock
-    monkeypatch.setattr(_ModelBackend, "_run", (run_mock := Mock()))
+    monkeypatch.setattr(_ModelBackend, '_run', (run_mock := Mock()))
     logging.basicConfig()
     setup_root_logging(_ModelBackend())
     dummy_charm_function()
@@ -72,9 +72,9 @@ def test_fixme(monkeypatch: Any, reset_warnings: Any):
 
 
 def test_another(monkeypatch: Any, reset_warnings: Any):
-    monkeypatch.setenv("JUJU_UNIT_NAME", "dummy")
+    monkeypatch.setenv('JUJU_UNIT_NAME', 'dummy')
     # I'd rather use pytest-subprocess instead of this crude mock
-    monkeypatch.setattr(_ModelBackend, "_run", (run_mock := Mock()))
+    monkeypatch.setattr(_ModelBackend, '_run', (run_mock := Mock()))
     logging.basicConfig()
     setup_root_logging(_ModelBackend())
     dummy_charm_function()
--- a/test/test_log_warning.py
+++ b/test/test_log_warning.py
@@ -1,0 +1,81 @@
+# Copyright 2020 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+import logging
+import sys
+import warnings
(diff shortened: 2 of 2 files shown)


ops/log.py
```diff
@@ -53,6 +53,7 @@ def setup_root_logging(
         debug: if True, write logs to stderr as well as to juju-log.
         exc_stderr: if True, write uncaught exceptions to stderr as well as to juju-log.
     """
+    logging.captureWarnings(True)
```
tonyandrewmeyer, line 56: I'm still of the opinion that we should have a custom formatter. There's so much that goes into the debug-log - I think we need to take care to try not to make things worse. I don't see the value in repeating the part of the code that triggered the warning - either the charm is open-source, in which case the code can be found fairly easily<sup>*</sup> via the file and line number, or it's not (is this ever the case?) and we shouldn't be leaking the code.

In either situation, my expectation would be that the charm should provide in the message instructions for the Juju admin (use a different action name/param, upgrade a related charm, etc). It seems doubtful that having the line of code will help with that. (Even the file and line don't seem helpful, except perhaps if you're communicating with the charm author and they haven't made it easy to describe).

As discussed in the other PR, I think the full filename is more than what's required, but also that trying to properly trim off the parts we don't care about is maybe too much to do right now.

<sup>*</sup> To be fair, it's harder than it should be to find which version of the charm is running, much of the time (charms could solve this by adding the VERSION file, but most don't, and there are documentation issues there). So file+line number is ideally combined with revision, which then gets used to (hopefully) look up a git tag.

test/test_log_warning.py
```diff
@@ -0,0 +1,81 @@
+# Copyright 2020 Canonical Ltd.
```
tonyandrewmeyer, line 1: Nit: it's 2024 :smile:

More importantly, do we want a new test module for this? It seems like a big add for a small feature, and test_log is pretty small (73 lines including whitespace, docs, and comments). I think there's an argument to be made about potentially splitting up some of the very large test files, even though it breaks the symmetry, but I don't think we would start with this one.

test/test_log_warning.py
```diff
@@ -0,0 +1,81 @@
...
+            warnings.warn(category(msg), stacklevel=2)
+            return f(*args, **kwargs)
+
+        return inner
+
+    return decorator
+
+
+if sys.version_info >= (3, 13):
+    from warnings import deprecated  # type: ignore
```
tonyandrewmeyer, line 41: This feels unPythonic to me, but I'd be interested in hearing counter-arguments.

In my experience, this is typically done like:

```python
try:
    from warnings import deprecated
except ImportError:
    # Handle Python < 3.13.
    # backport goes here
```

Even when it's not an import, e.g.:

```python
try:
    True, False
except NameError:
    True, False = 1, 0  # Handle Python < 2.2.2
```

dimaqq: try/except makes `pyright` evaluate the type of `deprecated`, which it cannot until we've got a Py 3.13 runtime (as in all devs + ci), and I couldn't find a good way to suppress this type interference, other than to hide the statement behind an `if` clause.

test/test_log_warning.py
dimaqq: Thanks, clearly I've copy-pasted this from another file...

Well, my thinking was that we've got some rather long test files (though `test_log.py` is on the shorter end of the spectrum).

I guess I preferred to draft this idea in isolation, and the content can be merged into `test_log.py`.

In the current form, maybe it's easier to review, as I'm using slightly different kind of mocking. I'm open to opinions on this subject 🙇🏻

test/test_log_warning.py
tonyandrewmeyer: Hmm, interesting. My thinking would be Pythonic+type:ignore > unPythonic+check-version. However, I could well be biased by my general distaste for typing in Python :smile:. I'm curious what the consensus here would be.

tonyandrewmeyer:
Oh, and the same comment I had on the other PR - if we forge ahead with this one, then there should be some documentation in the `setup_root_logging` docstring. We don't have any other mention of capturing logging in the API reference docs (at least, I don't recall any and searching for "debug-log" and "debug log" finds nothing), so that's probably it, but we should definitely make sure it's mentioned elsewhere (after whichever PR goes through).

test/test_log_warning.py
```diff
@@ -0,0 +1,81 @@
...
+
+def dummy_charm_function():
+    deprecated_library_function()
+
+
+@pytest.fixture
+def reset_warnings():
+    del logging.root.handlers[:]
+    logging.captureWarnings(False)
+    warnings._onceregistry.clear()  # type: ignore
```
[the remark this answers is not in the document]
dimaqq: DeprecationWarning should be issued only once, in which case, resetting or mocking the internal `warnings` storage should be needed.

However, in this case, this line doesn't actually do anything and I don't yet understand why.

Specific warnings issued contain the line number of the call site in the test file, which make the warnings unique, but also the storage is actually empty, so 🤷🏻

test/test_log_warning.py
```diff
@@ -0,0 +1,81 @@
...
+    return decorator
+
+
+if sys.version_info >= (3, 13):
+    from warnings import deprecated  # type: ignore
+
+
+@deprecated('Use new_library_function instead')
+def deprecated_library_function():
+    pass
```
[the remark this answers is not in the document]
dimaqq: My intention was to simulate some "ideal" library code that real charm may use that generates a warning.

With the `warnings.deprecated` coming in Py 3.13, I figured that could be a good fit.

That being said, indeed, there should be no real difference between a decorator and an explicit call to `warnings.warn()`.

dimaqq:
Based on the discussion, I feel that `test_main.py` approach as outlined in https://github.com/canonical/operator/pull/1240#discussion_r1634320963 is in fact cleaner.

This PR can be closed.

Keeping it around for a little bit in case something useful can be taken from this code, and will close later.

ops/log.py
dimaqq: I'm ambivalent on custom formatter.

Standard format means lesser mental load on library user; custom format is more concise, thus also lesser mental load.

Re "leaking" source code lines. If we went that way, we'd have to whiten the traceback in `ops/log.py: except_hook` too. Somehow I don't see us doing that at the moment.

test/test_log_warning.py
dimaqq: That's a discussion I'd like to have, because I could as well adopt the output of it when deprecating the framework methods that rely on AllWatcher.

dimaqq: P.S. I'd love to use:

```
try:
    from warnings import deprecated
except ImportError:
    from .utils.backport import deprecated
```

I just couldn't get the type checker happy.
The stdlib decorator is already included in typeshed, so maybe there's soon a way to resolve the types and then there's no need for discussion 🎉

PR #1259, opened 2024-06-12
fix: properly clean up after running setup_root_logging in test_log

tonyandrewmeyer:
Calling `setup_root_logging` changes `sys.excepthook`. The tests should undo this in the same way that they clear out the logging handlers that `setup_root_logging` adds.

### DIFF
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -15,6 +15,7 @@
 import io
 import logging
 import re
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Thanks

dimaqq (approved):

PR #1260, opened 2024-06-12
chore: make dependabot and update-charm-tests only happen monthly

benhoyt:
Unfortunately dependabot.yaml doesn't support saying "25th of each month", just "monthly" which is hard-coded to the 1st. But that's probably okay for that one.


tonyandrewmeyer (approved):
Thanks!

dimaqq (approved):

PR #1262, opened 2024-06-14
ci: validate PR title against conventional commit rules

dimaqq:
As documented in #1252

- types taken from the doc PR
- scopes were scraped from project PR history

### DIFF
--- a/.github/workflows/semantic-pr-check.yaml
+++ b/.github/workflows/semantic-pr-check.yaml
@@ -1,0 +1,40 @@
+---
+name: "Semantic PR Title Check"
+# Ensure that the PR title conforms to the Conventional Commits and our choice of types and scopes, so that library version bumps can be detected automatically
+
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: |
+            chore
+            ci
+            docs
+            feat
+            fix
+            perf
+            refactor
+            revert
+            test
+          scopes: |
+            harness
+            deps
+            testing
+            pebble
+            model
+            framework
+            typing
+            charm
+          requireScope: false
--- a/.github/workflows/validate-pr-title.yaml
+++ b/.github/workflows/validate-pr-title.yaml
@@ -34,5 +34,4 @@ jobs:
             test
           scopes: |
             harness
-            [a-z]{3,20}
           requireScope: false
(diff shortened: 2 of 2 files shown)


.github/workflows/semantic-pr-check.yaml
```diff
@@ -0,0 +1,38 @@
...
+          types: |
+            chore
+            ci
+            docs
+            feat
+            fix
+            perf
+            refactor
+          scopes: |
+            harness
```
benhoyt, line 30: We're just going to allow `harness` for now though, right? Per discussion of https://github.com/canonical/operator/pull/1252/files#diff-9769a3f5504af3cd735e08f078b94f827bdd51eacac2e9d80ac9761bc485171fR181

.github/workflows/semantic-pr-check.yaml
```diff
@@ -0,0 +1,38 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: |
+            chore
+            ci
+            docs
+            feat
+            fix
+            perf
+            refactor
```
benhoyt, line 28: Per https://github.com/canonical/operator/pull/1252 I think we want to add `revert` and `test`.

benhoyt (changes requested):
Nice idea, thanks. Couple of changes suggested to the lists, based on #1252.

.github/workflows/semantic-pr-check.yaml
dimaqq: Thanks, I've missed that.

.github/workflows/semantic-pr-check.yaml
dimaqq: Let me merge this in my fork to see what the meaning of `scopes: [a,b,c]; requireScope: false` is, that is if arbitrary, unlisted scopes are allowed.

dimaqq: done, added a sane catch-all

dimaqq:
Validated that this works: https://github.com/dimaqq/operator/pull/28

.github/workflows/validate-pr-title.yaml
```diff
@@ -0,0 +1,38 @@
...
+            docs
+            feat
+            fix
+            perf
+            refactor
+            revert
+            test
+          scopes: |
+            harness
+            [a-z]{3,20}
```
benhoyt, line 37: Shouldn't we prevent any other scope than `harness`? The PR title should leave it off in that case.

dimaqq: Let's chat at next standup.
I thought we'd want to allow free-form scope, for:
- outstanding #1148 
- upcoming work like `feat(notice)`
- hypothetical `feat(immutable)` work

I'm happy to go with whatever team decides, this is meant to be pragmatic after all.

dimaqq: Decided at the stand-up: remove free-form scope.

benhoyt (approved):
Thanks!

tonyandrewmeyer (approved):
Thanks!

Looks good to me. I did not look at the upstream source to verify that it's not doing anything bad (particularly with the token we are providing) - I'm not super well placed to do that in JS. The upstream repo seems fairly healthy, which is a positive indicator.

PR #1266, opened 2024-06-21
docs: add a basic security policy

tonyandrewmeyer:
Adds a simple security policy, so that users can easily find out how to privately report security issues.

The policy states that 2.x will get security updates, which seems reasonable to me, but we could make that more recent versions if that was better.

The policy offers reporting via GitHub (which would need to be turned on) and to the security@ubuntu.com address - I think it's important to still offer an email (particularly encrypted email) mechanism, not just the GitHub one.

This is based on the [LXD policy](https://github.com/canonical/lxd/blob/main/SECURITY.md), and the [work to develop a Canonical security policy template](https://warthogs.atlassian.net/browse/SEC-4238) (internal link only, sorry).

See also [this Mattermost discussion](https://chat.canonical.com/canonical/pl/gnk4rsorrpgr3yetka9suh5bpa) (also internal only, sorry).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,7 +11,7 @@ create the issue, affected versions, and, if known, mitigations for
 the issue.
 
 The easiest way to report a security issue is through
-[GitHub](https://github.com/canonical/operator/security/advisories/new). See
+[GitHub's security advisory for this project](https://github.com/canonical/operator/security/advisories/new). See
 [Privately reporting a security
 vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
 for instructions.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,12 +2,7 @@
 
 ## Supported versions
 
-All ops 2.x versions are currently supported with security updates.
-
-| Version | Supported |
-| - | - |
-| 2.x.y | :white_check_mark: |
-| < 2.0 | :x: |
+All ops 2.x versions released within the last year are currently supported with security updates.
 
 ## Reporting a vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -23,8 +23,8 @@ for instructions.
 
 The ops GitHub admins will be notified of the issue and will work with you
 to determine whether the issue qualifies as a security issue and, if so, in
-which component. We will then handle figuring out a fix, getting a CVE
-assigned and coordinating the release of the fix.
+which component. We will then figure out a fix, get a CVE
+assigned, and coordinate the release of the fix.
 
 You may also send email to security@ubuntu.com. Email may optionally be
 encrypted to OpenPGP key
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,7 +14,7 @@ The easiest way to report a security issue is through
 [GitHub's security advisory for this project](https://github.com/canonical/operator/security/advisories/new). See
 [Privately reporting a security
 vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
-for instructions.
+for instructions on reporting using GitHub's security advisory feature.
 
 The ops GitHub admins will be notified of the issue and will work with you
 to determine whether the issue qualifies as a security issue and, if so, in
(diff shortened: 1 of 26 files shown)


tonyandrewmeyer:
@lucistanescu and @setharnold would you mind having a look over this?

setharnold:
I like it, it could certainly work as-is.

It might be worth trying to say what inputs are completely trusted or which inputs are untrusted, to try to scale expectations of how it could and should be used, but this is already a good start.

Thanks

SECURITY.md
```diff
@@ -0,0 +1,41 @@
...
+The easiest way to report a security issue is through
+[GitHub](https://github.com/canonical/operator/security/advisories/new). See
+[Privately reporting a security
+vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
+
+The ops GitHub admins will be notified of the issue and will work with you
+to determine whether the issue qualifies as a security issue and, if so, in
+which component. We will then handle figuring out a fix, getting a CVE
+assigned and coordinating the release of the fix.
```
benhoyt, line 27: More active:

```suggestion
which component. We will then figure out a fix, get a CVE
assigned, and coordinate the release of the fix.
```

benhoyt (approved):
I don't really know what we're aiming for here (security team's input should help their), but this looks like a good start to me. One micro-nit comment.

tonyandrewmeyer:
> I don't really know what we're aiming for here

Personally, just to have an easy way for people to figure out how to report security issues, without having to go through the "this is from Canonical, I should look for a Canonical system" thought path.

@benhoyt are you ok with saying that 2.x is supported for security fixes? It's < 2 years ago, so my thinking was that if there was something that needed to be addressed it would likely not be too difficult to port it back there, and then do 13+ point releases (and yank the bad releases). A non-trivial amount of work, but not huge, and given the likelihood it seemed ok to me. If you'd rather be more aggressive, maybe we could make it one year, or something like that? The downside of that approach is needing to update the file each month if we're going to explicitly list versions.

benhoyt:
> are you ok with saying that 2.x is supported for security fixes?

My first thought was that that's fine and seems quite reasonable to me. However, then I realized that this would mean doing actual patch releases for all the minor releases we've done, which is non-ideal. Then again, also pretty unlikely, as we push people to use the latest Ops.

I'm okay with it, but I do like putting a time-cap on it. To avoid constantly updating the text, we could just say "2.x versions in the last 12 months" or something.

SECURITY.md
```diff
@@ -0,0 +1,41 @@
+# Security policy
+
+## Supported versions
+
+All ops 2.x versions are currently supported with security updates.
+
+| Version | Supported |
+| - | - |
+| 2.x.y | :white_check_mark: |
+| < 2.0 | :x: |
```
tonyandrewmeyer, line 10:
```suggestion
All ops 2.x versions released within the last year are currently supported with security updates.
```

dimaqq (commented):
Sorry I forgot to click Submit earlier...

SECURITY.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Reporting a vulnerability
+
+Please provide a description of the issue, the steps you took to
+create the issue, affected versions, and, if known, mitigations for
+the issue.
+
+The easiest way to report a security issue is through
+[GitHub](https://github.com/canonical/operator/security/advisories/new). See
```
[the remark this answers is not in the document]
tonyandrewmeyer: What about:

```suggestion
[GitHub's security advisory for this project](https://github.com/canonical/operator/security/advisories/new). See
```

SECURITY.md
```diff
@@ -0,0 +1,36 @@
...
+[Privately reporting a security
+vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
```
tonyandrewmeyer, line 17: What about this?

```suggestion
for instructions on reporting using GitHub's security advisory feature.
```

dimaqq: I guess that also works.

I suppose a URL would also work instead of a link -- it's pretty clear what it is and where it leads.

dimaqq (approved):

PR #1268, opened 2024-06-21
fix: add checks and log_targets to ops.testing

### LINKED ISSUE #1112: Combine Pebble layers in testing get_plan; merge "checks" and "log-targets"

amandahla:
Add checks and log_targets to ops.testing add_layer so it is possible to test it via Harness.

Fixes #1112

### DIFF
--- a/.github/actions/update-charm-pins/main.py
+++ b/.github/actions/update-charm-pins/main.py
@@ -9,14 +9,14 @@ import sys
 from httpx import Client
 from ruamel.yaml import YAML
 
-yaml = YAML(typ="rt")
+yaml = YAML(typ='rt')
 yaml.indent(mapping=2, sequence=4, offset=2)
 
 github = Client(
-    base_url="https://api.github.com/repos",
+    base_url='https://api.github.com/repos',
     headers={
-        "Authorization": f'token {os.getenv("GITHUB_TOKEN")}',
-        "Accept": "application/vnd.github.v3+json",
+        'Authorization': f'token {os.getenv("GITHUB_TOKEN")}',
+        'Accept': 'application/vnd.github.v3+json',
     },
 )
 
@@ -27,29 +27,27 @@ def update_charm_pins(workflow):
         doc = yaml.load(file)
 
     # Assume the workflow has a single job or the first job is parameterized with charm repos
-    job_name = next(iter(doc["jobs"]))
+    job_name = next(iter(doc['jobs']))
 
-    for idx, item in enumerate(doc["jobs"][job_name]["strategy"]["matrix"]["include"]):
-        charm_repo = item["charm-repo"]
-        commit = github.get(f"{charm_repo}/commits").raise_for_status().json()[0]
-        data = github.get(f"{charm_repo}/tags").raise_for_status().json()
-        comment = " ".join(
-            [tag["name"] for tag in data if tag["commit"]["sha"] == commit["sha"]]
-            + [commit["commit"]["committer"]["date"]]
+    for idx, item in enumerate(doc['jobs'][job_name]['strategy']['matrix']['include']):
+        charm_repo = item['charm-repo']
+        commit = github.get(f'{charm_repo}/commits').raise_for_status().json()[0]
+        data = github.get(f'{charm_repo}/tags').raise_for_status().json()
+        comment = ' '.join(
+            [tag['name'] for tag in data if tag['commit']['sha'] == commit['sha']]
+            + [commit['commit']['committer']['date']]
         )
 
         # A YAML node, as opposed to a plain value, can be updated in place to tweak comments
-        node = doc.mlget(
-            ["jobs", job_name, "strategy", "matrix", "include", idx], list_ok=True
-        )
-        node["commit"] = commit["sha"]
-        node.yaml_add_eol_comment(comment, key="commit")
+        node = doc.mlget(['jobs', job_name, 'strategy', 'matrix', 'include', idx], list_ok=True)
+        node['commit'] = commit['sha']
+        node.yaml_add_eol_comment(comment, key='commit')
 
-    with open(workflow, "w") as file:
+    with open(workflow, 'w') as file:
         yaml.dump(doc, file)
 
 
-if __name__ == "__main__":
-    logging.basicConfig(level="INFO")
-    for workflow in " ".join(sys.argv[1:]).split():
+if __name__ == '__main__':
+    logging.basicConfig(level='INFO')
+    for workflow in ' '.join(sys.argv[1:]).split():
         update_charm_pins(workflow)
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1172,8 +1172,7 @@ class Check:
         for name, value in other.__dict__.items():
             if not value:
                 continue
-            else:
-                setattr(self.tcp, name, value)
+            setattr(self.tcp, name, value)
 
     def _merge(self, other: 'Check'):
         """Merges this check object with another check definition.
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1247,6 +1247,22 @@ class LogTarget:
         else:
             return NotImplemented
 
+    def _merge(self, other: 'LogTarget'):
+        """Merges this log target object with another log target definition.
+
+        For attributes present in both objects, the passed in log target
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value or name == 'name':
+                continue
+            if name == 'labels':
+                getattr(self, name).update(value)
+            elif name == 'services':
+                getattr(self, name).extend(value)
+            else:
+                setattr(self, name, value)
+
 
 class FileType(enum.Enum):
     """Enum of file types."""
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1135,6 +1135,64 @@ class Check:
         else:
             return NotImplemented
 
+    def _merge_exec(self, other: 'ExecDict') -> None:
+        """Merges this exec object with another exec definition.
+
+        For attributes present in both objects, the passed in exec
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value:
+                continue
+            if name == 'environment':
+                getattr(self.exec, name).update(value)
+            else:
+                setattr(self.exec, name, value)
+
+    def _merge_http(self, other: 'HttpDict') -> None:
+        """Merges this http object with another http definition.
+
+        For attributes present in both objects, the passed in http
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value:
+                continue
+            if name == 'headers':
+                getattr(self.http, name).update(value)
+            else:
+                setattr(self.http, name, value)
+
+    def _merge_tcp(self, other: 'TcpDict') -> None:
+        """Merges this tcp object with another tcp definition.
+
+        For attributes present in both objects, the passed in tcp
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value:
+                continue
+            else:
+                setattr(self.tcp, name, value)
+
+    def _merge(self, other: 'Check'):
+        """Merges this check object with another check definition.
+
+        For attributes present in both objects, the passed in check
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value or name == 'name':
+                continue
+            if name == 'http':
+                self._merge_http(value)
+            elif name == 'tcp':
+                self._merge_tcp(value)
+            elif name == 'exec':
+                self._merge_exec(value)
+            else:
+                setattr(self, name, value)
+
 
 class CheckLevel(enum.Enum):
     """Enum of check levels."""
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1148,7 +1148,7 @@ class Check:
                 continue
             if name == 'environment':
                 current = self.exec.get(name, {})
-                current.update(value)
+                current.update(value)  # type: ignore
                 self.exec[name] = current
             else:
                 self.exec[name] = value
@@ -1166,7 +1166,7 @@ class Check:
                 continue
             if name == 'headers':
                 current = self.http.get(name, {})
-                current.update(value)
+                current.update(value)  # type: ignore
                 self.http[name] = current
             else:
                 self.http[name] = value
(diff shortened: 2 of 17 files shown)


ops/pebble.py
```diff
@@ -1135,6 +1135,18 @@ def __eq__(self, other: Union['CheckDict', 'Check']) -> bool:
...
+    def _merge(self, other: 'Check'):
+        """Merges this check object with another check definition.
+
+        For attributes present in both objects, the passed in check
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value or name == 'name':
+                continue
+            else:
```
benhoyt, line 1147: We need to consider the "http", "tcp", and "exec" fields and merge those too. Something like this:

```python
if not value or name == 'name':
    continue
if name == 'http':
    self._merge_http(value)
elif name == 'tcp':
    self._merge_tcp(value)
elif name == 'exec':
    self._merge_exec(value)
else:
    setattr(self, name, value)

def _merge_http(self, other_dict):
    # handle "headers" specially, kind of like:
    # https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/plan/plan.go#L398

...
"""

Comment (benhoyt) [reply ops/pebble.py:1147]: Would you be willing to do the same for log-targets while you're in here? See also https://github.com/canonical/operator/pull/1194, which tried to do this, but got stale. The original PR owner hasn't responded for a long time, but -- speaking of "merging", maybe we could merge these efforts?

File: test/test_testing.py [hunk 2]
```diff
@@ -4073,6 +4073,226 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
             == plan.to_yaml()
         )
 
+    def test_add_layer_checks_combine_override_replace(self, client: _TestingPebbleClient):
```
benhoyt, line 4076: This is quite a lot of YAML. Can we minimize the test cases to test just the features we're adding?

amandahla: Good point, removed.

ops/pebble.py
amandahla: I added the _merge methods now and I'll start working on the log-targets, they look very similar to this 

BTW I ended up doing some mess while rebasing this branch so other commits are appearing right now. Is that ok? I believe once the PR is approved, will be squashed right?

amandahla: Done :)

amandahla: ah one more thing, about header wouldnt be enough doing the `getattr(self.http, name).update(value)` since is a ` Dict[str, str]`?

.github/actions/update-charm-pins/main.py
```diff
@@ -9,14 +9,14 @@
 from httpx import Client
 from ruamel.yaml import YAML
 
-yaml = YAML(typ="rt")
+yaml = YAML(typ='rt')
```
benhoyt, line 12: I think the changes to this file are probably good (more in line with our Ruff format). However, I don't think they belong in this PR. Can you please revert them here?

ops/pebble.py
```diff
@@ -1135,6 +1135,64 @@ def __eq__(self, other: Union['CheckDict', 'Check']) -> bool:
...
+        For attributes present in both objects, the passed in exec
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value:
+                continue
+            if name == 'environment':
+                getattr(self.exec, name).update(value)
+            else:
+                setattr(self.exec, name, value)
```
benhoyt, line 1150: So I think this is close, and the `.update()` is a good idea. However, there are a few issues here:

- `other` is a dict here, so it should just be `other.items()`, not `other.__dict__.items()`
- because it's a dict, we can't use `getattr` and `setattr`, `self.exec[name]` and `self.exec[name] = value` would work
- if the `environment` key isn't present in the current `self.exec` dict, we need to handle that case -- probably with `self.exec.get()`
- unfortunately you can't directly update, you have to reassign, for the case when it's using the `get` default
- I think we probably want `if self.exec is None: self.exec = {}` at the start of these methods, in case the base layer doesn't have it; seems unlikely to happen in practice, but still a good safeguard

Overall, I think we want something more like this:

```python
        if self.exec is None:
            self.exec = {}
        for name, value in other.items():
            if not value:
                continue
            if name == 'environment':
                current = self.exec.get(name, {})
                current.update(value)
                self.exec[name] = current
            else:
                self.exec[name] = value
```

Similar with `_merge_http` and `_merge_tcp`.

We should probably add tests of these cases too, to the merge test.

ops/pebble.py
```diff
@@ -1189,6 +1247,22 @@ def __eq__(self, other: Union['LogTargetDict', 'LogTarget']):
...
+        """Merges this log target object with another log target definition.
+
+        For attributes present in both objects, the passed in log target
+        attributes take precedence.
+        """
+        for name, value in other.__dict__.items():
+            if not value or name == 'name':
+                continue
+            if name == 'labels':
+                getattr(self, name).update(value)
```
benhoyt, line 1260: `self.labels` is actually an optional dict, but I think it should be a non-optional dict, to make this a bit simpler. I think we can just make that change. In `LogTarget.__init__`, could you please make this change? Change this:

```python
        if labels is not None:
            labels = copy.deepcopy(labels)
        self.labels: Optional[Dict[str, str]] = labels
```

to this:

```python
        if labels is not None:
            labels = copy.deepcopy(labels)
        else:
            labels = {}
        self.labels: Dict[str, str] = labels
```

You'll have to update an assert in `test_pebble.py` to `assert target.labels == {}` too.

ops/testing.py
```diff
@@ -3204,6 +3204,44 @@ def add_layer(
                         s._merge(service)
                     else:
                         layer.services[name] = service
+            for name, check in layer_obj.checks.items():
```
benhoyt, line 3207: Nit: could you please add a blank line between each top-level `for` loop to visually separate them a bit?

ops/testing.py
```diff
@@ -3204,6 +3204,44 @@ def add_layer(
...
+                if not check.override:
+                    raise RuntimeError(
+                        f'500 Internal Server Error: layer "{label}" must define'
+                        f'"override" for check "{name}"'
+                    )
+                if check.override not in ('merge', 'replace'):
+                    raise RuntimeError(
+                        f'500 Internal Server Error: layer "{label}" has invalid '
+                        f'"override" value on check "{name}"'
```
benhoyt, line 3216: The Pebble error message is actually "value for check" rather than "value on check". Could you please update these and the existing service one too?

test/test_testing.py
```diff
@@ -4073,6 +4073,228 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+                up:
+                    override: replace
+                    level: alive
+                    period: 30s
+                    threshold: 1
+                    exec:
+                        command: service nginx status
+            """),
+        )
+        plan = client.get_plan()
```
benhoyt, line 4093: It probably doesn't add much value to do this intermediate `get_plan()` and assertion -- can probably just drop those to shorten the test a bit.

test/test_testing.py
```diff
@@ -4073,6 +4073,228 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+        )
+        client.add_layer(
+            'foo',
+            """\
+            checks:
+              up:
+                override: merge
+                level: alive
+                period: 10s
+                threshold: 5
```
benhoyt, line 4164: Here's where we need significantly more test cases. Perhaps have one `exec` check, one `http`, and one `tcp`. And may need a test where the base layer doesn't have `exec.environment` and one where it does, so we're testing all the different cases in our new logic. Similar with `http.headers`.

test/test_testing.py
```diff
@@ -4073,6 +4073,228 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+    def test_add_layer_log_targets_combine_override_merge(self, client: _TestingPebbleClient):
+        client.add_layer(
+            'foo',
+            pebble.Layer("""\
+            log-targets:
+                baz:
+                    type: loki
+                    location: https://example123.com:3100/loki/api/v1/push
+                    services:
+                        - foo,
```
benhoyt, line 4264: I don't think there should be trailing commas here (I think YAML is treating them as part of the value).

benhoyt (changes requested):
Thanks for this. The structure is very good, but there are a few minor issues with the types and the merging logic.

.github/actions/update-charm-pins/main.py
tonyandrewmeyer: Opened [a PR to fix this](https://github.com/canonical/operator/pull/1278), which I've hit too - it happens if you use the `tox -e fmt` we provide :)

dimaqq:
I apologise for this project’s slightly unconventional settings wrt. quotation style and line length.

Please use `pre-commit` and it will auto fix the code un your branch

.github/actions/update-charm-pins/main.py
dimaqq: #1278 so that github links the two.

dimaqq: #1278 has been merged; an alternative to reverting this file would be to rebase.

test/test_testing.py
amandahla: Agree :)

test/test_testing.py
```diff
@@ -4073,6 +4073,295 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+            pebble.Layer("""\
+            log-targets:
+                baz:
+                    override: replace
+                    type: loki
+                    location: https://example.com:3100/loki/api/v1/push
+                    services:
+                        - foo
+                    labels:
+                        key: value,
```
benhoyt, line 4287:
```suggestion
                        key: value
```

test/test_testing.py
```diff
@@ -4073,6 +4073,295 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+            pebble.Layer("""\
+            log-targets:
+                baz:
+                    override: replace
+                    type: loki
+                    location: https://example123.com:3100/loki/api/v1/push
+                    services:
+                        - foo,
+                    labels:
+                        key1: value1,
```
benhoyt, line 4302:
```suggestion
                        key1: value1
```

test/test_testing.py
```diff
@@ -4073,6 +4073,295 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+            """),
+            combine=True,
+        )
+        plan = client.get_plan()
+        assert (
+            textwrap.dedent("""\
+            log-targets:
+              baz:
+                labels:
+                  key1: value1,
```
benhoyt, line 4312:
```suggestion
                  key1: value1
```

test/test_testing.py
benhoyt: It looks like this isn't resolved yet -- there are still trailing commas on these values.

benhoyt (approved):
Code looks good now, approving! Just a couple of nit issues in the tests. I'll also ask another Charm Tech team member to review.

amandahla:
off topic: Why the PR title validation fails if I change it to "fix: add checks and log_targets to ops.testing"?

test/test_testing.py
amandahla: My bad. Commas removed now.

benhoyt (approved):

benhoyt:
> off topic: Why the PR title validation fails if I change it to "fix: add checks and log_targets to ops.testing"?

@amandahla Must have been a glitch -- I just changed it back to that now, and it ran fine. We'll keep an eye on it.

benhoyt:
@amandahla We require two reviews for PRs on this repo, so I've asked one more Charm Tech reviewer to review this (probably in the next couple of days), but it's looking good!

test/test_testing.py
```diff
@@ -4073,6 +4073,295 @@ def test_add_layer_merge(self, client: _TestingPebbleClient):
...
+                labels:
+                  key1: value1
+                location: https://example123.com:3100/loki/api/v1/push
+                override: replace
+                services:
+                - foo
+                type: loki
+            """)
+            == plan.to_yaml()
+        )
```
tonyandrewmeyer, line 4320: This is asserting that the plan is equal to the second layer, right? Because `override: replace` was used.

I think the test would be clearer (and more resilient) if it made that explicit. Something like (untested):

```suggestion
        layer = pebble.Layer("""\
            log-targets:
                baz:
                    override: replace
                    type: loki
                    location: https://example123.com:3100/loki/api/v1/push
                    services:
                        - foo
                    labels:
                        key1: value1
            """
        )
        client.add_layer(
            'foo',
            layer,
            combine=True,
        )
        plan = client.get_plan()
        assert layer.to_yaml() == plan.to_yaml()
```

tonyandrewmeyer (approved):
A couple of small suggestions, but looks great, thanks!

test/test_testing.py
amandahla: I dont have a strong opinion about it so I don't mind changing it but to me the existing code looks explicit too :)
I follow the pattern in the test_add_layer_combine_override_replace. I'm afraid that changing just one of them would look confusing.

ops/testing.py
```diff
@@ -3205,6 +3206,46 @@ def add_layer(
...
+                    raise RuntimeError(
+                        f'500 Internal Server Error: layer "{label}" has invalid '
+                        f'"override" value for log target "{name}"'
+                    )
+                elif log_target.override == 'replace':
+                    layer.log_targets[name] = log_target
+                elif log_target.override == 'merge':
+                    if combine and name in layer.log_targets:
+                        lt = layer.log_targets[name]
+                        lt._merge(log_target)
```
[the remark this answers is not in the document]
benhoyt: Amanda changed them but then they were inconsistent, so I've changed all three to the one-line version to make them consistent and avoid the naming debate.

PR #1273, opened 2024-06-25
docs: add main to ops api reference

IronCore864:
The docstring of `main` is missing from RTD, this adds it back.

According to Dora, it would be nice to link to `main` in [here](https://juju.is/docs/sdk/ops#opsmain-1).

I don't have a very good solution to add it, did some testing, this is the best I can do, preview [here](https://ops--1273.org.readthedocs.build/en/1273/#module-ops.main).

Tried to put it in front of the first entry of the ops module but then the indentation for following items are wrong, since they are handled automatically by autodoc.

Also tried to add only the function `ops.main.main` instead of the module, in this way we can put it in front of [`ops.ActionEvent`](https://ops--1273.org.readthedocs.build/en/1273/#ops.ActionEvent), but it is still before [the big paragraph](https://ops--1273.org.readthedocs.build/en/1273/#module-ops) that comes from `__init__.py`, which looks even worse.

So, the current preview looks the best, although having `main` at the bottom seems not so good.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,7 +12,7 @@ ops module
 .. automodule:: ops
 
 
-`ops.main`
+ops.main
 -----------
 
 .. automodule:: ops.main
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,7 +12,7 @@ ops module
 .. automodule:: ops
 
 
-ops.main
+main
 -----------
 
 .. automodule:: ops.main
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,6 +12,12 @@ ops module
 .. automodule:: ops
 
 
+`ops.main`
+-----------
+
+.. automodule:: ops.main
+
+
 ops.pebble module
 =================
 


tmihoc:
Oh, I see now what you mean by it being rendered at the bottom:

image

So, before we had a gap but things looked fine. Now we've filled the gap but the placement is not ideal. Which one would I rather have? Definitely the latter. If `ops.main` is a construct in Ops, I expect to be able to find it in the search and to be able to link to it. So, I think this PR is in the right direction. Approved from my part, and thanks!

benhoyt (approved):
This seems fine to me, thanks! Not perfect placement, but seems like a good start!

I guess there's no way to insert `ops.main` into the list of autodoc'd members?

IronCore864:
> I guess there's no way to insert `ops.main` into the list of autodoc'd members?

I've also tried to do this but failed, no matter what I did, it wouldn't show up in the members...

dimaqq (approved):

dimaqq:
I too was wondering why reference for `main` was missing.

Thank you for adding it :)

tonyandrewmeyer (approved):

tonyandrewmeyer:
@IronCore864 if you rebase this the tests should pass now.

PR #1278, opened 2024-06-27
chore: run ruff format over charm pin update code

tonyandrewmeyer:
The only changes are from running `tox -e fmt`, and adding `ruff format --preview check` to the `tox -e lint` commands.

I don't think these changes are generally important, being in a separate file only used in CI, but ideally running `tox -e fmt` doesn't bring in unrelated changes when working on a branch, and we might as well be consistent.

### DIFF
--- a/.github/actions/update-charm-pins/main.py
+++ b/.github/actions/update-charm-pins/main.py
@@ -9,14 +9,14 @@ import sys
 from httpx import Client
 from ruamel.yaml import YAML
 
-yaml = YAML(typ="rt")
+yaml = YAML(typ='rt')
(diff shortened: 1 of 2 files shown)


dimaqq:
A worthy change… was the file excluded from pre commit and/or tox?

if so, let’s include it.

if it was included , then I’m doubly sorry for having merged it regardless!

tonyandrewmeyer:
> A worthy change… was the file excluded from pre commit and/or tox?
> 
> if so, let’s include it.

It was included in the formatting but not in the linting. I've added that too, now.

dimaqq (approved):

IronCore864 (approved):

PR #668, opened 2021-12-01
Add checks config fields and get_checks API

benhoyt:
Add support for the new backoff and checks-related fields in the plan configuration types, as well as support for the new `/v1/checks` API (`Client.get_checks` and `Container.get_checks` and `Container.get_check`).

See also the brief docs I'm adding in https://github.com/benhoyt/juju-docs/pull/3 (in addition to the API reference docs in this PR).

PR #1281, opened 2024-07-03
feat: add support for Pebble check-failed and check-recovered events

tonyandrewmeyer:
Two new event types are supported (actually getting these events will also require a Juju change):

* `PebbleChangeFailedEvent` - emitted when a Pebble check in a container hits the failure threshold
* `PebbleChangeRecoveredEvent` - emitted when a Pebble check in a container passes after previously hitting the failure threshold

Each event type has a single new attribute (in addition to those inherited from `WorkloadEvent`):

* `info` - this is a shortcut to Pebble's `get_check` method (not otherwise accessible from the charm), but will only query Pebble if more than the `name` is requested, and will cache the result on the event instance.

Two new enums are added to Pebble, to make it easier to work with these events, particularly with Harness:

* `ChangeStatus` - the possible values for `ops.pebble.Change.status` (note that this is the full list, but we only expect charms to see Doing, Done, and Error for now).
* `ChangeKind` - the possible values for `ops.pebble.Change.kind` (again, this is the current full list, but we only expect charms to see perform-check and recover-check at the moment).

`Harness.pebble_notify` is extended to support emitting the two new events. This requires implementation knowledge (that the check events are triggered by a Pebble notice of type change-updated), but provides minimal support for testing with Harness, and will be supplemented by user-friendly testing in Scenario.

The testing Pebble backend gains an implementation for `get_change` and `get_checks`. These are not used directly here, but will be used in Scenario, and it makes more sense to have the implementation in ops.

This is the ops portion of [OP046](https://docs.google.com/document/d/13ItH8l5ZSmmv9WqZXpqjV8EifZU5e3zxINiNLubnC68/edit). It requires a [corresponding change to Pebble](https://github.com/canonical/pebble/pull/444) and [change to Juju](https://github.com/juju/juju/pull/17655).


tonyandrewmeyer:
This is not particularly usable until the Juju and Pebble changes land, but it should be ok to review now (but not merge: we need to wait in case any changes in the Juju/Pebble side have impacts here).

dimaqq (commented):
Partial review from a mobile device

ops/charm.py
```diff
@@ -831,6 +831,55 @@ class PebbleCustomNoticeEvent(PebbleNoticeEvent):
...
+    Note that the check may have started passing by the time this event is
+    emitted (which will mean that a :class:`PebbleCheckRecoveredEvent` will be
+    emitted next). If the handler is executing code that should only be done
+    if the check is currently failing, check the current status with
+    ``event.info.status == ops.pebble.CheckStatus.DOWN``.
+    """
+
+
+class PebbleCheckRecoveredEvent(PebbleCheckEvent):
+    """Event triggered when a Pebble check passes after exceeding the failure threshold."""
```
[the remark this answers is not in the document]
tonyandrewmeyer: No, [there's a failure threshold, but no recovery threshold](https://github.com/canonical/pebble#layer-specification). So you get this event as soon as the test passes. The two distinctions I'm trying to make are:

* You don't get it when the check is first added if it's succeeding immediately (it's "recovered", not "succeeded").
* You don't get it when the check succeeds after failing, unless it failed long enough to hit the threshold and be considered down.

Maybe the problem is trying to fit that all into the one line. I'll try explaining those in the body and making the intro simpler.

ops/pebble.py
```diff
@@ -1471,6 +1471,15 @@ class NoticeType(enum.Enum):
     """Enum of notice types."""
 
     CUSTOM = 'custom'
+    """A custom notice reported via the Pebble client API or ``pebble notify``.
```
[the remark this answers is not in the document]
tonyandrewmeyer: Sphinx.

ops/pebble.py
```diff
@@ -1483,6 +1492,51 @@ class NoticesUsers(enum.Enum):
...
+    """Undo status means the change or task should be undone, probably due to an error."""
+
+    UNDOING = 'undoing'
+    """UndoingStatus means the change or task is being undone or an attempt was made to undo it."""
+
+    ERROR = 'error'
+    """Error status means the change or task has errored out while running or being undone."""
+
+    WAIT = 'wait'
+    """Wait status means the task was accomplished successfully but some
```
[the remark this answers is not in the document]
tonyandrewmeyer: This is briefly mentioned in the PR description. You shouldn't generally need to use these in the charm code - Juju does the work of figuring out that it's (perform-check, error) or (recover-check, done) and translates that into an appropriate event. We don't even give the notice key/change ID, so querying Pebble for the raw data would be tricky.

You do need to use them if you're writing Harness tests (as opposed to Scenario, where it will be much simpler). For that, you only really need those two combinations.

However, we do expose `get_notice` and `get_check` in `ops.pebble.Client` so it seems more correct to me that we let people use an enum rather than use a bunch of strings. It's quite a few lines of code, but should be nearly no maintenance burden (it seems unlikely the list of statuses will change, and if there are new change kinds then that's probably something we would want to add anyway) and doesn't change performance in any way. It's a few more options in an autocomplete, but I don't think that will really be impacting people.

I did expect push back on this, and I don't feel super strongly about it, so please do outline why it would be better to not do it if you disagree.

ops/model.py
```diff
@@ -3792,6 +3792,40 @@ def _ensure_loaded(self):
...
+    def __init__(self, container: Container, name: str):
+        self._container = container
+        self.name = name
+        self._info: Optional[ops.pebble.CheckInfo] = None
+
+    def __repr__(self):
+        return f'LazyCheckInfo(name={self.name!r})'
+
+    def __getattr__(self, item: str):
+        # Note: not called for defined attribute `name`.
```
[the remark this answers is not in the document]
tonyandrewmeyer: Sorry, not sure what you mean by this. Do you mean the `item` argument for `__getattr__`? That's a regular Python magic method, so it has the normal meaning, it's the (otherwise undefined) attribute that someone is trying to read.

ops/pebble.py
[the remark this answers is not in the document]
benhoyt: I thought about suggesting we only add the ones we need, but I think this is fine as is. It's just enum values, not complicated code, and these are very unlikely to change.

ops/model.py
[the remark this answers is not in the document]
dimaqq: Oh, I didn't even think of that, I guess the comment was redundant for me.
Maybe it's useful for someone else who reads this code some day, so I'm fine with the comment.

ops/pebble.py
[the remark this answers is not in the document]
dimaqq: I don't have a strong opinion against inclusion of these values.
I would maybe consider documenting which of these aren't expected to be visible to the ops API user, so that they don't write huge dispatch functions for every possible value.

benhoyt (approved):
Thanks!

test/test_testing.py
```diff
@@ -6813,6 +6833,48 @@ def _on_pebble_custom_notice(self, event: ops.PebbleCustomNoticeEvent):
...
+        harness.charm.observe_container_events('foo')
+
+        def get_change(_: ops.pebble.Client, change_id: str):
+            return ops.pebble.Change.from_dict({
+                'id': change_id,
+                'kind': pebble.ChangeKind.PERFORM_CHECK.value,
+                'summary': '',
+                'status': pebble.ChangeStatus.ERROR.value,
+                'ready': False,
+                'spawn-time': '2021-02-10T04:36:22.118970777Z',
```
dimaqq, line 6858: I'm curious about the significance of this timestamp 😆

ops/testing.py
```diff
@@ -3658,10 +3689,6 @@ def _notify(
 
         Return a tuple of (notice_id, new_or_repeated).
         """
-        if type != pebble.NoticeType.CUSTOM:
```
dimaqq, line 3661: I'm curious about the decision between widening the allow-list vs. not having a list at all.

ops/testing.py
```diff
@@ -3632,8 +3655,16 @@ def send_signal(self, sig: Union[int, str], service_names: Iterable[str]):
...
-        raise NotImplementedError(self.get_checks)  # type:ignore
+    def get_checks(
+        self, level: Optional[pebble.CheckLevel] = None, names: Optional[Iterable[str]] = None
+    ) -> List[pebble.CheckInfo]:
+        if names is not None:
+            names = frozenset(names)
+        return [
+            info
+            for info in self._check_infos.values()
+            if (level is None or level == info.level) and (names is None or info.name in names)
```
dimaqq, line 3666: Minor: I wish we could rename the local variable `info` so that it's clear it doesn't stand for e.g. logging level `INFO`.

ops/testing.py
```diff
@@ -3023,6 +3039,8 @@ def __init__(self, backend: _TestingModelBackend, container_root: pathlib.Path):
         self._exec_handlers: Dict[Tuple[str, ...], ExecHandler] = {}
         self._notices: Dict[Tuple[str, str], pebble.Notice] = {}
         self._last_notice_id = 0
+        self._changes: Dict[str, pebble.Change] = {}
+        self._check_infos: Dict[str, pebble.CheckInfo] = {}
```
dimaqq, line 3043: "CheckInfo" is a thing in pebble.
At the same time, `self._check_infos` just reads unclear/muddy.
I wish we could name things a bit better, but I guess in this case we have to keep pebble (or juju) nomenclature, don't we?
cc @benhoyt

ops/model.py
```diff
@@ -3792,6 +3792,40 @@ def _ensure_loaded(self):
...
+    """Provide lazily-loaded access to a Pebble check's info.
+
+    The attributes provided by this class are the same as those of
+    :class:`ops.pebble.CheckInfo`, however, the notice details are only fetched
+    from Pebble if necessary (and cached on the instance).
+    """
+
+    name: str
+    level: Optional[Union[pebble.CheckLevel, str]]
+    status: Union[pebble.CheckStatus, str]
```
dimaqq, line 3805: What is the reason for union types here? Is it because other, custom values are possible?

ops/charm.py
```diff
@@ -831,6 +831,60 @@ class PebbleCustomNoticeEvent(PebbleNoticeEvent):
...
+        self.info = model.LazyCheckInfo(workload, check_name)
+
+    def snapshot(self) -> Dict[str, Any]:
+        """Used by the framework to serialize the event to disk.
+
+        Not meant to be called by charm code.
+        """
+        d = super().snapshot()
+        d['check_name'] = self.info.name
+        return d
```
dimaqq, line 856: Super-minor: why `d`? because it's a dict? delta? data?

dimaqq: How about:

```suggestion
        return {**super().snapshot(), "check_name": self.info.name}
```

dimaqq (approved):

test/test_testing.py
tonyandrewmeyer: I was too lazy to generate one myself so copied it from one of the existing tests. I would guess it's from when they were added.

ops/testing.py
tonyandrewmeyer: Having the list there is annoying when adding new functionality - for example, the Juju PR goes first but has an ops charm in the tests, so has to patch all sorts of things. Basically, ops shouldn't be holding you back if Pebble has never functionality.

I don't really think we gain much by having a check here. If we want to warn people then it seems like static checking would be more approximate for this case.

@benhoyt might have more to say here: he added the check originally, but then also removed it in the PR (later reverted) that added support for change-updated-notice events.

ops/testing.py
tonyandrewmeyer: There was some discussion about the name in the spec process (in a review with @benhoyt and then in the last spec review meeting). `check_info` would be most accurate, but it's a bit long, and "check" is there in the event name. `check` is problematic because a `Check` class exists, and because "check" could be a verb, although you'd expect a method, not a property in that case. Containers have `get_check` and `get_checks` so this inconsistent with that, but as above, I think it's confusing that you don't get a `Check` from those. It could be `get_check_info as a method, with the name as another attribute, but that's less convenient (more explicit, but less consistent with notices).

I don't really buy the argument that someone would think it's `logging.info` or `logging.INFO`. Is there some reason to think people would expect that?

My preference is to stick with `info` but I'm open to other suggestions.

ops/testing.py
tonyandrewmeyer: Do you have a suggestion for a better name? I don't love `_check_infos` either, but it's accurate and consistent, and since it's private I don't care as much.

ops/charm.py
tonyandrewmeyer: I'm not a fan of the `d` either, but it's consistent with all of the others. @benhoyt what's your take on this? A separate PR to adjust them all? Have this one be inconsistent? Change them all in this PR as a drive-by? Leave it alone for now?

ops/model.py
tonyandrewmeyer: Here it's because this is really just a wrapper over `CheckInfo` so it needs to have the same types.

In the real `CheckInfo`, I believe it's so that if Pebble gains additional status or level values, ops will keep working. @benhoyt might have more to say on this?

ops/testing.py
benhoyt: Nothing to add -- what Tony said makes sense (and is why I removed the check in the now-reverted PR).

ops/testing.py
benhoyt: Is all Dima is asking about the local variable name? That would be fine to change -- but I disagree it's needed here. This is just a variable local to the list comprehension, so less context seems fine to me.

But as for the `.info` property, yeah, that's already been decided in the spec review process.

Yeah, let's stick with what Tony has here.

ops/charm.py
benhoyt: I like a separate PR to adjust them all would be fine. Probably `dct` or `snapshot`?

ops/model.py
benhoyt: Yeah, that's right -- if Pebble gains additional values but they haven't been added to Ops yet, we won't get exceptions. However, in retrospect I'm not sure it was a good choice. The "enum or str" type is very awkward.

ops/charm.py
tonyandrewmeyer: @dimaqq would you like to open one for that? I can do it if not.

ops/model.py
tonyandrewmeyer: I think `StrEnum|str` is a lot less awkward to deal with, so we could switch to that once 3.11 can be a minimum.

ops/testing.py
dimaqq: I first wrote a long justification for Tony's way (summary at the end), but since the ask is specifically for a suggestion or an alternative here it goes... please tread lightly:

### Alternative

Revisit https://github.com/canonical/operator/pull/668

Rename `CheckInfo` (which mirror go internal structure and a JSON block in the API response) to something like `CheckStatus` or `DetailedCheckStatus`, because that's what (I think) it is.

Now is probably the last opportunity to change names, given that this PR makes responding to check failures functional.

Caveats:
We'd have to make a call on offering an alias for (hypothetical?) existing users of the public `ops.pebble.*` API.

CheckInfoMapping is re-exported at top level, but CheckInfo is not; it seems that only jhack uses this type directly; Only sunbeam seems to use `.get_checks()`.

### Keep Current

The library's job is to expose specific juju and pebble concepts, so I guess we have to stay close to the original naming.

This particular member is "private", thus it's not a burden to our users, and it's annotated with a more readable type (I imagine the key is the check name that's defined in the pebble plan; and the value is a typed thing).

ops/charm.py
dimaqq: My vote is for an inline dictionary merge, as it obviates local variable entirely, while is just as readable.

ops/model.py
```diff
@@ -3792,6 +3792,40 @@ def _ensure_loaded(self):
...
+class LazyCheckInfo:
+    """Provide lazily-loaded access to a Pebble check's info.
+
+    The attributes provided by this class are the same as those of
+    :class:`ops.pebble.CheckInfo`, however, the notice details are only fetched
+    from Pebble if necessary (and cached on the instance).
+    """
+
+    name: str
+    level: Optional[Union[pebble.CheckLevel, str]]
```
dimaqq, line 3804: Looking through the implementation, I think that this field is always set, either a raw string, or an enum value, where the latter may be `UNSET`, but the field is never `None.

ops/testing.py
benhoyt: We discussed this briefly in our daily sync. I think you're right that `CheckStatus` is a better name, but we have 4 "FooStatus" types in pebble.py (and have for a long time), so if we changed one, it'd be inconsistent with the others. Renames like this can cost quite a lot (cf "integration" vs "relation"), so I think we should just stick with the status quo here. `CheckInfo` is not sooo bad.

ops/model.py
dimaqq: I guess this can't helped, resolving.

dimaqq: FWIW, my preference in a situation like this would be:

```py
status: CheckStatus|None
raw_status: str
```

That way, most callers use `.status`.
When a charmer tries to use the new pebble feature, they are nudged to upgrade ops by the static type checker.
A determined, enlightened caller who depends on unreleased pebble functionality can still get their custom value.


Another way would be to rely on Literals instead of Enums.

ops/model.py
tonyandrewmeyer: This is copied from `CheckInfo`, which should match. The docs there say:

> This can be :attr:`CheckLevel.ALIVE`, :attr:`CheckLevel.READY`, or None (level not set).

The `CheckInfo.__init__` accepts `None`, although I don't think anyone would normally be manually creating one.  `from_dict` will set it to `None` if there's no `level` key in the dictionary. In [the Pebble code](https://github.com/canonical/pebble/blob/ffb99d9d60faa946c1ccf6afd2d7c813b0c2030b/client/checks.go#L37) is seems like "unset" is the empty string (like the ops enum). It does seem like the [Pebble API](https://github.com/canonical/pebble/blob/ffb99d9d60faa946c1ccf6afd2d7c813b0c2030b/internals/daemon/api_checks.go#L56) will always return a "level" value.

@benhoyt do you know the story here? Was there an earlier version of Pebble that wouldn't necessarily have "level" in the check response? If so then probably we need to continue supporting that and it should stay as-is (although a comment like the one for change-id might be worthwhile). If "level" will have been there for all Pebble versions, then I think @dimaqq is correct that removing the ability for it to be `None` makes sense (I'd do that in a separate PR).

ops/testing.py
tonyandrewmeyer: Ah, I missed the "local" in the comment. Indeed, probably the intention was just the local variable name. And I guess then the reason that it brings logging to mind is that there's "info.level" which seems a bit "logging level"-y.

But for the local name then I would agree with Ben's first paragraph.

ops/model.py
benhoyt: No, the "level" field is optional in the check config (you don't have to set a level), and it's marked as "omitempty" for the JSON serialisation, meaning its omitted from the JSON if it's unset (empty string) in the plan.

tonyandrewmeyer: Ah, the `omitempty` is what I missed. Thanks! So this needs to stay as-is.

PR #1283, opened 2024-07-11
bug fix for 1.5

xtrusia:
I faced issue related two commits when I use 1.5.5

error: Cannot access attribute "nrpe_external_master_relation_changed" for class "CharmEvents"
AttributeError: 'CharmEvents' object has no attribute 'xxxx-relation-changed'

With these patches, it works fine.

Thanks

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -533,6 +533,14 @@ class ObjectEvents(Object):
         event_kinds = ', '.join(sorted(self._event_kinds()))
         return '<{}.{}: {}>'.format(k.__module__, k.__qualname__, event_kinds)
 
+    def __getattr__(self, name: str) -> Any:
+        """The existence of this method tells type checkers to allow dynamic attributes.
+
+        This allows charms to access dynamically-defined events such as
+        ``self.on.db_relation_joined`` without Mypy/Pyright whining.
+        """
+        return super().__getattribute__(name)
+
 
 class PrefixedEvents:
     """Events to be found in all events using a specific prefix."""
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -414,57 +414,102 @@ class Harness(Generic[CharmType]):
         """
         self._backend._cleanup()
 
-    def _create_meta(self, charm_metadata: Optional[YAMLStringOrFile],
-                     action_metadata: Optional[YAMLStringOrFile]) -> CharmMeta:
+    def _create_meta(self, charm_metadata_yaml: Optional[YAMLStringOrFile],
+                     action_metadata_yaml: Optional[YAMLStringOrFile]) -> CharmMeta:
         """Create a CharmMeta object.
 
         Handle the cases where a user doesn't supply explicit metadata snippets.
+        This will try to load metadata from ``<charm_dir>/charmcraft.yaml`` first, then
+        ``<charm_dir>/metadata.yaml`` if charmcraft.yaml does not include metadata,
+        and ``<charm_dir>/actions.yaml`` if charmcraft.yaml does not include actions.
         """
         filename = inspect.getfile(self._charm_cls)
         charm_dir = pathlib.Path(filename).parents[1]
 
+        charm_metadata: Optional[Dict[str, Any]] = None
+        charmcraft_metadata: Optional[Dict[str, Any]] = None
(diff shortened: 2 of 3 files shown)


benhoyt:
Hi @xtrusia. I would *much* rather not pull these commits into the 1.5 branch. That's very old by now, and we're really only doing security fixes there. Is there a reason you can't upgrade to ops 2.x? The [differences between 1.5 and 2.0](https://github.com/canonical/operator/releases/tag/2.0.0) are relatively minor, and it shouldn't take much work to get an ops 1.5 charm using the latest 2.x. Then you can be on the latest and get these changes automatically.

xtrusia:
> Hi @xtrusia. I would _much_ rather not pull these commits into the 1.5 branch. That's very old by now, and we're really only doing security fixes there. Is there a reason you can't upgrade to ops 2.x? The [differences between 1.5 and 2.0](https://github.com/canonical/operator/releases/tag/2.0.0) are relatively minor, and it shouldn't take much work to get an ops 1.5 charm using the latest 2.x. Then you can be on the latest and get these changes automatically.

hi thanks for the explanation
I'm trying to make lldpd charm (stable/bionic) support Bionic as the customer requested but there was python library dependency issue. e.g dataclasses,  Typed..

so by downgrading ops, reached to 1.5.5 

I guess 2.x doesn't support Bionic right?

Thanks a lot

benhoyt:
Ah, yes, if you need to support Bionic, you will need to stay on 1.5, as 2.x requires Python 3.8 (Bionic only has Python 3.6).

However, can you explain what the original errors were? Do you have a branch you can point me to where you get these errors and I can reproduce by running the unit tests locally?

```
error: Cannot access attribute "nrpe_external_master_relation_changed" for class "CharmEvents"
AttributeError: 'CharmEvents' object has no attribute 'xxxx-relation-changed'
```

dimaqq:
I could suggest an alternative approach, but I'd like to see the charm and its tests first.
Please link :)

xtrusia:
> Ah, yes, if you need to support Bionic, you will need to stay on 1.5, as 2.x requires Python 3.8 (Bionic only has Python 3.6).
> 
> However, can you explain what the original errors were? Do you have a branch you can point me to where you get these errors and I can reproduce by running the unit tests locally?
> 
> ```
> error: Cannot access attribute "nrpe_external_master_relation_changed" for class "CharmEvents"
> AttributeError: 'CharmEvents' object has no attribute 'xxxx-relation-changed'
> ```

@dimaqq @benhoyt Thanks, 
https://github.com/xtrusia/charm-lldpd/tree/stable/bionic
I also see an error in self.state.ready(I would like to know if i can just rempve it) . you can test it please let me know if there is anything I can help

I'm not the initial developer of this so my understanding could be limited

Thanks

dimaqq:
Summary of the issue in the charm repo, bionic branch:
1. all 13 unit tests fail with `AttributeError: 'CharmEvents' object has no attribute 'nrpe_external_master_relation_changed'`
2. `static-charm` tox target
   a. 2 cases of the above, and
   b. `Argument of type "Literal['ready']" cannot be assigned to parameter "key" of type "Literal['on']" in function "__getattr__"`

tonyandrewmeyer:
@xtrusia why use charmcraft.yaml and not metadata.yaml? It seems like it's less invasive for the charm to stick with the old file than to do a release of old ops that supports the new file.

dimaqq:
The Charm Tech team discussed this issue this morning, and our thinking mainly does like this:

Bionic has reached the end of normal support.
This library has moved on to newer set of Python versions.
Other tooling, as it turns out has moved on too.

Thus, I've opened a PR https://github.com/xtrusia/charm-lldpd/pull/1 that runs unit and static tests with the old toolset.

xtrusia:
> The Charm Tech team discussed this issue this morning, and our thinking mainly does like this:
> 
> Bionic has reached the end of normal support. This library has moved on to newer set of Python versions. Other tooling, as it turns out has moved on too.
> 
> Thus, I've opened a PR [xtrusia/charm-lldpd#1](https://github.com/xtrusia/charm-lldpd/pull/1) that runs unit and static tests with the old toolset.

Thanks a lot for this work you all. I'll test your PR(I'm sure it works) and adjust it for bionic.

PR #1285, opened 2024-07-17
ci: bump the Go version to match Pebble

tonyandrewmeyer:
[Pebble now requires Go 1.22](https://github.com/canonical/pebble/releases/tag/v1.15.0), so bump the CI real Pebble tests to match.

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -75,10 +75,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Set up Go 1.20
+      - name: Set up Go
         uses: actions/setup-go@v5
         with:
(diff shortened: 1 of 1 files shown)


dimaqq (approved):

benhoyt (approved):
Looks good, thanks

.github/workflows/framework-tests.yaml
```diff
@@ -75,10 +75,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Set up Go 1.20
+      - name: Set up Go
```
tonyandrewmeyer, line 78: Seems to me that there's no benefit in repeating the version here and it's just an easy place for it to get out of sync with the actual version just below.

PR #1290, opened 2024-07-19
fix: use temp dir for secret data

dimaqq:
Change the mechanism used to pass secret values from ops to hook tools from command line to files in temporary directory.

Removes the 2MB total secret size limit imposed by command line length

Hook tool calls juju agent which in turn reads these files, I've confirmed with the Juju team that juju agent and charm code are run as the same user.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -3580,6 +3580,16 @@ class _ModelBackend:
         id = list(info_dicts)[0]  # Juju returns dict of {secret_id: {info}}
         return SecretInfo.from_dict(id, typing.cast(Dict[str, Any], info_dicts[id]))
 
+    @contextlib.contextmanager
+    def _secrets_in_temp_files(self, content: Dict[str, str]) -> Generator[List[str], None, None]:
+        rv: List[str] = []
+        with contextlib.ExitStack() as temp_files:
+            for k, v in content.items():
+                temp = temp_files.enter_context(tempfile.NamedTemporaryFile(buffering=0))
+                temp.write(v.encode('utf-8'))
+                rv.append(f'{k}#file={temp.name}')
+            yield rv
+
     def secret_set(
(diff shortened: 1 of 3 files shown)


dimaqq:
Also tested manually to replacing ops in a running charm venv.

benhoyt (approved):
Yes, I like this approach a lot better, thanks! The code reads a lot more simply, and it's nice to avoid keeping N files open. If you've done manual testing, I'm good with this.

ops/model.py
```diff
@@ -3598,11 +3598,13 @@ def secret_set(
...
             args += ['--rotate', rotate.value]
-        if content is not None:
-            # The content has already been validated with Secret._validate_content
-            for k, v in content.items():
-                args.append(f'{k}={v}')
-        self._run_for_secret('secret-set', *args)
+        with tempfile.TemporaryDirectory() as tmp:
+            # The content is None or has already been validated with Secret._validate_content
+            for k, v in (content or {}).items():
+                with open(f'{tmp}/{k}', mode='w', encoding='utf-8') as f:
```
benhoyt, line 3604: At first I was concerned about slashes in keys, but we (and Juju) is very strict about keys (basically numbers and letters and hyphens) and they've already been checked.

tonyandrewmeyer (approved):
Did a quick check (Juju 3.6, LXD) and all seems good here too. Agreed that this approach seems generally nicer.

ops/model.py
dimaqq: Indeed.
I've checked that actually.
The only possible unaccounted for is a key that's very very long, exceeding directory entry name limit, as we don't limit the length in ops.
I figure it's OK to fail in that case.

dimaqq:
Fixes CVE-2024-41129

PR #1294, opened 2024-07-22
docs: escape <major> etc in hacking.md

dimaqq:
GH markdown converts angle-bracketed words into HTML tags, which, being unknown, are ignored by the browser.


benhoyt:
Thanks!

benhoyt (approved):

PR #1295, opened 2024-07-22
chore: 2.15.0

dimaqq:


CHANGES.md
```diff
@@ -1,3 +1,25 @@
+# 2.15.0 - 22 Jul 2024
+
+## Features
+
+* Add support for Pebble check-failed and check-recovered events (#1281)
+
+## Fixes
+
+* Use temp dir for secret data (#1290)
```
tonyandrewmeyer, line 9: I think this could probably be a bit more detailed? Maybe:

```suggestion
* Pass secret data to Juju via files, rather than as command-line values (#1290)
```

CHANGES.md
```diff
@@ -1,3 +1,25 @@
...
+* Add checks and log_targets to ops.testing (#1268)
```
tonyandrewmeyer, line 10: Maybe a bit more detailed with this one too?

```suggestion
* Include checks and log targets when merging layers in ops.testing (#1268)
```

tonyandrewmeyer (approved):
A couple of small suggested tweaks (I should really get in the habit of proposing these when doing the reviews of the actual PRs, so that they're good before we get to release notes).

I think the changelog really ought to mention (link to?) the advisory, though. Even if it's not public when the PR gets merged, as long as it works when the release is done and the advisory is published, that should be fine.

benhoyt (approved):
Agreed with Tony's feedback (especially having a bit more on check-failed/recovered vents), but approving optimistically.

IronCore864 (approved):

PR #1301, opened 2024-07-25
ci: add a workflow that runs the TIOBE quality checks

tonyandrewmeyer:
This workflow will run the TIOBE "TICS" quality checks and reports them through to the TIOBE server.

It's set to run on the first of each month, so that we have regular updates in TIOBE, but without this hitting every PR (the ruff checks we have are doing that work).

There's a manual way to do this and there's their packaged action. The action seems the cleaner choice, so I've gone with that. We still have to provide the coverage file - we could store that from the normal run and retrieve it here, which would save the resources of re-running the tests, but it feels like it is cleaner to keep these completely separate, which also works well with us splitting that out.


.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,32 @@
...
+          rm .coverage
+          mkdir .coverage
+          mv coverage.xml .coverage/
+
+      - name: TICS GitHub Action
+        uses: tiobe/tics-github-action@v3
+        with:
+          mode: qserver
+          viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
+          ticsAuthToken: ${{ secrets.TICS_AUTH_TOKEN }}
```
dimaqq, line 30: TICS is a clever little name that probably gets everyone confused...

then again, if we're already using this secret, so be it.

dimaqq (approved):

.github/workflows/tiobe.yaml
tonyandrewmeyer: We could change it (something in the secret here is broken - it works in my repo, but not here). Their suggestion was `TICSAUTHTOKEN` and I suggested that we break it up with the underscores. Maybe `TIOBE_AUTH_TOKEN` would be clearer? TICS is their product name, but I think we are generally calling it "TIOBE" or "TIOBE quality checks".

IronCore864 (approved):

PR #1303, opened 2024-07-31
docs: use Sphinx 8

dimaqq:
Don't merge!

I tried to used Sphinx 8, but `furo` (style sheet / look) has only recently begun supporting 8 and there's no release yet. If we wait a month to a quarter, we should be able move straight to 8. 

I was able to get Sphinx 7 to work, would need to clean up the PR if we decided to merge this.


benhoyt:
Would it make sense just to wait till we can use Sphinx 8, to avoid busy-work? In the meantime, we could relax 6.2.1 to 6.x.

dimaqq:
6.2.1 is the latest 6.x release, and Sphinx 6.x is no longer supported. In fact they rejected my bug report wrt. py 3.12.4.

dimaqq:
I'll try to update this to Sphinx 8.

IronCore864:
Looks nice, it seems nothing changed, which is good though!

dimaqq:
Nothing broke visually

https://pianomister.github.io/diffsite/?url1=https%3A%2F%2Fops--1303.org.readthedocs.build%2Fen%2F1303%2F&url2=https%3A%2F%2Fops.readthedocs.io%2Fen%2Flatest%2F

The only real difference detected by tooling is that inline code spans are 1px off, but I can't see this with my own eye.

The search behaviour is different:
- ops.readthedocs.io: 7 hits for "pebble", seems to search only text
- this branch: 240 hits, listing every `ops.pebble.This` and that.

dimaqq:
rebased on latest main, just in case

tonyandrewmeyer:
> * ops.readthedocs.io: 7 hits for "pebble", seems to search only text
> * this branch: 240 hits, listing every `ops.pebble.This` and that.

This is *much* better. For example, a while back we [had a complaint about searching for "secret" not providing useful results](https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$W3jzVqqcm84LMFLfmNGxB11ShXQSrjxme739ARXV83M?via=ubuntu.com&via=matrix.org), which was true. Now it gives you the most obvious result first and lots of other good ones afterwards.

pyproject.toml
```diff
@@ -24,11 +24,13 @@ dynamic = ["version"]
 [project.optional-dependencies]
 docs = [
     "canonical-sphinx-extensions",
+    # pre-release `furo` that supports Sphinx 8
+    #"git+https://github.com/pradyunsg/furo.git@2d3b6baa3371451d997d879a0aa61b3592ade638",
```
benhoyt, line 28: Can we get rid of this comment now?

pyproject.toml
```diff
@@ -24,11 +24,13 @@ dynamic = ["version"]
...
     "furo",
     "linkify-it-py",
     "myst-parser",
     "pyspelling",
-    "sphinx==6.2.1",
+    "sphinx ~= 8.0.0",
```
benhoyt, line 33: Should we lock in an exact version to avoid nuisance updates/changes, e.g., `8.0.2` (latest at this time)?

benhoyt (commented):
Ah yes, search is much better, thanks!

pyproject.toml
dimaqq: Sorry, I forgot to clean this up.

pyproject.toml
dimaqq: I don't think we need to.

docs/requirements.txt acts as a lock file.

A developer has to explicitly run `tox -e docs-deps` to regenerate that file and then open a PR for newer version to kick in. I don't expect nuisance here.

Meanwhile, dependabot is not aware of the causative connection between the two files. Restricting the version here won't help.

benhoyt: `~= 8.0.0` will allow docs-deps to upgrade to 8.1 and 8.2 (which should be compatible), but not 9.0, right? In that case, I think what you're saying makes sense.

benhoyt (approved):

pyproject.toml
tonyandrewmeyer: > `~= 8.0.0` will allow docs-deps to upgrade to 8.1 and 8.2 (which should be compatible), but not 9.0, right? 

I think the [compatibility clause](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release) is tighter than that because of the extra `.0`. If I understand correctly, `~=8.0.0` is the same as `>=8.0.0 and ==8.0.*`. For what you describe it should be `~8.0`.

It's an odd specification since if these are semver then 8.1.0 should be "compatible with" 8.0.0. But my understanding is that it's saying "compatible with 8.0" rather than "compatible with 8", even though it ought to have the same effect.

IronCore864 (approved):

pyproject.toml
dimaqq: TBH I'm fine with either `~= 8.0` or `~= 8.0.0`.

I suspect it doesn't matter which, given 6.x was used before.

PR #1311, opened 2024-08-08
docs: add example Juju version markers

### LINKED ISSUE #1305: Add minimum juju version to event docstring
Some events are only available from a certain version of juju.
https://github.com/canonical/operator/blob/86ca8e31ab01f405aab3d1d140e85f0423e114c6/ops/charm.py#L868
It would be handy to have it documented in the docstring.

tonyandrewmeyer:
Adds the version features were added in Juju to the API reference docs (other than ops.testing).

No version tags are added for anything older than 3.0 - meaning that if there's no version tag, then it was available in the oldest supported Juju version.

Unfortunately, the built-in Sphinx `versionadded` directive does not support customising the text, so while otherwise suitable, it results in "Added in version 3.0" (seems like it's referring to ops 3.0) or "Added in version Juju 3.0". Instead, a new pair of custom directives are added, `jujuversion` and `jujuremoved`. These are very heavily based on `versionadded` and `versionremoved` but have custom text (and we could customise them further, e.g. linking to revision notes) if we wanted to in future). In order to avoid having to also add custom CSS, they use the same classes as the `versionadded` and `versionremoved` directives.

The features being removed in Juju 4.0 also have `jujuremoved` tags added.

I also ran `ruff format` over [custom_conf.py](docs/custom_conf.py), which is why there are some unrelated formatting changes.

Fixes #1305

### DIFF
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -3,7 +3,10 @@ import datetime
 import pathlib
 import sys
 
-import sphinx.domains.changeset
+from docutils import nodes
+
+from sphinx import addnodes
+from sphinx.util.docutils import SphinxDirective
 
 import furo
 import furo.navigation
@@ -18,8 +21,8 @@ _old_compute_navigation_tree = furo._compute_navigation_tree
 
 def _compute_navigation_tree(context):
     tree_html = _old_compute_navigation_tree(context)
-    if not tree_html and context.get("toc"):
-        tree_html = furo.navigation.get_navigation_tree(context["toc"])
+    if not tree_html and context.get('toc'):
+        tree_html = furo.navigation.get_navigation_tree(context['toc'])
     return tree_html
 
 
@@ -29,14 +32,17 @@ furo._compute_navigation_tree = _compute_navigation_tree
 # "invalid signature for autoattribute ('ops.pebble::ServiceDict.backoff-delay')"
 import re  # noqa: E402
 import sphinx.ext.autodoc  # noqa: E402
+
 sphinx.ext.autodoc.py_ext_sig_re = re.compile(
-    r'''^ ([\w.]+::)?            # explicit module name
+    r"""^ ([\w.]+::)?            # explicit module name
           ([\w.]+\.)?            # module and/or class name(s)
           ([^.()]+)  \s*         # thing name
           (?: \((.*)\)           # optional: arguments
            (?:\s* -> \s* (.*))?  #           return annotation
           )? $                   # and nothing more
-          ''', re.VERBOSE)
+          """,
+    re.VERBOSE,
+)
 
 # Custom configuration for the Sphinx documentation builder.
 # All configuration specific to your project should be done in this file.
@@ -100,53 +106,43 @@ html_favicon = '.sphinx/_static/favicon.png'
 # (Some settings must be part of the html_context dictionary, while others
 #  are on root level. Don't move the settings.)
 html_context = {
-
     # Change to the link to the website of your product (without "https://")
     # For example: "ubuntu.com/lxd" or "microcloud.is"
     # If there is no product website, edit the header template to remove the
     # link (see the readme for instructions).
     'product_page': 'juju.is/docs/sdk',
-
     # Add your product tag (the orange part of your logo, will be used in the
     # header) to ".sphinx/_static" and change the path here (start with "_static")
     # (default is the circle of friends)
     'product_tag': '_static/tag.png',
-
     # Change to the discourse instance you want to be able to link to
     # using the :discourse: metadata at the top of a file
     # (use an empty value if you don't want to link)
     'discourse': 'https://discourse.charmhub.io/',
-
     # Change to the Mattermost channel you want to link to
     # (use an empty value if you don't want to link)
     'mattermost': '',
-
     # Change to the Matrix channel you want to link to
     # (use an empty value if you don't want to link)
     'matrix': 'https://matrix.to/#/#charmhub-charmdev:ubuntu.com',
-
     # Change to the GitHub URL for your project
     'github_url': 'https://github.com/canonical/operator',
-
     # Change to the branch for this version of the documentation
     'github_version': 'main',
-
     # Change to the folder that contains the documentation
     # (usually "/" or "/docs/")
     'github_folder': '/docs/',
-
     # Change to an empty value if your GitHub repo doesn't have issues enabled.
     # This will disable the feedback button and the issue link in the footer.
     'github_issues': 'enabled',
-
     # Controls the existence of Previous / Next buttons at the bottom of pages
     # Valid options: none, prev, next, both
-    'sequential_nav': "none"
+    'sequential_nav': 'none',
 }
 
 # If your project is on documentation.ubuntu.com, specify the project
 # slug (for example, "lxd") here.
-slug = ""
+slug = ''
 
 ############################################################
 ### Redirects
@@ -165,9 +161,7 @@ redirects = {}
 ############################################################
 
 # Links to ignore when checking links
-linkcheck_ignore = [
-    'http://127.0.0.1:8000'
-    ]
+linkcheck_ignore = ['http://127.0.0.1:8000']
 
 # Pages on which to ignore anchors
 # (This list will be appended to linkcheck_anchors_ignore_for_url)
@@ -199,7 +193,7 @@ custom_extensions = [
     'sphinx.ext.napoleon',
     'sphinx.ext.todo',
     'sphinx.ext.viewcode',
-    ]
+]
 
 # Add custom required Python modules that must be added to the
 # .sphinx/requirements.txt file.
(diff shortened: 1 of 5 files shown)


dimaqq (commented):
I'm very happy to see this, our users out to be delighted.

If we wanted to be pedantic, I'd go for this syntax:

```rst
    .. versionadded:: 2.15
       Requires Juju 3.5
```

Listing both this library version and Juju version.

ops/charm.py
```diff
@@ -998,6 +1011,8 @@ class SecretExpiredEvent(SecretEvent):
     This event is fired on the secret owner to inform it that the secret revision
     must be removed. The event will keep firing until the owner removes the
     revision by calling :meth:`event.secret.remove_revision() <ops.Secret.remove_revision>`.
+
+    *Added in Juju 3.0*
```
benhoyt, line 1015: I like this formatting the most (though I'd want a full stop at the end of the sentence -- Python docs have this too).

The "tip" definitely isn't right. The "note" is okay but the formatting is very big/heavy. So the plain italics seems best.

As you note, the `versionadded` directive leads to funny formatting with "New in version Juju". Also, it seems the intention with "versionadded" is for the version of *this* project, that is, Ops itself, so if we were going to add those we'd use it for that (but we've decided not to).

So I'd go with the plain italics: `*Added in Juju 3.0.*`

Alternatively, can we write our own `jujuversion` directive? Looks like we might be able to do that *relatively* easily: https://www.sphinx-doc.org/en/master/development/tutorials/extending_syntax.html ... not sure whether it's worth it over just an italics phrase though.

tonyandrewmeyer:
> If we wanted to be pedantic, I'd go for this syntax:
> 
> ```
>     .. versionadded:: 2.15
>        Requires Juju 3.5
> ```
> 
> Listing both this library version and Juju version.

We did previously have some (definitely not complete, and generally using text, not the directive) markers of when functionality was added in ops. [We removed that and adjusted the CHANGES file](https://github.com/canonical/operator/pull/1159) just a few months ago.

I do agree that it's both providing the most information and looks nice:

image

The main reason the ops version markers were removed (rather than extended) was to avoid people looking through to find a specific version of ops to use rather than just always using the latest version. I think a secondary one was that over time this does get quite noisy because almost everything ends up with a little tag. We also considered versioned API docs but decided that it would be best to only do that for major versions (ie. we should have both 2.{latest} and 3.{latest} versions if/when there's an ops 3).

If we do use this format, then it becomes less consistent to not have the markers on ops features that aren't also new Juju features (so almost everything ends up with a version label).

dimaqq:
docs build failed, it seems?

https://readthedocs.org/projects/ops/builds/25247433/

dimaqq:
`Unknown directive type "versionremoved"`

tonyandrewmeyer:
> `Unknown directive type "versionremoved"`

Yeah, that's this:

> (Also in version changes: newer Sphinx have versionremoved, which I've experimented with to see how that would look for the ones being removed in 4.0. With current Sphinx that generates an error, so I'll remove it before moving this PR from draft, but I'm leaving it in place in case people would like to examine that too).

ops/charm.py
tonyandrewmeyer: > Alternatively, can we write our own `jujuversion` directive? Looks like we might be able to do that _relatively_ easily: https://www.sphinx-doc.org/en/master/development/tutorials/extending_syntax.html ... not sure whether it's worth it over just an italics phrase though.

I somehow only just found this comment, which you had indeed made before mentioning it in the daily as well. Sorry for that!

docs/custom_conf.py
```diff
@@ -16,8 +21,8 @@
 
 def _compute_navigation_tree(context):
     tree_html = _old_compute_navigation_tree(context)
-    if not tree_html and context.get("toc"):
-        tree_html = furo.navigation.get_navigation_tree(context["toc"])
+    if not tree_html and context.get('toc'):
```
benhoyt, line 24: Do we need to add this to automatic Ruffing?

docs/custom_conf.py
```diff
@@ -334,3 +330,82 @@ def _compute_navigation_tree(context):
     ('py:class', 'ops.testing.CharmType'),
     ('py:obj', 'ops.testing.CharmType'),
 ]
+
+
+# This is very strongly based on
+# https://github.com/sphinx-doc/sphinx/blob/03b9134ee00e98df4f8b5f6d22f345cdafe31870/sphinx/domains/changeset.py#L46
+# Unfortunately, the built-in class is not easily subclassable without also
+# requiring extra CSS.
```
benhoyt, line 338: This is okay, but would it be easier to add said extra CSS?

ops/charm.py
```diff
@@ -354,6 +354,8 @@ class PreSeriesUpgradeEvent(HookEvent):
     callback method associated with this event, the administrator will initiate
     steps to actually upgrade the series.  After the upgrade has been completed,
     the :class:`PostSeriesUpgradeEvent` will fire.
+
+    .. jujuremoved:: 4.0
```
benhoyt, line 358: This is nice thinking ahead, but maybe we should wait till 4.0 is (closer to being) shipped.

ops/model.py
```diff
@@ -678,6 +683,8 @@ def open_port(
         Use :meth:`set_ports` for a more declarative approach where all of
         the ports that should be open are provided in a single call.
 
+        .. jujuversion:: 3.1
```
benhoyt, line 686: I think the ports functionality works quite a ways back for machines -- it was just the support on K8s that was added in 3.0.3. Or am I missing something?

benhoyt (commented):
Love it, thanks! A few comments, mainly wondering if we should remove the notes about Juju 4.0, which won't be shipped for many months yet.

docs/custom_conf.py
tonyandrewmeyer: Yeah, that's probably a good idea.

ops/model.py
tonyandrewmeyer: Hmm, the [Juju docs](https://juju.is/docs/juju/hook-tool#open-port) have a "Requires Juju 3.1+" note, but then the description does talk about how the behaviour was different in 2.9.

Should I check this with the Juju team, or are you confident that the note in the Juju docs should be fixed, and ours should just have a note about K8s charms getting this later and using jujuversion as a guard?

ops/charm.py
tonyandrewmeyer: I was thinking that it's better to give charmers as much notice as possible. What if the text was "Scheduled for removal in Juju 4.0" for now, and changed to the simpler "Removed in" once 4.0 makes it out of beta?

docs/custom_conf.py
tonyandrewmeyer: The code would then be able to be roughly what's in this [earlier commit](https://github.com/canonical/operator/pull/1311/commits/5476a0c00c3b9fc27f7e5b00ebeeb9f2018b23bb).

Then we'd need a custom CSS file in `docs/.sphinx/_static`, which I think could just have something like:

```css
div.jujuversion {
    border-color: #38a84d;
}
div.jujuversion p {
    margin-bottom: .125rem;
    margin-top: .125rem;
}
div.jujuversion .versionmodified {
    color: #21632c;
}
```

And the CSS would be added to the `custom_html_css_files` variable in `custom_conf.py`.

Our current method of pulling in changes from the starter pack involves replacing the `.sphinx` folder entirely, so we'd need to handle that somehow (either instructions in `HACKING.md`, or having tox copy the file there from somewhere safer). I'm hopeful that the moves towards [using an extension](https://discourse.canonical.com/t/this-month-in-documentation-july-2024/4045#starter-pack) will make everything much simpler for us in that regard, so perhaps this is only a short-term issue.

I'm not sure which I prefer (I don't really love either). Thoughts?

ops/charm.py
benhoyt: Fair enough. I like the "Scheduled for removal in Juju 4.0" -- that's nice.

docs/custom_conf.py
benhoyt: > Our current method of pulling in changes from the starter pack involves replacing the .sphinx folder entirely

Okay, given that, let's stick with what you have.

ops/model.py
tonyandrewmeyer: [Matrix](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$v_MYS-BDWb0AIUPClA64eWPEBsZ58zIQEzKSCQ1V5tU?via=ubuntu.com&via=matrix.org&via=matrix.debian.social)

benhoyt (approved):

docs/custom_conf.py
```diff
@@ -334,3 +330,85 @@ def _compute_navigation_tree(context):
     ('py:class', 'ops.testing.CharmType'),
     ('py:obj', 'ops.testing.CharmType'),
 ]
+
+
+# This is very strongly based on
```
dimaqq, line 335: I guess this approach is recommended at https://github.com/picnixz/sphinx/blob/8c6d234e961eaf945567755863d377fa27dc6f22/doc/development/tutorials/examples/todo.py#L41

dimaqq (approved):

PR #1313, opened 2024-08-13
feat: parse JUJU_* env in one place

IronCore864:
Parse all `JUJU_*` env vars in one place.

Closes https://github.com/canonical/operator/issues/1075.

---

Note: This draft does the minimum changes by creating a `dataclass`, no other logic in `main.py` is changed. Variables are not grouped into smaller `dataclasses`, and all types are string just as they were in `os.environ.get()`.

There will be another alternative draft with variable grouping and typing.

---

### 1 Grouping

I think it's fine without it but I'm not against it either. If there is a solid reason for grouping, I'm OK with it, but here are my thoughts:

1. Value added: It seems only `JUJU_NOTICE_*` and `JUJU_SECRET_*` can be grouped into a smaller dataclass. Maybe we can argue that `JUJU_RELATION` and `JUJU_RELATION_ID` can be grouped too, but then the first needs to be renamed to `JUJU_RELATION_NAME` which deviates from the acutal env variable name. So, most env vars can't really be grouped, which means the `_JujuContext` dataclass can't be shortened or improved readability-wise by grouping. And when using it, it's not too much difference, comparing `juju_context_obj.juju_notice_id` and `juju_context_obj.juju_notice.id`. So I think the value added by grouping is insignificant.
2. Logic: Without grouping, all variables are on the same level in parallel, similar to each other. If we group like `_JujuContext.notice.id` and `_JujuContext.juju_charm_dir`, it might indicate that `_JujuContext` owns `notice` and `juju_charm_dir`, they are on the same level, and there is another object that is `notice` who owns `id`, while in fact it's not the case, they are all env vars.
3. Last but not least, the source comes from [here](https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1394) where it's just a list of key/value pairs with no grouping, so I think maybe it's better to stay the same.

### 2 Type

I put all fields as str in this draft, because:

1. I wanted to put some fields as int in the first place but after investigating, only very few fields are actually non-str, like [`JUJU_SECRET_REVISION`](https://github.com/canonical/operator/blob/main/ops/main.py#L180); and some values are processed based on the env var, like [juju relation id](https://github.com/canonical/operator/blob/main/ops/main.py#L151). So, maybe it's easier to keep them as they are and handle the type conversion whenever needed (mostly aren't). There are on extra plus side and that is we don't need to do extra processing if the key is `JUJU_SECRET_REVISION`, keeping the `from_env` method simple without key-name related logic.
2. Since the whole `_JujuContext` dataclass comes from env vars, it's intuitive to think all fields are strings.
3. The [source](https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1394) is all strings.

Some explanation on `Optional` or not, and default values, becaues it may seem strange at first glace, and these logic is not from Juju but from `ops.main`, how we are using them currently:

1. `Optional`: Some fields are not `Optional` because in `main.py` the current usage of them is treating them like they surely exist (`os.environ[key]`), so I kept them as type `str` instead of `Optional[str]`.
2. Some fields must be `Optional` with None value because in the current logic there are tests like `if key in os.environ`.
3. Some fields are used as args to pass to other things and they can be `None` so I left these type all as `Optional` that can be `None`.


tonyandrewmeyer:
For what it's worth, I agree that it's not worth grouping (flat is better than nested). However, I do think it's worth doing the type conversion in the class, as well as any `split` type logic.

I'd also suggest that the `juju_` prefix for the attribute names isn't needed, since it's implied by the fact that it's in `_JujuContext`.

dimaqq:
Consider

```
	// relationId identifies the relation for which a relation hook is
	// executing. If it is -1, the context is not running a relation hook;
	// otherwise, its value must be a valid key into the relations map.
	relationId int
```

If we're going to cast or interpret the values, we gotta handle these special cases too. In this case, I think `-1 ==> None`.

ops/main.py
```diff
@@ -42,6 +43,38 @@
...
+@dataclasses.dataclass(frozen=True)
+class _JujuContext:
+    # source: https://juju.is/docs/juju/charm-environment-variables
+    # the HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    juju_relation: str = ''
+    juju_relation_id: str = ''
+    juju_secret_revision: str = ''
+    juju_workload_name: str = ''
```
benhoyt, line 55: Why are these `str = ''` and the others `Optional[str] = None`? I would think we should use `Optional[str] = none` for all.

ops/main.py
```diff
@@ -42,6 +43,38 @@
 logger = logging.getLogger()
 
 
+@dataclasses.dataclass(frozen=True)
+class _JujuContext:
+    # source: https://juju.is/docs/juju/charm-environment-variables
+    # the HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    juju_relation: str = ''
```
benhoyt, line 52: I think this should be called `relation_name` for clarity, even though the env var is `JUJU_RELATION`.

Also, I think Tony mentioned it already somewhere, but let's drop the `juju_` prefix, as that's true of all of them (and this is the "juju context" object).

ops/main.py
```diff
@@ -42,6 +43,38 @@
...
+    juju_relation_id: str = ''
```
benhoyt, line 53: And I think this one should be `relation_id: Optional[int] = None`, that is, the parsed and converted version.

ops/main.py
```diff
@@ -42,6 +43,38 @@
...
+    juju_notice_type: Optional[str] = None
+    juju_pebble_check_name: Optional[str] = None
+    juju_remote_app: Optional[str] = None
+    juju_remote_unit: Optional[str] = None
+    juju_secret_id: Optional[str] = None
+    juju_secret_label: Optional[str] = None
+    juju_storage_id: Optional[str] = ''
+
+    @classmethod
+    def from_environ(cls) -> '_JujuContext':
```
benhoyt, line 73: I think it'd be nicer (and easier to test) if `from_environ` took a mapping, so you could call it with a custom dict, or with `os.environ`.

ops/main.py
```diff
@@ -42,6 +43,38 @@
...
+        kwargs: Dict[str, Any] = {name: os.getenv(name.upper()) for name in cls.__annotations__}
```
benhoyt, line 74: This is clever, but we'll probably want to go with just custom code like you've done in #1314 as we may want to customise some of the names and types.

IronCore864:
After discussion, we will use a flat structure without nested objects, and we will put all fields as `Optional[str]` or `Optional[int]`.

IronCore864:
Notes:

1. `_JujuContext.from_environ(dict(os.environ))` looks weird, because strictly speaking, it's not `from_environ` but `from_dict`. The name `from_environ` is used to be consistent with a few other usages in the code base; a `Dict` is used instead of `os._Environ` per previous code review suggestions to make it easy to be mocked in tests.
2. Where to put `_JujuContext`: I think it makes sense to put it in `main.py` because it's first used there. But no matter where I put it, I can't bypass circular imports in `jujuversion.py`, so I left it where it's most logical and used `TYPE_CHECKING` to fix the import for typing checks. However, this breaks sphinx doc build. _How can we solve this?_

Changes:


`_JujuContext`:

- Every field is of type `Optional[str]` or `Optional[int]` and defaults to `None`, according to the code review and our daily discussion.
- The only two int-type fields are `relation_id` and `secret_revision`, where I did not handle the case that the env var's value exists but is not a number. Should we handle this case? Or let it throw a `ValueError` is OK?
- A few more env vars used in `model.py` and `JujuVersion` are added, except `JUJU_DEBUG_AT` (used in `framework.py`) and `OPERATOR_DISPATCH` (used in `main.py`), because they are not in the source [here](https://juju.is/docs/juju/charm-environment-variables) and [here](https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398).
- `_JujuContext` is instantiated only once in `_Manager`, then passed to `_Dispatcher` and `_ModelBackend`, the latter also passes it to `_Container`. These classes will pass it to `JujuVersion` too.

`JujuVersion`:

- The original `from_environ` method is renamed to `from_context` because it's not appropriate any more. `from_juju_context` was also considered, but `JujuVersion.from_juju_context(juju_context)` seemed too verbose and duplicated, hence `JujuVersion.from_context(juju_context)`.
- The original `if v is None: v = '0.0.0` logic is removed; default value is handled now in `_JujuContext`.


`model.py`:

- There were a few `os.environ` / `os.getenv` in it, all refactored to use `_JujuContext`.

`test/*`:

- Add tests for `JujuVersion.from_context`.
- Refactor tests to make them pass.
- Add tests for `_JujuContext.from_environ`.

ops/model.py
```diff
@@ -3250,7 +3256,7 @@ def relation_get(
             raise TypeError('is_app parameter to relation_get must be a boolean')
 
         if is_app:
-            version = JujuVersion.from_environ()
+            version = JujuVersion.from_context(self._juju_context)
```
benhoyt, line 3259: Maybe we should parse and store `_juju_version` once in the constructor?

ops/model.py
```diff
@@ -3141,19 +3142,21 @@ class _ModelBackend:
 
     def __init__(
         self,
+        juju_context: 'ops.main._JujuContext',
```
benhoyt, line 3145: @tonyandrewmeyer Will this signature change break ops-scenario? I don't really understand how this subclass can call `super().__init__()` without args here anyway: https://github.com/canonical/ops-scenario/blob/602f9b9ce0321ed383adc9f6de28047c3a4afac6/scenario/mocking.py#L101-L113

I'm not saying we shouldn't do it even if it does break Scenario, but we should think it through at any rate.

We could even make `juju_context` the final parameter, and make it optional, and if not provided this uses `_JujuContext` with `os.environ` by default. That would avoid lots of changes to our tests. Not as explicit, but seems a reasonable default.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    unit_name: Optional[str] = None
+
+    # env var JUJU_VERSION
+    version: Optional[str] = None
+
+    # env var JUJU_WORKLOAD_NAME
+    workload_name: Optional[str] = None
+
+    @classmethod
+    def from_environ(cls, env: Dict[str, Any]) -> '_JujuContext':
```
benhoyt, line 122: Yeah, I think you're right this should now be `from_dict`. However, can we make the type a Mapping so that we don't need `dict()` when passing `os.environ`?

ops/main.py
```diff
@@ -42,6 +43,114 @@
 logger = logging.getLogger()
 
 
+@dataclasses.dataclass(frozen=True)
+class _JujuContext:
+    # source: https://juju.is/docs/juju/charm-environment-variables
+    # the HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    # env var JUJU_ACTION_NAME
```
benhoyt, line 52: I'd prefer something a bit more descriptive for these. Perhaps follow a pattern like:

```
# The action's name, for example 'backup' (from JUJU_ACTION_NAME).
```

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    # env var JUJU_VERSION
+    version: Optional[str] = None
+
+    # env var JUJU_WORKLOAD_NAME
+    workload_name: Optional[str] = None
+
+    @classmethod
+    def from_environ(cls, env: Dict[str, Any]) -> '_JujuContext':
+        return _JujuContext(
+            action_name=env.get('JUJU_ACTION_NAME'),
```
benhoyt, line 124: For most of these should we use `env.get('JUJU_ACTION_NAME') or None` in case it's set but empty string?

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+            model_name=env.get('JUJU_MODEL_NAME'),
+            model_uuid=env.get('JUJU_MODEL_UUID'),
+            notice_id=env.get('JUJU_NOTICE_ID'),
+            notice_key=env.get('JUJU_NOTICE_KEY'),
+            notice_type=env.get('JUJU_NOTICE_TYPE'),
+            pebble_check_name=env.get('JUJU_PEBBLE_CHECK_NAME'),
+            relation_name=env.get('JUJU_RELATION'),
+            relation_id=int(env.get('JUJU_RELATION_ID', '').split(':')[-1])
+            if 'JUJU_RELATION_ID' in env
+            else None,
```
benhoyt, line 139: Can we add some parentheses to improve the formatting here? Also, similar to above, might be best to use `if env.get('JUJU_RELATION_ID')` in case it's set but empty string.
```suggestion
            relation_id=(int(env['JUJU_RELATION_ID'].split(':')[-1])
                         if env.get('JUJU_RELATION_ID')
                         else None),
```

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+            remote_app=env.get('JUJU_REMOTE_APP'),
+            remote_unit=env.get('JUJU_REMOTE_UNIT'),
+            secret_id=env.get('JUJU_SECRET_ID'),
+            secret_label=env.get('JUJU_SECRET_LABEL'),
+            secret_revision=int(env.get('JUJU_SECRET_REVISION', ''))
+            if 'JUJU_SECRET_REVISION' in env
+            else None,
+            storage_id=env.get('JUJU_STORAGE_ID'),
+            unit_name=env.get('JUJU_UNIT_NAME'),
+            version=env.get('JUJU_VERSION', '0.0.0'),
```
benhoyt, line 149: We should probably not have `0.0.0` as the default here, but use None correctly here. Can do that in `JujuVersion` or where it's used if we need to.

benhoyt (changes requested):
This looks like a great start, thanks. A few initial comments.

ops/model.py
tonyandrewmeyer: > @tonyandrewmeyer Will this signature change break ops-scenario?

It would break Scenario, yes. It wouldn't be too arduous to release a 6.x version that can handle it either way, and 7.0 could require ops 2.16 (it already needs ops 2.15).

> I don't really understand how this subclass can call `super().__init__()` without args here anyway: https://github.com/canonical/ops-scenario/blob/602f9b9ce0321ed383adc9f6de28047c3a4afac6/scenario/mocking.py#L101-L113

The other args have defaults :). I haven't checked, but I think that Scenario isn't passing those things through because it's handling the mocking in some other way.

ops/jujuversion.py
```diff
@@ -17,7 +17,10 @@
 import os
 import re
 from functools import total_ordering
-from typing import Union
+from typing import TYPE_CHECKING, Union
+
+if TYPE_CHECKING:
+    from ops.main import _JujuContext
```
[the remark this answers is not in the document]
IronCore864: If we remove `from_context`, circular imports won't exist.

However, if we want to add a docstring in `JujuVersion.from_environ` to let users know that it's better to use `JujuVersion(context.version)` where `context` is an instance of `_JujuContext`, sphinx doc build will still fail with a warning about not knowing what `_JujuContext` is, and in that way circular imports still exist.

Shall we not add this line of docstring, or move `_JujuContext` to `_private/jujucontext.py`?

ops/jujuversion.py
```diff
@@ -100,12 +103,21 @@ def __lt__(self, other: Union[str, 'JujuVersion']) -> bool:
 
     @classmethod
     def from_environ(cls) -> 'JujuVersion':
-        """Build a version from the ``JUJU_VERSION`` environment variable."""
+        """Build a version from the ``JUJU_VERSION`` environment variable.
+
+        Kept for backward compatibility. Use :meth:`from_context` instead.
```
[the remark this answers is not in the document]
IronCore864: Ah, this answers my previous question, too.

ops/model.py
```diff
@@ -2211,6 +2211,7 @@ def __init__(
         self, name: str, backend: '_ModelBackend', pebble_client: Optional[pebble.Client] = None
     ):
         self.name = name
+        self._juju_context = backend._juju_context
```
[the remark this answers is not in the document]
IronCore864: Ah, now I get it. I misunderstood it in the first place; I found that there was already a check on `supports_exec_service_context ` in `Container.exec`, so I thought it was already resolved.

How about storing `self.juju_version` in the init instead of just `self. supports_exec_service_context`? Because if the service context is not supported, the `RuntimeError` raised in `exec` would need to contain the version anyway. Or shall we remove the version from the warning?

ops/model.py
IronCore864: Yes, it makes sense because there are two occurrences of it already. Will do it.

ops/model.py
IronCore864: Yes, it's because other args are optional with defaults, I saw it in a bunch of tests where only unit_name is passed in.

Per the above discussion, I think @benhoyt's suggestion makes sense: I will make `juju_context` the final optional parameter and revert the test changes.

dimaqq: Let's make it keyword-only argument then

test/test_model.py
```diff
@@ -1105,7 +1111,13 @@ def assertBackendCalls(  # noqa: N802
...
 
     def test_run_error(self, fake_script: FakeScript):
-        model = ops.Model(ops.CharmMeta(), _ModelBackend('myapp/0'))
+        model = ops.Model(
+            ops.CharmMeta(),
+            _ModelBackend(
+                juju_context=ops.main._JujuContext.from_environ(dict(os.environ)),
+                unit_name='myapp/0',
+            ),
+        )
```
dimaqq, line 1120: This feels like a code smell.

If the test expects specific values in the environment, let's pass these values here, like `_ModelBackend(juju_context=_JujuContext("foo", "bar", other="baz"))`.

If the test doesn't expect specific values, it's better to supply a blank context.

If the test depends parent process environment, that's probably a bug.

IronCore864: I agree; this is solved by the previous comment from Ben, though. We are making `juju_context` optional with a default value coming from `os.environ`.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    secret_revision: Optional[int] = None
+
+    # env var JUJU_STORAGE_ID
+    storage_id: Optional[str] = None
+
+    # env var JUJU_UNIT_NAME
+    unit_name: Optional[str] = None
+
+    # env var JUJU_VERSION
+    version: Optional[str] = None
```
tonyandrewmeyer, line 116: What if this was actually `version: JujuVersion`? Or `version: Optional[JujuVersion] = None` if it can be missing, which I wouldn't think was actually the case?

That would avoid the import issues but also it should reduce the number of times `JujuVersion` needs to get instantiated. There would be `juju_context.version.supports_exec_service_context()` type calls instead.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    departing_unit: Optional[str] = None
+
+    # env var JUJU_DISPATCH_PATH
+    dispatch_path: Optional[str] = None
+
+    # env var JUJU_MODEL_NAME
+    model_name: Optional[str] = None
+
+    # env var JUJU_MODEL_UUID
+    model_uuid: Optional[str] = None
```
tonyandrewmeyer, line 74: Can these actually be missing in Juju 2.9+?

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    storage_id: Optional[str] = None
+
+    # env var JUJU_UNIT_NAME
+    unit_name: Optional[str] = None
+
+    # env var JUJU_VERSION
+    version: Optional[str] = None
+
+    # env var JUJU_WORKLOAD_NAME
+    workload_name: Optional[str] = None
```
tonyandrewmeyer, line 119: This is the one other attribute where alphabetical sorting doesn't put it with its friends (notices, check). But I'm not sure `pebble_workload_name` is a good choice.

ops/model.py
tonyandrewmeyer: I think it probably ought to be keyword-only (there seems to be no logical order here) but note that will also break anything using this. It'd be a pretty straightforward update for Scenario, but this is a public class, even though we say not to instantiate it directly in the docstring.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    action_name: Optional[str] = None
+
+    # env var JUJU_ACTION_UUID
+    action_uuid: Optional[str] = None
+
+    # env var JUJU_CHARM_DIR
+    charm_dir: Optional[str] = None
+
+    # env var JUJU_DEBUG
+    debug: Optional[str] = None
```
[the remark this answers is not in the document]
IronCore864: Previously the logic was [`debug = 'JUJU_DEBUG' in os.environ`](https://github.com/canonical/operator/blob/main/ops/main.py#L440C9-L440C43), it doesn't matter what value it is, as long as it's set, it's in debug mode. But I agree a bool defaults to false makes more sense, I will update it.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+
+    # env var JUJU_DEPARTING_UNIT
+    departing_unit: Optional[str] = None
```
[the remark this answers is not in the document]
IronCore864: `departing_unit_name` seems much better, and although `relation_departing_unit_name` is too long, after sorting the lines it sits nicely with other relation stuff. I will use the latter.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    # env var JUJU_ACTION_NAME
+    action_name: Optional[str] = None
+
+    # env var JUJU_ACTION_UUID
+    action_uuid: Optional[str] = None
+
+    # env var JUJU_CHARM_DIR
+    charm_dir: Optional[str] = None
```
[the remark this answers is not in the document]
IronCore864: Ah yes, this would be much better, since `context.charm_dir` is only used in that function. I can make this change.

After some investigation, I figured that if we remove `_get_charm_dir`, it will break scenario [here](https://github.com/canonical/ops-scenario/blob/main/scenario/ops_main_mock.py#L18). But I think it's still reasonable to remove `_get_charm_dir`, right?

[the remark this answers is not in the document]
tonyandrewmeyer: > After some investigation, I figured that if we remove `_get_charm_dir`, it will break scenario [here](https://github.com/canonical/ops-scenario/blob/main/scenario/ops_main_mock.py#L18). But I think it's still reasonable to remove `_get_charm_dir`, right?

I think it's fine. We can open PRs to adjust Scenario before the next ops release.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+
+    # env var JUJU_DEBUG
+    debug: Optional[str] = None
+
+    # env var JUJU_DEPARTING_UNIT
+    departing_unit: Optional[str] = None
+
+    # env var JUJU_DISPATCH_PATH
+    dispatch_path: Optional[str] = None
```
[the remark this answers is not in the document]
IronCore864: It's added [here](https://github.com/juju/juju/blame/8ad001fadb526945445506026aea5c828f9320d7/worker/uniter/runner/runner.go#L471) and from 2.7.4+ on it can't be None.

ops/main.py
```diff
@@ -42,6 +43,114 @@
...
+    secret_id: Optional[str] = None
+
+    # env var JUJU_SECRET_LABEL
+    secret_label: Optional[str] = None
+
+    # env var JUJU_SECRET_REVISION
+    secret_revision: Optional[int] = None
+
+    # env var JUJU_STORAGE_ID
+    storage_id: Optional[str] = None
```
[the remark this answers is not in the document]
IronCore864: Yes, the id is not used. I will change it to `storage_name` only.

ops/main.py
IronCore864: I'll leave it like this then, putting `workload_name` next to `pebble_check_name` seems weird since the whole fields are sorted (although we don't have to, since [the source](https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398) isn't sorted anyways.)

ops/main.py
IronCore864: I think it can't according to [here](https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398). I will update it and others from the source.

IronCore864:
db-charm-tests failed because [the charm imports JujuVersion from `ops.model`](https://github.com/canonical/postgresql-k8s-operator/blob/main/src/charm.py#L63) instead of `ops.jujuversion`.

ops/model.py
IronCore864: I've made it:

```python
def __init__(
    self,
    unit_name: Optional[str] = None,
    model_name: Optional[str] = None,
    model_uuid: Optional[str] = None,
    juju_context: Optional['ops.main._JujuContext'] = None,
):
```

and tested it, it did not break things. Making it as keyword-only also works, as in:

```python
def __init__(
    self,
    unit_name: Optional[str] = None,
    model_name: Optional[str] = None,
    model_uuid: Optional[str] = None,
    *,
    juju_context: Optional['ops.main._JujuContext'] = None,
):
```

But since the first version didn't break anything, I kept the first version for now. Feel free to reopen this conversation if I missed something.

ops/jujuversion.py
```diff
@@ -39,7 +39,10 @@ class JujuVersion:
         re.VERBOSE,
     )
 
-    def __init__(self, version: str):
+    def __init__(self, version: Union[str, None]):
+        if version is None:
+            version = '0.0.0'
+
```
tonyandrewmeyer, line 45: I think there was some discussion about this before, but there are a lot of resolved conversations to go through to find it, so apologies if this has already been discussed & decided.

It seems to me that when using Juju there is always a version, so the only time it's missing is when using Harness. So it would be better to have the version required here and the ugly `0.0.0` code in ops.testing.

tonyandrewmeyer (approved):
Looking good! A few small comments.

I tried this out on Juju 3.6, LXD, but only with a few events, not exhaustively. Everything looked good for the ones I tried.

ops/main.py
```diff
@@ -334,7 +455,8 @@ def _init_dispatch(self):
         JUJU_DISPATCH_PATH will be set to the wanted hook, e.g. hooks/install,
         in both cases.
         """
-        self._dispatch_path = Path(os.environ['JUJU_DISPATCH_PATH'])
+        assert self._juju_context.dispatch_path is not None
+        self._dispatch_path = Path(self._juju_context.dispatch_path)
```
benhoyt, line 459: I don't think this will be needed anymore once `_JujuContext.dispatch_path` is non-optional.

benhoyt (changes requested):
Getting close. Some minor nits, but one bigger question about whether we should move `_JujuContext` to a new `jujucontext.py` file.

ops/main.py
```diff
@@ -42,6 +43,134 @@
 logger = logging.getLogger()
 
 
+@dataclasses.dataclass(frozen=True)
```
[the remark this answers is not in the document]
tonyandrewmeyer: `init=False` would make that worse, no? Because then it will have `object.__init__` that takes no arguments, so you'd always have the defaults. Plus I think this would break the creation of the object in the classmethod, no?

The proper fix, imo, is that anything that isn't optional should either be fixed in `__post_init__` or (better) should have a `dataclasses.field` with an appropriate default factory.

ops/model.py
```diff
@@ -3144,16 +3144,21 @@ def __init__(
         unit_name: Optional[str] = None,
         model_name: Optional[str] = None,
         model_uuid: Optional[str] = None,
+        juju_context: Optional['ops.main._JujuContext'] = None,
     ):
+        self._juju_context = (
+            juju_context if juju_context else ops.main._JujuContext.from_dict(os.environ)
+        )
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yeah, on balance I think I agree that moving (both) would be best. If this was just specifying the type, then I think I'd rather leave it alone, but since we're allowing people to pass `None` and have it populated here I don't like the circle either.

ops/main.py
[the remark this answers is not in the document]
dimaqq: I thought we would like people to instantiate partial _JujuContext in tests.

At least in the tests for ops itself.

benhoyt:
Oh, one more thing I noticed, this use of `JUJU_DEBUG_AT` in framework.py:

```python
        # Parse the env var once, which may be used multiple times later
        debug_at = os.environ.get('JUJU_DEBUG_AT')
        if debug_at:
            self._juju_debug_at = {x.strip() for x in debug_at.split(',')}
        else:
            self._juju_debug_at: Set[str] = set()
```

Should this go in `_JujuContext` too?

ops/main.py
```diff
@@ -42,6 +43,134 @@
...
+
+    # The storage name, for example 'my-storage'
+    # if the original environment variable's value is 'my-storage/1' (from JUJU_STORAGE_ID).
+    storage_name: Optional[str] = None
+
+    # The name of the unit, for example 'myapp/0' (from JUJU_UNIT_NAME)
+    unit_name: str = ''
+
+    # The version of Juju, for example '3.4.0' (from JUJU_VERSION)
+    version: Optional[JujuVersion] = None
```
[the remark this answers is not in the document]
IronCore864: No, it can't be None; I only make it optional with a default None value so that it can be put in the right place alphabetically. And if I give it a default value like `JujuVersion('0.0.0')`, ruff would complain about it. Shall I make it non-optional without default value and put it as the first attribute of the class?

ops/main.py
```diff
@@ -42,6 +43,134 @@
...
+
+    # The action's uuid, for example '1' (from JUJU_ACTION_UUID).
+    action_uuid: Optional[str] = None
+
+    # The root directory of the charm where it is running,
+    # for example '/var/lib/juju/agents/unit-bare-0/charm' (from JUJU_CHARM_DIR).
+    # If JUJU_CHARM_DIR is None or set to an empty string,
+    # use Path(f'{__file__}/../../..') as default
+    # (assuming the '$JUJU_CHARM_DIR/lib/op/main.py' structure).
+    charm_dir: Path = Path()
```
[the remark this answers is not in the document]
IronCore864: Then it will be the same as `JujuVersion` (see the other comment above), Ruff will complain "Do not perform function call in dataclass defaults" Ruff[RUF009](https://docs.astral.sh/ruff/rules/function-call-in-dataclass-default-argument). Shall we do it?

ops/jujuversion.py
IronCore864: This logic was [already in the codebase](https://github.com/canonical/operator/blob/main/ops/jujuversion.py#L106) and it's there for a reason because if not, many test cases that don't care about JujuVersion hence don't set the `JujuVersion` env var will fail since it will be `None` which can't be matched against the SEMVER.

We can disallow `None` in the `__init__` here, but in `_JujuContext.from_dict`, I'd still pass in something like `version=env.get('JujuVersion', '0.0.0')`, not `version=env.get('JujuVersion', '')`. Is this OK?

ops/model.py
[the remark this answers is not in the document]
IronCore864: Got it, I will add an explicit `None` check, and move both `JujuVersion` and `_JujuContext` into `jujucontext.py`.

ops/main.py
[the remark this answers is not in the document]
tonyandrewmeyer: You'd do it like:

```suggestion
    charm_dir: Path = dataclasses.field(default_factory=lambda: Path(f'{__file__}'/../../..'))
```

Although I haven't checked that that's exactly the right path. But using `default_factory`.

ops/main.py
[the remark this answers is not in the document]
tonyandrewmeyer: You should definitely be able to do:

```suggestion
    version: JujuVersion = dataclasses.field(default_factory=lambda: JujuVersion('0.0.0'))
```

I still don't love the "0.0.0" version, but I suppose it's ok as a "doesn't have any of the new features" default, and probably better than forcing people to provide a version every time, when most of the time they probably don't care.

ops/jujuversion.py
tonyandrewmeyer: To me, it seems cleaner to have code that's only needed for tests in the testing code, not in the main code. (Yes, this is a change from before, but it's a positive change, in my opinion).

It makes sense for Harness to have a default version, just like it has a default unit name, app name, model UUID, and so on. We also need to do this for backwards compatibility with people's tests.

Not all tests do version comparisons, so it doesn't take much to do this change. This is very much just proof-of-concept not polished code, but with this diff the tests pass:

<details>
<summary>Rough diff for no default version outside testing</summary>

```diff
diff --git a/ops/jujuversion.py b/ops/jujuversion.py
index 1ddb3fc..8e25b3e 100644
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -39,10 +39,7 @@ class JujuVersion:
         re.VERBOSE,
     )
 
-    def __init__(self, version: Union[str, None]):
-        if version is None:
-            version = '0.0.0'
-
+    def __init__(self, version: str):
         m = self._pattern_re.match(version)
         if not m:
             raise RuntimeError(f'"{version}" is not a valid Juju version string')
diff --git a/ops/testing.py b/ops/testing.py
index cfd1282..2c135bd 100644
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -259,6 +259,8 @@ class Harness(Generic[CharmType]):
         actions: Optional[YAMLStringOrFile] = None,
         config: Optional[YAMLStringOrFile] = None,
     ):
+        if "JUJU_VERSION" not in os.environ:
+            os.environ["JUJU_VERSION"] = "0.0.0"
         self._juju_context = main._JujuContext.from_dict(os.environ)
         self._charm_cls = charm_cls
         self._charm: Optional[CharmType] = None
diff --git a/test/test_framework.py b/test/test_framework.py
index 7be174a..9d45ffd 100644
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -37,6 +37,8 @@ from test.test_helpers import FakeScript
 
 def create_model():
     """Create a Model object."""
+    if 'JUJU_VERSION' not in os.environ:
+        os.environ['JUJU_VERSION'] = '0.0.0'
     backend = _ModelBackend(unit_name='myapp/0')
     meta = ops.CharmMeta()
     model = ops.Model(meta, backend)
diff --git a/test/test_helpers.py b/test/test_helpers.py
index 84f9b20..505a7da 100644
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -88,6 +88,7 @@ def create_framework(
         os.environ['PATH'],
     ])
     os.environ['JUJU_UNIT_NAME'] = 'local/0'
+    os.environ['JUJU_VERSION'] = '0.0.0'
 
     tmpdir = pathlib.Path(tempfile.mkdtemp())
 
diff --git a/test/test_main.py b/test/test_main.py
index 3bba892..605914b 100644
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -1492,28 +1492,28 @@ class TestStorageHeuristics:
 
 class TestJujuContext:
     def test_both_str_and_int_fields_default_to_none(self):
-        juju_context = ops.main._JujuContext.from_dict({})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0'})
         assert juju_context.action_name is None
         assert juju_context.relation_id is None
 
     def test_parsing_int_fields(self):
-        juju_context = ops.main._JujuContext.from_dict({'JUJU_RELATION_ID': 'x:42'})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0', 'JUJU_RELATION_ID': 'x:42'})
         assert juju_context.relation_id == 42
 
     def test_parsing_secret_revision_as_int(self):
-        juju_context = ops.main._JujuContext.from_dict({'JUJU_SECRET_REVISION': '42'})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0', 'JUJU_SECRET_REVISION': '42'})
         assert juju_context.secret_revision == 42
 
     def test_parsing_juju_debug_as_bool(self):
-        juju_context = ops.main._JujuContext.from_dict({'JUJU_DEBUG': 'true'})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0', 'JUJU_DEBUG': 'true'})
         assert juju_context.debug is True
 
     def test_parsing_juju_charm_dir(self):
-        juju_context = ops.main._JujuContext.from_dict({'JUJU_CHARM_DIR': '/dir'})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0', 'JUJU_CHARM_DIR': '/dir'})
         assert juju_context.charm_dir == Path('/dir')
 
     def test_parsing_juju_charm_dir_not_set(self):
-        juju_context = ops.main._JujuContext.from_dict({})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0'})
         assert juju_context.charm_dir == Path(f'{__file__}/../../..').resolve()
 
     def test_parsing_juju_version(self):
@@ -1521,7 +1521,7 @@ class TestJujuContext:
         assert juju_context.version == JujuVersion('3.4.0')
 
     def test_parsing_storage_id_to_name(self):
-        juju_context = ops.main._JujuContext.from_dict({'JUJU_STORAGE_ID': 'my-storage/1'})
+        juju_context = ops.main._JujuContext.from_dict({'JUJU_VERSION': '0.0.0', 'JUJU_STORAGE_ID': 'my-storage/1'})
         assert juju_context.storage_name == 'my-storage'
 
     def test_parsing_all_str_fields(self):
@@ -1546,6 +1546,7 @@ class TestJujuContext:
         }
 
         env = {key: 'foo' for key in env_var_attr_mapping}
+        env['JUJU_VERSION'] = '0.0.0'
 
         juju_context = ops.main._JujuContext.from_dict(env)
```

[/details]

IronCore864:
> Oh, one more thing I noticed, this use of `JUJU_DEBUG_AT` in framework.py:
> 
> ```python
>         # Parse the env var once, which may be used multiple times later
>         debug_at = os.environ.get('JUJU_DEBUG_AT')
>         if debug_at:
>             self._juju_debug_at = {x.strip() for x in debug_at.split(',')}
>         else:
>             self._juju_debug_at: Set[str] = set()
> ```
> 
> Should this go in `_JujuContext` too?

I noticed this in the beginning but didn't add it, because `_JujuContext` is things from the source [here](https://juju.is/docs/juju/charm-environment-variables) but `JUJU_DEBUG_AT` isn't one of them.

Should I add it to `_JujuContext` anyway?

ops/main.py
```diff
@@ -42,6 +43,134 @@
...
+            notice_type=env.get('JUJU_NOTICE_TYPE') or None,
+            pebble_check_name=env.get('JUJU_PEBBLE_CHECK_NAME') or None,
+            relation_departing_unit_name=env.get('JUJU_DEPARTING_UNIT') or None,
+            relation_name=env.get('JUJU_RELATION') or None,
+            relation_id=(
+                int(env['JUJU_RELATION_ID'].split(':')[-1])
+                if env.get('JUJU_RELATION_ID')
+                else None
+            ),
+            remote_app=env.get('JUJU_REMOTE_APP') or None,
```
gboutry, line 156: `env.get(...) or None` is redundant since `dict.get` will return None is there's no key.

ops/main.py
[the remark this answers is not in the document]
IronCore864: > I thought we would like people to instantiate partial _JujuContext in tests.
> 
> At least in the tests for ops itself.

In the test, we can create a mocked dictionary then `_JujuContext.from_dict(d)`, it's even easier than monkey patching, and I think this is why Ben suggested passing a dict to `from_dict` in the first place.

ops/main.py
```diff
@@ -42,6 +43,134 @@
...
+            secret_revision=(
+                int(env['JUJU_SECRET_REVISION']) if env.get('JUJU_SECRET_REVISION') else None
+            ),
+            storage_name=(
+                env.get('JUJU_STORAGE_ID', '').split('/')[0]
+                if env.get('JUJU_STORAGE_ID')
+                else None
+            ),
+            unit_name=env.get('JUJU_UNIT_NAME', ''),
+            version=JujuVersion(env.get('JUJU_VERSION')),
```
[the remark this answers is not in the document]
IronCore864: See the previous comment, because there was already a "if None then defaults to 0.0.0" logic in JujuVersion, if we do this and disallow `None` in `JujuVersion`, it will break many test cases.

ops/main.py
IronCore864: It mainly handles the corner case where the key is set to an empty str, I learned this trick from Ben :)

```python
>>> import os
>>> os.environ['foo'] = ''
>>> foo = os.getenv('foo')
>>> foo
''
>>> assert foo is None
AssertionError
```

gboutry: Oh, makes sense then :)

ops/jujuversion.py
```diff
@@ -14,133 +14,7 @@
...
-        """Report whether this Juju version supports exec's service_context option."""
-        if (self.major, self.minor, self.patch) < (3, 1, 6):
-            # First released in 3.1.6
-            return False
-        if (self.major, self.minor, self.patch) == (3, 2, 0):
-            # 3.2.0 was released before Pebble was updated, but all other 3.2
-            # releases have the change (3.2.1 tag was never released).
-            return False
-        return True
+# For backward compatibility only; please use `from ops import JujuVersion`.
```
benhoyt, line 19: Now that I realize we need to keep this file anyway, I'm having second thoughts about moving JujuVersion to jujucontext.py -- it's confusing. Let's move JujuVersion back here, but keep `_JujuContext` in jujucontext.py. (We would still import `JujuVersion` in jujucontext.py, but not the other way around.)

Sorry to make busy-work!

ops/jujucontext.py
```diff
@@ -0,0 +1,345 @@
...
+class _JujuContext:
+    """_JujuContext collects information from environment variables named 'JUJU_*'.
+
+    Source: https://juju.is/docs/juju/charm-environment-variables.
+    The HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398.
+    Only a subset of the above source, because these are what are used in ops.
+    """
+
+    action_name: Optional[str] = None
+    """The action's name.
```
benhoyt, line 161: Field docstrings look nice, thanks!

ops/jujucontext.py
```diff
@@ -0,0 +1,345 @@
...
+
+    @classmethod
+    def from_dict(cls, env: Mapping[str, Any]) -> '_JujuContext':
+        return _JujuContext(
+            action_name=env.get('JUJU_ACTION_NAME') or None,
+            action_uuid=env.get('JUJU_ACTION_UUID') or None,
+            charm_dir=(
+                Path(f'{__file__}/../../..').resolve()
+                if not env.get('JUJU_CHARM_DIR')
+                else Path(env['JUJU_CHARM_DIR']).resolve()
```
benhoyt, line 313: Let's avoid the not/negative here (simpler, and to match the other similar constructs):

```suggestion
                Path(env['JUJU_CHARM_DIR']).resolve()
                if env.get('JUJU_CHARM_DIR')
                else Path(f'{__file__}/../../..').resolve()
```

ops/model.py
```diff
@@ -55,7 +55,11 @@
 import ops
 import ops.pebble as pebble
 from ops._private import timeconv, yaml
-from ops.jujuversion import JujuVersion
+from ops.jujucontext import JujuVersion, _JujuContext
+
+# JujuVersion is not used in ops.model, but there are charms that are importing JujuVersion
+# from ops.model, so we keep it here.
+_ = JujuVersion
```
benhoyt, line 62: Thanks!

benhoyt (changes requested):
This looks really good. Just a couple more requests for file moving (sorry for the back and forth!).

test/test_main.py
```diff
@@ -1468,19 +1468,102 @@ class TestStorageHeuristics:
...
+            'JUJU_NOTICE_TYPE': 'notice_type',
+            'JUJU_PEBBLE_CHECK_NAME': 'pebble_check_name',
+            'JUJU_DEPARTING_UNIT': 'relation_departing_unit_name',
+            'JUJU_RELATION': 'relation_name',
+            'JUJU_REMOTE_APP': 'remote_app_name',
+            'JUJU_REMOTE_UNIT': 'remote_unit_name',
+            'JUJU_SECRET_ID': 'secret_id',
+            'JUJU_SECRET_LABEL': 'secret_label',
+            'JUJU_UNIT_NAME': 'unit_name',
+            'JUJU_WORKLOAD_NAME': 'workload_name',
```
tonyandrewmeyer, line 1560: I'm a little confused by this - you're creating a dict that has unique values, but then in the test you only use the keys, set every value to "foo" and assert that the context attribute has the value "foo".

If you want to use "foo" for all the values, then this could just be a set rather than a dictionary. Or if you want to have unique values then it seems like you should just use `env_var_attr_mapping` and not create the `env` dict.

benhoyt:
Decision was to put JUJU_DEBUG_AT in `_JujuContext.debug_at`.

test/test_main.py
IronCore864: I intended to maintain an environment variable to _JujuContext attribute name mapping since we renamed a few fields and they are not the same as the env vars. But I agree it is confusing, let me rewrite this test so that it's more readable. Thanks for pointing it out!

ops/jujucontext.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+
+    workload_name: Optional[str] = None
+    """The name of the workload.
+
+    For example 'workload' (from JUJU_WORKLOAD_NAME).
+    """
+
+    @classmethod
+    def from_dict(cls, env: Mapping[str, Any]) -> '_JujuContext':
```
sed-i, line 185: Design question:

The half-baked idea I imagined was that `from_dict` returns a `Union` of types, and the caller always uses a [match](https://peps.python.org/pep-0636/) expression on the result. It would imply py3.10 (ubuntu 22.04) which isn't great, but I wonder (a) if it's a feasible design goal in the first place, and (b) if we should plan the design today with `match` in mind.

I.e. it would be great if this new class brought significant new added value over poking into `os.environ` directly.

sed-i (commented):
So happy to see this PR!

benhoyt (approved):

test/test_framework.py
```diff
@@ -84,7 +89,13 @@ class TestFramework:
     def test_deprecated_init(self, caplog: pytest.LogCaptureFixture):
         # For 0.7, this still works, but it is deprecated.
         with caplog.at_level(logging.WARNING):
-            framework = ops.Framework(':memory:', None, None, None)  # type: ignore
+            framework = ops.Framework(
+                ':memory:',  # type: ignore
+                None,  # type: ignore
+                None,  # type: ignore
+                None,  # type: ignore
+                juju_debug_at=_JujuContext.from_dict({'JUJU_VERSION': '0.0.0'}).debug_at,
```
tonyandrewmeyer, line 97: This seems a long-winded way of saying `set()`? Same below.

No need to change it if you have some reason for preferring this way.

ops/framework.py
```diff
@@ -650,13 +651,9 @@ def __init__(
...
-
-        # Parse the env var once, which may be used multiple times later
-        debug_at = os.environ.get('JUJU_DEBUG_AT')
-        if debug_at:
-            self._juju_debug_at = {x.strip() for x in debug_at.split(',')}
-        else:
-            self._juju_debug_at: Set[str] = set()
+        if juju_debug_at is None:
+            juju_debug_at = _JujuContext.from_dict(os.environ).debug_at
+        self._juju_debug_at = juju_debug_at
```
tonyandrewmeyer, line 656: This is the most backwards compatible way of doing this, and if you prefer then leave it like it is.

I think we could simplify it and just have:

```suggestion
        self._juju_debug_at = juju_debug_at or set()
```

Technically, `Framework` is public and we don't tell people not to instantiate it themselves in the docs, but it seems super unlikely that anyone is doing so, or at least doing so and also expecting debug-at values to be loaded.

What about this?

```suggestion
        if juju_debug_at is None:
            warnings.warn("In the future, a `None` value for `juju_debug_at` will have the same meaning as the empty set", warnings.DeprecationWarning)
            juju_debug_at = _JujuContext.from_dict(os.environ).debug_at
        self._juju_debug_at = juju_debug_at
```

test/test_helpers.py
```diff
@@ -105,7 +107,13 @@ class TestCharmEvents(ops.CharmEvents):
         meta = ops.CharmMeta()
     model = ops.Model(meta, _ModelBackend('local/0'))  # type: ignore
     # We can pass foo_event as event_name because we're not actually testing dispatch.
-    framework = ops.Framework(SQLiteStorage(':memory:'), tmpdir, meta, model)  # type: ignore
+    framework = ops.Framework(
+        SQLiteStorage(':memory:'),
+        tmpdir,
+        meta,
+        model,
+        juju_debug_at=_JujuContext.from_dict(os.environ).debug_at,
```
tonyandrewmeyer, line 115: Same "why not just `set()`" comment as elsewhere.

tonyandrewmeyer (approved):
Looks great!

I'm fine with this as-is, but left a couple of small suggestions - feel free to discard those if you prefer.

ops/framework.py
benhoyt: Yeah, good point. I strongly suspect we could get away with just the `juju_debug_at or set()` version.

ops/jujucontext.py
sed-i: Example: https://stackoverflow.com/a/71519690

IronCore864: If I understand correctly, there is already some similar logic in `main.py`, specifically, in [`_get_event_args`](https://github.com/canonical/operator/blob/main/ops/main.py#L154), where it tries to match the type of the event then pull related environment variables.

I think it would be great if we could somehow integrate it with JujuContext, maybe in the future we can build on top of this PR and implement that.

Regarding the scope of this PR, we had a short discussion today in the daily and we think this PR serves two purposes: 1, parse all ENV vars from a single place (previously it was scattered across `main.py` and `framework.py`); and 2, provide some unified object so that others can build on top of it if they want to implement some experiments.

So, for now, I'm sorry that there is no implementation of the event type matching thingy and we are going to merge this `_JujuContext`. Thanks!

test/test_framework.py
IronCore864: I used this way to mimic how the `_Manager` in `main.py` creates the `Framework` in the `_make_framework` function; also it shows some context of where the value comes from. But after a second look, I guess it's not super necessary, I will change it to `set()` (and other occurrences, too).

ops/framework.py
IronCore864: Got it, I have updated it to `juju_debug_at or set()`, let me run a final test before merging, especially the tests with other charms.

ops/jujucontext.py
tonyandrewmeyer: > The half-baked idea I imagined was that `from_dict` returns a `Union` of types, and the caller always uses a [match](https://peps.python.org/pep-0636/) expression on the result.

If the intention is to find out which event Juju has emitted, then it's much simpler to just look at that directly rather than trying to match against which context variables have been set. If the intention is to get back a collection of types (e.g. which fields are set are the same with different events - start and stop would be setting the same) then I'm not sure I really see the use-case.

I don't think it's good to have a hierarchy of context objects when there's already a 1:1 mapping of what that would be in the hierarchy of event classes. A `SecretChangedJujuContext` would only be useful for ` SecretChangedEvent`, for example.

As @IronCore864 mentioned, `_get_event_args` provides the "given a context, what are the arguments needed to create an event object". It would be simple enough for someone to subclass `_JujuContext` to add a `to_event` method (or whatever), if they wanted to build an "alt ops" package.

What would be most useful in that sort of situation, I believe, is being able to create `SecretChangedEvent` (and on on) objects without opting in to the whole framework/handle system (and registering, in particular). My feeling is that's what we could provide (not this cycle) to continue making it easier for people to explore alternative ways to do event handling without re-implementing everything.

> I.e. it would be great if this new class brought significant new added value over poking into `os.environ` directly.

"significant" is subjective, of course, but I think there is value in not just collecting all the uses together as @IronCore864 mentioned, but also doing all the conversion so that you have a bunch of Python objects rather than a bunch of strings. With this, plus the similar hook tool wrapping in the model/model backend (the border of those could be a bit cleaner) that's the main interface with Juju (other than Pebble).

test/test_framework.py
```diff
@@ -30,13 +30,16 @@
...
+from ops.jujucontext import _JujuContext
 from ops.model import _ModelBackend
 from ops.storage import JujuStorage, NoSnapshotError, SQLiteStorage
 from test.test_helpers import FakeScript
 
 
 def create_model():
     """Create a Model object."""
+    if 'JUJU_VERSION' not in os.environ:
+        os.environ['JUJU_VERSION'] = '0.0.0'
```
samuelallan72, line 42: @IronCore864 why does the test framework need to manipulate the environment now?  The new context appears to already pick a default if JUJU_VERSION is not present.

cc @gabrielcocenza

tonyandrewmeyer: > @IronCore864 why does the test framework need to manipulate the environment now? The new context appears to already pick a default if JUJU_VERSION is not present.

Note that this is the internal tests for ops themselves, not ops.testing (Harness). It seems like no-one else should really care what the internal ops tests are doing. Does this actually impact you, or is it just something you noticed while running the ops tests or similar?

It's being set here because `_JujuContext` expects the environment variable to be set, because all supported Juju versions so do.

gabrielcocenza: > Does this actually impact you, or is it just something you noticed while running the ops tests or similar?

Yes, it's affecting out unit tests that are checking the arguments that were called in subprocess. See an example at canonical/charm-simple-streams#22

tonyandrewmeyer: Ah, it's not the code linked above (your tests will never run that), it's [this similar code in ops.testing](https://github.com/canonical/operator/blob/cdfd10f34f2d8338fb18dea1185cdd70ea082252/ops/testing.py#L263).

We failed to consider this :disappointed:, and we must not have run this change against the full set of charms we know about :disappointed:. It would indeed be better to have Harness not change the environment. It's not immediately clear to me what we should do now: we could change the behaviour back so that it doesn't, but if everyone that this impacts has then already changed their code by the time that fix gets out, then it's just doubling up the impact (although it is cleaner in general).

I'll talk with the team about this later today.

tonyandrewmeyer: FYI, we're going to do an ops 2.16.1 change that reverts this behaviour.

gabrielcocenza: I confirm that now it's fixed :+1:

PR #1314, opened 2024-08-13
feat: poc parsing juju env in one place - alternative draft

IronCore864:
Parse all `JUJU_*` env vars in one place.

Closes https://github.com/canonical/operator/issues/1075.

Alternative version where variables are grouped with explicit types.

See the first version here: https://github.com/canonical/operator/pull/1313.

Typing is explicitly not `None` except `juju_charm_dir`, `juju_debug`, and `juju_dispatch_path` because the original code tests whether they are in `os.environ` or not.

### DIFF
--- a/ops/main.py
+++ b/ops/main.py
@@ -19,6 +19,7 @@ This is so that :code:`import ops` followed by :code:`ops.main(MyCharm)` works
 as expected.
 """
 
+import dataclasses
 import logging
 import os
 import shutil
@@ -42,6 +43,38 @@ CHARM_STATE_FILE = '.unit-state.db'
 logger = logging.getLogger()
 
 
+@dataclasses.dataclass(frozen=True)
+class _JujuContext:
+    # source: https://juju.is/docs/juju/charm-environment-variables
+    # the HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    juju_relation: str = ''
+    juju_relation_id: str = ''
+    juju_secret_revision: str = ''
+    juju_workload_name: str = ''
+    juju_action_name: Optional[str] = None
+    juju_action_uuid: Optional[str] = None
+    juju_charm_dir: Optional[str] = None
+    juju_debug: Optional[str] = None
+    juju_departing_unit: Optional[str] = None
+    juju_dispatch_path: Optional[str] = None
+    juju_notice_id: Optional[str] = None
+    juju_notice_key: Optional[str] = None
+    juju_notice_type: Optional[str] = None
+    juju_pebble_check_name: Optional[str] = None
+    juju_remote_app: Optional[str] = None
+    juju_remote_unit: Optional[str] = None
+    juju_secret_id: Optional[str] = None
+    juju_secret_label: Optional[str] = None
+    juju_storage_id: Optional[str] = ''
+
+    @classmethod
+    def from_environ(cls) -> '_JujuContext':
+        kwargs: Dict[str, Any] = {name: os.getenv(name.upper()) for name in cls.__annotations__}
+        return cls(**kwargs)
+
+
 def _exe_path(path: Path) -> Optional[Path]:
     """Find and return the full path to the given binary.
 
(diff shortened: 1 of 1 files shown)


ops/main.py
```diff
@@ -42,6 +43,87 @@
...
+@dataclasses.dataclass(frozen=True)
+class _JujuContext:
+    # source: https://juju.is/docs/juju/charm-environment-variables
+    # the HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    juju_action: '_JujuAction'
+    juju_charm_dir: Optional[str]
+    juju_debug: Optional[str]
+    juju_departing_unit: str
```
dimaqq, line 55: I think that this field should be scoped under relation

ops/main.py
```diff
@@ -42,6 +43,87 @@
...
+    juju_dispatch_path: Optional[str]
+    juju_notice: '_JujuNotice'
+    juju_pebble_check_name: str
+    juju_relation: '_JujuRelation'
```
dimaqq, line 59: I would prefer that the top-level field is optional, rather than its components

ops/main.py
```diff
@@ -42,6 +43,87 @@
...
+    juju_remote_app: str
+    juju_remote_unit: str
+    juju_secret: '_JujuSecret'
+    juju_storage_id: str
```
dimaqq, line 63: 🤔 this field is called "storage tag" in juju codebase. The change is from 3 years ago, maybe we can follow that lead? wdyt?

ops/main.py
```diff
@@ -42,6 +43,87 @@
...
+    juju_workload_name: str
```
dimaqq, line 64: I wonder if this should be scoped under "container"

ops/main.py
```diff
@@ -42,6 +43,87 @@
...
+    # the HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398
+    # only allow accesing these vars because these are the currently used vars in main.py
+
+    juju_action: '_JujuAction'
+    juju_charm_dir: Optional[str]
+    juju_debug: Optional[str]
+    juju_departing_unit: str
+    juju_dispatch_path: Optional[str]
+    juju_notice: '_JujuNotice'
+    juju_pebble_check_name: str
```
dimaqq, line 58: I imagine this too should be scoped under... a check? a container? a notice? Ugh this is getting difficult...

dimaqq (commented):
While in general, "flat is better than nested", I think that 1 level of nesting is just fine, assuming that the net difference is, roughly, writing something like

```
assert ctx.notice
if ctx.notice.type == "...": ...
```

instead of
```
assert ctx.notice_id
if ctx.notice_type == "...": ...
```

On the other hand, if nesting grew deeper, I wouldn't be so sure.

ops/main.py
tonyandrewmeyer: It indeed gets messy. It shouldn't really be under notice because notices are meant to be an implementation detail of check events. It's really check and notice both under container, but I don't think we want to be three levels deep (I'm not even convinced of two levels).

benhoyt: Yeah, I'm with Tony on this one, and this is part of the reason that hierarchies -- while seeming enticing at first -- are often not ideal. Sometimes it's unclear which hierarchy something belongs in, and sometimes a thing belongs in multiple hierarchies.

In addition, if we're going for nested, I don't like how we have to have all these different classes, like `_JujuRelation` (and it's actually not a relation, it's an "extended relation identifier" or something).

So I'm much more in favour of flat, 1-level for this, as #1313 does.

benhoyt: Additionally, we convert to "proper" types further down, like converting notice id/type/key to `LazyNotice` in the notice events. I think that's another reason to keep the types here flat and simple.

IronCore864:
After discussion, we will use a flat structure without nested objects, and we will put all fields as `Optional[str]` or `Optional[int]`. Closing this PoC in favour of [the other one](https://github.com/canonical/operator/pull/1313), changes will be made in the other PR.

PR #1315, opened 2024-08-15
ci: run tests on Py 3.12 and install test on Py 3.13

dimaqq:
Python 3.13 has reached rc1, this PR adds it to the test matrix.

Note: not using free-threaded builds at this point due to  https://github.com/actions/setup-python/issues/771 (not like we'd expect any difference, frankly)

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.10", "3.13"]
+        python-version: ["3.8", "3.10", "3.12"]
         exclude:
         - {python-version: "3.8", os: "macos-latest"}  # macos-14 is arm64, and there's no Python 3.8 build for arm64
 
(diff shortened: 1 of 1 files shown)


IronCore864 (approved):

benhoyt (changes requested):
Per discussion, we're going to add 3.12 to the places it's missing, but leave 3.13 for now. When 3.13 is out, we'll add it to the `pip install` block but not for running our tests.

dimaqq:
Aside: why do test one artefact, but not the other?
- `dist/ops-<version>.tar.gz` is tested
- `dist/ops-<version>-py3-none-any.whl` is not tested

Is it because `charmcraft pack` will use .gz and that's all we care about?

dimaqq:
Updated Python versions per review.

tonyandrewmeyer:
> Aside: why do test one artefact, but not the other?
> 
> * `dist/ops-<version>.tar.gz` is tested
> * `dist/ops-<version>-py3-none-any.whl` is not tested
> 
> Is it because `charmcraft pack` will use .gz and that's all we care about?
> 
> Full disclosure: I believe that the packages are equivalent for install purpose, even though the content is not equivalent: 97 files in a tarball, 26 files in a zip/whl. The differences are our tests, tox.ini, setup.cfg and pyproject.toml.

Yeah, I think that's the reason why: if we broke installing, it's hard to see how we would break wheel and not sdist or the other way around. The packaging tools could, but probably not something we do. So if there's no point checking both, then we might as well check the one that will be used when packing.

benhoyt (approved):
LGTM, thanks!

PR #1318, opened 2024-08-16
test: update the smoke test series/bases

tonyandrewmeyer:
A collection of fixes for the smoke test:

* Update the smoke test charm to use the preferred "import ops" style.
* Fix tox to use "allowlist_externals" rather than "whitelist_externals", which doesn't work in recent versions of tox.
* Remove a bash comment in the tox smoke env - this didn't really add much value and on my system failed (I think tox believed everything after the `#` was a comment in tox, rather than in the bash statement)
* Drop support for bionic (18.04) and xenial (16.04). Modern charmcraft can't pack on these systems, because it requires a newer version of pip than is available for the system Python. We don't really support these any more either - at least not for using current versions of ops.
* Add support for jammy (22.04)
* Add support for noble (24.04) - this required a substantial change because of the required changes to `charmcraft.yaml`. Instead of having all the bases specified in a `charmcraft.yaml` that's in version control (which works fine for the older three), have the test generate a `charmcraft.yaml` that matches the required style for that base.
* Use pytest.mark.parametrize rather than putting the build and deploy in a single test - this is nicer from a logging/progress indication point of view, and generally cleaner for what are essentially subtests. Since modern charmcraft is building each charm separately (and especially now that it's using a separate yaml for each, to handle Noble) there isn't really any functional difference in terms of the test.
* When deploying, use `base=` rather than the deprecated `series=`

To test, run `tox -e smoke` (you'll need to have `charmcraft`, `bash`, and a Juju controller available).


benhoyt (approved):
Thanks for this!

PR #1319, opened 2024-08-16
test: run pytest in parallel with pytest xdist

IronCore864:
Distributing tests across multiple CPUs to speed up test execution.

On my 10-core mac m1, time is reduced by a little more than a third (~4m39s to ~3m3s). According to GitHub Actions it also speeds up proportionally to that (see [this run](https://github.com/canonical/operator/actions/runs/10417030729/job/28850433850?pr=1313) without pytest xdist and [this run](https://github.com/canonical/operator/actions/runs/10417133297/job/28850745589?pr=1319) with it).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -73,10 +73,11 @@ deps =
     websocket-client==1.*
     coverage[toml]~=7.0
     pytest~=7.2
+    pytest-xdist~=3.6
     typing_extensions~=4.2
 commands =
     coverage run --source={[vars]src_path} \
-             -m pytest --ignore={[vars]tst_path}smoke -v --tb native {posargs}
+             -m pytest -n auto --ignore={[vars]tst_path}smoke -v --tb native {posargs}
     coverage report
 
 [testenv:pebble]
--- a/tox.ini
+++ b/tox.ini
@@ -64,7 +64,7 @@ commands =
     pyright {posargs}
 
 [testenv:unit]
-description = Run unit tests
+description = Run unit tests in parallel without coverage
 passenv =
     RUN_REAL_PEBBLE_TESTS
     PEBBLE
@@ -76,6 +76,22 @@ deps =
     typing_extensions~=4.2
 commands =
     pytest -n auto --ignore={[vars]tst_path}smoke -v --tb native {posargs}
+
+[testenv:coverage]
+description = Run unit tests with coverage
+passenv =
+    RUN_REAL_PEBBLE_TESTS
+    PEBBLE
+deps =
+    PyYAML==6.*
+    websocket-client==1.*
+    coverage[toml]~=7.0
+    pytest~=7.2
+    typing_extensions~=4.2
+commands =
+    coverage run --source={[vars]src_path} \
+             -m pytest --ignore={[vars]tst_path}smoke -v --tb native {posargs}
+    coverage report
 
 [testenv:pebble]
 description = Run real pebble tests
--- a/tox.ini
+++ b/tox.ini
@@ -71,14 +71,11 @@ passenv =
 deps =
     PyYAML==6.*
     websocket-client==1.*
-    coverage[toml]~=7.0
     pytest~=7.2
     pytest-xdist~=3.6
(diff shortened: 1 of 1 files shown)


benhoyt (commented):
Looks reasonable, and this takes the `tox -e unit` on my laptop from 1m26s to 21s, roughly 4x as fast, which is great. I've had a quick look at pytest-xdist project, and it seems reasonable -- definitely not a thorough look by any means.

One thing this breaks is the coverage report. Is that something we can fix? Or maybe we can talk about whether we care, and turn off coverage? I actually find it gets in the way most of the time anyway. :-)

```
/home/ben/w/operator/.tox/unit/lib/python3.12/site-packages/coverage/control.py:888: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")
unit: commands[1]> coverage report
Name                       Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------
ops/__init__.py                9      9      0      0     0%   44-323
ops/_private/__init__.py       0      0      0      0   100%
ops/_private/timeconv.py      60     60     28      0     0%   17-129
ops/_private/yaml.py           8      8      0      0     0%   17-33
ops/charm.py                 660    660    159      0     0%   17-1929
ops/framework.py             684    684    266      0     0%   17-1426
ops/jujuversion.py            79     79     48      0     0%   17-146
ops/lib/__init__.py          175    175     77      0     0%   23-289
ops/log.py                    26     26      4      0     0%   17-73
ops/main.py                  264    264     94      0     0%   22-566
ops/model.py                1546   1546    606      0     0%   17-3909
ops/pebble.py               1357   1357    427      0     0%   20-3361
ops/storage.py               158    158     30      0     0%   17-425
ops/testing.py              1632   1632    784      0     0%   17-3797
ops/version.py                 1      1      0      0     0%   20
----------------------------------------------------------------------
TOTAL                       6659   6659   2523      0     0%
```

dimaqq:
Indeed, coverage needs a special fix to combine the data from N processes into one.

Maybe we can try https://pytest-cov.readthedocs.io/en/latest/xdist.html ?

I've done similar with a hack in the past, the hard way; there's another, better way with pre- or post-fork hook now, but those are still kinda ugly.

tonyandrewmeyer:
Fwiw, coverage is run as part of the TIOBE reports and in the work-in-progress PR I have it running separately because it seemed cleaner/simpler than having it stored by one of the other runs and then loaded by the TIOBE one. So if we can get that running properly then we'd at least have a monthly coverage report (in the action results and in the TIOBE viewer).

IronCore864:
Per discussion at today's standup, we decided to remove coverage and do it in a separate PR/workflow.

benhoyt (approved):
Nice, thanks!

dimaqq (approved):
I'm fine with this overall, however consider Tony's #1301 where TIOBE run relies on `tox -e unit`.

Maybe we can offer a "covering" alternative, like `tox -e unit-coverage` or `tox -e coverage` which includes coverage dep, and runs tests "undistributed" under coverage?

WDYT?

Poking @tonyandrewmeyer in case he'd rather include the new tox environment in his PR instead.

tonyandrewmeyer:
> Maybe we can offer a "covering" alternative, like `tox -e unit-coverage` or `tox -e coverage` which includes coverage dep, and runs tests "undistributed" under coverage?

This sounds good to me. It could maybe also generate the XML coverage file that TIOBE wants (maybe even in the weird place they want it) which would simplify the workflow for TIOBE>

> Poking @tonyandrewmeyer in case he'd rather include the new tox environment in his PR instead.

Either's fine with me.

benhoyt:
That's a good idea. I'd go with just `tox -e coverage` -- I don't think we'll be tracking coverage for other tests anytime soon.

IronCore864:
Per the above discussion, I added `tox -e coverage` (the same as the old `tox -e unit` without parallelization).

PR #1323, opened 2024-08-20
docs: Live reload with sphinx-autobuild

dimaqq:
- [x] it works 🎉 
- [x] sphinx-autobuild was already included in pyproject.toml, was there already a way to rebuild on change?
- [x] let's paint that bike shed and decide on the `tox -e <name>` for this

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,6 +49,16 @@ commands_pre =
 commands =
     sphinx-build -W --keep-going docs/ docs/_build/html
 
+[testenv:docs-live]
+description = Live development: build the Sphinx docs with autoreloading enabled
+deps = {[testenv:docs]deps}
+commands_pre =
+    {[testenv:docs]commands_pre}
+# https://github.com/sphinx-doc/sphinx-autobuild/issues/170
+change_dir = {tox_root}/docs
+commands =
+    sphinx-autobuild ./ ./_build/html --watch . --port 8000
+
 [testenv:fmt]
 description = Apply coding style standards to code
 deps =


dimaqq:
WDYT folks, do we need it or not?

benhoyt:
I'm not opposed to it. I think I would have used this when working on docs(trings). `docs-live` is okay. Or maybe `docs-server`?

tonyandrewmeyer:
I'm not opposed to this, but I doubt I would use it. I tend to write, then do `tox -e docs` and read the output to make sure I didn't mess up a reference etc and then refresh the page in my browser. It doesn't seem like this would change that workflow enough that I would make use of it. But again, not opposed to it if it would be useful for others.

`docs-live-reload`?

docs/conf.py
```diff
@@ -2,9 +2,9 @@
 import sys
 import os
 
-sys.path.append('./')
+sys.path.append(os.path.abspath('.'))
```
IronCore864, line 5: May I ask why do we need this change? This file comes from the Canonical starter pack, so I think if it's absolutely necessary, we should leave it untouched and do things in `custom_conf.py`. Also, I had a look at Pebble docs where it also enabled `sphinx-autobuild`, but there wasn't this change and it worked.

dimaqq: docs/conf.py assumes that sphinx is run or loads config with a specific current working directory.

plain sphinx does indeed so that.

somehow sphinx-autobuild does not.

there's another way to fix this, if necessary.

dimaqq:
I've addressed the code comments, the change is smaller now.
I think it's good to go :)

benhoyt (approved):

tonyandrewmeyer (approved):
I still don't have any objection to this, but I also don't really see its value. I tried it out, thinking that I could live see any changes I made to the docs, but that isn't the case, because the docs are in `ops/` and it's watching for changes in `docs/`. So this would seem to only be useful when making changes to the custom config or the index, which are both pretty rare in my experience.

Could you please update the PR description to better explain what the PR does? Then it can also be the details for the commit to main, and they nicely match for future looking back.

benhoyt:
> the docs are in ops/ and it's watching for changes in docs/. So this would seem to only be useful when making changes to the custom config or the index, which are both pretty rare in my experience.

Oooh, I'd missed that. Yeah, I don't think it's worth it then. Can it watch `ops/` as well?

PR #1326, opened 2024-08-21
chore: standardise copyright headers

### LINKED ISSUE #1325: Remove copyright years from source files
Per team discussion, we're going to remove the copyright year (or year range in some cases) from the top of each source file, but leave the copyright notice itself, so the first line just says `# Copyright Canonical Ltd.`.

[This blog post](https://hynek.me/til/copyright-years/) notes that various large projects and companies have started doing this ([curl](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/), [React](https://github.com/facebook/react/pull/13593), [Amazon](https://chatter.quux.world/@mark/109621622761353381)), so there's a good precedent.

As mentioned in the [Curl commit])(https://github.com/curl/curl/commit/2bc1d775f510196154283374284f98d3eae03544), there's no legal need:

> - they are mostly pointless in all major jurisdictions
> - many big corporations and projects already don't use them
> - saves us from pointless churn
> - git keeps history for us

james-garner-canonical:
Copyright notice line only needs file creation year (not range).

Full boilerplate from LICENSE.txt should be used (rather than a one line note to see LICENSE).

Copyright notice process documented in HACKING.md.

Resolves #1325

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -190,6 +190,12 @@ Note that the commit messages to the PR's branch do not need to follow the
 conventional commit format, as these will be squashed into a single commit to `main`
 using the PR title as the commit message.
 
+## Copyright
+
+The format for copyright notices is documented in the [LICENSE.txt](LICENSE.txt).
+New files should begin with a copyright line with the current year (e.g. Copyright 2024 Canonical Ltd.) and include the full boilerplate (see APPENDIX of [LICENSE.txt](LICENSE.txt)).
+The copyright information in existing files does not need to be updated when those files are modified -- only the initial creation year is required.
+
 # Documentation
 
 In general, new functionality
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -1,3 +1,17 @@
+# Copyright 2019 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # ruff: noqa
 import datetime
 import pathlib
--- a/ops/_private/__init__.py
+++ b/ops/_private/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2019-2024 Canonical Ltd.
+# Copyright 2019 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -1,4 +1,4 @@
-# Copyright 2019-2021 Canonical Ltd.
+# Copyright 2019 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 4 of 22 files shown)


tonyandrewmeyer (approved):
LGTM, thanks!

I wonder if it's worth adding a small note somewhere in HACKING.md explaining what we do, to avoid people adding them back later? What do you think?

james-garner-canonical:
> I wonder if it's worth adding a small note somewhere in HACKING.md explaining what we do, to avoid people adding them back later? What do you think?

Technically the instructions we're currently following are documented in LICENSE.txt, but it might be worth adding a note in HACKING.md saying not to change existing notices and to follow those instructions for new files.

james-garner-canonical:
I've noticed that the yaml files under test/charms/test_main/ don't have copyright notices.

Nor do any of the files under docs/ (conf.py, custom_conf.py, index.rst and requirements.txt) or test/bin/ (bash and bat scripts).

Also project .md files and pyproject.toml, but I don't think those would typically require them.

Should any of these have copyright notices added? I'm thinking at least the .yaml and .py files should have them added.

benhoyt:
> I've noticed that the yaml files under test/charms/test_main/ don't have copyright notices.
> 
> Nor do any of the files under docs/ (conf.py, custom_conf.py, index.rst and requirements.txt) or test/bin/ (bash and bat scripts).
> 
> Also project .md files and pyproject.toml, but I don't think those would typically require them.
> 
> Should any of these have copyright notices added? I'm thinking at least the .yaml and .py files should have them added.

I think let's do the yaml files and the docs/*.py files, but not the Markdown/ReST/txt files/.

sed-i:
Not sure about ruff, but with flake8 we could have a regex for the copyright
https://github.com/canonical/operator/blob/7a5513accdd4acf97d929009f8d33be991d9f2b1/pyproject.toml#L95

https://github.com/canonical/alertmanager-k8s-operator/blob/b8d96dd65d83f7af611c23034cc49ed1da4390b6/pyproject.toml#L35

tonyandrewmeyer:
> I think let's do the yaml files and the docs/*.py files, but not the Markdown/ReST/txt files/.

While you're doing this, could you take a little look into why `ruff lint` isn't flagging those doc files? `CPY` is enabled as a check, and at a glance I don't see that disabled for the `docs` folder, and I thought we just recently enabled linting for custom_conf.py.

By the way, for docs/conf.py, I feel like if we add a copyright notice to this it would be better to also submit that upstream, because we're just copying this file into our repo. It's copied from Canonical, so pasting our own copyright on it is probably fine, but really it belongs at the source.

benhoyt:
> By the way, for docs/conf.py, I feel like if we add a copyright notice to this it would be better to also submit that upstream, because we're just copying this file into our repo. It's copied from Canonical, so pasting our own copyright on it is probably fine, but really it belongs at the source.

Oh yes, I'd overlooked that. I think that (at least in this PR) we shouldn't modify anything from upstream, or it'll make updating too messy. Let's just do our files in this PR (custom_config.py?).

james-garner-canonical:
> While you're doing this, could you take a little look into why `ruff lint` isn't flagging those doc files?

There is a global ruff: noqa directive in  both docs/conf.py and docs/custom_conf.py

james-garner-canonical:
> > By the way, for docs/conf.py, I feel like if we add a copyright notice to this it would be better to also submit that upstream, because we're just copying this file into our repo. It's copied from Canonical, so pasting our own copyright on it is probably fine, but really it belongs at the source.
> 
> Oh yes, I'd overlooked that. I think that (at least in this PR) we shouldn't modify anything from upstream, or it'll make updating too messy. Let's just do our files in this PR (custom_config.py?).

I wasn't sure about docs/conf.py due to this. The current version was added without customisation, but was subsequently modified slightly in one commit (noqa added and rst_epilog removed). I won't modify it further in this PR.

benhoyt (approved):
Looks good to me -- thanks for this!

PR #1330, opened 2024-08-22
refactor: enable linting of docs/custom_conf.py

james-garner-canonical:
This PR reverts the only edit to docs/conf.py, which was made in [this commit](https://github.com/canonical/operator/compare/9b2b3341ddfae77da27933297d80579b14aabdf1...8bae12d15013cf2d2b4119568a36ab55c6f01b1a), and enables linting of docs/custom_conf.py

The resulting linting errors have been resolved primarily via `tool.ruff.lint.per-file-ignores`, with one inline noqa (`copyright` variable shadows builtin), and the removal of a type annotation not in scope (`Node`). It didn't seem worth spending significant time to add docstrings and resolve the other errors at this point. In fact, I'm not entirely sure whether it's worth merging in these changes, as it sounds like we've yet to settle on how the docs code will be handled with respect to upstream going forward, but I figured I might as well make a PR for review anyway in case it's helpful.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-# ruff: noqa
 import sys
 import os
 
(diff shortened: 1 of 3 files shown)


benhoyt (approved):
Thanks -- I was a bit on the fence about just ignoring all these warnings, but I think each of them seems reasonable to me in the context of `custom_conf.py`.

james-garner-canonical:
I agree that it's not ideal to disable so many warnings. Hopefully at least having the other linting checks enabled for the file will be helpful in future.

dimaqq (approved):

PR #1332, opened 2024-08-23
test: bump pyright to 1.1.377

dimaqq:
In preparation to enabling `reportUnnecessaryTypeIgnoreComment`, bumping pyright to today's latest.

super-tox run: OK (tested on host, macos)
- 58 charms pass against canoincal/operator@main
- 58 charms pass against dimaqq/operator@bump-pyright

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,6 +141,12 @@ ignore = [
     # Manual dict comprehension.
     "PERF403",
 
+    # Convert Dict[X] to dict[x], which we cannot do while we support Python 3.8
+    "UP006",
+
+    # Convert Union[X, Y] to X | Y, which we cannot do while we support Python 3.8
+    "UP007",
+
     # Convert {} from `TypedDict` functional to class syntax
     # Note that since we have some `TypedDict`s that cannot use the class
     # syntax, we're currently choosing to be consistent in syntax even though
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,12 +141,6 @@ ignore = [
     # Manual dict comprehension.
     "PERF403",
 
-    # Convert Dict[X] to dict[x], which we cannot do while we support Python 3.8
-    "UP006",
-
-    # Convert Union[X, Y] to X | Y, which we cannot do while we support Python 3.8
-    "UP007",
-
     # Convert {} from `TypedDict` functional to class syntax
     # Note that since we have some `TypedDict`s that cannot use the class
     # syntax, we're currently choosing to be consistent in syntax even though
@@ -166,6 +160,11 @@ ignore = [
     # Unnecessary `tuple` call (rewrite as a literal)
     "C408",
 ]
+
+[tool.ruff.lint.pyupgrade]
+# Preserve types, even if a file imports `from __future__ import annotations`.
+# Since we target py 3.8, we want to keep using List[str] and Optional[str]
+keep-runtime-typing = true
 
 [tool.ruff.lint.per-file-ignores]
 "test/*" = [
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -381,10 +381,7 @@ class _TestMain(abc.ABC):
             class ThisCharmEvents(MyCharmEvents):
                 pass
 
-            class _HasStored(typing.Protocol):
-                _stored: ops.StoredState
-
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -141,6 +141,12 @@ ignore = [
     # Manual dict comprehension.
     "PERF403",
 
+    # Convert Dict[X] to dict[x], which we cannot do while we support Python 3.8
+    "UP006",
```
tonyandrewmeyer, line 145: For this, and the next one, shouldn't the target version already be skipping these? That's generally better than manually maintaining the upgrade list.

tonyandrewmeyer (commented):
Is the switch to the future import required for the pyright change? I'm happy to make that change, but if it's not part of the required changes could you add some info about it to the PR description?

benhoyt (commented):
Nice, thanks!

Can we also make sure that charms that use MyPy still type check okay? I did a fair bit of work when updating the types to ensure that charms could use either Pyright or MyPy. We should probably have a some kind of automated test for this.

test/test_main.py
```diff
@@ -381,10 +381,7 @@ def _read_and_clear_state(
             class ThisCharmEvents(MyCharmEvents):
                 pass
 
-            class _HasStored(typing.Protocol):
-                _stored: ops.StoredState
-
-            class Charm(self.charm_module.Charm, _HasStored):  # type: ignore
```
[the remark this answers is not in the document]
dimaqq: I thought that I removed the attribute **declaration** only.

The test still passes, so maybe I was wrong why this Protocol was added in the first place?

I guessed that the protocol was added because we don't actually know anything about the Charm as it was loaded dynamically and a kludge was needed, e.g. because `._stored` was accessed right below, or some such.

Let's cover this during the stand-up, I'm happy to learn the back story.

pyproject.toml
dimaqq: It should, but it doesn't.

Let me double-check, I may have left this line when I was trying to upgrade `ruff` at the same time.

tonyandrewmeyer: Hmm, [looks like](https://docs.astral.sh/ruff/settings/#target-version) we should remove target-version in favour of our Python version specification.

tonyandrewmeyer: Hmm, [looks like](https://docs.astral.sh/ruff/settings/#target-version) we should remove target-version in favour of our Python version specification.

test/test_main.py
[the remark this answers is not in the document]
tonyandrewmeyer: > I thought that I removed the attribute **declaration** only.

Ah, you did, sorry. I misread before.

> The test still passes, so maybe I was wrong why this Protocol was added in the first place?

I'm not sure why it exists, or why things are good without it.

> Let's cover this during the stand-up, I'm happy to learn the back story.

Hopefully, Ben knows :)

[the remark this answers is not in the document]
tonyandrewmeyer: > I thought that I removed the attribute **declaration** only.

Ah, you did, sorry. I misread before.

> The test still passes, so maybe I was wrong why this Protocol was added in the first place?

I'm not sure why it exists, or why things are good without it.

> Let's cover this during the stand-up, I'm happy to learn the back story.

Hopefully, Ben knows :)

pyproject.toml
dimaqq: Right, so I was getting a tox lint warning in `ops/pebble.py` which I couldn't figure a good way around.

My solution was to enable `from __future__ import annotations` and update all the type hints. A huge mechanical change. Somehow ruff then applies these two rules that surely should be gated by py 3.10+, and normally they are, but not when `__future__` is enabled.

I suppose I could try to fix that original single linting issue some other way, maybe with I could simply suppress it?

tonyandrewmeyer: Or maybe we should be setting [keep-runtime-typing](https://docs.astral.sh/ruff/settings/#lint_pyupgrade_keep-runtime-typing)?

I'm in favour of having the future import (until it's no longer required because it's the standard behaviour in the earliest version we support). It would be nice if `ruff` still applied the Python version around the pyupgrade checks, so we don't need to remember to go remove those when we move to a newer Python. But it's not that bad to need to do it.

benhoyt (approved):

IronCore864 (approved):

PR #1333, opened 2024-08-23
ci: make Pyright report unnecessary type ignore comments

dimaqq:
Parent: https://warthogs.atlassian.net/browse/CHARMTECH-232

Covered in the parent epic:

- charmcraft analyse will validate that charm Python entrypoint calls ops.main
- it is static type checker’s job to ensure correct arguments to ops.main
- today, many charms do and have to do ops.main(ThisCharm)  # type: ignore 
- thus, type hints need to be improved so that typpe: ignore can be dropped

 

Plan in this story:

- upgrade pyright (not sure if strictly necessary, #1332)
- enabled pyright setting reportUnnecessaryTypeIgnoreComment
- include code that calls ops.main(SomeCharm) in every which legal way in ops tests somewhere
   - pyright validates the positive case, that ops.main can be called with a charm class
- include code that calls ops.main() without a charm class
   - pyright now validates that such invocation is detected as an error
   - thus, this invocation is annotated with # type: ignore or similar
   - with the setting above, pyright also enforce that the annotation is actually needed

Ref: "How to write negative type tests?" https://github.com/microsoft/pyright/discussions/8803

Part 2, based on #1332 

MyPy checks, pinned mypy:
- [x] `nginx-ingress-integrator-operator` mypy 1.6.1 OK
- [x] `sdcore-webui-operator` mypy 1.11.1 OK

Super-tox, running on host (macos)
- 58 charms pass against canonical/operator@main
- 58 charms pass against dimaqq/aoperator@pyright-reportUnnecessaryTypeIgnoreComment

### DIFF
--- a/ops/__init__.py
+++ b/ops/__init__.py
@@ -175,7 +175,7 @@ __all__ = [  # noqa: RUF022 `__all__` is not sorted
 # isort:skip_file
 
 # Import pebble explicitly. It's the one module we don't import names from below.
-from . import pebble  # type: ignore
+from . import pebble
 
 # Also import charm explicitly. This is not strictly necessary as the
 # "from .charm" import automatically does that, but be explicit since this
@@ -184,7 +184,7 @@ from . import charm  # type: ignore # noqa: F401 `.charm` imported but unused
 
 # Import the main module, which we've overriden in main.py to be callable.
 # This allows "import ops; ops.main(Charm)" to work as expected.
-from . import main  # type: ignore
+from . import main
 
 # Explicitly import names from submodules so users can just "import ops" and
 # then use them as "ops.X".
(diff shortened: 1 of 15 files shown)


benhoyt (commented):
Similar to my comment on #1332, let's just ensure that removing these doesn't mess with charms that use MyPy for type checking.

dimaqq:
I've updated the description.

tonyandrewmeyer (approved):

dimaqq:
re failing o11y tests, it's an upstream issue, a test dep is not pinned:
https://github.com/canonical/prometheus-k8s-operator/pull/638
https://github.com/canonical/alertmanager-k8s-operator/pull/286

benhoyt (approved):
Nice cleanup, thanks!

PR #1338, opened 2024-08-26
feat: add the description field to SecretInfo

### LINKED ISSUE #1289: Expose "description" in SecretInfo
The `SecretInfo` that's returned from `Secret.get_info()` does not include the `description` field. This is included in `secret-info-get` - we're just ignoring it at the moment. It is possible to set the description in `Secret.set_info()` - it seems reasonable that charms should be able to access this information as well.

tonyandrewmeyer:
The `SecretInfo` class exposed all the fields from `secret-info-get` exception `description`. This PR adds that field as well.

Fixes #1289.


benhoyt (approved):
Thanks!

PR #1339, opened 2024-08-26
Fix main split up tests

dimaqq:
Same as #1331, but the tests are split up:

separate files for runtime and type tests

separate tests for recommended and legacy invocation at run time.


dimaqq:
I've merged this into the other PR

PR #1342, opened 2024-08-26
fix: correct the signature of .events()

### LINKED ISSUE #1341: `Framework.ObjectEvents.events` has wrong signature
The signature of the `.events()` method is:

```python
    def events(self) -> Dict[str, EventSource]:
        """Return a mapping of event_kinds to bound_events for [...]
```

so what is it, EventSource or bound_events?
Turns out in practice it is a mapping of str to BoundEvent (docstring is correct, type annotation is wrong).

image

PietroPasotti:
Fixes #1341

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -485,7 +485,7 @@ class ObjectEvents(Object):
                 event_kinds.append(attr_name)
         return event_kinds
 
-    def events(self) -> Dict[str, EventSource]:
+    def events(self) -> Dict[str, BoundEvent]:
         """Return a mapping of event_kinds to bound_events for all available events."""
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
Thanks! Curious that pyright doesn't complain about this.

PietroPasotti:
I think it didn't because all attributes that were accessed were coincidentally common between the two types :)
they both talked like ducks

tonyandrewmeyer (approved):
LGTM, thanks!

The main case seems very straightforward - the code is clearly wanting the bound event - it's even asserting on it. The error message probably works either way, but I'm +0 on this adjustment.

The framework one should indeed be working with bound events - the class has the source, the instance has the bound event.

benhoyt (approved):
Thanks!

Issue #1343, opened 2024-08-28
`status_set`: document (or link to) the list of valid values


sed-i:
Currently, the docstring doesn't mention that "status" is limited to a set of valid values.
https://github.com/canonical/operator/blob/796918744ef259df0a449a4366b2c184da732029/ops/model.py#L3405-L3410

benhoyt:
Hey Leon, this is the docstring of a method on a private (underscore-prefixed `_ModelBackend`) class, that we see as an implementation detail of Ops. I don't mind adding the set of values it's limited to, but I'm wondering why this is important in the documentation for a method on a private class?

tonyandrewmeyer:
Same question about why it's worthwhile (given that if someone is building something on top of ops so has a legitimate reason to use a private class, they ought to know enough about Juju to know the possible values), but for what it's worth, I'd suggest we would do this by just type annotating `status`, [identically to how it already is in Harness's backend](https://github.com/canonical/operator/blob/8bc7ad8ddf9a0fb9f90c2b6c7c3583467cc040f5/ops/testing.py#L2492).

tonyandrewmeyer:
This was fixed in #1354.

PR #1354, opened 2024-09-03
chore: Document and validate settable status values in _ModelBackend.set_status

james-garner-canonical:
Document the status values that are valid arguments for `_ModelBackend.status_set`, as requested in #1343, via type annotation. The `status` argument was previously type hinted only as `str`. However, this argument can really only be one of four valid statuses that Juju's `status-set` will accept `('active', 'blocked', 'maintenance', 'waiting')`. This new type information is propagated to `StatusBase` and relevant tests, including `_TestingModelBackend`.

Since the arguments to `status_set` are already partially validated,  (`is_app` being a bool in `status_set`, `status` and `message` being strings in `Application.status` and `Unit.status`), and and the valid values for `status` are now encoded and associated with this method, we can check that `status` is valid in `status_set` and raise a `ModelError` immediately if not, rather than catching `status-set`'s exit code and only then raising a `ModelError`.

Tests are updated to account for this (`test_collect_status_priority` split into `test_collect_status_priority_valid` and `test_collect_status_priority_invalid` since `status-set` isn't called in the invalid case, and `test_local_set_invalid_status` is updated to account for this too).

Since `is_app` and `status` are both validated in `status_set`, it makes sense to also validate the type of `message` here, removing redundant checks in `Application.status` and cleaning up a `fixme` in `Unit.status`. A test is added to cover this (`test_status_set_message_not_str_raises`).

Finally, fix a couple of broken unit tests that interact with `StatusBase` and `status_set` (`test_base_status_instance_raises` and `test_status_set_is_app_not_bool_raises` respectively).


dimaqq:
```
  /home/runner/work/operator/operator/src/charm.py:119:33 - error: Argument of type "str" cannot be assigned to parameter "name" of type "_StatusName" in function "from_name"
    Type "str" is not assignable to type "_StatusName"
      "str" is not assignable to type "Literal['active']"
      "str" is not assignable to type "Literal['blocked']"
      "str" is not assignable to type "Literal['maintenance']"
      "str" is not assignable to type "Literal['waiting']"
      "str" is not assignable to type "Literal['error']"
      "str" is not assignable to type "Literal['unknown']" (reportArgumentType)
```

Given that it's an observability job that fails, I suspect that this change is a breaking change subject to how charmers use our code today.

Please check.

P.S. I recommend amazing https://github.com/tonyandrewmeyer/charm-analysis to try new code across some 120~160 charms.

james-garner-canonical:
This PR breaks a type checking test in canonical/prometheus-k8s-operator where StatusBase.from_name(name, message) [is called](https://github.com/canonical/prometheus-k8s-operator/blob/27025e5d840ec09331608b398cf4e49e5fe256e9/src/charm.py#L119). The error is because `name` is typed as `str` in this charm, while this PR changes the type of the `name` argument in `from_name` from `str` to `Literal['active', 'blocked', 'maintenance', 'waiting', 'error', 'unknown']`.

Since `StatusBase.from_name` is part of the public api of ops, this isn't good. On the other hand, `from_name` does [currently](https://github.com/canonical/operator/blob/e56686444e3d1504c1785262a1bb3610e1c5182a/ops/model.py#L1888) raises a `KeyError` at runtime if called with a `name` that isn't one of those six.

Solutions here include:

1. change the type of the `name` argument of `from_name` back to `str`, and `cast`/etc internally (it's not our users' problem)
2. require users to either more strongly type their arguments, use a `cast`, or use a `type: ignore[asignment]`
2a. make it easier to use the type `Literal['active', 'blocked', 'maintenance', 'waiting', 'error', 'unknown']` externally (e.g. expose a `model.StatusName` type alias in the public api)
2b. submit a PR to canonical/prometheus-k8s-operator to help with this

I'd lean in favour of (1), and will look into modifying this PR to avoid changes to the type signature of the public api at this point.

james-garner-canonical:
> P.S. I recommend amazing https://github.com/tonyandrewmeyer/charm-analysis to try new code across some 120~160 charms.

Thanks, I can see that `StatusBase.from_name` is used in three more charms (see below), as well as in Ben's test-charms repository. I've reverted the signature of this method. I don't think any other changes affect the public API.

Charms that use `StatusBase.from_name`:
```
prometheus-k8s-operator/src/charm.py
sunbeam-charms/ops-sunbeam/ops_sunbeam/compound_status.py
jenkins-k8s-operator/src/charm.py
loki-k8s-operator/src/charm.py
test-charms/statustest/src/statuspool.py
```

dimaqq:
Looking at loki example, what they do:

```py
class CompositeStatus(TypedDict):
    """Per-component status holder."""

    # These are going to go into stored state, so we must use marshallable objects.
    # They are passed to StatusBase.from_name().
    k8s_patch: Tuple[str, str]
    config: Tuple[str, str]
    rules: Tuple[str, str]
    retention: Tuple[str, str]

...


    def _on_k8s_patch_failed(self, event: K8sResourcePatchFailedEvent):
        self._stored.status["k8s_patch"] = to_tuple(BlockedStatus(cast(str, event.message)))

...

def to_status(tpl: Tuple[str, str]) -> StatusBase:
    """Convert a tuple to a StatusBase, so it could be used natively with ops."""
    name, message = tpl
    return StatusBase.from_name(name, message)

...

    def _on_collect_unit_status(self, event: CollectStatusEvent):
        # "Pull" statuses
        # TODO refactor _configure to turn the "rules" status into a "pull" status.

        # "Push" statuses
        for status in self._stored.status.values():
            event.add_status(to_status(status))
```

Arguably that code could be refactored, though it's unclear exactly how.

One option would be to change str to a union of literals...

Another option would be observe that `_stored` could hypothetically persist over charm upgrades and therefore ops library upgrades and downgrades, and the run-time check against supported values is therefore needed should the allow list ever change.

In any case, good call to keep this PR small and leave more extensive changes for later / slower process.

james-garner-canonical:
Interesting observation about data persisting across library versions. I definitely wouldn't want to remove any runtime checks when adding type checks (though here it would just fail with a relatively traceable KeyError on dictionary access).

The data flow you've pointed out in loki could indeed simply be annotated with `Literal`s, which would make it compatible with both the current api and the breaking change (now reverted). Defining the `Literal['active', 'blocked', 'maintenance', 'waiting', 'error', 'unknown']` is a bit painful for users, and isn't  very 'single source of truth', especially when thinking about changing ops/juju versions, so I wouldn't want to recommend this pattern to library users generally.

I wonder if it would be worth making the `StatusName` (and `SettableStatusName`?) -- type aliases for `Literal`s -- part of the public api, if not in this PR, then perhaps in future.

ops/model.py
```diff
@@ -3413,6 +3426,10 @@ def status_set(self, status: str, message: str = '', *, is_app: bool = False) ->
         """
         if not isinstance(is_app, bool):
             raise TypeError('is_app parameter must be boolean')
+        if not isinstance(message, str):
+            raise TypeError('message parameter must be a string')
```
tonyandrewmeyer, line 3430: Is there any point having this, given the check immediately below?

ops/model.py
```diff
@@ -3413,6 +3426,10 @@ def status_set(self, status: str, message: str = '', *, is_app: bool = False) ->
...
+        if status not in _SettableStatusNames:
+            raise ModelError(f'status must be in {_SettableStatusNames}, not {status}')
```
tonyandrewmeyer, line 3432: Maybe this should be `InvalidStatusError` rather than `ModelError`?

Also, while I'm here - I noticed when looking over the docs that we say that `ModelError` will be raised if using `ErrorStatus` or `UnknownStatus` to set - could you fix that to say `InvalidStatusError` (if I'm correct in thinking that is the case - it certainly seems like it from the rest of the docs).

dimaqq:
115 out of 115 (100%) runs passed. 🚀 

I would have imagined at least some charm to explicitly set 
"error" for good measure, which is now disallowed, but alas! no such thing 😅

ops/model.py
james-garner-canonical: This checks the message argument, while the following check is for the status argument

ops/model.py
```diff
@@ -50,6 +51,7 @@
     Type,
     TypedDict,
     Union,
+    get_args,
```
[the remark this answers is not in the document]
dimaqq: me too

ops/model.py
dimaqq: +1 on InvalidStatusError from me.

While this could (pedantically) be considered a breaking change at the API level (someone may have called this function with `status="error"` and got away with it), it seems that none's doing that, according to super-tox. Arguably it wouldn't be very useful due to juju hook tool behaviour either.

Pls grep known charm code for `"unknown"` and `"error"` before resolving.

test/test_model.py
```diff
@@ -2951,7 +2945,18 @@ def test_local_get_status(self, fake_script: FakeScript, name: str):
...
         for is_app_v in [None, 1, 2.0, 'a', b'beef', object]:
             with pytest.raises(TypeError):
-                self.backend.status_set(ops.ActiveStatus, is_app=is_app_v)  # type: ignore
+                self.backend.status_set(
+                    'active',
+                    is_app=is_app_v,  # type: ignore[assignment]
+                )
+
+    def test_status_set_message_not_str_raises(self):
+        for message in [None, 1, 2.0, True, b'beef', object]:
```
dimaqq, line 2954: My gut tells me you meant:

```suggestion
        for message in [None, 1, 2.0, True, b'beef', object()]:
```

ops/model.py
```diff
@@ -68,7 +70,10 @@
 _StorageDictType = Dict[str, Optional[List['Storage']]]
 _BindingDictType = Dict[Union[str, 'Relation'], 'Binding']
 
-_StatusDict = TypedDict('_StatusDict', {'status': str, 'message': str})
+_SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
+_SettableStatusNames: Tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
```
dimaqq, line 74: I'm a bit uneasy about how similar the two look.
Do you think it would be better, if the constant was _ALL_UPPER_CASED to distinguish it from the _CamelCased type?

dimaqq: P.S. I'm also a bit on the fence about the "_Settable" prefix, it seems quirky, like what does "settable" mean in abstract, without reading the rest of the code?

I feel like maybe this is inverted? WDYT about:
- the user-visible type would be called `_StatusName`, and
- the internal union would be called e.g. `_StatusNameOrInternal` or some such?

ops/model.py
dimaqq: P.S. given that it's a string, wdyt about:

```suggestion
            raise ModelError(f'status must be in {_SettableStatusNames}, not {status!r}')
```

dimaqq (approved):

ops/model.py
```diff
@@ -423,9 +428,12 @@ def status(self, value: 'StatusBase'):
         if not self._backend.is_leader():
             raise RuntimeError('cannot set application status as a non-leader unit')
 
-        for _key in {'name', 'message'}:
-            assert isinstance(getattr(value, _key), str), f'status.{_key} must be a string'
-        self._backend.status_set(value.name, value.message, is_app=True)
+        self._backend.status_set(
+            typing.cast(_SettableStatusName, value.name),  # TODO: narrower type for value
```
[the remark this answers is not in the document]
james-garner-canonical: My thinking was that we could eventually narrow the type of value to help with static checking, as you've suggested, or with an intermediate base class for settable `Status` types that takes care of the type of `status.name`.

The major blocker for this currently is that the getter needs to return `StatusBase`, since it should be valid to learn that Juju has set your status to `error` or `unknown` via this mechanism, right?

Python type checkers don't seem to fully support having [different setter and getter signatures for properties](https://github.com/python/mypy/issues/3004), so even if we use a narrower type for the setter, users won't see errors when they write e.g. `my_unit.status = UnknownStatus()` (though this *will* explode at runtime).

In the meantime, I'll just leave the `cast` and remove the `TODO`.

test/test_model.py
james-garner-canonical: To be honest I just copied the test cases from `test_status_set_is_app_not_bool_raises` and changed the `str` one to a `bool`. I'll update them both to avoid future puzzlement.

ops/model.py
james-garner-canonical: I agree on changing to `InvalidStatusError`, especially as it's a subclass of `ModelError`, so it's unlikely to break anybody's `try ... except ...`

Anything in particular you'd like me to grep for? unknown and error are such generic terms that they're widely used in many codebases. I did find one interesting result in namespace-node-affinity-operator, where `unit.status` is set to `ErrorStatus` on [this line](https://github.com/canonical/namespace-node-affinity-operator/blob/3b24b7c166461c77ff313400d5089e8136099cae/src/charm.py#L187):
```
            self.model.unit.status = ErrorStatus("K8S resources removal failed")
```
This will currently cause the charm to raise an unhandled `ModelError`, and with this PR would instead cause it to raise an unhandled `InvalidStatusError`.

ops/model.py
james-garner-canonical: I agree that the name is a bit strange, especially out of context. `_StatusDict` was already used in `model.py` (with `str` typed 'status' key), while `_StatusName` was used only in tests (and contained 5 of the 6 Juju statuses), while the `Settable` prefix is just what I came up with, and I'd definitely be happy to change it to something better.

Perhaps marking the non-settable names explicitly as read-only would improve things. Does this help readability enough, do you think?
```python
_ReadOnlyStatusName = Literal['error', 'unknown']
_SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
_StatusName = Union[_SettableStatusName, _ReadOnlyStatusName]
_StatusDict = TypedDict('_StatusDict', {'status': _StatusName, 'message': str})
_SETTABLE_STATUS_NAMES: Tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
```
I'm wary about making the full union name too unfriendly as it reflects the `name` attribute type for the user facing type `StatusBase`, which includes `ErrorStatus` and `UnknownStatus`. While these can't be used to `set` the status without causing an error, you could definitely `get` such a status.

Maybe a suffix is better? Would improve discoverability at least.
```python
_StatusNameReadOnly = Literal['error', 'unknown']
_StatusNameSettable = Literal['active', 'blocked', 'maintenance', 'waiting']
_StatusName = Union[_StatusNameSettable, _StatusNameReadOnly]
_StatusDict = TypedDict('_StatusDict', {'status': _StatusName, 'message': str})
_SETTABLE_STATUS_NAMES: Tuple[_StatusNameSettable, ...] = get_args(_StatusNameSettable)
```

james-garner-canonical:
> Generally looks good - a couple of small notes/queries.
> 
> One other thing I noticed while reviewing the docs to see if they fit with this: the `add_status` method accepts both `error` and `unknown`, and the doc lists those. It reads very much like you can add those statuses, but especially for `error` if you include that then it'll definitely fail because it will "win" the evaluation and then can't be set.
> 
> Maybe it makes this PR too big to adjust that, but if so we should probably open a ticket about that too.

I've opened an issue to discuss this (#1356).

ops/model.py
tonyandrewmeyer: Oops, totally overlooked that :)

james-garner-canonical:
I'm going to defer work on (https://github.com/canonical/operator/issues/1356 -- i.e.  making it an error to call `add_status` with `ErrorStatus` or `UnknownStatus`) to a future PR. Pending finalising the name for the `_StatusName` / `_SettableStatusName` type aliases (see https://github.com/canonical/operator/pull/1354#discussion_r1742903566), I don't anticipate doing further work on this PR in the meantime. There isn't really any rush for getting it finalised, but I would welcome any feedback on the names. For ease of reference, I'll post my two ideas below, but I'm open to suggestions.

```python
_ReadOnlyStatusName = Literal['error', 'unknown']
_SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
_StatusName = Union[_SettableStatusName, _ReadOnlyStatusName]
_StatusDict = TypedDict('_StatusDict', {'status': _StatusName, 'message': str})
_SETTABLE_STATUS_NAMES: Tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
```
Prefixes, but slightly more comprehensible in context?

```python
_StatusNameReadOnly = Literal['error', 'unknown']
_StatusNameSettable = Literal['active', 'blocked', 'maintenance', 'waiting']
_StatusName = Union[_StatusNameSettable, _StatusNameReadOnly]
_StatusDict = TypedDict('_StatusDict', {'status': _StatusName, 'message': str})
_SETTABLE_STATUS_NAMES: Tuple[_StatusNameSettable, ...] = get_args(_StatusNameSettable)
```
Are suffixes better here?

ops/model.py
benhoyt: I'm not too worried about these, given they're internal (though see my comment elsewhere -- maybe `StatusName` should be public, given that it's the type annotation of a public field).

ops/model.py
```diff
@@ -1866,10 +1877,10 @@ class StatusBase:
     directly use the child class such as :class:`ActiveStatus` to indicate their status.
     """
 
-    _statuses: Dict[str, Type['StatusBase']] = {}
+    _statuses: Dict[_StatusName, Type['StatusBase']] = {}
 
-    # Subclasses should override this attribute
-    name = ''
+    # Subclasses must provide this attribute
+    name: _StatusName
```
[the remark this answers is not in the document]
james-garner-canonical: I think it's ok to make it public, and I think it would be good to do so. This is very much a read only property, so the narrower type is fine. No one should be setting this attribute, and no one should be subclassing this base class, especially not with a non-`StatusName` "name".

The following type check won't break since name is a union of literal strings.
```python
my_var: str = status_object.name
```

The signature of `StatusBase.from_name` hasn't changed -- the `name` argument is still typed as `str`. Ideally someday it would be typed as `StatusName`, but for now this isn't changed, so nothing should break.

I do think that I should update the `StatusBase.from_name` docstring to list all six valid names.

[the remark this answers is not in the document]
benhoyt: Per discussion in daily sync, we'll make `StatusName` public.

james-garner-canonical:
I think this is ready for a final review and merge if anyone wants to take another look and provide the second approval (@tonyandrewmeyer, @benhoyt).

ops/charm.py
```diff
@@ -1079,12 +1079,17 @@ class CollectStatusEvent(LifecycleEvent):
 
     The order of priorities is as follows, from highest to lowest:
 
-    * error
+    * error (Note: will always result in an InvalidStatusError!)
```
benhoyt, line 1082: I wonder if we should just remove the "error" and "unknown" lines from this bulleted list, as you'd never want to use them with collect-status / `add_status`. If that's the case, I think we should just remove the new explanatory paragraph too -- it probably just adds confusion, given you can't use those.

ops/model.py
```diff
@@ -69,7 +71,11 @@
 _StorageDictType = Dict[str, Optional[List['Storage']]]
 _BindingDictType = Dict[Union[str, 'Relation'], 'Binding']
 
-_StatusDict = TypedDict('_StatusDict', {'status': str, 'message': str})
+_ReadOnlyStatusName = Literal['error', 'unknown']
+_SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
+StatusName = Union[_SettableStatusName, _ReadOnlyStatusName]
```
benhoyt, line 76: We need to add this to `ops/__init__.py`, so it's accessible from there and shows up in the [docs](https://ops--1354.org.readthedocs.build/en/1354/search.html?q=statusname&check_keywords=yes&area=default).

ops/model.py
```diff
@@ -1911,7 +1923,8 @@ def from_name(cls, name: str, message: str):
         does not have an associated message.
 
         Args:
-            name: Name of the status, for example "active" or "blocked".
+            name: Name of the status, one of:
+                "active", "blocked", "maintenance", "waiting", "error", or "unknown".
```
benhoyt, line 1927: I think we should revert this change, as the range of values is now shown in the type, [for example](https://ops--1354.org.readthedocs.build/en/1354/#ops.StatusBase.name)

benhoyt (commented):
Looks good, just a couple of minor comments.

ops/charm.py
benhoyt: Decision is to revert this change in favour of https://github.com/canonical/operator/issues/1356

ops/model.py
benhoyt: Nah, I'd missed which `name` this was - ignore.

james-garner-canonical:
Discussed changes have been made, happy to approve and merge @benhoyt ?

benhoyt (approved):
Yep, looks good, thanks!

PR #1363, opened 2024-09-05
chore: re-enable test now that Pebble directory permissions are fixed

tonyandrewmeyer:
https://github.com/canonical/pebble/issues/80 is fixed (almost three years ago!) so we should enable this test again, as far as I can understand.


benhoyt (approved):
Thanks!

dimaqq (approved):
🎉

PR #1364, opened 2024-09-06
fix: remote unit data is available in relation-departed

### LINKED ISSUE #1109: remote unit data is inaccessible in relation-departed
The [docs](https://juju.is/docs/sdk/relation-name-relation-departed-event#heading--relation-data) say:

> A unit’s relation settings persist beyond its own departure from the relation
[...]
> During a relation-departed hook, relation settings can still be read (with relation-get) [...] All units will still be able to see all other units

However, it's not possible (without using a private attribute) to access remote data from ops.

This [pair of example charms](https://gist.github.com/tonyandrewmeyer/695a91eff9eae093a073683739b42aef) demonstrate the issue:

* 'provider-charm' puts some example data into the local unit relation databag on `relation-joined` and does nothing else.
* 'require-charm' logs that data on `relation-changed` (just to show it's accessible normally) and attempts to log that data on `relation-departed`. However, that will fail, unless we get the data with manually creating a `RelationDataContent` object, which requires using `model._backend`.

To use the charms:

1. `charmcraft pack` each charm, and `juju deploy` them into a model (I used k8s but I believe this will be the case for machine as well)
2. `juju integrate` the two charms together
3. Wait for everything to settle down (you'll see that the data is in the relation databag from the `relation-changed` debug-log output)
4. Use `juju remove-relation` to remove the integration between the applications
5. You should see `departed: foo` and `departed (using private): foo` in `juju debug-log`, but instead will see `departed: COULD NOT GET IT.` and `departed (using private): foo`.

The reason for this is that `Relation.units` is populated by running Juju's `relation-list`, which doesn't have the departing relation in the `relation-departed` event. `Relation.data` is a `RelationData` object, and the `._data` attribute (which `__getitem__` exposes) is populated with (a) data for the local unit, (b) data for the local app, and (c) data for each unit in the relation's `.units`. In (e.g.) `relation-changed`, this means that the remote unit's data will be there, but in `relation-departed` it will not.

Juju's `relation-get` will still provide access to the data, and this is meant to be available (according to the docs linked above and also [the Juju team](https://matrix.to/#/!xdClnUGkurzjxqiQcN:ubuntu.com/$R8_gCGrfhjcqM6k_zIaVFcoiWX6sw02aSmoxUcVxsnM?via=ubuntu.com&via=matrix.org&via=fsfe.org)). However, the only method ops provides to use `relation-get` is via the `RelationDataContent` class (which needs the backend passed to instantiate an object) or `RelationData` (which needs the unit to be in `relation-list` as described above).

It seems like we should make this data available. It seems cleanest for it to be available in `Relation.data` like it would be outside of `relation-departed` - the departed unit is available in the environment and in the `event` object, so there would need to be a bit of work to get it into `Relation.data` without passing too much context around, but it's certainly feasible.

tonyandrewmeyer:
During relation-departed the remote unit data should still be available to the charm, but the relation is no longer included in `relation-list`, so it's not available by the usual means (but is available when using the hook tool directly).

When a remote unit name is provided in the Juju context, ensure that unit is included in the relation units.

Fixes #1109


tonyandrewmeyer:
This can be done much more simply by adding this at line 208 of `_main.py`:

```python
        if relation:
            relation.data._data[kwargs['unit']] = ops.model.RelationDataContent(
                relation, kwargs['unit'], model._backend
            )
```

Pros/cons compared to the current state of the branch (I've manually tested that both work, but unit tests are non-trivial so want to finalise on an approach before committing to those):
 * This is far less invasive, and it feels wrong that `Model` has the remote unit name (although it's not particularly different from knowing the broken relation ID, which we already did).
 * This changes the `event.relation` object but if other code does `model.get_relation` during the relation-departed event, they don't get the change. It feels like `get_relation` shouldn't need to know anything about the current event but it also feels wrong for the two objects to be different.

The more I work on this, the more it feels like we're working around Juju behaviour and perhaps we should be advocating for a change in Juju instead (to only change `relation-list` at the conclusion of `relation-broken`).

dimaqq:
@tonyandrewmeyer still draft or ready for review and possibly merge?

tonyandrewmeyer:
@benhoyt, @dimaqq I'm not sure why I assigned you for review when this was still in draft, sorry. Anyway, it's ready for review now.

ops/_main.py
```diff
@@ -365,14 +365,19 @@ def _make_storage(self, dispatcher: _Dispatcher):
...
 
         model = _model.Model(
-            self._charm_meta, self._model_backend, broken_relation_id=broken_relation_id
+            self._charm_meta,
+            self._model_backend,
+            broken_relation_id=broken_relation_id,
+            # Similarly, if we are in a RelationDeparted event, we need to provide
+            # access to the remote relation data, even though the relation will not
+            # be returned from the `relation-list` hook tool.
+            remote_unit_name=self._juju_context.remote_unit_name,
```
benhoyt, line 380: Do we not need to use similar logic as for relation-broken here? "if relation-departed event, use this, otherwise None" If not, why not? Or is `juju_context.remote_unit_name` only not None if in a relation-departed event?

test/test_model.py
```diff
@@ -4410,5 +4410,19 @@ def test_error(self, fake_script: FakeScript, model: ops.Model):
         assert str(excinfo.value) == 'ERROR cannot access cloud credentials\n'
 
 
+@pytest.mark.skipif(
+    not hasattr(ops.testing, 'Context'), reason='requires optional ops[testing] install'
+)
+def test_departing_unit_in_relations():
```
benhoyt, line 4416: Does this test succeeding mean that Scenario uses `_make_framework`?

benhoyt (approved):
Just a couple of questions.

Presumably (given the test) this behaviour will be fixed in (Scenario) unit tests? If so, let's be sure to run it using super-tox across lots of charm tests.

ops/_main.py
tonyandrewmeyer: I believe it's not None in any relation event, but in events other than relation-departed (and likely relation-broken) when it's set the `relation-list` output will include that unit, so the `.units` will already include it. That's a set (of cached objects) so adding the one specified in the environment to the ones from `relation-list` is a no-op outside of relation-departed.

It seemed cleaner to just provide it any time it was set, but I can adjust it so that it's only done on relation-departed if you prefer. The end behaviour will be the same.

test/test_model.py
tonyandrewmeyer: Yes. Scenario uses (a subclass of) `_Manager`. As far as possible, it tries to use the same ops mechanics, other than the Juju hook tools.

ops/_main.py
benhoyt: I'm fine to leave it as is then, but perhaps adjust the comment to clarify.

tonyandrewmeyer:
Hi @dimaqq - friendly reminder to review this when you have some time available. Thanks!

ops/_main.py
```diff
@@ -365,14 +365,24 @@ def _make_storage(self, dispatcher: _Dispatcher):
...
 
         model = _model.Model(
-            self._charm_meta, self._model_backend, broken_relation_id=broken_relation_id
+            self._charm_meta,
+            self._model_backend,
+            broken_relation_id=broken_relation_id,
+            # In a RelationDeparted event, the unit is not included in the Juju
+            # `relation-list` output, but the charm still has access to the remote
+            # relation data. To provide the charm with a mechanism for getting
+            # access to that data, we include the remote unit in the set of units.
```
dimaqq: **minor:** I'm not sure what "the set of units" encompasses here.

test/test_model.py
```diff
@@ -4410,5 +4410,19 @@ def test_error(self, fake_script: FakeScript, model: ops.Model):
         assert str(excinfo.value) == 'ERROR cannot access cloud credentials\n'
 
 
+@pytest.mark.skipif(
+    not hasattr(ops.testing, 'Context'), reason='requires optional ops[testing] install'
```
dimaqq, line 4414: **praise:** I kinda like testing ops itself using Scenario!

test/test_model.py
```diff
@@ -4410,5 +4410,19 @@ def test_error(self, fake_script: FakeScript, model: ops.Model):
...
+)
+def test_departing_unit_in_relations():
+    ctx = ops.testing.Context(
+        ops.CharmBase, meta={'name': 'mycharm', 'requires': {'db': {'interface': 'db'}}}
+    )
+    rel = ops.testing.Relation('db', remote_units_data={0: {}}, remote_app_name='db')
+    state_in = ops.testing.State(relations={rel})
+    with ctx(ctx.on.relation_departed(rel, remote_unit=1), state_in) as mgr:
+        mgr.run()
+        assert {unit.name for unit in mgr.charm.model.relations['db'][0].units} == {'db/0', 'db/1'}
```
dimaqq: **discussion:** the way the test is written, it's unclear what it being tested.

Perhaps this is indeed the best way to test the way potentially-missing-remote unit is propagated. In that case some comments would help.

dimaqq (approved):
**question:** at high level, don't we have the same problem with both `JUJU_REMOTE_APP` and `JUJU_REMOTE_UNIT` ?

ops/_main.py
tonyandrewmeyer: `Relation.units: set[ops.Unit]`. I'll tweak the comment.

test/test_model.py
tonyandrewmeyer: Added comments.

tonyandrewmeyer:
> **question:** at a high level, don't we have the same problem with both `JUJU_REMOTE_APP` and `JUJU_REMOTE_UNIT` ?

`JUJU_REMOTE_APP` should always be set in all relation events (see the [Juju bug](https://bugs.launchpad.net/juju/+bug/1960934)).

`JUJU_REMOTE_UNIT` is still set in relation-departed. The issue is only that the unit is not included in `relation-list` in that event (the environment variables are fine, it's the output from the hook tool that's missing what ops needs).

dimaqq:
> `JUJU_REMOTE_UNIT` is still set in relation-departed. The issue is only that the unit is not included in `relation-list` in that event (the environment variables are fine, it's the output from the hook tool that's missing what ops needs).

As, sorry, I conflated two issues.

Presumably the provider side of an inter-app relation also needs to list all related apps.

I just want to confirm that this issue (remote missing in listing) only applies to units and not apps.

tonyandrewmeyer:
> I just want to confirm that this issue (remote missing in listing) only applies to units and not apps.

Yes, the app is always there other than in the separate known issue around JUJU_REMOTE_APP.

dimaqq:
Cross-referencing https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$soyBslfVVkROMq-FOPTTq_prEMxdSdSPrAcmVnjIZz4?via=ubuntu.com&via=matrix.org&via=laquadrature.net

ops/_main.py
addyess: @benhoyt and @tonyandrewmeyer i don't know how you'd like me to report this...

i've been hitting an issue i think this may have created when trying to use 2.23.0 and 3.0.0

I have a charm `k8s` that relates to another application on multiple relations:
* `containerd`
* `k8s-cluster`

during the `containerd_relation_joined` event with `k8s-worker/0` , if i look at the `model.get_relation("k8s-cluster").units` -- i see `k8s-worker/0`.  

but `k8s-worker/0` hasn't yet run it's `relation_joined` joined event. 


because it shows up in the `Relation.units`, when i try access the data `relation.data[unit]`, i end up with a `KeyError` because that unit is not yet joined on the `k8s-cluster` relation. 



i was able to change this line of code to :

```python
            remote_unit_name=broken_relation and self._juju_context.remote_unit_name
```

and the issue goes away.  I know that's probably not the right fix -- but i'm not sure i understand the value of `Relation.units` showing a unit that isn't actually joined to a relation.

tonyandrewmeyer: Thanks @addyess ! I'll take a look at this later today and probably open a new issue.

tonyandrewmeyer: https://github.com/canonical/operator/issues/1917

PR #1365, opened 2024-09-06
ci: periodically run the unit tests of all GitHub-hosted published charms

tonyandrewmeyer:
Once a month, before the typical release time, run a workflow that tests all charms published on CharmHub where the source location is in the charm metadata and that location is in a `canonical` GitHub repository.

This is much the same as the existing data, hello, and observability tests, but at a larger scale and without the expectation that all of the tests would pass in every PR. The intention is that this gives us an insight into when ops changes break the most important charms - not that we would necessarily ensure that the tests were always passing (but we would at least look into it).

This also tests the `charmcraft init` profiles (other than the framework profiles, which don't start out with usable tests) to detect breakage there.

To avoid manually maintaining the list of charms, a script is added that will update the workflow file with the latest list, pulled from CharmHub. This is not currently automatically run.


benhoyt:
Nice! Do you have an example run?

tonyandrewmeyer:
> Nice! Do you have an example run?

Yes, [on my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/10730039376).

There are a few failures:

* temporal-k8s (haven't dug into why)
* temporal (looks like basically the same as the -k8s one)
* data-platform-tests (needs this [PR](https://github.com/canonical/data-platform-libs/pull/120), which I guess I should try pushing someone to merge)
* mysql-router-k8s (96 failures of one test - haven't dug into this either)

.github/update-published-charms-tests-workflow.py
```diff
@@ -0,0 +1,144 @@
+#! /usr/bin/env python
+
+# /// script
```
benhoyt, line 3: Can this be under the Copyright header? Seems like that should be as close to the top as possible, for consistency.

benhoyt: How do you run this script? Do we want a Tox command, or just a comment here with the `uv` or whatever incantation?

.github/update-published-charms-tests-workflow.py
```diff
@@ -0,0 +1,144 @@
...
+    workflow['jobs']['charm-tests']['strategy']['matrix']['include'] = [
+        {'charm-repo': f'canonical/{charm}'} for charm in charms if charm and charm not in SKIP
+    ]
+    with WORKFLOW.open('w') as f:
+        yaml.dump(workflow, f)
+    # yaml.safe_load transforms "on" to "true". See https://github.com/yaml/pyyaml/issues/376
+    # The 'run' step that patches the requirements files also ends up looking
+    # rather unfriendly, but it's still functionally the same, so works.
+    with WORKFLOW.open('r') as f:
+        content = f.read().replace('true:', 'on:')
```
benhoyt, line 138: Just in case?

```suggestion
        content = f.read().replace('\ntrue:', '\non:')
```

.github/workflows/published-charms-tests.yaml
```diff
@@ -0,0 +1,104 @@
...
+      with:
+        repository: ${{ matrix.charm-repo }}
+    - name: Checkout the operator repository
+      uses: actions/checkout@v4
+      with:
+        path: myops
+    - name: Install patch dependencies
+      run: pip install poetry~=1.6
+    - name: Update 'ops' dependency in test charm to latest
+      run: "rm -rf myops/test\nif [ -e \"test-requirements.txt\" ]; then\n  sed -i\
```
benhoyt, line 21: I see your comment above, but this is going to be almost impossible to edit/maintain. Is there some way to have this in a separate file? Or have the source of it be in the update-published-charms script file in a reasonable format?

benhoyt (commented):
Thanks! Left a few comments. Have you looked into the failures? Be nice to start with a green slate if we can.

.github/update-published-charms-tests-workflow.py
tonyandrewmeyer: > Can this be under the Copyright header? Seems like that should be as close to the top as possible, for consistency.

Sure. I haven't seen these used much yet (now that both pipx and uv have support I imagine they'll get more common), so wasn't really sure where the right placing was. I can't see any [ordering instructions in the docs](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata) - it seems like maybe you can put the comment anywhere as long as it's top-level. I'll move the copyright above.

> How do you run this script? Do we want a Tox command, or just a comment here with the `uv` or whatever incantation?

I simplified it to only use `pyyaml` and the standard lib so if you're in an environment with the ops dependencies then you can just `python update-published-charms-tests-workflow.py`. If you're not then using `uv run` will handle getting pyyaml, and I think maybe pipx can be used too (but I'm not sure what the incantation is for a single-file run).

I can add in instructions, or if it's preferred I can drop the `pyyaml` requirement and instead of parsing the YAML and then dumping it out again, I can just open it as text and do a basic text replacement. I have to do that once anyway because of the 'on:' to 'true:' quirk. Any preference?

tonyandrewmeyer:
> Have you looked into the failures? Be nice to start with a green slate if we can.

* data-platform-libs: I need to convince someone to merge my ~9 month old PR.
* temporal[-k8s]: I can open a PR to get them updated to more recent ops - it should just be expecting the checks in the Pebble plan
* mysql-router-k8s: I'll look into it - they have thousands of tests that default to running 120 at a time which does not work well on my system but I guess here I can just run the failing one to investigate.

.github/update-published-charms-tests-workflow.py
benhoyt: I think we should do basic text replacement, but not really for the dependencies reason -- but because it will solve the `run` formatting shenanigans, too.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: Hmm, fair point. Maybe it is best to go back to reading and patching the file as text rather than YAML? I did look into whether I could get pyyaml to roundtrip this as it was but couldn't find any option that would do that without having to write something custom for dumping.

tonyandrewmeyer:
> * data-platform-libs: I need to convince someone to merge my ~9 month old PR.

Rebased the [PR](https://github.com/canonical/data-platform-libs/pull/120) and slightly improved it, and posted in Matrix to see if someone can work with me to get it merged.

> * temporal[-k8s]: I can open a PR to get them updated to more recent ops - it should just be expecting the checks in the Pebble plan

* [temporal-k8s-operator](https://github.com/canonical/temporal-k8s-operator/pull/41)
* [temporal-ui-k8s-operator](https://github.com/canonical/temporal-ui-k8s-operator/pull/13)
* temporal-worker-k8s-operator: this needs the [pytest-interface-test PR](https://github.com/canonical/pytest-interface-tester/pull/21) to land, and I think that will solve it, or it might then need some updates - I'll look after that PR is merged.

> * mysql-router-k8s: I'll look into it - they have thousands of tests that default to running 120 at a time which does not work well on my system but I guess here I can just run the failing one to investigate.

Ah, this is the one about the change for relation-broken not including the broken relation in the list any more. That's actually the Juju behaviour by now, so the charm really ought to be upgraded. This is non-trivial, but I'll take a look tomorrow and see if I can figure it out in an hour or so.

.github/workflows/published-charms-tests.yaml
```diff
@@ -0,0 +1,124 @@
...
+          fi
+          if [ -e "requirements-charmcraft.txt" ]; then
+            sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements-charmcraft.txt
+            echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements-charmcraft.txt
+          fi
+          if [ -e "requirements.txt" ]; then
+            sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
+            echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
+          else
+            sed -i -e "s/^ops[ ><=].*/ops = {path = \"myops\"}/" pyproject.toml
```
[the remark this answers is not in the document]
tonyandrewmeyer: > How do we know `pyproject.toml` exists in this branch? If there's no requirements.txt we assume pyproject -- is that safe enough?

If the charm is using poetry or uv then `pyproject.toml` should exist and have the dependencies. I'm not entirely sure for other tools like hatch.

I guess this should really look for `poetry.lock` and do the pyproject.toml edit plus poetry lock if it exists, and fail otherwise, and then if there are failures we can add support for other tools if published charms adopt them.

For what it's worth, the dependency working group's spec currently requires a requirements.txt.

.github/update-published-charms-tests-workflow.py
```diff
@@ -0,0 +1,132 @@
+#! /usr/bin/env python
```
dimaqq, line 1: It wasn't obvious to me that this script needed to be run manually 🙈

.github/workflows/tiobe.yaml
```diff
@@ -3,6 +3,7 @@ name: TIOBE Quality Checks
 on:
   schedule:
     - cron:  '0 7 1 * *'
+  workflow_dispatch:
```
dimaqq, line 6: unrelated change

dimaqq (approved):

.github/workflows/tiobe.yaml
tonyandrewmeyer: Yeah, I'm not sure what the story is with this, because main@HEAD has:

```yaml
on:
  workflow_dispatch:
  schedule:
    - cron:  '0 7 1 * *'
```

So I don't understand why there's no complaint about a conflict, or the `workflow_dispatch` line being both above and below (and I could just remove the below one, which I must have had in an earlier version in my fork). I'll try to figure out what's happening here.

.github/update-published-charms-tests-workflow.py
tonyandrewmeyer: What would you recommend here? I could add a section to `CONTRIBUTING.md` that explains that this exists and how to update it, or I could just add that in the docstring for the script itself. Or I guess it could be run automatically on a schedule, which is probably cleanest, but I wasn't sure if we were sold enough on this idea to do that straight off.

benhoyt: Yeah, I think manual updates are a good start. We can always run this automatically later if desired.

benhoyt (approved):
Nice, thanks. Let's see how this goes!

PR #1366, opened 2024-09-06
docs: note about repeatedly setting secret value in Juju 3.6

dimaqq:
https://warthogs.atlassian.net/browse/CHARMTECH-171

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -279,7 +279,7 @@ class Model:
         again, unless ``refresh=True`` is used, or :meth:`Secret.set_content`
         has been called.
 
-        .. jujuversion:: 3.0
+        .. jujuadded:: 3.0
             Charm secrets added in Juju 3.0, user secrets added in Juju 3.3
 
         Args:
@@ -464,7 +464,7 @@ class Application:
     ) -> 'Secret':
         """Create a :class:`Secret` owned by this application.
 
-        .. jujuversion:: 3.0
+        .. jujuadded:: 3.0
 
         Args:
             content: A key-value mapping containing the payload of the secret,
@@ -1233,7 +1233,7 @@ class Secret:
     All secret events have a :code:`.secret` attribute which provides the
     :class:`Secret` associated with that event.
 
-    .. jujuversion:: 3.0
+    .. jujuadded:: 3.0
         Charm secrets added in Juju 3.0, user secrets added in Juju 3.3
     """
 
@@ -2927,7 +2927,7 @@ class Container:
     def get_notice(self, id: str) -> pebble.Notice:
         """Get details about a single notice by ID.
 
-        .. jujuversion:: 3.4
+        .. jujuadded:: 3.4
 
         Raises:
             ModelError: if a notice with the given ID is not found
@@ -2952,7 +2952,7 @@ class Container:
         See :meth:`ops.pebble.Client.get_notices` for documentation of the
         parameters.
 
-        .. jujuversion:: 3.4
+        .. jujuadded:: 3.4
         """
         return self._pebble.get_notices(
             users=users,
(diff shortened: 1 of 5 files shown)


dimaqq:
@tonyandrewmeyer we don't have a magical "Changed in Juju X.Y" on par with "Added in Juju / Removed in Juju" do we?

tonyandrewmeyer:
> @tonyandrewmeyer we don't have a magical "Changed in Juju X.Y" on par with "Added in Juju / Removed in Juju" do we?

No, but if you look at the end of docs/custom_conf.py you'll see that it's trivial to add.

ops/model.py
```diff
@@ -1445,6 +1445,9 @@ def set_content(self, content: Dict[str, str]):
         secret no longer exists, this method will succeed, but the unit will go
         into error state on completion of the current Juju hook.
 
+        Changed in Juju 3.6: setting the content to the same exact value no
+        longer creates a new secret revision.
+
```
tonyandrewmeyer, line 1450: I think we must have had a discussion about adding this to the docs, since there's a Jira ticket, but I can't remember the details of that, sorry. It seems like this really belongs in the _Juju_ docs, not the _ops_ docs, because it's a Juju behaviour change. Like perhaps a note in the [secret lifecycle doc](https://juju.is/docs/juju/secret#heading--secret-lifecycle), and we link to that? Or [secret-set](https://juju.is/docs/juju/hook-tool#secret-set) and link to that?

On the other hand, if this opened with "This will create a new secret revision, if the content differs from the latest revision, and will notify all units ..." then I wouldn't have objected to that and thought it had too much Juju specific information. So maybe it is fine?

I think having a Sphinx directive for Juju changed as you hinted at in another comment would be better than just having it like this.

I don't love "same exact", maybe:

```suggestion
        Changed in Juju 3.6: setting the content with the same content as the latest
        revision longer creates a new secret revision.
```

dimaqq:
Refactored to use custom sphinx markup.

P.S. I'm not too happy with redundancy in "Setting the content with the same content as ..." phrase. Would someone have a better idea?

ops/model.py
```diff
@@ -1445,6 +1445,10 @@ def set_content(self, content: Dict[str, str]):
         secret no longer exists, this method will succeed, but the unit will go
         into error state on completion of the current Juju hook.
 
+        .. jujuchanged:: 3.6
+            Setting the content with the same content as the latest revision
+            longer creates a new secret revision.
+
```
james-garner-canonical, line 1451: Should it be 'no longer creates'?

Rephrasing idea to eliminate redundancy, feel free to disregard
```
            A new secret revision will *not* be created if the content being set
            is identical to the latest revision.
```

dimaqq: Done, thank you!

tox.ini
```diff
@@ -54,10 +54,8 @@ description = Live development: build the Sphinx docs with autoreloading enabled
 deps = {[testenv:docs]deps}
 commands_pre =
     {[testenv:docs]commands_pre}
-# https://github.com/sphinx-doc/sphinx-autobuild/issues/170
-change_dir = {tox_root}/docs
 commands =
-    sphinx-autobuild ./ ./_build/html --watch . --port 8000
+    sphinx-autobuild docs/ docs/_build/html --watch ops/ --port 8000 {posargs}
```
tonyandrewmeyer, line 58: Does this mean that the live reload will now update if the content changes, rather than just the config?

tonyandrewmeyer (approved):
This still feels like it belongs in the Juju docs and not the ops docs, but I'm ok with adding it.

tox.ini
dimaqq: it should be on both, content and settings.
`--watch` means additional directories to watch, according to usage.

james-garner-canonical (approved):

PR #1369, opened 2024-09-11
refactor: move the content of ops.testing to ops._private.harness

tonyandrewmeyer:
Moves the content of `ops/testing.py` to `ops/_private/harness.py`, and then re-exposes all the names in the `ops.testing` namespace (other than standard library and third-party modules).

This allows other testing frameworks (such as Scenario) to be exposed in the `ops.testing` namespace while also building on top of the existing testing code.


tonyandrewmeyer:
PRs to stop charms using the private class:
* [prometheus--k8s](https://github.com/canonical/prometheus-k8s-operator/pull/639)

tonyandrewmeyer:
> PRs to stop charms using the private class:
> 
> * [prometheus-k8s](https://github.com/canonical/prometheus-k8s-operator/pull/639)
> * [grafana-k8s](https://github.com/canonical/grafana-k8s-operator/pull/347)
> * [postgresql](https://github.com/canonical/postgresql-operator/pull/617)
> * [postgresql-k8s](https://github.com/canonical/postgresql-k8s-operator/pull/692)
> * [mysql](https://github.com/canonical/mysql-operator/pull/519)

Unfortunately, these are not the only culprits, there's also:

* mimir-k8s-operator/lib/charms/harness_extensions/v0/evt_sequences.py
* minio-operator/tests/unit/test_charm.py
* notebook-operators/charms/jupyter-controller/tests/unit/test_operator.py
* opensearch-operator/tests/helpers.py
* pgbouncer-operator/tests/unit/helpers.py
* mongodb-operator/tests/unit/helpers.py
* sunbeam-charms/ops-sunbeam/ops_sunbeam/test_utils.py
* seldon-core-operator/tests/unit/test_operator.py
* loki-k8s-operator/tests/unit/test_provider.py
* loki-k8s-operator/tests/unit/test_charm.py

My feeling is that by accessing a private name and not fixing that after the issue in ops was fixed, the charmers should take responsibility for fixing this themselves (it's fairly straightforward). However, if people want to argue that we should offer up PRs for the above as well, I can do that.

ops/testing.py
```diff
@@ -14,3823 +14,52 @@
...
+    # These names are exposed for backwards compatibility.
+    'CharmBase',
+    'CharmMeta',
+    'CharmType',
+    'Container',
+    'ExecArgs',
+    'ExecHandler',
+    'ExecProcess',
+    'ExecResult',
+    'Harness',
```
benhoyt, line 55: Presumably `Harness` should be in the top group, in other words, the "Expose the Harness functionality" group. I wonder if these groups got a bit messed up -- it seems there should be a lot more in that group.

benhoyt (commented):
Do you think you could rebase/redo this PR into two commits: one with a pure rename of `ops/testing.py` to `ops/_private/harness.py`, and one with any actual changes? Otherwise it's very difficult to see the actual changes.

ops/testing.py
tonyandrewmeyer: Yeah, the auto-sorting breaks things if it's in one list, so I have it in two now. I put the PR back into draft until things were resolved - I wasn't expecting reviews until I un-drafted it.

benhoyt: Oops -- my bad. I was just going through my email notifications and missed that it was a draft.

ops/testing.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2021 Canonical Ltd.
```
benhoyt, line 1: Year added = 2024?

ops/testing.py
```diff
@@ -0,0 +1,64 @@
...
+    YAMLStringOrFile,
+    charm,
+    framework,
+    model,
+    pebble,
+    storage,
+)
+
+# The Harness testing framework.
+_ = ActionFailed
```
benhoyt, line 42: What's the difference/advantage of doing this over `__all__`?

ops/testing.py
```diff
@@ -0,0 +1,64 @@
...
+_ = CharmType
+_ = ExecArgs
+_ = ExecHandler
+_ = ExecResult
+_ = Harness
+_ = ReadableBuffer
+_ = YAMLStringOrFile
+
+# Names exposed for backwards compatibility
+_ = CharmBase
```
benhoyt, line 54: I guess they don't hurt too much, but it is a pity we need these here. Have you tested using super-tox without them? If they're used, which ones?

benhoyt (commented):
Looks straight-forward enough -- just a couple of questions for now.

ops/testing.py
tonyandrewmeyer: Yeah, it wasn't clear what to do with that. There is a file `ops/testing.py` that has existed since 2021. The content of the new file `ops/_private/harness.py` has existed since 2021. I wasn't sure what years to put in. If it's purely on the content then I suppose 2024 is more correct for `ops/testing.py` but it's also a bunch of imports, not really copyright-able anyway.

Happy to do this if you prefer:

```suggestion
# Copyright 2024 Canonical Ltd.
```

ops/testing.py
tonyandrewmeyer: `__all__` causes Sphinx to get very confused, and I couldn't figure out any way to keep `__all__` and de-confuse it.

ops/testing.py
tonyandrewmeyer: I haven't, partly because there is definitely breakage on ~12 charms because they're using `_TestingModelBackend`, so I would need to work out if it's additional breakage or only that.

I still feel like the right-thing-to-do-(tm) is to expose these, just like we did with `main` and `model`, even if we don't know of any specific cases where it would break things, because the API was that these names were available, even if we really hope or expect that no-one was using that.

If the team (or team leader :smile:) strongly prefers that we don't have these, then I can spend the time to figure out if there will be known breakage.

benhoyt: Yep, that's fair enough -- let's leave it as is.

ops/testing.py
benhoyt: Yep, I think the 2024 makes more sense as it's really a completely new file.

benhoyt (commented):
Okay, I'm happy with the changes themselves, but I'm concerned about what this means for our API reference docs. It seems like PR means there are now no docs for `ops.testing` -- is that intentional? https://ops--1369.org.readthedocs.build/en/1369/#module-ops.testing

tonyandrewmeyer:
> Okay, I'm happy with the changes themselves, but I'm concerned about what this means for our API reference docs. It seems like PR means there are now no docs for `ops.testing` -- is that intentional?

No, when I checked what the docs were like I loaded the wrong file, so I missed this. I'll figure out how to fix it.

benhoyt (approved):
Looks good to me, thanks!

dimaqq (approved):

dimaqq:
pls rebase too

PR #1370, opened 2024-09-13
feat: add a JujuVersion property for Pebble log forwarding to Loki

### LINKED ISSUE #1230: Add pebble log forwarding support check to `jujuversion`
Following up on #1074, should we add something to `ops/jujuversion.py`?
E.g. `supports_pebble_log_forwarding`.
I realize it's a pebble feature, but since the two are in lockstep then perhaps it makes sense?

tonyandrewmeyer:
Adds a new property to `JujuVersion` indicating if the version of Pebble bundled with Juju supports log forwarding to Loki, with labels.

See discussion in #1230 around which version of log forwarding should be the cutoff for support, which landed on when labels were added.

Fixes #1230

### DIFF
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -144,3 +144,13 @@ class JujuVersion:
             # releases have the change (3.2.1 tag was never released).
             return False
         return True
+
+    @property
+    def supports_pebble_log_forwarding(self) -> bool:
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Looks good to me, thanks!

dimaqq (approved):

PR #1373, opened 2024-09-17
fix: Secret.set_info and Secret.set_content can be called in the same hook

### LINKED ISSUE #1288: Cannot call Secret.set_info and Secret.set_content in the same hook
If `Secret.set_info()` and `Secret.set_content()` are both called in the same hook, then whichever is first is ignored.

I expect that from a Juju point of view, this is because `secret_set` has been called twice and so it assumes that the later call replaces the earlier one, rather than trying to merge the two together. This is particularly unexpected behaviour in ops because we have two different methods exposing parts of `secret_set`, so a charmer doesn't necessarily know that it's the same underlying Juju hook tool.

This can be reproduced with this small charm:

```python
class SecretInfoAndContentSetCharm(ops.CharmBase):
    """Charm the application."""

        super().__init__(framework)
        framework.observe(self.on.set_action, self._on_set)

        self.unit.add_secret(label="my-secret", content={"foo": "bar"})

    def _on_set(self, event):
        secret = self.model.get_secret(label="my-secret")
        secret.set_content(event.params)
        secret.set_info(description="hello world")
```

If you call the action (`juju run unit-name/0 set one=two`) with the code like it is above, then the description will be set, but the content will not. If you swap the code so that the `set_content` and `set_info` calls are around the other way, then the action will set the content but not the description.

tonyandrewmeyer:
Each call to `secret-set` replaces the values in any previous `secret-set` call in the same hook. Since both `Secret.set_content` and `Secret.set_info` use `secret-set`, this means that doing both in the same hook would only retain the change from whichever was done last.

This PR changes the model backend to cache the secret metadata (from `set_info`) and add it into any subsequent `secret-set` calls. Although the metadata is only available to the secret owner, it does not seem so private that it is unsafe to keep it in memory for the duration of the hook execution.

There is an additional behaviour change here: previously calling `set_info` twice in the same hook would 'undo' setting metadata. For example:

```python
secret.set_info(label="foo")
secret.set_info(description="bar")
# With main, the secret will end up with a description and no label
# With the PR branch, the secret will end up with a description and a label
```

I believe this is a bug fix also, because it seems likely that a charmer would expect both values to be set with code such as the above, and the documentation states:

> Once attributes are set, they cannot be unset.

For the secret content, I believe we should not cache this in the charm process's memory, so this PR only sets a sentinel that the content has been set, and if there is a subsequent `set_info` call, the content is retrieved via a `secret-get` call (I believe this is from the in-memory cache in the unit agent, but, in any case, it's the un-committed value in an existing location, so does not increase the secret content exposure).

There is an example charm in #1288 that can be used to verify the behaviour in a real Juju model.

No updates to Harness because it already has the behaviour that we're changing to.

Fixes #1288

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -109,10 +109,6 @@ _NetworkDict = TypedDict(
 )
 
 
-class _SecretSetSequenceError(Exception):
-    """Raised when calling secret-set without content after previously calling it with content."""
-
-
 logger = logging.getLogger(__name__)
 
 MAX_LOG_LINE_LEN = 131071  # Max length of strings to pass to subshell.
@@ -1517,7 +1513,6 @@ class Secret:
                 effect only after the currently-scheduled rotation.
 
         Raises:
-            TypeError: if no attributes are provided.
             ModelError: if used after previously using :meth:`set_content` in the
                 same hook.
         """
@@ -1527,18 +1522,13 @@ class Secret:
             )
         if self._id is None:
             self._id = self.get_info().id
-        try:
-            self._backend.secret_set(
-                typing.cast(str, self.id),
-                label=label,
-                description=description,
-                expire=_calculate_expiry(expire),
-                rotate=rotate,
-            )
-        except _SecretSetSequenceError:
-            raise ModelError(
-                "Can't set secret info after setting content in the same hook."
-            ) from None
+        self._backend.secret_set(
+            typing.cast(str, self.id),
+            label=label,
+            description=description,
+            expire=_calculate_expiry(expire),
+            rotate=rotate,
+        )
 
     def grant(self, relation: 'Relation', *, unit: Optional[Unit] = None):
         """Grant read access to this secret.
@@ -3694,9 +3684,8 @@ class _ModelBackend:
         elif 'rotate' in self._secret_cache[id]:
             args += ['--rotate', self._secret_cache[id]['rotate'].value]
         if content is None and 'content' in self._secret_cache[id]:
-            raise _SecretSetSequenceError(
-                'secret-set called without content, and previously used in this hook'
-            )
+            # Get the previous content from the unit agent's cache.
+            content = self.secret_get(id=id, peek=True)
         elif content is not None:
             self._secret_cache[id]['content'] = object()
         with tempfile.TemporaryDirectory() as tmp:
--- a/ops/model.py
+++ b/ops/model.py
@@ -1511,10 +1511,6 @@ class Secret:
             expire: New expiration time (or timedelta from now) to apply.
             rotate: New rotation policy to apply. The new policy will take
                 effect only after the currently-scheduled rotation.
-
-        Raises:
-            ModelError: if used after previously using :meth:`set_content` in the
-                same hook.
         """
         if label is None and description is None and expire is None and rotate is None:
             raise TypeError(
--- a/ops/model.py
+++ b/ops/model.py
@@ -109,6 +109,10 @@ _NetworkDict = TypedDict(
 )
 
 
+class _SecretSetSequenceError(Exception):
+    """Raised when calling secret-set without content after previously calling it with content."""
+
+
 logger = logging.getLogger(__name__)
 
 MAX_LOG_LINE_LEN = 131071  # Max length of strings to pass to subshell.
@@ -1511,20 +1515,30 @@ class Secret:
             expire: New expiration time (or timedelta from now) to apply.
             rotate: New rotation policy to apply. The new policy will take
                 effect only after the currently-scheduled rotation.
+
+        Raises:
+            TypeError: if no attributes are provided.
+            ModelError: if used after previously using :meth:`set_content` in the
+                same hook.
         """
         if label is None and description is None and expire is None and rotate is None:
             raise TypeError(
-                'Must provide a label, description, expiration time, ' 'or rotation policy'
+                'Must provide a label, description, expiration time, or rotation policy'
             )
         if self._id is None:
             self._id = self.get_info().id
-        self._backend.secret_set(
-            typing.cast(str, self.id),
-            label=label,
-            description=description,
-            expire=_calculate_expiry(expire),
-            rotate=rotate,
-        )
+        try:
+            self._backend.secret_set(
+                typing.cast(str, self.id),
+                label=label,
+                description=description,
+                expire=_calculate_expiry(expire),
+                rotate=rotate,
+            )
+        except _SecretSetSequenceError:
+            raise ModelError(
+                "Can't set secret info after setting content in the same hook."
+            ) from None
 
     def grant(self, relation: 'Relation', *, unit: Optional[Unit] = None):
         """Grant read access to this secret.
@@ -3680,7 +3694,9 @@ class _ModelBackend:
         elif 'rotate' in self._secret_cache[id]:
             args += ['--rotate', self._secret_cache[id]['rotate'].value]
         if content is None and 'content' in self._secret_cache[id]:
-            raise ValueError('secret-set called without content, and previously used in this hook')
+            raise _SecretSetSequenceError(
+                'secret-set called without content, and previously used in this hook'
+            )
         elif content is not None:
             self._secret_cache[id]['content'] = object()
         with tempfile.TemporaryDirectory() as tmp:
--- a/ops/model.py
+++ b/ops/model.py
@@ -340,7 +340,9 @@ class _ModelCache:
     def __init__(self, meta: 'ops.charm.CharmMeta', backend: '_ModelBackend'):
         self._meta = meta
         self._backend = backend
-        self._secret_set_cache: Dict[str, Dict[str, Any]] = collections.defaultdict(dict)
+        self._secret_set_cache: collections.defaultdict[str, Dict[str, Any]] = (
+            collections.defaultdict(dict)
+        )
         self._weakrefs: _WeakCacheType = weakref.WeakValueDictionary()
 
     @typing.overload
@@ -1290,7 +1292,7 @@ class Secret:
         id: Optional[str] = None,
         label: Optional[str] = None,
         content: Optional[Dict[str, str]] = None,
-        _secret_set_cache: Optional[Dict[str, Dict[str, Any]]] = None,
+        _secret_set_cache: Optional['collections.defaultdict[str, Dict[str, Any]]'] = None,
     ):
         if not (id or label):
             raise TypeError('Must provide an id or label, or both')
@@ -1300,7 +1302,9 @@ class Secret:
         self._id = id
         self._label = label
         self._content = content
-        self._secret_set_cache = _secret_set_cache or collections.defaultdict(dict)
+        self._secret_set_cache: collections.defaultdict[str, Dict[str, Any]] = (
+            collections.defaultdict(dict) if _secret_set_cache is None else _secret_set_cache
+        )
 
     def __repr__(self):
         fields: List[str] = []
--- a/ops/model.py
+++ b/ops/model.py
@@ -39,6 +39,7 @@ from typing import (
     Any,
     BinaryIO,
     Callable,
+    ClassVar,
     Dict,
     Generator,
     Iterable,
@@ -1265,6 +1266,8 @@ class Secret:
 
     _key_re = re.compile(r'^([a-z](?:-?[a-z0-9]){2,})$')  # copied from Juju code
 
+    _secret_set_cache: ClassVar[Dict[str, Dict[str, Any]]] = collections.defaultdict(dict)
+
     def __init__(
         self,
         backend: '_ModelBackend',
@@ -1484,7 +1487,26 @@ class Secret:
         self._validate_content(content)
         if self._id is None:
             self._id = self.get_info().id
-        self._backend.secret_set(typing.cast(str, self.id), content=content)
+
+        # If `_backend.secret_set` has already been called, this call will
+        # overwrite anything done in the previous call (even if it was in a
+        # different event handler, as long as it was in the same Juju hook
+        # context). We want to provide more predictable behavior, so we cache
+        # the values and provide them in subsequent calls.
+        cached_data = self._secret_set_cache[self._id]
+        expire = _calculate_expiry(cached_data['expire']) if 'expire' in cached_data else None
+        # Don't store the actual secret content in memory, but put a sentinel
+        # there to indicate that the content has been set during this hook.
+        cached_data['content'] = object()
+
+        self._backend.secret_set(
+            typing.cast(str, self.id),
+            content=content,
+            label=cached_data.get('label'),
+            description=cached_data.get('description'),
+            expire=expire,
+            rotate=cached_data.get('rotate'),
+        )
         # We do not need to invalidate the cache here, as the content is the
         # same until `refresh` is used, at which point the cache is invalidated.
 
@@ -1518,8 +1540,29 @@ class Secret:
             )
         if self._id is None:
             self._id = self.get_info().id
+
+        # If `_backend.secret_set` has already been called, this call will
+        # overwrite anything done in the previous call (even if it was in a
+        # different event handler, as long as it was in the same Juju hook
+        # context). We want to provide more predictable behavior, so we cache
+        # the values and provide them in subsequent calls.
+        cached_data = self._secret_set_cache[self._id]
+        label = cached_data.get('label') if label is None else label
+        cached_data['label'] = label
+        description = cached_data.get('description') if description is None else description
+        cached_data['description'] = description
+        expire = cached_data.get('expire') if expire is None else expire
+        cached_data['expire'] = expire
+        rotate = cached_data.get('rotate') if rotate is None else rotate
+        cached_data['rotate'] = rotate
+        # Get the previous content from the unit agent's cache.
+        content = (
+            self._backend.secret_get(id=self._id, peek=True) if 'content' in cached_data else None
+        )
+
         self._backend.secret_set(
             typing.cast(str, self.id),
+            content=content,
             label=label,
             description=description,
             expire=_calculate_expiry(expire),
@@ -3218,7 +3261,6 @@ class _ModelBackend:
         self._is_leader: Optional[bool] = None
         self._leader_check_time = None
         self._hook_is_running = ''
-        self._secret_cache: Dict[str, Dict[str, Any]] = collections.defaultdict(dict)
 
     def _run(
         self,
@@ -3661,32 +3703,12 @@ class _ModelBackend:
         args = [id]
         if label is not None:
             args.extend(['--label', label])
-            self._secret_cache[id]['label'] = label
-        elif 'label' in self._secret_cache[id]:
-            args.extend(['--label', self._secret_cache[id]['label']])
         if description is not None:
             args.extend(['--description', description])
-            self._secret_cache[id]['description'] = description
-        elif 'description' in self._secret_cache[id]:
-            args.extend(['--description', self._secret_cache[id]['description']])
         if expire is not None:
             args.extend(['--expire', expire.isoformat()])
-            self._secret_cache[id]['expire'] = expire
-        elif 'expire' in self._secret_cache[id]:
-            args.extend(['--expire', self._secret_cache[id]['expire'].isoformat()])
         if rotate is not None:
             args.extend(['--rotate', rotate.value])
-            self._secret_cache[id]['rotate'] = rotate
-        elif 'rotate' in self._secret_cache[id]:
-            args.extend(['--rotate', self._secret_cache[id]['rotate'].value])
-
-        if content is not None:
-            # Don't store the actual secret content in memory, but put a sentinel there to indicate
-            # that the content has been set during this hook.
-            self._secret_cache[id]['content'] = object()
-        elif content is None and 'content' in self._secret_cache[id]:
-            # Get the previous content from the unit agent's cache.
-            content = self.secret_get(id=id, peek=True)
 
         with tempfile.TemporaryDirectory() as tmp:
             # The content is None or has already been validated with Secret._validate_content
--- a/ops/model.py
+++ b/ops/model.py
@@ -14,6 +14,7 @@
 
 """Representations of Juju's model, application, unit, and other entities."""
 
+import collections
 import dataclasses
 import datetime
 import enum
@@ -3217,6 +3218,7 @@ class _ModelBackend:
         self._is_leader: Optional[bool] = None
         self._leader_check_time = None
         self._hook_is_running = ''
+        self._secret_cache: Dict[str, Dict[str, Any]] = collections.defaultdict(dict)
 
     def _run(
         self,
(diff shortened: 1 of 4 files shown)


ops/model.py
```diff
@@ -3640,12 +3642,31 @@ def secret_set(
...
+            self._secret_cache[id]['rotate'] = rotate
+        elif 'rotate' in self._secret_cache[id]:
+            args += ['--rotate', self._secret_cache[id]['rotate'].value]
+        if content is None and 'content' in self._secret_cache[id]:
+            # Get the previous content from the unit agent's cache.
+            content = self.secret_get(id=id, peek=True)
+        elif content is not None:
+            # Don't store the actual secret content in memory, but put a sentinel there to indicate
+            # that the content has been set during this hook.
+            self._secret_cache[id]['content'] = object()
```
dimaqq, line 3669: This was a bit hard to understand... wdyt if the clauses were reordered?

```py
sec = self._secret_cache[id]  # debatable

if content is not None:
    sec["content-set"] = True
elif sec.get("content-set"):
    content = ...
```

ops/model.py
```diff
@@ -3640,12 +3642,31 @@ def secret_set(
...
         if label is not None:
             args.extend(['--label', label])
+            self._secret_cache[id]['label'] = label
+        elif 'label' in self._secret_cache[id]:
+            args.extend(['--label', self._secret_cache[id]['label']])
         if description is not None:
             args.extend(['--description', description])
+            self._secret_cache[id]['description'] = description
+        elif 'description' in self._secret_cache[id]:
+            args.extend(['--description', self._secret_cache[id]['description']])
```
dimaqq, line 3652: Pls consider adding blank line to separate `if/elif` blocks from independent `if`'s.

dimaqq (approved):

ops/model.py
tonyandrewmeyer: This does match the if statements above, so I'll adjust to match with that.

To me, it made more sense to have them in the order that things would happen, rather than the first clause being something that never happens before the second clause is triggered.

ops/model.py
tonyandrewmeyer: This method is basically a big bunch of if statements processing the arguments, followed by the bit that does the content and the actual set call. Adding in a blank line between each of the ifs would make for a very broken up chunk of code, and it doesn't seem more readable to me. All the arg checking is done with separate if statements, but is all _logically_ grouped.

ops/model.py
```diff
@@ -3640,12 +3642,33 @@ def secret_set(
...
         if rotate is not None:
             args += ['--rotate', rotate.value]
+            self._secret_cache[id]['rotate'] = rotate
+        elif 'rotate' in self._secret_cache[id]:
+            args += ['--rotate', self._secret_cache[id]['rotate'].value]
+
+        if content is not None:
+            # Don't store the actual secret content in memory, but put a sentinel there to indicate
+            # that the content has been set during this hook.
+            self._secret_cache[id]['content'] = object()
```
benhoyt, line 3667: I wonder: does this matter, especially given that hooks are short-lived (which this cache depends on, right)? We already cache the content on a `Secret` object -- though I guess the lifetime of those are shorter-lived?

test/test_model.py
```diff
@@ -3816,6 +3816,62 @@ def test_set_info(self, model: ops.Model, fake_script: FakeScript):
...
+        calls = fake_script.calls(clear=True)
+        assert calls == [
+            ['secret-set', f'secret://{model._backend.model_uuid}/q', ANY],
+            ['secret-get', f'secret://{model._backend.model_uuid}/q', '--peek', '--format=json'],
+            [
+                'secret-set',
+                f'secret://{model._backend.model_uuid}/q',
+                '--description',
+                description,
+                ANY,
```
benhoyt, line 3837: Huh, TIL you can use `ANY` in normal `assert` statements / equality comparisons. I would have thought it was only for `assert_called_with` and co. Cool!

benhoyt (commented):
I like the fix, however, it seems to me we should do it in class `Model`, rather than in `_ModelBackend`. They're supposed to be to 1:1 wrappers around the Juju hook tools. Might be good to discuss during daily.

test/test_model.py
jameinel: It is a little ugly that this file now does:
```
from unittest.mock import ANY, MagicMock, patch
```

and
```
from typing import (
    Any,
```

So you have both an ANY and an Any and they aren't the same thing nor can be used in the same way.

https://docs.python.org/3/library/unittest.mock.html#any

I guess the implementation is just a type who always returns True for any equality test:
https://github.com/python/cpython/blob/c7a94e77bc61ed6a1ec1419748f2eaefa67e0c26/Lib/unittest/mock.py#L2462-L2474

jameinel (commented):
FWIW, if people feel like the charm hook:
```
secret-set $ID --label=foo
secret-set $ID --description="my description"
```
Should also preserve the information, then this really should be a Juju fix, and not the ops framework caching the information to pass it in a second time. 
We might want to have ops try the caching thing to make it easier to support multiple versions of Juju, but we should be pushing to have the functionality changed in Juju not just in ops.

jameinel:
also, what is the behavior if you:
```
secret-set $ID foo=bar
secret-set $ID baz=bing
```
I *think* the behavior is that content is replaced, and not additive. 

As demonstrated with:
```
$ juju exec --unit ubuntu/0 "secret-set secret:cr87vfcuth6s6ib4phog foo=bar; secret-set secret:cr87vfcuth6s6ib4phog baz=bing"
$ juju show-secret cr87vfcuth6s6ib4phog --reveal
cr87vfcuth6s6ib4phog:
  revision: 52
  owner: ubuntu-lite
  label: foo
  created: 2024-08-29T14:03:42Z
  updated: 2024-09-17T23:20:21Z
  content:
    baz: bing
```

Note that this *is* different from `relation-set`. Which instead uses:
```
Setting an empty string causes the setting to be removed. Duplicate settings
are not allowed.
```
But there if you do:
```
$ juju exec --unit dummy-source/0 "relation-set -r :0 foo=bar; relation-set -r :0 baz=bing"
$ juju exec --unit dummy-source/0 "relation-get -r :0 - dummy-source/0"
baz: bing
egress-subnets: 10.190.220.126/32
foo: bar
ingress-address: 10.190.220.126
private-address: 10.190.220.126
```

I don't think it is predictable that `secret-set` behaves differently than `relation-set`, but I'm not sure if that ship has sailed.

jameinel:
Partly the reason why secret-set has to behave differently is that we don't want to cache the known existing values to do the merge operation because of their sensitivit. So we push that onto the charm author. It could be better to find a way to keep doing that for labels, etc.

Though in my testing, calling `secret-set $ID --label=a`; `secret-set $ID --description=blah` does what you would expect:

```
$ juju exec --unit ubuntu-lite/0 "secret-set secret:cr87vfcuth6s6ib4phog --label='foo'; secret-set secret:cr87vfcuth6s6ib4phog --description='some text'"
$ juju show-secret cr87vfcuth6s6ib4phog --reveal
cr87vfcuth6s6ib4phog:
  revision: 52
  owner: ubuntu-lite
  description: some text
  label: foo
  created: 2024-08-29T14:03:42Z
  updated: 2024-09-17T23:27:38Z
  content:
    baz: bing
```

The content hasn't been updated, but the label and description are changed.

test/test_model.py
tonyandrewmeyer: For what it's worth, the introduction of `unittest.mock.ANY` was done a couple of months ago - I was just following the previous pattern. I would have personally had them both living in their namespaces so that it was `typing.Any` and `mock.ANY` (or maybe `unittest.mock.ANY`).

I'm fine reworking the tests to not use `ANY` (it does seem a bit odd to be using a `unittest.mock` object when we're not actually doing the mocks with `unittest.mock`), for example:

```python
assert calls[0][:-1] == ['secret-set', f'secret://{model._backend.model_uuid}/q', ANY]
assert calls[1] == ['secret-get', f'secret://{model._backend.model_uuid}/q', '--peek', '--format=json']
assert calls[2] == [
    'secret-set',
    f'secret://{model._backend.model_uuid}/q',
    '--description',
    description,
    ANY,
]
```

But presumably I should then adjust the existing tests as well? Or I could change the import and adjust based on that.

ops/model.py
tonyandrewmeyer: Hooks are typically short-lived, and this cache indeed is just in memory so only lives as long as the model backend does.

With the `Secret` object, we cache when you *get* the (tracked) content, but we don't when you *set* the (new latest) content. I also feel like it's also more predictable that the content might be cached on the `Secret` object, rather than on the model backend (which might be processing multiple events, for example).

My preference is to not expand the caching of secret content (this would be doubly: both new content on set, and also the model backend rather than the secret object).

ops/model.py
```diff
@@ -3640,12 +3642,33 @@ def secret_set(
...
         if expire is not None:
             args.extend(['--expire', expire.isoformat()])
+            self._secret_cache[id]['expire'] = expire
+        elif 'expire' in self._secret_cache[id]:
+            args.extend(['--expire', self._secret_cache[id]['expire'].isoformat()])
         if rotate is not None:
             args += ['--rotate', rotate.value]
+            self._secret_cache[id]['rotate'] = rotate
+        elif 'rotate' in self._secret_cache[id]:
+            args += ['--rotate', self._secret_cache[id]['rotate'].value]
```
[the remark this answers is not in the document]
tonyandrewmeyer: Odd indeed. Both the `extend` and the `+=` were done in the same commit, so I'm not sure why there's a mix. I'll change them to be consistent.

tonyandrewmeyer:
> Though in my testing, calling `secret-set $ID --label=a`; `secret-set $ID --description=blah` does what you would expect:
> 
> ```
> $ juju exec --unit ubuntu-lite/0 "secret-set secret:cr87vfcuth6s6ib4phog --label='foo'; secret-set secret:cr87vfcuth6s6ib4phog --description='some text'"
> $ juju show-secret cr87vfcuth6s6ib4phog --reveal
> cr87vfcuth6s6ib4phog:
>   revision: 52
>   owner: ubuntu-lite
>   description: some text
>   label: foo
>   created: 2024-08-29T14:03:42Z
>   updated: 2024-09-17T23:27:38Z
>   content:
>     baz: bing
> ```
> 
> The content hasn't been updated, but the label and description are changed.

Isn't that because `label` is special? For example (this is Juju 3.6b1 with LXD, if that matters):

```
$ juju exec --unit secret-info-and-content-set/5 "secret-set secret:crl1juhtvhl39c5lur90 --expire=3600s; secret-set secret:crl1juhtvhl39c5lur90 --description='some text'"
$ juju show-secret crl1juhtvhl39c5lur90 --reveal
crl1juhtvhl39c5lur90:
  revision: 1
  owner: secret-info-and-content-set/5
  description: some text
  label: my-secret
  created: 2024-09-18T00:08:27Z
  updated: 2024-09-18T00:09:53Z
  content:
    foo: bar
```

No expiry, but the description is set. And in the other order:

```
$ juju exec --unit secret-info-and-content-set/5 "secret-set secret:crl1juhtvhl39c5lur90 --description='some other text'; secret-set secret:crl1juhtvhl39c5lur90 --expire=1800s;"
$ juju show-secret crl1juhtvhl39c5lur90 --reveal
crl1juhtvhl39c5lur90:
  revision: 1
  expires: 2024-09-18T00:41:10Z
  owner: secret-info-and-content-set/5
  description: some text
  label: my-secret
  created: 2024-09-18T00:08:27Z
  updated: 2024-09-18T00:11:10Z
  content:
    foo: bar
```

The description hasn't changed, but the expiry is set. Compared to the single call where both update:

```
$ juju exec --unit secret-info-and-content-set/5 "secret-set secret:crl1juhtvhl39c5lur90 --description='some other text still' --expire=900s;"
$ juju show-secret crl1juhtvhl39c5lur90 --reveal
crl1juhtvhl39c5lur90:
  revision: 1
  expires: 2024-09-18T00:28:10Z
  owner: secret-info-and-content-set/5
  description: some other text still
  label: my-secret
  created: 2024-09-18T00:08:27Z
  updated: 2024-09-18T00:13:10Z
  content:
    foo: bar
```

test/test_model.py
benhoyt: Let's do `from unittest import mock` so we get `mock.ANY` and the like -- probably best to do that in a follow-up PR for cleanliness if you prefer.

jameinel:
> No expiry, but the description is set. And in the other order:

sigh... that seems quite inconsistent. Do we have a bug on Juju about this?
I understand why we don't want to merge secret content, but for everything else I'd rather have multiple calls actually not replace valid values with empty values.

test/test_model.py
tonyandrewmeyer: Opened #1379, and I'll adjust this one after that gets merged (which I assume will be quicker than this one).

tonyandrewmeyer:
> I like the fix, however, it seems to me we should do it in class `Model`, rather than in `_ModelBackend`.

The catch with that is that `Model` doesn't do `secret-set` calls, `Secret` does. `Secret` does not currently have a link to a `Model` object, only to a `_ModelBackend`. If we want this done in `Model`, then that would mean adjusting `Secret` to get access to a `Model` object (which would also mean adjusting `Unit` and `Application`, which create `Secret` objects).

We can't do it in individual `Secret` objects because you could have two `Secret` objects that are the same underlying secret and they should also presumably not clobber each other. I guess it could be done at the `Secret` class level?

> They're supposed to be to 1:1 wrappers around the Juju hook tools.

This is a pretty convincing argument.

> Might be good to discuss during daily.

Sure :). Also about whether we should be caching content when setting and John's comments :).

tonyandrewmeyer:
> Do we have a bug on Juju about this?

No. We (Charm-Tech) did have a talk about this issue broadly with other people in the Juju team (it would either have been at the Charm-Tech/Juju APAC sync or Charm-Tech/Ian sync, I can't remember which). The conclusion there was that `secret-set` behaves consistently within itself (each call replaces any previous call in the same hook execution), even though it's different from `relation-set` (as you described), and the problem arises from ops splitting `secret-set` up into two separate methods, one for the metadata and one for the content. So we felt like this should be fixed in ops, because ops was at fault for the limitation of not being able to set both metadata and content in the same hook.

I didn't realise that `label` was different at the time - I only found that out while testing things while working on this PR.

> Partly the reason why secret-set has to behave differently is that we don't want to cache the known existing values to do the merge operation because of their sensitivity.

It's cached somewhere though, right? Because if I do a `secret-set` and then a `secret-get --peek` and then crash the hook I get the new content back in the peek but it's not saved. Is the secret backend handling this rather than Juju? Like Juju does a transaction with the secret backend that then isn't committed/is rolled back if handling the hook fails?

Even if the caching is in a transaction of sorts in the secret backend, doesn't that still make it available to Juju? `secret-get --peek` works somehow :).

> I understand why we don't want to merge secret content, but for everything else I'd rather have multiple calls actually not replace valid values with empty values.

I can open a bug for this and let the Juju'ers hammer out whether it should happen :).

jameinel:
On Tue, Sep 17, 2024 at 9:15 PM Tony Meyer ***@***.***> wrote:

> Do we have a bug on Juju about this?
>
> No. We (Charm-Tech) did have a talk about this issue broadly with other
> people in the Juju team (it would either have been at the Charm-Tech/Juju
> APAC sync or Charm-Tech/Ian sync, I can't remember which). The conclusion
> there was that secret-set behaves consistently within itself (each call
> replaces any previous call in the same hook execution), even though it's
> different from relation-set (as you described), and the problem arises
> from ops splitting secret-set up into two separate methods, one for the
> metadata and one for the content. So we felt like this should be fixed in
> ops, because ops was at fault for the limitation of not being able to set
> both metadata and content in the same hook.
>
> I didn't realise that label was different at the time - I only found that
> out while testing things while working on this PR.
>
> Partly the reason why secret-set has to behave differently is that we
> don't want to cache the known existing values to do the merge operation
> because of their sensitivity.
>
> It's cached somewhere though, right? Because if I do a secret-set and
> then a secret-get --peek and then crash the hook I get the new content
> back in the peek but it's not saved. Is the secret backend handling this
> rather than Juju? Like Juju does a transaction with the secret backend that
> then isn't committed/is rolled back if handling the hook fails?
>
> Even if the caching is in a transaction of sorts in the secret backend,
> doesn't that still make it available to Juju? secret-get --peek works
> somehow :).
>

I think it is cached only in the memory of the unit agent, which is why it
rolls back if you crash the hook.
John
=:->

ops/model.py
tonyandrewmeyer: The code was moved from here anyway :)

ops/model.py
tonyandrewmeyer: Discussed during daily and there were no objections to avoiding additional caching of sensitive content.

tonyandrewmeyer:
> Tony will try the dict at the class level, if that doesn't seem great, then explicitly wire the secret_cache dict through Model/Application/Unit/Secret.

The implementation, including ops tests themselves, is ok ([commit](https://github.com/canonical/operator/pull/1373/commits/3752955b8c00bbfe72dc3ad120235faa6e84bcff).

For charm unit tests, Scenario can just reset the cache for each `run`. Harness is much trickier, because there isn't really a defined way of saying "this is a new Juju hook execution". We could reset each time you make a new `Harness` object (or each time the charm is instantiated, which is roughly the same) but it does increase the behaviour that is unexpectedly different from running against an actual Juju. I'm concerned that this would end up strangely leaking secret info/metadata between tests, although maybe the answer is to just make a new `Harness` every time you want a new Juju hook execution.

I'll have a look at the alternative of passing a cache down to `Application` and `Unit` and through those to `Secret`.

benhoyt (approved):
Looks good to me, thanks! Can we make sure to super-tox this?

benhoyt:
> I'll have a look at the alternative of passing a cache down to Application and Unit and through those to Secret.

Sorry, I jumped the gun on that review. Thanks!

tonyandrewmeyer:
> I'll have a look at the alternative of passing a cache down to `Application` and `Unit` and through those to `Secret`.

There was less wiring here than I anticipated, because I forgot we already have `_ModelCache`, and it seems reasonable to me to store the cache there. I think this version is much better.

> Can we make sure to super-tox this?

Yes, will do. I didn't with the first version because a `_ModelBackend` change wouldn't change anything for unit tests, but this change potentially does.

tonyandrewmeyer:
> > Can we make sure to super-tox this?
> 
> Yes, will do. I didn't with the first version because a `_ModelBackend` change wouldn't change anything for unit tests, but this change potentially does.

There are no new failures. That's to be expected, I think, because the Harness behaviour was already what we are changing this to, so as long as `Model` is now doing what `Harness` already did, the test behaviour isn't going to change.

All of the `set_info` use I can find is in the tls-certificates lib, and that does do both `set_content` and `set_info` together (although perhaps not in an active code path, based on the comment from #1288?). As far as I can tell, the existing code will not work correctly, and this change will make it work correctly, but I haven't found any tests around it (and if there were tests, then you'd think that someone would have noticed this already).

tonyandrewmeyer:
I opened [a Juju ticket](https://bugs.launchpad.net/juju/+bug/2081034) where the Juju team can decide if they want to keep the existing `secret-set` behaviour or improve it to match the new ops behaviour.

benhoyt (commented):
Looks good overall -- couple of potential minor issues.

ops/model.py
```diff
@@ -304,7 +305,13 @@ def get_secret(self, *, id: Optional[str] = None, label: Optional[str] = None) -
             # Canonicalize to "secret:<id>" form for consistency in backend calls.
             id = Secret._canonicalize_id(id, self.uuid)
         content = self._backend.secret_get(id=id, label=label)
-        return Secret(self._backend, id=id, label=label, content=content)
+        return Secret(
```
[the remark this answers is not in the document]
tonyandrewmeyer: Nice catch, thanks. I had noticed that one earlier but then totally forgot to grep again for all the `Secret(` cases.

ops/model.py
```diff
@@ -1279,6 +1300,7 @@ def __init__(
         self._id = id
         self._label = label
         self._content = content
+        self._secret_set_cache = _secret_set_cache or collections.defaultdict(dict)
```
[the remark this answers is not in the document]
tonyandrewmeyer: > Wouldn't this have to be an `is None` check, otherwise if the passed-in cache dict is empty it'll evaluate to false here and always create a new dict.

Yes, good catch, thanks!

> I guess we could make it a required argument after `backend` instead, and make it non-optional?

Although the docs say

> This class should not be instantiated directly

that would still break backwards compatibility, so I'd prefer to not do that.

test/test_model.py
```diff
@@ -3823,6 +3823,62 @@ def test_set_info(self, model: ops.Model, fake_script: FakeScript):
...
+                f'secret://{model._backend.model_uuid}/q',
+                '--description',
+                description,
+            ],
+            [
+                'secret-set',
+                f'secret://{model._backend.model_uuid}/q',
+                '--description',
+                description,
+                mock.ANY,
```
dimaqq, line 3874: I guess this represents secret content, but not obvious on the first read.

Wdyt about adding a comment, for example like so?

```suggestion
                mock.ANY,  # foo#file=<path with `bar` content>
```

test/test_model.py
```diff
@@ -3823,6 +3823,62 @@ def test_set_info(self, model: ops.Model, fake_script: FakeScript):
...
+            ],
+        ]
+        # For this test we don't need to check that the content was in the temporary file, but we
+        # do want to be certain that there was a file passed to the call (but don't care about
+        # what the name of the file was).
+        assert calls[1][-1].startswith('foo#file=') and calls[1][-1].endswith('/foo')
```
dimaqq, line 3880: The extensive comment here suggest that a small refactor is in order.
Besides, if the content is validate in the other tests, there's no need for it to behere.

I do wonder about something else though:
What if user calls:

```py
.set_content({"foo": "bar"})
.set_content({"baz": "qux"})
```

Do we aggregate or not?
If we do in ops, the test needs to check that.

if we don't, perhaps it's better to be explicit about that too.

dimaqq (changes requested):
Not a request for change, rather a critical question about juju agent behaviour

ops/model.py
```diff
@@ -1483,7 +1509,26 @@ def set_content(self, content: Dict[str, str]):
...
+        # If `_backend.secret_set` has already been called, this call will
+        # overwrite anything done in the previous call (even if it was in a
+        # different event handler, as long as it was in the same Juju hook
+        # context). We want to provide more predictable behavior, so we cache
+        # the values and provide them in subsequent calls.
+        cached_data = self._secret_set_cache[self._id]
+        expire = _calculate_expiry(cached_data['expire']) if 'expire' in cached_data else None
+        # Don't store the actual secret content in memory, but put a sentinel
+        # there to indicate that the content has been set during this hook.
+        cached_data['content'] = object()
```
dimaqq, line 1522: I think GitHub ate my home^H^H^H^Hprevious comment

Is this safe in this scenario:

```py
.set_content(...)  # passed via temp file to juju agent
# temp dir ctx manager exits, removes the temp files

.set_info(...)  # another hook tool invocation
```

Does juju agent combine the value with the info in this case?

If it doesn't I guess we've got no choice to cache the value somewhere, either in memory or in longer-lived temporary directory/file.

tonyandrewmeyer: > Is this safe in this scenario:
> 
> ```python
> .set_content(...)  # passed via temp file to juju agent
> # temp dir ctx manager exits, removes the temp files
> 
> .set_info(...)  # another hook tool invocation
> ```
> 
> Does juju agent combine the value with the info in this case?

No (it never combines), but it reads from the file(s) used in the `set_content` during the `secret-set` call. There's no need for those files to stick around after the `secret-set` is finished. Otherwise our use of temp files wouldn't work - we'd have to clean them up after the hook completed somehow.

Ops does (in this branch) combine the two, though. It gets the content from `secret-get` and passes it back (in brand new temporary files).

It is inefficient to do `set_info()` and then `set_content()` compared with a single `secret-set` call - and you're better off doing the `set_info` first if you can, because then at least you don't have ops writing out to the files more than once (and Juju reading from them). I played around with having a new method that let you do both or extending `set_info` to let you pass content as well, but a new method felt too heavy unless we were going to deprecate the other two, and `set_info` seems confusingly named if it can do content as well. I think we'll want to redesign this eventually, but not any time soon.

test/test_model.py
tonyandrewmeyer: There are 8 other cases of `mock.ANY` in these tests, all, I believe, meaning "file with secret content". It seems odd to comment this one and not any of the others, particularly this in this test it does actually explain what it is by verifying it right after that line, which the others don't (although they generally validate the content in the file).

I'm not opposed to this, but it does seem like it should be done consistently, and it feels like that would be better in another PR (like the one that changed `ANY` to `mock.ANY`).

tonyandrewmeyer: I've reworked the tests to not use `mock.ANY`. It's a bit wordier and it's no longer consistent with the other tests, but it avoids all the issues.

test/test_model.py
tonyandrewmeyer: Added a test for this, thanks!

dimaqq (approved):

PR #1379, opened 2024-09-18
refactor: keep the unittest.mock names in the 'mock' namespace

tonyandrewmeyer:
No functional changes (including to the tests) - this just adjusts the imports so that the names from `unittest.mock` are imported into a `mock` namespace, rather than all into the module namespace.

Primarily, this ensures that `typing.Any` and `unittest.mock.ANY` are clearly distinguished by more than case, but it's also generally tidier ("Namespaces are one honking great idea").

See discussion in #1373.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -24,7 +24,7 @@ import typing
 import unittest
 from collections import OrderedDict
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


benhoyt (approved):

james-garner-canonical (approved):
Looks good to me

PR #1381, opened 2024-09-18
feat: optionally install Scenario with `ops[testing]` and expose the names in ops.testing

tonyandrewmeyer:
Add a new optional install `testing`, e.g. `pip install ops[testing]`. This pulls in a compatible version of `ops-scenario`, and exposes the Scenario names in the `ops.testing` namespace, alongside Harness.

`pip install ops[harness]` is also supported to ease the (far off, presumably 3.0) transition to Harness being moved out of the base install. It currently installs no extra dependencies, so is the same as `pip install ops` but a forward-looking charm would use `pip install ops[harness]` (or `pip install ops[harness, testing]`) if using Harness.

Requires ops-scenario 7.0.5, which has the required adjustments to support insertion into `ops.testing`.

The `ActionFailed` name exists in both `ops._private.harness` and `scenario.context`. This is handled by adjusting the Harness class to support the functionality of both and monkeypatching that into Scenario until Scenario starts using it. It's compatible with both Harness and Scenario, but will have empty data in an attribute (which attribute depends on which framework is used).

The `Container` name in `ops.testing`, which is only present for backwards compatibility, is also overwritten if ops-scenario is installed. If anyone is actually using `ops.testing.Container` instead of `ops.Container` then they'll need to fix their code before using `ops[testing]` (or ops-scenario combined with the release of ops with this change).

A very basic unit test is added to make sure that Scenario tests work (all the actual Scenario tests are in the ops-scenario repo) if ops-scenario/ops[testing] is installed (this is the case for `tox -e unit`). A test is also added to ensure that all of the Scenario names are documented, since `automodule` isn't used any more.

Also adjusts the documentation to include the new framework in ops.testing.


tonyandrewmeyer:
@PietroPasotti @benhoyt @dimaqq I imagine this PR will have a bit of back-and-forth before it's completely done. Would you please give it an initial look-over? Probably focusing on the main approaches rather than specifics for now, and then once everyone is happy with that there can be a second review for approval that goes deeper. Thanks!

tonyandrewmeyer:
(I'll look into the failing test with Python 3.10 - I hadn't checked different Pythons locally, so will do that).

dimaqq:
CI fail: `TypeError: EventBase.__init__() got an unexpected keyword argument 'app'`

3.12 was not run / did not get to complete because 3.10 failed.

tonyandrewmeyer:
> CI fail: `TypeError: EventBase.__init__() got an unexpected keyword argument 'app'`
> 
> 3.12 was not run / did not get to complete because 3.10 failed.

Yeah, the tests all run locally for me with 3.8, 3.9, 3.10, 3.11, 3.12. Still looking into this, but it seems unlikely that it needs to block a high-level review.

docs/requirements.txt
```diff
@@ -73,6 +73,10 @@ mdurl==0.1.2
     # via markdown-it-py
 myst-parser==4.0.0
     # via ops (pyproject.toml)
+ops==2.16.1
```
dimaqq, line 76: 🤔 does ops belong here? I'd imagine that docs are generated from the current repo.

Does it hurt? e.g. if someone is working on a branch, could this pull in different version of code?

ops/testing.py
```diff
@@ -38,8 +41,144 @@
...
-_ = ActionFailed
+# If the 'ops.testing' optional extra is installed, make those
+# names available in this namespace.
+try:
+    _version = _get_package_version('ops-scenario')
+except _PackageNotFoundError:
+    from ops._private.harness import ActionFailed  # type: ignore
+else:
+    if _version and int(_version.split('.', 1)[0]) >= 7:
+        from scenario import (
```
dimaqq, line 52: 🤔 don't we want to hide the name "scenario" from like everywhere?

dimaqq:
A point for discussion about the overall approach:
- do we want to keep separate pypi:ops-scenario going forward?
- or do shall we vendor scenario code and rename things?

docs/requirements.txt
tonyandrewmeyer: > 🤔 does ops belong here? I'd imagine that docs are generated from the current repo.

This file is auto-generated. ops ends up here because there's a circle: `ops-scenario` depends on `ops` (that could possibly be removed in time, but it would be a bunch of refactoring work) and `ops[testing]` depends on `ops-scenario`. The docs need to have `ops-scenario` installed to be able to pull in all the docstrings, and installing `ops-scenario` installs `ops`.

> Does it hurt? e.g. if someone is working on a branch, could this pull in different version of code?

Good question! It seems to work ok - I edited one of the ops modules locally and did `tox -e docs` and it picked up my local changes. I'm not totally sure about when readthedocs does the build - the config has:

```yaml
python:
  install:
    - method: pip
      path: .
      extra_requirements:
        - docs
```

So it's ignoring docs/requirements.txt I believe. I'm not sure what would happen if a branch of `ops` had a version number incompatible with the requirements in `ops-scenario` - probably an error? But that does seem like to be something we can manage ourselves, especially once we've merged the repos.

ops/testing.py
tonyandrewmeyer: > 🤔 don't we want to hide the name "scenario" from like everywhere?

I think we want (or are expected to) hide it so that it's not used in charms, in our docs, etc. Essentially: you can build charms and write tests for them without ever knowing the name "Scenario".

It's not 100% hidden, e.g. you can do this:

```python
>>> repr(ops.testing.Context)
"<class 'scenario.context.Context'>"
```

Although arguably `Context.__repr__` could be set to something better. But there's still things like:

```python
>> ops.testing.State.__module__
'scenario.state'
```

I don't feel that the "I want to reserve the name Scenario" comment means that we need to entirely wipe all traces of it.

tonyandrewmeyer:
> * do we want to keep separate pypi:ops-scenario going forward?
> * or do shall we vendor scenario code and rename things?

We have to have a separate package in order to (cleanly) have one set of code that gets used when building charms and an augmented set of code that gets used when testing charms. I very strongly feel that the test framework should not be in the bundled charm. I think the correct way to do that is to have a package for testing (eventually two: one for Harness and one for Scenario) rather than things like charmcraft ripping out the folder when packing.

We are intending to add the contents of the ops-scenario repository to the operator repository, but my preference is to do that as a separate PR after this one. We would then publish two packages from the operator repository. I still feel that's going to be a better maintenance burden than having two repos. I don't really see a benefit in renaming things as part of this, but that can be discussed when we get to it.

We could publish Scenario under a new name, like `ops-testing`. I don't see much value in that either, though, and I like having a small nod to the history of how this framework came to be.

tonyandrewmeyer:
I can reproduce the failing test locally if I turn pytest-xdist off (probably other small numbers work too). I assume some other test is leaking into the one that's failing, so it only fails when both of those tests are executed in the same runner.

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+  Juju events. This is currently included with a base ``ops`` install.
+
+.. note::
+    Unit testing is only one aspect of a comprehensive testing strategy. For more
+    on testing charms, see `Charm SDK | Testing <https://juju.is/docs/sdk/testing>`_.
+
+State-transition testing
+------------------------
+
+Write tests that declaratively define the Juju state all at once, define the
```
benhoyt, line 55: This reads a bit odd, like a subject is missing. I think we should say "State-transition tests allow you to ..." or something.

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+Juju context against which to test the charm, and fire a single event on the
+charm to execute its logic. The tests can then assert that the Juju state has
+changed as expected.
```
benhoyt, line 58: There's a lot of text after this point before the developer can see anything in action. I think it'd be great to show a tiny example right here:

```python
def test_start():
    ctx = Context(MyCharm)
    state = State(...something small here...)
    out = ctx.run(ctx.on.start(), state)
    assert isinstance(out.unit_status, ActiveStatus)
```

Maybe we want *something* really simple in the `State` so readers can see that.

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+
+These tests are 'state-transition' tests, a way to test isolated units of charm
+functionality (how the state changes in reaction to events). They are not
+necessarily tests of individual methods or functions (but might be, depending on
+the charm's event observers); they are testing the 'contract' of the charm: given
+a certain state, when a certain event happens, the charm should transition to a
+certain (likely different) state. They do not test against a real Juju
+controller and model, and focus on a single Juju unit, unlike integration tests.
+For simplicity, we refer to them as 'unit' tests in the charm context.
```
benhoyt, line 67:
```suggestion
These 'state-transition' tests give charm authors a way to test
how the state changes in reaction to events. They are not
necessarily tests of individual methods or functions;
they are testing the 'contract' of the charm: given
a certain state, when a certain event happens, the charm should transition to
another state. Unlike integration tests, they do not test using a real Juju
controller and model, and focus on a single Juju unit.
For simplicity, we refer to them as 'unit' tests.
```

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+
+Writing these tests should nudge you into thinking of a charm as a black-box
+input->output function. The input is the union of an `Event` (why am I, charm,
+being executed), a `State` (am I leader? what is my integration data? what is my
+config?...) and the charm's execution `Context` (what integrations can I have?
+what containers can I have?...). The output is another `State`: the state after
+the charm has had a chance to interact with the mocked Juju model and affect the
+state.
```
benhoyt, line 75:
```suggestion
Writing these tests should nudge you into thinking of a charm as a black-box
'input to output' function. The inputs are:

- Event: why am I, the charm, being executed
- State: am I the leader? what is my integration data? what is my config?
- Context: what integrations can I have? what containers can I have?

The output is another `State`: the state after
the charm has interacted with the mocked Juju model.
The output state is the same type of data structure as the input state.
```

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+
+.. image:: https://raw.githubusercontent.com/canonical/ops-scenario/main/resources/state-transition-model.png
+   :alt: Transition diagram, with the input state and event on the left, the context including the charm in the centre, and the state out on the right
+
+Writing unit tests for a charm, then, means verifying that:
+
+- the charm does not raise uncaught exceptions while handling the event
+- the output state (as compared with the input state) is as expected.
```
benhoyt, line 83: I would reverse those two bullet points as the "output state" one is by far the most important.

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+
+A test consists of three broad steps:
+
+- **Arrange**:
+    - declare the context
+    - declare the input state
+- **Act**:
+    - select an event to fire
+    - run the context (i.e. obtain the output state, given the input state and the event)
```
benhoyt, line 92: "run the context" sounds odd. Maybe this could just be one bullet point with "run a specific Juju event"?

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+- **Assert**:
+    - verify that the output state (as compared with the input state) is how you expect it to be
```
benhoyt, line 94:
```suggestion
    - verify that the output state is what you expect it to be
```

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+    - verify that the charm has seen a certain sequence of statuses, events, and `juju-log` calls
+    - optionally, you can use a context manager to get a hold of the charm instance and run
```
benhoyt, line 96: Maybe drop this bullet point altogether? We don't need to mention this more advanced feature right here.

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+      assertions on APIs and state internal to it.
+
+The most basic scenario is one in which all is defaulted and barely any data is
```
benhoyt, line 99: The wording here is a bad odd/casual. But maybe if we start with a tiny example above, we don't need this paragraph or example at all?

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+status is `unknown`. With that, we can write the simplest possible test:
+
+.. code-block:: python
+
+    def test_base():
+        ctx = Context(MyCharm)
+        out = ctx.run(ctx.on.start(), State())
+        assert out.unit_status == UnknownStatus()
+
+.. autoclass:: ops.testing.ActionFailed
```
benhoyt, line 110: It's a pity we have to list these out manually? Is there any way to avoid that, so that we don't forget to add new classes here when we add them to the code?

benhoyt (commented):
Adding some further comments on the docs.

ops/charm.py
```diff
@@ -240,7 +241,7 @@ class StartEvent(HookEvent):
     """Event triggered immediately after first configuration change.
 
     This event is triggered immediately after the first
-    :class:`ConfigChangedEvent`. Callback methods bound to the event should be
+    :class:`~ops.ConfigChangedEvent`. Callback methods bound to the event should be
```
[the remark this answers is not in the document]
tonyandrewmeyer: Scenario copies various docstrings from ops objects, particularly the statuses and the events. When there's something like `:class:ConfigChangedEvent` in the Scenario objects and Sphinx does `autoclass`, it can't find the `ConfigChangedEvent` object, so fails. With the leading `ops.` Sphinx knows where to find them even when the docstring is being used outside of ops. The `~` tells Sphinx to just display the name and not the `ops.`.

The docstring copying code could try to be smarter and auto-fix these (it's probably just replacing ":class:" with "class:ops." (and the backticks), with maybe some false positives to work around), or Scenario could have its own docstrings (maybe just saying "go read the ops class for more details").

Discussed this in a call and Ben's ok with leaving them as-is for the moment, now that it's clearer what they are there for.

ops/testing.py
```diff
@@ -14,8 +14,11 @@
 
 """Infrastructure to build unit tests for charms using the ops library."""
 
+import typing as _typing
+from importlib.metadata import PackageNotFoundError as _PackageNotFoundError
+from importlib.metadata import version as _get_package_version
```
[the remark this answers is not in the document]
tonyandrewmeyer: It was because I couldn't get Sphinx to properly work with `__all__` when using `automodule`. However, I had to give up using `automodule` anyway, because I couldn't find a clean way to get it to use `ops.testing` rather than `ops._private.harness` and `scenario` for references.

So that means that `__all__` is actually viable - I'll go back to that. I'll also rely on `__all__` defining the API and not bother importing `typing as _typing` and so on.

ops/testing.py
```diff
@@ -38,8 +41,144 @@
...
+            StoredState,
+            SubordinateRelation,
+            TCPPort,
+            UDPPort,
+            UnitID,
+            UnknownStatus,
+            WaitingStatus,
+            errors,
+        )
+        from scenario.context import CharmEvents
```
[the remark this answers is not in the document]
tonyandrewmeyer: I think it was accidentally not included in the `scenario` top level: https://github.com/canonical/ops-scenario/pull/200

docs/index.rst
tonyandrewmeyer: I do agree. I couldn't get `automodule` to work and also use the `ops.testing` name rather than leaking `ops._private.harness` or `scenario`. It's somewhat working against the `automodule` intention since the usual desired behaviour is to document anything defined in the module but not anything that the module pulls in to use - but here we have basically nothing defined in the module, and want to document (almost) everything that's pulled in.

There could be some combination of Sphinx configuration options that make it work, but I wasn't able to find them with the limited amount of time I allowed for it. It's also possible that we could create our own `automodule` directive (subclassing or patching the original one, probably) that would solve our issues - but this seemed like it would be a lot more work.

For now, I've added a test_infra test to make it less likely that we forget to maintain this, and a comment about why we do this, and hopefully someone in the future will make this nicer.

tonyandrewmeyer:
@tmihoc the code changes won't be interesting to you, but you might like to review the doc changes? It's probably easiest to look over the [generated docs](https://ops--1381.org.readthedocs.build/en/1381/) since it pulls in the definitions from ops-scenario as well (which you already reviewed, of course).

docs/index.rst
```diff
@@ -34,7 +34,148 @@ ops.pebble module
...
+  extra to use this framework; for example: ``pip install ops[testing]``
+* Harness, which provides an API similar to the Juju CLI. This is a legacy
+  framework, and has issues, particularly with resetting the charm state between
+  Juju events. This is currently included with a base ``ops`` install.
+
+.. note::
+    Unit testing is only one aspect of a comprehensive testing strategy. For more
+    on testing charms, see `Charm SDK | Testing <https://juju.is/docs/sdk/testing>`_.
+
+State-transition testing
```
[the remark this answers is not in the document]
tonyandrewmeyer: I've split it out - you might want to take another look now?

ops/testing.py
```diff
@@ -56,27 +61,140 @@
...
+    'ReadableBuffer',
+    'YAMLStringOrFile',
+]
+
+# If the 'ops.testing' optional extra is installed, make those
+# names available in this namespace.
+try:
+    _version = importlib.metadata.version('ops-scenario')
+except importlib.metadata.PackageNotFoundError:
+    pass
```
benhoyt, line 83: Do we need `_version = None` here, so the `if _version` below doesn't raise a `NameError`?

docs/harness.rst
```diff
@@ -0,0 +1,31 @@
+:orphan:
+
+.. _harness:
+
+Harness Unit Test Framework
+===========================
+
+.. deprecated:: 2.17
+    The Harness framework is deprecated and will be moved out of the base
+    install in a future ops release.
```
benhoyt, line 10:
```suggestion
    install in an ops release in the future. Charm authors that don't want to upgrade
    will still be able to use it with ``pip install ops[harness]``.
```

docs/state-transition-testing.rst
```diff
@@ -0,0 +1,120 @@
+:orphan:
+
+.. _state-transition-tests:
+
+Unit Test Framework
```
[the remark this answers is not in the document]
benhoyt: Tony and I roughly agree on changing the menu to this:

* API reference
* Unit testing (was "Scenario")
* Harness (legacy unit testing)

Perhaps also say `State-transition tests (previously known as "Scenario") ...` in the page content.

benhoyt (commented):
Overall looking great, just comments on doc that we've discussed.

ops/_private/harness.py
```diff
@@ -161,18 +172,35 @@ class ActionOutput:
...
+    def __init__(
+        self,
+        message: str,
+        output: typing.Optional[ActionOutput] = None,
+        *,
+        state: typing.Optional['State'] = None,
+    ):
         self.message = message
-        self.output = output
+        self.output = ActionOutput([], {}) if output is None else output
```
dimaqq, line 202: This would be the same, wouldn't it?

```suggestion
        self.output = output or ActionOutput([], {})
```

ops/__init__.py
```diff
@@ -14,7 +14,7 @@
 
 """The API to respond to Juju events and manage the application.
 
-This module provides code freatures to your charm, including:
+This module provides code features to your charm, including:
```
dimaqq, line 17: I'm pretty sure "core features" was intended...

dimaqq (approved):

ops/testing.py
tonyandrewmeyer: No, because the `if _version` is in the try's `else` block, so if the exception is raised it won't get to there.

docs/index.rst
```diff
@@ -7,10 +7,18 @@ The `ops` library is a Python framework for writing and testing Juju charms.
...
 
-- :ref:`ops_main_entry_point`, used to initialise and run your charm;
 - :ref:`ops_module`, the API to respond to Juju events and manage the application;
+- :ref:`ops_main_entry_point`, used to initialise and run your charm;
 - :ref:`ops_pebble_module`, the Pebble client, a low-level API for Kubernetes containers;
-- :ref:`ops_testing_module` frameworks for unit testing charms in a simulated environment;
+- the APIs for unit testing charms in a simulated environment:
+
+  - :doc:`State-transition testing </state-transition-testing>`. This is the
+    recommended approach.
```
benhoyt, line 16:
```suggestion
    recommended approach (it was previously known as 'Scenario').
```

docs/index.rst
```diff
@@ -7,10 +7,18 @@ The `ops` library is a Python framework for writing and testing Juju charms.
...
+  - :doc:`Harness </harness>`. This is a deprecated framework, and has issues,
+    particularly with resetting the charm state between Juju events. It will be
+    moved out of the base ``ops`` install in an ops release in the future. Charm
```
benhoyt, line 19:
```suggestion
    moved out of the base ``ops`` install in an ops release in the future; at that point, charm
```

benhoyt: It also seems like too much information for this point in the docs. Can we move everything from "It will be moved out..." to the Harness page?

benhoyt (approved):
Approving! Left a couple of minor take-or-leave comments.

docs/index.rst
tonyandrewmeyer: Happy to move it - in terms of "at that point": I updated the description to include this but didn't mention it in a comment: I added the ability to do `pip install ops[harness]` now (this was in the spec, but I had forgotten). This eases transition in the future, because charms can use `pip install ops[harness]` now and it should seamlessly work when it becomes separated.

PR #1382, opened 2024-09-19
fix: Use a TypeVar to fix annotations for StatusBase.register

### LINKED ISSUE #1380: StatusBase.register decorator obscures the type of decorated classes
`StatusBase.register` is typed as follows:
```python
class StatusBase:
    ...
    @classmethod
    def register(cls, child: Type['StatusBase']):
        ...
        return child
```
And used like so:
```python
@StatusBase.register
class ActiveStatus(StatusBase):
    ...
```
Which leads to:
```python
reveal_type(ActiveStatus)   # Type of "ActiveStatus" is "type[StatusBase]"
```

I believe the solution is to use a type variable with a bound of `Type[StatusBase]` instead and annotate the return type, e.g.
```python
_StatusBaseType = typing.TypeVar('_StatusBaseType', bound='Type[StatusBase'])

class StatusBase:
    ...
    @classmethod
    def register(cls, child: _StatusBaseType) -> _StatusBaseType:
        ...
        return child
```
Which results in the expected:
```python
reveal_type(ActiveStatus)   # Type of "ActiveStatus" is "type[ActiveStatus]"
```

james-garner-canonical:
Previously `StatusBase.register` obscured the type of decorated classes. The classes would all end up typed as `Type[StatusBase]`, and when instantiated they would be typed as `StatusBase` instead of their actual class (see #1380). This PR fixes the typing to align with the actual decorator behaviour, which returns the decorated class unmodified.

---

This change will produce a type checking error if you were doing something like this (as one of our tests did):
```python
status = BlockedStatus("msg")
...
status = ActiveStatus()
```
Here the author's intent is for `status` to have a type of `StatusBase` (or they're making a mistake). Previously, `status` would implicitly be typed as `StatusBase` due to the `StatusBase.register` class decorator applied to `BlockedStatus` (etc) changing its type to `Type[StatusBase]`. Now `status` is instead accurately typed as `BlockedStatus`, so `status = ops.ActiveStatus()` is an error, and the author would need to write this instead:
```python
status: StatusBase = BlockedStatus("msg")
...
status = ActiveStatus()
```

---

Initially this PR broke docs generation, because sphinx has trouble with type variables. I think this open [issue](https://github.com/sphinx-doc/sphinx/issues/10974) for sphinx (Nov 2022) is related. See also the related known [issue](https://github.com/sphinx-doc/sphinx/issues/10785) with autodoc (Aug 2022). Our current workaround is to list our type variables in the `nitpick_ignore`s in `docs/custom_conf.py`.

This is what the generated docs look like. I put a copy of the TypeVar declaration in the docstring since sphinx refuses to link to it.
image

---

Resolves #1380.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -78,7 +78,7 @@ _StatusDict = TypedDict('_StatusDict', {'status': StatusName, 'message': str})
 _SETTABLE_STATUS_NAMES: Tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
 
 # for the StatusBase.register class decorator
-_StatusBaseType = typing.TypeVar('_StatusBaseType', bound=Type['StatusBase'])
+_StatusClass = typing.TypeVar('_StatusClass', bound='StatusBase')
 
 # mapping from relation name to a list of relation objects
 _RelationMapping_Raw = Dict[str, Optional[List['Relation']]]
@@ -1940,8 +1940,19 @@ class StatusBase:
             return cls._statuses[typing.cast(StatusName, name)](message)
 
     @classmethod
-    def register(cls, child: _StatusBaseType) -> _StatusBaseType:
-        """Register a Status for the child's name."""
+    def register(cls, child: Type[_StatusClass]) -> Type[_StatusClass]:
+        """Class decorator to register a subclass for lookup using :meth:`from_name`.
+
+        .. code-block:: python
+
+            _StatusClass = typing.TypeVar('_StatusClass', bound='StatusBase')
+
+        Args:
+            child: A subclass of StatusBase, with a ``name`` attribute of type ``str``.
+
+        Returns:
+            The decorated class, unmodified.
+        """
         if not (hasattr(child, 'name') and isinstance(child.name, str)):
             raise TypeError(
                 f"Can't register StatusBase subclass {child}: ",
--- a/ops/model.py
+++ b/ops/model.py
@@ -77,6 +77,9 @@ StatusName = Union[_SettableStatusName, _ReadOnlyStatusName]
 _StatusDict = TypedDict('_StatusDict', {'status': StatusName, 'message': str})
 _SETTABLE_STATUS_NAMES: Tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
 
+# for the StatusBase.register class decorator
+_StatusBaseType = typing.TypeVar("_StatusBaseType", bound="Type[StatusBase]")
+
 # mapping from relation name to a list of relation objects
 _RelationMapping_Raw = Dict[str, Optional[List['Relation']]]
 # mapping from container name to container metadata
@@ -1937,7 +1940,7 @@ class StatusBase:
             return cls._statuses[typing.cast(StatusName, name)](message)
 
     @classmethod
-    def register(cls, child: Type['StatusBase']):
+    def register(cls, child: _StatusBaseType) -> _StatusBaseType:
         """Register a Status for the child's name."""
         if not (hasattr(child, 'name') and isinstance(child.name, str)):
             raise TypeError(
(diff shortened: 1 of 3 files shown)


dimaqq:
/home/docs/checkouts/readthedocs.org/user_builds/ops/checkouts/1382/ops/model.py:docstring of ops.model.StatusBase.register:1: WARNING: py:class reference target not found: ops.model._StatusBaseType [ref.class]

in the docs build 🤔

benhoyt (approved):
Looks good to me thanks. Yeah, I agree with you that `register` should be private (or maybe not even exist), but I think we should probably just leave it as is for now.

james-garner-canonical:
Thanks @benhoyt. Do you think it's worth adding a note to the `register` docstring that the method shouldn't be called? Maybe:
```
Note: this method is intended for internal use only, and is used to make the six valid Juju statuses available by name.
```

benhoyt:
Yeah, sounds good -- go for it.

benhoyt (approved):

ops/model.py
```diff
@@ -1937,8 +1940,22 @@ def from_name(cls, name: str, message: str):
...
+        .. code-block:: python
+
+            _StatusClass = typing.TypeVar('_StatusClass', bound='StatusBase')
+
+        Args:
+            child: A subclass of StatusBase, with a ``name`` attribute of type ``str``.
+
+        Returns:
+            The decorated class, unmodified.
+        """
```
dimaqq, line 1958: Should the name of the method get an underscore, if it's for internal use only?

cc @benhoyt for the balance of very hypothetically breaking a public-ish method on a public class.

james-garner-canonical: I think it would be nice if it had one, and it doesn't seem to be used elsewhere, but perhaps that's out of scope for this PR?

dimaqq: Ah, I see this has been discussed already.

ops/model.py
```diff
@@ -1937,8 +1940,22 @@ def from_name(cls, name: str, message: str):
             return cls._statuses[typing.cast(StatusName, name)](message)
 
     @classmethod
-    def register(cls, child: Type['StatusBase']):
-        """Register a Status for the child's name."""
+    def register(cls, child: Type[_StatusClass]) -> Type[_StatusClass]:
+        """Class decorator to register a subclass for lookup using :meth:`from_name`.
+
+        Note: this method is intended for internal use only.
+        It is used to make the six valid Juju statuses available by name.
```
tonyandrewmeyer, line 1947: I would drop the "six" - there's no current intention to add a 7th as far as I know, but if that does happen then it seems very likely that the "six" here would get missed when we added support since nothing else in this method would need to change. I don't the number is particularly important in the doc.

```suggestion
        It is used to make the valid Juju statuses available by name.
```

james-garner-canonical:
Worth considering [Dima's suggestion](https://github.com/canonical/operator/issues/1380#issuecomment-2362783689) to just not use `StatusBase.register` as a decorator. Then it can be typed as just
```python
@classmethod
def register(cls, child: Type['StatusBase']) -> None:
    ...
```
and called somewhere else. My only concern here is where to call it. Directly after each subclass, by convention? All in one place after all the subclasses? Arguably this is the perfect use case for `__init_subclass__`.

ops/model.py
```diff
@@ -1937,8 +1940,22 @@ def from_name(cls, name: str, message: str):
...
+
+        .. code-block:: python
+
+            _StatusClass = typing.TypeVar('_StatusClass', bound='StatusBase')
```
tonyandrewmeyer, line 1951: We have a few other places where we have types referred to in the docs but that aren't in the docs (_ModelBackend, _ModelCache, etc). I think this is like that, in that we don't expect people to be using this so providing the exact type in the docs isn't that important. So I would personally leave this bit out and just have the `_StatusClass` not linked or defined in the docs.

tonyandrewmeyer (approved):

james-garner-canonical:
Closed in favour of #1383

PR #1383, opened 2024-09-20
fix: fix type of StatusBase subclasses by calling StatusBase.register in __init_subclass__

james-garner-canonical:
This is a simple alternative to #1382, following @dimaqq's suggestion to just not use `StatusBase.register` as a decorator. In this PR we instead do so via `__init_subclass__`.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1907,6 +1907,9 @@ class StatusBase:
             raise TypeError('cannot instantiate a base class')
         self.message = message
 
+    def __init_subclass__(cls):
+        StatusBase.register(cls)
+
     def __eq__(self, other: 'StatusBase') -> bool:
         if not isinstance(self, type(other)):
             return False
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer (approved):
I like it - I think there was a clear expectation that every subclass would register so doing it automatically, while a behaviour change, is ok.

I couldn't find any charms with `(StatusBase`, which isn't a perfect search for subclassing, but I think it's a good sign that no-one is doing that.

I like that `register` remains unchanged - we just ignore the returned class. So no need for deprecation warnings etc, and if people are registering different types of class (no idea why) then that would work. And we already allowed registering twice, so anyone doing this:

```python
@Statusbase.register
class OtherStatus(StatusBase):
    name = "other"
```

Is still fine.

james-garner-canonical:
Thanks, @tonyandrewmeyer. I also tried `ag 'class .*\(.*StatusBase'` to see if anyone was subclassing `StatusBase` and didn't find anything.

If we wanted to deprecate register, we could move the functionality to `_register`, call that in `__init_subclass__` and `register`, and add a deprecation warning to `register`. If there's interest in doing that, I'm happy to add it to the PR.

benhoyt (approved):
This approach is nice and simple -- looks good to me!

PR #1384, opened 2024-09-22
refactor: deprecate StatusBase.register decorator

james-garner-canonical:
As `StatusBase.__init_subclass__` is now used to automatically register status classes by name, and the old `StatusBase.register` decorator (1) shouldn't have been public, (2) shouldn't have been used externally anyway, (3) doesn't appear to be used in charms currently, and (4) isn't needed internally, let's deprecate it.

See also discussion on #1380.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1941,27 +1941,8 @@ class StatusBase:
 
     @classmethod
     def register(cls, child: Type['StatusBase']):
-        """Class decorator to register a subclass for lookup using :meth:`from_name`.
-
-        Note: this method is deprecated. Registration is now automatic via __init_subclass.
-        Also, this decorator obscures the class type when used as a decorator.
-
-        Note: this method is intended for internal use only.
-        It is used to make the valid Juju statuses available by name.
-
-        Args:
-            child: A subclass of StatusBase, with a ``name`` attribute of type ``str``.
-
-        Returns:
-            The decorated class, unmodified.
-        """
-        warnings.warn(
-            'StatusBase.register should not be called. It is intended'
-            'for internal use only and is superseded by automatic subclass'
-            'registration via __init_subclass__',
-            DeprecationWarning,
-            stacklevel=2,
-        )
+        """.. deprecated:: 2.17.0 Deprecated - this was for internal use only."""
+        warnings.warn('StatusBase.register is for internal use only', DeprecationWarning)
         cls._register(child)
         return child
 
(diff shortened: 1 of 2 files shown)


ops/model.py
```diff
@@ -1941,14 +1941,38 @@ def from_name(cls, name: str, message: str):
 
     @classmethod
     def register(cls, child: Type['StatusBase']):
-        """Register a Status for the child's name."""
+        """Class decorator to register a subclass for lookup using :meth:`from_name`.
```
benhoyt, line 1944: This all seems a bit heavy, and actually makes it *more* prominent in the docs. Can we shorten this right up, and add a Sphinx deprecation label, like [`Storage.id`](https://github.com/canonical/operator/blob/20c962360d19f0802c98bbf1878631e4f9122bcb/ops/model.py#L2178), so it'll show up nicely [in the docs](https://ops--1384.org.readthedocs.build/en/1384/#ops.Storage.id)?

I would recommend something more like:

```python
@classmethod
def register(cls, child: Type['StatusBase']):
    """.. deprecated:: 2.17.0 Deprecated - this was for internal use only."""
    warnings.warn('StatusBase.register is for internal use only', DeprecationWarning)
    cls._register(child)
    return child
```

benhoyt (approved):
Works for me, thanks!

test/test_model.py
```diff
@@ -830,6 +830,9 @@ class NoNameStatus(ops.StatusBase):  # pyright: ignore[reportUnusedClass]
             class NonStringNameStatus(ops.StatusBase):  # pyright: ignore[reportUnusedClass]
                 name = None  # pyright: ignore[reportAssignmentType]
 
+        with pytest.deprecated_call():
```
tonyandrewmeyer, line 833: It feels like this would more properly be in a new (very small) test of its own, rather than being in the "base status instance" test.

```suggestion
    def test_statusbase_register_is_deprecated(self):
        with pytest.deprecated_call():
```

tonyandrewmeyer (approved):
LGTM! Small test suggestion and a question for Ben.

ops/model.py
```diff
@@ -1941,14 +1941,19 @@ def from_name(cls, name: str, message: str):
 
     @classmethod
     def register(cls, child: Type['StatusBase']):
-        """Register a Status for the child's name."""
+        """.. deprecated:: 2.17.0 Deprecated - this was for internal use only."""
+        warnings.warn('StatusBase.register is for internal use only', DeprecationWarning)
```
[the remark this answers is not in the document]
benhoyt: Ah yeah, that was an oversight, `stacklevel=2` sounds good to me.

PR #1385, opened 2024-09-22
test: properly mock the environment

tonyandrewmeyer:
These two tests change `os.environ` without restoring it afterwards, which means that the changes leak into other tests. Change that by using `monkeypatch.setenv`.


tonyandrewmeyer:
This should fix the issue in #1381.

benhoyt (approved):

james-garner-canonical (approved):

Issue #1356, opened 2024-09-04
Clarity around `add_status`, `collect_app_status` and `collect_unit_status` regarding `ErrorStatus` and `UnknownStatus`


james-garner-canonical:
While improving the type checking and documentation of the status types that can be set by charms (#1354), we noticed an incongruity in how `error` and `unknown` status are treated. While they will both raise errors if used to set the unit or app status (and `ErrorStatus` and `UnknownStatus` document this), the documentation around `add_status` doesn't clearly reflect this.

> ... the `add_status` method accepts both `error` and `unknown`, and the doc lists those. It reads very much like you can add those statuses, but especially for `error` if you include that then it'll definitely fail because it will "win" the evaluation and then can't be set.

_@tonyandrewmeyer in https://github.com/canonical/operator/pull/1354#pullrequestreview-2278753728_

We could just add a bit more documentation here, but it seems peculiar that calling `add_status(ErrorStatus('message')` doesn't raise an immediate error -- instead it's guaranteed to raise an error at status collection time, as the status is internally set to the highest priority status (and `error` is the highest priority). Of course the semantics are different, and perhaps you could use this as a sort of deferring of the exception till after other handlers have had a chance to do something, but it feels off.

It seems a bit more legitimate to call `add_status(UnknownStatus('message'))`, since another event callback could add a higher priority status (`unknown` is the lowest priority), and no error would be raised, but still seems like a strange pattern.

Looking at the 150+ charms, I can find `add_status` being called a reasonable amount with the settable status types (`ActiveStatus`, `BlockedStatus`, `MaintenanceStatus`, `WaitingStatus`), but never with `ErrorStatus` or `UnknownStatus`.

Perhaps this is a good opportunity to make it an error to call `add_status` with `ErrorStatus` or `UnknownStatus`?

Having said that, the inability to set `ErrorStatus` makes me wonder what the current best practice is for a charm to give up and require manual intervention, and if documentation could be improved there?

tonyandrewmeyer:
> Perhaps this is a good opportunity to make it an error to call `add_status` with `ErrorStatus` or `UnknownStatus`?

I think so. My guess is that this was simply missed when collect-status was added, so it simply included all statuses and how they would be ranked, without considering that it's not actually possible to set `error` or `unknown`.

> Having said that, the inability to set ErrorStatus makes me wonder what the current best practice is for a charm to give up and require manual intervention, and if documentation could be improved there?

[...]

PR #1386, opened 2024-09-23
feat: make it an error to call CollectStatusEvent.add_status with error or unknown

james-garner-canonical:
It seems like it should immediately be an error to add an `ErrorStatus` or `UnknownStatus` with `add_status`, since these can never be set as the status, and will cause an error at the end of the event if they are the highest priority status added (always the case with `ErrorStatus` as it has the highest priority, while with `UnknownStatus` only if it is the only status type added, as it has the lowest priority). See also issue #1356.

This draft PR makes this relatively simple runtime change, and then investigates the changes needed to support this via type checking.

There ends up being more type checking code than runtime code, with two new public type aliases (`SettableStatus` and `SettableStatusName` -- better names might be better?).

The biggest offender is `from_name`, which previously took a `str` argument and (implicitly) returned an `UnknownStatus` or another `StatusBase`, but which we now need to know whether the status returned will be valid for use with `add_status` or not. This PR uses `@overload` to accomplish this, but possibly there's a better way (a `TypedDict` for `StatusBase._statuses` perhaps?).

One concern I have is if people are using loosely typed strings to call `from_name`, or loosely typed collections of statuses when calling `add_status`, they'll now have a lot of type checking errors.

One possible solution would be to just make the runtime changes and leave the typing alone for now.


tonyandrewmeyer (commented):
I think the overloads are fine - yes, there are 18 lines, but it's readable and clear what it's doing.

> One concern I have is if people are using loosely typed strings to call from_name, or loosely typed collections of statuses when calling add_status, they'll now have a lot of type checking errors.

It would be good to check for this (e.g. `super-tox -e` with the various names charms use for static checking). I'm generally ok with going to the stricter types both for the ability to have more static checking and for any autocomplete improvements it brings (probably not much here, since these objects are practically identical), but I don't think we want to annoy charmers if there isn't a clear win.

ops/charm.py
```diff
@@ -1105,12 +1105,10 @@ class CollectStatusEvent(LifecycleEvent):
 
     The order of priorities is as follows, from highest to lowest:
 
-    * error
     * blocked
     * maintenance
     * waiting
     * active
-    * unknown
```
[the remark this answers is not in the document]
benhoyt: Yeah, I think removing these from the list is good, but agree with Tony it's probably worth adding a sentence like that.

benhoyt (commented):
Sorry to be a wet blanket, but I'm not a fan of this. I think it's a lot of implementation noise for the sake of something that's likely not an issue in practice. I'd much prefer we keep the types simple and do the runtime checks.

However, I suggest we do that a different way. For one thing, this only covers `CollectStatusEvent.add_status`, not `app.status = ...` or `unit.status = ...`, and we'd want the same check for those.

If we put the check in `status_set`, we'd mirror the check that Juju's `status-set` hook tool already does. In fact, for the `_TestingModelBackend` used by harness, we already do this:

```python
    def status_set(self, status: '_StatusName', message: str = '', *, is_app: bool = False):
        if status in [model.ErrorStatus.name, model.UnknownStatus.name]:
            raise model.ModelError(
                f'ERROR invalid status "{status}", expected one of'
                ' [maintenance blocked waiting active]'
            )
```

(It would seem better to do "not in allowed" rather than "in disallowed", but that's a minor detail.)

In other words, as long as someone has a unit test with the code path in question, `Harness` will already raise an error and flag this.

It looks like Scenario doesn't do this check (@tonyandrewmeyer can correct me if I'm wrong), but that could easily be added, and I'd be in support of that.

Overall it's a trade-off: we want reasonable types, but they come at a cost, and I think we need to be careful here.

james-garner-canonical:
> I think it's a lot of implementation noise for the sake of something that's likely not an issue in practice

Yeah I think it's a lot of typing ceremony for not much value added. I've taken out the type checking changes in favour of just runtime checks.

`status_set` is actually already checked at runtime, but I think it's still worth checking in `add_status` to make it easier to get to the the offending line of code from the exception, since an error on the automatic `status_set` call after the `collect-status` event won't point directly to the bad `add_status` call.

tonyandrewmeyer:
> It looks like Scenario doesn't do this check (@tonyandrewmeyer can correct me if I'm wrong), but that could easily be added, and I'd be in support of that.

It doesn't, and I'd be happy have that added too (it would be in mocking.py's `set_status` I imagine). Feel free to open a PR!

benhoyt:
> status_set is actually already checked at runtime, but I think it's still worth checking in add_status to make it easier to get to the the offending line of code from the exception, since an error on the automatic status_set call after the collect-status event won't point directly to the bad add_status call.

Ah, that's a goo point -- thanks.

benhoyt (approved):
Thanks, I'm happy with this much more limited change.

benhoyt:
I've also opened an [issue in ops-scenario](https://github.com/canonical/ops-scenario/issues/202) to make its `status_set` do the same thing.

james-garner-canonical:
Are we happy to merge the changes in as is? The current changes (1 approval) are an update to the docstring for `CollectStatusEvent`, and making it an immediate `InvalidStatusError` to call `add_status` with `ErrorStatus` or `UnknownStatus`, rather than waiting for an `InvalidStatusError` to be raised when the status is set at the end of the `collect-status` event (after the charm's observer for `CollectStatusEvent` has returned).

ops/charm.py
```diff
@@ -1105,12 +1105,13 @@ class CollectStatusEvent(LifecycleEvent):
...
+
+    It is an error to call :meth:`add_status` with an instance of
+    :class:`ErrorStatus` or :class:`UnknownStatus`.
```
tonyandrewmeyer, line 1114: I would use the `Raises:` style to document this (which will also mean we can get rid of that line with all the ignores, I think, as a side-effect bonus).

Something like:

```
Raises:
    InvalidStatusError: if :class:`ErrorStatus` or :class:`UnknownStatus` are added
```

tonyandrewmeyer (approved):
Looks good to me, thanks!

One small doc suggestion.

ops/charm.py
james-garner-canonical: I'm drawing a blank, sorry, which is the line with the ignores?

tonyandrewmeyer: It's [right above](https://github.com/canonical/operator/blob/98a4e4f44d37d3361a253bc1eff49eee2ca6d140/ops/charm.py#L1145) where you would put the `Raises:` lines.

james-garner-canonical: I wonder if the `raises` would be best in the `add_status` docstring, but then that isn't so good for addressing the "why aren't all six statuses listed in the priority order" question ...
image
vs
image

benhoyt: I think the "raises" belongs on the `add_status` docstring, as that's what will actually raise. I think it's fine without those in the bulleted list -- in fact, perhaps clearer.

tonyandrewmeyer: Oh, good point that these are two different docstrings, I somehow missed that.

Should `Raises: ...` should go into `add_status`? We try to be complete in documenting the exceptions that can be raised, but only for those that we expect the charm may want to catch (typically: something unexpected has gone wrong). We don't document the ones like `TypeError` that are really a bug in the charm. I think this does fall into that category (unless the charm was getting the status from somewhere external but that seems very unlikely and should be handed in the charm code).

So I take back my "do this with Raises:" suggestion, sorry, and I would stick with what you originally had.

james-garner-canonical: Funny timing, I just pushed the change removing the sentences from the class documentation and adding `Raises:` to `add_status`. I'll just revert that to what I had before and merge.

I had a look at removing those `noqa` directives, but it seems like `ruff` really doesn't like `return` in a code sample in a docstring.

benhoyt (approved):
Fine with me, thanks!

PR #1387, opened 2024-09-24
ci: run the smoke tests on a schedule

### LINKED ISSUE #1322: Run 'tox -e smoke' in CI
We currently don't run the smoke tests (which at the moment is just packing and deploying on supported bases) in CI, which means that these tests don't get exercised regularly.

We should add a workflow that runs `tox -e smoke` - note that this requires a real Juju, so the environment will need to install Juju, charmcraft, and a local cloud of some sort (e.g. microk8s/k8s or LXD), and bootstrap a controller.

We need to figure out:
* Do we want to just pick a Juju and charmcraft version and use that, or do we want to put those in a matrix (perhaps Juju 2.9, 3.5 (until 3.6 is released), and 4.0, with the latest charmcraft 2.x and 3.x)?
* What should trigger this? A schedule (perhaps just before we expect to release, like the charm dep checks), or only on a merge to main, or on each PR? The tests will be slow, but wouldn't have to be required.

tonyandrewmeyer:
We have some basic smoke tests, which are currently the only tests we regularly run against an actual Juju model. The tests deploy a basic charm and verify that it goes to `active` status (which happens in an `install` handler). However, these are currently only run if someone manually runs `tox -e smoke`. This PR adds running the tests on a schedule (currently monthly, in the days before we will likely do a release).

This tests:
 * Juju 3.5 (latest stable 3.x - I suggest we *change* this to 3.6 when that's released, rather than *adding* 3.6). The smoke tests do not currently work with Juju 2.9 (I'm not sure if it's worth adding that at this point), but if that changes, it should only require a matrix change in the workload. The smoke tests use python-libjuju, which doesn't yet support Juju 4.0, but when that changes, this should also only require a matrix change.
 * Charmcraft 2.x and 3.x.
 * LXD and MicroK8s.

The smoke test is changed to use a plain `subprocess.run` to pack the charm, rather than use pytest-operator for this. The `pytest-operator` approach doesn't work out-of-the-box in the workflow, and it seems cleaner to just do the pack ourselves.

There's an [example run on my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/11006199263/job/30560942709).

This could be optimised in the future to store the packed charms and share them across Juju versions and cloud types, but it doesn't seem worth doing that for now.

Fixes #1322

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -1,0 +1,36 @@
+name: ops Smoke Tests
+
+on:
+  workflow_dispatch:
+    schedule:
+      - cron:  '0 7 25 * *'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        juju-version: ['2.9', '3', '4.0']
+        charmcraft-version: ['2.x', '3.x']
+
+    steps:
+      - name: Set up Juju
+        run: sudo snap install juju --channel=${{ matrix.juju-version }}
+
+      - name: Set up LXD
+        uses: canonical/setup-lxd@ce8decb3609c0a03c5abd5034d02a6c145e2076f
+        with:
+          channel: 5.0/stable
+
+      - name: Bootstrap Juju controller
+        run: juju bootstrap localhost
+
+      - name: Install charmcraft
+        run: sudo snap install charmcraft --channel=${{ matrix.charmcraft-version }}
+
+      - name: Install tox
+        run: pip install tox~=4.2
+
+      - name: Run smoke tests
+        run: tox -e smoke
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -12,7 +12,8 @@ jobs:
     strategy:
       matrix:
         # pylibjuju does not currently support Juju 4.x
-        juju-version: ['2.9', '3.5']
+        # The smoke tests do not yet work on Juju 2.9.
+        juju-version: ['3.5']
         charmcraft-version: ['2.x', '3.x']
         cloud: ['lxd', 'microk8s']
 
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -29,6 +29,12 @@ jobs:
       - name: Install charmcraft
         run: sudo snap install charmcraft --channel=${{ matrix.charmcraft-version }} --classic
 
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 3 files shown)


dimaqq (approved):

tox.ini
```diff
@@ -147,6 +148,8 @@ commands =
     python -m build --sdist --outdir={toxinidir}/test/charms/test_smoke/
     # Inject the tarball into the smoke test charm's requirements.
     bash -c 'echo "./$(ls -1 ./test/charms/test_smoke/ | grep tar.gz)" > ./test/charms/test_smoke/requirements.txt'
+    # If a specific Juju version is set, then make sure we are using that version of pylibjuju.
+    bash -c 'if [ -n "$JUJU_VERSION" ]; then pip install "juju ~= $JUJU_VERSION"; fi'
```
benhoyt, line 152: Clever!

.github/workflows/smoke.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Set up Microk8s
+        if: matrix.cloud == 'microk8s'
+        uses: balchua/microk8s-actions@v0.3.2
+        with:
+          channel: '1.26-strict/stable'
+          devMode: 'true'
+          addons: '["dns", "hostpath-storage"]'
+
+      - name: Set up Juju (classic)
+        if: matrix.juju-version == '2.9'
```
benhoyt, line 40: Shall we just drop the "if 2.9" stuff here, seeing we're not running these on 2.9? Or do you think we should add that in the near future?

benhoyt (approved):
Nice! Thanks a lot for this. Just one optional nit comment.

.github/workflows/smoke.yaml
tonyandrewmeyer: Yeah, good question. I started out planning to do 2.9, 3.[latest stable], and 4.0, thinking that would be a good range of coverage. 2.9 because it's still in "security fix support" and seems to have non-trivial use still, 3.[latest stable] because it seems unlikely that we'd break deploying only for one 3.x, and 4.0 because it's good to be proactive. I then ran into a bunch of issues and ended up dropping down to just 3.x.

`tox -e smoke` passes for me (except for trying to deploy on 24.04) locally, so I think it shouldn't be too much work to get it working in CI as well. I'd like to use a quite small amount of time to try to get that going and if not dropping it would make sense yes. I could cut it from this PR and add it back later if that was preferred.

benhoyt: Sounds good. I'm happy either way -- no specific urgency on this.

PR #1388, opened 2024-09-24
ci: spell-check the code as part of linting

tonyandrewmeyer:
Add a `codespell` run as part of the `tox -e lint` environment, to pick up any spelling errors that might be missed in review. It's also added to the pre-commit checks, for anyone using those.

The PR also makes 4 corrections, which codespell found. Two are debatable (whether or not to hyphenate) but two are definite errors.

This adds a new (lint-time) dependency of codespell - it's also used in the default charmcraft profile, I've used it before, and it has a [good Snyk score](https://snyk.io/advisor/python/codespell) so it seems ok to me. I have lightly skimmed the code, but not read it in depth.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.5.0
     hooks:
       - id: check-added-large-files
       - id: check-ast
@@ -24,3 +24,10 @@ repos:
         args: [ --preview ]
       - id: ruff-format
         args: [ --preview ]
+  # Spellcheck the code.
+  - repo: https://github.com/codespell-project/codespell
+    rev: v2.2.4
+    hooks:
+    - id: codespell
+      additional_dependencies:
+        - tomli
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -6,7 +6,7 @@ In the interest of fostering an open and welcoming environment, we as
 contributors and maintainers pledge to making participation in our project and
 our community a harassment-free experience for everyone, regardless of age, body
 size, disability, ethnicity, sex characteristics, gender identity and expression,
-level of experience, education, socio-economic status, nationality, personal
+level of experience, education, socioeconomic status, nationality, personal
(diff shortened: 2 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -24,3 +24,10 @@ repos:
         args: [ --preview ]
       - id: ruff-format
         args: [ --preview ]
+  # Spellcheck the code.
+  - repo: https://github.com/codespell-project/codespell
```
benhoyt, line 28: Is this, or can we set this, to British English (Canonical style)? At a cursory look "socio-economic" is more common in British English, so that made me wonder.

benhoyt (approved):
Looks reasonable to me. Out of interest, what does this check? Just code comments and strings?

tonyandrewmeyer:
> Out of interest, what does this check? Just code comments and strings?

I think it's doing a regular expression match for `[\w\-'’]+` by default (you can customise this). The doc says:

>> Fix common misspellings in text files. It's designed primarily for checking misspelled words in source code (backslash escapes are skipped), but it can be used with other files as well. It does not check for word membership in a complete dictionary, but instead looks for a set of common misspellings. Therefore it should catch errors like "adn", but it will not catch "adnasdfasdf". This also means it shouldn't generate false-positives when you use a niche term it doesn't know about.

I think this 'look for misspellings' is how it manages to avoid having to look specifically for words in comments, strings, and so on. If I add a line `adn = 42` then it'll flag that as a potential misspelling. There are ways to disable the check for specific lines, but if we start having to do that all over the place then I'd rather get rid of it or use a different tool. I think that it found 2 (to 4) legitimate mistakes and had no false positives is a positive sign, though.

.pre-commit-config.yaml
tonyandrewmeyer: My understanding is that it will currently accept both en-UK and en-US spellings, so it fails to help with that. There is an option to make en-UK spellings an error with the en-US the suggestion, but there doesn't seem to be the reverse option, which is what we would want.

There's a lot of customisation possible in terms of which dictionaries are used, so maybe it's possible to eliminate en-US? I'm not totally sure, and the docs don't seem to say anything about it.

dimaqq (approved):

Issue #1287, opened 2024-07-18
Simplify Pebble enum|str types to just enum


benhoyt:
Per our discussion at Charm Tech daily sync today, we'd like to simplify the various pebble enum types from `Union[EnumType | str]` to just `EnumType`, for example, `ServiceInfo.current` now has type `Union[ServiceStatus, str]`, which can be awkward for us (and for charms) to deal with.

The original reason I made them `|str` was so that if a charm is deployed to a newer version of Juju, which has a newer version of Pebble that add a new value to the enum, code won't fail when the enum is being created. To create the value we have code like this:

```python
        try:
            current = ServiceStatus(d['current'])
        except ValueError:
            current = d['current']
```

But that means code that uses these fields have to have awkward types and sometimes have to do awkward things like `use(info.current) if isinstance(info.current, str) else use(info.current.value)`.

So we'd like to simplify the type of these to just the enum type, for example just `ServiceStatus` in this example.

To avoid the charm code raising an exception, we want to add a new value `UNKNOWN = 'unknown'` to all the enum types that are used in this way. If we got an invalid or unknown string value from Pebble, we'd set it to that instead, and give a warning. The code to create the value from Pebble JSON would become something like:

```python
        try:
            current = ServiceStatus(d['current'])
        except ValueError:
            warnings.warn(f'Unknown ServiceStatus value {d["current"]!r}')
            current = ServiceStatus.UNKNOWN
```

(Note that `FileType` already has an `UNKNOWN` value, which is handy.)

benhoyt:
@james-garner-canonical Assigning this to you to see if you can squeeze it in before The Hague.

james-garner-canonical:
Due to backwards compatibility, we can't make this planned change (see discussion on PR #1400).

PR #1391, opened 2024-09-25
feat: remove enum unions from use in ServiceInfo

james-garner-canonical:
Working towards issue #1287, this draft PR removes the use of `str` and `enum` member unions connected to the `ops.pebble.ServiceInfo` class.

In this PR I try to remove all such unions to simplify types. However, I wonder if it would be better to continue to allow the use of both `str` and `enum` members as arguments in some cases (constructors) where removing the union might result in a breaking api change.

---

Previously a `str` argument that an `ops` `enum` doesn't know about would result in the use of the `str`, while a known argument would use the corresponding `enum` member. Instead, we now use an `UNKNOWN` enum member in this case.

This is a behaviour change, but shouldn't be a breaking change, because the only time you should be using such values would be if you were developing for a more recent Pebble version than `ops.pebble` knows about. Since `ops.pebble` is presumable up to date currently, this will only be the case on older `ops` versions, which this PR does not apply to. However, it will be worth mentioning this behaviour change in the appropriate release notes.

An example of the core behaviour change, in the `ServiceInfo.from_dict` method:
```diff
 try:
     startup = ServiceStartup(d['startup'])
 except ValueError:
-    startup = d['startup']
+    warnings.warn(f'Unknown ServiceStartup value {d["startup"]!r}')
+    startup = ServiceStartup.UNKNOWN
...
 return cls(
     name=d['name'],
     startup=startup,
     current=current,
 )
```

---

This leads us to the `__init__` method, where we also have unions to remove. I want to highlight this change because it's potentially breaking.

```diff
     def __init__(
         self,
         name: str,
-        startup: Union[ServiceStartup, str],
-        current: Union[ServiceStatus, str],
+        startup: ServiceStartup,
+        current: ServiceStatus,
     ):
         self.name = name
         self.startup = startup
         self.current = current
```

This is a breaking change, as the `pebble` module is public (explicitly exported in `ops.__init__.__all__`), and `ServiceInfo` is a public member of that module. Previously a user could expect to be able to call `ServiceInfo("foo", startup="enabled", current="active")` if they wanted to. However:

1. It never worked correctly with string arguments. Note how `startup` and `current` are assigned in `__init__` without any runtime validation. This leads to unexpected behaviour here:
```python
  def is_running(self) -> bool:
      """Return True if this service is running (in the active state)."""
      return self.current == ServiceStatus.ACTIVE
```
This may have been written with the assumption that `ServiceStatus.ACTIVE == 'active'`, the way it would with a normal class (is this how enums in other languages most commonly work?). If that were the case, then you'd expect self.current being a union of `str` and `enum` members to just work. Instead, `ServiceStatus.ACTIVE` is an object of `type` `<enum 'ServiceStatus'>` and `ServiceStatus.ACTIVE.value == "hello"`, so this will return `False` if `self.current == 'active'`.

2. Constructing a `ServiceInfo` from arbitrary arguments isn't really the intended use case, as you want information about actually running services. You're more likely to call `ops.pebble.get_services`, or a bit lower level, call `ops.pebble.ServiceStatus.from_dict` with a dict representation of the services.

This leads me to think this breaking change may be ok. Currently the change will only result in a type checking error, which if ignored will lead to the same (broken) runtime behaviour as previously.

We could avoid the breaking change and handle things better at runtime (fixing the issue with `is_running`) by continuing to accept unions for the `__init__` args and turning them into `enum` members before assigning them. The `from_dict` logic could be moved to `__init__` to handle this.

---

The final instance of `str` and `enum` member unions related to this class is `_ServiceInfoDict`. Removing this union leads to another potentially breaking change.

`_ServiceInfoDict` is part of the signature of the public `ServiceInfo.from_dict` method, and also used to annotate a value returned from an external Pebble call, which could never actually contain Python `enum` members. Without introducing more type variables, the correct change when removing these unions should therefore be:
```diff
 _ServiceInfoDict = TypedDict(
     '_ServiceInfoDict',
-    {'startup': Union['ServiceStartup', str], 'current': Union['ServiceStatus', str], 'name': str},
+    {'startup': str, 'current': str, 'name': str},
 )
```
But again this is technically a breaking change, as previously a user could expect to validly call:
```python
ServiceInfo.from_dict({"name": "foo", "startup": ServiceStartup.ENABLED, "current": ServiceStatus.ACTIVE})
```
This still works at runtime, because `ServiceStartup(ServiceStartup.ENABLED)` and `ServiceStartup('enabled')` both produce `ServiceStartup.ENABLED`, but potentially introduces a new type checking error for users. Again though, using a hand crafted dictionary rather than one that ultimately comes from Pebble itself is less likely here (what about unit tests?).

We could avoid a breaking change here by continuing to accept the original `_ServiceInfoDict` type in `from_dict`.

---

Do we want to remove all the uses of `str` and `enum` member unions as a goal in itself to simplify typing? If so, this may lead to some breaking changes.

Alternatively, if the goal is just to switch to using an `UNKNOWN` enum member instead of passing along unknown string values, then I think we'll need to keep `str` and`enum` member unions in some places.

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -197,7 +197,7 @@ _AuthDict = TypedDict(
 
 _ServiceInfoDict = TypedDict(
     '_ServiceInfoDict',
-    {'startup': Union['ServiceStartup', str], 'current': Union['ServiceStatus', str], 'name': str},
+    {'startup': str, 'current': str, 'name': str},
 )
 
 # Callback types for _MultiParser header and body handlers
@@ -350,6 +350,18 @@ class _NotProvidedFlag:
 
 
 _not_provided = _NotProvidedFlag()
+
+
+class _Unknown:
+    """Get a value that will never be in Pebble's json.
+
+    Used as the value for UNKNOWN in enums, to avoid silent success
+    on collision with a string returned from Pebble, e.g. if
+    'unknown' value was added on Pebble's end as a valid value.
+    """
+
+    def __repr__(self):
+        return f'{self.__module__}.{self.__class__.__name__}()'
 
 
 class _UnixSocketConnection(http.client.HTTPConnection):
@@ -1038,6 +1050,7 @@ class ServiceStartup(enum.Enum):
 
     ENABLED = 'enabled'
     DISABLED = 'disabled'
+    UNKNOWN = _Unknown()
 
 
 class ServiceStatus(enum.Enum):
@@ -1046,6 +1059,7 @@ class ServiceStatus(enum.Enum):
     ACTIVE = 'active'
     INACTIVE = 'inactive'
     ERROR = 'error'
+    UNKNOWN = _Unknown()
 
 
 class ServiceInfo:
@@ -1054,8 +1068,8 @@ class ServiceInfo:
     def __init__(
         self,
         name: str,
-        startup: Union[ServiceStartup, str],
-        current: Union[ServiceStatus, str],
+        startup: ServiceStartup,
+        current: ServiceStatus,
     ):
         self.name = name
         self.startup = startup
@@ -1071,11 +1085,13 @@ class ServiceInfo:
         try:
             startup = ServiceStartup(d['startup'])
         except ValueError:
-            startup = d['startup']
+            warnings.warn(f'Unknown ServiceStartup value {d["startup"]!r}')
+            startup = ServiceStartup.UNKNOWN
         try:
             current = ServiceStatus(d['current'])
         except ValueError:
-            current = d['current']
+            warnings.warn(f'Unknown ServiceStatus value {d["current"]!r}')
+            current = ServiceStatus.UNKNOWN
         return cls(
             name=d['name'],
             startup=startup,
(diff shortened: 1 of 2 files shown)


benhoyt:
I lean towards simplifying them all -- changing them all to just the enum (or just the str in the case of the dict). People shouldn't be doing this, and as you show, in the cases where we do allow a union things like `is_running` are half broken anyway. I suspect the `_ServiceInfoDict` allowing enum types at all was a copy-n-paste issue; it seems like a bug.

Let's at least try that and see what it looks like, and run super-tox over the diff to see if there are any failures.

ops/pebble.py
```diff
@@ -352,6 +352,18 @@ class _NotProvidedFlag:
 _not_provided = _NotProvidedFlag()
 
 
+class _Unknown:
```
benhoyt, line 355: I think we should avoid this and just use `'unknown'` in the enum definition. If Pebble does have "unknown" (`FileType` already has an "unknown" value), nothing bad happens -- it's just that the caller can't distinguish between the two, which doesn't seem like it would be an issue in practice. There's also the warning log for debugging.

james-garner-canonical: Note that the warning log won't be emitted if Pebble has "unknown" and "unknown" is in the `enum`. If that's ok, then I'll remove the sentinel class. I hadn't really thought about what I'd do in the case of `FileType.UNKNOWN`, so having it always be the string "unknown" makes that easier.

PR #1320, opened 2024-08-16
docs: add top-level intro and module-level intros

dimaqq:
Docs: Ops: Add top-level intro, clean up ops.module intro in light of that, flesh out ops.pebble and ops.testing intros a bit more

https://warthogs.atlassian.net/browse/CHARMTECH-183

- [x] "home page" library reference
- [x] ops module
- [x] ops.main entry point
- [x] ops.pebble module
- [x] ops.testing module

PR #1392, opened 2024-09-26
docs: move Pebble to a separate page

tonyandrewmeyer:
Pros:
 * Significantly reduces the length of the right-hand sidebar for the main API reference
 * Clearly delineates ops.pebble, which could legitimately used independently of ops as a Python Pebble API
 * Separates duplication between ops.Container and ops.pebble into separate pages

Cons:
* Browser search-on-page is less useful (although we already started this process by splitting ops.testing out)
* Existing deep links into the docs will break (we should be able to find and fix the ones on juju.is/docs at least)

### DIFF
--- a/docs/pebble.rst
+++ b/docs/pebble.rst
@@ -1,6 +1,6 @@
 .. _ops_pebble_module:
 
-ops.pebble
-==========
+Pebble client
+=============
 
 .. automodule:: ops.pebble
--- a/docs/pebble.rst
+++ b/docs/pebble.rst
@@ -1,0 +1,6 @@
+.. _ops_pebble_module:
+
+ops.pebble
+==========
+
+.. automodule:: ops.pebble
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,7 +9,7 @@ The library (`available on PyPI`_) provides:
 
 - :ref:`ops_module`, the API to respond to Juju events and manage the application;
 - :ref:`ops_main_entry_point`, used to initialise and run your charm;
-- :ref:`ops_pebble_module`, the Pebble client, a low-level API for Kubernetes containers;
+- :doc:`ops.pebble </pebble>`, the Pebble client, a low-level API for Kubernetes containers;
 - the APIs for unit testing charms in a simulated environment:
 
   - :doc:`State-transition testing </state-transition-testing>`. This is the
@@ -30,6 +30,7 @@ integrations with other services, and making the charm easier to test.
    :maxdepth: 2
 
    self
+   pebble
    state-transition-testing
    harness
 
@@ -55,15 +56,6 @@ legacy main module
 
 .. automodule:: ops.main
    :noindex:
-
-.. _ops_pebble_module:
-
-ops.pebble
-----------
-
-.. automodule:: ops.pebble
-
-.. _ops_testing_module:
 
 
 Indices
--- a/ops/__init__.py
+++ b/ops/__init__.py
@@ -14,7 +14,7 @@
 
 """The API to respond to Juju events and manage the application.
 
-This module provides core features to your charm, including:
+This API provides core features to your charm, including:
 
 - :class:`~ops.CharmBase`, the base class for charms and :class:`~ops.Object`,
   the base class for charm libraries.


tonyandrewmeyer:
@tmihoc feel free to provide an opinion here too, if you have time & want to :)

benhoyt:
@tonyandrewmeyer RTD docs are failing to build. I'll wait to review till you've fixed that so I can compare with https://ops.readthedocs.io/en/latest/

tonyandrewmeyer:
> @tonyandrewmeyer RTD docs are failing to build. I'll wait to review till you've fixed that so I can compare with https://ops.readthedocs.io/en/latest/

Huh, they built without error locally, and I didn't get an email about the failure; weird. Thanks for letting me know - I'll figure out what's going on.

tonyandrewmeyer:
@benhoyt should be ready now, sorry.

docs/pebble.rst
```diff
@@ -0,0 +1,6 @@
+.. _ops_pebble_module:
+
+ops.pebble
```
benhoyt, line 3: This shows up as "ops.pebble" in the sidebar, which doesn't fit the others. Can we make it "Pebble client"?

benhoyt (changes requested):
One change requested. Otherwise, I'm happy with this. Let's just follow up with fixing the links in the Juju.is docs.

benhoyt (approved):

james-garner-canonical (approved):
This change makes sense to me, and looks good when built.

dimaqq (commented):
👍🏻

tmihoc:
> @tmihoc feel free to provide an opinion here too, if you have time & want to :)

Giving that the Pebble module is indeed quite independent, I think the move makes sense. 

Unrelatedly: Looking at https://ops.readthedocs.io/en/latest/index.html, the Navigation, there are two things that I'd do differently:

(1) [API reference](https://ops.readthedocs.io/en/latest/index.html): API for what? What's the connection between this module and the Pebble module? (In short, I wish we had a better title, though not sure what.)
(2) [Unit testing (was: Scenario)](https://ops.readthedocs.io/en/latest/state-transition-testing.html): The "was" there is not great. I'd rephrase as "Unit testing (previously, "Scenario"). There too, I'm wondering: How does this section/title fit with the logic for the other sections/titles? We should try to have a coherent strategy that correctly represents the pieces and the relationships between them.

tonyandrewmeyer:
> (1) nitpick: [Unit testing (was: Scenario)](https://ops.readthedocs.io/en/latest/state-transition-testing.html): The "was" there is not great. I'd rephrase as "Unit testing (previously, "Scenario").

This is intended to only be there for a few months, for what it's worth, to help out all the people that know of Scenario understand that it's the same thing. It could be "previously", although it does then wrap to two lines, which is not ideal.

(2) a bigger concern: [API reference](https://ops.readthedocs.io/en/latest/index.html), [Pebble client](https://ops.readthedocs.io/en/latest/pebble.html#), [Unit testing (was: Scenario)](https://ops.readthedocs.io/en/latest/state-transition-testing.html), [Harness (legacy unit testing)](https://ops.readthedocs.io/en/latest/harness.html): The titles don't help me see the pieces and the relationships between them clearly. API reference = for what?

It could be "Ops API reference" I guess? If you look down from the top of the page, it says "The Ops Library", then "The ops library documentation" and then "API reference". 

> How does it connect to the Pebble client or the Unit testing or the Harness bits?

I think Pebble is tricky because ops wraps the Pebble client in the `Container` object, so for the most part charmers should not need to worry about what's in "Pebble client" because they should be using what's in the "Container" section of "API reference". However, `Container` leaks a bunch of ops.pebble stuff (error classes, layers, etc) so it's not entirely wrapped, and charmers do need to go to "Pebble client" for those (and "client" in that case is somewhat wrong).

We could introduce a layer, like:

* Development
 * API reference
 * Pebble
* Testing
 * Unit testing (previously: Scenario)
 * Harness (legacy unit testing)

Would that help?

> I don't have a clear suggestion but I think we should take a moment to think about this.

Sure :)

tmihoc:
> It could be by namespace, but we decided to merge the two APIs in ops.testing, so that would be:
> 
> * ops
> * ops.pebble
> * ops.testing
> * ops.testing (legacy API)

This makes the most sense to me and I just typed something like this too on your other PR: https://github.com/canonical/operator/pull/1320#issuecomment-2376265039 

With a small difference:

- ops (natural order:)
  - ops.main entrypoint
  - ops.main (alphabetical order:)
    - ActionEvent
    - ...
  - ops.pebble
  - ops.testing
  - ops.testing.harness (legacy)

PR #1349, opened 2024-08-29
docs: typo on CharmBase inheritance example

theofpa:
typo: the CharmBase inheritance example should be an object, not a function

PR #1394, opened 2024-09-26
chore: update changelog and bump version number for 2.17.0 release

james-garner-canonical:
Also add 2.16.1 to CHANGES.md and fix links in 2.16.0 changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,7 +42,7 @@
 
 ## Chores
 
-* note juju version on legacy workaround in (#1355)
+* Note Juju version on legacy workaround (#1355)
 * re-enable test now that Pebble directory permissions are fixed in (#1363)
 * generate warnings for events that will be removed in Juju 4.0 in (#1374)
 * update charm pins in (#1390)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,38 +1,97 @@
+# 2.17.0 - 26 Sep 2024
+
+## Features
+
+* optionally install Scenario with `ops[testing]` and expose the names in ops.testing in (#1381)
+* add a JujuVersion property for Pebble log forwarding to Loki in (#1370)
+* expand the secret ID out to the full URI when only given the ID in (#1358)
+* pre-emptively raise `InvalidStatusError` instead of waiting for Juju:
+    * make it an error to call `CollectStatusEvent.add_status` with error or unknown in (#1386)
+    * document and validate settable status values in `_ModelBackend.set_status` in (#1354)
+
+## Fixes
+
+* rework `ops.main` type hints to allow different flavours (callable class) in (#1345)
+* fix type of `StatusBase` subclasses by calling `StatusBase.register` in `__init_subclass__` in (#1383)
+* `Secret.set_info` and `Secret.set_content` can be called in the same hook in (#1373)
+
+## Documentation
+
+* add top-level intro and module-level intros in (#1320)
+* update the links to the Pebble docs in (#1362)
+* note about repeatedly setting secret value in Juju 3.6 in (#1366)
+* `config-changed` is triggered by Juju trust in (#1357)
+* typo on `CharmBase` inheritance example by @theofpa in (#1349)
+
+## Continuous Integration
+
+* periodically run the unit tests of all GitHub-hosted published charms in (#1365)
+* update the TIOBE reporting for the changes in coverage calculation in (#1367)
+* spell-check the code as part of linting in (#1388)
+* run the smoke tests on a schedule in (#1387)
+
+## Testing
+
+* properly mock the environment in (#1385)
+
+## Refactoring
+
+* move the content of `ops.testing` to `ops._private.harness` in (#1369)
+* keep the `unittest.mock` names in the 'mock' namespace in (#1379)
+* deprecate `StatusBase.register` decorator in (#1384)
+
+## Chores
+
+* note juju version on legacy workaround in (#1355)
+* re-enable test now that Pebble directory permissions are fixed in (#1363)
+* generate warnings for events that will be removed in Juju 4.0 in (#1374)
+* update charm pins in (#1390)
+* update charm pins in (#1375)
+* bump version in (#1348)
+* bump canonical/setup-lxd version in (#1350)
+* bump actions/attest-build-provenance from 1.4.0 to 1.4.2 in (#1351)
+
+# 2.16.1 - 5 Sep 2024
+
+## Fix
+
+* avoid changing `os.environ` in Harness in (#1359)
+
 # 2.16.0 - 29 Aug 2024
 
 ## Features
 
-* Add the description field to SecretInfo in https://github.com/canonical/operator/pull/1338
+* Add the description field to SecretInfo in (#1338)
 
 ## Refactor
 
-* Parse JUJU_* environment variables in one place in https://github.com/canonical/operator/pull/1313
+* Parse JUJU_* environment variables in one place in (#1313)
 
 ## Fixes
 
-* Fix reading Juju secret expiry dates in https://github.com/canonical/operator/pull/1317
-* Correct the signature of .events() in https://github.com/canonical/operator/pull/1342
+* Fix reading Juju secret expiry dates in (#1317)
+* Correct the signature of .events() in (#1342)
 
 ## Documentation
 
-* Security policy change to only support each active major release in https://github.com/canonical/operator/pull/1297
-* Add Juju version markers in https://github.com/canonical/operator/pull/1311
-* Use Sphinx 8 in https://github.com/canonical/operator/pull/1303
-* Live reload documentation with sphinx-autobuild in https://github.com/canonical/operator/pull/1323
+* Security policy change to only support each active major release in (#1297)
+* Add Juju version markers in (#1311)
+* Use Sphinx 8 in (#1303)
+* Live reload documentation with sphinx-autobuild in (#1323)
 
 ## Tests
 
-* Update the smoke test series/bases in https://github.com/canonical/operator/pull/1318
-* Run pytest in parallel with pytest xdist in https://github.com/canonical/operator/pull/1319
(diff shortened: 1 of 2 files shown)


james-garner-canonical:
The bits of the release notes that aren't in this PR should also be proofread, so I'll post them here.

I was intending to just use `2.17.0` as the "Release Title" unless someone has a suggestion for a summary to follow that.

## What's Changed

This release features the ability to install Scenario with `ops[testing]`, improvements in the use of secrets (expanding the ID to the full URI and fixing the inability to call `set_info` and `set_content` in the same hook), and JujuVersion property for Pebble log forwarding to Loki.

Additionally, `ops.main` is now type hinted correctly and will no longer require a `type: ignore` directive from users. `StatusBase` subclasses will have the correct inferred type now. `InvalidStatusError`s will now be raised pre-emptively in some cases that would result in a `ModelError` after a round-trip to Juju.

Among other changes, documentation has seen various improvements including a new top-level intro, while CI improvements include new scheduled smoke and unit tests, running published charms with the `main` branch of `ops`.

## New Contributors

* @theofpa made their first contribution in https://github.com/canonical/operator/pull/1349

CHANGES.md
```diff
@@ -1,38 +1,97 @@
+# 2.17.0 - 26 Sep 2024
+
+## Features
+
+* optionally install Scenario with `ops[testing]` and expose the names in ops.testing in (#1381)
```
tonyandrewmeyer, line 5: This is probably missing from HACKING.md: we put a capital letter for each of the bullet points (but annoyingly conventional commits wants a lower-case letter, so it's a manual job).

(Same comment for all the bullets).

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+* add a JujuVersion property for Pebble log forwarding to Loki in (#1370)
```
tonyandrewmeyer, line 6: We strip out the "in" at the end as well, e.g.

```suggestion
* add a JujuVersion property for Pebble log forwarding to Loki (#1370)
```

Same comment for all the bullets.

Tbh, I think the link/don't link aspect could be reconsidered. Feel free to bring it up in a team meeting.

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+* expand the secret ID out to the full URI when only given the ID in (#1358)
+* pre-emptively raise `InvalidStatusError` instead of waiting for Juju:
+    * make it an error to call `CollectStatusEvent.add_status` with error or unknown in (#1386)
+    * document and validate settable status values in `_ModelBackend.set_status` in (#1354)
+
+## Fixes
+
+* rework `ops.main` type hints to allow different flavours (callable class) in (#1345)
```
tonyandrewmeyer, line 14: I feel like the impact of this won't be understood, even though this is the PR title and it does do what it says.

Maybe:

```suggestion
* Change ops.main() so that you don't need to `type: ignore` it (#1345)
```

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+## Continuous Integration
+
+* periodically run the unit tests of all GitHub-hosted published charms in (#1365)
+* update the TIOBE reporting for the changes in coverage calculation in (#1367)
+* spell-check the code as part of linting in (#1388)
+* run the smoke tests on a schedule in (#1387)
+
+## Testing
+
+* properly mock the environment in (#1385)
```
tonyandrewmeyer, line 35: This is my bad PR title sorry.

```suggestion
* Fix tests that leaked environment variables (#1385)
```

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+## Chores
+
+* note juju version on legacy workaround in (#1355)
+* re-enable test now that Pebble directory permissions are fixed in (#1363)
+* generate warnings for events that will be removed in Juju 4.0 in (#1374)
+* update charm pins in (#1390)
+* update charm pins in (#1375)
+* bump version in (#1348)
+* bump canonical/setup-lxd version in (#1350)
+* bump actions/attest-build-provenance from 1.4.0 to 1.4.2 in (#1351)
```
tonyandrewmeyer, line 52: My preference is that housecleaning stuff like this is not included (you can read the git log for that). I think that this is the team decision also, although someone can correct me if not.

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+
+## Refactoring
+
+* move the content of `ops.testing` to `ops._private.harness` in (#1369)
+* keep the `unittest.mock` names in the 'mock' namespace in (#1379)
+* deprecate `StatusBase.register` decorator in (#1384)
+
+## Chores
+
+* note juju version on legacy workaround in (#1355)
```
tonyandrewmeyer, line 45:
```suggestion
* Note Juju version on legacy workaround (#1355)
```

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+* update charm pins in (#1375)
+* bump version in (#1348)
+* bump canonical/setup-lxd version in (#1350)
+* bump actions/attest-build-provenance from 1.4.0 to 1.4.2 in (#1351)
+
+# 2.16.1 - 5 Sep 2024
+
+## Fix
+
+* avoid changing `os.environ` in Harness in (#1359)
```
tonyandrewmeyer, line 58: Let's use the same text as in the release:

```suggestion
* Don't alter os.environ when creating a Harness (#1359)
```

CHANGES.md
```diff
@@ -1,38 +1,97 @@
...
+
 # 2.16.0 - 29 Aug 2024
 
 ## Features
 
-* Add the description field to SecretInfo in https://github.com/canonical/operator/pull/1338
+* Add the description field to SecretInfo in (#1338)
```
tonyandrewmeyer, line 64: I would dump all the "in"s in these too, sorry.

CHANGES.md
james-garner-canonical: And here I went and corrected some of my accidentally capitalised PRs to lowercase to match the others ...

CHANGES.md
james-garner-canonical: I noticed that other releases didn't include these items, but I also didn't see anything about removing them in HACKING.md. I'm happy to remove them.

james-garner-canonical: Which I've now done

tonyandrewmeyer:
> I was intending to just use `2.17.0` as the "Release Title" unless someone has a suggestion for a summary to follow that.

Hmm, we have had "names" except for 2.1.0 and 2.16.0. I think we should be consistent either way. I'm pretty sure that the names don't show up anywhere other than in the GitHub releases page (they don't show in the PyPI release history, for example), although we do use them for the title of the Discourse post as well. I'd be +0 on removing them.

If we keep them then I think you could use "2.17.0: new unit testing API, and Secret fixes" or something like that. The `ops[testing]` optional install and Scenario classes in `ops.testing` is a pretty big deal.

> ## What's Changed
> This release features the ability to install Scenario with `ops[testing]`, improvements in the use of secrets (expanding the ID to the full URI and fixing the inability to call `set_info` and `set_content` in the same hook), and JujuVersion property for Pebble log forwarding to Loki.

The first two of these are pretty significant, so I would suggest a bigger blurb (but then I worked on them, so if this is just bias, call that out!).

Maybe:

```markdown
This release adds a state-transition testing API for unit tests, formerly known as Scenario (many thanks to @PietroPasotti for his work developing and maintaining the library). Harness is still available, but we encourage all charmers to use the new API for unit tests going forward. The API is an optional install, for example: `pip install ops[testing]`, so that the test API code does not get bundled into the charms. All of the classes that you would previously find when using `ops-scenario` in the `scenario` namespace can then be found in `ops.testing`. For example:

from ops import testing

state = ctx.run(ctx.on.start(), testing.State(leader=True))
assert state.unit_status == testing.ActiveStatus()

There are also two improvements to Juju Secrets: firstly, the `.id` of a `Secret` will always include the model UUID, so it can be reliably used for cross-model relations. A more significant change is that, previously, if a charm called `set_contents()` and `set_info()` in the same hook (even if not the same event handler, or in a deferred event handler) only the last call would have any effect. This is now changed so that the calls accumulate - you can set both content and metadata in the same hook, and also if you `set_contents()` twice the contents will be combined. The Juju team intend to make this the `secret-set` behaviour in the future as well.
```

> Additionally, `ops.main` is now type hinted correctly and will no longer require a `type: ignore` directive from users. `StatusBase` subclasses will have the correct inferred type now. `InvalidStatusError`s will now be raised pre-emptively in some cases that would result in a `ModelError` after a round-trip to Juju.

I think I would skip this for the release notes (the list of changes is there if they want to see it), but use it for the Discourse post (where the changes are not all listed).

> Among other changes, documentation has seen various improvements including a new top-level intro, while CI improvements include new scheduled smoke and unit tests, running published charms with the `main` branch of `ops`.

Similarly, I would definitely skip this bit for the release notes but use it in the Discourse post. But for the last phrase, "running the unit tests of all charms publicly listed on CharmHub and source-hosted on GitHub against the `main` branch of `ops".

CHANGES.md
tonyandrewmeyer: > And here I went and corrected some of my accidentally capitalised PRs to lowercase to match the others ...

Maybe we should ignore the conventional commits rule going forward and have "feat: Adds a new Foo to the Bar class"? I like that more, personally, and it makes this bit a little less tedious.

I say this every month, but I also need to remember to think about "reviewing" the PR title when I do reviews so that I suggest improvements then rather than when we come to the release.

CHANGES.md
tonyandrewmeyer: > I noticed that other releases didn't include these items, but I also didn't see anything about removing them in HACKING.md.

Good point. Probably should be part of:

>> 6. Where appropriate, collapse multiple tightly related bullet points into a single point that refers to multiple commits.

Feel free to open a PR for HACKING.md improvements as well - it's always most insightful when people do their first release.

dimaqq (approved):

IronCore864 (approved):

PR #1395, opened 2024-09-26
chore: bump version to 2.17.1.dev0

james-garner-canonical:


ops/version.py
```diff
@@ -17,4 +17,4 @@
 This module is NOT to be used when developing charms using ops.
 """
 
-version: str = '2.17.0'
+version: str = '2.17.1.dev0'
```
tonyandrewmeyer, line 20: So we don't actually have any process around this, but what we have done so far is bump the minor version, with the expectation that the next release will have at least one feature (so would be `2.18.0.dev0` in this case). However, that's certainly not guaranteed, and the next one will be a shorter turnaround because of the sprint, I presume.

To a certain extent it doesn't matter, it's just an indicator that it's not a released version and not the version that's from PyPI. We used to have a complicated git hash thing and simplified to this since when you're using ops from a checkout you mostly don't care exactly which revision it is (or if you do, you installed it with that revision specified).

So I'm fine with this, but just FYI the more common number would have been 18 :)

tonyandrewmeyer (approved):

ops/version.py
benhoyt: Yeah, I think we should change this to 2.18.0.dev0 just to set the more common precedent and avoid confusion.

dimaqq (approved):

PR #1397, opened 2024-09-26
docs: use MaintenanceStatus for local issues

tonyandrewmeyer:
`WaitingStatus` should be used when waiting on an integrated charm. `MaintenanceStatus` should be used when waiting for something local to the charm. We missed an example in `can_connect()` when clarifying the statuses earlier in the year.


james-garner-canonical (approved):
Makes sense, and this is just changing the example in the docs to reflect what's recommended and used elsewhere.

dimaqq (approved):

PR #1399, opened 2024-09-27
docs: explicitly document that collect-status is is run on *every* hook

### LINKED ISSUE #1265: Should collect-status events only be triggered after observed events?
In our discussion of @tonyandrewmeyer's change-updated spec [OP045](https://docs.google.com/document/d/1EgRwMarsAOB04zWdSpcHbP0VykRApxQlyNFF1oC2Gzw/edit), we noted that collect-status events are currently run at the end of *every* hook. This is as designed and [documented](https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent) as "Event triggered at the end of every hook to collect statuses for evaluation."

However, @hpidcock noted that this is probably not ideal, and they should only be triggered after every hook that the charm is observing (with `Framework.observe`). For one thing, it's easier to cause an infinite loop of events if the charm is doing something odd in collect-status which triggers another event.

We should think through the pros/cons of this and decide whether to change it or not (changing it would strictly-speaking be backwards incompatible, but that may be acceptable here).

james-garner-canonical:
We decided in #1265 that, for backwards compatibility, the behaviour of collect-status won't be changing. That is, it will continue to run at the end of every hook, even if the charm does not provide any observers for that hook. This PR makes this clear in the docs. Closes #1265.

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -1100,7 +1100,12 @@ class CollectStatusEvent(LifecycleEvent):
     way after the end of every hook, it should observe the
     :attr:`collect_app_status <CharmEvents.collect_app_status>` or
     :attr:`collect_unit_status <CharmEvents.collect_unit_status>` event,
-    respectively.
+    respectively. Note that this event is triggered at the end of *every* hook,
+    so any observer of these events will be run on every hook, regardless of
+    whether the charm observes that hook event. For example, even if you don't
+    observe :attr:`config_changed <CharmEvents.config_changed>`, your collect status
+    observer will run at the end of every :attr:`config_changed <CharmEvents.config_changed>`
+    event.
 
     The framework will trigger these events after the hook code runs
     successfully (``collect_app_status`` will only be triggered on the leader
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -1100,12 +1100,10 @@ class CollectStatusEvent(LifecycleEvent):
     way after the end of every hook, it should observe the
     :attr:`collect_app_status <CharmEvents.collect_app_status>` or
     :attr:`collect_unit_status <CharmEvents.collect_unit_status>` event,
-    respectively. Note that this event is triggered at the end of *every* hook,
-    so any observer of these events will be run on every hook, regardless of
-    whether the charm observes that hook event. For example, even if you don't
-    observe :attr:`config_changed <CharmEvents.config_changed>`, your collect status
+    respectively. Note that observing the specific hook event is not required. For example, even
+    if you don't observe :attr:`config_changed <CharmEvents.config_changed>`, your collect status
     observer will run at the end of every :attr:`config_changed <CharmEvents.config_changed>`
-    event.
+    event, as well as all others.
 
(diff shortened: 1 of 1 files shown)


ops/charm.py
```diff
@@ -1100,7 +1100,12 @@ class CollectStatusEvent(LifecycleEvent):
     way after the end of every hook, it should observe the
     :attr:`collect_app_status <CharmEvents.collect_app_status>` or
     :attr:`collect_unit_status <CharmEvents.collect_unit_status>` event,
-    respectively.
+    respectively. Note that this event is triggered at the end of *every* hook,
+    so any observer of these events will be run on every hook, regardless of
+    whether the charm observes that hook event. For example, even if you don't
+    observe :attr:`config_changed <CharmEvents.config_changed>`, your collect status
+    observer will run at the end of every :attr:`config_changed <CharmEvents.config_changed>`
+    event.
```
tonyandrewmeyer, line 1108: I feel like we're maybe now going *too* far. There are 3 "every hook" comments in the doc now :)

Maybe:
```suggestion
    respectively. Note that observing the specific hook event is not required. For example, even
    if you don't observe :attr:`config_changed <CharmEvents.config_changed>`, your collect status
    observer will run at the end of every :attr:`config_changed <CharmEvents.config_changed>`
    event, as well as all others.
```

tonyandrewmeyer (commented):
It's good but I feel like we are maybe *over*emphasising things now. I've suggested a small tweak that cuts it down a bit.

ops/charm.py
james-garner-canonical: Thanks, that's much better

tonyandrewmeyer (approved):

ops/charm.py
```diff
@@ -1100,7 +1100,10 @@ class CollectStatusEvent(LifecycleEvent):
     way after the end of every hook, it should observe the
     :attr:`collect_app_status <CharmEvents.collect_app_status>` or
     :attr:`collect_unit_status <CharmEvents.collect_unit_status>` event,
-    respectively.
+    respectively. Note that observing the specific hook event is not required. For example, even
```
benhoyt, line 1103: This seems a bit confusing to me, in part because it's unclear what "the specific hook event" is referring to. I also don't think it needs to be in this first paragraph.

I wonder if we could make it even more succinct, and drop the example (hopefully we can make the language clear enough we don't need an example for this?). What about tweaking the paragraph below like so:

> The framework will trigger these events after the hook code runs
> successfully, on every Juju event whether it's `observe`d or not
> (``collect_app_status`` will only be triggered on the leader unit). ...

benhoyt: Or as a new sentence:

> The framework will trigger these events after the hook code runs
> successfully (``collect_app_status`` will only be triggered on the leader
> unit). **This happens on every Juju event whether it was `observe`d or not.**

james-garner-canonical: Sounds good, that's very concise and clear

benhoyt (approved):

PR #1400, opened 2024-09-27
feat: remove pebble enum | str unions

james-garner-canonical:
Resolving #1287,  this PR removes the use of `str` and `enum` member unions connected to the `pebble` module.

Generally this just involves typing `dict`s generated by pebble as containing `str` values, and using the `enum` types exclusively in `ops`. The major change is that unknown values from pebble result in an `UNKNOWN` member of the `enum`, rather than being passed around as `str`s.

Reviewers will notice some block comments calling attention to potential issues with the current code, questions about my design choices, and suggestions for further changes. Feedback on these areas would be especially appreciated.


ops/pebble.py
```diff
@@ -1054,15 +1056,20 @@ class ServiceInfo:
...
         self.current = current
 
     def is_running(self) -> bool:
         """Return True if this service is running (in the active state)."""
+        if not isinstance(self.current, ServiceStatus):
+            raise ValueError(
+                'self.current must be a ServiceStatus member (e.g. ServiceStatus.ACTIVE)'
+                f', not {self.current}'
+            )
         return self.current == ServiceStatus.ACTIVE
```
james-garner-canonical, line 1073: This comparison would be silently `False` if `self.current` was accidentally a `str`.

ops/pebble.py
```diff
@@ -1098,11 +1107,16 @@ def __init__(self, name: str, raw: Optional[CheckDict] = None):
...
-            level = dct.get('level', '')
-        self.level = level
+            warnings.warn(f'Unknown CheckLevel value {dct_level!r}')
+            self.level = CheckLevel.UNKNOWN
+        #
+        # these are all Optional in CheckDict and here, why '' instead of None?
+        # note that all falsey values are filtered out in to_dict
         self.period: Optional[str] = dct.get('period', '')
         self.timeout: Optional[str] = dct.get('timeout', '')
         self.threshold: Optional[int] = dct.get('threshold')
```
james-garner-canonical, line 1122: In general a lot of values are marked as `Optional` when they could perhaps be simply left out of the `TypedDict`. I haven't fully checked what the values received from Pebble for these dicts can be -- can these items be the `go` version of `None` (`null`? `nil`?). It seems like if so, we should use `None` instead of `''`, and if not we shouldn't use `Optional`.

ops/pebble.py
```diff
@@ -1124,7 +1138,20 @@ def __init__(self, name: str, raw: Optional[CheckDict] = None):
...
+        # c1 = Check.from_dict(d)
+        # c2 = Check.from_dict(c1.to_dict())
+        # c1.level == CheckLevel.UNKNOWN
+        # c2.level == CheckLevel.UNSET
+        #
+        # so do we want to just use 'unknown' for CheckLevel.UNKNOWN here?
+        # we still get a warning on making `c1` but not on `c2`, but maybe that's ok
+        # but is it useful to have 'unknown' in the dict?
+        # It'll error if that value makes it to pebble, right?
         fields = [
```
james-garner-canonical, line 1155: I'm not really sure what behaviour we want here with `CheckLevel.UNKNOWN`.

ops/pebble.py
```diff
@@ -1205,9 +1232,20 @@ def _merge(self, other: Check):
...
                 continue
+            # how to handle CheckLevel.UNKNOWN?
+            #
+            # it would be skipped here if we used other.to_dict
+            #
+            # previously it would have been <some-unknown-value-string>
+            # while we might want <some-unknown-value-string> to overwrite,
+            # would we ever want Checklevel.UNKNOWN to overwrite?
+            if value is CheckLevel.UNKNOWN:
+                continue
```
james-garner-canonical, line 1248: I think we should just use `other.to_dict()` here, is there a reason we weren't already?

ops/pebble.py
```diff
@@ -3046,6 +3105,9 @@ def get_checks(
         """
         query: Dict[str, Any] = {}
         if level is not None:
+            # is this bad now that value could be 'unknown'?
+            # should we explicitly avoid it here?
+            # if level is not None and level.value is not CheckLevel.UNKNOWN:
             query['level'] = level.value
```
james-garner-canonical, line 3111: Another case where it's not clear how to handle `CheckLevel.UNKNOWN` on the outgoing side.

ops/pebble.py
```diff
@@ -3098,7 +3160,7 @@ def get_notices(
         *,
         users: Optional[NoticesUsers] = None,
         user_id: Optional[int] = None,
-        types: Optional[Iterable[Union[NoticeType, str]]] = None,
+        types: Optional[Iterable[NoticeType]] = None,
         keys: Optional[Iterable[str]] = None,
     ) -> List[Notice]:
         """Query for notices that match all of the provided filters.
```
james-garner-canonical, line 3166: I wonder if this will break any users.

test/test_pebble.py
```diff
@@ -443,8 +443,9 @@ def test_file_info_from_dict(self):
         d['user'] = 'bob'
         d['group-id'] = 34
         d['group'] = 'staff'
-        info = pebble.FileInfo.from_dict(d)
-        assert info.type == 'foobar'
+        with pytest.warns(UserWarning):
+            info = pebble.FileInfo.from_dict(d)
+        assert info.type == pebble.FileType.UNKNOWN
```
james-garner-canonical, line 448: Should we be using a different kind of warning here?

ops/_private/harness.py
```diff
@@ -3863,7 +3863,7 @@ def _notice_matches(
         # For example: if user_id filter is set and it doesn't match, return False.
         if user_id is not None and not (notice.user_id is None or user_id == notice.user_id):
             return False
-        if types is not None and notice.type not in types:
+        if types is not None and notice.type.value not in types:
             return False
```
james-garner-canonical, line 3867: `types` is a collection of `str`s. I think this is mimicking the logic of `pebble`, so I think it makes sense for us to use `notice.type.value` here. Was this always silently broken? Or silently worked because `notice.type` was often a `str` -- while now it's supposed to always be an `enum` member?

ops/pebble.py
james-garner-canonical: Would a better solution be to switch to using `enum.StrEnum` instead of `enum.Enum`? This would seem to match the intent of the original code better. With `StrEnum`, `ServiceStatus.ACTIVE == 'active'`, so passing around strings instead of enum members would silently succeed instead of silently failing in the absence of a type check.

dimaqq: StrEnum is a py 3.11 feature, isn't it?

We've got to support py 3.8 because Juju supports specific Ubuntu releases as bases and that's where the Python interpreter comes from.

ops/pebble.py
dimaqq: There's not a single user of .get_notices() yet.
Source: charm-analysis and github search.

ops/_private/harness.py
dimaqq: I think you can answer that question, as notices are not really used by charmers yet.
Which means that the question can be rephrased as "do we have a test for this harness feature?" and "does that test actually work?".

dimaqq:
Only superficial comments from me, I didn't read the change properly yet.

ops/pebble.py
james-garner-canonical: Good catch, @dimaqq, thanks.

ops/pebble.py
```diff
@@ -134,7 +134,7 @@
     'CheckDict',
     {
         'override': str,
-        'level': Union['CheckLevel', str],
+        'level': str,
```
tonyandrewmeyer, line 137: The charms define `CheckDict`s (as part of `Layer`s), right? So this isn't something that's just coming back from Pebble.

So if there's charm code like:

```python
{
...
    "checks": {
        "override": True,
        "level": pebble.CheckLevel.READY,
        ...
    },
...
}
```

then wouldn't this cause problems? I think it's pretty common to be building layers from dictionaries rather than using `Layer` and `Check` and `Service` classes (etc) explicitly.

ops/pebble.py
```diff
@@ -1071,11 +1078,13 @@ def from_dict(cls, d: _ServiceInfoDict) -> ServiceInfo:
         try:
             startup = ServiceStartup(d['startup'])
         except ValueError:
-            startup = d['startup']
+            warnings.warn(f'Unknown ServiceStartup value {d["startup"]!r}')
```
tonyandrewmeyer, line 1081: It would be nice if these warnings helped guide people towards solving the problem somehow. I assume our expectation here is that the charm is using a version of Pebble that is more recent than the version of ops knows of, and some new value was added to Pebble (rather than someone is passing in data that didn't come from Pebble, for example).

So perhaps the solution should generally be "update ops"? In which case we could perhaps say something like "try updating to the latest version of ops"?

tonyandrewmeyer (commented):
This is a fairly light review - I agree with much of Ben's comments, and it's clear that you've discussed it together and will be making adjustments, so I'm assuming it's better for me to just do this lightly now and come back after those are done and do a more thorough review then.

ops/pebble.py
james-garner-canonical: @benhoyt what do you think about changing these warning messages to provide problem solving help, and if so what would be the right message? "Unknown value blah, the ops version may be lower than the pebble version" or something?

ops/model.py
```diff
@@ -3047,7 +3047,7 @@ def get_notices(
         *,
         users: Optional[pebble.NoticesUsers] = None,
         user_id: Optional[int] = None,
-        types: Optional[Iterable[Union[pebble.NoticeType, str]]] = None,
+        types: Optional[Iterable[pebble.NoticeType]] = None,
```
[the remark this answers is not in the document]
james-garner-canonical: The code currently handles both the str and enum member cases. Maybe this is a case where we should keep the `Union` for a better user facing experience, @benhoyt?

Or if we are going ahead with removing the `Union`, then I guess I should change the code to stop sneakily handling strings as well as enum members.

Note that @dimaqq and @benhoyt already indicated that they weren't worried about breaking things with this function signature change in a [different thread](https://github.com/canonical/operator/pull/1400#discussion_r1782031819).

ops/pebble.py
benhoyt: Yeah, I think something (succinct) like that would be good. Perhaps don't compare ops version with pebble version, as those aren't related. Maybe what Tony said, or "; do you need to update ops?"

ops/model.py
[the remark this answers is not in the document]
benhoyt: Yeah, I'm not worried about breaking things. But I do sort of agree with Tony that `types=('custom', 'change-updated')` is nicer than `types=(pebble.NoticeType.CUSTOM, pebble.NoticeType.CHANGE_UPDATED)`. Plus, I guess we shouldn't get in the habit of sneaking breaking changes in, even if here we think it's fine. Let's revert (just this type signature) to the union, given that and Tony's (fair) comment.

ops/pebble.py
james-garner-canonical: You're right, this seems problematic, I was assuming these dictionaries would only come from pebble, but I notice that `CheckDict` is a public type. @benhoyt, another case where we should keep the `Union` to avoid breaking changes?

benhoyt: Hmm, interesting. Would Tony's code actually work, or would the `to_dict` then yield `"level": "CheckLevel.READY"` or something. Yeah, looks like it:

```
>>> print(yaml.dump(d))
exec:
  command: echo foo
http:
  url: https://example.com/
level: !!python/object/apply:ops.pebble.CheckLevel
- ready
override: replace
period: 10s
tcp:
  port: 80
threshold: 5
timeout: 3s
```

So I think this is a plain bug right now (in the `CheckDict` type annotation). Would you agree @tonyandrewmeyer?

tonyandrewmeyer: For all of the changes proposed here, if it currently doesn't work, then I'm fine with the theoretical backwards compatibility break, since in practice it never could have been used, and you could argue that the bug was suggesting it could be.

james-garner-canonical: As far as I can tell, both previously and currently, you can pass in a `CheckLevel` member instead of a string, and it will just work. So the annotation change that Tony is highlighting is just making it a type checking error to use an enum member in the dict instead of a `str`. If this is the case, then we should probably allow the `Union` here, right?

Both previously and currently, we essentially do `CheckLevel(check_dict['level'])` in the `Check` constructor, and (e.g.) `CheckLevel(CheckLevel.READY)` just returns `CheckLevel.READY`.

`CheckLevel.to_dict` always expected that `Check.level` could be a `CheckLevel` member, and used `.value` in that case.

How can I reproduce your failing example, @benhoyt?

benhoyt: I did it like this:

```
>>> d = {
...             'override': 'replace',
...             'level': 'alive',
...             'period': '10s',
...             'timeout': '3s',
...             'threshold': 5,
...             # Not valid for Pebble to have more than one of http, tcp, and exec,
...             # but it makes things simpler for the unit tests.
...             'http': {'url': 'https://example.com/'},
...             'tcp': {'port': 80},
...             'exec': {'command': 'echo foo'},
...         }
>>> 
>>> d['level'] = pebble.CheckLevel.READY
>>> import yaml
>>> print(yaml.dump(d))
exec:
  command: echo foo
http:
  url: https://example.com/
level: !!python/object/apply:ops.pebble.CheckLevel
- ready
override: replace
period: 10s
tcp:
  port: 80
threshold: 5
timeout: 3s
```

Which I presume is what happens when you create a `pebble.Layer` and then pass it to `add_layer`: it calls `layer.to_yaml()`. And `to_yaml` does `yaml.safe_dump(self.to_dict())`.

james-garner-canonical: Thanks, @benhoyt.

Looks like `Pebble.Client.add_layer` does go through `Layer` if you pass a `dict`:
```python
elif isinstance(layer, dict):
    layer_yaml = Layer(layer).to_yaml()
```
Where `Layer`'s constructor then passes any `CheckDict`s through `Check`:
```python
self.checks = {name: Check(name, check) for name, check in d.get('checks', {}).items()}
```
And `Check` now (and previously on main) does the equivalent of:
```python
self.level = CheckLevel(check_dict['level'])  # actually using get + try..except for default/unknown
```
Finally, `Layer.to_yaml` calls `Check.to_dict`, which uses `self.level.value`, and we get a `str` in our yaml the end.

So it seems that it's always worked if `check_dict['level']` is a `CheckLevel` member or a `str`, and it continues to work at runtime, just with a type checking error now. I think that `CheckDict.level` should continue to be typed as `Union[CheckLevel, str]` to remove that type checking error for backwards compatibility.

This doesn't really seem like a bad pattern that we should try to steer charm users away from either, does it?
```python
{
    ...
    "checks": {
        "override": True,
        "level": pebble.CheckLevel.READY,
        ...
    },
    ...
}
```
I kind of prefer it to using strings, as my editor will complain about typos etc. Though you know, if we wanted to provide the same benefits to people using strings, we could consider eventually using `Literal`s instead of `str`s in the Pebble `TypedDict`s, but that's probably something to consider in a future issue.

james-garner-canonical: Sounds good. From a quick look at the charms we use for super-tox, no one seems to be using the `"level": pebble.CheckLevel.READY` style.
```bash
ag 'CheckLevel'
```
And using strings seems to be very common.
```bash
ag --case-sensitive '[\'"]level[\'"]:'
```

benhoyt: Decision after daily discussion is to go ahead with current approach: basically consider that passing `"level": pebble.CheckLevel.READY,` was a bug -- as the intent was these dicts should all be basic JSON/YAML-ifiable types.

james-garner-canonical: Note: Following a question in daily discussion, using (e.g.) `pebble.CheckLevel.READY` works at runtime, but not because of any explicit special casing of `str` vs `CheckLevel`.

It only works because of how enum constructors are happy to take either a value or a member as their argument (i.e. `CheckLevel('ready') == CheckLevel.READY == CheckLevel(CheckLevel.READY)`.

As decided, `level": pebble.CheckLevel.READY` will result in a type checking error, but we're not going to add a runtime error (just like we decided not to in other cases).

james-garner-canonical:
I think all reviewer concerns have now been addressed, so this PR should now be ready for final reviews.

ops/pebble.py
```diff
@@ -1308,8 +1318,8 @@ class FileInfo:
     name: str
     """Base name of the file."""
 
-    type: Union[FileType, str]
-    """Type of the file ("file", "directory", "symlink", etc)."""
+    type: FileType
+    """Type of the file (FileType.FILE, FileType.DIRECTORY, FileType.SYMLINK, etc)."""
```
benhoyt, line 1322: Nit: We shouldn't need to link to these, as the range of values is known from the type, and you can click and see the full list in the docs. I think either say `(file, directory, symlink, etc)` or drop them entirely (my slight preference).

ops/pebble.py
```diff
@@ -1496,6 +1507,9 @@ class NoticeType(enum.Enum):
     notices is the change ID.
     """
 
+    UNKNOWN = 'unknown'
+    """Used if we receive an unrecognised notice type (e.g. from future Pebble versions)."""
```
benhoyt, line 1511: Nit: I think in docs its better to spell out these abbreviations. For one thing, the difference between "e.g." and "i.e." can be confusing (e.g. => for example, i.e. => that is).

```suggestion
    """Used if we receive an unrecognised notice type (for example, from future Pebble versions)."""
```

benhoyt (approved):
I'm good with this, thanks! Just two nit comments.

ops/pebble.py
james-garner-canonical: Sure, happy to remove the examples. I made this change because strings are now invalid arguments, and only enum members are expected, but it's very redundant with the typing.

---

While fixing this I noticed that the following attribute has:
```python
size: Optional[int]
"""Size of the file (will be 0 if ``type`` is not "file")."""
```
Which I've changed to:
```python
"""Size of the file (will be 0 if :attr:`type` is not :attr:`FileType.FILE`)."""
```
Let me know if you'd prefer something different here. I thought it was worth linking as `type` appears after `size` in the docs, since attributes are sorted alphabetically.

james-garner-canonical: Btw `size` is another case where `Optional` feels a little bit off to me. Not something to address now, but maybe I should add this as an example in that issue. But maybe `None` is actually correct here?

In pebble we have:
```go
Size         *int64   `json:"size,omitempty"`
```
Does this omit if size is 0?

In ops we have:
```python
_FileInfoDict.size: NotRequired[Optional[int]]  # could be NotRequired[int]
```
```python
# in FileInfo.from_dict
size=d.get('size')  # None if missing -- should it be 0 instead? Depends on pebble behaviour
```
```python
FileInfo.size: Optional[int]  # could remove Optional here as well if 0 is default value
"""Size of the file (will be 0 if ``type`` is not "file")."""
```

benhoyt: Oh yes, it seems like the docstring is incorrect here. Pebble will omit it for non-files, so it'll come through as None. We should change the docstring to "will be None if ..." -- seems fine to do that in as a drive-by change.

tonyandrewmeyer (changes requested):
I ran super-tox (with `-e unit`) and found two failures:

synapse-operator gets 93 failing tests, that look like this:

```
  File "/home/tameyer/code/charm-analysis/.cache/synapse-operator/src/pebble.py", line 113, in check_nginx_ready
    return check.to_dict()
           ^^^^^^^^^^^^^^^
  File "/home/tameyer/code/charm-analysis/.cache/synapse-operator/.tox/unit/lib/python3.11/site-packages/ops/pebble.py", line 1138, in to_dict
    ('level', self.level.value),
              ^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'value'
```

github-actions-exporter-operator gets 5 failures that look like:

```
  File "/home/tameyer/code/charm-analysis/.cache/github-actions-exporter-operator/src/github_actions_exporter.py", line 31, in check_ready
    return check.to_dict()  # type: ignore
           ^^^^^^^^^^^^^^^
  File "/home/tameyer/code/charm-analysis/.cache/github-actions-exporter-operator/.tox/unit/lib/python3.11/site-packages/ops/pebble.py", line 1138, in to_dict
    ('level', self.level.value),
              ^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'value'
```

This doesn't necessarily block this, but we should definitely investigate what's going on there first (if we still want the changes as-is, then I think we should consider opening PRs to solve the issues for those two repos).

james-garner-canonical:
Thanks for catching that, @tonyandrewmeyer! I need to get in the habit of running super-tox more with your setup to diff between two branches and filter out all the unrelated failures.

`synapse-operator` is doing things [like](https://github.com/canonical/synapse-operator/blob/9311b48d1fd4eca7119a6a3fb2a04c2f24599637/src/pebble.py#L109-L113)
```python
    check = Check(synapse.CHECK_NGINX_READY_NAME)
    check.override = "replace"
    check.level = "ready"
    check.http = {"url": f"http://localhost:{synapse.SYNAPSE_NGINX_PORT}/health"}
    return check.to_dict()
```
as is `github-actions-exporter` ([link](https://github.com/canonical/github-actions-exporter-operator/blob/8debf0ed861dbcbc05f0f6ab7c5d79562cbea426/src/github_actions_exporter.py#L26C1-L31C43))
```python
    check = Check(CHECK_READY_NAME)
    check.override = "replace"
    check.level = "ready"
    check.tcp = {"port": GITHUB_METRICS_PORT}
    # _CheckDict cannot be imported
    return check.to_dict()  # type: ignore
```
Presumably to avoid having to use a dictionary to specify the arguments for `Check`
```
    def __init__(self, name: str, raw: Optional[CheckDict] = None):
```
This is ironic, because a `str` is fine for `level` in a `CheckDict`, and gets turned into a `CheckLevel` in `__init__`.

They get an error because they bypass the `__init__` logic and set the attribute directly, which used to be ok since we had logic for passing arbitrary strings to pebble in case of an unrecognised value -- that's what's being removed in this PR, and we're assuming internally that `Check.level` is a `CheckLevel`, not a `str`.

One interesting solution would be to define a custom `level` `@property` that takes care of the `str` -> `CheckLevel` conversion on assignment, if we want to support this non-idiomatic constructor pattern that they're using to get around our (arguably) non-idiomatic `__init__` signature. (The other classes tend to have a more standard `__init__` and a separate `from_dict` constructor.)

tonyandrewmeyer:
> Thanks for catching that, @tonyandrewmeyer! I need to get in the habit of running super-tox more with your setup to diff between two branches and filter out all the unrelated failures.

It gives me as a reviewer something to do :wink:.

> They get an error because they bypass the `__init__` logic and set the attribute directly, which used to be ok since we had logic for passing arbitrary strings to pebble in case of an unrecognised value -- that's what's being removed in this PR, and we're assuming internally that `Check.level` is a `CheckLevel`, not a `str`.

Would you say that it's wrong to do it the way they are rather than using the enum - that they were taking advantage of a bug rather than something that we were at least implying was correct? If so then I'm guessing the issue can be fixed with very small PRs on their end, which we could submit for them (and would be backwards compatible, so could be merged ahead of ops releasing).

> One interesting solution would be to define a custom `level` `@property` that takes care of the `str` -> `CheckLevel` conversion on assignment, if we want to support this non-idiomatic constructor pattern that they're using to get around our (arguably) non-idiomatic `__init__` signature. (The other classes tend to have a more standard `__init__` and a separate `from_dict` constructor.)

I don't think we want to do that. I'd rather provide some sort of object-creation-style API in a future release than patch around oddities in the dict-creation-style API we have now.

james-garner-canonical:
> Would you say that it's wrong to do it the way they are rather than using the enum - that they were taking advantage of a bug rather than something that we were at least implying was correct?

No, since the attributes are public, and the `level` attribute is typed as `CheckLevel | str`, you would expect that assigning a `str` to `check.level` would be valid.

> I'm guessing the issue can be fixed with very small PRs on their end, which we could submit for them (and would be backwards compatible, so could be merged ahead of ops releasing)

This is true though, so we could just try to fix these users preemptively (and fix any users super-tox didn't catch after next release, I guess). There are a couple of easy fixes:
1. Put the values they're setting in a dict and call the constructor with that dict. I think this is the 'right' way to use `Check`.
2. A perhaps less invasive change would be to just wrap the appropriate strings in `CheckLevel` calls.

> I don't think we want to do [the wacky property setter idea].

It's definitely a bit weird, and it's supporting a case we're trying to move away from in this PR (using strings and enums interchangeably).

It is a pretty clean way to support the public api that I'm only now realising we had exposed though -- that you could set various attributes with strings or enum members. Maximum backwards compatibility.

But it would be nicer if we could continue with our move to those attributes being simply typed, so I'd be in favour of trying to fix those charms preemptively.

benhoyt:
Hmmm, now that I see how this is being used, and it's in charm code, not just charm tests, I'm a lot more hesitant about preventing `check.level = 'ready'`. We *did* support this, whether we intended to or not. let's finalise the decision during daily.

james-garner-canonical:
In daily sync we decided to walk back the changes made in this PR because we hadn't considered the implications the various pebble related objects having public attributes (for example, `Check.level`).

To maintain backwards compatibility, we need to continue to let users assign `str`s to these attributes. We also need to support users that are expecting to see a an enum member on these attributes (for example, `CheckLevel.READY`). Therefore, public attributes will need to stay typed as `Union`s of `str` and enum types. 

All the attributes changed in this PR are technically public, so presumably all of these need to go back to being unions with `str` (unless @benhoyt is confident that user's wouldn't construct some of these classes manually?):
```python
Check.level: CheckLevel
CheckInfo.level: CheckLevel
CheckInfo.status: CheckStatus
FileInfo.type: FileType
Notice.type: NoticeType
ServiceInfo.startup: ServiceStartup
ServiceInfo.current: ServiceStatus
```

---

Supporting this typing change, we also made a runtime behaviour change: Turning unrecognised `str`s (for example from future pebble versions) into an `.UNKNOWN` member and emitting a warning, instead of just passing the `str` around.

This is a breaking behaviour change. One way that it could lead to an error would be if you get an unknown value from pebble and then try sending it back, you'll send back the value "unknown", which will be an error. It also required updating our tests, where we expected to be able to check an attribute and see our strange string, but instead we'd see the `.UNKNOWN` enum member.

**Do we still want to make this behaviour change?**

One concern is that the logic is in the various constructors, so we'll have a divergence in behaviour if user's instead set attributes directly, like in [the case](https://github.com/canonical/operator/pull/1400#issuecomment-2390510112) motivating us to walk back this PR.

This change might not really be adding any value since we can't guarantee via type annotations that the object's attributes will be enum members rather than strings.

---

The other change made in this PR is changing the types in `TypedDict`s to be `str` not `Union[str, SomeEnum]`.
1. Public `TypedDict`s affected: `CheckDict` -- we [decided](https://github.com/canonical/operator/pull/1400#discussion_r1785501966) to make this change despite it technically being breaking since no one seemed to be using the enum form, and the dict was intended to be serialisable. **Do we want to keep this?**
2. Private `TypedDict`s affected: `_ServiceInfoDict`, `_CheckInfoDict` -- **worth keeping these changes?**

benhoyt:
> unless @benhoyt is confident that user's wouldn't construct some of these classes manually?

I'm pretty confident these wouldn't be constructed in charm code, because they're only returned by pebble. But you had a good point that they may well be constructed (this way) in tests, so I think we should play it safe here too.

Regarding `.UNKNOWN` -- you have a good point about the round trip potentially breaking things. Let's be conservative and keep this as is.

For the public TypedDicts, I don't think we should change those for the same reasons. However, for the private TypedDicts it seems to me like we should fix those (keep those changes).

james-garner-canonical:
I've rolled back the changes made in this PR to only remove the union types from the private `TypedDict`s. No other typing or runtime changes remain. This requires updating two tests to pass static type checking (the actual tests pass without changes).

I double checked for uses of the constructors whose arguments use these private types in charm code, and did not find any uses conflicting with this typing change (see below). Nevertheless, this may possibly break someone's type checking tests the same way it broke ours (there are charms we don't (yet) cover with super-tox), and I guess we'd be in the wrong, as the constructors are all public -- unless the implied privacy of the argument type is sufficient as a warning.

`ServiceInfo.from_dict(cls, d: _ServiceInfoDict)`
One use found with
```bash
ag 'ServiceInfo.from_dict'
```

redis-k8s-operator/tests/unit/test_charm.py
```python
mock_info = {"name": "redis", "startup": "enabled", "current": "active"}
mock_service = ServiceInfo.from_dict(mock_info)
```
Doesn't conflict with changing type of values to `str`.

---

`CheckInfo.from_dict(cls, d: _CheckInfoDict)`
No uses foud with
```bash
ag 'CheckInfo.from_dict'
```

---

`FileInfo.from_dict(cls, d: _FileInfoDict)`
No uses foud with
```bash
ag 'FileInfo.from_dict'
```

benhoyt (approved):
This does seem reasonable to me, thanks.

I guess (for future reference) we should either have `_ServiceInfoDict` and the like public, as they're used in public methods, or make `from_dict` et al private...

tonyandrewmeyer (approved):
Sorry about all the changes this went through, but at least we do end up with some small improvements!

PR #1406, opened 2024-10-02
refactor: import the ops-scenario repository

tonyandrewmeyer:
Adds the content of the ops-scenario repository (including history) into the repository, for easier long-term maintenance. The previous ops-scenario content is located primarily in the new top-level `testing` folder (named for `ops[testing]`), other than some content that has been removed (e.g. separate `docs`, `.pre-commit-config.yaml` and so on) or merged (e.g. `tox.ini`).

There are more than 700 commits because the history is imported, and a *lot* of new code, so this will be an unusual one to review. I suggest starting with the most recent commits (basically everything after what's in ops-scenario) and then reviewing each commit individually. To be clear: I didn't touch any of the ones before this, they were generated with a `git merge --allow-unrelated-histories` from a clone of the canonical/ops-scenario repo.

The commits are:

1. [Moving everything other than .git in ops-scenario to ops-scenario/testing](https://github.com/canonical/operator/commit/e77a45df2a6f3d61f4af41a202a05dbc5b4c43d9)
2. [Copying ops-scenario to operator](https://github.com/canonical/operator/pull/1406/commits/eea56562abf02e2fbee44fc5851d6407e82cd6ea)
3. [Delete files that aren't needed any more, and move some others](https://github.com/canonical/operator/pull/1406/commits/c52067ec8a5e2172c8c765aec0112998b593b09d)
4. [Adjust .gitignore](https://github.com/canonical/operator/pull/1406/commits/819476f744ea34f02ac40ce306154718aa4d6179)
5. [Small linting changes to match with operator](https://github.com/canonical/operator/pull/1406/commits/18c57e7969f47e26e7c1b983dc4447da7ce1a912)
6. [Fix issues that our linter found](https://github.com/canonical/operator/pull/1406/commits/1633648f57120be6d36458e212072023e9b3b828) - mostly unused imports, but also a couple of shadowed tests and other small issues
7. [Add test-readme to operator/tox.ini, install ops-scenario for testing via the local copy](https://github.com/canonical/operator/pull/1406/commits/57a0e431ac47514b2180a8f082269adc2cffc7db)
8. [Adjust the ops-scenario pyproject.toml](https://github.com/canonical/operator/pull/1406/commits/42967b8c89494e08f5ef60bc7402a372321f0fa6)
9. [Drop out-of-date parts of ops-scenario pyproject.toml](https://github.com/canonical/operator/pull/1406/commits/6afc70f0e0947fccb24006ced8ff1513eb5e13cd)
10. [Remove an unnecessary CI workflow](https://github.com/canonical/operator/pull/1406/commits/f07dd9678bac735478f7f6c485e68a91fdb83e37)
11. [Partially adjust the CI workflow for publishing ops-scenario](https://github.com/canonical/operator/pull/1406/commits/58d461ccdc37e025baa595c55ea7b8eccac51e74) - this needs to switch to Trusted Publisher and maybe some other tweaks - it seems best finalised in a separate & dedicated PR
12. [A bunch of fixes so that tox -e static is happy with the ops-scenario code (not tests)](https://github.com/canonical/operator/pull/1406/commits/e075c3f60b3fb1ae6cf16313ead0e274d56eb45d) - this is mixed quality - there are too many `type: ignore` and `Any` uses, but it's reasonable and better than before and gets us to a place where the static checks pass with the same config as for the rest of the code. I feel like further improvements can be done separately (and also ideally type checking the ops-scenario tests, I guess).
13. [Small change to an ops-scenario test to not clash with the ops.charm name](https://github.com/canonical/operator/pull/1406/commits/61941ab99cd0348ebb9578a88eb9351346bbc0dc)
14. [Adjusts the ops-scenario tests to use relative imports for the test_helpers file](https://github.com/canonical/operator/pull/1406/commits/5d5a9c61bd9de066ca808ff810082e9ab4c5a0e5) - I'm not sure this is the best solution, but it seemed ok for now. We actually want to get rid of the `trigger` helper anyway, so maybe the problem just goes away in time.
15. [Adds a dependency for the unit tests](https://github.com/canonical/operator/pull/1406/commits/63df97ebfb8fa89d0f3e9522a109432a2163acb5)
16. [Adjusts pyproject.toml to handle the repo not being flat-layout any more, and static checks ops-scenario](https://github.com/canonical/operator/pull/1406/commits/550936f483299184355cfe6dae2c825ac4e60be8)
17. [Adjusts the ops-scenario tests to pass](https://github.com/canonical/operator/pull/1406/commits/07bbfca3ca9c84f98e2e3fb896e0872e36dbdd42) - these are puzzling. I don't understand why `pytest.raises` does not work for one set of tests (previously shadowed so not running), any advice on that would be great. I'm also not sure why the other tests weren't failing previously, since they were clearly wrong.
18. [Use the ops._private.harness.ActionFailed rather than defining a new one](https://github.com/canonical/operator/pull/1406/commits/6837528fc9c3324ffb8cfd156894feed4f609cde) (as discussed/planned previously - this is needed in this PR to avoid clashing in the tests).
19. [Fix consistency-checker looking for items in the state](https://github.com/canonical/operator/pull/1406/commits/c7abfa8ee3f302513484f5555ed51282f23dc670) - this is a bug (see https://github.com/canonical/ops-scenario/issues/199), and I thought it was required to fix one of the tests, but I think actually a different fix was required, but included this since it's needed rather than back it out.
20. [Ensure that we're testing against the local code and not code from PyPI](https://github.com/canonical/operator/pull/1406/commits/5d49f3021779e9f84034eda9101e69b64a1a95ab)
21. [Skip the weird pytest.raises issue mentioned above for now, rather than working around it](https://github.com/canonical/operator/pull/1406/commits/7d35baa3caf574a74d26e5263c5d6e486542a95b)
22. [Ignore the first two commits listed here in blame](https://github.com/canonical/operator/pull/1406/commits/05114b6be47e68b0d34bf6b107b4638add127398)

Note that I have *not* done the `"` to `'` conversion in this. It seems like that could be a separate and dedicated PR.

There's essentially no code changes, other than addressing issues that the operator tox `unit`, `lint`, or `static` found, just a bunch of moving things around.


testing/src/ops_main_mock.py
```diff
@@ -0,0 +1,288 @@
...
+import ops.framework
+import ops.model
+import ops.storage
+from ops import CharmBase
+
+# use logger from ops._main so that juju_log will be triggered
+try:
+    from ops._main import CHARM_STATE_FILE, _Dispatcher, _get_event_args  # type: ignore
+    from ops._main import logger as ops_logger  # type: ignore
+except ImportError:
```
PietroPasotti, line 21: the imports here should be simplified now that we know what ops version we're being shipped with

testing/src/ops_main_mock.py
```diff
@@ -0,0 +1,288 @@
...
+            ops.jujucontext._JujuContext.from_dict(os.environ),  # type: ignore
+        )
+    dispatcher.run_any_legacy_hook()
+
+    framework = setup_framework(charm_dir, state, event, context, charm_spec)
+    charm = setup_charm(charm_class, framework, dispatcher)
+    return dispatcher, framework, charm
+
+
+class Ops:
```
PietroPasotti, line 209: perhaps we can do that later, but now is a good time to trim down ops_main_mock and start using directly some of the newer stuff in ops

PietroPasotti (commented):
Just a quick look for now, will give some more thoughts later.
Something that comes to mind is: is the idea to do (eventually) `from ops.scenario import ...?` still on the table, or is the 'scenario' namespace going to remain toplevel?

testing/src/ops_main_mock.py
PietroPasotti: or, maybe not, since it's still an independent package that could be installed alongside older ops. nvm

tonyandrewmeyer:
> Something that comes to mind is: is the idea to do (eventually) `from ops.scenario import ...?` still on the table, or is the 'scenario' namespace going to remain toplevel?

As of ops 2.17 you can do this:

```python
from ops import testing
ctx.run(ctx.on.start(), testing.State())
```

The intention is to encourage use in that form.

testing/src/ops_main_mock.py
benhoyt: It's probably true that it could still theoretically be installed alongside older ops, but in that case why wouldn't you also just use an older ops-scenario version? I think part of the reason it's good in a single repo is so we can do atomic commits for things that use `_private` functions.

*However*, it might make sense to keep this PR limited to just copying code 1:1, and then subsequent PRs for updates to those files. Will make review saner I think, and mean we're not tempted to include other changes here.

tonyandrewmeyer: Yeah, I had to bump the minimum ops version in this PR so the compatibility code can be removed, but I would rather do it in another PR, since this one is already complicated to review.

testing/src/ops_main_mock.py
tonyandrewmeyer: I think it is good to do that soon, but not in this PR, which is already very large/complex.

benhoyt:
Thanks @tonyandrewmeyer, this is excellent work. A few thoughts/questions:

- During the import, did you scan through the individual commits to see if there were mistakes like large binary files committed, only to be deleted later? I've been burnt by this before, there was some 100MB+ binary added to the repo history. The main thing is: let's ensure the .git repo size is not too much larger than before (would you be able to provide before/after numbers?).
- Yeah, I think we should probably add those renames to the blame-ignore list.
- We can probably leave the puzzling pytest.raises thing for a separate PR just to keep this limited (TODO and commented-out code is fine for now).
- How do the imports like `from scenario.errors import NoObserverError` in `testing/src/runtime.py` work? In other words, where is that scenario name coming from? Should these be `from .errors import Foo` and the like? Or maybe not -- but I just don't understand how this is working as is.

tonyandrewmeyer:
> * During the import, did you scan through the individual commits to see if there were mistakes like large binary files committed, only to be deleted later?

No. I briefly scanned through the commits (I was originally thinking of ones that said "accidentally committed password") but only the title and that's not really going to tell the whole story. I *could* read through the details of all the commits, but that would take quite a while.

> The main thing is: let's ensure the .git repo size is not too much larger than before (would you be able to provide before/after numbers?).

A fresh ops-scenario clone is 2.6M (.git is 1.8M). operator/.git is 3.2M main@HEAD. operator/.git is 4.8M when cloned with `--single-branch --branch=import-ops-scenario` from my fork. So it seems ok - it is admittedly a ~50% increase in size, but my guess is that the only way around that would be to start squashing things, which seems very complex and time-consuming.

> * Yeah, I think we should probably add those renames to the blame-ignore list.

Since this will be a rebase-and-merge, does that mean I can do that now rather than in a follow-up PR because the hashes won't change?

> * We can probably leave the puzzling pytest.raises thing for a separate PR just to keep this limited (TODO and commented-out code is fine for now).

Ok, I can change that.

> * How do the imports like `from scenario.errors import NoObserverError` in `testing/src/runtime.py` work? In other words, where is that scenario name coming from? Should these be `from .errors import Foo` and the like? Or maybe not -- but I just don't understand how this is working as is.

1. The name comes from the `testing/pyproject.toml` configuration. If you do a `pip install .` inside of `testing`, you'll get a package called `scenario` installed.
2. They probably *should* be relative imports, but I felt like that could wait for a follow-up PR.
3. Doing `pip install .[testing]` at the top level might not be doing what we want for the tests - I think maybe it's pulling `ops-scenario` from PyPI so testing against a mixture of the latest release (when importing from `scenario`) and the local code (everything else). It probably needs to be `pip install .` done in both places instead, maybe with a specific test for `pip install ops[testing]`. I hadn't thought about this - I'll look into it.

benhoyt:
Thanks!

I think 50% bigger is totally reasonable. I was thinking of 10x or worse blow-up. I don't think we need to do anything further here. If there were passwords, they'd been in the public ops-scenario repo already. :-)

> Since this will be a rebase-and-merge, does that mean I can do that now rather than in a follow-up PR because the hashes won't change?

Yeah, that sounds right.

> The name comes from the testing/pyproject.toml configuration. If you do a pip install . inside of testing, you'll get a package called scenario installed. They probably should be relative imports, but I felt like that could wait for a follow-up PR.

Great, that sounds fine to me.

> Doing pip install .[testing] at the top level might not be doing what we want for the tests - I think maybe it's pulling ops-scenario from PyPI so testing against a mixture of the latest release (when importing from scenario) and the local code (everything else). It probably needs to be pip install . done in both places instead, maybe with a specific test for pip install ops[testing]. I hadn't thought about this - I'll look into it.

:+1:

testing/README.md
```diff
@@ -304,7 +304,7 @@ def test_relation_data():
         ),
     }
 
-# which is very idiomatic and superbly explicit. Noice.
+# which is very idiomatic and superbly explicit.
```
james-garner-canonical, line 307: This is funny

tonyandrewmeyer: It made me sad to remove it, but codespell complained and it seemed silly to keep it and put in a codespell bypass just for this.

testing/src/context.py
```diff
@@ -674,7 +674,7 @@ def run(self, event: "_Event", state: "State") -> "State":
             if self._action_failure_message is not None:
                 raise ActionFailed(
                     self._action_failure_message,
-                    state=self._output_state,
+                    state=self._output_state,  # type: ignore
```
james-garner-canonical, line 677: I guess it isn't worth the pain of bringing in `ops._private.harness.ActionOutput` here and `cast`ing or trying to use it instead of the `"State"` annotation for `_output_state`?

james-garner-canonical:
Thank you for a very clean series of commits, @tonyandrewmeyer.

The consistency checker fix seems simple enough to me that it would be fine to solve it in this PR even if it's not strictly needed to get everything that's currently working in scenario working here, especially since it's the last commit, so doesn't interfere with understanding the others.

More for my own understanding than anything, are these all the follow up issues coming out of this PR?
* finish adjusting the CI workflow for publishing ops-scenario - this needs to switch to Trusted Publisher and maybe some other tweaks -- high priority, as this is needed before next release?
* investigate puzzling pytest raises issue -- lower priority, since the test will be disabled for now
* type check the ops-scenario tests and reduce the use of `Any` and `# type: ignore` in ops-scenario -- lower priority, nice to have

I'll wait to hear back on the situation with tox potentially installing scenario from PyPI in testing. I wasn't clear on explanation of the `scenario` imports with regards to this, should they work if scenario isn't being installed from PyPI, or should they become relative imports?

tonyandrewmeyer:
[Ben]
> If there were passwords, they'd been in the public ops-scenario repo already. :-)

Yes, and if there was a "get rid of the password I committed earlier" commit I assume they would have been invalidated too. But security scanners tend to look through a repo's entire history and in my experience it's a pain to have to explain that a problem was already solved.

[James]
> More for my own understanding than anything, are these all the follow up issues coming out of this PR?
> 
> * finish adjusting the CI workflow for publishing ops-scenario - this needs to switch to Trusted Publisher and maybe some other tweaks -- high priority, as this is needed before next release?

Yes. It could be made to work by just adjusting the working directory and putting an appropriate secret into our repo, but I would rather go straight to Trusted Publisher and we also need to decide if ops-scenario will still publish automatically whenever the version is increased or if it moves to a manual process like ops.

> * investigate puzzling pytest raises issue -- lower priority, since the test will be disabled for now

Yes. I am super baffled by this. I'm hoping that if I come back to it in a few days it'll be really obvious.

> * type check the ops-scenario tests and reduce the use of `Any` and `# type: ignore` in ops-scenario -- lower priority, nice to have

Yes. I suspect there's a bunch of type annotation improvements that could be made in general - I don't think that was a focus for ops-scenario in general.

> I'll wait to hear back on the situation with tox potentially installing scenario from PyPI in testing. I wasn't clear on explanation of the `scenario` imports with regards to this, should they work if scenario isn't being installed from PyPI, or should they become relative imports?

I believe if I do the `pip install .` for both packages then they don't need to be relative, because everything will be installed just like as if from PyPI. But probably relative imports are the more correct way to go in general? We don't do that in ops either, though.

I think that is everything that directly arises from this PR. There are some other things that should be done (e.g. using the `_Manager` and `_JujuContext` classes that ops recently added, removing the compatibility for older versions of ops, but I think they all have separate tickets.

testing/src/context.py
tonyandrewmeyer: It is bringing in `ops._private.harness.ActionOutput`. pyright was complaining about `state` being a `State` rather than `State|None` (there's an assert just above this that forces that), which I didn't really understand since it seems like it should be fine to be *more* specific. I basically punted on this one, figuring that there needed to be a bunch of "decease type:ignore usage" work later anyway and it could be figured out then, since it doesn't really gain much to have this be checked.

james-garner-canonical: Because of the following lines, and the fact that you are already asserting that it's not `None`, I assumed that the type checker was complaining about it being a `State` rather than a `ActionOutput`, which is the type that `ActionFailed` expects, but I haven't looked deeper, so perhaps I'm wrong. And probably not worth the pain either way.

```
    from scenario.state import (
        ...
        State,
        ...
    )
```

```python
        self._output_state: Optional["State"] = None
```

james-garner-canonical:
Thanks, @tonyandrewmeyer. I don't have a strong opinion on relative imports, just wondering about the resolution for this:

> I think maybe it's pulling ops-scenario from PyPI so testing against a mixture of the latest release (when importing from scenario) and the local code (everything else).

If this is working correctly with `scenario` imports, then all good. Happy to approve if/when this is the case in current tests (is it already?).

benhoyt (approved):
Thanks for your work on this. Let's get final approval from @PietroPasotti as well.

testing/src/context.py
tonyandrewmeyer: `ActionFailed` is expecting a `State` (it expects an `ActionOutput|None` for the `output` argument, which defaults to `None`).

If I don't have the `type: ignore` then this is what I get:

```
/tmp/operator/testing/src/context.py
  /tmp/operator/testing/src/context.py:678:27 - error: Argument of type "State" cannot be assigned to parameter "state" of type "State | None" in function "__init__"
    Type "State" is incompatible with type "State | None"
      "scenario.state.State" is incompatible with "_private.harness.State"
      "State" is incompatible with "None" (reportArgumentType)
```

I think maybe this is because in `_private/harness` I import `State` from `ops.testing` if possible and if not (to handle people only installing `ops` and not `ops[testing]`) then I define a placeholder `State` class. This is not pretty but I wasn't sure how else I could have it not complain if ops-scenario wasn't installed (other than an ignore or an `Any`). I think maybe the error now is because it's still using that placeholder class, even though it should be able to get `State` from `ops.testing` (getting it from `scenario` doesn't help).

I'll leave it for someone else (including potentially future-me) to figure out.

james-garner-canonical: Oh my bad, this is what I get for looking at code on github instead of in vscode. Thanks for the explanation!

tonyandrewmeyer:
> If this is working correctly with `scenario` imports, then all good. Happy to approve if/when this is the case in current tests (is it already?).

It is the case now - I think it was mixing local and PyPI previously (a few small extra things to fix showed up afterwards also). It's not perfect now, because of the way that Tox doesn't know to regenerate an environment if it's a local install but it will work correctly and is ok for now I think.

tox.ini
```diff
@@ -166,7 +184,11 @@ allowlist_externals =
     mkdir
     cp
 deps =
-    -e .[testing]
+    # TODO: If the code in testing changes, it isn't reinstalled.
+    # -e doesn't install it with the right name. We should probably
+    # build a wheel and then use that across the environments,
+    # similar to what was done in the ops-scenario repository.
+    ./testing
```
james-garner-canonical, line 191: Is this set of deps intentionally missing `-e .`?

tonyandrewmeyer: Yes, see the comment above it :wink:.

tonyandrewmeyer: Oh wait, that wasn't "I didn't read the comment", that was "-e .", sorry!

james-garner-canonical: Yeah I just noticed that this one was different from the others where you made similar changes. This one is missing the `-e .` line that the others replaced `-e .[testing]` with.

tonyandrewmeyer: And the answer is yes, because it's only testing the `testing/README.md` file. Although thinking about it more, I suppose it ought to have the latest ops code as well, so I'll add it in.

I think we'll probably end up shrinking `testing/README.md` down with the examples there moving into how-tos (some has already) and the API reference docs (also) and so on, and then can remove this tox environment. It was super handy while I was doing the 7.0 work to avoid forgetting to update the README or doing it incorrectly, even though it's limited in what it checks.

james-garner-canonical (approved):

tox.ini
james-garner-canonical: Makes sense! Code snippets in readmes are nice, and making sure they run correctly is nice too.

testing/src/_consistency_checker.py
```diff
@@ -476,7 +478,7 @@ def check_secrets_consistency(
         return Results(errors, [])
 
     assert event.secret is not None
-    if event.secret not in state.secrets:
+    if event.secret.id not in {s.id for s in state.secrets}:
         secret_key = event.secret.id if event.secret.id else event.secret.label
```
PietroPasotti, line 481: if there is a chance that secret.id could be None (seems implied by this line) then we have to be careful about putting them in a set comprehension? or perhaps this line is old code that can now be removed

testing/README.md
PietroPasotti: hahaha I think I added that during an early live demo, too bad it has to go :)

PietroPasotti (approved):
Overall it looks great! Thanks for this humongous work :)

testing/src/_consistency_checker.py
```diff
@@ -309,7 +309,9 @@ def _check_storage_event(
             f"storage event {event.name} refers to storage {storage.name} "
             f"which is not declared in the charm metadata (metadata.yaml) under 'storage'.",
         )
-    elif storage not in state.storages:
+    elif (storage.name, storage.index) not in {
+        (s.name, s.index) for s in state.storages
+    }:
         errors.append(
             f"cannot emit {event.name} because storage {storage.name} "
```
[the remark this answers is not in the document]
tonyandrewmeyer: Good point. I'll do a follow-up PR to do that and add tests for this case, rather than making this one even larger.

testing/src/_consistency_checker.py
tonyandrewmeyer: I'll check on this, but I believe the event's secret should always have an ID and we seem to type it that way. Could indeed be older code.

tonyandrewmeyer:
@benhoyt would you be able to merge this please? You need to enable rebase & merge, do it, then turn that off again, for this special-case. Thanks!

benhoyt:
@tonyandrewmeyer I've enabled Rebase and Merge temporarily -- go ahead and rebase+merge this one (I'll let you do the honours!), and then I'll disable that setting again.

PR #1407, opened 2024-10-02
ci: configure the labels for dependabot PRs

tonyandrewmeyer:
For GitHub actions dependabot PRs: only use the `dependencies` label, rather than both that and `github_actions`.

For pip (Python) dependabot PRs: we only have security dependabot PRs enabled, so add the same label restriction (avoiding a `python` label) but this requires adding a section to configure dependabot, and then essentially disabling most of it by setting the maximum PRs to 0 (security PRs will still be opened).

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,3 +4,12 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"  # This is hard-coded to the 1st of the month
+    labels:
+      - "dependencies"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 0  # Security updates only
+    labels:
+      - "dependencies"


james-garner-canonical:
I know you said it should be straightforward, but I'm a little confused -- good thing you didn't request a review from me ;)

I do want to ask though -- is this changing the schedule for python security dependency updates from monthly to daily? Knowing earlier about security updates does seem like a good thing.

It also seems odd that just changing the labels requires adding a new `package-ecosystem` section. This is probably just my ignorance about dependabot.

tonyandrewmeyer:
> I do want to ask though -- is this changing the schedule for python security dependency updates from monthly to daily? Knowing earlier about security updates does seem like a good thing.

No. [Dependabot security updates](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates) are mostly separate from regular dependabot. They're mostly configured elsewhere and have a schedule managed by GitHub that can't be adjusted.

I'm not certain, but I think this slightly messy situation exists because the security updates system existed in GitHub before they acquired dependabot so it's two independently built systems that have been mushed together.

> It also seems odd that just changing the labels requires adding a new `package-ecosystem` section. This is probably just my ignorance about dependabot.

It's because the security updates can't be configured to do different things with labels, but [do inherit what's in the 'regular' dependabot config](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#impact-of-configuration-changes-on-security-updates), including [labels](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#labels). [The limit on PRs does not impact security](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit), so it's how you have the config for the 'regular' updates so the joint options are picked up without actually turning the 'regular' updates on. I believe the docs explicitly recommend this somewhere but couldn't find it with a quick search.

james-garner-canonical:
Thanks for the explanation, @tonyandrewmeyer, super helpful! I guess with security updates being managed separately from the dependabot schedule, and `open-pull-requests-limit: 0`, `schedule: interval: "daily"` has no true effect here, but is needed since it's a required field.

dimaqq (approved):

tonyandrewmeyer:
> I guess with security updates being managed separately from the dependabot schedule, and `open-pull-requests-limit: 0`, `schedule: interval: "daily"` has no true effect here, but is needed since it's a required field.

Yes, exactly.

PR #1410, opened 2024-10-04
docs: set the READTHEDOCS context variable

### LINKED ISSUE #1393: Update docs to handle readthedocs addons change (by October 7, 2024)
[Instructions available from rtd](https://about.readthedocs.com/blog/2024/07/addons-by-default/).

I took a quick look at our config and it seems like we don't have the snippets they're suggesting, so I assume we need to do this.

tonyandrewmeyer:
As of October 7th, the `html_context` dictionary is meant to contain a `READTHEDOCS` key set to True if that's in the environment variables, so this change does that.

We may see small changes in the visual look of the docs but everything should keep working with this small adjustment.

[Instructions from Read the Docs](https://about.readthedocs.com/blog/2024/07/addons-by-default/)

Fixes #1393

### DIFF
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -139,8 +139,8 @@ html_context = {
     'sequential_nav': 'none',
 }
 # Addons-by-default, see: https://about.readthedocs.com/blog/2024/07/addons-by-default/
-if os.environ.get("READTHEDOCS", "") == "True":
-    html_context["READTHEDOCS"] = True
+if os.environ.get('READTHEDOCS', '') == 'True':
+    html_context['READTHEDOCS'] = True
 
 # If your project is on documentation.ubuntu.com, specify the project
 # slug (for example, "lxd") here.
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import datetime
+import os
 import pathlib
 import sys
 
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
I read the linked page on read the docs and had a quick look at our `custom_conf.py`.

I'm guessing we need this change that Read the Docs suggests in "the most common changes you may need to do in your Sphinx's configuration file" (adding the `READTHEDOCS` environment variable to our `html_context` dict) because Spinx uses Jinja internally -- the change is commented in the article with the note "Tell Jinja2 templates the build is running on Read the Docs"?

benhoyt (approved):
Thanks!

tonyandrewmeyer:
> I'm guessing we need this change that Read the Docs suggests in "the most common changes you may need to do in your Sphinx's configuration file" (adding the `READTHEDOCS` environment variable to our `html_context` dict) because Spinx uses Jinja internally -- the change is commented in the article with the note "Tell Jinja2 templates the build is running on Read the Docs"?

Yes, I think so. My understanding is that rtd does some post processing (for the PR banner for example) that's moving to addons and so we need to have this signal there. I couldn't find anything more specific than what's in the linked post (although I admittedly didn't look for ages), and it's tricky to dig into since I'm expecting to see no changes and that could also be from doing something that's ignored...

james-garner-canonical (approved):
Thanks for the info, Tony. The change makes sense to me, and I don't think there's any harm in including the environment variable in the html_context if it's not used. At the very least it should future proof us to some extent if we lean on more rtd features in future.

PR #1412, opened 2024-10-08
chore: Python 3.13 has been released, remove allow-prerelease

dimaqq:
Python 3.13 has been released yesterday.
Let's enable it in our CI.

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.10", "3.12", "3.13"]
+        python-version: ["3.8", "3.10", "3.12", "3.13", "3.14.0a1"]
         exclude:
         - {python-version: "3.8", os: "macos-latest"}  # macos-14 is arm64, and there's no Python 3.8 build for arm64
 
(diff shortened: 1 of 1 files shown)


.github/workflows/framework-tests.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.10", "3.12"]
+        python-version: ["3.8", "3.10", "3.12", "3.13"]
```
benhoyt, line 46: I'm open to having a new discussion if you want, but on the [last PR](https://github.com/canonical/operator/pull/1315) that updated this we decided to run the tests just on the LTS versions (3.8, 3.10, 3.12), but do the `pip install` on 3.13 (which it already is, but we could remove the pre-release).

I think we should keep "test" and "test-real-pebble" on just 3.8/3.10/3.12, but update "lint" to 3.13 (it's currently still on 3.11).

That said, it might be a few days premature -- looks like the GitHub Action can't find Python 3.13 yet, hence the failed runs.

dimaqq:
I'm testing 3.14 alpha just for fun and will update the PR after.

dimaqq:
`["3.8", "3.10", "3.12", "3.13", "3.14.0-alpha.1"]` all passed.
reducing as suggested in the review.

benhoyt (approved):

IronCore864 (approved):

PR #1414, opened 2024-10-08
docs: fix rtd ad placement

IronCore864:
Fixing the issue where the ad doesn't show up on the left navigation bar even if there is empty space for it.

See the issue here: https://github.com/canonical/operator/issues/1393, and the related PR: https://github.com/canonical/operator/pull/1410.

Preview: https://ops--1414.org.readthedocs.build/en/1414/.

### DIFF
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -139,8 +139,15 @@ html_context = {
     'sequential_nav': 'none',
 }
 # Addons-by-default, see: https://about.readthedocs.com/blog/2024/07/addons-by-default/
-# if os.environ.get('READTHEDOCS', '') == 'True':
-#     html_context['READTHEDOCS'] = True
+if os.environ.get('READTHEDOCS', '') == 'True':
+    html_context['READTHEDOCS'] = True
+    html_context["current_version"] = "latest"
+    html_context["conf_py_path"] = "/docs/"
+    html_context["display_github"] = True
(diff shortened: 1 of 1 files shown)


dimaqq (approved):

docs/custom_conf.py
```diff
@@ -141,6 +141,14 @@
 # Addons-by-default, see: https://about.readthedocs.com/blog/2024/07/addons-by-default/
 if os.environ.get('READTHEDOCS', '') == 'True':
     html_context['READTHEDOCS'] = True
+    # The following are needed, see: https://github.com/pradyunsg/furo/blob/main/docs/conf.py#L135.
+    html_context['current_version'] = 'latest'
```
tonyandrewmeyer, line 145: I'm not certain about this. We never used to see a banner warning about "latest" and I don't think we want one. So maybe we want "stable" to be the current version?

Or would this remove that banner because it's telling rtd that "latest" is (essentially) "stable"? If so then it would be great indeed.

tonyandrewmeyer (approved):
Thanks for this!

docs/custom_conf.py
dimaqq: my 2c: let's merge it and see

benhoyt: My main concern: I don't want all our "/latest" docs URLs to break or go stale.

PR #1415, opened 2024-10-10
ci: disable the automatic ops-scenario releasing

tonyandrewmeyer:
This hasn't been fully updated to handle the location in the merged repo, but has been enough to wrongly release an `ops.version.version` release (and tag) on merge to main.

Disable this for now so that we don't accidentally have another `ops.version.version` release. I'll do a follow-up PR that properly implements ops-scenario releasing alongside ops releases.

### DIFF
--- a/.github/workflows/build_scenario_wheels.yaml
+++ b/.github/workflows/build_scenario_wheels.yaml
@@ -1,9 +1,10 @@
 name: Build ops-scenario wheels
 
 on:
-  push:
-    branches:
-      - main
+  workflow_dispatch
+#  push:
+#    branches:
+#      - main
 
 jobs:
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Fine with me. Could you also add a TODO comment to make it easier to find?

benhoyt:
I'm going to merge this without further review to avoid another `ops.version.version` release being created.

tonyandrewmeyer:
> I'm going to merge this without further review to avoid another `ops.version.version` release being created.

Thanks! I deleted the one (release and tag) that I triggered when merging the doc PR. In the few short hours since we did the repo merge I forgot that it would happen :disappointed:.

PR #1416, opened 2024-10-10
docs: clarify where StoredState is stored, and the upgrade behaviour

### LINKED ISSUE #1270: Documentation is missing for StoredState behavior
Documentation is missing for StoredState behavior, in particular across upgrades.
In charm code we import StoredState so seems reasonable to have it documented in the docstring there.

- For vm charms, there are no docs. What happens to stored state across upgrades? Does `use_juju_for_storage` have an impact?
- For k8s charms, info is sparse and fragmented, so need to parse the code every time just to be sure.

https://github.com/canonical/operator/blob/89ea7f4d56173ea314d59a271e4304b18cd129ed/ops/framework.py#L1037-L1038

https://github.com/canonical/operator/blob/89ea7f4d56173ea314d59a271e4304b18cd129ed/ops/main.py#L354-L355

https://github.com/canonical/operator/blob/89ea7f4d56173ea314d59a271e4304b18cd129ed/ops/main.py#L540-L543

tonyandrewmeyer:
Adjustments to the `StoredState` documentation:

* For `StoredDict`, `StoredList`, and `StoredSet`, note that charms would not normally create these directly and would use `StoredState` instead.
* Explain where the data is stored (there is one case that it doesn't mention, where there is already a unit-state DB locally, but I think that complicates things too much to specify here).
* Explain what happens when the charm is upgraded and recommend using a peer relation over stored state for K8s charms that want data preserved across upgrades.

As a related drive-by: remove the `jujuremoved` tag for `use_juju_for_storage`, since this won't actually be going away with Juju 4.0.

Fixes #1270

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -1132,14 +1132,9 @@ class BoundStoredState:
 class StoredState:
     """A class used to store data the charm needs, persisted across invocations.
 
-    Example::
-
-        class MyClass(ops.Object):
-            _stored = ops.StoredState()
-
     Instances of ``MyClass`` can transparently save state between invocations by
     setting attributes on ``_stored``. Initial state should be set with
-    ``set_default`` on the bound object, that is::
+    ``set_default`` on the bound object; for example::
 
         class MyClass(ops.Object):
             _stored = ops.StoredState()
@@ -1152,6 +1147,16 @@ class StoredState:
             def _on_seen(self, event):
                 self._stored.seen.add(event.uuid)
 
+    Data is stored alongside the charm (in the charm container for Kubernetes
+    sidecar charms, and on the machine for machine charms), except for podspec
+    charms and charms explicitly passing `True` for `use_juju_for_storage` when
+    running :meth:`ops.main` (in those cases, the data is stored in Juju).
+
+    For machine charms, charms are upgraded in-place on the machine, so the data
+    is preserved. For Kubernetes sidecar charms, when the charm is upgraded, the
+    pod is replaced, so any data is lost. When data should be preserved across
+    upgrades, Kubernetes sidecar charms should use a peer-relation for the data
+    instead of `StoredState`.
     """
 
     def __init__(self):
@@ -1245,7 +1250,12 @@ def _wrapped_repr(obj: '_StoredObject') -> str:
 
 
 class StoredDict(typing.MutableMapping[Hashable, Any]):
-    """A dict-like object that uses the StoredState as backend."""
+    """A dict-like object that uses the StoredState as backend.
+
+    Charms are not expected to use this class directly. Adding a
+    :class:`StoredState` attribute to a charm class will automatically use this
+    class to store dictionary-like data.
+    """
 
     def __init__(self, stored_data: StoredStateData, under: Dict[Hashable, Any]):
         self._stored_data = stored_data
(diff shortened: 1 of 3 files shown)


ops/framework.py
```diff
@@ -1245,7 +1250,12 @@ def _wrapped_repr(obj: '_StoredObject') -> str:
 
 
 class StoredDict(typing.MutableMapping[Hashable, Any]):
-    """A dict-like object that uses the StoredState as backend."""
+    """A dict-like object that uses the StoredState as backend.
+
+    Charms are not expected to use this class directly. Adding a
+    :class:`StoredState` attribute to a charm class will automatically use this
+    class to store dictionary-like data.
```
benhoyt, line 1257: How does Ops know whether to create a StoredDict, StoredList, or StoredSet when you create/access a `_stored` attribute?

tonyandrewmeyer: When you set the attribute [it gets wrapped](https://github.com/canonical/operator/blob/51ba92f06b7b5f2e87c32aa67fb411b583be47ea/ops/framework.py#L1109) and similarly unwrapped when [accessed](https://github.com/canonical/operator/blob/51ba92f06b7b5f2e87c32aa67fb411b583be47ea/ops/framework.py#L1123).

It does an `isinstance` check so I guess the doc should say "dictionary data" rather than 
dictionary-like" even though the `StoredDict` class is more generic.

benhoyt (approved):

ops/framework.py
```diff
@@ -1152,6 +1147,16 @@ def __init__(self, parent, key):
...
+    Data is stored alongside the charm (in the charm container for Kubernetes
+    sidecar charms, and on the machine for machine charms), except for podspec
+    charms and charms explicitly passing `True` for `use_juju_for_storage` when
+    running :meth:`ops.main` (in those cases, the data is stored in Juju).
+
+    For machine charms, charms are upgraded in-place on the machine, so the data
+    is preserved. For Kubernetes sidecar charms, when the charm is upgraded, the
+    pod is replaced, so any data is lost. When data should be preserved across
+    upgrades, Kubernetes sidecar charms should use a peer-relation for the data
+    instead of `StoredState`.
```
james-garner-canonical, line 1159: Random ideas -- I know that `use_juju_for_storage` isn't actually going away with 4.0, but do we want to continue to steer people away from it?

We could mention podspec charms going away in future? e.g.

"except for podspec  charms (deprecated in favour of sidecare charms), where the data is stored in Juju."

And maybe move the bit about `use_juju_for_storage` to *after* the good recommendation in the last paragraph for preserving data across upgrades? e.g. as a final sentence:

"To explicitly opt into Juju data storage, set `use_juju_for_storage` to `True` when running :meth:`ops.main` (but this is not recommended)."

james-garner-canonical (approved):

ops/framework.py
james-garner-canonical: podspec charms are deprecated, aren't they? Second guessing myself ..

tonyandrewmeyer: > podspec charms are deprecated, aren't they? Second guessing myself ..

Yes :). You have to find that info [in the Juju docs, though](https://juju.is/docs/sdk/charm-taxonomy#heading--podspec).

tonyandrewmeyer: I've adjusted it to mention podspec and `use_juju_for_storage` being deprecated - I agree that's worth doing.

I don't think the `use_juju_for_storage` mention should be at the end, though - even with a "this is not recommended" I think it promotes it too much saying that it's a way to opt in. We don't want people doing this (and they'll get a warning if they do) - it's mentioned here so people that are doing it know that it is a consequence, but if they want the storage to be in Juju then they should use a peer relation.

james-garner-canonical: Makes sense!

PR #1417, opened 2024-10-10
fix: raise ModelError on unknown/error status set in Scenario

tonyandrewmeyer:
This copies across the fix from Harness to also be present in Scenario: if `status_set` is passed anything other than the valid statuses, then `ModelError` is raised (to better mimic what the Juju hook would actually do).

Closes https://github.com/canonical/ops-scenario/issues/202 (rather than copying it over to this repo).

### DIFF
--- a/testing/src/mocking.py
+++ b/testing/src/mocking.py
@@ -368,15 +368,15 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
 
     def status_set(
         self,
-        status: _RawStatusLiteral,
+        status: _SettableStatusName,
         message: str = "",
         *,
         is_app: bool = False,
     ):
-        if status not in get_args(_SettableStatusName):
+        valid_names = get_args(_SettableStatusName)
+        if status not in valid_names:
             raise ModelError(
-                f'ERROR invalid status "{status}", expected one of '
-                f"[maintenance blocked waiting active]"
+                f'ERROR invalid status "{status}", expected one of [{", ".join(valid_names)}]',
             )
         self._context._record_status(self._state, is_app)
         status_obj = _EntityStatus.from_status_name(status, message)
--- a/testing/src/mocking.py
+++ b/testing/src/mocking.py
@@ -62,7 +62,6 @@ from scenario.state import (
     _EntityStatus,
     _port_cls_by_protocol,
     _RawPortProtocolLiteral,
-    _RawStatusLiteral,
 )
 
 if TYPE_CHECKING:  # pragma: no cover
--- a/testing/src/mocking.py
+++ b/testing/src/mocking.py
@@ -376,7 +376,7 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
         if status not in get_args(_SettableStatusName):
             raise ModelError(
                 f'ERROR invalid status "{status}", expected one of '
-                f'[maintenance blocked waiting active]'
+                f"[maintenance blocked waiting active]"
             )
         self._context._record_status(self._state, is_app)
         status_obj = _EntityStatus.from_status_name(status, message)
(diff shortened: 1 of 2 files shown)


testing/src/mocking.py
```diff
@@ -371,6 +373,11 @@ def status_set(
         *,
         is_app: bool = False,
     ):
+        if status not in get_args(_SettableStatusName):
+            raise ModelError(
+                f'ERROR invalid status "{status}", expected one of '
+                f"[maintenance blocked waiting active]"
```
benhoyt, line 379: Nit:

```suggestion
                f'[maintenance blocked waiting active]'
```

It's weird that this passes Ruff-ing, given our quote style.

testing/tests/test_e2e/test_status.py
```diff
@@ -169,3 +170,48 @@ def test_status_comparison(status):
...
+@pytest.mark.parametrize(
+    "status",
+    (
+        ActiveStatus("foo"),
+        WaitingStatus("bar"),
+        BlockedStatus("baz"),
+        MaintenanceStatus("qux"),
+    ),
+)
+def set_status_success(status: ops.StatusBase):
```
benhoyt, line 184:
```suggestion
def test_status_success(status: ops.StatusBase):
```

benhoyt (approved):
Couple of nit comments, but approving in advance.

testing/src/mocking.py
tonyandrewmeyer: I actually had to change it *to* this, having the single quotes first. I opened https://github.com/canonical/operator/issues/1419 as a result :)

Basically: Scenario used double-quotes and so all the code in `testing/` is expected to follow that pattern (and ruff is configured to know that, in the `testing/pyproject.toml`.

I think it's better to be consistent now (rather than putting in an ignore here) and then do the change in a big bang commit.

testing/tests/test_e2e/test_status.py
tonyandrewmeyer: Oops :blush:

testing/src/mocking.py
dimaqq: minor: a bit weirded out by the 2nd string being an f-string without interpolation.
I wonder if the allowed set can be computed rather than hard-coded, making the entire string shorter, fitting on one line, and making this discussion a moot point.

dimaqq:
Looks good!

testing/src/mocking.py
james-garner-canonical: I remember when I used pylint it would complain about f-strings without interpolation, but I guess ruff doesn't mind or we have it configured not to?

tonyandrewmeyer: See the [long discussion upstream](https://github.com/astral-sh/ruff/pull/10886) for details, but basically if it's a concatenated string then f is allowed if it's used in any of the parts, because you get nicer (subjectively) layout where the string starts match.

Dima's suggestion to avoid it rather than going one way sounds good to me. I'll look into that.

testing/src/mocking.py
```diff
@@ -371,6 +373,11 @@ def status_set(
         *,
         is_app: bool = False,
     ):
+        if status not in get_args(_SettableStatusName):
```
james-garner-canonical, line 376: Worth changing the type of the `status` argument as well?

james-garner-canonical (approved):
Looks good to me!

testing/src/mocking.py
james-garner-canonical: >if it's a concatenated string then f is allowed if it's used in any of the parts

Nice, that's what I always wanted when pylint complained

PR #1418, opened 2024-10-10
chore: remove compatibility code with older ops

tonyandrewmeyer:
As part of the merge of the `ops-scenario` repo into this one, `ops-scenario` now requires ops 2.17+. This has the added benefit that we can get rid of the compatibility code that supported ops 2.15 and ops 2.15.

With the pyright "detect ignores that aren't needed" check enabled again, some additional ignores can also be removed.

Closes https://github.com/canonical/ops-scenario/issues/192 (rather than copying it into this repo).

### DIFF
--- a/testing/src/context.py
+++ b/testing/src/context.py
@@ -43,7 +43,6 @@ from scenario.state import (
 
 if TYPE_CHECKING:  # pragma: no cover
     from ops._private.harness import ExecArgs
-
     from scenario.ops_main_mock import Ops
     from scenario.state import (
         AnyJson,
--- a/testing/src/context.py
+++ b/testing/src/context.py
@@ -41,14 +41,8 @@ from scenario.state import (
     _Event,
 )
 
-# TODO: Remove the line below after the ops compatibility code is removed.
-# pyright: reportUnnecessaryTypeIgnoreComment=false
-
 if TYPE_CHECKING:  # pragma: no cover
-    try:
-        from ops._private.harness import ExecArgs  # type: ignore
-    except ImportError:
-        from ops.testing import ExecArgs  # type: ignore
+    from ops._private.harness import ExecArgs
 
     from scenario.ops_main_mock import Ops
     from scenario.state import (
--- a/testing/src/mocking.py
+++ b/testing/src/mocking.py
@@ -24,12 +24,7 @@ from typing import (
 
 from ops import JujuVersion, pebble
 
-# TODO: Remove the line below after the ops compatibility code is removed.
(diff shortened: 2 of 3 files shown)


testing/src/mocking.py
```diff
@@ -24,12 +24,7 @@
...
 
-# TODO: Remove the line below after the ops compatibility code is removed.
-# pyright: reportUnnecessaryTypeIgnoreComment=false
-try:
-    from ops._private.harness import ExecArgs, _TestingPebbleClient  # type: ignore
-except ImportError:
-    from ops.testing import ExecArgs, _TestingPebbleClient  # type: ignore
+from ops._private.harness import ExecArgs, _TestingPebbleClient
 
 from ops.model import CloudSpec as CloudSpec_Ops
```
james-garner-canonical, line 29: How about this one?

testing/src/context.py
```diff
@@ -41,14 +41,8 @@
...
-# pyright: reportUnnecessaryTypeIgnoreComment=false
-
 if TYPE_CHECKING:  # pragma: no cover
-    try:
-        from ops._private.harness import ExecArgs  # type: ignore
-    except ImportError:
-        from ops.testing import ExecArgs  # type: ignore
+    from ops._private.harness import ExecArgs
 
     from scenario.ops_main_mock import Ops
```
james-garner-canonical, line 47: Do we want this blank line?

james-garner-canonical (approved):
Looks good to me, must feel good to remove all those conditional imports and `type: ignore`s. Just a couple of (very) minor formatting questions.

testing/src/context.py
tonyandrewmeyer: This one should be here, because we consider `ops` to be third-party modules and `scenario` is first-party, so there's a gap between them (and ruff will put it back if you take it out).

I guess now that we've put the code in the same repo, there's an argument that both ops and scenario should be considered the same level in this regard, but if we do that it would be better as a separate PR that makes the change everywhere. Since we still release two packages, I'm also not entirely convinced we should do that.

testing/src/mocking.py
tonyandrewmeyer: Probably not; removed.

testing/src/context.py
james-garner-canonical: Yeah I wasn't sure about this one. I don't think I've really knowingly used subpackages like this before, so I'm not really clear on what the relationship between them should be now.

dimaqq (approved):

PR #1431, opened 2024-10-10
refactor: fix the testing src-layout structure and use relative imports

tonyandrewmeyer:
The original commit merging ops-scenario into this repository had an error in the [src-layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) structure: the `src` folder should contain a folder for each of the importable packages, not have all of the individual modules.

This PR moves everything in `testing/src/` to `testing/src/scenario/` to fix this. This means that the build tool can find the package without any help, and makes editable installs work correctly, which removes a quirk in the tox dependency installs. It also makes it simpler to work on the project, since the `ops` and `scenario` packages are available in the expected format with the expected name (although src-layout encourages editable installs for development).

For the second part of the cleanup, the PR changes the `testing/src/scenario` imports to be relative for everything inside of that location. While adjusting these, most of the imports are adjusted to import from the top-level `ops` namespace rather than from individual modules (except for non-exported names).


benhoyt:
Oh, this is a pity, as it means the code is (even) more nested. I guess it can't be just `testing/*.py` or `scenario/*.py`, because they're two different PyPI packages and need their own pyproject.toml?

tonyandrewmeyer:
> Oh, this is a pity, as it means the code is (even) more nested. I guess it can't be just `testing/*.py` or `scenario/*.py`, because they're two different PyPI packages and need their own pyproject.toml?

We could go to a flat layout, which would remove the `src` layer (so instead of moving everything to `src/scenario` it would be renaming `src` to `scenario`). My understanding is that src-layout is considered best practice these days, but I don't have strong feelings about it.

benhoyt (approved):

tox.ini
```diff
@@ -105,11 +101,7 @@ deps =
     typing_extensions~=4.2
     jsonpatch~=1.33
     -e .
-    # TODO: If the code in testing changes, it isn't reinstalled.
-    # -e doesn't install it with the right name. We should probably
-    # build a wheel and then use that across the environments,
-    # similar to what was done in the ops-scenario repository.
-    ./testing
+    -e ./testing/.
```
dimaqq, line 104: `./testing/.` looks funky, but I wonder if this notation is required.

.gitignore
```diff
@@ -22,6 +22,7 @@ coverage.xml
 /build
 /docs/_build
 /testing/ops_scenario.egg-info/
+/testing/src/ops_scenario.egg-info/
```
dimaqq, line 25: WDYT about making this simpler?

```suggestion
ops_scenario.egg-info
```

dimaqq (approved):
getting rid of "scenario" all over the source code is a win :)

tox.ini
tonyandrewmeyer: > `./testing/.` looks funky, but I wonder if this notation is required.

It can't be `testing` (because that would look for a package on PyPI called "testing"). `./testing` and `testing/.` both work for me, at least with `uv pip`. I don't have a strong preference here - would you prefer either of those?

tonyandrewmeyer: Actually, now that `-e` is there, `testing` would work because the `-e` forces it to be a local reference since you can't have an editable PyPI install.

So it could be `-e testing` if you prefer that? There's a small risk that someone drops the `-e` or similar and causes problems in the future, but I think `testing` is a restricted PyPI name anyway, so people can't actually register something bad under it.

PR #1432, opened 2024-10-11
ci: adjust the release process to handle publishing ops and ops-scenario

tonyandrewmeyer:
Adds new publish workflows for ops-scenario (one for main PyPI, one for test.pypi.org), modelled on the ops workflows. These differ significantly from the old ops-scenario publish workflow (now deleted), which would automatically create a GitHub release and tag and publish to PyPI (with API credentials) on a push to main.

To avoid confusion, the publish workflows for `ops` are renamed (and while this is being done, the extension is changed from `yml` to `yaml` to match all the other files).

The publish workflows are trigged by new tags - `2.*` for `ops` (rather than a release being published) and `scenario-7.*` for `ops-scenario`. This allows releases that only include publishing one of the packages (for example, in a month that had no changes to one, or when a patch release is required).

The HACKING.md documentation is updated to reflect the changes in the release process.

A CHANGES.md is added for Scenario, backdated to 7.0.0 (although 7.0.0 is not entirely complete, but it does list the main changes).

Once this is merged, trusted publishing will need to be configured for the `ops-scenario` package, and the config for `ops` will need to be adjusted to refer to the new workflow filename.

Drive-bys:
 * The smoke test workflow had incorrect indentation for the `schedule` trigger.
 * Clarify when `(testing)` and `(harness)` scopes are used in conventional commits, now that we have Scenario in the repo.
 * Note the use of the various `.. juju` directives in documentation.
 * Minor clarifications in the release process based on comments from @james-garner-canonical when he did his first release.


HACKING.md
```diff
@@ -311,43 +322,61 @@ To make a release of the ops library, do the following:
...
    noteworthy changes.
-12. Have someone else in the Charm-Tech team proofread the release notes.
-13. When you are ready, click "Publish". (If you are not ready, click "Save as Draft".)
-
-This will trigger an automatic build for the Python package and publish it to
-[PyPI](https://pypi.org/project/ops/)) (authorisation is handled via a
+13. Have someone else in the Charm-Tech team proofread the release notes.
+14. When you are ready, click "Publish". (If you are not ready, click "Save as Draft".)
+15. If the release includes an `ops-scenario` package, then create a new tag in
+   the form `scenario-<major>.<minor>.<patch>`.
```
james-garner-canonical, line 358: How to create a (second) new tag? I'm hoping there's a way in the github web ui, but I can't figure it out. Normally I just tag stuff via the cli, but I don't think that's what we want here since it would require being able to push directly to ops (main).

HACKING.md
```diff
@@ -311,43 +322,61 @@ To make a release of the ops library, do the following:
...
+
+This will trigger automatic builds for the Python packages and publish them to
```
james-garner-canonical, line 360: Maybe we should be clearer here about what triggers the automatic builds -- tag creation, right? If we're releasing one of ops or scenario, then the sole tag is created on publishing the release. But if we're releasing both, then the ops tag is created on publishing a release (triggering an automatic build). While the second tag is going to be created after the release (triggering the second automatic build). Is that right?

james-garner-canonical (changes requested):

HACKING.md
tonyandrewmeyer: Hmm, yes, this is a problem. I hadn't thought about that - too used to having high access to a repo. Maybe we do have to have three types of tags after all? Or maybe we simplify things by just using one tag and always doing two releases, or the release workflow tries to be clever and see if the number has changed.

I'll need to think more about this.

james-garner-canonical: My gut feeling is that just making two separate releases with `2.x.y` or `scenario-a.b.c` tags respectively would be the way to go, but if there does exist an easy way to do what you originally proposed then that would be better.

tonyandrewmeyer: Maybe that would be the cleanest way of working in the system. The disadvantages would be that there are two sets of source artefacts (but I don't think anyone uses those anyway), the person doing the release would need to do a few things twice, and the release notes wouldn't be able to refer to both releases (although the Discourse post is probably the more true release notes, and could).

dimaqq (approved):

HACKING.md
james-garner-canonical: > The disadvantages would be that there are two sets of source artefacts
A bit redundant, but totally fine and correct imo

> the person doing the release would need to do a few things twice
Biggest downside! But maybe the duplicated work isn't too much? You'd have to separate out the commits between ops and opt-testing anyway, for example.

> the release notes wouldn't be able to refer to both releases
Which release notes? In github you would still start with auto generated release notes vs correct previous release tag and separate out the (non)testing ones (for example), right? A single discourse post when releasing both would make sense though.

Just my 2c

IronCore864 (approved):

HACKING.md
tonyandrewmeyer: > Which release notes? In github you would still start with auto generated release notes vs correct previous release tag and separate out the (non)testing ones (for example), right?

Yes. I meant that say there's some significant change that impacts both ops and ops-scenario: if there was one combined GitHub release, then that could have some note about it at the top, once, and it's probably one PR, so it would be one bullet point in the list below. With two GitHub releases, you need to call it out at the top of each of them, and the bullet point is in both of the lists.

Even for a little change this seems a bit annoying: most features are going to also include corresponding changes in Scenario. For example:

* Adds support for the new workload-ready Juju event (#19790125)

This ends up being in the GitHub release for `ops` and for `ops-scenario`, because for `ops` it's the feature itself, but then there's the support for testing it in `ops-scenario`.

It's not a blocker, it just means that it's not super straightforward to filter the lists for the two releases, particularly if you didn't work on or review the PR yourself.

james-garner-canonical: Oh yeah I didn't think of that, thanks for explaining. Kind of puts me off the idea of two truly separate releases.

Probably a bad idea: when we release both, we could just create a second dummy release for testing with a link to the ops release with combined release notes (with separate sections where appropriate). So basically your originally proposed workflow, but the 'create a second tag' step is done via creating a second release for the same commit.

HACKING.md
```diff
@@ -318,65 +318,152 @@ the build frontend is [build](https://pypi.org/project/build/).
...
-10. Check if there's a `chore: update charm pins` auto-generated PR in the queue. If it looks
-   good, merge it and check that tests still pass. If needed, you can re-trigger the
-   `Update Charm Pins` workflow manually to ensure latest charms and ops get tested.
-11. Add, commit, and push, and open a PR to get the changelog and version bump
+5. Have the release create a new tag, in the form `<major>.<minor>.<patch>` for
+   `ops` and `scenario-<major>.<minor>.<patch>` for `ops-scenario`. If releasing
+   both packages, use the ops tag.
+6. Use the "Generate Release Notes" button to get a copy of the changes into the
+   notes field. The 'Release Documentation' section below details the form that
+   the release notes and changelog should take.
```
james-garner-canonical, line 338: Worth adding a note about choosing the previous release tag to compare to?

I think the rule is: find the previous release tag for each thing you're releasing, and if there is more than one (i.e. an ops only release and a scenario only release), use the older of the two.

Say you had an ops release, then a scenario release, and are now doing a joint release -- you'd want to choose the ops release to generate release notes.

But you'd need to filter out any scenario PRs from before the last scenario release, since they'd have been covered already.

HACKING.md
```diff
@@ -318,65 +318,152 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+12. Announce the release on [Discourse](https://discourse.charmhub.io/c/framework/42) and [Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com).
+
+13. Open a PR to change the version strings to the expected
+   next version, with ".dev0" appended (for example, if 3.14.1 is the next
+   expected version, use `'3.14.1.dev0'`).
+
+## Release Documentation
+
+`ops` and `ops-scenario` releases have several documentation artifacts, each
+serving a separate purpose and covering a different level.
```
james-garner-canonical, line 376: This sentence left me wondering what these different artefacts are specifically and what the implications are for someone doing a release.

Aren't all our docs always tracking main, rather than specific releases?

Are the release artifacts changing from just being zipped source code?

Oh also today I learned that artifacts is US English, and artefacts is British English.

james-garner-canonical (commented):
Thanks Tony, looking good. Just a few more questions from me, sorry!

HACKING.md
```diff
@@ -318,65 +318,152 @@ the build frontend is [build](https://pypi.org/project/build/).
...
-
-This will trigger automatic builds for the Python packages and publish them to
-PyPI ([ops](https://pypi.org/project/ops/) and
-[ops-scenario](https://pypi.org/project/ops-scenario)) (authorisation is handled via a
-[Trusted Publisher](https://docs.pypi.org/trusted-publishers/) relationship).
+9. Save the release notes as a draft, and have someone else in the Charm-Tech
+   team proofread the release notes.
+10. If the release includes both `ops` and `ops-scenario` packages, then push a
+   new tag in the form `scenario-<major>.<minor>.<patch>`. This is done by
+   executing `git tag scenario-x.y.z`, then `git push upstream --tags` locally
```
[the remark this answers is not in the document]
tonyandrewmeyer:
```suggestion
   executing `git tag scenario-x.y.z`, then `git push upstream tag scenario-x.y.z` locally
```

HACKING.md
tonyandrewmeyer:
```suggestion
`ops` and `ops-scenario` releases have several documentation artefacts, each
serving a separate purpose and covering a different level.
```

tonyandrewmeyer: > This sentence left me wondering what these different artefacts are specifically and what the implications are for someone doing a release.

Hmm, I maybe need to use a different term then. They are the things outlined in the subsequent sections: raw logs, changelogs, release notes, release post.

> Aren't all our docs always tracking main, rather than specific releases?

Yes, except for these documents that are specifically about the release.

> Are the release artifacts changing from just being zipped source code?

The zipped and tar.gz repository dump (what GitHub calls "assets") stay the same: I don't think we can change those, and no-one should really be using them. The CI pipeline produces .tar.gz and wheels of the two packages (unchanged for `ops`, minor change for `ops-scenario` in that it previously was wheel-only) that are uploaded to PyPI and attestations for each of those (unchanged for `ops`, new for `ops-scenario`). We manually create the `CHANGES.md` files (unchanged for `ops`, new for `ops-scenario`), which are included in both the "assets" and PyPI artefacts. We also manually create the GitHub Releases text and Discourse post, which only live in those places.

Any suggestions for a better/clearer title here? Or clearer text?

james-garner-canonical: Oh I see now. I just immediately jumped to build artefacts. Probably just my unfamiliarity with terminology. Mentally substituting with "We produce several pieces of documentation for `ops` and `ops-scenario` releases, each serving ..." makes everything make sense for me.

HACKING.md
james-garner-canonical: We can add it to the release instructions! "If you're the first one to do a patch release, update the release instructions with what you did and remove this directive."

james-garner-canonical: Clarification: mostly joking. I'll hit approve now as I don't think any of this is worth blocking merging over (though you have two approvals anyway)

james-garner-canonical (approved):

benhoyt (changes requested):
I've had a look over this, and the changes themselves look okay to me, but I'm marking as "Request changes" as a "do not merge yet" signal, as I think we should have further discussion about whether this is the right approach as a team. In particular, the list of steps are getting quite unwieldy especially with the "if you're just releasing ops, do X" or "for ops-scenario, do Y" kind of stuff.

So let's discuss in The Hague and see if we can simplify. I'll add it to our list of discussion topics.

tonyandrewmeyer:
> So let's discuss in The Hague and see if we can simplify. I'll add it to our list of discussion topics.

I think the conclusion of that discussion was that we would look into further improvements and simplifications and automation.

In the meantime, I think we should merge this as the current state to work against, rather than what we have in main. It would also allow us to do the releasing in the meantime, even if it's not as ideal as we would perhaps like.

benhoyt (approved):
Per discussion, we're going to merge this and then future changes (Dima was going to look at that) will be on top of this.

PR #1433, opened 2024-10-11
chore: use annotations from the __future__

tonyandrewmeyer:
The goal here is to reduce the use of the name "Scenario" in the docs, to control where it appears when the docs are included at ops.readthedocs.io.

Using `from __future__ import annotations` seems to make Sphinx much happier when doing the class signatures, avoiding odd text like `~scenario.state.CloudCredential` instead of the expected link with text "CloudCredential" and destination that class.

When adding those imports, pyupgrade transformed all the type annotations. I've run all the tests with 3.8, 3.9, 3.10, 3.11, and 3.12 and they all pass, and I've manually tested in 3.8 and everything seems to work without any problems. I like this much more, so it seems like a win-win.

You may find it easier to review commit-by-commit, as [one commit is completely automated](https://github.com/canonical/operator/commit/7f250a1161c49c193018b90bd769fc88060fa4c3), running `pyupgrade`, and it is responsible for the majority of lines changed.

Also removes a couple of references to the name "Scenario", as the original intent of this change was to clean up references in the docs.

Migrated from https://github.com/canonical/ops-scenario/pull/203


james-garner-canonical (approved):
Looks good to me! The manual changes are very straight forward and make sense. I'll confess that I didn't read every line of the automated changes, but they look sane, and things aren't breaking.

dimaqq (approved):

Issue #1335, opened 2024-08-26
Upgrade ruff to latest


dimaqq:
We're on 0.4.5 and the newest today is 0.6.2.

Newer version produces a ton of warnings, but the changes are mostly mechanical.

- ruff aims to support py 3.10, thus issues many warnings on py 3.8 type hints that have to be suppressed
- some warnings can be auto-fixed by ruff itself
- a few warnings seems not entirely trivial, but I didn't have time for those

PR #1437, opened 2024-10-18
chore: upgrade ruff to v0.7.0

IronCore864:
Currently, we are using ruff 0.4.5 but recently 0.7.0 was released, this PR upgrades ruff to 0.7.0 for both formatting and linting.

Closes https://github.com/canonical/operator/issues/1335.

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -716,7 +716,7 @@ class Framework(Object):
             parent_path = parent.path
 
         kind_: str = kind or cls.handle_kind
-        self._type_registry[(parent_path, kind_)] = cls
+        self._type_registry[parent_path, kind_] = cls
         self._type_known.add(cls)
 
     def save_snapshot(self, value: Union['StoredStateData', 'EventBase']):
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -3216,8 +3216,6 @@ class _FilesParser:
         self._parser.feed(data)
 
     def _prepare_tempfile(self, filename: str):
-        # TODO: Remove noqa SIM115 when the following issue is fixed.
-        # https://github.com/astral-sh/ruff/issues/7313
         tf = tempfile.NamedTemporaryFile(delete=False)  # noqa: SIM115
         self._files[filename] = tf  # type: ignore # we have a custom protocol for it
         self.current_filename = filename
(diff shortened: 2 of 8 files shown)


benhoyt (commented):
Looks good overall, thanks. Regarding [this issue](https://github.com/astral-sh/ruff/issues/7313) / SIM115, unless I'm missing something, it looks like it's intentionally flagging those now, so I don't think we should have TODO comments on these, right?

We should just keep the `noqa: SIM115` comments as this will be always on now.

dimaqq:
> Looks good overall, thanks. Regarding [this issue](https://github.com/astral-sh/ruff/issues/7313) / SIM115, unless I'm missing something, it looks like it's intentionally flagging those now, so I don't think we should have TODO comments on these, right?
> 
> We should just keep the `noqa: SIM115` comments as this will be always on now.

I've opened https://github.com/canonical/operator/issues/1440 to fix this.
Fine to keep `# noqa: ...` or `# noqa: ...  # https://github.com/canonical/operator/issues/1440` until then.

dimaqq (changes requested):
I wonder why so few changes are required.

Perhaps it's because Tony already took care of a lot of stuff during the Scenario merge.

Or maybe it's an invitation for us re-enable some of the ignored codes?

Please check if there's some low-hanging fruit. 

For example, it seems that C408 no longer needs to be ignored.

tonyandrewmeyer:
> I wonder why so few changes are required.

I suspect that using `--preview` helps, because that lets us be a bit ahead of rules hitting stable. We also generally write well crafted and formatted code, so there's less for the linter/formatter to find problems with. Looking over the new rules and rule changes, there's a bunch that aren't related to anything we do (e.g. async stuff).

> Or maybe it's an invitation for us re-enable some of the ignored codes?

For background: most of the ones that are in pyproject.toml as ignored (in the first section) we want to keep ignoring. There are a few that have comments indicating that we'd want to consider enabling them in the future. I think mostly those will be better as separate PRs, though. If there are `noqa` pragmas in the code then those would be great to remove if that becomes feasible.

It would be good to check if there are any new rules/sets that we should consider enabling, either here or separately.

IronCore864:
Per the discussion above, we are keeping `noqa` for SIM115.

Per discussion in today's standup, we agreed to merge this PR first and then handle the refactor (try to remove rules that are not hitting anything or only affect a few places) in a separate PR.

dimaqq (approved):

benhoyt (approved):

PR #1438, opened 2024-10-18
chore: upgrade pyright to 1.1.385

IronCore864:
Upgrade pyright version to the latest (1.1.385).

### DIFF
--- a/test/test_pebble.py
+++ b/test/test_pebble.py
@@ -1359,7 +1359,7 @@ class MockClient(pebble.Client):
             raise resp
         if callable(resp):
             resp = resp()
-        return resp
+        return resp  # type: ignore
 
     def _request_raw(
         self,
--- a/tox.ini
+++ b/tox.ini
@@ -80,7 +80,7 @@ allowlist_externals = cp
 deps =
     PyYAML==6.*
     websocket-client==1.*
-    pyright==1.1.377
+    pyright==1.1.385
     pytest~=7.2
     typing_extensions~=4.2
     -e .


benhoyt (approved):
Wow, much simpler than previous Pyright upgrades! Maybe Python types are stabilising after all.

test/test_pebble.py
```diff
@@ -1359,7 +1359,7 @@ def _request(
             raise resp
         if callable(resp):
             resp = resp()
-        return resp
+        return resp  # type: ignore
```
tonyandrewmeyer, line 1362: Do we know why tge ignore ia needed now? Can we do anything else to avoid it?

tonyandrewmeyer (approved):

tonyandrewmeyer:
> Wow, much simpler than previous Pyright upgrades! Maybe Python types are stabilising after all.

Its ominous that one of the two features landed for 3.14 is type related though 😂.

test/test_pebble.py
IronCore864: I had done some testings before ignoring this, and I think it's because the `self.responses` is `typing.List[typing.Any]`.

If a response is an exception, it's raised. If it's a callable, it's called and the returned type is an object. If none of those cases are matched, it's returned.

So the returned result type is `object | Any`, which doesn't match the method's return type `typing.Dict[str, typing.Any]`.

Casting the final return would help, like `return typing.cast(typing.Dict[str, typing.Any], resp)`. I ignored it because I think it's simpler, since the return type is already annotated with this type.

I'm merging now with the ignore, if the type casting is preferred, I can change it.

tonyandrewmeyer: Ok, so probably the best thing would be to tighten up the annotation for `responses` (or not overload it as much) to avoid needing the ignore or a cast. I guess the newer pyright found this but the older one didn't for some reason.

Leaving this for another day seems fine - I don't think it's worth opening an issue or PR now.

benhoyt: Yeah, it's in a test file so I don't think worth spending more time on now.

dimaqq: When in doubt, do this:

```py
from typing import reveal_type

reveal_type(resp)
```

I've poked around for a couple of minutes and discovered that:
- if resp is callable, that callable always raises, and the line above is misleading
- it's trivial to set the correct type, (below) and even our lambdas are auto-detcted when because they call functions that raise uconditionally

```py
self.responses: list[Callable[[], NoReturn]|dict[str, Any]|Exception] = []
```

(code above assumes `from __future__ import annotations` and `from typing import ...`)

Issue #1346, opened 2024-08-28
Address CI warnings


dimaqq:
Annotations
4 warnings

static
The `python-version` input is not set. The version of Python currently in `PATH` will be used.

test-real-pebble (ubuntu-latest, 3.8)
Restore cache failed: Dependencies file is not found in /home/runner/work/operator/operator. Supported file pattern: go.sum

test-real-pebble (ubuntu-latest, 3.10)
Restore cache failed: Dependencies file is not found in /home/runner/work/operator/operator. Supported file pattern: go.sum

test-real-pebble (ubuntu-latest, 3.12)
Restore cache failed: Dependencies file is not found in /home/runner/work/operator/operator. Supported file pattern: go.sum

Sample run:
https://github.com/canonical/operator/actions/runs/10605060925?pr=1331

PR #1439, opened 2024-10-18
chore: fix ci warnings

IronCore864:
Fix Python version warnings and go cache warnings in github actions.

Closes https://github.com/canonical/operator/issues/1346.

On the go cache warning: By default [the cache is enabled](https://github.com/actions/setup-go/blob/main/action.yml#L15), but in our case, since we only need go to install Pebble and there is no `go.sum` file, there is no need for the cache, see the logic [in the code here](https://github.com/actions/setup-go/blob/main/src/main.ts#L67-L79). So I am disabling the cache explicitly, which should not make the actions run slower but should suppress the warnings.

See a successful run [here](https://github.com/canonical/operator/actions/runs/11402361559) and there is no more warnings.

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -30,6 +30,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up Python 3
         uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
 
       - name: Install tox
         run: pip install tox~=4.2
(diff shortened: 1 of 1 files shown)


.github/workflows/framework-tests.yaml
```diff
@@ -79,6 +81,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "1.22"
+          # To suppress the "Restore cache failed" error, since there is no go.sum file here.
+          cache: false
```
dimaqq, line 85: I guess there are 2 ways to go here:
1. no cache, always run pebble@master
2. `go get ...` and pin pebble

I think that @benhoyt intended the first option, so LGTM from me.
(Do speak up if that's not the case)

dimaqq (approved):

benhoyt (approved):
This looks good to me, thanks. Yep, I intended to always run pebble@master.

PR #1441, opened 2024-10-22
refactor: update linting with the latest ruff

IronCore864:
Update linting rules with the latest ruff version 0.7.0.

After upgrading ruff to the latest version [in this PR](https://github.com/canonical/operator/pull/1437), some rules can be removed since they are not hitting any code, and some other rules can be relatively easily refactored.

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3486,8 +3486,10 @@ class _TestingPebbleClient:
             raise self._api_error(404, f'stat {path}: no such file or directory')
         files = [file_path]
         if not itself:
-            with contextlib.suppress(NotADirectoryError):
+            try:
                 files = [file_path / file for file in os.listdir(file_path)]
+            except NotADirectoryError:
+                pass
 
         if pattern is not None:
             files = [file for file in files if fnmatch.fnmatch(file.name, pattern)]
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -14,7 +14,6 @@
 
 """Infrastructure to build unit tests for charms using the ops library."""
 
-import contextlib
 import dataclasses
 import datetime
 import fnmatch
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -14,6 +14,7 @@
 
 """Infrastructure to build unit tests for charms using the ops library."""
 
+import contextlib
 import dataclasses
 import datetime
 import fnmatch
@@ -39,6 +40,7 @@ from typing import (
     AnyStr,
     BinaryIO,
     Callable,
+    ClassVar,
     Dict,
     Generic,
     Iterable,
@@ -2225,7 +2227,7 @@ def _copy_docstrings(source_cls: Any):
 class _TestingConfig(Dict[str, Union[str, int, float, bool]]):
     """Represents the Juju Config."""
 
-    _supported_types = {
+    _supported_types: ClassVar[Dict[str, Any]] = {
         'string': str,
         'boolean': bool,
         'int': int,
@@ -2838,9 +2840,7 @@ class _TestingModelBackend:
         unit_secret = secret.owner_name == self.unit_name
         app_secret = secret.owner_name == self.app_name
 
-        if unit_secret or (app_secret and self.is_leader()):
-            return True
-        return False
+        return unit_secret or (app_secret and self.is_leader())
 
     def secret_info_get(
         self, *, id: Optional[str] = None, label: Optional[str] = None
@@ -3486,10 +3486,8 @@ class _TestingPebbleClient:
             raise self._api_error(404, f'stat {path}: no such file or directory')
         files = [file_path]
         if not itself:
-            try:
+            with contextlib.suppress(NotADirectoryError):
                 files = [file_path / file for file in os.listdir(file_path)]
-            except NotADirectoryError:
-                pass
 
         if pattern is not None:
             files = [file for file in files if fnmatch.fnmatch(file.name, pattern)]
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -22,6 +22,7 @@ import warnings
 from typing import (
     TYPE_CHECKING,
     Any,
+    ClassVar,
     Dict,
     List,
     Literal,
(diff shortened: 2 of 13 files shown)


ops/charm.py
```diff
@@ -1678,7 +1679,7 @@ class RelationMeta:
     ``metadata.yaml`` and used by the charm code if appropriate.
     """
 
-    VALID_SCOPES = ['global', 'container']
+    VALID_SCOPES: ClassVar[List[str]] = ['global', 'container']
```
dimaqq, line 1682: Arguably, it's a constant and could be a tuple or a frozenset instead of a list.
In that case, ruff won't bother us about ClassVar.

My reading is that ruff only insists that **mutable** class-level fields should be ClassVars.

It appears that no charm uses this thing (semantically a constant, but technically a public field on a public class).

Which means that we could change it's type.

cc @tonyandrewmeyer and @benhoyt

ops/_private/harness.py
```diff
@@ -2225,7 +2227,7 @@ def decorator(target_cls: Any):
 class _TestingConfig(Dict[str, Union[str, int, float, bool]]):
     """Represents the Juju Config."""
 
-    _supported_types = {
+    _supported_types: ClassVar[Dict[str, Any]] = {
```
dimaqq, line 2230: I wonder if ClassVar is correct here.
The field is only mutable because it's a dict.
Semantically, it's not meant to be modified.

ops/model.py
```diff
@@ -2023,7 +2024,7 @@ def _register(cls, child: Type['StatusBase']) -> None:
             )
         cls._statuses[child.name] = child
 
-    _priorities = {
+    _priorities: ClassVar[Dict[str, Any]] = {
```
dimaqq, line 2027: Same here wrt. mutability.

ops/charm.py
dimaqq: On the other hand, https://peps.python.org/pep-0526/ states that the intention of ClassVar is to prevent the field being accidentally overridden in an instance, regardless of mutability.

ops/_private/harness.py
```diff
@@ -3486,10 +3486,8 @@ def list_files(
             raise self._api_error(404, f'stat {path}: no such file or directory')
         files = [file_path]
         if not itself:
-            try:
+            with contextlib.suppress(NotADirectoryError):
                 files = [file_path / file for file in os.listdir(file_path)]
-            except NotADirectoryError:
-                pass
```
tonyandrewmeyer, line 3492: I'm not a fan of `contextlib.suppress`. I don't see how its use improves code at all - the reader needs to know that this exists and what it does, there's no improvement in indentation, it's two lines briefer but *longer* in terms of characters, and I don't think it's any more understandable (although that's subjective of course).

My preference would be to leave this as-is.

pyproject.toml
```diff
@@ -136,8 +136,6 @@ ignore = [
 
     # Return condition directly, prefer readability.
     "SIM103",
-    # Use contextlib.suppress() instead of try/except: pass
-    "SIM105",
```
tonyandrewmeyer, line 140: As noted above, I would prefer that we kept this.

pyproject.toml
```diff
@@ -146,17 +144,9 @@ ignore = [
...
-    "PERF401",
     # Manual dict comprehension.
     "PERF403",
 
-    # Convert Dict[X] to dict[x], which we cannot do while we support Python 3.8
-    "UP006",
-
-    # Convert Union[X, Y] to X | Y, which we cannot do while we support Python 3.8
-    "UP007",
-
```
tonyandrewmeyer, line 159: Why is it that we can enable these now? They used to generate warnings because we use e.g. `Union[str, int]`, and we still do need to support Python 3.8, and we haven't added the `__future__` annotations import yet.

pyproject.toml
```diff
@@ -167,14 +157,6 @@ ignore = [
 
     # `subprocess` call: check for execution of untrusted input
     "S603",
-    # Prefer `next(iter(info_dicts))` over single element slice
-    "RUF015",
```
tonyandrewmeyer, line 171: I agree that `next(generator expression)` is better than `[list comprehension][0]`, and that does seem to be where most of the changes are. I'm less convinced about `list(x)[0]` (although that ought to be `tuple(x)[0]` in my opinion).

Are switching this because we have decided that we like this style after all, or just because we want to follow the ruff defaults more closely?

pyproject.toml
IronCore864: I tried linting without these two rules and there were no warnings.

pyproject.toml
IronCore864: I think the ruff rule makes sense here, although my original idea was only to remove as many ignored rules as possible to stay consistent with the community. I will keep the next generator then.

ops/_private/harness.py
IronCore864: Yes, I think here it only indicates the attribute is a class variable and should not be set on instances of the class, the emphasis isn't on the variable/mutable part. I made this change so that I could remove the RUF012 rule from the ignore list. I will keep the type unchanged, same for other private class variables.

IronCore864:
I have made some refactoring according to the reviews. Summary:

- RUF015 (prefer `next(iter(info_dicts))` over single element array/tuple) is removed from the ignore list.
- SIM105 (use `contextlib.suppress()` instead of `try/except: pass`) is kept as it is, still ignored.
- RUF012 (mutable class attributes should be annotated with `typing.ClassVar`) is removed from the ignore list. Since the emphasis is on the class attribute (instead of instance attribute) part, not on the "mutable" part, I kept the type as it is.
- UP006 (convert Dict[X] to dict[x]) and UP007 (convert Union[X, Y] to X | Y) are removed from the ignore list because after testing without ignoring these two rules with Python 3.8, it seems there is no warning.

pyproject.toml
dimaqq: ruff uses target-version setting and doesn't enable UP007 for us because:
- py38
- from future not used

I wonder if perhaps version elicitation was broken before or what?

dimaqq (approved):

benhoyt (approved):

tonyandrewmeyer (approved):

PR #1443, opened 2024-10-30
ci: use the actual poetry command, rather than manually tweaking the file

tonyandrewmeyer:
When handling charms that used poetry, the workflow would do a clone of the ops repo, tweak the poetry.lock file to include it and then re-lock. This leaves a copy of the ops repo around, and sometimes the ops tests get picked up by pytest, which is not what we want here.

The PR changes this to use the proper `poetry add` functionality to add the appropriate version of ops, in the same way that the requirements.txt files are updated.

Additional related fixes:
 * The test runs for the charmcraft profiles also had the operator clone but never used it, so remove that.
 * The charmcraft profile can safely run the static checks (since the charms are so minimal), and that would have caught an issue previously, so add that.

### DIFF
--- a/.github/workflows/published-charms-tests.yaml
+++ b/.github/workflows/published-charms-tests.yaml
@@ -59,17 +59,11 @@ jobs:
         with:
           repository: ${{ matrix.charm-repo }}
 
-      - name: Checkout the operator repository
-        uses: actions/checkout@v4
-        with:
-          path: myops
-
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Thanks!

dimaqq (approved):

james-garner-canonical (approved):

Issue #1449, opened 2024-11-06
Expose in ops a jhack-eval-like entrypoint


PietroPasotti:
`jhack eval` is a command that allows you to evaluate python expressions in the context of a charm, without having to wait for the charm to receive an event.

`jhack eval traefik/0 "self.unit.name"` will print to stdout `traefik/0`.

This has proved to be extremely useful in debugging and troubleshooting.
It is implemented by scp'ing a file into the unit and then juju-exec'ing it. The file contains much of the machinery in ops.main.main, but never calls the `_emit_charm_event` function; instead, it eval's the user-provided expression while binding the following globals:

- `self`: the charm instance
- `ops`: the ops module, to let the user evaluate expressions such as `setattr(self.unit.status, ops.ActiveStatus())`
- `json`: the json module, to let the user output json instead of plain str/repr.

In certain environments where jhack is not available, it would be handy to have a way to ask the user to evaluate an expression for debugging purposes, to inspect the charm's current state outside of what the logging output shows.
It would be fantastic if we could offer a way for the user to achieve the same functionality without much hackery.

My proposal is to add to ops.main logic to skip event emission if the charm is being executed with a special envvar set.

The user would need to run:
`juju exec --unit traefik/0 OPS_EVAL_EXPR="self.unit.name" ./dispatch`

(or something like that).

When called with that envvar set, ops.main will not emit an event on the charm, but instead will evaluate the provided exception while binding the same globals that jhack exposes.

see https://github.com/canonical/operator/pull/1450 for a proof of concept implementation

dimaqq:
I would have thought that `setattr(self.unit.status, ops.ActiveStatus())` wouldn't actually work outside the context of a hook. Am I missing something?

benhoyt:
@dimaqq No, it looks like the `status-set` hook tool works fine outside the context of a hook (at least with `juju exec`):

```
$ juju exec --unit snappass-test/0 -- status-set blocked 'blocky!'
$ juju status
Model  Controller          Cloud/Region        Version    SLA          Timestamp
t      microk8s-localhost  microk8s/localhost  3.6-beta1  unsupported  15:47:44+13:00

App            Version  Status   Scale  Charm          Channel        Rev  Address         Exposed  Message
snappass-test           waiting      1  snappass-test  latest/stable    9  10.152.183.112  no       installing agent

Unit              Workload  Agent  Address       Ports  Message
snappass-test/0*  blocked   idle   10.1.164.163         blocky!
```

[...]

PR #1450, opened 2024-11-06
feat: ops eval hook

PietroPasotti:
Proposal to address #1449

### DIFF
--- a/ops/_main.py
+++ b/ops/_main.py
@@ -33,6 +33,7 @@ from ops.jujucontext import _JujuContext
 from ops.log import setup_root_logging
 
 CHARM_STATE_FILE = '.unit-state.db'
+CHARM_EVAL_EXPR_ENVVAR = 'CHARM_EVAL_EXPR'
 
 
 logger = logging.getLogger()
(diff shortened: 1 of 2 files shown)


ops/_main.py
```diff
@@ -523,15 +525,43 @@ def _commit(self):
         """Commit the framework and gracefully teardown."""
         self.framework.commit()
 
+    def _eval(self, expr: str):
+        """Eval an expression in the context of the charm and print the result to stdout."""
+        import json
+
+        globs = {'self': self.charm, 'ops': ops, 'json': json}
```
dimaqq, line 532: I feel a slippery slope here: start with self, add ops, then `json.dumps()` because surely it's useful, then maybe yaml? pathlib? os? stdlib operator?

benhoyt: Yeah, I agree. `self` seems fine as you kinda need the charm instance, `ops` seems reasonable, but anything beyond that I'd see if we can find a different way to import stuff. If we go the "exec" route (rather than "eval"), I guess you can do `import json; json.dumps(x)`.

ops/_main.py
```diff
@@ -20,6 +20,7 @@
 import subprocess
 import sys
 import warnings
+from optparse import Option
```
dimaqq, line 23: OT: unused, afaics

ops/_main.py
PietroPasotti: to be fair, jhack also has a sister command, `jhack script`, that allows loading in a python file with a `main(charm:Charm)` entrypoint for the kind of 'more complex' scenarios where you may need to import stuff and the like

PietroPasotti: eval is mostly for inspection, not so much mutation. json felt like a universal, low-barrier addition. Yaml I wouldn't object to, but pathlib, os, operator feel too much. It's not about allowing you to golf complex operations into a string, it's about inspecting live data

benhoyt:
Waiting on a spec for #1449, going to close for now.

PR #1456, opened 2024-11-18
chore: minor linting improvements to testing/src/scenario

tonyandrewmeyer:
We're not aiming to get 100% in the TIOBE scoring, but it's nice to get a higher score when it aligns with our existing style and doesn't involve significant changes. The `testing` folder already scores 95.99 but there's still some low-hanging fruit, which this PR addresses:

* Some basic import reordering, which is mostly because "ops" is first-party now.
* Clearing out outdated TODO/FIXME style comments (I've considered each of these, and the ones removed either are out of date, are not something we want to do now, or already have tickets).
* TIOBE's linter prefers multiple `if` statements rather than if/elif/else when the block unconditionally returns/raises. This is a debatable style choice, but where it seemed reasonable I've applied that here. I don't think we have a team style decision on this yet.
* Added module, class, method docstrings.


testing/src/scenario/state.py
```diff
@@ -1692,18 +1688,19 @@ class DeferredEvent:
     # them because they need a Handle.
     @property
     def name(self):
+        """A comparable name for the event."""
         return self.handle_path.split("/")[-1].split("[")[0]
 
 
 class _EventType(str, Enum):
-    framework = "framework"
```
james-garner-canonical, line 1699: Huh, I guess inheriting from `str` and `Enum` is a way to get `StrEnum` like behaviour for older pythons

james-garner-canonical (approved):

testing/src/scenario/_consistency_checker.py
```diff
@@ -170,14 +170,16 @@ def check_event_consistency(
...
     warnings: List[str] = []
 
-    # custom event: can't make assumptions about its name and its semantics
-    # todo: should we then just skip the other checks?
     if not event._is_builtin_event(charm_spec):
-        warnings.append(
-            "this is a custom event; if its name makes it look like a builtin one "
-            "(e.g. a relation event, or a workload event), you might get some false-negative "
-            "consistency checks.",
-        )
```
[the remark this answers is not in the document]
tonyandrewmeyer: I bundled it into the "remove outdated TODOs" item. Basically, these lines and the warning are all saying "if this is a custom event whose name looks like a Juju event, then the consistency checker is going to do the wrong thing.

For example, say you have a custom `database-relation-changed` event - the test framework can't really tell if that's a Juju relation-changed event for an endpoint called "database" or your custom thing.

My opinion (which I'm pushing into the PR) is that in these situations the best thing to do is just skip the consistency checks. I think that's a better choice than the false-negative (which I think would actually be false-positive) results and the warning. Basically: the todo's question of "should we just skip them" I think is answered "yes", so I did that.

testing/src/scenario/mocking.py
```diff
@@ -2,6 +2,11 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+"""Juju and Pebble mocking
+
+This module contains mocks for the Juju and Pebble APIs that are used by ops
+to interact with the Juju controller and the Pebble service manager."""
+
```
tonyandrewmeyer, line 9:
```suggestion
to interact with the Juju controller and the Pebble service manager.
"""

```

Yes, odd that the checker didn't find that.

testing/src/scenario/runtime.py
```diff
@@ -151,6 +153,7 @@ def __enter__(self):
         pass
 
     def emit(self):
+        """Emit the event."""
         self._has_emitted = True
```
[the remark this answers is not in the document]
tonyandrewmeyer: No, it's in a private class, so users don't interact with it at all. It's not a great docstring, but it's also scheduled for removal (we'll use the new(ish) _Manager class instead of this stuff) so I didn't take much care.

Basically, for the testing case we don't need to do anything here, other than record that happened.

I'll make it better.

testing/src/scenario/runtime.py
```diff
@@ -151,6 +153,7 @@ def __enter__(self):
         pass
 
     def emit(self):
+        """Emit the event."""
```
tonyandrewmeyer, line 156:
```suggestion
        """Emit the event.
        
        For the testing case, all that's required is recording that it was emitted.
        """
```

testing/src/scenario/runtime.py
```diff
@@ -254,7 +253,7 @@ def _get_event_env(self, state: "State", event: "_Event", charm_root: Path):
                 else:
                     logger.warning(
                         "remote unit ID unset; no remote unit data present. "
-                        "Is this a realistic scenario?",  # TODO: is it?
+                        "Is this a realistic scenario?",
```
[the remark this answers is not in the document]
tonyandrewmeyer: We don't have any extra information. I think it's up to the user to decide what is and what isn't a realistic scenario (state), assuming that it is actually possible in Juju. I think you could have this in Juju but it would be a somewhat broken state to be in, so the person writing the tests probably has made a mistake, but we're not 100% sure on that so it's better to warn than error out.

testing/src/scenario/state.py
tonyandrewmeyer: Yeah, that's what people usually do/did from my experience. It's not perfect, but I guess close enough that people found it ok.

dimaqq (approved):

IronCore864 (approved):

PR #1458, opened 2024-11-19
fix: make push_path open in binary mode so it works on non-text files

### LINKED ISSUE #1455: `Container.push_path()` only supports text files
As you can see in the linked code below, the `Container.push_path()` function only works with text files. Internally, it's calling `Container.push()`, but it doesn't allow customizing its parameters, so you can't specify *encoding*, *permissions*, *user_id* and so on.

https://github.com/canonical/operator/blob/77b85ee2ce8e9319f89aef443c7337684f4fd21d/ops/model.py#L2594-L2598

It would be great if at least the `encoding` parameter was surfaced in `push_path`, so that it's possible to get binary files from the charm container.

My current workaround:
```python
with open(binary_file, "rb") as f:
	container.push(dst_path, source=f)
```

benhoyt:
There was a bug in `push_path` where it called `open` with the default (text) mode on the local files, causing it to not work on binary files (more specifically, if they weren't valid UTF-8). This is the fix for that -- `push_path` should treat the files as raw bytes, so read in binary mode.

Also change the test to ensure binary / non-UTF-8 files work as expected.

Fixes #1455.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2662,7 +2662,7 @@ class Container:
                     if info.type is pebble.FileType.DIRECTORY:
                         self.make_dir(dstpath, make_parents=True)
                         continue
-                    with open(info.path) as src:
+                    with open(info.path, 'rb') as src:
                         self.push(
                             dstpath,
                             src,
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -1369,8 +1369,8 @@ def test_recursive_push_and_pull(case: PushPullCase):
     for file in case.files:
         fpath = os.path.join(push_src.name, file[1:])
         os.makedirs(os.path.dirname(fpath), exist_ok=True)
-        with open(fpath, 'w') as f:
-            f.write('hello')
+        with open(fpath, 'wb') as f:
+            f.write(b'push \xc3\x28')  # invalid UTF-8 to ensure binary works
     if case.dirs:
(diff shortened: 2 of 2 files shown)


james-garner-canonical (commented):
Looks good to me.

I wondered if defaulting to reading bytes would be problematic, but it's not, because previously with a text file `pebble.Client.push` would just read the text and encode it as bytes before sending (defaulting to UTF-8 -- and `Container.push_path` doesn't care about or allow the user to specify encoding anyway). Likewise, `Container.pull_path` always pulls with `encoding=None`, so it just works with the bytes directly as well.

So this is indeed just a straightforward bugfix.

benhoyt:
> Previously, line endings would get normalised to \n, since Python does that when opening in text mode.

Yeah, I'd consider that a subtle but nasty bug. I was involved in the original design of push and pull, and the intention with push and pull was very much (and especially the recursive push_path and pull_path versions) was copying files, not modifying the data or doing line ending normalisation. In Go it's all bytes anyway. In Python, push and pull themselves allow supply strings which are then encoded as UTF-8, but that's just to make passing strings/TextIO easier in Python, not for normalisation reasons.

I think one test is probably fine here as the intent is that there's only code path, so going to run with this. Thanks!

james-garner-canonical (approved):

PR #1461, opened 2024-11-25
chore: update charm pins

benhoyt:
This is an automated PR to update pins of the external repositories that the operator framework is tested against

PR #1463, opened 2024-11-27
ci: a better way than commenting out external repos

dimaqq:
Proposing a fix to https://github.com/canonical/operator/pull/1461 deleting commented out repos


benhoyt (approved):
Looks like a reasonable approach -- thanks.

tonyandrewmeyer (approved):
So I gather what will happen here is that the commit line will still get bumped, but it will remain disabled. That seems reasonable to me.

I do think we need to document this somewhere, though. That could just be a `disabled: false` on the ones that are ok, or maybe it should be in HACKING.md or the action's readme.md or somewhere? Otherwise once nothing is disabled, it won't be obvious that this is how you do it (although the `if` will still be there).

I didn't actually run the script to verify that this will probably work, but looking over the code it seems like it, and we'll see soon enough if not, and it wouldn't be disruptive anyway.

PR #1467, opened 2024-11-27
refactor!: use ops._main._Manager in Scenario

### LINKED ISSUE #1425: Remove testing.Ops, use ops.main.Manager instead
Now that large part of the ops_main_mock.Ops machinery [has been upstreamed into ops proper](https://github.com/canonical/operator/pull/1085), we should consider cleaning up ops_main_mock and see what can be simplified.

Moved from https://github.com/canonical/ops-scenario/issues/125

tonyandrewmeyer:
The ops._main._Manager class started out as the Ops class in Scenario, but went through significant modifications before it was added in ops. This PR merges those changes back into Scenario, so that it's also using the `_Manager` class. Other than generally simplifying the code, this also means that Scenario is even more consistent with using ops in production.

This includes some small refactoring in ops to make this cleaner:
* `ops._main._emit_charm_event` is only used by `_Manager`, so logically belongs inside that class. This also makes it much simpler to have a subclass override the behaviour.
* Allow passing a `JujuContext` to `_Manager`, so that it can be populated by something other than `os.environ`.
* Allow `_Manager` subclasses to override how the charm metadata is loaded.
* Allow `_Manager` subclasses to execute code after committing but before the framework is closed.
* Support a dispatch path that uses `_` rather than `-` in event names (Scenario consistently uses `_`, even though this would always be `-` from Juju - feel free to push back on this, requiring Scenario to undo the `_` conversion when populating the context).

The Scenario `runtime` and `ops_main_mock` modules are renamed to properly indicate that everything inside of them is an implementation detail and should be considered private by Scenario users. This is a breaking change but we're considering it a bug fix and not one that requires a major release, given that the documentation and top-level namespace were already clear on this.

Significant Scenario refactoring:
* UnitStateDB is moved from `runtime` to `ops_main_mock`, and it no longer (re) opens the SQLiteStorage - it instead takes a reference to the one that's in the framework, so that we're consistently using the same underlying storage, without relying on the filename being consistent (this eases moving to `:memory:` in the future).
* As far as I can tell, `runtime._OpsMainContext` is completely unused and missed being removed in previous refactoring, so it's entirely deleted.
* Loading and storing deferred events and stored state into the state database is moved from the runtime to the manager subclass.
* The `Ops` class survives, but is now a subclass of the `ops._main._Manager` class. A reasonable amount of functionality is now removed (because the parent handles it), and the remaining code is specifically for handling the testing framework.

A few type hints are added in some tests where I was working on fixing the test and it helped to do that.

Fixes #1425


ops/_main.py
```diff
@@ -515,19 +499,48 @@ def _emit(self):
...
+            juju_context: An instance of the _JujuContext class.
+        """
+        event_to_emit = self._get_event_to_emit(event_name)
+
+        # If the event is not supported by the charm implementation, do
+        # not error out or try to emit it. This is to support rollbacks.
+        if event_to_emit is not None:
+            args, kwargs = _get_event_args(self.charm, event_to_emit, self._juju_context)
+            logger.debug('Emitting Juju event %s.', event_name)
+            event_to_emit.emit(*args, **kwargs)
```
james-garner-canonical, line 528: Minor style change that I think is easier to follow, up to you
```suggestion
        if event_to_emit is None:
            # If the event is not supported by the charm implementation, do
            # not error out or try to emit it. This is to support rollbacks.
            return
        args, kwargs = _get_event_args(self.charm, event_to_emit, self._juju_context)
        logger.debug('Emitting Juju event %s.', event_name)
        event_to_emit.emit(*args, **kwargs)
```

ops/_main.py
```diff
@@ -515,19 +499,48 @@ def _emit(self):
...
+
     def _commit(self):
         """Commit the framework and gracefully teardown."""
         self.framework.commit()
 
+    def _close(self):
+        """Perform any necessary cleanup before the framework is closed."""
+        # Provided for child classes - nothing needs to be done in the base.
```
james-garner-canonical, line 536: Comment could go in docstring?

testing/src/scenario/_consistency_checker.py
```diff
@@ -170,16 +170,14 @@ def check_event_consistency(
...
-        # requirements for the custom event are (in terms of the state), so we
-        # skip everything here. Perhaps in the future, custom events could
-        # optionally include some sort of state metadata that made testing
-        # consistency possible?
-        return Results(errors, warnings)
+        warnings.append(
+            "this is a custom event; if its name makes it look like a builtin one "
+            "(e.g. a relation event, or a workload event), you might get some false-negative "
+            "consistency checks.",
+        )
```
james-garner-canonical, line 180: I like emitting a warning here, but given the comments (old and new), it kind of seems like returning early and skipping the other checks is correct.

Is the idea that we shouldn't be shielding users from the consequences of using a poorly named custom event that could clash with a relation/workload/action/storage event?

testing/src/scenario/_runtime.py
```diff
@@ -249,7 +136,7 @@ def _get_event_env(self, state: "State", event: "_Event", charm_root: Path):
                 else:
                     logger.warning(
                         "remote unit ID unset; no remote unit data present. "
-                        "Is this a realistic scenario?",
+                        "Is this a realistic scenario?",  # TODO: is it?
```
james-garner-canonical, line 139: Strong deja vu here haha, didn't we talk about this exact comment on another PR?

tonyandrewmeyer:
This maybe broke the [recent variadic type change](https://github.com/canonical/operator/pull/1445), since `ops._main._Manager` now sets the `.charm` attribute and the change didn't extend that far. I'll look into that.

testing/src/scenario/_ops_main_mock.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env python3
```
james-garner-canonical, line 1: This is actually a refactor of `ops_main_mock.py` and not a completely new file, right?

james-garner-canonical (commented):
Looks like a bunch of changes have come in while I'm reviewing, I'll just leave these comments for now and finish the review later.

tonyandrewmeyer:
> Looks like a bunch of changes have come in while I'm reviewing, I'll just leave these comments for now and finish the review later.

The changes are only merging in the charm type PR (well, rebasing to get that, but because I rename the file here it was a bit messy). No functional changes, although as noted above I wonder if I break that feature here and need to get the generic type over to main somehow.

testing/src/scenario/_ops_main_mock.py
tonyandrewmeyer: Yes. The first commit only renames the files and does nothing else so that it's easier to see, if you want to look that way. I'm unsure why the diff shows as changes for runtime, but not for ops_main_mock - maybe "too much" has changed?

testing/src/scenario/_runtime.py
tonyandrewmeyer: Ah, dammit, maybe my rebase/merge has gone wonky and is showing stuff already in main. I checked it had the changes from the latest PR that was merged and that tests passed, but not more. I'll look.

ops/_main.py
```diff
@@ -515,19 +499,48 @@ def _emit(self):
...
+        self._emit_charm_event(self.dispatcher.event_name)
         # Emit collect-status events.
         ops.charm._evaluate_status(self.charm)
 
+    def _get_event_to_emit(self, event_name: str) -> Optional[ops.framework.BoundEvent]:
+        try:
+            return getattr(self.charm.on, event_name)
+        except AttributeError:
+            logger.debug('Event %s not defined for %s.', event_name, self.charm)
+        # If the event is not defined, return None.
```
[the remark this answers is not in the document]
tonyandrewmeyer: I wonder if we should discuss implicit vs. explicit 'return None' for the team style guide. I had the feeling some linters don't like implicit, even though the pep8 guideline is to be explicit if there's any other return.

ops/_main.py
tonyandrewmeyer: This was code that was just moved, not added in this PR, so my inclination was to leave it alone rather than changing *and* moving it. But I agree it is nicer that way, so will do.

ops/_main.py
tonyandrewmeyer: I feel like it's not part of the documentation of the method, it's commenting on the (lack of) implementation, so a comment is more appropriate.

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: Yes, there was meant to be the warning and then a return.

PR #1468, opened 2024-11-27
fix: require the same object to be in the testing state as in the event

### LINKED ISSUE #1429: consistency checker gives false positives when working with copied objects
This is a regression from Scenario 6:

MWE:

```python
import ops
from scenario import Context
def test_relation_evt():
    ctx = Context(ops.CharmBase, meta={"name": "mateusz",
                                   "requires": {"a": {"interface": "b"}}})
    rel = Relation("a", interface="b")
    rel_cpy = replace(rel, remote_app_data={"foo": "bar"})

    ctx.run(ctx.on.relation_changed(rel_cpy), State(relations={rel}))
    # scenario.errors.InconsistentScenarioError: Inconsistent scenario.
    # The following errors were found: cannot emit a_relation_changed because 
    # relation 1 is not in the state.
```

culprit is this line in the consistency checker:

```
if event.relation not in state.relations:
    ... <mark as inconsistent>
```

Proposed solution:
```
if event.relation.id not in [r.id for r in state.relations]:
    ...
```

Moved from https://github.com/canonical/ops-scenario/issues/199

tonyandrewmeyer:
This was incorrectly changed during the import of the ops-scenario code to this repo. As discussed in #1429, having an object in the state that has the same identifier as the object in the event is insufficient and could end up being confusing: it should be the same actual object.

As well as reverting the previous change, adjust the error messages so that they're clearer if this case is triggered.

Fixes #1429

### DIFF
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -214,9 +214,11 @@ def _check_relation_event(
                 f"relation event should start with relation endpoint name. {event.name} does "
                 f"not start with {event.relation.endpoint}.",
             )
-        if event.relation.id not in {rel.id for rel in state.relations}:
+        if event.relation not in state.relations:
             errors.append(
-                f"cannot emit {event.name} because relation {event.relation.id} is not in the state.",
+                f"cannot emit {event.name} because relation {event.relation.id} is not in the "
+                f"state (a relation with the same ID is not sufficient - you must "
+                f"pass the object in the state to the event).",
             )
 
 
@@ -241,7 +243,8 @@ def _check_workload_event(
             if event.container not in state.containers:
                 errors.append(
                     f"cannot emit {event.name} because container {event.container.name} "
-                    f"is not in the state.",
+                    f"is not in the state (a container with the same name is not "
+                    f"sufficient - you must pass the object in the state to the event).",
                 )
             if not event.container.can_connect:
                 warnings.append(
@@ -310,12 +313,11 @@ def _check_storage_event(
             f"storage event {event.name} refers to storage {storage.name} "
             f"which is not declared in the charm metadata (metadata.yaml) under 'storage'.",
         )
-    elif (storage.name, storage.index) not in {
-        (s.name, s.index) for s in state.storages
-    }:
+    elif storage not in state.storages:
         errors.append(
             f"cannot emit {event.name} because storage {storage.name} "
-            f"is not in the state.",
+            f"is not in the state (an object with the same name and index is not "
+            f"sufficient - you must pass the object in the state to the event).",
         )
 
 
@@ -479,10 +481,12 @@ def check_secrets_consistency(
         return Results(errors, [])
 
     assert event.secret is not None
-    if event.secret.id not in {s.id for s in state.secrets}:
+    if event.secret not in state.secrets:
         secret_key = event.secret.id if event.secret.id else event.secret.label
         errors.append(
-            f"cannot emit {event.name} because secret {secret_key} is not in the state.",
+            f"cannot emit {event.name} because secret {secret_key} is not in the state "
+            f"(a secret with the same ID is not sufficient - you must pass the object "
+            f"in the state to the event).",
         )
     elif juju_version < (3,):
         errors.append(


james-garner-canonical (approved):
Makes sense, I see this is just rolling back an incorrect change and better documenting the behaviour in error messages

james-garner-canonical:
Why did a bunch of tests get cancelled tho?

tonyandrewmeyer:
> Why did a bunch of tests get cancelled tho?

Yeah, something weird is happening there - the cancelled ones are because one failed, which is for some strange "environment inconsistent" thing, which I have not seen before, and doesn't appear related to the changes.

I'll look into it (and it obviously blocks merging anyway).

tonyandrewmeyer:
> The code looks good to me, but I have to trust you on the minute semantics difference.

Basically, it's this:

In main:

```python
>>> import dataclasses
>>> import ops
>>> from ops import testing
>>> ctx = testing.Context(ops.CharmBase, meta={'name': 'n', 'requires': {'foo': {'interface': 'i'}}})
>>> r = testing.Relation('foo')
>>> r2 = dataclasses.replace(r, interface='x')
>>> r.id, r2.id
(1, 1)
>>> ctx.run(ctx.on.relation_changed(r), testing.State(relations={r2}))
State(config={}, relations=frozenset({Relation(endpoint='foo', interface='x', id=1, local_app_data={}, local_unit_data={'egress-subnets': '192.0.2.0', 'ingress-address': '192.0.2.0', 'private-address': '192.0.2.0'}, remote_app_name='remote', limit=1, remote_app_data={}, remote_units_data={0: {'egress-subnets': '192.0.2.0', 'ingress-address': '192.0.2.0', 'private-address': '192.0.2.0'}})}), networks=frozenset(), containers=frozenset(), storages=frozenset(), opened_ports=frozenset(), leader=False, model=Model(name='qzYc1M0vTmxETBYBRce4', uuid='888c4943-2e56-4913-bf1b-7bb760fbb562', type='kubernetes', cloud_spec=None), secrets=frozenset(), resources=frozenset(), planned_units=1, deferred=[], stored_states=frozenset({StoredState(name='_stored', owner_path=None, content={'event_count': 4}, _data_type_name='StoredStateData')}), app_status=UnknownStatus(), unit_status=UnknownStatus(), workload_version='')
>>> 
```

In the branch:

```python
>>> import dataclasses
>>> import ops
>>> from ops import testing
>>> ctx = testing.Context(ops.CharmBase, meta={'name': 'n', 'requires': {'foo': {'interface': 'i'}}})
>>> r = testing.Relation('foo')
>>> r2 = dataclasses.replace(r, interface='x')
>>> r.id, r2.id
(1, 1)
>>> ctx.run(ctx.on.relation_changed(r), testing.State(relations={r2}))
  File "/home/tameyer/code/operator/testing/src/scenario/context.py", line 665, in run
    with self._run(event=event, state=state) as manager:
  File "/usr/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/home/tameyer/code/operator/testing/src/scenario/context.py", line 687, in _run
    with runtime.exec(
  File "/usr/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/home/tameyer/code/operator/testing/src/scenario/runtime.py", line 435, in exec
    check_consistency(state, event, self._charm_spec, self._juju_version)
  File "/home/tameyer/code/operator/testing/src/scenario/_consistency_checker.py", line 125, in check_consistency
    raise InconsistentScenarioError(
scenario.errors.InconsistentScenarioError: Inconsistent scenario. The following errors were found: cannot emit foo_relation_changed because relation 2 is not in the state (a relation with the same ID is not sufficient - you must pass the object in the state to the event).
relation endpoint foo is not declared in metadata.
>>> 
```

In main, `r` and `r2` have the same ID, so they are somewhat the same relation as far as Juju would be concerned, but they are different objects in the test with different values for other attributes, so it's weird (and potentially problematic) that you can use one to 'bind' the event, and put the other one in the state.

tonyandrewmeyer:
> > Why did a bunch of tests get cancelled tho?
> 
> Yeah, something weird is happening there - the cancelled ones are because one failed, which is for some strange "environment inconsistent" thing, which I have not seen before, and doesn't appear related to the changes.
> 
> I'll look into it (and it obviously blocks merging anyway).

Not sure what happened here. One test failed with a weird environment error that seems unrelated to the changes, and when I re-ran it, it passed. Then there were several others that were cancelled since that one failed, but they all passed when I re-ran them.

I'm assuming this was some GitHub workflows glitch.

PR #1469, opened 2024-11-27
test: handle warnings generated by our own tests

### LINKED ISSUE #1408: Ignore Harness PendingDeprecationWarnings in our own tests
Running the tests currently generates a lot of warnings along the lines of:

```
test/test_testing.py::TestHarness::test_relation_events
  /tmp/operator/test/test_testing.py:745: PendingDeprecationWarning: Harness is deprecated; we recommend using state transition testing (previously known as 'Scenario') instead
    harness = ops.testing.Harness(
```

Obviously, this is a case where we do not want to switch to Scenario instead. We should be able to ignore this warning for the entire test/test_testing module.

tonyandrewmeyer:
Our tests currently end with a very large list of warnings. This PR fixes that:

* All of the Harness tests emit a `PendingDeprecationWarning` telling us that Harness is going away and we should use Scenario. This seems like something to simply ignore in our own tests, so I've added a `-W` argument to Python to silence them in bulk.
* One `test_main` test raises a warning about controller storage going away - there's another test that tests that warning happens, and the two tests are otherwise identical. I've merged the two - we don't need them both.
* Two of the secret tests check that `SecretInfo` objects can still be created without providing the model UUID. I've added a `pytest.warns` to (a) ensure that there's the warning, and (b) stop it appearing in the tox output.
* Scenario should provide the model UUID when creating `SecretInfo` objects. This is a one-line fix, so I've added it here - it stops deprecation warnings, but is really more the code than the tests, but since it's so small I think it's ok here.
* The `context.on` Scenario tests check that pre- and post- series upgrade events work, and those raise warnings. I've split them off to a separate test that verifies (and silences) the warning.

Fixes #1408

### DIFF
--- a/testing/tests/test_context_on.py
+++ b/testing/tests/test_context_on.py
@@ -1,5 +1,4 @@
 import copy
-import warnings
 
 import ops
 import pytest
@@ -81,9 +80,7 @@ def test_simple_deprecated_events(event_name, event_kind):
     ctx = scenario.Context(ContextCharm, meta=META, actions=ACTIONS)
     # These look like:
     #   ctx.run(ctx.on.pre_series_upgrade(), state)
-    with warnings.catch_warnings(
-        record=False, action="ignore", category=DeprecationWarning
-    ):
+    with pytest.warns(DeprecationWarning):
         with ctx(getattr(ctx.on, event_name)(), scenario.State()) as mgr:
             mgr.run()
             assert len(mgr.charm.observed) == 2
--- a/testing/tests/test_context_on.py
+++ b/testing/tests/test_context_on.py
@@ -1,4 +1,5 @@
 import copy
+import warnings
 
 import ops
 import pytest
@@ -55,8 +56,6 @@ class ContextCharm(ops.CharmBase):
         ("update_status", ops.UpdateStatusEvent),
         ("config_changed", ops.ConfigChangedEvent),
         ("upgrade_charm", ops.UpgradeCharmEvent),
-        ("pre_series_upgrade", ops.PreSeriesUpgradeEvent),
-        ("post_series_upgrade", ops.PostSeriesUpgradeEvent),
         ("leader_elected", ops.LeaderElectedEvent),
     ],
 )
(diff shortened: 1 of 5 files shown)


testing/tests/test_context_on.py
```diff
@@ -71,6 +70,27 @@ def test_simple_events(event_name, event_kind):
...
+        ("post_series_upgrade", ops.PostSeriesUpgradeEvent),
+    ],
+)
+def test_simple_deprecated_events(event_name, event_kind):
+    ctx = scenario.Context(ContextCharm, meta=META, actions=ACTIONS)
+    # These look like:
+    #   ctx.run(ctx.on.pre_series_upgrade(), state)
+    with warnings.catch_warnings(
+        record=False, action="ignore", category=DeprecationWarning
+    ):
```
james-garner-canonical, line 86: Why `warnings.catch_warnings` instead of `pytest.warns` here?

james-garner-canonical (approved):

testing/tests/test_context_on.py
tonyandrewmeyer: No reason, really, just that this was the first of these I did and I naturally reached for `warnings.catch_warnings` and then was reminded about `pytest.warns` later when I did the other ones because there was already some use of it. I'll switch this over: I think pytest.warns is the more appropriate choice in tests.

james-garner-canonical:
Validate PR title: I really think that `tests` is a better commit type than `test`, because to me the latter always feels like "this is just a test, not a for real". But here we are ...

Unit tests: looks like `catch_warnings` fails on older pythons, I guess the `__init__` form changed? This makes me think, we should update CI to not have unit tests cancel each other -- it's possible for them to pass on some python versions and fail on others, and it's good to know which ones -- and they're cheap and fast to run so cancelling doesn't really save much anyway.

testing/tests/test_context_on.py
james-garner-canonical: Sweet, looks like it was breaking on some python versions anyway

tonyandrewmeyer:
> Validate PR title: I really think that `tests` is a better commit type than `test`, because to me the latter always feels like "this is just a test, not a for real". But here we are ...

I feel like I get this wrong about 50% of the time :laughing:.

> Unit tests: looks like `catch_warnings` fails on older pythons, I guess the `__init__` form changed?

Yeah, a bunch of parameters were added in 3.11, and I'd forgotten that. Maybe I ought to default to running tox with 3.8 locally - it's much more likely that it would catch things than using newer Ptyhon.

> This makes me think, we should update CI to not have unit tests cancel each other -- it's possible for them to pass on some python versions and fail on others, and it's good to know which ones -- and they're cheap and fast to run so cancelling doesn't really save much anyway.

Interesting thought - feel free to add it to the team discussion topics. I do agree that at times (like this one) it's more informative to immediately see "ah, this is bad on [3.8 or macOS or ..]". At other times, when something is just generally wrong (hopefully only drafts, if one is properly running tox locally) it does feel wasteful to do all the matrix when it's going to fail every time.

james-garner-canonical (approved):

dimaqq (approved):
I wish there was a more elegant way, but I guess given that Harness is deprecated, this will do.

The downside is that we won't notice if Harness is accidentally used in new tests, isn't it? I supposed that would get caught by human review.

IronCore864 (approved):

PR #1470, opened 2024-11-27
docs: use the right ops-scenario for building the docs

tonyandrewmeyer:
This PR fixes the version of ops-scenario used when building the docs.

Previously, the published version of ops-scenario was used for generating the documentation, and the generated requirements file was used for local builds but not for read-the-docs.

That breaks if there are local incompatible changes (such as renaming a module), and is generally wrong anyway.

With this PR, the local version is loaded, and read-the-docs uses the same requirements file as building locally. A few minor type annotations were adjusted to remove remaining warnings (I don't know why Sphinx couldn't figure out where `Path` was coming from, but keeping it in the `pathlib` namespace is generally nicer anyway).

Split out from #1467.

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -9,12 +9,12 @@ from __future__ import annotations
 import dataclasses
 import datetime
 import inspect
+import pathlib
 import random
 import re
 import string
 from enum import Enum
 from itertools import chain
-from pathlib import Path, PurePosixPath
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -747,9 +747,9 @@ class Exec(_max_posargs(1)):
 class Mount(_max_posargs(0)):
     """Maps local files to a :class:`Container` filesystem."""
 
-    location: str | PurePosixPath
+    location: str | pathlib.PurePosixPath
     """The location inside of the container."""
-    source: str | Path
+    source: str | pathlib.Path
     """The content to provide when the charm does :meth:`ops.Container.pull`."""
 
 
@@ -1017,7 +1017,7 @@ class Container(_max_posargs(1)):
             infos[name] = info
         return infos
 
-    def get_filesystem(self, ctx: Context) -> Path:
+    def get_filesystem(self, ctx: Context) -> pathlib.Path:
         """Simulated Pebble filesystem in this context.
 
         Returns:
@@ -1314,7 +1314,7 @@ class Storage(_max_posargs(1)):
             return (self.name, self.index) == (other.name, other.index)
         return False
 
-    def get_filesystem(self, ctx: Context) -> Path:
+    def get_filesystem(self, ctx: Context) -> pathlib.Path:
         """Simulated filesystem root in this context."""
         return ctx._get_storage_root(self.name, self.index)
 
@@ -1325,7 +1325,7 @@ class Resource(_max_posargs(0)):
 
     name: str
     """The name of the resource, as found in the charm metadata."""
-    path: str | Path
+    path: str | pathlib.Path
     """A local path that will be provided to the charm as the content of the resource."""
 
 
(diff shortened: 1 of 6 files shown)


testing/src/scenario/state.py
```diff
@@ -9,12 +9,12 @@
 import dataclasses
 import datetime
 import inspect
+import pathlib
 import random
 import re
 import string
 from enum import Enum
 from itertools import chain
-from pathlib import Path, PurePosixPath
```
james-garner-canonical, line 17: Sad to see this change. I guess if it's the only way to make Sphinx happy ... I know we generally prefer to keep imports in their namespace, but
1. `from pathlib import Path` is so idiomatic
2. I kind of like seeing what's being used from an import in the imports
Ah well, not a blocker

testing/src/scenario/state.py
```diff
@@ -747,9 +747,9 @@ def _run(self) -> int:
 class Mount(_max_posargs(0)):
     """Maps local files to a :class:`Container` filesystem."""
 
-    location: str | PurePosixPath
+    location: str | pathlib.PurePosixPath
     """The location inside of the container."""
```
james-garner-canonical, line 751: Do you know why we need `PurePosixPath` here btw? Anything worth documenting (possibly in a future PR). Not a blocker, just stood out to me, and I'm curious

tox.ini
```diff
@@ -40,6 +40,7 @@ description = Compile the requirements.txt file for docs
 deps = pip-tools
 commands =
     pip-compile --extra=docs -o docs/requirements.txt pyproject.toml
+    python -c 'open("docs/requirements.txt", "a").write("./testing/\n")'
```
james-garner-canonical, line 43: Interesting that testing is specified relative to the parent directory of docs, but I assume this is correct

james-garner-canonical (approved):

testing/src/scenario/state.py
tonyandrewmeyer: This pre-dates me, but I assume we're using a pure path here because the path doesn't actually exist on the testing system, only instead of the (mock) container. POSIX because we're assuming a non-Windows image.

Or if the question is why not just `str` I assume it's to give people more flexibility, for those that like a plain `/charm-stuff/blah.txt` vs. those that like only using path objects. We could take a stand on that and only accept one, I imagine, but not until the next major version at this point.

tox.ini
tonyandrewmeyer: The actual `pip install` is done at the root level, so the path is relative to that. It does work locally (you can test with `tox -e docs`, optionally doing `tox -e docs-deps` first) and on rtd (you can look at the link from the passing CI step).

testing/src/scenario/state.py
tonyandrewmeyer: For what it's worth, I am very much on the side of it being an improvement anyway (a discussion for another day, I imagine) so I didn't look very hard at getting Sphinx to be happy with it as-is.

If you feel strongly about it, then I could try to see why Sphinx thought it was something defined in `scenario.state` and couldn't detect that it was coming from the standard lib.

testing/src/scenario/state.py
james-garner-canonical: Sorry for the ambiguity -- your first paragraph answers my question. Thanks. No documentation update should be needed as users will be aware that it's a mock container, and the `PurePosixPath` will rebel if people accidentally try to use concrete path operations. Resolving this.

testing/src/scenario/state.py
james-garner-canonical: All good, I don't think it's worth spending further time on this. Keen to chat about import styles another day though

dimaqq (approved):

PR #1471, opened 2024-11-28
ci: don't cancel other unit tests when one fails

james-garner-canonical:
A test can pass on one python/os version and fail on others, and it’s nice to see that in the outcome, rather than dig into cancelled tests which may or may not have hit the failing case.

Since unit tests are cheap and fast, we’re not really saving much by cancelling them – by the time one unit test run fails and asks the others to cancel, they’ll already have been spun up and run many of the tests, so cancelling them doesn't save much.

The pip-install and test-real-pebble jobs are faster than the test job, so this commit adds fail-fast: false to them too.

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -43,6 +43,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: ["3.8", "3.10", "3.12"]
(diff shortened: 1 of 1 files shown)


dimaqq (approved):

tonyandrewmeyer (approved):

benhoyt (approved):

james-garner-canonical:
We can wait till after the release to merge this

tonyandrewmeyer:
> We can wait till after the release to merge this

@james-garner-canonical want to push the big green button now? :smile:

dimaqq:
Or now :)

PR #1474, opened 2024-11-28
chore: run pre-commit on all files

dimaqq:
Apparently pre-commit was not run on a few occasions and here's the diff.

I'm not sure about dangling whitespace in the pebble test.

### DIFF
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -1810,7 +1810,7 @@ class TestBreakpoint:
             '-',
             '...foo',
             'foo.bar',
-            'bar--' 'FOO',
+            'bar--FOO',
             'FooBar',
             'foo bar',
             'foo_bar',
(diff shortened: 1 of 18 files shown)


dimaqq:
P.S. pre-commit also reports a few unused `# noqa: ...` that it doesn't fix automatically.

james-garner-canonical:
Definitely seems problematic that running pre-commit with our config can break our tests

test/test_framework.py
```diff
@@ -1810,7 +1810,7 @@ def test_breakpoint_good_names(self, request: pytest.FixtureRequest, name: str):
             '-',
             '...foo',
             'foo.bar',
-            'bar--' 'FOO',
+            'bar--FOO',
```
dimaqq, line 1813: I wonder if this was a bug, perhaps `'bar--', 'FOO'` was meant?

tonyandrewmeyer: I think so, looking at the [original commit](https://github.com/canonical/operator/commit/11a1849205d750e28aaa4a13938b5864659f928b). Nice find!

dimaqq:
I'll redo this from scratch at a later date.

dimaqq:
James took over this with from-future work.

PR #1475, opened 2024-11-28
chore: fix skipping disabled observability charms

dimaqq:
Took 5 attempts, but I got the syntax right after all.


tonyandrewmeyer (approved):
Assuming correct. Thanks!

james-garner-canonical (approved):
I see no reason to think that the syntax is incorrect this time

PR #1477, opened 2024-11-29
ci: add support for injecting the latest ops when uv is used

tonyandrewmeyer:
When `uv` is used (as with self-signed-certificates-operator), the published-charms-tests workflow doesn't know how to inject the latest version of ops (there's no requirements.txt file).

This PR updates the workflow to look for `uv.lock` and if it exists then it uses `uv` to change the `ops` source in `pyproject.toml`.

self-signed-certificates-operator doesn't use `uv` to do the install in `tox` so we have to modify the raw sources list rather than have it be uv-specific. However, this should work even for projects that do, since the raw list is the originator of the dependencies - we might need to re-lock if other projects handle this differently, but we don't for now, so I've avoided that.

I've tested this manually, and have it manually against the branch (where it did install ops from my branch).

### DIFF
--- a/.github/workflows/published-charms-tests.yaml
+++ b/.github/workflows/published-charms-tests.yaml
@@ -60,7 +60,7 @@ jobs:
           repository: ${{ matrix.charm-repo }}
 
       - name: Install patch dependencies
-        run: pip install poetry~=1.6
+        run: pip install poetry~=1.6 uv~=0.5
 
       - name: Update 'ops' dependency in test charm to latest
         run: |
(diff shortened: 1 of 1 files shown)


james-garner-canonical (approved):
Comparing the failures on your [manual run](https://github.com/tonyandrewmeyer/operator/actions/runs/12077415738) to the [previous one](https://github.com/tonyandrewmeyer/operator/actions/runs/12001552009), it looks like `self-signed-certificates-operator` passes now, and there are no new failures

dimaqq (approved):

PR #1478, opened 2024-11-29
chore: update and clarify release process

dwilding:
Changes:
- Drop "ops" from the release title - we don't seem to include "ops" in practice
- Use ops[testing] in the release title in case of a release that includes ops[testing]
- Choose the previous tag after setting the new tag - the GitHub UI forces this order
- Exclude chores from CHANGES.md - we clarified that decision yesterday
- For an ops-only release, recommend to store any scenario-only changes at the top of testing/CHANGES.md
- Recommend to add backticks for code formatting in CHANGES.md

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -327,9 +327,9 @@ To make a release of the `ops` and/or `ops-scenario` packages, do the following:
 2. Visit the [releases page on GitHub](https://github.com/canonical/operator/releases).
 3. Click "Draft a new release"
 4. The "Release Title" is the full version numbers of ops and/or ops-scenario,
-   in the form `ops <major>.<minor>.<patch> and ops-scenario <major>.<minor>.<patch>`
+   in the form `<major>.<minor>.<patch> and ops[testing] <major>.<minor>.<patch>`
    and a brief summary of the main changes in the release.
-   For example: `ops 2.3.12 Bug fixes for the Juju foobar feature when using Python 3.12`
+   For example: `2.3.12 Bug fixes for the Juju foobar feature when using Python 3.12`
 5. If the last release was for both `ops` and `ops-scenario`, leave the previous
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -327,22 +327,22 @@ To make a release of the `ops` and/or `ops-scenario` packages, do the following:
 2. Visit the [releases page on GitHub](https://github.com/canonical/operator/releases).
 3. Click "Draft a new release"
 4. The "Release Title" is the full version numbers of ops and/or ops-scenario,
-   in the form `ops <major>.<minor>.<patch> and ops-scenario <major>.<minor>.<patch>`
+   in the form `<major>.<minor>.<patch> and ops[testing] <major>.<minor>.<patch>`
```
tonyandrewmeyer, line 330: The catch here is that the package we are releasing is called `ops` or `ops-scenario`. Using "ops[testing]" here seems misleading.

In terms of leaving "ops" out when we're releasing both (or, I assume, only ops), I'm reluctantly on board with that, given that "ops" is the thing that brings together what's in this repo, and the most expected package.

benhoyt: Yeah, for this I'd probably stick with what we have given these are the PyPI package names.

benhoyt (approved):

HACKING.md
dwilding: If both ops and ops-scenario are being released (which should usually be the case, right?) then shall we go with option 1 or option 2:
1. `ops <major>.<minor>.<patch> and ops-scenario <major>.<minor>.<patch>`
2. `<major>.<minor>.<patch> and ops-scenario <major>.<minor>.<patch>` 

The difference being whether we mention `ops` at the beginning

benhoyt: Decision is to go with option 1 for clarity.

Issue #1434, opened 2024-10-11
Optimise Scenario 


PietroPasotti:
The [unit testing suite for traefik](https://github.com/canonical/traefik-k8s-operator/tree/main/tests/scenario) is probably the largest scenario test battery around.
Today I ran it and I realized it took over a minute and a half to complete, so I decided to profile it.

This is the result: 
![image](https://github.com/user-attachments/assets/0185dbe7-63ef-4cc7-92d2-cdc26f50123e)

to produce the profile, run with this tox env:

```
[testenv:profile-scenario-tests]
description = Profile scenario unit test battery
deps =
    pytest
    pytest-profiling
    ops[testing]>=2.17
    -r{toxinidir}/requirements.txt
commands =
    pytest -v --tb native {[vars]tst_path}/scenario --profile --profile-svg --log-cli-level=INFO -s {posargs}
```

There are some obvious candidates for optimization:
- using an in-memory mock for the sqlite db instead of using the real thing could shave off a good chunk of time spent in pointless IO
- A ridiculous amount of time is spent in `State.__new__`. Can we do something about that?
- how come mocking `juju-log` takes so long?
- A single list comprehension in state.py:143 takes 2 seconds of our time: can we lazify some of the code perhaps?

profiling scenario's own test suite yields a very similar picture:
![image](https://github.com/user-attachments/assets/537114b4-2d6b-4467-b54c-0e06d94ecae2)

tonyandrewmeyer:
> * using an in-memory mock for the sqlite db instead of using the real thing could shave off a good chunk of time spent in pointless IO

This indeed seems like an obvious win. Presumably we can just pass in `:memory:` instead of the unit state filename - trivial to do, and we're not testing sqlite so no downsides.

> * A ridiculous amount of time is spent in `State.__new__`. Can we do something about that?
> * A single list comprehension in state.py:143 takes 2 seconds of our time: can we lazify some of the code perhaps?

Both of these are the positional/keyword argument processing. I wonder if we can find a tidy way to have newer Python use the built-in support and only fall back to this on old Python.

> * how come mocking `juju-log` takes so long?

I wonder if this is also the same issue: lots of log lines and each one involves creating a `JujuLogLine` object.

benhoyt:
> Presumably we can just pass in :memory: instead of the unit state filename - trivial to do, and we're not testing sqlite so no downsides.

[...]

PR #1494, opened 2024-12-09
feat: use :memory: as the unit state storage location for ops.testing

tonyandrewmeyer:
This PR changes Scenario to always use `:memory:` as the location for the unit state database, to speed up running tests.

In most cases, the charm root location is a `TemporaryDirectory`, so the unit state database only exists during the event emission anyway. The database could be accessed if a specific root directory was provided (or inside the context manager context) but the correct way to interact with the database is through the State API (deferred events, saved state objects) so we don't want to leave that open for users.

Timing (best of 3):

|Suite|main|branch|
|-|-|-|
|operator unit tests (no concurrency)|4m11.923s|3m41.354s|
|traefik scenario tests|3m49.479s|1m56.795s|
|kafka-k8s-operator scenario tests|0m13.463s|0m13.236s|

Note that this [is already the case for Harness](https://github.com/canonical/operator/blob/0ad731a6a9118dfac240e84681f5c41810c4f3f2/ops/_private/harness.py#L309).

Refs #1434

### DIFF
--- a/testing/src/scenario/_ops_main_mock.py
+++ b/testing/src/scenario/_ops_main_mock.py
@@ -143,9 +143,7 @@ class Ops(_Manager):
 
     def _make_storage(self, _: _Dispatcher):
         # TODO: add use_juju_for_storage support
-        # TODO: Pass a charm_state_path that is ':memory:' when appropriate.
-        charm_state_path = self._charm_root / self._charm_state_path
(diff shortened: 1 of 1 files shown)


james-garner-canonical (approved):
Makes sense! And a big performance boost for traefik's tests it seems

dimaqq (approved):

PR #1496, opened 2024-12-10
fix: remove ops.main.main deprecation warning, and avoid warnings in action output

### LINKED ISSUE #1460: should actions leak deprecation warnings?
In my view, `juju actions` are the public facade of the charm. The user-facing API. 
A user doesn't necessarily want to see things such as internal deprecation warnings or other logs as part of the action output, or they might think something went wrong when running the action.

I'm not sure if this is a juju or an ops concern, but this user experience should be improved:

image

What do you think is the way forward on this? Silence all stdout when running an action? This warning will show up in the juju debug-log anyway, which is the only place I'd expect to see it.

tonyandrewmeyer:
Background:
* A `warnings.warn` call outputs to stderr by default.
* Juju captures stderr and when executing an action provides it to the user.
* We emit a DeprecationWarning when charms call `main()` other than `import ops ... ops.main()`
* DeprecationWarnings are ignored by default, *except* for issues in the `__main__` module.
* A warning like this is meant to be discovered during tests - but unit tests typically do not call `main()` or execute the `charm.py` file, so they will not trigger this warning. That means that we're relying on integration tests.

The above means that, currently, the deprecation warning is included in the action output - this warning is intended for the developer of the charm, not the user of the charm, so we want to change this.

Firstly, the `ops.main.main` deprecation warning is removed. It's too awkward to have a warning that gets emitted before the framework is set up. Instead, we'll rely on `charmcraft analyse` and static checks.

However, while that will solve the specific issue in #1460, we do also want to avoid any future leakage of warnings into action output if there are other warnings emitted that originate in the `__main__` model.

To that, rather than rely on Juju capturing the warning via stderr, we use the logging package to capture all warnings and redirect them to logging calls. These are then directed to Juju's debug-log via the existing mechanism. While we are doing this, we minimise the output to include references to the filename and line, but not the actual code itself (which may be sensitive).

When running (Scenario) tests, we ensure that the warning and logging systems use their default behaviour, so that pytest and other tools can work as expected.

Fixes #1460

### DIFF
--- a/ops/log.py
+++ b/ops/log.py
@@ -60,7 +60,6 @@ def setup_root_logging(
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
-    logging.captureWarnings(True)
 
     def custom_showwarning(
         message: typing.Union[str, Warning],
@@ -68,9 +67,10 @@ def setup_root_logging(
         filename: str,
         lineno: int,
         *_: typing.Any,
-    ) -> str:
-        """Like the default showwarning, but don't include the code."""
-        return f'{filename}:{lineno}: {category.__name__}: {message}'
+        **__: typing.Any,
+    ):
+        """Direct the warning to Juju's debug-log, and don't include the code."""
+        logger.warning('%s:%s: %s: %s', filename, lineno, category.__name__, message)
 
     warnings.showwarning = custom_showwarning
 
--- a/ops/log.py
+++ b/ops/log.py
@@ -23,7 +23,7 @@ from ops.model import _ModelBackend
 
 
 class JujuLogHandler(logging.Handler):
-    """A handler for sending logs and warnings to Juju via juju-log."""
+    """A handler for sending logs to Juju via juju-log."""
 
     def __init__(self, model_backend: _ModelBackend, level: int = logging.DEBUG):
         super().__init__(level)
--- a/ops/log.py
+++ b/ops/log.py
@@ -18,12 +18,18 @@ import logging
 import sys
 import types
 import typing
+import warnings
 
 from ops.model import _ModelBackend
 
 
+# We do this on module import because some warnings are issued before we set up
+# the framework, and we need to capture those as well.
+logging.captureWarnings(True)
+
+
 class JujuLogHandler(logging.Handler):
-    """A handler for sending logs to Juju via juju-log."""
+    """A handler for sending logs and warnings to Juju via juju-log."""
 
     def __init__(self, model_backend: _ModelBackend, level: int = logging.DEBUG):
         super().__init__(level)
@@ -41,12 +47,15 @@ class JujuLogHandler(logging.Handler):
 def setup_root_logging(
     model_backend: _ModelBackend, debug: bool = False, exc_stderr: bool = False
 ):
-    """Setup python logging to forward messages to juju-log.
+    """Setup Python logging to forward messages to juju-log.
 
     By default, logging is set to DEBUG level, and messages will be filtered by Juju.
     Charmers can also set their own default log level with::
 
       logging.getLogger().setLevel(logging.INFO)
+
+    Warnings issued by the warnings module are redirected to the logging system
+    and forwarded to juju-log, too.
 
     Args:
         model_backend: a ModelBackend to use for juju-log
@@ -56,6 +65,19 @@ def setup_root_logging(
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
+
+    def custom_warning_formatter(
+        message: typing.Union[str, Warning],
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        _: typing.Optional[str] = None,
+    ) -> str:
+        """Like the default formatter, but don't include the code."""
+        return f'{filename}:{lineno}: {category.__name__}: {message}'
+
+    warnings.formatwarning = custom_warning_formatter
+
     if debug:
         handler = logging.StreamHandler()
         formatter = logging.Formatter('%(asctime)s %(levelname)-8s %(message)s')
--- a/ops/log.py
+++ b/ops/log.py
@@ -59,7 +59,6 @@ def setup_root_logging(
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
-
     if debug:
         handler = logging.StreamHandler()
         formatter = logging.Formatter('%(asctime)s %(levelname)-8s %(message)s')
--- a/ops/log.py
+++ b/ops/log.py
@@ -62,12 +62,12 @@ def setup_root_logging(
     logger.addHandler(JujuLogHandler(model_backend))
 
     def custom_showwarning(
-        message: typing.Union[str, Warning],
+        message: typing.Union[Warning, str],
         category: typing.Type[Warning],
         filename: str,
         lineno: int,
-        *_: typing.Any,
-        **__: typing.Any,
+        file: typing.Optional[typing.TextIO] = None,
+        line: typing.Optional[str] = None,
     ):
         """Direct the warning to Juju's debug-log, and don't include the code."""
         logger.warning('%s:%s: %s: %s', filename, lineno, category.__name__, message)
--- a/ops/log.py
+++ b/ops/log.py
@@ -23,11 +23,6 @@ import warnings
 from ops.model import _ModelBackend
 
 
-# We do this on module import because some warnings are issued before we set up
-# the framework, and we need to capture those as well.
-logging.captureWarnings(True)
-
-
 class JujuLogHandler(logging.Handler):
     """A handler for sending logs and warnings to Juju via juju-log."""
 
@@ -65,18 +60,19 @@ def setup_root_logging(
     logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(JujuLogHandler(model_backend))
+    logging.captureWarnings(True)
 
-    def custom_warning_formatter(
+    def custom_showwarning(
         message: typing.Union[str, Warning],
         category: typing.Type[Warning],
         filename: str,
         lineno: int,
-        _: typing.Optional[str] = None,
+        *_: typing.Any,
     ) -> str:
-        """Like the default formatter, but don't include the code."""
+        """Like the default showwarning, but don't include the code."""
         return f'{filename}:{lineno}: {category.__name__}: {message}'
 
-    warnings.formatwarning = custom_warning_formatter
+    warnings.showwarning = custom_showwarning
 
     if debug:
         handler = logging.StreamHandler()
--- a/ops/log.py
+++ b/ops/log.py
@@ -18,14 +18,8 @@ import logging
 import sys
 import types
 import typing
-import warnings
 
 from ops.model import _ModelBackend
-
-
-# We do this on module import because some warnings are issued before we set up
-# the framework, and we need to capture those as well.
-logging.captureWarnings(True)
 
 
 class JujuLogHandler(logging.Handler):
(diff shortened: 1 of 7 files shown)


tonyandrewmeyer:
> The PR uses the second option, which seems cleaner - the main drawback is that if someone is running the charm with a custom warning filter, such as `-W error::DeprecationWarning`, then that won't match this warning.

Thinking on this more, I think it should be the other one. I'll switch over (and fix whatever the test issue is - maybe 3.8 compatibility since they passed locally) and un-draft. No need to review until then.

ops/main.py
```diff
@@ -37,9 +48,37 @@ def main(charm_class: Type[ops.charm.CharmBase], use_juju_for_storage: Optional[
...
+    # will get fixed before going into production.
+    frame = _top_frame()
+    assert frame is not None
+
+    class DeprecatedMainCharmBase(charm_class):
+        def __init__(self, *args: Any, **kwargs: Any):
+            super().__init__(*args, **kwargs)
+            logger.warning(
+                '%s:%s: DeprecationWarning: Calling `ops.main()` is deprecated, '
+                'call `ops.main()` instead',
```
dimaqq, line 77: This part is hilarious :)

ops/main.py
```diff
@@ -37,9 +48,37 @@ def main(charm_class: Type[ops.charm.CharmBase], use_juju_for_storage: Optional[
...
+    # and line number - we could wrap this in a custom formatter, but it's
+    # simpler to just use a logging.warning() call instead.
+    # Note that unit tests almost never call main(), so wouldn't show this
+    # warning, so we're relying on integration tests, which aren't going to be
+    # treating warning calls in any special way. That means that we can just
+    # output the warning directly as a log message, and the effect will be the
+    # same.
+    # Note also that this will be logged with every event. Our assumption is
+    # that this will be noticeable enough during integration testing that it
+    # will get fixed before going into production.
```
dimaqq, line 68: I'd like to see the assumptions in this comment validated, and the comment rewritten to maybe 1/3 of the size.

ops/main.py
```diff
@@ -37,9 +48,37 @@ def main(charm_class: Type[ops.charm.CharmBase], use_juju_for_storage: Optional[
...
+            super().__init__(*args, **kwargs)
+            logger.warning(
+                '%s:%s: DeprecationWarning: Calling `ops.main()` is deprecated, '
+                'call `ops.main()` instead',
+                frame.f_code.co_filename,
+                frame.f_lineno,
+            )
+
+    return _main.main(
+        charm_class=DeprecatedMainCharmBase, use_juju_for_storage=use_juju_for_storage
```
dimaqq, line 83: If it's used in this way, the new class should have a different name.

Alternatively, we could declare a new charm class dynamically with name being upstream name + some hack, and the deprecated base (or mixin?) added in.

ops/main.py
tonyandrewmeyer: Hey - maybe you missed the comment on the PR: I've switched it to draft while I adjust some things, and no need to review until then (moving to draft doesn't unsubscribe so de-selecting reviewers doesn't really do anything unfortunately).

ops/main.py
```diff
@@ -37,9 +48,37 @@ def main(charm_class: Type[ops.charm.CharmBase], use_juju_for_storage: Optional[
...
+    # Note that unit tests almost never call main(), so wouldn't show this
+    # warning, so we're relying on integration tests, which aren't going to be
+    # treating warning calls in any special way. That means that we can just
+    # output the warning directly as a log message, and the effect will be the
+    # same.
+    # Note also that this will be logged with every event. Our assumption is
+    # that this will be noticeable enough during integration testing that it
+    # will get fixed before going into production.
+    frame = _top_frame()
+    assert frame is not None
```
dimaqq, line 70: I'm a little confused how this is useful.

Given that these lines are evaluated at runtime for every legacy charm that uses newer ops, we surely expect the assert to pass. But then what is the assert for?

ops/main.py
dimaqq: Sorry for the early fly-by, I'm cleaning up my github backlog 🙇🏻 

In any case, I wanted to understand the 2 alternatives in the PR description.

dimaqq:
> The two options for this are:
> 
> - Customise the warnings formatting, just around this specific warning.
> - Output the warning in another way (printing or logging, for example).

I think there's a 3rd option: remove the warning entirely, and solve the ops.main.main -> ops.main some other way:
- better type hints (?)
- charmcraft lint
- github search, leading to opening PRs to charmers' repos
- or something

Let's brainstorm at the next standup, perhaps?

tonyandrewmeyer:
> I think there's a 3rd option: remove the warning entirely

I did think about this too. This one in particular is complicated because it's "outside" the charm, so it doesn't get picked up in unit tests and is also complicated to code around because the framework isn't set up yet.

I do think that we should have a PR that prevents the general issue of DeprecationWarnings that caused by code in src/charm.py ending up in action output. But we could potentially do that and also give up on this specific warning in favour of other approaches.

benhoyt:
Decision from Charm Tech is to remove the warning altogether to keep it simple. And then Charm Tech (whoever works on this) can open 2 or 3 PRs jubilantly (!) to fix this in the "culprit teams".

ops/main.py
tonyandrewmeyer: It was for pyright, because it couldn't infer that it would always be non-None in our case. But removed now.

ops/main.py
tonyandrewmeyer: The class did have a new name. I guess it could have been dynamic rather than fixed, but I'm not sure whether that would really have any impact. But gone now anyway.

ops/main.py
tonyandrewmeyer: Introduced by mistake :laughing:. But not relevant now anyway.

tonyandrewmeyer:
> And then Charm Tech (whoever works on this) can open 2 or 3 PRs jubilantly (!) to fix this in the "culprit teams".

Yoikes:

```
tameyer@tam-canoncial-1:~/code/charm-analysis/.cache$ rg 'from ops.main import main' -l | wc -l
312
```

benhoyt:
> Yoikes:
>
>     tameyer@tam-canoncial-1:~/code/charm-analysis/.cache$ rg 'from ops.main import main' -l | wc -l
>     312

Ouch. We could still let 2 or 3 teams know with PRs, but yeah, with those numbers, maybe that's not a good approach. I guess just a polite message in Charm Development and leave it at that?

Out of interest, are there 312 charms in your dir now, or does this occur more than once per charm?

tonyandrewmeyer:
> Out of interest, are there 312 charms in your dir now, or does this occur more than once per charm?

More than once. Some are documentation, but that really ought to get fixed too. Some are in libs, so there's some duplication that would remove the number of PRs, but be more annoying to make sure the change is in the right PR. Some repos also have quite a few charms in them, which ups the volume.

```
$ rg 'from ops.main import main' -l | grep -v 'lib/charms' | grep -v 'tests/integration' | wc -l
161
```

ops/log.py
```diff
@@ -56,6 +60,20 @@ def setup_root_logging(
...
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        *_: typing.Any,
+        **__: typing.Any,
+    ):
+        """Direct the warning to Juju's debug-log, and don't include the code."""
+        logger.warning('%s:%s: %s: %s', filename, lineno, category.__name__, message)
+
+    warnings.showwarning = custom_showwarning
```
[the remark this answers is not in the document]
tonyandrewmeyer: > Would this suppress warnings from third-party libraries in charmers' tests?

This doesn't really suppress the warnings, it just changes them to go through the logging system (I suppose if your log level is higher than `warning` then they are effectively suppressed, but I would argue that you have asked for that to happen).

This would apply to any warnings including third-party libraries, yes. However, I think that's the behaviour we want - if some third-party library generated a warning during an action, then I think that should end up in the Juju debug-log (or Loki or anywhere else people have directed the logging to) and should *not* end up in the action output.

This wouldn't happen with charm *unit* tests, because we (at least in Scenario) have a custom log/warning setup there - we don't do the special-casing for actions, and the debug-log entries end up in a side-effect collection in the context object. It would happen with *integration* tests but I'd argue that's what we want: the same behaviour as when the charm is run in production. I don't think people notice warnings in integration tests, but I think they often do in unit tests, so that's where it's important they show up.

If charms run their integration tests so that the charm is run with `-W error` (last time I looked, no-one does this, but I think it would be a good idea) then it does still apply (I manually checked this case when creating the PR), so the charm would crash, failing the test.

ops/log.py
```diff
@@ -56,6 +60,20 @@ def setup_root_logging(
...
+    def custom_showwarning(
+        message: typing.Union[str, Warning],
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        *_: typing.Any,
+        **__: typing.Any,
+    ):
+        """Direct the warning to Juju's debug-log, and don't include the code."""
+        logger.warning('%s:%s: %s: %s', filename, lineno, category.__name__, message)
```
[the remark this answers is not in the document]
tonyandrewmeyer: I don't like unstructured log entries that have newlines in them, and I don't think it's ideal to have code snippets in the Juju debug log. However, we do both of those for exceptions already, so perhaps that doesn't matter. However, I don't think it's that *useful* for warnings. The default warnings formatter looks like:

```
/tmp/fw.py:4: DeprecationWarning: Don't call foo(), use the new bar() instead
  warnings.warn("Don't call foo(), use the new bar() instead", DeprecationWarning)
```

That second line is telling me nothing that the first line hasn't already. Maybe the line has something other than the `warnings.warn` call that provides extra context, but in my experience that's unusual.

That would currently show something like:

```
unit-my-charm-0: 10:07:20 INFO unit.my-charm/0.juju-log charm.py:42: DeprecationWarning: Don't call foo(), use the new bar() instead
```

I feel like that's a better form for an unstructured log message.

ops/log.py
```diff
@@ -56,6 +60,20 @@ def setup_root_logging(
...
     logger.addHandler(JujuLogHandler(model_backend))
+
+    def custom_showwarning(
+        message: typing.Union[str, Warning],
+        category: typing.Type[Warning],
+        filename: str,
+        lineno: int,
+        *_: typing.Any,
+        **__: typing.Any,
+    ):
```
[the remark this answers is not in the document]
tonyandrewmeyer: > `warnings.showwarning` doesn't accept any more arguments, the `*_, **__` should be unused.

Well, it does accept `file` (the `*` isn't needed indeed, that was wrong).

> What am I missing here, potential future Python feature?

No, that I don't make use of `file`.

You're right that it's better to match the actual signature - maybe there is some positional argument usage that would break here, or lack of providing a value for line.

ops/log.py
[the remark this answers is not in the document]
dimaqq: I guess that depends on whether the end user looks at single messages are series of messages.

`Don't call foo(), use the new bar() instead` is great if there's a single call to foo(), and not so much if there are hundreds, or if there's no obvious call, as in a case of being called via `getattr(<computed>)()`.

I can see benefits for both alternatives, I guess.

test/test_main_invocation.py
```diff
@@ -52,9 +52,6 @@ def test_top_level_import(charm_env: None):
 def test_top_level_import_legacy_call(charm_env: None):
     import ops
```
dimaqq, line 54: My original intention was to actually test `ops.main.main(...)` as that remains a supported, albeit deprecated way.

I'd rather see `with...deprecated` removed, but `ops.main.main(CharmBase)` remain.

dimaqq (approved):

ops/log.py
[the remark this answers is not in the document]
tonyandrewmeyer: > `Don't call foo(), use the new bar() instead` is great if there's a single call to foo(), and not so much if there are hundreds, or if there's no obvious call, as in a case of being called via `getattr(<computed>)()`.

But you still have the filename and the line number. You don't have the exact line of code, but assuming you have access to it, it's trivial to open that file and go to that line number. You don't get a full traceback with the default formatter, so you only have the line the warning was pointing at - which is hopefully using a stacklevel that is one up from the actual `warning.warn` call.

For example:

```python
# charm.py
class MyCharm(ops.CharmBase):
    ...
    def _on_something(self, event):
        self.foo() # first one
        ...

    def _on_something_else(self, event):
        self.foo() # second one

    def foo(self):
        warnings.warn("don't foo(), do bar()", DeprecationWarning, stacklevel=2)
```

When `_on_something` is called, there will be a Juju log entry that points you at charm.py to the line with "# first one", and when `on_something_else` is called, there will be a Juju log entry that points you at charm.py to the line with "# second one".

If we include the actual code, like the default formatter, then you'd also have in the Juju debug log those actual lines of code, but I don't think this really gains you much (especially when it's a warning aimed at the developer, who will have access to the code) - and it saves having a multi-line log entry.

tonyandrewmeyer:
@benhoyt friendly reminder about this PR in case it's slipped off your radar :)

benhoyt (approved):
Yes, it had indeed slipped off my radar. I'm happy with this -- thanks!

Issue #1492, opened 2024-12-09
cleanup testing.Context signature


PietroPasotti:
Scenario's Context object was designed with the old 'metadata/actions/config.yaml`s in mind.
Nowadays we have a single charmcraft.yaml, which makes the signature obsolete and confusing.
I recently helped someone debug this code:
```
def test_foo():
    ctx = Context(MyCharm, config={"foo": "bar"})
    ctx.run(..., State())
```

which failed with a strange AttributeError caused by the charm attempting to `observe(self.on.foo_relation_changed)`.
The reason is that Context will only attempt to autoload meta/actions/config from the yaml files IF the user doesn't specify any of the `meta, config, actions` Context params.

The fix was:
```
def test_foo():
    ctx = Context(MyCharm)
    ctx.run(..., State(config={"foo": "bar"}))
```

But it was not obvious.

It's not clear that the 'config' param in Context doesn't mean "juju config" but "the config section in charmcraft.yaml".
Can we improve this without breaking backwards compatibility?

tonyandrewmeyer:
Some possible alternatives to #1497:

### Just warn

Something like this (with some of the better docs from #1497, a mention in the Context docs, and a test):

```diff
diff --git a/testing/src/scenario/context.py b/testing/src/scenario/context.py
index 411beed9..5f053527 100644
--- a/testing/src/scenario/context.py
+++ b/testing/src/scenario/context.py
@@ -510,6 +510,11 @@ class Context(Generic[CharmType]):
         else:
             if not meta:
                 meta = {"name": str(charm_type.__name__)}
+                warnings.warn(
+                    "No meta provided: only the charm name is set.",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
```

### Deprecate all three, add from_yaml to replace

Emit deprecation warnings if `meta`, `actions`, or `config` are provided and add something like:

```python
class Context[...]:
    ...
    @classmethod
    def from_yaml(charm_type: ..., charmcraft_yaml: dict) -> Context[...]:
        # private attributes to pass the things, or do it post __init__ or whatever
        return charm_type()

    @classmethod
    def from_legacy_yaml(charm_type: ..., meta: dict, *, config: dict|None, actions: dict|None) -> Context[...]:
        ...
```

Although the `from_` methods would need to also handle passing in other arguments for the `Context` `__init__` I guess.

### Deprecate all three, use charmcraft to replace

[...]

Issue #1424, opened 2024-10-10
Add support for passing charmcraft.yaml as metadata to testing.Context


tonyandrewmeyer:
## Description

Charmcraft.yaml (replacing metadata.yaml) now removes the need to have different files for metadata, config, and actions. When providing metadata in the charmcraft.yaml format, we get a `InconsistentScenarioError` error because the config option is not recognized.

```python
METADATA = yaml.safe_load(Path("tests/unit/charms/tls_certificates_interface/v4/dummy_requirer_charm/charmcraft.yaml").read_text())
state_in = scenario.State(
    config={"common_name": "example.com"},
)
ctx = scenario.Context(
    charm_type=DummyTLSCertificatesRequirerCharm,
    meta=METADATA,
)
```
## Logs

```
scenario.runtime.InconsistentScenarioError: Inconsistent scenario. The following errors were found: config option 'common_name' in state.config but not specified in config.yaml.
```

## Desired state

```python
self.ctx = scenario.Context(
    charm_type=DummyTLSCertificatesRequirerCharm,
    meta=METADATA,
)
```

## Workaround

```python
self.ctx = scenario.Context(
    charm_type=DummyTLSCertificatesRequirerCharm,
    meta=METADATA,
    config=METADATA["config"],
    actions=METADATA["actions"],
)
```

Moved from https://github.com/canonical/ops-scenario/issues/156

tonyandrewmeyer:
Aim to overload the `meta` argument to handle both metadata.yaml content and charmcraft.yaml content.

dimaqq:
Discussion:

How about we do something like this:

```py
# class ops.testing.Context
def __init__(self, charm_type, meta=None, *, config=None, actions=None):
    if config is None and meta is not None:
        config = meta.get("config")
    if actions is None and meta is not None:
        actions = meta.get("actions")
```

Observations:
* if you pass any of meta/config/actions, magic auto-loading is switched off, this approach is consistent with that.
* we’re relying on how charcraft splits singular charmcraft.yaml file into metadata/config/actions.yaml
* `meta` is a positional argument
* not introducing new arguments now is great
* @dimaqq: if meta is a large dict with non-metadata things like bases, then we are assuming that it’s actually the (new-style, complete, singular) charmcraft.yaml file

- [ ] we’d need to update the docstring with the above
- [ ] update our tests, cover the functionality above
- [ ] go through "how to manage config/actions" to see if the text needs an update

PR #1497, opened 2024-12-10
added error if meta not passed when config or actions are

PietroPasotti:
My 2c on a possible mitigation for #1492

### DIFF
--- a/testing/src/scenario/context.py
+++ b/testing/src/scenario/context.py
@@ -454,6 +454,10 @@ class Context(Generic[CharmType]):
     Use this when calling :meth:`run` to specify the event to emit.
     """
 
+    # def test_foo():
+    #     ctx = Context(MyCharm, config={"foo": "bar"})
+    #     ctx.run(..., State())
+
     def __init__(
         self,
         charm_type: type[CharmType],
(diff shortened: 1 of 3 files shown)


testing/src/scenario/context.py
```diff
@@ -503,13 +507,18 @@ def __init__(
...
+                    "Please parse and provide their contents manually via the `meta`, `config` "
+                    "and `actions` Context parameters.",
                 ) from e
 
         else:
             if not meta:
-                meta = {"name": str(charm_type.__name__)}
+                raise ContextSetupError(
+                    "Scenario doesn't support overriding `actions|config` without overriding `meta` too."
+                )
```
tonyandrewmeyer, line 521: And I suppose here also :disappointed: 

```suggestion
                raise ContextSetupError(
                    "To override `actions|config` you must also override `meta`."
                )
```

testing/tests/test_charm_spec_autoload.py
```diff
@@ -126,6 +126,23 @@ def test_no_meta_raises(tmp_path, legacy):
...
+    (
+        {"actions": {"foo": "bar"}},
+        {"config": {"foo": "bar"}},
+    ),
+)
+def test_partial_meta_raises(tmp_path, legacy, params):
+    with create_tempcharm(
+        tmp_path, legacy=legacy, meta={"type": "charm", "name": "sergio"}
+    ) as charm:
+        # metadata not found:
```
tonyandrewmeyer, line 141: I found this initially confusing because there's metadata two lines above, maybe this would be clearer?

```suggestion
        # metadata not found in YAML:
```

tonyandrewmeyer (commented):
What we lose here is being able to do simple contexts like:

```python
ctx = testing.Context(MyCharm, actions={'foo': {'description': '...'}})
```

Although this isn't that much longer:

```python
ctx = testing.Context(MyCharm, meta={'name': '...'}, actions={'foo': {'description': '...'}})
```

I think I'm ok with the backwards compatibility breakage - I think generally people are autoloading so this is unlikely to break for anyone. It doesn't seem like there's a backwards compatible option outside of docs or warnings.

An alternative to this would be that if meta/config/actions was `None` then we could attempt autoloading that specific piece. That does get a bit weird if they're all in charmcraft.yaml, though. Basically, meta/config/actions would be an override over the autoload, rather than an alternative. I think this is what the OP was expecting?

I wonder if we're going to have some backwards compatibility breakage if it would be worth doing more and trying to solve the single-YAML issue too (#1424).

benhoyt (approved):
I think this is reasonable. Re Tony's comment:

> Basically, meta/config/actions would be an override over the autoload, rather than an alternative. I think this is what the OP was expecting?

That sounds too magical to me. I think it should be one or the other, not an override of (some) parts -- I think that would just be confusing.

PR #1499, opened 2024-12-13
refactor: cache signature structure in ops.testing state classes

tonyandrewmeyer:
The Scenario code that limits the number of positional arguments (needed in Python 3.8 since the dataclasses there do not have `KW_ONLY`) is quite expensive. We have a method of providing this functionality in 3.8, but it is unfortunately expensive, and therefore a regression from older versions of Scenario.

I considered 4 options here:
* Manually write out the dozen or so `__init__` signatures (as we had at one point in the [original PR](https://github.com/canonical/ops-scenario/pull/137)). I'd rather not for the same reasons we made the decision at that time.
* Figure out a way to have a syntax that allows using regular `dataclasses.KW_ONLY` with modern Python and only use our custom code on older Python. I spent some time looking into this, and couldn't figure out a way to do it where the code still seemed readable and maintainable enough.
* Cache the most expensive work (what's in this PR).
* Don't bother doing anything now, and eagerly wait for the ability to drop 3.8 support. A roughly 5% performance regression felt worth fixing as long as the change is fairly straightforward (although this change only gets about half of that back).

The most expensive parts of the `__new__` method are the ones that work with `inspect`: `inspect.signature` in particular, but also getting the default value of the parameters. If we assume that no-one is run-time altering the signature of the class (I believe no-one should be) then the values of these never actually change, but we are currently calculating them every time an instance of the class is created. This PR changes that to cache those three values the first time they are needed.

There's one additional performance tweak in the branch that doesn't make a significant difference, but is trivial to do: when checking if the YAML files exist, skip the filesystem `exists()` call if we just created an empty temporary directory a few lines above, since we know that it will never exist.

A drive-by: I happened to notice while working on this branch Harness referring to `options.yaml`, which does not exist (any more?) as far as I know, so a docs tweak to address that.

Timing (best of 3):

|Suite|main|branch|
|-|-|-|
|operator unit tests (no concurrency)|165.25s|161.89s|
|traefik scenario tests|45.49|44.30s|
|kafka-k8s-operator scenario tests|4.48s|4.38s|

Refs #1434

### DIFF
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -224,7 +224,9 @@ class Runtime:
         actions_yaml = virtual_charm_root / "actions.yaml"
 
         metadata_files_present: Dict[Path, Optional[str]] = {
-            file: file.read_text() if file.exists() else None
+            file: file.read_text()
+            if charm_virtual_root_is_custom and file.exists()
+            else None
             for file in (metadata_yaml, config_yaml, actions_yaml)
         }
 
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -122,26 +122,40 @@ def _max_posargs(n: int):
 
         _max_positional_args = n
 
-        def __new__(cls, *args: Any, **kwargs: Any):
+        @classmethod
+        def _annotate_class(cls):
+            """Record information about which parameters are positional vs. keyword-only."""
+            if hasattr(cls, "_init_parameters"):
+                # We don't support dynamically changing the signature of a
+                # class, so we assume here it's the same as previously.
+                return
             # inspect.signature guarantees the order of parameters is as
             # declared, which aligns with dataclasses. Simpler ways of
             # getting the arguments (like __annotations__) do not have that
             # guarantee, although in practice it is the case.
-            parameters = inspect.signature(cls.__init__).parameters
-            required_args = [
+            cls._init_parameters = parameters = inspect.signature(
+                cls.__init__
+            ).parameters
+            cls._init_kw_only = {
+                name
+                for name in tuple(parameters)[cls._max_positional_args :]
+                if not name.startswith("_")
+            }
+            cls._init_required_args = [
                 name
                 for name in tuple(parameters)
-                if parameters[name].default is inspect.Parameter.empty
-                and name not in kwargs
-                and name != "self"
+                if name != "self"
+                and parameters[name].default is inspect.Parameter.empty
+            ]
+
+        def __new__(cls, *args: Any, **kwargs: Any):
+            cls._annotate_class()
+            required_args = [
+                name for name in cls._init_required_args if name not in kwargs
             ]
             n_posargs = len(args)
             max_n_posargs = cls._max_positional_args
-            kw_only = {
-                name
-                for name in tuple(parameters)[max_n_posargs:]
-                if not name.startswith("_")
-            }
+            kw_only = cls._init_kw_only
             if n_posargs > max_n_posargs:
                 raise TypeError(
                     f"{cls.__name__} takes {max_n_posargs} positional "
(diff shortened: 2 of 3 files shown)


james-garner-canonical (approved):
Looks good, clever and simple solution, and a nice performance win.

I was briefly confused when tracing the new vs old logic, because `required_args` in `__new__` stores required (as args -- not kwargs), while `cls._init_required_args` stores required (as args or kwargs). Idk if there's a clearer short name for `_init_required_args` though, and it seems easy enough to follow when not trying to go side by side (so fine for future readers).

testing/src/scenario/state.py
```diff
@@ -122,26 +122,40 @@ class _MaxPositionalArgs:
 
         _max_positional_args = n
 
-        def __new__(cls, *args: Any, **kwargs: Any):
+        @classmethod
+        def _annotate_class(cls):
+            """Record information about which parameters are positional vs. keyword-only."""
+            if hasattr(cls, "_init_parameters"):
+                # We don't support dynamically changing the signature of a
+                # class, so we assume here it's the same as previously.
```
dimaqq, line 130: This function will only ever be used on our own classes, will it not?

Maybe it's better to state that instead.

testing/src/scenario/state.py
```diff
@@ -122,26 +122,40 @@ class _MaxPositionalArgs:
...
             # getting the arguments (like __annotations__) do not have that
             # guarantee, although in practice it is the case.
-            parameters = inspect.signature(cls.__init__).parameters
-            required_args = [
+            cls._init_parameters = parameters = inspect.signature(
+                cls.__init__
+            ).parameters
+            cls._init_kw_only = {
+                name
+                for name in tuple(parameters)[cls._max_positional_args :]
```
dimaqq, line 141: I'm assuming this was auto-formatted by ruff, the dangling `:]` is kinda funny :]

dimaqq (approved):
High level: the time saved is marginal, but then code is rather straightforward and I'm not sure what would the alternative be, maybe custom types where the inspected bits are inlined or hardcoded?

Anyway +1 on this.

testing/src/scenario/_runtime.py
```diff
@@ -224,7 +224,9 @@ def _virtual_charm_root(self):
         actions_yaml = virtual_charm_root / "actions.yaml"
 
         metadata_files_present: Dict[Path, Optional[str]] = {
-            file: file.read_text() if file.exists() else None
+            file: file.read_text()
+            if charm_virtual_root_is_custom and file.exists()
+            else None
```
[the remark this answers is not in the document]
tonyandrewmeyer: Saving the calls, it's this bit in the PR description:

> There's one additional performance tweak in the branch that doesn't make a significant difference, but is trivial to do: when checking if the YAML files exist, skip the filesystem exists() call if we just created an empty temporary directory a few lines above, since we know that it will never exist.

It's not going to make any real performance difference, but I feel it's generally cleaner and better to be checking a bool before you do external things (and even really before calling a function), if you're joining with `and`.

testing/src/scenario/state.py
tonyandrewmeyer: It was, yeah - I lazily just wrote it out on one line and let `tox -e fmt` clean up.

Does the trailing happy face bother you? I could do something like move `tuple(parameters)` out to a separate line to get the dict comprehension to be on a single line, if that would be better.

testing/src/scenario/state.py
tonyandrewmeyer: It's tricky - `_max_posargs()` is clearly marked as being only for our use, but then most of the state components are public and you could, theoretically, mess about with them - I doubt a test would ever have reason to do this, but maybe some library built on top of Scenario might? I'll adjust the comment so that it covers both points.

PR #1504, opened 2024-12-17
test: add a small set of ops.testing benchmark tests

tonyandrewmeyer:
The PR:
 * Adds a new `tox -e benchmark` command.
 * Adds a new `test/test_benchmark.py` model.
 * Excludes the benchmark tests from `tox -e unit` and `tox -e coverage`

I wrote a small set of benchmark tests while working on #1434. It seems like it'd be worth keeping these - in the future, they could be expanded to include a small set of benchmark tests that target ops more specifically, and ideally we could have a CI workflow that failed if there were regressions (currently, running the tests just outputs the timing, and you can use the pytest-benchmark tools to do comparisons).


test/test_benchmark.py
```diff
@@ -0,0 +1,141 @@
...
+
+import ops
+from ops import testing
+
+sys.path.append(str(pathlib.Path(__file__).parent / 'charms' / 'test_benchmark' / 'src'))
+
+from bcharm import BenchmarkCharm
+
+
+def test_context_explicit_meta(benchmark):
```
benhoyt, line 38: Magical pytest import strikes again! I couldn't for the life of me see where this came from at first. All I want for Christmas is for pytest to start being explicit and acting like regular Python! :-)

tox.ini
```diff
@@ -104,7 +104,8 @@ deps =
     -e .
     -e testing
 commands =
-    pytest -n auto --ignore={[vars]tst_path}smoke -v --tb native \
+    pytest -n auto --ignore={[vars]tst_path}smoke --ignore={[vars]tst_path}test_benchmark.py \
+        -v --tb native \
```
benhoyt, line 108: Does pytest run tests in directories recursively by default? If not, would it be better to put these in a `test/benchmark/` directory to avoid having to ignore them in various places and keep them a bit more separate? It may also avoid the pages of "Skipping non-benchmark" output.

tox.ini
```diff
@@ -125,10 +126,30 @@ deps =
...
 
+[testenv:benchmark]
+description = Run benchmark tests
+passenv =
+    RUN_REAL_PEBBLE_TESTS
+    PEBBLE
+deps =
+    PyYAML==6.*
+    websocket-client==1.*
+    coverage[toml]~=7.0
```
benhoyt, line 143: Can we drop coverage for running benchmarks? In fact, could this slow things down / mess with the results?

tox.ini
```diff
@@ -125,10 +126,30 @@ deps =
...
+    pytest~=7.2
+    pytest-benchmark
+    typing_extensions~=4.2
+    jsonpatch~=1.33
```
benhoyt, line 147: Do we need jsonpatch here?

benhoyt (commented):
Good idea, thanks for committing these! Couple of minor comments/questions.

test/test_benchmark.py
tonyandrewmeyer: > Magical pytest import strikes again! I couldn't for the life of me see where this came from at first. All I want for Christmas is for pytest to start being explicit and acting like regular Python! :-)

:) I'm not a fan of this either. I could add a comment explaining that the fixture is automatically available via pytest-benchmark above the first test, or in the module docstring?

tox.ini
tonyandrewmeyer: Ah, good catch, thanks. I had lazily just copied the dependencies from the coverage environment. It shouldn't change the results because I'm not actually running `coverage` but it's indeed unnecessary.

tox.ini
tonyandrewmeyer: Right now, yes, because pytest goes through collecting all the tests and then pytest-benchmark marks lots of them as ones to skip - but we've still imported all the other test files where there are `import jsonpatch` lines.

If the test is in a separate folder instead, then I think that can be avoided.

tox.ini
tonyandrewmeyer: > Does pytest run tests in directories recursively by default? If not, would it be better to put these in a `test/benchmark/` directory to avoid having to ignore them in various places and keep them a bit more separate? 

Yeah, I wasn't sure where the best place for this was. I'm happy to move it to a separate directory - the tox command can point there specifically so any recursing wouldn't be an issue.

> It may also avoid the pages of "Skipping non-benchmark" output.

Yeah, definitely not a fan of those.

test/test_benchmark.py
benhoyt: That would be helpful, thanks.

benhoyt: I think above the first test would be more visible.

benhoyt (approved):

test/charms/test_benchmark/src/bcharm.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
```
dimaqq, line 1: How about something easier to understand in the directory listing than `bcharm` "be charmed?"

Given that it's the only file under `src`, why not `charm.py`?

If it must to be different, maybe `benchmark_charm.py`?

test/charms/test_benchmark/src/bcharm.py
```diff
@@ -0,0 +1,39 @@
...
+        framework.observe(self.on.stop, self._on_stop)
+        framework.observe(self.on.config_changed, self._on_config_changed)
+
+    def _on_update_status(self, _: ops.UpdateStatusEvent):
+        # Say a bunch of things.
+        for level in ('debug', 'info', 'warning', 'error'):
+            for i in range(50):
+                getattr(logger, level)('This is message %s', i)
+
+    def _on_stop(self, _: ops.StopEvent):
```
dimaqq, line 31: if that's only for update status to kick in, maybe replace `pass` with a docstring to that effect?

testing/tests/benchmark/test_testing.py
```diff
@@ -0,0 +1,142 @@
...
+    benchmark(mock_pytest)
+
+
+def test_lots_of_logs(benchmark):
+    ctx = testing.Context(BenchmarkCharm)
+    benchmark(ctx.run, ctx.on.update_status(), testing.State())
+    assert len(ctx.juju_log) > 200
+
+
+def ditest_full_state(benchmark):
```
dimaqq, line 103: I imagine this means "disabled test".

Should it be removed before merge?

tox.ini
```diff
@@ -104,7 +106,9 @@ deps =
     -e .
     -e testing
 commands =
-    pytest -n auto --ignore={[vars]tst_path}smoke -v --tb native \
+    pytest -n auto --ignore={[vars]tst_path}smoke \
+        --ignore={[vars]tst_path}benchmark --ignore={[vars]testing_tst_path}benchmark \
```
dimaqq, line 110: Can we use markers instead?

The magic paths are not very ergonomic, e.g. not easy to copy-paste into the terminal.

test/charms/test_benchmark/src/bcharm.py
tonyandrewmeyer: > Given that it's the only file under `src`, why not `charm.py`?

We annoyingly have ops/charm.py and also default to charms being src/charm.py. Most of the time this is fine, but in the tests ops is in the `PYTHONPATH` so ops/charm.py gets found with a plain `import charm`. I originally had a bunch of path mangling to get `src` in the right place so that it'd pick up src/charm.py instead but it ended up pretty messy (I don't recall the specifics now).

> If it must to be different, maybe `benchmark_charm.py`?

Sure.

testing/tests/benchmark/test_testing.py
tonyandrewmeyer: Ah, definitely not meant to be left like this. Thanks - will fix.

tox.ini
tonyandrewmeyer: That's true, but we are using them extensively throughout tox.ini already. I feel like it's better to remain consistent.

I wouldn't have any objection to a separate PR that removed them entirely, though. I don't think they add much value, and using explicit paths (for e.g. PYTHONPATH) and markers (for e.g. skipping types of test) could be a nice improvement.

testing/tests/benchmark/test_testing.py
```diff
@@ -0,0 +1,154 @@
...
+
+sys.path.append(
+    str(
+        pathlib.Path(__file__).parent.parent.parent.parent
+        / "test"
+        / "charms"
+        / "test_benchmark"
+        / "src"
+    )
+)
```
dimaqq, line 32: WDYT about one of these?

```suggestion
sys.path.append(str(pathlib.Path(__file__) / "../../../../test/charms/test_benchmark/src"))
```

or 

```suggestion
sys.path.append(str(pathlib.Path(__file__).parent / "../../../test/charms/test_benchmark/src"))
```

if a construct like `sub/dir/file.py/../../../another/dir` is too jarring...

I'm a little wary of changing import path at import time, as this could hypothetically affect other tests.

Alternatives would include:
- path tweak and import in a fixture, or
- explicit path setup in pytest config

testing/tests/benchmark/test_testing.py
```diff
@@ -0,0 +1,154 @@
...
+def test_run_no_observer(benchmark):
+    ctx = testing.Context(BenchmarkCharm)
+    benchmark(ctx.run, ctx.on.start(), testing.State())
+    assert len({e.handle.kind for e in ctx.emitted_events}) == 1
+
+
+def test_run_observed(benchmark):
+    ctx = testing.Context(BenchmarkCharm)
+    benchmark(ctx.run, ctx.on.stop(), testing.State())
+    assert len({e.handle.kind for e in ctx.emitted_events}) == 1
```
dimaqq, line 53: As a random developer who's reading these tests, I'm confused why that's so, or alternatively, what the intention of the test is.

The charm listen to `on_stop` but not to `on_start`.
Why is the handler count the same?

Would it be clearer to assert that `handlers == [this.specific.Handler]` instead?

Or is this about internal handling machinery and unrelated to what the charm listens on?

tox.ini
```diff
@@ -124,11 +128,30 @@ deps =
...
+[testenv:benchmark]
+description = Run benchmark tests
+passenv =
+    RUN_REAL_PEBBLE_TESTS
+    PEBBLE
+deps =
+    PyYAML==6.*
+    websocket-client==1.*
+    pytest~=7.2
+    pytest-benchmark
```
dimaqq, line 148: Maybe request a compatible version rather than any, wdyt?

```suggestion
    pytest-benchmark~=5.0
```

testing/tests/benchmark/test_testing.py
```diff
@@ -0,0 +1,154 @@
...
+def test_context_autoload_meta(benchmark):
+    ctx = benchmark(testing.Context, BenchmarkCharm)
+    assert isinstance(ctx, testing.Context)
+
+
+def test_many_tests_explicit_meta(benchmark):
+    def mock_pytest():
+        """Simulate running multiple tests against the same charm."""
+        for event in ("install", "start", "stop", "remove"):
+            for _ in range(5):
```
dimaqq, line 77: Given that benchmark, presumably, handles the repetition, what is the significance of going through exactly `5` inner loop iterations?

dimaqq (approved):

testing/tests/benchmark/test_testing.py
tonyandrewmeyer: The purpose of the tests is to check how long it takes Scenario to simulate running the event - when there's no observer in the first case, and when there is an observer in the second case. The assert is only there to make sure that the event did actually run, rather than it being a no-op (accidentally no-op'ing code makes benchmarks pleasingly fast). What is being asserted in these tests isn't the point of the test - the point is how long the test takes to run.

The count is the same because it's a *set*, so it's the count of *unique* events. The actual list of emitted events is as long as the benchmark repetition setting, so not really predictable in the test code. It's 1 and not 0 because it's the number of events, not the number of events handled. The `handle` attribute is a framework "handle" not meaning "observer function" (it's unfortunate that we use the same word for both things).

I'll add a comment to clarify.

testing/tests/benchmark/test_testing.py
tonyandrewmeyer: This test was badly constructed. I've adjusted it and also added a comment that outlines what it's doing.

testing/tests/benchmark/test_testing.py
tonyandrewmeyer: Agreed, it would be better to minimise the impact. I've moved it to a fixture, thanks.

testing/tests/benchmark/charms/benchmark_charm/requirements.txt
```diff
@@ -0,0 +1 @@
+ops ~= 2.17
```
dimaqq, line 1: This seems to specify `ops` from pypi and not the local checkout. Is that intended?

dimaqq (approved):

testing/tests/benchmark/charms/benchmark_charm/requirements.txt
tonyandrewmeyer: It doesn't get used, so it doesn't really matter. Scenario behaves differently if there's an actual charm folder to load the metadata from vs being passed the metadata, and the timing there was one of the things I was working on, which is why the test exists. The test could create a temporary directory and write the metadata there, but it seemed simpler to just have an actual charm folder with a real charm.

It's also patterned on the charms in the ops tests - for test_smoke it does actually matter what version of ops is there, because we pack and deploy the charm (tox builds ops and puts it there); for test_main I think it doesn't matter because we never pack.

I could just remove the requirements file, but it seems nice to be able to pack it and run it manually if you want to investigate things (but not so much to have a tox command for it).

PR #1505, opened 2024-12-18
docs: remove the publish badge from the README

tonyandrewmeyer:
This is currently broken (because the workflow was renamed), but rather than fix it, we've decided that it doesn't really add any value to have this badge, so we'll remove it instead.


james-garner-canonical (approved):

dimaqq (approved):

dimaqq:
Let's merge this, the end of the month is upon us!

PR #1509, opened 2024-12-18
chore: remove file that seems to be accidentally committed

tonyandrewmeyer:
The removed file has the same name as another one, but starts and ends with `#`, and has slightly different (I'm assuming outdated) content. I think this is the result of having the file open in some editor when something like `git add .` was done.


dwilding (approved):
I just had a look and came to the same conclusion :slightly_smiling_face:

IronCore864 (approved):

PR #1511, opened 2024-12-18
docs: remove tutorial chapters that are covered by the how-to guide

tonyandrewmeyer:
The (k8s) tutorial is currently quite long, and we'd like it to be shorter so that it's easier for charmers to get going.

The plan is that the [features](https://juju.is/docs/sdk/study-your-application#heading--features) described in the tutorial should be the chapters. Namely:

 * A simple web server with endpoints that can be used, including metrics
 * Integration with postgresql
 * An action to be able to manually interact with postgresql
 * COS integration, including logs and metrics

This PR removes the chapters:
 * Exposing the workload version (the existing how-to guide already covers this, and it's not essential for a charm)
 * Storing state (we would generally prefer that charms don't do this, so shouldn't encourage it, and it'll be covered by a how-to guide before this PR is ready for review)
 * Exposing a port via K8s (too small for a tutorial chapter, and the existing how-to guide already covers this - and was based mostly on this chapter, in the knowledge that it'd be replacing it)
 * Write Harness tests - the new unit testing (Scenario) chapter is merged into this. The testing chapters need to be migrated into sections at the end of each chapter, but that's work for another day and another person. This just stops teaching new charmers to write Harness tests
 * Publishing the charm to Charmhub. Now that this is specifically an ops tutorial, rather than a broader Juju SDK one, this isn't really part of our story, and is better placed in the Charmcraft documentation.

There are a few other small tidy-ups, particularly around whitespace and removing old commented text.

Fixes #1587

PR #1513, opened 2024-12-18
feat: add a docs link to the Harness deprecation warning

### LINKED ISSUE #1503: Improve Harness deprecation warning to help charmers know what to do next
The deprecation warning for Harness says:

> Harness is deprecated; we recommend using state transition testing (previously known as 'Scenario') instead

We should do a better job helping the charmer move forward from this. Just pointing to the docs is potentially enough (I don't think they were live when we added the warning). Maybe we also want to say to install `ops[testing]`, that the classes are still in `ops.testing`, and things like that as well.

Pointed out by @jameinel in Matrix.

tonyandrewmeyer:
Links to the documentation in the Harness pending deprecation warning, so that charmers know how to proceed, if they have no idea what "state transition testing" is (and don't go searching for "Scenario").

Fixes #1503

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -316,7 +316,7 @@ class Harness(Generic[CharmType]):
         )
 
         warnings.warn(
-            'Harness is deprecated. See more: "
+            'Harness is deprecated. See more: '
             'https://ops.readthedocs.io/en/latest/howto/write-unit-tests-for-a-charm.html',
             PendingDeprecationWarning,
             stacklevel=2,
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -316,7 +316,7 @@ class Harness(Generic[CharmType]):
         )
 
         warnings.warn(
-            'Harness is deprecated. See more: '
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
In another PR, I rename destination link, so leaving as draft until that is merged.

tonyandrewmeyer:
> In another PR (#1511), I rename the destination link, so leaving as draft until that is merged.

@IronCore864 did this for me in another PR (thanks!) so this should be good to go now.

ops/_private/harness.py
```diff
@@ -316,8 +316,8 @@ def __init__(
         )
 
         warnings.warn(
-            'Harness is deprecated; we recommend using state transition testing '
-            "(previously known as 'Scenario') instead",
+            'Harness is deprecated. See more: '
```
dwilding, line 319: I'm being picky, but the how-to page won't actually give any more information about Harness being deprecated :wink: 

How about:

```suggestion
            'Harness is deprecated. For the recommended approach, see: '
```

dwilding (approved):

IronCore864 (approved):

PR #1514, opened 2024-12-18
ci: correctly point PyPI publishing to the ops-scenario packages

tonyandrewmeyer:
`working-directory` is specifically for `run` steps. For the `gh-action-pypi-publish` action, `with: packages-dir` should be used.

### DIFF
--- a/.github/workflows/publish-ops-scenario.yaml
+++ b/.github/workflows/publish-ops-scenario.yaml
@@ -29,10 +29,10 @@ jobs:
         run: python -m build
         working-directory: ./testing
       - name: Attest build provenance
-        working-directory: ./testing
         uses: actions/attest-build-provenance@v1.4.3
         with:
           subject-path: 'testing/dist/*'
       - name: Publish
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
I'm not sure how to test this pre-merge (to avoid it being wrong again, although I've gone through the documentation more carefully this time). I can't run the action against this branch, and I can't run the action in my fork (because the trusted publishing system requires it to be from the real repo). Suggestions welcome!

james-garner-canonical (approved):
I guess the best way to test will be to run the test pypi job manually after merging this

dimaqq (approved):
re testing: push to test pypi from a branch with made up package version.

Edit:  I guess that requires setting secrets for the branch and possibly tweaking the test.pypi trusted publishing temprarily?

tonyandrewmeyer:
Tested and [works fine] :tada: - I'm reasonably confident that this means that publishing to actual PyPI will also be fine.

PR #1523, opened 2025-01-08
docs: use explicit framework param instead of *args

benhoyt:
A while back we decided to recommend using an explicit `framework` parameter instead `*args` to a charm's `__init__`. We've updated `charmcraft init` templates, but we forgot a few places in our docs. Update them to match here.

[Preview](https://ops--1523.org.readthedocs.build/en/1523/reference/ops.html#ops.CharmBase)

### DIFF
--- a/ops/__init__.py
+++ b/ops/__init__.py
@@ -20,8 +20,8 @@ This API provides core features to your charm, including:
   the base class for charm libraries.
 - :class:`~ops.framework.EventBase` class and individual event types, like
   the :class:`~ops.ActionEvent` class.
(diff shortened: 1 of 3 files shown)


benhoyt:
`test-real-pebble` failed once [here](https://github.com/canonical/operator/actions/runs/12679592014/job/35339603857?pr=1523). It looks like `test_checks_and_health` is timing-sensitive and should be improved. Re-running for now, and if it happens again we should open an issue and fix.

tonyandrewmeyer (approved):
Thanks!

dwilding (approved):
Nice one, thanks!

PR #1526, opened 2025-01-13
chore: update documentation dependencies

tonyandrewmeyer:
There's a security release for Jinja (not one that impacts our use from what I can tell, but nice to just move to the updated version anyway). While bumping that, bump all the other dependencies as well.

This is just the result of removing the requirements file and regenerating it with `tox -e docs-deps`.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,7 +6,7 @@
 #
 alabaster==1.0.0
     # via sphinx
-anyio==4.4.0
+anyio==4.8.0
     # via
     #   starlette
     #   watchfiles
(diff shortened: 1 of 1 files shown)


docs/requirements.txt
```diff
@@ -17,15 +17,15 @@ beautifulsoup4==4.12.3
     #   canonical-sphinx-extensions
     #   furo
     #   pyspelling
-bracex==2.5
+bracex==2.5.post1
```
james-garner-canonical, line 20: This jumped out at me. Looks like the `2.5.post1` release is just fixing the PyPI landing page. If this was manual, I'd suggest we should pin on 2.5 instead, but in this context I guess it just makes sense to use what pip outputs?

james-garner-canonical (approved):

docs/requirements.txt
tonyandrewmeyer: I don't think we should do anything differently just because the library uses an odd versioning scheme (it's not indicating a pre-release, which pip would have skipped). If they'd gone with the usual `2.5.1` I don't think we would have any hesitation using it, so it's really just naming.

james-garner-canonical: Makes sense, thanks

dimaqq: LOL at `Pump version to post so we can fix PyPI landing page` 🙃

tonyandrewmeyer:
@dimaqq you commented here but didn't approve/request changes. Are you ok with the PR?

dimaqq (approved):

PR #1535, opened 2025-01-21
chore: bump CI dependencies

tonyandrewmeyer:
* Bump grafana-k8s charm revision to the current latest, which should get the tests passing again and therefore unblock merging (a nicer pinning that includes the comment will automatically happen later in the month)
* Bump the database charm CI to use a newer poetry, which seems to be a requirement for newer versions of the DB charms

### DIFF
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -35,7 +35,7 @@ jobs:
           path: myops
 
       - name: Install patch dependencies
-        run: pip install poetry~=1.6
+        run: pip install poetry~=2.0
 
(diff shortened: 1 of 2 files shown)


james-garner-canonical (dismissed):

james-garner-canonical (commented):
I initially hit approve but it looks like this leads to more test failures?

tonyandrewmeyer:
> I initially hit approve but it looks like this leads to more test failures?

Yeah, seems like our command doesn't exist in the newer poetry. I'll remove that part of the PR and whoever solves things in the next automated one can handle that part.

dimaqq (approved):

james-garner-canonical (approved):

PR #1538, opened 2025-01-24
chore: make event_name mandatory in Framework

dimaqq:
I'm sure that the event name is always set (except some units tests)
Meanwhile we've got this signature: `Framework.__init__(..., event_name: str|None, ...)`
And therefore `self._event_name` is optional too.

This PR makes this mandatory.

Open question remaining: whether ops-scenario / ops version needs to be bumped to keep the two in sync.
That would be necessary if ops-scenario ships with harness code and ops is not.
(I don't recall if ops / ops['testing'] has reached that stage)

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -312,6 +312,7 @@ class Harness(Generic[CharmType]):
             self._charm_dir,
             self._meta,
             self._model,
+            '',
             juju_debug_at=self._juju_context.debug_at,
             # Harness tests will often have defer() usage without 'purging' the
             # deferred handler with reemit(), but still expect the next emit()
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -606,7 +606,7 @@ class Framework(Object):
         charm_dir: Union[str, pathlib.Path],
         meta: 'charm.CharmMeta',
         model: 'Model',
-        event_name: Optional[str] = None,
+        event_name: str,
         juju_debug_at: Optional[Set[str]] = None,
         skip_duplicate_events: bool = True,
     ):
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -73,6 +73,7 @@ def create_framework(
         charm_dir,
         meta=model._cache._meta if model else ops.CharmMeta(),
         model=model,  # type: ignore
+        event_name='',
         juju_debug_at=_JujuContext.from_dict(os.environ).debug_at,
     )
     request.addfinalizer(framework.close)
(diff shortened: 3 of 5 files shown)


benhoyt:
What's the motivation for this change? Unless we have a very compelling reason, I don't think we should do it, as it's a breaking change and will actually break some charm tests (GitHub Actions exporter, Prometheus K8s, Sunbeam charms, and Traefik Route):

```
~/w/charm-analysis/.cache$ rg 'Framework\(' --sort=path -A5
github-actions-exporter-operator/tests/unit/test_charm.py
40:        self.harness._framework = ops.framework.Framework(
41-            self.harness._storage, self.harness._charm_dir, self.harness._meta, self.harness._model
42-        )
43-        self.harness._charm = None
44-        self.harness.update_config({"github_webhook_token": "foo"})
45-        self.harness.enable_hooks()
--
68:        self.harness._framework = ops.framework.Framework(
69-            self.harness._storage, self.harness._charm_dir, self.harness._meta, self.harness._model
70-        )
71-        self.harness._charm = None
72-        self.harness.update_config({"github_webhook_token": ""})
73-        self.harness.enable_hooks()
--
118:        self.harness._framework = ops.framework.Framework(
119-            self.harness._storage, self.harness._charm_dir, self.harness._meta, self.harness._model
120-        )
121-        self.harness._charm = None
122-        new_webhook_token = token_hex(16)
123-        self.harness.update_config({"github_webhook_token": new_webhook_token})

prometheus-k8s-operator/tests/unit/test_remote_write.py
292:            self.harness._framework = framework.Framework(
293-                self.harness._storage,
294-                self.harness._charm_dir,
295-                self.harness._meta,
296-                self.harness._model,
297-            )

sunbeam-charms/ops-sunbeam/ops_sunbeam/test_utils.py
822:    harness._framework = framework.Framework(
823-        ops.storage.SQLiteStorage(":memory:"),
824-        harness._charm_dir,
825-        harness._meta,
826-        harness._model,
827-    )

traefik-route-k8s-operator/tests/unit/conftest.py
79:    harness._framework = framework.Framework(
80-        harness._storage, harness._charm_dir, harness._meta, harness._model
81-    )
82-
83-    harness._charm = None
84-    harness.begin()
```

dimaqq:
Oh, I didn't realise that charm tests instantiated the `ops.Framework` directly, I thought that surely they used the fake framework from Harness or something like that.

Given this, I guess the proposed change is a no-go.

Maybe let's discuss if we can get rid of this nit somehow... (docs? long deprecation period? a default like `event_name="he-be-dragons"`, smth else?).

I consider it a nit because a usable Framework, that is a framework that can call the user handlers for events, must have an event associated with it, or an event set to call handlers.

dimaqq:
Stand-up decision: let's not do this.
We'll need to clean up or rearchitect `._event_name` and `backend._hook_is_running` fields, but it doesn't seem worth it while Harness is still around.

PR #1541, opened 2025-01-27
ci: use Concierge to set up the smoke test environments

tonyandrewmeyer:
The smoke tests are currently failing because of `sg` behaviour changes in the `ubuntu:latest` runners. Rather than figure out what the solution for that is, this PR simplifies the entire environment setup to just use Concierge instead.

This also switches from testing against Juju 3.5 to 3/latest (which is 3.6), since 3.6 is available now and the intention was always to test against the latest 3.

A [run can be seen in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/13003299600).

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -13,47 +13,20 @@ jobs:
       matrix:
         # pylibjuju does not currently support Juju 4.x
         # The smoke tests do not yet work on Juju 2.9.
-        juju-version: ['3.6']
-        charmcraft-version: ['2.x', '3.x']
-        cloud: ['lxd', 'microk8s']
+        juju-channel: ['3/stable']
+        charmcraft-channel: ['2.x/stable', '3.x/stable']
+        preset: ['machine', 'microk8s']
 
     env:
-      JUJU_VERSION: "${{ matrix.juju-version }}"
+      CONCIERGE_JUJU_CHANNEL: "${{ matrix.juju-channel }}"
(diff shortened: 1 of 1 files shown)


james-garner-canonical (approved):
Wow this is a very nice simplification of the setup, looks like `concierge` does what it promises

.github/workflows/smoke.yaml
```diff
@@ -13,47 +13,20 @@ jobs:
...
-        charmcraft-version: ['2.x', '3.x']
-        cloud: ['lxd', 'microk8s']
+        juju-channel: ['3/stable']
+        charmcraft-channel: ['2.x/stable', '3.x/stable']
+        preset: ['machine', 'microk8s']
 
     env:
-      JUJU_VERSION: "${{ matrix.juju-version }}"
+      CONCIERGE_JUJU_CHANNEL: "${{ matrix.juju-channel }}"
+      CONCIERGE_CHARMCRAFT_CHANNEL: "${{ matrix.charmcraft-channel }}"
```
dimaqq, line 22: I figure that converting these to explicit command line arguments to concierge would be cleaner, more targeted and it makes the whole env block redundant.

tonyandrewmeyer: The env block felt clean to me, but no objection to switching it over.

benhoyt (approved):
Great, thanks for this!

@jnsgruk Speaking of Concierge... :-)

PR #1542, opened 2025-01-28
ci: bump poetry to 2.0 to match downstream

tonyandrewmeyer:
The charms have moved to Poetry 2.0, so we need to as well to remain compatible with the new options (and `--no-update` is no longer a flag, as it's the default behaviour).


james-garner-canonical (approved):
Thanks for the explanation, changes make sense, and the tests have all passed too

dimaqq (approved):

PR #1543, opened 2025-01-28
chore: fix compatibility with ops-scenario 7.0.5

tonyandrewmeyer:
The next version of ops-scenario renames the `runtime` module to `_runtime`. The `main` branch of `ops` currently expects that rename to have happened, but doesn't require it in the `[testing]` dependency. This means that `main@HEAD` `ops` doesn't work unless you also use `main@HEAD` of `ops-scenario`, which isn't the case in our compatibility checks - or in any charm that ends up with that combination.

We could fix this by adjusting the `ops` dependencies to require the newer `ops-scenario`, but that would really require releasing a new version of `ops-scenario`, and we're unlikely to be doing that before we also release `ops`.

The relevant code is only required until we release the next version of `ops-scenario` so for now just handle 7.0.5 as well as the newer code.

This will fix several of the "broad compatibility" charm CI tests.

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -177,14 +177,18 @@ else:
         # monkeypatch it in, so that the ops.testing.ActionFailed object is the
         # one that we expect, even if people are mixing Harness and Scenario.
         # https://github.com/canonical/ops-scenario/issues/201
+        # This will be the case in the next version of ops-scenario, so this
+        # code can be removed as of the release of ops-scenario after 7.0.5.
+        # Remember to bump the required version of `ops-scenario` in pyproject.toml
+        # at that time as well.
         try:
             import scenario._runtime as _runtime
         except ImportError:
-            import scenario.runtime as _runtime
+            import scenario.runtime as _runtime  # type: ignore
         import scenario.context as _context
 
         _context.ActionFailed = ActionFailed  # type: ignore[reportPrivateImportUsage]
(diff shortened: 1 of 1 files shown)


benhoyt (approved):

james-garner-canonical (approved):
Seems like a good solution in this case.

It kind of seems like we *should* be using `main@HEAD` of `ops-scenario` when running unreleased `main@HEAD` of `ops` though, shouldn't we? That way we can make changes to both (potentially in the same PR), test them, and release them simultaneously, bumping the requirements at that time.

I'm not sure what exactly we need to do to make that possible -- you've probably already thought about it, but I just thought I'd mention it.

dimaqq (approved):

dimaqq:
P.S. thank you, I've rebased my branch on top of yours and I can now test my random charm against `ops[tracing]`

tonyandrewmeyer:
> It kind of seems like we _should_ be using `main@HEAD` of `ops-scenario` when running unreleased `main@HEAD` of `ops` though, shouldn't we? That way we can make changes to both (potentially in the same PR), test them, and release them simultaneously, bumping the requirements at that time.
> 
> I'm not sure what exactly we need to do to make that possible -- you've probably already thought about it, but I just thought I'd mention it.

Yeah, I think we probably ought to, particularly as the use of ops-scenario increases. I have a [Jira ticket](https://warthogs.atlassian.net/browse/CHARMTECH-425) for getting some matrix testing in place, and it could probably be done as part of that (I've added a comment there).

PR #1545, opened 2025-01-29
ci: include Juju 2.9 in the smoke tests

tonyandrewmeyer:
Now that we're using Concierge running against Juju 2.9 works trivially, so we might as well include it for now, to avoid any breakage.

This also fixes the python-libjuju installation that broke in the last change (but still worked because the default is 3.x, which is what Juju 3.x needs).

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         # pylibjuju does not currently support Juju 4.x
         # The smoke tests do not yet work on Juju 2.9.
-        juju-channel: ['3/stable']
+        juju-channel: ['2/stable', '3/stable']
         charmcraft-channel: ['2.x/stable', '3.x/stable']
         preset: ['machine', 'microk8s']
 
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -12,7 +12,6 @@ jobs:
     strategy:
       matrix:
         # pylibjuju does not currently support Juju 4.x
-        # The smoke tests do not yet work on Juju 2.9.
         juju-channel: ['2.9/stable', '3/stable']
         charmcraft-channel: ['2.x/stable', '3.x/stable']
         preset: ['machine', 'microk8s']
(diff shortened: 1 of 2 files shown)


dimaqq (approved):

tox.ini
```diff
@@ -186,8 +185,8 @@ commands =
     python -m build --sdist --outdir={toxinidir}/test/charms/test_smoke/
     # Inject the tarball into the smoke test charm's requirements.
     bash -c 'echo "./$(ls -1 ./test/charms/test_smoke/ | grep tar.gz)" > ./test/charms/test_smoke/requirements.txt'
-    # If a specific Juju version is set, then make sure we are using that version of pylibjuju.
-    bash -c 'if [ -n "$JUJU_VERSION" ]; then pip install "juju ~= $JUJU_VERSION"; fi'
+    # Make sure we are using the appropriate version of pylibjuju.
+    bash -c 'JUJU_VERSION=$(juju --version | awk -F. "{print \$1 \".\" \$2}"); if [ -n "$JUJU_VERSION" ]; then pip install "juju ~= $JUJU_VERSION"; fi'
```
james-garner-canonical, line 189: Out of curiosity, where was `JUJU_VERSION` being set before?

.github/workflows/smoke.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         # pylibjuju does not currently support Juju 4.x
         # The smoke tests do not yet work on Juju 2.9.
-        juju-channel: ['3/stable']
+        juju-channel: ['2.9/stable', '3/stable']
```
james-garner-canonical, line 16: We can remove the comment that the smoke tests don't work on 2.9, right?

james-garner-canonical (approved):

.github/workflows/smoke.yaml
tonyandrewmeyer: Ha, yes 😊

tox.ini
tonyandrewmeyer: In [CI](https://github.com/canonical/operator/blob/e573f8f39c6b11470dcae3ac94ad798e4655ee91/.github/workflows/smoke.yaml#L21)

Issue #1516, opened 2024-12-20
Update empty Juju and Charmcraft links


dwilding:
As part of #1481, we have several empty links to Juju and Charmcraft docs that are commented out, ready for when the appropriate target pages are ready

**Action:** Uncomment the links and point them to the right places

To find the links, search for `UPDATE LINKS` in the doc source. [Details](https://github.com/canonical/operator/pull/1481#issuecomment-2547837136)

dwilding:
@tmihoc I'd be happy to work on this when the time comes, but let me know if you'd prefer to handle it

tmihoc:
@dwilding I can handle it. The Juju PR has been merged but we're still having some issues with the RTD end, so I don't have the links yet. I'll update them as soon as that's solved.

dwilding:
Cool, I just assigned it to you. Thank you very much!

PR #1546, opened 2025-01-29
docs: set up intersphinx and add links

tmihoc:
This PR sets up intersphinx for Juju and Charmcraft links and adds intersphinx-style links to Juju and Charmcraft in all the places we had bookmarked through UPDATE LINKS comments before (see https://github.com/canonical/operator/issues/1516 ).

### DIFF
--- a/docs/howto/manage-actions.md
+++ b/docs/howto/manage-actions.md
@@ -1,9 +1,6 @@
 (manage-actions)=
 # How to manage actions
-
-<!-- UPDATE LINKS
-> See first: [`juju` | Action](https://juju.is/docs/juju/action), [`juju` | Manage actions](https://juju.is/docs/juju/manage-actions), [`charmcraft` | Manage actions]()
--->
+> See first: {external+juju:ref}`Juju | Charm <action>`, {external+juju:ref}`Juju | Manage actions <manage-actions>`, {external+charmcraft:ref}`Charmcraft | Manage actions <manage-actions>`
 
 ## Implement the feature
 
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -295,7 +295,11 @@ autodoc_default_options = {
 
 # This config value contains the locations and names of other projects
 # that should be linked to in this documentation.
-intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3', None),
+    'juju': ('https://canonical-juju.readthedocs-hosted.com/en/latest/', None),
+    'charmcraft': ('https://canonical-charmcraft.readthedocs-hosted.com/en/stable/', None),
+    }
 
 # -- General configuration ---------------------------------------------------
 
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -298,7 +298,7 @@ autodoc_default_options = {
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'juju': ('https://canonical-juju.readthedocs-hosted.com/en/latest/', None),
-    'charmcraft': ('https://canonical-charmcraft.readthedocs-hosted.com/en/stable/', None),
+    'charmcraft': ('https://canonical-charmcraft.readthedocs-hosted.com/en/latest/', None),
     }
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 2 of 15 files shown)


docs/howto/manage-actions.md
```diff
@@ -1,9 +1,6 @@
 (manage-actions)=
 # How to manage actions
-
-<!-- UPDATE LINKS
-> See first: [`juju` | Action](https://juju.is/docs/juju/action), [`juju` | Manage actions](https://juju.is/docs/juju/manage-actions), [`charmcraft` | Manage actions]()
--->
+> See first: {external+juju:ref}`Juju | Charm <action>`, {external+juju:ref}`Juju | Manage actions <manage-actions>`, {external+charmcraft:ref}`Charmcraft | Manage actions <manage-charms>`
```
benhoyt, line 3: "See first" reads awkwardly to me. What about "Read first"?

benhoyt (commented):
Nice work -- thanks Dora (and Tony). Just one question about "See first" reading awkwardly (to me).

docs/howto/manage-actions.md
tmihoc: > "See first" reads awkwardly to me. What about "Read first"?

Open to either, but the current verb we use everywhere else is "See": "See also", "See more", "See first", "See next". Do you find all of those awkward or just this one?

tmihoc:
> There are a few links to sections in charmcraft docs that I think are not exported:
> 
> For these to work, I think we need to get the charmcraft docs to explicitly provide reference anchors for these.

I'm creating them in a parallel PR.

tonyandrewmeyer:
> I'm creating them in a parallel PR.

Ah, I have one open already, linked above: https://github.com/canonical/charmcraft/pull/2129

Feel free to do a different one if there's a better naming scheme or I've missed some or whatever and I can close mine.

tmihoc:
> > I'm creating them in a parallel PR.
> 
> Ah, I have one open already, linked above: [canonical/charmcraft#2129](https://github.com/canonical/charmcraft/pull/2129)
> 
> Feel free to do a different one if there's a better naming scheme or I've missed some or whatever and I can close mine.

My PR is already in review: https://github.com/canonical/charmcraft/pull/2128/files# 

Sorry, i should have mentioned this upfront.

docs/howto/manage-actions.md
benhoyt: Just this one. "See also" and "See more" are common and clear. But "See first" and "See next" I find unusual and unclear. I'm guessing "see first" means "read this first"? It's just not a phrase that I've heard before and it gives me pause. And "See next" (if I'm understanding what you mean by it) seems too close to "See also" -- more commonly for that I've seen just "Next", or maybe "Next up".

tmihoc: So, the problem is as follows. In any doc set there's a need to cross-reference. In the past we used to cross-reference without any system. Over the past few years I've been deliberate about keeping in-line references only between reference (=definition) docs. For everything else I have an out-of-line pattern with the phrases "See also", for 'lateral' kinds of links from, e.g., Manage models to the reference doc defining the notion of 'model', and "See more" for 'zoom-in' type links from, e.g., Add a model to details about the `juju add-model` command. This pattern isn't necessarily something you find anywhere else, but it's a pattern I've gotten good feedback on and that seems to be working exactly as intended.

image

Now, with the move to federated docs and the fact that for certain entities we have stories that start in Juju docs, are continued in Charmcraft docs, and then finished in Ops docs, it feels necessary to expand the "See ___" system in a way that would give people a clue about how the order in which the stories connect. E.g., the Manage configurations story in Charmcraft should tell people to look at Juju first and at Ops next.

image

If "first" and "next" don't work for you, I'm open to suggestion. I would emphasize however that the "See ___" system has worked well for now, so I'd prefer sticking with "See ___", and also that "Next up" doesn't really capture the intended meaning here.

On a different note: We should be able to launch the new docs next week (the web team is almost done with the entrypoint page). As such, I believe it's important to ensure that the various doc sets are properly connected. This PR achieves that. May I suggest we merge it as is? On my part I promise to keep thinking about this, to discuss it with the other TAs, and to also proactively seek user feedback on it.

benhoyt: Yes, it's totally fine to merge this now (and keep thinking about it). I guess I think it's not clear on its own, but in context it kinda works.

benhoyt (approved):

benhoyt:
We're happy to merge this, but we'd like to wait till the anchors in Charmcraft PR https://github.com/canonical/charmcraft/pull/2128 are merged, which should fix the docs build.

PR #1547, opened 2025-01-30
chore: update docs-deps step in HACKING.md for the release

dimaqq:
`tox -e docs-deps` produces subtly different result depending on which Python is used, as pip-compile uses that Python to resolve transient deps, so 3.10, 3.11, 3.12, 3.13 are all different.


benhoyt (approved):
David's away for another week, so reviewing this instead. This looks okay to me, at least as an interim solution. Maybe we'll switch to "uv everything" at some point soon...

tonyandrewmeyer (approved):

PR #1551, opened 2025-01-30
fix: require today's ops for today's ops-scenario

dimaqq:
and prep for release

the idea is that we'd:
- release this as `ops-scenario=7.1.1`
- yank `ops-scenario=7.1.0`

the users can then either stay on old ops and old ops-scenario, or upgrade both
the automatic version resolution should work, I think :)

### DIFF
--- a/testing/CHANGES.md
+++ b/testing/CHANGES.md
@@ -1,3 +1,9 @@
+# 7.1.1 - 30 Jan 2025
+
+## Fixes
+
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
> * yank `ops-scenario=7.1.0`

I'm not that convinced we need to yank, although not super opposed to it. If someone has moved to 7.1 already and done a more typical update of both ops and ops-scenario then there's no need to break that.

benhoyt (approved):
Thanks! I agree with Tony, probably not worth yanking.

Issue #1108, opened 2024-01-10
facilitate common exit hook pattern?


PietroPasotti:
Before the `collect_status` events were introduced, I could write:

```
class MyCharm(CharmBase):
    def __init__(self, framework):
        super().__init__(framework)
        
        for event in self.on.events().values():
            self.framework.observe(event, self._on_event)
```

and rest assured that `_on_event` would be called exactly once, no matter what event the charm was processing.

However, nowadays, `_on_event` will be called three times!

I wouldn't consider this a bug/regression because that pattern was not widely accepted/publicised, but it is still public API and many of my simpler tester charms that have a very simple holistic control logic implement it because it's quick and dirty.

No production charm I'm aware of is affected by this. Only testing code.

As a workaround, I can filter out the `collect_status` events and omit listening to those in my `observe` calls, but that is suddenly a lot of code for a semantically simple ask: whatever event I see, run this handler.

In some or many situations it would be acceptable to simply put the code in `__init__` instead of observing anything at all, since I want the code to execute unconditionally (and only once). But I can also imagine needing access to the event object.

Is it maybe a good idea to add a `self.framework.observe_any(self._handler)` or something similar to facilitate this use case? Or is it a use case we don't want to facilitate?

tonyandrewmeyer:
> Before the `collect_status` events were introduced, I could write:
> 
> ```
> class MyCharm(CharmBase):
>     def __init__(self, framework):
>         super().__init__(framework)
>         
>         for event in self.on.events().values():
>             self.framework.observe(event, self._on_event)
> ```
> 
> and rest assured that `_on_event` would be called exactly once, no matter what event the charm was processing.

As you note, the simple case is something like:

```python
class MyCharm(CharmBase):
    def __init__(self, framework):
        super().__init__(framework)
        self._on_event()
```

But there's no clean way to get an event object there, because the charm isn't yet handling an event. You said you can imagine wanting that object - does that mean you don't currently use it, even in the tester charms?

> As a workaround, I can filter out the `collect_status` events and omit listening to those in my `observe` calls, but that is suddenly a lot of code for a semantically simple ask: whatever event I see, run this handler.

I guess the main question I have is what use cases there are (especially outside tester charms) for "I want to run this bit of code for every hook event? That's a pretty wide range of things happening (e.g. stop and relation-changed don't seem like they would generally have common code, but maybe that's my inexperience?).

[...]

PR #1554, opened 2025-01-31
[POC] any-event

PietroPasotti:
brainstorming proof of concept API for introducing an 'any' event in ops

### Issue 1:
```
class MyCharm(ops.CharmBase):
    def __init__...
        ...
        self.reconcile()
    def reconcile(self):
        # holistic charm state reconciler
```
examples: 
- [parca-k8s-operator](https://github.com/canonical/parca-k8s-operator/blob/main/src/charm.py#L182) 
- 
### Issue 2:
```
class MyCharm(ops.CharmBase):
    def __init__...
        ...
        for event in [list of all possible events]:
            self.framework.observe(event, self._on_any)

    def _on_any(self, _event):
        # holistic charm state reconciler
```

### Proposed solution:
```
class MyCharm(ops.CharmBase):
    def __init__...
        ...
        self.framework.observe(self.on.any, self._on_any)

    def _on_any(self, _event):
        # holistic charm state reconciler
```

### DIFF
--- a/ops/_main.py
+++ b/ops/_main.py
@@ -488,6 +488,9 @@ class _Manager:
 
     def _emit(self):
         """Emit the event on the charm."""
+        # emit the 'any' event onto the charm
+        self._emit_charm_event("any")
+
         # TODO: Remove the collect_metrics check below as soon as the relevant
         #       Juju changes are made. Also adjust the docstring on
         #       EventBase.defer().
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -92,6 +92,21 @@ class _ContainerBaseDict(TypedDict):
 logger = logging.getLogger(__name__)
 
 
+class AnyEvent(EventBase):
+    """Event always emitted on the charm."""
+
+    def defer(self) -> NoReturn:
+        """The `any` event is not deferrable.
+
+        This is because these events are run alongside each event invocation,
+        so deferring would always end up simply doubling the work.
+
+        Raises:
+            RuntimeError: always.
+        """
+        raise RuntimeError('cannot defer `any` event')
+
+
 class HookEvent(EventBase):
     """Events raised by Juju to progress a charm's lifecycle.
 
@@ -1197,6 +1212,9 @@ class CharmEvents(ObjectEvents):
 
     # NOTE: The one-line docstrings below are copied from the first line of
     #       each event class's docstring. Please keep in sync.
+
+    any = EventSource(AnyEvent)
+    """Always triggered on the charm, regardless of the juju event being processed (see :class:`~ops.AnyEvent`)."""
 
     install = EventSource(InstallEvent)
     """Triggered when a charm is installed (see :class:`~ops.InstallEvent`)."""


benhoyt:
This is a duplicate of https://github.com/canonical/operator/issues/1108, and I think the reason we closed that one still mostly applies. However, we are seeing/promoting the holistic pattern more now, so a few additional thoughts:

* Doing it in `__init__` is okay, but is probably going to get you more events than you want. In addition, you probably don't want to reconcile on events like `secret-rotate` and other "non holistic" events. See [holistic vs delta doc](https://ops.readthedocs.io/en/latest/explanation/holistic-vs-delta-charms.html#when-to-use-the-holistic-approach).
* We still believe the for loop over explicit event names is the way to go. Part of the value of an event system is the producer (Juju, or sometimes ops) can produce new event types whenever it needs to, and the consumer will only get notified about those it's listening for. (Gustavo made this point fairly strongly when we were working on the `change-updated` stuff.)
* Even if we did something like this, what would "any" mean? You probably only want to listen to holistic-friendly events, which is roughly the same thing as defer-able. So perhaps an `observe_any_derrable` or some list which lists all the deferrable events?
* Another approach is opting into observing event groups, something like `self.framework.observe_groups(self._reconcile, relation=True, config=True)`. But this will still suffer from the problem of Juju adding new events to a group.

Regarding your solution in https://github.com/canonical/operator/issues/1108 and the three events you're seeing as a result of `collect_app_status` and `collect_unit_status`, you could filter out `LifecycleEvent` types. Something like:

```python
for event in self.on.events().values():
    if not issubclass(event.event_type, ops.LifecycleEvent):
         self.framework.observe(event, self._on_event)
```

I'll leave this issue open for comments for another week. And it might be good to have further discussion -- if so, let us know and we'll schedule a nice little discussion for Frankfurt.

PietroPasotti:
> * Doing it in `__init__` is okay, but is probably going to get you more events than you want. In addition, you probably don't want to reconcile on events like `secret-rotate` and other "non holistic" events. See [holistic vs delta doc](https://ops.readthedocs.io/en/latest/explanation/holistic-vs-delta-charms.html#when-to-use-the-holistic-approach).

I think the whole point of a well-written reconciler is that it truly is idempotent. There's no 'running it too often'. The only case in which there can be confusion is when the event doesn't match the state (e.g. you get a relation-removed, but the relation and its data are still readable).
I want it running on every single event, and possibly even more often than that (e.g. using a resurrect mechanism, in case the user has set update-status-hook-interval to something longer.

> * We still believe the for loop over explicit event names is the way to go. Part of the value of an event system is the producer (Juju, or sometimes ops) can produce new event types whenever it needs to, and the consumer will only get notified about those it's listening for. (Gustavo made this point fairly strongly when we were working on the `change-updated` stuff.)

I think this argument only holds for charms that don't use a reconciler. A holistic charm will only look at the parts of the state it knows, and if juju adds an event type it will only mean it will reconcile 'more often'. So long as the semantics of the state don't change we're safe.

> * Even if we did something like this, what would "any" mean? You probably only want to listen to holistic-friendly events, which is roughly the same thing as defer-able. So perhaps an `observe_any_derrable` or some list which lists all the deferrable events?

Side-note: we also realized along the way that the only domain where 'true delta charming' really is useful is machine charms: there 'start'/'remove' really mean something. In k8s world, where everything can churn at any moment and go up and down at any time, that's where holistic charming is more natural.
I don't think 'deferable-or-not' is the line we should be looking at. For charms like the ones we (o11y) write, even non-deferrable events are a good reason to run the reconciler.
The reason is that we don't really have a careful upgrade/shutdown path. Database charms are very different in that regard.
I wonder if we should be special-casing on the workload type instead. All charms I'm aware of except database charms don't care if they skip observing a 'remove' or a 'stop' or an 'upgrade'.
Could we call them stateful-workload-charms and talk about that instead?

> Regarding your solution in #1108 and the three events you're seeing as a result of `collect_app_status` and `collect_unit_status`, you could filter out `LifecycleEvent` types. Something like:

Yes, absolutely.

> I'll leave this issue open for comments for another week. And it might be good to have further discussion -- if so, let us know and we'll schedule a nice little discussion for Frankfurt.

I wonder if the workload-statefulness thing is a distinction that more people feel we should be philosophizing about; if so, I'd happily discuss it in frankfurt.



Since we already have a framework.on.commit we can (abuse) to do this, and the only difference with what we want is that we'd like our reconciler to run **before** all other events are emitted (and before the framework has closed all databases) rather than after... would you feel differently about adding a `framework.on.init` or similar?

benhoyt:
We discussed this briefly as a team, and we're going to close won't-fix on this one. We think an additional "pre" lifecycle event is somewhat interesting. But I think this proposal is best handled with further "holistic experimentation" like Carl's talking about, which could feed into an Ops 3.x design.

ca-scribner:
I realize I've arrived late :D but wanted to echo @PietroPasotti's comments.  Some charms benefit greatly from atomic event handling and those charms are served well by ops right now.  

Charms that are written as holistic reconcilers, like the ones @PietroPasotti (and I) make, are harder to write in the current event model because catching the proper events for a holistic charm requires detailed knowledge about ops.  Agreed this example [from @tonyandrewmeyer](https://github.com/canonical/operator/issues/1108#issuecomment-1885928739) to catch all the right events is simple:
```python
class MyCharm(CharmBase):
        super().__init__(framework)
        
        for event in (kind for kind in self.on.events().values() if isinstance(kind, ops.HookEvent)):
            self.framework.observe(event, self._on_event)
```
but I doubt many people outside charm tech would know its **correct** for the needs of a reconciler.  Catching the right events for a reconciler style charm should be accessible to beginners.

benhoyt:
Noted. I've added a link to this issue to an item we have on next cycle's roadmap for documenting holistic/reconciler charm patterns, for us to either reconsider this idea, or at least document the useful patterns.

PR #1559, opened 2025-02-03
fix: put the Juju version in the environment as well as the JujuContext

### LINKED ISSUE #1555: Scenario doesn't set JUJU_VERSION in os.environ
[This commit (49a9cec) in cos-lib](https://github.com/canonical/cos-lib/pull/121/commits/49a9cec45044e51fa0b93173a85e24f8cd308db9) is fixes our unit tests which is the result of a pin `ops[testing] < 2.18`.

A sample CI run is found [here](https://github.com/canonical/cos-lib/actions/runs/13061093223/job/36444016722#step:5:205) where:

```
self = <ops.model.RelationMapping object at 0x7fe1fd6ce520>
relation_name = 'peers'
ops/model.py:976: in _get_unique
    relations = self[relation_name]
```

which results in a `KeyError` because this does not exist.

tonyandrewmeyer:
Charms are expected to access the Juju environment variables via the mechanisms that ops provides. Generally, this is through the model and other attributes of the charm class. However, the Juju version is not available via the charm object, and instead is expected to be obtained by charms via `JujuVersion.from_environ()`.

Previously, we stopped Scenario from putting all the Juju environment variables in `os.environ`, and instead provided them directly to the `JujuContext` object. This works, except for the `JUJU_VERSION` case, as `from_environ` (as the name indicates) loads the version from `os.environ`, not from a `JujuContext` object.

This PR changes the behaviour so that `JUJU_VERSION` is set in both the context and the environment.

Fixes #1555.

### DIFF
--- a/testing/tests/test_runtime.py
+++ b/testing/tests/test_runtime.py
@@ -117,3 +117,19 @@ def test_env_clean_on_charm_error():
 
     # Ensure that the Juju environment didn't leak into the outside one.
     assert os.getenv("JUJU_REMOTE_APP", None) is None
+
+
+def test_juju_version_is_set_in_environ():
+    version = "2.9"
+
+    class MyCharm(ops.CharmBase):
+        def __init__(self, framework: ops.Framework):
+            super().__init__(framework)
+            framework.observe(self.on.start, self._on_start)
+
+        def _on_start(self, _: ops.StartEvent):
+            juju_version = ops.JujuVersion.from_environ()
+            assert juju_version == version
+
+    ctx = Context(MyCharm, meta={"name": "foo"}, juju_version=version)
+    ctx.run(ctx.on.start(), State())
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -6,6 +6,7 @@
 
 import copy
 import dataclasses
+import os
 import tempfile
 import typing
 from contextlib import contextmanager
@@ -319,6 +320,12 @@ class Runtime:
                 charm_root=temporary_charm_root,
             )
             juju_context = _JujuContext.from_dict(env)
+            # We need to set JUJU_VERSION in os.environ, because charms may use
+            # `JujuVersion.from_environ()` to get the (simulated) Juju version.
+            # All of the other environment variables are exposed to the charm
+            # through the framework machinery, so don't require this.
+            previous_juju_version = os.environ.get("JUJU_VERSION")
+            os.environ["JUJU_VERSION"] = env["JUJU_VERSION"]
 
             logger.info(" - entering ops.main (mocked)")
             from ._ops_main_mock import Ops  # noqa: F811
@@ -345,6 +352,10 @@ class Runtime:
                 ) from e
 
             finally:
+                if previous_juju_version is None:
+                    del os.environ["JUJU_VERSION"]
+                else:
+                    os.environ["JUJU_VERSION"] = previous_juju_version
                 logger.info(" - exited ops.main")
 
         context.emitted_events.extend(captured)


dimaqq:
Code-wise it's OK, but let's discuss at the stand-up first.

james-garner-canonical (commented):
This looks like it resolves the issue, though I wonder if there's a nice way to have the Juju version also be fully encapsulated by scenario for testing.

Maybe there's some benefit to encapsulating it at runtime as well? Could `ops` store `JUJU_VERSION` somewhere when it loads the other Juju context things, and could `from_environ` read it from there instead of from the environment directly?

I was going to approve with this comment, but since we're discussing it at daily I'll just leave this as a comment.

tonyandrewmeyer:
We discussed the approach as a team today and:

* We feel the `JujuVersion.from_environ().has_x` pattern isn't great. I'll open a separate PR to expose the JujuVersion (which we already have in the JujuContext object) from `Model`.
* Harness doesn't support setting a Juju version like Scenario does, so `from_environ` needs to read from the environment for backwards compatibility, so we can't simply replace it with reading from the context, even if the circular nature of that was acceptable.
* Monkeypatching `JujuVersion.from_environ` from Scenario rather than patching the environment doesn't seem significantly better, and in general we try to have Scenario run the actual Framework code where possible, which goes against that approach.

Example harness use:

```
openfga-operator/tests/unit/conftest.py
74:    return mocker.patch.dict(os.environ, {"JUJU_VERSION": "3.2.1"})

catalogue-k8s-operator/charm/tests/unit/test_charm.py
25:        os.environ["JUJU_VERSION"] = "3.0.3"

sunbeam-charms/ops-sunbeam/ops_sunbeam/test_utils.py
810:    os.environ["JUJU_VERSION"] = "3.4.4"

mysql-k8s-operator/tests/unit/test_upgrade.py
106:        with patch.dict(os.environ, {"JUJU_VERSION": "2.9.44"}):
118:        with patch.dict(os.environ, {"JUJU_VERSION": "3.1.5"}):

loki-k8s-operator/tests/unit/test_charm.py
119:        os.environ["JUJU_VERSION"] = "3.0.3"
179:        os.environ["JUJU_VERSION"] = "3.0.3"
273:        os.environ["JUJU_VERSION"] = "3.0.3"
310:        os.environ["JUJU_VERSION"] = "3.0.3"
404:        os.environ["JUJU_VERSION"] = "3.0.3"
447:        os.environ["JUJU_VERSION"] = "3.0.3"

glauth-k8s-operator/tests/unit/conftest.py
247:    return mocker.patch.dict(os.environ, {"JUJU_VERSION": "3.2.1"})
```

We concluded that putting `JUJU_VERSION` in the environment from Scenario is the best choice for now, and the new API will let us do this more cleanly in the future (once everyone moves to it).

testing/tests/test_runtime.py
```diff
@@ -117,3 +117,19 @@ def test_env_clean_on_charm_error():
...
+    class MyCharm(ops.CharmBase):
+        def __init__(self, framework: ops.Framework):
+            super().__init__(framework)
+            framework.observe(self.on.start, self._on_start)
+
+        def _on_start(self, _: ops.StartEvent):
+            juju_version = ops.JujuVersion.from_environ()
+            assert juju_version == version
+
+    ctx = Context(MyCharm, meta={"name": "foo"}, juju_version=version)
```
dimaqq, line 134: Minor: I think this would read clearer:

```suggestion
            assert ops.JujuVersion.from_environ() == "2.9"

    ctx = Context(MyCharm, meta={"name": "foo"}, juju_version="2.9")
```

dimaqq (approved):
Discussed at startup:
- use this environ patch
- in a separate PR
   - deprecate JujuVersion.from_environ
   - export .juju_version on Model

testing/tests/test_runtime.py
tonyandrewmeyer: I'm happy to collapse the 131 and 132 lines into a single line. I'd rather not repeat the `2.9` string, though.

james-garner-canonical (approved):

PR #1560, opened 2025-02-03
feat: support starting and stopping Pebble checks, and the checks enabled field

tonyandrewmeyer:
Add support for:

* The `startup` field in Pebble checks.
* The `start_checks` and `stop_checks` Pebble API calls.

Harness (and Scenario, mostly via re-using the Harness implementation) is adjusted to more closely simulate the Changes implementation of Pebble Checks, so that the 'if the change ID is the empty string, the check is inactive' behaviour can be simulated.

A subtle bug with notices and check-infos is also fixed: previously the mocked Pebble would gather all notices and check-infos from all containers in the state, instead of only those that are in the correct container.

[Pebble PR](https://github.com/canonical/pebble/pull/560)

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3386,10 +3386,15 @@ class _TestingPebbleClient:
             try:
                 info = self._check_infos[name]
             except KeyError:
+                status = (
+                    pebble.CheckStatus.INACTIVE
+                    if check.startup == pebble.CheckStartup.DISABLED
+                    else pebble.CheckStatus.UP
+                )
                 info = pebble.CheckInfo(
                     name,
                     level=check.level,
-                    status=pebble.CheckStatus.UP,
+                    status=status,
                     failures=0,
                     change_id=pebble.ChangeID(''),
                 )
@@ -3796,17 +3801,21 @@ class _TestingPebbleClient:
             if (level is None or level == info.level) and (names is None or info.name in names)
         ]
 
-    def start_checks(self, names: List[str]):
+    def start_checks(self, names: List[str]) -> List[str]:
         self._check_connection()
+        started: List[str] = []
         for name in names:
             if name not in self._check_infos:
                 raise self._api_error(404, f'cannot find check with name "{name}"')
             info = self._check_infos[name]
             if not info.change_id:
                 self._new_perform_check(info)
+                started.append(name)
+        return started
 
-    def stop_checks(self, names: List[str]):
+    def stop_checks(self, names: List[str]) -> List[str]:
         self._check_connection()
+        stopped: List[str] = []
         for name in names:
             if name not in self._check_infos:
                 raise self._api_error(404, f'cannot find check with name "{name}"')
@@ -3816,6 +3825,8 @@ class _TestingPebbleClient:
                 change.status = pebble.ChangeStatus.ABORT.value
                 info.status = pebble.CheckStatus.INACTIVE
                 info.change_id = pebble.ChangeID('')
+                stopped.append(name)
+        return stopped
 
     def notify(
         self,
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3187,7 +3187,6 @@ class _TestingPebbleClient:
                         failures=0,
                         threshold=3 if check.threshold is None else check.threshold,
                         startup=check.startup,
-                        change_id=pebble.ChangeID(str(uuid.uuid4())),
                     )
                     self._check_infos[name] = info
                 if not info.change_id:
@@ -3389,6 +3388,7 @@ class _TestingPebbleClient:
                     failures=0,
                     change_id=pebble.ChangeID(''),
                 )
+                self._check_infos[name] = info
             info.level = check.level
             info.threshold = 3 if check.threshold is None else check.threshold
             info.startup = check.startup
@@ -3837,6 +3837,7 @@ class _TestingPebbleClient:
             if info.change_id:
                 change = self._changes[info.change_id]
                 change.status = pebble.ChangeStatus.ABORT.value
+                info.status = pebble.CheckStatus.INACTIVE
                 info.change_id = pebble.ChangeID('')
 
     def notify(
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3175,37 +3175,42 @@ class _TestingPebbleClient:
             if startup == pebble.ServiceStartup.ENABLED:
                 self._service_status[name] = pebble.ServiceStatus.ACTIVE
 
+    def _new_perform_check(self, info: pebble.CheckInfo) -> pebble.Change:
+        now = datetime.datetime.now()
+        change = pebble.Change(
+            pebble.ChangeID(str(uuid.uuid4())),
+            pebble.ChangeKind.PERFORM_CHECK.value,
+            summary=info.name,
+            status=pebble.ChangeStatus.DOING.value,
+            tasks=[],
+            ready=False,
+            err=None,
+            spawn_time=now,
+            ready_time=None,
+        )
+        info.change_id = change.id
+        info.status = pebble.CheckStatus.UP
+        info.failures = 0
+        self._changes[change.id] = change
+        return change
+
     def replan_services(self, timeout: float = 30.0, delay: float = 0.1):
         for name, check in self._render_checks().items():
-            if check.startup != pebble.CheckStartup.DISABLED:
-                info = self._check_infos.get(name)
-                if info is None:
-                    info = pebble.CheckInfo(
-                        name=name,
-                        level=check.level,
-                        status=pebble.CheckStatus.UP,
-                        failures=0,
-                        threshold=3 if check.threshold is None else check.threshold,
-                        startup=check.startup,
-                    )
-                    self._check_infos[name] = info
-                if not info.change_id:
-                    info.status = pebble.CheckStatus.UP
-                    info.failures = 0
-                    now = datetime.datetime.now()
-                    change = pebble.Change(
-                        pebble.ChangeID(str(uuid.uuid4())),
-                        pebble.ChangeKind.PERFORM_CHECK.value,
-                        summary=name,
-                        status=pebble.ChangeStatus.DOING.value,
-                        tasks=[],
-                        ready=False,
-                        err=None,
-                        spawn_time=now,
-                        ready_time=now,
-                    )
-                    self._changes[change.id] = change
-                    info.change_id = change.id
+            if check.startup == pebble.CheckStartup.DISABLED:
+                continue
+            info = self._check_infos.get(name)
+            if info is None:
+                info = pebble.CheckInfo(
+                    name=name,
+                    level=check.level,
+                    status=pebble.CheckStatus.UP,
+                    failures=0,
+                    threshold=3 if check.threshold is None else check.threshold,
+                    startup=check.startup,
+                )
+                self._check_infos[name] = info
+            if not info.change_id:
+                self._new_perform_check(info)
         return self.autostart_services(timeout, delay)
 
     def start_services(
@@ -3393,21 +3398,8 @@ class _TestingPebbleClient:
             info.threshold = 3 if check.threshold is None else check.threshold
             info.startup = check.startup
             if info.startup != pebble.CheckStartup.DISABLED and not info.change_id:
-                info.status = pebble.CheckStatus.UP
-                info.failures = 0
-                change = pebble.Change(
-                    pebble.ChangeID(str(uuid.uuid4())),
-                    pebble.ChangeKind.PERFORM_CHECK.value,
-                    summary=name,
-                    status=pebble.ChangeStatus.DOING.value,
-                    tasks=[],
-                    ready=False,
-                    err=None,
-                    spawn_time=datetime.datetime.now(),
-                    ready_time=None,
-                )
-                self._changes[change.id] = change
-                info.change_id = change.id
+                self._new_perform_check(info)
+
 
     def _render_services(self) -> Dict[str, pebble.Service]:
         services: Dict[str, pebble.Service] = {}
@@ -3812,21 +3804,7 @@ class _TestingPebbleClient:
                 raise self._api_error(404, f'cannot find check with name "{name}"')
             info = self._check_infos[name]
             if not info.change_id:
-                info.status = pebble.CheckStatus.UP
-                info.failures = 0
-                change = pebble.Change(
-                    pebble.ChangeID(str(uuid.uuid4())),
-                    pebble.ChangeKind.PERFORM_CHECK.value,
-                    summary=name,
-                    status=pebble.ChangeStatus.DOING.value,
-                    tasks=[],
-                    ready=False,
-                    err=None,
-                    spawn_time=datetime.datetime.now(),
-                    ready_time=None,
-                )
-                self._changes[change.id] = change
-                info.change_id = change.id
+                self._new_perform_check(info)
 
     def stop_checks(self, names: List[str]):
         self._check_connection()
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3178,7 +3178,18 @@ class _TestingPebbleClient:
     def replan_services(self, timeout: float = 30.0, delay: float = 0.1):
         for name, check in self._render_checks().items():
             if check.startup != pebble.CheckStartup.DISABLED:
-                info = self._check_infos[name]
+                info = self._check_infos.get(name)
+                if info is None:
+                    info = pebble.CheckInfo(
+                        name=name,
+                        level=check.level,
+                        status=pebble.CheckStatus.UP,
+                        failures=0,
+                        threshold=3 if check.threshold is None else check.threshold,
+                        startup=check.startup,
+                        change_id=pebble.ChangeID(str(uuid.uuid4())),
+                    )
+                    self._check_infos[name] = info
                 if not info.change_id:
                     info.status = pebble.CheckStatus.UP
                     info.failures = 0
@@ -3403,7 +3414,6 @@ class _TestingPebbleClient:
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, service in layer.services.items():
-                # TODO: merge existing services https://github.com/canonical/operator/issues/1112
                 services[name] = service
         return services
 
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3176,6 +3176,13 @@ class _TestingPebbleClient:
                 self._service_status[name] = pebble.ServiceStatus.ACTIVE
 
     def replan_services(self, timeout: float = 30.0, delay: float = 0.1):
+        for name, check in self._render_checks().items():
+            if check.startup != pebble.CheckStartup.DISABLED:
+                info = self._check_infos[name]
+                if not info.change_id:
+                    info.status = pebble.CheckStatus.UP
+                    info.failures = 0
+                    info.change_id = pebble.ChangeID("replanned")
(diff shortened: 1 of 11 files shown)


ops/_private/harness.py
```diff
@@ -3176,6 +3176,26 @@ def autostart_services(self, timeout: float = 30.0, delay: float = 0.1):
                 self._service_status[name] = pebble.ServiceStatus.ACTIVE
 
     def replan_services(self, timeout: float = 30.0, delay: float = 0.1):
+        for name, check in self._render_checks().items():
+            if check.startup != pebble.CheckStartup.DISABLED:
```
dimaqq, line 3180: Is this for Scenario or only Harness?

tonyandrewmeyer: Both, Scenario re-uses most of the mock Pebble client.

Note that this is still draft, there's a bunch of stuff I haven't pushed yet, particularly around testing.

ops/_private/harness.py
```diff
@@ -3346,12 +3376,44 @@ def add_layer(
...
+                )
+                self._changes[change.id] = change
+                info.change_id = change.id
+
     def _render_services(self) -> Dict[str, pebble.Service]:
         services: Dict[str, pebble.Service] = {}
         for key in sorted(self._layers.keys()):
             layer = self._layers[key]
             for name, service in layer.services.items():
-                # TODO: merge existing services https://github.com/canonical/operator/issues/1112
```
tonyandrewmeyer, line 3354: Drive-by: that ticket is closed, and it looks like this is already done.

ops/pebble.py
```diff
@@ -1103,6 +1106,11 @@ def __init__(self, name: str, raw: Optional[CheckDict] = None):
         except ValueError:
             level = dct.get('level', '')
         self.level = level
+        try:
+            startup: Union[CheckStartup, str] = CheckStartup(dct.get('startup', ''))
```
benhoyt, line 1110: I realise we decided not to remove the union for existing fields, but for new fields shall we simplify and just use `startup: CheckStartup` (where one of the options is `''`)? It would avoid awkward types and awkward `isinstance` checks like we see below -- at least for the new field.

ops/pebble.py
```diff
@@ -1407,12 +1427,21 @@ class CheckInfo:
     This can be :attr:`CheckLevel.ALIVE`, :attr:`CheckLevel.READY`, or None (level not set).
     """
 
+    startup: Union[CheckStartup, str]
```
benhoyt, line 1430: I think the type here should definitely be just `CheckStartup`, right, as it's always going to be set when returned from Go in the check info?

Oh, I guess there's the case where it's not present (older versions of Pebble) -- but in that case, it's ENABLED, right?

ops/pebble.py
```diff
@@ -1442,10 +1471,17 @@ def __init__(
...
     ):
         self.name = name
         self.level = level
-        self.status = status
+        self.startup = startup
+        if change_id:
+            self.status = status
+        else:
+            # No change ID means the check is inactive, or that the user has an
+            # old version of Pebble. We assume the former.
```
benhoyt, line 1483: Hmm, I wonder if we can avoid this. If the `startup` field is present, we know it's a newer version of Pebble. Can we use that fact?

ops/pebble.py
```diff
@@ -1475,6 +1516,7 @@ def __repr__(self):
             'CheckInfo('
             f'name={self.name!r}, '
             f'level={self.level}, '
+            f'startup={self.startup}, '
```
benhoyt, line 1519: Ben, repeat after me: "dataclasses are good, use them next time". :-)

ops/pebble.py
```diff
@@ -2126,7 +2168,9 @@ def autostart_services(self, timeout: float = 30.0, delay: float = 0.1) -> Chang
         return self._services_action('autostart', [], timeout, delay)
 
     def replan_services(self, timeout: float = 30.0, delay: float = 0.1) -> ChangeID:
-        """Replan by (re)starting changed and startup-enabled services and wait for them to start.
+        """Replan by (re)starting changed and startup-enabled services and checks.
```
benhoyt, line 2171: It's a pity I called this `replan_services`, but oh well, at least it's called just `replan` for the `Container` method. Not worth deprecating/aliasing here, I think.

ops/pebble.py
```diff
@@ -2335,6 +2379,19 @@ def _wait_change_using_polling(
 
         raise TimeoutError(f'timed out waiting for change {change_id} ({timeout} seconds)')
 
+    def _checks_action(self, action: str, checks: Iterable[str]) -> FrozenSet[str]:
```
benhoyt, line 2382: I would prefer using `List[str]` here (and on the `Container` methods) instead of frozenset, as basic types seem nicer (and the caller could actually mutate it or `.sort` it if they want). Also, this would be the first use of frozenset in pebble.py or model.py.

Actually, that makes me think, can we make this actually an ordered list on the Go side, so it's guaranteed ordered by name? Basically add a `sort.Strings` call before returning from `StartChecks` and `StopChecks`. Those kinds of consistency guarantees are nice (for testing if nothing else). It would also mean an ordered `list[str]` is probably the better type anyway.

ops/pebble.py
```diff
@@ -3052,6 +3109,28 @@ def get_checks(
...
         return [CheckInfo.from_dict(info) for info in resp['result']]
 
+    def start_checks(self, checks: Iterable[str]) -> FrozenSet[str]:
+        """Start checks by name.
+
+        Args:
+            checks: Non-empty list of checks to start.
+
+        Returns:
+            Set of check names that were started.
```
benhoyt, line 3119: Can we clarify with something like "Checks that were already started won't appear in this list."? (Here and in model.py)

ops/_private/harness.py
```diff
@@ -3176,6 +3176,36 @@ def autostart_services(self, timeout: float = 30.0, delay: float = 0.1):
                 self._service_status[name] = pebble.ServiceStatus.ACTIVE
 
     def replan_services(self, timeout: float = 30.0, delay: float = 0.1):
+        for name, check in self._render_checks().items():
+            if check.startup != pebble.CheckStartup.DISABLED:
```
benhoyt, line 3180: Nit: maybe `if check.startup == DISABLED: continue` to avoid excess indentation?

ops/_private/harness.py
```diff
@@ -3743,6 +3805,41 @@ def get_checks(
...
+    def start_checks(self, names: List[str]):
+        self._check_connection()
+        for name in names:
+            if name not in self._check_infos:
+                raise self._api_error(404, f'cannot find check with name "{name}"')
+            info = self._check_infos[name]
+            if not info.change_id:
+                info.status = pebble.CheckStatus.UP
+                info.failures = 0
+                change = pebble.Change(
```
benhoyt, line 3817: Perhaps factor this out (occurs 3 times) into `change = self._perform_check_change()` or `change = self._new_change(pebble.ChangeKind.PERFORM_CHECK` or something?

benhoyt (changes requested):
This is good work -- thanks especially for updating the testing stuff (and testing the testing stuff!).

A few minor comments, some debatable, and one that would require a Pebble tweak.

ops/pebble.py
tonyandrewmeyer: Yeah, that was my thinking as well.

ops/pebble.py
tonyandrewmeyer: Ah nice suggestion, thanks!

ops/pebble.py
tonyandrewmeyer: Sure - I wondered about that but thought maybe it would be better to just be consistent. But happy to go with the simpler case.

ops/pebble.py
tonyandrewmeyer: > I would prefer using List[str] here (and on the Container methods) instead of frozenset, as basic types seem nicer

Interesting: I would consider frozenset a "basic type" :). 

> (and the caller could actually mutate it or .sort it if they want)

I don't think the caller should be mutating it, which is why I went with frozenset and not set. You're getting back a collection of check names that were started/stopped, why would you add/remove items from that collection? If you're later using this to have some other meaning, then I think the code would be clearer if you explicitly put in a `list(result)`.

Since these are check names ordering alphabetically isn't illogical, and I can see how that might be nice for displaying the results in a log message or similar. I feel that checks are not inherently ordered, though - they are not ordered in the layer/plan, for example, and they aren't started or stopped in alphabetical order, and I don't think they are run in alphabetical order. More importantly, you can't have duplicates in the results, and I think the return type should indicate that.

> Also, this would be the first use of frozenset in pebble.py or model.py.

I'm not sure why this matters? This isn't adding a dependency or even a new import or anything like that. frozensets have been in Python since 2.4, and aren't some obscure feature people aren't familiar with. There's no division in pebble.py, would that mean it's a bad idea to have ` / `? That seems the same level of additional complexity to me.

I agree that in pebble.Client, `List` is more correct, since the client is generally returning "raw" content from the Pebble API, and the API is limited to JSON.

However, for model.Container:
 * `get_services` returns a `ServiceInfoMapping` (with the type annotation saying it's a Mapping)
 * `get_checks` returns a `CheckInfoMapping` (with the type annotation saying it's a CheckInfoMapping)
 * `list_files` does return a `List`, but I think people expect directory listings to be lists, even though there's not often any logical ordering, and often duplicates are not allowed
 * `get_notices` returns a `List`, but I'd argue (very belatedly!) that's not the best choice

Note that `get_services` and `get_checks` return frozen containers, so a frozen return for start/stop would be consistent with that (although getting data and the response from an action are not identical actions).

I think:
* `get_checks` should probably be annotated with Mapping to be consistent with `get_services` - from a typing point of view it's important that it's a mapping, and `CheckInfoMapping` is just the concrete implementation of that.
* `get_notices` should return a frozenset and be typed as Collection (or at least return a frozen list and be typed as sequence), but it's too late to change that
* `start_checks` and `stop_checks` should return a frozenset and be typed as Collection (I think I could be convinced that it should be a frozen list and be typed as sequence)

benhoyt: Thanks -- you've convinced me that frozenset is okay and basic. :-)

Agreed that we should leave it as List[str] for the more "raw" pebble.py.

Per voice discussion: I was ready to agree that frozenset was the right choice for the Container methods, but then the order is useful here, as we agreed it would mostly be used for logging, so having a stable order is useful. So our decision was List[str] here too.

Thanks for the Pebble tweak over at https://github.com/canonical/pebble/pull/572 to order the result. Such stability is always nicer for tests/logs.

benhoyt (approved):
Looks good to me know, thanks!

testing/tests/test_e2e/test_state.py
```diff
@@ -251,6 +253,17 @@ def pre_event(charm: CharmBase):
...
 
+@pytest.mark.parametrize("id", ("", None, "28"))
+def test_checkinfo_changeid(id: Optional[str]):
+    info = CheckInfo("foo", change_id=ops.pebble.ChangeID(id))
+    if id is None:
+        assert info.change_id, "None should result in a random change_id"
+        info2 = CheckInfo("foo")
+        assert info.change_id != info2.change_id
+    else:
+        assert info.change_id == ops.pebble.ChangeID(id)
```
dimaqq, line 264: If None is important, let's split this test into two, as there's almost no commonality vis a vis "" or "28.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -546,3 +546,89 @@ def _on_bar_check_failed(self, event):
...
+    ctx = Context(MyCharm, meta={"name": "foo", "containers": {"foo": {}}})
+    container = Container("foo", can_connect=True)
+    state_out = ctx.run(
+        ctx.on.pebble_ready(container), state=State(containers={container})
+    )
+    chk1_info = state_out.get_container("foo").get_check_info("chk1")
+    assert chk1_info.status == pebble.CheckStatus.UP
+
+
+def test_pebble_start_check():
```
dimaqq, line 572: This test could be combined with the test above and parametrised on chk1 plan value.

That's assuming I'm not blind and the rest of the test is actually the same.

testing/src/scenario/state.py
```diff
@@ -1047,6 +1069,13 @@ def get_filesystem(self, ctx: Context) -> pathlib.Path:
         """
         return ctx._get_container_root(self.name)
 
+    def get_check_info(self, name: str) -> CheckInfo:
+        """Get the check info for a check by name."""
+        for check_info in self.check_infos:
+            if check_info.name == name:
```
dimaqq, line 1075: 🤔 `check_infos` is a frozen set.
CheckInfo doesn't implement custom `__hash__` by name only.
So, there could be multiple infos with the same name and let's say different levels, statuses or ids.

Is it meaningful to return a "random" check info in such case?
Or are we relying on unique names by construction somehow?

testing/src/scenario/state.py
```diff
@@ -865,13 +865,17 @@ class CheckInfo(_max_posargs(1)):
...
+
     status: pebble.CheckStatus = pebble.CheckStatus.UP
     """Status of the check.
 
     :attr:`ops.pebble.CheckStatus.UP` means the check is healthy (the number of
     failures is fewer than the threshold), :attr:`ops.pebble.CheckStatus.DOWN`
     means the check is unhealthy (the number of failures has reached the
-    threshold).
+    threshold), and :attr:`ops.pebble.CheckStatus.INACTIVE` means the check is
+    not currently running.
```
dimaqq, line 878: It's a tad unclear to me whether "not currently running" means:
- disabled, or
- last check finished, next check not started yet

test/test_pebble.py
```diff
@@ -2886,13 +2921,15 @@ def test_get_checks_all(self, client: MockClient):
...
                     'threshold': 2,
+                    'change-id': '1',
                 },
                 {
                     'name': 'chk2',
                     'level': 'alive',
                     'status': 'down',
                     'failures': 5,
                     'threshold': 3,
+                    'change-id': '3',
```
dimaqq, line 2932: Are these ids important?
The test doesn't seem to validate these values.

ops/pebble.py
```diff
@@ -2335,6 +2369,19 @@ def _wait_change_using_polling(
 
         raise TimeoutError(f'timed out waiting for change {change_id} ({timeout} seconds)')
 
+    def _checks_action(self, action: str, checks: Iterable[str]) -> List[str]:
+        if isinstance(checks, (str, bytes)) or not hasattr(checks, '__iter__'):
+            raise TypeError(f'checks must be of type Iterable[str], not {type(checks).__name__}')
```
dimaqq, line 2374: If each check name is a `str`, why do we type-check against `bytes`?
I recall there's a lot of `str/bytes` duality in pebble, though I thought that was merely historic.

ops/pebble.py
```diff
@@ -2335,6 +2369,19 @@ def _wait_change_using_polling(
...
+
+        checks = tuple(checks)
+        for chk in checks:
+            if not isinstance(chk, str):
+                raise TypeError(f'check names must be str, not {type(chk).__name__}')
+
+        body = {'action': action, 'checks': checks}
+        resp = self._request('POST', '/v1/checks', body=body)
```
dimaqq, line 2382: P.S. If any check name is actually bytes, it will not get serialised as JSON, raising a TypeError instead.

ops/pebble.py
```diff
@@ -1461,20 +1485,28 @@ def from_dict(cls, d: _CheckInfoDict) -> CheckInfo:
             status = CheckStatus(d['status'])
         except ValueError:
             status = d['status']
+        change_id = ChangeID(d['change-id']) if 'change-id' in d else None
```
dimaqq, line 1488: Perhaps:

```suggestion
        change_id = ChangeID(d.get('change-id'))
```

ops/pebble.py
```diff
@@ -1461,20 +1485,28 @@ def from_dict(cls, d: _CheckInfoDict) -> CheckInfo:
...
+        if not change_id and 'startup' in d:
+            # This is a version of Pebble that supports stopping checks, which
+            # means that the check is inactive if it has no change ID.
+            logger.debug('Check %s has raw status %s, but is inactive.', d['name'], status)
```
dimaqq, line 1492: I'm a bit unclear if this is an expected condition or merely hypothetical.

dimaqq (approved):
Looks good, leaving minor comments about docstrings and tests.

testing/src/scenario/state.py
tonyandrewmeyer: > It's a tad unclear to me whether "not currently running" means:
> 
> * disabled, or
> * last check finished, next check not started yet

A check never "finishes" (individual changes or tasks might), it's always running (unless it's stopped). I'll try to tweak the wording without diverging from the Pebble terms or saying "inactive means inactive" :)

ops/pebble.py
tonyandrewmeyer: That would result in `ChangeID(None)` if `change-id` wasn't present in the dictionary, which won't work, because `ChangeID` is a string subclass.

ops/pebble.py
tonyandrewmeyer: This is definitely expected - it's the case whenever you have stopped a check. Would you rather the debug line is removed? It was handy while I was testing things, but shouldn't be generally useful.

test/test_pebble.py
tonyandrewmeyer: It's important that they aren't the empty string or missing (or None, although that shouldn't happen). I can add in validation that they are as expected, although that is covered by other tests.

ops/pebble.py
tonyandrewmeyer: This was to be consistent with `_services_action` and `send_signal`.

However, you're right - you can get a series of str via `for x in some_str` but you can't get a series of str via `for x in some_bytes`, so it's more useful to error out in the first test than when looking at what would be the first byte.

ops/pebble.py
tonyandrewmeyer: That couldn't happen, though, because there's a check for each individual check name being a string above this (so that you get an error if you pass in `list[int]` or something).

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer: They are testing distinct behaviours, though, so my preference is to keep them separate. However, they are very similar as you say - I've adjusted them instead so that the start check is more clearly testing two different parts, first that a startup:disabled check doesn't start, and then that it can be started.

ops/pebble.py
dimaqq: Well if I could wish for things... I would rather Pebble version detection was more explicit and Pebble features handling would be similar to Juju features handling.

In the short term though, yes I think the debug line should not be there, as DEBUG level messages are forwarded to Juju by default.

ops/pebble.py
dimaqq: Ah! sorry that somehow escaped my attention.

ops/pebble.py
tonyandrewmeyer: > I would rather Pebble version detection was more explicit and Pebble features handling would be similar to Juju features handling.

Yeah, I had wondered for a bit if I needed to do a `system_info()` call but decided against it, and then Ben pointed out that the presence/absence of the `startup` field was enough to determine whether the minimal version was met in this case.

It does seem like having the Pebble `Client` know which version it's connected to could be generally useful, although you don't have to be able to connect when instantiating the class, so it'd have to be able to be "don't know" for some of the time.

> In the short term though, yes I think the debug line should not be there, as DEBUG level messages are forwarded to Juju by default.

Ah good point, I had forgotten about that. Will do.

PR #1563, opened 2025-02-04
feat: expose the Juju version via Model objects

tonyandrewmeyer:
Charm code currently needs to do `JujuVersion.from_environ()` to get the Juju version. This isn't very 'ops-y' - we expose all the other Juju state through the `Model` object, and it also means that charms instantiate `JujuVersion` objects multiple times - even though the framework has actually done this already when parsing the Juju environment variables.

This PR exposes the Juju version via `Model.juju_version`, and deprecates the `JujuVersion.from_environ()` method (actual removal will presumably be in 3.0, a long time off).

This also centralises all the Juju environment variable processing in the `JujuContext` object.

### DIFF
--- a/ops/jujuversion.py
+++ b/ops/jujuversion.py
@@ -16,6 +16,7 @@
 
 import os
 import re
+import warnings
 from functools import total_ordering
 from typing import Union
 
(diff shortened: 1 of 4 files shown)


benhoyt (approved):
This looks like a good improvement to me, thanks!

dimaqq (approved):

PR #1573, opened 2025-02-11
fix: expose the mocked Juju environment variables via os.environ again

### LINKED ISSUE #1572: ops-testing should expose the mocked JUJU_ environment variables in os.environ
A recent change to how envvars are set before charm initialization in `ops.testing._runtime` has broken some of our unittests.
The reason is that [our charm](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/110) uses a decorator that patches the charm's `__init__` method to set up some stuff and [as part of that polls os.environ to extract the `JUJU_DISPATCH_PATH`](https://github.com/canonical/tempo-k8s-operator/blob/main/lib/charms/tempo_k8s/v1/charm_tracing.py#L505), which, however, is no longer set on the environment at that point in time.

I think in general it shouldn't be expected that charms only use the framework to access data provided by envvars. 
I know of ops.Event-agnostic code, for example, that uses the dispatch path from the environment to determine whether we are in a relation-broken context.

In general I think we should aim at keeping the unittesting context a charm is executed with as close as possible to the one it will have when running on juju. On juju there'll be envvars, why shouldn't there be the same ones when testing locally?

tonyandrewmeyer:
Rather than only put `JUJU_VERSION` into `os.environ`, copy the entire contents of the mock Juju environment variables (and after the run, put the environment dictionary back to the same state it was originally).

In general, we feel that charms should be using ops to access this information, rather than going directly to the raw environment variable, but this was too large a change to make in a refactoring bugfix release, and should wait for an 8.x release (and also ensuring that there are reasonable ways to get everything from ops - for example, the event name is perhaps too tricky to get globally right now).

Fixes #1572

### DIFF
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -352,14 +352,10 @@ class Runtime:
                 ) from e
 
             finally:
-                for key, value in os.environ.copy().items():
-                    if key in previous_env:
-                        os.environ[key] = value
-                    else:
+                for key in tuple(os.environ):
+                    if key not in previous_env:
                         del os.environ[key]
-                for key, value in previous_env.items():
-                    if key not in os.environ:
-                        os.environ[key] = value
+                os.environ.update(previous_env)
                 logger.info(" - exited ops.main")
 
         context.emitted_events.extend(captured)
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -322,10 +322,10 @@ class Runtime:
             juju_context = _JujuContext.from_dict(env)
             # We need to set JUJU_VERSION in os.environ, because charms may use
             # `JujuVersion.from_environ()` to get the (simulated) Juju version.
-            # All of the other environment variables are exposed to the charm
-            # through the framework machinery, so don't require this.
-            previous_juju_version = os.environ.get("JUJU_VERSION")
-            os.environ["JUJU_VERSION"] = env["JUJU_VERSION"]
+            # For consistency, we put all the other ones there too, although we'd
+            # like to change this in the future.
+            previous_env = os.environ.copy()
+            os.environ.update(env)
 
             logger.info(" - entering ops.main (mocked)")
             from ._ops_main_mock import Ops  # noqa: F811
@@ -352,10 +352,14 @@ class Runtime:
                 ) from e
 
             finally:
-                if previous_juju_version is None:
-                    del os.environ["JUJU_VERSION"]
-                else:
-                    os.environ["JUJU_VERSION"] = previous_juju_version
+                for key, value in os.environ.copy().items():
+                    if key in previous_env:
+                        os.environ[key] = value
+                    else:
+                        del os.environ[key]
+                for key, value in previous_env.items():
+                    if key not in os.environ:
+                        os.environ[key] = value
                 logger.info(" - exited ops.main")
 
         context.emitted_events.extend(captured)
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Code change looks good to me -- thanks. Two questions:

1) Have you tested that this fixes the broken charm tests mentioned in #1572? If not, could we double-check that and note it here.
2) Do you think we should do a point release of ops/ops-scenario after this is merged?

tonyandrewmeyer:
> 1. Have you tested that this fixes the broken charm tests mentioned in [ops-testing should expose the mocked JUJU_ environment variables in os.environ #1572](https://github.com/canonical/operator/issues/1572)?

Yes, although it also needs:

```diff
--- a/tests/scenario/test_charm_tracing.py
+++ b/tests/scenario/test_charm_tracing.py
@@ -29,7 +29,7 @@ from charms.tempo_coordinator_k8s.v0.tracing import (
 from ops import EventBase, EventSource, Framework
 from ops.charm import CharmBase, CharmEvents
 from scenario import Context, Relation, State
-from scenario.runtime import UncaughtCharmError
+from scenario.errors import UncaughtCharmError
 
 logger = logging.getLogger(__name__)
 
```

The only other charm that I have locally that needs this is `tempo-k8s-operator` (which is archived). We could re-export things from `_runtime` to `runtime` but I feel like that would somewhat negate the point of moving everything to the private module anyway. You can get `UncaughtCharmError` from `scenario.errors` in 7.0.0, the recent change is *not* being able to also get it from `scenario.runtime`. My preference here would be to keep things as they are.

> 2. Do you think we should do a point release of ops/ops-scenario after this is merged?

I feel this ties into the general "do we wait for the end of the month" discussion. This doesn't seem as urgent as breaking deferring in some cases - it's only testing, and it's only this one lib as far as we know - but I'm also not opposed to releasing fixes as we get to them.

testing/src/scenario/_runtime.py
```diff
@@ -352,10 +352,14 @@ def exec(
...
-                else:
-                    os.environ["JUJU_VERSION"] = previous_juju_version
+                for key, value in os.environ.copy().items():
+                    if key in previous_env:
+                        os.environ[key] = value
+                    else:
+                        del os.environ[key]
+                for key, value in previous_env.items():
+                    if key not in os.environ:
+                        os.environ[key] = value
```
james-garner-canonical, line 362:
```suggestion
                for key, value in previous_env.items():
                        os.environ[key] = value
                for key in tuple(os.environ):
                    if key not in previous_env:
                        del os.environ[key]
```

james-garner-canonical (approved):

testing/src/scenario/_runtime.py
james-garner-canonical: Oh hold up, how about this:
```suggestion
                for key in tuple(os.environ):
                    if key not in previous_env:
                        del os.environ[key]
                os.environ.update(previous_env)
```

benhoyt:
@tonyandrewmeyer Regarding whether we ship a new version, it's debatable, but I'd say this falls into "regression" territory and so we should ship a new (patch?) version.

The scenario.errors vs scenario.runtime change, I agree with you there. That would be an easy one for them to fix -- or just avoid -- in their tests (the environment one less so).

tonyandrewmeyer:
> @tonyandrewmeyer Regarding whether we ship a new version, it's debatable, but I'd say this falls into "regression" territory and so we should ship a new (patch?) version.

Ok. At the moment the only changes in main are docs so I can do this without needing to create a separate branch.

testing/src/scenario/_runtime.py
tonyandrewmeyer: Ah, yes, that's much nicer, thanks!

Issue #817, opened 2022-08-22
Add item assignment support for 'RelationData' object


sed-i:
It is common to see the following relation data update patterns in charm code:

```python
for relation in self._charm.model.relations[self._relation_name]:
	relation.data[self._charm.app]["first"] = "foo"
	relation.data[self._charm.app].update({"second": "bar"})
```

However, attempting to assign relation data in a "single source of truth" approach like so

```python
app_data = {"first": "foo", "second": "bar"}
for relation in self._charm.model.relations[self._relation_name]:
	relation.data[self._charm.app] = app_data
```

errors out with `TypeError: 'RelationData' object does not support item assignment`, and indeed `__setitem__` [isn't implemented](https://github.com/canonical/operator/blob/ac7b9f0944d7a313234b43ebda95df483dd17329/ops/model.py#L856).

It could be handy if it was possible to overwrite the entire thing, instead of relying on `update` + manual tidying up of leftovers.

That being said, NOT supporting item assignments may have the advantage of preventing unintentional overwrites in some cases.

PietroPasotti:
An issue I can see with that is that there'd be an asymmetry between `__getitem__(item:str) -> RelationDataContents` and `__setitem__(str, Dict[str,str])`. One might then reasonably expect to get a dict instead of `RelationDataContents`.

So, if we go for this, I'd be in favour of adding this functionality as `RelationDataContents.replace()`.

Related: I often see various implementations of `wipe_all_data()`. Which we could implement as `__delitem__`, or, better, as a `RelationDataContents.wipe()` (or `clear`, `erase`, etc...) method.

PietroPasotti:
@niemeyer would you be on board with adding the following two methods?

```
RelationDataContents.clear()  # deletes all key/value pairs
RelationDataContents.replace(items: Dict[str,str])  # equivalent to clear(); update(items)
```

rwcarlsen:
An issue with using `__setitem__` is ambiguity w.r.t. existing keys - are they erased?  Are they merged/added to the new ones?  I'm not convinced this is worth saving the one `for k, v in app_data.items():` line

benhoyt:
Per Madrid discussion, this seems reasonable (to add the `__setitem__` assignment ability). The raw dict would be validated and converted to a proper `RelationDataContent` object. Let's do a proof of concept at some point and discuss further.

benhoyt:
Didn't get to this in 24.10. We'd like to consider this as part of the relation data validation work planned for 25.04

[...]

PR #1578, opened 2025-02-13
feat: multiple possible implementations of assignment feature

james-garner-canonical:
This draft PR sketches 3 approaches to resolving [#817](https://github.com/canonical/operator/issues/817).

v0: Keep a reference to the `backend` in `RelationData`. Use it in `__setitem__` to construct a new `RelationDataContent` object, and add this to `RelationData` with the provided `key`. But do we want to support adding new entries like this?

v1: In `__setitem__`, raise a `KeyError` if `key` isn't already stored. Get the current `RelationDataContent` for `key`, call `RelationDataContent.clear()` (already provided by inheritance from `MutableMapping` -- the provided implementation will call `RelationDataContent`'s custom `__delitem__`), and then call `RelationDataContent.update(...)`.

v2: Just add the convenience method `RelationDataContent.replace`, which just calls `clear` and `update`. `relation_data['foo'] = {...}` not supported, instead call `relation_data['foo'].replace({...})`. Same semantics as v1, but no surprising than `KeyError` on `__setitem__`.

What would be the best way to test this, btw? Just unit tests (using scenario)?

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1760,6 +1760,7 @@ class RelationData(Mapping[Union['Unit', 'Application'], 'RelationDataContent'])
 
     def __init__(self, relation: Relation, our_unit: Unit, backend: '_ModelBackend'):
         self.relation = weakref.proxy(relation)
+        self._backend = backend  # for __setitem__ v0
         self._data: Dict[Union[Unit, Application], RelationDataContent] = {
             our_unit: RelationDataContent(self.relation, our_unit, backend),
             our_unit.app: RelationDataContent(self.relation, our_unit.app, backend),
@@ -1784,6 +1785,19 @@ class RelationData(Mapping[Union['Unit', 'Application'], 'RelationDataContent'])
 
     def __getitem__(self, key: Union['Unit', 'Application']) -> 'RelationDataContent':
         return self._data[key]
+
+    def __setitem__(self, key: Union['Unit', 'Application'], value: Dict[str, str]) -> None:
+        # v0 -- make a new RelationDataContent on assignment
+        content = RelationDataContent(relation=self.relation, entity=key, backend=self._backend)
+        content.update(value)
+        self._data[key] = content
+        return
+        # v1 -- replace contents of existing RelationDataContent object
+        if key not in self._data:
+            raise KeyError(key, 'Assignment only supported for known keys.')
+        content = self._data[key]
+        content.clear()
+        content.update(value)
 
     def __repr__(self):
         return repr(self._data)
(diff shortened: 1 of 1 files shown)


benhoyt:
My take would be option v1, because `RelationData.__init__` already sets up keys for the apps and units, so they should exist already.

I suppose the type on the `value` argument could be `Mapping[str, str]`?

ops/model.py
```diff
@@ -1785,6 +1786,19 @@ def __iter__(self):
     def __getitem__(self, key: Union['Unit', 'Application']) -> 'RelationDataContent':
         return self._data[key]
 
+    def __setitem__(self, key: Union['Unit', 'Application'], value: Dict[str, str]) -> None:
+        # v0 -- make a new RelationDataContent on assignment
+        content = RelationDataContent(relation=self.relation, entity=key, backend=self._backend)
+        content.update(value)
+        self._data[key] = content
```
tonyandrewmeyer, line 1793: I feel like this will end up being problematic because people might keep references to the `RelationDataContent` object, and you could then end up with two with different values, both meant to be accurately referring to what is/will be in Juju. There isn't a simple way to say "your old object is now obsolete".

ops/model.py
```diff
@@ -1785,6 +1786,19 @@ def __iter__(self):
...
+        return
+        # v1 -- replace contents of existing RelationDataContent object
+        if key not in self._data:
+            raise KeyError(key, 'Assignment only supported for known keys.')
```
tonyandrewmeyer, line 1797: Same comment as Ben here: the keys are app/unit, so it seems like they would always be present, unless I'm missing something.

You would want to avoid things like `relation.data['foo'] = {...}` or relation.data[Unit('made-up-unit')] = {...}`, but I think you'd want the error message to be clearer than just "known keys".

ops/model.py
```diff
@@ -1785,6 +1786,19 @@ def __iter__(self):
     def __getitem__(self, key: Union['Unit', 'Application']) -> 'RelationDataContent':
         return self._data[key]
 
+    def __setitem__(self, key: Union['Unit', 'Application'], value: Dict[str, str]) -> None:
```
tonyandrewmeyer, line 1789: I think this would be better as a `Mapping[str, str]` than `Dict`. For one thing, that makes the `__getitem__` vs. `__setitem__` types a bit more similar, so I think you would be less likely to expect to get a `dict` from `__getitem__`, and more likely to expect a `Mapping`, which is what you get (although it's more specific and also mutable).

tonyandrewmeyer (commented):
I agree that v1 seems best, although v2 is ok.

For testing, if you mean e2e style testing where you actually have a charm, then I'd write a basic charm that uses the one of these you pick to update the data (perhaps just an action that copies the event params into the unit databag, and another action that gets the unit databag as the results). You can then write a unit test that makes sure it works, yes, but also you can test it our for real.

In terms of tests for the feature in ops itself, it looks like test_model mostly uses Harness. So I guess you could just do a similar thing, or a similar thing with Scenario. I think you could probably also just create a RelationData object yourself and test things in the usual way.

james-garner-canonical:
Closing this draft for #1590 1586

PR #1583, opened 2025-02-19
docs: small updates to machine charm tutorial

dwilding:
This PR is for capturing small updates to [Write your first machine charm](https://ops.readthedocs.io/en/latest/tutorial/write-your-first-machine-charm.html) while I'm working on creating my own sample machine charm.

Summary of changes:

- Adjusted the "What you'll need section" to tidy up the language & formatting, and to clarify that familiarity with Juju is not an absolute requirement.

- When using `curl` to test the service after the initial deployment, we don't need to `juju exec` into the unit first. Instead, we can do `curl http://10.122.219.101:8080`, where 10.122.219.101 is the IP of the unit. I've updated the command and the preceding text to match.

- Added missing context to the note that comes after using `curl` to test the service. This note was supposed to say that the reader should follow the four steps in case something went wrong after deploying the charm. I added this context.

    I also removed a commented-out part about inspecting hooks. In the newly-added context, I mention about `juju debug-log`, which is probably more useful than inspecting hooks.

**[Preview doc](https://ops--1583.org.readthedocs.build/en/1583/tutorial/write-your-first-machine-charm.html)**

### DIFF
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -243,8 +243,8 @@ Online
 ```{note}
 
 If the Juju status doesn't look right, for example if you see an "error" status instead of "active",
-there might be an issue with the charm code. You can use a debug session (`juju debug-log`) to look
-for error reports from Python. After you've identified the issue:
+there might be an issue with the charm code. You can use a debug session (`juju debug-log`) to get
+more detailed information about the issue. After you've identified the issue:
 
 1. Fix the code in `src/charm.py`.
 2. Rebuild the charm: `charmcraft pack`.
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -17,7 +17,7 @@ When it's available in Juju.
 - Familiarity with Linux.
 - Familiarity with the Python programming language, including Object-Oriented Programming and event handlers.
 
-It will also help if you're familiar with Juju , but don't worry if you're new to Juju. This tutorial will guide you though each step.
+It will also help if you're familiar with Juju , but don't worry if you're new to Juju. This tutorial will guide you through each step.
 
 **What you'll do:**
 
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -17,7 +17,7 @@ When it's available in Juju.
 - Familiarity with Linux.
 - Familiarity with the Python programming language, including Object-Oriented Programming and event handlers.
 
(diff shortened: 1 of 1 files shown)


docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -12,10 +12,12 @@ When it's available in Juju.
...
-- A workstation, e.g., a laptop, with amd64 architecture and which has sufficient resources to launch a virtual machine with 4 CPUs, 8 GB RAM, and 50 GB disk space
-- Familiarity with Linux
-- Familiarity with Juju.
-- Familiarity with object-oriented programming in Python
+
+- A workstation. For example, a laptop with an amd64 architecture. You'll need sufficient resources to launch a virtual machine with 4 CPUs, 8 GB RAM, and 50 GB disk space.
+- Familiarity with Linux.
+- Familiarity with the Python programming language, including Object-Oriented Programming and event handlers.
+
+It will also help if you're familiar with Juju , but don't worry if you're new to Juju. This tutorial will guide you through each step.
```
tonyandrewmeyer, line 20:
```suggestion
It will also help if you're familiar with Juju, but don't worry if you're new to Juju. This tutorial will guide you through each step.
```

docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -230,39 +232,26 @@ Machine  State    Address         Inst id        Base          AZ  Message
...
 
 ```{note}
 
-1. Fix the code in `src/charm.py`.
-2. Rebuild the charm: `charmcraft pack`
-3. Refresh the application from the repacked charm: `juju refresh microsample --path=./microsample-vm_ubuntu-22.04-amd64.charm --force-units`
-4. Let the model know the issue is resolved (fixed): `juju resolved microsample/0`.
+If the Juju status doesn't look right, for example if you see an "error" status instead of "active",
+there might be an issue with the charm code. You can use a debug session (`juju debug-log`) to look
+for error reports from Python. After you've identified the issue:
```
tonyandrewmeyer, line 247: "error reports from Python" doesn't seem right to me. I guess in some ways it's Python providing the error, but in other ways it's ops or Juju.

What about something like "to get more detailed information about the problem. After ..."?

tonyandrewmeyer (approved):
Looks good, thanks!

A couple of small suggestions, but happy with it as-is.

docs/tutorial/write-your-first-machine-charm.md
dwilding: Nice suggestion, thanks! I've updated in 2e85081 (using "issue" instead of "problem" to stay consistent).

IronCore864 (approved):

PR #1585, opened 2025-02-24
ci: run tox static environment for charms that have it

james-garner-canonical:
This PR adds a steps to the `charm-tests` job defined in the `published-charm-tests` workflow to run the charm's `static` `tox` environment. The first step checks if the `static` environment exists, and the second step runs it if it does exist. Additionally, output and step ids are added to allow these steps to runeven if the unit test step fails, but not if there was a failure in an earlier step (installing dependencies, etc.) and not if the job was cancelled. Therefore a given job may be marked as a failure overall if both or either of the `static` and `unit` tests fail (if both fail, two failures will be listed for the job on the summary page for the workflow).

### DIFF
--- a/.github/workflows/published-charms-tests.yaml
+++ b/.github/workflows/published-charms-tests.yaml
@@ -39,7 +39,6 @@ jobs:
           - charm-repo: canonical/mongodb-operator
           - charm-repo: canonical/mysql-router-k8s-operator
           - charm-repo: canonical/namecheap-lego-k8s-operator
-            run-static: true
           - charm-repo: canonical/nginx-ingress-integrator-operator
           - charm-repo: canonical/oathkeeper-operator
           - charm-repo: canonical/oauth2-proxy-k8s-operator
@@ -48,12 +47,10 @@ jobs:
           - charm-repo: canonical/pgbouncer-k8s-operator
           - charm-repo: canonical/ranger-k8s-operator
           - charm-repo: canonical/route53-lego-k8s-operator
-            run-static: true
           - charm-repo: canonical/s3-integrator
           - charm-repo: canonical/saml-integrator-operator
           - charm-repo: canonical/seldon-core-operator
           - charm-repo: canonical/self-signed-certificates-operator
-            run-static: true
           - charm-repo: canonical/smtp-integrator-operator
           - charm-repo: canonical/superset-k8s-operator
           - charm-repo: canonical/temporal-admin-k8s-operator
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
> not if there was a failure in an earlier step (i.e. the unit test step was skipped)

I'm not an expert with GitHub workflows, so this might be a dumb question, but is this just future-proofing? There's no way that the unit tests can be skipped as it currently exists, right?

I'm also sure sure if this makes sense - typically you'd do linting then static then unit tests - why skip the static tests if the unit tests are skipped? Wouldn't the opposite make more sense in general? But here, don't we want to know about both types of issue?

> and not if the job was cancelled.

Same caveat here, but if the job is cancelled, doesn't the runner take care of not doing anything more?

> There's just one major problem -- the marking has been done manually in this PR by adding `run-static: true` to four charms, which will be overwritten when we next run `update-published-charms-tests-workflow.py`. What should we do here? The simplest approach might be to add the manually marked charms to a list in the python script, have the script generate the `run-static` entries, and extend the regex used to substitute the repo lines to handle `run-static` lines.

I'd rather have the opposite approach to this - something like "if there seems to be a 'static' command, run it". That can be something like `tox list --no-desc | grep '^static$'`, no?

james-garner-canonical:
> is this just future-proofing? There's no way that the unit tests can be skipped as it currently exists, right?

I believe that if there's an error running one of the previous steps, the subsequent steps will be skipped. So the idea is that if any of the setup steps fail, we won't do any unit testing or linting (because unit testing is skipped, and linting checks if it was skipped). Otherwise we'll do both -- unit testing failing won't stop the linting.

This seems better to me than setting `continue-on-error` on unit testing, as that results in the step always being marked as a success.

> if the job is cancelled, doesn't the runner take care of not doing anything more

In general yes. However, if we want to avoid the `continue-on-error` solution, one way would be to mark linting as `always()` running -- but this will run even if the workflow is cancelled, so we use `!cancelled()` instead.

>  typically you'd do linting then static then unit tests

The reason I decided to have the lint step detect if the static step is skipped, rather than the reverse, is because we'll skip the static step in many cases when there's no static environment, whereas we're always running the unit tests.

> I'd rather have the opposite approach to this - something like "if there seems to be a 'static' command, run it". That can be something like tox list --no-desc | grep '^static$', no?

I think we'd also want to detect what the `static` environment is doing -- currently most of them are running `bandit`, which we presumably don't care about. I guess we could just run `static` anyway and try to push charmers to change their environments. Or we could try inspecting what the static environment does programmaticly, but that seems painful. Manually marking the charms that have static environments that run type checking was an attempt to get us just the tests we care about at the moment.

Lmk if you want to sync to talk about this. It would be nice if we didn't have to manually flag charms as having static environments that are doing type checking. But at the same time, it doesn't seem too bad to me to start with a manual list and drop it in future.

james-garner-canonical:
Based on discussion with @tonyandrewmeyer, I've changed the workflow to always run the static environment if it exists -- sample run [here](https://github.com/james-garner-canonical/operator/actions/runs/13513690700).

.github/workflows/published-charms-tests.yaml
```diff
@@ -85,11 +85,29 @@ jobs:
           fi
 
       - name: Install dependencies
+        id: deps
         run: pip install tox~=4.2
 
+        # the following test steps will only run if the last setup step (deps) succeeded
+
       - name: Run the charm's unit tests
+        if: ${{ !cancelled() && steps.deps.outcome == 'success' }}
```
dimaqq, line 94: Is this the default behaviour?
- when one step fails, the rest are not run
- if the job is cancelled, the rest is not run

james-garner-canonical: Yes, this `if` is unnecessary for the unit tests step, but I figured it might be nicer to use the same condition for all the post-deps steps in case they get re-ordered at some point

james-garner-canonical:
Ready for review. A test run for the current changes can be seen [here](https://github.com/james-garner-canonical/operator/actions/runs/13663448044).

.github/workflows/published-charms-tests.yaml
```diff
@@ -95,11 +95,30 @@ jobs:
           fi
 
       - name: Install dependencies
+        id: deps
         run: pip install tox~=4.2
 
+        # the following test steps will only run if the last setup step (deps) succeeded
```
tonyandrewmeyer, line 101: Nit:
```suggestion
        # The following test steps will only run if the last setup step (deps) succeeded.
```

benhoyt (approved):
Looks good, thanks. Acting as second reviewer so we can get this merged and then test on https://github.com/canonical/operator/pull/1606

PR #1586, opened 2025-02-24
feat: efficient implementation of RelationDataContent.update

james-garner-canonical:
This PR makes it efficient to add/update/delete multiple keys and values in your relation data, by overriding the default implementation of `RelationDataContent.update` (provided by `MutableMapping`) with an efficient implementation that calls the `relation-set` hook tool only once. The changes to internals that back the new `update` implementation are exercised by existing tests interacting with `RelationDataContent.update` and `RelationDataContent.__setitem__`.

Both Harness and Scenario's model backend implementations require updates to accommodate this. The Harness backend is exercised in the existing Ops tests. The Scenario updates are exercised in new tests added to `testing/tests/test_e2e/test_relations.py`.

---

Original description, now outdated, but retained to explain the original motivation for this work:

This PR resolves [#817](https://github.com/canonical/operator/issues/817) by adding `RelationData.__setitem__(self, key, value)`. This allows users to write `my_relation[app_or_unit] = {'key-value': 'pairs'}`, instead of having to fetch the `RelationDataContent` object with `my_relation[app_or_unit]` and then update its keys and values directly.

The implementation of `RelationData.__setitem__` gets the current `RelationDataContent` for the `key`, calls `RelationDataContent.clear()` (provided by `RelationDataContent`'s inheritance from `MutableMapping` -- the inherited implementation will call `RelationDataContent`'s custom `__delitem__`), and then calls the existing `RelationDataContent.update(value)` method. This ensures that validation and updating the backend occurs exactly as it would have if the user interacted with `RelationDataContent` directly. `RelationData.__setitem__` will raise a `KeyError` if the `key` is not actually a known `Application` or `Unit`.


ops/model.py
```diff
@@ -1790,6 +1790,14 @@ def __iter__(self):
     def __getitem__(self, key: Union['Unit', 'Application']) -> 'RelationDataContent':
         return self._data[key]
 
+    def __setitem__(self, key: Union['Unit', 'Application'], value: Mapping[str, str]) -> None:
+        try:
+            content = self._data[key]
+        except KeyError:
+            raise KeyError(f'Key "{key}" is not a known unit or app.') from None
```
benhoyt, line 1797: The message/str of a standard KeyError is `'a'` (including quotes). How about we kinda match that with the following:

```python
raise KeyError(f'{key!r} is not a known unit or app')
```

ops/model.py
```diff
@@ -1790,6 +1790,14 @@ def __iter__(self):
...
+        content.clear()
+        content.update(value)
```
benhoyt, line 1799: Will this call `__setitem__` and hence the `backend.update_relation_data` hook tool for every existing item, and then again for every new item? I wonder if we can do that a bit more efficiency by only updating the new or changed items, and deleting the ones that are going away?

Might want to expand our unit test for the different cases too.

test/test_model.py
```diff
@@ -270,6 +270,22 @@ def test_update_app_relation_data(
...
+        original_data = {'hello': 'world'}
+        new_data = {'goodnight': 'moon'}
+        relation_id = harness.add_relation(relation_name, 'remote', app_data=original_data)
+        relation = harness.model.get_relation(relation_name)
+        assert relation is not None
+        assert relation.data[relation.app] == original_data
+        # new relation data can be set via RelationData.__setitem__
+        relation.data[relation.app] = new_data
+        assert relation.data[relation.app] == new_data
+        assert harness.get_relation_data(relation_id, relation.app) == new_data
```
benhoyt, line 284: Do we also want to assert that the `RelationDataContent` object (id) hasn't changed?

ops/model.py
james-garner-canonical: Yes it will. I was digging into the internals a bit to see if there was a more efficient way to do things, but I guess `backend.update_relation_data` is ultimately the performance bottleneck since it results in a separate process -- and now I see what you mean. Instead of `clear` we can just delete the keys that aren't in the new mapping, which could already be a big win on its own. Secondly, we could update only the key-value pairs that need updating -- if the values are always strings or simple enough data types then that should be safe just with testing for equality.

tonyandrewmeyer: Wouldn't it be better to add a new method on `RelationDataContent` that would do all the updates in one `relation-set` call? Via a file if you're concerned about the length of the command.

james-garner-canonical: Oh yeah that would be ideal, I didn't know `relation-set` could do that, thanks!

tonyandrewmeyer: [Docs](https://canonical-juju.readthedocs-hosted.com/en/3.6/user/reference/hook-commands/list-of-hook-commands/relation-set/) if you need them :)

IronCore864 (approved):

james-garner-canonical:
I've broken the tests. I'm going to iterate on the efficient approach to unify the single/multiple update code paths as much as possible, and then I'll update the test side of things (may need some little modifications to harness/scenario).

testing/src/scenario/mocking.py
```diff
@@ -390,7 +392,8 @@ def relation_set(self, relation_id: int, key: str, value: str, is_app: bool):
             tgt = relation.local_app_data
         else:
             tgt = relation.local_unit_data
-        tgt[key] = value
+        for key, value in data.items():
+            tgt[key] = value
```
tonyandrewmeyer, line 396: Does Scenario currently have a bug in that the mocked `relation_set` doesn't do the "empty value means remove" behaviour?

ops/model.py
```diff
@@ -1936,11 +1954,29 @@ def update(self, other: typing.Any = (), /, **kwargs: str):
         super().update(other, **kwargs)
 
     def __delitem__(self, key: str):
-        self._validate_write(key, '')
```
tonyandrewmeyer, line 1939: Should this not be replaced with validate write access?

ops/model.py
```diff
@@ -1790,6 +1790,13 @@ def __iter__(self):
     def __getitem__(self, key: Union['Unit', 'Application']) -> 'RelationDataContent':
         return self._data[key]
 
+    def __setitem__(self, key: Union['Unit', 'Application'], value: Mapping[str, str]) -> None:
+        try:
+            content = self._data[key]
+        except KeyError:
+            raise KeyError(f'{key!r} is not a known unit or app') from None
+        content._replace(value)
```
tonyandrewmeyer, line 1798: A nit, but maybe `_replace` is not the right name given that there's an `_update` that works only on the local cache? Or maybe `_update` should be `_update_cache`? Feel free to ignore this -- I don't feel strongly about it, it just seems a bit off (but probably more the original name than your one).

test/test_model.py
```diff
@@ -270,6 +271,26 @@ def test_update_app_relation_data(
             rel.data[harness.model.app].update(*args, **kwargs)
             assert harness.get_relation_data(relation_id, harness.model.app) == {'foo': 'baz'}
 
+    def test_set_app_relation_data(self, harness: ops.testing.Harness[ops.CharmBase]):
+        relation_name = 'db1'
+        original_data = {'deleted': 'item', 'unchanged': 'item', 'changed': 'original-value'}
+        new_data = {'new': 'item', 'unchanged': 'item', 'changed': 'new-value'}
```
tonyandrewmeyer, line 277: Nit: probably nice to give deleted and new unique values, in case of some weird bug where values get put against the wrong keys.

test/test_model.py
```diff
@@ -270,6 +271,26 @@ def test_update_app_relation_data(
...
+        # TODO: break this down into individual test cases? parametrize test?
```
tonyandrewmeyer, line 278: I'd say individual cases probably, unless it would parametrize easily, but it doesn't seem so at a glance.

ops/model.py
```diff
@@ -1936,11 +1954,29 @@ def update(self, other: typing.Any = (), /, **kwargs: str):
         super().update(other, **kwargs)
 
     def __delitem__(self, key: str):
-        self._validate_write(key, '')
         # Match the behavior of Juju, which is that setting the value to an empty
         # string will remove the key entirely from the relation data.
         self.__setitem__(key, '')
 
+    def _replace(self, data: Mapping[str, str]) -> None:
```
tonyandrewmeyer, line 1961: Both the docstring and the code here are very readable, nice work!

tonyandrewmeyer (commented):
I think this is definitely the right approach. It's looking really great!

On testing:
 * You should definitely add a testing/tests test, probably in test_relations. Basically the same kind of thing as you have for harness/test_testing, but for Scenario. I'd also include one that checks that removing a key works (which I think will fail currently!).
 * I don't think there's a test for the "I gave setitem an empty dict" case raising an error. It would be good to have one if it's not too complicated.
 * I don't think there's any relation-set checking in test_main at the moment, but it feels like this is maybe a good opportunity to add some, which could include this. Basically just verifying that relation-get and relation-set are called properly, including in this new flow. At the moment there's quite an assumption that it works ok since the tests (like much of test_model!) rely on Harness.

ops/model.py
james-garner-canonical: Good call, how about `_replace_data` and `_update_cache`?

ops/model.py
james-garner-canonical: Because `__delitem__` just calls `__setitem__`, and `__setitem__` first calls `_validate_write`, this is just removing what seems to be a completely  redundant check.

testing/src/scenario/mocking.py
james-garner-canonical: Definitely looks like it. Good catch! Thanks.

ops/model.py
```diff
@@ -1931,12 +1942,27 @@ def __getitem__(self, key: str) -> str:
...
+        for k, v in dict(data).items():
+            self[k] = v
+        for k, v in kwargs.items():
+            self[k] = v
+        """
+        data = dict(data)
+        data.update(kwargs)
+        changes = {k: v for k, v in data.items() if k not in self or v != self[k]}
+        if not changes:  # no changes or deletions
+            return  # content is already exactly as requested
```
james-garner-canonical, line 1960: Should we filter changes and exit early like this? If so, should we also filter out deletion of keys that already don't exist?
```py
changes = {
    k: v for k, v in data.items()
    if (k not in self and v != '') or (k in self and v != self[k])
}
```

testing/src/scenario/mocking.py
james-garner-canonical: Reposting from matrix:

Curiously, we seem to get the correct result for deletion via setting the value to an empty string using the scenario mock backend's `relation_set` without special casing empty string therein -- if we check at the end of the mocked `relation_set`, the dict is full of empty string values. When we actually retrieve the relation data via `get_relation` after `Context.run` in the new tests below, the keys with empty string values have been deleted.

So that's nice, I guess, though I'm completely baffled as to where it's happening

ops/model.py
james-garner-canonical: Or the slightly spicier:
```py
changes = {k: v for k, v in data.items() if (v != self[k] if k in self else v !='')}```

Comment (tonyandrewmeyer) [reply ops/model.py:1960]: I think the spicier one would fall under this [style guide decision](https://github.com/canonical/operator/blob/main/STYLE.md#avoid-nested-comprehensions-and-generator-expressions), even though it's not exactly what's described there.

But exiting early is good I think - although maybe after the validation, so that you consistently get an error if you don't have access, even if you're setting to the same thing?

And filtering out the 'change' of deleting something that isn't there is probably good too.

File: testing/tests/test_e2e/test_relations.py [hunk 11]
```diff
@@ -88,6 +90,124 @@ def pre_event(charm: CharmBase):
     )
 
 
+@pytest.mark.parametrize(
+    ("original_data", "new_data", "result_data"),
+    [
+        (  # populate the relation data from scratch
```
tonyandrewmeyer, line 96: What about using these comments as the ID for the test? `test_relation_set[populate the relation data from scratch]` is a lot clearer than whatever auto ID this is going to end up as.

testing/tests/test_e2e/test_relations.py
```diff
@@ -88,6 +90,124 @@ def pre_event(charm: CharmBase):
...
+
+    ctx = Context(
+        Charm,
+        meta={
+            "name": "charm-name",
+            "peers": {relation_name: {"interface": "interface-name"}},
+        },
+    )
+    rel_in = PeerRelation(endpoint=relation_name, local_unit_data=original_data)
+    state = ctx.run(ctx.on.update_status(), State(relations=[rel_in]))
```
tonyandrewmeyer, line 206: Nit:
```suggestion
    state = ctx.run(ctx.on.update_status(), State(relations={rel_in}))
```

testing/tests/test_e2e/test_relations.py
```diff
@@ -88,6 +90,124 @@ def pre_event(charm: CharmBase):
...
+    rel_out = cast(PeerRelation, state.get_relation(rel_in.id))
```
tonyandrewmeyer, line 207: Do you need the cast? `RelationBase` has `.local_unit_data` so it seems like it shouldn't be needed. Is this maybe a hangover from when I wrongly pointed you towards `.peers_data`?

testing/tests/test_e2e/test_relations.py
```diff
@@ -88,6 +90,124 @@ def pre_event(charm: CharmBase):
...
+            {"NEW-KEY-1": "NEW-VAL-1", "old-key-2": "NEW-VAL-2", "old-key-3": ""},
+            {
+                "old-key-1": "old-val-1",
+                "NEW-KEY-1": "NEW-VAL-1",
+                "old-key-2": "NEW-VAL-2",
+            },
+        ),
+    ],
+)
+def test_relation_set(
```
tonyandrewmeyer, line 182: I know it's not functionality you're adding in this PR, but could you add another `relation_set` test for the non-update way of setting/deleting relation data? Like:

```python
def _on_update_status(self, _: ops.UpdateStatusEvent):
    rel = self.model.get_relation(relation_name)
    assert rel is not None
    data = rel.data[self.unit]
    data['new'] = 'new'
    data['changed'] = 'changed'
    del data['to-remove']
```

Maybe parametrized but probably all three could be tested in a single test.

I would probably call that `test_relation_set_single` and this `test_relation_set_update` or something.

testing/tests/test_e2e/test_relations.py
james-garner-canonical: Good catch, that's exactly why the unnecessary `cast` is there

testing/tests/test_e2e/test_relations.py
tonyandrewmeyer: Actually, I see there's a `test_relation_set` in test_state, which is doing this although in a much older style. So it's probably not necessary.

james-garner-canonical: Ah, I already wrote it -- keep or throw it away?

testing/src/scenario/mocking.py
tonyandrewmeyer: `_GenericLazyMapping`'s `_data` property calls `_load`. In production, that calls out to `relation-get` and so you get a new dictionary every time. In Scenario, `relation_get` returns the dictionary from the `RelationBase` object, and it's the same dictionary object every time, not a copy. That means that the local cache and the object on the state are the same, so it's actually updated twice.

That means you can do:

```diff
diff --git a/testing/src/scenario/mocking.py b/testing/src/scenario/mocking.py
index 6819f3f..77b8432 100644
--- a/testing/src/scenario/mocking.py
+++ b/testing/src/scenario/mocking.py
@@ -392,8 +392,6 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
             tgt = relation.local_app_data
         else:
             tgt = relation.local_unit_data
-        for key, value in data.items():
-            tgt[key] = value
 
     def secret_add(
         self,
```

And your tests still pass (although test_state's `test_relation_set` does not).

This has a bit of an icky feel, but changing it is probably too much for this PR. Although, while it works when `update` is used, I'm not sure it works when using the `del data['key']` style (and it doesn't look like the test_state test checks that?). Feel free to punt all that to another issue.

testing/tests/test_e2e/test_relations.py
tonyandrewmeyer: I think keep it, it's nicer than the old style one in test_state, and it feels like one belongs in test_relations anyway.

tonyandrewmeyer:
The docs fail because it looks like your code is a "block quote". You can fix it like:

```diff
diff --git a/ops/model.py b/ops/model.py
index 61ed63d..2e564af 100644
--- a/ops/model.py
+++ b/ops/model.py
@@ -1947,11 +1947,12 @@ class RelationDataContent(LazyMapping, MutableMapping[str, str]):
     ) -> None:
         """Efficiently write multiple keys and values to the databag.
 
-        Has the same ultimate result as this, but uses a single relation-set call:
-        for k, v in dict(data).items():
-            self[k] = v
-        for k, v in kwargs.items():
-            self[k] = v
+        Has the same ultimate result as this, but uses a single relation-set call::
+
+            for k, v in dict(data).items():
+                self[k] = v
+            for k, v in kwargs.items():
+                self[k] = v
         """
         data = dict(data)
         data.update(kwargs)
```

Although I think maybe it's only the extra `:` that's strictly needed.

The PR title needs updating, as you did with the description.

tonyandrewmeyer:
We probably need a comment on #817 that explains why `__setitem__` isn't being done after all. I think this will still auto-close that issue, and it probably should instead be closed as not-completed.

testing/tests/test_e2e/test_relations.py
james-garner-canonical: To complicate things a little bit, it fails with the current scenario code (where update passed) -- it results in the empty string values staying in the dict.

testing/src/scenario/mocking.py
james-garner-canonical: Mentioned in the other thread, but this might be the best place to discuss it -- the single version fails how we initially expected the update version to fail with current scenario code.

testing/src/scenario/mocking.py
```diff
@@ -379,7 +379,9 @@ def status_set(
     def juju_log(self, level: str, message: str):
         self._context.juju_log.append(JujuLogLine(level, message))
 
-    def relation_set(self, relation_id: int, key: str, value: str, is_app: bool):
+    def relation_set(
+        self, relation_id: int, data: Mapping[str, str], is_app: bool
+    ) -> None:
```
tonyandrewmeyer, line 384: I wonder if our/charmers lives would be easier if we made this backwards compatible. The `relation_set` call in `model.py` already passes `data` as a keyword arg, so this could be:

```suggestion
    # TODO: Remove the backwards compatibility `key` and `value` arguments and make
    # make the signature the same as in model.py once we no longer support old ops-scenario.
    def relation_set(
        self, relation_id: int, key: str, value: str, is_app: bool, data: Mapping[str, str]=None,
    ) -> None:
        if data is None:
            data = {key: value}
```

I think this would give an easier upgrade path, with the next release of `ops-scenario` supporting both old and new `ops`, and then after that transition version we can bump the required version of `ops-scenario` for `ops[testing]` to be newer and drop this.

Maybe there's a way to put in the right versions in the `ops[testing]` and `ops-scenario` dependencies that avoids this, but I suspect that may be complicated.

testing/src/scenario/mocking.py
tonyandrewmeyer: Yeah, so I believe it *is* a Scenario bug. We can fix it in another PR, or you could drive-by fix it here.

testing/src/scenario/mocking.py
james-garner-canonical: We could do this, yeah. We'd need `key` and `value` to be optional though, and I guess `is_app` as well.

testing/src/scenario/mocking.py
james-garner-canonical: Isn't it weird that it's a bug with assignment but not with `update` though? If it's that we're calling `RelationDataContent.update`, which called `_update_cache`, which handles popping empty string values, wouldn't assignment via `RelationDataContent.__setitem__` do this too?

testing/src/scenario/mocking.py
tonyandrewmeyer: Ah indeed, sorry, was just quickly writing it out and didn't think through all the specifics. I don't really like doing this, but I feel like it's probably going to be less pain to have it for a version or two.

james-garner-canonical: Tbh I'm kind of -1 on this. Wouldn't it be better to have `ops[testing]` require the latest `ops-scenario`, and have `ops-scenario` require the latest `ops`? Whenever both are being released, this means both of them have the dependency version in the `pyproject.toml` to match the release version of the other. I guess currently charmers can sometimes get away with updating one without the other, but I don't think we should try and guarantee that will work, neither in general, nor for a defined period like 2 months. If you're bumping one, you can (and should) bump the other. But won't most charmers just bump `ops` and use the version of `ops-scenario` that comes with it?

tonyandrewmeyer: Well, I'm pretty sure there are still charms that use `ops-scenario` specifically for tests, rather than `ops[testing]`, and so you can easily in that situation have `ops==2.20.0` and `ops-scenario==7.2.0` and that'll break running the tests but pip won't complain.

We do want people to be using `ops[testing]` and we do want people to always be on the latest version of both `ops` and `ops-scenario`, so it's not terrible if we do this. I don't think we need to guarantee that all ops 2.x will work with ops-scenario 7.x, but I'm not super keen on them being in lockstep so that you always bump both, either.

Anyway, your call here - if you'd rather bump the versions, then that's fine (we'll need to make sure we track that somewhere so it doesn't get forgotten).

test/test_model.py
```diff
@@ -513,7 +531,6 @@ def test_relation_data_del_missing_key(self, harness: ops.testing.Harness[ops.Ch
                 ('relation_ids', 'db1'),
                 ('relation_list', relation_id),
                 ('relation_get', relation_id, 'myapp/0', False),
-                ('update_relation_data', relation_id, harness.model.unit, 'port', ''),
```
james-garner-canonical, line 516: Deleting an already missing key no longer hits the backend

testing/src/scenario/mocking.py
james-garner-canonical: We've switched to releasing ops and ops-scenario in lockstep with exact version requirements (thanks, Tony!), so backwards compatibility code is no longer necessary here.

testing/src/scenario/mocking.py
tonyandrewmeyer: Since `__setitem__` (now?) does call `update()`, the Scenario code definitely does nothing. The tests all pass (and manually checking things are fine) with:

```diff
diff --git a/testing/src/scenario/mocking.py b/testing/src/scenario/mocking.py
index 6819f3f..27b8630 100644
--- a/testing/src/scenario/mocking.py
+++ b/testing/src/scenario/mocking.py
@@ -383,6 +383,7 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
         self, relation_id: int, data: Mapping[str, str], is_app: bool
     ) -> None:
         self._check_app_data_access(is_app)
+        return
         relation = self._get_relation_by_id(relation_id)
         if is_app:
             if not self._state.leader:
```

However, I'm fine with just leaving this as-is for this PR. It's great that we have proper tests for it so that we can do better things in the future.

tonyandrewmeyer (approved):
Looks really great - thanks for making those changes!

It's a shame that we can't really do the originally requested feature, but I think this is a nice performance improvement, and it generally makes the code nicer since it follows Juju more closely.

testing/src/scenario/mocking.py
james-garner-canonical: I've added a comment for now. My instinct here is that scenario should either model the cache vs databag separation properly by giving `RelationDataContent` a copy of the backend dict on request and only updating it here, or perhaps it should drop modeling the backend bit to make it less dependent on ops internals ... not sure which.

PR #1589, opened 2025-02-26
feat: add the ability to emit custom events in unit tests

### LINKED ISSUE #1421: Restore support for running custom events in Scenario
In Scenario 6, it's possible to run custom events:

```python
ctx = Context(...)
state_in = scenario.State()
state_out = ctx.run("my_charm_lib.on.foo", state_in)
assert state_out...
```

In Scenario 7.0, this has been removed: firstly because we got rid of the 'stringly-typed' events, and, more significantly, because we felt that the level of abstraction for Scenario is at the "Juju event". The intention was that people would run the Juju event that would cause the custom event to be emitted, e.g. `ctx.on.relation_changed()` might end up causing a `DatabaseCreated` event to be emitted.

@gruyaume [pushed back on this](https://matrix.to/#/!xdClnUGkurzjxqiQcN:ubuntu.com/$nFBZ-L75xwUYMM1fMgR_x8au14g7kr2c-38pARyIg2g?via=ubuntu.com&via=matrix.org&via=matrix.debian.social):

> our charm unit tests shouldn't know about library specificities. Our charm does not look into the relation databags, it only interfaces with the library. Taking the approach you are proposing entails our charm unit test to replicate the behavior of libraries that we don't have ownership over.

I feel like the charm author probably does need to know which Juju events will cause the custom events to be able to manually use the charm and to be able to write integration tests, and that you shouldn't need to manage the relation databags.

However, there are potentially many events and you have to know which ones might be triggered for what's happening. For example, with the data_platform `DatabaseRequires` class from data-platform-libs, it probably needs something like:

```python
def test_credentials():
    ctx = Context(DemoAPICharm)
    db_rel = Relation("database")
    container = Container("webapp", can_connect=True)
    state = State(relations={db_rel})
    state = ctx.run(ctx.on.relation_created(), state)
    state = ctx.run(ctx.on.relation_joined(), state)
    state = ctx.run(ctx.on.relation_changed(), state)
    state = ctx.run(ctx.on.secret_changed(), state)

    assert state.get_container("webapp")... # check that the config file is there or whatever
```

Maybe tests that rely on custom events should be [Catan](https://github.com/pietroPasotti/catan) tests? I think what I want to be testing is that when I integrate the two charms the state of the charm I'm writing ends up the way I expect.

However, maybe we should add back some way to run custom events explicitly. @gruyaume also needed the ability to provide custom arguments for creating the event, which Scenario 6 can't do.

Maybe this looks like this?

```python
ctx = Context(...)
state_in = State(...)
state_out = ctx.run(ctx.on.custom_event('database.on.database_ready', x, y=42), state_in)
assert state_out...
```

`custom_event` would be a static method on `CharmEvents` with a signature like `(attribute: str, *args, **kwargs)` and it would pass `*args` and `**kwargs` through to the custom event class.

Moved from https://github.com/canonical/ops-scenario/issues/177

tonyandrewmeyer:
In Scenario 7.0, the ability to emit custom events was removed, because we felt that the proper way to test these was to emit the Juju event that would trigger them (e.g. relation-changed). However, custom events are essentially the API for (relation-style) charm libraries, and charms should not need to know how to populate the relation data to generate them - only that they will get a specific event, with specific attributes.

Scenario 6 and earlier could emit custom events, but couldn't provide any arguments/snapshot for the event, and the user had to provide a string that identified an attribute on the charm, which doesn't match with the non-stringly-typed events in Scenario 7. (This functionality is still present, but requires creating `_Event` objects manually, which are now private).

Several API choices were considered for emitting custom events, and this PR differs from the last discussion that was held. The main challenge is that the ops framework essentially writes a notice of an event (and the snapshot for that event) into the unit state storage, and then retrieves notices (and snapshots) and creates the actual event object to pass to the observing method. This exists to support deferring, but is used even if the event is not deferred. This requires a `Handle`, and in particular a path for the `Handle`. In practice, the path is not determined until the charm object is instantiated (specifically until the library's `Object` is instantiated), and that is before the `ctx.run` call, so it is difficult to get matching `Handle` paths, particularly with only the event object class, which was the planned API. In addition, the stored observers don't know what type of object will be emitted, only the event source, which makes something like `ctx.on.custom(CustomEvent)` difficult to match against the observer list.

The API added in this PR assumes a charm and library structure like:

```python
# lib/charms/my_lib_charm/v0/lib.py

class MyEvent(ops.EventBase):
    def __init__(self, handle: ops.Handle, arg0: str = "", arg1: int = 0):
        super().__init__(handle)
        self.arg0 = arg0
        self.arg1 = arg1

    def snapshot(self):
        base = super().snapshot()
        base.update({"arg0": self.arg0, "arg1": self.arg1})
        return base

    def restore(self, snapshot: dict[str, typing.Any]):
        super().restore(snapshot)
        self.arg0 = snapshot["arg0"]
        self.arg1 = snapshot["arg1"]

class MyEvents(ops.ObjectEvents):
    foo_changed = ops.EventSource(MyEvent)

class MyConsumer(ops.Object):
    on = MyEvents()

    def __init__(self, charm: ops.CharmBase, relation_name: str):
        super().__init__(charm, relation_name)

# src/charm.py

from my_lib_charm.v0.lib import MyConsumer, MyEvent

class MyCharm(ops.CharmBase):
        super().__init__(framework)
        self.consumer = MyConsumer(self, "name-of-a-relation")
        framework.observe(self.consumer.on.foo_changed, self._on_event)

    def _on_event(self, event: MyEvent):
        ...
```

Most significantly, the charm must have an attribute that is an `Object` object, which itself has an attribute that is an `ObjectEvents` object, and that must be the same type as the one passed to `ctx.run`. I believe this should cover the standard patterns that libraries use, and if we need to support others in the future we can extend the API (making it more complex, but able to accommodate more variation).

This allows emitting a custom event like:

```python
state_in = testing.State()
state_out = ctx.run(ctx.on.custom(MyConsumer.on.foo_changed, "foo", arg1=42), state_in)
```

The charm author already needs to know about `MyConsumer.on.foo_changed`, because they're creating a `MyConsumer` instance in the charm, and using `on.foo_changed` for the observing. They also need to know the type of event that they'll get in their observer method, which implies knowing the arguments/keyword arguments that are used to create an object of that type. (Note that the Handle doesn't need to be passed - the framework takes care of that).

(I went through a lot of alternatives to the above - I'm happy to discuss those or outline some if anyone is interested. I also considered asking libraries to provide support for testing, but I'd prefer to avoid that so that use of existing libraries can be tested without those needing updates, and having to get buy-in for those.).

The implementation looks at the charm object's attributes for an `Object`, and then looks for an `ObjectEvents` attribute of that, trying to find one that matches the one provided in the `custom()` call, and then basically swaps out the unbound event that (e.g.) `MyConsumer.on.foo_changed` has (missing the framework and handle) for the bound one that `MyConsumer(...).on.foo_changed` provides. If no match is found, it falls through to the regular search and will error out saying that the source can't be found.

This is not dissimilar to the Scenario 6 approach, which used the (string) event path to navigate through attributes of the charm, but rather than have to provide the "attribute path" as a string, the charm author provides the object that is used (essentially) in the observe call, and Scenario searches for one that matches.

For the event arguments (other than the `Handle`), the implementation simply bypasses the normal code that figures those out for regular events. This seems uncontroversial - there isn't really any other way to get these, unless we had the user provide a snapshot dictionary instead, which seems less user-friendly (it's more knowing the implementation of the event than simply its signature).

Some libraries have custom events that take ops objects (particularly `ops.Relation`) as an argument. The test code cannot pass these to the `custom()` call, because the framework is needed to get those objects - the test code should be working with the Scenario options (such as `testing.Relation`) instead. To facilitate this, Scenario transforms Scenario objects in the args/kwargs into the (mocked) ops equivalents, when the framework is available.

There are two "on"s in `ctx.on.custom(MyConsumer.on.foo_changed)` (assuming that the `ObjectEvents` is called "on", as is usual), which I don't love, but I prefer this to `ctx.custom(MyConsumer.on.foo_changed)` or similar.

Some libraries have custom events that inherit from a `HookEvent` subclass (particularly `RelationEvent`). We advise against this, and do not support it with the feature in this PR. Most of those events also need the relation data to be populated as well, so the charm test already needs to populate a `scenario.Relation` and should trigger the custom event by running the originating Juju event (such as `relation_changed`).

Note that if you are testing a *library* then you should not be emitting custom events in the tests, you should always emit the triggering Juju ones (having set up the relation data appropriately), and verify that the custom events are *emitted* correctly.

Drive-bys:
 * Fixed an outdated docstring.
 * Removed `_Event.args` and `_Event.kwargs`, which haven't been used for a while.
 * A bunch of tests in `test_context_on` had the wrong name for an object - mostly this didn't break anything, but it made the tests confusing, and there was one case where it was a problem.
 * To simplify the new tests, added future annotations and switch to "import ops" for the deferred tests.

Fixes #1421


tonyandrewmeyer:
@PietroPasotti if you have some time, I'd appreciate your feedback here too, thanks!

dimaqq:
High-level comment: given how extensive the PR description needs to be, please consider copying chunks of it into the appropriate sections in the docs. I would expect most description paragraphs to appear in the PR body somewhere.

The docstring is a good start, but I guess we need add something to the how-to, reference and and possibly the testing tutorial?

testing/tests/test_context_on.py
```diff
@@ -115,9 +117,9 @@ def test_simple_secret_events(as_kwarg, event_name, event_kind, owner):
         mgr.run()
         assert len(mgr.charm.observed) == 2
         assert isinstance(mgr.charm.observed[1], ops.CollectStatusEvent)
-        mgr = mgr.charm.observed[0]
-        assert isinstance(mgr, event_kind)
-        assert mgr.secret.id == secret.id
+        evt = mgr.charm.observed[0]
+        assert isinstance(evt, event_kind)
+        assert evt.secret.id == secret.id
```
dimaqq, line 122: Cheers!
I think I went a tad further here: https://github.com/canonical/operator/pull/1527/files#diff-104ae483d187a2a70cb0600c23ab87c65b3604059e64c0b695c51c77c128af06 though an improvement is still an improvement.

testing/src/scenario/context.py
```diff
@@ -365,6 +365,79 @@ def action(
...
+        The arguments for the custom event are the arguments for the original
+        Juju event, combined with any that are passed here.
+        """
+        event_kwargs = {}
+        if from_juju_event:
+            name = from_juju_event.name
+            if from_juju_event.storage is not None:
+                event_kwargs["storage"] = from_juju_event.storage
+            if from_juju_event.relation is not None:
+                event_kwargs["relation"] = from_juju_event.relation
```
dimaqq, line 410: It would be weird to have both set, wouldn't it?

dimaqq (approved):
This is a solid piece of work!

testing/src/scenario/context.py
tonyandrewmeyer: Yeah, most of these are actually groups, where it makes no sense to get both of them. It felt simpler in this code to just copy everything that was set. I could instead add an `_Event.merge` or something like that, which would look at the type of event and only copy the relevant ones, or raise an exception if it was inconsistent. Or `_Event` could itself have a `__post_init__` that verified consistency? Or maybe there should be `_Event` subclasses for the different types rather than the catch-all it currently is?

Not sure if it's worth doing any of these in this PR, wdyt?

testing/src/scenario/_ops_main_mock.py
```diff
@@ -160,6 +169,22 @@ def _make_storage(self, _: _Dispatcher):
...
+                if not isinstance(attr, ops.Object):
+                    continue
+                for sub_attr_name in dir(attr):
+                    sub_attr = getattr(attr, sub_attr_name)
+                    if not isinstance(sub_attr, ops.ObjectEvents):
+                        continue
+                    emitter_type = type(self.event.custom_event_source.emitter)
+                    if emitter_type is not type(sub_attr):
+                        continue
+                    return getattr(sub_attr, self.event.custom_event_source.event_kind)
```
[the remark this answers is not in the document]
tonyandrewmeyer: If you have something like:

```python
# lib/charms/my_lib_charm/v0/lib.py

class MyEvent(ops.EventBase):
    pass

class MyEvents(ops.ObjectEvents):
    foo_changed = ops.EventSource(MyEvent)

class MyConsumer(ops.Object):
    on = MyEvents()

    def __init__(self, charm: ops.CharmBase, relation_name: str):
        super().__init__(charm, relation_name)

# lib/charms/other_lib_charm/v0/lib.py

class OtherEvent(ops.EventBase):
    pass

class OtherEvents(ops.ObjectEvents):
    foo_changed = ops.EventSource(OtherEvent)

class OtherConsumer(ops.Object):
    on = OtherEvents()

    def __init__(self, charm: ops.CharmBase, relation_name: str):
        super().__init__(charm, relation_name)

# src/charm.py

from my_lib_charm.v0.lib import MyConsumer, MyEvent
from other_lib_charm.v0.lib import OtherConsumer, OtherEvent

class MyCharm(ops.CharmBase):
        super().__init__(framework)
        self.consumer1 = MyConsumer(self, "name-of-a-relation")
        self.consumer2 = OtherConsumer(self, "name-of-a-relation")
        framework.observe(self.consumer1.on.foo_changed, self._on_event)
        framework.observe(self.consumer2.on.foo_changed, self._on_event)

    def _on_event(self, event: MyEvent | OtherEvent):
        ...
```

Then it's fine, because you'll do:

```python
ctx.run(ctx.on.custom(OtherConsumer.on.foo_changed), state)
```

And Scenario will be looking for an emitter that matches `OtherConsumer.on.foo_changed` - it's the type of the emitter that matters, not that both are called `foo_changed`. Even if the library is subclassing another emitter, then it should be ok I believe, because it's doing `type(x) is type(y)` rather than `isinstance`.

I'll add a test for this to be sure that it's ok.

The `event_name` that gets passed comes from the `framework.event_name` attribute and that is originally the *Juju* event name. When custom events really run, they skip this part, because the `.emit()` method is called directly on the `BoundEvent`, but in Scenario you're asking for the event to be emitted in the same style as if it was a Juju event.

testing/tests/test_context_on.py
tonyandrewmeyer: Ah, I hadn't noticed that when I was looking over the tracing PR. Those changes are not actually related to the tracing, right? What about you/I cut those out of that PR and put them in a separate one, and then it reduces the size of the tracing one, and I can rebase/merge here and avoid the drive-by in this one too.

benhoyt: Always a fan of reducing big PRs, thanks!

benhoyt:
Per discussion, this is looking good, but we're going to refine the API to use `*args` and `**kwargs`, and rename `from_juju_event` to `from_` (and indicate in the docs that it's only for poorly-designed custom events that do this weird inheritance from HookEvent):

```python
ctx.on.custom(MyConsumer.on.foo_changed, "foo", arg1=42)
ctx.on.custom(DatabaseRequires.on.database_created, from_=ctx.on.relation_changed(relation))
```

tonyandrewmeyer:
Of the 33 libraries with custom events I've looked at, 12 have custom events that inherit from some `HookEvent` subclass (typically `RelationEvent`, or a `RelationEvent` subclass). We strongly imply that you shouldn't do this in the docs - we say in the `EventBase` doc that you *should* subclass it to create a custom event (but don't say explicitly only that one and not one of its subclasses), and the `HookEvent` docs say that it's for events raised by Juju (but don't say explicitly that this should be directly raised, or that it should apply to subclasses). Our how-to guide says that it should be an `EventBase` subclass (but not that it can't be a subclass of an `EventBase` subclass).

A couple of libraries have a base custom event class that automatically does the sensible thing for snapshot & restore. If we offered that as `ops.CustomEvent(ops.EventBase)` and told everyone they must inherit from that, we would at least be much clearer about what we believe you should do.

Over a third doing this feels like a lot - even though 33 libraries is not exhaustive, assuming that the proportion reflects the whole population.

9/12 of the ones that inherit from `Relation` make use of `self.relation` in a method or property of the custom event class. Theoretically, this isn't needed - you could just populate all of the values on init (I don't think any were *changing* relation data, just reading it). However, there could be performance advantages (unlikely to be significant), and security advantages (although the secrets ought to be in Juju secrets, so you could safely store the secret ID and get the content on demand), and if the custom event is deferred then you'll get the newer version of the databag (hard to know if this is an advantage). For these, the charm test will have to populate the relation databag, and so there's not a lot of advantage of running the custom event rather than the original Juju one.

Again, I don't think we do much guiding on this, and we probably should.

There are at least 3 custom events that inherit from `EventBase` specifically but do pass in a `Relation` object. This is also not currently possible. However, I think this could easily be fixed, where if you pass in a `testing.Relation` (maybe also `testing.Container`, `testing.Secret`, and so on) as an argument then Scenario converts it into the ops version of the object (which Scenario can do because at the time of the actual emit it has access to the model, unlike the charm test code, but also to all its own mock infrastructure).

I feel there are three pathways here:

1. Custom events inheriting from `RelationEvent` is common, so we should support testing that.
2. Almost all custom events inheriting from `RelationEvent` are unusable without setting up the relation data, so if the lib goes this way then you should just run the underlying Juju event.
3. We want to drive libs towards a better way, so we only support the "pure" approach where a custom event has no access to anything from whatever Juju event triggered it, unless it's explicitly passed in via `__init__` (no `HookEvent` subclassing, no reading the environment variables to get Juju info, etc). We only support testing this and try to drive the libraries that are currently incompatible to change.

I lean towards (3).

benhoyt:
> I feel there are three pathways here:
>
> 1. Custom events inheriting from RelationEvent is common, so we should support testing that.
> 2. Almost all custom events inheriting from RelationEvent are unusable without setting up the relation data, so if the lib goes this way then you should just run the underlying Juju event.
> 3. We want to drive libs towards a better way, so we only support the "pure" approach where a custom event has no access to anything from whatever Juju event triggered it, unless it's explicitly passed in via __init__ (no HookEvent subclassing, no reading the environment variables to get Juju info, etc). We only support testing this and try to drive the libraries that are currently incompatible to change.
>
> I lean towards (3).

I agree. Though these aren't really three pathways. Though you could argue these aren't three pathways: we "support" 1 via 2, and we document that 3 is the way to go.

If people do try to test using `ctx.on.custom()` with a custom event that's a HookEvent subclass, can we make it fail with a helpful error message?

Having a `CustomEvent` base class is interesting, but I kinda think unless we enforce it (which we can't without breaking users), some will just ignore it and keep inheriting from HookEvent anyway. So I don't think that's worth it.

So the final signature will be `ctx.on.custom(MyConsumer.on.foo_changed, "foo", arg1=42)`, right?

testing/src/scenario/context.py
tonyandrewmeyer: No longer relevant, since this piece of the feature was removed.

testing/tests/test_context_on.py
dimaqq: #1596 for deferred events

dimaqq: #1597 for this file

tonyandrewmeyer:
> High-level comment: given how extensive the PR description needs to be, please consider copying chunks of it into the appropriate sections in the docs. I would expect most description paragraphs to appear in the PR body somewhere.

The first three paragraphs don't seem necessary for the docs: they're just context about what existed before and an explanation for why this differs from what was discussed when previously designing the changes here. I don't think anyone will care much about those going forward (and there is always the description in the merge commit if needed).

The code example and how it works should be covered by the docstring (and the second one has been removed now that the API has been simplified).

Then there's a bunch more about why I think this is the best design choice. I suppose this could all be wrapped into an explanation doc, but I don't think people would generally be interested once the PR is closed.

> The docstring is a good start, but I guess we need add something to the how-to, reference and and possibly the testing tutorial?

Reference is automatically updated from the docstring, so covered. I've updated the manage libraries how-to to include this (as well as some other tidying up).

I'm reluctant to touch the tutorial because (a) there's the project this cycle to move testing to each individual chapter, and I don't want to preempt that work, and (b) because the tutorial uses the data-platform-libs library, and that's one where you can't actually use the testing approach that this PR adds, because you need the relation data supplied in the test, and because the events are `RelationEvent` subclasses. Maybe the COS chapter could have tests like this - those libraries tend to be more friendly to this approach, but I'd still prefer to leave the tutorial test updates for that roadmap item.

testing/tests/test_context_on.py
tonyandrewmeyer: Thanks for that! I'll pull those into this PR once they're merged.

tonyandrewmeyer:
@dwilding would you mind having a look over the changes to the [how-to](https://ops--1589.org.readthedocs.build/en/1589/howto/manage-libraries.html) and the [new reference doc](https://ops--1589.org.readthedocs.build/en/1589/reference/ops-testing.html#ops.testing.CharmEvents.custom)?

I'd like the reference doc to be as good as possible in this PR. For the how-to, it'd be nice if it's better than what we have now, but I think trying to make it a really great how-to (probably actually two separate ones) could be done as a different piece of work.

testing/src/scenario/context.py
```diff
@@ -365,6 +365,50 @@ def action(
...
+    @staticmethod
+    def custom(
+        event: ops.BoundEvent,
+        *args: Any,
+        **kwargs: Any,
+    ):
+        """Event triggered by a charm library.
+
+        For example, for a library that emits a ``HappenedEvent`` event via a
+        ``MyConsumer`` object, and charm code like::
```
dwilding, line 377: I found this slightly difficult to parse at first. How about this (also removing "via"):

```suggestion
        For example, suppose that a library uses a ``MyConsumer`` object to emit an
        ``ItHappened`` event, with charm code::
```

testing/src/scenario/context.py
```diff
@@ -365,6 +365,50 @@ def action(
...
+
+            class MyCharm(ops.CharmBase):
+                def __init__(self, framework: ops.Framework):
+                    super().__init__(framework)
+                    self.source = MyConsumer(self, 'source')
+                    framework.observe(self.source.on.it_happened, self._on_happened_event)
```
dwilding, line 383: Are you deliberately using `on_happened_event` vs `it_happened` to help distinguish the event name from the handler name? I think would have expected the names to match.

testing/src/scenario/context.py
```diff
@@ -365,6 +365,50 @@ def action(
...
+
+        Emit this event with::
```
dwilding, line 385: Suggest making this more explicit:

```suggestion
        To emit a `HappenedEvent` event::
```

testing/src/scenario/context.py
```diff
@@ -365,6 +365,50 @@ def action(
...
+
+            ctx.run(ctx.on.custom(
+                MyConsumer.on.it_happened, 1, 2, foo='bar'
+            ), state)
+
+        Custom events do not have access to the Juju context of the originating
```
dwilding, line 391: "originating" vs "original"? I'm not sure I have a preference. Would "upstream" be a more appropriate word to use?

testing/src/scenario/context.py
```diff
@@ -365,6 +365,50 @@ def action(
...
+        hook event.
+
+        Any additional arguments are passed through when instantiating the event
+        provided to observing handlers. Any of these arguments that are State
+        components (such as :attr:`ops.testing.Relation`) will be converted to
+        their ops counterparts (such as :attr:`ops.Relation`).
```
dwilding, line 397: Something seems off with the first sentence. How about this?

```suggestion
        Any additional arguments are passed through to the observing handlers
        when instantiating the event. Any of these arguments that are State
        components (such as :attr:`ops.testing.Relation`) will be converted to
        their ops counterparts (such as :attr:`ops.Relation`).
```

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
 (manage-libraries)=
 # Manage libraries
+
 > See first: {external+charmcraft:ref}`Charmcraft | Manage libraries <manage-libraries>`
 
+## Use a library
 
-## Write a library
+In your `src/charm.py`, observe the custom events it puts at your disposal. For example, a database library may have provided a  `ready` event -- a high-level wrapper around the relevant `juju` relation events -- so you use it to manage the database relation in your charm:
```
dwilding, line 8: I'd suggest a few adjustments, mainly to make it clearer what "it" refers to:

```suggestion
In `src/charm.py`, observe the custom events that the library provides. For example, a database library may have provided a  `ready` event -- a high-level wrapper around the relevant `juju` relation events. You can use the `ready` event to manage the database relation in your charm:
```

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
...
+        super().__init__(framework)
+        self.database = DatabaseRequirer(self, 'db-relation')
+        framework.observe(self.database.on.ready, self._on_db_ready)
+
+    def _on_db_ready(self, event: DatabaseReadyEvent):
+        secret_content = event.credential_secret.get_content()
+        ...
+```
+
+The tests for the charm that uses this library look like:
```
dwilding, line 26: Minor suggestion, since "this" could be misinterpreted as referring to the charm what we just showed.

```suggestion
The tests for the charm that uses the database library look like:
```

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
...
-    def __init__(self, charm, ...):
-        [...]
-        self.framework.observe(self.charm.on.database_relation_changed, self._on_db_changed)
-    
-    def _on_db_changed(self, e):
-        if [...]:  # check remote end has sent valid data
-            self.on.ready.emit()
+## Write a library
+
+When you're writing libraries, instead of callbacks, use custom events; this results in a more `ops`-native-feeling API. A custom event is, from a technical standpoint, an [](ops.EventBase) subclass that can be emitted at any point throughout the charm's lifecycle. These events are totally unknown to Juju. They are essentially charm-internal, and can be useful to abstract certain conditional workflows and wrap the top level Juju event so it can be observed independently.
```
dwilding, line 44: This paragraph really helped me get the point of how a library works. Nice :smile: 

A couple of nit suggestions:
```suggestion
When you're writing libraries, instead of callbacks, use custom events. This results in a more Ops-native feeling API. From a technical standpoint, a custom event is an [](ops.EventBase) subclass that can be emitted to Ops at any point throughout the charm's lifecycle. These events are totally unknown to Juju. They are essentially charm-internal, and can be useful to abstract certain conditional workflows and wrap the top level Juju event so it can be observed independently.
```

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
...
+
+```{important}
+Custom events must inherit from `EventBase`, but not from an `ops` subclass of
```
dwilding, line 47: I'd probably change `ops` to Ops here, for consistency.

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
...
+    
+    def _on_db_changed(self, event: ops.RelationChangedEvent):
+        if remote_data_is_valid(event.relation):
+            secret = ...
+            self.on.ready.emit(credential_secret=secret)
+```
+
+## Write tests
+
+### Test initialisation
```
dwilding, line 94: I first thought that this would be about initialising the test framework, so how about "Test that the library initialises"?

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
...
+
+In `tests/unit/test_my_lib.py`, add an initial test that a charm can initialise
```
dwilding, line 96: Small suggestion to remove "initial" and follow the pattern used later in the doc:

```suggestion
In `tests/unit/test_my_lib.py`, add a test that validates that a charm can initialise
```

docs/howto/manage-libraries.md
```diff
@@ -1,100 +1,144 @@
...
+### Test custom endpoint names
+
+If `DatabaseRequirer` is a relation endpoint wrapper, a frequent pattern is to
+allow customising the name of the endpoint that the object is wrapping.
 
-#### Relation endpoint wrapper lib
+> Examples: [`traefik's ingress-per-unit`](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v1/ingress_per_unit.py) lib
 
-If `MyObject` is a relation endpoint wrapper such as [`traefik's ingress-per-unit`](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v1/ingress_per_unit.py) lib, a frequent pattern is to allow customizing the name of the endpoint that the object is wrapping. We can write a test like so:
+In our `tests/unit/test_my_lib.py` file, add a test that validates that custom
```
dwilding, line 134: our --> your for consistency with the rest of the doc:

```suggestion
In your `tests/unit/test_my_lib.py` file, add a test that validates that custom
```

docs/howto/manage-libraries.md
```diff
@@ -129,104 +174,26 @@ def context(my_charm_type):
...
 
-### Advanced use cases
-
-#### Testing internal (charm-facing) library APIs
+### Test that the custom event is emitted
 
-Suppose that `MyObject` has a `data` method that exposes to the charm a list containing the remote databag contents (the `my-data-N` we have seen above).
-We can use `Context` as a context manager to run code within the lifetime of the Context like so:
+To verify that the library does emit the custom event appropriately, in the
+`tests/unit/test_my_lib.py` file, add the test:
```
dwilding, line 186: Recommend a small adjustment to match the rest of the doc:

```suggestion
To verify that the library does emit the custom event appropriately,
add a test in your `tests/unit/test_my_lib.py` file:
```

docs/howto/manage-libraries.md
tonyandrewmeyer:
```suggestion
A unit test for the charm that uses the database library looks like:
```

docs/howto/manage-libraries.md
tonyandrewmeyer: I thought that we were meant to be standardising on `ops` but happy to go with this.

```suggestion
Custom events must inherit from `EventBase`, but not from an Ops subclass of
```

docs/howto/manage-libraries.md
tonyandrewmeyer:
```suggestion
### Test that the library initialises
```

testing/src/scenario/context.py
tonyandrewmeyer: It's the current Juju event, during which the library has emitted the custom event. "original" to me feels more like a first event, like `install` is the "original" Juju event. "originating" felt like it was saying it was the Juju event that the custom event originates from.

Maybe "upstream" is clearer? Maybe "triggering"? Maybe "current" (except that's only true if you ignore deferring)? Maybe "of the Juju hook event that caused the library to emit the event"?

testing/src/scenario/context.py
tonyandrewmeyer: "passed through to the observing handlers" sounds to me like you'd get a handler signature like `def _on_event(self, event, extra_arg1, extra_arg2)`. The extra arguments here are provided when creating the custom event (the event `__init__` looks like `def __init__(self, handle, extra_arg1, extra_arg2)`).

What about something like this?

```suggestion
        If the custom event takes more arguments than an :attr:`ops.Handle`,
        pass these as additional arguments to this method. Any of these arguments that are State
        components (such as :attr:`ops.testing.Relation`) will be converted to
        their ops counterparts (such as :attr:`ops.Relation`).
```

testing/src/scenario/context.py
tonyandrewmeyer: There's no requirement that the `on` attribute has the same name as the event class that it emits. That is usually the case, other than some normalisation, e.g. `on.start`->`StartEvent`, `on.update_status`->`UpdateStatusEvent`, `source.on.sources_changed`->`GrafanaSourcesChanged` (note with this example, there's an additional "Grafana" and no "Event").

I have no objection to adjusting the names here so that they are more aligned. I was trying to avoid being too specific, like calling it `database_ready` and `DatabaseReadyEvent` and `_on_database_ready`, but maybe it is best to just do that?

tonyandrewmeyer:
@benhoyt this should be ready for your review now, I think. Thanks!

testing/src/scenario/context.py
dwilding: Yeah, I think I prefer more realistic names like `DatabaseReadyEvent` etc. Good idea. Thanks for explaining - you're right that I mistakenly expected the event name to match the `on` attribute.

testing/src/scenario/context.py
dwilding: Let's stick with "originating", as I didn't have a problem understanding what you meant by it :slightly_smiling_face:

testing/src/scenario/context.py
dwilding: Ah right, I get the point now. The handler would access the extra arguments as part of the event data. I like your suggestion. Would it be possible to change "this method" to `custom()`, to make the point even clearer?

benhoyt (approved):
Looks good to me, now. And thanks for the thorough docs updates.

dimaqq:
(3) for me too, pls.

I wouldn't be totally against raising an exception for bad subclass in tests, but am a bit vary of breaking existing code/tests. So maybe raising an exception when a test uses "instantiates" a custom event in question?

tonyandrewmeyer:
> I wouldn't be totally against raising an exception for bad subclass in tests, but am a bit vary of breaking existing code/tests. So maybe raising an exception when a test uses "instantiates" a custom event in question?

Yeah, the approach here is to raise when you call `custom` in your test with a custom event source that isn't compatible. Since it's all new, it won't break anything - it just prevents you using the new functionality unless you change the lib.

testing/src/scenario/context.py
```diff
@@ -365,6 +365,51 @@ def action(
...
+        For example, suppose that a library uses a ``DatabaseRequirer`` object
+        to emit a ``DatabaseReadyEvent`` event, with charm code::
+
+            class MyCharm(ops.CharmBase):
+                def __init__(self, framework: ops.Framework):
+                    super().__init__(framework)
+                    self.db = DatabaseRequirer(self, 'db-relation')
+                    framework.observe(self.db.on.ready, self._on_db_ready)
+
+        To emit a `DatabaseReadyEvent` event::
```
dwilding, line 385: Adding extra backticks:

```suggestion
        To emit a ``DatabaseReadyEvent`` event::
```

testing/src/scenario/context.py
```diff
@@ -365,6 +365,51 @@ def action(
...
+
+            ctx.run(ctx.on.custom(
+                DatabaseRequirer.on.ready, 1, 2, foo='bar'
+            ), state)
+
+        Custom events do not have access to the Juju context of the originating
+        hook event.
+
+        If the custom event takes more arguments than an :attr:`ops.Handle`,
+        pass these as additional arguments to `custom()`. Any of these
```
dwilding, line 395: Adding extra backticks:

```suggestion
        pass these as additional arguments to ``custom()``. Any of these
```

dwilding (approved):
Very nice, thank you @tonyandrewmeyer! I spotted a couple of formatting issues in the description of `custom()` - I've added specific comments.

testing/src/scenario/context.py
tonyandrewmeyer: Bah, too much switching between Markdown and ReST. I ought to have noticed this, though. Thanks!

testing/tests/test_context_on.py
```diff
@@ -344,3 +347,333 @@ def test_relation_departed_event():
...
+        assert subordinaterelation is not None
+        self.subordinaterelation = subordinaterelation
+        self.notice = ops.pebble.Notice(
+            snapshot["notice_id"],
+            None,
+            "",
+            snapshot["notice_key"],
+            datetime.datetime.now(),
+            datetime.datetime.now(),
+            datetime.datetime.now(),
```
dimaqq, line 463: Minor: is the value significant?
I wonder if testing with a fixed value would be cleaner than 3 teenily-tinily different current values.

dimaqq (approved):

testing/tests/test_context_on.py
tonyandrewmeyer: Nothing in the tests use any of the times, and the three timestamps could be different in practice, but I agree it's tidier for this test to use one value.

PR #1590, opened 2025-02-26
ci: handle presence/absence of "static" and "static-charm" envs

tonyandrewmeyer:
The observability charms we test against now have a mixture of environment names, so check whether each is present before trying to run them.

### DIFF
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -57,9 +57,11 @@ jobs:
             echo "::set-output name=exists::false"
           fi
 
-      - name: Run the charm's scenario unit tests
-        if: ${{ steps.check-tox-env-scenario.outputs.exists == 'true' && !(matrix.disabled) }}
-        run: tox -vve scenario
+      # alertmanager-k8s has old Scenario tests, so we can't do this until those
+      # are updated.
+#      - name: Run the charm's scenario unit tests
+#        if: ${{ steps.check-tox-env-scenario.outputs.exists == 'true' && !(matrix.disabled) }}
+#        run: tox -vve scenario
 
       - name: Check if 'static' tox environment exists
         id: check-tox-env-static
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -52,9 +52,9 @@ jobs:
         id: check-tox-env-scenario
         run: |
           if tox --listenvs | grep -q "^scenario$"; then
-            echo "::set-output name=exists::true"
+            echo "exists=true" >> "$GITHUB_OUTPUT"
           else
-            echo "::set-output name=exists::false"
+            echo "exists=false" >> "$GITHUB_OUTPUT"
           fi
 
       # alertmanager-k8s has old Scenario tests, so we can't do this until those
@@ -67,9 +67,9 @@ jobs:
         id: check-tox-env-static
         run: |
           if tox --listenvs | grep -q "^static$"; then
-            echo "::set-output name=exists::true"
+            echo "exists=true" >> "$GITHUB_OUTPUT"
           else
-            echo "::set-output name=exists::false"
+            echo "exists=false" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Run the charm's static analysis checks
@@ -80,9 +80,9 @@ jobs:
         id: check-tox-env-static-charm
         run: |
           if tox --listenvs | grep -q "^static-charm$"; then
-            echo "::set-output name=exists::true"
+            echo "exists=true" >> "$GITHUB_OUTPUT"
           else
-            echo "::set-output name=exists::false"
+            echo "exists=false" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Run the charm's static (charm) analysis checks
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -51,6 +51,7 @@ jobs:
       - name: Check if 'scenario' tox environment exists
         id: check-tox-env-scenario
         run: |
+          tox --listenvs
           if tox --listenvs | grep -q "^scenario$"; then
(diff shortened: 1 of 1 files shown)


.github/workflows/observability-charm-tests.yaml
```diff
@@ -48,6 +48,43 @@ jobs:
...
+#        if: ${{ steps.check-tox-env-scenario.outputs.exists == 'true' && !(matrix.disabled) }}
+#        run: tox -vve scenario
+
+      - name: Check if 'static' tox environment exists
+        id: check-tox-env-static
+        run: |
+          if tox --listenvs | grep -q "^static$"; then
+            echo "::set-output name=exists::true"
+          else
+            echo "::set-output name=exists::false"
```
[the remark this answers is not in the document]
dimaqq: Actually, since the output is consumed in the same job, and steps of a single job run in the same vm, you may as well:
- touch a magic file
- set an env var (see below)

```yaml
jobs:
  set_and_use_env_vars:
    runs-on: ubuntu-latest
    steps:
      - name: Set environment variable
        run: echo "MY_ENV_VAR=myValue" >> $GITHUB_ENV

      - name: Use environment variable
        run: |
          echo "The value of MY_ENV_VAR is $MY_ENV_VAR"
```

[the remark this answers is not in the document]
james-garner-canonical: I used the `echo` env var approach [here](https://github.com/canonical/operator/pull/1585/files. You can test for it in `if` using `env.MY_ENV_VAR`.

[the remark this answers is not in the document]
james-garner-canonical: I guess attaching to the step output is nicer here since we have a number of conditional steps depending on the presence of different environments. Though `HAS_SCENARIO`, `HAS_STATIC` and `HAS_STATIC_CHARM` could be fine.

.github/workflows/observability-charm-tests.yaml
```diff
@@ -48,6 +48,43 @@ jobs:
...
+          fi
+
       - name: Run the charm's static analysis checks
-        if: ${{ !(matrix.disabled) }}
+        if: ${{ steps.check-tox-env-static.outputs.exists == 'true' && !(matrix.disabled) }}
```
james-garner-canonical, line 76: How does `matrix.disabled` work? Will this step still run if (e.g.) unit tests fail?

james-garner-canonical, line 76: How does `matrix.disabled` work? Will this step still run if (e.g.) unit tests fail?

tonyandrewmeyer: Unless I'm missing something, the behaviour of `matrix.disabled` doesn't change here. It's hard-coded in the matrix values, so the behaviour doesn't change based on any passes/fails. It's also used to skip all of the testing sets so if it was true then the unit tests can't fail, they'll always be skipped (I guess the checks for environments could be skipped if everything is disabled, since they won't be used).

.github/workflows/observability-charm-tests.yaml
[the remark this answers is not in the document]
tonyandrewmeyer: Thanks both! I prefer the output over the envvar, so that the code that sets the value can be identical, and it seems cleaner to me, since it's not used for anything other than the GitHub workflow itself. I've moved to the recommended way.

I do hope that all of this can be removed in the not-distant future, if we have a spec that says "have at least these 5 commands that do these specific things executed in one of these 3 ways", and then we can convince the owners of the charms we regularly test against to adopt that.

.github/workflows/observability-charm-tests.yaml
tonyandrewmeyer: Somehow there seems to be an identical comment on the same lines - resolving this one.

james-garner-canonical: > Unless I'm missing something, the behaviour of matrix.disabled doesn't change here.

I don't really know what does in the first place, a quick google search wasn't helpful. I don't really understand what you mean about it being hard-coded. It doesn't look like it's a field we set in the matrix, so I'm guessing it's a built-in thing?

What I'm really wondering though is, do earlier steps failing (e.g. unit tests) stop static tests from running, and if so is that what we want for this workflow?

tonyandrewmeyer: > I don't really know what does in the first place, a quick google search wasn't helpful. I don't really understand what you mean about it being hard-coded. It doesn't look like it's a field we set in the matrix, so I'm guessing it's a built-in thing?

No, it's a Charm-Tech thing specific to these workflows. See https://github.com/canonical/operator/pull/1463 (note [this comment](https://github.com/canonical/operator/pull/1463#pullrequestreview-2463197427) pointing out that it'll be confusing when nothing is disabled unless we document it ...).

> What I'm really wondering though is, do earlier steps failing (e.g. unit tests) stop static tests from running, and if so is that what we want for this workflow?

It's whatever the default workflow behaviour is, and shouldn't be changed by this PR. I think the answer is that they do exit, yes. If you think that should be changed then it should be a separate PR IMO (and apply equally to the other similar workflows I would think).

james-garner-canonical: > note this comment pointing out that it'll be confusing when nothing is disabled unless we document it ...

prophetic!

> It's whatever the default workflow behaviour is, and shouldn't be changed by this PR

Fair enough, thanks, I was mainly wondering because I figured I should be using the same approach (I will use output instead of env I think), but that makes sense here.

Responding to this an earlier comment:
> I guess the checks for environments could be skipped if everything is disabled, since they won't be used

I agree, they could also be behind if: `${{ !(matrix.disabled) }}`, and probably the matrix.disabled condition should come before steps.check-tox-env-static.outputs.exists here. Not a big deal though

james-garner-canonical (approved):
Looks good, I'll follow the same approach in the broad charm tests since there are some using `static-charm` there too

IronCore864 (approved):

PR #1596, opened 2025-02-28
refactor: when testing deferral, don't observe life cycle events

dimaqq:
Pulled out from the OTEL branch.
Since life cycle events can't be deferred anyway, let's not observe them in these tests.
I hope this makes the tests more focused, and perhaps easier to maintain.

Inspired by #1589

### DIFF
--- a/testing/tests/test_e2e/test_deferred.py
+++ b/testing/tests/test_e2e/test_deferred.py
@@ -1,13 +1,12 @@
 import pytest
 from ops.charm import (
     CharmBase,
-    CollectStatusEvent,
     RelationChangedEvent,
     StartEvent,
     UpdateStatusEvent,
     WorkloadEvent,
 )
-from ops.framework import Framework
+from ops.framework import Framework, LifecycleEvent
 
 from scenario import Context
 from scenario.state import Container, Relation, State, _Event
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (changes requested):
The change looks good to me - but the PR title seems wrong. This isn't ignoring lifecycle events (which would be a bad change I think), this is making the test code do the same thing but be more maintainable.

dimaqq:
I've rewritten the PR title and description.

I think what motivated my change was that I was confused by "defer by count" in the first place.

It felt like a total coincidence that CollectStatus was not deferred.

tonyandrewmeyer:
> I've rewritten the PR title and description.
> 
> I think what motivated my change was that I was confused by "defer by count" in the first place.
> 
> It felt like a total coincidence that CollectStatus was not deferred.

Sorry, this is mostly my bad. I somehow overlooked before the new if statement, so I was only seeing the change in how the observed events are extracted and that didn't line up with the title.

However, it's not really a coincidence, it's careful manipulation of the counts so that the lifecycle ones aren't deferred (the test would error if that happened, since defer raises).

I don't think the "defer by count" is great for maintainability - this change probably does improve things a bit so that when figuring out counts you can skip over the lifecycle ones.

tonyandrewmeyer (approved):

james-garner-canonical (approved):

PR #1599, opened 2025-02-28
fix: in tests, raise the same error as Pebble when an invalid service is provided

### LINKED ISSUE #1310: Align the exception raised from Harness when a service does not exist
Basically, we want to [fix this TODO](https://github.com/canonical/operator/blob/449793f9867eba11dc2e7e1c28ea3dca1e1da231/ops/testing.py#L3123) (and the related following ones).

When starting, stopping, restarting a service, if the service does not exist, then Harness will raise a `RuntimeError`, but if this happens outside of tests, then `ops.pebble.APIError` will be raised.

It's better if the Harness exceptions match the ones you normally get, so that the charm doesn't have to catch exceptions just to make test behaviour match the normal behaviour.

Note that Scenario also uses the ops.testing Pebble client, so this will also adjust Scenario as well.

tonyandrewmeyer:
This PR addresses the TODO and notes in the start/stop/restart service methods of Harness (and via inheritance, Scenario).

* Raise an APIError rather than RuntimeError, with the same message as Pebble, when a service doesn't exist.
* Return a ChangeID, now that we're keeping track of these.

Also: check for an empty list of services and raise the same APIError as Pebble in that case.

Fixes #1310

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3213,75 +3213,102 @@ class _TestingPebbleClient:
                 self._new_perform_check(info)
         return self.autostart_services(timeout, delay)
 
+    def _new_service_change(self, kind: pebble.ChangeKind, services: List[str]) -> pebble.Change:
+        now = datetime.datetime.now()
+        if kind == pebble.ChangeKind.START:
+            action = 'Start'
+        elif kind == pebble.ChangeKind.STOP:
+            action = 'Stop'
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Nice fix, thank you!

james-garner-canonical (approved):
Looks good, sorry for the delay in reviewing this

PR #1604, opened 2025-03-03
docs: add section about services with long startup time

### LINKED ISSUE #1402: Write how-to doc about how to handle workloads with variable startup times
This is in leiu of initial-delay (which we [decided not to implement](https://github.com/canonical/pebble/issues/145#issuecomment-2314206029)). Basically a short doc that describes the "recommendation" in the first point of [that comment](https://github.com/canonical/pebble/issues/145#issuecomment-2314206029).

benhoyt:
Referencing the issue, I don't think this needs a whole how-to, just a section with a couple of paragraphs seems sufficient.

Fixes #1402

### DIFF
--- a/docs/howto/run-workloads-with-a-charm-kubernetes.md
+++ b/docs/howto/run-workloads-with-a-charm-kubernetes.md
@@ -513,7 +513,13 @@ If there are no checks configured, Pebble returns HTTP 200 so the liveness and r
 
 Consider the K8s liveness success (`level=alive` check) to mean "Pebble is alive" rather than "the application is fully alive" (and failure to mean "this container needs to die"). For charms that take a long time to start, you should not have a `level=alive` check (if Pebble's running, it will report alive to K8s), and instead use an ordinary Pebble check (without a `level`) in conjunction with `on-check-failure: restart`. That way Pebble itself has full control over restarting the service in question.
 
-### Test checks
+### Services with long startup time
+
+When a K8s liveness probe (a `level-alive` check) succeeds, you should consider it to mean "Pebble is alive" rather than "workload is alive". Similarly, probe failure mean "this container needs to be restarted".
+
+This means that for workloads that take a long or indefinite period of time to start, you should not have a `level=alive` check for that service. Instead, use an ordinary Pebble check without a level in conjunction with `on-check-failure: restart` on the service. That way Pebble itself has control over restarting the service in question.
+
+### Testing with checks
 
 > Added in ops 2.17
 
--- a/docs/howto/run-workloads-with-a-charm-kubernetes.md
+++ b/docs/howto/run-workloads-with-a-charm-kubernetes.md
@@ -517,7 +517,7 @@ Consider the K8s liveness success (`level=alive` check) to mean "Pebble is alive
 
 When a K8s liveness probe (a `level=alive` check) succeeds, you should consider it to mean "Pebble is alive" rather than "the workload is alive". Similarly, a liveness probe failure means "this container needs to be restarted" rather than an issue with the workload.
 
-This means that for workloads that take a long or indefinite period of time to start, you should not have a `level=alive` check for the service. Instead, use a Pebble check without a level and specify `on-check-failure: restart` for the service. That way Pebble itself has control over restarting the service.
+This means you should not usually have a `level=alive` check for a service in a charm. This is especially important for workloads that take a long or indefinite period of time to start. Instead, use a Pebble check without a level and specify `on-check-failure: restart` for the service. That way Pebble itself has control over restarting the service.
 
 ### Testing with checks
 
--- a/docs/howto/run-workloads-with-a-charm-kubernetes.md
+++ b/docs/howto/run-workloads-with-a-charm-kubernetes.md
@@ -515,7 +515,7 @@ Consider the K8s liveness success (`level=alive` check) to mean "Pebble is alive
 
 ### Services with long startup time
 
-When a K8s liveness probe (a `level-alive` check) succeeds, you should consider it to mean "Pebble is alive" rather than "workload is alive". Similarly, probe failure mean "this container needs to be restarted".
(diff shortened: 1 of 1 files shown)


docs/howto/run-workloads-with-a-charm-kubernetes.md
```diff
@@ -513,7 +513,13 @@ If there are no checks configured, Pebble returns HTTP 200 so the liveness and r
 
 Consider the K8s liveness success (`level=alive` check) to mean "Pebble is alive" rather than "the application is fully alive" (and failure to mean "this container needs to die"). For charms that take a long time to start, you should not have a `level=alive` check (if Pebble's running, it will report alive to K8s), and instead use an ordinary Pebble check (without a `level`) in conjunction with `on-check-failure: restart`. That way Pebble itself has full control over restarting the service in question.
 
-### Test checks
+### Services with long startup time
+
+When a K8s liveness probe (a `level-alive` check) succeeds, you should consider it to mean "Pebble is alive" rather than "workload is alive". Similarly, probe failure mean "this container needs to be restarted".
```
dwilding, line 518: Some small suggestions/fixes:

```suggestion
When a K8s liveness probe (a `level=alive` check) succeeds, you should consider it to mean "Pebble is alive" rather than "the workload is alive". Similarly, a liveness probe failure means "this container needs to be restarted" rather than an issue with the workload.
```

docs/howto/run-workloads-with-a-charm-kubernetes.md
```diff
@@ -513,7 +513,13 @@ If there are no checks configured, Pebble returns HTTP 200 so the liveness and r
...
+
+This means that for workloads that take a long or indefinite period of time to start, you should not have a `level=alive` check for that service. Instead, use an ordinary Pebble check without a level in conjunction with `on-check-failure: restart` on the service. That way Pebble itself has control over restarting the service in question.
```
dwilding, line 520: I think it's fine to say "the service" instead of "that service" and "the service in question". Also I'd suggest adjusting the middle sentence slightly for readability:

```suggestion
This means that for workloads that take a long or indefinite period of time to start, you should not have a `level=alive` check for the service. Instead, use a Pebble check without a level and specify `on-check-failure: restart` for the service. That way Pebble itself has control over restarting the service.
```

dwilding (approved):
Suggested some tweaks, but good for me - thanks!

benhoyt:
@tonyandrewmeyer Ping for review on this small docs update.

tonyandrewmeyer:
Apologies for the slow review!

A few thoughts:

1. Aren't we essentially saying here "don't use level=alive in charms"? The text here says to not use one if the workload takes a long time to start, but it seems like it's also saying that it doesn't make sense in general, because it's saying "Pebble is alive". If it does make sense, then an example of when it does would help, I think.
2. My recollection was that we wanted to have some guidance on how to handle workloads that take a long time to start up. Using `on-check-failure: restart` is definitely better than `level=alive`, since Pebble will restart the service rather than K8s restarting the container. However, you're still in a bad place, aren't you? You have to put in some combination of period+failures that mean that the service will be running within `period*failure` seconds. What do we actually suggest when there's an indeterminate amount of time for it to start? Maybe we have no answer for this now and need to wait on something like JU138? Or can we suggest something now?

benhoyt:
> 1. Aren't we essentially saying here "don't use level=alive in charms"? The text here says to not use one if the workload takes a long time to start, but it seems like it's also saying that it doesn't make sense in general, because it's saying "Pebble is alive". If it does make sense, then an example of when it does would help, I think.

Good point. I've rearranged the second paragraph a bit to make this clearer.

> 2. My recollection was that we wanted to have some guidance on how to handle workloads that take a long time to start up. Using on-check-failure: restart is definitely better than level=alive, since Pebble will restart the service rather than K8s restarting the container. However, you're still in a bad place, aren't you? You have to put in some combination of period+failures that mean that the service will be running within period*failure seconds. What do we actually suggest when there's an indeterminate amount of time for it to start? Maybe we have no answer for this now and need to wait on something like JU138? Or can we suggest something now?

True, but at least you have a fair bit of control over that (unless the K8s probe settings, which are controlled by Juju), and you're actually checking whether the service is up/down, rather than just the process running. I don't think we really have a better answer now, so should leave this until we get something like JU138 (hah, thanks for reminding me about that spec).

benhoyt:
@tonyandrewmeyer See above -- ready for re-review at some point.

tonyandrewmeyer (approved):

tonyandrewmeyer:
> > 2. My recollection was that we wanted to have some guidance on how to handle workloads that take a long time to start up.
[...]
> I don't think we really have a better answer now, so should leave this until we get something like JU138

Ok, fair enough. I think it would be nice if we did have some suggestions (for example, you could start with a layer that has the check with a low threshold and period and no service, and then in the `check-failed` event add the service without `on-check-failure` set and adjust the threshold and period, and then in the `check-recovered` event add `on-check-failure`; or if the workload has a hook for it you could use a custom notice to adjust when the workload is ready; probably other ideas not coming to mind immediately). However, this how-to probably isn't the place for those, and it's perhaps better to just focus on a good solution instead of providing workarounds.

benhoyt:
> Ok, fair enough. I think it would be nice if we did have some suggestions (for example, you could start with a layer that has the check with a low threshold and period and no service, and then in the check-failed event add the service without on-check-failure set and adjust the threshold and period, and then in the check-recovered event add on-check-failure; or if the workload has a hook for it you could use a custom notice to adjust when the workload is ready; probably other ideas not coming to mind immediately). However, this how-to probably isn't the place for those, and it's perhaps better to just focus on a good solution instead of providing workarounds.

That's an interesting idea. However, I'm wary of suggesting a relatively complicated solution like that without having proven it in the wild. I'm going to leave this as is for now, can always improve further later or push for a JU138-like feature.

PR #1606, opened 2025-03-04
chore: use relative imports

benhoyt:
Per discussion at daily. I thought this would be a trivial job, but the Ops imports are a bit of a mess, so it was a bit more work than usual.

As Dima pointed out, we have to be careful in public modules (like `ops.charm`) not to remove existing names or introduce new names. I think I've done that right, but please review that carefully. The one exception is when in `model.py` or similar I've not worried about removing `ops`, because if you're writing `ops.charm.ops.foo` you deserve an `AttributeError`, at least. :-)

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -54,7 +54,8 @@ counts: Optional[Tuple[str, int]]
 
 ### Use relative imports inside a package
 
-In code inside a package (directory), use relative `.` imports. For example, in the `ops` directory of the Ops codebase:
+When writing code inside a package (a directory containing an `__init__.py` file), use relative imports with a `.` instead of absolute imports.
+For example, within the `ops` package:
 
 **Don't:**
 
--- a/STYLE.md
+++ b/STYLE.md
@@ -52,6 +52,27 @@ counts: Optional[Tuple[str, int]]
 ```
 
 
+### Use relative imports inside a package
+
+In code inside a package (directory), use relative `.` imports. For example, in the `ops` directory of the Ops codebase:
+
+**Don't:**
+
+```python
+from ops import charm
+```
+
+**Do:**
+
+```python
+from . import charm
+
+# Or, if you need to avoid adding the public name "charm" to the namespace:
+
+from . import charm as _charm
+```
+
+
 ### Avoid nested comprehensions and generator expressions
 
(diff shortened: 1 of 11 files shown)


james-garner-canonical:
Looks good. I'm thinking it would be nice to see a run of the published charm tests in this branch (you can run it in your fork) -- but if possible, after we merge in running published charms' static tests #1585 as well.

ops/testing.py
```diff
@@ -37,29 +37,22 @@
...
-    RelationNotFoundError,
-    RelationRole,
     YAMLStringOrFile,
-    charm,
-    framework,
-    model,
-    pebble,
-    storage,
 )
+from .charm import CharmBase, CharmMeta, RelationRole
```
dimaqq, line 53: What if someone imports `ops.testing.charm` today 🤔 
Let's validate that none actually uses the 5 disappearing re-exports.

benhoyt: No, they're still imported -- see line 40. Does that help?

dimaqq: Oh sorry. I didn't realise that `ops._private.harness.charm is ops.charm`...

STYLE.md
```diff
@@ -52,6 +52,27 @@ counts: Optional[Tuple[str, int]]
 ```
 
 
+### Use relative imports inside a package
+
+In code inside a package (directory), use relative `.` imports. For example, in the `ops` directory of the Ops codebase:
```
dimaqq, line 57:
```suggestion
When writing code inside a package (a directory containing an `__init__.py`), use relative imports with a `.` instead of absolute imports.
For example, within the `ops` package:
```

dimaqq (approved):

benhoyt:
@james-garner-canonical Good call. I've run it here https://github.com/benhoyt/operator/actions/runs/13667501618 -- exact same 7 failures as the [run on your branch](https://github.com/james-garner-canonical/operator/actions/runs/13663448044), so I think we're good here.

james-garner-canonical (approved):
I didn't catch anything wrong in my read through of the changes, and it's a great sign that all the tests (including `published-charms-tests`) haven't changed. Approving pending merge resolution.

PR #1607, opened 2025-03-04
ci: add compatibility testing for ops and ops-scenario versions

tonyandrewmeyer:
It's important that we know which versions of `ops` work with which versions of `ops-scenario`, so that we can set the version in the `ops[testing]` optional dependency and the dependencies of `ops-scenario` correctly.

For example, `ops-scenario` 7.2.0 broke compatibility with earlier versions of ops (which don't have the Pebble check startup support) but we missed bumping the required version, which means that you can install a broken combination, like `ops==2.18.1` and `ops-scenario==7.2.0`.

We also want to know when the main version of `ops` is going to break compatibility with older versions of `ops-scenario`, as with the changes from https://github.com/canonical/operator/pull/1586, so we can make sure we set the version correctly before releasing.

Issues with `main` of `ops` and `main` of `ops-scenario` should be covered by the regular test runs.

This PR runs the unit tests (of both `ops` and `ops-scenario`) for all the combinations of versions:
* That aren't explicitly prohibited by the dependency specifications in the projects
* That are after the merge into one repository (mostly just to keep this simple)
* That aren't ops 2.17.0, which doesn't use the local code for `tox -e unit`
* That aren't known failures (Scenario 7.2.0 and older ops)

The workflow is set to run once a month around release time, like the other compatibility checks, and can be manually run.

The workflow clones and checks out the specified ops branch/tag, and then clones the repo a second time, only including the top-level `testing` folder, checked out to the specified ops-scenario tag, and then replaces the top-level `testing` folder with that one. This means that we have both the branch Scenario code and the branch Scenario tests.

[Example run in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/13647822571).

### DIFF
--- a/.github/workflows/ops-scenario-matrix.yaml
+++ b/.github/workflows/ops-scenario-matrix.yaml
@@ -1,0 +1,60 @@
+name: ops/ops-scenario compatibility
+
+on:
+  workflow_dispatch:
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        ops: ["main", "2.17.1", "2.18.0", "2.18.1", "2.19.0"]
+        scenario: ["7.1.1", "7.1.2", "7.1.3", "7.2.0"]
+        exclude:
+        - {ops: "2.17.1", scenario: "7.1.1"} # ops-scenario 7.1.1 explicitly depends on ops 2.18
+        - {ops: "2.17.1", scenario: "7.1.2"} # ops-scenario 7.1.2 explicitly depends on ops 2.18
+        - {ops: "2.17.1", scenario: "7.1.3"} # ops-scenario 7.1.2 explicitly depends on ops 2.18
+        - {ops: "2.17.1", scenario: "7.2.0"} # Missed requiring ops 2.19.0 for CheckStartup
+        - {ops: "2.18.0", scenario: "7.2.0"} # Missed requiring ops 2.19.0 for CheckStartup
+        - {ops: "2.18.1", scenario: "7.2.0"} # Missed requiring ops 2.19.0 for CheckStartup
+
+    steps:
+      - name: Expand ref
+        id: set-ref
+        run: |
+          if [ "${{ matrix.ops }}" == "main" ]; then
+            echo "refs=refs/heads/main" >> "$GITHUB_OUTPUT"
+          else
+            echo "refs=refs/tags/${{ matrix.ops }}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Checkout ops
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
> I have one big concern with maintainability going forward -- isn't `exclude` going to grow exponentially, since a new `scenario` that requires `ops>=2.X` is going to have to add a line for every listed `ops<2.X`?

Yes. My hope is that we don't have that often - but I'm not sure that this goal is shared across Charm-Tech. I do think we would likely prune the lists periodically, or perhaps we should only have the versions from the minimum in the latest release up? That would be 7.0.5 and up for Scenario and 2.18.0 and up for ops.

> Would it be better if `exclude` only contained 'bad' releases where we missed bumping the requirements correctly, and otherwise the job exits early with success if the requirements forbid the combination?

This does sound like a nice improvement. Any chance you know how to do this off the top of your head? If not, I can look into it.

> Another future nice to have would be if the `ops` and `scenario` matrix could be populated automatically -- it seems to be possibly to both [list github releases](https://github.com/marketplace/actions/list-github-releases) and [dynamically generate the matrix](https://stackoverflow.com/a/67180216). This would drop a manual step from the release process.

Yeah, I thought about that too, and wasn't sure whether it would be worth it. Rather than getting the releases from GitHub or from PyPI, I think it'd be simplest to just get the list of tags.

james-garner-canonical:
I guess changes to `ops` requiring a higher scenario version, and vice versa will hopefully be rare. I do think we'd want to keep older ops and scenarios in the lists to make sure we've fenced them off properly via dependency specification from the latest releases if they'll break them -- but maybe that only requires the one before the current minimum? I'm not sure here.

>> Would it be better if exclude only contained 'bad' releases where we missed bumping the requirements correctly, and otherwise the job exits early with success if the requirements forbid the combination?

> This does sound like a nice improvement. Any chance you know how to do this off the top of your head? If not, I can look into it.

I'd guess something like `continue-on-error` in a step that tries to `pip install`, then a subsequent step that checks the output (what output exactly? I'm not sure) and either sets a flag to perform subsequent steps if it was a success, skip subsequent steps if it's a failure because of an `ops` and `ops-scenario` clash, and exit with an error if it was some other failure ... but maybe (hopefully) there's a better way to probe for dependency clashes?

>> Another future nice to have would be if the ops and scenario matrix could be populated automatically ...

> Yeah, I thought about that too, and wasn't sure whether it would be worth it. Rather than getting the releases from GitHub or from PyPI, I think it'd be simplest to just get the list of tags.

I don't think it's a blocker at this point. Tags sounds like a good idea though.

tonyandrewmeyer:
Charm-Tech discussed this today and decided that (at least for now) we're going to release `ops` and `ops-scenario` together, always requiring the matching versions. That means that the compatibility tests won't show anything interesting.

PR #1608, opened 2025-03-04
ci: use the latest version of ops-scenario in the compatibility tests

tonyandrewmeyer:
We currently patch in the latest version of `ops`, but should really be also patching in the latest version of `ops-scenario` to make sure that we aren't breaking compatibility specifically in tests (at least as long as the latest versions are used).

[Observability run in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/13648333151)
[Broad compatibility run in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/13648336581)

### DIFF
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -32,13 +32,16 @@ jobs:
       - name: Install dependencies
         run: pip install tox~=4.2 uv~=0.6
 
-      - name: Update 'ops' dependency in test charm to latest
+      - name: Update 'ops' and 'ops-scenario' dependencies in test charm to latest
         run: |
           if [ -e "requirements.txt" ]; then
             sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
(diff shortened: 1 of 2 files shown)


james-garner-canonical (approved):
Thanks for doing this, @tonyandrewmeyer. Looks like the charms that are failing the `published-charms-tests` are the same ones that were already failing, so that's good.

One thing I was wondering about here is if we'd want/need to patch the `ops` `pyproject.toml` requirements here for charms that just use `ops[testing]`.  I guess adding the specific commit version to the charm's requirements overrules that anyway, so probably not needed.

tonyandrewmeyer:
> One thing I was wondering about here is if we'd want/need to patch the `ops` `pyproject.toml` requirements here for charms that just use `ops[testing]`. I guess adding the specific commit version to the charm's requirements overrules that anyway, so probably not needed.

At the moment we're basically assuming that one of the files that we are patching is a lock file, and that the install is done from there, so `ops[testing]` isn't relevant, since it wouldn't be in a lock file.

This definitely doesn't have to be the case - there might not be a lock file, only a regular dependencies file, and the tox configuration might have the dependencies (like ops annoyingly does at the moment), or there might be some other thing.

We do [tell people to have a lock file](https://docs.google.com/document/d/11JsKoxtIdFGy5LK9OYSMfMCR87hPpNt1tT8avoTy6RU), but that's a new requirement, so it's not clear if it's done by all the published charms.

I feel like keeping with the assumption that one of these is a lock file is ok for this PR and if we do find that it's missing some cases then we should look at opening a PR against those charms perhaps.

IronCore864 (approved):

PR #1616, opened 2025-03-06
docs: fix Scenario example

dimaqq:
Fixing a bug in our examples.
I wonder if perhaps this was the Scenario 6 way or something?
In that case there could be other bugs lurking around.

### DIFF
--- a/docs/howto/write-scenario-tests-for-a-charm.md
+++ b/docs/howto/write-scenario-tests-for-a-charm.md
@@ -56,7 +56,7 @@ def test_charm_runs():
 
     # Assert:
     #  Verify that the output state looks like you expect it to.
-    assert state_out.status.unit.name == 'unknown' 
+    assert state_out.unit_status.name == 'unknown'
 ```
 
 > See more: 


dwilding:
Thanks Dima. I'm working through this doc as we speak, so I'll hold off reviewing for an hour or two in case I spot anything else we should add to this PR.

docs/howto/write-scenario-tests-for-a-charm.md
```diff
@@ -56,7 +56,7 @@ def test_charm_runs():
 
     # Assert:
     #  Verify that the output state looks like you expect it to.
-    assert state_out.status.unit.name == 'unknown' 
+    assert state_out.unit_status.name == 'unknown'
```
dwilding, line 59: Should we have it as:

```suggestion
    assert state_out.unit_status == testing.UnknownStatus()
```

james-garner-canonical (approved):
Look good to me. Let's go with David's suggestion to use `UnknownStatus` if possible.

dwilding:
@dimaqq, @james-garner-canonical, I'm confused about the example at the end of the doc. Copying it here for reference:

```python
# For example, suppose that the charm we're testing uses the KubernetesServicePatch.
# To update the test above to mock that object, modify the test file to contain:

import pytest
from unittest import patch

@pytest.fixture
def my_charm():
    with patch("charm.KubernetesServicePatch"):
        yield MyCharm

# Then you should rewrite the test to pass the patched charm type to the Context,
# instead of the unpatched one. In code:

def test_charm_runs(my_charm):
    # Arrange: 
    #  Create a Context to specify what code we will be running
    ctx = Context(my_charm)
    # ...
```

Here's what I'm confused about:

1. The "patch" in `KubernetesServicePatch` has nothing to do with patching in the context of testing, right? It looks to me like an unfortunate use of the same term.

2. We need an `import charm` somewhere, right? Or maybe `from charm import MyCharm`.

3. The patching in `my_charm()` doesn't appear to actually do anything. Shouldn't we be doing something like:

    ```python
    @pytest.fixture
    def my_charm():
        with patch("charm.KubernetesServicePatch") as mocked:
            # some code to prepare `mocked`
            yield MyCharm

4. In `test_charm_runs()`, does it matter whether we do `ctx = Context(my_charm)` or `ctx = Context(MyCharm)`? In my (limited) testing, both seem to work. Is there a reason to prefer one over the other?

dwilding (approved):
Approving so that this PR isn't blocked. We can handle my questions in a follow-up if needed.

tonyandrewmeyer:
> from unittest import patch

Note that this should be `from unittest.mock import patch`. Personally, I think it's better to use `monkeypatch` with `pytest` rather than mix `unittest` with `pytest`, but I think that's mostly personal taste.

> 1. The "patch" in `KubernetesServicePatch` has nothing to do with patching in the context of testing, right? It looks to me like an unfortunate use of the same term.

Correct.

> 2. We need an `import charm` somewhere, right? Or maybe `from charm import MyCharm`.

You don't for the patching. `unittest.mock.patch` doesn't need the thing it's patching to be a name in the local namespace, it just has to be locally importable. However, to yield `MyCharm` you do need to import that, yes.

> 3. The patching in `my_charm()` doesn't appear to actually do anything. Shouldn't we be doing something like:
>    ```python
>    @pytest.fixture
>    def my_charm():
>        with patch("charm.KubernetesServicePatch") as mocked:
>            # some code to prepare `mocked`
>            yield MyCharm
>    ```

It depends what `KubernetesServicePatch` does. `unittest.mock.patch` will replace it with ` MagicMock`, which roughly means that anything the charm does with it will not fail but also not do anything. It's presumably meant to be the one from [this lib](https://github.com/canonical/kratos-operator/blob/6363da5539c635ace5dbd7118fd32b14cb15691d/lib/charms/observability_libs/v0/kubernetes_service_patch.py#L112). That means the charm will have some code like:

```
self.service_patcher = KubernetesServicePatch(
        self, [(f"{self.app.name}", 443, 443, 30666)], "LoadBalancer"
    )
```

By replacing `KubernetesServicePatch` with a `MagicMock`, `self.service_patcher` will also end up as a `MagicMock`, and that line won't do anything else - unpatched, it would do some observe calls and other things.

The intention here isn't to mock `KubernetesServicePatch` to give a specific result when called, it's to just null it out.

> 4. In `test_charm_runs()`, does it matter whether we do `ctx = Context(my_charm)` or `ctx = Context(MyCharm)`? In my (limited) testing, both seem to work. Is there a reason to prefer one over the other?

If you do `ctx = Context(MyCharm) you're using the unpatched charm. If you do `ctx = Context(my_charm)` you're using the patched charm. I don't know litekube well enough to know what creating a `Service` object (which is one thing that `KubernetesServicePatch.__init__` does) will do if you're not actually inside a K8s cluster, but I'm guessing it may crash.

In your limited testing, did your charm class have a line like the `self.service_patcher` one above? If not, then it wouldn't matter whether you used the patched charm or the unpatched one, because neither would be actually using what's being patched.

>    [Later edit] Continuing from question 3, I guess `my_charm()` shouldn't yield `MyCharm`. It should yield `mocked`. And `my_charm()` should be called something like `kubernetes_service_patch` instead. Then `test_charm_runs()` needs to be:
>    ```python
>    def test_charm_runs(kubernetes_service_patch):
>        # Arrange: 
>        #  Create a Context to specify what code we will be running
>        ctx = Context(MyCharm)
>        # Act:
>        # ...
>        # Assert:
>        # some code to assert how `kubernetes_service_patch` (aka `mocked` in the fixture) was used
>    ```

You could do it this way -- it would end up the same in the end.

If I was rewriting this chapter I would probably leave all this out entirely. It's teaching Python, not Scenario or ops.

dwilding:
Thank you @tonyandrewmeyer, this helps my understanding a lot! One point I still don't fully get is the patched vs unpatched charm question. In my testing, I thought the charm was using the thing that was being patched, and even then the unpatched charm worked. My interpretation was that the context (from `with`) that was being yielded was the key thing. Anyway, I've switched to using monkeypatch, which I'm finding easier to understand.

For an immediate action, I think the example in the doc should import the charm, otherwise it's broken for testing an existing charm type.

@dimaqq how about adding an extra import in the example, like this: https://github.com/dwilding/canonical-operator/commit/818e8b22ad17553efa454b1995d407ff9b81c1e1#diff-de92ec4d4ada4fd379bf89a5248fef84a3eacd4e67c571c6b690a5341c8cbb13

I agree that ideally the whole example should be improved/removed. As that would be future work (not to block this PR), I've added a comment in @IronCore864's [testing improvement plan](https://docs.google.com/document/d/1TQmxTn_VltOALJ19oMDtLxq763FNzhgJeRLC-3EGSp0/edit?disco=AAABe1xdj4g) so that we can find this discussion at a later date.

tonyandrewmeyer:
> One point I still don't fully get is the patched vs unpatched charm question. In my testing, I thought the charm was using the thing that was being patched, and even then the unpatched charm worked. My interpretation was that the context (from `with`) that was being yielded was the key thing. 

"Patched charm" was maybe a bad way to put it. The patch applies for the context of the `with` statement, and the fixture yields the charm within that context, so while the charm from the fixture is used, the patch is applied. When the fixture code resumes after the `yield` (after the text using the fixture completes) the patching is automatically undone. So it's not really the "patched charm", it's the "charm provided while the patching is happening".

> Anyway, I've switched to using monkeypatch, which I'm finding easier to understand.

:tada:

PR #1617, opened 2025-03-07
feat: ops.testing.State components are less mutable

### LINKED ISSUE #1605: should scenario's deepcopy be deeper?
one of the premises of scenario is that state-in doesn't get mutated. However, it turns out it is!
`copy.deepcopy(state)` doesn't recurse into `Relation, Network...`.

Test that fails on `main` and IMHO should pass:

```python
def test_state_immutable(mycharm):
    def event_handler(charm: CharmBase, _):
        rel = charm.model.get_relation("foo")
        rel.data[charm.app]["a"] = "b"
        rel.data[charm.unit]["c"] = "d"

    mycharm._call = event_handler

    relation_in = Relation(
        endpoint="foo",
        remote_app_name="remote",
    )
    state = State(relations={relation_in}, leader=True)

    out = trigger(
        state,
        event="start",
        charm_type=mycharm,
        meta={
            "name": "foo",
            "requires": {"foo": {"interface": "bar"}},
        },
    )

    relation_out = out.get_relation(relation_in.id)
    assert relation_in is not relation_out  # this raises!
    assert not relation_in.local_app_data  # this too!
```

tonyandrewmeyer:
This PR has two changes to make the `testing.State` less mutable (or more immuntable as my Kiwi fingers keep typing).

* All of the components that have a dictionary or list attribute now deepcopy that attribute in `__postinit__`. The type annotations are also changed to `Mapping` and `Sequence` so static type checking should detect when charm tests try to alter these (the actual underlying types are still the ones passed, and default to dict and list).
* In `State.__deepcopy__` we deepcopy all of the items inside the collections (the relations, the containers, and so forth) - this means that all of these items are separate in the input and output states. This also includes the config.

This also fixes a small bug that was found by the new tests: `State.get_storage` was wrongly comparing the index of the storage object to the *`str.index` function* (so would never match) rather than to the index of the `Storage` object in the state. The `| None` type annotation is also removed, as the index is `int`, not `int | None`, so it doesn't make sense to pass `None` in as the index.

Fixes #1605

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 import copy
 import dataclasses
 import datetime
-import functools
 import inspect
 import pathlib
 import random
@@ -112,24 +111,17 @@ _SECRET_EVENTS = {
 }
 
 
-def _detach_dicts_and_lists(cls):
-    original_post_init = getattr(cls, "__post_init__", None)
-
-    @functools.wraps(cls)
-    def __post_init__(self):
-        if original_post_init:
-            original_post_init(self)
-        # We don't have a frozendict to freeze any dictionaries, and while we
-        # could freeze a list into a tuple, we would break tests that currently
-        # assert that they get a list, so we can't do that until 8.0. That means
-        # we can't actually freeze the content here, but we can at least
-        # disassociate it from the original object.
-        for attr, value in self.__dict__.items():
-            if isinstance(value, (dict, list)):
-                object.__setattr__(self, attr, copy.deepcopy(value))
-
-    cls.__post_init__ = __post_init__
-    return cls
+def _deepcopy_mutable_fields(obj: object):
+    # We don't have a frozendict to freeze any dictionaries, and while we
+    # could freeze a list into a tuple, we would break tests that currently
+    # assert that they get a list, so we can't do that until 8.0. That means
+    # we can't actually freeze the content here, but we can at least
+    # disassociate it from the original object.
+    for attr, value in obj.__dict__.items():
+        if isinstance(value, (dict, list)):
+            # We expect that the obj is a frozen dataclass, so have to use
+            # object.__setattr__ to bypass the frozen check.
+            object.__setattr__(obj, attr, copy.deepcopy(value))
 
 
 # This can be replaced with the KW_ONLY dataclasses functionality in Python 3.10+.
@@ -234,7 +226,6 @@ class JujuLogLine:
 
 
 @dataclasses.dataclass(frozen=True)
-@_detach_dicts_and_lists
 class CloudCredential(_max_posargs(0)):
     __doc__ = ops.CloudCredential.__doc__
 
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
I'll also run this against the compatibility tests and with super_tox, but it should be reviewable now without needing to wait for that.

tonyandrewmeyer:
The regular compatibility tests (hello, db, obvs) pass, as do the [published charms](https://github.com/tonyandrewmeyer/operator/actions/runs/13713506395) ones.

tonyandrewmeyer:
There are no additional `unit`, `static`, or `scenario` failures with the charms I have locally.

testing/src/scenario/state.py
```diff
@@ -110,6 +112,26 @@
 }
 
 
+def _detach_dicts_and_lists(cls):
```
benhoyt, line 115: The name "detach" seems a bit odd to me -- why not `_copy_dicts_and_lists`? Or perhaps `_deepcopy_mutable_fields`?

testing/src/scenario/state.py
```diff
@@ -212,20 +234,22 @@ class JujuLogLine:
 
 
 @dataclasses.dataclass(frozen=True)
+@_detach_dicts_and_lists
```
benhoyt, line 237: Not sure, but what do you think about just adding `__post_init__` on these explicitly? It's 2 lines (plus a blank line) instead of 1, but perhaps clearer (and we wouldn't need to call the original `__post_init__`, as there isn't one, or we'd be adding to the existing one):

```python
def __post_init__(self):
    _deepcopy_mutable_fields(self)
```

testing/src/scenario/state.py
```diff
@@ -1502,6 +1534,24 @@ def __post_init__(self):
             if not isinstance(val, frozenset):
                 object.__setattr__(self, name, frozenset(val))
 
+    def __deepcopy__(self, memo: dict[int, Any] | None = None) -> State:
```
benhoyt, line 1537: Can you add a comment explaining why this class needs a `__deepcopy__` (when others don't)?

benhoyt:
Maybe it *is* time to revisit Pietro's base class for the state dataclasses... but maybe call it `MarvelBase` instead of `DCBase`? ;-) Seriously though: I was the one who pushed pretty hard to remove it, but I'm open to revisiting that if this `__post_init__` logic really is common to all State classes. I just think we need a better base class name, maybe just `StateBase`. Perhaps discuss after daily sync today?

testing/src/scenario/state.py
```diff
@@ -1502,6 +1539,31 @@ def __post_init__(self):
...
+        new_state = copy.copy(self)
+        for attr in (
+            "relations",
+            "containers",
+            "storages",
+            "networks",
+            "opened_ports",
+            "secrets",
+            "resources",
+            "stored_states",
```
james-garner-canonical, line 1559: Suggestion: move "networks" between "relations" and "containers" so the order matches the order in the class definition.

Maybe move the "config" copy above this loop for the same reason.

Should "deferred" be copied too?

james-garner-canonical (approved):

testing/src/scenario/state.py
james-garner-canonical: Makes sense. It not being copied made me wonder if it was a shared queue or something scary like that.

PR #1621, opened 2025-03-10
chore: bump doc dependencies

tonyandrewmeyer:
There's a security alert for Jinja 3.1.5. As far as I can tell, it's not relevant to our use (which is only in doc generation anyway), but it's nice to move to the fixed version even if just to silence the warning systems.

While we're doing this, bump the other doc dependencies (this is just generating a fresh `requirements.txt` via `uvx --python=3.11 tox -e docs-deps`).

I haven't read through all of the changes in each of the bumps, but for the ones that aren't micro bumps, certifi seems safe (and that's calendar versioning, not semver), sphinx-notfound-page seems safe, the starlette changes don't seem relevant, and the websockets changes don't seem relevant to docs (it's less clear for ops itself).

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -10,9 +10,9 @@ anyio==4.8.0
     # via
     #   starlette
     #   watchfiles
-babel==2.16.0
+babel==2.17.0
     # via sphinx
-beautifulsoup4==4.12.3
+beautifulsoup4==4.13.3
     # via
     #   canonical-sphinx-extensions
     #   furo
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
Strange, the docs build failed when installing/upgrading pip and setuptools. Will look into that.

IronCore864 (approved):

tonyandrewmeyer:
> Strange, the docs build failed when installing/upgrading pip and setuptools. Will look into that.

Seems like this is some sort of PyPI issue - it's happening all over the place.

james-garner-canonical (approved):

PR #1624, opened 2025-03-11
feat: add a remove_revision() method to SecretRemoveEvent and SecretExpiredEvent

### LINKED ISSUE #1309: Wrap event.secret.remove_revision(event.revision) into a simpler API
When handling the `secret-remove` event, all charms should have code like:

```python
event.secret.remove_revision(event.revision)
```

I don't think charms should be working directly with secret revisions at all. A way to move towards that would be to provide some syntactic sugar for the above, which can be done as a simple backwards-compatible improvement. Perhaps:

```python
event.remove_revision()
```

tonyandrewmeyer:
In general, charms should treat secret revisions as an internal Juju detail - they only need to work with the tracked and latest versions, and don't need to know what numbers those happen to be.

To encourage this behaviour, this PR adds a `remove_revision` method to the secret-remove and secret-expired events, where charms are expected to call `secret-remove`. The method needs no arguments because the event already has the secret and revision in question.

Fixes #1309

### DIFF
--- a/docs/howto/manage-secrets.md
+++ b/docs/howto/manage-secrets.md
@@ -177,7 +177,7 @@ class MyDatabaseCharm(ops.CharmBase):
 
     def _on_secret_remove(self, event: ops.SecretRemoveEvent):
         # All observers are done with this revision, remove it:
-        event.secret.remove_revision(event.revision)
+        event.remove_revision()
 ```
 
 ### Revoke a secret
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -1012,7 +1012,7 @@ class SecretRemoveEvent(SecretEvent):
     inform the secret owner that the old revision can be removed.
 
     After any required cleanup, the charm should call
-    :meth:`event.secret.remove_revision() <ops.Secret.remove_revision>` to
+    :meth:`event.remove_revision() <ops.SecretRemoveEvent.remove_revision>` to
     remove the now-unused revision. If the charm does not, then the event will
     be emitted again, when further revisions are ready for removal.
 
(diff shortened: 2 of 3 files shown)


IronCore864 (approved):

dimaqq:
High-level comment: I wish we could provide some advice to reconciler charmers.

Would that be smth like "observe this event, call .remove_revision(), then call reconcile()"?

tonyandrewmeyer:
> High-level comment: I wish we could provide some advice to reconciler charmers.
> 
> Would that be smth like "observe this event, call .remove_revision(), then call reconcile()"?

For `secret-remove`, handling should be entirely outside of the holistic approach. The charm should be observing `secret-changed` using their holistic handler that does the big reconcile. Once that's complete, so the secret is no longer used anywhere, then `secret-remove` is emitted, and at that point there should be no need to do a further reconcile. In some situations you might want to do some tidying up (like removing credentials locally - not *use* of the credentials, but the credential itself) but hopefully you're just calling `remove_revision()`.

`secret-expired` is more complicated. We should probably have a whole section in the secrets how-to on using secret expiry. The simple case, which I think is maybe the only one that's actually in use, is that the expiration is set externally (like you got given a certificate that expires at some point). I think you'd want to handle that in your collect status handler (so not in a reconcile holistic handler, but still somewhat holistically) putting in warnings about the expiry coming up with instructions about how to provide new content when the deadline is close (probably also logging that so it shows up in Loki or similar), and then after expiry putting something about degraded status in there.

If the charm is responsible for getting a replacement secret itself, then presumably you want to do that *before* it's expired. So your reconcile/holistic code should always be checking if the expiry is "soon" and when it is, getting a new secret and updating to that revision. Then you get a `secret-remove` event *before* the expiry date, and you never actually deal with `secret-expired` (unless your method of getting new secret content continually fails, in which case you're back to the collect status approach above).

dimaqq:
At the very least let's create doc issues for these ☝🏻

dimaqq (approved):

PR #1627, opened 2025-03-13
fix: assorted fixes for Pebble layer merging in Harness and Scenario

tonyandrewmeyer:
* When merging `pebble.Check`s, an unset `startup` should not override a set `startup`. (Once set you can't go back to unset - you need to explicitly use enabled).
* When rendering services, checks, and log targets from a series of layers, Harness and Scenario were combining after sorting the items by name. Although Pebble uses filename sorting when reading initial layers from a directory, future layers are combined in the order they are added via the API. For Harness and Scenario, we need to use the order in which they were added - if the caller wants to simulate loading from the filesystem, they need to do that by calling `add_layer` in that order. The plan itself still sorts the services, checks, and log targets.
* When rendering services, checks, and log targets, if there are multiple services/checks/log-targets with the same name, and `override` is set to `merge`, the service/check/log-target needs to merge rather than replace. This was already done when adding layers with the same name (in `add_layer`) but not when there were differently named layers.
* The Scenario `Container.plan` property was ignoring any defined checks and log targets. Those are now included.
* The Scenario `Container` service, check, and log target rendering also needed the (no) sorting and merging fixes from above.
* When adding a layer in Harness or Scenario, update any existing `CheckInfo`s to match changes to the plan - in real Pebble the startup, level, and threshold fields are always pulled from the plan but we bake them in when testing, so need to make sure they continually track what's in the combined plan.

Also adds a `__repr__` for `pebble.Plan` - this helped while debugging issues, and it seems reasonable to keep it.

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3319,7 +3319,7 @@ class _TestingPebbleClient:
         raise NotImplementedError(self.wait_change)
 
     def _update_check_infos_from_plan(self):
-        # In testing, the check info is has the level, threshold, and startup
+        # In testing, the check info has the level, threshold, and startup
         # from the check baked in, so we need to update the info when the plan
         # changes.
         for check in self.get_plan().checks.values():
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3291,6 +3291,18 @@ class _TestingPebbleClient:
     ) -> pebble.Change:
         raise NotImplementedError(self.wait_change)
 
+    def _update_check_infos_from_plan(self):
+        # In testing, the check info is has the level, threshold, and startup
+        # from the check baked in, so we need to update the info when the plan
+        # changes.
+        for check in self.get_plan().checks.values():
+            info = self._check_infos.get(check.name)
+            if not info:
+                continue
+            info.level = check.level
+            info.threshold = 3 if check.threshold is None else check.threshold
+            info.startup = check.startup
+
     def add_layer(
         self,
         label: str,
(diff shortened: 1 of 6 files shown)


dimaqq:
LGTM. May I assume that these semantics were validated against real pebble?

tonyandrewmeyer:
> May I assume that these semantics were validated against real pebble?

Yes. Sorry, I should have said that.

ops/_private/harness.py
```diff
@@ -3318,6 +3318,18 @@ def wait_change(
     ) -> pebble.Change:
         raise NotImplementedError(self.wait_change)
 
+    def _update_check_infos_from_plan(self):
+        # In testing, the check info is has the level, threshold, and startup
```
benhoyt, line 3322:
```suggestion
        # In testing, the check info has the level, threshold, and startup
```

ops/_private/harness.py
```diff
@@ -3421,39 +3433,52 @@ def add_layer(
                 info = pebble.CheckInfo(
                     name,
                     level=check.level,
+                    startup=check.startup,
                     status=status,
+                    threshold=3 if check.threshold is None else check.threshold,
```
benhoyt, line 3438: Can we make this magic number a named (non-public) constant, as it's used in a couple of places now?

ops/pebble.py
```diff
@@ -1210,14 +1213,19 @@ def _merge(self, other: Check):
         attributes take precedence.
         """
         for name, value in other.__dict__.items():
-            if not value or name == 'name':
+            if value is None or value == '' or name == 'name':
```
benhoyt, line 1216: I was just about to suggest `if not value or name == 'name'` as a simpler way of capturing both None and empty string, and then I realised that the previous version was exactly that. Out of interest, why this change?

test/test_testing.py
```diff
@@ -7119,6 +7119,180 @@ def test_stop_then_start(self, request: pytest.FixtureRequest):
...
 
+    @pytest.mark.parametrize(
+        'combine,new_layer_name',
+        [
+            (False, 'new-layer'),
+            (True, 'base'),
+            (
+                True,
+                'new-layer',
+            ),  # This doesn't have anything to combine with, but for completeness.
```
benhoyt, line 7130: The formatting here is awkward -- took me a moment to figure that this was just another tuple, not nested. What about this?
```suggestion
            # This doesn't have anything to combine with, but for completeness:
            (True, 'new-layer'),
```

testing/src/scenario/state.py
```diff
@@ -1046,12 +1052,36 @@ def __post_init__(self):
...
             for name, service in layer.services.items():
-                services[name] = service
+                if name in services and service.override == "merge":
+                    services[name]._merge(service)
+                else:
+                    services[name] = service
         return services
 
+    def _render_checks(self) -> Dict[str, pebble.Check]:
+        # copied over from ops.testing._TestingPebbleClient._render_checks()
```
benhoyt, line 1064: Too tricky to reuse the same/similar code from `_private/harness.py`?

benhoyt (approved):
Looks good to me -- approving. Few minor/comments and questions, but happy for you to address as you see fit and merge.

dwilding (approved):
Thanks @tonyandrewmeyer, with these changes my unit test is doing the expected thing!

IronCore864 (approved):

ops/pebble.py
tonyandrewmeyer: It was originally because I went down a wrong path when figuring out what was going wrong (I thought false-y values were being skipped when they should get merged because they were not "no value").

I don't love that `period` and `timeout` can be both `None` and `""`, and they seem to mean the same thing: not present in the layer. `""` seems Go'ish and `None` seems Pythonic and we're allowing either for some reason.

Of the other values, `threshold` can be `0` and will be `None` when not in the layer, and those are distinct values. Pebble accepts a value of `0` for threshold (and the docs don't say it's not acceptable, though they do for other fields). So the replacement line was meant to fix that corner case, even though I'm not sure what a `0` value for `threshold` would actually mean.

However, while I did check that Pebble accepts the value, I didn't check this one case of merging, and it's a bit odd.

```yaml
# zero.yaml
checks:
  chk1:
    override: replace
    threshold: 1
    exec:
      command: /bin/true
```

```yaml
# zero2.yaml
checks:
  chk1:
    override: merge
    threshold: 0
```

Adding a second layer with `0` leaves it unchanged:

```
$ go run ./cmd/pebble add base zero.yaml
Layer "base" added successfully from "zero.yaml"
$ go run ./cmd/pebble add base2 zero2.yaml
Layer "base2" added successfully from "zero2.yaml"
$ go run ./cmd/pebble plan
checks:
    chk1:
        override: replace
        threshold: 1
        exec:
            command: /bin/true
```

Adding that again with `--combine` resets it to the default (I somewhat see how this works, but it's a bit unexpected, particularly that it doesn't happen if done to the first layer).

```
$ go run ./cmd/pebble add base2 zero2.yaml --combine
Layer "base2" added successfully from "zero2.yaml"
$ go run ./cmd/pebble plan
checks:
    chk1:
        override: replace
        threshold: 3
        exec:
            command: /bin/true
```

So `threshold` *should* be ignored when `0` ... except in some cases :laughing:. I feel like disallowing a `0` threshold in Pebble is perhaps the most correct fix. For this code, I'll revert it back to what it was and add a comment.

testing/src/scenario/state.py
tonyandrewmeyer: "copied over" is a little imprecise (I borrowed the wording from the `render_services` above). One uses `self.checks` (and so on) and the other `self._checks` (and so on). I could pass in the attribute name, I guess.

I'm reluctant to make Scenario *more* dependant on Harness, since the aim is that eventually Harness breaks off into `ops-harness` and `ops-scenario` depends only on `ops` and not that new package. On the other hand, there's plenty of dependency at the moment, so it wouldn't be making things significantly worse.

I'll have a little play with it.

benhoyt: Ah got it -- in that case, I think it's fine as is. "A little copying is better than a little dependency" :-)

Issue #1427, opened 2024-10-10
Add support for testing pebble services interaction


tonyandrewmeyer:
There is currently no way in ops-testing to test a charm interaction with a pebble service, like restarting it for example: `self._container.restart(self._service_name)`.

The only way to accomplish this is to mock `ops.model.Container.restart`.

It would be nice to have a better way to do this, maybe as an automatic mock in the output state?

```
...

state_out = self.ctx.run(container.pebble_ready_event, state_in)

assert state_out.containers[0].services["ausf"].restarted
```

Moved from https://github.com/canonical/ops-scenario/issues/34

benhoyt:
Let's have someone spend 1-2h before end of Sept to see if we can determine from charm unit tests whether this is "missing" or how they're doing it now. @tonyandrewmeyer to do this.

@dimaqq Also to provide his higher-level idea.

sinapah:
I also think this would be very helpful as it may be valuable to know whether a restart has happened. A specific use case  would be when you are asserting that an event results in a restart of a particular service. Somewhat related would be to be able to get an equivalent of `pebble changes` so you can see all the replans and restarts that took place.

tonyandrewmeyer:
Thanks @sinapah ! I think that's a fairly indicative example of where things are now. You can test to a certain level, if you're willing to assume things like "replan means services go to active" but you can't test more complex cases.

From what I've found, mostly charms just don't test this right now (or test things via integration tests instead).

We don't have space for this in 26.04 at the moment, but I think it's worth keeping this open for when we do have time.

dimaqq:
From memory, charmers come to this wanting to test code like:
- on specific event, we know what things have changed
- new plan is generated
- service is explicitly restarted
- let's assert that it was restarted

However, a reconciler charm would behave more like:
- on any event
- plan is generated
- plan is applied
- container-side pebble decides if service is actually restarted

The earlier is more amenable to testing, but is it the charm style that we want to encourage?

The latter (I think) is already testable via "assert applied_plan == expected_plan".
Testing that deeper would require a custom workload mock, which would need API design:
- given env var X in the plan, health checks are expected to pass
- in some corner case, health checks are expected to fail
- if service A (of N available) gets restarted as part of replan, expect a custom notice later

PR #1630, opened 2025-03-13
feat: check that the check-infos in testing.Container match the plan

tonyandrewmeyer:
Add some consistency checks for the `CheckInfo`s in `testing.Container`:

* If there is a `CheckInfo`, then the plan must have a check by that name, or Pebble would never have generated that info.
* The threshold, startup, and level values of the `CheckInfo` come from the check definition itself, so must match the plan.

### DIFF
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -685,14 +685,12 @@ def check_containers_consistency(
                 )
                 continue
             plan_check = plan.checks[check.name]
-            for attr_from_plan in ("level", "startup", "threshold"):
-                if getattr(check, attr_from_plan) != getattr(
-                    plan_check, attr_from_plan
-                ):
+            for attr in ("level", "startup", "threshold"):
+                if getattr(check, attr) != getattr(plan_check, attr):
                     errors.append(
                         f"container {container.name!r} has a check {check.name!r} with a "
-                        f"different {attr_from_plan!r} ({getattr(check, attr_from_plan)}) "
-                        f"than the plan ({getattr(plan_check, attr_from_plan)}).",
+                        f"different {attr!r} ({getattr(check, attr)}) "
+                        f"than the plan ({getattr(plan_check, attr)}).",
                     )
 
     return Results(errors, [])
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -670,7 +670,7 @@ def check_containers_consistency(
             f"Missing from metadata: {diff}.",
         )
 
-    # If you have check-infos, then they must match the plan.
+    # If you have check-infos, then they must match the computed plan.
     for container in state.containers:
         plan = container.plan
         for check in container.check_infos:
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -683,21 +683,16 @@ def check_containers_consistency(
                     f"container {container.name!r} has a check {check.name!r} "
                     f"but the {plan_has}.",
                 )
-            else:
-                if check.level != plan.checks[check].level:
+                continue
+            plan_check = plan.checks[check.name]
+            for attr_from_plan in ("level", "startup", "threshold"):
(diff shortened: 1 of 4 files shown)


testing/tests/test_consistency_checker.py
```diff
@@ -85,46 +85,40 @@ def test_workload_event_without_container():
...
-    )
-    assert_inconsistent(
-        State(containers={Container("foo", check_infos={check}), Container("bar")}),
-        _Event(
-            "bar-pebble-check-recovered", container=Container("bar"), check_info=check
-        ),
-        _CharmSpec(MyCharm, {"containers": {"foo": {}, "bar": {}}}),
+
+
+def test_check_info_container_matches_event():
```
tonyandrewmeyer, line 90: These are the same as above but:
* They add they layer, which they probably should have been doing before, and now have to.
* They're better labelled, so it's easier to see what inconsistency is being checked with each one.
* They aren't in the "no container" test, which they shouldn't have been.
* They're done in a loop rather than repeating for recovered and failed events.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -481,8 +481,18 @@ def _on_check_failed(self, event):
             infos.append(event.info)
 
     ctx = Context(MyCharm, meta={"name": "foo", "containers": {"foo": {}}})
-    check = CheckInfo("http-check", failures=7, status=pebble.CheckStatus.DOWN)
-    container = Container("foo", check_infos={check})
+    layer = pebble.Layer({
```
tonyandrewmeyer, line 484: All the tests in this file are unchanged, except that they add in the layer that really ought to have already been there.

testing/src/scenario/_consistency_checker.py
```diff
@@ -670,6 +670,31 @@ def check_containers_consistency(
...
+                    f"container {container.name!r} has a check {check.name!r} "
+                    f"but the {plan_has}.",
+                )
+                continue
+            plan_check = plan.checks[check.name]
+            for attr_from_plan in ("level", "startup", "threshold"):
+                if getattr(check, attr_from_plan) != getattr(
+                    plan_check, attr_from_plan
+                ):
+                    errors.append(
```
dimaqq, line 692: I guess that's the common case.

I do wonder though, what if someone tests charm receiving check infos while the plan is being changed, is it hypothetically possible to get info from the previous plan version?

dimaqq: Aside... do we support `Mock()` for these leaf fields?
I could imagine someone not bothering with exact field values in their tests.

dimaqq (approved):

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: > I do wonder though, what if someone tests charm receiving check infos while the plan is being changed, is it hypothetically possible to get info from the previous plan version?

We don't support using Scenario to test this. Right now, the expectation is that there is a static input state, and that state includes Pebble. It doesn't have much detail, but expanding this sort of functionality is what https://github.com/canonical/operator/issues/1427 Is intended for.

testing/src/scenario/_consistency_checker.py
```diff
@@ -670,6 +670,31 @@ def check_containers_consistency(
...
+                errors.append(
+                    f"container {container.name!r} has a check {check.name!r} "
+                    f"but the {plan_has}.",
+                )
+                continue
+            plan_check = plan.checks[check.name]
+            for attr_from_plan in ("level", "startup", "threshold"):
+                if getattr(check, attr_from_plan) != getattr(
+                    plan_check, attr_from_plan
+                ):
```
benhoyt, line 691: Nit: This `if` statement doesn't format nicely -- the getattrs are unbalanced, making it harder to read. I think it'd be fixed if you just called this local var `attr` (which seems enough here?). Alternatively, something like:

```python
check_value = getattr(...)
plan_value = getattr(...)
if check_value != plan_value:
    ...
```

benhoyt (approved):

PR #1631, opened 2025-03-14
feat: run deferred events with fresh charm instances

### LINKED ISSUE #1174: Run deferred Juju events on a fresh charm instance
The main benefit here would be that it would be simpler to reason about (Juju) event behaviour, because it would be consistent both when an event was deferred and when it is executed normally.

This would not apply to custom events, which are executed synchronously, and are already different from Juju events. This would also not apply to `LifecycleEvents` (framework generated events), because those cannot be deferred.

We will need to use the same charm instance for multiple deferred (event, handler) pairs when more than one handler defers the event. If something like this happens:

* original Juju event: observer A defers, observer B does not
* next Juju event of same type: deferred observer A defers, observer A defers, observer B defers

I don't think we can tell that these were different Juju events, and we would maybe collapse the duplicates (?) and probably have to use the same charm instance for both observers. We would otherwise need to add some sort of ID to the deferred event to be able to group them correctly.

I believe the cleanest change here is to lift out the deferred `reemit` from `_Manager` in [ops/main.py], and handle the deferred events in `main()`, creating new `_Managers` for each one (we need to check that `setup_root_logging` can be called twice, and avoid calling legacy hooks twice - probably the legacy hook should be moved out of `_Manager` too). This provides a fresh `Model`, `Framework` and `_Dispatcher` as well, which is also most consistent (and avoids having to have the framework forget/unobserve the old charm).

There would be some small behaviour changes here: the pre-commit and commit events would get emitted twice (just as if the event had not been deferred, but not like now), which I think is fine, and we would collect the status twice (again just as if the event had not been deferred, but not like now), which might break some expectations.

See also the discussion in #736. Note that we will not be able to make Harness mimic this behaviour (because it would break too many existing charm tests) but we will be encouraging using Scenario, which can handle it (right now Scenario copies the ops behaviour, saving the deferred events to a queue that ops then runs ahead of the Juju event, but the Scenario API would not need changing in order to match the new ops behaviour).

tonyandrewmeyer:
Create a new charm object for each emission of a deferred event, and for the event that triggered the ops process.

When using the `testing.Context` as a context manager, the charm that is made available is the one for the primary event. If there is demand in the future, we could add a mechanism for accessing the other charm objects for manipulation prior to running the deferred events, or inspection after.

This will break charms that are relying on the charm object to be the same object for deferred and the primary event. However, we have always told charmers to not do that, and it's risky behaviour because it'll differ based on whether it's the non-deferred or deferred handler call, so we're ok with this change.

The other significant change for charms is that, because we need to commit the framework before finishing up with each charm, the `pre-commit` and `commit` Lifecycle events are emitted after each deferred event as well as after the primary event. I'm only aware of two charms that use these:
* ops-sunbeam, which uses `commit` to put statuses in stored state. Running more than once should be fine, and the performance impact not noticeable.
* `statustest` (of @benhoyt's test-charms collection), which seems almost identical to the sunbeam one. I assume this isn't really used, and it wouldn't break anyway.

There are three main parts of this implementation:
* Fairly minor adjustments to `_main`, particularly `_Manager`, so that the charm is passed around and the framework is accessed via the charm, rather than creating the framework and charm once, and the actual creation of each charm prior to re-emitting.
* Functionality to "forget" charms. The framework registers event types on the class, not the instance, so creating multiple charm instances, even on different `Framework`s, is not normally possible (Scenario gets around this by wrapping the event source in a new class for each run). `CharmBase` gains a `_destory_charm` function that undoes all of the event registration done in `__init__` (for all the dynamic events from relations, containers, and so forth). I considered using `__del__` for this, but it's awkward to cleanly remove everything when destruction is happening, and we don't really want charmers to start creating more than one charm instance, so this leaves the functionality private.
* Reworking the Scenario "capture events" mechanism, to use a `Framework` subclass, rather than a context manager. This needed changes to handle capturing deferred events, but it also fixes a bug. Previously, the `Framework` *class* was patched, with methods that used local variables to determine which events would be captured: this meant that the `capture_events()` context manager was not threadsafe.

Fixes #1174


ops/_main.py
```diff
@@ -364,31 +364,35 @@ class _Manager:
...
         self,
         charm_class: Type['_charm.CharmBase'],
-        model_backend: Optional[_model._ModelBackend] = None,
+        juju_context: _JujuContext,
         use_juju_for_storage: Optional[bool] = None,
         charm_state_path: str = CHARM_STATE_FILE,
-        juju_context: Optional[_JujuContext] = None,
     ):
-        if juju_context is None:
-            juju_context = _JujuContext.from_dict(os.environ)
```
tonyandrewmeyer, line 385: Ops and Scenario are the only ones creating `_Manager` objects, and now that we're forcing alignment of versions of `ops` and `ops-scenario`, this compatibility code can be removed to simplify the class.

testing/tests/test_e2e/test_event.py
```diff
@@ -90,9 +90,10 @@ def _foo(self, e):
         ctx.on.start(), State(deferred=[_Event("update-status").deferred(MyCharm._foo)])
     )
 
-    assert len(ctx.emitted_events) == 5
```
tonyandrewmeyer, line 93: Right after this we check for the exact contents of the list, so there doesn't seem to be any point checking the length first.

testing/src/scenario/_ops_main_mock.py
```diff
@@ -98,7 +107,7 @@ def apply_state(self, state: "State"):
             db.save_snapshot(stored_state._handle_path, stored_state.content)
 
 
-class Ops(_Manager):
+class Ops(_Manager, Generic[CharmType]):
```
tonyandrewmeyer, line 110: This fixes a complaint in `_runtime.py`.

tonyandrewmeyer:
@benhoyt would you mind giving this a 'pre-review' - no need to go over it line-by-line, but check whether the general approach seems reasonable?

testing/src/scenario/state.py
```diff
@@ -2055,7 +2055,11 @@ def _is_builtin_event(self, charm_spec: _CharmSpec[CharmType]) -> bool:
         return self._path.type is not _EventType.CUSTOM
 
     def deferred(self, handler: Callable[..., Any], event_id: int = 1) -> DeferredEvent:
-        """Construct a DeferredEvent from this Event."""
+        """Construct a DeferredEvent from this Event.
+
+        The charm class passed to the :class:`Context` must be observing the
+        event this DeferredEvent is based on.
```
tonyandrewmeyer, line 2061: I believe this is a new requirement (a test that did not do this started failing after the changes). However, I feel that it should always have been the case, and if charms are dynamically observing events, then they need to have static versions of those for the unit tests.

Without this restriction, the type doesn't get registered.

ops/framework.py
```diff
@@ -475,6 +475,23 @@ def define_event(cls, event_kind: str, event_type: 'Type[EventBase]'):
         event_descriptor.__set_name__(cls, event_kind)
         setattr(cls, event_kind, event_descriptor)
 
+    @classmethod
+    def undefine_event(cls, event_kind: str):
```
benhoyt, line 479: Can we make this non-public?

Actually, I wonder -- instead of undefining/forgetting, which seems error-prone, can we instead change Ops to have the event definitions go on the instance rather than the class? What we be involved in that?

benhoyt (commented):
One comment about whether we can restructure to avoid the "forgetting" (which seems error-prone to me).

Regarding this:

> This will break charms that are relying on the charm object to be the same object for deferred and the primary event. However, we have always told charmers to not do that, and it's risky behaviour because it'll differ based on whether it's the non-deferred or deferred handler call, so we're ok with this change.

I agree. But can we get any more verification of this? Perhaps (once we've finalised the approach) run several charms' integration tests to get a better idea.

Can we check with the ops-sunbeam folks about the `commit` change, and make sure the original authors of that approach seem okay with this?

The `reemit` change is an optional argument so shouldn't break even those existing users, right?

ops/framework.py
benhoyt: If we could achieve this, it would get rid of the wrapping hacks in Scenario too, right?

tonyandrewmeyer:
> > This will break charms that are relying on the charm object to be the same object for deferred and the primary event. However, we have always told charmers to not do that, and it's risky behaviour because it'll differ based on whether it's the non-deferred or deferred handler call, so we're ok with this change.
> 
> I agree. But can we get any more verification of this? Perhaps (once we've finalised the approach) run several charms' integration tests to get a better idea.

I can try that, yes. The trouble I've had previously is that the most common use of `defer` is in the data platform charms, and it's challenging to run their integration tests with a custom branch of ops.
 
> Can we check with the ops-sunbeam folks about the `commit` change, and make sure the original authors of that approach seem okay with this?

Sure/
 
> The `reemit` change is an optional argument so shouldn't break even those existing users, right?

Yes, it's entirely backwards compatible. I mentioned it because it means we support that going forward, and it's not strictly necessary (since `_Manager` could just use a private `Framework` method).

benhoyt:
> I can try that, yes. The trouble I've had previously is that the most common use of defer is in the data platform charms, and it's challenging to run their integration tests with a custom branch of ops.

Well, let's try it on some non-data charms before merge, and then a couple of data platform ones after merge but before release. That would be easier, right? We can always revert or fix before release.

Regarding `reemit` and supporting that new arg. That's a good point. I think I'd prefer to use the private method in that case, rather than exporting an new arg that others won't / probably shouldn't use.

ops/framework.py
tonyandrewmeyer: > Can we make this non-public?

Yes, it's only called by the non-public `CharmBase` method. `unregister_type`, too.

> Actually, I wonder -- instead of undefining/forgetting, which seems error-prone, can we instead change Ops to have the event definitions go on the instance rather than the class? What we be involved in that?

I wondered about that. It seemed like a quite significant change, although it would make everything simpler. It's hard to know how this would impact charms -- I would hope that no-one is relying on that behaviour, but I'm also not sure how to rigourously check for it. I'm not sure if unit tests would show issues, and I don't have a good solution for running integration tests in bulk yet.

I believe the motivation for making it a class event was that the handle path is something like `MyCharm/on` and if you have multiple instances they have the same path, and so you can't tell which one the path is referring to. The PR (as is) does break this a bit, in that there are multiple instances using the same path, but the code makes sure that there aren't multiple instances at the same time, so there's still only one observer it could be.

We could change the handle path to identify instances, but that would have to survive across hooks somehow, so probably mean providing a per-instance key that doesn't change, and it seems like this ends up not really helping with the change to how deferring is handled.

If we stored the `EventSource` on the `ObjectEvents` instance, I think charms would not really notice, because everything is getting stored on the `CharmEvents` instance that is a class attribute of `CharmBase`, so it's still class based. Something like `CharmEvents.on.peer_relation_changed` wouldn't work any more, but that's probably ok?

It's probably libraries where there's more risk, because they are creating new `ObjectEvents` outside of `CharmEvents`. As far as I can recall, none of them are adding event sources inside of instance methods (but I'd have to look again to be sure), so they wouldn't change much, although you would be able to do things you couldn't before, and I think that could include things that break (multiple `ObjectEvents` within a class, maybe?).

TL;DR: I'm confident that it could work for the straightforward cases, and it would be simpler code. I'm not confident that I'd be able to predict how it would impact existing charms and what kind of mistakes or patterns it would enable in the future.

> If we could achieve this, it would get rid of the wrapping hacks in Scenario too, right?

Yes. Scenario would also be able to use the undefine/unregister methods that this PR introduces if we kept things that way. It seemed like it was simpler to avoid doing that in the same PR.

benhoyt: Okay, thanks. In that case I believe it's better to stick with the approach you've got.

ops/_main.py
```diff
@@ -492,7 +492,10 @@ def _make_storage(self):
             store = _storage.SQLiteStorage(charm_state_path)
         return store
 
-    def _make_framework(self, event_name: str):
+    def _make_framework(self, *args: Any, **kwargs: Any):
```
benhoyt, line 495: Nit: It seems to me this method's not worth it anymore, plus it just reduces type safety.

ops/_main.py
```diff
@@ -471,87 +504,106 @@ def _make_framework(self, dispatcher: _Dispatcher):
...
         """Commit the framework and gracefully teardown."""
-        self.framework.commit()
+        framework.commit()
 
     def _close(self):
         """Perform any necessary cleanup before the framework is closed."""
         # Provided for child classes - nothing needs to be done in the base.
 
+    def run_deferred(self):
+        """Emit and then commit the framework.
```
benhoyt, line 573: This docstring summary line doesn't seem to match the function name.

benhoyt (approved):
Looks good to me now, thanks

ops/_main.py
tonyandrewmeyer: I've added a comment explaining why it's there: it's needed because the Scenario `_Manager` subclass (`_Ops`) needs to provide a different framework class (I feel this is tidier than having Scenario monkeypatch new methods in).

I could have a class/instance variable for the class to use, but then they'd need to have the same signature. The Scenario one currently uses the `Context`, for knowing which events to 'capture' and to put the captured events into the context. I could have the class capture all events and then filter them afterwards, and store them on the `CapturingFramework` instance and pull them from there after the events. Or `CapturingFramework` could have something like a `set_context()` method that could be used to add the context after the framework is instantiated but before the `run`. Would you prefer either of those?

If it stays the same, I could bring back the specific args (reducing type safety is definitely a valid point). I can just have `_make_framework` take all the framework arguments and pass those through.

By the way, I do see increasing customisation of `_Ops` vs. `_Manager`. I wonder if at some point it's better to simplify `_Manager` so that it's not very customisable via subclassing and have `_Ops` implement the entire functionality itself (the same purpose and maybe even a similar API, but entirely separate code).

benhoyt: Nah, this is fine for now -- thanks for the comment.

ops/_main.py
```diff
@@ -49,27 +49,27 @@ def _exe_path(path: Path) -> Optional[Path]:
 
 
 def _create_event_link(
-    charm: '_charm.CharmBase',
+    charm_dir: Path,
```
dimaqq, line 52: Minor and probably a separate PR: we're supposed to express this as `pathlib.Path`, aren't we?

dimaqq (commented):
About multiple lifecycle events: what happens if the collected status values are inconsistent?

dimaqq:
Or did I perhaps misunderstand, and it's pre-commit and commit events are multiplied, but collect app/unit status events are not?

test/test_main.py
```diff
@@ -233,7 +233,7 @@ def __init__(self, framework: ops.Framework):
                     ops.main(MyCharm)
 
         assert mock_charm_event.call_count == 1
-        return mock_charm_event.call_args[0][0]
+        return mock_charm_event.call_args[0][1]
```
dimaqq, line 236: Can you explain this change?

ops/charm.py
```diff
@@ -1376,6 +1376,31 @@ def __init__(self, framework: Framework):
...
             )
 
+    def _destroy_charm(self):
+        for relation_name in self.framework.meta.relations:
+            relation_name = relation_name.replace('-', '_')
+            self.on._undefine_event(f'{relation_name}_relation_created')
+            self.on._undefine_event(f'{relation_name}_relation_joined')
+            self.on._undefine_event(f'{relation_name}_relation_changed')
+            self.on._undefine_event(f'{relation_name}_relation_departed')
+            self.on._undefine_event(f'{relation_name}_relation_broken')
```
dimaqq, line 1386: Somehow I'm not too happy with this.
Is there a better way?

For example, remembering the events that were defined?
A context manager?
Or removing events by name or pattern matching?
Or removing all events except the hardcoded set / allow list?
Or simply resetting the whole `self.on`?

dimaqq (commented):
Reviewed about 50% so far.

ops/_main.py
tonyandrewmeyer: Yes - this was using an existing import, but switching it is fine with me.

test/test_main.py
tonyandrewmeyer: The signature of `_emit_charm_event` in `ops._main` has changed from: `_emit_charm_event(self, event_name: str)` to `_emit_charm_event(self, charm: _charm.CharmBase, event_name: str)`. That means that the name is now `[1]` rather than `[0]`.

I could switch the arguments around so that `event_name` is still first, but it seemed more consistent to have the charm first, and also more logical in that the name is "emit charm event" so "charm then event" as the argument order aligns.

tonyandrewmeyer:
> About multiple lifecycle events: what happens if the collected status values are inconsistent?
> Or did I perhaps misunderstand, and it's pre-commit and commit events are multiplied, but collect app/unit status events are not?

Yes, you get a pre-commit and commit after every event, but there's still just one collect-status run. If there was multiple collect-status emits then you'd just get the status after the first event until it was replaced with a status set from the next event - but it seems unlikely that there would be changes here, so I think it's better to take the performance win and only do that once.

For pre-commit and commit, mostly they're unused, but since we need to commit the framework it seems like we have to emit those events as well.

Possibly the confusion comes from the Sunbeam charms using the commit event to do status management.

ops/charm.py
tonyandrewmeyer: > For example, remembering the events that were defined?

The events that were defined are the ones in the meta list - it seems to me that also putting them in some other list would just be duplication.

> A context manager?

The defining is done in the `CharmBase` `__init__`, and I don't think we can change that and keep backwards compatibility. So the defining couldn't be in an `__enter__`. I guess main could do something like:

```python
            with self._make_charm(event_handle.kind) as charm:
                charm.framework._reemit_single_path(event_path)
                self._commit(charm.framework)
                self._close()
```

And the `__exit__` could call `_destroy_charm`. That doesn't seem significantly better to me, though. The calls to `_undefine_event` could go into the `__exit__`, but I think it's better to have them right next to the code that does the defining - for example, if we end up adding a new dynamic event then it's more likely that the person doing that will realise that they need to undefine it as well than if the code is in `_main.py`.

> Or removing events by name or pattern matching?

This is by name. I'd rather be explicit (iterating through the meta list in exactly the same way that `__init__` does) than do something like `for event in self.on.events()` with `if event.kind.endswith('relation_created')` or something like that. 

> Or removing all events except the hardcoded set / allow list?

This is roughly:

```python
for event in self.on.events():
    if event.kind not in ('install', 'start', 'etc'):
        self.on._undefine_event(event)
```

We could probably generate and save the set of events in `__init__`.

To me this feels less clean, because it's doing more than just the inverse of what `__init__` does, but it would have the advantage that if anyone had dynamically added to `on` those would get removed as well (but generally I think we should discourage that).

> Or simply resetting the whole `self.on`?

The problem is that these are defined on the `CharmEvents` **class**, so that means we'd need to have something to produce the class itself, *and* we need to do this in a backwards compatible way. This ends up being much messier (at least in the attempts I made at it).

tonyandrewmeyer: What's your view of this now?

testing/src/scenario/state.py
dimaqq: Do we wipe deferred events on charm upgrade?
I imagine that the unit state db is kept during the upgrade of vm charms.
In which case there's no hard guarantee.

test/test_main.py
dimaqq: thanks

test/charms/test_main/src/charm.py
```diff
@@ -123,6 +129,18 @@ def __init__(self, *args: typing.Any):
...
+            if isinstance(event, ops.LifecycleEvent) or name == 'custom':
+                continue
+            self.framework.observe(event, self._on_any_event)
+        if hasattr(self, 'charm_attribute'):
+            raise RuntimeError('charm instance was reused')
+
+    def _on_any_event(self, _: ops.EventBase):
+        # Note that doing this is bad behaviour: we're doing it here to make
+        # sure that the value is *not* retained.
+        self.charm_attribute = 'config-changed'
```
dimaqq, line 142: How about some variation of this?

```py
class MyCharm(ops.CharmBase):
    _test_event_seen: str | None = None  # sentinel for test only

    def __init__(self, *a):
        ...
        if self._test_event_seen is not None:
            raise RuntimeError('charm instance was reused')

    def _on_any_event(self, event: ops.EventBase):
        # Test sentinel: set to validate that state is not retained across events.
        self._test_event_seen = str(event)
```

test/charms/test_main/src/charm.py
```diff
@@ -26,6 +26,12 @@
 
 logger = logging.getLogger()
 
+# Note for ops developers: juju-log doesn't go anywhere useful much of the time
+# during unit tests, and test_main failures that subprocess out are often
+# difficult to debug. Uncomment this line to get more informative errors when
+# running the tests.
+# logger.addHandler(logging.StreamHandler(sys.stderr))
```
dimaqq, line 33: Thanks for this.
I've found out the hard way while debugging `ops[tracing]` test failures.

ops/framework.py
```diff
@@ -961,6 +998,7 @@ def _reemit(self, single_event_path: Optional[str] = None):
             try:
                 event = self.load_snapshot(event_handle)
             except NoTypeError:
+                logger.debug('Skipping notice %s - cannot find event class.', event_path)
```
dimaqq, line 1001: That's probably OK, given that it should not happen.
At the same time, recall that we (often? typically?) run with debug on, and this line would likely appear in e.g. action log output.

ops/framework.py
```diff
@@ -475,6 +475,23 @@ def define_event(cls, event_kind: str, event_type: 'Type[EventBase]'):
...
 
+    @classmethod
+    def _undefine_event(cls, event_kind: str):
+        """Remove the definition of an event on this type at runtime.
+
+        This undoes the effect of :meth:`define_event`. This is not intended
+        for use by charm authors, but rather for use by the ops library itself.
+        """
+        event_descriptor = getattr(cls, event_kind)
+        if hasattr(event_descriptor, 'framework') and event_descriptor.framework is not None:
```
dimaqq, line 486: Would this be more elegant / easier to read?

```suggestion
        if (framework := getattr(event_descriptor, 'framework', None)) is not None:
```

ops/charm.py
```diff
@@ -1376,6 +1379,12 @@ def __init__(self, framework: Framework):
                 f'{container_name}_pebble_check_recovered', PebbleCheckRecoveredEvent
             )
 
+    def _destroy_charm(self):
+        for event in self.on.events():
+            if event in self._static_events:
+                continue
```
dimaqq, line 1385: `set(...) - set(...)` may be more expressive.

ops/charm.py
```diff
@@ -1350,6 +1351,8 @@ def on(self) -> CharmEvents: ...  # noqa
     def __init__(self, framework: Framework):
         super().__init__(framework, None)
 
+        self._static_events: Set[str] = set(self.on.events())
```
dimaqq, line 1354: Off the top of my head I'm not sure what static means here.

- events like "start" -- sure
- events on relations that we pick up from metadata?
- custom events -- surely not

Perhaps a comment would help me.

dimaqq (approved):

dimaqq:
This branch probably needs main to be merged into, even if no conflicts are detected by GitHub.

testing/src/scenario/state.py
tonyandrewmeyer: > Do we wipe deferred events on charm upgrade?

It depends - on K8s the unit state DB is lost because you get a new container, but on machines it's in-place so they stay.

> In which case there's no hard guarantee.

I still feel that the charm code should handle this situation. But in any case, as per the comment in David's PR that took this fix, it should be handled in a separate issue if anyone does need this functionality. It's not really related to this PR or to that one, just triggered by it, and it's not what the test is intended to check.

testing/tests/test_emitted_events_util.py
```diff
@@ -1,83 +1,76 @@
...
-    def _on_foo(self, e):
+    def _on_foo(self, _: Foo):
         pass
 
 
 def test_capture_custom_evt_nonspecific_capture_include_fw_evts():
-    with capture_events(include_framework=True) as emitted:
-        trigger(State(), "start", MyCharm, meta=MyCharm.META)
+    ctx = Context(MyCharm, meta=MyCharm.META, capture_framework_events=True)
+    ctx.run(ctx.on.start(), State())
```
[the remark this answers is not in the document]
tonyandrewmeyer: Do you mean stop using the "trigger" helper? That can definitely be done everywhere. In general, I've been doing it when tests need to be updated, but there could be a single PR that gets rid of all the remaining ones (but yes agreed in a separate PR).

ops/framework.py
tonyandrewmeyer: Logging output shouldn't appear in action logs, only in juju-log, because we send it to juju-log rather than printing it to stdout/stderr.

We do essentially always have logging set to debug (all the filtering, including level, being done on the Juju side), so that's correct that it would always end up there. But I think that's ok - this really isn't expected to happen, so when it does having a log seems ok (it almost feels like it could be info, but I think most of the time you don't care at all).

ops/charm.py
tonyandrewmeyer: Added a comment.

ops/charm.py
tonyandrewmeyer: I'm not sure. It's briefer (one set of parentheses might not be required here):

```suggestion
        for event in (self._static_events - set(self.on.events())):
```

Personally, I feel like having the if/continue is more readable, as well as avoiding creating an additional set (although set-set does avoid the O(n) checking for membership).

ops/framework.py
tonyandrewmeyer: I really don't like the walrus operator, and feel like it almost never makes code more readable. However, I feel that's a me thing, so I'll make this change, thanks!

PR #1634, opened 2025-03-14
feat: port the venv hotfix for https://bugs.launchpad.net/juju/+bug/2058335

dimaqq:
https://github.com/canonical/tempo-coordinator-k8s-operator/blame/42bd218789af89fb349f87a70267e6f52123c616/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py#L237-L252

Moved out from #1612 

Ref: https://bugs.launchpad.net/juju/+bug/2058335

⚠️ discussion needed if we want to adopt this fix ⚠️ 

### Why adopt it

There are deployed apps that cannot be upgraded to `ops[tracing]` without it.

The fix on the Juju side works at deploy time, meaning that a machine charm that was originally deployed with older Juju is still in a broken state. The Juju change doesn't fix those.

### Why not adopt it

K8s charms are not affected, because a new pod is provisioned on charm upgrade.

There's a command-line `juju exec` fix for deployed machine charms (ssh into charm, find empty dirs, nuke them).

### DIFF
--- a/ops/__init__.py
+++ b/ops/__init__.py
@@ -187,6 +187,8 @@ __all__ = [  # noqa: RUF022 `__all__` is not sorted
 # isort:skip_file
 from typing import Optional, Type
 
+from . import _aaa_venv_workaround as _aaa_venv_workaround
+
 # Import pebble explicitly. It's the one module we don't import names from below.
 from . import pebble
 
--- a/ops/_aaa_venv_workaround.py
+++ b/ops/_aaa_venv_workaround.py
@@ -26,7 +26,10 @@ import shutil
 from collections import defaultdict
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


dimaqq:
Huh, fails on Py 3.8 with `AttributeError: 'PathDistribution' object has no attribute '_normalized_name'`. I'll need to dig into API differences between that and newer Pythons.

tonyandrewmeyer:
Worth a quick discussion in daily on Monday, I think.

dimaqq:
standup summary: team is not sure this warrants inclusion.
Dima will look into format repro of original issue later.

dimaqq:
I'm going to close this for now.

Issue #1623, opened 2025-03-11
Correct/clarify docs about container testing


dwilding:
Collecting a few issues that need fixing:

1. [Get started with charm testing](https://ops.readthedocs.io/en/latest/howto/get-started-with-charm-testing.html#the-testing-framework) shows how to write a unit test for Pebble-ready in a K8s charm, but there are some important parts missing:
    - Needs `container = testing.Container(...)` before creating `state_in`.
    - Needs to specify `containers={container}` when creating `state_in`.

2. In [How to write unit tests for a charm](https://ops.readthedocs.io/en/latest/howto/write-scenario-tests-for-a-charm.html), in the part about mocking beyond the sate, we should mention that `import charm` needs adding so that `MyCharm` is available.

3. The example for [testing.Container.mounts](https://ops.readthedocs.io/en/latest/reference/ops-testing.html#ops.testing.Container.mounts) is wrong because `Mount()` doesn't allow positional arguments.

4. The descriptions for [testing.Mount](https://ops.readthedocs.io/en/latest/reference/ops-testing.html#ops.testing.Mount) suggest that `Mount()` can only be used for files. Actually, it's possible to mount directories (as the example in [testing.Container.mounts](https://ops.readthedocs.io/en/latest/reference/ops-testing.html#ops.testing.Container.mounts) shows).

dwilding:
Items 1 and 2 have been fixed by @IronCore864's improvements to [How to write unit tests for a charm](https://ops.readthedocs.io/en/latest/howto/write-unit-tests-for-a-charm.html) (#1639). Items 3 and 4 have been fixed by my improvements to [`ops.testing.Container.mounts`](https://ops.readthedocs.io/en/latest/reference/ops-testing.html#ops.testing.Container.mounts) and [`ops.testing.Mount`](https://ops.readthedocs.io/en/latest/reference/ops-testing.html#ops.testing.Mount) (#1637).

PR #1637, opened 2025-03-18
docs: clarify how to use mounts in ops.testing.Container

dwilding:
This PR fixes items 3 and 4 in https://github.com/canonical/operator/issues/1623. Changes:

- In the description of `ops.testing.Mount`, changed "file" to "path". I verified that files and directories can both be mounted.
- In the description of `mounts` in `ops.testing.Container`:
    - Simplified the example to show one file and one directory.
    - Added the missing `location` and `source` keywords.
    - Removed the part of the comment that duplicated the example in the docstring.

**[Preview build](https://ops--1637.org.readthedocs.build/en/1637/reference/ops-testing.html#ops.testing.Container.mounts)**

A couple of things I'm not 100% sure about:

- Is it worth wrapping the `source` args in `pathlib.Path`? `Mount()` accepts a string or a Path.
- Is it worth keeping the remaining part of the comment (with more details about pushing and pulling)?

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -796,7 +796,7 @@ class Exec(_max_posargs(1)):
 
 @dataclasses.dataclass(frozen=True)
 class Mount(_max_posargs(0)):
-    """Maps local files to a :class:`Container` filesystem."""
+    """Maps a local path to a :class:`Container` filesystem."""
 
     location: str | pathlib.PurePosixPath
     """The location inside of the container."""
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -997,11 +997,11 @@ class Container(_max_posargs(1)):
             'bin': Mount(location='/bin', source=pathlib.Path('/path/to/local/bin')),
         }
 
-    The simulated container filesystem will have symlinks to `/path/to/local/foo.py` and
-    `/path/to/local/bin` at the specified locations.
+    When your charm runs, the simulated container filesystem will have symlinks to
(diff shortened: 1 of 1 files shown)


IronCore864 (approved):

testing/src/scenario/state.py
```diff
@@ -982,36 +982,24 @@ class Container(_max_posargs(1)):
...
         mounts = {
-            'foo': Mount('/home/foo', pathlib.Path('/path/to/local/dir/containing/bar/py/')),
-            'bin': Mount('/bin/', pathlib.Path('/path/to/local/dir/containing/bash/and/baz/')),
+            'foo': Mount(location='/home/foo.py', source=pathlib.Path('/path/to/local/foo.py')),
+            'bin': Mount(location='/bin', source=pathlib.Path('/path/to/local/bin')),
         }
     """
+    # when the charm runs `pebble.pull`, it will return .open() from one of those paths.
+    # when the charm pushes, it will either overwrite one of those paths (careful!) or it will
+    # create a tempfile and insert its path in the mock filesystem tree
```
james-garner-canonical, line 1002: Should this just go in the docstring? Do you know when it will overwrite your local data and when it will create a tempfile?

james-garner-canonical (approved):
Looks good to me, no need for further docstring enhancements if you want to close this off

testing/src/scenario/state.py
dwilding: Replying for the record - thanks for helping to figure this out!

As we found, local data can be overwritten because `location` inside the temporary filesystem gets symlinked to the `source` path that was specified. I've added a part to the docstring to help make people aware of this.

PR #1640, opened 2025-03-20
chore: remove now-unnecessary testing/tox.ini

benhoyt:
Deleted testing/tox.ini, which includes:

- `unit`: run via top-level `tox -e unit` (and you can easily narrow them down by specifying just the `testing` dir)
- `lint`: covered by top level tox
- `static`: ditto
- `lint-tests`: ditto
- `fmt`: ditto
- `docs-deps` and `docs`: no longer needed
- `test-readme`: available in the top-level tox

Also in this PR:

- Point to top-level HACKING.md in CONTRIBUTING.md.
- Get `tox -e test-readme` working. Copied a snippet from one of the tests.
- Remove old badge links from CONTRIBUTING.md and README.md


tonyandrewmeyer:
> Kept:
> 
> * `test-readme`: this is the only one remaining here. We should run this in CI, but it's not worth it if the README is mostly going away soon.

Agreed that it probably will go away soon. But we could remove it here, because it's at the top level.

testing/README.md
```diff
@@ -649,9 +644,20 @@ check doesn't have to match the event being generated: by the time that Juju
...
+    "checks": {"http-check": {"override": "replace", "startup": "enabled", "failures": 7}},
+})
+check_info = scenario.CheckInfo(
+    "http-check",
+    status=pebble.CheckStatus.UP,
+    level=layer.checks["http-check"].level,
+    startup=layer.checks["http-check"].startup,
+    threshold=layer.checks["http-check"].threshold,
+)
+container = scenario.Container("my_container", check_infos={check_info}, layers={"layer1": layer})
```
tonyandrewmeyer, line 660: Thanks - I missed this in the recent consistency check PR. A good argument for having this environment run automatically (and a better one for removing these examples from the README).

tonyandrewmeyer (approved):
Thanks! I'd prefer to just delete `testing/tox.ini`, but this is fine too.

IronCore864 (approved):

PR #1643, opened 2025-03-23
chore: bump pre-commit versions, plus small fixes

tonyandrewmeyer:
Possibly simplest to review commit by commit.

This bumps the versions of the pre-commit tools.

A few small issues are also fixed (this is somewhat a subset of #1474, skipping all the docs and controversial changes).

* A few strings are formatted more cleanly (including where previous tooling has put in `" "` breaks).
* A fix to a framework test (found in #1474).
* A fix to a Harness test.
* Removing shebangs in Scenario code (none of the files are executable).


ops/_private/harness.py
```diff
@@ -1796,8 +1796,7 @@ def revoke_secret(self, secret_id: str, observer: AppUnitOrName):
         # Model secrets:
         if secret.owner_name in [self.model.app.name, self.model.unit.name]:
             raise RuntimeError(
-                f'Secret {secret_id!r} owned by the charm under test, "'
-                f"can't call revoke_secret"
+                f'Secret {secret_id!r} owned by the charm under test, "can\'t call revoke_secret'
```
james-garner-canonical: I don't understand why there's a double quote at the start of the word `can't` -- a mistake?

ops/_private/harness.py
```diff
@@ -1796,7 +1796,7 @@ def revoke_secret(self, secret_id: str, observer: AppUnitOrName):
         # Model secrets:
         if secret.owner_name in [self.model.app.name, self.model.unit.name]:
             raise RuntimeError(
-                f'Secret {secret_id!r} owned by the charm under test, "can\'t call revoke_secret'
+                f"Secret {secret_id!r} owned by the charm under test, can't call revoke_secret"
```
james-garner-canonical, line 1799: Ah, it's dropped in this commit

test/test_framework.py
```diff
@@ -1982,7 +1982,8 @@ def test_breakpoint_good_names(self, request: pytest.FixtureRequest, name: str):
             '-',
             '...foo',
             'foo.bar',
-            'bar--' 'FOO',
+            'bar--',
+            'FOO',
```
james-garner-canonical, line 1986: Reckon it's worth keeping something like the 'bar--FOO' case? Probably unnecessary, but separate logical test cases might be clearer:
- trailing hyphen 'bar-'
- all caps 'FOO'
- multiple consecutive hyphens 'foo--bar'
- mixed case 'fOo-BaR'

james-garner-canonical (approved):
I went commit by commit. Looks good. The accidental 'bar--' 'FOO' concatenation was interesting in that the tests work either way, which is surprising at first. Left a suggestion for perhaps clearer test cases, but not a blocker.

benhoyt:
Happy with this. I'm not quite sure though -- why aren't these things caught in CI by the ruff checks there?

james-garner-canonical:
I think it's because this jumps to using the latest `0.11` ruff version, while we're using `0.7` in ci currently, which I should make a PR to bump. Merging the fixes here first should make that smoother, thanks, Tony!

test/test_framework.py
tonyandrewmeyer: If you look at line 1459 of the [original commit](https://github.com/canonical/operator/commit/11a1849205d750e28aaa4a13938b5864659f928b) it looks very much like a mistake.

Multiple consecutive hyphens are accepted by the regex we currently use. I'm not sure if there are any restrictions applied by Juju (the docs don't say either way).

`bar--` doesn't work because it ends with a hypen. `FOO` doesn't work because it has an upper-case character. `bar--FOO` doesn't work because of the upper-case character.

We could add more tests here, or arrange these differently, but I think that belongs in a different PR.

james-garner-canonical: Ah, if consecutive hyphens are actually ok so far as we know, then all I'd suggest here is to change 'bar--' to 'bar-' to avoid implying the opposite. Probably fine to make the one character change in this PR if it's worth doing at all -- I don't really think it's worth a separate PR.

test/test_testing.py
```diff
@@ -361,7 +361,7 @@ def _on_cluster_relation_changed(self, event: ops.EventBase):
         harness.set_leader(False)
         harness.update_relation_data(rel_id, 'test-app', {'k': 'v3'})
         assert backend.relation_get(rel_id, 'test-app', is_app=True) == {'k': 'v3'}
-        assert len(harness.charm.observed_events), 1
+        assert len(harness.charm.observed_events) == 1
```
dimaqq, line 364: ouch!

dimaqq (approved):

PR #1648, opened 2025-03-25
feat: expose the config metadata in CharmMeta

### LINKED ISSUE #1493: Feature: Add `config` property to `ops.CharmMeta`
There are situations where we want to compare against or set the default config values of a charm config option.

E.g. A user sets an invalid log level and so the charm sets the default value.

``` python
if some_config not in allowed_list:
    some_config = self.meta.config["some_config"]["default"]
```

If we we have string literals scattered throughout code that are supposed to "track" the config defaults then this is code duplication and could be solved by adding a config property to [ops.CharmMeta](https://ops.readthedocs.io/en/latest/#ops.CharmMeta). This currently supports most of the top-level keys of a charmcraft.yaml file and Scenario already effectively has this, so it's quite viable.

Currently this is achievable with loading the charmcraft.yaml yourself. Another way is to use stored state but both options enable varying coding styles across charms which could be solved with the `self.meta`.

tonyandrewmeyer:
Just like `metadata.yaml` and `actions.yaml` (whether individual files or in `charmcraft.yaml`), load the `config.yaml` and expose it in `CharmMeta`.

This adds loading to Harness. Although we're generally not adding functionality, Harness already loaded the meta to be able to handle working with config, so it was simpler to just pass it through to `CharmMeta` than to have it be oddly empty. The one backwards-incompatible change is that if you had a `config.yaml` that was missing the required `type` field for an option, Harness would previously raise a `RuntimeError` on `config_set` and now raises a `KeyError` when loading the meta. I feel like this is a reasonable change, and people shouldn't be testing for config that Juju wouldn't accept.

Scenario also is updated - it was also already setting up `config.yaml`, but now also passes it through to `CharmMeta`.

Fixes #1493


ops/charm.py
```diff
@@ -1896,6 +1922,29 @@ def __init__(self, name: str, raw: Optional[Dict[str, Any]] = None):
...
+class ConfigMeta:
+    """Object containing metadata about a config option."""
+
+    name: str
+    """Name of the config option."""
+
+    type: Literal['string', 'int', 'float', 'boolean', 'secret']
+    """Type of the config option."""
+
+    default: Any
```
benhoyt, line 1934: Could/should this be `Union[str, int, float, bool, None]`?

ops/charm.py
```diff
@@ -1896,6 +1922,29 @@ def __init__(self, name: str, raw: Optional[Dict[str, Any]] = None):
         self.additional_properties = raw.get('additionalProperties', True)
 
 
+class ConfigMeta:
```
benhoyt, line 1925: Should we make new classes like this dataclasses, or keep consistent with the others here?

benhoyt (approved):

ops/charm.py
```diff
@@ -1896,6 +1922,29 @@ def __init__(self, name: str, raw: Optional[Dict[str, Any]] = None):
...
+    default: Any
+    """Default value of the config option."""
+
+    description: str
+    """Description of the config option."""
+
+    def __init__(self, name: str, raw: Optional[Dict[str, Any]] = None):
+        raw = raw or {}
+        self.name = name
+        self.type = raw['type']
```
james-garner-canonical, line 1943: This will error if `raw` was `None`, which I guess means it shouldn't be `Optional`?

ops/charm.py
james-garner-canonical: We'd get a nice repr, but we'd have to do a `from_dict` style constructor, which maybe kind of evens out.

test/test_testing.py
```diff
@@ -1241,7 +1241,7 @@ def test_config_secret_option(self, request: pytest.FixtureRequest):
         assert harness.charm.changes == [{'name': 'config-changed', 'data': {'a': secret_id}}]
 
     def test_no_config_option_type(self):
-        with pytest.raises(RuntimeError):
+        with pytest.raises(KeyError):
```
james-garner-canonical, line 1244: We could explicitly raise a `RuntimeError` when the `'type'` key is missing to avoid this behaviour change?

ops/charm.py
tonyandrewmeyer: Nice catch, yes it should.

```suggestion
    default: Union[bool, int, float, str, None]
```

test/test_testing.py
tonyandrewmeyer: I wondered about that. This should ever happen in production because charmcraft validates that there's a `type` when you pack, so you'd have to be deploying a non-packed charm or have messed with the installed `config.yaml`. It seemed like `RuntimeError` was the wrong kind of error for this, but on balance probably it is better to just be consistent.

ops/charm.py
tonyandrewmeyer: Hmm, good catch, thanks. I was modelling this on `ActionMeta` but that can actually be just a name and nothing else. An option needs at least a type, so not passing a dictionary here doesn't make sense.

ops/charm.py
tonyandrewmeyer: I wondered about this too, and thought consistency was best since there didn't seem to be a clear benefit in a dataclass here. On the other hand, if we don't start using them for new things then we'll never have them - I could easily be convinced to switch here.

> We'd get a nice repr, but we'd have to do a `from_dict` style constructor, which maybe kind of evens out.

The regular `__init__` should be fine - `CharmMeta` can just have something like:

```python
    self.config = {
            name: ConfigMeta(
                name,
                type=config['type'],
                default=config.get('default'),
                description=config.get('description'),
            )
            for name, config in config_raw_.get('options', {}).items()
        }
```

No-one should need to create instances of this class manually, so we don't need to provide a `ConfigMeta(name, raw: dict)` signature.

test/test_testing.py
tonyandrewmeyer: This is definitely more annoying in the dataclass version, though...

ops/charm.py
tonyandrewmeyer: In https://github.com/canonical/operator/pull/1648/commits/3c653e64e6fa9e2cfaa59f5dd178c6193ad09e1d I changed it to a dataclass. Which version do you prefer? I think I do like the dataclass one more, except for the annoying bit checking for no `type`.

test/test_testing.py
james-garner-canonical: Could wrap `self.config = {...` in a `try ... except` to catch the `KeyError` instead?

james-garner-canonical: Or if the behaviour change is not a big deal, then just raise `KeyError`

tonyandrewmeyer: > Could wrap `self.config = {...` in a `try ... except` to catch the `KeyError` instead?

Yes, although then the error can't say which option is faulty.

james-garner-canonical: Ah, true, I retract that suggestion then.

ops/charm.py
benhoyt: I slightly prefer the dataclass version, and gets us moving in that direction. I'd probably do the "parsing" in a `_from_dict` method, but as you have it is fine too.

james-garner-canonical (approved):

PR #1649, opened 2025-03-25
docs: fix code snippet indentation

IronCore864:
Fix indentation. It used a tab before, now everything is updated to spaces.

Before: https://ops.readthedocs.io/en/latest/explanation/holistic-vs-delta-charms.html

After: https://ops--1649.org.readthedocs.build/en/1649/explanation/holistic-vs-delta-charms.html


dwilding (approved):

benhoyt (approved):
Thanks!

PR #1658, opened 2025-03-31
ci: post-release versioning, release process update + workflow fix

james-garner-canonical:
This PR bumps the post-release versions of `ops` and `ops-scenario`, and updates `HACKING.md` to reflect the process followed in this release. It also adds a fix for the failing `observability-charm-tests` (needed for the new versions in this PR), which should fix them for releases too, and updates the `docs-deps` tox environment that the docs build with the development version requirements.

Post-release, I've changed the dependency versions from `==$latest_release` to `==$dev_version`. To allow the docs to build with this change, I've modified the `docs-deps` tox environment to add the line `.` to the requirements before `./testing/`, which means that ops is installed from the local files first, then `ops-scenario`'s requirements requests the dev version when it's installed, and finds it already installed.

This PR also includes a fix to `observability-charm-tests` from https://github.com/dimaqq/operator/pull/33#discussion_r2018070161, without which the test fails due to the development version requirements. Dima's fix also works with exact unreleased dependency versions (like the dev versions used in this PR), which I believe allows us to re-enable the `observability-charm-tests` for releases.

### DIFF
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -40,7 +40,6 @@ jobs:
             sed -i -e "/^ops-scenario[ ><=]/d" -e "/^ops\[testing\][ ><=]/d" requirements.txt
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops-scenario&subdirectory=testing" >> requirements.txt
           elif [ -e "uv.lock" ]; then
-            set -vex
             uv remove ops --optional dev --frozen || echo "maybe ops[testing] is not a dev dependency"
             uv remove ops-scenario --optional dev --frozen || echo "maybe ops-scenario is not a dev dependency"
             uv remove ops --frozen
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -40,8 +40,12 @@ jobs:
             sed -i -e "/^ops-scenario[ ><=]/d" -e "/^ops\[testing\][ ><=]/d" requirements.txt
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops-scenario&subdirectory=testing" >> requirements.txt
           elif [ -e "uv.lock" ]; then
-            uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA --raw-sources
-            uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#subdirectory=testing --raw-sources
+            set -vex
+            uv remove ops --optional dev --frozen || echo "maybe ops[testing] is not a dev dependency"
+            uv remove ops-scenario --optional dev --frozen || echo "maybe ops-scenario is not a dev dependency"
+            uv remove ops --frozen
+            uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA --raw-sources --prerelease=allow
+            uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#subdirectory=testing --optional dev --raw-sources --prerelease=allow
           else
             echo "Error: no requirements.txt or uv.lock file found"
             exit 1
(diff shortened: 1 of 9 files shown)


.github/workflows/observability-charm-tests.yaml
```diff
@@ -40,8 +40,12 @@ jobs:
             sed -i -e "/^ops-scenario[ ><=]/d" -e "/^ops\[testing\][ ><=]/d" requirements.txt
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops-scenario&subdirectory=testing" >> requirements.txt
           elif [ -e "uv.lock" ]; then
-            uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA --raw-sources
-            uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#subdirectory=testing --raw-sources
+            set -vex
```
tonyandrewmeyer, line 43: In Dima's PR this was noted as being there just for working on things. Do we need it here all the time?

tox.ini
```diff
@@ -42,7 +42,7 @@ description = Compile the requirements.txt file for docs
 deps = pip-tools
 commands =
     pip-compile --extra=docs -o docs/requirements.txt pyproject.toml
-    python -c 'open("docs/requirements.txt", "a").write("./testing/\n")'
+    python -c 'open("docs/requirements.txt", "a").write(".\n./testing/\n")'
```
tonyandrewmeyer, line 45: Does this still pick up changes if you do `tox -e docs` and then change something and then `tox -e docs` again? I wonder if `-e` is needed to force an editable install (although I'm not sure whether rtd would be ok with that).

HACKING.md
```diff
@@ -368,20 +364,24 @@ To make a release of the `ops` and/or `ops-scenario` packages, do the following:
    and add it to `CHANGES.md`.
 10. For `ops`, change [version.py](ops/version.py)'s `version` to the
    appropriate string. For `ops-scenario`, change the version in
-   [testing/pyproject.toml](testing/pyproject.toml). Both packages use
-   [semantic versioning](https://semver.org/), and adjust independently
-   (that is: ops 2.18 doesn't imply ops-scenario 2.18, or any other number).
-11. Run `uvx -p 3.11 tox -e docs-deps` to recompile the `requirements.txt` file
+   [testing/pyproject.toml](testing/pyproject.toml). We use both
+   [semantic versioning](https://semver.org/) and lockstep releases, so if
+   one library requires a major or minor version bump, the other will too.
```
tonyandrewmeyer, line 369: Major, minor, or patch/macro/bugfix. So maybe it's better to just say "so if one library requires a version bump, the other will too".

HACKING.md
```diff
@@ -395,11 +395,13 @@ To make a release of the `ops` and/or `ops-scenario` packages, do the following:
...
 
-15. Announce the release on [Discourse](https://discourse.charmhub.io/c/framework/42) and [Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com).
+16. Announce the release on [Discourse](https://discourse.charmhub.io/c/framework/42) and [Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com).
 
-16. Open a PR to change the version strings to the expected
-   next version, with ".dev0" appended (for example, if 3.14.1 is the next
-   expected version, use `'3.14.1.dev0'`).
+17. Open a PR to change the version strings to the expected next version, with ".dev0" appended
+   (for example, if 3.14.0 is the next expected version, use `'3.14.0.dev0'`).
+   In this PR, also update the requirements in the respective pyproject.toml files to be >= the
```
tonyandrewmeyer, line 402: It seems that you went with `==` rather than this - I assume the files are correct and the docs outdated?

tox.ini
james-garner-canonical: Yes, it still works as is. I ran `tox -e docs`, edited a docstring, and then reran, and the docstring changes showed in the rendered docs.

HACKING.md
james-garner-canonical: Thanks, done.

HACKING.md
james-garner-canonical: Good catch -- yes the docs were outdated, fixed now, thanks.

.github/workflows/observability-charm-tests.yaml
james-garner-canonical: I added it here and removed the `||` for the removing `ops` step as I figured we'd want the chance to look into things if that line ever failed. I've removed it since apparently github sets `-e` anyway when using `bash` as the shell.

tonyandrewmeyer (approved):

benhoyt (approved):

Issue #1360, opened 2024-09-05
TestExec.test_no_wait_call may be flaky


dimaqq:
An example failure in the doc-only change:

https://github.com/canonical/operator/actions/runs/10713359939/job/29705252719

```py
=================================== FAILURES ===================================
__________________________ TestExec.test_no_wait_call __________________________
[gw0] linux -- Python 3.10.14 /home/runner/work/operator/operator/.tox/unit/bin/python
Traceback (most recent call last):
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/_pytest/runner.py", line 262, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/pluggy/_hooks.py", line 513, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/pluggy/_manager.py", line 120, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/pluggy/_callers.py", line 182, in _multicall
    return outcome.get_result()
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/pluggy/_result.py", line 100, in get_result
    raise exc.with_traceback(exc.__traceback__)
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/pluggy/_callers.py", line 103, in _multicall
    res = hook_impl.function(*args)
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/_pytest/runner.py", line 177, in pytest_runtest_call
    raise e
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/_pytest/runner.py", line 169, in pytest_runtest_call
    item.runtest()
  File "/home/runner/work/operator/operator/.tox/unit/lib/python3.10/site-packages/_pytest/python.py", line 1792, in runtest
    self.ihook.pytest_pyfunc_call(pyfuncitem=self)
[...]

PR #1664, opened 2025-04-02
fix: try to fix flaky pebble exec test

IronCore864:
Fix a flaky Pebble exec test.

Resolves https://github.com/canonical/operator/issues/1360.

After looking into it, I found that the flaky test happens because garbage collection (GC) can clean up other temporary objects when this test (`test_no_wait_call`) runs. If other temporary objects are cleaned up right before the `ExecProcess` object's cleanup method `__del__` is called, it makes the test sometimes catch the wrong ResourceWarning.

Since it's hard to reproduce this problem (not to mention consistently), I've been thinking about how to fix it. Here are a few ideas:

1. Force garbage collection: Call `gc.collect()` right after `del process`. Or even move the `client.exec` part outside the `with` block to shorten the time in the `with` block before GC happens.
  - This would lower the chance of catching unrelated warnings.
  - Problem: This is still not a valid solution, because we can't be sure that no other GC happens inside the `with` block before our forced cleanup.
2. Turn off automatic GC: Disable automatic GC temporarily during the test and re-enable it afterward.
  - This would reduce the chance of catching other warnings caused by automatic GC.
  - Problem: Even with automatic GC off, reference-counting GC can still happen, and we might still catch unwanted warnings. So, this (and the previous idea) is still non-deterministic and could still be flaky- maybe a lot less flaky, but still non-deterministic.
3. Filter Warnings (best fix?): Only check for the `ExecProcess` warnings by messages and ignore others.
  - This guarantees we catch the right warning, no matter when GC runs, so it is probably the most deterministic way.

So, in this PR, I'm suggesting the last choice.

### DIFF
--- a/test/test_pebble.py
+++ b/test/test_pebble.py
@@ -3346,8 +3346,13 @@ class TestExec:
         with pytest.warns(ResourceWarning) as record:
             process = client.exec(['true'])
             del process
+
+        # GC might collect other warnings. Filter the warnings to only
+        # those from ExecProcess.
+        warnings = [r for r in record if 'ExecProcess' in str(r.message)]
+        assert len(warnings) == 1
         assert (
-            str(record[0].message)
+            str(warnings[0].message)
             == 'ExecProcess instance garbage collected without call to wait() or wait_output()'
         )
 
--- a/test/test_pebble.py
+++ b/test/test_pebble.py
@@ -3347,8 +3347,8 @@ class TestExec:
             process = client.exec(['true'])
             del process
 
-        # GC might collect other warnings. Filter the warnings to only
-        # those from ExecProcess.
+        # GC may trigger unrelated warnings (for example, from TemporaryDirectory).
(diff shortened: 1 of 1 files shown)


test/test_pebble.py
```diff
@@ -3347,8 +3347,8 @@ def test_no_wait_call(self, client: MockClient):
             process = client.exec(['true'])
             del process
 
-        # GC might collect other warnings. Filter the warnings to only
-        # those from ExecProcess.
+        # GC may trigger unrelated warnings (for example, from TemporaryDirectory).
+        # Filter the warnings to only those from ExecProcess.
```
dimaqq, line 3351: I'd say that this second comment line is redundant.

dimaqq (approved):

benhoyt (approved):

james-garner-canonical (approved):
Looks good to me based on your thorough explanation, thanks for this fix!

Issue #1324, opened 2024-08-20
Remove legacy code that sets up event symlinks


benhoyt:
Per discussion with @jameinel, these haven't been needed since `JUJU_DISPATCH_PATH` was invented, since about Juju 2.7. Let's get rid of the code in main.py.

dimaqq:
There's also this on the code path:

```
# When a charm is force-upgraded and a unit is in an error state Juju
# does not run upgrade-charm and instead runs the failed hook followed
# by config-changed. Given the nature of force-upgrading the hook setup
# code is not triggered on config-changed.
#
# 'start' event is included as Juju does not fire the install event for
# K8s charms (see LP: #1854635).
```

The referenced ticket, https://bugs.launchpad.net/juju/+bug/1854635

Is stamped as resolved in Juju 2.7.6 and 2.8.0

benhoyt:
@dwilding still planning to do this this cycle.

PR #1665, opened 2025-04-02
chore: remove code for creating event symlinks

dwilding:
Per https://github.com/canonical/operator/issues/1324, we can remove code from [`main.py`](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/ops/_main.py) that creates symlinks. My interpretation is that we can assume that Juju is providing a dispatch path, so we can also remove "no dispatch" tests.

I've started by removing what I think can be removed at minimum, but I've run into a failing test that I don't really understand. The failing test is [`test_emitted_deferred`](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/testing/tests/test_e2e/test_event.py) in the Scenario tests.

This test started failing after I removed [`_setup_event_links()`](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/ops/_main.py#L92-L112) - not when I removed only the inner call to `_create_event_link()`. From what I can tell, removing `charm.on.events()` on [this line](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/ops/_main.py#L109) caused the test to fail, and re-inserting `dummy = charm.on.update_status` is enough to get the test to pass again. Is there some kind of side-effect going on here?

I've hit the limit of my understanding so will ask for help!


dwilding:
Adding a comment to record my previous issue:

> I've started by removing what I think can be removed at minimum, but I've run into a failing test that I don't really understand. The failing test is [`test_emitted_deferred`](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/testing/tests/test_e2e/test_event.py) in the Scenario tests.

> This test started failing after I removed [`_setup_event_links()`](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/ops/_main.py#L92-L112) - not when I removed only the inner call to `_create_event_link()`. From what I can tell, removing `charm.on.events()` on [this line](https://github.com/canonical/operator/blob/2eab3c9d41597ed924593e41ecf2c9aa4a5314ae/ops/_main.py#L109) caused the test to fail, and re-inserting `dummy = charm.on.update_status` is enough to get the test to pass again. Is there some kind of side-effect going on here?

This is fixed by [`e98ac60`](https://github.com/canonical/operator/pull/1665/commits/e98ac60b56c999bf73a36041df6030f4f7491dd3) - thank you @tonyandrewmeyer!

ops/_main.py
```diff
@@ -399,19 +317,14 @@ def __init__(
 
         # Set up dispatcher, framework and charm objects.
         self.dispatcher = _Dispatcher(self._charm_root, self._juju_context)
-        self.dispatcher.run_any_legacy_hook()
+        self.dispatcher.run_any_legacy_hook()  # TODO(dwilding): I guess we need to keep this
```
dwilding, line 320: To discuss: I assume we need to keep this. Because even though we can assume that Juju is providing a dispatch path, the charm itself might contain legacy hooks.

ops/_main.py
```diff
@@ -221,25 +157,7 @@ def __init__(self, charm_dir: Path, juju_context: _JujuContext):
...
-        # by config-changed. Given the nature of force-upgrading the hook setup
-        # code is not triggered on config-changed.
-        #
-        # 'start' event is included as Juju does not fire the install event for
-        # K8s charms https://bugs.launchpad.net/juju/+bug/1854635, fixed in juju 2.7.6 and 2.8
-        if self.event_name in ('install', 'start', 'upgrade_charm') or self.event_name.endswith(
-            '_storage_attached'
-        ):
-            _setup_event_links(self._charm_dir, charm, self._juju_context)
+            self._init_legacy()  # TODO(dwilding): Can we remove this whole path?
```
dwilding, line 160: To discuss: Can we remove this whole path, as we're assuming that Juju is providing a dispatch path?

benhoyt: Yes, we can remove this whole path. I think we could change this to just the following:

```
assert self._juju_context.version.is_dispatch_aware(), self._juju_context.version
self._init_dispatch()
```

Or just inline `_init_dispatch()` right into `__init__` here.

ops/_main.py
benhoyt: I'm pretty sure we can remove this too, as no Ops charm should have legacy hook files. @jameinel Does that sound right to you?

ops/_main.py
dwilding: In that case can we remove the whole concept of `is_dispatch_aware`?

benhoyt: Yes, at least in `_main.py`. We can't remove `JujuVersion.is_dispatch_aware` for backwards-compatibility reasons, though.

ops/_main.py
benhoyt: John said:

> I would not expect that most people writing ops code would have a shell script that they also wanted to run.
> However, if there were a file in hooks/* when we designed the work, it is the responsibility of ops to run that, because Juju will only call dispatch. And why would you have a file there that doesn't do anything?
> My guess is that if you do an audit, you'll find that several charms happen to have hooks/ but they'll mostly be the stub hooks that we used for Juju 2.7 compatibility (that didn't understand dispatch)
> I would consider it a breaking change, but reasonable to do in a release, or at least bring up to teams to see if anyone is needing that functionality, and then announce it widely that "ops version X will be dropping support for legacy hooks/".

Which is a good point. However, I've done a quick check of the 159 charms in my `charm-analysis/.cache` directory, and the only ones with a `hooks` directory are reactive or non-ops charms. See https://pastebin.canonical.com/p/WhHznyVYGG/ ... no `src` or Ops in sight. So I think John's right it's a breaking change, but it seems highly unlikely to affect anyone given that.

@tonyandrewmeyer Given that "due dilegence", what do you think about just dropping this (without a major version bump)?

dwilding: Thanks Ben & John for the advice. I'll pick this up after Tony's had a chance to look. Keeping this in draft for now.

tonyandrewmeyer: I'd be very surprised if there was any use, although if there was it does seem like it would be very significant breakage (probably a good thing, since integration tests ought to catch it). However, it does seem like a violation of our promise to remove it, especially without any warning.

What do we gain from removing it now rather than slapping a warning on it and removing it in 3.0? This code is rarely touched, so it doesn't save much in maintenance burden. How much of a performance cost is it supporting this?

benhoyt: Yeah, that's fair push-back. Let's add a deprecation warning and leave it then.

ops/_main.py
dwilding: I just ran into something that's relevant to this discussion: https://github.com/canonical/operator/issues/1677

ops/_main.py
```diff
@@ -266,7 +181,7 @@ def run_any_legacy_hook(self):
 
         argv = sys.argv.copy()
         argv[0] = str(dispatch_path)
-        logger.info('Running legacy %s.', self._dispatch_path)
+        logger.info('Running legacy %s (legacy hooks are deprecated).', self._dispatch_path)
```
dwilding, line 184: I'm saying "(legacy hooks are deprecated)" rather than simply "(deprecated)" to avoid someone thinking that a particular hook is deprecated.

dwilding:
I'm requesting several pairs of eyes on this, to make extra sure that I didn't accidentally break something. Thanks all!

ops/_main.py
```diff
@@ -217,43 +151,24 @@ def __init__(self, charm_dir: Path, juju_context: _JujuContext):
         self._charm_dir = charm_dir
         self._exec_path = Path(self._juju_context.dispatch_path or sys.argv[0])
 
-        dispatch = charm_dir / 'dispatch'
-        if self._juju_context.version.is_dispatch_aware() and _exe_path(dispatch) is not None:
-            self._init_dispatch()
-        else:
-            self._init_legacy()
+        # Grab the wanted hook from JUJU_DISPATCH_PATH, e.g. hooks/install.
```
benhoyt, line 154: Nit: "wanted hook" reads a bit funny here. How about:
```suggestion
        # Grab the correct hook from JUJU_DISPATCH_PATH, e.g. hooks/install.
```

ops/_main.py
```diff
@@ -217,43 +151,24 @@ def __init__(self, charm_dir: Path, juju_context: _JujuContext):
...
-        ):
-            _setup_event_links(self._charm_dir, charm, self._juju_context)
+        self._set_name_from_path(self._dispatch_path)
 
     def run_any_legacy_hook(self):
         """Run any extant legacy hook.
 
-        If there is both a dispatch file and a legacy hook for the
-        current event, run the wanted legacy hook.
+        If there is a legacy hook for the current event, run the wanted legacy hook.
```
benhoyt, line 167:
```suggestion
        If there is a legacy hook for the current event, run it.
```

benhoyt (approved):
Looks good to me, thanks. Less code to maintain, and the unit tests run almost twice as fast. (14s instead of 26s on my machine)

dimaqq:
Would it make sense to remove `is_dispatch_aware()` from `ops/jujuversion.py`?
Or do we want to keep it lest someone somewhere uses it?
There no hits in our collection of charms or github search, but who knows, right?

cc @tonyandrewmeyer

benhoyt:
@dimaqq I think you're right that it's highly unlikely any charms use it (it was probably just the framework), but I don't see any need to remove it and make that small breaking change -- it's not getting in the way.

testing/tests/test_e2e/test_event.py
```diff
@@ -77,7 +77,11 @@ def test_emitted_deferred():
     class MyCharm(CharmBase):
         META = {"name": "joop"}
 
-        def _foo(self, e):
+        def __init__(self, framework: ops.Framework):
+            super().__init__(framework)
+            framework.observe(self.on.update_status, self._on_update_status)
+
+        def _on_update_status(self, _: ops.UpdateStatusEvent):
```
dimaqq, line 84: This change seems unrelated:
- it's a test for Scenario, not ops
- I'm unsure about possible original quirks for this test, cc @tonyandrewmeyer 

The change seems reasonable, but I wonder about the minute semantics, specifically having a deferred event with a handler that is no longer "registered". I suppose it's something that could happen on charm upgrade.

ops/_main.py
dimaqq: If the legacy hooks are deprecated, WDYT about upgrading this to a `warnings.warning`?

dimaqq (commented):
Code changes look fine.
Let's wrangle the details and it should be good to go.

test/test_main.py
```diff
@@ -243,12 +229,11 @@ def fake_script(request: pytest.FixtureRequest):
 
 class _TestMain(abc.ABC):
     @abc.abstractmethod
-    def _setup_entry_point(self, directory: Path, entry_point: str):
+    def _setup_entry_point(self):
```
dimaqq, line 232: I wonder if we could get rid of the ABC class, given that this class tree is much smaller/simpler now.

test/test_main.py
```diff
@@ -1402,12 +1190,10 @@ def test_crash_action(self, request: pytest.FixtureRequest, fake_script: FakeScr
         assert "'foo' not found in 'bar'" in stderr
 
 
-class TestMainWithDispatchAsScript(_TestMainWithDispatch):
+class TestMainWithDispatchAsScript(_TestMain):
```
dimaqq, line 1193: I'm a bit unclear if perhaps dispatch-is-a-script should be the default, as ops-based charms don't have a custom dispatch (they could, but not a single charm in our collection does that), and charmcraft generates dispatch which is a bash script...

But then... I see references to DISPATCH_PATH from year 2020 in charmcraft, and yet there's a PR from Aug 2024 that "adds dispatch if it doesn't exist". So is it a long-standing feature or a new feature?

testing/tests/test_e2e/test_event.py
tonyandrewmeyer: It's needed as a fix at the moment (David borrowed it from my PR). If we want to support this in the future, then we should have a specific Scenario ticket for it.

ops/_main.py
```diff
@@ -217,43 +151,24 @@ def __init__(self, charm_dir: Path, juju_context: _JujuContext):
...
 
-    def ensure_event_links(self, charm: '_charm.CharmBase'):
-        """Make sure necessary symlinks are present on disk."""
-        if self.is_dispatch_aware:
-            # links aren't needed
-            return
+        if 'OPERATOR_DISPATCH' in os.environ:
+            logger.debug('Charm called itself via %s.', self._dispatch_path)
+            raise _Abort(0)
+        os.environ['OPERATOR_DISPATCH'] = '1'
```
tonyandrewmeyer, line 160: Do we have a test for this behaviour? I couldn't find one, but test_main is complicated, so I could have missed it.

tonyandrewmeyer (approved):
Looks good, thanks!

I agree with the suggestion about making the deprecation message an actual warning.

testing/tests/test_e2e/test_event.py
dwilding: This change was needed because the Scenario test started failing after I removed the code that created symlinks. The symlink code was introspecting the charm to find which events to create symlinks for, and there was some interaction with Scenario. I don't fully understand the interaction, but Tony helped me fix the Scenario test.

ops/_main.py
dimaqq: Arguably, that's not the subject of this PR, whether or not there is a test.

dimaqq (approved):

test/test_main.py
dwilding: Yeah, I guess this could be simpler now, as only two methods are implemented differently in `TestMainWithDispatchAsScript` vs `TestMainWithDispatchAsSymlink`. We also use the ABC pattern in test_storage, so I think I'll keep it as-is for the sake of merging. But if you/others feel it would be clearer a different way, I'd be happy to do a follow-up PR.

dwilding:
Thanks all for the reviews! This turned out to be more complicated than expected (don't they often :slightly_smiling_face:)

I've changed the deprecation warning from this:

```python
# ... some checks to make sure we should attempt to run the hook

logger.info('Running legacy %s (legacy hooks are deprecated).', self._dispatch_path)
```

to this:

```python
warnings.warn('Legacy hooks are deprecated.', category=DeprecationWarning)

# ... some checks to make sure we should attempt to run the hook

logger.info('Running legacy %s.', self._dispatch_path)
```

Hopefully this looks good. I'd be happy to revise further in a follow-up PR if needed.

PR #1667, opened 2025-04-04
chore: build docs using tox

dimaqq:
Harmonise docs build between local and rtd.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,6 +9,6 @@ build:
        - asdf plugin add uv
        - asdf install uv latest
        - asdf global uv latest
-  build:
-    html:
-       - uvx tox -e docs
+    build:
+      html:
+         - uvx tox -e docs
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,14 +1,14 @@
 version: 2
 
-python:
-  install:
-    - requirements: docs/requirements.txt
-
 build:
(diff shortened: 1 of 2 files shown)


benhoyt (commented):
Can you explain what the create_environment and the `READTHEDOCS_OUTPUT` do?

dimaqq:
Create environment allows us to install more tools.
I'm adding `uv` in order to `uvx tox`.

The env var tells us where RTD wants the build output.
https://docs.readthedocs.com/platform/stable/config-file/v2.html
https://docs.readthedocs.com/platform/stable/build-customization.html#where-to-put-files

dimaqq:
It does seem a little magical...

If the env var name doesn't give enough clue, I can add a comment to the env var in the `pass_env` statement. wdyt?

james-garner-canonical (approved):
Looks good to me, and the docs appear to be building correctly still. Neat that readthedocs provides `asdf` for build time setup of tools. I was also curious about the `create_environment` name -- it's a pre-defined job name available in builds.job (see [docs](https://docs.readthedocs.com/dev/latest/design/future-builder.html#build-jobs)).

dimaqq:
I've documented the magic env var.
Wrt. the create_environment stage, I feel it's self-documenting; at least I can't come up with something interesting to say, if I were to add a comment.

benhoyt (approved):
Thanks!

PR #1670, opened 2025-04-08
docs: clarify guidance about designing Python modules

dwilding:
This PR clarifies the guidance in [Design your Python modules](https://ops.readthedocs.io/en/latest/howto/write-and-structure-charm-code.html#design-your-python-modules).

Main things that I'm clarifying:

- When to use private vs public methods
- What would go in `<workload>.py`

**[Preview docs build](https://ops--1670.org.readthedocs.build/en/1670/howto/write-and-structure-charm-code.html#design-your-python-modules)**

### DIFF
--- a/docs/howto/write-and-structure-charm-code.md
+++ b/docs/howto/write-and-structure-charm-code.md
@@ -165,8 +165,10 @@ Arrange the methods of this class in the following order:
 1. An `__init__` method that observes all relevant events and instantiates objects.
    For example, in a Kubernetes charm:
    ```python
-   framework.observe(self.on["workload_container"].pebble_ready, self._on_pebble_ready)
-   self.container = self.unit.get_container("workload-container")
+   def __init__(self, framework: ops.Framework):
+       super().__init__(framework)
+       framework.observe(self.on["workload_container"].pebble_ready, self._on_pebble_ready)
+       self.container = self.unit.get_container("workload-container")
    ```
 2. Event handlers, in the order that they're observed in `__init__`.
    Make the event handlers private.
--- a/docs/howto/write-and-structure-charm-code.md
+++ b/docs/howto/write-and-structure-charm-code.md
@@ -174,8 +174,7 @@ Arrange the methods of this class in the following order:
 2. Event handlers, in the order that they're observed in `__init__`.
    Make the event handlers private.
    For example, `def _on_pebble_ready(...)` instead of `def on_pebble_ready(...)`.
-3. Other helper methods, which can be public. If you're writing a charm that will be extended
-   by other charms, you could use private helper methods instead.
+3. Other helper methods, which may be private or public.
 
 ```{admonition} Best practice
 :class: hint
--- a/docs/howto/write-and-structure-charm-code.md
+++ b/docs/howto/write-and-structure-charm-code.md
@@ -229,6 +229,9 @@ class DemoServerCharm(ops.CharmBase):
         self.unit.set_workload_version(version)
         self.unit.status = ops.ActiveStatus()
 
+    # Put helper methods here.
+    # If a method doesn't depend on Ops, put it in src/demo_server.py instead.
+
 
 if __name__ == "__main__":  # pragma: nocover
     ops.main(DemoServerCharm)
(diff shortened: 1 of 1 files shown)


dwilding:
@james-garner-canonical, @tonyandrewmeyer, this PR includes some clarifications to the Ops docs about when to use private vs public methods (plus some other improvements), following on from our recent discussions. Please could you review when you have time? Thanks!

docs/howto/write-and-structure-charm-code.md
```diff
@@ -157,29 +157,114 @@ control, so that exact versions of charms can be reproduced.
...
+   def __init__(self, framework: ops.Framework):
+       super().__init__(framework)
+       framework.observe(self.on["workload_container"].pebble_ready, self._on_pebble_ready)
+       self.container = self.unit.get_container("workload-container")
+   ```
+2. Event handlers, in the order that they're observed in `__init__`.
+   Make the event handlers private.
+   For example, `def _on_pebble_ready(...)` instead of `def on_pebble_ready(...)`.
+3. Other helper methods, which can be public. If you're writing a charm that will be extended
+   by other charms, you could use private helper methods instead.
```
james-garner-canonical, line 178: I feel like we could be clearer here what exactly we recommend and whether it matters.

Maybe we want something like:
```suggestion
3. Other helper methods, which should be public. If you're writing a charm class designed for subclassing
   by other charms, you should instead make all the methods that subclasses shouldn't rely on private.
```

I suggest waiting for Tony's recommendation on this one tho.

james-garner-canonical (approved):

docs/howto/write-and-structure-charm-code.md
```diff
@@ -157,29 +157,114 @@ control, so that exact versions of charms can be reproduced.
 (design-your-python-modules)=
 ## Design your Python modules
 
-In your `src/charm.py` file, define the charm class and manage the charm's interface
-with Juju: set up the event observation and add the handlers. For each workload
-that the charm manages, add a `{workload}.py` file that contains methods for
-interacting with the workload.
+In your `src/charm.py` file, define a class for managing the charm's interface with Juju.
```
tonyandrewmeyer, line 160: I realise this was there previously, but I wonder if we should avoid the word "interface", since that has a specific meaning in Juju. Maybe "interaction"?

docs/howto/write-and-structure-charm-code.md
```diff
@@ -157,29 +157,114 @@ control, so that exact versions of charms can be reproduced.
...
+        demo_server.install()
+
+    def _on_start(self, event: ops.StartEvent):
+        """Handle start event."""
+        self.unit.status = ops.MaintenanceStatus("starting server")
+        demo_server.start()
+        version = demo_server.get_version()
+        self.unit.set_workload_version(version)
+        self.unit.status = ops.ActiveStatus()
+
```
tonyandrewmeyer, line 232: I wonder if we should have a helper function here just to make the example more complete. I'm not sure what it would be, though.

docs/howto/write-and-structure-charm-code.md
tonyandrewmeyer: I feel like what we are trying to say is "other than event handlers, use your own judgement for public/private names". A charm class isn't intended for use in other code, so questions about what constitutes the public API don't really make sense, because there is no public API. For a machine/k8s base class, that's almost certainly within a team, so should be decided internally. For other special-cases like the 12 factor charms, they feel too specific to be called out here.

What about just not telling people how to name their other names? Do people need guidance on this?

tonyandrewmeyer (approved):

docs/howto/write-and-structure-charm-code.md
dwilding: Great point. Changed in [`bb5d761`](https://github.com/canonical/operator/pull/1670/commits/bb5d761e3af516a6c73cf4aab53a7b33ce8a0024)

docs/howto/write-and-structure-charm-code.md
dwilding: I agree, but I can't think of one either. I don't really want to artificially make the example harder to understand. What do you think about this comment that I just added?

```py
...

        """Handle start event."""
        self.unit.status = ops.MaintenanceStatus("starting server")
        demo_server.start()
        version = demo_server.get_version()

    # Put helper methods here.
    # If a method doesn't depend on Ops, put it in src/demo_server.py instead.


    ops.main(DemoServerCharm)
```

tonyandrewmeyer: Sure, looks good.

Issue #1594, opened 2025-02-28
Add support for Pebble identities


benhoyt:
We should add Ops support for Pebble identity-related APIs: add, update, replace, and remove.

Design decisions:

* Function names: probably `Pebble` and `Container` methods matching those on the Go client: `get_identities`, `add_identities`, `replace_identities`, `remove_identities`.
* Or do we just want to add `get_identities` and `replace_identities` for now, since replacing is idempotent (which is what we want for charms) and also allows removal?
* Should the argument be `Mapping[str, pebble.Identity]`, strongly typed dataclasses matching the Go [client.Identity](https://pkg.go.dev/github.com/canonical/pebble@v1.19.0/client#Identity) type? In which case we'll need to add new types if we add new identity types, but that's probably okay (happens very slowly).

Typical call might look like this:

```python
hashed = subprocess.run(['openssl', 'passwd', '-6'], capture_output=True, check=True).stdout
model.add_secret(hashed_password) # pseudo-code (but store it somewhere)
container.replace_identities({
    "web": pebble.Identity(access=pebble.MetricsAccess, basic=pebble.BasicIdentity(password=hashed)),
})
```

For documenting the above, generated API reference docs is probably fine. We should also write a doc for how to use them (with /v1/metrics) in a charm -- I'll open a separate issue for that.

benhoyt:
Keeping in 25.04 for now, but if it bumps to 25.10 that's okay too.

james-garner-canonical:
Idea: don't provide the passthrough methods on `ops.Container` unless there's a compelling reason to (argument typing, for example?), and just let our users call `container.pebble.{add,get,replace,remove}_identities`.

IronCore864:
> Idea: don't provide the passthrough methods on `ops.Container` unless there's a compelling reason to (argument typing, for example?), and just let our users call `container.pebble.{add,get,replace,remove}_identities`.

In the beginning, I thought we didn't need to have `container.get_identities`, but after a closer look, it seems the container has many methods that are already on the pebble client, so I added them to the container.

That said, we should not add them to the container because, unlike push/pull/get_plan/add_layer/etc, identities don't logically belong to the container.

I still have them on the container in my [first draft here](https://github.com/canonical/operator/pull/1672).

Issue #1595, opened 2025-02-28
Add how-to for using Pebble's metrics endpoint in charm


benhoyt:
After Ops has support for identities (https://github.com/canonical/operator/issues/1594), you'll be able to add a "metrics" identity to allow Prometheus or whatever tool to hit Pebble's metrics endpoint. But we should write a how-to explaining the various steps involved: generating a password, using `Container.replace_identities`, adding a secret, and setting up Prometheus with the secret.

PR #1672, opened 2025-04-09
feat: pebble identities

IronCore864:
Add support to Pebble identities. See [here](https://github.com/canonical/operator/issues/1594) for design-related discussions.

Details:
- Only add `get_identities`, `remove_identities`, and `replace_identities`. The latter is idempotent and can be used for upsert/delete.
- Support both typed dict and strongly typed dataclasses according to our discussion.

Closes https://github.com/canonical/operator/issues/1594.


ops/model.py
```diff
@@ -3152,6 +3152,33 @@ def get_notices(
             keys=keys,
         )
 
+    def get_identities(self) -> Dict[str, pebble.Identity]:
```
benhoyt, line 3155: Let's go with James's suggestion and not add these to Container for now. I don't think they'll be commonly-used.

ops/pebble.py
```diff
@@ -563,6 +576,10 @@ def __str__(self):
         return message
 
 
+class IdentityError(Error):
```
benhoyt, line 579: I see your note about why you added this, but realistically the server will never return a null identity, so let's remove this error and just call from_dict and let it raise whatever it raises.

ops/pebble.py
```diff
@@ -181,6 +181,19 @@
     total=False,
 )
 
+LocalIdentityDict = typing.TypedDict('LocalIdentityDict', {'user-id': int})
+BasicIdentityDict = typing.TypedDict('BasicIdentityDict', {'password': str})
+
+IdentityDict = typing.TypedDict(
+    'IdentityDict',
+    {
+        'access': str,
```
benhoyt, line 190: What about using `Literal['untrusted', 'metrics', 'read', 'admin']`?

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+
+@dataclasses.dataclass
+class Identity:
+    """Pebble identity configuration."""
+
+    access: Union[IdentityAccess, str]
+    local: Optional[LocalIdentity] = None
+    basic: Optional[BasicIdentity] = None
+
+    def __post_init__(self) -> None:
```
benhoyt, line 2030: This is a nice touch, thanks.

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    def from_dict(cls, d: BasicIdentityDict) -> BasicIdentity:
+        """Create new BasicIdentity from dict parsed from JSON."""
+        return cls(password=d['password'])
+
+
+@dataclasses.dataclass
+class Identity:
+    """Pebble identity configuration."""
+
+    access: Union[IdentityAccess, str]
```
benhoyt, line 2026: I think for new types we should just make this `access: IdentityAccess`. We've regretted using the union for existing types.

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
         return self.read(n)
 
 
+class IdentityAccess(enum.Enum):
```
benhoyt, line 1989: How about we make this a StrEnum, but that's not available on Python <3.10, so you can [do this](https://stackoverflow.com/questions/75040733/is-there-a-way-to-use-strenum-in-earlier-python-versions/75040847):

```suggestion
class IdentityAccess(str, enum.Enum):
```

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    """Enum of identity access levels."""
+
+    ADMIN = 'admin'
+    READ = 'read'
+    METRICS = 'metrics'
+    UNTRUSTED = 'untrusted'
+
+
+@dataclasses.dataclass
```
benhoyt, line 1998: For most other dataclasses, we've used `frozen=True`. What do you think?

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    def __post_init__(self) -> None:
+        """Validate that at least one of local or basic is provided."""
+        if self.local is None and self.basic is None:
+            raise ValueError('at least one of "local" or "basic" must be provided')
+
+    @classmethod
+    def from_dict(cls, d: IdentityDict) -> Identity:
+        """Create new Identity from dict parsed from JSON."""
+        if 'access' not in d:
+            raise ValueError('"access" key is required in IdentityDict')
```
benhoyt, line 2039: In other similar methods, we've just let the `KeyError` when accessing `d['access']` below be raised, so I think we can drop this.

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+
+        try:
+            access = IdentityAccess(d['access'])
+        except ValueError:
```
benhoyt, line 2043: We'll be able to drop this try/except too, if we get rid of the union type.

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+            access = d['access']
+
+        local = (
+            LocalIdentity.from_dict(d['local'])
+            if 'local' in d and d['local'] is not None
```
benhoyt, line 2048: Equivalently:
```suggestion
            if d.get('local')
```

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+            if 'basic' in d and d['basic'] is not None
+            else None
+        )
+
+        return cls(access=access, local=local, basic=basic)
+
+    def to_dict(self) -> IdentityDict:
+        """Convert this identity its dict representation."""
+        result: Dict[str, Any] = {
+            'access': self.access.value if isinstance(self.access, IdentityAccess) else self.access
```
benhoyt, line 2062: And without the union, and with the str-enum, this will just be `'access': self.access`

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+            BasicIdentity.from_dict(d['basic'])
+            if 'basic' in d and d['basic'] is not None
+            else None
+        )
+
+        return cls(access=access, local=local, basic=basic)
+
+    def to_dict(self) -> IdentityDict:
+        """Convert this identity its dict representation."""
+        result: Dict[str, Any] = {
```
benhoyt, line 2061: Could the type of this be `IdentityDict` to avoid the cast further down?

ops/pebble.py
```diff
@@ -1969,6 +1986,88 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+            'access': self.access.value if isinstance(self.access, IdentityAccess) else self.access
+        }
+        if self.local is not None:
+            result['local'] = {'user-id': self.local.user_id}
```
benhoyt, line 2065: Even though they're small, these should probably have parallel `to_dict` methods on the individual types, eg `LocalIdentity.to_dict`.

ops/pebble.py
```diff
@@ -3211,6 +3310,49 @@ def get_notices(
...
+        """
+        resp = self._request('GET', '/v1/identities')
+        result = resp['result']
+
+        identities: Dict[str, Identity] = {}
+        for name, identity in result.items():
+            if identity is None:
+                raise IdentityError(f'server returned null identity {name!r}')
+            identities[name] = Identity.from_dict(identity)
+        return identities
```
benhoyt, line 3330: Seeing we're removing IdentityError, this can probably become a dict comprehension:
```suggestion
        return {name: Identity.from_dict(d) for name, d in result.items()}
```

ops/pebble.py
```diff
@@ -3211,6 +3310,49 @@ def get_notices(
...
+        identities_dict: Dict[str, Optional[IdentityDict]] = {}
+        for name, identity in identities.items():
+            if identity is None or isinstance(identity, dict):
+                identities_dict[name] = identity
+            elif isinstance(identity, Identity):
+                identities_dict[name] = identity.to_dict()
+            else:
+                raise TypeError(
+                    f'identity must be a dict or pebble.Identity, not {type(identity).__name__}'
+                )
```
benhoyt, line 3351: I'd probably simplify by inverting the if/else logic and dropping the explicit TypeError for simplicity (we rely on static type checking a lot):

```suggestion
        identities_dict = {
            name: identity.to_dict() if isinstance(identity, Identity) else identity
            for name, identity in identities.items()
        }
```

benhoyt (commented):
Looks like a good start. I do think we should go with James's suggestion and just add these to the `pebble.Client` class, at least for now.

ops/pebble.py
IronCore864: This is because previously I set `total=False` for the typed dict, since the `local`/`basic` keys are not mandatory, it's possible for either of them to be empty. But I think I can set `total=True` then make `local`/`basic` optional.

ops/pebble.py
IronCore864: This is used to remove type check error, because `d['local']` is typed as `Optional[LocalIdentityDict]` but `LocalIdentity.from_dict()` expects a non-None dict.

ops/pebble.py
IronCore864: This is because `total=False` for `IdentityDict`, since the `local`/`basic` keys are not mandatory, it's possible for either of them to be empty. And if `access` is not mandatory, `d['access']` would cause an error `Could not access item in TypedDict
  "access" is not a required key in "IdentityDict", so access may result in runtime exceptionPylance[reportTypedDictNotRequiredAccess](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportTypedDictNotRequiredAccess)`

IronCore864: I set `total=False` for `IdentityDict`, because `local`/`basic` is not mandatory, one of them can be empty. Since `total=False`, `access` is not a required key in `IdentityDict` any more, so here is only to make sure the below `d['access']` won't cause static check error (access may result in runtime exception (reportTypedDictNotRequiredAccess))

ops/pebble.py
IronCore864: See the comment above, this is because `total=False`, if I type it as `IdentityDict`, here I must set basic/local as well. Type casting is slightly better than setting them to None/empty dict then deleting the key later if they are empty.

ops/pebble.py
IronCore864: This was to follow the same logic/coding style in `add_layer`, but I do agree that the suggested code looks much cleaner.

ops/pebble.py
IronCore864: Since `get_identities` returns objects, and `replace_identities` can take objects, if not frozen, users can get -> modify -> replace easily:

```python
identities = client.get_identities()
identities['bob'].local.user_id = 1000  # update
client.replace_identities(identities)  # replace
```

If it's frozen, users would have to `to_dict`, modify the dict, then `from_dict`, then replace.

Is this a legit use case?

test/test_pebble.py
benhoyt, line 1: It would be good to add a couple of tests to `test_real_pebble.py` against a real Pebble daemon.

test/test_pebble.py
```diff
@@ -3185,6 +3185,73 @@ def test_get_notices_filters(self, client: MockClient):
...
+        identities = client.get_identities()
+        assert len(identities) == 2
+        assert identities['alice'] == pebble.Identity.from_dict({
+            'access': 'admin',
+            'local': {'user-id': 42},
+        })
+        assert identities['web'] == pebble.Identity.from_dict({
+            'access': 'metrics',
+            'basic': {'password': 'hashed password'},
+        })
```
benhoyt, line 3208: Would be simpler, and you could avoid the `len == 2` check, to assert on the result dict directly:
```suggestion
        assert identities == {
             'alice': pebble.Identity.from_dict({
                 'access': 'admin',
                 'local': {'user-id': 42},
             },
             'web': pebble.Identity.from_dict({
                 'access': 'metrics',
                 'basic': {'password': 'hashed password'},
             },
        }
```

test/test_pebble.py
```diff
@@ -3185,6 +3185,73 @@ def test_get_notices_filters(self, client: MockClient):
...
+        ]
+
+    def test_replace_identities(self, client: MockClient):
+        client.responses.append({
+            'status': 'OK',
+            'status-code': 200,
+            'type': 'sync',
+        })
+        identities: typing.Dict[str, typing.Union[pebble.IdentityDict, pebble.Identity, None]] = {
+            'web': {'access': 'metrics', 'basic': {'password': 'hashed password'}},
```
benhoyt, line 3221: Would be good to have one entry that's an `Identity`, one that's an `IdentityDict`, and one that's `None`, to test each of those cases.

test/test_pebble.py
```diff
@@ -3185,6 +3185,73 @@ def test_get_notices_filters(self, client: MockClient):
...
+                {
+                    'action': 'replace',
+                    'identities': {
+                        'web': {'access': 'metrics', 'basic': {'password': 'hashed password'}},
+                    },
+                },
+            ),
+        ]
+
+        client.responses.append({
```
benhoyt, line 3238: Why do we need to test `get_identities` again in `test_replace_identities`?

test/test_pebble.py
```diff
@@ -3893,3 +3960,54 @@ def recv():
...
+        assert identity.to_dict() == {
+            'access': 'metrics',
+            'basic': {'password': 'hashed password'},
+        }
+
+    def test_no_access(self):
+        with pytest.raises(ValueError) as excinfo:
+            raw: pebble.IdentityDict = {'local': {'user-id': 42}}
+            pebble.Identity.from_dict(raw)
+        assert str(excinfo.value) == '"access" key is required in IdentityDict'
```
benhoyt, line 4002: Let's not test the exception message, just that `ValueError` is raised. Similar below.

ops/_private/harness.py
```diff
@@ -4048,3 +4048,11 @@ def _notice_matches(
         if keys is not None and notice.key not in keys:
             return False
         return True
+
```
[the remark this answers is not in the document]
tonyandrewmeyer: Identities belong to a specific Pebble, so we'd add them to `Container` and then implement the pebble methods in the mock pebble client. That's mostly done in Harness (and inherited by Scenario) but if we want to avoid Harness supporting them could be just in Scenario.

ops/pebble.py
```diff
@@ -1969,6 +1982,91 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
         return self.read(n)
 
 
+class IdentityAccess(str, enum.Enum):
```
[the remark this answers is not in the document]
tonyandrewmeyer: I'd lean towards consistency. It's probably not going to change how people use them in practice, but it is part of the public API for the type, and having one work differently seems bad.

ops/_private/harness.py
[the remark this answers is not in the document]
benhoyt: Thanks. @IronCore864 I think we should do this, but I'm happy to do it in a subsequent PR to keep things moving.

test/test_pebble.py
IronCore864: After adding a real pebble test and using the API in the tests, I also made a small change, which is the type annotation of the param of `replace_identity` (from `Dict` to `Mapping`), because otherwise after calling `get_identities` which returns `Dict[str, Identity]` and updating some of the content, it can't be used directly in `replace_identity` without type-casting:

```
Argument of type "dict[str, Identity]" cannot be assigned to parameter "identities" of type "Dict[str, IdentityDict | Identity | None]" in function "replace_identities"
  "dict[str, Identity]" is not assignable to "Dict[str, IdentityDict | Identity | None]"
    Type parameter "_VT@dict" is invariant, but "Identity" is not the same as "IdentityDict | Identity | None"
    Consider switching from "dict" to "Mapping" which is covariant in the value typePylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

And I think calling get, update, then typecast before replace could be some unnecessary overhead for the users, hence the above-mentioned change.

ops/pebble.py
```diff
@@ -1969,6 +1983,95 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    def to_dict(self) -> LocalIdentityDict:
+        """Convert this local identity to its dict representation."""
+        return {'user-id': self.user_id}
+
+
+@dataclasses.dataclass
+class BasicIdentity:
+    """Basic identity configuration (for HTTP basic authentication)."""
+
+    password: str
```
tonyandrewmeyer, line 2015: Currently, the structure of the classes and the docs make it seem like you'll get a hash here when you do a query to Pebble to get identities, which would be bad. You actually get five `*`, which is good, but not explained anywhere here.

Is there any point having an attribute that will be `*****` for every identity? It's not a password, and it's not even a salted hash, it's just a fixed string. I don't see how this will be useful to anyone retrieving identities.

I feel this is worse when adding/updating identities get mixed in, because the same dict field is being used for a static mask string and also a salted hash - one has no security value at all and one is extremely high value, and they are being mixed into the same location.

ops/pebble.py
```diff
@@ -3211,6 +3314,34 @@ def get_notices(
...
+        """
+        resp = self._request('GET', '/v1/identities')
+        result = resp['result']
+        return {name: Identity.from_dict(d) for name, d in result.items()}
+
+    def replace_identities(
+        self, identities: Mapping[str, Union[IdentityDict, Identity, None]]
+    ) -> None:
+        """Replace the named identities in Pebble with the given ones.
+
```
tonyandrewmeyer, line 3331: This "don't have a remove method" decision seems like it's "because we don't have to have one". Is there more to it than that? I don't like that the Pebble CLI is more user-friendly than the Python client. It seems like it would be trivial to have a `remove_identities` method that just called replaced under the hood with appropriate None values, and that would be user-friendly and not seem like exposing an implementational detail.

tonyandrewmeyer (changes requested):
I have concerns about how the identity objects are being used to both hold sensitive and retrieve (mostly) non-sensitive data. I think we need to think more about how the API will be used, and potentially reconsider some of this.

ops/_private/harness.py
```diff
@@ -4048,3 +4048,11 @@ def _notice_matches(
...
+    def get_identities(self) -> Dict[str, pebble.Identity]:
+        raise NotImplementedError(self.get_identities)
```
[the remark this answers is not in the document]
benhoyt: Yeah, that's super-weird, but it's like the others. I've removed this in https://github.com/canonical/operator/commit/b55453e9d5e3c109326f65fc445867fdefb8372b

ops/pebble.py
```diff
@@ -1969,6 +1983,95 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    ADMIN = 'admin'
+    READ = 'read'
+    METRICS = 'metrics'
+    UNTRUSTED = 'untrusted'
+
+
+@dataclasses.dataclass
+class LocalIdentity:
+    """Local identity configuration (for ucrednet/UID authentication)."""
```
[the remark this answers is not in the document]
benhoyt: Good point. We call it "peer credential" and "UID" in our docs. Maybe `(for Unix socket / UID authentication)`?

ops/pebble.py
benhoyt: Hmm, yes, I think these are valid questions. However, this is the way the Pebble API works, so I think the method should and these types probably reflect that. However, we should document it better, something like "When writing, this is a SHA512-crypt hashed password. When reading, it is returned as `*****`."

ops/pebble.py
benhoyt: Yeah, that's fair. It's not hard to add `remove_identities` (though I'd have it call the "remove" Pebble API rather than going via replace).

@tonyandrewmeyer Do you think we should also add `add_identities` and `update_identities` too, for completeness? I think those are almost always better done with `replace_identities`, especially in a charm context, where you want idempotency.

tonyandrewmeyer: I wouldn't object to `add` and `update` but I don't feel those are as necessary. A single method for add/update seems clean and not-common, and indeed there's some idempotency benefits. It was specifically remove having the special "put in None" that felt like bad DX to me.

test/test_real_pebble.py
```diff
@@ -308,6 +308,43 @@ def test_log_forwarding(self, client: pebble.Client):
...
         assert plan.log_targets['pretend-loki'].labels == {'foo': 'bar'}
 
+    def test_identities(self, client: pebble.Client):
+        identities = client.get_identities()
+        assert len(identities) == 0
+
+        identities = {
+            'web': pebble.Identity.from_dict({
+                'access': 'metrics',
+                'basic': {'password': 'hashed password'},
```
benhoyt, line 318: Per discussion, let's make this an actual `$6$...` style SHA512 crypt hash.

ops/pebble.py
IronCore864: Per discussion in the daily, we decided to add `remove_identities` with the parameter being a set of strings (names of the identities to be removed). I will implement this and add some unit tests and adjust the real pebble test accordingly.

dimaqq:
Adjacent to the proposed change, here's a concern Tiexin and I discovered during out 1-1.

We're making this change so that charmers create a user that the telemetry stack can log into pebble over http/tcp with to read the metrics data.

For that to be useful, the charm needs to determine the pod hostname `socket.getfqdn()` and the port that the specific workload container's Pebble listens on.

It seems that Juju already pass unique port numbers to Pebble to the different containers in the pod:

```
# charm
root           1       0  0 Apr06 ?        00:05:30 /charm/bin/pebble run --http :38812 --verbose

# workload
root           1       0  0 Apr06 ?        00:07:32 /charm/bin/pebble run --create-dirs --hold --http :38813 --verbose
```

How should the charm read this port number though?

There doesn't seem to be a Pebble CLI command that returns this port number.
I couldn't find the port number in the API either.

My current best hack is this:
```py
>>> container.pull("/proc/1/cmdline").read().split("\0")
['/charm/bin/pebble', 'run', '--create-dirs', '--hold', '--http', ':38813', '--verbose', '']
```

benhoyt:
@dimaqq Yeah, that's more relevant for the [upcoming metrics how-to](https://github.com/canonical/operator/issues/1595), but it's a good point. I've asked Harry on Matrix > Juju.

ops/pebble.py
```diff
@@ -3211,6 +3317,44 @@ def get_notices(
         resp = self._request('GET', '/v1/notices', query)
         return [Notice.from_dict(info) for info in resp['result']]
 
+    def get_identities(self) -> Dict[str, Identity]:
```
dimaqq, line 3320: Please merge latest `main` (or rebase, whichever) and add instrumentation on this the the other two methods.

ops/_private/harness.py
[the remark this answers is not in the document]
dimaqq: Should I take it that there's going to be a separate PR for `ops.testing`?

ops/pebble.py
```diff
@@ -181,6 +182,19 @@
     total=False,
 )
 
+LocalIdentityDict = typing.TypedDict('LocalIdentityDict', {'user-id': int})
+BasicIdentityDict = typing.TypedDict('BasicIdentityDict', {'password': str})
+
+IdentityDict = typing.TypedDict(
```
dimaqq, line 188: This is a bit confusing, could we have better names?

I made the mistake of thinking that `IdentityDict` is like a base class for `LocalIdentityDict`. I imagine some charmers may make the same mistake.

ops/pebble.py
```diff
@@ -3211,6 +3317,44 @@ def get_notices(
...
+
+        Returns:
+            A dict mapping identity names to :class:`Identity` objects.
+        """
+        resp = self._request('GET', '/v1/identities')
+        result = resp['result']
+        return {name: Identity.from_dict(d) for name, d in result.items()}
+
+    def replace_identities(
+        self, identities: Mapping[str, Union[IdentityDict, Identity, None]]
```
dimaqq, line 3331: Which do we propose the charmers to use: `IdentityDict` or `Identity`?

If it's the earlier, I'd be against exposing `'access': Literal['untrusted', 'metrics', 'read', 'admin']` to our charmers.

Recall that a literal is not a constant, it's a verbatim string.

For example, the following will be rejected by the type checker and there's no decent work-around:

```py
id1 = json.loads(json.dumps({"access": "read", ...}))
```

If some charm gets identities over a relation, they won't be able to apply them to pebble in a way that static type checker recognises.

Only hard-coded identities will be OK, and even that with some caveats.

And if the recommended way is `Identity` dataclass, then that should go first.

dimaqq (changes requested):
This PR needs some example how the feature is supposed to be used.

That could be a test charm, perhaps later used for integration tests.

Or it could be a code sample in documentation.

ops/pebble.py
```diff
@@ -181,6 +182,19 @@
     total=False,
 )
 
+LocalIdentityDict = typing.TypedDict('LocalIdentityDict', {'user-id': int})
+BasicIdentityDict = typing.TypedDict('BasicIdentityDict', {'password': str})
```
dimaqq, line 186: Is this missing the `username` field?

https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit?tab=t.0

dimaqq: Ah I think I see it now, the `username` is implied from the identities **dictionary key**.
Right?

This is a bit... unconventional.
I understand that Golang implementation doing that way.
A YAML config file is likewise OK.
That use in the Pebble HTTP API is a bit questionable...
And having this in the Python API gives me the jibbies.

It's a bit of a slippery slope.
@benhoyt WDYT?

ops/pebble.py
benhoyt: I think the existing names are okay, as they match the [Pebble Go client's naming](https://pkg.go.dev/github.com/canonical/pebble@v1.20.0/client#Identity): `Identity` is the "root", and then we have `LocalIdentity` and `BasicIdentity` (plus the "Dict" suffix for the plain dict types).

ops/pebble.py
benhoyt: Yes, I'd originally proposed having the `username` field explicit, but we decided in spec review to not have it for now, and just use the dict key. We would potentially add it as a field later (at the Pebble level) if users wanted the basic auth username to differ from the identity name, but for now they have to be the same.

So I'd like to stick with the Pebble semantics here in Python too.

benhoyt:
@dimaqq Let's talk about what's remaining at daily sync. Note that we plan docs/examples in the upcoming [metrics how-to](https://github.com/canonical/operator/issues/1595) later.

ops/pebble.py
```diff
@@ -3211,6 +3317,44 @@ def get_notices(
         resp = self._request('GET', '/v1/notices', query)
         return [Notice.from_dict(info) for info in resp['result']]
 
+    def get_identities(self) -> Dict[str, Identity]:
+        """Get all identities in Pebble.
+
+        Returns:
+            A dict mapping identity names to :class:`Identity` objects.
+        """
```
tonyandrewmeyer, line 3325: We should add in a `.. jujuadded:: 3.6.x` with the right version here and for `replace_identities` and `remove_identities` - I don't think we need that for any of the typed dicts/dataclasses, just for the three `Client` methods.

ops/pebble.py
dimaqq: I'm against exposing literal types in the public API.

### The case against closed types

A future version of Pebble may introduce another access level. Consider a charm that reads identities from Pebble. If the `identity.access` is a closed type: an Enum (or a Literal, though type-checkers vary here), that allows the charmer to write a charm that may crash on the new value and still pass static analyser.

Would we pass the new value to the charmer as is on the read side?

Or would we filter out the identities with unknown, new access levels?

### The case against custom types

A charm uses a bunch of Python packages. Consider the case where ops exposes access level as `Literal`, Prometheus exposes a very similar access level as an `Enum`, MySQL exposes a different `Enum`, a charm lib exposes a different `Literal`.

Pretty soon, the charm code will be mostly translation from MySQL enums to Prometheus enums, from ops literals to charm lib literals, from literals to enums and back.

That's not what we want to force upon charmers, is it?

### Literal specifically is not ergonomic

The literal type hint requires the value to be written verbatim in the code. It does not mean a value that equals the declared constant.

```py
from typing import Literal

def foo(arg: Literal['untrusted', 'metrics', 'read', 'admin']) -> None: ...

# OK
foo("metrics")

# OK
arg = "metrics"
foo(arg)

# Not OK
args = ["metrics"]
foo(*args)

# Not OK
args = ["metrics"]
foo(args[0])
```

The support for type narrowing is pretty decent in Pyright, but not perfect; It's not uniform across Pyright and Mypy or in fact across versions of Pyright.

ops/pebble.py
IronCore864: As discussed in the daily, this has been resolved.

ops/pebble.py
IronCore864: As discussed in the daily, this has been resolved.

ops/pebble.py
benhoyt: > If the identity.access is a closed type: an Enum ... that allows the charmer to write a charm that may crash on the new value and still pass static analyser.

Ah yes, this is true and something I hadn't considered. If the upgrade Juju to a new version of Pebble that includes another type, `get_identities` as we have it now will raise an exception on this line:

```python
access = IdentityAccess(d['access'])
```

@james-garner-canonical We were going to add `UNKNOWN` values to our enums a while back -- did we decide not to do that in the end? What do you think about adding an `UNKNOWN` value here? Or if it's a `StrEnum` (see below), can we just set it to `'unknown'`?

On the type checking side, I don't mind the `Literal` -- if we add a new name and they want access to it when type checking, they can easily upgrade Ops to the latest. But on the runtime side I agree with you.

@IronCore864 In an earlier comment I suggested making this a `StrEnum` (or `(str, Enum)` as we're on Python 3.8) -- do you intend to update this? That helps prevent the "different types of enum" problem (though I don't think that's as big an issue as Dima indicates, as this is a Pebble-specific type).

james-garner-canonical: We were going to add `UNKNOWN` to the enums in #1400 (which aimed to eliminate str | enum union types), but opted not to (after walking back the scope of the PR due to backwards compatibility concerns). The concern with `UNKNOWN` specifically was round-trip safety -- that is, if we convert some new value 'foo' from Pebble into 'unknown' for our users, then if they send it back to Pebble it will error.

ops/pebble.py
```diff
@@ -181,6 +182,19 @@
...
 
+LocalIdentityDict = typing.TypedDict('LocalIdentityDict', {'user-id': int})
+BasicIdentityDict = typing.TypedDict('BasicIdentityDict', {'password': str})
+
+IdentityDict = typing.TypedDict(
+    'IdentityDict',
+    {
+        'access': Literal['untrusted', 'metrics', 'read', 'admin'],
+        'local': Optional[LocalIdentityDict],
+        'basic': Optional[BasicIdentityDict],
```
james-garner-canonical, line 193: Should these be `Optional` or `NotRequired`?

james-garner-canonical: I see we have `total=False` below -- is the intention here that 'access' is required, but the others are not? Currently all keys are optional, and 'local' and 'basic' can be `None` or their respective typed dicts.

ops/pebble.py
```diff
@@ -181,6 +182,19 @@
     total=False,
 )
 
+LocalIdentityDict = typing.TypedDict('LocalIdentityDict', {'user-id': int})
```
james-garner-canonical, line 185: I also wonder if these new `TypedDict` should be defined only at `TYPE_CHECKING` time. This may somewhat address @dimaqq's concern about naming, as it makes it clear they're not for use as runtime types / inheritance.

ops/pebble.py
james-garner-canonical: Note that `dataclass(frozen=True)` does support this pattern with `dataclasses.replace`, but it's pretty unwieldy with nested structures like these (example below), so I agree that not using `frozen` here seems like the right call
```py
identities = client.get_identities()
bob = identities['bob']
new_local = dataclasses.replace(bob.local, user_id=1000)
identities['bob'] = dataclasses.replace(bob, local=new_local)
client.replace_identities(identities)  # replace
```

ops/pebble.py
```diff
@@ -1973,6 +1987,98 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    user_id: int
+
+    @classmethod
+    def from_dict(cls, d: LocalIdentityDict) -> LocalIdentity:
+        """Create new LocalIdentity from dict parsed from JSON."""
+        return cls(user_id=d['user-id'])
+
+    def to_dict(self) -> LocalIdentityDict:
+        """Convert this local identity to its dict representation."""
+        return {'user-id': self.user_id}
```
james-garner-canonical, line 2012: Sorry if this has already been litigated, but do we need `to_dict` methods? `dataclasses.asdict` should do the same thing, though it won't be typed as `LocalIdentityDict` -- but if this is primarly for internal use then that's probably fine, and I suspect we want users to work with `LocalIdentity` objects instead of converting them to dicts.

ops/pebble.py
james-garner-canonical: I think this is a case that would be cleaner if we used `Required` / `NotRequired` instead of `total=False` and `Optional` for the `TypedDict`s.

ops/pebble.py
james-garner-canonical: Tiexin reminded me in conversation that we use `to_dict` to (1) convert names between snake and kebab case, and (2) filter out Falsey values, so I think we 100% want to keep these.

ops/pebble.py
IronCore864: @benhoyt regarding `(str, Enum)` -- in that discussion, Tony thought keeping it consistent with the existing code base, which doesn't do this, is more important, so I didn't change it. Shall we do it?

ops/pebble.py
dimaqq: From stand-up: we can always reorder options in a Union. closing this thread for now.

dimaqq:
pls update harness and tests: `AttributeError: module 'ops.pebble' has no attribute 'IdentityDict'`

dimaqq:
You may want to update the ops and ops-scenario (if any) in the tox environment:

```
[testenv:benchmark]
description = Run benchmark tests
passenv =
    RUN_REAL_PEBBLE_TESTS
    PEBBLE
deps =
    PyYAML==6.*
    websocket-client==1.*
    pytest~=7.2
    pytest-benchmark~=5.0
    typing_extensions~=4.2
    -e .
    -e testing
    -e tracing
commands =
    pytest -v --tb native \
        {[vars]tst_path}benchmark \
        {[vars]testing_tst_path}benchmark \
        {posargs}

[testenv:pebble]
description = Run real pebble tests
allowlist_externals = pebble
                      mkdir
                      bash
setenv =
  PEBBLE=/tmp/pebble
  RUN_REAL_PEBBLE_TESTS=1
deps =
    PyYAML==6.*
    websocket-client==1.*
    coverage[toml]~=7.0
    pytest~=7.2
    typing_extensions~=4.2
commands =
    bash -c "umask 0; (pebble run --http=':4000' --create-dirs &>/dev/null & ) ; sleep 1; pytest -v --tb native test/test_real_pebble.py {posargs} ; killall -y 3m pebble"
```

Note how the benchmarks and unit test were updated to use `-e .`
Maybe that's the problem.

ops/pebble.py
benhoyt: Sorry, I'd totally forgotten about that. In any case, I do think there's a real problem here that Dima's surfaced. Let's discuss as a team next Tuesday.

ops/jujuversion.py
```diff
@@ -127,6 +127,14 @@ def has_controller_storage(self) -> bool:
         """Report whether this Juju version supports controller-side storage."""
         return (self.major, self.minor, self.patch) >= (2, 8, 0)
 
+    def has_pebble_identities(self) -> bool:
```
benhoyt, line 130: Thanks for adding these. We've been inconsistent in naming these over time (and whether they're functions or properties). However, I think we should follow the more recent convention: move these to the button, call them `supports_x` instead of `has_x`, and make them properties instead of methods.

james-garner-canonical:
Idea:
- Type the new `pebble.Client` methods as accepting only `Identity`, not `IdentityDict`.
- Make all the new `{to,from}_dict` methods private.

This makes the new `TypedDict`s purely an internal implementation detail, so they can be private too.

This in turn means we don't need to expose the `Literal` of access strings to users. Instead we only expose the `IdentityAccess` enum. We can add `IdentityAccess.UNKNOWN` to future proof against reading new values from newer Pebbles.

The new `Identity` type feels a little more complicated than users might expect, with the nested `LocalIdentity` and `BasicIdentity` types. I think it would be nice if there was just one obvious way to work with them.

If it turns out that users really want to work with the dict versions, it's totally backwards compatible to later allow functions to also accept `IdentityDict`, or to expose `{to,from}_dict` methods publicly.

IronCore864:
Sample usage (for discussion about type checking):

```python
from ops.pebble import Client, IdentityAccess, IdentityDict

client = Client('/home/ubuntu/PEBBLE_HOME/.pebble.socket')

def _identity_to_removed() -> IdentityDict:
    return {
        "access": "admin",
        "local": {
            "user-id": 42,
        }
    }

def _identity_to_removed_2() -> IdentityDict:
    access = IdentityAccess.ADMIN.value
    return {
        "access": access,
        "local": {
            "user-id": 42,
        }
    }

client.replace_identities({
    'bob': _identity_to_removed()
})

client.replace_identities({
    'bob': _identity_to_removed_2()
})

client.replace_identities({
    'bob': {
        "access": IdentityAccess.ADMIN.value,
        "local": {
            "user-id": 42,
        }
    }
})

client.replace_identities({
    'bob': {
        "access": "admin",
        "local": {
            "user-id": 42,
        }
    }
})
```

ops/pebble.py
benhoyt: - update enum classes to `(str, Enum)` pattern so we avoid runtime exceptions -- catch enum init error and set field to `str`
- later, we can consider a nicer-typed API on `Container`

benhoyt:
We're decided to move `FooDict` types back to top level (not under `if TYPE_CHECKING`).

ops/pebble.py
benhoyt: The idea with the `str, Enum`-style enum was that we can now simplify this type to `access: IdentityAccess` as `IdentityAccess` is a subclass of `str`.

ops/pebble.py
```diff
@@ -1973,6 +1987,101 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    def __post_init__(self) -> None:
+        """Validate that at least one of local or basic is provided."""
+        if self.local is None and self.basic is None:
+            raise ValueError('at least one of "local" or "basic" must be provided')
+
+    @classmethod
+    def from_dict(cls, d: IdentityDict) -> Identity:
+        """Create new Identity from dict parsed from JSON."""
+        if 'access' not in d:
+            raise KeyError('"access" key is required in IdentityDict')
```
benhoyt, line 2055: I think we can drop this again, as `KeyError` will be raised by the `d['access']` expression below.

ops/pebble.py
```diff
@@ -1973,6 +1987,101 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+
+        try:
+            access = IdentityAccess(d['access'])
+        except ValueError:
+            access = d['access']
```
benhoyt, line 2060: Great, thanks!

test/test_pebble.py
```diff
@@ -3185,6 +3185,82 @@ def test_get_notices_filters(self, client: MockClient):
...
+                'alice': {'access': 'admin', 'local': {'user-id': 42}},
+                'web': {'access': 'metrics', 'basic': {'password': 'hashed password'}},
+            },
+            'status': 'OK',
+            'status-code': 200,
+            'type': 'sync',
+        })
+
+        identities = client.get_identities()
+        assert identities == {
```
benhoyt, line 3200: I think we should use the dataclass here, rather than `from_dict`, as we don't need to be testing that as well here.

I'd also suggest we add a `str` test to test the `(str, Enum)`-style enum:

```
assert identities['alice']['access'] == 'admin'
assert identities['web']['access'] == 'metrics'
```

benhoyt (changes requested):
Couple of minor changes requested, and then I think we're good to go.

ops/pebble.py
IronCore864: These two changes are to make pyright happy.

We could not drop `if 'access' not in d`, because `access` is not a required key in `IdentityDict` ([reportTypedDictNotRequiredAccess](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportTypedDictNotRequiredAccess)), also see [here](https://github.com/canonical/operator/pull/1672#discussion_r2036576287).

The reason for ` access: Union[IdentityAccess, str]` is that access might be a string, so pyright would complain when we do `return cls(access=access, ...)`.

benhoyt: Okay, yep, fine to leave the `raise KeyError` then.

On the type annotation -- yes, that's expected, but we want to simplify the typing for the user and pretend. So I think we want to silent Pyright and do this:

```python
try:
    access = IdentityAccess(d['access'])
except ValueError:
    access = typing.cast(IdentityAccess, d['access'])
```

benhoyt (approved):

ops/pebble.py
```diff
@@ -182,6 +182,19 @@
...
 
+LocalIdentityDict = typing.TypedDict('LocalIdentityDict', {'user-id': int})
+BasicIdentityDict = typing.TypedDict('BasicIdentityDict', {'password': str})
+
+IdentityDict = typing.TypedDict(
+    'IdentityDict',
+    {
+        'access': Literal['untrusted', 'metrics', 'read', 'admin'],
+        'local': Optional[LocalIdentityDict],
+        'basic': Optional[BasicIdentityDict],
```
james-garner-canonical, line 193: These should be `NotRequired`, right? I'm guessing we just went back to `Optional` by mistake when moving out of `typing.TYPE_CHECKING`.

It should be possible to do `NotRequired` here. For example, we could do the `class` form of `TypedDict` (since all the keys are valid Python identifiers), meaning the types would be in annotation scopes, so we could just stringify `NotRequired` (e.g. `'NotRequired[LocalIdentityDict]'`).

ops/pebble.py
```diff
@@ -2067,7 +2074,7 @@ def from_dict(cls, d: IdentityDict) -> Identity:
 
     def to_dict(self) -> IdentityDict:
         """Convert this identity to its dict representation."""
-        result: Dict[str, Any] = {'access': self.access.value}
+        result: Dict[str, Any] = {'access': self.access}
```
james-garner-canonical, line 2077: With `NotRequired` as suggested above, you should be able to use the `IdentityDict` annotation here and drop the cast.

ops/pebble.py
```diff
@@ -1973,6 +1987,101 @@ def read1(self, n: int = -1) -> Union[str, bytes]:
...
+    def to_dict(self) -> BasicIdentityDict:
+        """Convert this basic identity to its dict representation."""
+        return {'password': self.password}
+
+
+@dataclasses.dataclass
+class Identity:
+    """Pebble identity configuration."""
+
+    access: IdentityAccess
```
james-garner-canonical, line 2042: I think we might still need a union here tbh ... `IdentityAccess` is a `str`, yes, but a `str` is not an `IdentityAccess`.

Using `Union[IdentityAccess, str]` here would mean we could drop the `cast` in `from_dict`, and would allow users to write `Identity('admin', ...)` if that's what we want.

What using the `(str, Enum)` instead of `Enum` gets us is that an enum member and an unknown string value are both `str` and can be worked with the same way.

What do you think @benhoyt ?

benhoyt: I see what you mean. I was thinking we can pretend, but you're probably right that it's best to be accurate. Sorry about the flip-flop @IronCore864, but let's go back to `Union[IdentityAccess, str]` like James suggests. @james-garner-canonical are you happy with this PR with that change? (Tiexin has had to do a lot of back and forth, so let's only comment on any breaking concerns at this point.)

james-garner-canonical: I think we need to make sure the typing for `IdentityDict` is correct. Currently we have `total=False` and `local` and `basic` typed as `Optional`. So `{}` is a valid `IdentityDict`, and for some `IdentityDict` `d`, `d['local']` will be typed as `LocalIdentityDict | None`, which seems undesirable, and hard to fix in a backwards compatible way. I'd like to get this right -- working on it now.

ops/pebble.py
```diff
@@ -181,6 +182,19 @@
...
+
+IdentityDict = typing.TypedDict(
+    'IdentityDict',
+    {
+        'access': Literal['untrusted', 'metrics', 'read', 'admin'],
+        'local': Optional[LocalIdentityDict],
+        'basic': Optional[BasicIdentityDict],
+    },
+    total=False,
+)
```
james-garner-canonical, line 196: This is my suggestion to use `NotRequired`. I think we really do want these values to be `NotRequired` in the `TypedDict`, and I also think we don't want them to be `Optional`.

This change may require an additional ignore in tests (where we intentionally use a bad dict), and should let us drop the `cast` in `from_dict`.

However, I'm not 100% convinced that we have the type of `access` correct here and in `class Identity`, so more to come.

```suggestion

class IdentityDict(typing.TypedDict):
    """TypedDict for Pebble identy."""

    access: Literal['untrusted', 'metrics', 'read', 'admin']
    local: NotRequired[LocalIdentityDict]
    basic: NotRequired[BasicIdentityDict]
```

james-garner-canonical (changes requested):

ops/pebble.py
IronCore864: James' suggestions are merged in this PR, resolving this comment now.

james-garner-canonical (approved):

IronCore864:
All comments resolved, merging now, the failed pack test is due to known reasons.

tonyandrewmeyer (approved):

PR #1675, opened 2025-04-10
chore: dummy PR to test CI

dimaqq:
I have grafana-k8s unit tests failing...
differently in GHA and locally as well.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,4 +1,4 @@
-DUMMY CHANGE DONT'T MERGE
+DUMMY CHANGE DON'T MERGE
 
 # Setting up a Dev Environment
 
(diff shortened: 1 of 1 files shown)


dimaqq:
I believe that the 3rd party charm tests fail dues to this:
[before]   + pyright==1.1.398
                                                                                                                           [after]⌑   + pyright==1.1.399

dimaqq:
Upstream PR to pin/restrict pyright: https://github.com/canonical/grafana-k8s-operator/pull/414

We could possibly work around this in our CI:
- we want to replace the ops that the charm project uses (ok)
- but we don't want to re-lock other dependencies (somehow new pyright got used)

dimaqq:
cc @tonyandrewmeyer in case you have ideas about patching github ops into charm repos in some better way.

PR #1683, opened 2025-04-17
ci: don't pin release jobs to github environments

dimaqq:
The `ops` and `ops-scenario` release jobs don't specify the github (secret) environments, so let's remove these from the `ops-tracing` release jobs.

### DIFF
--- a/.github/workflows/publish-ops-scenario.yaml
+++ b/.github/workflows/publish-ops-scenario.yaml
@@ -14,6 +14,7 @@ jobs:
   build-n-publish:
     name: Build and Publish ops-scenario to PyPI
     runs-on: ubuntu-latest
+    environment: publish-pypi
     permissions:
       id-token: write
       attestations: write
--- a/.github/workflows/publish-ops-tracing.yaml
+++ b/.github/workflows/publish-ops-tracing.yaml
@@ -13,6 +13,7 @@ jobs:
     uses: ./.github/workflows/hello-charm-tests.yaml
   release:
     runs-on: ubuntu-latest
+    environment: publish-pypi
     permissions:
       id-token: write
       attestations: write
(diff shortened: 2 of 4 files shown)


benhoyt:
Does this need to be `publish-pypi` now for all three projects? Because I changed the environment for `ops` and `ops-scenario` in PyPI to `publish-pypi`? I'd forgotten one had to update the GitHub Actions YAML.

dimaqq:
> Does this need to be `publish-pypi` now for all three projects? Because I changed the environment for `ops` and `ops-scenario` in PyPI to `publish-pypi`? I'd forgotten one had to update the GitHub Actions YAML.

Only if some secret is set in that environment, like `PYPI_API_TOKEN`.
I figured that's not the case since the job doesn't reference any `${{ secrets.AAA }}`.

tonyandrewmeyer:
I think it would be better to standardise on all three using the `publish-pypi` environment rather than no environment. We can hopefully continue to avoid secrets (I think we could possibly hide the tiobe one from the publish workflow by putting a dummy one with the same name in the publish environment, but it doesn't seem like we need to protect it). But we might want to use the environment for other restrictions, like branch protection, in the future, I think.

Overall, I'm in favour of going more restrictive for the publishing workflows rather than less, even if we're not enforcing anything extra yet.

dimaqq:
OK, let me change that.

dimaqq:
I've changed this so:
- real publishing requires access to "publish-pypi" environment
- test publishing doesn't have restriction

PYPI settings:
- [x] ops-tracing
- [ ] ops-scenario (Ben pls check)
- [ ] ops (Ben pls check)

tonyandrewmeyer (approved):

benhoyt (approved):
Looks good to me, thanks. And I've double-checked the GitHub and PyPI settings and they match.

PR #1684, opened 2025-04-17
ci: don't run tests within the publish job

dimaqq:
Assuming that I understood Ben correctly about his earlier conversation with Tony, we have CI running in the PR context and on the main branch after each PR is merged.

Thus, we need not run the tests at the start of the publish job.

I'm not sure if this was about all tests or only integration tests.

I've added a step into the release process to wait after the release PR is merged to make sure tests are indeed passing.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -376,7 +376,7 @@ To make a release of the `ops` and/or `ops-scenario` packages, do the following:
    using the same Python version as specified in the `.readthedocs.yaml` file.
 12. Add, commit, and push, and open a PR to get the `CHANGES.md` update, version bumps,
    and doc requirement bumps into main (and get it merged).
-13. Wait until the tests pass after the PR is merged. CI should within 10 minutes.
+13. Wait until the tests pass after the PR is merged. It takes around 10 minutes.
    If the tests don't pass at the tip of the main branch, do not release.
 14. When you are ready, click "Publish". GitHub will create the additional tag.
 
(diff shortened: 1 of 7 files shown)


benhoyt (approved):
Thanks!

HACKING.md
```diff
@@ -376,7 +376,9 @@ To make a release of the `ops` and/or `ops-scenario` packages, do the following:
    using the same Python version as specified in the `.readthedocs.yaml` file.
 12. Add, commit, and push, and open a PR to get the `CHANGES.md` update, version bumps,
    and doc requirement bumps into main (and get it merged).
-13. When you are ready, click "Publish". GitHub will create the additional tag.
+13. Wait until the tests pass after the PR is merged. CI should within 10 minutes.
```
tonyandrewmeyer, line 379: I think there's a word missing here. "Complete" or "finish" maybe?

tonyandrewmeyer (commented):
Won't this fail on the "needs" line because we no longer include the jobs? It seems like the "needs" line should be cleared/removed as well, unless I'm missing something.

james-garner-canonical:
I'm unsure about this -- we're removing automatic tests blocking a bad release, in favour  of a manual step in the release process (waiting for tests to finish on `main`). Additionally, it seems like our tests only run on push to `main`, but what about releases from other branches, like the releases from `2.19-maintenance`?

dimaqq:
Re maintenance branches: those branches should have CI enabled on push to said branches.

HACKING.md
dimaqq: Rephrased, thanks!

dimaqq:
Good catch, @tonyandrewmeyer wrt. "needs".
Your needs are very important to us 🤣

tonyandrewmeyer (approved):
I feel this is ok because we're keeping main green (and would check any maintenance branch) and these tests are against external systems not ops itself.

However, if there's disagreement, I think waiting until we have our release discussion in Frankfurt would also be fine.

james-garner-canonical (approved):

PR #1686, opened 2025-04-21
ci: ops[tracing] integration tests

dimaqq:
Add integration tests, testing a sample charm that effectively uses `ops[tracing]` (code from current checkout) against COS Tempo (k8s) in a couple scenarios:
- charm -- (HTTP) -- tempo
- charm -- (HTTPS) -- tempo

Following tests were envisioned, but are deferred to a separate PR:
- charm -- (HTTP) -- traefik -- tempo
- charm -- (HTTPS) -- traefik -- tempo


test/integration/conftest.py
```diff
@@ -0,0 +1,130 @@
...
+s3_integrator_blocked = functools.partial(app_is, app='s3-integrator', status='blocked')
+
+
+@pytest.fixture
+def juju() -> Generator[jubilant.Juju, None, None]:
+    with jubilant.temp_model() as j:
+        j.deploy(**TEMPO)
+        j.deploy(**TEMPO_WORKER)
+        j.deploy(**MINIO)
+        j.deploy('s3-integrator')
```
dimaqq, line 86: @benhoyt WDYT about this idiom?
Would it be a good idea to add the kwargs types to jubilant itself?

The reason I'm doing this is because I wanted to have compact set of jubilant invocations, but Python doesn't provide a way to extract kwargs type from existing function signature.

benhoyt: I don't actually like it -- I think it makes the code less clear, as 1) you have to refer up to the const definition to see what's actually happening, and 2) it requires the big TypedDict definition.

I think there's nothing wrong with plain old (and type safe) function calls:

```python
juju.deploy(
    charm='tempo-coordinator-k8s',
    app='tempo',
    channel='edge',
    trust=True,
    resources={
        'nginx-image': 'ubuntu/nginx:1.24-24.04_beta',
        'nginx-prometheus-exporter-image': 'nginx/nginx-prometheus-exporter:1.1.0',
    },
)
juju.deploy(
    charm='tempo-worker-k8s',
    app='tempo-worker',
    channel='edge',
    config={'role-all': True},
    trust=True,
    resources={'tempo-image': 'docker.io/ubuntu/tempo:2-22.04'},
}
```

If you *really* want them on one line, I'd use functions:

```python
def deploy_tempo(juju):
    juju.deploy(
        charm='tempo-coordinator-k8s',
        app='tempo',
        channel='edge',
        trust=True,
        resources={
            'nginx-image': 'ubuntu/nginx:1.24-24.04_beta',
            'nginx-prometheus-exporter-image': 'nginx/nginx-prometheus-exporter:1.1.0',
        },
    )
...
deploy_tempo(juju)
```

test/charms/test_integration/charmcraft.yaml
```diff
@@ -0,0 +1,43 @@
...
+type: charm
+title: Tracing tester
+summary: Test tracing
+description: Test tracing
+
+base: ubuntu@24.04
+build-base: ubuntu@24.04
+platforms:
+    amd64:
+    arm64:
```
benhoyt, line 12: Do we need to build for both platforms?

test/integration/conftest.py
```diff
@@ -0,0 +1,146 @@
...
+import pytest
+
+
+@pytest.fixture
+def juju() -> Generator[jubilant.Juju, None, None]:
+    """Make a Juju model with the tracing part of COS ready."""
+    with jubilant.temp_model() as j:
+        deploy_tempo(j)
+        deploy_tempo_worker(j)
+        j.deploy('minio', config={'access-key': 'accesskey', 'secret-key': 'mysoverysecretkey'})
```
benhoyt, line 34: I'd prefer to see `juju.deploy()` instead of `j.deploy()` -- it reads more clearly. I know some tools complain about the shadowed name, but that's kind of an issue with how Pyright's fixture naming works. So I'd just `noqa` or turn off that warning and rename this to `juju`.

test/integration/conftest.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@pytest.fixture(scope='session')
+def build_charm(
+    pytestconfig: pytest.Config,
+) -> Generator[Callable[[], str], None, None]:
+    """Build the test charm and provide the artefact path.
+
+    Starts building the tracing-tester charm early.
+    Call the fixture value to get the built charm file path.
```
benhoyt, line 83: This is clever! Though perhaps a bit too clever... how much time does it actually save? Won't the first test have to wait for this anyway? Unless it's saving us a lot of time, I'd prefer just `charmcraft pack` in the GitHub actions, or a simple `subprocess.run()` here.

test/integration/conftest.py
```diff
@@ -0,0 +1,146 @@
...
+    yield tracing_test_charm
+    if proc.returncode is None:
+        proc.kill()
+    out, err = proc.communicate()
+    logging.info('`charmcraft pack` stdout follows:\n%s', out)
+    logging.info('`charmcraft pack` stderr follows:\n%s', err)
+
+
+def app_is(s: jubilant.Status, app: str, status: str):
+    return next(iter(s.apps[app].units.values())).workload_status.current == status
```
benhoyt, line 110: This `app_is` helper seems misnamed with what it actually does -- check *unit* status.

- Do we actually want to be testing unit status here? Would `jubilant.all_active` et al work? (They test both app and unit status)
- It seems clearer to me to put these as lambdas inline in the test. That way a person maintaining the test doesn't have to go two layers deeper to figure out what it actually does.

test/integration/conftest.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+# These could use jubilant.all_active(s, apps=["minio"]).
+# Keeping custom code for now, as the above API may change.
+minio_active = functools.partial(app_is, app='minio', status='active')
```
benhoyt, line 115: With the upcoming changes, this can be spelled:

```python
minio_active = lambda s: jubilant.all_active(s, 'minio')
# or, if you really just want to look at unit status
minio_active = lambda s: s.apps['minio'].units['minio/0'].is_active
```

test/integration/test_tracing.py
```diff
@@ -0,0 +1,110 @@
...
+    juju.integrate('tracing-tester', 'tempo')
+    status = juju.wait(jubilant.all_active)
+
+    spans = wait_spans(status.apps['tempo-worker'].address, test=lambda s: 'ops.main' in str(s))
+    assert 'ops.main' in [s['name'] for s in spans]
+
+    event_names = [e['name'] for e in get_events(spans)]
+    assert 'StartEvent' in event_names
+
+    checkpoint = time.time()
```
benhoyt, line 39: Does this require the clocks being in sync? Seems likely given these are ephemeral VMs for CI, but I wonder if we can avoid by setting `since` to the timestamp of the last span received?

test/integration/test_tracing.py
```diff
@@ -0,0 +1,110 @@
...
+    pass
+
+
+def wait_spans(
+    address: str,
+    since: float = 0,
+    *,
+    test: Callable[[list[dict[str, Any]]], bool],
+    timeout: float = 60,
+):
```
benhoyt, line 76: The signature seems slightly odd to me, as the `test` callable is required. And it seems like `since` should be keyword-only. How about calling it `ready` and making the signature slightly more like `Juju.wait`:

```python
def wait_spans(
    address: str,
    ready: Callable[[list[dict[str, Any]]], bool],
    *,
    since: float = 0,
    timeout: float = 60,
): ...

wait_spans(address, lambda spans: 'foo' in str(spans))
# or
wait_spans(address, ready=lambda spans: 'foo' in str(spans))
```

test/integration/test_tracing.py
```diff
@@ -0,0 +1,110 @@
...
+    logging.warning('giving up on waiting for spans')
+    return spans
+
+
+def get_spans(address: str, since: float = 0):
+    with httpx.Client(base_url=f'http://{address}:3200/api/', timeout=5) as client:
+        r = client.get('search?tags=service.name=tracing-tester')
+        r.raise_for_status()
+        spans: list[dict[str, Any]] = []
+        for t in r.json()['traces']:
```
benhoyt, line 94: I think `trace` and `response` would be clearer names here.

test/integration/test_tracing.py
```diff
@@ -0,0 +1,110 @@
...
+            if float(t['startTimeUnixNano']) / 1e9 < since:
+                continue
+            r = client.get(f'traces/{t["traceID"]}')
```
benhoyt, line 97: Is there any way to get more than one at once?

test/integration/test_tracing.py
benhoyt: Also, I think the type signature should have a return type. I wonder why Pyright didn't flag that?

benhoyt (commented):
Looks like a good start, thanks!

test/charms/test_integration/charmcraft.yaml
dimaqq: Tempo charm is only published for amd64, so I guess it makes sense to match that in this tester charm, even if it's platform-agnostic.

test/integration/test_tracing.py
dimaqq: Yes this requires clocks to be in sync; they should be between the test host and the k8s that runs on the test host.

There's no guarantee that we've gotten all spans, so I wouldn't want to depend on the "start of last span that we did get".

In this specific test, it's not a very big deal, because no actions are run before this point, however, I kindof like the idiom of scoping the spans we are looking at.

A full set of traces and spans is more than a screenful even for a simple test like this. I'd rather help whoever may need to debug this if/when there's a potential regression.

test/integration/test_tracing.py
dimaqq: The return type is inferred from `rv: SomeType = []; ...; return rv`.

test/integration/test_tracing.py
dimaqq: Nope, there's no batch endpoint.

Note that there really isn't a "trace" object, rather, a trace is a series of batches (typically 1) with series of scope spans (typ 1) with series of spans (many) all of the latter sharing the same `.traceID` attribute.

In real world, a trace is distributed across microservices, while a span belongs to a single microservice. The spans are sent over in spare time, and there's no concept of "having received all spans that belong to trace", rather, "here are the spans I have received so far" and "is the root span finished", while child spans are allowed to outlive their parents and the root.

test/integration/conftest.py
dimaqq: Note that charmcraft suffers from cold start on CI, and GitHub runners are slow.

My local cold start `charmcraft pack` result: `Executed in  128.52 secs`

I would expect GHA to take about 3x longer, so 6 minutes?

Meanwhile the total test run is ~22 minutes today, half of it is taken by concierge, and the other 10 minutes by juju deploy / wait, with tempo charm being particularly slow because it restarts itself.

benhoyt: But won't the test(s) have to wait until the pack finishes anyway, so it ends up being no better than just running `charmcraft pack` before the tests?

test/integration/test_tracing.py
benhoyt: Yep, sounds good.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,26 @@
...
+        preset: ['microk8s']  # ['k8s'] https://github.com/jnsgruk/concierge/issues/51
+        test-group: [1,2,3,4]
+
+    steps:
+      - run: sudo snap install --classic concierge
+      - run: sudo concierge prepare --juju-channel=3/stable --charmcraft-channel=3.x/stable -p "${{ matrix.preset }}"
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+      # Note that the test group count must match the groups, and there must be enough tests.
+      - run: uvx tox -e integration -- --test-group-count=4 --test-group=${{ matrix.test-group }}
```
benhoyt, line 26: I don't think this test grouping thing works very well if we're using module-scoped fixtures (as there'd be a lot of repeated work). Personally I'd prefer to see tests split up across multiple files and have it more explicitly done by filename -- would be easier to see what's failed and diagnose issues.

dimaqq: I'm specifically not using module-scoped fixtures.
This pytest plugin actually has a setting to group tests by files.
I'm not using it here, because my tests are independent and they are meant to be independent: they test different deployment options.

test/integration/conftest.py
dimaqq: charmcraft pack is a session fixture, which means it's executed first
https://docs.pytest.org/en/stable/reference/fixtures.html#fixture-instantiation-order

then the `juju` fixture is run, which spins up the bulk of the model.

then the test is run, which waits for charmcraft to complete (typically it's already complete).

test/integration/conftest.py
```diff
@@ -0,0 +1,140 @@
+# Copyright 2025 Canonical Ltd.
```
james-garner-canonical, line 1: The fixtures in this file are very specific to tracing, but if we add more integration tests, the path of least resistance is probably adding more fixtures to `conftest.py`. The right call at that point would probably be to move these fixtures into `test_tracing`, or namespace them in some other way. Reckon it's worth taking any steps along those lines in this PR? I don't think this is a blocker, but it might be nice if we're planning to add more integration tests in future.

test/integration/test_tracing.py
```diff
@@ -0,0 +1,122 @@
...
+    assert 'StartEvent' in event_names
+
+    checkpoint = time.time()
+    juju.run('tracing-tester/0', 'one', params={'arg': 'oneoneoneone'})
+    spans = wait_spans(trace_api, ready=lambda spans: 'one_action' in str(spans), since=checkpoint)
+
+    # A span generated by ops
+    assert 'one_action: TracingTesterCharm' in [span['name'] for span in spans]
+    # A span generated by the test charm
+    assert 'on action' in [span['name'] for span in spans]
```
james-garner-canonical, line 47: `one_action` and `on action` seem similar enough to be confusing

test/integration/test_tracing.py
```diff
@@ -0,0 +1,122 @@
...
+    trace_api = status.apps['tempo-worker'].address
+
+    spans = wait_spans(trace_api, ready=lambda spans: 'ops.main' in str(spans))
+    assert 'ops.main' in [span['name'] for span in spans]
+
+    event_names = [event['name'] for event in get_events(spans)]
+    assert 'StartEvent' in event_names
+
+    checkpoint = time.time()
+    juju.run('tracing-tester/0', 'one', params={'arg': 'oneoneoneone'})
```
james-garner-canonical, line 41: I spend an embarassing amount of time trying to figure out how the action name was ending up in the span multiple times when I read the assert line haha
```py
assert 'oneoneoneone' in json.dumps(action_span)
```

Maybe a variable would be good?
```suggestion
    arg_value = 'oneoneoneoneone!!!!1'
    juju.run('tracing-tester/0', 'one', params={'arg': arg_value})
```
And later:
```py
assert arg_value in json.dumps(action_span)
```

Maybe `action_span_name = 'on action'` would be a good idea too. I'd be pretty tempted by `action_name = 'one'` and using `f'{action_name}_action'` over `'one_action'` too tbh.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: ['microk8s']  # ['k8s'] https://github.com/jnsgruk/concierge/issues/51
+        test-group: [1, 2, 3, 4]
```
james-garner-canonical, line 18: I'd kind of prefer manual test grouping of some sort. I think I remember Ben making a comment somewhere along the same lines. I guess the intention right now is to run each of the four tests in their own job in a way that doesn't require manually updating the workflow or `tox.ini` when adding new tests in future, which does seem nice.

james-garner-canonical (commented):
The tests themselves seem good. Just a few comments on the CI/organisation and naming.

docs/reference/ops-tracing.rst
```diff
@@ -14,6 +14,16 @@ Open Telemetry resource attributes
...
 
+
+Security considerations
+-----------------------
+
+The trace data can be sent out over HTTP or HTTPS. If your charm uses the
+``ops.tracing.Tracing()`` object, the protocol is determined by the URL that
+the charm tracing integration counterpart posts in the databag.
+
+This release supports TLS 1.2 and 1.3 for HTTPS connections.
```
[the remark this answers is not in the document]
james-garner-canonical: Ah we're just lowering the minimum version, I see.

test/integration/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+import subprocess
+from typing import Callable, Generator
+
+import jubilant
+import minio
+import pytest
+
+
+@pytest.fixture
+def juju() -> Generator[jubilant.Juju, None, None]:
```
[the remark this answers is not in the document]
dimaqq: yes, each test represents a different deployment scenario.

tox.ini
```diff
@@ -209,6 +211,35 @@ commands =
...
+
+    # Build `ops` and `ops-tracing` source tarballs.
+    bash -c 'for pkg in . tracing; do \
+                 (cd "$pkg" && uv build --sdist --out-dir={toxinidir}/test/charms/test_integration/); \
+             done'
+
+    # Write both tarballs into requirements.txt (one per line, with leading ./)
+    bash -c 'find {toxinidir}/test/charms/test_integration -maxdepth 1 -name "*.tar.gz" -print \
+        | sed "s|^.*/|./|" \
+        > {toxinidir}/test/charms/test_integration/requirements.txt'
```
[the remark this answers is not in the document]
dimaqq: I wonder if it's easier to re-write this in Python.

As it stands now, shell is needed to expand things like `*.charm` and `ops*.tar.gz`.

Tox doesn't glob natively.
So it's either `sh -c` or `find -name | xargs rm` or in fact Python.

test/charms/test_integration/charmcraft.yaml
```diff
@@ -0,0 +1,42 @@
+---
+name: tracing-tester
```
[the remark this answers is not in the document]
dimaqq: Maybe!
I was thinking that there could be more charms here eventually...
Let me think about this, perhaps it's easier to rename things a little bit.

docs/reference/ops-tracing.rst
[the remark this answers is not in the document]
dimaqq: yep, also I've merged this change in a separate PR #1705

test/integration/conftest.py
dimaqq: I figure that's something to solve if or when more integration tests are added.
Maybe we'd move the whole shebang to a subdirectory, or maybe we'd rename individual fixtures.

test/integration/test_tracing.py
dimaqq: fair point, changing

test/integration/test_tracing.py
dimaqq: sure, going to change this up.

.github/workflows/integration.yaml
dimaqq: Yeah, it's not ideal.
Some CI systems allow dynamic dispatch:
- `pytest --collect-only`
- N * `pytest somefile.py:some_test`

But GHA matrix is static, so that won't fly.
I guess we could make this a little better by using **more** test groups than there are tests and then short-circuiting the empty groups.
That felt a bit too magical to roll into a PR with useful tests though.

Let me lighten up the missing tests for now.

dimaqq:
I've now removed `on: pull_request` trigger, the latest run was all green:
https://github.com/dimaqq/operator/actions/runs/14707549352

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,42 @@
...
+        preset: ['k8s', 'microk8s']
+        test-group: [1, 2, 3, 4]
+
+    steps:
+      - run: |
+           # Work around https://github.com/jnsgruk/concierge/issues/30
+           sudo apt-get remove -y docker-ce docker-ce-cli containerd.io 
+           sudo rm -rf /run/containerd 
+      - run: sudo snap install --classic concierge
+      - run: >
```
benhoyt, line 26: Huh, TIL about `>`. YAML is so powerful/complicated!

benhoyt (changes requested):
Thanks for this. Nice end-to-end tests, and looks like a good start!

The one thing I'd like to see cleaned up is how we run the tests, and avoid pytest-test-groups -- see my comment on the thread.

Also, could you please add a PR description with a high-level overview of what we're testing here?

test/integration/conftest.py
```diff
@@ -0,0 +1,202 @@
...
```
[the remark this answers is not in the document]
dimaqq: That's Python 3.9+

I'm happy to convert integration and smoke tests to higher Python, but in a separate PR.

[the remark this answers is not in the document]
benhoyt: Perhaps, but it works fine on Python 3.8 projects if you're using `from __future__ import annotations` (which you are). Though I notice we're not type-checking this whole folder, so it wouldn't matter anyway.

.github/workflows/integration.yaml
dimaqq: That's a fair point, I don't like these 1,2,3,4 much either.

I'll experiment with better visibility in https://github.com/dimaqq/operator/pull/41 and remove group distribution for now... After all there are only 2 tests so far.

test/charms/test_tracing/src/charm.py
```diff
@@ -0,0 +1,52 @@
...
+            self, 'charm-tracing', ca_relation_name='receive-ca-cert'
+        )
+        self.framework.observe(self.on.start, self._on_start)
+        self.framework.observe(self.on.one_action, self._on_action)
+        self.framework.observe(self.on.two_action, self._on_action)
+        self.framework.observe(self.on.one_action, self._on_action)
+
+    @tracer.start_as_current_span('start')
+    def _on_start(self, event: ops.StartEvent):
+        if self.unit.is_leader():
```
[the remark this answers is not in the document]
dimaqq: App status vs unit status.

test/integration/conftest.py
```diff
@@ -0,0 +1,202 @@
...
+        # - new pod is scheduled by k8s
+        # - check own stateful set
+        #
+        # This process may take a while.
+
+        juju.wait(jubilant.all_active)
+
+        yield juju
+
+        print(juju.debug_log())
```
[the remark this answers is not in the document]
dimaqq: I found that in github actions specifically, having the debug on successful completion is very useful:
- to learn thing (ok maybe once?)
- to compare good and bad run
- to make manual check against false positives possible

test/integration/conftest.py
```diff
@@ -0,0 +1,202 @@
...
+            {'access-key': 'accesskey', 'secret-key': 'mysoverysecretkey'},
+        )
+
+        # Tempo goes through a cycle of:
+        # - update own stateful set
+        # - kill own pod
+        # - new pod is scheduled by k8s
+        # - check own stateful set
+        #
+        # This process may take a while.
```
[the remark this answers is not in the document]
dimaqq: Looks like that takes about half of the default timeout.

```
...
2025-04-29T23:50:15.8063021Z + .apps['tempo-worker'].units['tempo-worker/0'].workload_status.current = 'maintenance'
...
...
...
2025-04-29T23:51:37.0457228Z + .apps['tempo-worker'].units['tempo-worker/0'].workload_status.message = 'metrics-generator disabled. No prometheus remote-write relation configured on the coordinator'
```

Happens to work OK both in local testing and GHA, but I felt I had to call this out to any future maintainer.

.github/workflows/integration.yaml
dimaqq: Done: I've removed groups for now; they can always be bolted on later.

dimaqq:
Latest run: https://github.com/dimaqq/operator/actions/runs/14744203529
(there was one transient failure, if we get more of these, a longer wait timeout may be needed)

benhoyt (approved):
Thanks. Looks good! Can you please add a PR description before merge? Just a high-level description of what's being tested.

dimaqq:
The current approach (without groups) fails repeatedly on CK8s with:
```
2025-04-30T06:15:31.8259692Z       app_status=StatusInfo(
2025-04-30T06:15:31.8259909Z         current='waiting',
2025-04-30T06:15:31.8260412Z         message='running PreBind plugin "VolumeBinding": binding volumes: provisioning failed for PVC "minio-data-bfa19afc-minio-0"',
2025-04-30T06:15:31.8260926Z         since='30 Apr 2025 06:09:32Z',
2025-04-30T06:15:31.8261143Z       ),
2025-04-30T06:15:31.8261298Z       units={
2025-04-30T06:15:31.8261467Z         'minio/0': UnitStatus(
2025-04-30T06:15:31.8261688Z           workload_status=StatusInfo(
2025-04-30T06:15:31.8261931Z             current='waiting',
2025-04-30T06:15:31.8262421Z             message='running PreBind plugin "VolumeBinding": binding volumes: provisioning failed for PVC "minio-data-bfa19afc-minio-0"',
2025-04-30T06:15:31.8262955Z             since='30 Apr 2025 06:09:32Z',
2025-04-30T06:15:31.8263181Z           ),
2025-04-30T06:15:31.8263360Z           juju_status=StatusInfo(
2025-04-30T06:15:31.8263580Z             current='allocating',
2025-04-30T06:15:31.8264235Z             message='running PreBind plugin "VolumeBinding": binding volumes: provisioning failed for PVC "minio-data-bfa19afc-minio-0"',
2025-04-30T06:15:31.8264770Z             since='30 Apr 2025 06:09:32Z',
2025-04-30T06:15:31.8264999Z           ),
2025-04-30T06:15:31.8265158Z           leader=True,
2025-04-30T06:15:31.8265364Z           open_ports=['9000/TCP', '9001/TCP'],
2025-04-30T06:15:31.8265621Z           provider_id='minio-0',
2025-04-30T06:15:31.8265825Z         ),
2025-04-30T06:15:31.8265977Z       },
```

... when running the second test out of the two.

I'm not quite sure how to debug that, and I wonder if perhaps running two tests in series just somehow uses up all the available disk/volume space or something?

tox.ini
```diff
@@ -116,6 +116,7 @@ deps =
 commands =
     pytest -n auto \
         --ignore={[vars]tst_path}smoke \
+        --ignore={[vars]tst_path}integration \
```
tonyandrewmeyer, line 119: Not for this PR, but I think we definitely want to reorganise our tests and tox some time in 25.10, so that we're more "run these tests" and less "run all the tests except a,b,c,d".

test/charms/test_tracing/src/charm.py
```diff
@@ -0,0 +1,51 @@
...
+            self.app.status = ops.ActiveStatus('ok')
+        self.unit.status = ops.ActiveStatus('ok')
+
+    @tracer.start_as_current_span('custom trace on any action')
+    def _on_action(self, event: ops.ActionEvent):
+        time.sleep(1)
```
tonyandrewmeyer, line 45: Is this so that the "since checkpoint" bit works? I don't love sleeps in test code, partly because they're slower than needed, but also because the period is somewhat of a magic number.

Since this is a standalone test and not a production system hooked up to a real trace consumer, couldn't we say "since an hour ago" and avoid the sleep? Or would that not work for some reason?

test/integration/conftest.py
```diff
@@ -0,0 +1,202 @@
...
+
+        juju.wait(jubilant.all_active)
+
+        yield juju
+
+        print(juju.debug_log())
+
+
+@pytest.fixture(scope='session')
+def charm_dir(pytestconfig: pytest.Config) -> Generator[pathlib.Path, None, None]:
```
tonyandrewmeyer, line 80: I think future annotations lets you use the shorter return type if you prefer. Fine with me either way.

tonyandrewmeyer (approved):
Very nice, thanks!

I'm assuming you'll be able to resolve the resourcing-related failures referenced in the recent comment without significant changes.

james-garner-canonical (approved):

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,34 @@
...
+permissions: {}
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: ['k8s', 'microk8s']
```
james-garner-canonical, line 17: > The current approach (without groups) fails repeatedly ... when running the second test out of the two.

Worth adding a matrix to `-k` the two individual tests? I realise that adds extra work when adding a new integration test, but maybe it's fine.

```suggestion
        preset: ['k8s', 'microk8s']
        test: ['test_direct_connection', 'test_with_tls']
```

benhoyt: I think Dima's right that it's not worth this for just two tests. We can always rearrange as we add more.

james-garner-canonical: The motivation for my suggestion is purely as a fix for Dima's comment:

> The current approach (without groups) fails repeatedly ... when running the second test out of the two.
> I'm not quite sure how to debug that, and I wonder if perhaps running two tests in series just somehow uses up all the available disk/volume space or something?

I agree that this isn't worthwhile for just two tests otherwise.

test/charms/test_tracing/src/charm.py
dimaqq: Ah, no, this can be removed.
I originally wanted to check/test span durations, but that would be an overkill.

test/integration/conftest.py
dimaqq: Let me check that, you may be right. It's a 3.9+ feature, but future annotations stringify the type hints and these are not used at run time, so it should work.

.github/workflows/integration.yaml
dimaqq: I've kicked off a test with `-k` in a separate branch. Fingers crossed.

tox.ini
dimaqq: yep, that's for a separate effort

test/charms/test_tracing/src/charm.py
dimaqq: removed, all good.

test/integration/conftest.py
dimaqq: done, thank you!

.github/workflows/integration.yaml
dimaqq: works, finally! 😓

dimaqq:
Test run: https://github.com/dimaqq/operator/actions/runs/14771689094

PR #1687, opened 2025-04-22
chore: use annotations from the future for ops/model.py

james-garner-canonical:
This PR adds `from __future__ import annotations` to `ops/model.py` and updates the annotations accordingly.

[Here is a diff](https://github.com/james-garner-canonical/operator/compare/2025-04/chore/test-pyupgrade..james-garner-canonical:operator:2025-04/chore/future-annotations-ops-model?expand=1) showing the manual changes layered on top of those that running `ruff` with `pyupgrade` enabled gives us -- the automated changes haven't changed types outside of annotations (casts, type aliases, generic inheritance), while this PR does.

Type aliases have been annotated with `TypeAlias` (imported from `typing_extensions` in `if typing.TYPE_CHECKING`) and had their annotations stringified  if needed.

I intend to update annotations in other `ops` modules, but I figured it was worth breaking it into smaller chunks for review. Once all the modules are updated we can add the [type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) and [future-annotations](https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa) rulesets to our ruff linting, and disable `tool.ruff.lint.pyupgrade.keep-runtime-typing` to keep everything good automatically going forward.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -72,11 +72,19 @@ _ = JujuVersion
 # a k8s spec is a mapping from names/"types" to json/yaml spec objects
 K8sSpec = Mapping[str, Any]
 
+_StorageDictType: TypeAlias = 'dict[str, list[Storage] | None]'
+_BindingDictType: TypeAlias = 'dict[str | Relation, Binding]'
+
 _ReadOnlyStatusName = Literal['error', 'unknown']
 _SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
 StatusName: TypeAlias = '_SettableStatusName | _ReadOnlyStatusName'
 _StatusDict = TypedDict('_StatusDict', {'status': StatusName, 'message': str})
 _SETTABLE_STATUS_NAMES: tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
+
+# mapping from relation name to a list of relation objects
+_RelationMapping_Raw: TypeAlias = 'dict[str, list[Relation] | None]'
+# mapping from container name to container metadata
+_ContainerMeta_Raw: TypeAlias = 'dict[str, _charm.ContainerMeta]'
 
 # relation data is a string key: string value mapping so far as the
 # controller is concerned
@@ -563,8 +571,7 @@ class Unit:
         self._collected_statuses: list[StatusBase] = []
 
         if self._is_our_unit and hasattr(meta, 'containers'):
-            # mapping from container name to container metadata
-            containers: dict[str, _charm.ContainerMeta] = meta.containers
+            containers: _ContainerMeta_Raw = meta.containers
             self._containers = ContainerMapping(iter(containers), backend)
 
     def _invalidate(self):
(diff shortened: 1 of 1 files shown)


ops/model.py
```diff
@@ -72,19 +71,11 @@
 # a k8s spec is a mapping from names/"types" to json/yaml spec objects
 K8sSpec = Mapping[str, Any]
 
-_StorageDictType = Dict[str, Optional[List['Storage']]]
-_BindingDictType = Dict[Union[str, 'Relation'], 'Binding']
```
james-garner-canonical, line 76: These aliases are used only once, so they've been in-lined to facilitate modernising the annotations.

ops/model.py
```diff
@@ -72,19 +71,11 @@
...
 _ReadOnlyStatusName = Literal['error', 'unknown']
 _SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
 StatusName = Union[_SettableStatusName, _ReadOnlyStatusName]
 _StatusDict = TypedDict('_StatusDict', {'status': StatusName, 'message': str})
-_SETTABLE_STATUS_NAMES: Tuple[_SettableStatusName, ...] = get_args(_SettableStatusName)
-
-# mapping from relation name to a list of relation objects
-_RelationMapping_Raw = Dict[str, Optional[List['Relation']]]
-# mapping from container name to container metadata
-_ContainerMeta_Raw = Dict[str, '_charm.ContainerMeta']
```
james-garner-canonical, line 87: These aliases are used only once, so they've been in-lined (with their respective comments) to facilitate modernising the annotations.

james-garner-canonical:
This is the only remaining use of `Dict`. It's referred to several times, but it's also a tiny annotation so it could be easily in-lined. On the other hand, maybe the descriptive name is worth preserving.
```py
# relation data is a string key: string value mapping so far as the
# controller is concerned
_RelationDataContent_Raw = Dict[str, str]
```
Options:
1. in-line
2. use `TypeAlias`
3. leave as is

---

This alias seems like it would be worth in-lining despite being used several times, as the name doesn't add anything over `type[Unit] | type[Application]`, but it's public, so I haven't touched it. It doesn't seem to be used in charm code as of a few months ago, so perhaps it's safe to change.
```py
UnitOrApplicationType = Union[Type['Unit'], Type['Application']]
```

benhoyt:
Will review. [pyupgrade](https://github.com/asottile/pyupgrade) is the tool I was thinking of. Might be worth comparing with this or trying on future files?

benhoyt (approved):
This looks good to me. Assuming you get pyupgrade working half-decently, would be good to compare it with this just to ensure we haven't missed anything or made any mistakes (assuming pyupgrade is mistake-free!).

james-garner-canonical:
> Assuming you get pyupgrade working half-decently, would be good to compare it with this just to ensure we haven't missed anything or made any mistakes (assuming pyupgrade is mistake-free!).

I can confirm that running the `pyupgrade` checks with `ruff` results in no complaints or changes for `ops/model.py` (disabling our current setting of `tool.ruff.lint.pyupgrade.keep-runtime-typing`)

dimaqq (approved):
Thank you!

PR #1689, opened 2025-04-22
docs: add best practices about status

dwilding:
This PR adds a section called "Handle status" to [How to write and structure charm code](https://ops.readthedocs.io/en/latest/howto/write-and-structure-charm-code.html).

### DIFF
--- a/docs/howto/write-and-structure-charm-code.md
+++ b/docs/howto/write-and-structure-charm-code.md
@@ -269,6 +269,46 @@ def get_version() -> str:
     return response.text
 ```
 
+(handle-status)=
+## Handle status
+
+Your charm should report its status to Juju. The best approach depends on the charm's complexity:
+
+  - If units don't need to coordinate, use [`self.unit.status`](ops.Unit.status) to report the
+    status of each unit independently. For example:
+
+    ```python
+    def _on_start(self, event: ops.StartEvent):
+        """Handle start event."""
+        self.unit.status = ops.MaintenanceStatus("starting server")
+        # ...
+        # Start the workload and wait for it to be ready.
+        # ...
+        self.unit.status = ops.ActiveStatus()
+    ```
+
+    When your charm code sets `self.unit.status`, Ops immediately sends the unit status to Juju.
+
+    Juju aggregates the unit statuses into an overall application status.
+
+    ```{admonition} Best practice
+    :class: hint
+
+    In practice, it's often awkward to decide which status to report. Instead of writing
+    (and duplicating) complex conditional logic to set `self.unit.status`, write a handler for
+    [`collect_unit_status`](ops.CharmEvents.collect_unit_status) that tells Ops about the
+    relevant statuses. Ops will send the highest priority status to Juju. For an example, see
+    [](#integrate-your-charm-with-postgresql-update-unit-status).
+    ```
+
+  - If units need to coordinate, the leader unit should report the application status.
+    In your charm code, observe [`collect_app_status`](ops.CharmEvents.collect_app_status) or set
+    [`self.app.status`](ops.Application.status).
+
+    Each unit should still report its own status, but Juju won't aggregate the unit statuses.
+
+> See also: {external+juju:ref}`Juju | Status <status>`
+
 (handle-errors)=
 ## Handle errors
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -230,6 +230,7 @@ The diagram below illustrates the workflow for the case where the database relat
 
 ![Integrate your charm with PostgreSQL](../../resources/integrate_your_charm_with_postgresql.png) 
 
+(integrate-your-charm-with-postgresql-update-unit-status)=
 ## Update the unit status to reflect the relation state
 
 Now that the charm is getting more complex, there are many more cases where the unit status needs to be set. It's often convenient to do this in a more declarative fashion, which is where the collect-status event can be used.


docs/howto/write-and-structure-charm-code.md
```diff
@@ -269,6 +269,46 @@ def get_version() -> str:
     return response.text
 ```
 
+(handle-status)=
+## Handle status
+
+Your charm should report its status to Juju. The best approach depends on the charm's complexity:
+
+  - If units don't need to coordinate, use [`self.unit.status`](ops.Unit.status) to report the
```
benhoyt, line 277: This seems a bit backwards to me: we start by telling them to use `self.unit.status` but then say the best practice is not doing that, it's actually using `collect_unit_status`. I think we should reverse that: recommend `collect_unit_status` with an example, but then say something like "if you need something more fine-grained, you can set the status directly with `self.unit.status = ...`".

tonyandrewmeyer: Agreed. I think I muddle the conversation in daily earlier in the week because I was behind on Matrix messages and didn't realise there were two questions. I think our conclusions were:

* `collect-*-status` is always the pattern we recommend. It's ok to not use it for simple cases, but we should always prefer it in our docs, even in simple cases like the tutorial, because it's better to start people on that track than potentially confuse them by switching later (and then there's the "when you do pass simple" question).
* If your charm is expected to have more than one unit, you probably want to also set the app status rather than leaving it to Juju.

I suggest what we do here is say that you should use `collect-unit-status`, except when you need the status to update *during* the hook, typically setting maintenance status or active status with a degraded service message. The example would be similar, but the active status would be done in a collect_unit_status handler.

docs/howto/write-and-structure-charm-code.md
```diff
@@ -269,6 +269,46 @@ def get_version() -> str:
...
+    def _on_start(self, event: ops.StartEvent):
+        """Handle start event."""
+        self.unit.status = ops.MaintenanceStatus("starting server")
+        # ...
+        # Start the workload and wait for it to be ready.
+        # ...
+        self.unit.status = ops.ActiveStatus()
+    ```
+
+    When your charm code sets `self.unit.status`, Ops immediately sends the unit status to Juju.
```
tonyandrewmeyer, line 290: I think we need to extend this to explain that collect-status updates the status in Juju at the end of the hook.

docs/howto/write-and-structure-charm-code.md
```diff
@@ -269,6 +269,46 @@ def get_version() -> str:
...
+
+    Juju aggregates the unit statuses into an overall application status.
```
tonyandrewmeyer, line 292: I was going to suggest that this link to Juju to explain how the status is chosen, but then when I found the relevant page I noticed that it's the same as the "see also" below. We probably don't want to link to the same page twice, even if one link is to a specific section. Any ideas on how to indicate that you'll find the info there? If not, I'm fine with this as-is.

docs/howto/write-and-structure-charm-code.md
```diff
@@ -269,6 +269,46 @@ def get_version() -> str:
...
+
+    ```{admonition} Best practice
+    :class: hint
+
+    In practice, it's often awkward to decide which status to report. Instead of writing
```
tonyandrewmeyer, line 297: I like the best practice notes to be as small as possible, so that they'll be easier to consume in a single list of items to check. What do you think about removing the first sentence?

docs/howto/write-and-structure-charm-code.md
```diff
@@ -269,6 +269,46 @@ def get_version() -> str:
...
+    (and duplicating) complex conditional logic to set `self.unit.status`, write a handler for
+    [`collect_unit_status`](ops.CharmEvents.collect_unit_status) that tells Ops about the
+    relevant statuses. Ops will send the highest priority status to Juju. For an example, see
+    [](#integrate-your-charm-with-postgresql-update-unit-status).
+    ```
+
+  - If units need to coordinate, the leader unit should report the application status.
+    In your charm code, observe [`collect_app_status`](ops.CharmEvents.collect_app_status) or set
+    [`self.app.status`](ops.Application.status).
```
tonyandrewmeyer, line 306: I wonder if we can point out here (a small example with a comment?) that the charm can unconditionally observe collect-app-status because it'll only be emitted for the leader. That is, you can do:

```python
framework.observe(self.on.collect_app_status, self._on_app_status)
```

Rather than needing to do:

```python
if self.unit.is_leader:
    framework.observe(self.on.collect_app_status, self._on_app_status)
```

docs/howto/write-and-structure-charm-code.md
```diff
@@ -269,6 +269,83 @@ def get_version() -> str:
...
+    def _on_collect_status(self, event: ops.CollectStatusEvent):
+        if "port" not in self.config:
+            event.add_status(ops.BlockedStatus("no port specified"))
+            return
+        event.add_status(ops.ActiveStatus())
+```
+
+In your handler for `collect_unit_status`, you can call `add_status()` multiple times. Ops will send the highest priority status to Juju. This reduces the amount of logic you need to write to decide which status to report. For more information, see [`CollectStatusEvent`](ops.CollectStatusEvent).
+
+Your handler for `collect_unit_status` won't have access to data about the event that was handled before Ops triggered `collect_unit_status`.
```
dwilding, line 301: Not super happy with this part. Since `collect_unit_status` is triggered after _any_ event (not only observed/handled ones), the phrase "the event that was handled before..." isn't totally right.

benhoyt: Yeah, agreed this could use improvement. Maybe "data about the main Juju event (hook) being handled"?

benhoyt (approved):
Looks in pretty good shape, thanks!

tonyandrewmeyer (approved):

PR #1692, opened 2025-04-23
docs: how to manage metrics

IronCore864:
How to create a "basic" type identity and access the metrics endpoint in a charm context.

Closes https://github.com/canonical/operator/issues/1595.


dwilding:
Thanks @IronCore864, I've finished reviewing. I started drafting suggestions, but it was getting messy because I was trying to suggest a different overall sequence. So I've created a separate branch to show my thinking:
https://github.com/dwilding/canonical-operator/blob/review-manage-metrics/docs/howto/manage-metrics.md

I found it confusing that we create a secret and hardcode the secret ID as the default value of `metrics-secret-id`. Surely that default value would only be useful within the same Juju instance - or is that what we want? I think it's clearer to start with all the charm code, then show all the `juju ...` commands that would be run by whoever is deploying the charm.

What do you think about my revised version of the doc?

A couple of extra questions:

- In `_on_secret_changed`, do we need to check the secret ID against the expected ID before trying to extract the username and password?
- Instead of storing the secret ID in a config option, would it be simpler/clearer to store the secret name?

IronCore864:
> Thanks @IronCore864, I've finished reviewing. I started drafting suggestions, but it was getting messy because I was trying to suggest a different overall sequence. So I've created a separate branch to show my thinking: https://github.com/dwilding/canonical-operator/blob/review-manage-metrics/docs/howto/manage-metrics.md

It's not easy to do a full diff with your suggested version, but I get the idea: You prefer to show the code first before explanation, while my approach is a step-by-step guide from the user's perspective, what they would do logically and chronologically if they were to add secrets. I will think about which way is better, then come up with a final draft. Maybe I will do a more high-level introduction, then a step-by-step how-to guide so that the users can follow and understand the logic behind it better.

> * In `_on_secret_changed`, do we need to check the secret ID against the expected ID before trying to extract the username and password?

I am not sure if we need to check the secret ID, because I see in the manage secrets how-to, it doesn't.

> * Instead of storing the secret ID in the config option, would it be simpler/clearer to store the secret name in the config option?

After some research, I do not think this is possible. When getting the secret, we can only use either [ID or label](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Model.get_secret), and the [label is only local to the charm](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Secret.label). When [adding the user secret](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/add-secret/), you can't specify a label. Maybe there is an approach that I'm not aware of?

IronCore864:
After some rework, I combined my original piece with David's suggestion. This is my preferred logic: lay out the big picture before concrete stuff, then a detailed step-by-step guide following an order that a real user would do when adding secrets/configurations.

dwilding:
I'm still confused about the order of steps - I'll send you a message and leave a follow-up here later.

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,163 @@
...
+
+See more:
+
+- [Service discovery within a Kubernetes cluster](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/)
+
+### Through an Ingress
+
+To access the metrics endpoint from outside the Kubernetes cluster, use an Ingress.
+
+Use the service `<charm-name>` service in the Ingress (which is also automatically created) instead of the `<charm-name>-endpoints` service, as the latter is a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) and doesn't have a ClusterIP.
```
dwilding, line 133: Suggestion for consistency with the Kubernetes section:

```suggestion
Use the service `<charm-name>` service in the Ingress (which is also created by Juju) instead of the `<charm-name>-endpoints` service, as the latter is a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) and doesn't have a ClusterIP.
```

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,163 @@
...
+
+- To expose the Pebble HTTP port, use [`ops.Unit.set_ports`](ops.Unit.set_ports) in your charm code:
```
dwilding, line 135: I think it's fine to drop the full path:

```suggestion
- To expose the Pebble HTTP port, use [`set_ports`](ops.Unit.set_ports) in your charm code:
```

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,163 @@
...
+
+```python
+  self.unit.set_ports(38813)
+```
+
+- To create an Ingress, use the following example (assuming the charm and service name is `my-charm`):
```
dwilding, line 141: I wasn't sure where I would apply this example (but I'm not very experienced with K8s). Would this be appropriate?

```suggestion
- To create an Ingress, use the following cluster configuration (assuming the charm and service name is `my-charm`):
```

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,163 @@
+# How to Manage Metrics
```
dwilding, line 1:
```suggestion
# How to manage metrics
```

dwilding (approved):
Thank you @IronCore864 for helping me to work through my confusion and making some changes to the doc. Now that the doc includes something about how the charm user would set the config option (and now that the config option is not the default value of the secret ID) I'm good with the sequence of steps.

I think it makes sense while developing the charm. A couple of additional sentences in the bullet points at the beginning would help to reinforce what a charm user would need to do - see my suggestions.

Nice work, and I really appreciate your patience as my first review wasn't as helpful as it could have been!

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,163 @@
...
+- {external+juju:ref}`Juju | config <command-juju-config>`
+- {external+juju:ref}`juju grant-secret <command-juju-grant-secret>`
+
+## Access the metrics endpoint
+
+### Within the same Kubernetes cluster
+
+Deploying the charm causes Juju to automatically create a Kubernetes service named `<charm-name>-endpoints` within the Kubernetes cluster. Use this service to connect to Pebble within each workload container. Pebble's HTTP port for the first workload container is `38813`.
+
+For example, if you deploy a charm named `my-charm` in the `test` namespace, access the metrics endpoint at `my-charm-endpoints.test.svc.cluster.local:38813/v1/metrics`. You'll need to use HTTP basic authentication with the username and password that you specified in the `juju add-secret` command.
```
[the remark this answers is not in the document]
IronCore864: I'll leave out the protocol, a few reasons: 
- protocol isn't part of DNS
- consistent with the K8s doc
- we might support TLS in the future

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,163 @@
...
+
+See more:
+
+- [Service discovery within a Kubernetes cluster](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/)
+
+### Through an Ingress
```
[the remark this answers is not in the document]
IronCore864: An Ingress resource is the most accurate, but since we don't say "accessing a Pod resource", I guess it's OK to say just an "Ingress".

docs/howto/manage-metrics.md
IronCore864: I'll use "Ingress resource example" to be consistent with the [K8s doc](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource).

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,167 @@
...
+        access="metrics", basic=ops.pebble.BasicIdentity(password=sha512_crypt.hash(password))
+      ),
+    }
+    self.container.pebble.replace_identities(identities)
+    logger.debug("New metrics username: %s", username)
+
+  ...
+```
+
+The password of the Pebble identity is stored as a hash, which we generate using `sha512_crypt.hash()` from `passlib.hash`.
```
benhoyt, line 80: Could we link to docs for this function, or at least the library?

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,167 @@
...
+After deploying the charm, grant access to the user secret:
+
+```bash
+juju grant-secret metrics-user-password <charm-name>
+```
+
+> See more:
+> - {external+juju:ref}`Juju | config <command-juju-config>`
+> - {external+juju:ref}`Juju | deploy <command-juju-deploy>`
+> - {external+juju:ref}`Juju | grant-secret <command-juju-grant-secret>`
```
benhoyt, line 115: I think config and deploy are such basic commands they don't need to be linked, and maybe we can link grant-secret in the text above: `... [grant access](...) to the user secret`.

docs/howto/manage-metrics.md
```diff
@@ -0,0 +1,167 @@
...
+my-charm-endpoints.test.svc.cluster.local:38813/v1/metrics
+```
+
+You'll need to use HTTP basic authentication with the username and password that you specified in the `juju add-secret` command.
+
+> See more: [Service discovery within a Kubernetes cluster](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/)
+
+### Through an Ingress
+
+To access the metrics endpoint from outside the Kubernetes cluster, use an Ingress.
```
benhoyt, line 135: It might help with David's point above if we link the word Ingress. Can we link it to https://kubernetes.io/docs/concepts/services-networking/ingress/?

benhoyt (commented):
This is good, thank you! A few minor comments.

docs/howto/manage-metrics.md
```diff
@@ -53,13 +58,16 @@ class MyCharm(ops.CharmBase):
...
     # - Created a secret with keys 'username' and 'password'
     # - Stored the secret ID in the 'metrics-secret-id' configuration option
+    if not self.config.get('metrics-secret-id'):
+      return
     secret_id = str(self.config["metrics-secret-id"])
     secret = self.model.get_secret(id=secret_id)
     content = secret.get_content()
-    username, password = content["username"], content["password"]
-    self._replace_identities(username, password)
+    self._replace_identities(content["username"], content["password"])
```
benhoyt, line 66: Thanks. Let's make the same change below for consistency.

benhoyt (approved):
Approving -- looks pretty good. Just left a couple of nit comments.

IronCore864:
All comments resolved, preview [here](https://ops--1692.org.readthedocs.build/en/1692/howto/manage-metrics.html), looking good. Failed observability charm tests in the CI unrelated to this change, merging now.

PR #1695, opened 2025-04-25
test: fix overly specific test that fails sometimes with tracing

james-garner-canonical:
This PR fixes a test that fails intermittently on main. It failed because it asserted that the last two log lines were specific ones, while the addition of tracing means that sometimes (but not always), tracing log lines are emitted afterwards instead. This PR modifies the test to only assert that both lines are present in the logs instead.

### DIFF
--- a/testing/tests/test_e2e/test_juju_log.py
+++ b/testing/tests/test_e2e/test_juju_log.py
@@ -31,6 +31,8 @@ def mycharm():
 def test_juju_log(mycharm):
     ctx = Context(mycharm, meta=mycharm.META)
     ctx.run(ctx.on.start(), State())
-    assert ctx.juju_log[-2] == JujuLogLine(level='DEBUG', message='Emitting Juju event start.')
-    assert ctx.juju_log[-1] == JujuLogLine(level='WARNING', message='bar!')
+    line = JujuLogLine(level='DEBUG', message='Emitting Juju event start.')
+    assert line in ctx.juju_log
+    lines = ctx.juju_log[ctx.juju_log.index(line) :]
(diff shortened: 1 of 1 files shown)


testing/tests/test_e2e/test_juju_log.py
```diff
@@ -31,6 +31,8 @@ def _on_event(self, event):
 def test_juju_log(mycharm):
     ctx = Context(mycharm, meta=mycharm.META)
     ctx.run(ctx.on.start(), State())
-    assert ctx.juju_log[-2] == JujuLogLine(level='DEBUG', message='Emitting Juju event start.')
-    assert ctx.juju_log[-1] == JujuLogLine(level='WARNING', message='bar!')
+    line = JujuLogLine(level='DEBUG', message='Emitting Juju event start.')
+    assert line in ctx.juju_log
+    lines = ctx.juju_log[ctx.juju_log.index(line) :]
+    assert JujuLogLine(level='WARNING', message='bar!') in lines
```
dimaqq, line 37: I don't think the order of log lines matters here.
How about this instead?

```suggestion
    assert JujuLogLine(level='DEBUG', message='Emitting Juju event start.') in ctx.juju_log
    assert JujuLogLine(level='WARNING', message='bar!') in ctx.juju_log
```

dimaqq (approved):

dimaqq:
P.S. PR title prefix is `test:` not `tests:` 🤷🏻

testing/tests/test_e2e/test_juju_log.py
james-garner-canonical: Done

dwilding (approved):

PR #1696, opened 2025-04-25
chore: (mostly) automated migration to use annotations from the future

james-garner-canonical:
This PR migrates to use `from __future__ import annotations`, and use new-style annotations (like `X | Y | None`) in type annotations. This migration is mostly automatic, and does not migrate the style of type expressions used in other contexts like: `typing.cast` (though it does stringify these), or inheriting from generic classes, or type aliases (which can be stringified to use new-style annotations).

Manual changes are:
- adding `from __future__ import annotations` to every Python file (except empty `__init__.py` files)
- inlining one type alias which `pyright` and `flake8` had contradictory recommendations on stringification for (see [my comment](https://github.com/canonical/operator/pull/1696/files#r2059564820) on relevant lines)
- removing `annotations` from the namespace contents that `test_ops_testing_docs` checks that documentation is generated for (see [my comment](https://github.com/canonical/operator/pull/1696/files#r2059566099) on the relevant lines)

---

As with the recent quote style change PR, this PR contains just the automated (and minor manual) changes above. A follow-up PR will update `pyproject.toml` and add this PR to the git blame ignore list.

Additional PRs may follow to manually eliminate the remaining uses of old-style typing imports like `Union` (etc) in type aliases, casts, etc.


ops/model.py
```diff
@@ -335,21 +335,17 @@ def get_cloud_spec(self) -> CloudSpec:
...
         self._meta = meta
         self._backend = backend
         self._secret_set_cache: collections.defaultdict[str, dict[str, Any]] = (
             collections.defaultdict(dict)
         )
-        self._weakrefs: _WeakCacheType = weakref.WeakValueDictionary()
+        # (entity type, name): instance.
+        self._weakrefs: weakref.WeakValueDictionary[
+            tuple[UnitOrApplicationType, str], Unit | Application | None
+        ] = weakref.WeakValueDictionary()
```
james-garner-canonical, line 348: This is a manual change, resolving contradictory recommendations for strigification of `tuple`. In the original context, `pyright` wants it stringified because it's not in an annotation, while `flake8` does not, because it's behind `if typing.TYPE_CHECKING`. Since this type alias is private, type-checking only, and only used once, the cleanest solution seemed to be to just inline the annotation.

test/test_infra.py
```diff
@@ -67,6 +68,8 @@ def test_ops_testing_doc():
     expected_names.update(
         f'errors.{name}' for name in dir(ops.testing.errors) if not name.startswith('_')
     )
+    # we don't document `from __future__ import annotations`
+    expected_names.discard('errors.annotations')
```
james-garner-canonical, line 72: Manual change because `errors` now has `annotations` in its namespace due to `from __future__ import annotations`.

tonyandrewmeyer (approved):
Thanks for taking care of this!

It feels a bit odd to add the import for the files where nothing else changes, but I guess it's good to have it consistent across all the code.

benhoyt (approved):
Looks good to me. Thanks for breaking up into multiple manual-vs-automatic commits for easier review. I think it's fine to push this PR through and do an git-blame-ignore in a follow-up PR.

james-garner-canonical (commented):
Comment before merging main

PR #1697, opened 2025-04-25
full charm vroot copy

### LINKED ISSUE #1673: [feat] scenario: make a more complete copy of the charm's root
My charm's src directory looks like this:
```
src
├── charm.py
├── config.py
├── grafana_dashboards
│     └── overview-dashboard.json
├── loki_alert_rules
└── prometheus_alert_rules
        ├── otelcol_exporter.rules
        ├── otelcol_hardware.rules
        └── otelcol_receiver.rules
```
Scenario: The charm works as expected, so I write my first scenario test and it fails 😢.

```
scenario.errors.UncaughtCharmError: Uncaught exception (<class 'FileNotFoundError'>) in operator/charm code: FileNotFoundError(2, 'No such file or directory')
```
I realize I can solve this by adding an extra LOC in charm.py to create the dirs with `src_path.mkdir(parents=True, exist_ok=True)`. As a charm author I realize that adding code to the charm to make tests pass is bad practice. From the ops-scenario docs, I found that I can [create the virtual charm root](https://github.com/canonical/ops-scenario?tab=readme-ov-file#the-virtual-charm-root), like this:

```python
@pytest.fixture(scope="function")
def context(otelcol_charm, tmp_path):
    src_dirs = ["grafana_dashboards", "loki_alert_rules", "prometheus_alert_rules"]
    for src_dir in src_dirs:
        path_to_create = tmp_path / "src" / src_dir
        path_to_create.mkdir(parents=True, exist_ok=True)
    return Context(charm_type=otelcol_charm, charm_root=tmp_path)
```

All works and I feel more professional, but this makes me question whether ops-scenario should attempt to obtain this virtual charm root automatically. It can be quite the time sink solving this without guidance since the solution is not evident.

Little birdy told me that it would require this code `scenario._runtime.Runtime._virtual_charm_root` to have its functionality updated to make this happen. This would avoid having to create the virtual charm root as a workaround.

ping @PietroPasotti

PietroPasotti:
Fixes #1673 

This PR adds a step in `scenario._runtime.Runtime._virtual_charm_root` to do a recursive copy of the actual charm root into the Context-managed virtual charm root before (possibly over-)writing the metadata yamls and yielding the tempdir.

In a follow-up PR, I'd like to clean up `scenario._runtime.Runtime._virtual_charm_root` to more clearly separate the two 'charm-virtual-root-is-custom' branches using contextmanagers. I didn't do it now to keep this PR lean, but my fingers did itch.

### DIFF
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -6,9 +6,10 @@
 import copy
 import dataclasses
 import os
+import shutil
 import tempfile
 import typing
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
@@ -212,12 +213,17 @@ class Runtime:
         spec = self._charm_spec
 
         if charm_virtual_root := self._charm_root:
+            ctx = nullcontext()
             charm_virtual_root_is_custom = True
             virtual_charm_root = Path(charm_virtual_root)
         else:
-            charm_virtual_root = tempfile.TemporaryDirectory()
-            virtual_charm_root = Path(charm_virtual_root.name)
+            ctx = tempfile.TemporaryDirectory()
+            virtual_charm_root = Path(ctx.name)
             charm_virtual_root_is_custom = False
+
+            if spec.is_autoloaded:
+                # initialize the virtual charm root by making a recursive copy of the real charm root.
+                shutil.copytree(spec.get_charm_root(spec.charm_type), virtual_charm_root)
 
         metadata_yaml = virtual_charm_root / 'metadata.yaml'
         config_yaml = virtual_charm_root / 'config.yaml'
@@ -258,7 +264,8 @@ class Runtime:
         config_yaml.write_text(yaml.safe_dump(spec.config or {}))
         actions_yaml.write_text(yaml.safe_dump(spec.actions or {}))
 
-        yield virtual_charm_root
+        with ctx:
+            yield virtual_charm_root
 
         if charm_virtual_root_is_custom:
             for file, previous_content in metadata_files_present.items():
@@ -266,10 +273,6 @@ class Runtime:
                     file.unlink()
                 else:
                     file.write_text(previous_content)
-
-        else:
-            # charm_virtual_root is a tempdir
-            typing.cast(tempfile.TemporaryDirectory, charm_virtual_root).cleanup()  # type: ignore
 
     @contextmanager
     def _exec_ctx(self, ctx: 'Context'):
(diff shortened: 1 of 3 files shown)


PietroPasotti:
also, not sure what's going on, but in order to run anything with `tox` I need to comment out line 207 of tox.ini.
It doesn't seem to like `"{print \$1 \".\" \$2}"`.

image

testing/src/scenario/_runtime.py
```diff
@@ -212,13 +213,18 @@ def _virtual_charm_root(self):
...
         else:
-            charm_virtual_root = tempfile.TemporaryDirectory()
-            virtual_charm_root = Path(charm_virtual_root.name)
+            ctx = tempfile.TemporaryDirectory()
+            virtual_charm_root = Path(ctx.name)
             charm_virtual_root_is_custom = False
 
+            if spec.is_autoloaded:
+                # initialize the virtual charm root by making a recursive copy of the real charm root.
+                shutil.copytree(spec.get_charm_root(spec.charm_type), virtual_charm_root)
```
tonyandrewmeyer, line 226: I wonder if for the autoload case we ought to try to follow the charmcraft behaviour as defined in charmcraft.yaml (assuming it exists). For `python`, `uv`, and `poetry` plugins, that would mean copying only the `src` and `lib` trees, I think, and either requiring the test to handle anything else or maybe some support for the `dump` plugin behaviour. I can't find good docs for the `charm` plugin, but I think it's basically what's here (I thought there was an ignore file).

On the other hand, maybe having extra files is fine, and so it'd be better to keep it simpler and copy everything always.

As an unrelated second point, I wonder how expensive this ends up being. We probably can't sym/hard link since the charm might modify the files (unless we did that just for `lib` which is likely to have a lot of content but also seems like it ought to be read-only). This is likely to run at least once in every Scenario test, so even a small cost is going to quickly add up. The existing benchmark tests might show some info, although none of those would be copying much.

tonyandrewmeyer (commented):
My main concern here is the performance impact, since most tests autoload and we'll be copying the whole tree every time there's a new `Contest`. What are your thoughts on that?

tonyandrewmeyer:
> also, not sure what's going on, but in order to run anything with `tox` I need to comment out line 207 of tox.ini. It doesn't seem to like `"{print \$1 \".\" \$2}"`.

This works for me (and everyone else in Charm-Tech as far as I know; it's been there for a while), including when using Python 3.10. I'm using tox 4.25, is that what you have?

benhoyt:
We're going to close this PR based on https://github.com/canonical/operator/issues/1673#issuecomment-2883172223 but leave the issue open for future work.

PR #1701, opened 2025-04-28
feat: support for relation data schema as Python classes

tonyandrewmeyer:
**This is in draft until the [spec](https://docs.google.com/document/d/1G0xfWJXomzzH2VIY1trFq4b5ybMEEjV0w_aReLWyjCs/edit?tab=t.0) is accepted, so review should wait until it is moved from draft.**

A new base class, `ops.DatabagBase` is added. This currently provides *no* additional functionality, but we ask charmers to inherit from it (mix-in style when using Pydantic BaseModels) so that we can cleanly add new functionality (such as built-in basic validation or perhaps export to Starlark) in the future.

The core functionality is a new `load_data` method on `ops.Relation`. This takes a class (which should inherit from `ops.DatabagBase` as noted above) and returns an instance of that class with the current relation data loaded into it. Additional arguments are also able to be passed in if the class contains more than just the relation data.

In addition:
* At the end of the hook (right after collecting statuses), the current state of any loaded databag is sent through to Juju.
* When sending data to Juju, and when loading it into the class, we automatically serialise/deserialise to/from JSON.
* When using the existing `Relation.data` object to get a `RelationDataContent` object, that object will return values from any class that has had the values loaded, so that mixing the old and new style works.
* When using the existing `Relation.data` object to get a `RelationDataContent` object, and setting values, that object will also update any class that has had the values loaded, so that mixing the old and new style works.

A new exception, `ops.InvalidSchemaError` is also added. This is raised from any `ValueError` (or subclass) when instantiating the class. In the related config and action parameter PRs, we catch this error and exit with a zero code.

The implementation is intended to be duck-typed, in that any class that accepts the data as keyword arguments and raises `ValueError` (or a subclass) in instantiation and/or setting attributes will work. However, the tests verify four specific cases:
* A generic Python class (no inheritance other than `object`).
* Standard library dataclasses (I personally believe these are the best choice for most cases).
* Pydantic dataclasses

We don't explicitly promise that these exact cases work, but they should continue to do so while they rely on the current implementation. This leaves it open to support or not other cases (e.g. a potential Pydantic 3.x, `attrs`, and so on).

In order to test that these 4 types work, the unit tests add specific tests if Pydantic is installed, and the `tox` environment for `static` and `unit` install Pydantic. Ops itself does *not* require Pydantic (at run time or for tests).


ops/model.py
```diff
@@ -922,7 +932,7 @@ def __init__(
         self._backend = backend
         self._cache = cache
         self._broken_relation_id = broken_relation_id
-        self._data: _RelationMapping_Raw = {r: None for r in relations_meta}
+        self._data: _RelationMapping_Raw = dict.fromkeys(relations_meta)
```
tonyandrewmeyer, line 935: This was done automatically by the format (or pre-commit?) tooling.

ops/model.py
```diff
@@ -349,6 +358,7 @@ def __init__(self, meta: _charm.CharmMeta, backend: _ModelBackend):
         self._secret_set_cache: collections.defaultdict[str, dict[str, Any]] = (
             collections.defaultdict(dict)
         )
+        self._databag_obj_cache: dict[tuple[int, Union[Application, Unit]], Any] = {}
```
tonyandrewmeyer, line 361: Now that the future annotations are in this file, this can be:

```suggestion
        self._databag_obj_cache: dict[tuple[int, Application | Unit], Any] = {}
```

ops/model.py
```diff
@@ -1787,19 +1863,28 @@ class RelationData(Mapping[Union[Unit, Application], 'RelationDataContent']):
...
+        relation: Relation,
+        our_unit: Unit,
+        backend: _ModelBackend,
+        cache: _ModelCache | None = None,
+    ):
         self.relation = weakref.proxy(relation)
-        self._data: dict[Unit | Application, RelationDataContent] = {
-            our_unit: RelationDataContent(self.relation, our_unit, backend),
-            our_unit.app: RelationDataContent(self.relation, our_unit.app, backend),
+        self._data: dict[Union[Unit, Application], RelationDataContent] = {
```
tonyandrewmeyer, line 1874: Missed in the merge:

```suggestion
        self._data: dict[Unit | Application, RelationDataContent] = {
```

ops/model.py
```diff
@@ -2186,7 +2301,7 @@ class Resources:
 
     def __init__(self, names: Iterable[str], backend: _ModelBackend):
         self._backend = backend
-        self._paths: dict[str, Path | None] = {name: None for name in names}
+        self._paths: dict[str, Path | None] = dict.fromkeys(names)
```
tonyandrewmeyer, line 2304: Also done via the tooling.

ops/model.py
```diff
@@ -1771,6 +1782,71 @@ def remote_model(self) -> RemoteModel:
             self._remote_model = RemoteModel(uuid=d['uuid'])
         return self._remote_model
 
+    def load_data(
+        self,
+        cls: type[_InterfaceType],
+        app_or_unit: Union[Unit, Application],
+        *args: Any,
+        decoder: Callable[[Any], Any] | None = None,
+        encoder: Callable[[Any], Any] | None = None,
```
tonyandrewmeyer, line 1791: Comment from spec review: these could be tighter. The return type will be `str` because that's all that Juju will accept. The argument probably has to stay as `Any`, I think.

ops/model.py
```diff
@@ -1771,6 +1782,71 @@ def remote_model(self) -> RemoteModel:
...
+            return obj
+        data = copy.deepcopy(kwargs)
+        if decoder is None:
+            decoder = json.loads
+        content = {k: decoder(v) for k, v in self.data[app_or_unit].items()}
+        data.update(**content)
+        try:
+            obj = cls(*args, **data)
+        except ValueError as e:
+            self._backend.status_set('waiting', f'Error in relation data: {e}')
```
tonyandrewmeyer, line 1845: This should maybe be done at the end when _main catches the exception, so that if the charm catches it the status isn't set (which would send through immediately to Juju, and might override a status the charm already set).

ops/charm.py
```diff
@@ -1448,6 +1449,29 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
...
+        if encoder is None:
+            encoder = json.dumps
+        if hasattr(obj, '__dataclass_fields__'):
+            # A standard library dataclass or Pydantic dataclass.
+            fields = (field.name for field in dataclasses.fields(obj))
+            values = dataclasses.asdict(obj)
+        elif hasattr(obj, 'model_fields'):
+            # A Pydantic model.
+            fields = obj.model_fields
+            values = obj.model_dump()
```
tonyandrewmeyer, line 1463: For what it's worth, the obs charms pass in extra args here, `mode="json", by_alias=True, exclude_defaults=True`

ops/model.py
```diff
@@ -1771,6 +1782,71 @@ def remote_model(self) -> RemoteModel:
...
+                    f'it is already loaded into {type(obj).__name__}'
+                )
+            return obj
+        data = copy.deepcopy(kwargs)
+        if decoder is None:
+            decoder = json.loads
+        content = {k: decoder(v) for k, v in self.data[app_or_unit].items()}
+        data.update(**content)
+        try:
+            obj = cls(*args, **data)
```
tonyandrewmeyer, line 1843: Look into `cls.model_validate_json(json.dumps(data)) `

test/test_databag.py
```diff
@@ -0,0 +1,527 @@
...
+        def databag_class(self) -> type[ops.DatabagBase]:
+            return MyPydanticDataclassDatabag
+
+    class MyPydanticDatabag(pydantic.BaseModel, ops.DatabagBase):
+        foo: str
+        baz: list[str] = pydantic.Field(default_factory=list)
+        bar: int = pydantic.Field(default=0, ge=0)
+        quux: Nested = pydantic.Field(default_factory=Nested)
+
+        class Config:
```
tonyandrewmeyer, line 265: Consider:

```
model_config = ConfigDict(
            # tolerate additional keys in databag
            extra="ignore",
            # Allow instantiating this class by field name (instead of forcing alias).
            populate_by_name=True,
        )
```

ops/charm.py
```diff
@@ -1435,6 +1435,49 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
         unit.status = model.StatusBase._get_highest_priority(unit._collected_statuses)
 
 
+def _juju_fields(cls: type[object]) -> dict[str, str]:
```
tonyandrewmeyer, line 1604: This is a copy of the code from the actions PR. The only difference is that it uses `if dataclasses.is_dataclass(cls):` rather than `try/except TypeError`, which is cleaner. When this is merged it'll improve the actions PR too, without changing the behaviour.

tonyandrewmeyer:
@dwilding I think we should entirely rework the how-to guides for "libraries", "interfaces", and "relations", figuring out a better structure, aligning more to the actual charming use-cases, removing duplicate content, making them clearer guides, and so on. I've added some mention of the new methods in those but it seems like it would be better to do the rework and include it then rather than try to shove it in as-is. I'll add this as a topic for our next chat.

ops/model.py
```diff
@@ -1767,6 +1774,136 @@ def remote_model(self) -> RemoteModel:
...
+        The raw Juju relation data is passed to the data class's ``__init__``
+        method as keyword arguments, with values decoded using the provided
+        decoder function, or :func:`json.loads` if no decoder is provided.
+
+        Any additional positional or keyword arguments will be passed through to
+        the data class ``__init__``.
+
+        Args:
+            cls: A class that will accept the Juju relation data as keyword
+                arguments, and raise ``ValueError`` if validation fails.
```
benhoyt, line 1842: I think we should add something like "This is typically a Pydantic base model or a dataclass."

ops/model.py
```diff
@@ -1767,6 +1774,136 @@ def remote_model(self) -> RemoteModel:
...
+        **kwargs: Any,
+    ) -> _T:
+        """Load the data for this relation into an instance of a data class.
+
+        The raw Juju relation data is passed to the data class's ``__init__``
+        method as keyword arguments, with values decoded using the provided
+        decoder function, or :func:`json.loads` if no decoder is provided.
+
+        Any additional positional or keyword arguments will be passed through to
+        the data class ``__init__``.
```
benhoyt, line 1837: Might be good to have a tiny example of usage here? Don't even need to have the data class definition. Something like:

```
Example::

    relation = self.model.get_relation('db')
    data = relation.load(DatabaseModel, self.app)
    secret_id = data.credentials
```

ops/model.py
```diff
@@ -1767,6 +1774,136 @@ def remote_model(self) -> RemoteModel:
...
+    ):
+        """Save the data from the provided data class object to the Juju relation data.
+
+        Args:
+            obj: an object with attributes to save to the relation data.
+            app_or_unit: The destination in which to save the data to save. This
+                can be either a :class:`Unit` or :class:`Application` instance.
+            encoder: An optional callable that will be used to encode each field
+                before passing to Juju. If not provided, :func:`json.dumps` will
+                be used.
```
benhoyt, line 1916: I think a tiny example would be good here too.

benhoyt (approved):
I haven't gone through the tests closely (another reviewer should though :-) but this looks good to me, and very much in keeping with the other PRs. Couple of minor comments re docs/examples.

james-garner-canonical:
> The errors argument for load primarily exists for consistency with the errors argument of load_config and load_params. There is no errors argument for save: the charm is always expected to handle any issues.

Do you think it would make future maintenance and documentation/recommendations cleaner if we dropped the errors argument for relations? There are already some asymmetries between relation data and the other two (no schema generation for one thing), and I'm not sure that going to blocked on load will be the right choice enough of the time to justify having this option:

> ... if the error is with remote data that most likely shouldn't change the local unit status, and even when handling local data status is complicated and a simple "bad data in one relation means blocked" will only be true in simple situations

That said, if you think the symmetry does make it significantly easier for all involved, I'm on board with keeping it

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+        *args: Any,
+        decoder: Callable[[str], Any] | None = None,
+        errors: Literal['raise', 'blocked'] = 'raise',
+        **kwargs: Any,
+    ) -> _T:
+        """Load the data for this relation into an instance of a data class.
+
+        The raw Juju relation data is passed to the data class's ``__init__``
+        method as keyword arguments, with values decoded using the provided
+        decoder function, or :func:`json.loads` if no decoder is provided.
```
james-garner-canonical, line 1812: Will we document anywhere publicly that dataclasses and Pydantic models are special cased to only be passed the keys corresponding to their fields?

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+        if decoder is None:
+            decoder = json.loads
+        for key, value in sorted(self.data[app_or_unit].items()):
+            value = decoder(value)
+            if fields is None:
+                data[key] = value
+            else:
+                if key not in fields:
+                    continue
+                data[fields[key]] = value
```
james-garner-canonical: I wonder if this is significantly better? Reject if not
```suggestion
            elif key in fields:
                data[fields[key]] = value
```

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+            raise _Abort(0) from e
+
+    def save(
+        self,
+        obj: object,
+        app_or_unit: Unit | Application,
+        *,
+        encoder: Callable[[Any], str] | None = None,
+    ):
+        """Save the data from the provided data class object to the Juju relation data.
```
james-garner-canonical: Is 'data class object' likely to be confusing? Maybe it will nudge people towards `dataclasses` which would be nice.

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
             self._remote_model = RemoteModel(uuid=d['uuid'])
         return self._remote_model
 
+    def load(
+        self,
+        cls: type[_T],
+        app_or_unit: Unit | Application,
```
james-garner-canonical: Thought: wdyt about flipping the order of these arguments (and likewise for `save`)? The app or unit seems conceptually to be at the top level for me.

I guess the real problem is two positional arguments without a natural ordering. Making one of them keyword only doesn't seem great though. Maybe something like `Relation.data[app_or_unit].load(cls)` or `Relation.typed_data[app_or_init].load(cls)` would be nice.

Not too big a deal though, if you think the current ordering is natural enough that's fine with me -- IDEs shold make it easy enough for users anyway.

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+            fields = {k: k for k in get_type_hints(obj.__class__)}
+            values = {field: getattr(obj, field) for field in fields}
+
+        # Encode each value, and then pass it over to Juju.
+        data: dict[str, str] = {}
+        for attr, field in sorted(fields.items()):
+            if not isinstance(field, str):
+                raise ValueError(
+                    f'The value for "{field}" must be a string, not {type(field).__name__}'
+                )
```
james-garner-canonical: Perhaps a `TypeError` would be more appropriate here? Also I wonder if it would be more helpful to raise this error where `fields` is created -- I'm guessing that a Pydantic model would probably complain if `alias` wasn't a string, so it's probably only needed after `alias = ...` in the `dataclass` branch.

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+            app_or_unit: The destination in which to save the data to save. This
+                can be either a :class:`Unit` or :class:`Application` instance.
+            encoder: An optional callable that will be used to encode each field
+                before passing to Juju. If not provided, :func:`json.dumps` will
+                be used.
+
+        Raises:
+            ValueError: if the encoder does not return a string.
+            RelationNotFoundError: if the relation does not exist.
+            ModelError: if the charm does not have permission to write to the
```
james-garner-canonical: We're going via the `RelationDataContent` object, so would this always be a `RelationDataAccessError`?

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+                relation data.
+        """
+        if encoder is None:
+            encoder = json.dumps
+
+        # Determine the fields, which become the Juju keys, and the values for
+        # each field.
+        fields: dict[str, str] = {}  # Class attribute name: Juju key.
+        if dataclasses.is_dataclass(obj):
+            assert not isinstance(obj, type)
```
james-garner-canonical: A little confusing on its own. I guess type checking wise an `object` can be a class.

Interesting that this must come after `dataclasses.is_dataclass` to work -- that feels like a type checker bug (`not isinstance(obj, type) and dataclasses.is_dataclass(obj)` should be equivalent to `dataclasses.is_dataclass(obj) and not isinstance(obj, type)`, but they result in different revealed types).

Maybe this `assert` is cleaner on the line before we need it (the `asdict` call)? I'm happy for it to remain here if you prefer that though. Either way, a comment might be nice, maybe something like `# dataclass instance, not dataclass itself`

ops/model.py
```diff
@@ -1767,6 +1774,149 @@ def remote_model(self) -> RemoteModel:
...
+                # Pydantic takes care of the alias.
+                fields[field.alias or name] = field.alias or name  # type: ignore
+            values = obj.model_dump(mode='json', by_alias=True, exclude_defaults=False)  # type: ignore
+        else:
+            # If we could not otherwise determine the fields for the class,
+            # store all the fields that have type annotations. If a charm needs
+            # a more specific set of fields, then it should use a dataclass or
+            # Pydantic model instead.
+            fields = {k: k for k in get_type_hints(obj.__class__)}
+            values = {field: getattr(obj, field) for field in fields}
```
james-garner-canonical, line 1952: Interestingly, this will raise an attribute error for any annotations that don't have values. I wonder if it would be better to skip such fields ... though `dataclasses.asdict` also raises an `AttributeError` if one of the `dataclasses.fields` doesn't have a value, so ultimately I reckon this is fine as is.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+
+@dataclasses.dataclass
+class Nested:
+    sub: int = 28
+
+
+class DatabagProtocol(Protocol):
+    foo: str
+    baz: list[str]
+    bar: int
```
james-garner-canonical: Very minor and also silly: how about swapping these variable names, unless doing 'foo, baz, bar' instead of 'foo bar, baz' is intentional ... don't worry if it's a pain though

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+
+    def _on_relation_changed(self, _: ops.RelationChangedEvent) -> None:
+        raise NotImplementedError('databag class must implement this')
+
+    @property
+    def databag_class(self) -> type[DatabagProtocol]:
+        raise NotImplementedError('databag_class must be set in the subclass')
+
+    encoder: Callable[[Any], str] | None = None
+    decoder: Callable[[str], Any] | None = None
```
james-garner-canonical: Minor: attributes are typically defined before `__init__`, aren't they?

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+nested_decode = functools.partial(json.loads, object_hook=json_nested_hook)
+
+
+class MyCharm(BaseTestCharm):
+    @property
+    def databag_class(self) -> type[DatabagProtocol]:
+        return MyDatabag
+
+    encoder: Callable[[Any], str] | None = nested_encode
+    decoder: Callable[[str], Any] | None = nested_decode
```
james-garner-canonical: Minor: ditto

ops/model.py
tonyandrewmeyer: Fair question :). Currently, we don't, in config, actions, or here. It felt like that was an implementation detail, but I guess as soon as it's in a release, it's part of the API and can't really change in a backwards-incompatible way. Now that the 'don't know' case is 'give everything' it does feel like something that's more easily documented and less likely to change. I was also originally trying to avoid mentioning Pydantic explicitly but that bird has flown now.

Would you say that the right thing to do here would be to add something like this to `CharmBase.load_config`, `ActionEvent.load_params`, and `Relation.load`?

> For dataclasses and Pydantic `BaseModel` subclasses, only {options, parameters, fields} that have a matching field in the class are passed as arguments. Pydantic fields that have an `alias`, or dataclasses that have a `metadata{'alias'=}`, will have the Juju {option, parameter, field} name transformed to the alias when setting the class field.

And for `Relation.save`:

> For dataclasses and Pydantic `BaseModel` subclasses, only the class's fields will be saved through to the relation data. Pydantic fields that have an `alias`, or dataclasses that have a `metadata{'alias'=}`, will have the object's value saved until the alias name. For other classes, all of the object's attributes that have a class type annotation will be saved through to the relation data.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+        self.other = other
+
+    # This is pretty quirky, but we need `fooBar` to be in the type annotations
+    # and we need it to return the value of `foo_bar` to correctly save back to
+    # Juju. Other than being ugly to look at, this means that the class offers
+    # `.fooBar` as well as `.foo_bar`. In practice, the expectation is that
+    # charms that need aliases like this should use dataclasses or pydantic.
+    # We have this here so that we can still have the standard set of four
+    # classes being tested.
+    fooBar: int = property(lambda self: self.foo_bar)  # type: ignore  # noqa: N815
```
james-garner-canonical, line 290: Wow! That's a heck of a workaround, nice.

Minor: would be good to put both annotations together imo.

ops/model.py
james-garner-canonical: Yes, that sounds good to me. Maybe make {options, parameters, fields} more specific? 'Juju config opions', 'Juju action parameters', 'relation data keys/items/entries'. Maybe 'passed as keyword arguments'?

The alias explanation is a little confusing to me. Maybe 'will look up the alias value in the [Juju config options] instead'? Perhaps followed by an example.

For saving relation data, something that came up (review comment pending) is whether we should attribute error for vanilla classes if an annotation key isn't actually set on the object (current behaviour). I leaned towards keeping this behaviour, but the explanation (your proposed explanation) seems cleaner if we filter it to attributes that are actually set.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+
+
+@dataclasses.dataclass
+class _DataclassesAlias:
+    foo_bar: int = dataclasses.field(default=42, metadata={'alias': 'fooBar'})
+    other: str = 'baz'
+
+
+_alias_data_classes: list[type[object]] = [_Alias, _DataclassesAlias]
```
james-garner-canonical: I guess `data_classes` means 'relation data' classes, but how about just `_alias_classes` to avoid confusion with `dataclasses`.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+        foo_bar: int = pydantic.Field(42, alias='fooBar')
+        other: str = pydantic.Field('baz')
+
+    _alias_data_classes.extend([_PydanticDataclassesAlias, _PydanticBaseModelAlias])
+
+
+@pytest.mark.parametrize('relation_data', [{}, {'fooBar': '24'}])
+@pytest.mark.parametrize('data_class', _alias_data_classes)
+def test_relation_load_custom_naming_pattern(
+    relation_data: dict[str, str], data_class: type[_AliasProtocol]
```
james-garner-canonical: Maybe `relation_data_class`?

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+    def __init__(self, foo: str):
+        self.foo = foo
+
+
+@dataclasses.dataclass
+class _DataclassesOneString:
+    foo: str
+
+
+_one_string_data_classes: list[type[object]] = [_OneString, _DataclassesOneString]
```
james-garner-canonical:
```suggestion
_one_string_classes: list[type[object]] = [_OneString, _DataclassesOneString]
```

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+        return str(data)
+
+    def custom_decode(data: str) -> Any:
+        if hasattr(data, '__getitem__'):
+            return data[::-1]
+        return data
+
+    class Charm(charm_class):
+        def _on_relation_changed(self, event: ops.RelationChangedEvent):
+            data = event.relation.load(self.databag_class, self.app, decoder=custom_decode)
```
james-garner-canonical, line 555: Let's add an assert on the line below to make this easier to follow. I guess `data.foo` is `'eulav'` here.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+    ): ...
+
+
+class BaseTestCharmCommonTypes(ops.CharmBase):
+    @property
+    def databag_class(self) -> type[CommonTypesProtocol]:
+        raise NotImplementedError('databag_class must be set in the subclass')
+
+    encoder: Callable[[Any], str] | None = None
+    decoder: Callable[[Any], str] | None = None
```
james-garner-canonical: Minor: attributes first?

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+
+
+class CommonTypes(BaseTestCharmCommonTypes):
+    @property
+    def databag_class(self):
+        class Data:
```
james-garner-canonical: Can we unnest this and follow the convention used in the other tests? I'd only define the class here if you're using the `self` value from `databag_class(self)` here somehow.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+    NZ = 'New Zealand'
+    JP = 'Japan'
+    CN = 'China'
+
+
+class CommonTypesProtocol(Protocol):
+    url: str
+    ip: ipaddress.IPv4Address | ipaddress.IPv6Address | str
+    network: ipaddress.IPv4Network | ipaddress.IPv6Network | str
+    origin: Country | str
```
james-garner-canonical: Should these all have `| str`, or will that be validated away?

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+            obj,
+            (
+                ipaddress.IPv4Address,
+                ipaddress.IPv6Address,
+                ipaddress.IPv4Network,
+                ipaddress.IPv6Network,
+            ),
+        ):
+            return str(obj)
+        # Not an IP, but for the tests simplest to just include it here.
```
james-garner-canonical: Does this read better if we rename the class and drop the comment? e.g. `CommonTypesEncoder`

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+@staticmethod
+def json_ip_hook(dct: dict[str, Any]) -> dict[str, Any]:
+    for key, value in dct.items():
+        try:
+            if '/' in value:
+                dct[key] = ipaddress.ip_network(value, strict=False)
+            else:
+                dct[key] = ipaddress.ip_address(value)
+        except ValueError:  # noqa: PERF203
+            pass
```
james-garner-canonical, line 691: I guess we get a `ValueError` if it's just not an IP address? Maybe:
```suggestion
            pass  # not an IP address
```

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+
+        return Data
+
+    @staticmethod
+    def _str_encoder(x: Any) -> str:
+        if isinstance(x, Country):
+            return json.dumps(x.value)
+        return json.dumps(str(x))
+
+    encoder: Callable[[Any], str] | None = _str_encoder
```
james-garner-canonical: Do we need the type annotation if this is annotated in the base class? Also I guess the type checker complains, but if it's possible to just do `def encoder` that would be nice.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+        class Data:
+            url: str
+            ip: ipaddress.IPv4Address | ipaddress.IPv6Address
+            network: ipaddress.IPv4Network | ipaddress.IPv6Network
+            origin: Country | str
+
+        return Data
+
+    encoder: Callable[[Any], str] | None = json_ip_encode
+    decoder: Callable[[str], Any] | None = json_ip_decode
```
james-garner-canonical: Annotations unneeded? Attributes first in the class definition?

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+
+json_ip_encode = functools.partial(json.dumps, cls=IPJSONEncoder)
+json_ip_decode = functools.partial(json.loads, object_hook=json_ip_hook)
+
+
+class CommonTypesDataclasses(BaseTestCharmCommonTypes):
+    @property
+    def databag_class(self):
+        @dataclasses.dataclass
+        class Data:
```
james-garner-canonical: This is a lot simpler than the vanilla class case, but I'd still rather not nest this definition.

test/test_model_relation_data_class.py
```diff
@@ -0,0 +1,900 @@
...
+def test_relation_common_types(charm_class: type[BaseTestCharmCommonTypes]):
+    class Charm(charm_class):
+        def __init__(self, framework: ops.Framework):
+            super().__init__(framework)
+            framework.observe(self.on['db'].relation_changed, self._on_relation_changed)
+
+        def _on_relation_changed(self, event: ops.RelationChangedEvent):
+            data: CommonTypesProtocol = event.relation.load(
+                self.databag_class, event.app, decoder=self.decoder
+            )
```
james-garner-canonical, line 802: Worth doing any `assert`s at this point? e.g. `assert isinstance(data.url, IP...)`

tonyandrewmeyer:
> Do you think it would make future maintenance and documentation/recommendations cleaner if we dropped the errors argument for relations? 
> [...]
> That said, if you think the symmetry does make it significantly easier for all involved, I'm on board with keeping it

Fair question here too :). I added this at @benhoyt's suggestion (in Frankfurt) for what it's worth.

If you're using collect-status, which is what we're recommending in general, then I think you're more likely to have:

```python
def _on_something(...):
    try:
        data = relation.load(MyClass, self.unit)
    except ValueError:
        return

def _on_collect_status(self, event: ops.CollectStatusEvent):
    try:
        self.model.get_relation(x).load(MyCharm, self.unit)
    except ValueError:
        event.add_status(...)
```

Than have the collect-status use `self.model.get_relation(x).load(MyCharm, self.unit, errors='blocked')` - it seems strange to have an explicit status set in collect-status. If you're using collect-status then you almost certainly don't want the load where you use the data to do `errors='blocked'` because the next event will override that.

Even when not using collect-status, it's unclear whether the blocked status would always be the one you wanted, because it's not clear whether how often it would mean the entire app is "blocked". If we did have more complex statuses (like a per-relation status) then it seems like it would be much more viable.

I wondered about an `errors: Literal['raise'] = raise` to have consistency (the non-'raise' options differ already) and an indication that we might offer more in the future. But it seemed like the consistency there doesn't offer that much and just leaving it out would be simpler.

Overall, I'm happy to remove it, but if @benhoyt prefers that it stays, I'm fine with that too.

james-garner-canonical (approved):
Look good overall. Some nitpicking about things in the tests that I found made them harder to read. A couple of thoughts about errors that should be addressed: what error to raise and where to raise from if the alias isn't a string (on save), and whether to raise an attribute error when saving vanilla classes if there's an annotation without an actual value.

james-garner-canonical:
Regarding the `errors` argument, your post makes me lean even more towards removing it (we can always add it with a default if we do want to make it customisable eventually), but let's get @benhoyt's take.

ops/model.py
tonyandrewmeyer: Any suggestions for alternative phrasing? Just "from the object to the Juju relation data" seems like it's missing an important piece. I don't really want to say "from the dataclass or Pydantic `BaseModel` subclass" and be that specific.

benhoyt:
> But it seemed like the consistency there doesn't offer that much and just leaving it out would be simpler.

Yeah, I agree we should probably remove `errors` for now. As James points out, it's easy to add as a keyword arg with a default later if we need it.

ops/model.py
tonyandrewmeyer: > Thought: wdyt about flipping the order of these arguments (and likewise for `save`)? The app or unit seems conceptually to be at the top level for me.

I read it as "save {data} into {app-or-unit}", but I can see how "save into {app-or-unit} the values from {data}" could be expected. It's somewhat based on the order in `relation.data[{app-or-unit}]` although there "data" is the actual characters "data" rather than an object, of course.

Your point about which databag it's going into being the higher-level choice is quite compelling.

The charms with existing patterns like this tend to bypass this question by having the save (dump) method on the data class instead, but we decided against that approach (and don't have a class to put it on any more either).

> I guess the real problem is two positional arguments without a natural ordering. Making one of them keyword only doesn't seem great though.

`rel.save(self.app, obj=data)` isn't terrible - I think charmers do expect to have to specify which databag so having that as position is ok. I don't really like `rel.save(obj, unit_or_app=self.app)` or `rel.save(obj=data, unit_or_app=self.app)`.

What about `rel.save(obj, to=self.app)`? It still has the object as positional but reads nicely.

A @benhoyt opinion here would be good too, please.

> Maybe something like `Relation.data[app_or_unit].load(cls)` or `Relation.typed_data[app_or_init].load(cls)` would be nice.

This idea came up in Frankfurt as well (we probably should have done that discussion in the second week so you could be there). It felt (to Ben and to me and at least some others) too magic given that `Relation.data` is already quite a specialised mapping.

A `Relation.typed_data[app_or_unit]` as a place to hold `load` and `save` didn't come up. I guess `typed_data` is a mapping of `Application | Unit` to a new type of object that just has those two methods on it? I like how it looks similar to what we already have, but I don't like introducing something just to hold the methods.

It feels like it's trying to be `Relation.save[app_or_unit](obj)` -- and I really don't like that :)

james-garner-canonical: `rel.save(self.app, obj=data)` does seem ok. The name would need to be different for load though, e.g. `rel.load(self.app, cls=my_class)`. Would we be want to make this the order? Would we want to make the second argument keyword only?

I'm on board with ruling out `typed_data` (and `data`) for this. I don't like `unit_or_app` as a keyword argument either. I quite like the `to` keyword argument for `save`  but it kind of implies a `from` argument for `load` (and `from` is a Python keyword).

ops/model.py
benhoyt: Yeah, could just drop "data class": "Save the data from the provided object to the Juju relation data."

ops/model.py
benhoyt: I think what we have is good for a couple of reasons:

- For `load(MyClass, app_or_unit)`, it puts the class first, which matches `load_config(MyClass)` and `load_params(MyClass)`
- `save(obj, app_or_unit)` makes it mimic the order of `load(cls, app_or_unit)`
- It's the same order as `json.dump(obj, fp)` and similar `dump` functions

james-garner-canonical: Ok, let's leave the ordering as is then, though I wouldn't object to a nicer argument name than `app_or_unit` in case people do want to use it as a keyword argument.

ops/model.py
tonyandrewmeyer: > A little confusing on its own. I guess type checking wise an `object` can be a class.

Yes. The [docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.is_dataclass) even have something fairly similar to this (except I know it will be an obj here, so I'm asserting).

> Maybe this `assert` is cleaner on the line before we need it (the `asdict` call)? I'm happy for it to remain here if you prefer that though. Either way, a comment might be nice, maybe something like `# dataclass instance, not dataclass itself`

I like it together because what I really want it to say is "if this is a dataclass obj". I can add a comment.

I could make it `if dataclasses.is_dataclass(obj) and not isinstance(obj, type)` if you feel it's better all in the `if` rather than partly in the `assert`. I suppose someone could mistakenly pass in the class like they do for `load`, but if that happens then I don't think they should be falling through to the "generic Python class" code.

james-garner-canonical: I agree that the `assert` (or some other violent error) is better than including it in the `if`. Happy for it to stay where it is with a comment added, thanks.

ops/model.py
tonyandrewmeyer: I could have sworn that I took `app_or_unit` from something that already existed in ops, but I can't find anywhere now, so maybe that was all in my head! app_or_unit is too duplicative of the type annotation so isn't a great name anyway. Looking at the existing names in ops, `entity` would be the most consistent, but tbh I like `to` more, but yes `from_` is always awkward.

tonyandrewmeyer: > * It's the same order as `json.dump(obj, fp)` and similar `dump` functions

Ah yes, this was another motivator, I had forgotten that, thanks! Although it's also one of the ones where I always struggle to remember which order the args are in :)

james-garner-canonical: I think `app_or_unit` is used in `jubilant`, maybe that's what you were thinking of. I like `to` for `save`, but then would you do `from_` for `load`?

ops/model.py
tonyandrewmeyer: I'd prefer to leave it as-is. I feel like a strong advantage of the current approach is its simplicity, particularly compared to your 'create a new dataclasses type' and my 'combine __dict__ and __annotations and then filter them'.

I think if you have a class attribute with a type and it's not set that is most likely to be a mistake where you've forgotten to remove the attribute or have a typo somewhere. This seems pretty likely to fail during testing (or maybe even during static types if you're also trying to use it) which seems ok.

james-garner-canonical: Sounds good to me

test/test_model_relation_data_class.py
tonyandrewmeyer: Not an IP address or an IP network, depending on whether the string has `/` in it. Maybe it's just that I've used the ipaddress module (and its ancestor, ipaddr) for ages so I'm super familiar with it, but isn't `x_type(value)` raising a `ValueError` self-explanatory that `value` wasn't an `x` type?

ops/model.py
tonyandrewmeyer: Does it look ok now?

ops/model.py
tonyandrewmeyer: > I like `to` for `save`, but then would you do `from_` for `load`?

Yeah, I don't love `from_`. I'm not a big fan of `entity`, either - other than in the ops internals, I don't think that's the normal term.

It seems like the charming term would be "databag" - I'm not sure if that's an official term or not. I'll see what Juju says.

ops/model.py
tonyandrewmeyer: Yes, `TypeError` would be better, thanks - that was leftover from when I was trying to have `ValueError` consistently raised everywhere, but that's not really relevant any more.

tonyandrewmeyer: In terms of when it can happen - this is on the field *value*, not the *key*, so the aliasing isn't important. This happens if you have a bad encoder (rather than json.dumps) that returns something non-string.

However, the static checks ought to take care of it, and then `Relation._validate_write_content` will do this check (I didn't originally go through a path that would hit that, but it will now). So I'll just remove it.

ops/model.py
tonyandrewmeyer: Yes, I guess so, assuming that we are completely re-implementing the Juju relation access model.

ops/model.py
james-garner-canonical: I think the alias explanation could use a little more. I guess that will be in common with the other PRs too, but I think I only understand it because I've looked at the code, specifically the idea that the alias is the name in the Juju data source.

ops/model.py
james-garner-canonical: I guess `ModelError` covers us if we let one through.

docs/howto/manage-relations.md
```diff
@@ -156,7 +180,8 @@ def _update_configuration(self, _: ops.Eventbase):
     if not db_relation:
         # We’re not integrated with the database charm yet.
```
dwilding: I saw you changed several `’`, so might as well change this too:

```suggestion
        # We're not integrated with the database charm yet.
```

docs/howto/manage-relations.md
```diff
@@ -156,7 +180,8 @@ def _update_configuration(self, _: ops.Eventbase):
...
         return
-    secret_id = db_relation.data[self.model.app]['credentials']
+    data = db_relation.load(DatabaseProviderAppData, self.app)
+    secret_id = data.credentials
     if not secret_id:
         # The credentials haven’t been added to the relation by the remote app yet.
```
dwilding:
```suggestion
        # The credentials haven't been added to the relation by the remote app yet.
```

docs/howto/manage-relations.md
```diff
@@ -106,12 +121,21 @@ To do additional setup work when each unit joins the relation (both when the cha
 framework.observe(self.on.smtp_relation_joined, self._on_smtp_relation_joined)
 ```
 
+In `src/charm.py`, create a class that defines the schema for the relation data.
+For example:
+
+```python
+class SMTPProviderUnitData(pydantic.BaseMode):
+    smtp_credentials: str = pydantic.Field(description="A Juju secret ID")
+```
```
dwilding, line 130: I got a bit confused at first, because this schema is almost identical to the one you added above. But I can see that you needed to do it this way, because the existing example for relation_joined is different from the existing example for relation_created.

I'm sure we can improve the overall structure in the future. For now, for a small edit to help avoid confusion, how about changing "To do additional setup work when each unit joins the relation" (above) to something like:

```
To provide SMTP credentials to each unit that joins the `smtp` relation
```

docs/howto/manage-relations.md
```diff
@@ -67,35 +67,50 @@ Other than this, implement a subordinate relation in the same way as any other r
...
 
+> See more: [](manage-libraries-write-a-library)
+
 (set-up-a-relation)=
 ##### Set up a relation
 
-To do initial setup work when a charm is first integrated with another charm (or, in the case of a peer relation, when a charm is first deployed) your charm will need to observe the relation-created event. For example, a charm providing a database relation might need to create the database and credentials, so that the requirer charm can use the database. In the `src/charm.py` file, in the `__init__` function of your charm, set up `relation-created` event observers for the relevant relations and pair those with an event handler.
+To do initial setup work when a charm is first integrated with another charm (or, in the case of a peer relation, when a charm is first deployed) your charm will need to observe the relation-created event. For example, a charm providing a database relation might need to create the database and credentials, so that the requirer charm can use the database.
+
+In the `src/charm.py` file, in the `__init__` function of your charm, set up `relation-created` event observers for the relevant relations and pair those with an event handler.
```
dwilding: I found the existing page a bit confusing because earlier we show `db` defined as the requirer, but here we've switched to talking about the providing charm. (The confusion isn't caused by your edits.) How about further changing this part help clarify the assumption:

```suggestion
In the `src/charm.py` file of the charm that's providing the relation, in the `__init__` function, set up `relation-created` event observers for the relevant relations and pair those with an event handler.
```

docs/howto/manage-libraries.md
```diff
@@ -208,3 +209,122 @@ def test_ready_event():
...
+
+### Write a library that manages a relation interface
+
+First, for a new interface, design the interface and the process related charms
+will use to populate the relation data. In the simplest case, the providing
+charm might populate the local app data when the relation is created, but a
+conversation between the charms is more common, where the requiring charm will
+populate its local data with a request, and the providing charm will use that to
+provide a suitable response. In more complex cases, this conversation might have
+multiple stages.
```
dwilding: A couple of suggestions for easier reading (to my eye):

```suggestion
First, for a new interface, design the interface and the process that related charms
will use to populate the relation data. In the simplest case, the providing
charm might populate the local app data when the relation is created, but a
conversation between the charms is more common. In a conversation, the requiring charm will
populate its local data with a request, and the providing charm will use that to
provide a suitable response. In more complex cases, this conversation might have
multiple stages.
```

docs/howto/manage-libraries.md
```diff
@@ -208,3 +209,122 @@ def test_ready_event():
...
+> See more: {ref}`manage-interfaces`
+
+If the library is implementing an existing interface, find the schema and
+details of the processes by which relation data is added on Charmhub in the
+integrations tabs, or by navigating to `https://charmhub.io/integrations/{integration-name}`.
+Alternatively, the interface details can be found in the
+[charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces)
+repository.
+
+In the documentation for the interface, find the description of the various
```
dwilding: Following the trail through Charmhub, I'm not totally sure what "the processes by which relation data is added" refers to. Maybe this would be fine (also including a couple more small suggestions):

```suggestion
If the library is implementing an existing interface, find the interface documentation by following links from the Integrations tab on Charmhub, or by navigating to `https://charmhub.io/integrations/{integration-name}`.
Alternatively, the interface documentation can be found in the
[charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces)
repository.

In the interface documentation, find the description of the various
```

docs/howto/manage-libraries.md
```diff
@@ -208,3 +209,122 @@ def test_ready_event():
...
+    )
+```
+
+```{tip}
+Ops defaults to transparently serialising each field to and from JSON, so you
+do not need to wrap each field in `pydantic.Json`.
+```
+
+In your `src/charm.py`, use the class you created to load and save data from the
+relation.
```
dwilding:
```suggestion
In your `src/charm.py` file, use the class you created to load and save data from the
relation:
```

docs/howto/manage-libraries.md
```diff
@@ -208,3 +209,122 @@ def test_ready_event():
...
+
+class TracingProviderAppData(pydantic.BaseModel):
+    receivers: list[Receiver] = pydantic.Field(
+        description="A list of enabled receivers in the form of the protocol they use and their resolvable server url.",
+    )
+```
+
+```{tip}
+Ops defaults to transparently serialising each field to and from JSON, so you
+do not need to wrap each field in `pydantic.Json`.
```
dwilding: Is the point here that all data in databags is stored as JSON? At first I was slightly confused by this tip, because JSON hadn't been mentioned up until this point. I'm also wondering what "defaults to" means - would I be able to disable it if I wanted? How about something like this (if correct):

```suggestion
Internally, Juju stores databag fields as JSON strings. Ops automatically serialises each field to and from JSON, so you
do not need to wrap each field in `pydantic.Json`.
```

ops/model.py
tonyandrewmeyer: Juju don't endorse "databag" as a term. Joe said that these are "unit relation settings" and "application relation settings". I don't like `rel.load(data, settings=self.unit)` at all.

james-garner-canonical: Ideas from conversation:
- `src` (for `load`) and `dst` (for `save`)
- `owner` for both (`load` the remote app's databag, `save` to my unit's databag, etc)
- maybe `scope` if `owner` isn't quite right

test/test_model_relation_data_class.py
tonyandrewmeyer: I'll move it, but I think it does belong in either place. With the other type annotations because it's a type annotation but after `__init__` because it's a method on the class.

docs/howto/manage-libraries.md
tonyandrewmeyer: > Is the point here that all data in databags is stored as JSON?

Yes. In particular, that at the moment the schemas (as found in libraries and charm-relation-interfaces) tend to look like:

```python
class TracingRequirerData(BaseModel):
    receivers: Json[List[str]] = Field(
        ..., description="List of protocols that the requirer wishes to use."
    )
```

But ops is taking care of the JSON serialisation and deserialisation so there's no need for the `Json` that's there.

> I'm also wondering what "defaults to" means - would I be able to disable it if I wanted?

Yes, `load` takes an `decoder` and `save` takes an `encoder` argument, and they default to `json.loads` and `json.dumps` but can be anything else you want.

> Internally, Juju stores databag fields as JSON strings.

So this isn't correct. Juju just has str:str and knows nothing at all about what that string might be. It's Ops that's defaulting to encoding/decoding each field to/from JSON.

What about something like this?

```suggestion
The Ops [](ops.Relation.load) and [](ops.Relation.save) methods serialise and deserialise the values of each field, and default to using JSON, so you do not need to wrap each field in `pydantic.Json`.
```

Or perhaps it's not needed at all? It's really mainly there for those people that are migrating from existing usage. Maybe it should be in a migration tip instead or something like that?

ops/model.py
james-garner-canonical: Updated explanation looks good to me =)

docs/howto/manage-libraries.md
dwilding: Thanks for explaining. So actually Ops _must_ serialise somehow for non string types, because Juju only sees strings. I'd be good to go with your suggestion, but would probably lean towards leaving this out unless we actively hear that people need it clarifying.

james-garner-canonical: Let's keep at least some mention of the default JSON (de)serialization and the implications for typing. @tonyandrewmeyer's latest suggestion looks good to me.

Note that this behaviour differs from the existing, untyped relation data, which is just returned as a `str:str` mapping, and which only permits writing strings (any (de)serialization is handled by the charmer separately).

ops/model.py
tonyandrewmeyer: I have `src` and `dst` at the moment. If anyone hates those or has something that they much prefer, speak now!

docs/howto/manage-libraries.md
tonyandrewmeyer: Ok, added it back. Let's look at this more closely when we rework these library/relation/interface guides.

PR #1703, opened 2025-04-28
feat: support for action parameter schema as Python classes

tonyandrewmeyer:
**This is in draft until the [spec](https://docs.google.com/document/d/1AiNvr1xfh078ieRczCCc9HSx4VW-dDJdF2p9FS2E_do/edit?usp=sharing) is accepted, so review should wait until it is moved from draft.**

A new class `ops.ActionBase` is added. This is to be used (potentially in mix-in style) for all classes that provide the action parameters as a schema. The class provides functionality to convert the Python classes to a Juju `actions.yaml` file (with the addition of a `yaml.safe_dump`). Classes can override the `to_juju_schema()` method if they need to provide YAML in a different way (or call `super().to_juju_schema()` and tweak the provided schema.

A new method is added to `ops.ActionEvent`: `load_params`, which is the custom class equivalent of the `.params` property. The charmer passes a `ActionBase` subclass to this method and gets back an instance of the class with the Juju action parameters loaded into it.

A new exception, `ops.InvalidSchemaError` is also added. This is raised from any `ValueError` (or subclass) when instantiating the class. We catch this error and exit with a zero code, in the same way as `_Abort(0)`, except that this exception is public, so that charmers can catch it and do custom error handling if they choose. When raising the error, we also set the action to failed using the `str()` of the error in the message.

* Pydantic dataclasses

A new `ops._private.attrdocs` module is added (this is also present -- with exactly the same implementation -- in the related config PR). This provides support for getting 'docstrings' of attributes.


tonyandrewmeyer:
Moving back to draft, since this will need similar changes as the config one with the decision to move out the schema generation.

tonyandrewmeyer:
Merging this with #1702.

PR #1705, opened 2025-04-28
fix: allow TLS 1.2 in ops-tracing

dimaqq:
Relax our, client-side TLS version requirement to allow TLS 1.2 or 1.3.

This roughly corresponds to allowing "Intermediate" config in addition to "Modern" in the Mozilla SSL Configuration Generator

Ref: https://ssl-config.mozilla.org/

Although https://github.com/canonical/tempo-coordinator-k8s-operator/pull/147 has been merged, I don't have clarity on all the trace data sinks and their deployment options.

Additionally some versions of the telemetry stack are in production and we'd have to analyse those very carefully if we enforce TLS 1.3 from the get-go, lest we force COS deployment upgrades merely because the first "random" application charm is switched to `ops[tracing]`.

### DIFF
--- a/docs/reference/ops-tracing.rst
+++ b/docs/reference/ops-tracing.rst
@@ -14,6 +14,16 @@ Open Telemetry resource attributes
 - ``service.instance.id`` the unit number, like ``0``.
 - ``service.charm`` the charm class name, like ``DbCharm``.
 
+
+Security considerations
+-----------------------
+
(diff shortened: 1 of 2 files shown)


james-garner-canonical (approved):
This change sounds reasonable to me, and it would be good to get it out before the release.

tonyandrewmeyer (approved):
Works for me. I feel 1.2 is still ok, even if not ideal, although I'm not keeping up with the TLS world like I used to.

PR #1710, opened 2025-04-30
docs: link to our docs from the top of our readme

james-garner-canonical:
A charming survey response suggested linking to our docs from the top of our readme. I think this is a good idea. Currently I believe the only link is at the end of the (now quite long) readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ The `ops` library is a Python framework for developing and testing Kubernetes an
 
 > - `ops` is  [available on PyPI](https://pypi.org/project/ops/).
 > - The latest version of `ops` requires Python 3.8 or above.
+> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, reference, and more.
 
 ## Give it a try
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The `ops` library is a Python framework for developing and testing Kubernetes an
 
 > - `ops` is  [available on PyPI](https://pypi.org/project/ops/).
 > - The latest version of `ops` requires Python 3.8 or above.
-> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, reference, and more.
+> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, the library reference, and more.
 
 ## Give it a try
 


tonyandrewmeyer (commented):
Two concerns:

* The README is structured the way it is because there was an effort to unify READMEs across the charming ecosystem (for example, compare with the Juju or Charmcraft ones). If the docs link should be at the top, I feel it's probably best for the TAs to agree on this and we all make the change.
* Do we need the link both at the top and the bottom? I'd rather move it than add a link, if we do want it at the top.

james-garner-canonical:
> The README is structured the way it is because there was an effort to unify READMEs across the charming ecosystem (for example, compare with the Juju or Charmcraft ones). If the docs link should be at the top, I feel it's probably best for the TAs to agree on this and we all make the change.

The overall structure is nice, and I think it's good that our READMEs follow a similar structure. I don't think adding a docs link after the PyPI link necessarily needs cross team buy-in, but if the consensus is otherwise then I'm happy to close this PR.

> Do we need the link both at the top and the bottom? I'd rather move it than add a link, if we do want it at the top.

I don't think there's any harm in the redundancy, and I think the README is long enough that both are useful.

benhoyt (commented):
I'm definitely in favour of this change -- I was looking around for this on the README just this morning, and it felt weird that I couldn't find it (I didn't even think to scroll to the bottom). I think having it in both places, and *slightly* different from other teams/projects is okay. I know David has some larger "overhauling" in mind for later.

README.md
```diff
@@ -6,6 +6,7 @@ The `ops` library is a Python framework for developing and testing Kubernetes an
 
 > - `ops` is  [available on PyPI](https://pypi.org/project/ops/).
 > - The latest version of `ops` requires Python 3.8 or above.
+> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, reference, and more.
```
dwilding, line 9: How about a small change?

```suggestion
> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, the library reference, and more.
```

dwilding (approved):
I'm good with this change. Although I think even more helpful would be if we add the doc link into the About section of the repo, as with Pebble:

Screenshot from 2025-05-06 07-18-18

PR #1711, opened 2025-05-01
revert: run deferred events with fresh charm instances

tonyandrewmeyer:
Revert #1631 (f97e707).

An issue that was missed was that because the framework doesn't provide a recommended way to run code prior to the event handler (an inverse of the `commit` event) charms are doing this work in `__init__`. With the change from #1631, that code would be run more than once per hook if there are any queued notices, which may be problematic. We also don't provide a mechanism for knowing if a handler is running from a queued notice or as a new event (other than looking in the environment), so charms couldn't trigger only on one of the events (although this would be simpler to solve).

For now, we're reverting this change. We'll discuss further whether we abandon this change, find an alternative implementation, or provide alternatives to running this type of code in `__init__` and do the change later.

### DIFF
--- a/ops/_main.py
+++ b/ops/_main.py
@@ -18,11 +18,11 @@ from __future__ import annotations
 
 import logging
 import os
-import pathlib
 import shutil
 import subprocess
 import sys
 import warnings
+from pathlib import Path
 from typing import Any, Union, cast
 
 from . import charm as _charm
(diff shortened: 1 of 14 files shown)


dimaqq:
First impressions: man I wish the cleanups and import renames were a separate PR...

james-garner-canonical (approved):
Revert looks good (I ran through it side by side with the original PR). A shame to lose the import cleanups/etc, but it is what it is.

dimaqq (approved):
Look good!

Issue #1718, opened 2025-05-07
Refresh K8s charm tutorial, with source files in Ops repo


dwilding:
This issue is for tracking several related improvements to the [Kubernetes charm tutorial](https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/index.html):

1. Do a clean play through of the tutorial and fix/clarify any inconsistencies.
2. Put the charm files in the Ops repo, in a new directory called `examples`.
    - Have a directory corresponding to the charm files as of the end of each chapter.
    - Add a CI workflow that runs the unit tests in each directory.
3. Decide what to do with [FastAPI-Monitoring.json.tmpl](https://github.com/canonical/juju-sdk-tutorial-k8s/raw/refs/heads/05_cos_integration/src/grafana_dashboards/FastAPI-Monitoring.json.tmpl), which is referenced in [the last chapter of the tutorial](https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.html#import-the-grafana-interface-libraries-and-set-up-the-grafana-dashboards).
4. Archive the [juju-sdk-tutorial-k8s](https://github.com/canonical/juju-sdk-tutorial-k8s) repo.
    - At the end of each chapter, link to the relevant directory within `examples`.

PR #1719, opened 2025-05-07
docs: update code in K8s tutorial, with source in repo (part 1)

dwilding:
This PR starts work on #1718:

- Updates the charm code and tests in [Create a minimal Kubernetes charm](https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.html), after I did some light cleanup:
    - Made sure that the code passes our linting rules
    - Replaced some old links in docstrings/comments
- Adds a directory `examples/k8s-1-minimal`, which contains the charm that you'd have after completing the chapter. I cross-checked this against [juju-sdk-tutorial-k8s](https://github.com/canonical/juju-sdk-tutorial-k8s/tree/01_create_minimal_charm).
- Adds a workflow to run the charm's unit tests.

**[Preview doc build](https://ops--1719.org.readthedocs.build/en/1719/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.html)** | **[Preview of example directory](https://github.com/dwilding/canonical-operator/tree/add-tutorial-charm-1/examples)**

---

Decisions (for the record):

- Unlike [tox.ini in juju-sdk-tutorial-k8s](https://github.com/canonical/juju-sdk-tutorial-k8s/blob/01_create_minimal_charm/tox.ini), the charm in the tutorial shouldn't include the targets `lint`, `format`, and `static`. I won't add those targets to `tox.ini` for the charms within `examples`. In the future, we'd like to update the tutorial to start with `charmcraft init`, at which point `tox.ini` will automatically have the extra targets.

Future work:

- I'll open a separate PR for the code & tests in the other chapters. When all the code is in the repo, we can update each chapter to point to the corresponding place in `examples` instead of juju-sdk-tutorial-k8s.
- When I've worked through the entire tutorial, I'll open a separate PR for any general fixes/clarifications.


.github/workflows/example-charm-tests.yaml
```diff
@@ -0,0 +1,35 @@
+name: Example Charm Tests
```
james-garner-canonical, line 1: How about we drop this workflow and drop the `--ignore` added to our `tox.ini` in this PR, and just run the example charms unit tests as part of our unit tests?

dwilding: I figured it would be more realistic to run the unit tests that the charm defines - but I'll defer on what the right approach would be!

When I try to run the charm's unit tests as part of our unit tests, I get an error: https://gist.github.com/dwilding/376d776b22411336d1e50f16ec052a2f

benhoyt: Decision: these charm tests aren't ops tests, so can't be run the same way. Current approach is fine.

benhoyt (approved):

.github/workflows/example-charm-tests.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - examples/k8s-1-minimal
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Set up Python 3
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
```
benhoyt, line 29: 3.11 is an odd choice. Should probably change to a matrix of 3.10 and 3.12. @dwilding

dwilding: Done in #1732

PR #1720, opened 2025-05-07
chore: update style guide

IronCore864:
Update the style guide, add most content in the Pebble repo and link it there.

Based on a discussion with David, we agreed to have fewer places to store the style guide for docs. Meaning we should not store it in both the ops repo and the Pebble repo, but rather, store it in one place and link it in the other.

After discussion, we think maybe the Pebble repo should be the right place, reasons:

- Examples are all based on Pebble.
- Keep the existing "British English" item in the ops repo, then use a link to link to the Pebble repo's `STYLE.md`. This makes British English an exception in the ops repo to the main style guide.

If we prefer to keep the docs guide in the ops repo, I can rework the PRs.

The other PR in the pebble repo:

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -137,9 +137,4 @@ if status is not pebble.ServiceStatus.ACTIVE:
 
 It's a bit less clear when we're dealing with code and APIs, as those normally use US English, for example, `pytest.mark.parametrize`, and `color: #fff`.
 
-
-### Spell out abbreviations
-
-Abbreviations and acronyms in docstrings should usually be spelled out, for example, "for example" rather than "e.g.", "that is" rather than "i.e.", "and so on" rather than "etc", and "unit testing" rather than UT.
-
-However, it's okay to use acronyms that are very well known in our domain, like HTTP or JSON or RPC.
+See more doc style guide [here](https://github.com/canonical/pebble/blob/master/STYLE.md#docs-and-docstrings).


STYLE.md
```diff
@@ -137,9 +137,4 @@ if status is not pebble.ServiceStatus.ACTIVE:
 
 It's a bit less clear when we're dealing with code and APIs, as those normally use US English, for example, `pytest.mark.parametrize`, and `color: #fff`.
 
-
-### Spell out abbreviations
-
-Abbreviations and acronyms in docstrings should usually be spelled out, for example, "for example" rather than "e.g.", "that is" rather than "i.e.", "and so on" rather than "etc", and "unit testing" rather than UT.
-
-However, it's okay to use acronyms that are very well known in our domain, like HTTP or JSON or RPC.
+See more doc style guide [here](https://github.com/canonical/pebble/blob/master/STYLE.md#docs-and-docstrings).
```
dwilding, line 140: It might be more obvious to put this at the top of the "Docs and docstrings" section - something like:

```
## Docs and docstrings

See the [Pebble doc style guide](https://github.com/canonical/pebble/blob/master/STYLE.md#docs-and-docstrings).

### Use British English
...
```

dwilding (approved):

benhoyt:
@IronCore864 @dwilding I'm not sure about this change. Ops is kind of Charm Tech's "main" repo, so it seems to me if anywhere, the docs style guide should belong there. If there's disagreement, seeing it's almost the engineering sprint, perhaps we can discuss and finalise next week? I've added to our discussion list.

IronCore864:
I agree that the ops repo is the "main" one. Although all the examples in the doc guide are from Pebble, I don't think it's a big problem, because they are all examples, so it's OK for them to live in the ops repo. I don't have a strong feeling about it.

dwilding:
Same - I'm good if the doc guidance goes in Ops instead of Pebble.

benhoyt:
Sounds good -- let's do that then. We may want to tweak some of the "obviously Pebble" examples to be more Ops-y, but we can cover that in review.

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
 
 ## Docs and docstrings
 
-### Use British English
+### Tones
```
benhoyt, line 134:
```suggestion
### Tone
```

benhoyt: Actually, similar to the Pebble PR, I think we should drop this section header and move the `####` to `###` to avoid too much nesting.

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+#### Avoid negative, be positive
```
benhoyt, line 136:
```suggestion
#### Avoid the negative, state conditions in the positive
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+State conditions positively as what should happen, rather than what shouldn't. Focus on desired behaviour, and frame instructions and conditions around the expected or successful outcome, not failure cases.
```
benhoyt, line 138:
```suggestion
State conditions positively: what should happen, rather than what shouldn't. Focus on desired behaviour, and frame instructions and conditions around the expected or successful outcome, not failure cases.
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Example:
+
+- Avoid: "If the command doesn’t exit within 1 second, the start is considered successful." (Netagive)
```
benhoyt, line 142:
```suggestion
- Avoid: "If the command doesn’t exit within 1 second, the start is considered successful." (Negative)
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+- Prefer: "If the command stays running for the 1-second window, the start is considered successful." (Positive)
+
+Exceptions: Only use negative phrasing (e.g., "if X does not happen") when:
```
benhoyt, line 145: Elsewhere in this doc we say to avoid Latin abbreviations like "e.g." :-). I think we can just drop this parenthesis:
```suggestion
Exceptions: Only use negative phrasing when:
```

Please replace "e.g." with "for example" throughout.

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+#### Avoid passive, be active
+
+Use the active voice as much as possible.
+
+We should only use the passive voice when we don't know or care about who performed the action.
+
+Example:
+
+- Avoid: "A minimum check is created using the default values"
```
benhoyt, line 158: Should these two lines say "minimal" instead of "minimum"?

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+- Prefer: "We create a minimum check, using the default values"
+
+#### Avoid subjective, be objective
+
+Instead of using words like "easy," "simple," or "just," which are very subjective and assume the reader's skill level, describe the action directly by stating what to do and focusing on concrete steps.
```
benhoyt, line 163:
```suggestion
Avoid using words like "easy", "simple", or "just", which are subjective and assume the reader's skill level. Instead, describe the action directly by stating what to do and focusing on concrete steps.
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Examples:
+
+- Avoid: "This can be done easily using ..."
+- Prefer: "This can be done using ..."
```
benhoyt, line 168:
```suggestion
- Avoid: "This can be done easily using ..."
- Prefer: "You can do this using ..."
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+- Avoid: "This can be easily configured by ..."
+- Prefer: "We can configure this using ..."
+
+- Avoid: "Simply run the command ..."
+- Prefer: "Run the command ..."
+
+### Structure
```
benhoyt, line 176: Let's again avoid the additional nesting level

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+#### Be short
+
+For example, a very long introduction part of a how-to guide or a tutorial probably isn't necessary; it's good if we are writing a blog post or a book, but not so much for a tutorial/how-to guide type of doc.
```
benhoyt, line 180:
```suggestion
For example, avoid long introductions in a how-to guide or a tutorial. They might be appropriate if we were writing a blog post or a book, but not for documentation.
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+- Avoid: "This can be easily configured by ..."
+- Prefer: "We can configure this using ..."
+
+- Avoid: "Simply run the command ..."
+- Prefer: "Run the command ..."
+
+### Structure
+
+#### Be short
```
benhoyt, line 178: "Be short" sounds odd. And let's merge this section into the next one, as conciseness is the same / overlaps.

```suggestion
### Be concise and avoid repetition
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+For example, a very long introduction part of a how-to guide or a tutorial probably isn't necessary; it's good if we are writing a blog post or a book, but not so much for a tutorial/how-to guide type of doc.
+
+#### Be concise and reduce repetition
+
+Trim repetitive titles: Avoid repeating the name if the surrounding content already provides context.
```
benhoyt, line 184: What is "the name"? Maybe join the following "For example" paragraph with this one.

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+For example, in the doc "how to use pebble", the section names can simply be "API", "CLI commands", "Features", etc., instead of "Pebble API", "Pebble CLI commands", "Pebble features".
```
benhoyt, line 186:
```suggestion
For example, in a document called "How to use Pebble", the section names can simply be "API", "CLI commands", and "Features", instead of "Pebble API", "Pebble CLI commands", and "Pebble features".
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+#### Order: alphanumeric or logical
+
+We can either use an alphanumeric order, or a logical order, depending on the context.
+
+Alphanumeric order is better in the case of a reference doc or a list (e.g., an Enum).
+
+Example (Enum):
+
+- Prefer: `["autostart", "replan", "restart", "start", "stop"]`
+- Avoid: `["start", "stop", "restart", "replan", "autostart"]`
```
benhoyt, line 197:
```suggestion
- Prefer: autostart, replan, restart, start, stop
- Avoid: start, stop, restart, replan, autostart
```

Also, speaking of order, can we use the same order for Avoid and Prefer (Avoid first, then Prefer) throughout, like the Pebble doc?

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Example (A list of commands):
+
+Avoid:
```
benhoyt, line 201:
```suggestion
Avoid (arbitrary order):
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+- `pebble ls`
+- `pebble mkdir`
+- `pebble rm`
+- `pebble push`
+- `pebble pull`
+
+Prefer:
```
benhoyt, line 209:
```suggestion
Prefer (alphabetical order):
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+#### Spell out abbreviations
 
-### Spell out abbreviations
+Abbreviations and acronyms in docstrings should usually be spelled out:
 
-Abbreviations and acronyms in docstrings should usually be spelled out, for example, "for example" rather than "e.g.", "that is" rather than "i.e.", "and so on" rather than "etc", and "unit testing" rather than UT.
+- "for example" rather than "e.g."
+- "that is" rather than "i.e."
+- "and so on" rather than "etc"
+- "unit testing" rather than UT, and so on
```
benhoyt, line 237: Thanks for this change -- it does make it more readable!

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
 
 However, it's okay to use acronyms that are very well known in our domain, like HTTP or JSON or RPC.
+
+#### Use sentence case in headings
+
+`## Use sentence case for headings`, instead of `## Use Title Case for Headings`.
+
+#### Be consistent
+
+Be consistent in every possible way as much as possible, and stay context-consistent.
```
benhoyt, line 247: This is a funny redundant sentence "in every possible way" and "as much as possible". What about just dropping it?

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Choice of words: For example, if the whole document uses "mandatory", you probably shouldn't use "required" in a newly added paragraph. For another example, if the whole doc uses "list foo" when adding new content, don't use "get a list of bar".
+
+Headings/lists: For example, if a doc is called "How to use Pebble", with sections "API", "CLI commands", "Health checks", and "Specification", use the heading "Environment variables" for a new section about environment variables, instead of the heading "Pebble environment variables".
```
benhoyt, line 251: This isn't needed, as we've already covered it above.

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+#### Be precise
+
+Be precise in names and verbs. Use precise verbs to describe the behaviour. For example, the appropriate description of `/v1/services` is "list services", while "get a service" is probably a better fit for `/v1/services/{name}`.
+
+Split distinct ideas: Use separate sentences/clauses and avoid cramming. Don’t merge unrelated details (e.g., parameter format + default behaviour) into a single phrase.
```
benhoyt, line 257:
```suggestion
Split distinct ideas: Use separate sentences or clauses, and avoid cramming. Don’t merge unrelated details into a single phrase (for example, parameter format and default behaviour).
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Example:
+
+- Avoid: "The names of the services to get, a comma-separated string. If empty, get all services."
+- Prefer: "The names of the services to get. Specify multiple times for multiple values. If omitted, the method returns all services." (In three short precise sentences we've covered what the parameter does, usage details, and behavioral details.)
+
+Example:
+
+- Avoid: "For reference information about the API, see [API and clients] and [API]." (What's the difference of these two?)
```
benhoyt, line 266:
```suggestion
- Avoid: "For reference information about the API, see [API and clients] and [API]." (It's hard to tell the difference of these two.)
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+Example:
+
+- Avoid: "Pebble provides a comprehensive health check feature"
+- Prefer: "Pebble provides a health check feature"
+
+Let features speak for themselves, describe actual capabilities and use measurable terms when possible (e.g., "supports 3 types of checks" instead of "versatile checking")
+
+Future-proofing:
+
+- Avoid absolute terms unless permanently true
```
benhoyt, line 282: What do you mean by this? Can you give a brief example?

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+- Qualify with "currently" when describing evolving features
```
benhoyt, line 283: I actually think we should drop this. It's usually just okay to omit "currently" and have the current docs reflect the current state. Otherwise if we were consistent there'd be "currently" everywhere.

Alternatively we could give a specific example of when this would apply.

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+#### Avoid implementation-specific terminology
+
+Use generic descriptions instead of what's specific to the code. Focus on behaviour, not internal implementation: Describe what it does (e.g., "returns an error message") rather than how a particular implementation works.
```
benhoyt, line 287:
```suggestion
Focus on behaviour, not internal implementation: Describe what it does (e.g., "returns an error message") rather than how a particular implementation works.
```

The "returns an error message" example here doesn't fit what you're talking about very well. Can we find a clearer one?

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+Example:
+
+- Avoid: "`Change.Err` will be non-empty if the change had an error." (Code-specific)
+- Prefer: "The `err` field in the response will contain an error message if the operation failed." (General)
+
+Exception: If documenting a client library (e.g., Go/Python SDKs), implementation details are appropriate.
+
+#### Articles
+
+"a" or "the", generic or specific, choose carefully. When describing generic parameters or behavior:
```
benhoyt, line 298:
```suggestion
Choose between the "a" or "the" articles carefully.
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+- Avoid: "Restart the service when the health check fails."
+- Prefer: "Restart a service when the health check fails." (When describing a generic behaviour, no specific service is implied.)
+
+
+### Code blocks
+
+- Consistency: the styles of code blocks and terminal output samples should be the same, at least within the same document.
+- Preferred style: Use `{code-block}` when showing files. Use `{terminal}` (`.rst` style) when showing commands or terminal output. Avoid using `{code-block}` for commands or terminal output, unless it's consistent with the existing content in the same document.
+- Highlighting: Use `:emphasize-lines: 8-10` with `{code-block}` for highlighting lines in files when necessary. Don't add an inconsistent `{code-block}` so that you can highlight lines in commands or terminal output. Instead, use `{terminal}` along with helpful words and comments.
```
benhoyt, line 311:
```suggestion
- Highlighting: Use `:emphasize-lines: 8-10` with `{code-block}` for highlighting lines in files when necessary. Don't add an inconsistent `{code-block}` just so that you can highlight lines in commands or terminal output. Instead, use `{terminal}` along with helpful words and comments.
```

STYLE.md
```diff
@@ -131,15 +131,188 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+### YAML
+
+- Types: Handle strings and integers carefully. For example, if the field "change" is a string in the code, use `change: "1"` in the YAML example, not `change: 1`.
```
benhoyt, line 315: This is more of a code review point, not a style guide thing. I think we should drop it.

benhoyt (changes requested):
Thanks for this. Looks good overall. Again, lots of comments, mostly minor style and grammar things.

STYLE.md
IronCore864: I removed this item and added an example for the "currently" one.

dimaqq:
Adding Tony, while Ben is away.
The docs LGTM :)

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
 
 However, it's okay to use acronyms that are very well known in our domain, like HTTP or JSON or RPC.
+
+### Use sentence case in headings
+
+`## Use sentence case for headings`, instead of `## Use Title Case for Headings`.
+
+### Be consistent
+
+Choice of words: For example, if the whole document uses "mandatory", you probably shouldn't use "required" in a newly added paragraph. For another example, if the whole doc uses "list foo" when adding new content, don't use "get a list of bar".
```
dwilding: Very minor, for consistency :slightly_smiling_face: 

```suggestion
Choice of words: For example, if the whole document uses "mandatory", you probably shouldn't use "required" in a newly added paragraph. For another example, if the whole document uses "list foo" when adding new content, don't use "get a list of bar".
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+### Avoid the negative, state conditions in the positive
+
+State conditions positively: what should happen, rather than what shouldn't. Focus on desired behaviour, and frame instructions and conditions around the expected or successful outcome, not failure cases.
+
+Example:
+
+- Avoid: "If the command doesn’t exit within 1 second, the start is considered successful." (Negative)
+- Prefer: "If the command stays running for the 1-second window, the start is considered successful." (Positive)
+
+Exceptions: Only use negative phrasing when:
```
dwilding: Very minor - I think it's better like:

```suggestion
Only use negative phrasing when:
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+Examples:
+
+- Avoid: "This can be done easily using ..."
+- Prefer: "You can do this using ..."
+
+- Avoid: "This can be easily configured by ..."
+- Prefer: "We can configure this using ..."
+
+- Avoid: "Simply run the command ..."
+- Prefer: "Run the command ..."
```
dwilding, line 174: These examples work better in the source than when displayed in GitHub. In GitHub, it's not possible to distinguish the different pairs. How about this:

```suggestion
- "You can do this using ..." rather than "This can be done easily using ..."
- "This can be easily configured by ..." rather than "We can configure this using ..."
- "Simply run the command ..." rather than "Run the command ..."
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+### Articles
+
+Choose between the "a" or "the" articles carefully.
+
+- Avoid: "The format of the duration string is a sequence of decimal numbers"
+- Prefer: "The format of a duration string is a sequence of decimal numbers." (Describing a generic parameter, not a specific one.)
+
+- Avoid: "Restart the service when the health check fails."
+- Prefer: "Restart a service when the health check fails." (When describing a generic behaviour, no specific service is implied.)
```
dwilding: The different groups don't render well in GitHub. How about this:

```suggestion
When describing a generic parameter:

- Avoid: "The format of the duration string is a sequence of decimal numbers."
- Prefer: "The format of a duration string is a sequence of decimal numbers."

When describing a generic behaviour:

- Avoid: "Restart the service when the health check fails."
- Prefer: "Restart a service when the health check fails." (No specific service is implied.)
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Trim repetitive titles: Avoid repeating the name if the surrounding content already provides context.
+
+For example, in a document called "How to use Pebble", the section names can simply be "API", "CLI commands", and "Features", instead of "Pebble API", "Pebble CLI commands", and "Pebble features".
+
+### Order: alphanumeric or logical
+
+We can either use an alphanumeric order, or a logical order, depending on the context.
+
+Alphanumeric order is better in the case of a reference doc or a list (for example, an Enum).
```
dwilding, line 188: I'd probably go with "alphabetical" as it sounds more natural, but either works

```suggestion
### Order: alphabetical or logical

We can either use an alphabetical order, or a logical order, depending on the context.

Alphabetical order is better in the case of a reference doc or a list (for example, an Enum).
```

dwilding (approved):
Nice one, just thought of a few minor things as I was reading through.

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
 
 ## Docs and docstrings
 
+### Avoid the negative, state conditions in the positive
+
+State conditions positively: what should happen, rather than what shouldn't. Focus on desired behaviour, and frame instructions and conditions around the expected or successful outcome, not failure cases.
+
+Example:
+
+- Avoid: "If the command doesn’t exit within 1 second, the start is considered successful." (Negative)
```
tonyandrewmeyer: We usually use plain `'` and let tooling move to smart quotes. I guess this is mostly to be consumed in the original format, but I feel we'd be better sticking to that:

```suggestion
- Avoid: "If the command doesn't exit within 1 second, the start is considered successful." (Negative)
```

We also use the plain `'` elsewhere in the doc, so it's more consistent.

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+### Avoid passive, be active
+
+Use the active voice as much as possible.
+
+We should only use the passive voice when we don't know or care about who performed the action.
+
+Example:
+
+- Avoid: "A minimal check is created using the default values"
+- Prefer: "We create a minimal check, using the default values"
```
tonyandrewmeyer: Personally, for something like this I would go with:

```suggestion
- Prefer: "Create a minimal check, using the default values"
```

So active voice but also imperative mood.

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+- Avoid: "This can be easily configured by ..."
+- Prefer: "We can configure this using ..."
+
+- Avoid: "Simply run the command ..."
+- Prefer: "Run the command ..."
+
+### Be concise and avoid repetition
+
+For example, avoid long introductions in a how-to guide or a tutorial. They might be appropriate if we were writing a blog post or a book, but not for documentation.
```
tonyandrewmeyer: I think we want to avoid these in reference docs too -- but I think that sometimes longer form is ok with explanation. I also feel that blog posts and books can be documentation.

What about something like:

```suggestion
For example, avoid long introductions. In tutorials, how-to guides, and reference documentation, get the reader to the core of the content as quickly as possible.
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+### Be consistent
+
+Choice of words: For example, if the whole document uses "mandatory", you probably shouldn't use "required" in a newly added paragraph. For another example, if the whole doc uses "list foo" when adding new content, don't use "get a list of bar".
+
+### Be precise
+
+Be precise in names and verbs. Use precise verbs to describe the behaviour. For example, the appropriate description of `/v1/services` is "list services", while "get a service" is probably a better fit for `/v1/services/{name}`.
+
+Split distinct ideas: Use separate sentences or clauses, and avoid cramming. Don’t merge unrelated details into a single phrase (for example, parameter format and default behaviour).
```
tonyandrewmeyer:
```suggestion
Split distinct ideas: Use separate sentences or clauses, and avoid cramming. Don't merge unrelated details into a single phrase (for example, parameter format and default behaviour).
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Example:
+
+- Avoid: "The names of the services to get, a comma-separated string. If empty, get all services."
+- Prefer: "The names of the services to get. Specify multiple times for multiple values. If omitted, the method returns all services." (In three short precise sentences we've covered what the parameter does, usage details, and behavioral details.)
+
+Example:
+
+- Avoid: "For reference information about the API, see [API and clients] and [API]." (It's hard to tell the difference of these two.)
+- Prefer: "For an explanation of API access levels, see [API and clients]. For the full API reference, see [API reference]." (Clear)
```
tonyandrewmeyer: Let's be consistent here and make "Clear" a sentence.
```suggestion
- Prefer: "For an explanation of API access levels, see [API and clients]. For the full API reference, see [API reference]." (Clear.)
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+Split distinct ideas: Use separate sentences or clauses, and avoid cramming. Don’t merge unrelated details into a single phrase (for example, parameter format and default behaviour).
+
+Example:
+
+- Avoid: "The names of the services to get, a comma-separated string. If empty, get all services."
+- Prefer: "The names of the services to get. Specify multiple times for multiple values. If omitted, the method returns all services." (In three short precise sentences we've covered what the parameter does, usage details, and behavioral details.)
+
+Example:
+
+- Avoid: "For reference information about the API, see [API and clients] and [API]." (It's hard to tell the difference of these two.)
```
tonyandrewmeyer:
```suggestion
- Avoid: "For reference information about the API, see [API and clients] and [API]." (It's hard to tell the difference between these two.)
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+Future-proofing: Qualify with "currently" when describing evolving features. For example: "Currently, the only supported log target type is Grafana Loki".
+
+### Avoid implementation-specific terminology
+
+Focus on behaviour, not internal implementation: Describe what it does rather than how a particular implementation works.
+
+Example:
+
+- Avoid: "`Change.Err` will be non-empty if the change had an error." (Code-specific)
+- Prefer: "The `err` field in the response will contain an error message if the operation failed." (General)
```
tonyandrewmeyer: I feel that the problem here is that it's essentially a tautology, rather than being code specific. I think a better example in this space would be:

```suggestion
- Avoid: "The `Change` ID will be empty when the check is stopped." (Implementation-specific.)
- Prefer: "The `Status` will be 'inactive' when the check is stopped." (General.)
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Exception: If documenting a client library (for example, Go/Python SDKs), implementation details are appropriate.
```
tonyandrewmeyer: It does feel a bit awkward to have this here, when all three packages in this repository are essentially client libraries (or frameworks, which is essentially the same).

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+### Code blocks
+
+- Consistency: the styles of code blocks and terminal output samples should be the same, at least within the same document.
+- Preferred style: Use `{code-block}` when showing files. Use `{terminal}` (`.rst` style) when showing commands or terminal output. Avoid using `{code-block}` for commands or terminal output, unless it's consistent with the existing content in the same document.
+- Highlighting: Use `:emphasize-lines: 8-10` with `{code-block}` for highlighting lines in files when necessary. Don't add an inconsistent `{code-block}` just so that you can highlight lines in commands or terminal output. Instead, use `{terminal}` along with helpful words and comments.
+
+### YAML
+
+- Use quotes for strings: This is especially important if a string contains special characters or starts with a number.
+- Indentation: Always use spaces and be consistent with the number of spaces throughout the same file (usually, this should be two or four).
```
tonyandrewmeyer: I'd be in favour of us picking one (probably 2?) and going with that rather than just saying to be consistent.

```suggestion
- Indentation: Always use spaces and be consistent with the number of spaces throughout the same file (two, unless the file already uses a different number).
```

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+- Prefer: "The format of a duration string is a sequence of decimal numbers." (Describing a generic parameter, not a specific one.)
+
+- Avoid: "Restart the service when the health check fails."
+- Prefer: "Restart a service when the health check fails." (When describing a generic behaviour, no specific service is implied.)
+
+
+### Code blocks
+
+- Consistency: the styles of code blocks and terminal output samples should be the same, at least within the same document.
+- Preferred style: Use `{code-block}` when showing files. Use `{terminal}` (`.rst` style) when showing commands or terminal output. Avoid using `{code-block}` for commands or terminal output, unless it's consistent with the existing content in the same document.
```
tonyandrewmeyer: I think the intention here is to choose between `terminal` and `code-block`, but it seems like it's saying that the `{code-block}` style should always be used. We definitely don't do that now -- we have lots of triple-backtick followed by language, and I think those are generally fine, particularly for simple cases.

I think this would be better extended to cover the common cases, particularly how to format a code block in API reference code (ReST in a docstring) and in general docs (Myst-markdown, in .md files) with 2-3 small examples. Focusing particularly on Python/Go code more than shell commands (or having that as well, but definitely with Python and/or Go, maybe one YAML example).

I think we should also link to https://mystmd.org/guide/code so that it's easy to find the reference for what can be done with code blocks (outside of the docstrings).

I think we eventually want to adopt more use of pulling in code from external files (Charmcraft does a bunch of this quite effectively, I think), so we could maybe also explain that, but perhaps that's better left for another day.

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+- Highlighting: Use `:emphasize-lines: 8-10` with `{code-block}` for highlighting lines in files when necessary. Don't add an inconsistent `{code-block}` just so that you can highlight lines in commands or terminal output. Instead, use `{terminal}` along with helpful words and comments.
```
tonyandrewmeyer: I'm not particularly sold on code highlighting in short code blocks in docs. Do we really want to recommend this?

tonyandrewmeyer (approved):
I agree with most of the suggestions here, but they don't feel very specific to Ops (or Pebble, or Charm Tech), more just general good documentation practice (or perhaps Canonical documentation practice).

I'm not sure what motivated adding these. Maybe this is way too late to ask this question, but is there a reason to add these to a Charm Tech repo rather than link out to this content in a Canonical location, and only have things that are quite specific to our team/repos here?

(Much like we don't describe what Pythonic code is, we have an automated tool that validates it and link out to other documentation, and only document where we deviate or have more specific guidelines).

I've made a few comments/suggestions, but approving with the assumption that either those will be resolved or we can tweak this going forward.

STYLE.md
```diff
@@ -131,15 +131,172 @@ if status is not pebble.ServiceStatus.ACTIVE:
...
+
+Use the active voice as much as possible.
+
+We should only use the passive voice when we don't know or care about who performed the action.
+
+Example:
+
+- Avoid: "A minimal check is created using the default values"
+- Prefer: "We create a minimal check, using the default values"
+
```
dwilding: Building on Tony's suggestion, how about:

```suggestion
Use the active voice as much as possible, especially in direct instructions to the reader.

We should only use the passive voice when we don't know or care about who performed the action.

Example:

- Avoid: "A minimal check is created using the default values"
- Prefer: "Create a minimal check, using the default values"

Writing "We create ..." rather than "Create ..." in explanatory documentation is also fine.

```

STYLE.md
dwilding: I like this suggestion, but I still feel like it's worth making a distinction between technical documentation and other types of content. How about this:

```suggestion
For example, avoid long introductions. In tutorials, how-to guides, and reference documentation, get the reader to the core of the content as quickly as possible. Even in explanatory documentation, stay focused on the most relevant points.

If you want to write a summary of a broader topic, for general interest, Discourse is a better place.
```

STYLE.md
dwilding: I would go with `)` over `.)` in the other instances, but that's personal taste :wink: For consistency, I agree with Tony

STYLE.md
dwilding: Ah, I actually interpreted this differently. I read it as "don't use `Change` because that's a class name and the user would more likely access it as `my_change.err`". Maybe I was reading too much into it. Tony's suggestion makes sense - given my interpretation, how about:

```suggestion
- Avoid: "The `Change` ID will be empty when the check is stopped." (Implementation-specific.)
- Prefer: "The `status` of the check will be 'inactive' when the check is stopped." (General.)
```

STYLE.md
dwilding: Maybe the best solution is to drop this sentence? Even libraries/frameworks will have their own implementation details (`_foo` stuff) which we try to avoid documenting, so the core advice of avoiding implementation-specific terminology still stands.

STYLE.md
dwilding: I think I need to look into this point again. In my head `{code-block}` and three backticks are synonymous. But it sounds like not. I'll educate myself and leave another comment!

STYLE.md
dwilding: "when necessary" --> "if highlighting is the clearest way to make your point" ?

STYLE.md
IronCore864: In another comment from Ben, he prefers to keep it as the same style (Avoid first, then Prefer), so I'm keeping the format here, otherwise it would not be consistent with other examples.

STYLE.md
IronCore864: Alphanumeric is suggested by Ben in a previous comment so I'm gonna keep his version.

STYLE.md
IronCore864: Changed to "... highlighting lines in relatively long code blocks when necessary".

STYLE.md
IronCore864: How about "use triple-backtick followed by language or `{code-block}`"?

Link to https://mystmd.org/guide/code added.

dwilding: Ah thanks - the links says:

> You can use either `code` or `code-block` directive documented above or even a normal Markdown code block. All implementations in MyST are resolved to the same code type in the abstract syntax tree.

I think our default approach should be triple-backticks, as `{code-block}` messes with syntax highlighting (at least in VS Code).

IronCore864: OK, I updated it to triple backticks, and will rerun the build.

IronCore864:
Updated and fixed all suggestions in the comments. Merging now.

benhoyt (approved):

PR #1721, opened 2025-05-08
ci: pin workflows by hash

james-garner-canonical:
This PR updates our uses of actions in our actions and workflows to specify them by their hash, rather than by tag. This is a good idea because tags are movable, and so could result in the execution of malicious code in our workflows if a bad commit is added to any of those actions.

I believe that dependabot can be configured to open PRs to bump the commit hash (and version comment) when there are new releases, though I'm not sure what will happen with our current configuration.

### DIFF
--- a/.github/actions/update-charm-pins/action.yaml
+++ b/.github/actions/update-charm-pins/action.yaml
@@ -17,7 +17,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
       with:
         python-version: "3.12"
 
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -19,7 +19,7 @@ jobs:
             commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
     steps:
       - name: Checkout test charm repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
         with:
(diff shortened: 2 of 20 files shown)


tonyandrewmeyer:
I think with this change we can tighten the zizmor policy (perhaps even totally removing zizmor.yml to go back to the default behaviour), and perhaps even run at a stricter level. Would you be able to look into that?

james-garner-canonical:
Somehow the Python 3.8 unit tests on ubuntu [failed with a runner crash, potentially a segmentation fault](https://github.com/canonical/operator/actions/runs/14898042979/job/41844312517?pr=1721), so that's weird. They passed when rerunning the job.

james-garner-canonical:
> I think with this change we can tighten the zizmor policy (perhaps even totally removing zizmor.yml to go back to the default behaviour), and perhaps even run at a stricter level. Would you be able to look into that?

Looks like we can indeed drop our `zizmore.yml` config file. The [zizmor docs](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) indicate that official actions (like those beginning `actions/`) can be pinned to a version tag with the default settings -- would we want to use version tags in this case, and hashes only for third party actions?

benhoyt (approved):

tonyandrewmeyer (approved):

tonyandrewmeyer:
> Looks like we can indeed drop our `zizmor.yml` config file.

Nice :). I was wondering if we could switch persona (to pedantic or maybe even auditor) but it doesn't seem like it's worth it with the current rules (and might be annoying in one specific case with moving tags).

> The [zizmor docs](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) indicate that official actions (like those beginning `actions/`) can be pinned to a version tag with the default settings -- would we want to use version tags in this case, and hashes only for third party actions?

I'd say let's go with what you have now, and if we find it annoying that we're prompted to update too often (or falling behind becomes annoying) then we could loosen it up later.

PR #1726, opened 2025-05-09
fix: fix link in breakpoint output, remove link from Harness error message

dwilding:
This PR addresses a couple of issues:

- In the welcome message that's shown when a breakpoint is triggered, there's a link to https://juju.is/docs/sdk/debugging. This link is broken. I think a suitable replacement would be [`juju debug-code`](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/debug-code/).

- Harness raises an error if the charm tries to update an unknown config option. The error message includes a link to https://juju.is/docs/sdk/config, which redirects to [https://ops.readthedocs.io/en/latest/howto/manage-configurations.html](https://ops.readthedocs.io/en/latest/howto/manage-configurations.html). I was going to replace the link, but since this is Harness, I then figured that removing the link would be the more maintainable option.

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -2284,8 +2284,6 @@ class _TestingConfig(Dict[str, Union[str, int, float, bool]]):
             raise RuntimeError(
                 f'Unknown config option {key}; '
                 'not declared in `config.yaml`.'
-                'For help, see:'
-                'https://ops.readthedocs.io/en/latest/howto/manage-configurations.html'
             )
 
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
Thanks! The Harness error should maybe be charmcraft.yaml or mention both but like the link I think it's fine like this since Harness is deprecated.

benhoyt (approved):

dimaqq:
belated LGTM

PR #1733, opened 2025-05-15
fix: remove apparently unused class in harness

james-garner-canonical:
While investigating Harness code, we found this apparently unused (private) class. It (rightly) does not appear to be used in any charms.


tonyandrewmeyer (approved):
:tada:

dimaqq (approved):
💯 🚀

PR #1739, opened 2025-05-20
chore: bump documentation dependencies

tonyandrewmeyer:
Regenerate the documentation dependencies (`rm docs/requirements.txt; uvx --python=3.11 tox -e docs-deps`).

This is primarily to deal with `h11`, which has a security issue that's fixed in the new release (I don't believe we are impacted, but updating is safest and silences the alert).

Rather than only bumping `h11`, bump the rest as well. The ones that are not bug-fix releases only are:

* [anyio](https://anyio.readthedocs.io/en/stable/versionhistory.html) (seems unlikely to impact doc generation)
* certifi (the usual cert updates)
* [click](https://click.palletsprojects.com/en/stable/changes/) (drops Python 3.8 support but for docs that should be fine already, nothing else seems to impact us, awful release notes)
* [importlib-metadata](https://github.com/python/importlib_metadata/blob/main/NEWS.rst) (really just bug fixes)
* opentelemetry-api
* [lxml](https://github.com/lxml/lxml/releases) (apparently a security fix -- not sure why this isn't picked up)
* [markdown](https://github.com/Python-Markdown/markdown/releases/tag/3.8) (really just bug fixes)
* [packaging](https://packaging.pypa.io/en/stable/changelog.html) (shouldn't impact docs)
* [snowballstemmer](https://github.com/snowballstem/snowball/blob/master/NEWS) (the Python changes are just admin, the stemming changes are complex but I'm willing to assume would only improve things for us, however we are using this)
* [soupsieve](https://facelessuser.github.io/soupsieve/about/changelog/#27) (just new selectors)
* [sphinxext-opengraph](https://github.com/sphinx-doc/sphinxext-opengraph/releases/tag/v0.10.0) (nothing that should impact us, although the 3.12 support is good)
* [typing-extensions](https://github.com/python/typing_extensions/releases)
* [urllib3](https://github.com/urllib3/urllib3/releases/tag/2.4.0) (nothing that should impact us)

The two I wonder about are:
* typing-extensions: maybe this should always match what we use in static checking?
* opentelemetry-api: maybe this should always match what ops uses?

[Preview build](https://ops--1739.org.readthedocs.build/en/1739/)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,30 +6,30 @@
 #
 alabaster==1.0.0
     # via sphinx
-anyio==4.8.0
+anyio==4.9.0
     # via
     #   starlette
     #   watchfiles
 babel==2.17.0
     # via sphinx
-beautifulsoup4==4.13.3
+beautifulsoup4==4.13.4
     # via
     #   canonical-sphinx-extensions
     #   furo
     #   pyspelling
 bracex==2.5.post1
     # via wcmatch
-canonical-sphinx-extensions==0.0.23
+canonical-sphinx-extensions==0.0.27
     # via ops (pyproject.toml)
-certifi==2025.1.31
+certifi==2025.4.26
     # via requests
-charset-normalizer==3.4.1
+charset-normalizer==3.4.2
     # via requests
-click==8.1.8
+click==8.2.0
     # via uvicorn
 colorama==0.4.6
     # via sphinx-autobuild
-deprecated==1.2.15
+deprecated==1.2.18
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Thanks for the details!

dimaqq (approved):

dimaqq:
lxml was not picked up, because the new release was made to include newer libxml2.
It was libxml2 that was affected by CVE(s), https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=libxml2

PR #1734, opened 2025-05-16
docs: update code in K8s tutorial, with source in repo (part 2)

dwilding:
This PR adds the remaining example charms from the Kubernetes charm tutorial, continuing from #1718.

At the same time, I did a fair bit of cleanup: linting, formatting, replacing links to juju.is, making comments/docstrings consistent, and fixing small mistakes in the docs.

One significant change to the tutorial is that I've switched to using `charmcraft fetch-libs`, with libs specified in charmcraft.yaml.

To help with reviewing the changes, I've added comments at points that are particularly significant.

PR #1741, opened 2025-05-21
feat: support for config schema as Python classes

tonyandrewmeyer:
A new method is added to `ops.CharmBase`: `load_config`, which is the custom class equivalent of the `.config` property. The charmer passes a class to this method and gets back an instance of the class with the Juju config loaded into it.

The `load_config` method also converts `secret` type config from a `str` to an `ops.Secret`, since doing anything interesting with the secret option requires getting the actual secret. This is a 'lazy' Secret, like `SecretEvent.secret`, that does not have any data from Juju (not even validation that the secret can be accessed) until it is used.

Two error-handling modes are provided for `load_config`. In the default, "raise", any exceptions raised creating the class are re-raised and must be handled by the charm (or left for the charm to go to error state). The alternative is "blocked", where `load_config` will catch any `ValueError` (or subclass) and exit the charm with a zero exit status after setting an appropriate blocked status. This mode is suited for holistic charms (where handling the specific event is not necessary) that do not have complex status needs.

* Pydantic dataclasses
* Pydantic 2.x BaseModel classes (we'll use these in our documentation, as they're in common use).

This PR includes reference docs ([Preview](https://ops--1741.org.readthedocs.build/en/1741/reference/ops.html#ops.CharmBase.load_config)) and updates the configuration how-to ([Preview](https://ops--1741.org.readthedocs.build/en/1741/howto/manage-configuration.html)). A follow-up PR is planned that would make use of this functionality in the zero-to-hero tutorial - it will be simplest to do that after #1734 and any follow-up PRs are complete, to avoid conflicts.

More details are available in the [spec](https://docs.google.com/document/d/1vj2pTvVRXClCrYM7tUjft3ngISSsFdBQD1B-CzOG1Vs/edit?usp=sharing).


ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+
+        config: dict[str, bool | int | float | str | model.Secret] = kwargs.copy()
+        fields = set(_juju_option_names(cls))
+        for key, value in self.config.items():
+            attr = key.replace('-', '_')
+            if not attr.isidentifier():
+                if errors == 'raise':
+                    raise ValueError(f'Invalid attribute name {attr}')
+                self.unit.status = model.BlockedStatus(f'Invalid attribute name {attr}')
+                raise _Abort(0)
```
tonyandrewmeyer, line 1476: When I hear back from the Juju team about what form an option can take I'll either remove this or add a comment explaining why it's needed.

ops/charm.py
```diff
@@ -1438,13 +1541,15 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
...
-    class is mostly for the framework to understand what the charm has defined.
+    This is read from ``metadata.yaml``, ``config.yaml``, and ``actions.yaml``.
+    Generally charms will define this information, rather than reading it at
+    runtime. This class is mostly for the framework to understand what the charm
+    has defined.
 
     Args:
         raw: a mapping containing the contents of metadata.yaml
         actions_raw: a mapping containing the contents of actions.yaml
+        config_raw: a mapping containing the contents of config.yaml
```
tonyandrewmeyer, line 1552: This is leftover from #1648.

test/test_charm.py
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
...
+        type: string
+        default: foo
+    my-secret:
+        type: secret
+"""
+
+
+@pytest.mark.parametrize('charm_class', _test_classes)
+def test_config_init(charm_class: type[ops.CharmBase], request: pytest.FixtureRequest):
+    harness = testing.Harness(charm_class, config=_config)
```
tonyandrewmeyer, line 1441: @benhoyt I'll use Scenario for the new package. For this one, with tests in `test_charm`, would you still prefer that I rewrite these to use Scenario?

docs/howto/manage-configurations.md
```diff
@@ -30,21 +49,28 @@ In the `src/charm.py` file of the charm project, in the `__init__` function of t
 self.framework.observe(self.on.config_changed, self._on_config_changed)
 ```
 
-Then, in the body of the charm definition, define the event handler. Here you may want to read the current configuration value, validate it (Juju only checks that the *type* is valid), and log it, among other things. Sample code for an option called `server-port`, with type `int`, and default value `8000`:
+Also in the `__init__` function, load the config into the config class that you
+defined. Pass `errors='blocked'` to have the charm exit after setting a blocked
+status if the configuration doesn't validate against the class you defined.
```
benhoyt, line 54: Maybe add something like "The default is `errors='raise'`, which means that `load_config` will raise the validation error."

docs/howto/manage-configurations.md
```diff
@@ -30,21 +49,28 @@ In the `src/charm.py` file of the charm project, in the `__init__` function of t
 self.framework.observe(self.on.config_changed, self._on_config_changed)
 ```
 
-Then, in the body of the charm definition, define the event handler. Here you may want to read the current configuration value, validate it (Juju only checks that the *type* is valid), and log it, among other things. Sample code for an option called `server-port`, with type `int`, and default value `8000`:
+Also in the `__init__` function, load the config into the config class that you
```
benhoyt, line 52: Is doing this in `__init__` the first/main pattern we want to recommend? I was expecting just to see `config = self.load_config(...)` in the config-changed handler.

docs/howto/manage-configurations.md
```diff
@@ -30,21 +49,28 @@ In the `src/charm.py` file of the charm project, in the `__init__` function of t
...
         return
-    
-    logger.debug("New application port is requested: %s", port)
-    self._update_layer_and_restart(None)
+    logger.info('Changing wiki name to %s', name)
+    self.set_wiki_name(name)
 ```
 
-> See more: [](ops.CharmBase.config)
+> See more: [](ops.CharmBase.load_config)
```
benhoyt, line 73: Do we want to link both `.config` and `.load_config`, or mention the "raw" `.config` anywhere above? It seems to me we should.

ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+
+        The raw Juju config is passed to the config class's ``__init__``, as
+        keyword arguments, with the following changes:
+
+        * ``secret`` type options have a :class:`model.Secret` value rather
+          than the secret ID. Note that the secret object is not validated by
+          Juju at this time, so may raise :class:`SecretNotFoundError` when it
+          is later used if the secret does not exist or the unit does not have
+          permission to access it.
+        * dashes in names converted to underscores.
```
benhoyt, line 1443:
```suggestion
        * dashes in names are converted to underscores.
```

ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+        """Load the config into an instance of a config class.
+
+        The raw Juju config is passed to the config class's ``__init__``, as
+        keyword arguments, with the following changes:
+
+        * ``secret`` type options have a :class:`model.Secret` value rather
+          than the secret ID. Note that the secret object is not validated by
+          Juju at this time, so may raise :class:`SecretNotFoundError` when it
+          is later used if the secret does not exist or the unit does not have
+          permission to access it.
```
benhoyt, line 1442:
```suggestion
          is used later (if the secret does not exist or the unit does not have
          permission to access it).
```

ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+        passed through to the config class ``__init__``.
+
+        Args:
+            cls: A class that will accept the Juju options as keyword arguments.
+            errors: what to do if the config is invalid. If ``blocked``, the
+                charm will exit successfully (this informs Juju that
+                the event was handled and it will not be retried) after setting
+                an appropriate blocked status. If ``raise``, ``load_config``
+                will not catch any exceptions, leaving the charm to handle
+                errors.
```
benhoyt, line 1455:
```suggestion
            errors: what to do if the config is invalid. If ``blocked``, this
                will set the unit status to blocked with an appropriate message
                and then exit successfully (this informs Juju that
                the event was handled and it will not be retried).
                If ``raise``, ``load_config``
                will not catch any exceptions, leaving the charm to handle
                errors.
```

ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+            # that using model.get_secret so that it's entirely lazy, in the
+            # same way that SecretEvent.secret is.
+            if option_type and option_type.type == 'secret':
+                assert isinstance(value, str)  # Juju will have made sure of this.
+                value = model.Secret(
+                    backend=self.model._backend,
+                    id=value,
+                    _secret_set_cache=self.model._cache._secret_set_cache,
+                )
+            config[attr] = value
```
benhoyt, line 1490: How will a secret field be type-annotated in the config class? As `ops.Secret`? Is looking it up in meta better than relying on that fact?

test/test_charm.py
benhoyt: No, this is fine then in `test_charm.py`

benhoyt (changes requested):

docs/howto/manage-configurations.md
tonyandrewmeyer: I'm the "don't do things in __init__" guy, but I think this is something better there than in either the holistic handler or a config-changed handler.

If the config doesn't match the schema and you're using `errors='blocked'` then I think you want to stay in blocked until the Juju user changes the config. If you have the loading in config-changed, then when the next event (update-status for example) comes along, you won't load the config and the blocked status will get replaced with something else (unless you're *also* loading it in collect-status).

What we really want is for this to run before the config is actually changed, but since that's not possible, we want to get it as early as we can, and that's `__init__`.

ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+                raise
+            # We exit with a zero code because we don't want Juju to retry
+            # (the config needs to be fixed by the Juju user), and we don't
+            # want the status we just set to be overridden by an error
+            # status.
+            self.unit.status = model.BlockedStatus(f'Invalid config: {e}')
+            raise _Abort(0) from e
+
+
+def _juju_option_names(cls: type[object]) -> Generator[str]:
```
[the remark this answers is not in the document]
tonyandrewmeyer: There was a bunch of discussion earlier (particularly, but not only, with @james-garner-canonical) about this. I do feel it's most convenient for the charmer if we allow the config class to not have to have every Juju option as a field (without having to tack on a `**kwargs` in `__init__`).

The most obvious use-case is if you have a 12-factor charm and want to extend it, adding an extra couple of options, then it's much more convenient to have a config class that has the ones you're adding, rather than all the 12-factor options as well (and then have to update that whenever you update the 12-factor dependency). If the 12-factor base adopted this feature then I guess you could inherit from their config class, which is perhaps cleanest, but it still seems nice to let people only model a subset.

With the default `__init__` from standard library dataclasses, we could use the `__init__` signature instead of doing it this way, but the Pydantic BaseModel class has a `*args, **kwargs` style signature so doesn't provide what we need.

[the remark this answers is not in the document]
benhoyt: Makes sense, thanks.

ops/charm.py
tonyandrewmeyer: Presumably as `ops.Secret`, yes. If we look it up in meta then we know that Juju has done some basic validation on the secret, and I think that's better. If we base it on the type annotation of the attribute then people could do things like have a Juju type of string and as long as it was in the form of a secret, it'd still work. I think that's too cutesy/hack-y and we should force people to use the real Juju secret type.

ops/charm.py
tonyandrewmeyer: No response, but I've tested things out myself and it seems like anything that's both a valid Go string and a valid YAML field name is fine. There are lots of things that work fine in Juju but are not valid Python identifiers, like:

* `t$`
* `class`
* `a b`
* `2`
* `1a`

These are all fine as dictionary keys (as in `Model.config`) but not really as class/object attribute names. You can do a `**kwargs` and `setattr` to get them into the `__dict__` but it's pretty weird.

However, I guess you already know that you've chosen inappropriate config names when you write the class, and this was predominately there to protect against bad naming functions, which we don't provide any more. So I'll remove it.

ops/charm.py
```diff
@@ -1417,6 +1423,103 @@ def config(self) -> model.ConfigData:
...
+    """Iterates over all the option names to include in the config YAML."""
+    try:
+        yield from (field.name for field in sorted(dataclasses.fields(cls)))  # type: ignore
+    except TypeError:
+        pass
+    else:
+        return
```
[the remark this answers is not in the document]
tonyandrewmeyer: What specifically do you find awkward?

Is it the `else: return`? I could put the return inside the first block; I think it's more correct to be in the `else`, but I don't object to it there. So:

```suggestion
    try:
        yield from (field.name for field in sorted(dataclasses.fields(cls)))  # type: ignore
        return
    except TypeError:
        pass
```

Or is it that it's creating a generator and then doing a `yield from`, rather than a `yield from` with an existing iterator? I wish `dataclasses.fields` returned a dictionary, so that it could be `yield from sorted(dataclasses.fields(cls))`. I like how the `yield from` makes it clear that it's just wrapping another iterator, but your version does avoid creating the extra generator.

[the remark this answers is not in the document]
benhoyt: It was both things that I think made it more complicated than necessary. The `else: return` just seems unnecessarily confusing, even if arguably correct. And I think of `yield from` as something to use when you have an existing list or generator, so it seems weird to build a generator only to use `yield from` to un-generator it; also the one-liner aspect feels unnecessarily terse.

I don't have super-strong feelings about it, just trying to keep things simple.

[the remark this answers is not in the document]
benhoyt: I'm fine with that too.

ops/charm.py
```diff
@@ -1417,6 +1423,108 @@ def config(self) -> model.ConfigData:
         """A mapping containing the charm's config and current values."""
         return self.model.config
 
+    def load_config(
+        self,
+        cls: type[_ConfigType],
+        *args: Any,
+        errors: Literal['raise', 'blocked'] = 'raise',
+        **kwargs: Any,
+    ) -> _ConfigType:
```
james-garner-canonical, line 1432: I wonder if it would be a better separation of concerns for the API if we accepted a `factory: Callable[..., _ConfigType]` instead of a `cls: type[_ConfigType]`. This would let us drop `*args` and `**kwargs` from this method's signature (they can be handled by the user's factory, `functools.partial`, etc).

The new `load_config` signature could look like this:
```py
    def load_config(
        self,
        factory: Callable[..., _ConfigType],
        errors: Literal['raise', 'blocked'] = 'raise',
    ) -> _ConfigType:
```
Example:
```py
# old
load_config(MyConfigClass, 1, 2, 3, foo=4)
# new
factory = partial(MyConfigClass, 1, 2, 3, foo=4)  # type checking error if MyConfigClass doesn't accept those arguments
load_config(factory)
``` 

Instead of `_juju_option_names` (which requires the class), we could use `inspect.signature` to pass only the config options the user's callable accepts (or the whole config, if it accepts variadic keyword arguments). For example, after building the config dict, the kwargs to pass to the factory in our `try: return` could be constructed like this:
```py
params = inspect.signature(factory).parameters
if not any(param.kind is inspect.Parameter.VAR_KEYWORD for param in params.values()):
    # filter config to the keyword arguments that the factory accepts
    kinds = {inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY}
    names = {name for name, param in params.items() if param.kind in kinds}
    config = {arg, value for arg, value in config.items() if arg in names}
```
This would allow us to drop the `_juju_option_names` function from this PR, and should support the same use cases of 12 factor charms that care about only a subset of the config, or multiple small config classes.

tonyandrewmeyer: I'd be more of a fan of using `inspect.signature` if we weren't pushing pydantic models, where the `__init__` uses `**`. The signature of the class doesn't:

```python
>>> class C(pydantic.BaseModel):
...     x: str = 'foo'
...     y: int = pydantic.Field(42)
...     z: list[str] = pydantic.Field(default_factory=list)
...     
>>> inspect.signature(C)
<Signature (*, x: str = 'foo', y: int = 42, z: list[str] = <factory>) -> None>
>>> inspect.signature(C.__init__)
<Signature (self, /, **data: 'Any') -> 'None'>
```

But I'm not sure how much either of these is part of the pydantic API as such and how much just implementation detail (as opposed to standard library dataclasses, where the `__init__` signature is very explicitly part of the API).

`signature` is also a bit slow (Scenario had this issue):

```python
$ uvx --python=3.10 python -m timeit -s 'from inspect import signature
def x(a, b=1): pass
' 'signature(x)'
20000 loops, best of 5: 14.2 usec per loop
````

But for this use-case that probably doesn't matter.

I feel the three important cases are:
 * A simple dataclass or Pydantic base model class is simple to use.
 * Passing in extra arguments, possibly only used to transform one (e.g. passing in model to get a relation) is not difficult.
 * The 12-factor case (especially) and the 'config subset' case (more generally) is not difficult.

I don't believe that the `_juju_options_names` function is problematic as an implementation detail. It has three pretty straightforward methods that are easily explained and easily justified for getting the set of fields.

In terms of usage:

```
# Simple
@dataclasses.dataclass(frozen=True)  # Nothing really changes with a BaseModel subclass.
class Config:
    log_level: str
    port: int = 8000

typed_config = self.load_config(Config)  # current
typed_config = self.load_config(Config)  # proposed

# Extra args
@dataclasses.dataclass(frozen=True)  # A BaseModel subclass can be similar.
class Config:
    log_level: str
    use_tls: bool
    model: dataclasses.InitVar[ops.Model]
    def __init__(self, *, model: ops.Model, log_level: str, use_tls: bool):
        # get a relation, check something in the databag to validate the use_tls setting is ok

typed_config = self.load_config(Config, model=self.model)  # current
typed_config = self.load_config(functools.partial(Config, model=self.model))  # proposed

# subset of options

class Config:
    log_level: str
    # The next two lines only needed in proposed.
    def __init__(self, log_level: str, **_):
        object.__setattr__(self, 'log_level', log_level)
        
class Config2(pydantic.BaseModel):
    log_level: str
    # The next two lines only needed in proposed.
    class Config:
        extra = "ignore"

# I suppose the factory could also discard the extra args instead.
```

There are two other important factors:
* The signature would ideally be consistent across `load_config`, `load_params`, and `Relation.load`. For actions, I don't think there's a use for only loading a subset of params (you might want `**` but that would be if you have the `additionalParameters` set to `True` and then you're presumably actually using them). For actions, I think there are use-cases for passing extra args, and it'd end up looking like config. For `Relation.load` it's a lot more complex because the databag potentially gets filled over multiple changes. It seems like the proposed signature would be ok there too.
* I haven't given up on the schema generation piece, which seems to me to be the most valuable part. Adding a layer between the load and the class that isn't there in the generation doesn't seem great. A 1-1 between 'this is the YAML you get generated' and 'these are the keyword arguments you get' is a valuable property in my opinion. That can still be true with a factory but it relies on people using the factory in a specific way.

Comment (james-garner-canonical) [reply ops/charm.py:1432]: I think what I find the most convincing is that it's good for the object construction logic to mirror the schema generation logic, since we do want to add schema generation.

I agree that `inspect.signature` is straying towards the dark side of Python, but I'm not convinced that using `dir` and `__annotations__` is any cleaner or less error prone (I've just added a comment pointing out an inconsistency I noticed in `_juju_option_names`). Regarding explainability, I actually think that using the callable's signature (with `inspect`) has an edge over the current approach: "It will be called with the config items that match its signature, or all of the config items if it takes **kwargs." vs "It will be called with the config items that match its `fields`/`model_fields` (if it's a `dataclass` or Pydantic model), or some combination of its annotations and class attributes (if it's neither)."  But if we can streamline the handling of the basic class case, I think that would make `inspect.signature` much less attractive -- e.g. we could align the basic class behaviour with `dataclasses` behaviour.

I don't think the performance of `inspect` would be too problematic if we did want to use it here, since we already import it at runtime in `framework.py`, so we're paying the module initialization cost either way, but ultimately I'm happy to close off this idea in favour of the current approach.

PS: regarding the 'subset of options' examples, note that if we used `inspect.signature` as proposed then the additional lines wouldn't be needed -- I believe they're only needed if we drop both `inspect.signature` and `_juju_option_names`. Actually I think the extra lines are actually only needed for the `dataclass` case, since the default `extra` behaviour for `pydantic` is `'ignore'`.

File: test/test_charm.py [hunk 14]
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
 name: my-charm
 """)
     assert meta.charm_user == 'root'
+
+
+JUJU_TYPES = typing.Union[bool, int, float, str]
```
james-garner-canonical, line 1266: Nit: could inline this, it'd be pretty concise with future annotations `bool | int | float | str`

test/test_charm.py
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
...
+class BaseTestCharm(ops.CharmBase):
+    _config_errors: typing.Literal['blocked', 'raise'] | None = None
+
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        if self._config_errors:
+            self.typed_config = self.load_config(self.config_class, errors=self._config_errors)
+        else:
+            self.typed_config = self.load_config(self.config_class)
+        # These should not have any type errors.
```
james-garner-canonical, line 1328: Any reason we don't interact with `my_bool` here?

test/test_charm.py
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
...
+        if my_int < 0:
+            raise ValueError('my_int must be zero or positive')
+        self.my_int = my_int
+        if not isinstance(my_str, str):
+            raise ValueError('my_str must be a string')
+        self.my_str = my_str
+        if my_secret is not None:
+            if not isinstance(my_secret, ops.Secret):
+                raise ValueError('my_secret must be a secret')
+            self.my_secret = my_secret
```
james-garner-canonical, line 1316: I'm curious why we do all this validation in this case, but only a tiny subset of it for the `dataclass`?

test/test_charm.py
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
...
+            raise ValueError('my_int must be zero or positive')
+
+
+class MyDataclassCharm(BaseTestCharm):
+    @property
+    def config_class(self) -> type[_ConfigProtocol]:
+        return MyDataclassConfig
+
+
+_test_classes: list[type[ops.CharmBase]] = [MyCharm, MyDataclassCharm]
```
james-garner-canonical, line 1371:
```suggestion
_test_classes: list[type[BaseTestCharm]] = [MyCharm, MyDataclassCharm]
```

test/test_charm.py
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
...
+    my-str:
+        type: string
+        default: foo
+    my-secret:
+        type: secret
+"""
+
+
+@pytest.mark.parametrize('charm_class', _test_classes)
+def test_config_init(charm_class: type[ops.CharmBase], request: pytest.FixtureRequest):
```
james-garner-canonical, line 1440:
```suggestion
def test_config_init(charm_class: type[BaseTestCharm], request: pytest.FixtureRequest):
```
I think this type annotation should eliminate the need for the `type: ignore` and `typing.cast` when getting `typed_config` in the test body. Maybe an explicit annotation in the body of `BaseTestCharm` would be needed too (e.g. `typed_config: _ConfigProtocol`), but it might just work

ops/charm.py
benhoyt: One thing we could do (for now at least) is only support dataclasses and pydantic classes (with `model_fields`), and error on anything else. Realistically that's all we're suggesting anyway, and it's always much easier to loosen a restriction later than tighten it.

ops/charm.py
```diff
@@ -1417,6 +1423,108 @@ def config(self) -> model.ConfigData:
...
+    # We filter out any ClassVar annotations, since they wouldn't have different
+    # values for different instances of the class. We also filter out any
+    # attributes with leading underscores, and any attributes that are methods.
+    attrs.extend((a for a, t in cls.__annotations__.items() if get_origin(t) is not ClassVar))
+    for attr in sorted(set(attrs)):
+        # If the attribute comes from __annotations__, and doesn't have a
+        # default value, then it's not accessible with hasattr
+        if attr.startswith('_') or (hasattr(cls, attr) and callable(getattr(cls, attr))):
+            continue
+        yield attr
```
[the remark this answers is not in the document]
tonyandrewmeyer: Interesting idea to use a transformed version. It reminds me of `dataclass_transform`, which I think we'd want to try to use (or perhaps tell charmers to use) if we had a newer Python in order to make things clearer to type checkers.

> It seems tricky to get this right.

I'm not sure I agree with this. I wouldn't say that my code is typically so perfect that one bug means that getting it right is tricky :laughing:.

My expectation is that dataclasses and pydantic base model subclasses will make up essentially all use. However, I don't think forcing those is a good idea - for example, say someone wants to use `attrs` then I think that should 'just work'. The simple `dir()` approach works fine there even when you're not using type annotations, whereas the "turn it into a dataclass" breaks (because of conflicting `__slots__` I think).

I think there's merit in the fallback behaviour being as simple as possible. I lean more towards removing special-cases like `ClassVar` rather than adding more magic - it's also simpler to handle getting too *many* arguments than too *few*.

test/test_charm.py
```diff
@@ -1253,3 +1261,316 @@ def test_meta_charm_user_default():
...
+def test_config_init_non_default(charm_class: type[ops.CharmBase], request: pytest.FixtureRequest):
+    harness = testing.Harness(charm_class, config=_config)
+    request.addfinalizer(harness.cleanup)
+    harness.update_config({
+        'my-bool': True,
+        'my-float': 2.71,
+        'my-int': 24,
+        'my-str': 'bar',
+    })
+    harness.begin()
```
james-garner-canonical, line 1466: Is the difference between this test and the previous one purely that harness sets up a simulated config changed event for the charm? (And the config is slightly different).

And because these test charms do all their logic in `__init__` and don't observe `config-changed`, there's basically no difference on the charm side?

james-garner-canonical: Ah no wait, in the previous test there's no config data at all (just the config schema), and we're doing the default values case -- that should have been obvious to me, sorrr.

test/test_charm.py
tonyandrewmeyer: I was trying to make these classes somewhat look like they would in actual use (other than the generic "my_x" names), so it's there for that purpose rather than strictly for the test itself, apart from the "positive, non-zero" check, which is there to verify that we can do extra validation.

But that is a fair point about this one having runtime type checking and the dataclass not having it. It shouldn't be required, because Juju is ensuring that the type is correct. So these should just be asserts I guess.

test/test_charm.py
tonyandrewmeyer: `my_bool` doesn't have a default value (in the Juju schema), so unless the user passes in a value it's missing from `Charmbase.config` (because it's missing from `config-get`). "This attribute may or may not be present" doesn't really translate directly to Python data classes so I felt like `None` would be the appropriate way to represent it (since you can't set the other types to None/null).

I could add in some sort of `if my_bool is not None: ...` here if you'd prefer. I think it's also less clear what you'd do to force a type error when it's not a `bool`, since Python lets you treat just about anything as a Boolean - `not my_bool` or `if my_bool` and so on would be perfectly valid even if `my_bool` was a float.

test/test_charm.py
tonyandrewmeyer: Thanks - I didn't originally have a `BaseTestCharm` (I had 4 separate charms all inheriting directly from CharmBase) and missed this when doing the refactoring.

ops/charm.py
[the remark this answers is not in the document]
tonyandrewmeyer: I've changed this so that it doesn't use `__annotations__` directly, and is simpler in general, only ignoring methods (which seem to me to be clearly not fields) and anything with a leading underscore (maybe more questionable, but I think a reasonable default, and also conveniently gets rid of all the magic methods).

ops/charm.py
tonyandrewmeyer: > it's always much easier to loosen a restriction later than tighten it.

I meant to make a point like this above too: I think one of the strongest arguments in favour of dropping the pass-through `*args` and `**kwargs` is that we can easily add that later, whereas dropping them is essentially impossible. If they're gone then it might as well be a factory rather than the class, regardless of whether we're limiting the arguments that get passed through or not.

> My pitch here is that we should use dataclasses.fields on non-dataclasses too, so we follow the exact same logic for regular classes as we do for dataclasses -- see https://github.com/canonical/operator/pull/1741#discussion_r2108223904.

My argument against that is that the very first attempt I made at using that, with `attrs`, failed. So I have no confidence that it will work a lot of the time, and I don't want a fallback for the fallback.

So: I'm convinced enough to drop `*` and `**`, but still strongly in favour of keeping a simple-as-possible non-`inspect-signature` fallback to limit which fields are passed through.

@benhoyt and @james-garner-canonical you've both at times been happy with resolving this but then the discussion has continued, so I'm not sure whether that's still the case or not. If you'd like something other than the above, let's finalise this in or after daily today.

ops/charm.py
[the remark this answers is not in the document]
james-garner-canonical: As far as I can tell, `attrs` follows the same behaviour as `dataclasses` here -- for the example class `C`, only `'b'` is added to the generated `__init__` method, just like `dataclasses`.

I think from a user facing perspective, the cleanest explanation (and least magical feeling behaviour) is for both `dataclasses` and all other classes (that aren't Pydantic models), to follow the same logic -- that of `dataclasses.fields`.

I think that whether it's better to do this by rolling our own `dataclasses`-like logic like the current approach or by calling `dataclasses.fields` on non-dataclasses is a separate question (implementation detail). There are clearly unexpected gotchas either way. I lean towards calling `dataclasses.fields`, but I'd be happy if we just made the current approach align with the `dataclass` behaviour.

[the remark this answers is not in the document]
james-garner-canonical: Thanks for spotting the `__slots__` issue btw. That comes up when trying to use `cls.__dict__` in `type` whenever `__slots__` is present (it's not specifically `dataclasses` problem).  It can be solved by filtering the items in `__slots__` out of `__dict__` for `type`. Just for reference, the full implementation for calling `dataclasses.fields` would then be:
```py
slots = getattr(cls, '__slots__', {})
cls_dict = {k: v for k, v in cls.__dict__.items() if k not in slots}
cpy = type(cls.__name__, cls.__bases__, cls_dict)
fields = [field.name for field in dataclasses.fields(dataclasses.dataclass(cpy))]
yield from sorted(fields)
```

ops/charm.py
tonyandrewmeyer: > So: I'm convinced enough to drop * and **, but still strongly in favour of keeping a simple-as-possible non-inspect-signature fallback to limit which fields are passed through.

Argh, except that I found the flaw in this plan when trying to implement it. If it's `Callable[..., _ConfigType]` then the method doesn't have the class object to pass into `_juju_option_names` so `dataclasses.fields` and `.model_fields` can't really be used. So I'm back to keeping it as it is in the branch now. I'll put it on the discussion list for daily.

james-garner-canonical: Sorry if my first reply on this thread made things unclear. I was convinced at that time to keep the current signature of `load_config`. It sounded like Ben was convinced too. I believe we're all in agreement on not using `inspect`. That means we need to pass the class to `load_config` rather than a factory. I think that in turn means we shouldn't drop the passthrough `*args`, `**kwargs`.

benhoyt: Yep, I think I'm happy with what's in the branch now. However, let's the 3 of us finalise after daily just to be sure we're all on the same page.

test/test_charm.py
james-garner-canonical: That's fair. I guess we could us assertions here (e.g. `if my_bool is not None: assert my_bool is True or my_bool is False`, but I'm fine with leaving it as is.

james-garner-canonical (approved):

tonyandrewmeyer:
Oops, that was not meant to be a close...

test/test_charm.py
```diff
@@ -1253,3 +1261,310 @@ def test_meta_charm_user_default():
...
+
+    def __post_init__(self):
+        if self.my_int < 0:
+            raise ValueError('my_int must be zero or positive')
+
+
+class MyDataclassCharm(BaseTestCharm):
+    @property
+    def config_class(self) -> type[_ConfigProtocol]:
+        return MyDataclassConfig
```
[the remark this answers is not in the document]
james-garner-canonical: I had forgotten about that test (sorry), but yes, I do think it would be worth testing with all 4 types. 

Thinking out loud, maybe easiest to parametrize that test with the existing test classes (and different expected outcomes?) and use a larger `_config` than define partial versions of the 4 types, but either way works.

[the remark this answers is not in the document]
james-garner-canonical: Perfect!

Issue #1655, opened 2025-03-28
Rework the README into a streamlined set of steps for trying Ops


dwilding:
The README (which also [displays on PyPI](https://pypi.org/project/ops/)) could be more approachable for people who want to quickly get the point of Ops and try out Ops.

Mainly, I think it would be great if the README had a tight, self-contained set of steps for making a simple charm and showing what the charm does when deployed & tested. We should be able to take advantage of Concierge and an improved charmcraft profile (upcoming) to limit the number of steps.

benhoyt:
Definitely worth discussing how to improve the README. However, I'm a little skeptical that "self-contained steps for making a simple charm" should be in the README. Shouldn't that be in a tutorial doc? And if our existing tutorials are too "un-tight", should we focus our effort there, or add another basic tutorial?

dwilding:
Let's discuss further, to clarify the best way to make use of the README. I do agree with this sentiment:

> And if our existing tutorials are too "un-tight", should we focus our effort there, or add another basic tutorial?

With the caveat that a tutorial is intended to be for learning. What I'd like to see _somewhere_ is an answer to the question "I just want to have a basic charm deployed from source on my machine, so that I can inspect it myself" - I guess what I have in mind is more like a script than a tutorial or steps that need to be followed manually.

dwilding:
I'm tagging this one as roadmap, as the README currently relies on getting an httpbin charm from `charmcraft init`. We're planning to remove this profile in Charmcraft 4.0 (see https://github.com/canonical/charmcraft/pull/2295). So we'll either need to reference the same charm from within the Ops repo (see #1743) or switch to a different example.

Related: https://github.com/canonical/operator/issues/1803

dwilding:
Per Dima's point on https://github.com/canonical/operator/issues/1817, while doing this work we should decide whether the README should include unit testing.

benhoyt:
@dwilding Do you still think this needs doing? Should we consider it for the 26.04 roadmap?

dwilding:
My current thinking is, let's make the README less like a tutorial (in contrast to my original idea). Better for people to work through our actual tutorials. I think the README can be cut down to a survey of the [httpbin demo charm](https://github.com/canonical/operator/tree/main/examples/httpbin-demo): here's where Ops is used, here's where unit testing happens, etc. We've made a good number of improvements to that charm since migrating it from Charmcraft.

dwilding:
[...]

PR #1743, opened 2025-05-21
docs: add httpbin charm from Charmcraft as an example charm

dwilding:
In Charmcraft 4.0, we're planning to remove the `simple` profile of `charmcraft init` (see https://github.com/canonical/charmcraft/pull/2295). This PR adds `simple`'s charm as a demo of Ops, so that the charm isn't lost.

I took the charm that was generated by `charmcraft init` and changed/removed placeholder comments & files that were more suitable for a template charm. I also made sure that the charm is formatted correctly for the Ops repo and passes our linting.

Within the charm directory, I tested that the following work:

- `tox -e static` - For now, this runs static checks on _src/charm.py_ only. When we make future improvements including switching to uv (see #[1805](https://github.com/canonical/operator/issues/1805)) we could expand `static` to cover the testing code too.

- `tox -e unit` and `tox -e integration`

- Manual pack and deploy, as documented in _examples/README.md_ and (for the time being) _README.md_.

I also added the charm to our CI job that runs unit tests for the example charms.

---

Follow-up items in future PRs:

- https://github.com/canonical/operator/issues/1655
- https://github.com/canonical/operator/issues/1805


examples/httpbin-demo/charmcraft.yaml
```diff
@@ -0,0 +1,76 @@
...
+  # An OCI image resource for each container listed above.
+  # You may remove this if your charm will run without a workload sidecar container.
+  httpbin-image:
+    type: oci-image
+    description: OCI image for httpbin
+    # The upstream-source field is ignored by Juju. It is included here as a
+    # reference so the integration testing suite knows which image to deploy
+    # during testing. This field is also used by the 'canonical/charming-actions'
+    # Github action for automated releasing.
+    upstream-source: kennethreitz/httpbin
```
dimaqq, line 76: This is 6 years old and only support AMD64.

dwilding: Well spotted, thanks. I've tracked this for future improvement in https://github.com/canonical/operator/issues/1805

docs/howto/legacy/write-legacy-unit-tests-for-a-charm.md
```diff
@@ -5,7 +5,7 @@
 
 ## Testing basics
 
-Here’s a minimal example, taken from the `charmcraft init` template with some additional comments:
```
dwilding, line 8: Removing this because `charmcraft init` is not going to produce the same charm after this month.

examples/httpbin-demo/requirements.txt
```diff
@@ -0,0 +1 @@
+ops < 4
```
dimaqq, line 1: 😆

examples/httpbin-demo/src/charm.py
```diff
@@ -0,0 +1,94 @@
...
+            # The config is good, so update the configuration of the workload
+            container = self.unit.get_container('httpbin')
+            # Push an updated layer with the new config
+            try:
+                container.add_layer('httpbin', self._pebble_layer, combine=True)
+                container.replan()
+            except ops.pebble.ConnectionError:
+                # We were unable to connect to the Pebble API, so we defer this event
+                self.unit.status = ops.MaintenanceStatus('waiting for Pebble API')
+                event.defer()
```
james-garner-canonical, line 64: Should we take this opportunity to avoid recommending `defer`? @tonyandrewmeyer

examples/httpbin-demo/src/charm.py
```diff
@@ -0,0 +1,94 @@
...
+                return
+
+            logger.debug("Log level for gunicorn changed to '%s'", log_level)
+            self.unit.status = ops.ActiveStatus()
+        else:
+            # In this case, the config option is bad, so block the charm and notify the operator.
+            self.unit.status = ops.BlockedStatus(f"invalid log level: '{log_level}'")
```
james-garner-canonical, line 71: Should we also take this opportunity to recommend the collect status pattern?

examples/httpbin-demo/tests/integration/test_charm.py
```diff
@@ -0,0 +1,45 @@
...
+from pytest_operator.plugin import OpsTest
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path('./charmcraft.yaml').read_text())
+APP_NAME = METADATA['name']
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
```
james-garner-canonical, line 30: Would be nice to rewrite this with `jubilant`.

james-garner-canonical (commented):
I've made some suggestions for where we might want to use current best practices. Happy for these to be deferred to follow up issues if you prefer.

examples/httpbin-demo/src/charm.py
tonyandrewmeyer: The main reason I'd like to have this charm still exist somewhere is so that anyone that is currently using `charmcraft init --profile=simple` to do a basic charming demo can just replace that by cloning `canonical/operator` and `cd examples/httpbin-demo`.

So I'm not overly concerned about making this a great example of how to write a charm, more that it's still available with essentially the same functionality. I think if you want to know how to write a simple charm you should work through the K8s tutorial (or use the folders that have that pre-made for you).

On the other hand, I guess if it's in our repo it implies some sort of blessing. So I'm not opposed to reworking this to not use `defer()`. I think it'd be better as a follow-up PR, though.

examples/httpbin-demo/tests/integration/test_charm.py
tonyandrewmeyer: I do think this would be worth doing. If people are using the tests as part of their "this is charming" demo, I don't object to forcing them to demo Jubilant rather than python-libjuju/pytest-operator.

But also in a follow-up PR, in my opinion.

dwilding: Good suggestion, thanks. I've added this to the catch-all issue for improvements: https://github.com/canonical/operator/issues/1805

examples/httpbin-demo/src/charm.py
dwilding: Also added this to the catch-all issue for improvements: https://github.com/canonical/operator/issues/1805

examples/httpbin-demo/src/charm.py
dwilding: Does no comment from @tonyandrewmeyer imply do it? :wink:

Switching to collect status should a relatively quick refactor. But as I start to work on this, I see something else that will need to be refactored: `event.workload` vs `self.unit.get_container('httpbin')`.

So I'd say let's merge this PR as-is, then I'll do that refactoring in a follow-up PR over the next couple of days.

tonyandrewmeyer: > Does no comment from @tonyandrewmeyer imply do it? 😉

It was meant to imply the same answer as for the other thread :) - probably worth doing but probably in a later PR.

james-garner-canonical (approved):

PR #1744, opened 2025-05-21
feat: add juju topology labels

### LINKED ISSUE #1729: implement Juju (COS) topology labels for tracing

dimaqq:
Add Juju topology labels as used by the Observability team.

Fixes #1729

### DIFF
--- a/docs/reference/ops-tracing.rst
+++ b/docs/reference/ops-tracing.rst
@@ -8,12 +8,24 @@
 Open Telemetry resource attributes
 ----------------------------------
 
+The following well-known and additional resource attributes are set:
+
 - ``service.namespace`` the UUID of the Juju model.
 - ``service.namespace.name`` the name of the Juju model.
 - ``service.name`` the application name, like ``user_db``.
 - ``service.instance.id`` the unit number, like ``0``.
 - ``service.charm`` the charm class name, like ``DbCharm``.
 
+The `Juju topology attributes <https://discourse.charmhub.io/t/juju-topology-labels/8874>`_ are also set:
+
+- ``charm_type`` the charm class name, like ``DbCharm``.
+- ``juju_model`` the name of the Juju model.
+- ``juju_model_uuid`` the UUID of the Juju model.
+- ``juju_application`` the application name, like ``user_db``.
+- ``juju_unit`` the unit name, like ``user_db/0``.
+
+Note that ``charm_name`` is not available, because tracing subsystem is initialised
+before the charm metadata is loaded.
 
 Security considerations
 -----------------------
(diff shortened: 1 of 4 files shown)


docs/reference/ops-tracing.rst
```diff
@@ -8,12 +8,24 @@
 Open Telemetry resource attributes
 ----------------------------------
 
+The following well-known and additional resource attributes are set:
```
tonyandrewmeyer, line 11: Is "well-known" here meaning it's like a defacto standard, almost everyone sets these? Or is this like the well-known URI where it's actually part of a strict definition somewhere?

If there's a link that can be provided that explains that (some of) these are typically set, like you have for the Juju topology ones, that would be really handy I think.

Right now, it seems like there's considerable overlap between the `service.` ones and the Juju topology ones. I'm guessing that there's a valid reason for that -- basically that the `service.` ones are expected by something, and that Canonical has standardised on the Juju topology ones -- but it'd be nice to be clear on the reason for the duplication/overlap, which I think a link would provide.

dimaqq: Well-known as in the attribute keys are suggested/defined upstream, by OpenTelemetry.

james-garner-canonical (approved):
Looks like a good and straightforward change. If we go with loading the charm name from metadata early I guess that should be a follow-up PR?

Echoing Tony in that a few words (or a link) on where the `service.` names come from and explaining the overlap with the Juju ones would be a nice addition to the reference doc.

docs/reference/ops-tracing.rst
```diff
@@ -8,12 +8,24 @@
...
 
+The `Juju topology attributes <https://discourse.charmhub.io/t/juju-topology-labels/8874>`_ are also set:
+
+- ``charm_type`` the charm class name, like ``DbCharm``.
+- ``juju_model`` the name of the Juju model.
+- ``juju_model_uuid`` the UUID of the Juju model.
+- ``juju_application`` the application name, like ``user_db``.
+- ``juju_unit`` the unit name, like ``user_db/0``.
+
+Note that ``charm_name`` is not available, because tracing subsystem is initialised
```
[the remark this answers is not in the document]
dimaqq: done

docs/reference/ops-tracing.rst
dimaqq: I've update the text, PTAL.

dimaqq:
@simskij I've added the charm name too now.

dimaqq:
P.S. I've removed `service.charm` as it's redundant, non-standard, and as far as I can tell none uses it yet.

PR #1746, opened 2025-05-22
docs: configurations -> configuration

benhoyt:
This PR changes "Manage configurations" to "Manage configuration", as the latter sounds much better to me. And we always call it "charm configuration" (singular) or "charm config", not "charm configs".

### DIFF
--- a/docs/explanation/testing.md
+++ b/docs/explanation/testing.md
@@ -98,7 +98,7 @@ Integration tests typically take significantly longer to run than unit tests.
 * Packing and deploying the charm
 * Charm actions
 * Charm relations
-* Charm configurations
+* Charm configuration
 * That the workload is up and running, and responsive
 * Upgrade sequence
   * Regression test: upgrade stable/candidate/beta/edge from charmhub with the locally-built charm.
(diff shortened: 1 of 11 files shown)


dwilding (approved):

dwilding:
Thanks @benhoyt! I pushed a commit to change a few more instances of "configurations". I also pushed a commit to revert the change to the Charmcraft link at the top of [Manage configuration](https://ops--1746.org.readthedocs.build/en/1746/howto/manage-configuration.html), because that page is still called "Manage configurations". I'll open a PR to see whether they can change their page too.

I notice that in Juju, the example of how to do an intersphinx link uses exactly the target we're changing! See [conf.py](https://github.com/juju/juju/blob/f867c8b18cb4f50613a41535279c8a3969e8f439/docs/conf.py#L220). @tmihoc you may want to change `manage-configurations` to `manage-configuration` there.

tonyandrewmeyer (approved):
Seems reasonable to me. It would be nice if this was consistent across the ecosystem as David mentions, but everyone following this change seems a good direction.

PR #1748, opened 2025-05-22
ci: remove explicit scopes and update HACKING.md

### LINKED ISSUE #1727: Remove use of PR title scopes
Decision at sprint: Ops is too small a repo and no automation uses them, so let’s not use scopes. Validation should error if you use one (and we should remove them from our docs).

benhoyt:
Per discussion at the sprint, we consider Ops too small a repo to use scopes, and no automation uses them, so we decided not not use scopes, and have validation error if you use one.

Fixes #1727

### DIFF
--- a/.github/workflows/validate-pr-title.yaml
+++ b/.github/workflows/validate-pr-title.yaml
@@ -32,4 +32,4 @@ jobs:
             refactor
             revert
             test
-          scopes: ""
+          disallowScopes: ".*"
--- a/.github/workflows/validate-pr-title.yaml
+++ b/.github/workflows/validate-pr-title.yaml
@@ -32,8 +32,4 @@ jobs:
             refactor
             revert
             test
-          scopes: |
-            harness
-            testing
-            tracing
-          requireScope: false
+          scopes: ""
--- a/HACKING.md
+++ b/HACKING.md
@@ -183,25 +183,18 @@ Pull requests should have a short title that follows the
 * revert
 * test
 
-At present, we only add a scope in these cases:
-
-* If the PR is limited to optional features, use one of these scopes to clarify context:
-  * `(harness)` for changes in `ops/_private/harness.py`
-  * `(testing)` for changes in `testing/`
-  * `(tracing)` for changes in `tracing/`
-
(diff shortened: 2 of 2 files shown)


.github/workflows/validate-pr-title.yaml
```diff
@@ -32,8 +32,4 @@ jobs:
             refactor
             revert
             test
-          scopes: |
-            harness
-            testing
-            tracing
-          requireScope: false
+          scopes: ""
```
benhoyt, line 35: This works. I've added a scope "(foo)" and CI fails: https://github.com/canonical/operator/actions/runs/15196060716/job/42740323103?pr=1748

```
Error: Unknown scope "foo" found in pull request title
```

benhoyt:
I tried to test this, but CI on the PR seems to run against the workflow on main (I tried scope "foo", and it still said "Scope must match one of: harness, testing, tracing"), so it has to be merged first. I'll test it after merge, unless someone can think of a simple way to test it here.

tonyandrewmeyer (approved):
Merging into a fork seems like it would work for testing, but testing after merge seems fine.

benhoyt:
Thanks, good idea. I've tested now on https://github.com/benhoyt/operator/pull/8, and `disallowScopes: ".*"` seems to work (but `scopes: ""` didn't).

HACKING.md
```diff
@@ -183,26 +183,19 @@ Pull requests should have a short title that follows the
...
-Use of these scopes makes it clear when a change is limited to test tooling or optional features, rather than core runtime behavior.
-
-For example:
+Some examples:
 
 * feat: add the ability to observe change-updated events
 * fix!: correct the type hinting for config data
-* docs(harness): clarify the types of exceptions that Harness.add_user_secret may raise
-* ci(testing): adjust the workflow that publishes ops-scenario
+* docs: clarify the types of exceptions that Harness.add_user_secret may raise
```
dwilding, line 190: Shall we take this opportunity to remove Harness from the example? (This is the only place that Harness is mentioned in HACKING.md.) To semi randomly pick a similar commit message, how about this:

```suggestion
* docs: clarify how to use mounts in ops.testing.Container
```

dwilding (approved):

PR #1754, opened 2025-05-28
fix: correctly remove suffix from event string in ops.testing

james-garner-canonical:
In `ops.testing`, `scenario.state` contains logic that removes a suffix string from the event name. With the previous `rsplit` based implementation, suffix removal would fail if the suffix occurred multiple times in the string, since it would split the string into more than two parts.

For example, this would lead to failure when running `ctx.on.action('my-action')`, as the string `my_action_action` would be split into `'my', '', ''`, and the code expected the result of splitting to always be two items (the prefix, and an empty string).

This PR sets `maxsplit=1` for `rsplit` to handle this case.

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1851,8 +1851,7 @@ class _EventPath(str):
         )
         # TODO: when we drop Python 3.8, we can change the whole if-else below to
         # instance.prefix = string.removesuffix(suffix)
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
Diff simplified, thanks @dimaqq

benhoyt (approved):

PR #1756, opened 2025-05-29
feat: support for config schema as Python classes

tonyandrewmeyer:
A new method is added to `ops.ActionEvent`: `load_params`, which is the custom class equivalent of the `.params` property. The charmer passes a class to this method and gets back an instance of the class with the Juju action parameters loaded into it.

Two error-handling modes are provided for `load_params`. In the default, "raise", any exceptions raised creating the class are re-raised and must be handled by the charm (or left for the charm to go to error state). The alternative is "fail", where `load_params` will catch any `ValueError` (or subclass) and exit the charm with a zero exit status after setting an appropriate failure message. This mode should be suitable for the majority of actions.

* Pydantic dataclasses
* Pydantic 2.x BaseModel classes (we'll use these in our documentation, as they're in common use).

This PR includes reference docs ([Preview]()) and updates the configuration how-to ([Preview]()). A follow-up PR is planned that would make use of this functionality in the zero-to-hero tutorial - it will be simplest to do that after #1734 and any follow-up PRs are complete, to avoid conflicts.

More details are available in the [spec](https://docs.google.com/document/d/1AiNvr1xfh078ieRczCCc9HSx4VW-dDJdF2p9FS2E_do/edit?usp=sharing).

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -25,7 +25,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     ClassVar,
-    Dict,
+    Generator,
     List,
     Literal,
     Mapping,
@@ -33,6 +33,7 @@ from typing import (
     Optional,
     TextIO,
     TypedDict,
+    TypeVar,
     cast,
 )
 
@@ -110,6 +111,9 @@ class HookEvent(EventBase):
     - Storage events
     - Metric events
     """
+
+
+_ActionType = TypeVar('_ActionType')
 
 
 class ActionEvent(EventBase):
@@ -226,6 +230,59 @@ class ActionEvent(EventBase):
             message: Optional message to record why it has failed.
         """
         self.framework.model._backend.action_fail(message)
+
+    def load_params(
+        self,
+        cls: type[_ActionType],
+        *args: Any,
+        errors: Literal['raise', 'fail'] = 'raise',
+        **kwargs: Any,
+    ) -> _ActionType:
+        """Load the action parameters into an instance of an action class.
+
+        The raw Juju action parameters are passed to the action class's
+        ``__init__`` method as keyword arguments, with the following changes:
+        * dashes in names are converted to underscores.
+
+        Any additional positional or keyword arguments will be passed through to
+        the action class ``__init__``.
+
+        Args:
+            cls: A class that will accept the Juju parameters as keyword
+                arguments, and raise ``ValueError`` if validation fails.
+            errors: what to do if the parameters are invalid. If ``fail``, this
+                will set the action to failed with an appropriate message and
+                then immediately exit. If ``raise``, ``load_params`` will not
+                catch any exceptions, leaving the charm to handle errors.
+            args: positional arguments to pass through to the action class.
+            kwargs: keyword arguments to pass through to the action class.
+
+        Returns:
+            An instance of the action class with the provided parameter values.
+
+        Raises:
+            ValueError: if ``errors`` is set to ``raise`` and instantiating the
+                action class raises a ValueError.
+        """
+        try:
+            fields = set(_juju_fields(cls))
+        except ValueError:
+            fields = None
+        params: dict[str, Any] = kwargs.copy()
+        for key, value in self.params.items():
+            attr = key.replace('-', '_')
+            if fields is not None and attr not in fields:
+                continue
+            params[attr] = value
+        try:
+            return cls(*args, **params)
+        except ValueError as e:
+            if errors == 'raise':
+                raise
+            self.fail(f'Error in action parameters: {e}')
+            from ._main import _Abort
+
+            raise _Abort(0) from e
 
     def __repr__(self):
         return f'<{self.__class__.__name__} {self.id=} via {self.handle}>'
@@ -1435,6 +1492,28 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
         unit.status = model.StatusBase._get_highest_priority(unit._collected_statuses)
 
 
+def _juju_fields(cls: type[object]) -> Generator[str]:
+    """Iterates over all the field names to include when loading into a class.
+
+    Raises:
+        ValueError: if unable to determine which fields to include
+    """
+    # Dataclasses:
+    try:
+        fields = [field.name for field in dataclasses.fields(cls)]  # type: ignore
+    except TypeError:
+        pass
+    else:
+        yield from sorted(fields)
+        return
+    # Pydantic models:
+    if hasattr(cls, 'model_fields'):
+        yield from sorted(cls.model_fields)  # type: ignore
+        return
+    # It's not clear, so give up.
+    raise ValueError('Unable to find field list')
+
+
 class CharmMeta:
     """Object containing the metadata for the charm.
 
@@ -1697,12 +1776,12 @@ class CharmMeta:
         meta = yaml.safe_load(metadata)
         raw_actions = {}
         if actions is not None:
-            raw_actions = cast('Optional[Dict[str, Any]]', yaml.safe_load(actions))
+            raw_actions = cast('dict[str, Any] | None', yaml.safe_load(actions))
             if raw_actions is None:
                 raw_actions = {}
         raw_config = {}
         if config is not None:
-            raw_config = cast('Optional[Dict[str, Any]]', yaml.safe_load(config))
+            raw_config = cast('dict[str, Any] | None', yaml.safe_load(config))
             if raw_config is None:
                 raw_config = {}
         return cls(meta, raw_actions, raw_config)
--- a/test/test_charm.py
+++ b/test/test_charm.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 import functools
 import pathlib
 import tempfile
@@ -23,8 +24,16 @@ from pathlib import Path
 import pytest
 import yaml
 
+try:
+    import pydantic
+    import pydantic.dataclasses
+except ImportError:
+    pydantic = None
+
 import ops
 import ops.charm
+from ops import testing
+from ops._main import _Abort
 from ops.model import ModelError, StatusName
 
 from .test_helpers import FakeScript, create_framework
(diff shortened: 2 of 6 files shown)


tonyandrewmeyer:
@dwilding this one also has some doc adjustments, if you'd like to review those.

ops/charm.py
```diff
@@ -1697,12 +1776,12 @@ def from_yaml(
...
         raw_actions = {}
         if actions is not None:
-            raw_actions = cast('Optional[Dict[str, Any]]', yaml.safe_load(actions))
+            raw_actions = cast('dict[str, Any] | None', yaml.safe_load(actions))
             if raw_actions is None:
                 raw_actions = {}
         raw_config = {}
         if config is not None:
-            raw_config = cast('Optional[Dict[str, Any]]', yaml.safe_load(config))
+            raw_config = cast('dict[str, Any] | None', yaml.safe_load(config))
```
tonyandrewmeyer, line 1784: Somehow missed in an earlier PR, but linting complains here.

test/test_charm.py
```diff
@@ -1253,3 +1263,447 @@ def test_meta_charm_user_default():
...
+            type: integer
+            default: 42
+"""
+    harness = testing.Harness(charm_class, actions=schema)
+    request.addfinalizer(harness.cleanup)
+    harness.begin()
+    # Harness raises an exception when fail() is called, so within Harness, we
+    # don't actually get the _Abort raised.
+    with pytest.raises(_Abort):
+        harness.run_action('my-action', params={'my-str': 'foo', 'my-int': -1})
```
james-garner-canonical, line 1491: This is confusing for me.

1. `ActionEvent.load_params` will call `self.fail`, then `raise _Abort(0)`
2. After the event handler returns, `Harness.run_action` raises `ActionFailed` if `failure_message` is set
3. Harness doesn't catch `_Abort`
4. `_Abort` is raised before the event handler returns (so it never returns)

So it seems like the code here is correctly catching `_Abort` (and perhaps could `assert` that `exit_code == 0`), and the comment is incorrect (and should be removed).

Is that right?

james-garner-canonical (commented):
Looks good. A couple of questions on the implementation, importantly:
- should `load_params` just require that the passed `cls` accepts all the params?
- if so, presumably this PR would lose `_juju_fields`
- if not, question: will this `_juju_fields` be the same function as #1741's `_juju_option_names` once PRs are merged?

Additionally, I've made some suggestions in the tests regarding things I found confusing when reading them.

test/test_charm.py
```diff
@@ -1253,3 +1263,447 @@ def test_meta_charm_user_default():
...
+        my-int:
+            type: integer
+            default: 42
+        my-float:
+            type: float
+            default: 3.14
+        my-list:
+            type: list
+"""
+    harness = testing.Harness(charm_class, actions=schema)
```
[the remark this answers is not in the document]
tonyandrewmeyer: > I know that @tonyandrewmeyer and @benhoyt settled on using Harness in these tests, I believe for consistency with the existing tests in these suites.

Yes, I originally wrote them using Harness because that's what all the ops tests use (if any framework). Then there was a request to rewrite them using Scenario in the original PR. Then when I split the PR and they now landed in test_charm.py I felt that consistency made sense so proposed leaving them as Harness, which Ben agreed to.

> I wonder if our policy going forward should be to use Scenario for new tests (unless explicitly testing Harness itself), to facilitate gradually migrating off Harness.

I did get a consensus that we are *open* to writing new tests using Scenario rather than Harness. If you'd like to go beyond "open" towards something stronger, then I think daily would be the place to bring it up and that we'd add it to the style guide or contributing doc if we do end up going that way.

ops/charm.py
```diff
@@ -1435,6 +1492,28 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
         unit.status = model.StatusBase._get_highest_priority(unit._collected_statuses)
 
 
+def _juju_fields(cls: type[object]) -> Generator[str]:
```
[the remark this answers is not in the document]
tonyandrewmeyer: No. `_juju_option_names` seemed inappropriate (as was the docstring) when used here, so I renamed it to be more generic so that it can work with all the PRs.

My intention was to resolve this when I merge main into this branch after the other PR is merged into main, along with any other conflicts. That seemed simpler than going back and changing it in the other PR in advance of something that's only currently proposed. If you'd rather I can make the other PR use this name as well.

test/test_charm.py
```diff
@@ -1253,3 +1263,447 @@ def test_meta_charm_user_default():
...
+            params = event.load_params(SnapshotAction, errors='raise')
+            event.log(f'Generating snapshot into {params.filename}')
+            success = self.do_snapshot(
+                filename=params.filename,
+                kind=params.compression.kind,
+                quality=params.compression.quality,
+            )
+            if not success:
+                event.fail('Failed to generate snapshot.')
+                return
```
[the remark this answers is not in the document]
tonyandrewmeyer: I was keeping this consistent with the example in the how-to, because I wanted to be confident that worked, since we don't currently test any of the code in the how-tos.

test/test_charm.py
```diff
@@ -14,6 +14,8 @@
```
james-garner-canonical, line 14: One more thing: these tests seem lengthy enough that it might be cleaner to pop them in a new file (e.g. `test/test_typed_action_params.py`). Just an idea. What do you think?

test/test_charm.py
```diff
@@ -1253,3 +1263,447 @@ def test_meta_charm_user_default():
...
+def test_action_extra_args(request: pytest.FixtureRequest):
+    @dataclasses.dataclass
+    class Action:
+        a: int
+        b: float
+        c: str
+
+        @classmethod
+        def _param_names(cls):
+            yield 'b'
```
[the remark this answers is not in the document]
tonyandrewmeyer: Ah, that's leftover from the old implementation that I missed removing, sorry.

test/test_charm.py
```diff
@@ -1253,3 +1263,447 @@ def test_meta_charm_user_default():
...
+        harness.run_action('my-action', params={'my-str': 'foo', 'my-int': -1})
+
+
+def test_action_custom_naming_pattern(request: pytest.FixtureRequest):
+    @dataclasses.dataclass(frozen=True)
+    class Act1:
+        foo_bar: int = 42
+        other: str = 'baz'
+
+        def __init__(self, fooBar: int = 42, other: str = 'baz'):  # noqa: N803
```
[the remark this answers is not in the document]
tonyandrewmeyer: I believe it's a bad practice to have defaults in the YAML and not have them in the Python class. I feel strongly that the YAML and Python should match.

If I remove the defaults from the Python class and leave them in the YAML, then I have to pass the default values in to the Harness call - this is what would happen with Juju, but I don't think it makes the test any simpler or cleaner.

The class attribute defaults could be dropped without changing the rest of the code, but then the `__init__` and the YAML and the class are all slightly unaligned, other than the deliberate misalignment of a different naming scheme.

So I would strongly prefer to keep this as-is.

test/test_charm.py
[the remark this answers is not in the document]
james-garner-canonical: That makes sense. I'd be in favour of either treating this like just another test and stripping out the irrelevant code (my initial suggestion), or adding a note somewhere (maybe the test docstring + a comment in the how-to if that's possible) that this is a test of code from that specific how-to.

ops/charm.py
```diff
@@ -227,6 +231,59 @@ def fail(self, message: str = ''):
...
+        cls: type[_ActionType],
+        *args: Any,
+        errors: Literal['raise', 'fail'] = 'raise',
+        **kwargs: Any,
+    ) -> _ActionType:
+        """Load the action parameters into an instance of an action class.
+
+        The raw Juju action parameters are passed to the action class's
+        ``__init__`` method as keyword arguments, with the following changes:
+        * dashes in names are converted to underscores.
```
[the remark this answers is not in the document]
tonyandrewmeyer: Fair call; I was falling too heavily on the "make this like the config PR" :smile:.

test/test_charm.py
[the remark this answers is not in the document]
james-garner-canonical: Thanks. Interesting how different our instincts are here. If the schema is the source of truth, I'd avoid encoding the defaults anywhere on the Python side if possible, especially if they're always expected to be overridden by the explicitly passed Juju defaults at runtime.

Question: when the schema generation is added, would you tweak this test to generate the schema yaml string from the Python class?

test/test_charm.py
```diff
@@ -1253,3 +1263,442 @@ def test_meta_charm_user_default():
...
+
+    harness = testing.Harness(RaiseCharm, actions=action_yaml)
+    request.addfinalizer(harness.cleanup)
+    harness.begin()
+    with pytest.raises(ValueError):
+        harness.run_action('my-action', params={'my-str': 'foo', 'my-int': -1})
+
+
+def test_action_custom_naming_pattern(request: pytest.FixtureRequest):
+    class Act1:  # noqa: B903
```
james-garner-canonical, line 1522: If you'd prefer a non-frozen `dataclass` over this `noqa`, I'd be on board with that, up to you.

test/test_charm.py
tonyandrewmeyer: If that means rewriting the tests in Scenario, I'd prefer to not spend my time doing that right now. If that would mean doing the same in the config PR, I'd also prefer to leave that in an approved state rather than circling back to the start of all this one more time.

We have a very 1:1 module:test-module pattern at the moment, other than main. I think I'd rather we discussed breaking that pattern as a team and decide how we'll change (if we do want to) than just start changing it here.

james-garner-canonical: Definitely not proposing rewriting anything. I'll add an item to daily to discuss breaking the module:test-module pattern. Even if we do decide to break that pattern going forward, I don't think it should mean you need to break the pattern in this PR if you'd prefer not to.

ops/charm.py
```diff
@@ -227,6 +231,59 @@ def fail(self, message: str = ''):
...
+
+        Returns:
+            An instance of the action class with the provided parameter values.
+
+        Raises:
+            ValueError: if ``errors`` is set to ``raise`` and instantiating the
+                action class raises a ValueError.
+        """
+        try:
+            fields = set(_juju_fields(cls))
```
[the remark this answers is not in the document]
tonyandrewmeyer: > Being able to accept only a subset of the options makes sense for config (particularly the 12 factor app case), but does it make sense for actions? Maybe we should just pass all the Juju params for actions, requiring that the class passed to `ActionEvent.load_params` accepts all the params defined for that action.

I do think that almost all of the time you'll want to take them all, and also that almost all the time charms aren't using `additionalProperties=true` so they won't need to accept additional keyword arguments.

However, a strong focus throughout this whole project has been keeping the use of schema classes consistent across the three use-cases: config, actions, relation data. I do feel like that's still worthwhile.

Given that `_juju_fields` (under whatever name) will exist in ops, if it's not used here, that's only removing the try/except and the later `if fields`. I don't feel that makes the implementation significantly simpler, and it also makes the implementation less consistent with `load_config` (and later, `Relation.load`).

In terms of usability, if you're wanting all the params, you're presumably defining them all in your dataclass or model fields, so you get the behaviour you expect in either case. If you don't define them all in your class then it seems to me that you're expecting to not get them all passed in (rather than expecting an error), so you get the behaviour you want as well.

So given that we gain consistency in charmer experience and consistency in implementation, and we don't gain anything by not doing this, I'm not sure what the argument against is. Could you please elaborate on what we'd gain by dropping this behaviour?

james-garner-canonical (approved):
Looks good, thanks for addressing all my concerns. Happy with only passing the parameters defined in the config class for consistency with the typed config and relation data.

test/test_charm.py
```diff
@@ -1253,3 +1263,442 @@ def test_meta_charm_user_default():
...
+        foo_bar: int = 42
+        other: str = 'baz'
```
james-garner-canonical, line 1524: I'm actually now surprised that this worked properly when it was a `dataclass`, because shouldn't `_juju_fields` have resulted in filtering out the `fooBar` argument? In fact, I think it wasn't working properly, `_juju_fields` *was* resulting in `fooBar` being filtered out, but the default value in the `__init__` was being used, which seems extremely bad given what the test appeared to be expressing.

Recommendation: do this test with both default and non-default params. e.g. parametrize on the second argument of `run_action`, with values `{}` and `{'fooBar': 43}`.

Recommendation: do this test with all 4 types.

tonyandrewmeyer: Nice catch, thanks. I indeed wrongly converted the test from the old system where you could just customise what values come through.

Should be fine now:

```
[gw4] [ 12%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_PydanticDataclassesAlias-fb_value0] 
[gw7] [ 25%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_PydanticBaseModelAlias-fb_value1] 
[gw6] [ 37%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_PydanticDataclassesAlias-fb_value1] 
[gw2] [ 50%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_DataclassesAlias-fb_value0] 
[gw1] [ 62%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_Alias-fb_value1] 
[gw0] [ 75%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_Alias-fb_value0] 
[gw3] [ 87%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_DataclassesAlias-fb_value1] 
[gw5] [100%] PASSED test/test_charm.py::test_action_custom_naming_pattern[_PydanticBaseModelAlias-fb_value0] 
```

test/test_charm.py
```diff
@@ -1253,3 +1263,469 @@ def test_meta_charm_user_default():
...
+
+    class _PydanticBaseModelAlias(pydantic.BaseModel):
+        foo_bar: int = pydantic.Field(42, alias='fooBar')
+        other: str = pydantic.Field('baz')
+
+    _alternate_name_configs.extend([_PydanticDataclassesAlias, _PydanticBaseModelAlias])
+
+
+@pytest.mark.parametrize('fb_value', [{}, {'fooBar': 24}])
+@pytest.mark.parametrize('action_config', _alternate_name_configs)
```
james-garner-canonical: Sorry to nitpick names, but how about `action_params`, `action_class` and `_alias_action_classes`.

ops/charm.py
```diff
@@ -1435,6 +1494,45 @@ def _evaluate_status(charm: CharmBase):  # pyright: ignore[reportUnusedFunction]
         unit.status = model.StatusBase._get_highest_priority(unit._collected_statuses)
 
 
+def _juju_fields(cls: type[object]) -> dict[str, str]:
+    """Iterates over all the field names to include when loading into a class.
+
+    Any Juju names that are not in the returned dictionary should not be passed to
+    the class. Names that are in the dictionary are mapped to the argument name;
+    in most cases this is the same string, but for aliases will differ.
```
james-garner-canonical, line 1609: Somehow I end up confused as to the directionality of the mapping, can we clarify this here? It's Juju parameter name -> class attribute, right?

test/test_charm.py
```diff
@@ -1253,3 +1263,469 @@ def test_meta_charm_user_default():
...
+        harness.run_action('my-action', params={'my-str': 'foo', 'my-int': -1})
+
+
+class _Alias:  # noqa: B903
+    foo_bar: int = 42
+    other: str = 'baz'
+
+    def __init__(self, fooBar: int, other: str):  # noqa: N803
+        self.foo_bar = fooBar
+        self.other = other
```
james-garner-canonical: If we're imagining that `_Alias` is the source of truth for the Juju metadata here, then shouldn't the attribute be `fooBar: int = 42`?

If we're imagining that `_Alias` is just the consumer of the data from Juju, then wouldn't it be more natural to drop the class attributes+annotations altogether?

test/test_charm.py
```diff
@@ -1253,3 +1263,469 @@ def test_meta_charm_user_default():
...
+
+
+@dataclasses.dataclass(frozen=True)
+class _DataclassesAlias:
+    foo_bar: int = dataclasses.field(default=42, metadata={'alias': 'fooBar'})
+    other: str = 'baz'
```
james-garner-canonical, line 1898: I'm not 100% sold on special casing metadata values for this case, though I'm not totally opposed to it. I do find the idea of following the `pydantic.dataclass` behaviour reasonable, like how we might want to follow the behaviour of a `dataclasses.dataclass` when handling a regular class.

But we *could* assume that the `dataclasses.dataclass` case here should be handled by the user, like we do for the vanilla class above. For example:

```suggestion
@dataclasses.dataclass
class _DataclassesAlias:
    fooBar: dataclasses.InitVar[int] = dataclasses.field(default=42)
    other: str = 'baz'
    def __post_init__(self, fooBar: int):
        self.foo_bar = fooBar
```

test/test_charm.py
tonyandrewmeyer: > If we're imagining that `_Alias` is the source of truth for the Juju metadata here, then shouldn't the attribute be `fooBar: int = 42`?

If the attribute is `fooBar` then we might as well just drop this class from the test, because it's not aliasing anything. Now that we don't have the base class, it's not certain how the classes override the generated schema, which they definitely need to do in some cases (especially with config to handle types that Juju doesn't understand, unless those are always just "string"). If we still allow the class to define a `to_juju_schema` then it could be customised in there.

> If we're imagining that `_Alias` is just the consumer of the data from Juju, then wouldn't it be more natural to drop the class attributes+annotations altogether?

I guess so. It feels like a more obvious place to document the API of the class, particularly when you add in attribute pseudo-docstrings, but that could all be in `__init__`.

test/test_charm.py
tonyandrewmeyer: I somewhat agree that it'd be better to only provide the built-in behaviour. The class would need to look like this I believe (I've stripped out the defaults):

```python
class _DataclassesAlias:
    fooBar: dataclasses.InitVar[int]
    foo_bar: int = dataclasses.field(init=False)
    other: str

    def __post_init__(self, fooBar: int):
        object.__setattr__(self, 'foo_bar', fooBar)
```

I'm not a huge fan of the duplication there and it's definitely uglier, but it is more correctly plain standard library Python.

However, that means that I have to add something like this into the `_juju_fields` method:

```python
        # We also need to allow any InitVars, because those are allowed.
        for name, field in get_type_hints(cls).items():
            if 'InitVar' in str(field):  # typing.get_origin doesn't know dataclasses.InitVar
                yield name
```

Otherwise, when we get a `fooBar` from Juju it's not in the fields, so we don't pass it in. I would be somewhat ok with this if `typing.get_origin(some_initvar_field)` returned `InitVar` but it doesn't know it so returns `None`, so we have to fall back to some ugly stringifying.

So, while I was on board enough to make this work, I would much prefer to go back to having `metadata={'alias': x}` work with standard library dataclasses. Thoughts?

james-garner-canonical: I'm on board with going back to supporting `'alias'` in `metadata` for `dataclasses.dataclass`, thanks for investigating this.

PR #1766, opened 2025-05-30
ci: persist credentials for update-charm-pins workflow

benhoyt:
This workflow is failing due to a permissions error on the `git push`. Remove the `persist-credentials: false` line on checkout.


tonyandrewmeyer (commented):
Thanks for this - and apologies for missing it when doing the earlier changes.

The unit test failure seems unrelated. I assume the docs failure just needs the latest changes from main.

If I run zizmor against this branch I get one medium alert. I think we need to add an ignore marker to indicate that we're ok with the credentials being persisted.

Have you tested this? I assume the branch is in canonical rather than your fork to make that easier, but I don't see it in the actions list.

benhoyt:
Given @james-garner-canonical is fixing Zizmor separately, I'm just going to merge this PR.

PR #1769, opened 2025-05-30
ci: configure and conform with zizmor

james-garner-canonical:
This PR adds `.github/zizmor.yml` to configure `zizmor` to allow actions from the `actions`, `github`, and `pypa` orgs to be pinned by tags (e.g. `v1`) the same way that actions from the `actions` org are permitted by default. All other actions are required to be pinned by hash by default.

The zizmor version is also pinned consistently for CI and pre-commit.

Credential persistence is set explicitly where required, to conform with zizmor.

In parallel, the repo settings have been updated to require zizmor to pass on PRs before merging.

Zizmor docs: https://docs.zizmor.sh/audits/#rulesunpinned-usesconfigpolicies

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -20,7 +20,7 @@ jobs:
           persist-credentials: false
 
       - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb  # v6.1.0
 
       - name: Run zizmor
         run: uvx zizmor@v1.9.0 --format=sarif . > results.sarif


james-garner-canonical:
Reckon it's worth encoding the defaults in the `zizmor.yml` for easy reference and to avoid them changing under us?
That would be:
```yaml
rules:
  unpinned-uses:
    config:
      policies:
        "actions/*": ref-pin
        "github/*": ref-pin
        "*": hash-pin
```

.github/workflows/update-charm-tests.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
+        with:  # zizmor: ignore[artipacked]
           token: ${{ secrets.UPDATE_CHARM_PINS_ACCESS_TOKEN }}
```
james-garner-canonical, line 18: Isn't the solution here to use `persist-credentials: false` instead of the ignore? Or do we need it in other steps? I guess we can repeat the `token:` in the steps that need it? Or just use `persist-credentials: true`?

https://docs.zizmor.sh/audits/#remediation

benhoyt (approved):
Looks good to me. Do you mind pushing a zizmor failure so we can test that and see what it looks like?

benhoyt:
Looks good, thanks!

.github/workflows/zizmor.yaml
```diff
@@ -23,12 +23,6 @@ jobs:
...
-        run: uvx zizmor --format sarif . > results.sarif
+        run: uvx zizmor@v1.8.0 . --format=github
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Upload SARIF file
-        uses: github/codeql-action/upload-sarif@v3
-        with:
-          sarif_file: results.sarif
-          category: zizmor
```
tonyandrewmeyer, line 34: What's the motivation for turning off reporting results? I'd much rather we kept that integration.

james-garner-canonical: The job doesn't fail when running in `zizmor` mode. (See [important box at the bottom of this section](https://docs.zizmor.sh/usage/#use-in-github-actions).) That's why we didn't notice we were having all kinds of failures. You can have it block PRs still somehow with [github rulesets](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/about-code-scanning-alerts#pull-request-check-failures-for-code-scanning-alerts), but Ben and I thought that this would be a lighter weight approach.

Do we use the upload for anything? Another option would be to do:
1. run in sarif mode
2. upload
3. run again in regular mode

benhoyt: @tonyandrewmeyer Where did those "reports" go?

tonyandrewmeyer: > The job doesn't fail when running in `zizmor` mode. (See [important box at the bottom of this section](https://docs.zizmor.sh/usage/#use-in-github-actions).) That's why we didn't notice we were having all kinds of failures.

Ah, I was confused about why there were failures showing in the security results but not against any PRs (but hadn't had time to figure out why). That explains it.

> You can have it block PRs still somehow with [github rulesets](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/about-code-scanning-alerts#pull-request-check-failures-for-code-scanning-alerts), but Ben and I thought that this would be a lighter weight approach.

I'd rather we turned that on. It's a pretty straightforward adjustment to the branch protection rules, and then we're using the tool as intended, and are also setup for any other security scanning tools that come along (for example, we currently do security scanning with internal-only Canonical tooling, but hopefully that moves to something that can be used publicly eventually).

tonyandrewmeyer (changes requested):
I'd like to keep the security reporting. If we're going to drop that, let's have a team discussion first at least.

.github/workflows/zizmor.yaml
james-garner-canonical: Thanks for bumping this. Will the branch protection rule status show prior to trying to merge, like CI failures would? If so, then I'm on board with that. If not, I'd prefer the generate report, upload, then produce errors approach. Can discuss in daily.

tonyandrewmeyer: > Will the branch protection rule status show prior to trying to merge, like CI failures would?

Yes:

image

Followed by:

image

(The PR is in my fork if you want to see it live).

james-garner-canonical: Excellent, thanks! In that case I'm fully on board with this approach, no need to discuss in daily. Will fix.

tonyandrewmeyer (approved):
Thanks!

PR #1772, opened 2025-05-30
docs: update HACKING.md regarding contributing and tests

james-garner-canonical:
This PR updates `HACKING.md` to document that we will prefer scenario tests (unless there's good reason not to), and to document that large collections of new, related tests can go in separate modules.

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -20,7 +20,7 @@ jobs:
           persist-credentials: false
 
       - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb  # v6.1.0
 
       - name: Run zizmor
         run: uvx zizmor --format sarif . > results.sarif


HACKING.md
```diff
@@ -168,6 +168,14 @@ For coding style, we follow [PEP 8](https://peps.python.org/pep-0008/) as well a
 as the [ops library reference](https://ops.readthedocs.io/en/latest/reference/index.html)
 is automatically generated from Python docstrings.
 
+Changes should include tests. Where reasonable, prefer to write 'Scenario' tests via [ops.testing](https://ops.readthedocs.io/en/latest/reference/ops-testing.html) over legacy [ops.testing.Harness](https://ops.readthedocs.io/en/latest/reference/ops-testing-harness.html) tests. Note that these tests will require the `ops-scenario` package to be installed. Running tests with `tox -e unit` will take care of this automatically. These tests should be decorated so that they're skipped if `ops-scenario` isn't installed, like so:
```
benhoyt, line 171: Couple of tweaks. And I don't think we need to mention that ops-scenario is required, as it's required now, and it's taken care of automatically as you say.

```suggestion
Changes should include tests. Where reasonable, prefer to write 'Scenario' tests using [ops.testing](https://ops.readthedocs.io/en/latest/reference/ops-testing.html) instead of legacy [ops.testing.Harness](https://ops.readthedocs.io/en/latest/reference/ops-testing-harness.html) tests. These tests should be decorated so that they're skipped if `ops-scenario` isn't installed, like so:
```

benhoyt: Hmm, isn't this required already? Why not just require ops-scenario when running testing, and not bother with the skipif decorator (which will then be required on many new tests).

HACKING.md
```diff
@@ -168,6 +168,14 @@ For coding style, we follow [PEP 8](https://peps.python.org/pep-0008/) as well a
...
+```py
+@pytest.mark.skipif(
+    not hasattr(ops.testing, 'Context'), reason='requires optional ops[testing] install'
+)
+```
+Tests should go in the test module corresponding to the code. For example, a feature added in `ops/main` would go in `test/test_main`. However, when adding a large number of logically related tests, consider putting these in their own file, named accordingly. For example, if adding a feature `foo` in `ops/main`, the tests might go in `test/test_main_foo`.
```
benhoyt, line 177:
```suggestion
Tests should go in the test module corresponding to the code. For example, a feature added in `ops/main.py` would go in `test/test_main.py`. However, when adding a large number of logically related tests, consider putting these in their own file, named accordingly. For example, if adding a feature `foo` in `ops/main.py`, the tests might go in `test/test_main_foo.py`.
```

HACKING.md
tonyandrewmeyer: It's not required if you specifically run the tests in the top-level `test` folder. If you run both those tests and the ones in `testing/tests` (which the top-level `tox -e unit` does) then it is required (and tox takes care of that for you).

However, I agree that if we're going to start regularly adding tests that use Scenario with the expectation that eventually all the Harness use is replaced (outside test_testing) then a skip marker will be too much noise, and we just require `ops[testing]` to run any of the ops unit tests.

tonyandrewmeyer:
> In my mind, Scenario = "testing individual charms", so I'm confused about why we'd recommend Scenario for testing Ops itself.

Is Harness "testing individual charms" in your mind as well? If so, there's lots of examples of using Harness outside of test_testing that you can see (mostly in test_model).

> Is there an existing Scenario test for Ops that I could check out to help me understand what this means in practice?

No, because until recently we didn't accept that.

<details>
<summary>Bad example since I already typed it out :joy:</summary>

But here's an [example of a test that uses Harness](https://github.com/canonical/operator/blob/9b712fe27dd8d41788e22e6315dbad883c53ab1d/test/test_model.py#L231C5-L241C103), somewhat randomly picked:

```python
    def test_invalid_type_relation_data(self, harness: ops.testing.Harness[ops.CharmBase]):
        relation_id = harness.add_relation('db1', 'remoteapp1')
        harness.add_relation_unit(relation_id, 'remoteapp1/0')

        with pytest.raises(ops.RelationDataError):
            with harness._event_context('foo_event'):
                harness.update_relation_data(relation_id, 'remoteapp1/0', {42: 'remoteapp1-0'})  # type: ignore

        with pytest.raises(ops.RelationDataError):
            with harness._event_context('foo_event'):
                harness.update_relation_data(relation_id, 'remoteapp1/0', {'foo': 42})  # type: ignore```

And here's how I would write it using Scenario:

```python
    # TODO: Giving these nicer names would be good.
    @pytest.mark.parametrize('data', [
        {42: {'remoteapp1-0'}},
        {0: {'foo': 42}},
    ])
    def test_invalid_type_relation_data(self, data):
        ctx = testing.Context(ops.CharmBase, meta={'name': 'foo'})
        relation = testing.Relation('db1', interface='remoteapp1', remote_units_data=data)  # type: ignore
        state_in = testing.State(relations={relation})
        with pytest.raises(ops.RelationDataError):
            ctx.run(ctx.on.relation_changed(relation), state_in)
        with pytest.raises(ops.RelationDataError):
            ctx.run(ctx.on.relation_changed(relation), state_in)
```

And then I'd realise that I can't do that because Scenario has earlier checks that the relation data is valid :joy:.

</details>

So here's an [example test that uses Harness](https://github.com/canonical/operator/blob/9b712fe27dd8d41788e22e6315dbad883c53ab1d/test/test_model.py#L461):

```python
    def test_relation_data_access_peer_leader(self, harness: ops.testing.Harness[ops.CharmBase]):
        r_id = harness.add_relation('db2', 'myapp')
        harness.add_relation_unit(r_id, 'myapp/1')  # peer!
        harness.update_relation_data(r_id, 'myapp', {'foo': 'bar'})
        with harness._event_context('foo_event'):
            # leaders can read
            harness.set_leader(True)
            relation = harness.model.get_relation('db2')
            assert relation is not None and relation.app is not None
            assert relation.data[relation.app]['foo'] == 'bar'
```

And this could be done using Scenario instead:

```python
def test_relation_data_access_peer_leader(self):
    ctx = testing.Context(ops.CharmBase, meta={'name': 'foo'})
    relation = testing.PeerRelation('db2', peers_data={1: {'foo': 'bar'}})
    state_in = testing.State(leader=True, relations={relation})
    # Leaders can read:
    with ctx.run(ctx.on.update_status(), state_in) as mgr:
        relation = mgr.charm.model.get_relation('db2')
        assert relation is not None and relation.app is not None
        assert relation.data[relation.app]['foo'] == 'bar'
```

HACKING.md
```diff
@@ -168,6 +168,14 @@ For coding style, we follow [PEP 8](https://peps.python.org/pep-0008/) as well a
...
+
 For more advice about contributing documentation, see [Contributing documentation](#contributing-documentation).
```
dwilding, line 171: I think this sentence has become orphaned by accident. It was supposed to follow on from the part ending "is automatically generated from Python docstrings."

james-garner-canonical:
Ready for second review @tonyandrewmeyer @benhoyt

benhoyt (approved):

tonyandrewmeyer (approved):

PR #1778, opened 2025-06-01
fix: correctly load an empty Juju config options map

### LINKED ISSUE #1775: Recent versions of ops fail when there are no config options
In my charmcraft.yaml I have no config options:
```yaml
config:
  options:
```

As a result, the install hook fails (ops-2.22.0 and ops-2.21.1):
```
  File "/var/lib/juju/agents/unit-swfd-0/charm/venv/lib/python3.12/site-packages/ops/charm.py", line 1609, in __init__
    for name, config in config_raw_.get('options', {}).items():
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
```

There are no errors with ops-2.17.

tonyandrewmeyer:
There are many forms that "no Juju config options" can take, for example:

* No `config` section in charmcraft.yaml, and no config.yaml
* An empty `config` section in charmcraft yaml
* An empty config.yaml
* A `config` section in charmcraft yaml with an empty `options` subsection
* No `config` section in charmcraft.yaml, and an empty `options` map in config.yaml

For the cases where there *was* an `options` map, but it was empty, we were failing to load the meta. Add more extensive tests to cover the different cases, and fix loading in that case.

Fixes #1775

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -1541,7 +1541,7 @@ class CharmMeta:
     ):
         raw_: dict[str, Any] = raw or {}
         actions_raw_: dict[str, Any] = actions_raw or {}
-        config_raw_: dict[str, Any] = config_raw or {}
+        options_raw: dict[str, Any] = (config_raw or {}).get('options', {}) or {}
 
         # When running in production, this data is generally loaded from
         # metadata.yaml. However, when running tests, this data is
@@ -1606,7 +1606,7 @@ class CharmMeta:
         # This is predominately for backwards compatibility with Harness. In a
         # real Juju environment this shouldn't be possible, because charmcraft
         # validates the config when packing.
-        for name, config in config_raw_.get('options', {}).items():
+        for name, config in options_raw.items():
             if 'type' not in config:
                 raise RuntimeError(
                     f'Incorrectly formatted config in YAML, option {name} is '
@@ -1619,7 +1619,7 @@ class CharmMeta:
                 default=config.get('default'),
                 description=config.get('description'),
             )
-            for name, config in config_raw_.get('options', {}).items()
+            for name, config in options_raw.items()
         }
         self.containers = {
             name: ContainerMeta(name, container)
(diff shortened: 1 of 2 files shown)


dimaqq:
**note:**

> with an empty options subsection

➡️ an empty-looking (actually null) options subsection

ops/charm.py
```diff
@@ -1541,7 +1541,7 @@ def __init__(
     ):
         raw_: dict[str, Any] = raw or {}
         actions_raw_: dict[str, Any] = actions_raw or {}
-        config_raw_: dict[str, Any] = config_raw or {}
+        options_raw: dict[str, Any] = (config_raw or {}).get('options', {}) or {}
```
dimaqq: **nit:** perhaps the following is more accurate:

```suggestion
        options_raw: dict[str, Any] = (config_raw or {}).get('options') or {}
```

tonyandrewmeyer:
> **question:** does charmcraft accept all of these? I wouldn't want to see ops accepting a wider set than charmcraft.

Yes (I tested that they all pack and deploy and that `config-get` returns `{}`), with one exception. The `config.yaml` file is optional as is the `config` section in `charmcraft.yaml`. The `options` field is optional within the config (the docs say otherwise, but it works), and you can have a zero-item `options` (the docs don't say this explicitly, but it works).

The exception is an empty `config.yaml`. However, this was covered by the existing test and existing code, so I don't think we should remove it. The argument for removing it would be that it fails to pack so it's not very useful, but potentially it's used in tests (one ought to pass `None`, but passing `''` works), and it keeps things consistent with `charmcraft.yaml` where it does work.

ops/charm.py
james-garner-canonical: +1 on Dima's suggestion. I wouldn't mind a comment too, e.g.: `# config_raw['options'] might be None`

EDIT: or `# config_raw might be {'options': None}`

james-garner-canonical (approved):

dwilding (approved):

Issue #1780, opened 2025-06-03
Broken 'Manage your deployment environment' link to Juju docs from tutorials


sinapah:
### Issue

In the [`Set up your development environment`](https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/set-up-your-development-environment.html) section of the [`Write your first machine charm`](https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/index.html) tutorial, the link [`Juju | Manage your deployment environment > Set things up`](https://documentation.ubuntu.com/juju/3.6/howto/manage-your-deployment/manage-your-deployment-environment/#set-things-up) points to a non-existing page. 

### Suggested Fix

It seems like the correct link is [`Set up your deployment`](https://documentation.ubuntu.com/juju/3.6/howto/manage-your-deployment/#set-up-your-deployment). 

---
*Reported from: https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/set-up-your-development-environment.html*

tonyandrewmeyer:
It's odd that an external:refs link can break like this without the docs build failing. I wonder if the refs inventory was outdated?

tonyandrewmeyer:
Yes, looks like there was a [change to the Juju docs a few hours ago that would have broken this](https://github.com/juju/juju/commit/873061b0b0f57b87b9fb233935e36acdb81155a2).

tonyandrewmeyer:
Thanks for the report @sinapah!

PR #1781, opened 2025-06-03
fix: update moved Juju link.

tonyandrewmeyer:
The Sphinx reference for the "set up your deployment" doc section in Juju changed a few hours ago, which broke two links from our documentation. This PR updates the reference so that they work again.

* From the tutorial
* From the how to for writing integration tests

A bunch of trailing whitespace is also removed via pre-commit. I suggest ignoring whitespace when you review.

Fixes: #1780

### DIFF
--- a/docs/howto/write-integration-tests-for-a-charm.md
+++ b/docs/howto/write-integration-tests-for-a-charm.md
@@ -19,7 +19,7 @@ The instructions all use the Juju `python-libjuju` client, either through the `p
 
 In order to run integrations tests you will need to have your environment set up with `tox` installed.
 
-> See more: {external+juju:ref}`How to set up your development environment <manage-your-deployment-environment>`
+> See more: {external+juju:ref}`Set up your deployment <set-up-your-deployment>`
 
 ## Prepare the `tox.ini` configuration file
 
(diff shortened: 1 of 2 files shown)


docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -19,7 +19,7 @@ The instructions all use the Juju `python-libjuju` client, either through the `p
 
 In order to run integrations tests you will need to have your environment set up with `tox` installed.
```
dwilding: Not part of this PR, but it occurs to me that we should emphasize that _Juju_ should be installed, not only tox. Reading this, I may think "Oh, I already have tox, no need to follow those setup instructions". But then integration tests won't work because I don't have Juju.

Do you want to update to something like "with Juju and `tox` installed"? Or I can do follow-up PR

tonyandrewmeyer:
Oops, sorry @jameinel I didn't mean to ask you for a review here.

docs/howto/write-integration-tests-for-a-charm.md
tonyandrewmeyer: That seems small enough to slip into here. I've added this - feel free to suggest better wording, of course.

james-garner-canonical (approved):

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Looks great, thanks :slightly_smiling_face:

PR #1784, opened 2025-06-04
fix: fix type annotation of container check_infos in ops.testing

dwilding:
Testing code such as `container = testing.Container("db", check_infos={check_info})` was failing type checks, with this error:

```
error: Argument of type "set[CheckInfo]" cannot be assigned to parameter "check_infos" of type "frozenset[CheckInfo]" in function "__init__"
    "set[CheckInfo]" is not assignable to "frozenset[CheckInfo]" (reportArgumentType)
```

This PR fixes the type annotation of `check_infos` so that non-frozen sets are acceptable.

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1016,7 +1016,7 @@ class Container(_max_posargs(1)):
     notices: Sequence[Notice] = dataclasses.field(default_factory=list)
     """Any Pebble notices that already exist in the container."""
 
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
> It's annoying that our own checks didn't detect this. We do run pyright over the code in `testing/src` but not `testing/tests`. We should really do that - I'll open a ticket for it.

#1785.

james-garner-canonical (approved):

PR #1787, opened 2025-06-04
fix: Juju allows access to the remote app databag in relation-broken, so Harness should too

### LINKED ISSUE #888: Should relation data be visible to charm code from within a `relation-broken` hook?

tonyandrewmeyer:
Removes an outdated (or perhaps always wrong) check in Harness that prevents accessing the remote app databag during relation-broken events. Juju does allow this, so Harness should have the same behaviour.

This check is specific to Harness and there's no equivalent check in Scenario.

Fixes #888

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -2476,13 +2476,6 @@ class _TestingModelBackend:
         if not isinstance(is_app, bool):
             raise TypeError('is_app parameter to relation_set must be a boolean')
 
-        if 'relation_broken' in self._hook_is_running and not self.relation_remote_app_name(
-            relation_id
-        ):
-            raise RuntimeError(
-                'remote-side relation data cannot be accessed during a relation-broken event'
-            )
-
         if relation_id not in self._relation_data_raw:
             raise RelationNotFoundError(relation_id)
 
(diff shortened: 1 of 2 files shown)


james-garner-canonical:
I will confess I haven't fully gotten my head around this issue. If this Harness check is currently incorrect, I agree we should remove it. Can we run super tox with this change and see if it causes any failures?

tonyandrewmeyer:
> I will confess I haven't fully gotten my head around this issue.

There are lots of words in #888, and it's a muddled issue because it combines at least three problems and was originally just a question.

However, all that can be ignored for this PR. This PR is just what it says in the title: Juju lets you access the remote app databag in relation-broken (this was said by John in #888 but I've also verified it with Juju 3.6 locally), but Harness had a check that prevented that. Harness should model the actual Juju behaviour, so the PR removes that check.

> Can we run super tox with this change and see if it causes any failures?

Sure, will do.

benhoyt (approved):

tonyandrewmeyer:
> > Can we run super tox with this change and see if it causes any failures?
> 
> Sure, will do.

No new failures.

PR #1788, opened 2025-06-05
docs: clarify that group args require user args

james-garner-canonical:
It is an error at runtime to provide only the group to Pebble. This PR documents this.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2634,9 +2634,9 @@ class Container:
             user_id: User ID (UID) for file.
             user: Username for file. User's UID must match user_id if both are
                 specified.
-            group_id: Group ID (GID) for file.
+            group_id: Group ID (GID) for file. May only be specified along with user_id or user.
             group: Group name for file. Group's GID must match group_id if
-                both are specified.
+                both are specified. May only be specified along with user_id or user.
         """
         self._pebble.push(
             str(path),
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (commented):
Quick comment before a complete review: should the `user_id` and `user` words be wrapped in double backticks? I see that other use of `user_id` and `group_id` aren't, so maybe not?

@dwilding do we have a style guide position on this?

james-garner-canonical:
I also just realised that there's a bit more I wanted to document around this behaviour, so I'm going to move this PR back to draft.

dwilding:
> I prefer using italics (putting them in asterisks) as it's a lot less intrusive visually, and it's also the style the Python stdlib docs use

I'd be happy with us adopting that style, as these arguments are solid. Previously, I've tended to use italics rather than code font for file paths too, which I think makes them more pleasant to read.

For this PR, I'd be inclined to make user_id, user, etc italic rather than waiting. I feel they do need something to set them apart from the text. And we've already used this style elsewhere on the same page (e.g., https://ops--1788.org.readthedocs.build/en/1788/reference/ops.html#ops.Model.get_secret).

james-garner-canonical:
Done, I went with backticks for consistency with the existing style (e.g. the `Raises` section of the `push` docstrings).

Just for the record, I personally greatly prefer `code style` to *italics* for variables in documentation. I'll admit it doesn't look the greatest in our current docs theming, but there's always the hope that that can improve. I'm not opposed to migrating our to *italics* in future if that's the consensus though.

PR #1791, opened 2025-06-05
fix: support newer OpenTelemetry SDK

dimaqq:
I'm proposing this as a hot-fix to support both older and newer style of the OpenTelemetry SDK.

Upstream change that broke our code: https://github.com/open-telemetry/opentelemetry-python/pull/4580

### DIFF
--- a/tracing/ops_tracing/_backend.py
+++ b/tracing/ops_tracing/_backend.py
@@ -74,11 +74,17 @@ def _create_provider(resource: Resource, charm_dir: pathlib.Path) -> TracerProvi
 
 def get_exporter() -> BufferingSpanExporter | None:
     """Get our export from OpenTelemetry SDK."""
+    exporter = None
     try:
+        # opentelemetry-sdk versions xx ~ xx
         exporter = get_tracer_provider()._active_span_processor.span_exporter  # type: ignore
     except AttributeError:
-        # The global tracer provider was not configured by us and has a wrong processor.
-        return None
+        pass
+    try:
+        # opentelemetry-sdk versions yy ~ yy
+        exporter = get_tracer_provider()._active_span_processor._batch_processor._exporter
+    except AttributeError:
+        pass
     if not exporter or not isinstance(exporter, BufferingSpanExporter):
         return None
     return exporter
--- a/tracing/ops_tracing/_backend.py
+++ b/tracing/ops_tracing/_backend.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import pathlib
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


dimaqq:
Further discussion: https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$ZcWPuTNKswybevvFUKzwSubMxKZj9p7oSafvSCAnVnE?via=ubuntu.com&via=matrix.org&via=laquadrature.net

tracing/ops_tracing/_backend.py
```diff
@@ -74,13 +75,19 @@ def _create_provider(resource: Resource, charm_dir: pathlib.Path) -> TracerProvi
 
 def get_exporter() -> BufferingSpanExporter | None:
     """Get our export from OpenTelemetry SDK."""
-    try:
+    exporter = None
+
+    # opentelemetry-sdk < 1.34.0
+    with contextlib.suppress(AttributeError):
```
tonyandrewmeyer: I don't really understand why `contextlib.suppress` was added to the stdlib. I don't get how this is any better than try/except/pass, and it's a thing that many people need to learn as opposed to a simple try/except that almost any Python dev understands. We have [disabled](https://github.com/canonical/operator/blob/26b504273f9404bd43b9589d1205873df9dd3ef0/pyproject.toml#L154) linting that recommends this.

However, I guess I don't object to it, since it is in the stdlib, so smarter people than me presumably had good reason for it. So this is just your old-man-yells-at-cloud for the day.

tonyandrewmeyer (commented):
I still don't really like that we're using private attributes, since it seems like it keeps the likelihood of problems like this happening. If we pinned to specific versions we could avoid that, but that's generally bad practice for libraries.

It seems like this is worth doing now to get a quick fix out and we can figure out if there's something better we can do longer term afterwards.

I haven't managed to try this out yet - will finish that first thing tomorrow I hope.

tracing/ops_tracing/_backend.py
dimaqq: Sure, let me copy-paste that setting to the tracing subdir.

dimaqq: done, thanks, I was wondering about that recommended simplification.

james-garner-canonical (approved):
+1 on figuring out something else for the long term, but agreed on getting this fix in now

benhoyt:
I don't think we should merge this workaround, but rather figure out a way of not using those private attributes. Your option 2 in the chat sounds reasonable to me: "use globals--record our exporter when we set the library-global tracer provider, and get a reference to own exporter from own global". Let's discuss further in daily and see if we can come to consensus.

PR #1792, opened 2025-06-05
docs: small adjustments to the 'how to trace charm code' doc

tonyandrewmeyer:
A few small adjustments to the how-to guide for tracing, where I ran into trouble trying to do this:

* Typo: "configuration"
* Remove the table of contents, because that's provided in the right-hand sidebar.
* Provide an example for the charmcraft YAML.
* Provide an example for instantiating the tracing object.
* Tweak the wording of the list of traces ops will generate.
* Explain that after you've added the tracing you still need to integrate with a tracing provider before you can really see anything, and provide an example of doing that.
* Move the note about using the context manager rather than the decorator into a tip box.
* Provide an example of adding a custom span and a custom event.
* Remove a duplicate instruction to add `ops[tracing]` in the dependencies.

### DIFF
--- a/docs/howto/trace-the-charm-code.md
+++ b/docs/howto/trace-the-charm-code.md
@@ -8,14 +8,6 @@ execution using OpenTelemetry.
 Refer to `ops.tracing` reference for the canonical usage example, configuration
 options and API details.
 
-This guide covers:
-- [Adding tracing to your charm](#getting-started-tracing-a-charm)
-- [Creating custom spans and events](#custom-spans-and-events)
-- [Adding tracing to charm libraries](#adding-tracing-to-charm-libraries)
-- [Migrating from the `charm_tracing` charm library](#migrating-from-charm_tracing-charm-library)
-- [How and when to use the lower-level API](#lower-level-api)
-
-(getting-started-tracing-a-charm)=
 ## Getting started
 
 To enable basic tracing:
--- a/docs/howto/trace-the-charm-code.md
+++ b/docs/howto/trace-the-charm-code.md
@@ -5,15 +5,15 @@
 `ops.tracing.Tracing`, allowing you to observe and instrument your charm's
 execution using OpenTelemetry.
 
-Refer to `ops.tracing` reference for the canonical usage example, configuration
-options and API details.
+Refer to the `ops.tracing` reference for the canonical usage example, configuration
+options, and API details.
 
 ## Getting started
 
 To enable basic tracing:
 
-- In your charm's `pyproject.toml` or `requirements.txt`, add `ops[tracing]` as a dependency, for example
-- In your `charmcraft.yaml`, declare the tracing and (optionally) ca relations
+- In your charm's `pyproject.toml` or `requirements.txt`, add `ops[tracing]` as a dependency
+- In your `charmcraft.yaml`, declare the tracing and (optionally) ca relations, for example:
 
 ```yaml
 requires:
@@ -27,7 +27,7 @@ requires:
     optional: true
 ```
 
-- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object
+- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object, for example:
 
 ```python
 class MyCharm(ops.CharmBase):
@@ -50,6 +50,25 @@ At this point, Ops will trace:
 This provides coarse-grained tracing, focused on the boundaries between the
 charm code and the external processes.
 
+When you deploy your charm, until it is integrated with an app providing the `tracing` relation
+(and optionally the `certificate_transfer` relation), the traces will be buffered in a tracing
+database on the unit. When the charm is successfully integrated with the `tracing` provider,
+the buffered traces and new traces will be sent to the tracing destination. For Kubernetes
+charms, if the container is recreated, any buffered traces will be lost. Ops will buffer traces
+for a reasonable period of time and amount of space.
+
+For example, to send traces to [Grafana Tempo](https://grafana.com/docs/tempo/latest/) from
+a charm named `my-charm`, assuming that
+[Charmed Tempo HA](https://discourse.charmhub.io/t/charmed-tempo-ha/15531) has already been
+deployed:
+
+TODO: what are the actual names for tempo's tracing and certificate_transfer?
+
+```bash
+juju deploy my-charm
+juju integrate my-charm tempo
+```
+
 (custom-spans-and-events)=
 ## Custom spans and events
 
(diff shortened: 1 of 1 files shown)


docs/howto/trace-the-charm-code.md
```diff
@@ -5,36 +5,75 @@
...
+juju integrate my-charm tempo
+```
+
 (custom-spans-and-events)=
 ## Custom spans and events
 
+TODO: why do we need to depend on opentelemetry when ops does? Is it the newer version? If so why not just have ops depend on that?
+
 - In your charm's `pyproject.toml` or `requirements.txt`, add
   `opentelemetry-api ~= 1.30.0` dependency.
```
tonyandrewmeyer: @dimaqq do you actually have to add `opentelemetry-api ~= 1.30.0` as a dependency? Or will the opentelemetry-api dependency in ops provide this for you?

dimaqq: `ops[tracing]` will bring that in, sure.

I was thinking "explicit is better than implicit" though.
Perhaps this can be footnoted instead, like "if you need to control the opentelementry API version, do this..."

docs/howto/trace-the-charm-code.md
```diff
@@ -5,38 +5,71 @@
...
-Refer to `ops.tracing` reference for the canonical usage example, ocnfiguration
-options and API details.
-
-This guide covers:
-- Adding tracing to your charm
-- Creating custom spans and events
-- Adding tracing to charm libraries
-- Migrating from the `charm_tracing` charm library
-- How and when to use the lower-level API
+Refer to the `ops.tracing` reference for the canonical usage example, configuration
```
dwilding: How about making `ops.tracing` a link. Because it's pointing to the module, I think the neatest way is to use `{ref}`:

```suggestion
Refer to the {ref}`ops_tracing` reference for the canonical usage example, configuration
```

docs/howto/trace-the-charm-code.md
```diff
@@ -5,38 +5,71 @@
...
 
 To enable basic tracing:
 
-- In your charm's `pyproject.toml` or `requirements.txt`, add `ops[tracing]` dependency
-- In your `charmcraft.yaml`, declare the tracing and (optionally) ca relations
-- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object
-
-At this point, the Ops library will be traced:
+- In your charm's `pyproject.toml` or `requirements.txt`, add `ops[tracing]` as a dependency
+- In your `charmcraft.yaml`, declare the tracing and (optionally) ca relations, for example:
```
dwilding: A couple of minor suggestions. I probably give different suggestions each time I review something (!) but I feel like we should probably use one of these two constructions:

* In `foo.bar`
* In your `foo.bar` file

```suggestion
- In `pyproject.toml` or `requirements.txt`, add `ops[tracing]` as a dependency
- In `charmcraft.yaml`, declare the tracing and (optionally) certificate_transfer relations, for example:
```

docs/howto/trace-the-charm-code.md
```diff
@@ -5,38 +5,71 @@
...
+    interface: tracing
+    limit: 1
+    optional: true
+  receive-ca-cert:
+    interface: certificate_transfer
+    limit: 1
+    optional: true
+```
+
+- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object, for example:
```
dwilding: Minor suggestions

```suggestion
- In your charm's `__init__` method, instantiate the `ops.tracing.Tracing` object, for example:
```

docs/howto/trace-the-charm-code.md
```diff
@@ -5,38 +5,71 @@
...
+        ...
+```
+
+At this point, Ops will trace:
 - The `ops.main()` call
-- The events that the Ops library emits
-- The hook tools called by the charm code
-- The Pebble API access by the charm code
+- Events that Ops emits, including all the Juju events
+- Hook tools called by the charm code
```
dwilding: We barely talk about "hook tools" in the Ops docs, since the framework wraps them. Is there a different way to put this that would still be useful? Something like this?

```suggestion
- Ops calls that inspect and update Juju (also called "hook tools")
```

docs/howto/trace-the-charm-code.md
```diff
@@ -5,38 +5,71 @@
...
 
 This provides coarse-grained tracing, focused on the boundaries between the
 charm code and the external processes.
 
+When you deploy your charm, until it is integrated with an app providing the `tracing` relation
+(and optionally the `certificate_transfer` relation), the traces will be buffered in a tracing
+database on the unit. When the charm is successfully integrated with the `tracing` provider,
+the buffered traces and new traces will be sent to the tracing destination. For Kubernetes
+charms, if the container is recreated, any buffered traces will be lost. Ops will buffer traces
+for a reasonable period of time and amount of space.
```
dwilding: For the implementation of this, is there actually a time limit, or is it all determined by storage? I'm wondering whether it might be better to say something like "Ops allocates a reasonable amount of storage for the buffered traces." Also I might split the paragraph, to make key points easier to spot:

```suggestion
When you deploy your charm, until it is integrated with an app providing the `tracing` relation
(and optionally the `certificate_transfer` relation), the traces will be buffered in a tracing
database on the unit. Ops allocates a reasonable amount of storage for the buffered traces.

When the charm is successfully integrated with the `tracing` provider,
the buffered traces and new traces will be sent to the tracing destination.

For Kubernetes charms, if the container is recreated, any buffered traces will be lost.
```

docs/howto/trace-the-charm-code.md
```diff
@@ -47,31 +80,60 @@ charm code and the external processes.
...
   `opentelemetry-sdk`, `opentelemetry-proto`, `opentelemetry-exporter-*`,
   `opentelemetry-semantic-conventions` and add `ops[tracing]` instead.
 - In your repository, remove the `charm_tracing` charm library.
-- In your charm code, remove `@trace_charm` decorator and its helpers: the 
+- In your charm code, remove the `@trace_charm` decorator and its helpers: the
   `tracing_endpoint` and `server_cert` properties or methods.
-- In your charm's `pyproject.toml` or `requirements.txt`, add `ops[tracing]` dependency
 - In your `charmcraft.yaml`, take note of the tracing and (optionally) ca relation names.
-- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object
+- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object,
```
dwilding:
```suggestion
- In your charm's `__init__` method, instantiate the `ops.tracing.Tracing` object,
```

dwilding (approved):
Looks good, added some small suggestions and questions about a couple of parts

docs/howto/trace-the-charm-code.md
```diff
@@ -5,38 +5,73 @@
...
 
 To enable basic tracing:
 
-- In your charm's `pyproject.toml` or `requirements.txt`, add `ops[tracing]` dependency
-- In your `charmcraft.yaml`, declare the tracing and (optionally) ca relations
-- In your charm class `__init__`, instantiate the `ops.tracing.Tracing(...)` object
-
-At this point, the Ops library will be traced:
+- In `pyproject.toml` or `requirements.txt`, add `ops[tracing]` as a dependency
+- In `charmcraft.yaml`, declare the tracing and (optionally) certificate_transfer relations, for example:
```
dimaqq: Maybe we ought to be more explicit here:

```suggestion
- In `charmcraft.yaml`, declare the charm_tracing and (optionally) certificate_transfer relations, for example:
```

The reason being that charm_tracing is/should be typically separate from workload tracing.

dimaqq: Or maybe like this?

```suggestion
- In `charmcraft.yaml`, declare the relations for tracing and (optionally) certificate_transfer interfaces, for example:
```

docs/howto/trace-the-charm-code.md
```diff
@@ -47,31 +82,60 @@ charm code and the external processes.
...
+                try:
+                    subprocess.run('/path/to/migrate.sh', capture_output=True, check=True)
+                except subprocess.CalledProcessError:
+                    opentelemetry.trace.get_current_span().add_event(
+                        'db-migrate-failed',
+                        {'attempt': attempt},
+                    )
+                    time.sleep(10 ** attempt)
+                else:
+                    break
```
dimaqq:
```suggestion
    def migrate_db(self):
        with tracer.start_as_current_span('migrate-db') as span:
            for attempt in range(3):
                try:
                    subprocess.check_output('/path/to/migrate.sh')
                except subprocess.CalledProcessError:
                    span.add_event('db-migrate-failed', {'attempt': attempt})
                    time.sleep(10 ** attempt)
                else:
                    break
```

which ties nicely with the tip to use context managers over decorators.

dimaqq: P.S. `subprocess.run('/path/to/migrate.sh', capture_output=True, check=True)` -> `subprocess.check_output('/path/to/migrate.sh')`

Issue #1779, opened 2025-06-03
Add arg to filter output of `ObjectEvents.events()`


sed-i:
Currently, the best practice for funneling events to the same observer is ([ref](https://github.com/canonical/operator/issues/1108#issuecomment-1885928739)):

```python
for event in (kind for kind in self.on.events().values() if isinstance(kind, ops.HookEvent)):
    self.framework.observe(event, self._on_event)
```

where `events()` is defined here:
https://github.com/canonical/operator/blob/1d234866918983301be01627d6d9fa16926c6915/ops/framework.py#L492

In this approach:
1. It is "all or nothing": we can't funnel just "storage" events, or just "workload" events, or just "relation" events.
2. The manual filtering is a bit of boilerplate we'd need to repeat everywhere.

It would be handy if `events()` had an optional filtering arg, for example:
```python
self.on.events()  # All events, including FrameworkEvents
self.on.events(ops.HookEvent)  # All "hook" events
self.on.events(ops.WorkloadEvent)  # Only workload events
```

This way we'd be able to limit the scope of each reconciler method while keeping the code ordering more explicit than the usual "order of observe() calls.

benhoyt:
We kind of like this idea -- adding an optional "class filter" parameter to `on.events()` seems the best proposal yet. That said, I don't see how it gains much over an `if` statement -- it's slightly less verbose, but also less clear, which doesn't seem like a win. This seems simple and more clear to me:

```python
for event in self.on.events().values():
    if issubclass(event.event_type, ops.RelationEvent):
        self.framework.observe(event, self._on_event)
```

sed-i:
- It's longer, spreading a bit of boilerplate.
  - `for event in self.on.events(ops.RelationEvent).values():`, vs
  - `for event in self.on.events().values(): if issubclass(event.event_type, ops.RelationEvent):`.
- The explicit `issubclass` seems a bit like a bigger impl detail leak?
- With *args support, `self.on.events(T1, T2)` seems more readable than `issubclass(event.event_type, (T1, T2))`.

benhoyt:
That's fair! On balance, especially with the need for `.event_type` in there, I'm inclined to add this. I'd rather not allow `*args`, though -- let's just take a single `classinfo` arg and allow it to be a single class or tuple of classes, like `issubclass` does. So I think the signature would become:

```python
def events(self, classinfo: type[EventBase] | tuple[type[EventBase], ...]) -> dict[str, BoundEvent]:
    ...
```

[...]

PR #1793, opened 2025-06-05
Event filtering

PietroPasotti:
Implemented event filtering as per #1779

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -489,9 +489,21 @@ class ObjectEvents(Object):
                 event_kinds.append(attr_name)
         return event_kinds
 
-    def events(self) -> dict[str, BoundEvent]:
-        """Return a mapping of event_kinds to bound_events for all available events."""
-        return {event_kind: getattr(self, event_kind) for event_kind in self._event_kinds()}
+    def events(self, *event_type: type[EventBase]) -> dict[str, BoundEvent]:
+        """Return a mapping of event names to BoundEvent instances the framework can observe.
+
+        Args:
+            event_type: An optional tuple of event types to narrow the selection.
+                        If omitted, all events will be returned.
+        """
+        events = {event_kind: getattr(self, event_kind) for event_kind in self._event_kinds()}
+        if event_type:
+            return {
+                event_kind: event
+                for event_kind, event in events.items()
+                if issubclass(event.event_type, event_type)
+            }
+        return events
 
     def __getitem__(self, key: str) -> PrefixedEvents:
         return PrefixedEvents(self, key)
--- a/test/test_events.py
+++ b/test/test_events.py
@@ -1,0 +1,81 @@
+# Copyright 2025 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from __future__ import annotations
+
+import pytest
+
+import ops
+from ops import Framework
+from ops.testing import CharmEvents, Context, State
+
+
+def collect_filtered(*event_type, meta: dict | None = None) -> tuple[ops.EventBase, ...]:
+    ctx = Context(ops.CharmBase, meta={'name': 'ben', **(meta or {})})
+    with ctx(ctx.on.update_status(), State()) as mgr:
+        return tuple(e.event_type for e in mgr.charm.on.events(*event_type).values())  # type: ignore
+
+
+def test_filtering_null():
+    base_events = set(
+        subc
+        for subc in ops.HookEvent.__subclasses__()
+        # exclude non-leaves
+        if not len(subc.__subclasses__()) > 1
+    ).union(
+        # only lifecycle event that is automatically added to charm.on
+        {ops.CollectStatusEvent}
+    )
+    secret_events = set(subc for subc in ops.SecretEvent.__subclasses__())
+    assert set(collect_filtered()) == base_events.union(secret_events)
+
+
+def test_filtering_secret():
+    secret_events = set(subc for subc in ops.SecretEvent.__subclasses__())
+    assert set(collect_filtered(ops.SecretEvent)) == secret_events
+
+
+@pytest.mark.parametrize(
+    'evt_type', (ops.StartEvent, ops.UpgradeCharmEvent, ops.SecretExpiredEvent)
+)
+def test_filtering_single(evt_type):
+    assert set(collect_filtered(evt_type)) == {evt_type}
+
+
+@pytest.mark.parametrize(
+    'filter_arg, event_to_emit, expect_run',
+    (
+        ((ops.RelationEvent,), CharmEvents.update_status(), False),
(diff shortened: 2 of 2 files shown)


ops/framework.py
```diff
@@ -489,9 +489,21 @@ def _event_kinds(self) -> list[str]:
...
         return event_kinds
 
-    def events(self) -> dict[str, BoundEvent]:
-        """Return a mapping of event_kinds to bound_events for all available events."""
-        return {event_kind: getattr(self, event_kind) for event_kind in self._event_kinds()}
+    def events(self, *event_type: type[EventBase]) -> dict[str, BoundEvent]:
+        """Return a mapping of event names to BoundEvent instances the framework can observe.
+
+        Args:
+            event_type: An optional tuple of event types to narrow the selection.
```
tonyandrewmeyer, line 496: This isn't actually passing a tuple, it's *args. @benhoyt indicated he'd rather have a tuple (so keep the doc as-is and change the signature to be `event_type: tuple[type[EventBase]]|None=None` (maybe `event_types`?). It does seem like a tuple rather than *args leaves things more open to having other arguments in the future, even though `events(ops.RelationEvent, ops.WorkloadEvent)` does read nicer than `events((ops.RelationEvent, ops.WorkloadEvent))`.

test/test_events.py
```diff
@@ -0,0 +1,81 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from __future__ import annotations
+
+import pytest
+
+import ops
+from ops import Framework
```
tonyandrewmeyer: This line seems a bit pointless? We can just write `ops.Framework`, right? It's only used in a single place, so it ends up being fewer characters too :)

tonyandrewmeyer (commented):
Nice, thanks for doing this!

I think our (just decided in the past week) guidelines for tests say that the new test file should be named `test/test_framework_events.py`.

ops/framework.py
tonyandrewmeyer: Hmm, writing out the example, though, it's much uglier with a single type, `events((ops.RelationEvent, ))`.

Maybe it could be `event_types: tuple[type[EventBase]]|type[EventBase]|None=None` allowing either a single type or a tuple of types? Or maybe we should reconsider `*event_types`?

benhoyt: Sorry if I was unclear, but what I was keen on was allowing either a single class or a tuple of classes, just like `issubclass` allows. That makes it the same as `issubclass`, and simple for the single-class case as well as clear for multiple classes; it also allows for additional positional arguments in future (not that I think that's likely).

The type annotation would be `event_type: type[EventBase] | tuple[type[EventBase], ...] | None = None`, and usage would be:

```python
on.events()
on.events(ops.RelationEvent)
on.events((ops.RelationEvent, ops.SecretEvent))
```

ops/framework.py
```diff
@@ -489,9 +489,21 @@ def _event_kinds(self) -> list[str]:
...
+                        If omitted, all events will be returned.
+        """
+        events = {event_kind: getattr(self, event_kind) for event_kind in self._event_kinds()}
+        if event_type:
+            return {
+                event_kind: event
+                for event_kind, event in events.items()
+                if issubclass(event.event_type, event_type)
+            }
+        return events
```
benhoyt, line 506: I don't have strong feelings, but it felt a bit weird not to filter it down in one for loop/comprehension:

```suggestion
        events = {}
        for kind in self._event_kinds():
            event = getattr(self, kind)
            if event_type is None or issubclass(event.event_type, event_type):
                events[kind] = event
        return events
```

Or perhaps:

```python
        events = {
            kind: getattr(self, kind) for kind in self._event_kinds()
            if event_type is None or issubclass(getattr(self, kind).event_type, event_type)
        }
        return events
```

test/test_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+from __future__ import annotations
+
+import typing
+
+import pytest
+
+import ops
+from ops import Framework
+from ops.testing import CharmEvents, Context, State
```
benhoyt: Similarly, we would advocate `from ops import testing` and `testing.State` for these (though I realise we haven't been consistent about that).

test/test_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def collect_filtered(
```
benhoyt: I think this helper function obscures the code in the tests and makes the actual usage of the `events()` call harder to see. I suggest we just put these couple of lines of code in each test, even if it means a tiny bit of duplication. In addition, it seems that the `meta` argument is not used; and I think the `**(meta or {})` construct is too clever anyway.

test/test_events.py
```diff
@@ -0,0 +1,87 @@
...
+    *event_type: type[ops.EventBase], meta: dict[str, typing.Any] | None = None
+) -> tuple[ops.EventBase, ...]:
+    ctx = Context(ops.CharmBase, meta={'name': 'ben', **(meta or {})})
+    with ctx(ctx.on.update_status(), State()) as mgr:
+        return tuple(e.event_type for e in mgr.charm.on.events(*event_type).values())  # type: ignore
+
+
+def test_filtering_null():
```
benhoyt: I find this test hard to follow. What's it actually testing, and can we rewrite it in a simpler, more imperative way?

benhoyt (changes requested):

ops/framework.py
dimaqq: Might as well `return {...compr...}` in the latest case.

Painting the walrus shed:
```py
return {kind: bound_event for kind in self._event_kinds()
    if (bound_event := getattr(self, kind)) and (event_type is None or issubclass(bound_event.event_type, event_type))}
```

test/test_framework_events.py
```diff
@@ -0,0 +1,86 @@
...
+        subc
+        for subc in ops.HookEvent.__subclasses__()
+        # exclude non-leaves
+        if not len(subc.__subclasses__()) > 1
+    ).union(
+        # only lifecycle event that is automatically added to charm.on
+        {ops.CollectStatusEvent}
+    )
+    secret_events = set(subc for subc in ops.SecretEvent.__subclasses__())
+    assert set(collect_filtered()) == base_events.union(secret_events)
```
dimaqq, line 45: This test is rather hard to read.

Can we have test case from the API user point of view?

Something that would look like:

```py
class MyCharm(...):
    def __init__(self, ...):
        super().__init__(...)
        for k, v in self.on.event(clever-args-here):
            ...
```

Then run it through Scenario, issue events and monitor some side-effect, like call count on the callback or what not.

Or maybe like this:

```
class MyCharm(...):
    def __init__(self, ...):
        super().__init__(...)
        assert self.on.events(clever-args-here) == {
            "foo": BoundEvent(self, ops.Foo),
            ...
        }
```

dimaqq (changes requested):

PietroPasotti:
hey all, sorry for letting this sit for so long, there were some team discussions to be had before we decided which way we wanted to go with this.
The current feeling is that this PR is obsolete.
[This](https://github.com/canonical/cos-reconciler/blob/init/cos_reconciler/main.py#L22) (or something like it) is what we're leaning towards at this point.

The rationale from moving away from the `ops.filter_me_all_events_of_type(ops.HookEvent)` is that if in the future juju were to add a new event type on which something special needs to happen (say, the next ActionEvent or the next custom notice), our charms might inadvertently observe it and that's bad.
So we'd rather maintain a more-or-less pinned list of 'base event types' we want to observe, and so long as ops doesn't subclass those in the future or change their semantics, we're safe and every new event that gets added to the list will be a conscious decision.

tonyandrewmeyer:
Thanks for the update! Closing as per the above - Charm Tech have also had some concerns about what happens if/when there are new events with reconcile approaches, and I think we're also happier with the explicit-include approach.

PR #1794, opened 2025-06-06
fix: restrict the version of a dependency, opentelemetry-sdk

dimaqq:
The newest release broke our code.
This is a quick fix that we could release to the charmers.

Upstream: https://github.com/open-telemetry/opentelemetry-python/issues/4616


james-garner-canonical (approved):
I'm on board with this as a hotfix, but I wonder if #1791 is the better hotfix if our long term plan doesn't involve restricting the version.

PR #1797, opened 2025-06-06
feat: provide a method to create a testing.State from a testing.Context

### LINKED ISSUE #1454: Add the ability to populate State from metadata
The `State` could have some default components based on the `Context` metadata, or perhaps there could be a `fill_from_context` method or similar.

For example, this could add:
* A `Container` for each container in the metadata (although you'd probably want to be able to specify whether `can_connect` should be `True` or `False`, or it seems like there might be a lot of awkward `dataclasses.replace` afterwards)
* `config` that's the same as the defaults in the metadata
* A default `Network` (there's essentially always a Juju network, right?)
* A `Storage` for each storage in the metadata (to the minimum side of the `multiple` value if provided)
* A `StoredState` for each `ops.StoredState` attribute on the `Context`'s charm class (not perfect since these can exist elsewhere, particularly in libs, but maybe a useful base?)

I don't think it could have relations (even if we do make it possible to have data-inaccessible relations, e.g. #1452) since there's a practical difference between a relation existing (from the point of view of the unit) and not existing, wider than just the data being available.

I don't think it could include resources, even though they're in the metadata and always (?) available, because it doesn't seem like there's any way to have a sensible default location for the content.

See [this Matrix discussion](https://matrix.to/#/!xdClnUGkurzjxqiQcN:ubuntu.com/$34JjNCfc9Fz7RbWkrkHrDr5c4Isp43GGW404BkiZgLo?via=ubuntu.com&via=matrix.org&via=fsfe.org) for more background.

tonyandrewmeyer:
A new `from_context` class method is added to `testing.State`, which returns a `State` object that is populated by the metadata, config, and charm class.

* The config is set to the default values (as defined in `charmcraft.yaml`)
* A `testing.Relation`, `testing.PeerRelation`, or `testing.SubordinateRelation` is added for each relation in `charmcraft.yaml`'s `peers`, `requires`, and `provides` fields. The relation has the endpoint and interface set, but no data. Note that relations already automatically have a `Network` defined. Note that this simulates `integrate` having already run.
* A `testing.Container` is added for each container in `charmcraft.yaml`'s `containers` field. The container's `can_connect` is set to `True`, unlike the default `False` with `Container(name)`; this simulates the container having already been created
* A `testing.Storage` is added for each storage in `charmcraft.yaml`'s `storage` field
* A `testing.StoredState` is added for each `ops.StoredState` object that is a class attribute of the charm

In some cases, this will be a more useful starting point for a `State` than a blank `State` as provided by `State()`.

The biggest issue is that because the `State` is immutable, it is then awkward to do `dataclasses.replace` calls to insert data and do other modifications. To help with this, extra/non-default components can be passed in, just as they would be to the regular `State()`. For example:

```python
relation = testing.Relation('db', local_app_data={'foo': 'bar'})
state_in = testing.State.from_context(ctx, leader=True, relations={relation})
```

Fixes #1454


testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
             r for r in self.relations if _normalise_name(r.endpoint) == normalized_endpoint
         )
 
+    @classmethod
+    def from_context(
+        cls,
+        ctx: Context[CharmType],
+        **kwargs: Any,
```
james-garner-canonical: We'd spell these out in the final version, right?

testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+    ) -> State:
+        """Create a State from the charm context.
+
+        The initial data is loaded from the metadata in the context:
+
+        * `config` is loaded from the metadata's `config` field
+        * `relations` are loaded from the metadata's `requires`, `provides`, and `peers` fields
+        * `containers` are loaded from the metadata's `containers` field, and
+          are created with `can_connect=True`
```
james-garner-canonical, line 1804: As the odd one out in this list, I wonder if `can_connect` should be a possible keyword argument too?

testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+            for option, details in options.items():
+                if option not in generated_config and 'default' in details:
+                    generated_config[option] = details['default']
+        relations = kwargs.pop('relations', set())
+        containers = kwargs.pop('containers', set())
+        storages = kwargs.pop('storage', set())
+        if meta:
+            for relation_type in ('requires', 'provides', 'peers'):
+                for endpoint, details in meta.get(relation_type, {}).items():
+                    if any(rel for rel in relations if rel.endpoint == endpoint):
```
```suggestion
                    if any(rel.endpoint == endpoint for rel in relations):
```

testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+                        continue
+                    if relation_type == 'peers':
+                        relation_class = PeerRelation
+                    elif details.get('scope') == 'container':
+                        relation_class = SubordinateRelation
+                    else:
+                        relation_class = Relation
+                    relations.add(relation_class(endpoint, details['interface']))
+            for container_name in meta.get('containers', {}):
+                if any(c for c in containers if c.name == container_name):
```
```suggestion
                if any(c.name == container_name for c in containers):
```

testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+                    continue
+                containers.add(Container(name=container_name, can_connect=True))
+            for storage_name in meta.get('storage', {}):
+                if any(s for s in storages if s.name == storage_name):
```
```suggestion
                if any(s.name == storage_name for s in storages):
```

testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+        generated_config = kwargs.pop('config', {})
+        if config:
+            options = config.get('options', {})
+            for option, details in options.items():
+                if option not in generated_config and 'default' in details:
+                    generated_config[option] = details['default']
+        relations = kwargs.pop('relations', set())
+        containers = kwargs.pop('containers', set())
+        storages = kwargs.pop('storage', set())
+        if meta:
```
```suggestion
        if meta:  # generate relations, containers, and storages from meta
```

testing/src/scenario/state.py
```diff
@@ -1687,6 +1687,92 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+            if isinstance(value, ops.StoredState):
+                if any(
+                    ss
+                    for ss in stored_states
+                    if ss.name == attr and ss.owner_path == ctx.charm_spec.charm_type.handle_kind
+                ):
+                    continue
+                stored_states.add(
+                    StoredState(attr, owner_path=ctx.charm_spec.charm_type.handle_kind)
+                )
```
```suggestion
                owner_path = ctx.charm_spec.charm_type.handle_kind
                if any(s.name == attr and s.owner_path == owner_path for s in stored_states):
                    continue
                stored_states.add(StoredState(attr, owner_path=owner_path))
```

testing/src/scenario/state.py
tonyandrewmeyer: Is this not fairly obvious from the subsequent code? The comment gives me LLM vibes. If you really feel it's not clear without it I'll add it, but I'm gently pushing back.

testing/src/scenario/state.py
tonyandrewmeyer: A `containers_can_connect=False` seems ok, I guess, although it's unlike the others. I don't love that there is an odd case, but I think the automatic containers are far less useful if not connected, and doing a dataclasses.replace to change them all is also messy.

testing/src/scenario/state.py
tonyandrewmeyer: I was thinking this was the final version :)

I guess if all the arguments from `State.__init__` were copied here that provides better IDE experience and better static checking. I assume that's your motivation? The duplication, particularly since there `__init__` is autogenerated, is annoying but it does seem worthwhile.

testing/src/scenario/state.py
james-garner-canonical: Yeah I'm not experienced enough to know if being able to get containers that can't connect from this method would be useful for testing or not. If you think it would be, then I'd be in favour of adding the keyword argument, since they're already a bit of a special case documentation wise. If you don't think they would be, then we can always add the argument later if people ask for it.

testing/src/scenario/state.py
james-garner-canonical: The code is straightforward enough, but there's enough of it that it took me till `relations.add` to work out what all the preceding bits were for. But also, isn't `ctx.charm_spec.meta` always a dict? In which case it might be cleaner just to drop this line altogether and move the `containers = ...` and `storages = ...` lines above their respective `for` loops.

testing/src/scenario/state.py
tonyandrewmeyer: I'd prefer to leave it until people ask. I think if you're doing tests where the container can't connect you're more likely to be working with a minimal state and not need everything, or (particularly for something like a collect-status test) need to have all sorts of variations with relations not integrated, storage not set up, and so on.

james-garner-canonical: Sounds good to me

testing/src/scenario/state.py
james-garner-canonical: Yes, the motivation for spelling out all the kwargs is to provide a better ux.

Unfortunately there doesn't (yet??) seem to be a good way to express 'these are the `kwargs` from function x' (let alone '+/- this or that arg') with type hints to avoid having to duplicate all the arguments.

I forgot that this is a `dataclass` though, it does seem a shame to have to wrangle all the dataclass fields into regular arguments. Let me think about this. Does `kwargs` here correspond to exactly the arguments for `State`?

james-garner-canonical: Perhaps the most concise way to handle this would be to define a `TypedDict`. I don't love the duplication however we do this, but this should avoid messing with your implementation, where kwargs is used very cleanly internally, as it just touches the type signature.
```py
if TYPE_CHECKING:
    from typing_extensions import Unpack
    ...

    class StateKwargs(TypedDict, total=False):
        config: dict[str, str | int | float | bool]
        relations: Iterable[RelationBase]
        networks: Iterable[Network]
        containers: Iterable[Container]
        storages: Iterable[Storage]
        opened_ports: Iterable[Port]
        leader: bool
        model: Model
        secrets: Iterable[Secret]
        resources: Iterable[Resource]
        planned_units: int
        deferred: Sequence[DeferredEvent]
        stored_states: Iterable[StoredState]
        app_status: _EntityStatus
        unit_status: _EntityStatus
        workload_version: str
```
```suggestion
        **kwargs: Unpack[StateKwargs],
```

dimaqq (approved):
I think we'd need to update some doc to surface this pattern.

Given 100 lines of implementation, I'd expect more tests, as there's bound to be quirks for all those if's, if-else's and if any(...)'s.

+1 on James's TYPE_CHECKING kwargs definition.

testing/src/scenario/state.py
tonyandrewmeyer: Ah nice, thanks. I had implemented it the naive way by duplicating it all but this looks much nicer, I'll switch it over.

testing/tests/test_e2e/test_state.py
```diff
@@ -649,3 +651,153 @@ def _on_start(self, event: ops.StartEvent):
...
+            'relreq': {'interface': 'across'},
+            'sub': {'interface': 'below', 'scope': 'container'},
+        },
+    }
+    rel = rel_type(endpoint)
+    ctx = Context(ops.CharmBase, meta=meta)
+    state = State.from_context(ctx, relations={rel})
+    # The passed in relation should be used, not the one from the context, which
+    # means the interface should be None, as it wasn't provided, but is in the
+    # context/meta.
```
dimaqq: Would it be possible to instantiate it as smth like `rel_type(endpoint, interface=None)` to make this more explicit?
The comment can then be reduced to "Ensure that relation declared in metadata is overridden".

testing/tests/test_e2e/test_state.py
```diff
@@ -649,3 +651,153 @@ def _on_start(self, event: ops.StartEvent):
...
+
+def test_state_from_context_skip_exiting_container():
+    meta = {
+        'name': 'sam',
+        'containers': {'container': {}},
+    }
+    container = Container(name='container')
+    ctx = Context(ops.CharmBase, meta=meta)
+    state = State.from_context(ctx, containers={container})
+    assert state.get_container(container.name).can_connect is False
```
dimaqq, line 782: This too, either make it somehow explicit, or add a comment why it can't be connected, I'm guessing as opposed to the metadata declaration that could be?

testing/tests/test_e2e/test_state.py
```diff
@@ -649,3 +651,153 @@ def _on_start(self, event: ops.StartEvent):
...
+
+def test_state_from_context_skip_exiting_storage():
+    meta = {
+        'name': 'sam',
+        'storage': {'storage': {}},
+    }
+    storage = Storage(name='storage')
+    ctx = Context(ops.CharmBase, meta=meta)
+    state = State.from_context(ctx, storages={storage})
+    assert state.get_storage(storage.name, index=storage.index) == storage
```
dimaqq, line 798: This looks a bit odd... aren't the metadata declaration and storage instantiation equivalent?
Could we add some field that's different between the two?

testing/tests/test_e2e/test_state.py
```diff
@@ -649,3 +651,153 @@ def _on_start(self, event: ops.StartEvent):
...
+
+
+def test_state_from_context_skip_exiting_stored_state():
```
dimaqq, line 802: It's not clear from the test body whether the stored state is replaced or merged.

james-garner-canonical (approved):

testing/src/scenario/state.py
```diff
@@ -1760,6 +1781,93 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+                if any(rel.endpoint == endpoint for rel in relations):
+                    continue
+                if relation_type == 'peers':
+                    relation_class = PeerRelation
+                elif details.get('scope') == 'container':
+                    relation_class = SubordinateRelation
+                else:
+                    relation_class = Relation
+                relations.add(relation_class(endpoint, details['interface']))
+        containers = set() if containers is None else containers
```
dimaqq: **discussion:**

The RHS types don't seem to match here.
The function kwarg is an iterable, could be a list or a tuple or generator.
It could be smth like:
```suggestion
        containers = set(containers or ())
```

Then again, the State object's members are all `Iterable[this]` and `Iterable[that]` and we never snapshot the arguments, so maybe it's fine the way it is?

tonyandrewmeyer: Ah, you're right - this will work if people pass in sets as we want them to, or if they pass something else and we don't have extra ones from the context, but will fail if they pass something else and we do add, because we call `.add`.

The 'is None' business was trying to avoid creating new sets if an empty one was passed, but that's not really necessary.

Switched it over to the syntax you suggested, and also added a test for this case.

PR #1798, opened 2025-06-06
fix: don't use private OpenTelemetry API

dimaqq:
The recent opentelemetry-sdk release broke our private API access.

Upstream: https://github.com/open-telemetry/opentelemetry-python/issues/4616

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2654,12 +2654,14 @@ class Container:
             make_dirs: If True, create parent directories if they don't exist.
             permissions: Permissions (mode) to create file with (Pebble default
                 is 0o644).
-            user_id: User ID (UID) for file.
-            user: Username for file. User's UID must match user_id if both are
-                specified.
-            group_id: Group ID (GID) for file.
-            group: Group name for file. Group's GID must match group_id if
-                both are specified.
+            user_id: User ID (UID) for file. If neither ``group_id`` nor ``group`` is provided,
+                the group is set to the user's default group.
+            user: Username for file. User's UID must match ``user_id`` if both are
+                specified. If neither ``group_id`` nor ``group`` is provided,
+                the group is set to the user's default group.
+            group_id: Group ID (GID) for file. May only be specified with ``user_id`` or ``user``.
+            group: Group name for file. Group's GID must match ``group_id`` if
+                both are specified. May only be specified with ``user_id`` or ``user``.
         """
         self._pebble.push(
             str(path),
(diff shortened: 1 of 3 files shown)


tracing/ops_tracing/_backend.py
```diff
@@ -67,23 +70,12 @@ def setup(juju_context: _JujuContext, charm_class_name: str) -> None:
 def _create_provider(resource: Resource, charm_dir: pathlib.Path) -> TracerProvider:
     """Create the OpenTelemetry tracer provider."""
     # Separate function so that it's easy to override in tests
-    exporter = BufferingSpanExporter(charm_dir / BUFFER_FILENAME)
-    span_processor = BatchSpanProcessor(exporter)
+    global _exporter
+    _exporter = BufferingSpanExporter(charm_dir / BUFFER_FILENAME)
+    span_processor = BatchSpanProcessor(_exporter)
     return TracerProvider(resource=resource, active_span_processor=span_processor)  # type: ignore
```
benhoyt: Not a change in this PR, but I'm curious why there's a `type: ignore` comment here?

benhoyt (approved):
This seems like a nice fix to me (and even simplification), thanks!

james-garner-canonical (approved):

tonyandrewmeyer (approved):
Definitely preferable to the private opentelemetry use :)

The code changes look good to me. I've only done a limited test of actually using it in a real charm, as I'm still figuring out getting end-to-end tracing working on a sample charm. I'm trusting that you or another review has done full tests that cover that side of things.

james-garner-canonical:
> I'm trusting that you or another review has done full tests that cover that side of things.

I haven't!

tracing/ops_tracing/_backend.py
dimaqq: The kwarg is declared as:

```py
        active_span_processor: Union[
            SynchronousMultiSpanProcessor, ConcurrentMultiSpanProcessor, None
        ] = None,
```

Where

```py
class SynchronousMultiSpanProcessor(SpanProcessor):
    """Implementation of class:`SpanProcessor` that forwards all received
    events to a list of span processors sequentially.
```

We're using only one processor, so I short-circuited that.
The official way to set things up looks something like this:

```py
    _exporter = BufferingSpanExporter(charm_dir / BUFFER_FILENAME)
    span_processor = BatchSpanProcessor(_exporter)
    provider = TracerProvider(resource=resource)
    provider.add_span_processor(span_processor)
    return provider
```

Maybe it's worth setting things up this longer way, to be as robust as we can in case OpenTelemetry's internal change... given that sometimes even public methods disappear between versions?

dimaqq:
I'm happy to upgrade my toy charm to this branch and test.

dimaqq:
Scratch the above, we've got an integration test with a simple charm.

tracing/ops_tracing/_backend.py
dimaqq: James and I have talked this through and I'll change our API usage to follow the official example, as that's least likely to break in some future SDK version.

Issue #1799, opened 2025-06-08
_JujuContext.version default factory has no effect in practice.


dimaqq:
Our current code:

```py
@dataclasses.dataclass(frozen=True)
class _JujuContext:
    ...
    version: JujuVersion = dataclasses.field(default_factory=lambda: JujuVersion('0.0.0'))
```

However, the class is never instantiated directly, instead it's create from the environment, like so:

```py
    @classmethod
    def from_dict(cls, env: Mapping[str, Any]) -> _JujuContext:
        return _JujuContext(
            ...
            version=JujuVersion(env['JUJU_VERSION']),
            ...
        )
```

If `JUJU_VERSION` is missing in the environment, the factory method dies with `KeyError`.
And if it is present, the value is used.

The default factory is never used.

dimaqq:
I figure this is a valid Sunday item.

tonyandrewmeyer:
> If `JUJU_VERSION` is missing in the environment, the factory method dies with `KeyError`.

This isn't true after the changes in #1840.

> The default factory is never used.

None of the default values are currently used, because we're only using `from_dict` to create these objects. However, we are considering making the class public in the future, and I don't see any value in removing all the default values now just to add them back at that time. Even if we never do that, do we really gain anything?

I suggest we close this PR and if/when we make `_JujuContext` public look at the default values at that time.

james-garner-canonical:
Unassigning myself from this as this should be resolved by the API changes in the spec to make `_JujuContext` public. I'm leaving the issue open, but no objections if it gets closed.

tonyandrewmeyer:
Closing as I believe #1996 makes this out-of-date.

PR #1801, opened 2025-06-08
chore: reset JUJU_VERSION after each unit test that sets it

### LINKED ISSUE #1800: More JUJU_VERSION needs to be plugged in ops unit tests
The full unit test run succeeds:

`pytest --ignore=test/smoke --ignore=test/integration --ignore=test/benchmark --ignore=testing/tests/benchmark --ignore=tracing/test/ --ignore=examples/ -v --tb native -W 'ignore:Harness is deprecated:PendingDeprecationWarning'`

However, selective run fails:

`pytest --ignore=test/smoke --ignore=test/integration --ignore=test/benchmark --ignore=testing/tests/benchmark --ignore=tracing/test/ --ignore=examples/ -v --tb native -W 'ignore:Harness is deprecated:PendingDeprecationWarning' -k service_context_not_su`

```py
  File "/code/operator/test/test_model.py", line 1730, in container
    backend = MockPebbleBackend('myapp/0')
  File "/code/operator/ops/model.py", line 3379, in __init__
    juju_context = _JujuContext.from_dict(os.environ)
  File "/code/operator/ops/jujucontext.py", line 229, in from_dict
    version=JujuVersion(env['JUJU_VERSION']),
                        ~~~^^^^^^^^^^^^^^^^
  File "<frozen os>", line 716, in __getitem__
KeyError: 'JUJU_VERSION'
```

dimaqq:
Fixes #1800

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -46,6 +46,11 @@ def fake_script(request: pytest.FixtureRequest) -> FakeScript:
     return FakeScript(request)
 
 
+@pytest.fixture
+def fake_juju_version(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv('JUJU_VERSION', '0.0.0')
+
+
 class TestModel:
     @pytest.fixture
     def harness(self):
@@ -1042,7 +1047,7 @@ class TestModel:
 
         self.assertBackendCalls(harness, [])
 
-    def test_storage(self, fake_script: FakeScript):
+    def test_storage(self, fake_script: FakeScript, fake_juju_version: None):
         meta = ops.CharmMeta()
         raw: ops.charm._StorageMetaDict = {
             'type': 'test',
@@ -1132,7 +1137,7 @@ class TestModel:
     ):
         assert expected == harness._get_backend_calls(reset=reset)
 
-    def test_run_error(self, fake_script: FakeScript):
+    def test_run_error(self, fake_script: FakeScript, fake_juju_version: None):
         model = ops.Model(ops.CharmMeta(), _ModelBackend('myapp/0'))
         fake_script.write('status-get', """echo 'ERROR cannot get status' >&2; exit 1""")
         with pytest.raises(ops.ModelError) as excinfo:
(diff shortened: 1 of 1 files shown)


test/test_model.py
```diff
@@ -2730,8 +2735,8 @@ class TestModelBackend:
     def backend(self):
         backend_instance = getattr(self, '_backend', None)
         if backend_instance is None:
-            os.environ['JUJU_VERSION'] = '0.0.0'
-            self._backend = _ModelBackend('myapp/0')
+            with mock.patch.dict(os.environ, {'JUJU_VERSION': '0.0.0'}):
```
tonyandrewmeyer: It feels a little odd to be mixing monkeypatch and mock here. Can monkeypatch be layered with the same thing being patched? If so, wouldn't it be cleaner to use that here too?

james-garner-canonical:
I understand the issue being fixed in `def backend`, but why do so many other tests need the new `fake_juju_version` fixture?

test/test_model.py
dimaqq: A cleaner way would be to convert `def backend(...)` into a fixture.
I can give it a go.

dimaqq:
> I understand the issue being fixed in `def backend`, but why do so many other tests need the new `fake_juju_version` fixture?

Because the code under test relies on `JUJU_VERSION` to be set, see https://github.com/canonical/operator/issues/1799

test/test_model.py
james-garner-canonical: A backend fixture seems like a good idea to me. It would be nice if it could replace the `_fake_juju_version` fixture, since I believe that is used solely to allow `_ModelBackend` to be called. Alternatively, wdyt about a non-fixture helper function (e.g. `get_backend`) as a place to put the environment patching?

dimaqq: Done now, at the cost of refactoring some tests.

dimaqq:
I'm hoping to merge this before UV #1804

benhoyt (approved):

PR #1802, opened 2025-06-08
docs: in the "manage the charm version" how-to, give an example of using override-build

tonyandrewmeyer:
Inspired by the [parca-k8s charm](https://github.com/canonical/parca-k8s-operator/blob/afdc784503a0ea7ac52172f417fd0a379b137632/charmcraft.yaml#L26), include an example in the how-to guide for managing the charm version of using a override-build step in the parts to automatically set the version.

### DIFF
--- a/docs/howto/manage-the-charm-version.md
+++ b/docs/howto/manage-the-charm-version.md
@@ -23,12 +23,12 @@ $ cat version
 0522e1fd009dac78adb3d0652d91a1e8ff7982ae
 ```
 
-Typically, your publishing or packing workflow takes care of updating file file, so
+Typically, your publishing or packing workflow takes care of updating this file, so
 that it will automatically match the revision that you're creating and publishing.
 Generally, using a version control revision is the best choice, as it unambiguously
 identifies the code that was used to build the charm.
 
-One way to achieve this is to include in your `charmcraft.yaml` a parts build
+One way to achieve this is to modify your `charmcraft.yaml` file to include a parts build
 override, such as:
 
 ```yaml
@@ -39,7 +39,7 @@ parts:
     build-packages: [git]
     build-snaps: [astral-uv]
     override-build: |
-      craftctl default
+      craftctl default  # Run the default build steps.
       git describe --always > $CRAFT_PART_INSTALL/version
 ```
 
--- a/docs/howto/manage-the-charm-version.md
+++ b/docs/howto/manage-the-charm-version.md
@@ -23,13 +23,24 @@ $ cat version
 0522e1fd009dac78adb3d0652d91a1e8ff7982ae
 ```
 
-```{note}
-
-Normally, your publishing workflow would take care of updating this file, so
-that it will automatically match the revision that you're publishing. Generally,
-using a version control revision is the best choice, as it unambiguously
+Typically, your publishing or packing workflow takes care of updating file file, so
+that it will automatically match the revision that you're creating and publishing.
+Generally, using a version control revision is the best choice, as it unambiguously
 identifies the code that was used to build the charm.
 
+One way to achieve this is to include in your `charmcraft.yaml` a parts build
+override, such as:
+
+```yaml
+parts:
+  charm:
+    source: .
+    plugin: uv
+    build-packages: [git]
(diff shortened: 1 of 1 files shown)


docs/howto/manage-the-charm-version.md
```diff
@@ -23,13 +23,24 @@ $ cat version
 0522e1fd009dac78adb3d0652d91a1e8ff7982ae
 ```
 
-```{note}
-
-Normally, your publishing workflow would take care of updating this file, so
-that it will automatically match the revision that you're publishing. Generally,
-using a version control revision is the best choice, as it unambiguously
+Typically, your publishing or packing workflow takes care of updating file file, so
```
dwilding:
```suggestion
Typically, your publishing or packing workflow takes care of updating this file, so
```

docs/howto/manage-the-charm-version.md
```diff
@@ -23,13 +23,24 @@ $ cat version
...
+that it will automatically match the revision that you're creating and publishing.
+Generally, using a version control revision is the best choice, as it unambiguously
 identifies the code that was used to build the charm.
 
+One way to achieve this is to include in your `charmcraft.yaml` a parts build
```
dwilding: This alternative phasing flows better to my ear:

```suggestion
One way to achieve this is to modify your `charmcraft.yaml` file to include a parts build
```

docs/howto/manage-the-charm-version.md
```diff
@@ -23,13 +23,24 @@ $ cat version
...
+
+```yaml
+parts:
+  charm:
+    source: .
+    plugin: uv
+    build-packages: [git]
+    build-snaps: [astral-uv]
+    override-build: |
+      craftctl default
```
dwilding: This is the first time I've encountered `craftctl default`. Do you think we should include a comment, something like:

```suggestion
      craftctl default  # Run the default build steps.
```

dwilding (approved):
This looks like a helpful update, thanks!

docs/howto/manage-the-charm-version.md
```diff
@@ -23,13 +23,24 @@ $ cat version
...
+```yaml
+parts:
+  charm:
+    source: .
+    plugin: uv
+    build-packages: [git]
+    build-snaps: [astral-uv]
+    override-build: |
+      craftctl default  # Run the default build steps.
+      git describe --always > $CRAFT_PART_INSTALL/version
```
dimaqq, line 43: I've just left a comment on the PR that introduced that change: https://github.com/canonical/parca-k8s-operator/pull/440/files#r2136690910

In short:

- `git describe --always` find nearest **annotated** tag in the branch history.
- `git describe --always --tags` finds nearest tag of any kind.
- `git describe --always --all` finds nearest tag or branch or remote of any kind.

Example from the main branch of our repo:
```command
⋊> dima@bb ⋊> /c/operator on friday-use-uv ◦ git describe                                                                                                                                                   09:07:00
1.1.0-1426-g16d2d0fb
⋊> dima@bb ⋊> /c/operator on friday-use-uv ◦ git describe --tags                                                                                                                                            09:07:02
2.22.0-36-g16d2d0fb
```

If we don't recommend `--tags` we're making an assumption about charmers' repo setups.

In general, `git describe` is clever and gets the job done; however, the output takes time to understand. I'm on the fence if we should recommend it.

tonyandrewmeyer: This is meant to be more about how you can use `parts` with an override to automate it, rather than exactly what it contains.

What about `git rev-parse HEAD`, which would continue with the example above it? It gives you the exact revision (assuming it's not dirty). It's not as human-readable but if you're looking to reproduce a build I imagine you're comfortable enough getting a specific revision.

dimaqq: Looking though https://github.com/search?q=%22%3E+%24CRAFT_PART_INSTALL%2Fversion%22&type=code two types seem to be pretty common, like our company/craft idioms perhaps?
- `git describe --always` (more common)
- `git -C $CRAFT_PROJECT_DIR rev-parse --short HEAD`

So maybe let's keep it as currently in the PR.

PR #1816, opened 2025-06-11
chore: test RTD with shell update

dimaqq:


dimaqq:
```
# https://app.readthedocs.org/projects/ops/builds/28475659/

uv tool update-shell 0s Exit: 2
--
1 | error: The executable directory /home/docs/.local/bin is not in PATH, but the current shell could not be determined

```

nope, that didn't work.

Issue #633, opened 2021-10-05
ops.model.Model.config is misleading


aieri:
The [documentation](https://ops.readthedocs.io/en/latest/index.html#ops.model.Model.config) states that `ops.model.Model.config`

> Return a mapping of config for the current application

This is unexpected... given the name I would have thought `ops.model.Model.config` returned the model config instead.

Wouldn't it feel more natural if the app config was returned by `ops.model.Application.config`?

jtroup:
Curious why this was closed?

jnsgruk:
I spoke with @aieri separately and we talked about a slightly different use case which is likely more a feature request/discussion on Juju itself - which is the ability for charms to read the "model config" rather than just their own config - which if implemented would likely require some consideration on this issue :)

PR #1819, opened 2025-06-12
feat: add CheckInfo.successes field and .has_run property

benhoyt:
This adds a `successes` attribute to the `pebble.CheckInfo` class (see canonical/pebble#633).

It also adds a `CheckInfo.has_run` property (bikeshed the name!) which returns True if the check has been run at all (either successes and failures greater than zero). This raises `NotImplementedError` if the version of Pebble you're running against doesn't support the `successes` field -- that's a bit brutal, but we don't actually know the answer in that case.

I've added unit tests, and tested it manually against old and new versions of Pebble.

This PR also includes a drive-by to make `test/pebble_cli.py` work more like Pebble: it uses `/var/lib/pebble/default/.pebble.socket` by default if `PEBBLE` isn't set. This makes manual testing with `pebble_cli` simpler.

This should only be merged after https://github.com/canonical/pebble/pull/633 is finalised and merged.


benhoyt:
The failing test-real-pebble tests will come right once canonical/pebble#633 is merged.

james-garner-canonical (approved):
Looks good to me, assuming everything on the Pebble side shakes out this way. I don't have a better name idea for `has_run`, but I think it's a fine name.

test/pebble_cli.py
```diff
@@ -163,11 +169,11 @@ def main():
 
     socket_path = args.socket
     if socket_path is None:
-        pebble_env = os.getenv('PEBBLE')
-        if not pebble_env:
-            print('cannot create Pebble client (set PEBBLE or specify --socket)', file=sys.stderr)
-            sys.exit(1)
+        pebble_env = os.getenv('PEBBLE') or '/var/lib/pebble/default'
```
tonyandrewmeyer, line 172: This seems unrelated to the PR. Is it a drive-by? Could you mention it in the description and perhaps explain why it's being added?

ops/pebble.py
```diff
@@ -1521,11 +1536,26 @@ def __repr__(self):
...
+
+        Raises:
+            NotImplementedError: If running against a version of Pebble too
+                old to support the "successes" field.
+        """
+        if self.successes is None:
+            raise NotImplementedError(
+                'Pebble version doesn\'t support "successes" (introduced in Pebble 1.23.0)'
+            )
+        return self.successes > 0 or self.failures > 0
```
dimaqq, line 1558: **nit:** strictly speaking, "failures" is an optional field too.

dimaqq:
`AssertionError: assert (None is not None)`

Most useful error ever 🏆 

But then it is from the Pebble tests :)

ops/pebble.py
tonyandrewmeyer: An optional field, but the `__init__` sets it to `0` if not present, and it's typed as `int` not `int|None`, so that shouldn't matter here.

ops/pebble.py
```diff
@@ -1451,6 +1452,17 @@ class CheckInfo:
     :attr:`CheckStatus.INACTIVE` means the check is not running.
     """
 
+    successes: int | None
+    """Number of times this check has succeeded.
+
+    This is reset when the check succeeds again after the check's failure
+    threshold was reached. This will be zero if the check has never run, or
+    has never run successfully.
```
tonyandrewmeyer: Unless the behaviour in the Pebble PR changes, I think it would be better to say something like:

```suggestion
    This will be zero if the check has never run successfully after being started,
    either on Pebble startup through `startup: enabled`, or via
    :meth:`ops.Pebble.Client.start_checks`. It is reset to one when the check
    succeeds after the check's failure threshold was reached.
```

ops/pebble.py
```diff
@@ -1521,11 +1536,26 @@ def __repr__(self):
...
             f'startup={self.startup}, '
             f'status={self.status}, '
+            f'successes={self.successes}, '
             f'failures={self.failures}, '
             f'threshold={self.threshold!r}, '
             f'change_id={self.change_id!r})'
         )
 
+    @property
+    def has_run(self):
```
tonyandrewmeyer, line 1547: I think the name is fine. I was originally concerned about it reading as "check info has run", but all the other attributes are about the check too, so it aligns with those. Looking over our usage in docs etc we generally write `chk = c.get_check(x)` or `chk_info = c.get_check(x)` rather than `info = c.get_check(x)` and `if chk.has_run:` or `if chk_info.has_run:` both seem fine to me.

tonyandrewmeyer (approved):
Looks good to me, assuming the Pebble PR doesn't change in a way that would impact this.

As mentioned before, I think Scenario should also be updated. I feel like that should be a small amount of work so could fit in here, but am also fine if we leave it for a follow-up.

test/pebble_cli.py
benhoyt: Yeah, sorry, that's a drive-by to make manual testing a bit easier. I've updated the PR description.

ops/pebble.py
benhoyt: Good call -- updated.

dimaqq:
> The failing test-real-pebble tests will come right once #633 is merged.

https://github.com/canonical/pebble/pull/633 is meant, not #633

benhoyt:
@tonyandrewmeyer I've added Scenario support per our discussion the other day (I think!), so I'll go ahead and merge, but feel free to review post-merge when I'm away and fix it up if anything looks off.

PR #1820, opened 2025-06-12
docs: replace Harness example and fix links

### LINKED ISSUE #1817: README should not contain instructions for Harness tests.
Maybe the unit test instructions can be simply removed in the scope of #1655 

If we want unit testing example, let's update that to Scenario.

dwilding:
Fixes #1817. This is a quick PR to replace the Harness example in the Ops README by the latest (Scenario) unit test that comes from `charmcraft init`. We're planning to rework the README at a later date (see https://github.com/canonical/operator/issues/1655) but I think this is a reasonable temporary solution.

I've also fixed a couple of links to the Ops reference docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ def test_httpbin_pebble_ready():
     assert state_out.unit_status == testing.ActiveStatus()
 ```
 
-> See more: [`ops.testing.Harness`](https://ops.readthedocs.io/en/latest/#ops.testing.Harness)
+> See more: [`ops.testing`](https://ops.readthedocs.io/en/latest/reference/ops-testing.html)
 
 
 Explore further, start editing the files, or skip ahead and pack the charm:
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ class OpsExampleCharm(ops.CharmBase):
         self.unit.status = ops.ActiveStatus()
 ```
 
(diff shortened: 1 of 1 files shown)


james-garner-canonical (approved):
A nice improvement, thanks!

README.md
```diff
@@ -86,46 +86,49 @@ class OpsExampleCharm(ops.CharmBase):
...
-        updated_plan = self.harness.get_container_pebble_plan("httpbin").to_dict()
-        # Check we've got the plan we expected
-        self.assertEqual(expected_plan, updated_plan)
-        # Check the service was started
-        service = self.harness.model.unit.get_container("httpbin").get_service("httpbin")
-        self.assertTrue(service.is_running())
-        # Ensure we set an ActiveStatus with no message
-        self.assertEqual(self.harness.model.unit.status, ops.ActiveStatus())
+import ops
+import ops.pebble
```
benhoyt: Note that there's no need for this line, as `pebble` is already [imported in `ops/__init__.py`](https://github.com/canonical/operator/blob/3d9f3570c5eae148bb19cf4c9b09de41956811cb/ops/__init__.py#L195).

dwilding: I've just removed it, thanks! I've also removed it from the same charm that we're moving from Charmcraft to Ops (in https://github.com/canonical/operator/pull/1743).

PR #1821, opened 2025-06-12
ci: use Charmcraft 3.x for smoke testing 20.04 and 22.04

tonyandrewmeyer:
In #1782, we removed Charmcraft 2.x from the smoke test matrix. However, the smoke test code was still specifying that 20.04 and 22.04 should use Charmcraft 2, so when the test ran a `KeyError` would be raised.

This PR changes those bases to use Charmcraft 3.

[The branch runs successfully in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/15601083025) - I must have missed testing that last time.

### DIFF
--- a/test/smoke/test_smoke.py
+++ b/test/smoke/test_smoke.py
@@ -71,8 +71,8 @@ def pack(charm_dir: pathlib.Path):
 @pytest.mark.parametrize(
     'base,charmcraft_version,name',
     (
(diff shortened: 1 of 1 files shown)


james-garner-canonical (approved):

dimaqq:
I've just started debugging this fail on my branch, only to come and see it already fixed.
Magic!

PR #1823, opened 2025-06-13
chore: fix pebble tests

dimaqq:
#1804 introduced a false positive to pebble tests: pytest failures are ignored.

This PR updates tox to modern pattern (commands_pre/commands/commands_post) where the errors are surfaced correctly.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -155,16 +155,16 @@ base=pebble
 
 [testenv:pebble]
 description = Run real pebble tests
-allowlist_externals = pebble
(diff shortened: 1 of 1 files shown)


dimaqq:
I've also tested this change manually by injecting a failure into one of the tests.
Failure was reported correctly, both as text and return code.

james-garner-canonical (approved):
These changes seem correct to me, thanks. Also today I learned that you don't need a space before the `&` when backgrounding a process.

PR #1828, opened 2025-06-16
fix: Do not return this unit in a mocked peer relation

### LINKED ISSUE #1827: Relation.units is inconsistent between ops and ops[testing] for peer relations
When an application with a peer relation accesses that peer relation and try to get its units, the returned set will not contain the current unit. Consider the following charm code:

```
peer_relation = self.model.get_relation("peers")
logger.error("%s", peer_relation.units)
```

When this code executes in a running charm with 1 unit, this will be the output:

```
ERROR unit.vault-k8s/0.juju-log set()

```

When scaled to 3 units, and running on unit 0, the output will look something like this:

```
ERROR unit.vault-k8s/0.juju-log vault-peers:0: {<ops.model.Unit vault-k8s/1>, <ops.model.Unit vault-k8s/2>}
```

When trying to unit test for this behaviour with `ops[testing]`, the current unit is listed:

```
ERROR    vault.juju_facade:juju_facade.py:542 {<ops.model.Unit vault-k8s/0>}
```

This was tested with the following versions:

- ops 2.19.4
- ops-scenario 7.19.4
- juju 3.6.7

ghislainbourgeois:
This PR filters out the current unit from being returned by Scenario in the case of a peer relation, to align with Juju.

This PR assumes that the behavior reported in the issue is consistent across Juju versions.

Fixes #1827

### DIFF
--- a/testing/src/scenario/mocking.py
+++ b/testing/src/scenario/mocking.py
@@ -294,7 +294,12 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
         relation = self._get_relation_by_id(relation_id)
 
         if isinstance(relation, PeerRelation):
-            return tuple(f'{self.app_name}/{unit_id}' for unit_id in relation.peers_data)
+            this_unit = int(self.unit_name.split('/')[-1])
+            return tuple(
+                f'{self.app_name}/{unit_id}'
+                for unit_id in relation.peers_data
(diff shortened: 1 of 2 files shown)


james-garner-canonical (approved):
Agreed, thanks for this PR!

PR #1831, opened 2025-06-17
feat: add a helper to generate a Layer from rockcraft.yaml

### LINKED ISSUE #1629: Add a testing.Container layer from rockcraft.yaml
As we see & encourage adoption of having an initial Pebble layer for charms inside of a rock, charm unit tests increasingly need to have that layer in their `testing.Container`.

This is roughly:

```python
with open('/path/to/rockcraft.html') as f:
    rc = yaml.safe_load(f)
layer_dict = {'services': rc.get('services', {}), 'checks': rc.get('checks', {})}
layer = pebble.Layer(layer_dict)
```

There's enough boilerplate there that it feels like it might be worthwhile providing a wrapper for this - and that might also continue to encourage this pattern.

We can't do this automatically as with `charmcraft.yaml`, since we don't know where the `rockcraft.yaml` file is likely to be located. Adding a "add_layer_from_rock" method to `Container` is probably not the right approach, since they are meant to be immutable - unless it was a classmethod that returned a new `Container`. Perhaps a `rockcraft.yaml` could be passed to `__init__`, or maybe a top-level helper utility that returned a new `pebble.Layer`.

tonyandrewmeyer:
This PR adds a new `layer_from_rockcraft` method to Scenario, to allow test code like:

```python
ctx = testing.Context(RockyCharm)
container = testing.Container('workload', layers=[testing.layer_from_rockcraft('../rock/rockcraft.yaml')])
...
```

Fixes #1629


test/test_infra.py
```diff
@@ -56,7 +56,7 @@ def test_import(mod_name: str, tmp_path: pathlib.Path):
 )
 def test_ops_testing_doc():
     """Ensure that ops.testing's documentation includes all the expected names."""
-    prefix = '.. autoclass:: ops.testing.'
+    prefix = ('.. autoclass:: ops.testing.', '.. automethod:: ops.testing.')
```
james-garner-canonical: Want to inline this at its single use site?

testing/tests/test_e2e/test_state.py
```diff
@@ -654,6 +658,143 @@ def _on_start(self, event: ops.StartEvent):
...
+        rockcraft_path = f.name
+        layer = layer_from_rockcraft(rockcraft_path)
+    assert layer.summary == rockcraft['summary']
+    if 'description' in rockcraft:
+        description = (
+            f'{rockcraft["description"]} (built from the rockcraft.yaml at {rockcraft_path})'
+        )
+    else:
+        description = f'(no description) (built from the rockcraft.yaml at {rockcraft_path})'
+    assert layer.description == description
```
james-garner-canonical: WDYT about loosening this up, maybe something like:
```suggestion
    if 'description' in rockcraft:
        assert description in layer.description
    assert rockcraft_path in description
```

james-garner-canonical (approved):
Looks good, a couple of thoughts on tests, feel free to reject them

IronCore864 (approved):

test/test_infra.py
tonyandrewmeyer: Yeah, I guess so. It used to be used twice but I had to drop that when I changed it to a tuple and used it in startswith.

PR #1832, opened 2025-06-17
fix: testing.PeerRelation properly defaults to no peers

tonyandrewmeyer:
The intention of Scenario's `PeerRelation` is that it defaults to no peers (the charm only has the one unit). In addition, `PeerRelation.peers_data` should only contain unit data for the *peers*, not for the unit that is under test - that is found in `local_unit_data`.

We were breaking these rules when creating a `PeerRelation` without an explicit `peers_data`. We would add in the default Juju relation data - but that was already present in `local_unit_data`, and we used `0` as the unit ID. In most cases, `0` is the unit being tested, so this data was mostly ignored and unable to be reached, although it was the underlying cause of the bug fixed in #1828. The worse situation is if someone provides a unit ID to `Context`, other than 0, and has a `PeerRelation` with default `peers_data`, the charm would 'magically' have two units.

The main fix here is that `peers_data` should be the empty dictionary by default. If charmers want peers, then they need to pass in an appropriate dictionary (and handle the default Juju keys themselves if relevant).

Since it's very simple to get `peers_data` wrong and put in data that should be in `local_unit_data`, the PR also adds a consistency check that verifies that everything is ok and suggests a fix if not.

### DIFF
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -66,6 +66,7 @@ def check_consistency(
     event: _Event,
     charm_spec: _CharmSpec[Any],
     juju_version: str,
+    unit_id: int,
 ):
     """Validate the combination of a state, an event, a charm spec, and a Juju version.
 
@@ -111,6 +112,7 @@ def check_consistency(
             event=event,
             charm_spec=charm_spec,
             juju_version=juju_version_,
+            unit_id=unit_id,
         )
         errors.extend(results.errors)
         warnings.extend(results.warnings)
@@ -536,6 +538,7 @@ def check_relation_consistency(
     state: State,
     event: _Event,
     charm_spec: _CharmSpec[CharmType],
+    unit_id: int,
     **_kwargs: Any,
 ) -> Results:
     """Check the consistency of any relations in the :class:`scenario.State`."""
@@ -597,6 +600,18 @@ def check_relation_consistency(
             errors.append('duplicate endpoint name in metadata.')
             break
         seen_endpoints.add(endpoint)
+
+    # make sure that a peer relation doesn't have data for its own unit in peers_data
+    for relation in state.relations:
+        if not isinstance(relation, PeerRelation):
+            continue
+        if unit_id in relation.peers_data:
+            errors.append(
+                f'`peers_data` should only contain data for other units, not '
+                f'for the unit under test (unit {unit_id}). '
+                f'Instead of `peers_data={{{unit_id}: x, other_unit: y}}`, use '
+                f'`peers_data={{other_unit: y}}, local_unit_data=x`.',
+            )
 
     return Results(errors, [])
 
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -68,7 +68,7 @@ def check_consistency(
     juju_version: str,
     unit_id: int,
 ):
-    """Validate the combination of a state, an event, a charm spec, and a Juju version.
+    """Validate the combination of a state, event, charm spec, Juju version, and unit.
 
     When invoked, it performs a series of checks that validate that the state is
     consistent with itself, with the event being emitted, the charm metadata,
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -711,7 +711,12 @@ class PeerRelation(RelationBase):
     """A relation to share data between units of the charm."""
 
     peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(default_factory=dict)
-    """Current contents of the peer databags."""
+    """Current contents of the peer databags.
+
+    Note that this does not include data for the unit being tested. Data for
+    that unit must be added to :attr:`RelationBase.local_unit_data`. To control
+    which unit is being tested, pass ``unit_id`` to the :class:`Context`.
+    """
     # Consistency checks will validate that *this unit*'s ID is not in here.
 
     def __hash__(self) -> int:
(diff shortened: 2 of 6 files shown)


testing/src/scenario/state.py
```diff
@@ -710,9 +710,7 @@ def remote_unit_name(self) -> str:
 class PeerRelation(RelationBase):
     """A relation to share data between units of the charm."""
 
-    peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(
-        default_factory=lambda: {0: _DEFAULT_JUJU_DATABAG.copy()},
-    )
+    peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(default_factory=dict)
     """Current contents of the peer databags."""
     # Consistency checks will validate that *this unit*'s ID is not in here.
```
tonyandrewmeyer, line 720: :exploding_head: - this comment was already here, but I've checked and the consistency check was not, so I guess someone in the future went back to put it in there to remind us to do it now?

james-garner-canonical (commented):
This seems like a good fix, thanks.

I think it would be worth running super tox to see if anyone gets hit with `InconsistentScenarioError` due to this change, or if anyone had been asserting on those bad default values, or `peers_data` being truthy, etc

I also wonder if it's worth checking whether [this new test](https://github.com/canonical/operator/pull/1828/files#diff-8a5a0358397e9f28c451313f8d64eab95a4742852de78325934596631d806e79) now passes with [this check](https://github.com/canonical/operator/pull/1828/files#diff-bae0f1aa53467e743617ad1a56c5a8b7058a45b49b8564d4ff23e42054317d38) removed -- though perhaps it's worth retaining the check anyway?

tonyandrewmeyer:
> I think it would be worth running super tox to see if anyone gets hit with `InconsistentScenarioError` due to this change, or if anyone had been asserting on those bad default values, or `peers_data` being truthy, etc

Sure, can do.

> I also wonder if it's worth checking whether [this new test](https://github.com/canonical/operator/pull/1828/files#diff-8a5a0358397e9f28c451313f8d64eab95a4742852de78325934596631d806e79) now passes with [this check](https://github.com/canonical/operator/pull/1828/files#diff-bae0f1aa53467e743617ad1a56c5a8b7058a45b49b8564d4ff23e42054317d38) removed -- though perhaps it's worth retaining the check anyway?

Ah, thanks - I meant to address that in the PR description and forgot. The test does pass without the change, because `peers_data` is now correctly empty. The consistency check should make it impossible to get in that situation - unless you turn the consistency checks off. I feel like it's mostly harmless to keep it in place and does provide addition protection. But I'm also open to removing it to keep things clean and more minimal.

testing/src/scenario/state.py
```diff
@@ -710,9 +710,7 @@ def remote_unit_name(self) -> str:
 class PeerRelation(RelationBase):
     """A relation to share data between units of the charm."""
 
-    peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(
-        default_factory=lambda: {0: _DEFAULT_JUJU_DATABAG.copy()},
-    )
+    peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(default_factory=dict)
     """Current contents of the peer databags."""
```
dwilding: How about expanding the docstring to something like this (suitably wrapped):

```suggestion
    """Current contents of the peer databags, not including the databag for the unit under test. To access the databag for the unit under test, use ``local_unit_data``."""
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -66,6 +66,7 @@ def check_consistency(
     event: _Event,
     charm_spec: _CharmSpec[Any],
     juju_version: str,
+    unit_id: int,
 ):
     """Validate the combination of a state, an event, a charm spec, and a Juju version.
```
dwilding: Minor - Would this need to be updated too, if there's space?

```suggestion
    """Validate the combination of a state, event, charm spec, Juju version, and unit.
```

tonyandrewmeyer:
> > I think it would be worth running super tox to see if anyone gets hit with `InconsistentScenarioError` due to this change, or if anyone had been asserting on those bad default values, or `peers_data` being truthy, etc
> 
> Sure, can do.

No additional failures with `-e unit` or `-e scenario`.

dwilding (approved):
This makes sense to me. Added a couple of comments about docstrings.

james-garner-canonical (approved):
> The consistency check should make it impossible to get in that situation - unless you turn the consistency checks off.

Is that a user facing setting? If so then that seems like a good reason to keep the additional check so that we match the Juju behaviour in more cases.

> I feel like it's mostly harmless to keep it in place and does provide addition protection. But I'm also open to removing it to keep things clean and more minimal.

Otherwise I'd be inclined to remove it to keep the code as simple as possible.

I'll leave the decision to you.

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: Sure. This is private but you’re right better updated.

tonyandrewmeyer: Sure. This is private but you’re right better updated.

tonyandrewmeyer:
> Is that a user facing setting?

Yes, although via a fairly obscure environment variable.

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: Sure. This is private but you’re right better updated.

tonyandrewmeyer: Ugh, note to self, don't reply to comments on my phone while waiting for the movie to start. Apologies for the duplicates!

Issue #1805, opened 2025-06-10
Improve the httpbin charm


dwilding:
There are several things to improve about the httpbin charm that we're adding to Ops in #1743. It's probably best to create separate PRs for these improvements when working on them.

- [x] Switch to uv and tox-uv. In fact, switch to the project setup that will be produced by Charmcraft 4. This will be helpful in view of https://github.com/canonical/operator/issues/1994
    https://github.com/canonical/operator/pull/2011

- [x] Improve how we're returning a layer. See [#1116](https://github.com/canonical/charmcraft/issues/1116)
    https://github.com/canonical/operator/pull/2003

- [x] Switch the integration test to Jubilant
    https://github.com/canonical/operator/pull/1987

- [x] Add an integration test to check that the charm blocks when misconfigured
    https://github.com/canonical/operator/pull/2002

- [x] Add tox environments for formatting and linting? I removed these when I ported the charm from Charmcraft because the Ops repo has its own formatting and linting, but perhaps it's better for the charm to define them. (I actually forgot to remove `format` and `lint` from the list of default environments in tox.ini)
    https://github.com/canonical/operator/pull/1972

- [x] Use `load_config()`.
    https://github.com/canonical/operator/pull/1852

- [x] Rework the charm code to avoid using defer()
    https://github.com/canonical/operator/pull/1833

dwilding:
I've removed "Rework the charm code to avoid using defer()" from this issue. It's getting done in https://github.com/canonical/operator/pull/1833

PR #1833, opened 2025-06-17
docs: improve httpbin charm by removing defer() and adding collect_status

dwilding:
This PR makes a few improvements to the [httpbin demo charm](https://github.com/canonical/operator/tree/main/examples/httpbin-demo).

Main improvements:
- Added a class attribute for the workload container
- Made sure that log-level is always validated before being put into a Pebble layer
- Removed `defer()` and instead retry the container calls a few times
- Use collect_status to report the unit status in a consolidated way

Drive-bys:
- Added constants `CONTAINER_NAME` and `SERVICE_NAME` to help differentiate between different uses of `httpbin`
- Added a unit test for the case where an uppercase log level is configured
- Made the formatting of comments consistent
- As in #1734, specified a minimum Ops version and adjusted the formatting of tox.ini

### DIFF
--- a/examples/httpbin-demo/src/charm.py
+++ b/examples/httpbin-demo/src/charm.py
@@ -37,13 +37,13 @@ class HttpbinDemoCharm(ops.CharmBase):
         framework.observe(self.on.collect_unit_status, self._on_collect_status)
         framework.observe(self.on[CONTAINER_NAME].pebble_ready, self._on_httpbin_pebble_ready)
         framework.observe(self.on.config_changed, self._on_config_changed)
-        self.log_level = str(self.config['log-level']).lower()  # str() is for the type checker.
         self.container = self.unit.get_container(CONTAINER_NAME)
 
     def _on_collect_status(self, event: ops.CollectStatusEvent):
         """Report the status of the workload (runs after each event)."""
-        if self.log_level not in VALID_LOG_LEVELS:
-            event.add_status(ops.BlockedStatus(f"invalid log level: '{self.log_level}'"))
+        log_level = str(self.config['log-level']).lower()
+        if log_level not in VALID_LOG_LEVELS:
+            event.add_status(ops.BlockedStatus(f"invalid log level: '{log_level}'"))
         try:
             if not self.container.get_service(SERVICE_NAME).is_running():
                 # We can connect to Pebble in the container, but the service hasn't started yet.
@@ -64,8 +64,11 @@ class HttpbinDemoCharm(ops.CharmBase):
 
     def _on_httpbin_pebble_ready(self, event: ops.PebbleReadyEvent):
         """Define and start a workload using the Pebble API."""
+        log_level = str(self.config['log-level']).lower()
+        if log_level not in VALID_LOG_LEVELS:
+            return
         # Add initial Pebble config layer using the Pebble API.
-        self.container.add_layer('httpbin', self._pebble_layer, combine=True)
+        self.container.add_layer('httpbin', self._make_pebble_layer(log_level), combine=True)
         # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
         self.container.replan()
         # In rare cases, these calls could fail because the workload container became unavailable.
@@ -74,14 +77,17 @@ class HttpbinDemoCharm(ops.CharmBase):
 
     def _on_config_changed(self, event: ops.ConfigChangedEvent):
         """Handle changed configuration."""
-        if self.log_level not in VALID_LOG_LEVELS:
+        log_level = str(self.config['log-level']).lower()
+        if log_level not in VALID_LOG_LEVELS:
             return
         # Update the configuration of the workload.
         # We might not be able to access the workload container yet, so we'll try a few times.
         max_attempts = 3
         for attempt in range(max_attempts):
             try:
-                self.container.add_layer('httpbin', self._pebble_layer, combine=True)
+                self.container.add_layer(
+                    'httpbin', self._make_pebble_layer(log_level), combine=True
+                )
                 self.container.replan()
             except (ops.pebble.APIError, ops.pebble.ConnectionError):  # noqa: PERF203 (try-except in loop)
                 logger.info('Unable to reconfigure gunicorn (attempt %d)', attempt + 1)
@@ -93,10 +99,9 @@ class HttpbinDemoCharm(ops.CharmBase):
             # We expect that there'll be a pebble-ready event in the future,
             # which will configure and start the workload.
             return
-        logger.debug("Log level for gunicorn changed to '%s'", self.log_level)
+        logger.debug("Log level for gunicorn changed to '%s'", log_level)
 
-    @property
-    def _pebble_layer(self) -> ops.pebble.LayerDict:
+    def _make_pebble_layer(self, log_level: str) -> ops.pebble.LayerDict:
         """Return a dictionary representing a Pebble layer."""
         return {
             'summary': 'httpbin layer',
@@ -107,7 +112,7 @@ class HttpbinDemoCharm(ops.CharmBase):
                     'summary': 'httpbin',
                     'command': 'gunicorn -b 0.0.0.0:80 httpbin:app -k gevent',
                     'startup': 'enabled',
-                    'environment': {'GUNICORN_CMD_ARGS': f'--log-level {self.log_level}'},
+                    'environment': {'GUNICORN_CMD_ARGS': f'--log-level {log_level}'},
                 }
             },
         }


examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
...
-                self.unit.status = ops.MaintenanceStatus('waiting for Pebble API')
-                event.defer()
-                return
-
-            logger.debug("Log level for gunicorn changed to '%s'", log_level)
-            self.unit.status = ops.ActiveStatus()
-        else:
-            # In this case, the config option is bad, so block the charm and notify the operator.
-            self.unit.status = ops.BlockedStatus(f"invalid log level: '{log_level}'")
+        if self.log_level.lower() not in VALID_LOG_LEVELS or not self.container.can_connect():
```
dwilding: Per [discussion in Matrix](https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$3axOQJ_cYBN5MjW6OC6ApEjI5WarPkhAfdQuwad_ISk?via=ubuntu.com&via=matrix.org&via=laquadrature.net), should this `can_connect()` check include a retry with delay?

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
 
     def __init__(self, framework: ops.Framework):
         super().__init__(framework)
+        framework.observe(self.on.collect_unit_status, self._on_collect_status)
         framework.observe(self.on['httpbin'].pebble_ready, self._on_httpbin_pebble_ready)
         framework.observe(self.on.config_changed, self._on_config_changed)
+        self.log_level = str(self.config['log-level'])  # str() is here for the type checker.
```
james-garner-canonical: I wonder if the `.lower()` call go here instead? What's actually allowed for the log level? e.g. is `DeBuG` ok?

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
...
+        if not self.container.can_connect():
+            event.add_status(ops.MaintenanceStatus('waiting for container'))
+            return
+        try:
+            if not self.container.get_service('httpbin').is_running():
+                event.add_status(ops.MaintenanceStatus('waiting for workload'))
+                return
+        except ops.ModelError:
+            event.add_status(ops.MaintenanceStatus('waiting for Pebble'))
+            return
```
james-garner-canonical: Is this better without  the early `return`s, since the idea is that collect status will prioritise the statuses added with `add_status`?

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
...
+            event.add_status(ops.BlockedStatus(f"invalid log level: '{self.log_level}'"))
+            return
+        if not self.container.can_connect():
+            event.add_status(ops.MaintenanceStatus('waiting for container'))
+            return
+        try:
+            if not self.container.get_service('httpbin').is_running():
+                event.add_status(ops.MaintenanceStatus('waiting for workload'))
+                return
+        except ops.ModelError:
```
james-garner-canonical: Why do we get a `ModelError` here?

examples/httpbin-demo/src/charm.py
james-garner-canonical: I suspect so, but I'll let @tonyandrewmeyer make the call here. If so, would we want to use `tenacity`, or just use a loop?

examples/httpbin-demo/src/charm.py
tonyandrewmeyer: More radical idea:

```python
self.config = self.load_config(HttpBinConfig, errors='blocked')
```

With a class like:

```python
class HttpBinConfig:
    """Juju config options for the HttpBin charm."""

    log_level: str = 'info'
    """Configures the log level of gunicorn.

    Acceptable values are: "debug", "info", "warning", "error" and "critical"
    """

    def __post_init__(self):
        log_level = self.log_level.lower()
        if log_level not in VALID_LOG_LEVELS:  # That could probably be a class attribute.
            raise ValueError(f'`log-level` must be one of {", ".join(VALID_LOG_LEVELS)}, not {log_level}')
        object.__setattr__(self, 'log_level', log_level)
```

And then it's accessed as `self.config.log_level` when needed.

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
...
         framework.observe(self.on['httpbin'].pebble_ready, self._on_httpbin_pebble_ready)
         framework.observe(self.on.config_changed, self._on_config_changed)
+        self.log_level = str(self.config['log-level'])  # str() is here for the type checker.
+        self.container = self.unit.get_container('httpbin')
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent):
+        """Report the status of the workload (runs after each event)."""
+        if self.log_level.lower() not in VALID_LOG_LEVELS:
+            event.add_status(ops.BlockedStatus(f"invalid log level: '{self.log_level}'"))
+            return
```
tonyandrewmeyer: If we go with my more radical suggestion, this wouldn't be needed. Or if we think it would be better we could remove the `errors=`, catch the `ValueError` in `__init__` and then also have here something like:

```python
try:
    self.load_config(HttpBinConfig)
except ValueError as e:
    event.add_status(ops.BlockedStatus(str(e)))
```

examples/httpbin-demo/src/charm.py
tonyandrewmeyer: Yes, all the returns in collect-status should be removed - the idea is that you add all the relevant statuses and let ops pick which one to use.

examples/httpbin-demo/src/charm.py
tonyandrewmeyer: I really like `tenacity` and I've used it a lot in the past. However, it's a big library to include if you have simple needs. So I would generally avoid using it in charms.

I don't love `can_connect()` because it introduces a (extraordinarily unlikely) race condition: Pebble might be able to connect at the time of the `can_connect()` call and then fail in the following call (a millisecond or two later).

My preference is for a pattern like:

```python
for attempt in range(5):
    try:  # the linter will complain about try/except in a loop, so will need a noqa
        self.container.add_layer(...)
        self.container.replan()
    except (ops.pebble.APIError, ops.pebble.ConnectionError):
        logger.info(...)
        time.sleep(2**attempt)  # or whatever formula you like that will give reasonable times
    else:
        break
else:
    logger.warning(...)
    # We expect that there will be a pebble-ready event in the future that will handle this.
    return
```

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
...
-        # Get a reference the container attribute on the PebbleReadyEvent
-        container = event.workload
-        # Add initial Pebble config layer using the Pebble API
-        container.add_layer('httpbin', self._pebble_layer, combine=True)
+        # Add initial Pebble config layer using the Pebble API.
+        self.container.add_layer('httpbin', self._pebble_layer, combine=True)
         # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
-        container.replan()
-        self.unit.status = ops.ActiveStatus()
+        self.container.replan()
```
tonyandrewmeyer: Technically, you can get a pebble-ready and then fail to talk to Pebble (the container was up and then by the time you got the event it was down, or network issues, or whatever). Like with the `can_connect()` race condition, this should be extremely unlikely.

Personally, I think given how unlikely it is, it's reasonable to have the code like you have it. If it fails, the unit goes to error state and someone investigates (which they probably should, because your cloud probably has issues if this is happening) or Juju automatically retries if that setting is enabled. Some charmers strongly dislike error status and try to avoid it at all costs, but I feel like this is a legitimate error status.

But I would probably add a comment saying something like this. For example, "If we cannot connect to Pebble, let the unit go into error status, where this hook will be retried or the Juju user will investigate".

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +31,44 @@ class HttpbinDemoCharm(ops.CharmBase):
...
+            return
+        if not self.container.can_connect():
+            event.add_status(ops.MaintenanceStatus('waiting for container'))
+            return
+        try:
+            if not self.container.get_service('httpbin').is_running():
+                event.add_status(ops.MaintenanceStatus('waiting for workload'))
+                return
+        except ops.ModelError:
+            event.add_status(ops.MaintenanceStatus('waiting for Pebble'))
```
tonyandrewmeyer: Nit: it feels like the Juju user probably shouldn't need to know a lot about Pebble (although once they start to debug they certainly would) so maybe "waiting for the workload container" would be more the terms they'd expect?

examples/httpbin-demo/tests/unit/test_charm.py
```diff
@@ -86,6 +86,23 @@ def test_config_changed_valid_cannot_connect():
...
+    container = testing.Container('httpbin', can_connect=True)
+    state_in = testing.State(containers={container}, config={'log-level': 'DEBUG'})
+
+    # Act:
+    state_out = ctx.run(ctx.on.config_changed(), state_in)
+
+    # Assert:
+    updated_plan = state_out.get_container(container.name).plan
+    gunicorn_args = updated_plan.services['httpbin'].environment['GUNICORN_CMD_ARGS']
+    assert gunicorn_args == '--log-level DEBUG'
```
tonyandrewmeyer: I'm confused by this one (and haven't run the tests locally, sorry). Doesn't the log level get `lower()`d? It seems like this should be `debug` not `DEBUG`.

tonyandrewmeyer (commented):
Thanks for doing this! A few suggestions, including one you should feel free to reject to use the new config class functionality.

examples/httpbin-demo/src/charm.py
james-garner-canonical: I'm in favour of using `load_config`, but I think that isn't possible here till the next release (which we could wait for), or we update the CI to pull `ops` from `main` when running the example charm tests -- or I guess we could just update the `pyproject.toml` for this charm to specify `main`?

tonyandrewmeyer: Yeah, I was assuming a "wait and require 2.23", but fair point that it's still ~2 weeks away and @dwilding (presumably) doesn't want to add a blocked PR.

So I guess my suggestion would be we do this but in a follow-up PR, but as part of that, don't take too much care to improve the config handling because we know an improvement is coming.

(I'm happy to open one for using it at the end of the month if @dwilding doesn't want the extra work).

james-garner-canonical: How about if @dwilding is on board with using `load_config`, we temporarily change this charm's `requirements.txt` to specify `main` in this PR?
```
ops @ git+https://github.com/canonical/operator@main
```
 That way the post-release follow-up PR would only need to change that.

dwilding: I'm not sure how I feel about this. I agree with switching to `load_config`, in the spirit of keeping our example charms fresh on the latest features. I'm less sure about requiring Ops from main, even temporarily, as we wouldn't want someone to think that we recommend doing it for a real charm. I'll work on the other updates and see how I feel after that!

examples/httpbin-demo/src/charm.py
dwilding: In [the doc](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Container.get_service) it says that `get_service()` raises `ModelError` if the service is not found, which I'm interpreting as Pebble not being fully ready (although `can_connect()` has already returned True, so Pebble must be somewhat ready - modulo Tony's point about race conditions). Would it be better to take a different approach?

examples/httpbin-demo/tests/unit/test_charm.py
dwilding: It only gets `lower()`d when checking validity. The value that's actually set in the Pebble layer is whatever the Juju user provided. That's how the charm previously worked, but it does feel unnecessary. I agree with James's suggestion of applying `lower()` when initially grabbing the value from config.

examples/httpbin-demo/src/charm.py
dwilding: Resolved by adding this comment:

```
# In rare cases, these calls could fail because the workload container became unavailable.
# If this happens, we'll let the unit go into error status. The hook will be retried, or
# the Juju user can investigate.
```

examples/httpbin-demo/src/charm.py
dwilding: Resolved per Tony's suggestion:

```python
# Update the configuration of the workload.
# We might not be able to access the workload container yet, so we'll try a few times.
max_attepts = 3
for attempt in range(max_attepts):
    try:
        self.container.add_layer('httpbin', self._pebble_layer, combine=True)
        self.container.replan()
    except (ops.pebble.APIError, ops.pebble.ConnectionError):  # noqa: PERF203 (try-except in loop)
        logger.info('Unable to reconfigure gunicorn (attempt %d)', attempt + 1)
        time.sleep(2**attempt)
    else:
        break
else:
    logger.warning('Unable to reconfigure gunicorn after %d attempts.', max_attepts)
    # We expect that there'll be a pebble-ready event in the future,
    # which will configure and start the workload.
    return
```

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +32,61 @@ class HttpbinDemoCharm(ops.CharmBase):
...
-        # Do some validation of the configuration option
-        if log_level in VALID_LOG_LEVELS:
-            # The config is good, so update the configuration of the workload
-            container = self.unit.get_container('httpbin')
-            # Push an updated layer with the new config
+        if self.log_level.lower() not in VALID_LOG_LEVELS:
+            return
+        # Update the configuration of the workload.
+        # We might not be able to access the workload container yet, so we'll try a few times.
+        max_attepts = 3
```
tonyandrewmeyer: mmmmm 😉

examples/httpbin-demo/src/charm.py
dwilding: We need to keep the return inside `if not self.container.can_connect()` though - otherwise `self.container.get_service('httpbin')` will fail with a Pebble API error won't it?

Maybe the better approach here is to remove `can_connect()`, as on config changed, and directly try to see whether the service is running. We can catch whichever errors come back. No need for any returns.

examples/httpbin-demo/src/charm.py
dwilding: Passed all checks, nothing to see here :hear_no_evil:

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +32,55 @@ class HttpbinDemoCharm(ops.CharmBase):
...
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent):
+        """Report the status of the workload (runs after each event)."""
+        if self.log_level.lower() not in VALID_LOG_LEVELS:
+            event.add_status(ops.BlockedStatus(f"invalid log level: '{self.log_level}'"))
+        try:
+            if not self.container.get_service('httpbin').is_running():
+                event.add_status(ops.MaintenanceStatus('waiting for workload'))
+        except (ops.pebble.ConnectionError, ops.ModelError):
+            event.add_status(ops.MaintenanceStatus('waiting for workload container'))
```
dwilding: @james-garner-canonical & @tonyandrewmeyer would you mind taking a look at this updated logic? I think we need to catch `ops.ModelError` in case the service is not found. I don't think we need to catch `ops.pebble.APIError`, as `get_service` only asks the Pebble API for a list of services, which shouldn't return an error. What do you think?

examples/httpbin-demo/src/charm.py
dwilding: I prefer to leave it as-is for now, so that we don't need to require Ops from main. I've created #1839 for one of us to work on later

examples/httpbin-demo/requirements.txt
```diff
@@ -1 +1 @@
-ops < 4
+ops>2,<4
```
dwilding, line 1: Adding a minimum version (as in #1734) in anticipation of switching to `load_config` after the next release (#1839)

examples/httpbin-demo/tox.ini
```diff
@@ -48,13 +48,12 @@ deps =
...
     -r {tox_root}/requirements.txt
 commands =
-    coverage run --source={[vars]src_path} \
-                 -m pytest \
-                 --tb native \
-                 -v \
-                 -s \
-                 {posargs} \
-                 {[vars]tests_path}/unit
+    coverage run --source={[vars]src_path} -m pytest \
```
dwilding, line 51: Matching `unit` and `integration` changes from #1734

dwilding:
This is ready for re-review - thanks!

examples/httpbin-demo/src/charm.py
tonyandrewmeyer: I don't see how you'd get a `ModelError`. Typically, `ModelError` is something you'd get when using a hook tool - Pebble doesn't know anything about `ModelError`s, and I don't see any code that would convert a Pebble error to a `ModelError` in this case.

It's unlikely but [possible](https://github.com/canonical/pebble/blob/1593ac65235b3a64b8b09f5cec24b9414169af24/internals/daemon/api_services.go#L43) that getting the list of services fails.

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +32,55 @@ class HttpbinDemoCharm(ops.CharmBase):
...
         """Handle changed configuration."""
-        # Fetch the new config value
-        log_level = cast('str', self.model.config['log-level']).lower()
-
-        # Do some validation of the configuration option
-        if log_level in VALID_LOG_LEVELS:
-            # The config is good, so update the configuration of the workload
-            container = self.unit.get_container('httpbin')
-            # Push an updated layer with the new config
+        if self.log_level.lower() not in VALID_LOG_LEVELS:
```
james-garner-canonical:
```suggestion
        if self.log_level not in VALID_LOG_LEVELS:
```

examples/httpbin-demo/src/charm.py
```diff
@@ -32,43 +32,55 @@ class HttpbinDemoCharm(ops.CharmBase):
...
+            return
+        # Update the configuration of the workload.
+        # We might not be able to access the workload container yet, so we'll try a few times.
+        max_attempts = 3
+        for attempt in range(max_attempts):
```
james-garner-canonical, line 86: Very nitpicky and possibly a matter of taste, but I'd inline `max_attempts`
```suggestion
        for attempt in range(3):
```
Oh unless you want to add it to the logging like 'attempt 1 of 3' kind of thing.

examples/httpbin-demo/src/charm.py
dwilding: Oops, thanks

james-garner-canonical (approved):

examples/httpbin-demo/src/charm.py
james-garner-canonical: Feel free to leave it as is tho

examples/httpbin-demo/src/charm.py
dwilding: For `ModelError`, I was going off the doc: https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Container.get_service

tonyandrewmeyer: Ah, my bad, I read it as [`get_services` (with the 's')](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Container.get_services). `get_service` [indeed raises ModelError if you pass it a bad name](https://github.com/canonical/operator/blob/ff47e4100aa79ec1705617db380ebd0bb4d2953e/ops/model.py#L2704). So we should catch that too, yes.

james-garner-canonical: Or should we not catch `ModelError` and let it take us to error status instead, since it's an error if we got the service name wrong?

dwilding: To James's question - that's one thing I'm unclear about. Is a service name mismatch between charm code and layer definition the only reason we'd get `ModelError`? Is it possible for, e.g., Pebble to send over an empty list of services before it's fully ready?

examples/httpbin-demo/src/charm.py
dwilding: I'm using `max_attempts` in the logging message if all attempts fail:

```python
logger.warning('Unable to reconfigure gunicorn after %d attempts.', max_attempts)
```

So I'll leave it as-is, unless you feel that this isn't necessary

james-garner-canonical: Sounds good, I didn't notice that

examples/httpbin-demo/src/charm.py
dwilding: I thought about this more and decided to go with a variation on James's suggestion. For the "unexpected" errors, `pebble.APIError` and `ModelError`, catch the error then log the error and re-raise it.

```python

try:
    if not self.container.get_service(SERVICE_NAME).is_running():
        # We can connect to Pebble in the container, but the service isn't running.
        event.add_status(ops.MaintenanceStatus('waiting for workload'))
except ops.pebble.ConnectionError:
    # We can't connect to Pebble in the container.
    event.add_status(ops.MaintenanceStatus('waiting for workload container'))
except ops.pebble.APIError:
    # It's technically possible (but unlikely) for Pebble to have an internal error.
    logger.error('Unable to fetch service info from Pebble')
    raise
except ops.ModelError:
    # We asked for the status of a service that doesn't exist in Pebble.
    # If this error happens, it means there's a bug in the charm.
    logger.error("Unable to find service '%s' in service info", SERVICE_NAME)
    raise
```

It's possible for the collect-status handler to run before the charm has added a layer to Pebble (e.g., in the case of an invalid config value). So I've updated the unit tests to provide a mock layer to the container, which ensures that `get_service()` can always find the service.

dwilding: Please ignore my previous comment. I realised that the approach doesn't work. `ModelError` will be raised as part of normal operation because the workload doesn't have a default Pebble layer (it's not a rock). In other words, anytime the collect-status handler runs before the charm has added a layer, `get_service()` will be able to connect to Pebble, but there won't be any services yet, so we'll get `ModelError`.

I've changed the logic to this:

```python
try:
    if not self.container.get_service(SERVICE_NAME).is_running():
        # We can connect to Pebble in the container, but the service hasn't started yet.
        event.add_status(ops.MaintenanceStatus('waiting for workload'))
except ops.ModelError:
    # We can connect to Pebble in the container, but the service doesn't exist. This is
    # most likely because we haven't added a layer yet.
    event.add_status(ops.MaintenanceStatus('waiting for workload container'))
except ops.pebble.ConnectionError:
    # We can't connect to Pebble in the container. This is most likely because the
    # container hasn't started yet.
    event.add_status(ops.MaintenanceStatus('waiting for workload container'))
except ops.pebble.APIError:
    # It's technically possible (but unlikely) for Pebble to have an internal error.
    logger.error('Unable to fetch service info from Pebble')
    raise
```

And I've reverted the mock layer in the unit tests.

examples/httpbin-demo/src/charm.py
```diff
@@ -17,58 +17,83 @@
...
 class HttpbinDemoCharm(ops.CharmBase):
     """Charm the service."""
 
     def __init__(self, framework: ops.Framework):
         super().__init__(framework)
-        framework.observe(self.on['httpbin'].pebble_ready, self._on_httpbin_pebble_ready)
+        framework.observe(self.on.collect_unit_status, self._on_collect_status)
+        framework.observe(self.on[CONTAINER_NAME].pebble_ready, self._on_httpbin_pebble_ready)
         framework.observe(self.on.config_changed, self._on_config_changed)
+        self.log_level = str(self.config['log-level']).lower()  # str() is for the type checker.
```
dimaqq: @tonyandrewmeyer do we advocate for or against accessing config in `__init__`?

examples/httpbin-demo/src/charm.py
```diff
@@ -17,58 +17,83 @@
...
-        # Fetch the new config value
-        log_level = cast('str', self.model.config['log-level']).lower()
-
-        # Do some validation of the configuration option
-        if log_level in VALID_LOG_LEVELS:
-            # The config is good, so update the configuration of the workload
-            container = self.unit.get_container('httpbin')
-            # Push an updated layer with the new config
+        if self.log_level not in VALID_LOG_LEVELS:
+            return
```
dimaqq: **minor:**
Given that this is a delta charm, I'm not sure how I feel about the indirection of saving the log level in an instance attribute at init time only to read it later here.

examples/httpbin-demo/src/charm.py
```diff
@@ -17,58 +17,83 @@
...
+        event.add_status(ops.ActiveStatus())
 
     def _on_httpbin_pebble_ready(self, event: ops.PebbleReadyEvent):
         """Define and start a workload using the Pebble API."""
-        # Get a reference the container attribute on the PebbleReadyEvent
-        container = event.workload
-        # Add initial Pebble config layer using the Pebble API
-        container.add_layer('httpbin', self._pebble_layer, combine=True)
+        # Add initial Pebble config layer using the Pebble API.
+        self.container.add_layer('httpbin', self._pebble_layer, combine=True)
```
dimaqq: **issue:**

If this even fires when the log level is outside the allowed set, this code path will set the pebble layer and start the workload with bad log level value.

examples/httpbin-demo/src/charm.py
tonyandrewmeyer: Personally, I like charms doing as little as possible in `__init__`, but the Charm Tech position (from the discussion in Frankfurt) is that they can do anything they like.

There's going to be a follow-up to this PR that uses `load_config` instead, so my preference is still to leave things as they are and then finalise config access in that PR.

dwilding:
Per discussion at Charm Tech standup, I've added log-level validation at the top of the pebble-ready handler. I've also removed `self.log_level`. Instead, the logic for loading and validating log-level is duplicated everywhere it's needed. This is in anticipation of @tonyandrewmeyer switching to [`load_config`](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.CharmBase.load_config) after that's released this month.

One other adjustment: I changed `_pebble_layer` from a property to a normal method (now called `_make_pebble_layer`) so that it can take log-level as an argument.

dimaqq (approved):

dwilding:
Merging - thanks all for the help. We can make further improvements in follow-up PRs (some are already tracked in #1839 and #1805).

PR #1837, opened 2025-06-18
fix: only provide the units belonging to the app in Relation.units

### LINKED ISSUE #1109: remote unit data is inaccessible in relation-departed

tonyandrewmeyer:
The previous PR for #1109 wrongly added the `REMOTE_UNIT` unit to all relations, but it should only be ones where the app matches (and in particular, is never for peer relations). This fixes that, and adds an extra unit test for verification.

To be extra sure, an integration test is added that verifies that the correct units are returned. The test/integration conftest was quite specific to tracing, so this also makes it more generalised.

I'd like to have a few integration tests for ops, and the door has been opened with tracing, but I'm happy to discuss this if people feel it's not worth doing for this case (I had to more-or-less make it anyway to verify the fix). It does form quite a significant portion of the changes here.

Fixes #1109


test/integration/test_relation.py
```diff
@@ -0,0 +1,71 @@
...
+    ingress_units = set(status.get_units(ingress))
+    task = juju.run('test-relation/0', 'get-units')
+    assert task.success, task.message
+    ops_units = json.loads(task.results['units'])
+
+    assert all(
+        unit.startswith(f'{app}/')
+        for rel, app in [('db', db), ('ingress', ingress), ('peer', 'test-relation')]
+        for unit in ops_units[rel]
+    )
```
james-garner-canonical: One minor suggestion
```suggestion
    assert all(unit.startswith('db/') for unit in ops_units[db])
    assert all(unit.startswith('ingress/') for unit in ops_units[ingress])
    assert all(unit.startswith('peer/') for unit in ops_units['test-relation])
```

ops/model.py
```diff
@@ -1763,7 +1763,11 @@ def __init__(
 
         # In relation-departed `relation-list` doesn't include the remote unit,
         # but the data should still be available.
-        if _remote_unit is not None:
+        if (
+            _remote_unit is not None
+            and not is_peer
+            and _remote_unit.name.startswith(relation_name)
```
dimaqq: This looks weird to me.

I would have imagined that unit name and relation names to be incomparable.

Can you provide examples for these strings?

What about the slashes, as in `"application/0".startswith("app")` but these are different apps, so it should be something like `"application/0".startswith("app/")`.

tonyandrewmeyer: Ah, nice catch, thanks. Yes, the test was failing to catch this because the remote app name was set to the same name as the relation name. I've fixed the test and the code, and also handled the edge case.

ops/model.py
```diff
@@ -1777,6 +1772,15 @@ def __init__(
         self.app = typing.cast('Application', app)
         self.data = RelationData(self, our_unit, backend)
 
+        # In relation-departed `relation-list` doesn't include the remote unit,
+        # but the data should still be available.
+        if (
+            _remote_unit is not None
+            and not is_peer
+            and _remote_unit.name.startswith(f'{self.app.name}/')
```
dimaqq, line 1780: **curiosity:** is it possible to for an app to be related to itself over a non-peer relation anyway?

test/integration/conftest.py
```diff
@@ -29,64 +29,90 @@
...
-            access_key='accesskey',
-            secret_key='mysoverysecretkey',
-            secure=False,
-        )
 
-        found = mc_client.bucket_exists('tempo')
-        if not found:
-            mc_client.make_bucket('tempo')
+@pytest.fixture
+def tracing_juju(juju: jubilant.Juju) -> Generator[jubilant.Juju]:
```
dimaqq, line 40: **praise:** thank you for refactoring this file!

ops/model.py
tonyandrewmeyer: No, I don't think so.

test/integration/conftest.py
```diff
@@ -29,64 +29,90 @@
...
-    """Prepare and return the test charm directory.
+def tracing_charm_dir(pytestconfig: pytest.Config) -> Generator[pathlib.Path]:
+    """Prepare and return the test_tracing charm directory.
 
     Builds and injects `ops` and `ops-tracing` from the local checkout in to the
     charm's dependencies. Cleans up afterwards.
     """
-    charm_dir = pytestconfig.rootpath / 'test/charms/test_tracing'
+    charm_dir = pytestconfig.rootpath / 'test/charms/test_tracing'  # type: ignore
+    assert isinstance(charm_dir, pathlib.Path)
```
dimaqq, line 95: Weirdly I don't need `type: ignore` or `isinstance` check here.

If it's really needed, why not flip this like so?

```suggestion
    assert isinstance(pytestconfig.rootpath, pathlib.Path)
    charm_dir = pytestconfig.rootpath / 'test/charms/test_tracing'
```

tonyandrewmeyer: Maybe this is because of a newer pyright? This was done 3 months ago and I think we bumped the version since then.

No objections to opening a new PR that makes improvements here.

PR #1840, opened 2025-06-19
fix: in meter-status-changed JUJU_VERSION is not set

### LINKED ISSUE #1836: Ops assumes JUJU_VERSION is always set
This might be a bug in Juju itself, however we shouldn't be failing charms because of this, either, and we should deal with the fact that some versions of Juju will have this bug.

Specifically, in the `meter-status-changed` hook (which still exists in Juju 3.6, and probably used to be hard ignored in ops, but is now triggering with the latest ops), I see this in my logs:
```
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed Traceback (most recent call last):
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed   File "/var/lib/juju/agents/unit-ss6-19/charm/src/charm.py", line 27, in <module>
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed     ops.main(SimpleSubordinateOperatorCharm)
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed   File "/var/lib/juju/agents/unit-ss6-19/charm/venv/lib/python3.10/site-packages/ops/__init__.py", line 356, in __call__
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed     return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed   File "/var/lib/juju/agents/unit-ss6-19/charm/venv/lib/python3.10/site-packages/ops/_main.py", line 473, in main
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed     manager = _Manager(charm_class, use_juju_for_storage=use_juju_for_storage)
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed   File "/var/lib/juju/agents/unit-ss6-19/charm/venv/lib/python3.10/site-packages/ops/_main.py", line 280, in __init__
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed     juju_context = _JujuContext.from_dict(os.environ)
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed   File "/var/lib/juju/agents/unit-ss6-19/charm/venv/lib/python3.10/site-packages/ops/jujucontext.py", line 229, in from_dict
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed     version=JujuVersion(env['JUJU_VERSION']),
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed   File "/usr/lib/python3.10/os.py", line 680, in __getitem__
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed     raise KeyError(key) from None
unit-ss6-19: 10:22:10 WARNING unit.unit-ss6-19.meter-status-changed KeyError: 'JUJU_VERSION'
```

I'm guessing the issue is that we used to see that "meter-status-changed" was being called, which was one of the "restricted context" hooks, and used to hard nope in ops. But now that it is still moving forward, we see that JUJU_VERSION isn't getting set in that context, and then hard dying because we assume the environment variable is always set.

This was with Juju 3.6.7 and ops 2.22.0 according to my uv.lock file.

tonyandrewmeyer:
Despite the [documentation](https://documentation.ubuntu.com/juju/3.6/reference/hook/) saying that `JUJU_VERSION` is set in the environment for all hooks, it is not for the `meter-status-changed` or `collect-metrics` events.

Restore the older behaviour (prior to #1313) where if `JUJU_VERSION` is not available we set a version of `0.0.0`. We can drop this once ops requires Juju 4 and above.

In addition, in the restricted context the `is-leader` hook tool is not available, which means that collect-status fails (before emitting any events, because it can't figure out which events to emit). We could just do the unit status in this case, but it seems simpler to just skip collect status in this case (which is only for two events no-one should be using any more).

Finally, the PR adds `meter-status-changed` to the list of restricted context events, which means that we won't run any deferred events during those hooks (as with `collect-metrics`), and will error out cleanly if `use_juju_for_storage` is used. Not emitting deferred events is a breaking change, but definitely should be done - it's extremely likely that the handler would crash if unexpectedly running in a restricted context. Similarly, for the `use_juju_for_storage` case, the charm would have less gracefully crashed anyway, when ops tried to use storage itself.

Fixes #1836

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -803,7 +803,6 @@ class _TestMain(abc.ABC):
             VERSION_LOGLINE,
             ['juju-log', '--log-level', 'DEBUG', '--', 'Emitting Juju event collect_metrics.'],
             ['add-metric', '--labels', 'bar=4.2', 'foo=42'],
-            ['is-leader', '--format=json'],
         ]
         calls = fake_script.calls()
 


tonyandrewmeyer:
@jameinel feel free to also review if you'd like, but no problem just leaving it for the others :)

tonyandrewmeyer:
The observability failure looks unrelated. I'll look into that separately - maybe an upstream change?

ops/_main.py
```diff
@@ -413,8 +413,11 @@ def _emit(self):
 
         # Emit the Juju event.
         self._emit_charm_event(self.dispatcher.event_name)
-        # Emit collect-status events.
-        _charm._evaluate_status(self.charm)
+        # Emit collect-status events. In a restricted context, we can't run
+        # is-leader, so can't do the full evaluation. Skip it rather than
+        # only running the unit status.
+        if not self.dispatcher.is_restricted_context():
+            _charm._evaluate_status(self.charm)
```
james-garner-canonical, line 420: I'm a little concerned about skipping `collect-status` in cases where it might have previous been executed without error. Is this not actually an issue, or should we investigate further whether `collect_metrics` would have errored out here?

I'm also a bit concerned even if it would have errored out, if that error would have happened *after* the charm's handler executed. If the error would have been after running the user's code, when trying to actually set the status, it seems like it might be better for backwards compatibility to move this check till then -- so the charm's collect status handler would execute, but we'd either swallow setting the status at all, or gracefully handle the `is-leader` failure.

tonyandrewmeyer: > I'm a little concerned about skipping collect-status in cases where it might have previous been executed without error. Is this not actually an issue, or should we investigate further whether collect_metrics would have errored out here?

With current ops, collect-metrics fails:

```
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics Traceback (most recent call last):
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics   File "/var/lib/juju/agents/unit-logenv-2/charm/./src/charm.py", line 33, in <module>
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics     ops.main(LogenvCharm)
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics   File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/__init__.py", line 356, in __call__
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics     return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics   File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/_main.py", line 473, in main
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics     manager = _Manager(charm_class, use_juju_for_storage=use_juju_for_storage)
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics   File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/_main.py", line 280, in __init__
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics     juju_context = _JujuContext.from_dict(os.environ)
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics   File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/jujucontext.py", line 229, in from_dict
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics     version=JujuVersion(env['JUJU_VERSION']),
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics   File "/usr/lib/python3.10/os.py", line 680, in __getitem__
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics     raise KeyError(key) from None
unit-logenv-2: 16:28:11 WARNING unit.unit-logenv-2.collect-metrics KeyError: 'JUJU_VERSION'
```

So the charm's handler wouldn't run. That would work in older ops before we required `JUJU_VERSION`.

With the `JUJU_VERSION` fix, it crashes still:

```
unit-logenv-2: 16:31:41 ERROR unit.unit-logenv-2.juju-log Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/model.py", line 3422, in _run
    result = subprocess.run(args, **kwargs)  # type: ignore
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-logenv-2/is-leader', '--format=json')' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

  File "/var/lib/juju/agents/unit-logenv-2/charm/./src/charm.py", line 33, in <module>
    ops.main(LogenvCharm)
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/__init__.py", line 356, in __call__
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/_main.py", line 475, in main
    manager.run()
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/_main.py", line 459, in run
    self._emit()
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/_main.py", line 405, in _emit
    _charm._evaluate_status(self.charm)
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/charm.py", line 1426, in _evaluate_status
    if charm.framework.model._backend.is_leader():
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/model.py", line 3562, in is_leader
    is_leader = self._run('is-leader', return_output=True, use_json=True)
  File "/var/lib/juju/agents/unit-logenv-2/charm/venv/ops/model.py", line 3424, in _run
ops.model.ModelError: ERROR leadership status unknown: not implemented for restricted context not implemented
```

I still get this error running `collect-metrics`:

```
$ juju collect-metrics logenv/2
failed to collect metrics: could not read stderr
```

But I am able to get the event to run, and have metrics show up in `juju metrics`. And the error definitely applies there.

> I'm also a bit concerned even if it would have errored out, if that error would have happened after the charm's handler executed. If the error would have been after running the user's code, when trying to actually set the status, it seems like it might be better for backwards compatibility to move this check till then -- so the charm's collect status handler would execute, but we'd either swallow setting the status at all, or gracefully handle the is-leader failure.

It's not `status-set` that fails (well, it might fail as well - I didn't check). It's `is-leader`. That's the [first line](https://github.com/canonical/operator/blob/ff47e4100aa79ec1705617db380ebd0bb4d2953e/ops/charm.py#L1592) of evaluating the status, well before the events are emitted. So if the charm has had no chance to run code in the collect-status handlers.

james-garner-canonical: Thanks for the thorough investigation! Since the charm has no chance to run the collect status code, I think you've made the right call with the current changes here.

james-garner-canonical (approved):

dimaqq:
**Discussion:** (from Tony and Dima's 1-1)

1. If this has been broken since `ops==2.5.0` (or so), should we fix it at all?
   * Tony reports that dispatch failures for these events are ignored by the unit agent
   * `ops.tracing.Tracing()` charm lib always runs `is_leader()` in `__init__`, guaranteeing an error exit
   * The `CollectStatus` event mechanics calls `is_leader()` which probably results in error exit 
3. Would it be better to exit early on these events and not let the reconciler charms enter `__init__`?
4. Is `0.0.0` actually a good default? Our Juju version feature detection would be all off.
5. Would it be better to remember last known Juju version and set that instead?

dimaqq:
@jnsgruk please add your thoughts ☝🏻

dimaqq:
Poke @jameinel who brought this up orignially.

tonyandrewmeyer:
> **Discussion:** (from Tony and Dima's 1-1)
> 
> 1. If this has been broken since `ops==2.5.0` (or so), should we fix it at all?
>    
>    * Tony reports that dispatch failures for these events are ignored by the unit agent
>    * `ops.tracing.Tracing()` charm lib always runs `is_leader()` in `__init__`, guaranteeing an error exit
>    * The `CollectStatus` event mechanics calls `is_leader()` which probably results in error exit
> 2. Would it be better to exit early on these events and not let the reconciler charms enter `__init__`?
> 3. Is `0.0.0` actually a good default? Our Juju version feature detection would be all off.
> 4. Would it be better to remember last known Juju version and set that instead?

Some follow-up on this:

* I think we should fix this, and we should fix it in 2.x not wait for 3.x, because it's mostly relevant to older charms/Jujus/bases. If it's in the final 2.x then older charms can upgrade from a <2.5 ops to the latest ops, getting various other fixes and improvements, even if they are stuck on older bases. I also feel like it's just wrong that we crash on these events even though it doesn't send the unit to error.
* It's true that this ends up changing behaviour, because if your charm was previously getting the metrics events it was crashing before and now won't -- so a charm that does the same code for every event will have that code run for two additional events rather than crashing. However, I don't think the charms that are written in that style are using these metric events, and they should also be aware of the risk of additional events and be able to handle that. If they can't handle running the reconcile method in restricted context events then they should either not support them or early exit in those cases (e.g. check if you get a `ModelError` checking for leadership).
* Even if Juju 3.6.x stops emitting these, I think it's worthwhile to have it fixed in ops, because we definitely have ops charms that are needing to support older versions of Juju.
* I can see the argument that you can't explicitly observe `meter-status-changed` so ops should exit on that event. However, this feels like a much more significantly breaking change to me. For `collect-metrics`, you can observe that, so I very strongly don't want us to treat its `__init__` differently than other events.

I don't think there's a 'correct' answer to "what is the fallback Juju version" -- that's one of the reasons it was so nice to (incorrectly) see that it was always provided so we could always set it to what Juju gave us. `0.0.0` has the advantage that you're never going to try to use a feature you don't have, and that it's what we had previously. It's only relevant in these metrics events, so I think that wrongly saying features aren't available is probably ok, because lots of features aren't available. I think using 2.9 or 3.6 would be too risky, and it's not worth the overhead of recording the version (even though that would be most likely correct) just for these events.

PR #1842, opened 2025-06-20
feat: expose trace data in testing

dimaqq:
This PR exposes `Context.trace_data` in Scenario similar to `Context.juju_log`.

The intention is to enable charm and charm library authors to validate custom, meaningful traces in unit tests.

Check list:
- [x] code update
- [x] minimal tests
- [x] new unit test for scenario itself

Out of scope:
- doc update: I'll make a separate PR for that


dimaqq:
@tonyandrewmeyer an early review would be welcome.

It took me a while to wire this all up, so you keen eye is particularly needed in that area.

testing/src/scenario/_ops_main_mock.py
```diff
@@ -270,6 +272,28 @@ def _close(self):
...
+            # FIXME: think if this is worth the effort:
+            # OpenTelemetry SDK stores spans chronologically by end time.
+            # This means that `ops._main` span is last.
+            #
+            # We'd want to offer trace data in a consistent order,
+            # ideally chronologically by start_time.
+            # However, a unit test could be run with time frozen.
+            #
+            # `reversed` ensures that outermost spans appear earlier...
+            # but inverts the order of consecutive spans...
```
dimaqq: I'm thinking that charmers would realistically want to find their spans by name, not by order in the list.

They may be interested in span A is parent of B (by unique ids), or span B contained within A (by timestamps) relationships.

They could be also interested in charm span A happened within the context of ops span E which is indicative of responding to a specific emitted event.

Absolute position would be an anti-patterns, because a change in Ops may "move" some charm span.

So, perhaps, helper functions are the way to go, rather than offering trace data in a defined order.

testing/src/scenario/_runtime.py
```diff
@@ -345,6 +347,9 @@ def exec(
                 raise UncaughtCharmError(f'Uncaught {type(e).__name__} in charm, try "exceptions [n]" if using pdb on Python 3.13+. Details: {e!r}') from e  # fmt: skip
 
             finally:
+                if ops:
+                    ops._destroy()
```
tonyandrewmeyer: This makes me think that `destroy` should be public (within the private `Ops` and `_Manager` classes).

testing/src/scenario/_ops_main_mock.py
```diff
@@ -270,6 +272,28 @@ def _close(self):
...
+
+            # Minor discussion: what it takes to type-check cleanly.
+            #
+            # .start_time is always set, but it's declared nullable.
+            self.trace_data = sorted(reversed(data), key=lambda s: s.start_time or 0)
+            from operator import attrgetter
+
+            # this option type checks cleanly though
+            self.trace_data = sorted(reversed(data), key=attrgetter('start_time'))
```
tonyandrewmeyer: I've always personally quite liked using the `operator` methods, and I think here avoiding the 'or 0' is nice.

I do get the point about the `lambda` version being understandable without needing to know about the `operator` module, though. So I'm slightly in favour of the latter, but don't object to the former.

testing/src/scenario/_ops_main_mock.py
tonyandrewmeyer: I'm not an expert here, but I agree that it seems like ordering wouldn't be something that a charm test would typically be checking (the ops[tracing] tests themselves might, but charms should be able to trust that the ordering makes sense).

I think that if temporal checks are wanted ('contains', 'overlaps', etc) then helpers are indeed the way to go.

dimaqq:
@tonyandrewmeyer do you reckon there's something we could realistically do about this:

```py
(Pdb) [s.name for s in ctx.trace_data]
['start: Tracing', 'start: Runtime._wrap.<locals>.WrappedCharm', 'collect_unit_status: Runtime._wrap.<locals>.WrappedCharm', 'ops.main']
```

The span names that ops issues are inconsistent between prod and Scenario.

Do you reckon it's realistic to hack the unit tests runtime wrap logic to preserve the charm name?

dimaqq:
☝🏻 https://github.com/dimaqq/operator/pull/64

tonyandrewmeyer:
> @tonyandrewmeyer do you reckon there's something we could realistically do about this:
> 
> ```python
> (Pdb) [s.name for s in ctx.trace_data]
> ['start: Tracing', 'start: Runtime._wrap.<locals>.WrappedCharm', 'collect_unit_status: Runtime._wrap.<locals>.WrappedCharm', 'ops.main']
> ```
> 
> The span names that ops issues are inconsistent between prod and Scenario.

I wondered about pulling the code from #1631, which solves the same problem that Scenario is working around in a perhaps cleaner way (unregistering the things that get registered, rather than moving everything one level lower in the hierarchy). However, I think that's probably not worth doing, at least within the framework code -- it's a bunch of extra code that would only be used for testing. It could be in Scenario (a Framework subclass perhaps - and I do want to grab the `CapturingFramework` out of that PR and open a fresh PR just for that).

> Do you reckon it's realistic to hack the unit tests runtime wrap logic to preserve the charm name?

I think this is probably fine to do now, given the context. Scenario is already trying to pretend that it's the same charm, so I don't think anyone would be confused by the name being the same.

testing/src/scenario/_runtime.py
dimaqq: done, thanks!

testing/src/scenario/_ops_main_mock.py
dimaqq: This is no longer relevant, as I've decided not to order the spans.

dimaqq:
@dwilding the `ReadableSpan` business is because it's a type hint, thus imported only `if TYPE_CHECKING`, which we need as I don't want to put hard dependency on the OTEL SDK in Scenario itself.

testing/src/scenario/_ops_main_mock.py
```diff
@@ -127,14 +127,16 @@ def __init__(
         self.context = context
         self.charm_spec = charm_spec
         self.store = None
+        self.trace_data = []
```
tonyandrewmeyer, line 130: One belated question here: do we definitely want to bind ourselves to using the otel object here? Mostly (but not entirely) in Scenario we create classes that live in `testing` and wrap/mock external classes. We could do that here - have a `testing.Trace` class that had all/most of the `ReadableSpan` properties and convert to those when this list has items added to it.

In general, I like avoiding tying ourselves to specific external classes. However, in this case, ops itself has already done that, so perhaps there's no point being more generic in Scenario.

However, if there's something about the `ReadableSpan` class that you don't like, or functionality you would like it to have, that would be a motivator for doing this.

dimaqq: My thinking was that it's better to expose the data in OTEL format.

We could wrap or replace the OTEL types, depending on the level of detail we want to expose.

OTEL spans in memory comprise these fat objects:
- ReadableSpan
- SpanContext
- Resource
- Event
- InstrumentationInfo
- InstrumentationScope
- Link (we don't support these yet)

Enums:
- Kind
- Status

As well as a bunch of misc. fields:
- values for trace_flags enum, as ints
- is_remote, is_valid and other helper properties
- timestamps, int nanoseconds since epoch
- attributes as Python dicts

Replicating all these seems like quite a bit of work for relatively little gain.

We could serialise each span (native functionality), or in fact a collection of spans (same code as exporter) into JSON (or py dicts) and get rid of all the custom types.

In that case, our types are clean (Python dicts and lists), but unit test authors will carry more burden:
- no IDE help when writing assertions
- no type hints to validate tests themselves
- minor: more work to validate containment
- minor: the data shape is weirder in the case of a collection of spans

dimaqq: At standup: let's export the OTEL types, but pair that with good advice in the doc PR, hoping that charmers will copy-paste our examples and change only span and attribute names.

testing/tests/test_e2e/test_trace_data.py
```diff
@@ -0,0 +1,53 @@
...
+    'name': 'traced_charm',
+    'requires': {'charm-tracing': {'interface': 'tracing', 'limit': 1}},
+}
+
+
+class TracedCharm(ops.CharmBase):
+    META = {
+        'name': 'traced_charm',
+        'requires': {'charm-tracing': {'interface': 'tracing'}},
+    }
```
james-garner-canonical: Is this used?

dimaqq: Evidently not! It would not work without `limit: 1`. Good catch!

dimaqq: fixed!

james-garner-canonical (approved):

james-garner-canonical:
No changes to request.

I was a bit hesitant over whether exposing the `opentelementary` types is the right call, but I'm happy with it on balance. Type hints for `Context.trace_data` when `ops[testing]` is installed but `ops[tracing]` isn't show as `list[Unknown]`, but this seems fine since charms should only be accessing `trace_data` if they're using `tracing`.

I haven't got my head fully around all the details of the tracing/testing machinery at play, but the intent is clear and the tests are clean and validate the intent.

Look good to me :)

PR #1850, opened 2025-06-25
ci: Add contributor license agreement check

tonyandrewmeyer:
It seems like the `canonical` organisation is now forcing all repositories to have a passing CLA check, so check if adding that workflow here solves that issue.

This is the workflow that the action recommends, with minor security improvements.


james-garner-canonical:
Looks good to me, and the `cla-check` passes. Resolves #1851

james-garner-canonical (approved):

.github/workflows/cla.yaml
```diff
@@ -0,0 +1,14 @@
...
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  cla-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if CLA signed
+        uses: canonical/has-signed-canonical-cla@5d1443b94417bd150ad234a82fe21f7340a25e4d  # v2.1.0
```
dimaqq, line 14: **minor:**
I wonder if we could trust the `canonical` org and reference this action by branch like `@v2`

dimaqq (approved):

.github/workflows/cla.yaml
```diff
@@ -0,0 +1,14 @@
+name: cla-check
+
+on:
+  pull_request:
+  workflow_dispatch:
```
dimaqq, line 5: This seems a bit odd... what would be checked?

Perhaps authors of the last NN commits?

.github/workflows/cla.yaml
james-garner-canonical: I'd be on board with an entry for `canonical` to the [.github/zizmor.yml policies](https://github.com/canonical/operator/blob/main/.github/zizmor.yml#L4) being added in this PR.

.github/workflows/cla.yaml
tonyandrewmeyer: I was thinking it would be useful if it failed, then the contributor signed the CLA, and we could run it manually. But maybe we can just re-run the failed check, so that would be sufficient.

.github/workflows/cla.yaml
tonyandrewmeyer: I wondered about that. You'd hope that we could. It feels pretty broad, but I guess it's sending the wrong message if we don't :)

tonyandrewmeyer:
It looks like maybe an org level one is meant to run. I've asked in MM for information. Maybe it's failing to run unless we duplicate it? That would be weird. I see that pattern in other repos, though.

tonyandrewmeyer:
Apparently [this is being run at an org level](https://chat.canonical.com/canonical/pl/ift588iejb8hfcprsf7yomwmnw) so no need to add it ourselves. I assume they will sort out the issue with it not running.

PR #1852, opened 2025-06-26
docs: use load_config in the httpbin example charm

### LINKED ISSUE #1839: Switch to load_config in httpbin demo charm
To keep our example charms fresh on the latest Ops features, we should update the httpbin charm to use the upcoming `load_config` feature to validate the config option (as documented in [Manage configuration](https://ops.readthedocs.io/en/latest/howto/manage-configuration.html)).

For more details, see:
* https://github.com/canonical/operator/pull/1833#discussion_r2153217277
* https://github.com/canonical/operator/pull/1833#discussion_r2153220879

When working on this, don't forget to bump the minimum Ops version in requirements.txt.

tonyandrewmeyer:
Adds a basic config class to the httpbin example charm, and uses `load_config` rather than `config`. The class is a stdlib dataclass rather than a `pydantic.BaseModel` because it seems pointless to require the large `pydantic` dependency for such a small use-case, where there's no value added other than recommending Pydantic.

Fixes #1839


tonyandrewmeyer:
Leaving in draft until the required version of ops is released.

examples/httpbin-demo/requirements.txt
```diff
@@ -1 +1 @@
-ops>2,<4
+ops>2.23,<4
```
james-garner-canonical:
```suggestion
ops>=2.23,<4
```

dwilding: Should we wait for Ops 3 and do the same as we decided for the Charmcraft profiles?

```suggestion
ops>=3,<4
```

dwilding (approved):
Nice one, thanks!

examples/httpbin-demo/requirements.txt
tonyandrewmeyer: I don't think there's any need to block this charm working on Ops 2.x, but happy to hear arguments otherwise.

james-garner-canonical: I'm in favour of keeping it working with `2.23` until there's a reason not to.

dwilding: I don't have a strong argument :smile: My only argument would be, I look at 2.23 and don't really understand why we picked this particular version. It's not the latest (at least, it won't be in the future) and it's not allowing earlier version of Ops 2. So it feels slightly confusing, which I think we ought to avoid for a demo

Would you be OK if we add a comment? Something like this:

```
ops>=2.23,<4  # Version 2.23 introduced the load_config() method.
```

PR #1855, opened 2025-06-26
feat: release automation script

IronCore864:


IronCore864:
## Test in a mock repo

### Existing old release

old-release

### Release

```bash
$ python release.py
ℹ️ INFO: Latest version in branch "main"": 0.1.0
ℹ️ INFO: Suggested new version: 0.1.1
Input version to release (press enter to use suggested version 0.1.1), or 'c' to cancel:
⚠️ WARNING: Check out the releases page: https://github.com/IronCoreWorks/release-automation/releases before confirming!
Confirm creating release '0.1.1' on branch 'main'? [y/N]: y
ℹ️ INFO: Draft release created: https://github.com/IronCoreWorks/release-automation/releases/tag/untagged-6661b7012148818ad142.
================================================================================
Formatted release notes:
================================================================================
## What's Changed

### Features
* Add changes.md in https://github.com/IronCoreWorks/release-automation/pull/6

### Documentation
* Add version in https://github.com/IronCoreWorks/release-automation/pull/3

### Tests
* Add test in https://github.com/IronCoreWorks/release-automation/pull/4

**Full Changelog**: https://github.com/IronCoreWorks/release-automation/compare/0.1.0...0.1.1
================================================================================
The automatically generated title is: 0.1.1
Enter release title, press Enter to keep the auto-generated title:
> 0.1.1 Test Release

Enter release summary (multi-line supported, type '.' on a new line to finish):
The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog.
.
ℹ️ INFO: Release title and notes updated.
ℹ️ INFO: Updated CHANGES.md with new release notes.
ℹ️ INFO: Updated ops/version.py to release version: 0.1.1
ℹ️ INFO: Updated testing/pyproject.toml to release version: 0.1.1
ℹ️ INFO: Updated tracing/pyproject.toml to release version: 0.1.1
Switched to a new branch 'release-prep-0.1.1'
[release-prep-0.1.1 fd5967f] chore: prepare release 0.1.1
 4 files changed, 17 insertions(+), 3 deletions(-)
            Enumerating objects: 16, done.
Counting objects: 100% (16/16), done.
Delta compression using up to 10 threads
Compressing objects: 100% (6/6), done.
Writing objects: 100% (9/9), 1.80 KiB | 1.80 MiB/s, done.
Total 9 (delta 4), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (4/4), completed with 4 local objects.
        remote:
remote: Create a pull request for 'release-prep-0.1.1' on GitHub by visiting:
remote:      https://github.com/IronCore864/release-automation/pull/new/release-prep-0.1.1
remote:
To github.com:IronCore864/release-automation.git
 * [new branch]      release-prep-0.1.1 -> release-prep-0.1.1
                                                                                
ℹ️ INFO: Created PR: https://github.com/IronCoreWorks/release-automation/pull/10
```

### Automated new release

new-release

### Automated release PR

release pr

### Post-release

```bash
$ python release.py --post-release
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
Unpacking objects: 100% (1/1), 907 bytes | 453.00 KiB/s, done.
From github.com:IronCoreWorks/release-automation
   fa80ad9..a4d32b4  main       -> upstream/main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
Updating fa80ad9..a4d32b4
Fast-forward
 CHANGES.md             | 14 ++++++++++++++
 ops/version.py         |  2 +-
 testing/pyproject.toml |  2 +-
 tracing/pyproject.toml |  2 +-
 4 files changed, 17 insertions(+), 3 deletions(-)
ℹ️ INFO: Updated ops/version.py to post-release version: 0.1.2.dev0
ℹ️ INFO: Updated testing/pyproject.toml to post-release version: 0.1.2.dev0
ℹ️ INFO: Updated tracing/pyproject.toml to post-release version: 0.1.2.dev0
Switched to a new branch 'post-release'
[post-release 5d204b5] chore: post release
 3 files changed, 3 insertions(+), 3 deletions(-)
Enumerating objects: 14, done.
Counting objects: 100% (14/14), done.
Delta compression using up to 10 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (8/8), 1.75 KiB | 1.75 MiB/s, done.
Total 8 (delta 3), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
remote:
remote: Create a pull request for 'post-release' on GitHub by visiting:
remote:      https://github.com/IronCore864/release-automation/pull/new/post-release
remote:
To github.com:IronCore864/release-automation.git
 * [new branch]      post-release -> post-release
ℹ️ INFO: Created PR: https://github.com/IronCoreWorks/release-automation/pull/11
```

### Post-release PR

post release pr

release.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def format(self, record: logging.LogRecord):
+        """Format the log record based on its level."""
+        if record.levelno == logging.INFO:
+            self.setFormatter(logging.Formatter('ℹ️ INFO: %(message)s'))  # noqa: RUF001
+        elif record.levelno == logging.WARNING:
+            self.setFormatter(logging.Formatter('⚠️ WARNING: %(message)s'))
+        elif record.levelno == logging.ERROR:
+            self.setFormatter(logging.Formatter('❌ ERROR: %(message)s'))
+        return super().format(record)
```
dimaqq: That's a very roundabout way of doing this.

The handler's formatter is bounced around on each record.

If you are determined to go this way, there must be an `else:` clause to support other, unlisted log levels.

I would recommend setting a **custom formatter** on the **stock handler** instead.

Here's one example: https://github.com/dimaqq/python-libjuju/blob/cd5930c85d0f8bf5417bc9394d2c767ecc6e5e85/example.py#L63-L93

release.py
```diff
@@ -0,0 +1,498 @@
...
+
+
+logger.addHandler(MyHandler())
+
+if 'GITHUB_TOKEN' not in os.environ:
+    logger.critical('Environment variable GITHUB_TOKEN not set.')
+    exit(1)
+
+OPS_VERSION_STR = r'version: str = \'(\d+\.\d+\.\d+(?:\.dev\d+)?)\''
```
dimaqq: I would recommend start and end of line markers to be more precise.

```suggestion
OPS_VERSION_STR = r'^version: str = \'(\d+\.\d+\.\d+(?:\.dev\d+)?)\'$'
```

`re.xxx` calls may need the `re.MULTILINE` flag.

Without these markers, other matches are possible, for example:

```py
testing/src/scenario/_runtime.py
62:        juju_version: str = '3.0.0',
```

release.py
```diff
@@ -0,0 +1,498 @@
...
+    changes = format_changes(categories, tag)
+    update_changes_file(changes, 'CHANGES.md')
+
+    update_versions_for_release(tag)
+
+    new_branch = f'release-prep-{tag}'
+    subprocess.run(['/usr/bin/git', 'checkout', '-b', new_branch], check=True)
+    changed_files = ['CHANGES.md', *VERSION_FILES]
+    for file in changed_files:
+        subprocess.run(['/usr/bin/git', 'add', file], check=True)
```
dimaqq: I think that we need to validate that each file was in fact changed.
`git add` returns OK status even if there are no changes.

Perhaps it's better to fail early in the `update_versions_**` or `update_ops/pyproject_version` functions.

release.py
```diff
@@ -0,0 +1,498 @@
...
+    new_branch = 'post-release'
+    subprocess.run(['/usr/bin/git', 'checkout', '-b', new_branch], check=True)
+    for file in VERSION_FILES:
+        subprocess.run(['/usr/bin/git', 'add', file], check=True)
+    subprocess.run(['/usr/bin/git', 'commit', '-m', 'chore: post release'], check=True)
+    subprocess.run(['/usr/bin/git', 'push', 'origin', new_branch], check=True)
+    pr = repo.create_pull(
+        title='Post Release',
+        body='This PR updates the version files after the release.',
+        head=f'{gh_client.get_user().login}:{new_branch}',  # "your_username:new_branch"
```
dimaqq: Corner cases:
- `post-release` branch exists locally
- `post-release` branch exists on GitHub but not locally

Perhaps the script could take a stance here--either fail, on `check=True` (current behaviour) or overwrite local and remote branches with this name, or generate a random name, like `f"post-release-{some_hex}"` (maybe secrets.token_hex() or in fact current git commit)

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 1
+revision = 2
```
dimaqq: **Off-topic:** we all need to upgrade our `uv`

dimaqq (changes requested):
Early review

uv.lock
dimaqq: Ah, nevermind.,
I get the same lock version bump if I remove the file and re-lock.
It seems that uv keeps current format if it can, and the only difference between the formats is the addition of the PYPI timestamp.

release.py
```diff
@@ -0,0 +1,848 @@
...
+
+    subprocess.run(['/usr/bin/git', 'checkout', '-b', new_branch], check=True)
+    for file in VERSION_FILES.values():
+        subprocess.run(['/usr/bin/git', 'add', file], check=True)
+    subprocess.run(
+        ['/usr/bin/git', 'commit', '-m', 'chore: update versions after release'], check=True
+    )
+    subprocess.run(['/usr/bin/git', 'push', 'origin', new_branch], check=True)
+    pr = repo.create_pull(
+        title='Post Release',
```
tonyandrewmeyer: This will have to be something like `chore: adjust versions after release` so that it fits with the conventional commits.

release.py
```diff
@@ -0,0 +1,848 @@
...
+import github.GitRelease
+import github.Repository
+
+
+class SymbolFormatter(logging.Formatter):
+    """Custom logging formatter to use symbols for log levels."""
+
+    from typing import ClassVar
+
+    levels: ClassVar[dict[str, str]] = {
```
james-garner-canonical: My suggestion here would be to just define this in the module scope (e.g. `_LOG_LEVELS = {...`), no type annotation required.

If you want to keep it in the class body, then I think that using `Mapping[str, str]` would be cleaner than the `ClassVar`. The `from typing import ...` line should go at the top of the script with the other imports.

release.py
```diff
@@ -0,0 +1,848 @@
...
+from datetime import datetime
+from pathlib import Path
+
+import github
+import github.GitRelease
+import github.Repository
+
+
+class SymbolFormatter(logging.Formatter):
+    """Custom logging formatter to use symbols for log levels."""
```
tonyandrewmeyer: Do we need this? Isn't it sufficient to just use a simple formatter?

release.py
```diff
@@ -0,0 +1,848 @@
...
+    )
+    args = parser.parse_args()
+
+    if args.publish:
+        publish_draft_release(owner=args.owner, repo_name=args.repo)
+        logger.info(
+            'Draft release published. Please run this script with --post-release '
+            'to update the version files.'
+        )
+        exit(0)
```
tonyandrewmeyer: I'm not that convinced we want to have a script do this. Is there any advantage over opening the release in GitHub and pushing the button there? I think we should retain a human element, and this means someone does read the content, double-check the settings, and so on, and can easily tweak things if necessary. If everything is ok, it's just opening one page and pushing one button, arguably less work than running the script.

release.py
```diff
@@ -0,0 +1,848 @@
...
+    subprocess.run(['/usr/bin/git', 'push', 'origin', new_branch], check=True)
+    pr = repo.create_pull(
+        title='chore: adjust versions after release',
+        body='This PR updates the version files after the release.',
+        head=f'{gh_client.get_user().login}:{new_branch}',  # "your_username:new_branch"
+        base=branch,
+    )
+    logger.info(f'Created PR: {pr.html_url}')
+    countdown('Opening the PR in: ', 3)
+    webbrowser.open(pr.html_url)
```
tonyandrewmeyer: Do we need to do this? The automation is creating the PR, so why do I need to look at it in a browser? The reviewers need to, but that's not me.

release.py
```diff
@@ -0,0 +1,848 @@
...
+    subprocess.run(['/usr/bin/git', 'push', 'origin', new_branch], check=True)
+    pr = repo.create_pull(
+        title=f'chore: update changelog and versions for {tag} release',
+        body=f'This PR prepares the release of version {tag}.',
+        head=f'{gh_client.get_user().login}:{new_branch}',  # "your_username:new_branch"
+        base=branch,
+    )
+    logger.info(f'Created PR: {pr.html_url}')
+    countdown('Opening the PR in: ', 3)
+    webbrowser.open(pr.html_url)
```
tonyandrewmeyer: I can see more benefit here in opening it, in that you might want to tweak things, but personally I would rather that the script didn't open a browser and if I want to do that I click the link.

release.py
```diff
@@ -0,0 +1,848 @@
...
+        'docs': [],
+        'test': [],
+        'ci': [],
+    }
+    full_changelog_line = None
+
+    for line in release_notes.splitlines():
+        line = line.strip()
+        if line.startswith('* '):
+            match = re.match(r'\* (\w+): (.*) by @\w+ in (.*)', line)
```
tonyandrewmeyer: I think we want to keep (in the release notes, not so much the changelog) acknowledgement of external contributions, don't we?

release.py
```diff
@@ -0,0 +1,848 @@
...
+        logger.info('Please merge "update charm pins" PRs first:')
+        pr = open_prs[0]
+        pr_url = f'#{pr.number} - {pr.html_url}'
+        logger.info(f'  {pr_url}')
+        if len(open_prs) > 1:
+            logger.info(
+                f'Note that there are {len(open_prs) - 1} more open update charm pins PRs.'
+            )
+        countdown('Exiting and opening the charm pins PR in: ', 3)
+        webbrowser.open(pr.html_url)
```
tonyandrewmeyer: Also here, I would rather just click the link. But if everyone else likes their browser being opened then I can live with that.

release.py
```diff
@@ -0,0 +1,848 @@
...
+
+        if confirm == 'y':
+            break
+
+        logger.info("Let's try again...")
+
+    return new_version
+
+
+def get_new_scenario_version_post_release() -> str:
```
tonyandrewmeyer: I would be fine if we removed this and made it `{ops_major+5}.{ops_minor}.{ops.bugfix}`.

Or if we are ok with ops-scenario bumping a major version without ops, then I think we should handle that for ops-tracing too. But I think we are sticking with no major version changes for now?

release.py
```diff
@@ -0,0 +1,848 @@
...
+
+    return new_tag
+
+
+def create_draft_release(
+    repo: github.Repository.Repository, tag: str, branch: str
+) -> github.GitRelease.GitRelease | None:
+    """Create a draft release with auto-generated notes."""
+    max_retries = 3
+    for attempt in range(1, max_retries + 1):
```
tonyandrewmeyer: Why is this needed? We don't have the retries everywhere else - is creating a release unreliable?

tonyandrewmeyer (commented):
A few comments, but in general I think the primary test is a release, and the last two, which used this, seemed fine, and if we think there should be tweaks as we each use this to do a release, doing it at the time seems fine.

release.py
```diff
@@ -0,0 +1,848 @@
...
+VERSION_FILES = {
+    'ops/src': 'ops/version.py',
+    'ops/pyproject': 'pyproject.toml',
+    'testing': 'testing/pyproject.toml',
+    'tracing': 'tracing/pyproject.toml',
+    'uvlock': 'uv.lock',
+}
+
+
+def get_latest_release_tag(repo: github.Repository.Repository, branch_name: str) -> str | None:
```
james-garner-canonical, line 56: If the latest release was a maintenance release (e.g. 2.23.1), won't this return it even if the new release is to be from the main branch? What do you think about inspecting tags directly instead of releases here?

release.py
james-garner-canonical: I'm not sure what we agreed on last, but I'd quite like this too, and now that we have this automation, doing so won't be an onerous part of the release process (just an onerous part of the automation development =).

release.py
```diff
@@ -0,0 +1,848 @@
...
+    full_changelog_line = None
+
+    for line in release_notes.splitlines():
+        line = line.strip()
+        if line.startswith('* '):
+            match = re.match(r'\* (\w+): (.*) by @\w+ in (.*)', line)
+            if match:
+                category = match.group(1)
+                description = match.group(2).strip()
+                description = description[0].upper() + description[1:]
```
james-garner-canonical:
```suggestion
                description = match.group(2).strip().capitalize()
```

release.py
```diff
@@ -0,0 +1,848 @@
...
+                pr_link = match.group(3).strip()
+                if category in categories:
+                    categories[category].append((description, pr_link))
```
james-garner-canonical: Suggestion: put the `definition` and `pr_link` creation in here too (behind the `if category in categories`).

release.py
```diff
@@ -0,0 +1,848 @@
...
+    existing_content = ''
+    if os.path.exists(file):
+        with open(file) as f:
+            existing_content = f.read()
+
+    new_content = changes + existing_content
+
+    with open(file, 'w') as f:
+        f.write(new_content)
+
```
james-garner-canonical: We should probably use `pathlib` for path stuff. Then this could be, e.g.:
```suggestion
    existing_content = path.read_text() if path.exists() else ''
    path.write_text(changes + existing_content)
```

release.py
```diff
@@ -0,0 +1,848 @@
...
+    """Update the uv.lock file with the new versions."""
+    subprocess.run(['uv', 'lock'], check=True)  # noqa: S607
+
+
+def parse_scenario_version():
+    """Parse the current scenario version from pyproject.toml."""
+    file = VERSION_FILES['testing']
+    file_path = Path(file)
+    content = file_path.read_text()
+    match = re.search(r'version = "' + VERSION_REGEX + '"', content)
```
james-garner-canonical: Does a raw f-string work here (and elsewhere where we use this pattern)?
```suggestion
    match = re.search(rf'version = "{VERSION_REGEX}"', content)
```

release.py
```diff
@@ -0,0 +1,848 @@
...
+gh_client = github.Github(auth=auth)
+
+
+VERSION_REGEX = r'(\d+\.\d+\.\d+(?:\.dev\d+)?)'
+VERSION_FILES = {
+    'ops/src': 'ops/version.py',
+    'ops/pyproject': 'pyproject.toml',
+    'testing': 'testing/pyproject.toml',
+    'tracing': 'tracing/pyproject.toml',
+    'uvlock': 'uv.lock',
```
james-garner-canonical: Speaking of `pathlib`, maybe it would pay to make these `pathlib.Path` where they're defined?
```suggestion
VERSION_FILES = {
    'ops/src': Path('ops/version.py'),
    'ops/pyproject': Path('pyproject.toml'),
    'testing': Path('testing/pyproject.toml'),
    'tracing': Path('tracing/pyproject.toml'),
    'uvlock': Path('uv.lock'),
```

release.py
```diff
@@ -0,0 +1,848 @@
...
+        else:
+            suggested_tag = bump_minor_version(latest_tag)
+            logger.info(f'Suggested new version: {suggested_tag}')
+
+    while True:
+        user_input = input(
+            f'Input tag to release (press enter to use the suggested tag '
+            f"{suggested_tag or 'required'}), or 'c' to cancel: "
+        ).strip()
+
```
james-garner-canonical: I don't think the prompt looks quite right in the case where there isn't a suggested tag. How about this:
```suggestion
    tag_prompt = f' (press enter to use the tag {suggested_tag})' if suggested_tag else ''
    prompt = f"Input the new tag for the release{tag_prompt}, or 'c' to cancel:\n"
    while True:
         user_input = input(prompt).strip()
```

release.py
```diff
@@ -0,0 +1,848 @@
...
+        'test': [],
+        'ci': [],
+    }
+    full_changelog_line = None
+
+    for line in release_notes.splitlines():
+        line = line.strip()
+        if line.startswith('* '):
+            match = re.match(r'\* (\w+): (.*) by @\w+ in (.*)', line)
+            if match:
```
james-garner-canonical:
```suggestion
        if match := re.match(r'^\* (\w+): (.*) by @\w+ in (.*)', line.strip()):
```
And dedent accordingly.

release.py
```diff
@@ -0,0 +1,848 @@
...
+
+def format_release_notes(
+    categories: dict[str, list[tuple[str, str]]], full_changelog: str | None
+) -> str:
+    """Format for release notes.
+
+    Results in a Markdown formatted string with sections for each commit type.
+    If `full_changelog` is provided, it is appended at the end.
+    """
+    res = "## What's Changed\n\n"
```
james-garner-canonical: Minor: it's much more efficient in Python to use `str.join` instead of repeated `+=`. I doubt efficiency is a concern here, but to me it looks more idiomatic to write:
```suggestion
    lines = ['## What's Changed', '', '',]
```
And at the end:
```py
return '\n'.join(lines)
```

release.py
```diff
@@ -0,0 +1,848 @@
...
+        r'ops==' + VERSION_REGEX,
+        f'ops=={ops_version}',
+        updated,
+    )
+    # Fail if nothing is changed.
+    if content == updated:
+        logger.error(f'No changes made to {file}. Check the versions.')
+        exit(1)
+    file_path.write_text(updated)
+    logger.info(f'Updated {file} to version {ops_version}')
```
james-garner-canonical: There seems to be a lot of duplication here that could perhaps be factored out cleanly before we merge this.

release.py
james-garner-canonical: +1 for a link instead, please

dimaqq:
Re PR description:

> git command available locally

I imagine it's pretty typical though. We all use git.

The PR doesn't seem to shell out to `git` or `gh`.

Could it be that the PR description is out of date?

tox.ini
```diff
@@ -190,3 +190,21 @@ commands =
...
+commands = python release.py
+
+[testenv:publish-release]
+description = Publish the draft release
+dependency_groups = release
+passenv = GITHUB_TOKEN
+commands = python release.py --publish
+
+[testenv:post-release]
+description = Make a release
```
benhoyt: This looks like it's needs updating to "Perform post-release actions" or similar.

release.py
```diff
@@ -0,0 +1,848 @@
...
+        '--repo',
+        '-r',
+        help='Repository name (e.g. "operator")',
+        default='operator',
+    )
+    parser.add_argument(
+        '--owner',
+        '-o',
+        help='Owner name (e.g. "Canonical")',
+        default='Canonical',
```
benhoyt: Shouldn't this be lowercase `canonical`? And maybe we can just hard-code repo and owner, given this is a very specific script?

release.py
IronCore864: In my testing and debugging, I found it hard to go through all those outputs, so I adopted this symbol formatter from [here](https://github.com/dimaqq/python-libjuju/blob/cd5930c85d0f8bf5417bc9394d2c767ecc6e5e85/example.py#L63-L93) based on Dima's suggestion. Now that the script is steady and requires less debugging, I am OK to drop this. Reverting it to a simple formatter, something like:

```python
formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')

logger = logging.getLogger('release')
logger.setLevel(logging.DEBUG)

handler = logging.StreamHandler()
handler.setFormatter(formatter)

logger.addHandler(handler)
```

release.py
IronCore864: My idea is that the person who's creating the draft PR might want to have a look at it before asking others to review, and it needs to be opened in the browser to assign reviewers anyway, so I added this, but it's more or less redundant, because the PR url is already printed out anyway. I will remove this.

release.py
IronCore864: Per [discussion here](https://docs.google.com/document/d/1fkoqp1A2SMiOnwHubOJVRkuD1GpvcYxSDmfExYIZO88/edit?tab=t.0) (internal link), we have agreed on "Drop the GH usernames (even if they come from outside the team)."

release.py
IronCore864: If there is no major version change, it is fine to use a suggested version, either from `{ops_major+5}.{ops_minor}.{ops.bugfix}`, or parse what's in the `testing/pyproject.toml` file and bump the minor version (part of the logic in this function).

Since this script was created during the 2.x -> 3 release, I added an optional override. There will still be an automatically suggested version in case you don't want to input anything, so I am keeping this logic for now.

release.py
IronCore864: I added this because in my network environment, I met failures here. I added retries elsewhere, making the script much more complicated, so I reverted most changes and kept only this one. But I agree it should be consistent, so I will remove this one too.

release.py
tonyandrewmeyer: Huh, so we did. I think it's a bad choice, but that's what we should do since it was the team decision.

release.py
IronCore864: I thought about many possibilities here:

- What if the latest tag/release is 2.23.1 and now we want 2.23.2 instead of 2.24.0?
- What if the latest tag/release does not follow SEMVER? For example, during the 3.0 release, Dima made a manual attempt to publish tracing with a `.post1` release. I'm not saying it's wrong to do so, just that it's a possibility and it's quite probable too)

No matter what we try to do here, it can't always be "correct", that's why the value from here is only used as a "suggested version", asking the user to confirm, with a possibility to override manually.

I think this is fine according to the discussion we had in Frankfurt, since we agreed on "asks for confirmation", and an option to manually specify a version.

release.py
IronCore864: The reason behind my current implementation is that `parse_release_notes` only parses PRs into categories, then the result is used later for two purposes:
- When creating a release note, use the result from `parse_release_notes`, and format it as to release note (`format_release_notes`).
- When updating `CHANGES.md`, use the result from `parse_release_notes`, and format it accordingly; that's why the description and PR links are handled in `format_changes`.

release.py
IronCore864: I had a draft without duplication in the first place, but it requires many if/elses, since:

- For each module, we need to decide if we are only updating a source file or also `pyproject.toml`
- For some modules, it only needs to update the version of itself, but for others, it also needs to update the ops version.

So I ended up with the current version, with duplicated code but much more readable.

release.py
IronCore864: upper/lower: updated to lower.

Hard-code V.S. default values: when testing, I need to create a repo under `my-org/repo`, then fork it to `my-gh-handle/repo`, then make the release from the latter, so I did the defaults with a possibility to override for local testing.

release.py
james-garner-canonical: Sorry, my suggestion was confusingly worded. Here's what I'm suggesting:
```py
category = match.group(1)
if category in categories:
    description = match.group(2).strip()
    description = description[0].upper() + description[1:]
    pr_link = match.group(3).strip()
    categories[category].append((description, pr_link))
```

release.py
james-garner-canonical: That makes sense as an overall philosophy.

On further reflection, I see how `'maintenance' not in branch_name` is trickier to handle since the branch name doesn't contain a version.

Perhaps this logic would be easier to follow if structured like this:
```py
releases = repo.get_releases()
known_version = (
    branch_name.removesuffix('-maintenance')
    if branch_name.endswith('-maintenance')
    else None
)
for release in releases:
    if release.draft:
        continue
    if known_version and known_version not in release.tag_name:
        continue
    return release.tag_name
return None
```

release.py
james-garner-canonical: I don't mind having three distinct top-level functions like this (`update_ops_version`, `update_testing_version`, `update_tracing_version`), that sounds like a good call to me. I do think that some internal deduplication could be nice. I'm imagining something like:
```py
def update_tracing_version(version: str) -> None:
    update_pyproject_versions(VERSION_FILES['tracing'], version, deps={'ops': version})

def update_pyproject_versions(path: Path, version: str, deps: dict[str, str]) -> None:
    content = path.read_text()
    updated = re.sub(rf'version = "{VERSION_REGEX}"', f'version = "{version}"', content)
    for pkg, pkg_version in deps.items():
        updated = re.sub(rf'{pkg}=={VERSION_REGEX}', f'{pkg}=={pkg_version}', updated)
    if content == updated:
        logger.error(f'No changes made to {path}. Check the versions.')
        exit(1)
    path.write_text(updated)
    logger.info(f'Updated {path} to version {version}')
```

release.py
```diff
@@ -0,0 +1,732 @@
...
+
+import github
+import github.GitRelease
+import github.Repository
+
+formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger('release')
+logger.setLevel(logging.DEBUG)
+handler = logging.StreamHandler()
+handler.setFormatter(formatter)
```
dimaqq: **cleanup:**

Is the debug level set at logger and not globally because the imported libraries log too much at that level?

I see a bunch of `logger.info(...)` but no `logger.debug(...)` in this file, so maybe this setup doesn't actually make a difference?

release.py
```diff
@@ -0,0 +1,732 @@
...
+logger.addHandler(handler)
+
+
+if 'GITHUB_TOKEN' not in os.environ:
+    logger.critical('Environment variable GITHUB_TOKEN not set.')
+    exit(1)
```
dimaqq: **cleanup:** more elegantly expressed as:

```suggestion
    raise SystemExit('Environment variable GITHUB_TOKEN not set.')
```

release.py
```diff
@@ -0,0 +1,732 @@
...
+
+auth = github.Auth.Token(os.environ['GITHUB_TOKEN'])
+gh_client = github.Github(auth=auth)
```
dimaqq, line 43: **note:** while this makes sense for scripts, hard dependency on the env var also means that this file cannot be imported in other contexts, e.g. in unit tests.

I suppose it's OK to keep it as is.

release.py
```diff
@@ -0,0 +1,732 @@
...
+    update_draft_release(release, title, notes)
+
+    changes = format_changes(categories, tag)
+    update_changes_file(changes, 'CHANGES.md')
+
+    update_versions_for_release(tag)
+
+    new_branch = f'release-prep-{tag}'
+    subprocess.run(['/usr/bin/git', 'checkout', '-b', new_branch], check=True)
+    changed_files = ['CHANGES.md', *[str(path) for path in VERSION_FILES.values()]]
```
dimaqq, line 490: **discussion:** this seems to the only place where `VERSION_FILES['uvlock']` is used.

I wonder if the concepts of "files we must change" and "files that are changed as a side-effect" could be separated clearer, or perhaps we can just hardcode every path, because it's a script.

release.py
```diff
@@ -0,0 +1,732 @@
...
+        pr = open_prs[0]
+        pr_url = f'#{pr.number} - {pr.html_url}'
+        logger.info(f'  {pr_url}')
+        if len(open_prs) > 1:
+            logger.info(
+                f'Note that there are {len(open_prs) - 1} more open update charm pins PRs.'
+            )
+
+
+def draft_release(owner: str, repo_name: str, branch: str):
```
dimaqq: **suggestion:** perhaps clearer names for base and PR branches would help reading the code.

Here readability is particularly important as there are no unit tests for this script.

```suggestion
def draft_release(owner: str, repo_name: str, base_branch: str):
```

release.py
dimaqq: IMO, this function should return the latest tag (non-empty string) or raise an error.

Returning `None` doesn't make sense, as there's always some release in the given base branch history.

Example:

```lisp
* 4d2fa49c (origin/main, origin/HEAD, david/main, david/HEAD, main) docs: update style guide (#1720)
...
* ce160b0d chore: adjust versions after release (#1879)
* 5e0d22b0 (tag: 3.0.0, tiexin/main, tiexin/HEAD) chore: update changelog and versions for 3.0.0 release (#1874)
* 0ee01ea1 ci: disable non-essential checks that need adjustments to work with a Python 3.10 minimum (#1878)
* c456fda0 chore: remove 3.8 and update to 3.10 (#1873)
* 3a69b71d chore: post release version update (#1866)
| * 51b49ac1 (upstream/2.23-maintenance, 2.23-maintenance) docs: update title and edit links for Ops 2.23 docs (#1885)
| * d61cba09 chore: version bump for 2.23 maintenance branch and fix observability ci charm tests (#1871)
| * 9b65f0b8 docs: add notice about Ops 2 and Ops 3 (#1867)
|/
* 4b43f2c6 docs: be consistent with recommending self.app and self.unit (#1856)
* af8c84a5 (tag: 2.23.0.post1) ci: hotfix, publish job for ops-tracing (#1865)
* ec2dc3f7 (tag: 2.23.0) chore: update changelog and versions for 2.23.0 release (#1864)
```

If we're at main HEAD, we walk back to `5e0d22b0 (tag: 3.0.0,`.
If we're at maintenance HEAD, we walk back to `af8c84a5 (tag: 2.23.0.post1)`

P.S. I didn't invent this, this is what other tools, like `google/release-please` do.

The subtle difference is operating on releases (GH API + depagination) vs tags (local checkout, must use smth like `git fetch --all --tags`).

release.py
IronCore864: Yes, I will update it to INFO. Thanks for pointing this out.
I had some debugging stuff, but I don't think they are extremely helpful, so I removed them.

release.py
IronCore864: My idea is "just hardcode every path", but since these paths are used in multiple places besides "git add/commit", I hard-coded them as dict values to eliminate magic numbers.

release.py
IronCore864: Thanks, will also update `post_release` as well.

release.py
james-garner-canonical: Since Dima has endorsed using tags, I'd like to echo that suggestion.

release.py
```diff
@@ -0,0 +1,694 @@
...
+    lines: list[str] = []
+    while True:
+        line = input()
+        if line.strip() == '.':
+            break
+        lines.append(line)
+
+    summary = '\n'.join(lines).strip()
+    summary = summary.lstrip('\n')
+    summary = summary.rstrip('\n') + '\n'
```
james-garner-canonical: Strip will remove any leading or trailing newlines.
```suggestion
    summary = '\n'.join(lines).strip() + '\n'
```

release.py
IronCore864: The tag returned here is only used to calculate a "suggested version", which needs to be user-confirmed, and if it's None, the user has to provide some value as the tag of the new release.

If there is a new branch, such as 3.0.0-maintenance, and we haven't made any release in that branch yet, it makes sense to return "None" instead of raising/exiting, because then the user can continue manually inputting a tag and making a release, right?

The `known_version` from James, in my opinion, makes the code easier to read. I will make that change, but maybe rename it to `maintenance_branch_version` instead of `known_version`, because when it's the main branch and `known_version` is `None` seem weird to me.

IronCore864:
Another test after refactoring according to all comments above:

```bash
$ git branch
  main

$ python release.py --owner ironcoreworks --repo release-automation  # testing org and repo
2025-07-10 11:13:14,849 - INFO - Latest tag in branch "main": 0.1.0
2025-07-10 11:13:14,850 - INFO - Suggested new version: 0.2.0
Input the new tag for the release (press enter to use the tag 0.2.0), or 'c' to cancel:
3.2.0
2025-07-10 11:13:17,939 - WARNING - Check out the releases page: https://github.com/ironcoreworks/release-automation/releases before confirming!
Confirm creating tag '3.2.0' on branch 'main'? [y/N]: y
2025-07-10 11:13:19,675 - INFO - Draft release created: https://github.com/IronCoreWorks/release-automation/releases/tag/untagged-5a42876b7b5b428d3b6b.
================================================================================
Formatted release notes:
================================================================================
## What's Changed


### Features
* Add changes.md in https://github.com/IronCoreWorks/release-automation/pull/6
* Update release automation script and versions in https://github.com/IronCoreWorks/release-automation/pull/20
* Update in https://github.com/IronCoreWorks/release-automation/pull/23
* Update release script in https://github.com/IronCoreWorks/release-automation/pull/24

### Documentation
* Add version in https://github.com/IronCoreWorks/release-automation/pull/3

### Tests
* Add test in https://github.com/IronCoreWorks/release-automation/pull/4

**Full Changelog**: https://github.com/IronCoreWorks/release-automation/compare/0.1.0...3.2.0
================================================================================
2025-07-10 11:13:19,676 - INFO - The automatically generated title is: 3.2.0
Enter release title, press Enter to keep the auto-generated title:
> Test release, ops 3.2.0, tracing 3.2.0, testing/scenario 8.2.0

Enter release summary (multi-line supported; type '.' on a new line to finish):
This is a release for testing the automation script.

The previous tag from my repo is 0.1.0, hence the suggested version is 0.2.0, which is overriden by values provided by the user.

A quick brown fox jumps over something that I forgot.

Put a single dot at the final line to finish drafting the introduction paragraph.
.
2025-07-10 11:13:38,740 - INFO - Release updated: https://github.com/IronCoreWorks/release-automation/releases/tag/untagged-01c17942f2ee6832d913.
2025-07-10 11:13:38,742 - INFO - Updated CHANGES.md with new release notes.
2025-07-10 11:13:38,742 - INFO - Suggested new scenario version: 8.2.0
Input new scenario version (press enter to use the suggested version 8.2.0), or 'c' to cancel:
Confirm using scenario version '8.2.0'? [y/N]: y
2025-07-10 11:13:47,429 - INFO - Updated ops/version.py to version 3.2.0
2025-07-10 11:13:47,430 - INFO - Updated pyproject.toml to version 8.2.0
2025-07-10 11:13:47,430 - INFO - Updated testing/pyproject.toml to version 8.2.0
2025-07-10 11:13:47,431 - INFO - Updated tracing/pyproject.toml to version 3.2.0
Using CPython 3.12.9
Resolved 149 packages in 304ms
Updated ops-scenario v8.2.0.dev0 -> v8.2.0
Updated ops-tracing v3.2.0.dev0 -> v3.2.0
Switched to a new branch 'release-prep-3.2.0'
[release-prep-3.2.0 0a8f4d0] chore: prepare release 3.2.0
 6 files changed, 26 insertions(+), 9 deletions(-)
Enumerating objects: 20, done.
Counting objects: 100% (20/20), done.
Delta compression using up to 10 threads
Compressing objects: 100% (8/8), done.
Writing objects: 100% (11/11), 2.03 KiB | 2.03 MiB/s, done.
Total 11 (delta 6), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (6/6), completed with 6 local objects.
remote:
remote: Create a pull request for 'release-prep-3.2.0' on GitHub by visiting:
remote:      https://github.com/IronCore864/release-automation/pull/new/release-prep-3.2.0
remote:
To github.com:IronCore864/release-automation.git
 * [new branch]      release-prep-3.2.0 -> release-prep-3.2.0
2025-07-10 11:13:53,908 - INFO - Created PR: https://github.com/IronCoreWorks/release-automation/pull/26
2025-07-10 11:13:53,908 - INFO - Draft release created. Please merge the version bump PR, publish the draft release, thenrun this script with --post-release to perform post-release actions.
```

release.py
```diff
@@ -0,0 +1,694 @@
...
+                f'Note that there are {len(open_prs) - 1} more open update charm pins PRs.'
+            )
+
+
+def draft_release(owner: str, repo_name: str, base_branch: str):
+    """Create a draft release, update changelog, and create a PR for the release."""
+    org = gh_client.get_organization(owner)
+    repo = org.get_repo(repo_name)
+
+    check_update_charm_pins_prs(repo)
```
tonyandrewmeyer, line 463: I wonder if we might need an override for this, but I think it's fine to leave as-is for now, and we can add that if it does come up in the future.

tonyandrewmeyer (changes requested):
I'm happy with this as-is. I'm sure we'll want to tweak things in the future, but this seems a good starting point. It's only for our use, so we have more flexibility than with most changes.

The one change I'd like is for HACKING.md to be updated to include instructions for running this script and remove instructions for all the things this script does.

release.py
```diff
@@ -0,0 +1,694 @@
...
+
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+import re
+import subprocess
+from datetime import datetime
+from pathlib import Path
```
benhoyt: [Our style](https://github.com/canonical/operator/blob/main/STYLE.md#import-modules-not-objects) is `import package`:
```suggestion
import datetime
import pathlib
```

release.py
```diff
@@ -0,0 +1,694 @@
...
+    'ops/pyproject': Path('pyproject.toml'),
+    'testing': Path('testing/pyproject.toml'),
+    'tracing': Path('tracing/pyproject.toml'),
+    'uvlock': Path('uv.lock'),
+}
+
+
+def get_latest_release_tag(repo: github.Repository.Repository, branch_name: str) -> str | None:
+    """Get the latest release tag from the repository."""
+    releases = repo.get_releases()
```
benhoyt, line 58: Are we certain this returns the releases in newest-first order?

release.py
```diff
@@ -0,0 +1,694 @@
...
+
+        return release.tag_name
+
+    return None
+
+
+def bump_minor_version(version: str) -> str:
+    """Bump minor version."""
+    major, minor, patch = map(int, version.split('.'))
+    return f'{major}.{minor + 1}.{patch}'
```
benhoyt: Would we want to reset the patch version to 0 here? It would seem odd to bump from say 2.23.4 to 2.24.4.

release.py
```diff
@@ -0,0 +1,694 @@
...
+    else:
+        logger.info(f'Latest tag in branch "{branch_name}": {latest_tag}')
+        if not re.match(r'^\d+\.\d+\.\d+$', latest_tag):
+            logger.info('Latest tag is not in format X.Y.Z.')
+        else:
+            suggested_tag = bump_minor_version(latest_tag)
+            logger.info(f'Suggested new version: {suggested_tag}')
+
+    tag_prompt = f' (press enter to use the tag {suggested_tag})' if suggested_tag else ''
+    prompt = f"Input the new tag for the release{tag_prompt}, or 'c' to cancel:\n"
```
benhoyt: I don't think we need special handling of "c" to cancel. It's common practice to press Ctrl-C to cancel these kind of scripts (which Python would handle automatically).

release.py
```diff
@@ -0,0 +1,694 @@
...
+    while True:
+        user_input = input(prompt).strip()
+        if user_input.lower() == 'c':
+            logger.warning('Release canceled.')
+            exit(0)
+
+        new_tag = user_input if user_input else suggested_tag
```
benhoyt: Nit, equivalent to:
```suggestion
        new_tag = user_input or suggested_tag
```

release.py
```diff
@@ -0,0 +1,694 @@
...
+
+        if not re.match(r'^\d+\.\d+\.\d+$', new_tag):
+            logger.error('Error: Tag must be in format X.Y.Z')
+            continue
+
+        release_page = f'https://github.com/{owner}/{repo.name}/releases'
+        logger.warning(f'Check out the releases page: {release_page} before confirming!')
+
+        confirm = (
+            input(f"Confirm creating tag '{new_tag}' on branch '{branch_name}'? [y/N]: ")
```
benhoyt: You can get rid of all the extra quotes and have Python add them for you by using the `!r` format (repr). Similar elsewhere:

```suggestion
            input(f'Confirm creating tag {new_tag!r} on branch {branch_name!r}? [y/N]: ')
```

release.py
```diff
@@ -0,0 +1,694 @@
...
+    try:
+        release = repo.create_git_release(
+            tag=tag,
+            name=tag,
+            draft=True,
+            generate_release_notes=True,
+            target_commitish=branch,
+        )
+        return release
+    except Exception as e:
```
benhoyt: I would prefer to simplify and just to let this exception bubble up and exit. We're exiting anyway, and it avoids the try/except boilerplate here and the check in the caller, and provides a full stack trace for the developer using the tool.

release.py
```diff
@@ -0,0 +1,694 @@
...
+        'docs': [],
+        'test': [],
+        'ci': [],
+    }
+    full_changelog_line = None
+
+    for line in release_notes.splitlines():
+        if match := re.match(r'^\* (\w+): (.*) by @\w+ in (.*)', line.strip()):
+            category = match.group(1)
+            if category in categories:
```
benhoyt, line 180: Should we have an `else` and print a log message if we run into a category we don't expect? Or is that not an issue?

release.py
```diff
@@ -0,0 +1,694 @@
...
+            res += f'## {commit_type_to_category(commit_type)}\n\n'
+
+            for description, pr_link in items:
+                pr_num = '?'
+                match = re.match(r'https?://[^ ]+/pull/(\d+)', pr_link)
+                if match:
+                    pr_num = match.group(1)
+                res += f'* {description} (#{pr_num})\n'
+
+            res += '\n'
```
benhoyt: Let's use a similar approach to `format_release_notes` with a `lines` list and `'\n'.join(lines)` at the end.

release.py
```diff
@@ -0,0 +1,694 @@
...
+    suggested_version, _ = parse_scenario_version()
+    logger.info(f'Suggested new scenario version: {suggested_version}')
+
+    while True:
+        user_input = input(
+            f'Input new scenario version (press enter to use the suggested version '
+            f"{suggested_version or 'required'}), or 'c' to cancel: "
+        ).strip()
+
+        if user_input.lower() == 'c':
```
benhoyt: Similar here -- I think we should drop this 'c' handling and just let the user press Ctrl-C.

release.py
```diff
@@ -0,0 +1,694 @@
...
+    content = file_path.read_text()
+    match = re.search(rf'version = "{VERSION_REGEX}"', content)
+    if not match:
+        raise ValueError(f'Could not find version string in {file_path}')
+    version_str = match.group(1)
+    base_version, dev_suffix = parse_version(version_str)
+    return base_version, dev_suffix
+
+
+def get_new_scenario_version_for_release() -> str:
```
benhoyt: Can we skip/automate this step entirely? Don't we want the scenario version to always be exactly major+5, like 3.1.2 -> 8.1.2? (For now at least, and we can always adjust the code later if needed.)

release.py
```diff
@@ -0,0 +1,694 @@
...
+
+        if confirm == 'y':
+            break
+
+        logger.info("Let's try again...")
+
+    return new_version
+
+
+def get_new_scenario_version_post_release() -> str:
```
benhoyt: Similar here, can we completely automate this for the scenario version?

release.py
```diff
@@ -0,0 +1,694 @@
...
+def update_versions_for_release(tag: str):
+    """Update version files to the specified release version."""
+    scenario_version = get_new_scenario_version_for_release()
+    update_ops_version(tag, scenario_version)
+    update_testing_version(tag, scenario_version)
+    update_tracing_version(tag)
+    update_uv_lock()
+
+
+def get_new_version_post_release(repo: github.Repository.Repository, branch_name: str) -> str:
```
benhoyt, line 400: Can we completely automate this? Won't this always be minor+1 with the `.dev` suffix?

release.py
```diff
@@ -0,0 +1,694 @@
...
+    open_prs = [pr for pr in prs if pr.title == 'chore: update charm pins']
+    if open_prs:
+        logger.info('Please merge "update charm pins" PRs first:')
+        pr = open_prs[0]
+        pr_url = f'#{pr.number} - {pr.html_url}'
+        logger.info(f'  {pr_url}')
+        if len(open_prs) > 1:
+            logger.info(
+                f'Note that there are {len(open_prs) - 1} more open update charm pins PRs.'
+            )
```
benhoyt, line 449: Don't we want to exit at this point, and they re-run the script when they've merged them? (Or at least pause and wait for them to press enter or something.)

release.py
```diff
@@ -0,0 +1,694 @@
...
+    changes = format_changes(categories, tag)
+    update_changes_file(changes, 'CHANGES.md')
+
+    update_versions_for_release(tag)
+
+    new_branch = f'release-prep-{tag}'
+    subprocess.run(['/usr/bin/git', 'checkout', '-b', new_branch], check=True)
+    changed_files = ['CHANGES.md', *[str(path) for path in VERSION_FILES.values()]]
+    for file in changed_files:
+        subprocess.run(['/usr/bin/git', 'add', file], check=True)
```
benhoyt: Could do it in one operation instead, something like:
```suggestion
    subprocess.run(['/usr/bin/git', 'add', *changed_files], check=True)
```

release.py
```diff
@@ -0,0 +1,694 @@
...
+    ).stdout.strip()
+    exist = local_branch or remote_branch
+    if exist:
+        logger.error(
+            f'Branch "{new_branch}" already exists in the current repository. '
+            'Please double check and delete it first before post release'
+        )
+        exit(1)
+
+    subprocess.run(['/usr/bin/git', 'fetch', 'upstream'], check=True)
```
benhoyt: These days my `canonical/operator` remote is called `origin` (and `benhoyt` is mine), not `upstream`. Can we avoid hard-coding this?

Could we do `git checkout {base_branch}` and `git pull` instead? And do we need to do the same "update to latest" before `draft_release` too?

release.py
```diff
@@ -0,0 +1,694 @@
...
+    subprocess.run(['/usr/bin/git', 'merge', f'upstream/{base_branch}'], check=True)
+
+    org = gh_client.get_organization(owner)
+    repo = org.get_repo(repo_name)
+
+    update_versions_for_post_release(repo, base_branch)
+
+    subprocess.run(['/usr/bin/git', 'checkout', '-b', new_branch], check=True)
+    for file in [str(path) for path in VERSION_FILES.values()]:
+        subprocess.run(['/usr/bin/git', 'add', file], check=True)
```
benhoyt: Similar here, let's do this in one `git add` operation with `*files`.

release.py
```diff
@@ -0,0 +1,694 @@
...
+        post_release(owner=args.owner, repo_name=args.repo, base_branch=args.branch)
+        logger.info(
+            'Post-release actions completed. Please check and merge the created PR '
+            'for version updates.'
+        )
+        exit(0)
+
+    draft_release(owner=args.owner, repo_name=args.repo, base_branch=args.branch)
+    logger.info(
+        'Draft release created. Please merge the version bump PR, publish the draft release, then'
```
benhoyt:
```suggestion
        'Draft release created. Please merge the version bump PR, review and publish the draft release, then'
```

benhoyt (changes requested):
Looking close. A few changes requested (and I agree about updating HACKING.md). Let's try to get this merged by the end of this week.

release.py
IronCore864: This calls GitHub API `https://api.github.com/repos/OWNER/REPO/releases`, which [according to the official doc](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) does not seem to guarantee a specific sort order. Based on my testing in a few repos, it seems it's sorting by release date. Based on some more research, it seems it's sorting by `published_at`, but with `prerelease` records potentially appearing first ([source](https://github.com/orgs/community/discussions/21901#discussioncomment-3234702)).

release.py
IronCore864: We cannot skip it all, because the scenario version differs from the ops version.

This function parses the scenario version from `pyproject.toml` and removes the `.dev0` suffix, uses it as a suggested scenario version for the release, and if the user doesn't accept, they can override it by providing a manually input version for scenario. The logic isn't major +5.

release.py
IronCore864: This exists specifically for major releases (I made the script during 2.23 -> 3.0 release). If the previous version is 2.23, and now we want to release 3.0, a manual override is required. But I don't think we will make any major release soon, so I will fully automate this.

james-garner-canonical: Hmm, I kind of feel like the 'suggest a version and get user confirmation' ux is good here too. If we're worried about the volume of code, how about factoring out the user input loops into a helper function? Then the difference between having manual confirmation in this function or not would be just one line or so.

release.py
IronCore864: `base_branch` defaults to `main`, so `git checkout main` and `git pull` will get the forked repo's main, instead of `canonical/operator`'s main. I will make the remote name configurable.

IronCore864: Before `draft_release`: I assume `tox -e draft-release` is executed in the forked ops repo, checked out at latest main, with no changes in the staging area. The reason I did not do any operations before `draft_release` is that in case you are working on something, it might ruin the result. I will add some description in HACKING.md.

release.py
benhoyt: Yeah, if it's not guaranteed/specified in the docs, I think we should do a quick `.sort()` by `published_at`, to avoid surprises. Then again, they'll probably never change this ... it'd break too many people, so maybe it's okay.

james-garner-canonical: This github endpoint is paginated, so I guess calling sort on the `PaginatedList` type will pull down all the pages, though that's probably not a major concern.

Alternatives would be querying the git tags (e.g. as Dima pitched [here](https://github.com/canonical/operator/pull/1855/files#r2196235918)), or maybe pulling release info from PyPI (not sure how this works exactly).

For an interactive script like this I think the existing implementation is probably fine -- and we can always make fixes later if we get unusual versions suggested when running it.

benhoyt: Sounds reasonable to keep it as is, thanks James.

HACKING.md
```diff
@@ -234,63 +234,38 @@ the build frontend is [build](https://pypi.org/project/build/).
...
-    via a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/) relationship).
-    Note that it sometimes take a bit of time for the new releases to show up.
-
-    See [.github/workflows/publish-ops.yaml](.github/workflows/publish-ops.yaml) and
-    [.github/workflows/publish-ops-scenario.yaml](.github/workflows/publish-ops-scenario.yaml) for details.
-    (Note that the versions in the YAML refer to versions of the GitHub actions, not the versions of the ops  library.)
-
-    You can troubleshoot errors on the [Actions Tab](https://github.com/canonical/operator/actions).
-14. Announce the release on [Discourse](https://discourse.charmhub.io/c/framework/42)
+1. Make sure your forked `canonical/operator` repo's main branch is up-to-date with upstream, the staging area is clean, and you are at the main branch and at the root directory of the forked repo.
```
benhoyt: It seems odd to me that we require this for the main draft-release step, but do it automatically for the post-release one. If we can do it automatically, can we do it automatically for both, and drop this manual step?

HACKING.md
```diff
@@ -234,63 +234,38 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+2. Draft a release: Run: `tox -e draft-release`.
+
+    > This assumes a draft release on the main branch, and your forked remote name is `origin`, and the `canonical/operator` remote name is `upstream`.
+    > 
+    > If you have different settings, add parameters accordingly. For example, the following command assumes your forked remote name is `mine`, and `canonical/operator` remote name is `origin`:
```
benhoyt, line 240: Let's discuss in daily today if we can agree as a team what to call our remotes and whether there's a good way to automatically detect this.

HACKING.md
```diff
@@ -234,63 +234,38 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+    > 
+    > `tox -e draft-release -- --canonical-remote origin --fork-remote mine`
+
+3. Follow the steps of the `tox -e draft-release` output. You need to input the release title and an introduction section, which can be multiple paragraphs with empty lines in between. End the introduction section by typing a period sign (.) in a new line, then press enter.
+4. If drafting the release succeeds, a PR named "chore: update changelog and versions for X.Y.Z release" will be created. Get it reviewed and merged, then wait until the tests pass after merging. It takes around 10 minutes. If the tests don't pass at the tip of the main branch, do not continue.
+5. Go to the GitHub releases page, edit the latest draft release, and click "Publish release". GitHub will create the additional tag.
+
+    > Pushing the tags will trigger automatic builds for the Python packages and
+    > publish them to PyPI ([ops](https://pypi.org/project/ops/) and
+    > [ops-scenario](https://pypi.org/project/ops-scenario)) (authorisation is handled
```
benhoyt: And ops-tracing?

HACKING.md
```diff
@@ -234,63 +234,38 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+    > via a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/) relationship).
+    > Note that it sometimes take a bit of time for the new releases to show up.
```
benhoyt: I think we can leave out the "authorisation is handled" sentence from here. A releaser doesn't need to know all the mechanics.

HACKING.md
```diff
@@ -234,63 +234,38 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+    > 
+    > See [.github/workflows/publish-ops.yaml](.github/workflows/publish-ops.yaml) and
+    > [.github/workflows/publish-ops-scenario.yaml](.github/workflows/publish-ops-scenario.yaml) for details.
+    > (Note that the versions in the YAML refer to versions of the GitHub actions, not the versions of the ops  library.)
```
benhoyt: I think we can drop this parenthesis.

HACKING.md
```diff
@@ -234,63 +234,38 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+    >
+    > You can troubleshoot errors on the [Actions Tab](https://github.com/canonical/operator/actions).
+6. Announce the release on [Discourse](https://discourse.charmhub.io/c/framework/42)
```
benhoyt: And Matrix?

benhoyt (approved):
Looks good, thanks for the updates. Approving optimistically, but a few minor changes to HACKING.md suggested.

HACKING.md
IronCore864: Per discussion, we agreed to use origin as the remote name of the forked repo, and upstream for canonical/operator. I have already refactored the script so that they are not hard-coded, and default values are provided.

HACKING.md
```diff
@@ -234,63 +234,41 @@ the build frontend is [build](https://pypi.org/project/build/).
...
-    - in [pyproject.toml for `ops-tracing`](tracing/pyproject.toml), the `version` attribute and the required version for `ops`
+1. Draft a release: Run: `tox -e draft-release`.
+
+    > This assumes a draft release on the main branch, and your forked remote name is `origin`, and the `canonical/operator` remote name is `upstream`.
+    > 
+    > If you have different settings, add parameters accordingly. For example, the following command assumes your forked remote name is `mine`, and `canonical/operator` remote name is `origin`:
+    > 
+    > `tox -e draft-release -- --canonical-remote origin --fork-remote mine`
+    > 
+    > If you want to make a release on another branch, for example, on "2.23-maintenance", run:
```
tonyandrewmeyer: Do I still run the command with my local clone in the main branch? Or do I switch/checkout the branch I want to release first? Or will the script take care of it and it doesn't matter?

HACKING.md
```diff
@@ -234,63 +234,41 @@ the build frontend is [build](https://pypi.org/project/build/).
...
+4. Go to the GitHub releases page, edit the latest draft release, and click "Publish release". GitHub will create the additional tag.
+
+    > Pushing the tags will trigger automatic builds for the Python packages and
+    > publish them to PyPI ([ops](https://pypi.org/project/ops/) 
+    > ,[ops-scenario](https://pypi.org/project/ops-scenario), and 
+    > [ops-tracing](https://pypi.org/project/ops-tracing/)).
+    > Note that it sometimes take a bit of time for the new releases to show up.
+    > 
+    > See [.github/workflows/publish-ops.yaml](.github/workflows/publish-ops.yaml) and
+    > [.github/workflows/publish-ops-scenario.yaml](.github/workflows/publish-ops-scenario.yaml) for details.
```
tonyandrewmeyer: @dimaqq changed this underneath you :smile:. It's just one workflow now.

tonyandrewmeyer (approved):
A couple of small notes, but looks great, thanks!

HACKING.md
IronCore864: Updated accordingly.

HACKING.md
IronCore864: Got it, thanks. Merging main now.

IronCore864:
Merged main into this branch, resolved the conflicts in HACKING.md, also tweaked the description in HACKING.md a little bit. Merging now.

release.py
```diff
@@ -0,0 +1,609 @@
...
+    if exist:
+        logger.error(
+            f'Branch "{new_branch}" already exists in the current repository. '
+            'Please double check and delete it first before post release'
+        )
+        exit(1)
+
+    subprocess.run(['/usr/bin/git', 'fetch', canonical_remote], check=True)
+    subprocess.run(['/usr/bin/git', 'checkout', base_branch], check=True)
+    subprocess.run(['/usr/bin/git', 'merge', f'{canonical_remote}/{base_branch}'], check=True)
```
dimaqq: I would prefer this to be `git pull ...` or `git merge --ff-only ...`

Otherwise, a local main with extra commits will end up with those extra commits and a merge commit for upstream main, then these changes will be pushed to the PR, etc.

In other words, I'd rather this tox job failed early if my local main is weird.

release.py
```diff
@@ -0,0 +1,609 @@
...
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.strip()
+    remote_branch = subprocess.run(
+        ['/usr/bin/git', 'ls-remote', '--heads', fork_remote, new_branch],
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.strip()
```
dimaqq: **minor:** this could be expressed shorter with `subprocess.check_output`

The difference is in `stderr`, which I hope doesn't matter if that's only present on error.

release.py
```diff
@@ -0,0 +1,609 @@
...
+    exist = local_branch or remote_branch
+    if exist:
```
dimaqq: This could be simplified:

```suggestion
    if local_branch or remote_branch:
```

release.py
```diff
@@ -0,0 +1,609 @@
...
+        logger.error(
+            f'Branch "{new_branch}" already exists in the current repository. '
```
dimaqq: **minor:** the error is not quite accurate wrt. the check above.

release.py
```diff
@@ -0,0 +1,609 @@
...
+    'ops/pyproject': pathlib.Path('pyproject.toml'),
+    'testing': pathlib.Path('testing/pyproject.toml'),
+    'tracing': pathlib.Path('tracing/pyproject.toml'),
+    'uvlock': pathlib.Path('uv.lock'),
+}
+
+
+def get_latest_release_tag(repo: github.Repository.Repository, branch_name: str) -> str | None:
+    """Get the latest release tag from the repository."""
+    releases = repo.get_releases()
```
dimaqq, line 58: **minor:**
I think that this only returns 30 releases.
If we want more or all, we have to follow pagination.

So if we had 30 drafts or 30 maintenance releases up top, the limitation would cause this function to fail.
Likewise, if we have 30 mainline releases up top and need to make a maintenance release, this function won't find it.

I suppose we could live with this limitation.

release.py
```diff
@@ -0,0 +1,609 @@
...
+    )
+
+    for release in releases:
+        if release.draft:
+            continue
+
+        if maintenance_branch_version and maintenance_branch_version not in release.tag_name:
+            continue
+
+        return release.tag_name
```
dimaqq: This seems imprecise.

- suppose we've been passed `branch_name="1.5-maintenance"`
- now we've got `maintenance_branch_version="1.5"`
- and we encounter `release="3.1.5"`
- we'd call this the last release for the maintenance branch

One way to resolve this would be:

```suggestion
        if maintenance_branch_version and not release.tag_name.startswith(f"{maintenance_branch_version}."):
            continue

        return release.tag_name
```

Another would be to try to apply `packaging.version.parse()` or similar to the tag and process version objects rather than strings.

release.py
```diff
@@ -0,0 +1,609 @@
...
+VERSION_FILES = {
+    'ops/src': pathlib.Path('ops/version.py'),
+    'ops/pyproject': pathlib.Path('pyproject.toml'),
+    'testing': pathlib.Path('testing/pyproject.toml'),
+    'tracing': pathlib.Path('tracing/pyproject.toml'),
+    'uvlock': pathlib.Path('uv.lock'),
+}
+
+
+def get_latest_release_tag(repo: github.Repository.Repository, branch_name: str) -> str | None:
```
dimaqq, line 56: **Food for thought:** what if we just stripped `.dev0` from the current ops version in the branch?

Why bother going through GH releases, if we already have the next version (+dev0) in the source code?

release.py
```diff
@@ -0,0 +1,609 @@
...
+    return f'{int(major) + 5}.{minor}.{patch}.{dev0}'
+
+
+def update_versions_for_release(tag: str):
+    """Update version files to the specified release version."""
+    scenario_version = get_new_scenario_version(tag)
+    update_ops_version(tag, scenario_version)
+    update_testing_version(tag, scenario_version)
+    update_tracing_version(tag)
+    update_uv_lock()
```
dimaqq: **Bug:** I think this would set, e.g.:
- ops version to `3.1.0`
- scenario version to `8.1.0.dev0`

Specifically, "new scenario version" concept should be different between the release and post-release PRs.

release.py
IronCore864: As mentioned in another comment, although not officially stated, the return result is sorted by release date (unless there is a draft release). Also, even if this function doesn't find anything, the user can still provide an override. So I'm going to keep this as it is.

release.py
IronCore864: Both should work, and they should be the same. I think we should do both, do a comparison, if not the same, ask the user to override. But probably it's not necessary, since the return value is only used as a suggested next release version, which can be overridden. So I'm keeping it as it is for now.

release.py
```diff
@@ -0,0 +1,607 @@
...
+    lines = ["## What's Changed", '']  # Initialize lines.
+    for commit_type, items in categories.items():
+        if items:
+            lines.append(f'### {commit_type_to_category(commit_type)}')  # Add category header.
+            for description, pr_link in items:
+                lines.append(f'* {description} in {pr_link}')  # Add commit description.
+            lines.append('')  # Add a blank line after each category.
+    if full_changelog:
+        lines.append(full_changelog)  # Append the full changelog.
+    return '\n'.join(lines)  # Join the lines with newline characters.
```
dimaqq: **Minor:** consider removing these comments, it's pretty clear what the code does.

release.py
```diff
@@ -0,0 +1,607 @@
...
+def parse_scenario_version():
+    """Parse the current scenario version from pyproject.toml."""
+    file_path = VERSION_FILES['testing']
+    content = file_path.read_text()
+    match = re.search(rf'version = "{VERSION_REGEX}"', content)
+    if not match:
+        raise ValueError(f'Could not find version string in {file_path}')
+    version_str = match.group(1)
+    base_version, dev_suffix = parse_version(version_str)
+    return base_version, dev_suffix
```
dimaqq, line 376: **Very minor:** could be simplified.

This is a good example where type hint for function return type could be helpful.

dimaqq (approved):

PR #1857, opened 2025-06-27
ci: build and publish in one step

### LINKED ISSUE #1813: uv: update build and publish, can be done in one action (build all, push all)

dimaqq:
This PR needs updates in three PyPI and three Test PyPI projects' trusted publisher settings.
Let's hold off merging this until Ben is back.

---

Build and publish in one step to close #1813

- use a single workspace build command to build all three packages
   - we can do that since #1804 was merged
- rely on PYPI action attestations
   - the PYPI action notes that attestations are experimental, however they are on by default and they work fine:
   - See e.g. https://pypi.org/project/otlp-json/#otlp_json-1.0.0.tar.gz

### About attestations

Our current CI uses two actions, `actions/attest-build-provenance@v2` and `pypa/gh-action-pypi-publish@release/v1`, which generate two [sets of] attestations, for GitHub and PYPI, which look the same, but I think are not actually the same attestation.

https://pypi.org/project/ops/#ops-2.22.0.tar.gz
`Sigstore integration time: May 29, 2025, 1:18:42 PM`
Log Index [223561483](https://search.sigstore.dev/?logIndex=223561483)

https://github.com/canonical/operator/attestations/7358329
`Created
last month (Thu, 29 May 2025 04:18:34 GMT)`
        "logIndex": "223561324", (in the downloaded json)

The two attestations do include the same git commit hash and tag.

### Discussion

Apparently, there's `uv publish` command that also supports PYPI trusted publishing, but I figured that would be too much experimentation for one day.


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build-and-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb  # v6.1.0
```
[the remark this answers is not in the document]
dimaqq: @tonyandrewmeyer , the zizmor expert.
I'm going to disable caching in this action, on the assumption that for the release jobs specifically maybe we don't want something stale.
Still, doing so feels like sprinkling pixie dust.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: Publish
+"on":
+  push:
+    tags:
+      - '3.*'
```
james-garner-canonical: What would be the best approach for handling releases from the `2.*-maintenance` branches?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba  # v6.3.1
+        with:
+          enable-cache: false
+      - run: uv build --all
+      - uses: pypa/gh-action-pypi-publish@release/v1
```
james-garner-canonical: Does the publish action just work with all 3 built packages? If so, that's very handy.

.github/workflows/publish.yaml
dimaqq: That's a good point.
To keep PYPI settings simple, it's best to have one set of settings that covers both the main and maintenance branches.
I'll add the maintenance postfix.

.github/workflows/publish.yaml
dimaqq: The publish doesn't care, it pushes everything under `dist/`.
(TBH, I didn't validate that PyPI accepts all files, certainly not recently; I do know that multiple versions of the same package can be pushed in one go)

Each PYPI repo trusted publisher needs to be configured with this GHA workflow name though.

tonyandrewmeyer:
> Our current CI uses two actions, actions/attest-build-provenance@v2 and pypa/gh-action-pypi-publish@release/v1, which generate two [sets of] attestations, for GitHub and PYPI, which look the same, but I think are not actually the same attestation.

No, they are not. The intention is that we'll move to only providing the PyPI ones, but at the moment (a) [they are not considered stable](https://docs.pypi.org/attestations/producing-attestations/), (b) I haven't had time to try out the [consumption tooling](https://docs.pypi.org/attestations/consuming-attestations/), since it wasn't around when PyPI first added this, and (c) I'm not sure yet what SLSA level PyPI is providing. Right now, the GitHub attestations are more stable and more useful.

It's possible that all of this ends up being discarded when our new building/testing infrastructure is ready, which is one of the reasons I stopped putting much time into this.

.github/workflows/publish.yaml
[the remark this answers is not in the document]
tonyandrewmeyer: I think it's a legitimate reason to disable caching (which we did [already](https://github.com/canonical/operator/blob/4b43f2c6fb81dd9b53cfa98cdc7f7226d90bea23/.github/workflows/publish-ops-tracing.yaml#L22), so it's not a change). It's not about something being stale, it's about a cache injection attack, which is why it's specifically disabled for publishing. The [zizmor docs](https://docs.zizmor.sh/audits/#cache-poisoning) explain it pretty well.

We don't publiish often, so there's minimal benefit in caching for this workflow anyway.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: Publish
+"on":
+  push:
+    tags:
+      - '[1-3].*'
```
benhoyt, line 6: Probably better to use `[1-9]` to future proof?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+permissions:
+  id-token: write
+  attestations: write
+  contents: read
+
+jobs:
+  build-and-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-latest
```
tonyandrewmeyer, line 16: We should stick with specifying an environment:

```suggestion
    runs-on: ubuntu-latest
    environment: publish-pypi
```

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba  # v6.3.1
+        with:
+          enable-cache: false
+      - run: uv build --all
```
tonyandrewmeyer, line 25: I'd prefer that we still do both the GitHub and the PyPI attestation, until the PyPI one is further advanced. It doesn't really cost us anything other than a small amount of CI time, and there are advantages (as mentioned in the main thread).

```suggestion
      - run: uv build --all
      - uses: actions/attest-build-provenance@v2
        with:
          subject-path: 'dist/*'
```

.github/workflows/test-publish.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: Publish (test)
+"on":
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  attestations: write
```
tonyandrewmeyer: We don't need this for test publish.

```suggestion
```

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: Publish
+"on":
+  push:
+    tags:
+      - '[1-3].*'
+
+permissions:
+  id-token: write
+  attestations: write
```
tonyandrewmeyer, line 10: If we do drop the GitHub attestations, we should drop this line (but my preference is to keep them for now, as noted elsewhere).

tonyandrewmeyer (changes requested):
I love the move to a single pair of workflows, assuming we're continuing with always releasing all three packages at once.

I like the simplified workflows as well.

We do need to continue using a GitHub workflow environment (maybe we weren't doing that when you originally opened this PR?).

My main issue here is that I would really like us to continue with the GitHub attestations for now.

.github/workflows/publish.yaml
dimaqq: 🤔 I wonder... on one hand, less work two years from now.

On the other, if someone pushed a  8.1.0 tag (for scenario), I wonder if I would want this to job to get triggered.

Or maybe we can decide that it's kindof intended?

dimaqq:
> Assuming this works, looks okay to me. What needs to be done here on the PyPI side?

Today we have configuration like this (with variation in the workflow name):

Publisher | Details |  
-- | -- | --
GitHub | Repository: canonical/operator<br>Workflow: publish-ops-tracing.yaml<br>Environment name: publish-pypi

We'll keep this config block for all three PYPI and test PYPI repos, so that releases can be made from the maintenance granches.

We'd add this config (exact same workflow name) to all three PYPI repos:

Publisher | Details |  
-- | -- | --
GitHub | Repository: canonical/operator<br>Workflow: publish.yaml<br>Environment name: publish-pypi

And this config to Test PYPI:

Publisher | Details |  
-- | -- | --
GitHub | Repository: canonical/operator<br>Workflow: test-publish.yaml<br>Environment name:

.github/workflows/test-publish.yaml
dimaqq: Right... I kinda wished that publish and test-publish would be as close as possible, but maybe that's just not something we can do.

.github/workflows/publish.yaml
dimaqq: I've added the github attestations back.

.github/workflows/test-publish.yaml
dimaqq: done

.github/workflows/publish.yaml
dimaqq: done

.github/workflows/publish.yaml
dimaqq: done.

tonyandrewmeyer (approved):
Thanks!

PR #1860, opened 2025-06-30
Release Preparation: 2.23.0

IronCore864:
This PR prepares the release of version 2.23.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,57 @@
+# 2.23.0 - 30 June 2025
+
+## Features
+
+* Support for config schema as Python classes (#1741)
+* Support for action parameter schema as Python classes (#1756)
+* Ops[tracing] compatibility with jhack (#1806)
+* Support for relation data schema as Python classes (#1701)
+* Add CheckInfo.successes field and .has_run property (#1819)
+* Provide a method to create a testing.State from a testing.Context (#1797)
+* Expose trace data in testing (#1842)
+* Add a helper to generate a Layer from rockcraft.yaml (#1831)
+
+## Fixes
+
+* Correctly load an empty Juju config options map (#1778)
+* Fix type annotation of container check_infos in ops.testing (#1784)
+* Restrict the version of a dependency, opentelemetry-sdk (#1794)
+* Remote unit data is available in relation-departed (#1364)
+* Juju allows access to the remote app databag in relation-broken, so Harness should too (#1787)
+* Don't use private OpenTelemetry API (#1798)
+* Do not return this unit in a mocked peer relation (#1828)
+* Testing.PeerRelation properly defaults to no peers (#1832)
+* In meter-status-changed JUJU_VERSION is not set (#1840)
+* Only provide the units belonging to the app in Relation.units (#1837)
+
+## Documentation
+
+* Remove two best practices, and drop two to tips (#1758)
+* Update link to Charmcraft for managing app config (#1763)
+* Update link to Juju documentation for setting up deployment (#1781)
+* Fix external OTLP link (#1786)
+* Distribute the ops-scenario README content across the ops docs (#1773)
+* Improve testing.errors.UncaughtCharmError message (#1795)
+* In the "manage the charm version" how-to, give an example of using override-build (#1802)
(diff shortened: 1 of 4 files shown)


testing/pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.23.0.dev0"
+version = "2.23.0"
```
tonyandrewmeyer, line 11: We need to keep ops-scenario at 7.x :)

tonyandrewmeyer (changes requested):
This also needs to bump the dependencies for ops-scenario and ops-tracing, and the extra dependencies for ops.

dwilding:
For the PR title, how about:

```
chore: update changelog and versions for 2.23.0 release
```

CHANGES.md
```diff
@@ -1,3 +1,57 @@
+# 2.23.0 - 30 June 2025
+
+## Features
+
+* Support for config schema as Python classes (#1741)
+* Support for action parameter schema as Python classes (#1756)
+* Ops[tracing] compatibility with jhack (#1806)
```
dwilding, line 7: One edge case to consider (for the future)

```suggestion
* ops[tracing] compatibility with jhack (#1806)
```

CHANGES.md
```diff
@@ -1,3 +1,57 @@
...
+* Support for relation data schema as Python classes (#1701)
+* Add CheckInfo.successes field and .has_run property (#1819)
```
dwilding, line 9: Normally we'd use backticks:

```
Add `CheckInfo.successes` field and `.has_run` property
```

Maybe this should remain as a manual step?

CHANGES.md
dimaqq: I'm happy to let this pass, as `pip` and friends will happily install ops as Ops, and also the framework is called Ops.

Automation over precision in this case.

CHANGES.md
dimaqq: My 2c: this on the original PR author.

Let's aim for a good-enough result, not a perfect result.

dimaqq (changes requested):
Lock file needs to be updated.
This is always the case, when we change dependencies such as `ops-tracing==x.y.z` in ops and vice versa.

CHANGES.md
dwilding: I don't trust myself to be consistent with ' vs \` vs no formatting in PR titles ;)

I think this a reasonable manual clean-up step to keep - I'd consider it part of proof-reading the generated changelog. But ultimately agree, it's a nice-to-have, and good enough without

testing/pyproject.toml
IronCore864: Yes, I have also figured out this bug. We could not simply use the tag as the version for everything. For ops/tracing, it's OK, then for testing, we need to parse the value then bump the minor version. For post release, it's the same, we can't simply bump the latest tag then add `.dev0` suffix but it depends on which module it is. I have refactored this part. Will close this PR and test the script with a new draft release.

CHANGES.md
IronCore864: It's not easy for automation to add backticks since we need to know which word is code and which isn't. I will keep it as it is now, and maybe we could improve the original PR title in the future.

dimaqq:
> This also needs to bump the dependencies for ops-scenario and ops-tracing, and the extra dependencies for ops.

@IronCore864 please note our current HACKING.md wrt. versions:

> There will be a total of seven changes:
> 
> - in [ops/version.py for ops](https://github.com/canonical/operator/blob/main/ops/version.py), the version declared in the version variable
> - in [pyroject.toml for ops](https://github.com/canonical/operator/blob/main/pyproject.toml), the required versions for ops-scenario and ops-tracing
> - in [pyproject.toml for ops-scenario](https://github.com/canonical/operator/blob/main/testing/pyproject.toml), the version attribute and the required version for ops
> - in [pyproject.toml for ops-tracing](https://github.com/canonical/operator/blob/main/tracing/pyproject.toml), the version attribute and the required version for ops

IronCore864:
Closing this PR now, will try again using the updated version of the script.

PR #1867, opened 2025-07-01
docs: add notice about Ops 2 and Ops 3

dwilding:
This PR adds a notice at the top of the docs for Ops 2.23. The notice should help readers understand which version of Ops to use, and how to find the docs for the latest version of Ops.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,7 +10,14 @@ reference/index
 explanation/index
 ```
 
+```{important}
+This is the documentation for version 2 of Ops, which stopped being actively developed in June 2025.
 
+- If your charm needs to support Python 3.8 (Ubuntu 20.04), use Ops 2.
+- Otherwise, use Ops 3.
+
+Ops 3 removed support for Python 3.8, but is otherwise compatible with Ops 2. See the [latest documentation](https://ops.readthedocs.io).
+```
 
 The Ops library (`ops`) is a Python framework for writing and testing Juju charms.
 


docs/index.md
```diff
@@ -10,7 +10,14 @@ reference/index
 explanation/index
 ```
 
+```{important}
+This is the documentation for version 2 of Ops, which stopped being actively developed in June 2025.
```
tonyandrewmeyer: Maybe note that it'll still get security fixes and critical bug fixes, to reassure people that if they need to they can stay on this?

dimaqq (approved):

tonyandrewmeyer (approved):
Looks good, thanks!

I noticed that the version picker doesn't appear on ops.rtd - I think we deliberately hid that in the past because there was only one version. Maybe we should bring that back? Or if it'll come back when we move to rtd-hosted then that's soon enough I think.

docs/index.md
dwilding: I've updated the first part to this:

> This is the documentation for version 2 of Ops, which stopped being actively developed in June 2025. Ops 2 will continue to receive security and critical bug fixes.

Are you OK with not saying how long this will continue for? (We didn't mention a time frame in the release notes)

dwilding:
> I noticed that the version picker doesn't appear on ops.rtd

You preempted my follow-up question :slightly_smiling_face: 

I'll raise that for discussion after the release discussions have subsided

IronCore864 (approved):

docs/index.md
tonyandrewmeyer: > Are you OK with not saying how long this will continue for? (We didn't mention a time frame in the release notes)

It will definitely be for a year for [security](https://github.com/canonical/operator/blob/main/SECURITY.md), so we could link to that if we wanted. I suspect we might do longer than that in practice, but it'll depend how quickly people move. I don't think we've discussed how long we'll do the non-security fixes, so I think we should hold off promising anything for now (in the same way that "critical" is left as subjective).

dwilding: Thanks Tony. I'll leave it as-is until we've been able to discuss further.

PR #1866, opened 2025-06-30
Post Release

IronCore864:
This PR updates the version files after the release.

PR #1868, opened 2025-07-01
ci: fix ops and ops-scenario injection for uv charms

dimaqq:
On top of #1866 meant to fix virtual env in downstream charms we test against local ops and ops-scenario.

* observability charms
* published charms (this takes 6 hours to validate, maybe best as a separate PR?)


dimaqq:
@IronCore864 I propose that you cherry-pick the observability charms workflow fix.

Though not necessarily the published charms workflow--here I don't know what's better and won't know for sure for 6 hours.

I've kicked off the actions:
- https://github.com/dimaqq/operator/actions/runs/15991113860 (your version)
   - 8 failed, 1 test takes forever to time out (?)
- https://github.com/dimaqq/operator/actions/runs/15991096657 (this PR)
   - 8 failed, 1 test takes forever to time out (?)

So, it seems like published charms workflow change doesn't make any difference.

PR #1872, opened 2025-07-02
docs: update HACKING.md with how to bump Ops version in Charmcraft profiles

dwilding:
In https://github.com/canonical/charmcraft/pull/2312 we're switching Charmcraft profiles to use `pyproject.toml` for dependencies - specifically Ops - and include `uv.lock` in generated charms. After discussion within Charm Tech, we decided that we should update the deps and lockfile whenever there's a major new feature of Ops or we resolve a dependency issue.

This PR updates the Ops release process in HACKING.md with instructions for doing that.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -397,7 +397,7 @@ The point here is to encourage people to check out the full notes and to upgrade
 promptly, so ensure that you entice them with the best that the new versions
 have to offer.
 
-## Updating the Ops versions for Charmcraft profiles
+## Updating the Ops versions in the Charmcraft profiles
 
 The Charmcraft `kubernetes` and `machine` profiles specify a minimum Ops version in their `pyproject.toml` templates. If an Ops release includes a major new feature or resolves a dependency issue, open a PR to Charmcraft to increase the minimum Ops version in the profiles and refresh the `uv.lock` templates.
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -418,9 +418,9 @@ Next, do the following for the `kubernetes` profile:
 6. In `charmcraft/templates/init-kubernetes/uv.lock.j2`, replace `generated-temp` by `{{ name }}`.
 7. Delete the `generated-temp` directory.
 
-Repeat these steps for the `machine` profile, then commit your changes.
+You also need to update the `machine` templates. Run a diff between `charmcraft/templates/init-kubernetes/pyproject.toml.j2` and `charmcraft/templates/init-machine/pyproject.toml.j2`. If the files match, copy `pyproject.toml.j2` and `uv.lock.js` from the `kubernetes` profile to the `machine` profile. Otherwise, repeat the full process for the `machine` profile.
 
-You should have changed these files:
+Commit your changes. You should have changed these files:
 * charmcraft/templates/init-kubernetes/pyproject.toml.j2
 * charmcraft/templates/init-kubernetes/uv.lock.j2
 * charmcraft/templates/init-machine/pyproject.toml.j2
--- a/HACKING.md
+++ b/HACKING.md
@@ -418,7 +418,7 @@ Next, do the following for the `kubernetes` profile:
 6. In `charmcraft/templates/init-kubernetes/uv.lock.j2`, replace `generated-temp` by `{{ name }}`.
 7. Delete the `generated-temp` directory.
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -396,3 +396,19 @@ In the post, outline the key improvements both in `ops` and `ops-scenario`.
 The point here is to encourage people to check out the full notes and to upgrade
 promptly, so ensure that you entice them with the best that the new versions
 have to offer.
+
+# Updating the Ops version for Charmcraft profiles
+
+First set up a development version of Charmcraft per their CONTRIBUTING.md
```
dimaqq: I think the English style guide looks down on "per, etc., et.al."

HACKING.md
```diff
@@ -396,3 +396,19 @@ In the post, outline the key improvements both in `ops` and `ops-scenario`.
...
+
+Then for each profile (`kubernetes` and `machine`):
+
+1. Update Ops version in pyproject.toml.j2
+2. Make a directory called generated-temp
+3. Run `charmcraft init` inside generated-temp, specifying the profile
+    Use the development version of Charmcraft
```
dimaqq: Something is a little weird here.

dimaqq (commented):
Early fly-by.

dwilding:
This isn't ready for review yet @dimaqq, sorry, I should have left a note

dimaqq:
No worries, "Draft" is an indication enough.

HACKING.md
```diff
@@ -396,3 +396,32 @@ In the post, outline the key improvements both in `ops` and `ops-scenario`.
 The point here is to encourage people to check out the full notes and to upgrade
 promptly, so ensure that you entice them with the best that the new versions
 have to offer.
+
+## Updating the Ops versions for Charmcraft profiles
```
dwilding: This heading uses sentence case. The rest of the file uses a mix of sentence case and Title Case. Any objection if I include drive-by changes to make all headings sentence case?

dimaqq: I'd do that in a separate PR, because you'll most likely be tempted to fix language here and there next to headings!

dimaqq: Would this read more natural?

```suggestion
## Updating the Ops versions in the Charmcraft profiles
```

HACKING.md
```diff
@@ -396,3 +396,32 @@ In the post, outline the key improvements both in `ops` and `ops-scenario`.
...
+3. Inside `generated-temp`, run:
+    ```text
+    CHARMCRAFT_DEVELOPER=1 python -m charmcraft init --profile=kubernetes
+    ```
+4. Inside `generated-temp`, run `uv lock`.
+5. Copy `generated-temp/uv.lock` to `charmcraft/templates/init-kubernetes/uv.lock.j2`, overwriting the existing file.
+6. In `charmcraft/templates/init-kubernetes/uv.lock.j2`, replace `generated-temp` by `{{ name }}`.
+7. Delete the `generated-temp` directory.
+
+Repeat these steps for the `machine` profile, then commit your changes.
```
dimaqq: Today `pyproject.toml.j2` are verbatim copies.
I would expect same for the lock files.
Does it make sense to re-generate manually, or should it be better to just copy the content over?

HACKING.md
dwilding: Agree, done

HACKING.md
dwilding: I was wondering that too. I figured it's safer not to assume the files will always be the same between kubernetes/machine. But is that worrying too much about a hypothetical?

dimaqq: We've got a hand in the content of the templates, I figure we can decide this.

dwilding: Updated to this:

> You also need to update the `machine` templates. Run a diff between `charmcraft/templates/init-kubernetes/pyproject.toml.j2` and `charmcraft/templates/init-machine/pyproject.toml.j2`. If the files match, copy `pyproject.toml.j2` and `uv.lock.js` from the `kubernetes` profile to the `machine` profile. Otherwise, repeat the full process for the `machine` profile.

dimaqq (approved):

HACKING.md
dwilding: Whoops, I mean `uv.lock.j2` not `uv.lock.js`. Fixed

dwilding: I just realised that doing a diff won't work right, because one template file has already been updated. Changed to this:

> For the `machine` profile, modify the Ops version specifier in `charmcraft/templates/init-machine/pyproject.toml.j2`. Then run a diff between `.../init-machine/pyproject.toml.j2` and `.../init-kubernetes/pyproject.toml.j2`. If the files match, copy `pyproject.toml.j2` and `uv.lock.j2` from the `kubernetes` profile to the `machine` profile. Otherwise, repeat the full process for the `machine` profile.

dwilding: Gah, and of course there's no need to copy `pyproject.toml.j2` because the files match 🤦

I'm going to do something else for a while

IronCore864 (approved):

PR #1873, opened 2025-07-02
chore: remove 3.8 and update to 3.10

IronCore864:
Remove Python 3.8.

- Bump Python version to 3.10 in `pyproject.toml`
- Update tags that trigger publish to `3.*`
- Remove Python 3.8 (and 3.9 for some tests) from GitHub Actions
- Update pyright config to 3.10

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -47,9 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.10", "3.12"]
-        exclude:
-        - {python-version: "3.8", os: "macos-latest"}  # macos-14 is arm64, and there's no Python 3.8 build for arm64
+        python-version: ["3.10", "3.12"]
 
     steps:
       - uses: actions/checkout@v4
@@ -72,7 +70,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.8", "3.10", "3.12"]
+        python-version: ["3.10", "3.12"]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


IronCore864:
o11y tests failed because they all have `requires-python = "~=3.8"`

dimaqq (approved):

IronCore864:
Pack failed because it tries to pack for 20.04. Tried locally to pack an example charm and it is OK.

james-garner-canonical (approved):
Looks good. One small suggestion: add a `FIXME`/`TODO` comment to the `tool.ruff.target-version`. Optional.

amandahla:
Hi, is there a specific reason for this change?
We cant update Ops in this charm, for example
https://github.com/canonical/indico-operator/pull/644

tonyandrewmeyer:
> Hi, is there a specific reason for this change?

Python 3.8 has been [out of support](https://devguide.python.org/versions/) since October last year, and 20.04 is [out of standard support](https://ubuntu.com/about/release-cycle) as of May. The Python ecosystem is dropping support for Python 3.8, making it increasingly difficult to use a modern toolchain and rely on fixes for dependencies.

This is roughly the same timeframe as dropping support for 18.04 in Ops 2.0. That was released in January 2023, 18.04 was out of standard support in May 2023, and Python 3.5 out of support in September 2020.

[Juju 4 also drops support for 20.04](https://discourse.canonical.com/t/charming-engineering-lt-notes-4-jun-2025/5772?u=tony-meyer), with Juju 3 [expected to go into security fixes only in May next year](https://documentation.ubuntu.com/juju/3.6/reference/juju/juju-roadmap-and-releases/).

If your charms need to support 20.04 still, then Ops 2.23 is available, and will receive critical bug fixes and security fixes for at least a year (and there will presumably be some avenue for security fixes until 20.04 is completely unsupported in 2030). Our recommendation is that you branch your code, leaving the branch that supports 20.04 with Ops 2 and moving main to Ops 3, so that you can use new features -- but you can just remain on Ops 2 if a single code base is preferable.

Issue #1877, opened 2025-07-02
Do something with observability tests.


dimaqq:
The downstream repos require `python_version=">=3.8"` and thus ops3 cannot be installed in the virtualenv.

tonyandrewmeyer:
I expect we can just replace the python-requires just as we replace the ops version. We know that these charms will work with 3.10.

Issue #1876, opened 2025-07-02
Probably drop hello-kubecon test


dimaqq:
The repo is archived and the base is 20.04.
This can't work with ops 3.

benhoyt:
We've dropped this already, closing.

PR #1878, opened 2025-07-02
ci: disable non-essential checks that need adjustments to work with a Python 3.10 minimum

tonyandrewmeyer:
To be resolved by:
* Pack (hello-kubecon): https://github.com/canonical/operator/issues/1876
* Observability compatibility tests: https://github.com/canonical/operator/issues/1877

We're confident that nothing will have changed in Ops that breaks these, and they aren't required to merge, but we should get them green (by explicitly disabling them, as here) before we do the next release.

### DIFF
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -15,8 +15,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - charm-repo: jnsgruk/hello-kubecon
-            commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
+#          - charm-repo: jnsgruk/hello-kubecon
+#            commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
     steps:
       - name: Checkout test charm repository
         uses: actions/checkout@v4
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        include:
+        include: []
 #          - charm-repo: jnsgruk/hello-kubecon
 #            commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
     steps:
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -39,6 +39,7 @@ jobs:
         run: pip install tox~=4.2 uv~=0.6
 
       - name: Update 'ops' and 'ops-scenario' dependencies in test charm to latest
+        if: ${{ !(matrix.disabled) }}
         run: |
           if [ -e "requirements.txt" ]; then
             sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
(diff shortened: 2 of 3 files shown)


james-garner-canonical (approved):

.github/workflows/observability-charm-tests.yaml
```diff
@@ -36,6 +39,7 @@ jobs:
         run: pip install tox~=4.2 uv~=0.6
 
       - name: Update 'ops' and 'ops-scenario' dependencies in test charm to latest
+        if: ${{ !(matrix.disabled) }}
```
dwilding, line 42: Is it possible/worth using this pattern in charmcraft-pack.yaml too?

.github/workflows/charmcraft-pack.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - charm-repo: jnsgruk/hello-kubecon
-            commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
+#          - charm-repo: jnsgruk/hello-kubecon
+#            commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
```
dwilding, line 19: From #1876 and discussion elsewhere, it seems unlikely that we'd update this charm to work on 22.04. Shall we fully remove the check now, rather than commenting it out, then #1876 can (later) add a fresh check for one of our in-repo example charms?

dwilding (approved):

.github/workflows/charmcraft-pack.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
     strategy:
       matrix:
         include:
```
dimaqq: Do we need a value here?

```suggestion
        include: []
```

YAML `foo: ` means `foo: null` which is a different type from before.
I dunno how strict is GitHub when validating against the schema.

.github/workflows/observability-charm-tests.yaml
tonyandrewmeyer: I'm not sure. We added it to the third-party charms because sometimes they break out of our control (something in the upstream changes) and we wanted a way to skip them while still letting the pin update action do its magic (previously, we just commented items out).

Not in this PR, though.

I think in general if we want to check packing, we should do that with 1+ charms that are in our repo. We're already doing this in the smoke tests anyway, we could just promote those to be a release check, or re-use those, or use one/more of the example charms, etc. If we do that then we shouldn't need to disable them, because if they break, that's our fault.

.github/workflows/charmcraft-pack.yaml
dimaqq: +1 from me on removing this entire file.

.github/workflows/charmcraft-pack.yaml
dimaqq: https://github.com/canonical/operator/actions/runs/16018066229
```
[Invalid workflow file: .github/workflows/charmcraft-pack.yaml#L17](https://github.com/canonical/operator/actions/runs/16018066229/workflow)
The workflow is not valid. .github/workflows/charmcraft-pack.yaml (Line: 17, Col: 17): Unexpected value ''
```

.github/workflows/charmcraft-pack.yaml
tonyandrewmeyer: I'd prefer to keep this change minimal, just acknowledging explicitly that we understand why it's not working and that we're ok with that for the 3.0 release, and then dealing with the longer term fixes before the next release, but in the other ticket for it.

I think commenting it out is a clearer statement than removing the entire workflow. We aren't intending to stop testing packing, we're just going to update what we pack.

Issue #1880, opened 2025-07-02
Improve stock logger with a more useful formatter


addyess:
By default, the ops logger creates uses the Default Formatter which is lacking in detail when traversing large charms.  Its not always clear where the logger messages come from

Consider this bit of charm code:

```python

import logging

log = logging.getlogger(__name__)


...

    log.warning("I need attention")
```

It's a horrible log message right?  But when we look in the juju logs, there's no idea where it comes from:

```
unit-charm-0: 14:04:52 WARN unit.charm/0.juju-log I need attention
```

It would be nicer for the charm to allow for more custom logging.  

in my own charm I've instrumented this:

```python
def improve_root_logger():
    """Improve the JujuLogHandler to include logger name and lineno."""
    root_logger = logging.getLogger()
    for handler in root_logger.handlers:
        if isinstance(handler, ops.log.JujuLogHandler):
            handler.setFormatter(logging.Formatter("%(name)s:%(lineno)-5d %(message)s"))
```

so this message would now show in the logs like this:

```
unit-charm-0: 14:04:52 WARN unit.charm/0.juju-log my.pitiful.library:12345 I need attention
```

addyess:
i'd hoped to do this:

```diff
 @ ops/log.py:34 @ class JujuLogHandler(logging.Handler):
     def __init__(self, model_backend: _ModelBackend, level: int = logging.DEBUG):
         super().__init__(level)
         self.model_backend = model_backend
+        self.setFormatter(logging.Formatter("%(name)s:%(lineno)-5d %(message)s"))

     def emit(self, record: logging.LogRecord):
         """Send the specified logging record to the Juju backend.
```


but it blows up the unit tests

tonyandrewmeyer:
The data team also like an alternative format (for example, [charm-refresh includes the logger name](https://canonical-charm-refresh--5.com.readthedocs.build/latest/add-to-charm/log-format/)). It does seem like the logger name in particular is useful as soon as a charm is a non-trivial size.

It's not a significant amount of code to just have the charm do this itself (as in the description and in charm-refresh).

jhack ([example](https://github.com/canonical/jhack/issues/218)) and potentially other tooling relies on a consistent format. If we allow a custom formatter, unless we accept the formatter in the `ops.main` call, there will also be some inconsistency, with early log messages using one formatter and later ones a different one. We might want to distinguish between logs that ops emits and logs that the charm does.

[...]

PR #1881, opened 2025-07-02
feat: Provide an option to change the juju-log formatting

addyess:
Addresses #1880 

Provides the charmer to adjust the formatter to something more pleasant without forcing it on others. 

# Usage

```python
class MyCharm(ops.CharmBase):
    def __init__(self, *args):
        ops.log.JujuLogHandler.set_formatting(ops.log.JujuLogHandler.VERBOSE_FORMATTING)
```

### DIFF
--- a/ops/log.py
+++ b/ops/log.py
@@ -28,6 +28,17 @@ from .model import _ModelBackend
 class JujuLogHandler(logging.Handler):
     """A handler for sending logs and warnings to Juju via juju-log."""
 
+    DEFAULT_FORMATTING = None
+    VERBOSE_FORMATTING = '%(name)s:%(lineno)-5d %(message)s'
+
+    @classmethod
+    def set_formatting(cls, format: str | None = DEFAULT_FORMATTING):
+        """Adjust the log format output."""
+        root_logger = logging.getLogger()
+        for handler in root_logger.handlers:
+            if isinstance(handler, cls):
+                handler.setFormatter(logging.Formatter(format or cls.DEFAULT_FORMATTING))
+
     def __init__(self, model_backend: _ModelBackend, level: int = logging.DEBUG):
         super().__init__(level)
         self.model_backend = model_backend
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -82,6 +82,38 @@ class TestLogging:
         calls = backend.calls(clear=True)
         assert calls == [result]
 
+    @pytest.mark.parametrize(
+        'message,result',
+        [
+            ('critical', ('CRITICAL', 'critical')),
+            ('error', ('ERROR', 'error')),
+            ('warning', ('WARNING', 'warning')),
+            ('info', ('INFO', 'info')),
+            ('debug', ('DEBUG', 'debug')),
+        ],
+    )
+    def test_verbose_logging(
+        self,
+        backend: FakeModelBackend,
+        logger: logging.Logger,
+        message: str,
+        result: tuple[str, str],
+    ):
+        ops.log.setup_root_logging(backend)
+        ops.log.JujuLogHandler.set_formatting(ops.log.JujuLogHandler.VERBOSE_FORMATTING)
+        assert logger.level == logging.DEBUG
+        assert isinstance(logger.handlers[-1], ops.log.JujuLogHandler)
+
+        method = getattr(logger, message)
+        method(message)
+        calls = backend.calls(clear=True)
+        level, message = result
+        expected = re.compile(rf'{logger.name}:\d+\s+{message}')
+
+        actual_level, actual_message = calls[0]
+        assert level == actual_level
+        assert expected.match(actual_message)
+
     def test_handler_filtering(self, backend: FakeModelBackend, logger: logging.Logger):
         logger.setLevel(logging.INFO)
         logger.addHandler(ops.log.JujuLogHandler(backend, logging.WARNING))


addyess:
Thanks for a look @benhoyt or @dimaqq

dimaqq:
Ben is on vacation, so let's ping @tonyandrewmeyer instead.

ops/log.py
```diff
@@ -28,6 +28,17 @@
 class JujuLogHandler(logging.Handler):
     """A handler for sending logs and warnings to Juju via juju-log."""
 
+    DEFAULT_FORMATTING = None
+    VERBOSE_FORMATTING = '%(name)s:%(lineno)-5d %(message)s'
```
dimaqq, line 32: Hot take with a dash of sarcasm: expect less pretty output at 100K LOC?

ops/log.py
```diff
@@ -28,6 +28,17 @@
...
+
+    @classmethod
+    def set_formatting(cls, format: str | None = DEFAULT_FORMATTING):
+        """Adjust the log format output."""
+        root_logger = logging.getLogger()
+        for handler in root_logger.handlers:
+            if isinstance(handler, cls):
+                handler.setFormatter(logging.Formatter(format or cls.DEFAULT_FORMATTING))
```
dimaqq, line 40: I'm a little confused here. DEAULT is None, so why `format or DEFAULT`?
Is this to escape user supplying an empty string?

test/test_log.py
```diff
@@ -82,6 +82,38 @@ def test_default_logging(
...
+        ops.log.setup_root_logging(backend)
+        ops.log.JujuLogHandler.set_formatting(ops.log.JujuLogHandler.VERBOSE_FORMATTING)
+        assert logger.level == logging.DEBUG
+        assert isinstance(logger.handlers[-1], ops.log.JujuLogHandler)
+
+        method = getattr(logger, message)
+        method(message)
+        calls = backend.calls(clear=True)
+        level, message = result
+        expected = re.compile(rf'{logger.name}:\d+\s+{message}')
```
dimaqq, line 111: While I like the test, I would prefer a test vector.
Like the exact log line produced with default and verbose formats.

If that's somehow hard (e.g. time), then I'd like to see examples in a comment.

ops/log.py
```diff
@@ -28,6 +28,17 @@
 class JujuLogHandler(logging.Handler):
     """A handler for sending logs and warnings to Juju via juju-log."""
 
+    DEFAULT_FORMATTING = None
+    VERBOSE_FORMATTING = '%(name)s:%(lineno)-5d %(message)s'
+
+    @classmethod
+    def set_formatting(cls, format: str | None = DEFAULT_FORMATTING):
+        """Adjust the log format output."""
```
dimaqq, line 36: Better docstring is needed for a public function.

ops/log.py
dimaqq: Also docstring is needed under the public constant.

ops/log.py
dimaqq, line 40: I'm not a fan of this loop. Can this be better?

Also, it seems that early logging would use the default format and only later the format would be switched. Am I reading this right?

dimaqq (commented):
Please describe why a different format is needed?
What kind of format is needed?
Are only DEFAULT and VERBOSE going to be used, or do you envision a very custom format?

How would a given charm decide when to toggle the format?
Do you expect every app in a deployment to use the same format, or are disparate formats good enough?

Considering all the other ops users, I feel that there needs to be a good reason to tweak the format:
- today we don't make any promised on the exact format
- with this change, we'd on the hook to support the verbose format forever?

ops/log.py
tonyandrewmeyer: > Also, it seems that early logging would use the default format and only later the format would be switched. Am I reading this right?

I'm not a big fan of this either, although charms could obviously switch up the formatting at any point anyway.

I think perhaps we ought to be consistently using `getLogger(__name__)` in the ops modules (e.g. we do in charm and model, we don't in storage), and we should have a consistent format for those, which means we'd have a handler that is specific for those. We could then leave the root logger for charms to use and they can configure the root logger's (juju-log) handler's formatter to be whatever they want.

addyess: Yes, an empty string would render the logs silent. Every message would be:

```diff 
- unit-my-charm-0: 13:05:35 INFO unit.my-charm/0.juju-log something the charm wishes to log
+ unit-my-charm-0: 13:05:35 INFO unit.my-charm/0.juju-log
```

test/test_log.py
addyess: i think i didn't do precisely that because the line-number of the call to line 108

You end up with a log message that looks something like this.

```
test_log:108   critical
```

ops/log.py
addyess: It's fine... at least the PR gets ideas rolling.  I''m mostly just currently bummed by not having the following in my logs: `%(name)s` and `%(lineno)s`.  If you want to change anything around underneath -- fine with me.  It's just tough at the moment to not have these two items in my logs. 

Here's a sample from the k8s charm with and without it
```diff
- unit-k8s-0: 13:05:35 INFO unit.k8s/0.juju-log k8s/0(juju-04ff10-noble-0) current cluster-name=k8s-bd08d4f998c4e658c2c4139659f824ac
- unit-k8s-0: 13:06:24 INFO unit.k8s/0.juju-log Feature 'load-balancer' enabled=False,deployed=False,ver=v0.14.9-ck0,updated_at=2025-07-01 22:01:42+00:00: disabled
+ unit-k8s-0: 13:05:35 INFO unit.k8s/0.juju-log __main__:384   k8s/0(juju-04ff10-noble-0) current cluster-name=k8s-bd08d4f998c4e658c2c4139659f824ac
+ unit-k8s-0: 13:06:24 INFO unit.k8s/0.juju-log events.update_status:128   Feature 'load-balancer' enabled=False,deployed=False,ver=v0.14.9-ck0,updated_at=2025-07-01 22:01:42+00:00: disabled
```

addyess: all the stuff `juju-log` puts out front is *nice* i guess.
* I like the timestamp, 
* I like the level. 

I don't know why the unit name is in there two times?  `unit-k8s-0` and `unit-k8s/0.juju-log`.  It feels cluttered
But `ops` can't change any of that...

you have control over everything past `juju-log` here -- lets make the most of that real-estate

addyess:
Closed based on the comments on the issue leaning towards a more unified solution rather than quick-fix of my very mundane problem

PR #1853, opened 2025-06-26
docs: ops tracing how to

dimaqq:

PR #1883, opened 2025-07-03
chore: create a span for ops.tracing.Tracing init

dimaqq:
In #1853, I argue that non-trivial charm library init should be wrapped in a span.

This PR does so for our first party `ops.tracing.Tracing` charm lib.

### DIFF
--- a/tracing/ops_tracing/_api.py
+++ b/tracing/ops_tracing/_api.py
@@ -116,7 +116,7 @@ class Tracing(ops.Object):
             if relation.interface_name != 'tracing':
                 raise ValueError(
                     f"{tracing_relation_name=} {relation.interface_name=} when 'tracing' is"
-                    ' expected'
+                    f' expected'
                 )
 
             self._tracing = TracingEndpointRequirer(
@@ -145,7 +145,7 @@ class Tracing(ops.Object):
                 if relation.interface_name != 'certificate_transfer':
                     raise ValueError(
                         f'{ca_relation_name=} {relation.interface_name=} when'
-                        " 'certificate_transfer' is expected"
+                        f" 'certificate_transfer' is expected"
                     )
 
                 self._certificate_transfer = CertificateTransferRequires(charm, ca_relation_name)
--- a/tracing/ops_tracing/_api.py
+++ b/tracing/ops_tracing/_api.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import logging
 
+import opentelemetry.trace
 import ops
 
 from ._buffer import Destination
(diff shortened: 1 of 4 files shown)


tracing/ops_tracing/_api.py
```diff
@@ -91,68 +93,70 @@ def __init__(
...
+                    f"{tracing_relation_name=} {relation.role=} when 'requires' is expected"
                 )
-            if relation.interface_name != 'certificate_transfer':
+
+            if relation.interface_name != 'tracing':
                 raise ValueError(
-                    f"{ca_relation_name=} {relation.interface_name=} when 'certificate_transfer' "
-                    'is expected'
+                    f"{tracing_relation_name=} {relation.interface_name=} when 'tracing' is"
+                    ' expected'
```
tonyandrewmeyer: ruff might disagree with me here (and if so it should win), and I'm not sure if this is Charm Tech style or not, but my preference is to keep the same style strings when concatenating like this, unless infeasible.

```suggestion
                    f"{tracing_relation_name=} {relation.interface_name=} when 'tracing' is"
                    f" expected"
```

Same for the one below.

dimaqq: PTAL, I can do `f"foo''" f'bar'` but ruff won't let me use the "wrong" kind of quotation, like `"foo" 'bar'`.

tracing/ops_tracing/_api.py
```diff
@@ -91,68 +93,70 @@ def __init__(
...
-            self.charm.on.start,
-            self.charm.on.upgrade_charm,
-            self._tracing.on.endpoint_changed,
-            self._tracing.on.endpoint_removed,
-        ):
-            self.framework.observe(event, self._reconcile)
-
-        if ca_relation_name:
-            relation = self.charm.meta.relations.get(ca_relation_name)
+        with tracer.start_as_current_span('ops.tracing.Tracing'):
```
james-garner-canonical, line 96: WDYT about using this as a decorator for `__init__` instead?

dimaqq: We can do that, however, that has a couple of issues:
- if someone uses somewhat older opentelemetry-api in their charm deps, IDE and static analysers can't see the types/arguments... let me check that, maybe we're out of the woods since we dropped py 3.8 and 3.9
- an extra function call is shown in any traceback that includes the call to the decorated thing

james-garner-canonical: I see why you went with the context manager instead.

james-garner-canonical (approved):

dimaqq:
fyi, the opentelemetry-api typing bug was https://github.com/open-telemetry/opentelemetry-python/issues/3836

while it's not an issue since we moved to Py 3.10 lower bound, I would stay away from decorators for some time in case charmers use ops code as an example of the right way to instrument things.

PR #1887, opened 2025-07-04
fix: if an event ends with _Abort(0) tests should behave as if it ended successfully

tonyandrewmeyer:
If ops code raises `_Abort(0)`, then the event completes successfully as far as Juju is concerned. The fact that ops has shortcut things to jump right to exit isn't really relevant from the Juju point of view.

This means that for Scenario tests, we should treat `_Abort(0)` as a normal successful event completion, because it's simulating the events from the point of view of Juju.

Raising `_Abort` with any other exit code still ends up as `UncaughtCharmError`, as that indicates that something has gone wrong, and from the Juju perspective changes will be rolled back and the charm will end up in error status.

### DIFF
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -30,6 +30,7 @@ from ops import (
     PreCommitEvent,
 )
 from ops.jujucontext import _JujuContext
+from ops._main import _Abort
 from ops._private.harness import ActionFailed
 
 from .errors import NoObserverError, UncaughtCharmError
@@ -345,6 +346,10 @@ class Runtime:
 
             except (NoObserverError, ActionFailed):
                 raise  # propagate along
+            except _Abort as e:
+                # If ops raised _Abort(0) then we want to treat that as normal completion.
+                if e.exit_code != 0:
+                    raise
             except Exception as e:
                 # The following is intentionally on one long line, so that the last line of pdb
                 # output shows the error message (pdb shows the "raise" line).
--- a/testing/tests/test_runtime.py
+++ b/testing/tests/test_runtime.py
@@ -6,8 +6,9 @@ from tempfile import TemporaryDirectory
 import pytest
 
 import ops
+from ops._main import _Abort
 
-from scenario import Context
+from scenario import Context, ActiveStatus
 from scenario.state import Relation, State, _CharmSpec, _Event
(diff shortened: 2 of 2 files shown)


james-garner-canonical (approved):
As per discussion, this sounds like a nice improvement for scenario tests. The instances of `_Abort(0)` prior to the introduction of the typed data methods we recently added (and which we're supporting with this test) were only two cases:

1. When `'OPERATOR_DISPATCH'` was set in the environment
2. When running in a restricted context with `use_juju_for_storage`

I don't think either of these are modelled by scenario, is that right? In which case, this shouldn't change the behaviour of any tests other than those exercising the new typed data methods.

Nevertheless, it would be good to verify if this change does introduce any test changes -- perhaps by temporarily enabling Python 3.8 and running super-tox.

james-garner-canonical:
This change should be contributed to the Ops 2 maintenance branch too, since it's a fix, and would otherwise introduce a divergence in test behaviour.

tonyandrewmeyer:
> This change should be contributed to the Ops 2 maintenance branch too, since it's a fix,

It's a fix, but I wouldn't call it a "critical bugfix", which is what we promise for 2.x.

dimaqq (approved):

tonyandrewmeyer:
> 1. When `'OPERATOR_DISPATCH'` was set in the environment

In theory I think you could get this case with a Scenario test, if you have the old-school non-dispatch links set up, because Scenario does the ops `_Manager` `__init__`, which is where the dispatch is done.

However, I would argue that in the (extraordinarily unlikely) case that you have this in your test, the more correct behaviour is for it to exit successfully, as would happen within real Juju.

> 2. When running in a restricted context with `use_juju_for_storage`

A charm that uses `use_juju_for_storage=True` can run tests with Scenario, but that field is ignored and the data is stored as if it was False (there are not `state-set` or `state-get` calls). There's no `ctx.on.add_metrics` event and no `ctx.on.meter_status_changed` event on `ctx.on` so you can't test those events unless you create an `_Event` object yourself, which you are not meant to. So I think this doesn't apply, even if we add `juju_for_storage` testing support in the future.

> Nevertheless, it would be good to verify if this change does introduce any test changes -- perhaps by temporarily enabling Python 3.8 and running super-tox.

I'm positive no-one has tests that would be changed here.

PR #1889, opened 2025-07-04
chore: bump zizmor to 1.11

tonyandrewmeyer:
Bumps Zizmor to 1.11 both in pre-commit and the workflow. There are various [new checks](https://docs.zizmor.sh/release-notes/) as well as a fix mode, but no new issues are found with our workflows.

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -23,7 +23,7 @@ jobs:
         uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb  # v6.1.0
 
       - name: Run zizmor
-        run: uvx zizmor@v1.9.0 --format=sarif . > results.sarif
+        run: uvx zizmor@v1.11.0 --format=sarif . > results.sarif
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,6 @@ repos:
         - tomli
   # Check workflow security.
   - repo: https://github.com/woodruffw/zizmor-pre-commit
-    rev: v1.9.0
+    rev: v1.11.0
     hooks:
     - id: zizmor


james-garner-canonical (approved):

.github/workflows/zizmor.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb  # v6.1.0
 
       - name: Run zizmor
-        run: uvx zizmor@v1.9.0 --format=sarif . > results.sarif
+        run: uvx zizmor@v1.11.0 --format=sarif . > results.sarif
```
dwilding, line 26: Would it be better to use v1.11.0 or v1.11 in this context? (not implying anything - asking for my own education)

dwilding (approved):

.github/workflows/zizmor.yaml
tonyandrewmeyer: @james-garner-canonical pinned it to 1.9.0, and I was following that. We do want the pre-commit version to be the same as the one run in the workflow, otherwise we'll get annoying inconsistencies. Maybe something or someone wanted an exact pin inside of the workflow?

There is a [zizmor action](https://github.com/zizmorcore/zizmor-action) now, which we might want to switch to in the future. The default `version` value there is `latest`, so it seems like their position is to not pin?

dimaqq: May as well `uvx zizmor~=1.11 -V` --> 1.11.0 that way we don't have to make these manual bumps.

Also... WDYT about adding zizmor into some dependency group?
The source is already checked out, may as well `uv run` it instead of `uvx run??version`

dimaqq: Alternatively still:
```bash
cargo install zizmor
zizmor -V
```

It's all rust, the pypi package is py-any, with per-os and per-libc wheels... basically distribution on pypi is convenience only.

tonyandrewmeyer: It seems like just moving to the action would be the simpler change.

dimaqq: I've asked upstream: https://discord.com/channels/1366889069551812668/1366889070021841019/1390572122199298212

dimaqq: There doesn't seems to be a first-party zizmor action.
`uvx` is one the recommended ways to run it, so be it.

dimaqq (approved):
Fine with the change; we can always move zizmor into a dep group later in #1892

.github/workflows/zizmor.yaml
tonyandrewmeyer: > There doesn't seems to be a first-party zizmor action.

There's a link to it 6 comments up in this thread :smile:. It's also the [recommendation in their docs](https://docs.zizmor.sh/integrations/).

PR #1891, opened 2025-07-04
docs: Use config and action classes in the Kubernetes tutorial

tonyandrewmeyer:
1. In the config chapter, we add a config class (with the single option) and add validation for it. I've put the loading into the pebble layer property, as that's (initially) the only place that it's used and it seemed cleanest for the tutorial. The invalid value handling is then moved to a small except block in the update and restart code.
2. In following chapters, continue with the approach for config, but when we move to collect-status also add validation of the value there. In the observability chapter, also handle getting the config for metrics.
3. In the actions chapter, we add an action class (with the single parameter of the action), and use it when loading the parameters. This is a very minimal change, but demonstrates the use of the feature, and encourages continuing with it when there are more complex actions. This uses `errors='fail'` because that is appropriate for actions with most charms and is cleaner code -- in this specific case, Juju will prevent any invalid value, so there will never be a failure.

pre-commit also fixed some unrelated trailing whitespace - you probably want to review with whitespace changes removed.

[Preview](https://ops--1891.org.readthedocs.build/en/1891/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/index.html)

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
@@ -13,12 +13,12 @@ This document is part of a  series, and we recommend you follow it in sequence. 
 git clone https://github.com/canonical/juju-sdk-tutorial-k8s.git
 cd juju-sdk-tutorial-k8s
 git checkout 03_integrate_with_psql
-git checkout -b  04_create_actions 
+git checkout -b  04_create_actions
 ```
 
 ````
 
-A charm should ideally cover all the complex operational logic within the code, to help avoid the need for manual human intervention. 
+A charm should ideally cover all the complex operational logic within the code, to help avoid the need for manual human intervention.
 
 Unfortunately, that is not always possible. As a charm developer, it is thus useful to know that you can also expose charm operational tasks to the charm user by defining special methods called `actions`.
 
@@ -40,6 +40,22 @@ actions:
         type: boolean
         default: False
 ```
+
+## Define an action class
+
+Open your `src/charm.py` file, and add an action class that matches the definition you used in `charmcraft.yaml`:
+
+```python
+# Note that this action is also defined in charmcraft.yaml
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class GetDbInfoAction:
+    """Fetches database authentication information."""
+
+    show_password: bool
+    """Show username and password in output information."""
+```
+
+Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameters, and static type checkers are able to validate that we are using the parameters correctly.
 
 ## Define the action event handlers
 
@@ -69,7 +85,7 @@ def _on_get_db_info_action(self, event: ops.ActionEvent) -> None:
 
     Learn more about actions at https://ops.readthedocs.io/en/latest/howto/manage-actions.html
     """
-    show_password = event.params['show-password']  # see charmcraft.yaml
+    params = event.load_params(GetDbInfoAction, errors='fail')
     db_data = self.fetch_postgres_relation_data()
     if not db_data:
         event.fail('No database connected')
@@ -99,7 +115,7 @@ juju refresh \
 ```
 
 Next, test that the basic action invocation works:
- 
+
 ```text
 juju run demo-api-charm/0 get-db-info
 ```
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
@@ -54,7 +54,7 @@ class GetDbInfoAction:
     """Show username and password in output information."""
 ```
 
-Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameter, and static type checkers are able to validate that we are using the parameter correctly.
+We'll use [](ActionEvent.load_params) to create an instance of your config class from the Juju action event. This allows IDEs to provide hints when we are accessing the action parameter, and static type checkers are able to validate that we are using the parameter correctly.
 
 ## Define the action event handlers
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
@@ -94,7 +94,7 @@ def _on_get_db_info_action(self, event: ops.ActionEvent) -> None:
         'db-host': db_data.get('db_host', None),
         'db-port': db_data.get('db_port', None),
     }
-    if show_password:
+    if params.show_password:
         output.update({
             'db-username': db_data.get('db_username', None),
             'db-password': db_data.get('db_password', None),
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
@@ -55,7 +55,7 @@ class GetDbInfoAction:
     """Show username and password in output information."""
 ```
 
-Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameters, and static type checkers are able to validate that we are using the parameters correctly.
+Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameter, and static type checkers are able to validate that we are using the parameter correctly.
 
 ## Define the action event handlers
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
@@ -46,7 +46,6 @@ actions:
 Open your `src/charm.py` file, and add an action class that matches the definition you used in `charmcraft.yaml`:
 
 ```python
-# Note that this action is also defined in charmcraft.yaml
 @dataclasses.dataclass(frozen=True, kw_only=True)
 class GetDbInfoAction:
     """Fetches database authentication information."""
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -236,7 +236,7 @@ def app_environment(self) -> dict[str, str]:
 Finally, let's define the method that is called on the database created event:
 
 ```python
-def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+def _on_database_created(self, _: DatabaseCreatedEvent) -> None:
     """Event is fired when postgres database is created."""
     self._update_layer_and_restart()
 ```
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -19,7 +19,7 @@ This document is part of a  series, and we recommend you follow it in sequence. 
 git clone https://github.com/canonical/juju-sdk-tutorial-k8s.git
 cd juju-sdk-tutorial-k8s
 git checkout 02_make_your_charm_configurable
-git checkout -b 03_integrate_with_psql 
+git checkout -b 03_integrate_with_psql
 ```
 
 ````
@@ -109,7 +109,7 @@ and not
 from lib.charms.data_platform_libs...
 ```
 
-The former is not resolvable by default but everything works fine when the charm is deployed. Why? Because the `dispatch` script in the packed charm sets the `PYTHONPATH` environment variable to include the `lib` directory when it executes your `src/charm.py` code. This tells Python it can check the `lib` directory when looking for modules and packages at import time. 
+The former is not resolvable by default but everything works fine when the charm is deployed. Why? Because the `dispatch` script in the packed charm sets the `PYTHONPATH` environment variable to include the `lib` directory when it executes your `src/charm.py` code. This tells Python it can check the `lib` directory when looking for modules and packages at import time.
 
 If you're experiencing issues with your IDE or just trying to run the `charm.py` file on your own, make sure to set/update `PYTHONPATH` to include `lib` directory as well.
 
@@ -125,7 +125,7 @@ export PYTHONPATH=lib:$PYTHONPATH
 ### Add relation event observers
 
 Next, in the `__init__` method, define a new instance of the 'DatabaseRequires' class. This is required to set the right permissions scope for the PostgreSQL charm. It will create a new user with a password and a database with the required name (below, `names_db`), and limit the user permissions to only this particular database (that is, below, `names_db`).
-  
+
 
 ```python
 # The 'relation_name' comes from the 'charmcraft.yaml file'.
@@ -181,11 +181,12 @@ Our application consumes database authentication information in the form of envi
 @property
 def _pebble_layer(self) -> ops.pebble.Layer:
     """A Pebble layer for the FastAPI demo services."""
+    config = self.load_config(FastAPIConfig)
     command = ' '.join([
         'uvicorn',
         'api_demo_server.app:app',
         '--host=0.0.0.0',
-        f'--port={self.config["server-port"]}',
+        f'--port={config.server_port}',
     ])
     pebble_layer: ops.pebble.LayerDict = {
         'summary': 'FastAPI demo service',
@@ -242,7 +243,7 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
 
 The diagram below illustrates the workflow for the case where the database relation exists and for the case where it does not:
 
-![Integrate your charm with PostgreSQL](../../resources/integrate_your_charm_with_postgresql.png) 
+![Integrate your charm with PostgreSQL](../../resources/integrate_your_charm_with_postgresql.png)
 
 ## Update the unit status to reflect the relation state
 
@@ -261,9 +262,10 @@ And define a method that does the various checks, adding appropriate statuses. T
 
 ```python
 def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
-    port = self.config['server-port']
-    if port == 22:
-        event.add_status(ops.BlockedStatus('Invalid port number, 22 is reserved for SSH'))
+    try:
+        self.load_config(FastAPIConfig)
+    except ValueError as e:
+        event.add_status(ops.BlockedStatus(str(e)))
     if not self.model.get_relation('database'):
         # We need the user to do 'juju integrate'.
         event.add_status(ops.BlockedStatus('Waiting for database relation'))
@@ -281,23 +283,18 @@ def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
     event.add_status(ops.ActiveStatus())
 ```
 
-We also want to clean up the code to remove the places where we're setting the status outside of this method, other than anywhere we're wanting a status to show up *during* the event execution (such as `MaintenanceStatus`). In `_on_config_changed`, change the port 22 check to:
+We also want to clean up the code to remove the places where we're setting the status outside of this method, other than anywhere we're wanting a status to show up *during* the event execution (such as `MaintenanceStatus`). In `_update_layer_and_restart`, remove the lines:
 
 ```python
-    if port == 22:
-        # The collect-status handler will set the status to blocked.
-        logger.debug('Invalid port number: 22 is reserved for SSH')
-        return
-```
-
-And remove the following lines from `_update_layer_and_restart`:
-
-```python
-    self.unit.status = ops.ActiveStatus()
+self.unit.status = ops.ActiveStatus()
 ```
 
 ```python
-    self.unit.status = ops.MaintenanceStatus('Waiting for Pebble in workload container')
+self.unit.status = ops.MaintenanceStatus('Waiting for Pebble in workload container')
+```
+
+```python
+self.unit.status = ops.BlockedStatus(str(e))
 ```
 
 ## Validate your charm
@@ -341,20 +338,20 @@ Model        Controller           Cloud/Region        Version  SLA          Time
 charm-model  tutorial-controller  microk8s/localhost  3.0.0    unsupported  13:50:39+01:00
 
 App             Version  Status  Scale  Charm           Channel  Rev  Address         Exposed  Message
-demo-api-charm  0.0.9    active      1  demo-api-charm             1  10.152.183.233  no       
+demo-api-charm  0.0.9    active      1  demo-api-charm             1  10.152.183.233  no
 postgresql-k8s           active      1  postgresql-k8s  14/stable      29  10.152.183.195  no       Primary
 
 Unit               Workload  Agent  Address      Ports  Message
-demo-api-charm/0*  active    idle   10.1.157.90         
+demo-api-charm/0*  active    idle   10.1.157.90
 postgresql-k8s/0*  active    idle   10.1.157.92         Primary
 
 Relation provider              Requirer                       Interface          Type     Message
-postgresql-k8s:database        demo-api-charm:database        postgresql_client  regular  
-postgresql-k8s:database-peers  postgresql-k8s:database-peers  postgresql_peers   peer     
-postgresql-k8s:restart         postgresql-k8s:restart         rolling_op         peer 
+postgresql-k8s:database        demo-api-charm:database        postgresql_client  regular
+postgresql-k8s:database-peers  postgresql-k8s:database-peers  postgresql_peers   peer
+postgresql-k8s:restart         postgresql-k8s:restart         rolling_op         peer
 ```
 
-The relation appears to be up and running, but we should also test that it's working as intended. First, let's try to write something to the database by posting some name to the database via API using `curl` as below -- where `10.1.157.90` is a pod IP and `8000` is our app port. You can repeat the command for multiple names. 
+The relation appears to be up and running, but we should also test that it's working as intended. First, let's try to write something to the database by posting some name to the database via API using `curl` as below -- where `10.1.157.90` is a pod IP and `8000` is our app port. You can repeat the command for multiple names.
 
 ```text
 curl -X 'POST' \
@@ -519,7 +516,7 @@ When it's done, the output should show two passing tests:
 ...
   demo-api-charm/0 [idle] waiting: Waiting for database relation
 INFO     juju.model:model.py:2759 Waiting for model:
-  demo-api-charm/0 [idle] active: 
+  demo-api-charm/0 [idle] active:
 PASSED
 -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- live log teardown --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 INFO     pytest_operator.plugin:plugin.py:783 Model status:
@@ -528,12 +525,12 @@ Model            Controller       Cloud/Region        Version  SLA          Time
 test-charm-2ara  main-controller  microk8s/localhost  3.1.5    unsupported  09:45:56+02:00
 
 App             Version  Status  Scale  Charm           Channel    Rev  Address        Exposed  Message
-demo-api-charm  1.0.1    active      1  demo-api-charm               0  10.152.183.99  no       
-postgresql-k8s  14.7     active      1  postgresql-k8s  14/stable   73  10.152.183.50  no       
+demo-api-charm  1.0.1    active      1  demo-api-charm               0  10.152.183.99  no
+postgresql-k8s  14.7     active      1  postgresql-k8s  14/stable   73  10.152.183.50  no
 
 Unit               Workload  Agent  Address       Ports  Message
-demo-api-charm/0*  active    idle   10.1.208.77          
-postgresql-k8s/0*  active    idle   10.1.208.107         
+demo-api-charm/0*  active    idle   10.1.208.77
+postgresql-k8s/0*  active    idle   10.1.208.107
 
 INFO     pytest_operator.plugin:plugin.py:789 Juju error logs:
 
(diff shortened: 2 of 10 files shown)


docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -41,6 +41,31 @@ config:
...
+    server_port: int = 8000
+    """Default port on which FastAPI is available."""
+
+    def __post_init__(self):
+        """Validate the configuration."""
+        if self.server_port == 22:
+            raise ValueError('Invalid port number, 22 is reserved for SSH')
+```
+
+You will also need to add `import dataclasses` in the imports at the top of the file.
```
dwilding: I'd suggest switching this sentence round, to make it more obvious that it's a step to be followed.

```suggestion
Then, still in `src/charm.py`, add `import dataclasses` in the imports at the top of the file.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -41,6 +41,31 @@ config:
...
+## Define a configuration class
+
+Open your `src/charm.py` file, and add a configuration class that matches the configuration you added in `charmcraft.yaml`:
+
+```python
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FastAPIConfig:
+    """Configuration for the FastAPI demo charm.
+
+    Note that this configuration is also defined in charmcraft.yaml
```
dwilding: Very minor: I feel we can drop "Note that"

```suggestion
    This configuration is also defined in charmcraft.yaml
```

examples/k8s-2-configurable/src/charm.py
```diff
@@ -26,6 +27,22 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FastAPIConfig:
+    """Configuration for the FastAPI demo charm.
+
+    Note that this configuration is also defined in charmcraft.yaml
```
dwilding:
```suggestion
    This configuration is also defined in charmcraft.yaml
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -41,6 +41,31 @@ config:
...
+
+    def __post_init__(self):
+        """Validate the configuration."""
+        if self.server_port == 22:
+            raise ValueError('Invalid port number, 22 is reserved for SSH')
+```
+
+You will also need to add `import dataclasses` in the imports at the top of the file.
+
+Now that we have defined the configuration with a Python class, IDEs will use it to provide hints when we are accessing the configuration, and static type checkers are able to validate that we are using the config options correctly.
```
dwilding: There's only one option in this case

```suggestion
Now that we have defined the configuration with a Python class, IDEs will use it to provide hints when we are accessing the configuration, and static type checkers are able to validate that we are using the config option correctly.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -51,23 +76,16 @@ In the `__init__` function, add an observer for the `config_changed` event and p
 framework.observe(self.on.config_changed, self._on_config_changed)
 ```
 
-Now, define the handler, as below. First, read the `self.config` attribute to get the new value of the setting. Then, validate that this value is allowed (or block the charm otherwise). Next, let's log the value to the logger. Finally, since configuring something like a port affects the way we call our workload application, we also need to update our Pebble configuration, which we will do via a newly created method `_update_layer_and_restart` that we will define shortly.
+Now, define the handler, as below. Since configuring something like a port affects the way we call our workload application, we need to update our Pebble configuration, which we will do via a newly created method `_update_layer_and_restart` that we will define shortly.
```
dwilding: I'd probably simplify this part:

```suggestion
Now, define the handler, as below. Since configuring something like a port affects the way we call our workload application, we need to update our Pebble configuration.
```

Then before the Caution box, add a sentence like:

```
We'll define `_update_layer_and_restart` shortly.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -104,26 +122,48 @@ def _update_layer_and_restart(self) -> None:
...
+    except ValueError as e:
+        logger.error('Configuration error: %s', e)
+        self.unit.status = ops.BlockedStatus(str(e))
     except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
         logger.info('Unable to connect to Pebble: %s', e)
         self.unit.status = ops.MaintenanceStatus('Waiting for Pebble in workload container')
 ```
 
-Now, crucially, update the `_pebble_layer` property to make the layer definition dynamic, as shown below. This will replace the static port `8000` with `f"--port={self.config['server-port']}"`.
+When the config is loaded as part of creating the Pebble layer, if it is invalid (in our case, if the port is set to 22), then a `ValueError` will be raised. The `_update_layer_and_restart` method handles that by logging the issue and setting the status of the unit to blocked, letting the Juju user know that they need to take action.
```
dwilding: Very minor suggestions:

```suggestion
When the config is loaded as part of creating the Pebble layer, if the config is invalid (in our case, if the port is set to 22), then a `ValueError` will be raised. The `_update_layer_and_restart` method handles that by logging the error and setting the status of the unit to blocked, letting the Juju user know that they need to take action.
```

examples/k8s-3-postgresql/src/charm.py
```diff
@@ -31,6 +32,22 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FastAPIConfig:
+    """Configuration for the FastAPI demo charm.
+
+    Note that this configuration is also defined in charmcraft.yaml
```
dwilding:
```suggestion
    This configuration is also defined in charmcraft.yaml
```

examples/k8s-3-postgresql/src/charm.py
```diff
@@ -84,7 +94,7 @@ def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
         # If nothing is wrong, then the status is active.
         event.add_status(ops.ActiveStatus())
 
-    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+    def _on_database_created(self, _: DatabaseCreatedEvent) -> None:
```
dwilding, line 95: This change also needs making in `integrate-your-charm-with-postgresql.md` (code sample) and `k8s-4-action/src/charm.py`

examples/k8s-4-action/src/charm.py
```diff
@@ -31,6 +32,31 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FastAPIConfig:
+    """Configuration for the FastAPI demo charm.
+
+    Note that this configuration is also defined in charmcraft.yaml
```
dwilding:
```suggestion
    This configuration is also defined in charmcraft.yaml
```

examples/k8s-5-observe/src/charm.py
```diff
@@ -34,6 +35,31 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FastAPIConfig:
+    """Configuration for the FastAPI demo charm.
+
+    Note that this configuration is also defined in charmcraft.yaml
```
dwilding:
```suggestion
    That this configuration is also defined in charmcraft.yaml
```

examples/k8s-4-action/src/charm.py
```diff
@@ -31,6 +32,31 @@
...
+    def __post_init__(self):
+        """Validate the configuration."""
+        if self.server_port == 22:
+            raise ValueError('Invalid port number, 22 is reserved for SSH')
+
+
+# Note that this action is also defined in charmcraft.yaml
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class GetDbInfoAction:
+    """Fetches database authentication information."""
```
dwilding, line 52:
```suggestion
@dataclasses.dataclass(frozen=True, kw_only=True)
class GetDbInfoAction:
    """Fetches database authentication information.
    
    This action is also defined in charmcraft.yaml
    """
```

examples/k8s-5-observe/src/charm.py
```diff
@@ -34,6 +35,31 @@
...
+    def __post_init__(self):
+        """Validate the configuration."""
+        if self.server_port == 22:
+            raise ValueError('Invalid port number, 22 is reserved for SSH')
+
+
+# Note that this action is also defined in charmcraft.yaml
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class GetDbInfoAction:
+    """Fetches database authentication information."""
```
dwilding, line 55:
```suggestion
@dataclasses.dataclass(frozen=True, kw_only=True)
class GetDbInfoAction:
    """Fetches database authentication information.
    
    This action is also defined in charmcraft.yaml
    """
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -41,6 +41,22 @@ actions:
...
 
+## Define an action class
+
+Open your `src/charm.py` file, and add an action class that matches the definition you used in `charmcraft.yaml`:
+
+```python
+# Note that this action is also defined in charmcraft.yaml
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class GetDbInfoAction:
+    """Fetches database authentication information."""
```
dwilding: Add this to the docstring instead, as for the config option?

```suggestion
@dataclasses.dataclass(frozen=True, kw_only=True)
class GetDbInfoAction:
    """Fetches database authentication information.
    
    This action is also defined in charmcraft.yaml
    """
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -41,6 +41,22 @@ actions:
...
+
+    show_password: bool
+    """Show username and password in output information."""
+```
+
+Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameters, and static type checkers are able to validate that we are using the parameters correctly.
```
dwilding: I guess you might have been highlighting the more general point about validating action parameters? I think this is probably clearer though:

```suggestion
Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameter, and static type checkers are able to validate that we are using the parameter correctly.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -69,7 +85,7 @@ def _on_get_db_info_action(self, event: ops.ActionEvent) -> None:
 
     Learn more about actions at https://ops.readthedocs.io/en/latest/howto/manage-actions.html
     """
-    show_password = event.params['show-password']  # see charmcraft.yaml
+    params = event.load_params(GetDbInfoAction, errors='fail')
```
dwilding, line 87: This code sample seems to be missing a change: `show_password` below should be `params.show_password`

dwilding (approved):
Awesome, thank you! I spotted a couple of places that need to be updated for consistency and added a few very minor suggestions.

I'm wondering whether it would be better to have `_pebble_layer` take the port as an argument, moving `load_config` into `_update_layer_and_restart`. Which would mean changing `_pebble_layer` into a method, as in the [httpbin charm](https://github.com/canonical/operator/blob/main/examples/httpbin-demo/src/charm.py#L130-L144).

Is there a strong reason to prefer one approach over the other?

Personally, I find the httpbin way easier to follow. The logic is more explicit and I don't need to consider the possibility of a property raising an error.

tonyandrewmeyer:
> I'm wondering whether it would be better to have `_pebble_layer` take the port as an argument, moving `load_config` into `_update_layer_and_restart`. Which would mean changing `_pebble_layer` into a method, as in the [httpbin charm](https://github.com/canonical/operator/blob/main/examples/httpbin-demo/src/charm.py#L130-L144).

Yeah, I wondered about that too. I ended up leaving it as-is to be a smaller change, but as you're also suggesting that, I'm happy to go that way.

> Personally, I find the httpbin way easier to follow. The logic is more explicit and I don't need to consider the possibility of a property raising an error.

Yeah, I find that properties are often overused, and I feel that they should almost never raise exceptions, because that's generally unexpected by users when accessing an attribute (other than AttributeError, I guess) rather than clearly calling something.

I think for really simple cases, the Pebble layer as a property might make sense, but it's pretty common to be pulling things from lots of places (relations, storage, config, ...) so a method makes more sense when you get to that point, and having the tutorial start people heading to where we think they'll end up is good.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
tonyandrewmeyer: What do you think about instead moving the config one to a comment?

Once I get the feature landed, I'd like to have the tutorial also use the functionality to generate the config & action YAML from the Python classes. If that happens, then the docstring for the action class ends up being the description for the action YAML (this isn't relevant for config) and the note isn't relevant in that case.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
tonyandrewmeyer: Ah, yes, I noticed that too, and forgot to push that fix. Thanks!

examples/k8s-5-observe/src/charm.py
```diff
@@ -34,6 +35,31 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FastAPIConfig:
+    """Configuration for the FastAPI demo charm.
+
+    That this configuration is also defined in charmcraft.yaml
```
tonyandrewmeyer:
```suggestion
    This configuration is also defined in charmcraft.yaml
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -104,26 +124,48 @@ def _update_layer_and_restart(self) -> None:
...
-command = ' '.join([
-    'uvicorn',
-    'api_demo_server.app:app',
-    '--host=0.0.0.0',
-    f'--port={self.config["server-port"]}',
-])
+@property
+def _pebble_layer(self) -> ops.pebble.Layer:
+    config = self.load_config(FastAPIConfig)
+    """A Pebble layer for the FastAPI demo services."""
```
james-garner-canonical:
```suggestion
    """A Pebble layer for the FastAPI demo services."""
    config = self.load_config(FastAPIConfig)
```

examples/k8s-5-observe/src/charm.py
dwilding: Oops, yes

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
dwilding: Moving the config note to a comment sounds good :+1:

examples/k8s-2-configurable/src/charm.py
```diff
@@ -74,18 +84,22 @@ def _update_layer_and_restart(self) -> None:
...
+            logger.error('Configuration error: %s', e)
+            self.unit.status = ops.BlockedStatus(str(e))
         except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
             logger.info('Unable to connect to Pebble: %s', e)
             self.unit.status = ops.MaintenanceStatus('Waiting for Pebble in workload container')
 
     @property
     def _pebble_layer(self) -> ops.pebble.Layer:
+        config = self.load_config(FastAPIConfig)
         """A Pebble layer for the FastAPI demo services."""
```
james-garner-canonical:
```suggestion
        """A Pebble layer for the FastAPI demo services."""
        config = self.load_config(FastAPIConfig)
```

examples/k8s-2-configurable/src/charm.py
```diff
@@ -74,18 +84,22 @@ def _update_layer_and_restart(self) -> None:
             logger.info(f"Replanned with '{self.pebble_service_name}' service")
 
             self.unit.status = ops.ActiveStatus()
+        except ValueError as e:
```
james-garner-canonical: I worry that the source of these exceptions is a little hard to follow. It's not immediately intuitive that the bit that causes a `ValueError` here is the `self._pebble_layer` property access. Do you think something like this might be a better recommendation for people following this tutorial?
```py
try:
    layer = self._pebble_layer  # or make this a regular method instead of a property
except ValueError:
    ...
    return
try:
    self.container...
except (ops.pebble...
```

That said, I haven't quite got my head around how this control flow interacts with `collect_status`, so I need to think about this a bit more.

tonyandrewmeyer: Yes, David also suggested making this a regular method, and I wondered about it too while writing it, so I've done that.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tonyandrewmeyer: Ha, I totally didn't notice this, bad copy & pasting. But I fixed it when changing to a property.

tonyandrewmeyer:
@dwilding I've changed the Pebble layer property to a method, feel free to look again if you like, or just trust James to do that :)

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -41,6 +41,22 @@ actions:
         default: False
 ```
 
+## Define an action class
+
+Open your `src/charm.py` file, and add an action class that matches the definition you used in `charmcraft.yaml`:
+
+```python
+# Note that this action is also defined in charmcraft.yaml
```
james-garner-canonical: I wonder if it's nicer to drop this line in the tutorial?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -41,6 +41,22 @@ actions:
...
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class GetDbInfoAction:
+    """Fetches database authentication information."""
+
+    show_password: bool
+    """Show username and password in output information."""
+```
+
+Now that we have defined the action with a Python class, IDEs will use it to provide hints when we are accessing the action parameter, and static type checkers are able to validate that we are using the parameter correctly.
```
james-garner-canonical:
```suggestion
We'll use `ActionEvent.load_params` (linking to reference doc) to create an instance of your config class from the Juju action event. This allows IDEs to provide hints when we are accessing the action parameter, and static type checkers are able to validate that we are using the parameter correctly.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -196,17 +236,16 @@ def _pebble_layer(self) -> ops.pebble.Layer:
...
     }
     return ops.pebble.Layer(pebble_layer)
 ```
 
-Now, let's define this property such that, every time it is called, it dynamically fetches database authentication data and also prepares the output in a form that our application can consume, as below:
+Now, let's define this method such that, every time it is called, it dynamically fetches database authentication data and also prepares the output in a form that our application can consume, as below:
 
 ```python
-@property
 def app_environment(self) -> dict[str, str]:
```
james-garner-canonical:
```suggestion
def get_app_environment(self) -> dict[str, str]:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -281,23 +321,18 @@ def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
     event.add_status(ops.ActiveStatus())
 ```
 
-We also want to clean up the code to remove the places where we're setting the status outside of this method, other than anywhere we're wanting a status to show up *during* the event execution (such as `MaintenanceStatus`). In `_on_config_changed`, change the port 22 check to:
+We also want to clean up the code to remove the places where we're setting the status outside of this method, other than anywhere we're wanting a status to show up *during* the event execution (such as `MaintenanceStatus`). If you missed doing so above, in `_update_layer_and_restart`, remove the lines:
```
james-garner-canonical, line 324:
```suggestion
We also want to clean up the code to remove the places where we're setting the status outside of this method, other than anywhere we're wanting a status to show up *during* the event execution (such as `MaintenanceStatus`). If you missed doing so above, in `_update_layer_and_restart`, remove the line:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -41,6 +41,29 @@ config:
       type: int
 ```
 
+## Define a configuration class
+
+Open your `src/charm.py` file, and add a configuration class that matches the configuration you added in `charmcraft.yaml`:
+
+```python
+# This configuration is also defined in charmcraft.yaml
```
james-garner-canonical: I wonder if it's nicer to drop this line in the tutorial?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -41,6 +41,29 @@ config:
...
+
+    def __post_init__(self):
+        """Validate the configuration."""
+        if self.server_port == 22:
+            raise ValueError('Invalid port number, 22 is reserved for SSH')
+```
+
+Then, still in `src/charm.py`, add `import dataclasses` in the imports at the top of the file.
+
+Now that we have defined the configuration with a Python class, IDEs will use it to provide hints when we are accessing the configuration, and static type checkers are able to validate that we are using the config option correctly.
```
james-garner-canonical:
```suggestion
We'll use `load_config` (linking to reference doc) to create an instance of your config class from the Juju config data. This allows IDEs to provide hints when we are accessing the configuration, and static type checkers are able to validate that we are using the config options correctly.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -93,37 +111,60 @@ def _update_layer_and_restart(self) -> None:
...
-        self.unit.status = ops.ActiveStatus()
-    except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
-        logger.info('Unable to connect to Pebble: %s', e)
-        self.unit.status = ops.MaintenanceStatus('Waiting for Pebble in workload container')
+        self.unit.status = ops.MaintenanceStatus('Assembling Pebble layers')
+        try:
+            config = self.load_config(FastAPIConfig)
+        except ValueError as e:
+            logger.error('Configuration error: %s', e)
+            self.unit.status = ops.BlockedStatus(str(e))
```
james-garner-canonical, line 118: Should we set the status here, or leave this for `_on_collect_status`? Ignore this comment if we add that in a later chapter.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -93,37 +111,60 @@ def _update_layer_and_restart(self) -> None:
...
+            self.unit.status = ops.ActiveStatus()
+        except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
+            logger.info('Unable to connect to Pebble: %s', e)
+            self.unit.status = ops.MaintenanceStatus('Waiting for Pebble in workload container')
 ```
 
-Now, crucially, update the `_pebble_layer` property to make the layer definition dynamic, as shown below. This will replace the static port `8000` with `f"--port={self.config['server-port']}"`.
+When the config is loaded as part of creating the Pebble layer, if the config is invalid (in our case, if the port is set to 22), then a `ValueError` will be raised. The `_update_layer_and_restart` method handles that by logging the error and setting the status of the unit to blocked, letting the Juju user know that they need to take action.
+
+Now, crucially, update the `_get_pebble_layer` method to make the layer definition dynamic, as shown below. This will replace the static port `8000` with `f"--port={port}"`, using the port passed to the method.
```
james-garner-canonical:
```suggestion
Now, crucially, update the `_get_pebble_layer` method to make the layer definition dynamic, as shown below. This will replace the static port `8000` with the port passed to the method.
```

examples/k8s-4-action/src/charm.py
```diff
@@ -102,7 +119,7 @@ def _on_get_db_info_action(self, event: ops.ActionEvent) -> None:
 
         Learn more about actions at https://ops.readthedocs.io/en/latest/howto/manage-actions.html
         """
-        show_password = event.params['show-password']  # see charmcraft.yaml
+        params = event.load_params(GetDbInfoAction, errors='fail')
```
james-garner-canonical, line 122: What happens on fail here? I guess the current run of the charm code by the framework is aborted, but then the framework still runs the charm again for collect status?

examples/k8s-4-action/src/charm.py
```diff
@@ -133,26 +150,35 @@ def _update_layer_and_restart(self) -> None:
...
             logger.info("Added updated layer 'fastapi_demo' to Pebble plan")
 
             # Tell Pebble to incorporate the changes, including restarting the
             # service if required.
             self.container.replan()
             logger.info(f"Replanned with '{self.pebble_service_name}' service")
-
-            self.unit.status = ops.ActiveStatus()
+        except ValueError as e:
+            logger.error('Configuration error: %s', e)
```
james-garner-canonical: Where does this value error come from?

dwilding:
> @dwilding I've changed the Pebble layer property to a method, feel free to look again if you like, or just trust James to do that :)

Great! I just had a look through. In "Write unit tests" in "Make your charm configurable", we have this:

```
Since we added a new feature to configure `server-port` and use it in the `_pebble_layer` dynamically, we should write tests for the feature.
```

Now that the name of the method has changed, how about changing the sentence to this:

```
[...] use it in the Pebble layer dynamically [...]
```

examples/k8s-4-action/src/charm.py
tonyandrewmeyer: A failure message is set for the action and the hook exits. `collect-status` does not run, but it doesn't make any sense for the status to change as a result of the action failing.

examples/k8s-4-action/src/charm.py
tonyandrewmeyer: Missed removing this, thanks.

examples/k8s-4-action/src/charm.py
james-garner-canonical: Right, that makes sense, thanks for spelling it out for me!

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tonyandrewmeyer: We currently only switch to collect-status in the DB chapter (after this). That also explains removing all the existing status setting (apart from maintenance).

I think probably the tutorial would be simpler if we just used collect-status all along, but that should really be a separate PR.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tonyandrewmeyer: There are three lines, surely it should be "lines", not "line"?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
james-garner-canonical: Thanks for confirming that this is correct as is. I'd be in favour of using collect-status throughout, and I agree with  you that that's work for a separate PR.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
james-garner-canonical: You're right, my mistake (got confused by the subsequent changes somehow).

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tonyandrewmeyer: David had me change "options" to "option" previously, because in this specific case there is only one. I think he's right and we are wrong.

james-garner-canonical: I'm happy with `option` instead of `options`.

examples/k8s-5-observe/src/charm.py
```diff
@@ -149,26 +171,35 @@ def _update_layer_and_restart(self) -> None:
...
             logger.info("Added updated layer 'fastapi_demo' to Pebble plan")
 
             # Tell Pebble to incorporate the changes, including restarting the
             # service if required.
             self.container.replan()
             logger.info(f"Replanned with '{self.pebble_service_name}' service")
-
-            self.unit.status = ops.ActiveStatus()
+        except ValueError as e:
+            logger.error('Configuration error: %s', e)
```
james-garner-canonical: I should have pointed this out in the first pass, but I guess this needs go to too?
```suggestion
```

james-garner-canonical (approved):

PR #1893, opened 2025-07-04
chore: define defaults in only one place in _JujuContext

james-garner-canonical:
As #1799 points out, the `_JujuContext` class sets default values for its arguments in two places: in its `dataclass` init, and in the `from_dict` method. Setting defaults in two places is a bit of an anti-pattern, and in this case where `_JujuContext` is always constructed with `from_dict`, one set of defaults is redundant.

This PR drops the defaults defined in `_JujuContext.from_dict` -- if a default would have been defined, instead the argument is not passed, falling back to the default values/factories in the `dataclass` init method. In all cases, the default values were the same in both places.

A simpler implementation option would have been to drop the default values from the `dataclass` init, since `from_dict` always passed all arguments explicitly, but as pointed out in [this comment](https://github.com/canonical/operator/issues/1799#issuecomment-3034458405), we'd likely want to add these back if we made `_JujuContext` public (as we are considering doing). The approach taken in this PR will require no such changes, as the `dataclass` init defaults are now the single source of truth.

---

One problematic aspect of the approach taken here is that we lose the static guarantee that we're passing validly named keyword arguments to `_JujuContext`. Additionally, because keys are added to `kwargs` conditionally, we need to rely on exhaustive branch coverage to catch badly named keys in tests.

We could solve this statically by defining a `TypedDict`. This is a bit verbose, and begins to feel like a second source of truth, though the type checker will warn us if the `TypedDict` names and types diverge from those permitted by `_JujuContext`, so it's not really. There would have been no guaranteed warning if the default values diverged, so arguably this is an improvement. A separate PR (#1894) tries this approach.

---

Ultimately I wonder if both of these approaches introduce too much to gain too little, in which case we should opt to close these PRs and #1799, deferring any further work until (if) we decide to make `_JujuContext` public.

---

Minor: pyright complains in my editor about `default_factory=set`, because it can't infer the type of its contents -- since we're Python 3.10 now, we can just write `set[str]` at runtime.

### DIFF
--- a/ops/jujucontext.py
+++ b/ops/jujucontext.py
@@ -61,7 +61,7 @@ class _JujuContext:
     If true, write logs to stderr as well as to juju-log (from JUJU_DEBUG).
     """
 
-    debug_at: set[str] = dataclasses.field(default_factory=set)  # pyright: ignore[reportUnknownVariableType]
+    debug_at: set[str] = dataclasses.field(default_factory=set[str])
     """Where you want to stop when debugging.
 
     For example 'all' (from JUJU_DEBUG_AT).


ops/jujucontext.py
```diff
@@ -185,51 +185,45 @@ class _JujuContext:
...
-            unit_name=env.get('JUJU_UNIT_NAME', ''),
-            # The meter-status-changed event, triggered by `juju set-meter-status`,
-            # does not set JUJU_VERSION, but all other events do. When we drop support
-            # for Juju 2 and Juju 3 we can change this to always expect JUJU_VERSION,
-            # as that event no longer exists in Juju 4.
-            version=JujuVersion(env.get('JUJU_VERSION', '0.0.0')),
-            workload_name=env.get('JUJU_WORKLOAD_NAME') or None,
-        )
+        kwargs = {
+            var.removeprefix('JUJU_').lower(): value
```
dimaqq, line 189: **TIL**
> Added in version 3.9.

Finally we can do these great things!

ops/jujucontext.py
```diff
@@ -185,51 +185,45 @@ class _JujuContext:
...
+                'JUJU_PEBBLE_CHECK_NAME',
+                'JUJU_SECRET_ID',
+                'JUJU_SECRET_LABEL',
+                'JUJU_UNIT_NAME',
+                'JUJU_WORKLOAD_NAME',
+            )
+            if (value := env.get(var))
+        }
+        if juju_charm_dir := env.get('JUJU_CHARM_DIR'):
+            kwargs['charm_dir'] = Path(juju_charm_dir).resolve()
```
dimaqq, line 208: Note that there was a fall-back.
I think it was for local testing, perhaps?

ops/jujucontext.py
```diff
@@ -61,7 +61,7 @@ class _JujuContext:
     If true, write logs to stderr as well as to juju-log (from JUJU_DEBUG).
     """
 
-    debug_at: set[str] = dataclasses.field(default_factory=set)
+    debug_at: set[str] = dataclasses.field(default_factory=set[str])
```
dimaqq, line 64: ❤️

dimaqq (commented):
Btw., why can't we do:

```diff
-    version: JujuVersion = dataclasses.field(default_factory=lambda: JujuVersion('0.0.0'))
+    version: JujuVersion = JujuVersion('0.0.0')
```

ops/jujucontext.py
dimaqq: Ah, I see that there's a default factory. nice.
Actually, `.resolve()` doesn't make syscalls (apart from `getcwd()`), does it?
So why not have the default without a factory?

ops/jujucontext.py
```diff
@@ -185,51 +185,45 @@ class _JujuContext:
...
+                'JUJU_NOTICE_ID',
+                'JUJU_NOTICE_KEY',
+                'JUJU_NOTICE_TYPE',
+                'JUJU_PEBBLE_CHECK_NAME',
+                'JUJU_SECRET_ID',
+                'JUJU_SECRET_LABEL',
+                'JUJU_UNIT_NAME',
+                'JUJU_WORKLOAD_NAME',
+            )
+            if (value := env.get(var))
```
tonyandrewmeyer, line 205: Personally, I think the walrus operator generally makes code less readable and understandable. We don't use it at all in ops (outside testing and tracing) at the moment, so this is a new introduction.

Is there any advantage of doing it like this rather than the if being `if var in env` and the dict value being `env[var]`? I feel that's more understandable code.

ops/jujucontext.py
```diff
@@ -185,51 +185,45 @@ class _JujuContext:
...
+        }
+        if juju_charm_dir := env.get('JUJU_CHARM_DIR'):
+            kwargs['charm_dir'] = Path(juju_charm_dir).resolve()
+        if 'JUJU_DEBUG' in env:
+            kwargs['debug'] = True
+        if juju_debug_at := env.get('JUJU_DEBUG_AT'):
+            kwargs['debug_at'] = {x.strip() for x in juju_debug_at.split(',')}
+        if juju_departing_unit := env.get('JUJU_DEPARTING_UNIT'):
+            kwargs['relation_departing_unit_name'] = juju_departing_unit
+        if juju_relation := env.get('JUJU_RELATION'):
```
tonyandrewmeyer, line 215: I'm not much of a fan of all the walrus in these either, although they are at least a much more straightforward case. I would personally have a map of environment name to context attribute and just loop through that, but I can live with these, I suppose.

ops/jujucontext.py
james-garner-canonical: The reason for not doing `if var in env` then `env[var]` is that is changes the semantics of the preexisting code, where any falsey value would result in falling back to the default.

I understand that this dict is supposed from the environment, so all the values are strings, and so there are no falsey values, but I didn't dig to see if it was safe to make this assumption. It looks like we always use `os.environ` in `ops` itself, so it's just a matter of making sure that `harness` and `scenario` never include falsey values in their env dicts.

If we do want to take one of these PRs further, I'd be happy to make that change, which would drive the walrus to extinction.

ops/jujucontext.py
james-garner-canonical: That would be doable and preserve type checking with a `Literal`, so definitely a possibility if we take either PR further.

james-garner-canonical:
With the spec on exposing lower level internals like this now in review, I'm going to close this PR in favour of the outcome of that spec.

PR #1894, opened 2025-07-04
chore: define defaults in only one place in _JujuContext (+ TypedDict)

james-garner-canonical:
This PR builds on #1893, addressing #1799 and additionally solving the issue of missing static validation with the addition of a `TypedDict`. This approach is significantly more verbose than #1893, but safer.

### DIFF
--- a/ops/jujucontext.py
+++ b/ops/jujucontext.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 import dataclasses
 from pathlib import Path
-from typing import Any, Mapping
+from typing import Any, Mapping, TypedDict
 
 from .jujuversion import JujuVersion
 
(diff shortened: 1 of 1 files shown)


dimaqq:
It's elegant, but does duplicate code.
I wonder if it's actually worth it.

ops/jujucontext.py
```diff
@@ -185,51 +185,82 @@ class _JujuContext:
...
-            unit_name=env.get('JUJU_UNIT_NAME', ''),
-            # The meter-status-changed event, triggered by `juju set-meter-status`,
-            # does not set JUJU_VERSION, but all other events do. When we drop support
-            # for Juju 2 and Juju 3 we can change this to always expect JUJU_VERSION,
-            # as that event no longer exists in Juju 4.
-            version=JujuVersion(env.get('JUJU_VERSION', '0.0.0')),
-            workload_name=env.get('JUJU_WORKLOAD_NAME') or None,
-        )
+        kwargs: _JujuContextKwargs = {}
+        # simple keys that copy the environment variable without modification
```
tonyandrewmeyer:
```suggestion
        # Simple keys that copy the environment variable value without modification.
```

ops/jujucontext.py
```diff
@@ -185,51 +185,82 @@ class _JujuContext:
...
+            kwargs['pebble_check_name'] = juju_pebble_check_name
+        if juju_secret_id := env.get('JUJU_SECRET_ID'):
+            kwargs['secret_id'] = juju_secret_id
+        if juju_secret_label := env.get('JUJU_SECRET_LABEL'):
+            kwargs['secret_label'] = juju_secret_label
+        if juju_unit_name := env.get('JUJU_UNIT_NAME'):
+            kwargs['unit_name'] = juju_unit_name
+        if juju_workload_name := env.get('JUJU_WORKLOAD_NAME'):
+            kwargs['workload_name'] = juju_workload_name
+        # keys that do something a little fancier
```
tonyandrewmeyer:
```suggestion
        # Keys that do something a little fancier.
```

ops/jujucontext.py
```diff
@@ -185,51 +185,82 @@ class _JujuContext:
...
-            # The meter-status-changed event, triggered by `juju set-meter-status`,
-            # does not set JUJU_VERSION, but all other events do. When we drop support
-            # for Juju 2 and Juju 3 we can change this to always expect JUJU_VERSION,
-            # as that event no longer exists in Juju 4.
-            version=JujuVersion(env.get('JUJU_VERSION', '0.0.0')),
-            workload_name=env.get('JUJU_WORKLOAD_NAME') or None,
-        )
+        kwargs: _JujuContextKwargs = {}
+        # simple keys that copy the environment variable without modification
+        if juju_action_name := env.get('JUJU_ACTION_NAME'):
```
tonyandrewmeyer: Wouldn't this be simpler in a loop? Like:

```python
kwargs = {
    arg: env[key]
    for arg, key in simple_keys.items()
    if key in env
}

**Review by tonyandrewmeyer — COMMENTED** [review]
I like this more than the other PR, but I'm not convinced that it improves on the current code.

Also: what happens if we decide that the public API for `JujuContext` should be that you provide all values? Or if we want to have different defaults, or some values with no defaults in that case (e.g. that you must provide a version, since there is no sensible default)? It feels like we would end up undoing all this work in that case.

File: ops/jujucontext.py (hunk 3 above)
Comment (james-garner-canonical) [reply ops/jujucontext.py]: Yes, but I couldn't figure out a good way to get static analysis of the key names in that case, other than making a big `Literal`, which seemed unappealing -- but maybe that's the right way to do it ...

File: ops/jujucontext.py [hunk 4]
```diff
@@ -185,51 +186,77 @@ class _JujuContext:
...
+    secret_id: str
+    secret_label: str
+    secret_revision: int
+    storage_name: str
+    unit_name: str
+    version: JujuVersion
+    workload_name: str
+
+
+SimpleKeys: TypeAlias = typing.Literal[
```
dimaqq, line 248:
```suggestion
_SimpleKeys: TypeAlias = typing.Literal[
```

ops/jujucontext.py
```diff
@@ -185,51 +186,77 @@ class _JujuContext:
...
-            unit_name=env.get('JUJU_UNIT_NAME', ''),
-            # The meter-status-changed event, triggered by `juju set-meter-status`,
-            # does not set JUJU_VERSION, but all other events do. When we drop support
-            # for Juju 2 and Juju 3 we can change this to always expect JUJU_VERSION,
-            # as that event no longer exists in Juju 4.
-            version=JujuVersion(env.get('JUJU_VERSION', '0.0.0')),
-            workload_name=env.get('JUJU_WORKLOAD_NAME') or None,
-        )
+        kwargs: _JujuContextKwargs = {}
+        # Simple keys that copy the environment variable value without modification.
```
dimaqq, line 190:
```suggestion

```

ops/jujucontext.py
```diff
@@ -185,51 +186,77 @@ class _JujuContext:
...
+        simple_keys: tuple[SimpleKeys, ...] = typing.get_args(SimpleKeys)
+        for key in simple_keys:
+            if val := env.get(f'JUJU_{key.upper()}'):
+                kwargs[key] = val
+        # Keys that do something a little fancier.
```
dimaqq, line 195:
```suggestion

```

dimaqq (commented):
Code-wise this looks fine, though some cleanup is needed.

Whether we should do this at all... I'm not sure. I'd like to have specific warts fixed, but having two separate helpers where we must maintain the field lists in sync seems excessive.

As a straw-dog counterproposal, WDYT about:

```py
class _JujuContext:
    action_name: str | None = None
    ...

mutable = dataclasses.make_dataclass("MutableContext", fields(_JujuContext), bases=(), frozen=False)

context = mutable()
if val := env.get("JUJU_X"):
    context.x = int(val)

return _JujuContext(**dataclasses.asdict(context))
```

james-garner-canonical:
> Whether we should do this at all... I'm not sure. I'd like to have specific warts fixed, but having two separate helpers where we must maintain the field lists in sync seems excessive.

Yes, I'm not a big fan of the overall shape of this PR either. I'd be in favour of closing these PRs and the associated issue in favour of the changes associated with making this class public (see Tony's spec).

james-garner-canonical:
With the spec on exposing lower level internals like this now in review, I'm going to close this PR in favour of the outcome of that spec.

PR #1886, opened 2025-07-04
chore: remove impossible test

dimaqq:
Screenshot 2025-07-04 at 12 01 33

Job: https://github.com/canonical/operator/actions/runs/16064127575

My suggestion from #1878 did not outsmart GitHub.

PR #1895, opened 2025-07-04
ci: use httpbin demo charm for the Charmcraft pack test

dwilding:
This PR uses the [httpbin](https://github.com/canonical/operator/tree/main/examples/httpbin-demo) demo charm for the "Charmcraft Pack Test" check.

Since the charm doesn't use pyproject.toml (yet? see https://github.com/canonical/operator/issues/1805), I've kept the step that modifies requirements.txt. And since we're getting the `ops` package from a git repo, I've added a step that adds git as a build package in charmcraft.yaml. The hello-kubecon charm already had git specified as a build package.

### DIFF
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -39,7 +39,7 @@ jobs:
         run: sudo snap install yq
 
       - name: Add 'git' as a build package 
-      - run: |
+        run: |
           cd examples/httpbin-demo
           if yq 'has("parts")' charmcraft.yaml; then
             echo "'parts' already exists in charmcraft.yaml"
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -10,23 +10,16 @@ permissions: {}
 
 jobs:
   charmcraft-pack:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-22.04  # Not ubuntu-latest?
 
-    strategy:
-      matrix:
-        include: []
-#          - charm-repo: jnsgruk/hello-kubecon
-#            commit: eb3225860cc7db0e071d0ea6cd4a01b47e66b1e5  # 2025-03-27T12:05:21Z
     steps:
-      - name: Checkout test charm repository
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
         with:
-          repository: ${{ matrix.charm-repo }}
           persist-credentials: false
-          ref: ${{ matrix.commit }}
 
       - name: Update 'ops' dependency in test charm to latest
         run: |
+          cd examples/httpbin-demo
           sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
           if [ -z "$CLONE_SHA" ]
           then
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -41,7 +41,7 @@ jobs:
       - name: Add 'git' as a build package 
         run: |
           cd examples/httpbin-demo
-          if yq 'has("parts")' charmcraft.yaml; then
+          if yq -e 'has("parts")' charmcraft.yaml; then
             echo "'parts' already exists in charmcraft.yaml"
             exit 1
           fi
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -44,4 +44,6 @@ jobs:
         run: sudo snap install charmcraft --classic
 
       - name: Pack the charm
-        run: sudo charmcraft pack --verbose
+        run: |
+          cd examples/httpbin-demo
+          sudo charmcraft pack --verbose
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -41,11 +41,11 @@ jobs:
       - name: Add 'git' as a build package 
         run: |
           cd examples/httpbin-demo
-          if yq -e 'has("parts")' charmcraft.yaml; then
+          if yq --exit-status 'has("parts")' charmcraft.yaml; then
             echo "'parts' already exists in charmcraft.yaml"
             exit 1
           fi
-          yq -i '.parts.charm.build-packages = ["git"]' charmcraft.yaml
+          yq --inplace '.parts.charm.build-packages = ["git"]' charmcraft.yaml
           cat charmcraft.yaml
 
       - name: Set up LXD
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -35,6 +35,19 @@ jobs:
           CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
           CLONE_SHA: ${{ github.event.pull_request.head.sha }}
 
+      - name: Install yq
+        run: sudo snap install yq
+
+      - name: Add 'git' as a build package 
+      - run: |
+          cd examples/httpbin-demo
+          if yq 'has("parts")' charmcraft.yaml; then
+            echo "'parts' already exists in charmcraft.yaml"
+            exit 1
+          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/charmcraft-pack.yaml
```diff
@@ -42,6 +35,19 @@ jobs:
...
+        run: sudo snap install yq
+
+      - name: Add 'git' as a build package 
+        run: |
+          cd examples/httpbin-demo
+          if yq -e 'has("parts")' charmcraft.yaml; then
+            echo "'parts' already exists in charmcraft.yaml"
+            exit 1
+          fi
+          yq -i '.parts.charm.build-packages = ["git"]' charmcraft.yaml
```
dwilding: Tests passing! Thanks to James's doc, one I keep coming back to :pray: 
https://canonical-charmlibs.readthedocs-hosted.com/how-to/python-package/#git

.github/workflows/charmcraft-pack.yaml
```diff
@@ -10,23 +10,16 @@ permissions: {}
 
 jobs:
   charmcraft-pack:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-22.04  # Not ubuntu-latest?
```
dwilding: Should we change this?

dwilding:
@tonyandrewmeyer would you mind having a look at this? (No urgency)

It follows on from discussion that Dima, James, and I were having in #1886.

.github/workflows/charmcraft-pack.yaml
```diff
@@ -51,4 +57,6 @@ jobs:
         run: sudo snap install charmcraft --classic
 
       - name: Pack the charm
-        run: sudo charmcraft pack --verbose
+        run: |
+          cd examples/httpbin-demo
+          sudo charmcraft pack --verbose
```
james-garner-canonical: Does this actually need to run with `sudo`?

.github/workflows/charmcraft-pack.yaml
```diff
@@ -42,6 +35,19 @@ jobs:
           CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
           CLONE_SHA: ${{ github.event.pull_request.head.sha }}
 
+      - name: Install yq
+        run: sudo snap install yq
+
+      - name: Add 'git' as a build package 
+        run: |
+          cd examples/httpbin-demo
+          if yq -e 'has("parts")' charmcraft.yaml; then
```
james-garner-canonical: In the logs I see
```
false
Error: no matches found
```
Are these lines the output of this invocation of `yq`?

james-garner-canonical (approved):

.github/workflows/charmcraft-pack.yaml
dwilding: Yeah. Do you think that's going to be confusing in the future?

.github/workflows/charmcraft-pack.yaml
dwilding: Seems to work without sudo

.github/workflows/charmcraft-pack.yaml
dimaqq: Let's change that to `ubuntu-latest`.
It should be safe and generally good practice.

.github/workflows/charmcraft-pack.yaml
```diff
@@ -42,6 +35,19 @@ jobs:
           CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
           CLONE_SHA: ${{ github.event.pull_request.head.sha }}
 
+      - name: Install yq
+        run: sudo snap install yq
```
dimaqq, line 39: `yq` can also be apt-installed, but I guess either is fine.

.github/workflows/charmcraft-pack.yaml
dwilding: Done, thanks

.github/workflows/charmcraft-pack.yaml
james-garner-canonical: It certainly confused me a little, but I think it's acceptable. Interestingly, `false` is stdout and `Error: ...` is stderr.

Unrelated suggestion: How about using `--exit-status` instead of `-e`?

.github/workflows/charmcraft-pack.yaml
james-garner-canonical: Suggestion: `--inplace` instead of `-i`?

dwilding: Done, along with the suggestion about `--exit-status`. Good call, thanks

tonyandrewmeyer (approved):
I think switching to a local charm is definitely the right move, and httpbin-demo should be sufficient for now (I imagine that we'd want the example charms' integration tests to run at some point, and that might make this superfluous -- but maybe that will take too long, and we'd rather keep one specific one here).

We should also remove this file from `.github/workflows/update-charm-tests.yaml` since there's no revision to update any more.

dwilding:
> We should also remove this file from `.github/workflows/update-charm-tests.yaml` since there's no revision to update any more.

Done, thanks!

.github/workflows/charmcraft-pack.yaml
```diff
@@ -42,6 +35,19 @@ jobs:
...
+
+      - name: Add 'git' as a build package 
+        run: |
+          cd examples/httpbin-demo
+          if yq --exit-status 'has("parts")' charmcraft.yaml; then
+            echo "'parts' already exists in charmcraft.yaml"
+            exit 1
+          fi
+          yq --inplace '.parts.charm.build-packages = ["git"]' charmcraft.yaml
```
dimaqq, line 48: nit: probably not great to whack old build packages in case one is added later, but then we'd notice very quickly that this job fails.

dimaqq (approved):

.github/workflows/charmcraft-pack.yaml
dwilding: I'm only adding `parts` if it doesn't already exist ([this check](https://github.com/canonical/operator/pull/1895/files#diff-97d6c301c388a8acf64230f4a13a4be6a6743cea99a1f71d598d225ec75e0088R44)). So if we add a build package later, the job should fail immediately - if I've got my logic right!

dwilding:
I'm merging this now. We can revise further in a follow-up PR if needed. Thanks everyone!

.github/workflows/charmcraft-pack.yaml
```diff
@@ -42,6 +35,19 @@ jobs:
           CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
           CLONE_SHA: ${{ github.event.pull_request.head.sha }}
 
+      - name: Install yq
+        run: sudo snap install yq
+
+      - name: Add 'git' as a build package
```
dimaqq, line 41:
```suggestion
      - name: Add 'git' as a build package
```

I guess our yaml linter doesn't care about comment content... or maybe we ain't got no yaml linter.

dwilding: @tonyandrewmeyer I seem to remember that you caught a bunch of unnecessary whitespace with pre-commit recently. Would that have been useful here? Would you mind explaining how you do that?

tonyandrewmeyer: Run `uvx pre-commit install` in the repo(s) that you want to use pre-commit. If they've already got a `.pre-commit-config.yaml` file, then any time you commit the configured checks will be run (against the files being committed - anything with modifications that isn't in the commit is temporarily stashed, and anything without changes is ignored).

For trailing whitespace (and a few of the other checks) it'll auto-fix for you as well, so you can commit (get failure), verify the fix (if needed), git add, and then commit again.

I don't actually know how this works if you commit outside a terminal (using an IDE for example) but I assume it shows you the output if it fails or something?

james-garner-canonical: You can also invoke `pre-commit` manually at any time to run the checks and fixes just once (e.g. `uvx pre-commit`).

dwilding: Ah great, thank you! I confirmed that pre-commit will clean up this training space next time.

PR #1896, opened 2025-07-04
docs: add gosherve-demo example, from the hello-kubecon talk

### LINKED ISSUE #1876: Probably drop hello-kubecon test

tonyandrewmeyer:
This PR adds a new example charm, gosherve-demo-k8s, which is the charm from Jon Seager's Operator Day hello-kubecon demonstration, brought up to date with current charming practice.

I feel that there's value in retaining this charm as an example:
 * It uses slightly more functionality than the httpbin example (relation, action, storage).
 * It comes prepared to be used in a presentation, including talk structure and slide content, so can easily be used if someone wants a quick-to-prepare talk on Juju and writing charms (for a local meetup, for example).
 * Most of the work was already done for us.

The code started as an exact copy of the jnsgruk/hello-kubecon repository (although I think I messed up that commit and committed a link instead). I added in the original slides, and a markdown version of them. I then modernised the charm as outlined in the CHANGES.md document.

TODO:
* [ ] Bump to ingress v2
* [ ] Asserts in the integration tests

Fixes #1876


examples/gosherve-demo/src/charm.py
```diff
@@ -0,0 +1,125 @@
...
+            port=8080,
+            host=f'{self.app.name}-endpoints.{self.model.name}.svc.cluster.local',
+            strip_prefix=True,
+        )
+
+    def _on_install(self, _: ops.InstallEvent):
+        # Download the site:
+        self._fetch_site()
+
+    def _on_config_changed(self, _: ops.ConfigChangedEvent | ops.PebbleReadyEvent):
```
dimaqq, line 65: Would it be better to call this:

```suggestion
    def _reconcile(self, _: ops.ConfigChangedEvent | ops.PebbleReadyEvent):
```

examples/gosherve-demo/slides/gosherve-demo-slides.md
```diff
@@ -0,0 +1,908 @@
+# Charmed Operator Development Workshop
```
dimaqq, line 1: I wonder if we want to have the copy of the presentation in the monorepo.
Maybe we could reference the slides instead?

examples/gosherve-demo/src/charm.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        framework.observe(self.on.install, self._on_install)
+        framework.observe(self.on.config_changed, self._on_config_changed)
+        framework.observe(self.on['gosherve'].pebble_ready, self._on_config_changed)
+        framework.observe(self.on['pull_site'].action, self._pull_site_action)
+        self.ingress = IngressPerAppRequirer(
+            self,
+            port=8080,
+            host=f'{self.app.name}-endpoints.{self.model.name}.svc.cluster.local',
```
dimaqq, line 57: I would personally strip one element from own FQDN than build this from parts: https://github.com/dimaqq/hexanator/blob/main/src/charm.py#L16

examples/gosherve-demo/src/charm.py
```diff
@@ -0,0 +1,125 @@
...
+                    'override': 'replace',
+                    'summary': 'gosherve',
+                    'command': '/gosherve',
+                    'startup': 'enabled',
+                    'environment': {
+                        'REDIRECT_MAP_URL': config.redirect_map,
+                        'WEBROOT': '/srv',
+                    },
+                }
+            },
```
dimaqq, line 103: We're relying on K8s resolving the service name to all units' ip addresses in IngressPerAppRequirer.

Which means that K8s effectively uses readiness checks to route or not route requests to specific units.

Thus, ideally we'd specify the readiness check... unless those have special meaning in Juju or we can't control those through  Pebble.

examples/gosherve-demo/demo-routes
```diff
@@ -0,0 +1,13 @@
+ops-repo https://github.com/canonical/operator/
+ops https://ops.readthedocs.io/
```
dimaqq, line 2: Could be updated now :)

examples/gosherve-demo/pyproject.toml
```diff
@@ -0,0 +1,146 @@
...
+target-version = "py312"
+
+# Ruff formatter configuration
+[tool.ruff.format]
+quote-style = "single"
+exclude = [
+    "lib/*",  # Ignore libs fetched during our testing (or by a charmer)
+]
+
+[tool.ruff.lint]
```
dimaqq, line 66: WDYT about not specifying tool config here and relying on monorepo config instead?

dimaqq (commented):
Some fly-by comments, in case I forget later.

tonyandrewmeyer:
I've changed my mind here - I was going to be doing a small Juju/ops demo and thought I'd reuse this to save time but ended up not doing the demo, and I'm no longer sure that we should have this in examples rather than do fresh content or just use one of the tutorials.

PR #1897, opened 2025-07-07
fix: if self.app is not actually set avoid a new crash location

tonyandrewmeyer:
`Model.app` should always be set (and code comments explicitly state that). However, [a Juju bug where JUJU_REMOTE_APP is not set](https://bugs.launchpad.net/juju/+bug/1960934) means that is not always true.

We work around this [elsewhere](https://github.com/canonical/operator/blob/a15646475eab8d243ef03c31d0e76ccb02941050/ops/charm.py#L570), so we should do it when potentially adding in the departing unit as well. Once the upstream issue is solved, we should be able to remove this.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1781,6 +1781,10 @@ class Relation:
         if (
             _remote_unit is not None
             and not is_peer
+            # In practice, the "self.app will not be None" statement above is not
+            # necessarily true. Once https://bugs.launchpad.net/juju/+bug/1960934
+            # is resolved, we should be able to remove the next line.
+            and self.app is not None
             and _remote_unit.name.startswith(f'{self.app.name}/')
         ):
             self.units.add(_remote_unit)


ops/model.py
```diff
@@ -1781,6 +1781,10 @@ def __init__(
         if (
             _remote_unit is not None
             and not is_peer
+            # In practice, the "self.app will not be None" statement above is not
+            # necessarily true. Once https://bugs.launchpad.net/juju/+bug/1960934
+            # is resolved, we should be able to remove the next line.
+            and self.app is not None
```
dimaqq, line 1787: Actually on the second read, I'm not sure about this fix.

If I understand the referenced Juju bug correctly, the missing REMOTE_UNIT and REMOTE_APP always go hand in hand.

And if that's the case, `_remote_unit` check above should have caught it.

Unless... perhaps, `_remote_unit is not None` because `_remote_unit == ""` ?

tonyandrewmeyer: The reported error was line 1874 with:

> AttributeError: 'NoneType' object has no attribute 'name'

There are two `.name` attributes there, `_remote_unit` and `self.app` - the problem must be with one of those being `None`. If `_remote_unit` was `""` then the error would be `'str' object has no attribute 'name'`.

`_remote_unit` cannot have been `None`, because there's a `_remote_unit is not None` in the `if` statement, earlier than the `.name` access.

That means that, regardless of the underlying cause, the problem must have been that `self.app` was `None`, and adding in the `self.app is not None` must prevent this from happening.

It's possible that the referenced Juju bug is not the underlying cause, and that `Model.app` can be `None` in other situations, even though the existing code said that it couldn't be. I don't think it's worth trying to reproduce those cases at the moment, because while the referenced bug is not fixed, it definitely can be `None`, and this PR definitely fixed the case that was causing problems.

PR #1903, opened 2025-07-11
chore: minor type checking fixes

tonyandrewmeyer:
Fix a variety of small type annotation issues.

These are found with `ty` and missed by `pyright` but should be fixed both because they are issues and because it would be good to be more prepared for a stronger type checker in the future.  (Note that we can't move to `ty` yet because (a) it's marked as not ready, and (b) there are a bunch of false positives still.)

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3713,7 +3713,9 @@ class _TestingPebbleClient:
                     f'not {data.__class__.__name__}'
                 )
             else:
-                return io.StringIO(typing.cast('str', data))
+                # pyright doesn't understand that data must be str at this point, but ty does.
+                assert isinstance(data, str)
+                return io.StringIO(data)
 
     def exec(
         self,
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -2220,7 +2220,10 @@ def _record_calls(cls: Any):
     return cls
 
 
-def _copy_docstrings(source_cls: Any):
+_T = TypeVar('_T')
+
+
+def _copy_docstrings(source_cls: type[object]) -> Callable[[_T], _T]:
     """Copy the docstrings from source_cls to target_cls.
 
     Use this as:
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -3587,7 +3587,7 @@ class _TestingPebbleClient:
                 file_path.write_bytes(source)
             else:
                 # If source is binary, open file in binary mode and ignore encoding param
-                is_binary = isinstance(source.read(0), bytes)  # type: ignore
+                is_binary = isinstance(source.read(0), bytes)
                 open_mode = 'wb' if is_binary else 'w'
                 open_encoding = None if is_binary else encoding
                 with file_path.open(open_mode, encoding=open_encoding) as f:
@@ -3716,8 +3716,6 @@ class _TestingPebbleClient:
                     f'not {data.__class__.__name__}'
                 )
             else:
-                # pyright doesn't understand that data must be str at this point, but ty does.
-                assert isinstance(data, str)
                 return io.StringIO(data)
 
     def exec(
--- a/ops/log.py
+++ b/ops/log.py
@@ -82,7 +82,9 @@ def setup_root_logging(
         handler.setFormatter(formatter)
         logger.addHandler(handler)
 
-    def except_hook(etype: type[BaseException], value: BaseException, tb: types.TracebackType):
+    def except_hook(
+        etype: type[BaseException], value: BaseException, tb: types.TracebackType | None
+    ):
         logger.error('Uncaught exception while in charm code:', exc_info=(etype, value, tb))
         if exc_stderr:
             print(f'Uncaught {etype.__name__} in charm code: {value}', file=sys.stderr)
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1940,8 +1940,6 @@ def _websocket_to_writer(ws: _WebSocket, writer: _WebsocketWriter, encoding: str
             # Received "end" command (EOF signal), stop thread
             break
 
-        # pyright doesn't understand that this must be bytes now, but ty does.
-        assert isinstance(chunk, bytes)
         if encoding is not None:
             chunk = chunk.decode(encoding)
         writer.write(chunk)
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -1940,8 +1940,10 @@ def _websocket_to_writer(ws: _WebSocket, writer: _WebsocketWriter, encoding: str
             # Received "end" command (EOF signal), stop thread
             break
 
+        # pyright doesn't understand that this must be bytes now, but ty does.
+        assert isinstance(chunk, bytes)
         if encoding is not None:
-            chunk = typing.cast('bytes', chunk).decode(encoding)
+            chunk = chunk.decode(encoding)
         writer.write(chunk)
 
 
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -218,10 +218,10 @@ _ServiceInfoDict = TypedDict(
 
 
 class _BodyHandler(Protocol):
-    def __call__(self, data: bytes, done: bool = False) -> None: ...
+    def __call__(self, data: bytearray, done: bool = False) -> None: ...
 
 
-_HeaderHandler = Callable[[bytes], None]
+_HeaderHandler = Callable[[bytearray], None]
 
 # tempfile.NamedTemporaryFile has an odd interface because of that
 # 'name' attribute, so we need to make a Protocol for it.
@@ -230,7 +230,7 @@ _HeaderHandler = Callable[[bytes], None]
 class _Tempfile(Protocol):
     name = ''
 
-    def write(self, data: bytes): ...
+    def write(self, data: bytearray): ...
 
     def close(self): ...
 
@@ -2763,7 +2763,7 @@ class Client:
         elif isinstance(source, bytes):
             source_io: _AnyStrFileLikeIO = io.BytesIO(source)
         else:
-            source_io: _AnyStrFileLikeIO = source  # type: ignore
+            source_io: _AnyStrFileLikeIO = source
         boundary = binascii.hexlify(os.urandom(16))
         path_escaped = path.replace('"', '\\"').encode('utf-8')
         content_type = f'multipart/form-data; boundary="{boundary.decode("utf-8")}"'
@@ -3484,7 +3484,7 @@ class _FilesParser:
         # these, so we'll prime the parser buffer with that missing sequence.
         self._parser.feed(b'\r\n')
 
-    def _process_header(self, data: bytes):
+    def _process_header(self, data: bytearray):
         parser = email.parser.BytesFeedParser()
         parser.feed(data)
         self._headers = parser.close()
@@ -3504,7 +3504,7 @@ class _FilesParser:
 
         self._part_type = typing.cast('Literal["response", "files"]', name)
 
-    def _process_body(self, data: bytes, done: bool = False):
+    def _process_body(self, data: bytearray, done: bool = False):
         if self._part_type == 'response':
             self._response_data.extend(data)
             if done:
@@ -3655,15 +3655,15 @@ class _MultipartParser:
                         return  # terminal boundary reached
                 elif safe_bound > self._pos:
                     # write partial body data
-                    data = self._buf[self._pos : safe_bound]
+                    partial_data = self._buf[self._pos : safe_bound]
                     self._pos = safe_bound
-                    self._handle_body(data)
+                    self._handle_body(partial_data)
                     return  # waiting for more data
                 else:
                     return  # waiting for more data
 
 
-def _next_part_boundary(buf: bytes, marker: bytes, start: int = 0) -> tuple[int, int, bool]:
+def _next_part_boundary(buf: bytearray, marker: bytes, start: int = 0) -> tuple[int, int, bool]:
     """Returns the index of the next boundary marker in buf beginning at start.
 
     Returns:
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -4763,7 +4763,7 @@ class TestTestingPebbleClient:
         # It is a common mistake to pass just a name vs a list of names, so catch it with a
         # TypeError
         with pytest.raises(TypeError):
-            client.get_services('foo')
+            client.get_services('foo')  # type: ignore
 
     def test_get_services_subset(self, client: _TestingPebbleClient):
         client.add_layer(
@@ -4816,7 +4816,7 @@ class TestTestingPebbleClient:
         # Start service takes a list of names, but it is really easy to accidentally pass just a
         # name
         with pytest.raises(TypeError):
-            client.start_services('unknown')
+            client.start_services('unknown')  # type: ignore
 
     def test_start_service_no_services(self, client: _TestingPebbleClient):
         with pytest.raises(pebble.APIError):
@@ -4826,7 +4826,7 @@ class TestTestingPebbleClient:
         # Start service takes a list of names, but it is really easy to accidentally pass just a
         # name
         with pytest.raises(TypeError):
-            client.stop_services('unknown')
+            client.stop_services('unknown')  # type: ignore
 
     def test_stop_service_no_services(self, client: _TestingPebbleClient):
         with pytest.raises(pebble.APIError):
@@ -4951,7 +4951,7 @@ class TestTestingPebbleClient:
         # Restart service takes a list of names, but it is really easy to
         # accidentally pass just a name.
         with pytest.raises(TypeError):
-            client.restart_services('unknown')
+            client.restart_services('unknown')  # type: ignore
 
     def test_restart_service_no_services(self, client: _TestingPebbleClient):
         with pytest.raises(pebble.APIError):
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -4951,6 +4951,7 @@ class TestTestingPebbleClient:
         # Restart service takes a list of names, but it is really easy to
         # accidentally pass just a name.
         with pytest.raises(TypeError):
+            # This should have a type:ignore, but pyright gets that wrong (ty gets it right).
             client.restart_services('unknown')
 
     def test_restart_service_no_services(self, client: _TestingPebbleClient):
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -4951,7 +4951,6 @@ class TestTestingPebbleClient:
         # Restart service takes a list of names, but it is really easy to
         # accidentally pass just a name.
         with pytest.raises(TypeError):
-            # This should have a type:ignore, but pyright gets that wrong (ty gets it right).
             client.restart_services('unknown')
 
     def test_restart_service_no_services(self, client: _TestingPebbleClient):
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -226,7 +226,7 @@ class Runtime:
         config_yaml = virtual_charm_root / 'config.yaml'
         actions_yaml = virtual_charm_root / 'actions.yaml'
 
-        metadata_files_present: dict[Path | str | None] = {
+        metadata_files_present: dict[Path, str | None] = {
             file: file.read_text() if charm_virtual_root_is_custom and file.exists() else None
             for file in (metadata_yaml, config_yaml, actions_yaml)
         }
@@ -350,6 +350,9 @@ class Runtime:
                 # If ops raised _Abort(0) then we want to treat that as normal completion.
                 if e.exit_code != 0:
                     raise
+                # If _Abort was raised before we created the instance, we can't get the state, so
+                # still want to fail.
+                assert ops is not None
             except Exception as e:
                 # The following is intentionally on one long line, so that the last line of pdb
                 # output shows the error message (pdb shows the "raise" line).
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -342,17 +342,16 @@ class Runtime:
                     juju_context=juju_context,
                 )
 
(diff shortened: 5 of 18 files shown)


testing/tests/test_e2e/test_state.py
```diff
@@ -984,11 +998,11 @@ class Charm(ops.CharmBase):
...
-    assert len(state.relations) == 5
+    assert len(state.containers) == 2
     assert isinstance(state.relations, frozenset)
-    assert len(state.storages) == 1
+    assert len(state.relations) == 5
     assert isinstance(state.storages, frozenset)
-    assert len(state.stored_states) == 1
+    assert len(state.storages) == 1
     assert isinstance(state.stored_states, frozenset)
+    assert len(state.stored_states) == 1
```
tonyandrewmeyer, line 1008: This looks odd in the diff but what it's doing is moving the `isinstance` assertion above the `len` one, because `len(Iterable)` can't be done, so we need to verify that it has a `__len__` first.

ops/_private/harness.py
```diff
@@ -3713,7 +3713,9 @@ def _transform_exec_handler_output(
                     f'not {data.__class__.__name__}'
                 )
             else:
-                return io.StringIO(typing.cast('str', data))
+                # pyright doesn't understand that data must be str at this point, but ty does.
+                assert isinstance(data, str)
+                return io.StringIO(data)
```
james-garner-canonical: This is because of the whole "`bytes` in a type annotation is a shorthand for `bytes | bytearray | memoryview`" thing, which was then dropped in [PEP-688](https://peps.python.org/pep-0688/#no-special-meaning-for-bytes). We actually explicitly opt into the old behaviour in our `pyproject.toml` with `disableBytesTypePromotions = false`. This flag was [added to pyright in 2023](https://github.com/microsoft/pyright/issues/5697#issuecomment-1734159067) and defaults to `true` in strict mode. I assume that `ty` doesn't care about this flag, and follows the modern behaviour of *not* treating `bytes` as a type checking shorthand for `bytes | bytearray | memoryview`.

To be clear, with our current `pyright` settings, if we have `data: bytes`, then `reveal_type(data)` gives `bytes | bytearray | memoryview`. At runtime, `isinstance(data, bytes)` will be `False` for `bytearray` or `memoryview`, so `pyright` correctly infers that in the `else` branch, `data` could be one of those types.

If wanted to maintain backwards compatibility with this interpretation, the modern thing would be `data: str | collections.abc.Buffer`, and `isinstance(data, Buffer)`, but that type wasn't added till Python 3.12. We could instead  do `isinstance(data, (bytes, bytearray, memoryview))` with `data: bytes` and `disableBytesTypePromotion = false` or `data: bytes | bytearray | memoryview` and `disableBytesTypePromotion = true` (though we'd need to see where else changing this flag would matter. Note that if we went this route, we'd need to change the logic in this branch, since `memoryview` doesn't have a `decode` method.

On the other hand, at least in this method, we don't support `bytearray` or `memoryview` at runtime anyway, since we'll try to `io.StringIO` them and get an immediate `TypeError` (or `ValueError` because of `encoding` being set). So I think the correct fix is `disableBytesTypePromotion = true` and then just drop the `typing.cast` here -- no need for the assert. But we'd need to investigate all our other uses of `bytes` in type annotations and see if any of them need to become `bytes | bytearray | memoryview`.

ops/pebble.py
```diff
@@ -1940,8 +1940,10 @@ def _websocket_to_writer(ws: _WebSocket, writer: _WebsocketWriter, encoding: str
             # Received "end" command (EOF signal), stop thread
             break
 
+        # pyright doesn't understand that this must be bytes now, but ty does.
+        assert isinstance(chunk, bytes)
         if encoding is not None:
-            chunk = typing.cast('bytes', chunk).decode(encoding)
+            chunk = chunk.decode(encoding)
```
james-garner-canonical: This seems like the same issue as above. I think `typing.cast` can go away without the need for an `assert` if we set `disableBytesTypePromotion = true`, with presumably no concerns about user facing compatibility.

test/test_testing.py
```diff
@@ -4951,6 +4951,7 @@ def test_restart_service_str(self, client: _TestingPebbleClient):
         # Restart service takes a list of names, but it is really easy to
         # accidentally pass just a name.
         with pytest.raises(TypeError):
+            # This should have a type:ignore, but pyright gets that wrong (ty gets it right).
             client.restart_services('unknown')
```
james-garner-canonical: We can fix this by changing the definition of the `_copy_docstrings(source_cls)` decorator factory applied to `_TestingPebbleClient` like so:
```py
_T = typing.TypeVar('_T')
def _copy_docstrings(source_cls: type[object]) -> typing.Callable[[_T], _T]:
```

testing/src/scenario/_runtime.py
```diff
@@ -226,7 +226,7 @@ def _virtual_charm_root(self):
         config_yaml = virtual_charm_root / 'config.yaml'
         actions_yaml = virtual_charm_root / 'actions.yaml'
 
-        metadata_files_present: dict[Path | str | None] = {
+        metadata_files_present: dict[Path, str | None] = {
```
james-garner-canonical, line 229: This is presumably a mistake by me when updating our annotations. Crazy that `pyright` somehow doesn't complain about this.

testing/src/scenario/_runtime.py
```diff
@@ -350,6 +350,9 @@ def exec(
                 # If ops raised _Abort(0) then we want to treat that as normal completion.
                 if e.exit_code != 0:
                     raise
+                # If _Abort was raised before we created the instance, we can't get the state, so
+                # still want to fail.
+                assert ops is not None
```
james-garner-canonical: Hmm, so we can get `_Abort` even during the `Ops` call? (before creating the instance)

Raising an assertion error feels wrong in that case. I'd rather see:
```suggestion
                if ops is None:
                    raise
```

On the other hand, if we can't actually get `_Abort` from `Ops`, then I'd prefer to see the `try ... except` restructured here. For example, a nested `try`, `except _Abort` and `finally` for the cleanup conditional on `ops` existing could be wrapped around just `yield ops`.

testing/tests/test_e2e/test_actions.py
```diff
@@ -103,6 +104,7 @@ def handle_evt(_: CharmBase, evt: ActionEvent):
     ctx = Context(mycharm, meta={'name': 'foo'}, actions={'foo': {}})
     with pytest.raises(ActionFailed) as exc_info:
         ctx.run(ctx.on.action('foo'), State())
+    assert isinstance(exc_info.value, ActionFailed)
```
james-garner-canonical: `pyright` correctly infers this type. I wouldn't add this assertion, the type checker shouldn't need it, and at testing time it's entirely redundant with `pytest.raises(ActionFailed)`.

testing/tests/test_e2e/test_actions.py
```diff
@@ -123,6 +125,7 @@ def _on_foo_action(self, event):
     ctx = Context(MyCharm, meta={'name': 'foo'}, actions={'foo': {}})
     with pytest.raises(ActionFailed) as exc_info:
         ctx.run(ctx.on.action('foo'), State())
+    assert isinstance(exc_info.value, ActionFailed)
```
james-garner-canonical: > `pyright` correctly infers this type. I wouldn't add this assertion, the type checker shouldn't need it, and at testing time it's entirely redundant with `pytest.raises(ActionFailed)`.

Same comment.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -347,6 +347,7 @@ def test_exec_wait_error(charm_cls):
         proc = container.exec(['foo'])
         with pytest.raises(ExecError) as exc_info:
             proc.wait_output()
+        assert isinstance(exc_info.value, ExecError)
```
james-garner-canonical: > `pyright` correctly infers this type. I wouldn't add this assertion, the type checker shouldn't need it, and at testing time it's entirely redundant with `pytest.raises(ActionFailed)`.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -476,7 +477,10 @@ def _on_check_failed(self, event):
             infos.append(event.info)
 
     ctx = Context(MyCharm, meta={'name': 'foo', 'containers': {'foo': {}}})
-    layer = pebble.Layer({'checks': {'http-check': {'override': 'replace', 'startup': 'enabled'}}})
+    layer = pebble.Layer({
+        'checks': {'http-check': {'override': 'replace', 'startup': 'enabled', 'threshold': 3}}
+    })
+    assert layer.checks['http-check'].threshold is not None
```
james-garner-canonical, line 482: Expanding test coverage to get around a type checker complaint later in the test, nice!

testing/tests/test_e2e/test_play_assertions.py
```diff
@@ -33,7 +34,7 @@ def _on_event(self, event):
 
 def test_charm_heals_on_start(mycharm):
     def pre_event(charm):
-        pre_event._called = True
+        pre_event._called = True  # type: ignore
```
james-garner-canonical: Do these even get asserted on? Can we just drop them?

testing/tests/test_e2e/test_play_assertions.py
```diff
@@ -42,13 +43,15 @@ def call(charm, _):
             charm.unit.status = ActiveStatus('yabadoodle')
 
     def post_event(charm):
-        post_event._called = True
+        post_event._called = True  # type: ignore
```
james-garner-canonical: > Do these even get asserted on? Can we just drop them?

testing/tests/test_e2e/test_relations.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, framework: Framework):
         def _on_event(self, event):
             if self._call:
                 MyCharm.called = True
+                assert callable(MyCharm._call)
```
james-garner-canonical: Maybe better to add a type annotation to `_call` above, something like:
`_call: Callable[[Event], None] | None = None`

testing/tests/test_e2e/test_secrets.py
```diff
@@ -575,7 +576,7 @@ def _on_start(self, _):
 
 def test_no_additional_positional_arguments():
     with pytest.raises(TypeError):
-        Secret({}, {}, None)
+        Secret({}, {}, None)  # type: ignore
```
james-garner-canonical: Can we make this pass type checking by changing the arguments? e.g. `Secret({}, {})` or `Secret({}, None)` should pass type checking but fail at runtime due to `_max_posargs(1)`.

testing/tests/test_e2e/test_state.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, framework: Framework):
         def _on_event(self, event):
             if self._call:
                 MyCharm.called = True
+                assert callable(MyCharm._call)
```
james-garner-canonical: > Maybe better to add a type annotation to `_call` above, something like:
`_call: Callable[[Event], None] | None = None`

testing/tests/test_e2e/test_state.py
```diff
@@ -308,7 +310,7 @@ def test_model_positional_arguments():
 
 def test_container_positional_arguments():
     with pytest.raises(TypeError):
-        Container('', '')
+        Container('', '')  # type: ignore
```
james-garner-canonical:
```suggestion
        Container('', True)
```

testing/tests/test_e2e/test_state.py
```diff
@@ -812,15 +818,19 @@ class Charm(ops.CharmBase):
     )
     state = State.from_context(ctx)
     assert state.config == {'foo': 'bar'}
+    assert isinstance(state.containers, frozenset)
```
james-garner-canonical, line 821: Do we want to assert on this, given that the public type is `Iterable[...]`? A lot of the time we convert these things into a `tuple` in the test anyway, how about `containers = tuple(state.containers)` and likewise for all these other `frozenset` assertions?

testing/tests/test_runtime.py
```diff
@@ -158,4 +158,5 @@ def _on_start(self, _: ops.StartEvent):
     else:
         with pytest.raises(_Abort) as exc:
             ctx.run(ctx.on.start(), State())
+        assert isinstance(exc.value, _Abort)
```
james-garner-canonical: As mentioned earlier, I think that asserting on `exc.value` after `pytest.raises(...)` isn't desirable, and the type checker should already know the type (`pyright` seems to).

testing/tests/test_e2e/test_actions.py
tonyandrewmeyer: Fair enough.

```suggestion
```

testing/tests/test_e2e/test_actions.py
tonyandrewmeyer:
```suggestion
```

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer:
```suggestion
```

testing/tests/test_runtime.py
tonyandrewmeyer:
```suggestion
        with pytest.raises(_Abort) as exc:
            ctx.run(ctx.on.start(), State())
```

testing/tests/test_e2e/test_state.py
tonyandrewmeyer: These are a bit messy - I regret not going with a custom `__init__` for all these classes.

I feel that the API is that the attribute is a `frozenset` (or `collections.abc.Set`) but the `__init__` type is `Iterable`. Basically: when creating the object, we accept anything that be iterated, but when you access it, you'll get something that is immutable and set-like.

So I do feel that asserting that the passed object gets converted to `frozenset` is important (I'd be fine with asserting that it matched `collections.abc.Set`, but realistically it's going to be a `frozenset`).

testing/tests/test_e2e/test_secrets.py
tonyandrewmeyer:
```suggestion
        Secret({}, {})
```

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer: Mostly this makes me (again) feel that I got the testing story wrong for checks. Tricky to change at this stage, though, while maintaining backwards compatibility.

IronCore864 (approved):

ops/_private/harness.py
dimaqq: The PEP is stamped with Python 3.12.

Does us supporting both 3.10 and 3.12 mean that we need hacks, or does newest pyright with strict settings adhere to the PEP regardless of the runtime version?

ops/log.py
```diff
@@ -82,7 +82,9 @@ def custom_showwarning(
         handler.setFormatter(formatter)
         logger.addHandler(handler)
 
-    def except_hook(etype: type[BaseException], value: BaseException, tb: types.TracebackType):
+    def except_hook(
+        etype: type[BaseException], value: BaseException, tb: types.TracebackType | None
```
dimaqq, line 86: **TIL** https://github.com/python/typeshed/pull/6133

dimaqq (commented):
**praise:** ty fer doin dis!

ops/_private/harness.py
james-garner-canonical: Pyright lets you opt into the Python 3.12+ behaviour with `disableBytesTypePromotion = true`, so we don't need any hacks there -- just need to check if there are cases where we've promised to accept `bytearray` or `memoryview` and wouldn't just error out immediately at runtime (or rule out that charmers use those types with these methods).

testing/tests/test_e2e/test_state.py
james-garner-canonical: In that case, I'm on board with asserting on `frozenset` here, thanks for explaining.

ops/_private/harness.py
tonyandrewmeyer: It looks like things are fine:

```
ops/pebble.py
221:    def __call__(self, data: bytes, done: bool = False) -> None: ...
3664:def _next_part_boundary(buf: bytes, marker: bytes, start: int = 0) -> tuple[int, int, bool]:
```

Does `+` with bytes so can't be a memoryview. Does work with a bytearray, but we get our boundary from using the email package and that will only return actual bytes.

```
224:_HeaderHandler = Callable[[bytes], None]
233:    def write(self, data: bytes): ...
246:_AnyStrFileLikeIO = Union[_FileLikeIO[bytes], _FileLikeIO[str]]
248:_IOSource = Union[str, bytes, _AnyStrFileLikeIO]
2213:        data: bytes | Generator[bytes, Any, Any] | None = None,
2769:        def generator() -> Generator[bytes, None, None]:
2789:            content: str | bytes = source_io.read(self._chunk_size)
3485:    def _process_header(self, data: bytes):
3505:    def _process_body(self, data: bytes, done: bool = False):
3532:    def feed(self, data: bytes):
3461:    def __init__(self, boundary: bytes | str):
3571:        marker: bytes,
3617:    def feed(self, data: bytes):
```

We always serialise the provided data to JSON and then encode it. However, `push` takes a `str | bytes` (plus a "file like" protocol that can be `str` or `bytes`). However, this will fail with `TypeError` if passed a `memoryview` or `bytearray` because of being used as input into `io.StringIO` (as @james-garner-canonical mentioned above).

```
356:    def send_binary(self, payload: bytes): ...
358:    def recv(self) -> str | bytes: ...
1958:    def write(self, chunk: str | bytes) -> int:
1982:    def read(self, n: int = -1) -> str | bytes:
2011:        result: str | bytes = self.remaining[:n]
2015:    def read1(self, n: int = -1) -> str | bytes:
```

All for the websocket handling. From what I can tell, on the receiving side, we will always get actual bytes. On the writing side, we raise `TypeError` if there's any attempt to send something that isn't actually bytes.

```
2943:        stdin: bytes | BinaryIO | None = None,
2962:        stdin: str | bytes | TextIO | BinaryIO | None = None,
```

Will raise `TypeError` because it's not `str` or `bytes` and doesn't have `.read`.

```
2948:    ) -> ExecProcess[bytes]: ...
```

Covered in the `ExecProcess` lines, but basically, this is what we get from the websocket.

```
test/test_pebble.py
1416:_bytes_generator = typing.Generator[bytes, typing.Any, typing.Any]
1449:        data: bytes | _bytes_generator | None = None,
1461:    def __init__(self, headers: dict[str, str], body: bytes):
1519:    data: bytes
1520:    want_headers: list[bytes]
1521:    want_bodies: list[bytes]
````

Same as the real Pebble.

```
1619:            headers: list[bytes] = []
1620:            bodies: list[bytes] = []
1631:            def handle_body(data: bytes, done: bool = False):
2506:    def _test_push_bytes(self, client: MockClient, source: bytes | typing.IO[bytes]):
```

Byte literals in the test data.

```
3375:        self.sends: list[tuple[str, str | bytes]] = []
3376:        self.receives: list[str | bytes] = []
3378:    def send_binary(self, b: bytes):
3963:        def send_binary(b: bytes):
```

Same as the real Pebble.

```
test/test_testing.py
5046:        original_data: str | bytes,
```

Used with byte literals.

```
test/test_real_pebble.py
279:        reads: list[bytes] = list(process.stdout)
```

This is stdout from a subprocess, so will always be actual bytes.

```
ops/model.py
2805:        source: bytes | str | BinaryIO | TextIO,
```

Wrapper of the Pebble `push`.

```
3235:        stdin: bytes | BinaryIO | None = None,
3240:    ) -> pebble.ExecProcess[bytes]: ...
3254:        stdin: str | bytes | TextIO | BinaryIO | None = None,
```

Wrapper of the Pebble `exec`.

```
ops/_private/harness.py
79:ReadableBuffer = Union[bytes, str, StringIO, BytesIO, BinaryIO]
```

Used for the mocked Pebble.

```
144:    stdin: str | bytes | None
158:    stdout: str | bytes = b''
159:    stderr: str | bytes = b''
1994:        result: int | str | bytes | ExecResult | None = None,
```

For the Pebble mocking, so if it's ok there it should be ok here (we don't want tests to allow more than the real client).

```
2379:        self._resources_map: dict[str, tuple[str, str | bytes]] = {}
```

Output from `yaml.dump`.

```
3702:        self, data: str | bytes, encoding: str | None
3730:        stdin: str | bytes | TextIO | BinaryIO | None = None,
```

Also for the Pebble mocking.

```
tracing/ops_tracing/_buffer.py
188:    def pushpop(self, record: tuple[bytes, str] | None = None) -> tuple[int, bytes, str] | None:
```

Passed into `conn.execute`, comes from otlp_json.encode_spans, which returns a JSON string encoded to bytes.

```
tracing/ops_tracing/_export.py
130:    def do_export(self, buffered_id: int, data: bytes, mime: str) -> None:
```

Comes from the pushpop above, so the same.

```
testing/src/scenario/mocking.py
926:        stdin: str | bytes | TextIO | None = None,
```

This is the mocked Pebble `exec` so the same as there.

james-garner-canonical: Thanks for investigating this. I'm curious why a lot of the pebble type annotations have been changed to `bytearray` now.

tonyandrewmeyer: The parser takes bytes but internally it puts those into a bytearray and then processes that. Basically, all those types annotations of private stuff were relying on `bytes` also meaning `bytearray`.

james-garner-canonical: Ah I see, so I guess `bytes | bytearray` would also satisfy the type checker now that we have `disableBytesTypePromotion = true`. Happy for you to make the call on which annotation would be best here.

james-garner-canonical (approved):
Thanks for doing all this!

ops/_private/harness.py
tonyandrewmeyer: All the ones that are now `bytearray` I checked (by hand) and are actually only passing around `bytearray`s so that seems best.

PR #1904, opened 2025-07-14
docs: add a Security explanation doc

tonyandrewmeyer:
Adds a Security doc in the explanation section.

This docs covers:

* Cryptography use
* Hardening (not required)
* Information about the unit state and tracing databases
* Information about Scenario creating new directories when running tests
* Getting security updates
* Security risks introduced/changed when using Ops[etc]
* Good security practices for charm authors

The goal is for this doc to be generally useful for charm authors, but also to comply with [SEC0030](https://docs.google.com/document/d/12_hj3_xhaYWqKHsmnuRl5wQO3Es4GuHWtgrxMHG21FM/edit?tab=t.0#heading=h.jszeo1fa8onu) and [Canonical product security documentation guidance](https://docs.google.com/document/d/1QBVUxfOVaTXEbBUiiWMzYHiAxG79NIGS7YTVX_GNE4Q/edit?tab=t.0#heading=h.ikq8mvhgar8v).

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -2,6 +2,8 @@
 
 # Security
 
+When producing security documentation for your charm, it's important to consider the security aspects of the charm's dependencies. If there are any security questions that aren't answered here in terms of the use of cryptographic technology, security risks, hardening, and good practice, with regards to Ops, please reach out to the Charm Tech team, and we'll do our best to assist.
+
 ## Cryptographic technology
 
 The only case where Ops uses cryptography is for sending trace data, when a certificate authority provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing security](#ops_tracing_security) for more details.


tonyandrewmeyer:
@canonical/charm-tech I've added you all as reviewers: I'd like an extensive review given that this is security. Please confirm that the content I've added is correct, but also if you have additional suggestions for good practice, hardening, and so forth, that would be great.

docs/explanation/security.md
```diff
@@ -0,0 +1,61 @@
...
+-rw-------  1 root root  20K Jul 13 23:48 .unit-state.db
+```
+
+## ops[testing]
+
+When testing an event with [](ops.testing.Context), the mocked unit state database and tracing data are stored in memory. Each event creates a new charm directory, which is provided by [](tempfile.TemporaryDirectory).
+
+## Security updates
+
+We strongly recommend pinning `ops` (and `ops[harness,testing,tracing]` in your dev dependencies) in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. If your charm needs to support Ubuntu 20.04 (with Python 3.8), then this looks like `ops~=2.23`. Otherwise, this looks like `ops~=3.0` (bump the minor number to be high enough that you get all the features that the charm uses).
```
james-garner-canonical: Wdyt about recommending `ops~=3` instead of `ops~=3.0`?

james-garner-canonical: Also, should we recommend here using a lock file to make it easy to know from the source what exact version of `ops` is packed into each charm release (making it easier to update compromised versions etc)?

docs/explanation/security.md
```diff
@@ -0,0 +1,62 @@
...
+
+The one case where Ops uses cryptography is for sending trace data, when a certificates provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing - security considerations](#ops_tracing_security) for more details.
+
+## Hardening
+
+Hardening a charm that uses Ops is done in the same way as any other charm: no extra hardening steps are required as a result of using Ops.
+
+## The charm unit databases
+
+Ops stores state ([](ops.StoredState) objects and the defer notice queue) in a sqlite3 database alongside the charm (the unit's machine for machine charms, and the charm container for Kubernetes charms). The database is named `.unit-state.db` and is in the charm directory (this is set by Juju in the `JUJU_CHARM_DIR` environment variable, and typically looks like `/var/lib/juju/agents/unit-my-unit-0/charm`). Ops sets the permissions for the database to allow only reading and writing by the user running the charm. You shouldn't try to edit this file or change its permissions.
```
james-garner-canonical: Do you think it's worth mentioning here that both stored state and deferred events are actively recommended against?

Should the stored state docs additionally include a pointer to using Juju secrets for storing sensitive information?

docs/explanation/security.md
tonyandrewmeyer: > Wdyt about recommending `ops~=3` instead of `ops~=3.0`?

It [doesn't have any meaning](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release):
> This operator MUST NOT be used with a single segment version number such as ~=1.

`~3.0` means "any version compatible with 3.0", which is really just shorthand for something like `>= 3.0, <4`. We're not saying "use 3.0". This is also why I have the note about bumping the minor to be high enough to get the features you need.

> Also, should we recommend here using a lock file to make it easy to know from the source what exact version of ops is packed into each charm release (making it easier to update compromised versions etc)?

It's pretty strongly implied by "every time that you re-lock". But calling out lock files explicitly as good practice is a good suggestion, yes.

docs/explanation/security.md
tonyandrewmeyer: > Do you think it's worth mentioning here that both stored state and deferred events are actively recommended against?

I don't think so - I think there are legitimate reasons to use both, and the reasons against using them don't have anything to do with security.

> Should the stored state docs additionally include a pointer to using Juju secrets for storing sensitive information?

Do you mean not in this doc but in the `StoredState` docstring or the "how to store state" guide? I'm not opposed to that, but I don't think it belongs in this PR. I'm happy to review one you submit :wink:.

docs/explanation/security.md
james-garner-canonical: Ah right, I'd forgotten the semantics of `~=`.

docs/explanation/security.md
james-garner-canonical: My thinking with both of these suggestions is about the flow for a newish user who lands here looking for best practices, perhaps after completing a tutorial or two. If we don't want to call out non-security best practices directly in this doc, then maybe the `StoredState` link could point to https://ops.readthedocs.io/en/latest/explanation/storedstate-uses-limitations.html instead of to the `ops.StoredState` reference.

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+-rw-------  1 root root  20K Jul 13 23:48 .unit-state.db
+```
+
+## ops[testing]
+
+When testing an event with [](ops.testing.Context), the mocked unit state database and tracing data are stored in memory. Each event creates a new charm directory, which is provided by [](tempfile.TemporaryDirectory).
+
+## Security updates
+
+We strongly recommend pinning `ops` (and `ops[harness,testing,tracing]` in your dev dependencies) in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. If your charm needs to support Ubuntu 20.04 (with Python 3.8), then this looks like `ops~=2.23`. Otherwise, this looks like `ops~=3.0` (bump the minor number to be high enough that you get all the features that the charm uses).
```
dimaqq: **terminology nit:** I used to think that pinning meant specifying an exact version, while we are advising to use a version range in pyproject.toml and locking in the lock file. Maybe a better choice of words is possible?

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+* Access to trace data provides detailed information about the implementation of the charm and the events that it has processed. This is particularly the case when tracing has not been configured, as the charm will have a large amount of buffered trace data stored (when tracing is active, this will be regularly sent to the trace receiver and removed from the local database).
+
+Ops inherits all the risks of Juju executing charms (for example, injecting data into the charm context or through the use of hook commands). Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>` and {external+pebble:ref}` Pebble security <security>`.
+
+If a charm is integrated with a tracing receiver, this introduces the risk of outgoing traces being intercepted. Traces should not include any sensitive data, but intercepted traces can provide information about the structure of the charm and the events that the charm has processed, and blocking trace data can hide malicious activity.
+
+## Good practices
+
+* Never include any sensitive data in logs.
+* Never include any sensitive data in traces.
```
dimaqq, line 74: I would expand this to exclude sensitive data from exceptions, as they are are often both logged and included in trace data.

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
+(security)=
+
+# Security
+
+## Cryptographic technology
+
+Ops uses almost no cryptographic technology. Communication with Juju is done via reading environment variables ({external+juju:ref}`Hook <hook>`) and running processes in the charm environment ({external+juju:ref}`Hook command <list-of-hook-commands>`). There is no use of hashing or digital signatures.
```
IronCore864: These first two paragraphs sound a little bit weird/negative to me (maybe because I'm not a native speaker), because it starts by saying "ops doesn't use ...", then followed by "except this only case that ..."

How about reordering it to something like "The only case ops use ... is ...", then "except that, ops doesn't use ..., because ..."

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+If a charm is integrated with a tracing receiver, this introduces the risk of outgoing traces being intercepted. Traces should not include any sensitive data, but intercepted traces can provide information about the structure of the charm and the events that the charm has processed, and blocking trace data can hide malicious activity.
+
+## Good practices
+
+* Never include any sensitive data in logs.
+* Never include any sensitive data in traces.
+* Use {external+juju:ref}`Juju secrets <secret>` for storing and sharing sensitive data.
+* Juju users that integrate a charm with a tracing receiver should also integrate with a certificate provider, to ensure all traces are sent via HTTPS.
+* Charms should follow best practices for writing secure Python code.
+* Charms should have workflows that statically check for security issues (such as [ruff](https://docs.astral.sh/ruff/linter/), [bandit](https://bandit.readthedocs.io/en/latest/index.html), and [zizmor](https://docs.zizmor.sh/)).
```
dimaqq: Hmmm I dunno if I would recommend bandit.

IronCore864 (approved):

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+Ops inherits all the risks of Juju executing charms (for example, injecting data into the charm context or through the use of hook commands). Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>` and {external+pebble:ref}` Pebble security <security>`.
+
+If a charm is integrated with a tracing receiver, this introduces the risk of outgoing traces being intercepted. Traces should not include any sensitive data, but intercepted traces can provide information about the structure of the charm and the events that the charm has processed, and blocking trace data can hide malicious activity.
+
+## Good practices
+
+* Never include any sensitive data in logs.
+* Never include any sensitive data in traces.
+* Use {external+juju:ref}`Juju secrets <secret>` for storing and sharing sensitive data.
+* Juju users that integrate a charm with a tracing receiver should also integrate with a certificate provider, to ensure all traces are sent via HTTPS.
```
dimaqq: Tracing needs a CA provider, not a certificate provider.

Note the arguably misnamed `interface: certificate_transfer`. All that really does is the provider sets the CA list that the charm picks up and sets as the CA list to verify the HTTPS host against.

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
+(security)=
+
+# Security
+
+## Cryptographic technology
+
+Ops uses almost no cryptographic technology. Communication with Juju is done via reading environment variables ({external+juju:ref}`Hook <hook>`) and running processes in the charm environment ({external+juju:ref}`Hook command <list-of-hook-commands>`). There is no use of hashing or digital signatures.
+
+The one case where Ops uses cryptography is for sending trace data, when a certificates provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing - security considerations](#ops_tracing_security) for more details.
```
dimaqq: Wow, does a hash-ref work across files? Sweet!

dimaqq (commented):
security.md:47: WARNING: external std:ref reference target not found: security [ref.ref]
https://app.readthedocs.org/projects/ops/builds/28841948/

dimaqq:
Ahh, https://github.com/canonical/pebble/pull/657 needs to be merged first.

docs/explanation/security.md
tonyandrewmeyer: When Sphinx processes the file, but not in GitHub, sadly.

docs/explanation/security.md
tonyandrewmeyer: I think this doc should be limited to security. However, a mention at the end of the "good practices" section about how to find best practices for charming seems reasonable.

The how-to guide recommends reading that doc first. I'm not opposed to adding a link to the reference doc as well, just not in this PR :smile:.

docs/explanation/security.md
tonyandrewmeyer: I read "pinning" as restricting which versions can be used, which might be a range (pin to a range) or a single version (pin to an exact version). But fair call.

```suggestion
We strongly recommend restricting the version of `ops` (and `ops[harness,testing,tracing]` in your dev dependencies) in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. If your charm needs to support Ubuntu 20.04 (with Python 3.8), then this looks like `ops~=2.23`. Otherwise, this looks like `ops~=3.0` (bump the minor number to be high enough that you get all the features that the charm uses).
```

docs/explanation/security.md
tonyandrewmeyer:
```suggestion
* Never include any sensitive data in logs.
* Never include any sensitive data in traces.
* Never include any sensitive data in exceptions.
```

docs/explanation/security.md
tonyandrewmeyer:
```suggestion
* Juju users that integrate a charm with a tracing receiver should also integrate with a certificate authority provider, to ensure all traces are sent via HTTPS.
```

docs/explanation/security.md
tonyandrewmeyer: Because you'd recommend just enabling the flake8-bandit rules in ruff instead? Or for some other reason? I'd use ruff these days, but bandit's pretty commonly used across Canonical charms and it doesn't seem unreasonable to me to continue to use it if you haven't adopted ruff.

docs/explanation/security.md
dwilding: I agree with Tiexin. How about this structure (including some minor wording edits):

```
## Cryptographic technology

The only case where Ops uses cryptography is [... rest of that paragraph]

There is no use of hashing or digital signatures.

## Inter-process communication

Ops communicates with Juju by reading environment variables (hooks) and running processes in the charm environment (hook commands).
```

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Hardening
+
+Hardening a charm that uses Ops is done in the same way as any other charm: no extra hardening steps are required as a result of using Ops.
```
dwilding, line 23: How about including a link to https://documentation.ubuntu.com/juju/3.6/howto/manage-your-deployment/#harden-your-deployment, if it's not too broad?

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+
+## The charm unit databases
```
dwilding:
```suggestion
## Charm unit databases
```

docs/explanation/security.md
dwilding: Couple of minor suggestions:

```suggestion
The one case where Ops uses cryptography is for sending trace data, when a certificate authority provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing security](#ops_tracing_security) for more details.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+
+Ops stores state ([](ops.StoredState) objects and the defer notice queue) in a sqlite3 database alongside the charm (the unit's machine for machine charms, and the charm container for Kubernetes charms). The database is named `.unit-state.db` and is in the charm directory (this is set by Juju in the `JUJU_CHARM_DIR` environment variable, and typically looks like `/var/lib/juju/agents/unit-my-unit-0/charm`). Ops sets the permissions for the database to allow only reading and writing by the user running the charm. You shouldn't try to edit this file or change its permissions.
```
dwilding: I found this part a bit hard to read. I'd suggest expanding it a bit - something like this:

```suggestion
Ops stores state in a sqlite3 database named `.unit-state.db`. This database includes [](ops.StoredState) objects and the defer notice queue.

The state database is in the charm directory, which is set by Juju in the `JUJU_CHARM_DIR` environment variable and typically looks like `/var/lib/juju/agents/unit-my-unit-0/charm`. For a machine charm, this directory is on the unit's machine. For a Kubernetes charm, this directory is in the charm container.

Ops sets the permissions for the state database to allow only reading and writing by the user running the charm. You shouldn't try to edit this file or change its permissions.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+
+Ops buffers tracing data in a sqlite3 database named `.tracing-data.db` in the same directory as the state database. Trace data is stored in this database only if the `tracing` extra is not installed, but even if tracing has not been enabled through an integration with a trace receiver (this allows collecting traces prior to the integration).
```
dwilding: I'm confused by this part. How about this:

```suggestion
Ops buffers tracing data in a sqlite3 database named `.tracing-data.db` in the same directory as the state database. Trace data is stored in this database only if the `tracing` extra is installed. When `tracing` is installed, Ops buffers tracing data even if tracing has not been enabled through an integration with a trace receiver (this allows collecting traces prior to the integration).
```

dwilding: This doc uses a mix of "tracing data" and "trace data". I think we should pick one term. What's your preference @dimaqq? This is relevant for https://github.com/canonical/operator/pull/1853 too.

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+
+For example:
```
dwilding: The example doesn't follow on directly from the previous paragraph, so how about:

```suggestion
For example, the permissions of the databases are:
```

docs/explanation/security.md
dwilding: Minor suggestion, as the paragraph already has several parts in brackets:

```suggestion
We strongly recommend restricting the version of `ops` (and `ops[harness,testing,tracing]` in your dev dependencies) in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. If your charm needs to support Ubuntu 20.04 (with Python 3.8), then this looks like `ops~=2.23`. Otherwise, this looks like `ops~=3.0`. Set a minor version that includes all the features that the charm uses.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Security updates
+
+We strongly recommend pinning `ops` (and `ops[harness,testing,tracing]` in your dev dependencies) in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. If your charm needs to support Ubuntu 20.04 (with Python 3.8), then this looks like `ops~=2.23`. Otherwise, this looks like `ops~=3.0` (bump the minor number to be high enough that you get all the features that the charm uses).
+
+Your charm repository should have tooling configured so that any dependencies with security updates are detected automatically (such as [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates) or [Renovate](https://www.mend.io/renovate/)), prompting to you re-lock so that the charm will be built with the latest version.
+
+For information about supported versions and how to report security issues, please see [SECURITY.md](https://github.com/canonical/operator/blob/main/SECURITY.md).
```
dwilding:
```suggestion
For information about supported versions and how to report security issues, see [SECURITY.md](https://github.com/canonical/operator/blob/main/SECURITY.md).
```

docs/explanation/security.md
```diff
@@ -0,0 +1,63 @@
...
+* Access to trace data provides detailed information about the implementation of the charm and the events that it has processed. This is particularly the case when tracing has not been configured, as the charm will have a large amount of buffered trace data stored (when tracing is active, this will be regularly sent to the trace receiver and removed from the local database).
+
+Ops inherits all the risks of Juju executing charms (for example, injecting data into the charm context or through the use of hook commands). Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>` and {external+pebble:ref}` Pebble security <security>`.
+
+If a charm is integrated with a tracing receiver, this introduces the risk of outgoing traces being intercepted. Traces should not include any sensitive data, but intercepted traces can provide information about the structure of the charm and the events that the charm has processed, and blocking trace data can hide malicious activity.
```
dwilding: I had to read this a couple of times to fully understand the different aspects. How about reorganising like this, if I understood correctly:

```suggestion
Ops inherits the risks of Juju executing charms (for example, injecting data into the charm context or through the use of hook commands). Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>` and {external+pebble:ref}` Pebble security <security>`.

If a charm is integrated with a tracing receiver, Ops introduces the risk of outgoing traces being intercepted. Traces should not include any sensitive data, but intercepted traces can provide information about the structure of the charm and the events that the charm has processed. In addition, an attacker that blocked trace data could hide malicious activity.

Otherwise, Ops doesn't introduce any new security risks. Ops does expand the impact of the Juju risk of an attacker gaining access to the filesystem of the charm:

* Access to the deferred notice queue provides information about events that could not be immediately processed (this includes the event name and the event context provided in the hook environment at the time).
* Access to trace data provides detailed information about the implementation of the charm and the events that it has processed. This is particularly the case when tracing has not been configured, as the charm will have a large amount of buffered trace data stored (when tracing is active, this will be regularly sent to the trace receiver and removed from the local database).
```

dwilding (commented):
There were a couple of parts I didn't fully understand at first, so I've suggested some changes based on my interpretation. I'll review again after you've had a look, as you might need to correct some of my suggestions.

docs/explanation/security.md
tonyandrewmeyer:
```suggestion
Hardening a charm that uses Ops is done in the same way as any other charm: no extra hardening steps are required as a result of using Ops.

> See also: {external+juju:ref}`Juju | Harden your deployment <harden-your-deployment>`
```

dwilding: Looks good, thanks

docs/explanation/security.md
tonyandrewmeyer:
```suggestion
The only case where Ops uses cryptography is for sending trace data, when a certificate authority provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing security](#ops_tracing_security) for more details.

There is no use of hashing or digital signatures.

## Inter-process communication

Ops communicates with Juju by reading environment variables and running processes in the charm environment (hook commands).

> See also:
> - {external+juju:ref}`Juju | Hook <hook>`
> - {external+juju:ref}`Juju |Hook command <list-of-hook-commands>`
```

docs/explanation/security.md
```diff
@@ -0,0 +1,70 @@
+(security)=
+
+# Security
+
+## Cryptographic technology
+
+Ops uses almost no cryptographic technology. Communication with Juju is done via reading environment variables ({external+juju:ref}`Hook <hook>`) and running processes in the charm environment ({external+juju:ref}`Hook command <list-of-hook-commands>`). There is no use of hashing or digital signatures.
+
+The one case where Ops uses cryptography is for sending trace data, when a certificate authority provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing security](#ops_tracing_security) for more details.
```
tonyandrewmeyer:
```suggestion
```

docs/explanation/security.md
dwilding: Looks good. How about this for the links:

```
> See also:
> - {external+juju:ref}`Juju | Hook <hook>`
> - {external+juju:ref}`Juju | Hook command <list-of-hook-commands>`
```

docs/explanation/security.md
dimaqq: This is "tools, such as ..." section, we can totally list a subset of common practice.

In my view: ruff includes a bunch of rules; and I've never found bandit actually useful.

docs/explanation/security.md
```diff
@@ -0,0 +1,70 @@
+(security)=
+
+# Security
```
dimaqq, line 3: The scope is unclear: is this ops library security or wider charming security and end-to-end security incl. juju, deployment and incident response?

docs/explanation/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+If a charm is integrated with a tracing receiver, Ops introduces the risk of outgoing traces being intercepted. Traces should not include any sensitive data, but intercepted traces can provide information about the structure of the charm and the events that the charm has processed. In addition, an attacker that blocked trace data could hide malicious activity.
+
+Otherwise, Ops doesn't introduce any new security risks. Ops does expand the impact of the Juju risk of an attacker gaining access to the filesystem of the charm:
+
+* Access to the deferred notice queue provides information about events that could not be immediately processed (this includes the event name and the event context provided in the hook environment at the time).
+* Access to trace data provides detailed information about the implementation of the charm and the events that it has processed. This is particularly the case when tracing has not been configured, as the charm will have a large amount of buffered trace data stored (when tracing is active, this will be regularly sent to the trace receiver and removed from the local database).
+
+## Good practices
```
dimaqq, line 71: Depending on the answer to scope, this section may or may not belong here.

It could be dropped or moved to how-to perhaps.

docs/explanation/security.md
tonyandrewmeyer: Our instructions say:

> Typically this should be titled something like Security in <product>, Security overview or simply Security. Its purpose is to give the user a good general perspective on security, and help orient their thinking about it.

I prefer "Security" over "Security overview" (I don't think "overview" is adding anything there) or "Security in X" where "X" might be "Ops" or "ops, ops[testing], ops[tracing]", or "Charm SDK" (what the security team are looking for, but we're defining as including Jubilant and Concierge which will have their own pages).

Would you prefer "Security overview"? Or are you suggesting adding an intro paragraph that explains what this doc is?

docs/explanation/security.md
tonyandrewmeyer: The guidance for this page says to include:

> topics such as common risks, good practices, built-in protection and so on

So that's what it has :smile:.

I think some how-to guides would be nice, and the guidance suggests those too, but they aren't required, so I'd rather leave them for future work than include them in this cycle's roadmap item.

docs/explanation/security.md
tonyandrewmeyer: I used bandit a lot in the pre-ruff days, and I found it regularly useful. However, I'm fine with being a bit more opinionated and saying to just use ruff (and get the bandit rules that way).

docs/explanation/security.md
james-garner-canonical: I'm still in favour of changing this doc's link to the `StoredState` [reference](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.StoredState) to a link to the `StoredState` [explanation](https://ops.readthedocs.io/en/latest/explanation/storedstate-uses-limitations.html#storedstate-uses-limitations) (and perhaps adding a link to the defer [explanation](https://ops.readthedocs.io/en/latest/explanation/how-and-when-to-defer-events.html#how-and-when-to-defer-events) to the mention of defer). My thinking is that when it comes to the slightly footgunier parts of the API, it's a good idea to points readers to our recommendations, especially in a doc that is presumably for users trying to do the 'right thing'. Not a blocker if you don't think it's worth worrying about though.

james-garner-canonical (approved):

dwilding (approved):

docs/explanation/security.md
```diff
@@ -0,0 +1,78 @@
...
+
+Ops sets the permissions for the state database to allow only reading and writing by the user running the charm. You shouldn't try to edit this file or change its permissions.
+
+Ops buffers tracing data in a sqlite3 database named `.tracing-data.db` in the same directory as the state database. Trace data is stored in this database only if the `tracing` extra is installed. When `tracing` is installed, Ops buffers tracing data even if tracing has not been enabled through an integration with a trace receiver (this allows collecting traces prior to the integration).
+
+For example, the permissions of the databases are:
+
+```text
+-rw-r--r--  1 root root  32K Jul 13 23:48 .tracing-data.db
+-rw-------  1 root root  20K Jul 13 23:48 .unit-state.db
```
benhoyt, line 45: Not a critique of this doc, but: I guess it's intended that the tracing db is readable by group and all, whereas the unit state db is only r/w by root?

docs/explanation/security.md
```diff
@@ -0,0 +1,78 @@
...
+
+* Never include any sensitive data in logs.
+* Never include any sensitive data in traces.
+* Never include any sensitive data in exceptions.
+* Use {external+juju:ref}`Juju secrets <secret>` for storing and sharing sensitive data.
+* Juju users that integrate a charm with a tracing receiver should also integrate with a certificate authority provider, to ensure all traces are sent via HTTPS.
+* Charms should follow best practices for writing secure Python code.
+* Charms should have workflows that statically check for security issues (such as [ruff](https://docs.astral.sh/ruff/linter/) and [zizmor](https://docs.zizmor.sh/)).
+* Charm authors should exercise caution when considering adding dependencies to their charms.
+* Write the exact dependencies of the charm into a lock file (using `uv lock`, `poetry lock`, and so on) and commit that lock file to source control.
```
benhoyt: Nit:
```suggestion
* Write the exact dependencies of the charm into a lock file (using `uv lock`, `poetry lock`, or similar tool) and commit that lock file to source control.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,78 @@
...
+* Charm repositories should have tooling that automatically detects outdated dependencies, particularly missing security updates.
+* For Kubernetes charms, run the charm as a non-root user when possible (via the {external+charmcraft:ref}`charm-user key in charmcraft.yaml <charmcraft-yaml-key-charm-user>` and specifying the `uid` and `gid` for containers).
+* Charm authors should harden their workloads by default.
```
benhoyt: Can we add a "for example" sentence for what kind of thing this is referring to?

benhoyt (approved):
Looks good to me, but added a couple of minor comments.

docs/explanation/security.md
tonyandrewmeyer: The unit state db being only r/w by root is deliberate. @dimaqq you'd be best placed to say whether it's intended or not that the tracing db is readable by group/all.

docs/explanation/security.md
tonyandrewmeyer:
```suggestion
* Charm authors should harden their workloads by default. For example, see charm guidance for [postgresql-k8s](https://canonical-charmed-postgresql-k8s.readthedocs-hosted.com/14/explanation/security/), [kubeflow](https://discourse.charmhub.io/t/security/15935), or workload hardening guidance such as [Wordpress](https://developer.wordpress.org/advanced-administration/security/hardening/).
```

docs/explanation/security.md
tonyandrewmeyer: I think changing the link is the wrong thing to do - it's a link to explain what stored state is, and is in the context of what data is stored and the security implications of that.

But I've added a "See also" below this with links to the two explanation docs (I also renamed the files so that they can be referenced unambiguously).

tonyandrewmeyer:
@dimaqq are you ok with this as it is now?

docs/explanation/security.md
```diff
@@ -0,0 +1,82 @@
...
+* Charms should have workflows that statically check for security issues (such as [ruff](https://docs.astral.sh/ruff/linter/) and [zizmor](https://docs.zizmor.sh/)).
+* Charm authors should exercise caution when considering adding dependencies to their charms.
+* Write the exact dependencies of the charm into a lock file (using `uv lock`, `poetry lock`, or similar tool) and commit that lock file to source control.
+* Charm repositories should have tooling that automatically detects outdated dependencies, particularly missing security updates.
+* For Kubernetes charms, run the charm as a non-root user when possible (via the {external+charmcraft:ref}`charm-user key in charmcraft.yaml <charmcraft-yaml-key-charm-user>` and specifying the `uid` and `gid` for containers).
+* Charm authors should harden their workloads by default. For example, see charm guidance for [postgresql-k8s](https://canonical-charmed-postgresql-k8s.readthedocs-hosted.com/14/explanation/security/), [kubeflow](https://discourse.charmhub.io/t/security/15935), or workload hardening guidance such as [Wordpress](https://developer.wordpress.org/advanced-administration/security/hardening/).
```
dimaqq, line 84: **minor:** These best charming practices are quite outside of Ops scope.

docs/explanation/security.md
dimaqq: Those documents are not public.
I have access, but the wider community doesn't.

docs/explanation/security.md
dimaqq: I don't think this matters in practice.

In the k8s world, we've got a whole container for one charm.

In the VM world, another charm's workload could be running as a different uid, and that's probably the origin of unit state db being readable by root only (that is by any charm code deployed to this VM, and any workload that's not snapped or run as a different uid).

Maybe we'll hard the tracing db later.

docs/explanation/security.md
dimaqq: How about adding a paragraph between `# Security` and `## Cryptographic technology` and provide some context for our readers?

From a library user PoV, I'm not clear why this page is in Ops docs.
I'd say especially so for external readers who can't see "our instructions".

Perhaps the target audience are the "the security people"?

Some links would be useful too, something like "security at canonical" or "security in the Juju ecosystem" or maybe some process links?

dimaqq (approved):
Happy for this to be merged and improved later.

docs/explanation/security.md
```diff
@@ -0,0 +1,82 @@
+(security)=
+
+# Security
+
+## Cryptographic technology
+
+The only case where Ops uses cryptography is for sending trace data, when a certificate authority provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing security](#ops_tracing_security) for more details.
```
dimaqq, line 9: **nit:**

```suggestion
Ops uses cryptography is for sending trace data, when a certificate authority provider has been integrated with the charm. This is done using HTTPS, using either TLS 1.2 or 1.3, using the implementation provided by the standard library of the Python that is executing the charm. See [tracing security](#ops_tracing_security) for more details.
```

tonyandrewmeyer: It's important here to make the point that this is not *a* use of cryptography, it is *the only* use of crypto.

docs/explanation/security.md
tonyandrewmeyer: I'm not sure what you mean by this, sorry? I'm not expecting the wider public to review this PR. I don't think I've linked in the doc to anything internal?

docs/explanation/security.md
tonyandrewmeyer: Agreed, but ops.readthedocs is not just the docs for Ops, it's the main "how to develop and maintain charms" documentation point. We have more than just how to use Ops (and ops-scenario, and ops-tracing), we have general information about how you should make charms, linking out to other doc sites where needed. There also isn't really any other doc sites that this could live in.

A couple of these were requested from other reviewers, so I'd rather not remove them now.

docs/explanation/security.md
tonyandrewmeyer: > From a library user PoV, I'm not clear why this page is in Ops docs.

It's very common to need to know what crypto is used, particularly when building software used in the US, especially if it's used by government or military clients. I've regularly had to go find this information (before Canonical, not in a Juju context) - I'm sure the same is true for many people. As someone using charms, I'm not going to know to come looking here, but when the charm developers get asked (or proactively do their own docs) they will expect us to be explaining it.

The same is true for most of the rest: when I'm writing a charm and need to write up a threat model, or I need to fill out a risk register, or I need to make sure I've hardened the deployment, I need to get that information from my dependencies as well as my own code. Ops is obviously one of the major dependencies, and I'll come to the ops docs and search for "security", and find all (hopefully) the info I need.

I do think the good practice stuff could be better done in some small how-tos, but the important thing for the moment is to get a start in our docs. We can iterate and improve as usual.

> Perhaps the target audience are the "the security people"?

Well, we obviously have a requirement to comply with the security policies of Canonical. We're Canonical employees and this is an official Canonical product. So they are *an* audience. But I disagree that they are the only audience that would find value in this doc, or the *only* audience.

> Some links would be useful too, something like "security at canonical" or "security in the Juju ecosystem" or maybe some process links?

There are already links to both Juju and Pebble security docs. The only Canonical ones I can find (that are public) are either about Ubuntu (not entirely unrelated, but seems a stretch to link to) or are very much marketing/sales focused. I'm happy to add more if you have specific suggestions (we already did that with other reviewers), or we can have follow-up PRs for adding more.

PR #1905, opened 2025-07-14
feat: add security event logging

tonyandrewmeyer:
Adds logging of security events, as specified in [SEC0045](https://docs.google.com/document/d/1nInWP9pEEhloKMfgzDsd4Pub4530OxQcKKh716Gvxfk/edit?usp=drivesdk).

A new `_security_event` helper is added to `ops.log`. Security events must be logged as structured data, so the helper does this, and provides some fields common to all events (a timestamp, the model UUID, and a type of "security") as well as the provided event, level, and description.

Logged security events are:
* `sys_startup`, immediately after the logging to Juju is configured.
* `sys_shutdown`, as the very last action of `ops.main`.
* `sys_crash`, if the charm exits because of an uncaught exception.
* `sys_restart`, if the unit's `reboot` command is executed.
* `authz_fail`, if any hook command fails with error text containing "access denied" (case insensitive).
* `user_added` and `user_updated` if the Pebble client is used to add, update, or remove identities.

In order to have the model UUID to log within `ops.pebble`, a new optional `app_id` argument to `pebble.Client` is added, which is used for logging (and the model backend passes in the model UUID for this, in the charm context).

TODO:
- [ ] Tests - the existing tests are updated, which cover startup, shutdown, and crashing, but we should add tests for the other events as well.

See also [OP067 - Security Event Logging for Charm SDK](https://docs.google.com/document/d/1VsxGDxuUOo6A3PfT_qzFssR3qTOJInrHPORitnbtiDk/edit?usp=sharing)

### DIFF
--- a/ops/_main.py
+++ b/ops/_main.py
@@ -31,7 +31,7 @@ from . import model as _model
 from . import storage as _storage
 from ._private import tracer
 from .jujucontext import _JujuContext
-from .log import setup_root_logging
+from .log import _security_event, setup_root_logging
 from .version import version
 
 CHARM_STATE_FILE = '.unit-state.db'
@@ -298,6 +298,12 @@ class _Manager:
         # Do this as early as possible to be sure to catch the most logs.
         self._setup_root_logging()
 
+        _security_event(
+            f'sys_startup:{os.getuid()}',
+            level='DEBUG',
+            description=f'Starting ops framework for {charm_class.__name__}',
+        )
+
         self._charm_root = self._juju_context.charm_dir
         self._charm_meta = self._load_charm_meta()
         self._use_juju_for_storage = use_juju_for_storage
@@ -478,3 +484,8 @@ def main(charm_class: type[_charm.CharmBase], use_juju_for_storage: bool | None 
     finally:
         if manager:
             manager._destroy()
+        _security_event(
+            f'sys_shutdown:{os.getuid()}',
+            level='DEBUG',
+            description='Stopping ops framework',
+        )
--- a/ops/log.py
+++ b/ops/log.py
@@ -16,13 +16,17 @@
 
 from __future__ import annotations
 
+import datetime
+import json
 import logging
+import os
 import sys
 import types
 import typing
 import warnings
 
-from .model import _ModelBackend
+if typing.TYPE_CHECKING:
+    from .model import _ModelBackend
 
 
 class JujuLogHandler(logging.Handler):
@@ -87,5 +91,37 @@ def setup_root_logging(
         if exc_stderr:
             print(f'Uncaught {etype.__name__} in charm code: {value}', file=sys.stderr)
             print('Use `juju debug-log` to see the full traceback.', file=sys.stderr)
+        _security_event(
+            f'sys_crash:{etype.__name__}',
+            level='ERROR',
+            description=f'Uncaught exception in charm code: {value!r}.',
+        )
 
     sys.excepthook = except_hook
+
+
+def _security_event(event: str, *, level: str, description: str):
+    """Send a structured security event log to Juju, as defined by SEC0045.
+
+    Args:
+        event: the name of the event, in the format described by OWASP
+          https://cheatsheetseries.owasp.org/cheatsheets/Logging_Vocabulary_Cheat_Sheet.html
+        level: log level, such as 'DEBUG', 'INFO', or 'ERROR'
+        description: a free-form description of the event, meant for human
+          consumption. Includes additional details of the event that do not
+          fit in the event name.
+    """
+    logger = logging.getLogger(__name__)
+    data: dict[str, typing.Any] = {
+        # This duplicates the timestamp that will be in the Juju log, but is
+        # included so that applications that are pulling out only the structured
+        # data can still see the time of the event.
+        'datetime': datetime.datetime.now(datetime.timezone.utc).isoformat(),
+        # Similarly, this duplicates the level at which this is logged to Juju.
+        'level': level,
+        'type': 'security',
+        'appid': os.environ.get('JUJU_MODEL_UUID', 'unknown'),
+        'event': event,
+        'description': description,
+    }
+    logger.log(getattr(logging, level.upper()), json.dumps(data))
--- a/ops/model.py
+++ b/ops/model.py
@@ -61,6 +61,7 @@ from . import pebble
 from ._private import timeconv, tracer, yaml
 from .jujucontext import _JujuContext
 from .jujuversion import JujuVersion
+from .log import _security_event
 
 if typing.TYPE_CHECKING:
     from typing_extensions import TypeAlias
@@ -3421,6 +3422,19 @@ class _ModelBackend:
             try:
                 result = subprocess.run(args, **kwargs)  # type: ignore
             except subprocess.CalledProcessError as e:
+                if 'access denied' in e.stderr.lower():
+                    if args[0] in ('juju-log', 'action-fail', 'action-set'):
+                        loggable_args = ['<arguments stripped>']
+                    else:
+                        loggable_args = args[1:]
+                    _security_event(
+                        f'authz_fail:{args[0]}',
+                        level='WARNING',
+                        description=f'Hook command {args[0]!r} failed with error: {e.stderr!r}. '
+                        f'The command exited with code: {e.returncode}. '
+                        f'Arguments were: {loggable_args!r}. '
+                        f'Unit {"is" if self.is_leader() else "is not"} leader.',
+                    )
                 raise ModelError(e.stderr) from e
             if return_output:
                 if result.stdout is None:  # type: ignore
@@ -3756,7 +3770,7 @@ class _ModelBackend:
 
     def get_pebble(self, socket_path: str) -> pebble.Client:
         """Create a pebble.Client instance from given socket path."""
-        return pebble.Client(socket_path=socket_path)
+        return pebble.Client(socket_path=socket_path, app_id=self._juju_context.model_uuid)
 
     def planned_units(self) -> int:
         """Count of "planned" units that will run this application.
@@ -3819,6 +3833,13 @@ class _ModelBackend:
             return self._run(*args, return_output=return_output, use_json=use_json)
         except ModelError as e:
             if 'not found' in str(e):
+                _security_event(
+                    f'authz_fail:{args[0]}',
+                    level='WARNING',
+                    description=f'Hook-tool {args[0]!r} failed with error: {e.args[0]!r}. '
+                    f'Arguments were: {args[1:]!r}. '  # This never includes the secret content.
+                    f'Unit {"is" if self.is_leader() else "is not"} leader.',
+                )
                 raise SecretNotFoundError() from e
             raise
 
@@ -3951,6 +3972,11 @@ class _ModelBackend:
         return Port(protocol_lit, int(port))
 
     def reboot(self, now: bool = False):
+        _security_event(
+            f'sys_restart:{os.getuid()}',
+            level='INFO',
+            description=f'Rebooting unit {self.unit_name!r} in model {self.model_name!r}',
+        )
         if now:
             self._run('juju-reboot', '--now')
             # Juju will kill the Charm process, and in testing no code after
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2101,6 +2101,8 @@ class Client:
     connection attempt to Pebble; used by ``urllib.request.OpenerDirector.open``. It's not for
     methods like :meth:`start_services` and :meth:`replan_services` mentioned above, and it's not
     for the command execution timeout defined in method :meth:`Client.exec`.
+
+    The ``app_id`` parameter is used purely for logging purposes, and may be None if not required.
     """
 
     _chunk_size = 8192
@@ -2111,6 +2113,7 @@ class Client:
         opener: urllib.request.OpenerDirector | None = None,
         base_url: str = 'http://localhost',
         timeout: float = 5.0,
+        app_id: str | None = None,
     ):
         if not isinstance(socket_path, str):
             raise TypeError(f'`socket_path` should be a string, not: {type(socket_path)}')
@@ -2120,6 +2123,7 @@ class Client:
         self.opener = opener
         self.base_url = base_url
         self.timeout = timeout
+        self.app_id = app_id
 
     @classmethod
     def _get_default_opener(cls, socket_path: str) -> urllib.request.OpenerDirector:
@@ -2212,6 +2216,20 @@ class Client:
             raise ConnectionError(e.reason) from e
 
         return response
+
+    # This matches the function of the same name in ops.log. It's repeated here,
+    # because ops.pebble doesn't know anything about ops.
+    def _security_event(self, event: str, *, level: str, description: str):
+        """Send a structured security event log, as defined by SEC0045."""
+        data = {
+            'datetime': datetime.datetime.now(datetime.timezone.utc).isoformat(),
+            'type': 'security',
+            'appid': self.app_id or 'pebble',
+            'level': level,
+            'event': event,
+            'description': description,
+        }
+        logger.log(getattr(logging, level.upper()), json.dumps(data))
 
     def get_system_info(self) -> SystemInfo:
         """Get system info."""
@@ -3404,6 +3422,32 @@ class Client:
 
             body = {'action': 'replace', 'identities': identities_dict}
             self._request('POST', '/v1/identities', body=body)
+            for username, identity in identities_dict.items():
+                if identity is None:
+                    self._security_event(
+                        f'user_updated:{os.getuid()},{username}',
+                        level='INFO',
+                        description=f'Removing identity {username!r} from Pebble.',
+                    )
+                    continue
+                if 'local' in identity:
+                    user_id = identity['local'].get('user-id', 'unknown')
+                else:
+                    user_id = 'unknown'
+                if identity.get('basic') and identity.get('local'):
+                    identity_type = 'basic,local'
+                elif identity.get('basic'):
+                    identity_type = 'basic'
+                elif identity.get('local'):
+                    identity_type = 'local'
+                else:
+                    identity_type = 'unknown'
+                self._security_event(
+                    f'user_added:{os.getuid()},{username},'
+                    f'attributes[{identity["access"]}, {identity_type}, {user_id}]',
+                    level='INFO',
+                    description=f'Added or updated identity {username!r} in Pebble.',
+                )
 
     def remove_identities(self, identities: Iterable[str]) -> None:
         """Remove the named identities in Pebble.
@@ -3417,6 +3461,12 @@ class Client:
             identities_dict = {name: None for name in identities}
             body = {'action': 'remove', 'identities': identities_dict}
             self._request('POST', '/v1/identities', body=body)
+            for username in identities:
+                self._security_event(
+                    f'user_updated:{os.getuid()},{username}',
+                    level='INFO',
+                    description=f'Removing identity {username!r} from Pebble.',
+                )
 
 
 class _FilesParser:
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -18,6 +18,7 @@ import abc
 import dataclasses
 import importlib.util
 import io
+import json
 import logging
 import os
 import re
@@ -28,7 +29,7 @@ import tempfile
 import typing
 import warnings
 from pathlib import Path
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 
@@ -68,6 +69,7 @@ class EventSpec:
     remote_app: str | None = None
     remote_unit: str | None = None
     model_name: str | None = None
+    model_uuid: str | None = None
     set_in_env: dict[str, str] | None = None
     workload_name: str | None = None
     notice_id: str | None = None
@@ -414,6 +416,8 @@ class _TestMain(abc.ABC):
             event_dir = 'hooks'
         if event_spec.model_name is not None:
             env['JUJU_MODEL_NAME'] = event_spec.model_name
+        if event_spec.model_uuid is not None:
+            env['JUJU_MODEL_UUID'] = event_spec.model_uuid
 
         self._call_event(fake_script, Path(event_dir, event_filename), env)
         return self._read_and_clear_state(event_spec.event_name, env)
@@ -759,6 +763,9 @@ class _TestMain(abc.ABC):
             ['is-leader', '--format=json'],
         ]
         calls = fake_script.calls()
+        # Remove the security events.
+        calls.pop(1)
+        calls.pop(-1)
 
         assert calls == expected
 
@@ -775,6 +782,9 @@ class _TestMain(abc.ABC):
         )
 
         calls = fake_script.calls()
+        # Remove the security events.
+        calls.pop(1)
+        calls.pop(-1)
 
         custom_event_prefix = 'Emitting custom event <CustomEvent via Charm/on/custom'
         expected = [
@@ -862,15 +872,25 @@ class _TestMain(abc.ABC):
         if not re.match(pattern, warning_message):
             pytest.fail(f'Warning was not sent to debug-log: {calls!r}')
 
+    @patch('os.getuid', return_value=1000)
     @pytest.mark.usefixtures('setup_charm')
-    def test_excepthook(self, fake_script: FakeScript):
+    def test_excepthook(self, _: MagicMock, fake_script: FakeScript):
         with pytest.raises(subprocess.CalledProcessError):
             self._simulate_event(
                 fake_script,
-                EventSpec(ops.InstallEvent, 'install', set_in_env={'TRY_EXCEPTHOOK': '1'}),
+                EventSpec(
+                    ops.InstallEvent,
+                    'install',
+                    set_in_env={'TRY_EXCEPTHOOK': '1'},
+                    model_uuid='1234',
+                ),
             )
 
-        calls = [' '.join(i) for i in fake_script.calls()]
+        calls = fake_script.calls()
+        sec_start = calls.pop(1)
+        sec_crash = calls.pop(-1)
+        sec_end = calls.pop(-2)
+        calls = [' '.join(i) for i in calls]
 
         assert calls.pop(0) == ' '.join(VERSION_LOGLINE)
         assert 'Using local storage: not a Kubernetes podspec charm' in calls.pop(0)
@@ -884,6 +904,25 @@ class _TestMain(abc.ABC):
             calls[0],
         )
         assert len(calls) == 1, f'expected 1 call, but got extra: {calls[1:]}'
+
+        assert sec_start[:-1] == sec_end[:-1] == ['juju-log', '--log-level', 'DEBUG', '--']
+        assert sec_crash[:-1] == ['juju-log', '--log-level', 'ERROR', '--']
+        data_start = json.loads(sec_start[-1])
+        data_crash = json.loads(sec_crash[-1])
+        data_end = json.loads(sec_end[-1])
+        for data in (data_start, data_crash, data_end):
+            assert data['type'] == 'security'
+            assert data['appid'] == '1234'
+            assert 'datetime' in data
+        assert data_start['description'] == 'Starting ops framework for Charm'
+        assert data_start['event'] == 'sys_startup:1000'
+        assert (
+            data_crash['description']
+            == "Uncaught exception in charm code: RuntimeError('failing as requested')."
+        )
+        assert data_crash['event'] == 'sys_crash:RuntimeError'
+        assert data_end['description'] == 'Stopping ops framework'
+        assert data_end['event'] == 'sys_shutdown:1000'
 
     @pytest.mark.usefixtures('setup_charm')
     def test_sets_model_name(self, fake_script: FakeScript):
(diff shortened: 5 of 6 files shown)


ops/_main.py
```diff
@@ -478,3 +484,8 @@ def main(charm_class: type[_charm.CharmBase], use_juju_for_storage: bool | None
     finally:
         if manager:
             manager._destroy()
+        _security_event(
+            f'sys_shutdown:{os.getuid()}',
+            level='DEBUG',
+            description='Stopping ops framework',
+        )
```
benhoyt: If I understand correctly, this will add two new logs to the debug log for every hook. That will add a lot of noise to `juju debug-log` when diagnosing issues. Is that really what we want/need?

ops/_main.py
```diff
@@ -298,6 +298,12 @@ def __init__(
         # Do this as early as possible to be sure to catch the most logs.
         self._setup_root_logging()
 
+        _security_event(
+            f'sys_startup:{os.getuid()}',
```
benhoyt: I haven't read the spec, but it looks like this is an f-string with two `:`-separated components for every call to `_security_event`. Should it be two parameters, `type: str` and `data: typing.Any | None = None` or something? (And the data is converted using `str()`.)

```python
_security_event(
    'sys_startup',
    os.getuid(),
    level='DEBUG',
    description='...'),
)
```

ops/log.py
```diff
@@ -87,5 +91,37 @@ def except_hook(etype: type[BaseException], value: BaseException, tb: types.Trac
...
+        _security_event(
+            f'sys_crash:{etype.__name__}',
+            level='ERROR',
+            description=f'Uncaught exception in charm code: {value!r}.',
+        )
 
     sys.excepthook = except_hook
+
+
+def _security_event(event: str, *, level: str, description: str):
```
benhoyt: What about putting `level` first like `logger.log()` does?

Is *description* required?

ops/model.py
```diff
@@ -3421,6 +3422,19 @@ def _run(
             try:
                 result = subprocess.run(args, **kwargs)  # type: ignore
             except subprocess.CalledProcessError as e:
+                if 'access denied' in e.stderr.lower():
+                    if args[0] in ('juju-log', 'action-fail', 'action-set'):
```
benhoyt: Can you clarify why the args are stripped for these specific commands?

ops/model.py
```diff
@@ -3421,6 +3422,19 @@ def _run(
             try:
                 result = subprocess.run(args, **kwargs)  # type: ignore
             except subprocess.CalledProcessError as e:
+                if 'access denied' in e.stderr.lower():
```
benhoyt: Are we sure this is the only text from Juju that represents an auth failure?

ops/model.py
```diff
@@ -3756,7 +3770,7 @@ def add_metrics(
 
     def get_pebble(self, socket_path: str) -> pebble.Client:
         """Create a pebble.Client instance from given socket path."""
-        return pebble.Client(socket_path=socket_path)
+        return pebble.Client(socket_path=socket_path, app_id=self._juju_context.model_uuid)
```
benhoyt: This is called "app_id", but it's the model UUID. Is that intended?

ops/model.py
```diff
@@ -3819,6 +3833,13 @@ def _run_for_secret(
             return self._run(*args, return_output=return_output, use_json=use_json)
         except ModelError as e:
             if 'not found' in str(e):
+                _security_event(
```
benhoyt: Is "not found" an auth failure?

ops/log.py
```diff
@@ -87,5 +91,37 @@ def except_hook(etype: type[BaseException], value: BaseException, tb: types.Trac
...
+    logger = logging.getLogger(__name__)
+    data: dict[str, typing.Any] = {
+        # This duplicates the timestamp that will be in the Juju log, but is
+        # included so that applications that are pulling out only the structured
+        # data can still see the time of the event.
+        'datetime': datetime.datetime.now(datetime.timezone.utc).isoformat(),
+        # Similarly, this duplicates the level at which this is logged to Juju.
+        'level': level,
+        'type': 'security',
+        'appid': os.environ.get('JUJU_MODEL_UUID', 'unknown'),
```
benhoyt: I think you pointed this out when we talked, but this is called "appid" but is actually the model UUID. Should we use something else?

Also: it's unfortunate we have to use the environment variable here, but I guess there's not really a cleaner way.

Could we have `app_id` be an optional parameter, and if not set, it uses the environment variable? (See comment below about Pebble logging too.)

ops/log.py
```diff
@@ -87,5 +91,37 @@ def except_hook(etype: type[BaseException], value: BaseException, tb: types.Trac
...
+        'event': event,
+        'description': description,
+    }
+    logger.log(getattr(logging, level.upper()), json.dumps(data))
```
benhoyt: Do we need to write `logger.log(getattr(...), '%s', json.dumps(data))` here, to avoid `%`s in the data causing trouble with log formatting?

ops/pebble.py
```diff
@@ -2213,6 +2217,20 @@ def _request_raw(
 
         return response
 
+    # This matches the function of the same name in ops.log. It's repeated here,
+    # because ops.pebble doesn't know anything about ops.
+    def _security_event(self, event: str, *, level: str, description: str):
```
benhoyt: I wonder if it'd be better to pass in a logging callable to `Client()`. Like `Client(log_security_event=log._security_event)` or `Client()` to mean no logging (the current default). Otherwise we're adding a bunch of logging to people who are importing Ops just to use the pebble Client, which is not exactly a breaking change, but probably annoying.

ops/pebble.py
```diff
@@ -3404,6 +3422,32 @@ def replace_identities(self, identities: Mapping[str, IdentityDict | Identity |
 
             body = {'action': 'replace', 'identities': identities_dict}
             self._request('POST', '/v1/identities', body=body)
+            for username, identity in identities_dict.items():
+                if identity is None:
+                    self._security_event(
+                        f'user_updated:{os.getuid()},{username}',
```
benhoyt: Is the getuid UID needed here?

ops/pebble.py
```diff
@@ -3404,6 +3422,32 @@ def replace_identities(self, identities: Mapping[str, IdentityDict | Identity |
...
+                    identity_type = 'basic,local'
+                elif identity.get('basic'):
+                    identity_type = 'basic'
+                elif identity.get('local'):
+                    identity_type = 'local'
+                else:
+                    identity_type = 'unknown'
+                self._security_event(
+                    f'user_added:{os.getuid()},{username},'
+                    f'attributes[{identity["access"]}, {identity_type}, {user_id}]',
```
benhoyt: This reads rather awkwardly on two lines (looks like two args). Can we adjust? Maybe build it in a previous statement or similar.

ops/pebble.py
```diff
@@ -3404,6 +3422,32 @@ def replace_identities(self, identities: Mapping[str, IdentityDict | Identity |
...
+            for username, identity in identities_dict.items():
+                if identity is None:
+                    self._security_event(
+                        f'user_updated:{os.getuid()},{username}',
+                        level='INFO',
+                        description=f'Removing identity {username!r} from Pebble.',
+                    )
+                    continue
+                if 'local' in identity:
+                    user_id = identity['local'].get('user-id', 'unknown')
```
benhoyt: It's pretty unfortunate we have to do this kind of parsing of the inner objects. Can we simplify and just log the full object in the description? Or even just leave out the full object altogether? (And are there security concerns with logging the full object if there's a password hash involved?)

test/test_main.py
```diff
@@ -885,6 +905,25 @@ def test_excepthook(self, fake_script: FakeScript):
         )
         assert len(calls) == 1, f'expected 1 call, but got extra: {calls[1:]}'
 
+        assert sec_start[:-1] == sec_end[:-1] == ['juju-log', '--log-level', 'DEBUG', '--']
```
benhoyt: This is certainly fiddly, though I'm not sure I see an obviously better way.

benhoyt (commented):
Adding a bunch of questions.

ops/_main.py
tonyandrewmeyer: It's not in the spec, but the guidance is to follow OWASP, and that is the [format](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Vocabulary_Cheat_Sheet.html#the-vocabulary) they have. I like this suggestion, and if we do end up exposing this for charmers it's definitely worth doing. Will adjust.

ops/model.py
tonyandrewmeyer: No, I noticed that yesterday as well (in the spec). Will fix.

ops/model.py
tonyandrewmeyer: They all might contain sensitive data, so we should be cautious and not include the arguments. I've added a comment.

ops/log.py
tonyandrewmeyer: We shouldn't have to - you get the plain `%s` or whatever in the output if there are no args. But it does seem like it would be tidier, thanks!

ops/log.py
tonyandrewmeyer: Yeah, I've fixed this so that it's a combination of the model UUID and the unit name, which I think is probably the best option we have.

I agree about having to use the environment variable. I'd rather get it from JujuContext, but don't want to have that loaded in `log.py` and don't want people to have to pass it in. In some ways it would be better if it was stamped in when the framework starts, so that if charm code does `os.environ['JUJU_UNIT_NAME'] = ''` or something then that doesn't impact this, but since charms can execute code before `ops.main()` I don't think that really gains anything.

I'll think about the Pebble case and comment in that thread -- I assume that would be the only reason for allowing passing in an app ID.

ops/_main.py
tonyandrewmeyer: It's my understanding of what's needed. I'll [double check with the security team](https://chat.canonical.com/canonical/pl/dnh59qp6jprxdgtkytfdk9dzey).

ops/log.py
benhoyt: Ah, didn't realise about no args -- makes sense, thanks.

ops/model.py
tonyandrewmeyer: It [can be](https://github.com/canonical/operator/issues/1229#issuecomment-2148651925), yes.

ops/model.py
tonyandrewmeyer: I think so, but I'll verify before moving this out of draft, yes.

tonyandrewmeyer: I'll [summarise in the additional information part of the spec](https://docs.google.com/document/d/1VsxGDxuUOo6A3PfT_qzFssR3qTOJInrHPORitnbtiDk/edit?tab=t.0#heading=h.u6qxls6ui4zr).

ops/log.py
tonyandrewmeyer: > What about putting `level` first like `logger.log()` does?

I wondered about that too. It felt like the event was the primary bit of data here "log security event X" with the level and description additional information. I'm not sure that it's obvious that it's `log_security_event(level, event)` rather than `log_security_event(event, level)` and with both being strings static type checks wouldn't trigger errors if you get it the wrong way around.

Now that I've split `event` into two, it's probably fine to switch them around. I could make `level` a `Literal` rather than `str` too, since the chances of a new level being used seem tiny (and you can type: ignore if needed).

> Is _description_ required?

Every example in the OWASP list has a description, but it doesn't say anywhere that it's always present. The spec is silent on it other than deferring to the OWASP page.

I feel like we should encourage having a human-readable description for the event. I'd prefer to leave it as required now - if this remains as internal use only then we should always put in a description, so it should be required, and if we open it up to be public then I think we should start as required and could loosen that if charmers come with a compelling story.

ops/_main.py
tonyandrewmeyer: I do feel like the second component ought to be passed as a string rather than having the method do it, though. When this comes out of draft you can see whether you agree or not :)

ops/pebble.py
tonyandrewmeyer: The [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Vocabulary_Cheat_Sheet.html#user_createduseridnewuseridattributesonetwothree) docs have the user doing the adding as well as the user being added. `os.getuid()` seemed like a reasonable choice for the user doing the adding, without going to the effort of `getpwnam`.

In the charm context, this would presumably always be the user that's running the charm code. If using ops.pebble outside of a charm, it could be anyone, of course.

ops/pebble.py
tonyandrewmeyer: Probably annoying, but probably also by design, I think. We aren't adding security event logging only specifically in the charm context, we're meant to be adding it to our products, and I'd argue that even though ops.pebble lives in the ops package, it's a product on its own and so is still meant to produce security logging when used in that way.

Reproducing a version of the method in ops.pebble is annoying, though.

What if we passed in a callable, but instead of `None` meaning "no logging" it meant log to `logging.getLogger('pebble.security')`. If you really want it off you could then adjust that logger, but you'd get messages going up to the root logger by default (if you're using ops.pebble in isolation, you're already responsible for configuring the logging if you want output).

It's a bit more knowledge to pass in - a callable with a specific signature rather than an arbitrary string - but that seems ok.

I don't believe that adding additional logging messages can be considered a breaking change. *Maybe* something to avoid in a bugfix release, but completely fine in a minor release.

ops/pebble.py
tonyandrewmeyer: Having the hash in the log doesn't seem ideal, even though it ought to be safe. We could leave out the type and (where relevant) the user ID, but it does seem like that information would be useful if you are tracing some unexpected/bad activity.

What do you think about changing `BasicIdentity`'s `__str__` to always mask the password? `__repr__` would stay as it is, with the real value. If it's a `BasicIdentity` coming _from_ Pebble then there's no change, so it would only be for identities that the user created themselves. I think this would probably be good in general, to avoid accidentally outputting the hash.

If we did that then I think it would be reasonable to log the whole object.

benhoyt: Ah yeah, masking out the password in BasicIdentity's `__str__` is a good idea.

ops/log.py
```diff
@@ -87,5 +99,220 @@ def except_hook(etype: type[BaseException], value: BaseException, tb: types.Trac
...
+            'WARN',
+            _SecurityEventSystem.SYS_CRASH,
+            etype.__name__,
+            description=f'Uncaught exception in charm code: {value!r}.',
+        )
 
     sys.excepthook = except_hook
+
+
+class _SecurityEventAuthN(enum.Enum):
```
tonyandrewmeyer: If we keep the event logging functionality as internal use only, then we definitely want to strip out all of these that we don't use (most of them).

If we do expose `log_security_event` publicly for charms to use, then I think it is helpful to have these. I don't think we want to have them in the top-level `ops` namespace - maybe they could be used from `ops.log`? We could also cut down the list to the ones that the security spec requires, rather than the full OWASP set.

We could also put them all in one (very large) enum, which means there's only a single name to export, and then it probably could be in the top-level `ops` namespace. If we *don't* do that, then the prefixes, like `SYS_` can probably be removed, since they are duplicating information in the enum class name.

ops/log.py
```diff
@@ -87,5 +99,220 @@ def except_hook(etype: type[BaseException], value: BaseException, tb: types.Trac
...
+    _SecurityEventPrivilege,
+    _SecurityEventSensitive,
+    _SecurityEventSequence,
+    _SecurityEventSession,
+    _SecurityEventSystem,
+    _SecurityEventUser,
+]
+
+
+def _log_security_event(
```
tonyandrewmeyer, line 154: TBD if this ends up being `log_security_event` and added to the `ops` namespace. So far, I have one vote in favour and silence otherwise.

ops/log.py
```diff
@@ -89,5 +93,234 @@ def except_hook(
...
+          consumption. Includes additional details of the event that do not
+          fit in the event name.
+    """
+    logger = logging.getLogger(__name__)
+    for juju_handler in logger.handlers:
+        if isinstance(juju_handler, JujuLogHandler):
+            model_backend = juju_handler.model_backend
+            juju_context = model_backend._juju_context
+            app_id = f'{juju_context.model_uuid}-{juju_context.unit_name}'
+            juju_log = model_backend.juju_log
```
tonyandrewmeyer: This avoids having to get things from the environment and means we can directly use the hook command rather than go through the logger, which is the better approach I think.

ops/pebble.py
```diff
@@ -2073,6 +2073,9 @@ def to_dict(self) -> BasicIdentityDict:
         """Convert this basic identity to its dict representation."""
         return {'password': self.password}
 
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}(password='*****')"
+
```
tonyandrewmeyer, line 2078: Not required now that we aren't logging identity events, but I think a good change anyway.

ops/pebble.py
tonyandrewmeyer: Not relevant any more, since there's no logging in ops.pebble any more.

IronCore864 (approved):

ops/log.py
```diff
@@ -89,5 +93,77 @@ def except_hook(
...
+    SYS_RESTART = 'sys_restart'
+    SYS_CRASH = 'sys_crash'
+    SYS_MONITOR_DISABLED = 'sys_monitor_disabled'
+
+
+def _log_security_event(
+    # These are the OWASP log levels, which are not the same as the Juju or Python log levels.
+    level: typing.Literal['INFO', 'WARN', 'CRITICAL'],
+    event_type: _SecurityEvent | str,
+    event: str,
```
benhoyt: The name "event_type" makes sense, but "event" not so much -- this is actually the "data" or "argument", right? [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Vocabulary_Cheat_Sheet.html) calls it "additional data". So maybe `event_data` for this second arg?

ops/log.py
```diff
@@ -89,5 +93,77 @@ def except_hook(
...
+
+    AUTHZ_FAIL = 'authz_fail'
+    SYS_RESTART = 'sys_restart'
+    SYS_CRASH = 'sys_crash'
+    SYS_MONITOR_DISABLED = 'sys_monitor_disabled'
+
+
+def _log_security_event(
+    # These are the OWASP log levels, which are not the same as the Juju or Python log levels.
+    level: typing.Literal['INFO', 'WARN', 'CRITICAL'],
```
benhoyt: Feels a bit weird that this is a Literal but the next one is an `enum | str`. Maybe `level: _SecurityLevel` with `_SecurityLevel` being an enum?

ops/log.py
```diff
@@ -89,5 +93,77 @@ def except_hook(
...
+    Args:
+        level: log level of the security event (this is not the same as the Juju log level)
+        event_type: the event type, in the format described by OWASP
+          https://cheatsheetseries.owasp.org/cheatsheets/Logging_Vocabulary
+        event: the name of the event, in the format described by OWASP
+        description: a free-form description of the event, meant for human
+          consumption. Includes additional details of the event that do not
+          fit in the event name.
+    """
+    logger = logging.getLogger()
```
benhoyt: Could this whole block (getting `app_id` and `juju_log`) be factored out into a function which is `functools.cache`d?

ops/log.py
```diff
@@ -89,5 +93,77 @@ def except_hook(
...
+            'JujuLogHandler is not set up for the logger. '
+            'Call setup_root_logging() before logging security events.',
+            RuntimeWarning,
+        )
+        app_id = 'charm'
+
+        def juju_log(level: str, message: str):
+            logger.debug(message)
+
+    type = event_type if isinstance(event_type, str) else event_type.value
```
benhoyt, line 173: This dance is what makes Python enums annoying. Can we make the enum(s) inherit from `(enum.Enum, str)` instead?

ops/log.py
benhoyt: Yeah, I guess let's start with it private. Easy enough to make it public when charmers start needing it / asking for it.

ops/model.py
```diff
@@ -433,6 +434,12 @@ def status(self) -> StatusBase:
             return UnknownStatus()
 
         if not self._backend.is_leader():
+            _log_security_event(
+                'CRITICAL',
+                _SecurityEvent.AUTHZ_FAIL,
+                'status-get',
+                description='Attempted to get application status when not leader.',
```
benhoyt: To me the full stop `.` seems extraneous. We're not writing an essay or even sentences here -- it's a short, human-readable description phrase. What do you think?

ops/model.py
```diff
@@ -2784,7 +2797,18 @@ def stop_checks(self, *check_names: str) -> list[str]:
...
-        return self._pebble.stop_checks(check_names)
+        stopped_checks = self._pebble.stop_checks(check_names)
+        description = (
+            f'Asked to stop checks {",".join(check_names)} - '
+            f'stopped checks {",".join(stopped_checks)}'
+        )
+        _log_security_event(
+            'WARN',
+            _SecurityEvent.SYS_MONITOR_DISABLED,
+            f'{os.getuid()},{",".join(check_names)}',
```
benhoyt: Couple of things:
- I'm logging this in Pebble, so should we just not log it here?
- It feels weird to comma-separate the names inside an already-comma-separated field. If we do keep this, what about just logging once per (actually) stopped check?

ops/model.py
```diff
@@ -3636,6 +3660,32 @@ def _run(
             try:
                 result = subprocess.run(args, **kwargs)  # type: ignore
             except subprocess.CalledProcessError as e:
+                authz_messages = (
```
benhoyt: This is starting to get quite cluttered. Worth refactoring this into a `_log_security_error` helper method?

ops/model.py
```diff
@@ -3636,6 +3660,32 @@ def _run(
...
+                )
+                if any(message in e.stderr.lower() for message in authz_messages):
+                    # These commands may have sensitive data in their arguments, and do not
+                    # support reading the data from a file.
+                    if args[0] in ('juju-log', 'action-fail', 'action-set'):
+                        log_args = f'Arguments were: {args[1:]!r}. '
+                    else:
+                        log_args = ''
+                    base_cmd = os.path.basename(args[0])
+                    description = (
```
benhoyt: Do we really need the "Unit is (not) leader" in the description? (We could add any number of things.) How about terse-ifying this a bit, maybe `Hook command {base_cmd!r} failed with code {e.return_code}: {e.stderr.strip()!r}`

test/test_main.py
```diff
@@ -929,8 +939,20 @@ def test_excepthook(self, fake_script: FakeScript):
...
             calls[0],
         )
+        sec_crash = calls.pop(1)
         assert len(calls) == 1, f'expected 1 call, but got extra: {calls[1:]}'
 
+        assert sec_crash.startswith('juju-log --log-level TRACE --')
+        data_crash = json.loads(sec_crash.rsplit('--', 1)[-1])
+        assert data_crash['type'] == 'security'
+        assert data_crash['appid'] == '1234-test_main/0'
+        assert 'datetime' in data_crash
```
benhoyt: Maybe tighten this slightly to a regex match to a ISO UTC date?

test/test_model.py
```diff
@@ -2021,6 +2044,35 @@ def test_get_check(self, container: ops.Container):
...
+        container.stop_checks('c1', 'c2')
+        assert container.pebble.requests == [('stop_checks', ('c1', 'c2'))]  # type: ignore
+        calls = fake_script.calls(clear=True)
+        sec_log = calls.pop(0)
+        assert sec_log[:-1] == ['juju-log', '--log-level', 'TRACE', '--']
+        data = json.loads(sec_log[-1])
+        assert data['level'] == 'WARN'
+        assert data['type'] == 'security'
+        assert data['appid'] == '1234-myapp/0'
+        assert data['event'] == 'sys_monitor_disabled:1001,c1,c2'
```
benhoyt: 1001 is the current UID, right? Can we test against `os.getuid()` instead?

benhoyt (commented):
Looks good overall, but a few comments/suggestions.

ops/model.py
tonyandrewmeyer: I feel we are writing sentences (not essays). But I'm fine dropping it.

```suggestion
                description='Attempted to get application status when not leader',
```

ops/model.py
tonyandrewmeyer: > * I'm logging this in Pebble, so should we just not log it here?

In this case, I feel that there's value in having the information at multiple layers. If I'm doing incident response and I've seen that checks were disabled, then seeing this immediately afterwards tells me that it was most likely done in the charm code. It's a strong hint towards a legitimate stop-check and the lack of it is a strong hint towards the stop-check being done manually or at least outside of the charm code. It's definitely not perfect - `juju-log` is available to anything that can run processes in the unit, so you can throw fake security event logs in, and the charm code could be doing this directly on the `Client` methods. But I feel like it would be helpful information.

In general, when I've had to dig into suspected (and actual) security issues, having information at multiple layers has been really useful, particularly when the underlying cause has been bugs rather than malicious activity. I removed the duplication where it didn't seem to add much value, but I feel like this one does (if we do eventually add identity management at the `Container` level, I'd be tempted to add logs there too).

> * It feels weird to comma-separate the names inside an already-comma-separated field. If we do keep this, what about just logging once per (actually) stopped check?

Sure.

ops/model.py
tonyandrewmeyer: I had the "unit is (not) leader" part there because roughly half of the authz failures are because due to a non-leader trying to do something that needs the leader. It seemed simpler to always include it rather than include it only if the error message indicated a leadership issue.

I'm happy to make this more terse if you'd prefer that. I do think leadership should stay, but it can be `leader=x` terse. I'm ok dropping the args, except I think the difference between `secret-get` with and without `--reveal` is significant, so I'd special case that one.

test/test_model.py
tonyandrewmeyer: > 1001 is the current UID, right? Can we test against `os.getuid()` instead?

It's the mocked UID from the first line of the test. Testing against the mocked `getuid()` doesn't seem like the right thing to do to me. I can make it a variable used in both the mock and the test (I do generally like that, but it doesn't seem like it's the general pattern with ops -- we usually just duplicate everything).

ops/log.py
tonyandrewmeyer: Would you prefer that, or to do it like in Pebble, where there's `_security_event_warn`, `_security_event_info`, `_security_event_critical` wrappers? That would match `logger.info` and the like more closely.

ops/model.py
tonyandrewmeyer: > I'm ok dropping the args, except I think the difference between `secret-get` with and without `--reveal` is significant, so I'd special case that one.

Oops, I was getting mixed up with the CLI. `secret-get` is always revealing :)

ops/model.py
benhoyt: Yep, fair, thanks.

ops/model.py
benhoyt: Okay, fine to leave leader in, but I would like it more terse: information density is usually a bonus when reading logs (even/especially for humans).

test/test_model.py
benhoyt: Ah, got it, thanks.

tonyandrewmeyer: Or I guess I could remove the mocking and assert against the real `os.getuid()`. Is that what you were suggesting?

benhoyt: Yeah, asserting against `os.getuid()` what I was suggesting (sorry if unclear), but I'm fine with the current approach now that I realise it's a mocked number.

ops/log.py
tonyandrewmeyer: I've gone with making it an enum. If we do expose this for charms, then I'm not convinced we'd want to add three more names.

ops/log.py
tonyandrewmeyer: We could, but:

* That doesn't do anything for us, because the `str()` is "_SecurityEvent.X" so we have to post-process it anyway (unless we do a custom `__str__`). Inheriting from `str` means you can compare against strings, but we're not doing that anywhere.
* The only use case I can think of for these outside passing them as arguments to `_log_security_event` is if we expose the method to charms and they have tests that assert that the right event and level were passed - but even there I'd argue that they should test against the enum, not a plain string. We're not returning one of these to any code, so there are only two places where we have to do the `isinstance` check, and both are here in this method.

So it doesn't seem worth it to me.

tonyandrewmeyer:
In `debug-log`, this looks like:

```
unit-uptime-13: 13:24:03 TRACE unit.uptime/13.juju-log {"datetime": "2025-08-01T01:24:03.969173+00:00", "level": "WARN", "type": "security", "appid": "30c372a0-cf16-4602-8a17-112925b159ae-uptime/13", "event": "sys_crash:NameError", "description": "Uncaught exception in charm code: NameError(\"name 'crash' is not defined\")."}
```

ops/log.py
benhoyt: Yeah, fair. Okay, this is good for now. We can consider further what we want if/when we publicly expose this.

ops/log.py
benhoyt: Ah, fair enough. Okay, we'll leave as is, and consider further what we want if/when we make this public.

ops/model.py
```diff
@@ -2798,16 +2798,17 @@ def stop_checks(self, *check_names: str) -> list[str]:
...
-            'WARN',
-            _SecurityEvent.SYS_MONITOR_DISABLED,
-            f'{os.getuid()},{",".join(check_names)}',
-            description=description,
-        )
+        for check in check_names:
+            if check in stopped_checks:
+                description = f'Stopped check {check}'
+            else:
+                description = f'Failed to stop check {check}'
```
benhoyt: Note that it may not be a failure, but just that the check is already stopped. As such, I wouldn't log anything in this case.

benhoyt (approved):

test/test_model.py
```diff
@@ -1177,7 +1201,8 @@ def test_juju_version_from_model(self):
         assert model.juju_version == version
         assert isinstance(model.juju_version, ops.JujuVersion)
         # Make sure it's not being loaded from the environment.
-        assert JujuVersion.from_environ() == '0.0.0'
+        with pytest.warns(DeprecationWarning):
```
tonyandrewmeyer, line 1206: Drive-by to get the test output a little quieter.

PR #1906, opened 2025-07-15
ci: add SBOM generation and secscan workflow

tonyandrewmeyer:
This repo now has too many artifacts (3 PyPI packages each in two different formats each with two maintained versions) to manually request the SBOM generation and secscan. This PR adds configuration files for the [sbomber](https://github.com/canonical/sbomber) tool to automate those processes.

The tool can be run manually against the configuration files, but there is also a workflow that will do this automatically. It is currently set to be manually triggered, but the intention is that once it is verified to work it will also be added to the publish workflow, to completely automate the process.

The secscan tool can only be run in the Canonical network, so the workflow is set to use a hosted runner. This prevents testing the workflow in my fork.

### DIFF
--- a/.github/workflows/sbom-secscan.yaml
+++ b/.github/workflows/sbom-secscan.yaml
@@ -23,7 +23,7 @@ jobs:
           - name: Checkout security scanner
             uses: actions/checkout@v4
             with:
-              repository: tonyandrewmeyer/sbomber
+              repository: canonical/sbomber
               path: scanner
               token: ${{ secrets.SBOMBER_TOKEN }}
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Approving, with the understanding that this will automatically update the versions once our release automation is in place.

tonyandrewmeyer:
> Approving, with the understanding that this will automatically update the versions once our release automation is in place.

I promise to do a follow-up PR, assuming this all works as expected. And if it doesn't, then a different sort of follow-up PR, I guess :laughing:.

PR #1907, opened 2025-07-15
refactor: move the testing event capture into a framework subclass

tonyandrewmeyer:
This extracts the event-capturing `Framework` subclass out of #1631 (merged then reverted).

This replaces the context handler that monkeypatches `Framework` to capture events with a subclass of `Framework` that provides the same functionality. This is a simpler implementation that can't leak across tests and is thread-safe.

### DIFF
--- a/testing/src/scenario/_ops_main_mock.py
+++ b/testing/src/scenario/_ops_main_mock.py
@@ -12,7 +12,9 @@ import warnings
 from typing import (
     TYPE_CHECKING,
     Any,
+    Generic,
     Sequence,
+    cast,
 )
 
 import ops
@@ -23,6 +25,7 @@ from ops.framework import _event_regex
 from ops._main import _Dispatcher, _Manager
 from ops._main import logger as ops_logger
 
+from .context import Context
 from .errors import BadOwnerPath, NoObserverError
 from .logger import logger as scenario_logger
 from .mocking import _MockModelBackend
@@ -111,7 +114,7 @@ class UnitStateDB:
             db.save_snapshot(stored_state._handle_path, stored_state.content)
 
 
-class Ops(_Manager):
+class Ops(_Manager, Generic[CharmType]):
     """Class to manage stepping through ops setup, event emission and framework commit."""
 
     def __init__(
@@ -127,6 +130,7 @@ class Ops(_Manager):
         self.context = context
         self.charm_spec = charm_spec
         self.store = None
+        self.captured_events: list[ops.EventBase] = []
         self.trace_data = []
 
         try:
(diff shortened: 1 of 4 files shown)


testing/src/scenario/_ops_main_mock.py
```diff
@@ -163,6 +167,34 @@ def _load_charm_meta(self):
...
+            self._charm_root,
+            self._charm_meta,
+            model,
+            event_name=dispatcher.event_name,
+            juju_debug_at=self._juju_context.debug_at,
+            context=self.context,
+        )
+        framework.set_breakpointhook()
+        return framework
+
```
james-garner-canonical: Wdyt about a little refactor on the `ops._main._Manager` side to eliminate the need to duplicate all this? I reckon the way to go might be adding `ops._main._Manager._make_model`.

tonyandrewmeyer: Yeah, #1631 had a similar approach, adding:

```python
def _make_framework(self, *args: Any, **kwargs: Any):
        # A wrapper so that the testing subclasses can easily override which
        # framework class is created.
        return _framework.Framework(*args, **kwargs)
```

And the rest of the code was elsewhere and re-used.

For this PR I was trying to keep the changes more minimal so just duplicated it. Do you think it would be better to be a bit more invasive to avoid the duplication?

james-garner-canonical: Factoring out the framework bit specifically into an overridable method gets rid of all the duplicated code, which is very nice. I would like to see something like that in this PR.

---

It's a real shame that Python's typing machinery doesn't do a good job supporting idomatic use of `*args, **kwargs`, but I think the right thing to do here is to just use them and not worry too much about the typing.

---

Afaict there's no nice way to express the signature of `CapturingFramework` in general terms -- there's just no way to say 'the same arguments as Framework but with one extra keyword argument'. I suppose the 'right' thing to do for type checking purposes would be to totally duplicate the `Framework` arguments, but since it's private and used in only one place, I prefer `*args, **kwargs` here -- especially if we go with `_make_framework(self, *args: Any, **kwargs: Any)`.

It is possible to get type checking, autocompletions, etc. for callers of a wrapper function like `_make_framework` by using a little no-op decorator (see below), but I don't think it's worth it given that `_make_framework` is private and would only be called in one place.

```py
_P = typing.ParamSpec('_P')
_T = typing.TypeVar(_'T')
def _argtypes(_: Callable[_P, object], /):
    def decorator(fn: Callable[..., _T], /) -> Callable[_P, _T]:
        return fn
    return decorator

@_argtypes(_framework.Framework.__init__)
def _make_framework(self, *args: Any, **kwargs: Any):
    ...
```

tonyandrewmeyer: Yeah, there's some discussion in #1631 about using args/kwargs, where I convinced Ben that it was best to have it that way rather than repeating everything.

james-garner-canonical (approved):

PR #1914, opened 2025-07-18
test: replace Python version to 3.10 for observability charm tests

IronCore864:
Replace Python version to 3.10 for observability charm tests.

See discussion [here](https://github.com/canonical/operator/issues/1877#issuecomment-3026362687).

Closes https://github.com/canonical/operator/issues/1877.

Also, the logic for handling `requirements.txt` is removed, since all three charms from observability teams don't use it any more.

### DIFF
--- a/.github/workflows/observability-charm-tests.yaml
+++ b/.github/workflows/observability-charm-tests.yaml
@@ -39,12 +39,12 @@ jobs:
         if: ${{ !(matrix.disabled) }}
         run: |
           if [ -e "uv.lock" ]; then
-            sed -i 's/requires-python = ">=3.8, <4"/requires-python = ">=3.10, <4"/g' uv.lock
             uv remove ops --optional dev --frozen || echo "maybe ops[testing] is not a dev dependency"
             uv remove ops-scenario --optional dev --frozen || echo "maybe ops-scenario is not a dev dependency"
             uv remove ops --frozen
             uv add git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#subdirectory=testing --optional dev --raw-sources --prerelease=if-necessary-or-explicit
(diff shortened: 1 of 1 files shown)


.github/workflows/observability-charm-tests.yaml
```diff
@@ -41,19 +38,15 @@ jobs:
...
-            sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
-            echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
-            sed -i -e "/^ops-scenario[ ><=]/d" -e "/^ops\[testing\][ ><=]/d" requirements.txt
-            echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops-scenario&subdirectory=testing" >> requirements.txt
-          elif [ -e "uv.lock" ]; then
+          if [ -e "uv.lock" ]; then
             uv remove ops --optional dev --frozen || echo "maybe ops[testing] is not a dev dependency"
             uv remove ops-scenario --optional dev --frozen || echo "maybe ops-scenario is not a dev dependency"
             uv remove ops --frozen
+            sed -i 's/requires-python = "~=3.8"/requires-python = "~=3.10"/g' pyproject.toml
```
dimaqq, line 45: I guess that's one way of doing this!

dimaqq: Interestingly, the first charm in the list, declares only one base:

https://github.com/canonical/alertmanager-k8s-operator/blob/main/charmcraft.yaml#L21-L22

They could have bumped min python version to 3.12 🤔

PR #1916, opened 2025-07-23
ci: fixes for the sbom and secscan workflow, and trigger it on publishing

tonyandrewmeyer:
Connect the publishing workflow to also trigger SBOM generation and secscan for the new artifacts. Also adds instructions to the release process to copy the resulting reports to the correct location.

Also fixes a few issues with the existing workflow:
 * Explicitly install the libapt-pkg-dev package, which is required to install the Python apt package.
 * Upload the reports into separate wheel and sdist locations, to avoid clashes.
 * Remove the 2.23/7.23 versions -- those should be in the 2.23 branch.
 * Remove an old configuration setting from the wheel config.

Also makes use of the new ability of sbomber to automatically detect the version number, so that the manifest files don't need to be updated for each release.

For ops 2.23, ops-scenario 7.23, and ops-tracing 2.23, new releases will continue to use manual processes.

### DIFF
--- a/.sbomber-manifest-sdist.yaml
+++ b/.sbomber-manifest-sdist.yaml
@@ -10,30 +10,28 @@ clients:
 artifacts:
   - name: 'ops'
     type: 'sdist'
-    version: '3.0.0'
     compression: 'gz'
     ssdlc_params:
         name: 'ops'
-        version: '3.0.0'
+        version: ''
         channel: 'stable'
         cycle: '25.10'
 
   - name: 'ops-scenario'
     type: 'sdist'
-    version: '8.0.0'
     compression: 'gz'
     ssdlc_params:
         name: 'ops-scenario'
-        version: '8.0.0'
+        version: ''
         channel: 'stable'
         cycle: '25.10'
 
   - name: 'ops-tracing'
     type: 'sdist'
-    version: '3.0.0'
+    version: ''
     compression: 'gz'
     ssdlc_params:
         name: 'ops-tracing'
-        version: '3.0.0'
+        version: ''
         channel: 'stable'
         cycle: '25.10'
--- a/.sbomber-manifest-wheel.yaml
+++ b/.sbomber-manifest-wheel.yaml
@@ -10,27 +10,24 @@ clients:
 artifacts:
   - name: 'ops'
     type: 'wheel'
-    version: '3.0.0'
     ssdlc_params:
         name: 'ops'
-        version: '3.0.0'
+        version: ''
         channel: 'stable'
         cycle: '25.10'
 
   - name: 'ops-scenario'
     type: 'wheel'
-    version: '8.0.0'
     ssdlc_params:
         name: 'ops-scenario'
-        version: '8.0.0'
+        version: ''
         channel: 'stable'
         cycle: '25.10'
 
   - name: 'ops-tracing'
     type: 'wheel'
-    version: '3.0.0'
     ssdlc_params:
         name: 'ops-tracing'
-        version: '3.0.0'
+        version: ''
         channel: 'stable'
         cycle: '25.10'


tonyandrewmeyer:
This still has two TODO items, but they're both trivial (and could be done in a follow-up PR if the merge and security approval isn't ready in time). Requesting a review in advance of that to hopefully get this included in the process for July's release, and address any issues found in review before those minor adjustments.

HACKING.md
```diff
@@ -261,15 +261,16 @@ the build frontend is [build](https://pypi.org/project/build/).
     >
     > You can troubleshoot errors on the [Actions Tab](https://github.com/canonical/operator/actions).
 
-5. Announce the release on [Discourse](https://discourse.charmhub.io/c/framework/42) and
+5. In the [SBOM and secscan workflow in the Actions Tab](https://github.com/canonical/operator/actions/workflows/sbom-secscan.yaml), verify that there is a run for the new release. In the workflow run, there will be two artifacts produced, `secscan-report-upload-sdist` and `secscan-report-upload-wheel`. Download both of these, and then upload them to the [SSDLC Ops folder in Drive](https://drive.google.com/drive/folders/17pOwak4LQ6sicr6OekuVPMECt2OcMRj8?usp=drive_link). Open the artifacts and verify that the security scan has not found any vulnerabilities. If you are releasing from the 2.23-maintenance branch, then follow the manual process instead, for both [SBOM generation](https://library.canonical.com/corporate-policies/information-security-policies/ssdlc/ssdlc---software-bill-of-materials-(sbom)) and [security scanning](https://library.canonical.com/corporate-policies/information-security-policies/ssdlc/ssdlc---vulnerability-identification).
```
benhoyt, line 264: Oh, when discussing the other day whether to automate this I was thinking this was a once a cycle thing, or whenever the security team needed updates. I didn't realise it was every release. Do we *need* to do it every release, or just once a cycle (and ad-hoc other than that)?

dimaqq (approved):
Code looks good.

I'm not sure about the extra manual steps, but we can always start like this and improve later.

HACKING.md
dwilding, line 264: > Open the artifacts and verify that the security scan has not found any vulnerabilities

I haven't used artifacts like this before - is it self-explanatory how to verify the artifacts? From reading the instructions, my assumption is that they are text files, and there would be obvious warnings/errors if a vulnerability was found.

tonyandrewmeyer: You can [download an example here if you'd like](https://github.com/canonical/operator/actions/runs/16458125290) - I should have put that in the description.

They're HTML files, and hopefully look like this:

IMG_5142

I don't have a failing one at hand, but it's pretty obviously not the "nothing found" result like the above.

I think it would be nice if the tool left a comment or scan result in the repo if it found something, but I don't think I have time to add that feature at the moment.

tonyandrewmeyer: The [policy](https://library.canonical.com/corporate-policies/information-security-policies/ssdlc/ssdlc---vulnerability-identification) says:

> It is a requirement for each product and engineering group to complete vulnerability scanning, through Canonical's scanning range (hosted by Security Engineering), each cycle[1]
> [1] Additional security scanning tools built into the CI/CD are highly encouraged, but at this point not required.

But also:

> The minimum timeframe for rescanning should be monthly, but weekly or daily is much preferred. Infrastructure to automate this rescanning is recommended, to ensure reliability.

With the addition of the SSDLC ID Params, the doc says that the results are copied to a long term store. Nowhere in that doc does it say that you have to upload the results otherwise. I'll check with Ricardo about this when talking to him tomorrow -- perhaps we don't need to provide the results if we're using the params.

For SBOM, the [policy](https://library.canonical.com/corporate-policies/information-security-policies/ssdlc/ssdlc---software-bill-of-materials-(sbom)) says:

> Coming soon, by the end of 25.04:
> Automatic upload of SBOMs to Google Drive

However, I haven't seen any announcement of this. I can ask Ricardo about that too. It also doesn't specify anything about providing the results, other than saying that you should post on MM, but nothing about frequency.

The overall [SSDLC policy](https://library.canonical.com/corporate-policies/information-security-policies/secure-software-development-lifecycle-(ssdlc)-process) says that both security scanning and SBOM generation are required every cycle. Most of it seems very designed to per-cycle releases like Ubuntu.

We do ~6 releases per cycle - it doesn't seem like a big ask to download and upload the results as part of that, in my opinion, although it would certainly be better if it was automated on the other side.

I do strongly feel that we should be doing the security scan for every release (even though I'm not particularly convinced it'll find anything that dependabot wouldn't), and generating an SBOM for every release (since the GitHub SBOM generator will only generate the current SBOM).

tonyandrewmeyer:
> This still has two TODO items, but they're both trivial (and could be done in a follow-up PR if the merge and security approval isn't ready in time). Requesting a review in advance of that to hopefully get this included in the process for July's release, and address any issues found in review before those minor adjustments.

The auto-detect version numbers feature is merged, so that's incorporated in here now. No answer from the security team about making sbomber public yet, so that should wait for a future PR (assuming approval is granted -- if not, then switching from my token to a prints-charming one).

HACKING.md
benhoyt: Sounds reasonable, thanks. Be good to hear more details from Ricardo on some of those items, and in the meantime I'm happy to progress with the manual download+upload (per monthly release).

tonyandrewmeyer: TL;DR: near-term we should be able to drop this requirement, for this cycle let's continue with the process as described here.

--

For SBOM, Ricardo has to check with some other people and hasn't been able to do that yet, so for the moment the ask is that we continue to upload it. Once a cycle is sufficient, but if we have it generated for each release then uploading that is nice. This is a short-term requirement, as there is definitely automation coming that will handle this for us.

For the security scan, they have this information (because our automation uses the ID params feature) so we don't need to store the results separately. However, our process bundles this with the SBOM so we'll end up doing it until we don't need the SBOM one any more.

There is a request to make sure that the [security dashboard](https://docs.google.com/spreadsheets/d/1tj5SMFBFlNo1Eq6dtZFJgU39bCHZRkw90QEoftRhOtg/edit?gid=2129585703#gid=2129585703) is updated with the secscan results once per cycle. I don't think that belongs in HACKING.md, though - there's automation coming that will produce a Jira SSDLC ticket for us with a DoD each cycle, and I think that's a better place to track once-per-cycle work.

PR #1918, opened 2025-07-23
fix: only add the remote unit for departed and broken relation events, fix ordering

### LINKED ISSUE #1917: Framework wrongly inserts the remote unit in relation-joined events
From [a comment in #1364 by @addyess ](https://github.com/canonical/operator/pull/1364#discussion_r2220448462):

I have a charm k8s that relates to another application on multiple relations:

 * containerd
* k8s-cluster

During the containerd_relation_joined event with k8s-worker/0 , if I look at the model.get_relation("k8s-cluster").units -- I see k8s-worker/0. But k8s-worker/0 hasn't yet run it's relation_joined joined event.

Because it shows up in the Relation.units, when I try access the data relation.data[unit], I end up with a KeyError because that unit is not yet joined on the k8s-cluster relation.

I was able to change this line of code (380 in _main) to:

```python
            remote_unit_name=broken_relation and self._juju_context.remote_unit_name
```

and the issue goes away. I know that's probably not the right fix -- but i'm not sure I understand the value of Relation.units showing a unit that isn't actually joined to a relation.

(Quote ends).

This does seem incorrect behaviour. It should already be present in the appropriate early events and in changed, so we could just add it in departed and broken and that would avoid this.

tonyandrewmeyer:
We inject the `JUJU_REMOTE_UNIT` unit into the relation's unit so that the databag is available during relation-departed events. When this was added in #1364 the thinking was that it was cleaner to just always add that unit rather than check for the specific events. However, that was not correct: this means that the unit may be added before the data is available (in relation joined, for example).

This PR changes the injection so that it only happens in relation departed and relation broken events.

In addition, we inserted the unit into the relation *after* setting up the relation's `.data`, which meant that the data would not be accessible for the inserted unit. The PR adjusts the ordering so that the inserting is done first.

Fixes #1917

### DIFF
--- a/ops/_main.py
+++ b/ops/_main.py
@@ -370,7 +370,7 @@ class _Manager:
     def _make_framework(self, dispatcher: _Dispatcher):
         # If we are in a RelationBroken event, we want to know which relation is
         # broken within the model, not only in the event's `.relation` attribute.
-        if self._juju_context.dispatch_path.endswith(('-relation-broken', '_relation_broken')):
+        if dispatcher.event_name.endswith('_relation_broken'):
             broken_relation_id = self._juju_context.relation_id
         else:
             broken_relation_id = None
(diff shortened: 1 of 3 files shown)


ops/_main.py
```diff
@@ -375,19 +375,30 @@ def _make_framework(self, dispatcher: _Dispatcher):
...
+        # We also expect it to be available in RelationBroken events, so ensure
+        # that it's available then as well. For other relation events, the unit
+        # will either already be in the set via `relation-list` (such as in a
+        # RelationChanged event) or correctly not in the list yet because the
+        # relation has not been rully set up (such as in a RelationJoined event).
+        if self._juju_context.dispatch_path.endswith((
+            '-relation-departed',
+            '_relation_departed',
+            '-relation-broken',
+            '_relation_broken',
```
benhoyt: Why do we need to check for both `_` and `-` at this point? Won't it be consistent?

dimaqq (approved):

ops/_main.py
tonyandrewmeyer: I had lazily copied the pattern from the check above. I'll figure out which one it should be and make both consistent.

PR #1921, opened 2025-07-25
test: make scenario context app name and unit id public

IronCore864:


benhoyt:
I think Tony has already addressed this in #1920.

PR #1925, opened 2025-07-28
fix: add the remote unit to Relation.data but not Relation.units

tonyandrewmeyer:
In #1364 (to address #1109) we added the departed unit to `Relation.units` so that charms could access the relation data in the relation-departed event.

This was followed up by #1897, to fix an issue where the app was not available, and then #1918, to add the unit before populating `Relation.data`, and to only add the unit for `relation-departed` and `relation-broken`, rather than all events.

A problem that the original solution introduces is a subtle behaviour change in what `Relation.units` means. It's not explicit in the documentation, but previously it was always the same as the `relation-list` output. Changing which units are included could cause issues for holistic or reconcile charms that would now see the departed or broken unit as still part of the relation, until the event after relation-broken.

In this PR, the implementation is adjusted so that rather than adding the remote unit to `Relation.units`, it is added to `Relation.data`. This restores the original meaning of `Relation.units`, but still allows charms to access the data of the departed or broken unit - but the charm will need to do so explicitly (via `event.relation.unit` for example) rather than looping through `Relation.units`.

The fix from #1897 and the 'only for departed and broken' parts of #1918 are kept (the other part of #1918 is no longer relevant).

The tests that were added to validate the "adding an extra unit to `Relation.units`" behaviour are removed, but the test for validating that the data is accessible is kept (and adjusted to explicitly request the data).

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1786,9 +1786,11 @@ class Relation:
             and self.app is not None
             and _remote_unit.name.startswith(f'{self.app.name}/')
         ):
-            self.units.add(_remote_unit)
+            remote_unit = _remote_unit
+        else:
+            remote_unit = None
 
-        self.data = RelationData(self, our_unit, backend)
+        self.data = RelationData(self, our_unit, backend, remote_unit)
 
         self._remote_model: RemoteModel | None = None
 
@@ -1984,7 +1986,13 @@ class RelationData(Mapping[Union[Unit, Application], 'RelationDataContent']):
     :attr:`Relation.data`
     """
 
-    def __init__(self, relation: Relation, our_unit: Unit, backend: _ModelBackend):
+    def __init__(
+        self,
+        relation: Relation,
+        our_unit: Unit,
+        backend: _ModelBackend,
+        remote_unit: Unit | None = None,
+    ):
         self.relation = weakref.proxy(relation)
         self._data: dict[Unit | Application, RelationDataContent] = {
             our_unit: RelationDataContent(self.relation, our_unit, backend),
@@ -1998,6 +2006,9 @@ class RelationData(Mapping[Union[Unit, Application], 'RelationDataContent']):
             self._data.update({
                 self.relation.app: RelationDataContent(self.relation, self.relation.app, backend),
             })
+        # The unit might be departing or broken, so not in relation-list, but accessible.
+        if remote_unit is not None and remote_unit not in self._data:
+            self._data[remote_unit] = RelationDataContent(self.relation, remote_unit, backend)
 
     def __contains__(self, key: Unit | Application):
         return key in self._data
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -4426,15 +4426,15 @@ def test_departing_unit_data_available(fake_script: FakeScript):
     fake_script.write('relation-list', """echo '["db/0"]'""")
     fake_script.write('relation-get', """echo '{"db": "data"}'""")
 
-    model = ops.model.Model(
-        ops.charm.CharmMeta({'name': 'mycharm', 'requires': {'db': {'interface': 'db'}}}),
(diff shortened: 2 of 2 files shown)


tonyandrewmeyer:
@carlcsaposs-canonical if you have a free moment, could you see whether this change make sense to you? Thanks!

dimaqq:
Observability tests fail with two kinds of errors:
- `OSError: [Errno 8] Exec format error: '/home/runner/work/operator/operator/cos-tool-amd64'` -- are the tests run on arm?
- `ops/_main.py", line 165, in __init__    raise _Abort(0)` -- this is weird, that's the `OPERATOR_DISPATCH` trick to prevent recursion... but why does this happen in the same process, under Scenario? https://github.com/canonical/operator/blob/main/ops/_main.py#L163-L165

tonyandrewmeyer:
> Observability tests fail with two kinds of errors:

Yes, I'm looking in that. They aren't related to this PR.

> * `ops/_main.py", line 165, in __init__    raise _Abort(0)` -- this is weird, that's the `OPERATOR_DISPATCH` trick to prevent recursion... but why does this happen in the same process, under Scenario? https://github.com/canonical/operator/blob/main/ops/_main.py#L163-L165

This is happening when a test crashes, for subsequent tests. Fixing the crashing test stops it happening. It seems like the environment isn't being reset, maybe related to the recent change in handling `_Abort`. But again, not related to this PR.

IronCore864 (approved):

dimaqq:
I've made https://github.com/canonical/operator/issues/1926 for the last one.

benhoyt (approved):

carlcsaposs-canonical:
> @carlcsaposs-canonical if you have a free moment, could you see whether this change make sense to you? Thanks!

@tonyandrewmeyer I recently had a look at #1364, #1837, #1897, #1918, #1925, and #1928 and I'm a bit confused about the goal of these changes

---

Behavior I saw when testing juju (3.6.2, on k8s):
- (for peer and non-peer relations) after relation departed, the departed unit's databag remains accessible (via `relation-get`) indefinitely (until after the relation is broken)
  - for example, if a unit is scaled down, its databag continues to be available after the unit has been removed
- (for peer and non-peer relations) during & after relation departed, the departing unit is removed from `relation-list` output
- when a unit with the same name is added (e.g. scale down and scale up on k8s), the databag is cleared once the new unit has access to its databag

---

regarding our experience in practice, I'm not aware of case where we rely on data in a unit databag for cleanup (unlike app databags), so think that's why we haven't noticed juju's behavior here

regarding what I would expect in theory (imagining possible use cases):
- given the knowledge that juju ([theoretically](https://github.com/juju/juju/issues/20214)) guarantees that all units successfully execute relation-departed before any units get relation-broken, I would expect all databags to be accessible in relation departed, in case info in those databags is needed for cleanup.
  - (For example, in mysql router, we rely on data in our local app databag during relation-broken for cleanup. But that's not exactly the same case—in relation broken, I would only expect local databags to be accessible, not remote databags)
- however I wouldn't necessarily expect the departing unit in `relation-list` since we might not get another event after departed

---

my understanding of the current ops behavior:
- during relation-departed or relation-broken, for non-peer relations, the departing unit is added to relation.data if the departing unit is from this app

when does that happen?

(think this might've been what Dima was pointing out [here](https://github.com/canonical/operator/pull/1837#discussion_r2162824712))

for non-peer relations, my understanding is the departing unit is always from the remote app
- except if the current unit is tearing down, then departing unit is the current unit (by "current unit", I mean the unit that ops/hook tools are being executed on)
  - but in this case, the current unit would never be in `relation-list` anyways
- except maybe in subordinate charms, didn't check

so I'm confused about how the current ops behavior is any different from before (the totality of) #1364, #1837, #1897, #1918, #1925, and #1928

PR #1940, opened 2025-07-31
docs: update links and config for switch to documentation.ubuntu.com/ops

dwilding:
This PR changes all hardcoded doc links to point to https://documentation.ubuntu.com/ops/latest/ instead of https://ops.readthedocs.io/en/latest/. In addition to switching to documentation.ubuntu.com, we're dropping `en` from URLs.

This PR also changes the Sphinx config to make sure that all styles can be loaded.

---

Related PR targeted at 2.23-maintenance: https://github.com/canonical/operator/pull/1942. As soon as both PRs have merged, I'll set up redirects from ops.readthedocs.io to documentation.ubuntu.com/ops.

PR #1942, opened 2025-07-31
docs: for 2.23, update links and config for switch to documentation.ubuntu.com/ops

dwilding:
This PR changes _selected_ hardcoded doc links to point to https://documentation.ubuntu.com/ops/2.23/ or https://documentation.ubuntu.com/ops/latest/ (as appropriate) instead of https://ops.readthedocs.io/en/latest/.

Since the changes are targeted at 2.23-maintenance, I wanted to avoid unnecessary doc and code changes. So I'm only adjusting the links that people would first encounter in the README/PyPI/docs.

This PR also changes the Sphinx config to make sure that all styles can be loaded.

---

Related PR targeted at main: https://github.com/canonical/operator/pull/1940. As soon as both PRs have merged, I'll set up redirects from ops.readthedocs.io to documentation.ubuntu.com/ops.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 The `ops` library is a Python framework for developing and testing Kubernetes and machine [charms](https://charmhub.io/). While charms can be written in any language, `ops` defines the latest standard, and charmers are encouraged to use Python with `ops` for all charms. The library is an official component of the Charm SDK, itself a part of [the Juju universe](https://juju.is/).
 
 > - `ops` is  [available on PyPI](https://pypi.org/project/ops/).
-> - Version 2.23 of `ops` requires Python 3.8 or above.
+> - Version 2 of `ops` requires Python 3.8 or above.
 > - Read our [docs](https://documentation.ubuntu.com/ops/2.x/) for tutorials, how-to guides, the library reference, and more.
 
 ## Give it a try
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 The `ops` library is a Python framework for developing and testing Kubernetes and machine [charms](https://charmhub.io/). While charms can be written in any language, `ops` defines the latest standard, and charmers are encouraged to use Python with `ops` for all charms. The library is an official component of the Charm SDK, itself a part of [the Juju universe](https://juju.is/).
 
 > - `ops` is  [available on PyPI](https://pypi.org/project/ops/).
-> - The latest version of `ops` requires Python 3.8 or above.
-> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, the library reference, and more.
+> - Version 2.23 of `ops` requires Python 3.8 or above.
+> - Read our [docs](https://documentation.ubuntu.com/ops/2.23/) for tutorials, how-to guides, the library reference, and more.
 
 ## Give it a try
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -5,8 +5,8 @@
 The `ops` library is a Python framework for developing and testing Kubernetes and machine [charms](https://charmhub.io/). While charms can be written in any language, `ops` defines the latest standard, and charmers are encouraged to use Python with `ops` for all charms. The library is an official component of the Charm SDK, itself a part of [the Juju universe](https://juju.is/).
 
 > - `ops` is  [available on PyPI](https://pypi.org/project/ops/).
-> - The latest version of `ops` requires Python 3.8 or above.
-> - Read our [docs](https://ops.readthedocs.io/en/latest/) for tutorials, how-to guides, the library reference, and more.
+> - Version 2.23 of `ops` requires Python 3.8 or above.
```
dwilding: The Python version needs changing in main's README. I've created #1943 for that.

docs/custom_conf.py
```diff
@@ -154,7 +154,7 @@
 
 # If your project is on documentation.ubuntu.com, specify the project
 # slug (for example, "lxd") here.
-slug = ''
+slug = 'ops'
```
dwilding, line 157: This prevents unstyled error pages. See https://github.com/canonical/pebble/pull/573

README.md
tonyandrewmeyer: I know the docs are specifically "2.23", but I think this would be better as something like:

```suggestion
> - Version 2.x of `ops` requires Python 3.8 or above.
```

tonyandrewmeyer (approved):
I somewhat feel like these docs should be "documentation.ubuntu.com/ops/2/" rather than "2.23". They are specifically for the 2.23.x versions, but the stable/latest docs are 3.x rather than 3.1 (and these docs were 2.x rather than any specific 2, while they were stable/latest).

It feels like you would go to these if you're using *any* 2.x version, so a "2" label would make more sense. Somewhat like "https://docs.python.org/3/library/subprocess.html" is "3" and "https://docs.python.org/2/library/subprocess.html" is "2" even though they take me to the 3.13.x (soon 3.14.x) and 2.7.x docs.

If this is difficult to do or doesn't align with Canonical practice, that's fine. Just how it feels to me.

I think not updating the tutorials (including examples) is fine. We really want people to always use the latest of those.

I think updating the link in the the deprecation warning in Harness would still make sense here.

I think switching the links in the Scenario README (which appears on PyPI as well) would also make sense.

README.md
benhoyt: Yeah, agreed. I prefer "2.x" over "2.23" too.

benhoyt:
Yep, I agree with tony about avoiding "2.23" (even though we're extremely unlikely to have a 2.24), however, I'd prefer `documentation.ubuntu.com/ops/2.x` over `documentation.ubuntu.com/ops/2` -- does that work for you, @tonyandrewmeyer?

benhoyt (approved):
Other than the comment about preferring 2.x, I'm good with this.

tonyandrewmeyer:
> I'd prefer `documentation.ubuntu.com/ops/2.x` over `documentation.ubuntu.com/ops/2` -- does that work for you, @tonyandrewmeyer?

Sure.

README.md
dwilding: The note on the homepage of the 2.x docs says "version 2 of Ops" (see https://documentation.ubuntu.com/ops/2.x/), so I've changed this part of the README to also say "version 2". We could of course change both to 2.x - let me know if you'd prefer that

benhoyt: Yeah, I think "version 2.x" would be preferable on both.

dwilding:
> I think updating the link in the the deprecation warning in Harness would still make sense here.
> 
> I think switching the links in the Scenario README (which appears on PyPI as well) would also make sense.

Done, thanks. I updated the main doc links in both the Scenario README and the tracing README.

README.md
dwilding: And "Ops 3.x" instead of "Ops 3" I assume?

PR #1944, opened 2025-07-31
chore: don't use f-strings in logging in the release script

tonyandrewmeyer:
f-strings shouldn't be used in calls to logging.log. In practice, this isn't going to cause issues in this specific case, but it's good to stick to best practice, and this avoids complaints by linters.

### DIFF
--- a/release.py
+++ b/release.py
@@ -100,9 +100,9 @@ def get_new_tag_for_release(
 
     suggested_tag = ''
     if not latest_tag:
-        logger.info(f'No version tag found in branch "{branch_name}"')
+        logger.info('No version tag found in branch %r', branch_name)
     else:
-        logger.info(f'Latest tag in branch "{branch_name}": {latest_tag}')
+        logger.info('Latest tag in branch %r: %s', branch_name, latest_tag)
         if not re.match(r'^\d+\.\d+\.\d+$', latest_tag):
             logger.info('Latest tag is not in format X.Y.Z.')
         else:
@@ -114,7 +114,7 @@ def get_new_tag_for_release(
             else:
                 logger.info('Branch is main, bumping minor version of the latest tag.')
                 suggested_tag = bump_minor_version(latest_tag)
-            logger.info(f'Suggested new version: {suggested_tag}')
+            logger.info('Suggested new version: %s', suggested_tag)
 
     tag_prompt = f' (press enter to use the tag {suggested_tag})' if suggested_tag else ''
     prompt = f'Input the new tag for the release{tag_prompt}\n'
@@ -132,7 +132,7 @@ def get_new_tag_for_release(
             continue
 
         release_page = f'https://github.com/{owner}/{repo.name}/releases'
-        logger.warning(f'Check out the releases page: {release_page} before confirming!')
+        logger.warning('Check out the releases page: %s before confirming!', release_page)
 
         confirm = (
             input(f'Confirm creating tag {new_tag!r} on branch {branch_name!r}? [y/N]: ')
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -100,9 +100,9 @@ def get_new_tag_for_release(
 
     suggested_tag = ''
     if not latest_tag:
-        logger.info(f'No version tag found in branch "{branch_name}"')
+        logger.info('No version tag found in branch %r', branch_name)
     else:
-        logger.info(f'Latest tag in branch "{branch_name}": {latest_tag}')
+        logger.info('Latest tag in branch %r: %s', branch_name, latest_tag)
```
dwilding: Very minor: Is there a need to use `%r` for the branch name? The branch name is already an identifier (ish), so I don't feel that `branch 'foo'` vs `branch foo` adds much in the output.

release.py
```diff
@@ -419,16 +419,17 @@ def get_new_version_post_release(repo: github.Repository.Repository, branch_name
...
+    logger.info('Latest version in branch %r: %s', branch_name, latest_version)
 
     # Check if the latest version is in SEMVER, in case it has a suffix like 3.0.0.post1.
     if not re.match(r'^\d+\.\d+\.\d+$', latest_version):
         logger.error(
-            f'Latest version "{latest_version}" must be in format '
-            f'X.Y.Z. Please input the new version manually (remember to add the .dev0 suffix).'
+            'Latest version %r must be in format '
+            'X.Y.Z. Please input the new version manually (remember to add the .dev0 suffix).',
+            latest_version,
```
dimaqq: Maybe this could be fixed while at it:

Instructions unclear: X.Y.Z or X.Y.Z.dev0?

release.py
```diff
@@ -460,10 +461,10 @@ def check_update_charm_pins_prs(repo: github.Repository.Repository):
         logger.info('Please merge "update charm pins" PRs first:')
         pr = open_prs[0]
         pr_url = f'#{pr.number} - {pr.html_url}'
-        logger.info(f'  {pr_url}')
+        logger.info('  %s', pr_url)
         if len(open_prs) > 1:
             logger.info(
-                f'Note that there are {len(open_prs) - 1} more open update charm pins PRs.'
+                'Note that there are %s more open update charm pins PRs.', len(open_prs) - 1
```
dimaqq, line 475: **nit:** there ought not be multiple update charm pin PRs.
at least not targetting the main branch.

release.py
```diff
@@ -539,8 +540,8 @@ def post_release(
     ])
     if local_branch or remote_branch:
         logger.error(
-            f'Branch "{new_branch}" already exists.'
-            'Please double check and delete it first before post release'
+            'Branch %r already exists.Please double check and delete it first before post release',
```
dimaqq:
```suggestion
            'Branch %r already exists. Please double check and delete it first before post release',
```

dimaqq (approved):

release.py
tonyandrewmeyer: True, so we could probably remove this entire block, but that feels quite invasive a drive-by for a "don't use f-strings in logs" PR.

release.py
tonyandrewmeyer: I was leaving things untouched (well, this will swap `"` for `'`, but otherwise). I'm fine changing it:

```suggestion
        logger.info('Latest tag in branch %s: %s', branch_name, latest_tag)
```

release.py
tonyandrewmeyer: Is it clearer now?

Issue #1943, opened 2025-07-31
Update info in README about Python versions


dwilding:
In [README.md](https://github.com/canonical/operator/blob/main/README.md), we say:

> The latest version of ops requires Python 3.8 or above.

This needs updating to Python 3.10.

PR #1946, opened 2025-08-01
docs: Update the required Python version and note the 2.x documentation site

### LINKED ISSUE #1941: Update info in CONTRIBUTING.md about doc versions
In [CONTRIBUTING.md](https://github.com/canonical/operator/blob/main/CONTRIBUTING.md#how-to-document-version-dependencies) we say:

> We don't publish separate documentation for separate versions of Ops. The published docs at [ops.readthedocs.io](https://ops.readthedocs.io/en/latest/index.html) are always for the in-development (main branch) of Ops [...]

This needs to be updated, now that we publish separate docs for 2.23.

tonyandrewmeyer:
Two minor doc changes following on from the 3.0 release:

* Update the minimum Python version in the README.
* We do publish separate docs for separate versions of Ops now, but only major versions, and we don't generally backport doc changes.

Fixes #1941, #1943

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,9 @@ tox -e docs-live
 
 ## How to document version dependencies
 
-We don't publish separate documentation for separate versions of Ops. The published docs at [documentation.ubuntu.com/ops](https://documentation.ubuntu.com/ops/latest/) are always for the in-development (main branch) of Ops, and do not include any notes indicating changes or additions across Ops versions. We encourage all charmers to promptly upgrade to the latest version of Ops, and to refer to the release notes and changelog for learning about changes.
+We publish separate documentation for each major version of Ops. We generally only make improvements to the latest version of the docs, and only backport changes that are applicable only to the older version, or are critical for charming.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -91,7 +91,9 @@ tox -e docs-live
 
 ## How to document version dependencies
 
-We don't publish separate documentation for separate versions of Ops. The published docs at [documentation.ubuntu.com/ops](https://documentation.ubuntu.com/ops/latest/) are always for the in-development (main branch) of Ops, and do not include any notes indicating changes or additions across Ops versions. We encourage all charmers to promptly upgrade to the latest version of Ops, and to refer to the release notes and changelog for learning about changes.
+We publish separate documentation for each major version of Ops. We generally only make improvements to the latest version of the docs, and only backport changes that are applicable only to the older version, or are critical for charming.
```
dwilding: I suppose strictly speaking, if a doc change is only applicable to the older version, it wouldn't be in the latest docs in the first place - so there'd be nothing to backport. But I get what you mean. We update the 2.x docs in case of an Ops change that only applies to version 2.x. What do you think about this alternative wording?

```suggestion
We publish separate documentation for each major version of Ops. We generally only make improvements to the latest version of the docs. If an older version of Ops changes in a way that's only applicable to that version, we update the older version of the docs. We also update the older version of the docs if there's an improvement that's critical for charming.
```

dwilding (approved):
Thanks!

benhoyt (approved):

PR #1947, opened 2025-08-01
docs: be consistent with recommending self.config

dwilding:
This PR changes `self.model.config` to `self.config` in the docs. I'm also updating the description of ConfigData to be consistent with [Application](https://documentation.ubuntu.com/ops/latest/reference/ops.html#ops.Application).

### DIFF
--- a/docs/explanation/storedstate-guidance.md
+++ b/docs/explanation/storedstate-guidance.md
@@ -19,7 +19,7 @@ Here's a simplified charm code snippet that will allow us to toggle the state of
 
 ```python
 def _on_config_changed(self, event: ops.ConfigChangedEvent):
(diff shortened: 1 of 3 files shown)


IronCore864 (approved):

tonyandrewmeyer (approved):
Thanks! It's nice that we are recommending both `.config` and `.load_config` in `ConfigData` now.

I see one more case in `docs/howto/run-workloads-with-a-charm-kubernetes.md`, should we change that too?

dwilding:
> I see one more case in `docs/howto/run-workloads-with-a-charm-kubernetes.md`, should we change that too?

Huh, I just re-ran my search and found this too. No idea why I didn't spot this the first time round... Thanks!

Issue #1730, opened 2025-05-15
Upgrade docs starter pack to latest


benhoyt:

dwilding:
For comparison, here's what we did for Pebble: https://github.com/canonical/pebble/pull/615

benhoyt:
@dwilding What's left here? Do you think you'll be able to do the rest as a Friday item, or should we consider it for our 26.04 roadmap?

dwilding:
Switching the docs build to the starter pack Makefile and enabling the latest versions of doc checks - especially the new spelling check. Although, I might leave the new spelling check disabled until @dimaqq's looked at switching to en-US.

I'll see whether I can wrap it up this Friday.

dwilding:
This is still in progress. PRs so far:

- https://github.com/canonical/operator/pull/2073
- https://github.com/canonical/operator/pull/2060

PR #1951, opened 2025-08-04
docs: use latest styles from starter pack and remove .html extensions

dwilding:
This PR partially updates Ops to the latest version of the [doc starter pack](https://github.com/canonical/sphinx-docs-starter-pack) (starting work on https://github.com/canonical/operator/issues/1730).

Main public-facing changes:

- Some slight style changes. For example, the right-hand TOC has got a bit wider.

- Page URLs are now "clean" and don't end in `.html`. For example, `howto/manage-charms/` instead of `howto/manage-charms.html`. This PR doesn't apply redirects; I can set up redirects in RTD.

**[Preview build](https://canonical-ubuntu-documentation-library--1951.com.readthedocs.build/ops/1951/)**

The latest version of the starter pack moves the Canonical styles to a package called `canonical-sphinx`. This means that a lot of files can be removed from the Ops repo. I've tried to reduce to only the files that we currently need.

Some of the remaining files are shuffled around, to be in more logical/discoverable places. One significant change is that `docs/conf.py` and `docs/custom_conf.py` have merged into `docs/conf.py`.

The latest version of the starter pack includes several local checks and workflows, e.g., a new spelling check. I haven't included any of these yet. I want to limit the scope of this PR to what's changed about building the docs. This should be a good foundation for adopting more of the starter pack later.


docs/.readthedocs.yaml
```diff
@@ -5,6 +5,8 @@ build:
   tools:
     python: "3.11"
   jobs:
+    post_checkout:
+      - git fetch --unshallow || true
```
dwilding, line 9: This is needed because the "last updated" dates on pages are generated from the git history.

docs/_static/project_specific.css
```diff
@@ -0,0 +1,10 @@
+/* Define a star icon for use in admonitions */
```
dwilding, line 1: This CSS is not newly-added. I rescued it from .sphinx/_static/custom.css, which is being removed.

pyproject.toml
```diff
@@ -42,19 +42,11 @@ lint = [
 ]
 docs = [
     "ops[testing,tracing]",
-    "canonical-sphinx-extensions",
-    "furo",
-    "linkify-it-py",
-    "myst-parser; python_version >= \"3.10\"",
-    "pyspelling",
-    "sphinx ~= 8.0.0; python_version >= \"3.10\"",
```
dwilding, line 50: Now that we've switched Ops to Python 3.10+, we shouldn't need to specify Python versions for any of the docs dependencies.

tox.ini
```diff
@@ -48,14 +48,14 @@ passenv =
     # https://docs.readthedocs.com/platform/stable/build-customization.html#where-to-put-files
     READTHEDOCS_OUTPUT
 commands =
-    sphinx-build -W --keep-going docs/ {env:READTHEDOCS_OUTPUT:docs/_build}/html
+    sphinx-build -W --keep-going -b dirhtml docs/ {env:READTHEDOCS_OUTPUT:docs/_build}/html -d docs/.sphinx/.doctrees
```
dwilding, line 51: `-b dirhtml` enables the builder that produces `howto/manage-charms/` instead of `howto/manage-charms.html`.

`-d docs/.sphinx/.doctrees` puts the cached files (generated during the build) in `.sphinx` instead of in the build directory. I don't think this makes a practical difference to much, but it matches the latest starter pack.

pyproject.toml
james-garner-canonical: So I guess the maintenance branch stays on the old starter pack?

tox.ini
james-garner-canonical: Does the `-d` option affect our `clean` recipe, if we have one somewhere currently?

docs/conf.py
```diff
@@ -1,166 +1,603 @@
-import sys
```
james-garner-canonical, line 1: Are there any changes to this file that are worth calling out? I assume it's mostly appending the old `custom_conf.py` to the new `conf.py`, with some interleaving where `conf.py` now sets something earlier in the file. I guess all those `TODO` comments come from the new `conf.py`.

pyproject.toml
dwilding: That's what I'm thinking, yeah. I don't think it's worth changing.

james-garner-canonical (approved):
The new build seems to be working fine. I viewed a few pages side by side with the current site, and it all seems to be in order. The page specific modification dates are nice.

tox.ini
dwilding: We don't have a `clean` recipe, as far as I can tell. The Makefile from the starter pack would bring one in for us - or we could adapt the [charmlibs one](https://github.com/canonical/charmlibs/blob/main/docs.just#L24) if you think that would be handy

docs/conf.py
dwilding: The diff is hard to review, sorry. But you're right. The only functional change is that we're now generating a [sitemap](https://canonical-ubuntu-documentation-library--1951.com.readthedocs.build/ops/1951/sitemap.xml) (ignore all the `MANUAL`s in there - when pushed to RTD they will be replaced by the version). Otherwise, the config should be equivalent to before.

dwilding:
I should add, the Ops repo contains several hardcoded links of the form `https://documentation.ubuntu.com/ops/latest/reference/ops-testing.html`. I'll do a follow-up PR to change these to `https://documentation.ubuntu.com/ops/latest/reference/ops-testing/`

tox.ini
james-garner-canonical: Cool, no changes needed in this PR then, and we can look into what we want from the Makefile later.

docs/.readthedocs.yaml
dimaqq: We have "last updated" today.
After all, shallow checkout does have the last commit, AFAIK.

Are you saying that the new way produces own date for each page, so you need more than the last commit?

Or is the intention to avoid bumping the global last updated date when there are no doc changes?

dimaqq (approved):

docs/.readthedocs.yaml
dwilding: This is to enable per-page last updated, based on looking at when each doc source file was last updated. Currently, "last updated" is more like "site last built" and is the same for all pages.

benhoyt:
Nice work on this, thanks!

PR #1958, opened 2025-08-06
chore: jubilant smoke tests

dimaqq:
Switch our smoke tests to `jubilant`... well Tony's jubilant-backports.

Part of our efforts to move from pytest-operator to jubilant everywhere.

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -3,9 +3,7 @@ name: ops Smoke Tests
 on:
   workflow_dispatch:
   schedule:
-    - cron:  '0 7 25 * *'
-  # FIXME temporary for testing only
-  pull_request:
+    - cron: '0 7 25 * *'
 
 permissions: {}
 
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -4,6 +4,8 @@ on:
   workflow_dispatch:
   schedule:
     - cron:  '0 7 25 * *'
+  # FIXME temporary for testing only
+  pull_request:
 
 permissions: {}
 
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -22,16 +22,12 @@ jobs:
         preset: ['machine', 'microk8s']
 
     steps:
-      - name: Install concierge
-        run: sudo snap install --classic concierge
-
-      - name: Install Juju and tools
-        run: >
+      - run: sudo snap install --classic concierge
+      - run: >
           sudo concierge prepare
-            --juju-channel=${{ matrix.juju-channel }}
-            --charmcraft-channel=3.x/stable
-            -p "${{ matrix.preset }}"
-
+          --juju-channel=${{ matrix.juju-channel }}
+          --charmcraft-channel=3.x/stable
+          -p "${{ matrix.preset }}"
       - name: Checkout the repository
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -19,7 +19,6 @@ jobs:
         # pylibjuju does not support Juju 4.x
         # We need to migrate the test to use Jubilant for Juju>2.
         juju-channel: ['2.9/stable', '3/stable']
-        charmcraft-channel: ['3.x/stable']
         preset: ['machine', 'microk8s']
 
     steps:
@@ -27,7 +26,11 @@ jobs:
         run: sudo snap install --classic concierge
 
       - name: Install Juju and tools
-        run: sudo concierge prepare --juju-channel=${{ matrix.juju-channel }} --charmcraft-channel=${{ matrix.charmcraft-channel }} -p "${{ matrix.preset }}"
+        run: >
+          sudo concierge prepare
+            --juju-channel=${{ matrix.juju-channel }}
+            --charmcraft-channel=3.x/stable
+            -p "${{ matrix.preset }}"
 
       - name: Checkout the repository
         uses: actions/checkout@v4
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,12 +73,10 @@ benchmark = [
 integration = [
     "ops[testing,tracing]",
     "pytest~=8.4",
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -171,19 +171,21 @@ commands_post = killall -y 3m pebble
...
-    # Inject the tarball into the smoke test charm's requirements.
-    bash -c 'echo "./$(ls -1 ./test/charms/test_smoke/ | grep tar.gz)" > ./test/charms/test_smoke/requirements.txt'
-    # Make sure we are using the appropriate version of pylibjuju.
-    bash -c 'JUJU_VERSION=$(juju --version | awk -F. "{print \$1 \".\" \$2}"); if [ -n "$JUJU_VERSION" ]; then uv pip install "juju ~= $JUJU_VERSION"; fi'
-
+    # Clean any stale artefacts, in case these steps are run locally and repeatedly.
+    bash -c 'rm -vf test/charms/test_smoke/*.tar.gz'
+    bash -c 'rm -vf test/charms/test_smoke/*.whl'
+    bash -c 'rm -vf test/charms/test_smoke/requirements.txt'
+    # Build ops and freidns from this checkout
```
tonyandrewmeyer:
```suggestion
    # Build ops and friends from this checkout.
```

pyproject.toml
```diff
@@ -68,17 +67,16 @@ coverage = [
     "coverage[toml]~=7.0",
 ]
 benchmark = [
-    "pytest-benchmark~=5.0; python_version >= \"3.9\"",
+    "pytest-benchmark~=5.0",
 ]
 integration = [
     "ops[testing,tracing]",
     "pytest~=8.4",
-    "pytest-operator~=0.23",
```
tonyandrewmeyer, line 76: Does this also mean we can remove the pylibjuju requirement? It doesn't seem like it's used elsewhere other than in the examples.

test/smoke/test_smoke.py
```diff
@@ -63,41 +63,28 @@ def pack(charm_dir: pathlib.Path):
...
+
     charm = pack(pathlib.Path('./test/charms/test_smoke/'))
 
-    app = await ops_test.model.deploy(
-        charm, base=f'ubuntu@{base}', application_name=f'{name}-smoke'
-    )
-    await ops_test.model.wait_for_idle(timeout=600)
+    app = f'smoke{base.replace(".", "")}'
+    # --force is needed to deploy with --series=noble (24.04) on Juju 2.9
+    juju.deploy(charm, app=app, base=f'ubuntu@{base}', force=True)
```
james-garner-canonical, line 80: Not a blocker for this PR by any means, but I wonder if `jubilant.Juju.version` would be a desirable feature.

.github/workflows/smoke.yaml
```diff
@@ -17,16 +17,15 @@ jobs:
         # pylibjuju does not support Juju 4.x
```
tonyandrewmeyer: We can probably remove this comment now.

tonyandrewmeyer (approved):
LGTM, thanks!

test/smoke/test_smoke.py
```diff
@@ -63,41 +63,28 @@ def pack(charm_dir: pathlib.Path):
...
+    juju.wait(lambda status: jubilant_backports.all_active(status, app), timeout=600)
+
 
-    assert app.status == 'active', f"Base ubuntu@{base} failed with '{app.status}' status"
+@pytest.fixture(scope='module')
+def juju():
+    with jubilant_backports.temp_model() as juju:
+        yield juju
```
james-garner-canonical, line 87: How about emitting logs on failure like [here](https://documentation.ubuntu.com/jubilant/how-to/migrate-from-pytest-operator/#a-juju-model-fixture)?
```py
        if request.session.testsfailed:
            log = juju.debug_log(limit=1000)
            print(log, end='')
```

pyproject.toml
james-garner-canonical: EDIT: This is incorrect.

It doesn't look like it's mentioned elsewhere in `pyproject.toml`, so maybe just needs ` uv lock` to drop it from the lockfile?

test/smoke/test_smoke.py
tonyandrewmeyer: For what it's worth, `jubilant_backports.Juju` has both `cli_version` and `cli_major_version`.

pyproject.toml
tonyandrewmeyer: Line 79/80, no?

james-garner-canonical: Oops!

test/smoke/test_smoke.py
benhoyt: That's a good idea. I've opened https://github.com/canonical/jubilant/issues/178

benhoyt:
I forget exactly what we decided about the future of tonyandrewmeyer/jubilant-backports, but are we sure enough of it's future that we want to depend on it here? Do we concretely plan to move it to canonical/jubilant-backports and have some level of support for it?

tonyandrewmeyer:
> I forget exactly what we decided about the future of tonyandrewmeyer/jubilant-backports, but are we sure enough of it's future that we want to depend on it here? Do we concretely plan to move it to canonical/jubilant-backports and have some level of support for it?

We decided that we would create a branch in canonical/jubilant and redo the implementation there (a bunch of things can be simplified having it in the same code, but most of the work can be reused) and then release a jubilant29 (or similar) package. The expectation is that this will be done this cycle in the second half of my secondment since the Data team was keen on having this to enable them to move to Jubilant.

So jubilant-backports doesn't have a future beyond this cycle (and I'll presumably archive tonyandrewmeyer/jubilant-backports at that point too) but a Jubilant that works with Juju 2.9 does. It seems reasonable to me to use jubilant-backports for now and then swap it out with jubilant29 when that's available.

benhoyt:
Great, thanks for the reminder, and that all sounds sane to me.

test/smoke/test_smoke.py
dimaqq: Given that this is a smoke test, I'm not sure it's worth it.

PR #1959, opened 2025-08-07
docs: add related doc links to homepage

dwilding:
Inspired by https://github.com/juju/juju/pull/20358, this PR adds a "related links" block to the docs homepage, with links to related docs in the charming ecosystem.

### DIFF
--- a/docs/_static/project_specific.css
+++ b/docs/_static/project_specific.css
@@ -8,3 +8,13 @@ body {
     -webkit-mask-image: var(--icon-star);
     mask-image: var(--icon-star);
 }
+
+/* For the "related links" block, use the same spacing & sizing as the "contents" block */
+.relatedlinks-title-container {
+    padding: var(--toc-title-padding);
+    padding-top: var(--toc-spacing-vertical);
+}
+.relatedlinks {
+    font-size: var(--toc-font-size);
+    line-height: 1.3;
+}
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,3 +1,7 @@
+---
+relatedlinks: "[Charmcraft](https://documentation.ubuntu.com/charmcraft/stable/), [Charmlibs](https://canonical-charmlibs.readthedocs-hosted.com/), [Concierge](https://github.com/canonical/concierge), [Jubilant](https://documentation.ubuntu.com/jubilant/), [Juju](https://documentation.ubuntu.com/juju/3.6/), [Pebble](https://documentation.ubuntu.com/pebble/)"
+---
+
 # Ops documentation
 
 ```{toctree}


docs/_static/project_specific.css
```diff
@@ -8,3 +8,13 @@ body {
...
+
+/* For the "related links" block, use the same spacing & sizing as the "contents" block */
+.relatedlinks-title-container {
+    padding: var(--toc-title-padding);
+    padding-top: var(--toc-spacing-vertical);
+}
+.relatedlinks {
+    font-size: var(--toc-font-size);
+    line-height: 1.3;
+}
```
dwilding: @SecondSkoll I feel it would be nice if the spacing & sizing of the "related links" block matches the "contents" block. But there might be a history behind the [current styling](https://github.com/canonical/canonical-sphinx-extensions/blob/main/canonical-sphinx-extensions/related-links/_static/related-links.css). What's your opinion about me adding these customisations to canonical-sphinx-extensions?

SecondSkoll: It was all done by Ruth. I don't have context, current or historical as I haven't used it. Happy for you to commit changes upstream though.

docs/index.md
```diff
@@ -1,3 +1,7 @@
+---
+relatedlinks: "[Charmcraft](https://documentation.ubuntu.com/charmcraft/stable/), [Charmlibs](https://canonical-charmlibs.readthedocs-hosted.com/), [Concierge](https://github.com/canonical/concierge), [Jubilant](https://documentation.ubuntu.com/jubilant/), [Juju](https://documentation.ubuntu.com/juju/3.6/), [Pebble](https://documentation.ubuntu.com/pebble/)"
```
tonyandrewmeyer, line 2: Can these (other than Concierge) be interspinx links instead? Or do we not have a ref to link to that is the base page for each of the sites?

dwilding: Unfortunately not. As far as I understand the related-links extension, we need to use absolute URLs.

docs/_static/project_specific.css
dwilding: I've opened https://github.com/canonical/canonical-sphinx-extensions/pull/74

tmihoc:
Changes from the Juju implementation = improvements; mind if I steal them all? 

One challenge in Juju is that I want to (a) include the Juju ecosystem docs link (the one to juju.is/docs which shows all of these docs sets in context) as well as (b) group the Juju thingies. 

About (b): It's a funny one because Jubilant is a way to do Juju but it really only serves charmers. Maybe the right grouping here is into Use charms and Build charms. It would be an interesting way to bring the old logic back in, even as it remains a bit faulty -- as in, Juju is also relevant if you're building charms, at least, in the real-life testing phase, (which is maybe an argument to keep filing that under Use charms, though Juju things such as `debug-hook` are really for charmers, though there too maybe one could argue they're about troubleshooting your Juju deployment, albeit a corner that charmers will tend to engage with especially. Enough with the hairsplitting: I'll file Jubilant under Build charms and Juju remains solidly under Use charms.

dwilding:
> Changes from the Juju implementation = improvements; mind if I steal them all?

Please go ahead!

> I'll file Jubilant under Build charms and Juju remains solidly under Use charms.

That sounds good to me. One _could_ use Jubilant as a general-purpose way to drive Juju, but as developed it's intended for charm integration testing.

tonyandrewmeyer:
> That sounds good to me. One _could_ use Jubilant as a general-purpose way to drive Juju, but as developed it's intended for charm integration testing.

In breaking news, this came up in the Charm Tech team-only today, and the intention going forward is that we won't be pushing that line any more. It's still the primary intended use-case, and we were always open to it being used elsewhere, but we'll drop the language around integration testing being *the* use case.

Basically, in its short life so far, Jubilant got too popular to stay in its lane :wink:.

dwilding:
I've removed the CSS customisation from this PR. That's moved upstream, to https://github.com/canonical/canonical-sphinx-extensions/pull/74. When the canonical-sphinx-extensions package is next published, Ops docs will get the updated CSS on the next rebuild.

dimaqq (approved):

PR #1960, opened 2025-08-07
demo: system metrics while running tests

dimaqq:
Shows what the GitHub action runner/vm was up to during smoke test.


tonyandrewmeyer:
Might be interesting to combine with adding `--measure` to the pack command?

PR #1963, opened 2025-08-07
ci: add Juju 4/beta to the smoke test matrix

tonyandrewmeyer:
Now that we're using Jubilant rather than python-libjuju, we can smoke test against Juju 4 as well. The smoke test only packs (not relevant) and deploys, so it's not showing a great deal of what works/doesn't work in Juju 4, but it's worth knowing this little bit.

[Passing run in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/16816761841).

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         # pylibjuju does not support Juju 4.x
         # We need to migrate the test to use Jubilant for Juju>2.
-        juju-channel: ['2.9/stable', '3/stable']
+        juju-channel: ['2.9/stable', '3/stable', '4/beta']
         charmcraft-channel: ['3.x/stable']
         preset: ['machine', 'microk8s']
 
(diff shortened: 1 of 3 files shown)


benhoyt (approved):
Nice!

dimaqq:
Let's see if it actually works: https://github.com/dimaqq/operator/pull/70

tonyandrewmeyer:
> Let's see if it actually works: [dimaqq#70](https://github.com/dimaqq/operator/pull/70)

It does. There's a link to a passing run in the ticket description 😄.

dimaqq (approved):
OT: looks like I forgot to remove `juju` from pyproject.toml in #1958 

Given that we start using Juju 4 now, do you reckon it's worse it to drop the dep and relock?

tonyandrewmeyer:
> OT: looks like I forgot to remove `juju` from pyproject.toml in #1958
> 
> Given that we start using Juju 4 now, do you reckon it's worse it to drop the dep and relock?

Sure, done.

PR #1966, opened 2025-08-08
chore: smoke test with uv

dimaqq:
Build the smoke test charm with uv; adjust tox.ini and test_smoke.py to match.

Successful run: https://github.com/canonical/operator/actions/runs/16910650895/job/47911097256

(Except one Juju 4 got stuck removing the model, as it does at times).


tox.ini
```diff
@@ -177,19 +177,19 @@ allowlist_externals =
...
     bash
+    cp
     uv
 dependency_groups = integration
 commands =
     # Clean any stale artefacts, in case these steps are run locally and repeatedly.
-    bash -c 'rm -vf test/charms/test_smoke/*.tar.gz'
-    bash -c 'rm -vf test/charms/test_smoke/*.whl'
-    bash -c 'rm -vf test/charms/test_smoke/requirements.txt'
+    bash -c 'find test/charms/test_smoke -name '*.whl' -or -name '*.tar.gz' -or -name pyproject.toml -or -name uv.lock | xargs -r rm -vf'
```
james-garner-canonical: Do these nested single quotes work as intended?

dimaqq: that's a good question, how come this didn't fail outright?

dimaqq:
Weirdly, the tests pass for me locally, both vm and k8s; only tested juju 3.6

And they fail repeatedly on GHA, why?

vm:
```
+ .apps['smoke2204'].app_status.current = 'error'
+ .apps['smoke2204'].app_status.message = 'hook failed: "install"'
+ .apps['smoke2204'].units['smoke2204/0'].workload_status.current = 'error'
+ .apps['smoke2204'].units['smoke2204/0'].workload_status.message = 'hook failed: "install"'
+ .apps['smoke2204'].units['smoke2204/0'].juju_status.current = 'idle'
```

k8s:
```
- .apps['smoke2204'].app_status.current = 'error'
- .apps['smoke2204'].app_status.message = 'hook failed: "install"'
- .apps['smoke2204'].units['smoke2204/0'].workload_status.current = 'error'
- .apps['smoke2204'].units['smoke2204/0'].workload_status.message = 'hook failed: "install"'
+ .apps['smoke2204'].app_status.current = 'maintenance'
+ .apps['smoke2204'].app_status.message = 'installing charm software'
+ .apps['smoke2204'].units['smoke2204/0'].workload_status.current = 'maintenance'
+ .apps['smoke2204'].units['smoke2204/0'].workload_status.message = 'installing charm software'
...
10 minutes later: timeout!
```

.github/workflows/smoke.yaml
```diff
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 7 25 * *'
+  pull_request:
```
tonyandrewmeyer: What's the reasoning here? I think these are too heavy to do on PRs, and super unlikely to break. Monthly seems fine. Or if we are going to do them on PR then I think we should remove the scheduled run.

.github/workflows/smoke.yaml
```diff
@@ -18,6 +19,7 @@
         preset: ['machine', 'microk8s']
 
     steps:
+      - uses: catchpoint/workflow-telemetry-action@v2
```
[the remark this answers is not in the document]
tonyandrewmeyer: Is this just while it's a draft?

test/smoke/test_smoke.py
```diff
@@ -78,14 +78,18 @@ def test_smoke(juju: jubilant_backports.Juju, base: str):
     app = f'smoke{base.replace(".", "")}'
     # --force is needed to deploy with --series=noble (24.04) on Juju 2.9
     juju.deploy(charm, app=app, base=f'ubuntu@{base}', force=True)
-    juju.wait(lambda status: jubilant_backports.all_active(status, app), timeout=600)
+    juju.wait(lambda status: jubilant_backports.all_active(status, app), timeout=999)
```
tonyandrewmeyer: Is this required? 10 minutes is already a long time to wait for a chair that doesn't do anything.

tox.ini
```diff
@@ -177,19 +177,19 @@ allowlist_externals =
...
 dependency_groups = integration
 commands =
     # Clean any stale artefacts, in case these steps are run locally and repeatedly.
-    bash -c 'rm -vf test/charms/test_smoke/*.tar.gz'
-    bash -c 'rm -vf test/charms/test_smoke/*.whl'
-    bash -c 'rm -vf test/charms/test_smoke/requirements.txt'
+    bash -c "find test/charms/test_smoke -name '*.whl' -or -name '*.tar.gz' -or -name pyproject.toml -or -name uv.lock | xargs -r rm -vf"
     # Build ops and friends from this checkout.
-    uv build --all --out-dir=test/charms/test_smoke/
+    uv build --all --out-dir=test/charms/test_smoke
```
tonyandrewmeyer: Since we're not using the sdist at all, what about we just build the wheel, then we don't need to clean it up.

tox.ini
```diff
@@ -177,19 +177,19 @@ allowlist_externals =
...
     # Inject current ops into the smoke test charm's requirements.
-    bash -c 'basename test/charms/test_smoke/ops-3.*.whl >> test/charms/test_smoke/requirements.txt'
+    cp test/charms/test_smoke/reference-pyproject.toml test/charms/test_smoke/pyproject.toml
```
tonyandrewmeyer, line 191: If we just call the file pyproject.toml does that cause issues?

dimaqq: I wanted to check in a clean version, as it gets changed during `tox -e smoke`.

This way, I can configure `.gitignore` to ignore `pyproject.toml` but not the clean file.

In other words, if someone runs smoke tests locally, I wouldn't want to burden them with a dirty tree / have them accidentally commit the changed file.

The file after changes looks like this:

```toml
[project]
name = "test-smoke"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.10"
dependencies = [
    "ops",
]

[tool.uv.sources]
ops = { path = "ops-3.2.0.dev0-py3-none-any.whl" }
```

The ops version is variable month-to-month.

dimaqq:
Juju 4 funnily enough showed me why the tests are failing.
- in my local, the built ops gets added to charm's pyproject.toml and uv.lock
- in gha actions, the build ops gets added to repo root pyproject.toml and uv.lock

I'm a bit at a loss why that happens, but at least I now know what happens.

test/smoke/test_smoke.py
dimaqq: No, it's not I was trying to figure out what I messed up.

.github/workflows/smoke.yaml
dimaqq: Sorry I should have marked this up.
While the PR is a draft, I'm adding this trigger to test what I'm doing. This bit needs to be reverted before the PR is merged.

dimaqq: I've cleaned this up now.

test/smoke/test_smoke.py
dimaqq: Reverted, thanks.

tox.ini
dimaqq: Done, thanks.

test/smoke/test_smoke.py
```diff
@@ -82,10 +84,14 @@ def test_smoke(juju: jubilant_backports.Juju, base: str):
...
 @pytest.fixture(scope='module')
-def juju():
+def juju(request: pytest.FixtureRequest):
     with jubilant_backports.temp_model() as juju:
         yield juju
 
+        if request.session.testsfailed:
+            log = juju.debug_log(limit=1000)
+            print(log, end='')
+
```
dimaqq, line 94: I'm not totally happy with this: https://github.com/canonical/jubilant/discussions/182

Then again, having this added and running against Juju 4 (but not Juju 3) showed me what was wrong in my PR when the other, the above block in this file got lost in a bad merge. So I figure it's better to keep it.

.github/workflows/smoke.yaml
```diff
@@ -10,6 +10,7 @@ permissions: {}
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 20  # Juju 4 sometimes gets stuck destroying the test model
```
dimaqq: 🤔 weirdly this did not work:
- https://github.com/canonical/operator/actions/runs/16910908904/job/47912015547
- I've manually cancelled the workflow when Juju 4 / microk8s had been running for 33 minutes.

tox.ini
```diff
@@ -179,19 +179,19 @@ allowlist_externals =
...
 dependency_groups = integration
 commands =
     # Clean any stale artefacts, in case these steps are run locally and repeatedly.
-    bash -c 'rm -vf test/charms/test_smoke/*.tar.gz'
-    bash -c 'rm -vf test/charms/test_smoke/*.whl'
-    bash -c 'rm -vf test/charms/test_smoke/requirements.txt'
+    bash -c "find test/charms/test_smoke -name '*.whl' -or -name '*.tar.gz' -or -name pyproject.toml -or -name uv.lock | xargs -r rm -vf"
     # Build ops and friends from this checkout.
-    uv build --all --out-dir=test/charms/test_smoke/
+    uv build --out-dir=test/charms/test_smoke --wheel .
```
tonyandrewmeyer, line 189: I assume the removal of `--all` is because the smoke test only actually needs ops, not ops-tracing and certainly not ops-scenario. If so, then probably the comment above (the "and friends") should be updated too.

test/charms/test_smoke/.gitignore
```diff
@@ -0,0 +1,5 @@
+pyproject.toml
+uv.lock
+*.tar.gz
+*.whl
+*.charm
```
tonyandrewmeyer: `.tar.gz` and `.charm` would be ignored anyway from the top-level ignore file, and it seems reasonable to add `.whl` there as well, and then only have the unique ignores here.

If you'd rather have something that's more self-contained leaving as-is works for me too.

test/charms/test_smoke/reference-pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+name = "test-smoke"
+version = "0.1.0"
+description = "Add your description here"
```
tonyandrewmeyer: It feels like we could shove a description in here or remove this field (if I'm remembering correctly that description is optional).

tonyandrewmeyer (approved):
Looks great, thanks!

.github/workflows/smoke.yaml
dimaqq: I've figured this out:
Workflow-wide timeout is only used by self-hosted runners.
So I've switched to timeout in the key step.

PR #1967, opened 2025-08-08
chore: upgrade pyright

### LINKED ISSUE #1953: Upgrade pyright and fix type hints
Ops 3.x dropped py 3.8 which means we can unpin pyright and use the latest.

That produces a bunch of errors though:

```py
(ops) 🦐/c/operator (chore-unpin-pyright)> tox -e static
static: uv-sync> uv sync --locked --python-preference system --no-default-groups --group static --group unit -p /Users/dima/.local/share/uv/tools/tox/bin/python3
static: commands[0]> pyright
/code/operator/ops/_private/harness.py
  /code/operator/ops/_private/harness.py:2333:5 - error: Type of "grants" is partially unknown
    Type of "grants" is "dict[Unknown, Unknown]" (reportUnknownVariableType)
  /code/operator/ops/_private/harness.py:2334:5 - error: Type of "user_secrets_grants" is partially unknown
    Type of "user_secrets_grants" is "set[Unknown]" (reportUnknownVariableType)
/code/operator/ops/jujucontext.py
  /code/operator/ops/jujucontext.py:64:5 - error: Type of "debug_at" is partially unknown
    Type of "debug_at" is "set[Unknown]" (reportUnknownVariableType)
/code/operator/ops/model.py
  /code/operator/ops/model.py:3236:9 - error: Overload 1 for "exec" overlaps overload 2 and returns an incompatible type (reportOverlappingOverload)
  /code/operator/ops/model.py:4376:5 - error: Type of "attributes" is partially unknown
    Type of "attributes" is "dict[Unknown, Unknown]" (reportUnknownVariableType)
  /code/operator/ops/model.py:4384:5 - error: Type of "redacted" is partially unknown
    Type of "redacted" is "list[Unknown]" (reportUnknownVariableType)
  /code/operator/ops/model.py:4422:5 - error: Type of "ca_certificates" is partially unknown
    Type of "ca_certificates" is "list[Unknown]" (reportUnknownVariableType)
/code/operator/ops/pebble.py
  /code/operator/ops/pebble.py:1666:5 - error: Type of "last_data" is partially unknown
    Type of "last_data" is "dict[Unknown, Unknown]" (reportUnknownVariableType)
  /code/operator/ops/pebble.py:2913:9 - error: Overload 1 for "exec" overlaps overload 2 and returns an incompatible type (reportOverlappingOverload)
  /code/operator/ops/pebble.py:3200:13 - error: Type of "process" is partially unknown
    Type of "process" is "ExecProcess[Unknown]" (reportUnknownVariableType)
/code/operator/test/test_charm.py
  /code/operator/test/test_charm.py:1705:5 - error: Type of "my_list" is partially unknown
    Type of "my_list" is "list[Unknown]" (reportUnknownVariableType)
/code/operator/test/test_model_relation_data_class.py
  /code/operator/test/test_model_relation_data_class.py:137:5 - error: Type of "baz" is partially unknown
    Type of "baz" is "list[Unknown]" (reportUnknownVariableType)
/code/operator/test/test_pebble.py
  /code/operator/test/test_pebble.py:3498:49 - error: Type of "excinfo" is partially unknown
    Type of "excinfo" is "ExceptionInfo[ExecError[Unknown]]" (reportUnknownVariableType)
  /code/operator/test/test_pebble.py:4033:23 - error: Argument of type "dict[str, IdentityAccess | LocalIdentityDict]" cannot be assigned to parameter "d" of type "IdentityDict" in function "from_dict"
    Type "Literal[IdentityAccess.ADMIN]" is not assignable to type "Literal['untrusted', 'metrics', 'read', 'admin']"
      "Literal[IdentityAccess.ADMIN]" is not assignable to "Literal['untrusted']"
      "Literal[IdentityAccess.ADMIN]" is not assignable to "Literal['metrics']"
      "Literal[IdentityAccess.ADMIN]" is not assignable to "Literal['read']"
      "Literal[IdentityAccess.ADMIN]" is not assignable to "Literal['admin']" (reportArgumentType)
  /code/operator/test/test_pebble.py:4058:23 - error: Argument of type "dict[str, IdentityAccess | BasicIdentityDict]" cannot be assigned to parameter "d" of type "IdentityDict" in function "from_dict"
    Type "Literal[IdentityAccess.METRICS]" is not assignable to type "Literal['untrusted', 'metrics', 'read', 'admin']"
      "Literal[IdentityAccess.METRICS]" is not assignable to "Literal['untrusted']"
      "Literal[IdentityAccess.METRICS]" is not assignable to "Literal['metrics']"
      "Literal[IdentityAccess.METRICS]" is not assignable to "Literal['read']"
      "Literal[IdentityAccess.METRICS]" is not assignable to "Literal['admin']" (reportArgumentType)
/code/operator/test/test_real_pebble.py
  /code/operator/test/test_real_pebble.py:171:49 - error: Type of "excinfo" is partially unknown
    Type of "excinfo" is "ExceptionInfo[ExecError[Unknown]]" (reportUnknownVariableType)
  /code/operator/test/test_real_pebble.py:188:49 - error: Type of "excinfo" is partially unknown
    Type of "excinfo" is "ExceptionInfo[ExecError[Unknown]]" (reportUnknownVariableType)
/code/operator/test/test_testing.py
  /code/operator/test/test_testing.py:6405:49 - error: Type of "excinfo" is partially unknown
    Type of "excinfo" is "ExceptionInfo[ExecError[Unknown]]" (reportUnknownVariableType)
18 errors, 0 warnings, 0 informations
static: exit 1 (3.04 seconds) /code/operator> pyright pid=12326
  static: FAIL code 1 (3.06=setup[0.02]+cmd[3.04] seconds)
  evaluation failed :( (3.09 seconds)
```

Tried with `pyright==1.1.403` which was latest when I wrote this.

Probably a great Friday item.

dimaqq:
Since we've dropped Python 3.8, we can use latest pyright again.

Mostly, it just has gotten stricter.

But there are a couple of bits that are worth reviewing, see my comments.

Fixes #1953


ops/_private/harness.py
```diff
@@ -2330,8 +2330,8 @@ class _Secret:
     label: str | None = None
     description: str | None = None
     tracked: int = 1
-    grants: dict[int, set[str]] = dataclasses.field(default_factory=dict)
-    user_secrets_grants: set[str] = dataclasses.field(default_factory=set)
+    grants: dict[int, set[str]] = dataclasses.field(default_factory=dict[int, set[str]])
+    user_secrets_grants: set[str] = dataclasses.field(default_factory=set[str])
```
dimaqq, line 2334: This is a typical example of pyright strict mode getting stricter.

ops/model.py
```diff
@@ -3269,7 +3269,7 @@ def exec(
         stdin: bytes | BinaryIO | None = None,
         stdout: BinaryIO | None = None,
         stderr: BinaryIO | None = None,
-        encoding: None = None,
+        encoding: None,
```
dimaqq, line 3272: We were doing it wrong before.

2 overloads must be distinguishable, while we had default value (different ones) in two overloads, and that was the only different field.

test/test_pebble.py
```diff
@@ -4030,7 +4030,7 @@ def test_local_identity_from_dict(self):
 
     def test_local_identity_from_dict_with_access_enum(self):
         identity = pebble.Identity.from_dict({
-            'access': pebble.IdentityAccess.ADMIN,
+            'access': pebble.IdentityAccess.ADMIN,  # type: ignore
```
dimaqq, line 4034: This is of potential concern to our users.

If we want them to call `Identity.from_dict` both with strings and enum members and have that type check strictly, we may have to do some extra work.

For now, I've decided not to adjust our public types, not within the scope of this PR anyway.

pyproject.toml
```diff
@@ -51,7 +51,7 @@ docs = [
 ]
 static = [
     "ops[testing,tracing]",
-    "pyright==1.1.385",
+    "pyright~=1.1",
```
tonyandrewmeyer, line 54: I believe the reason we pinned this one that tightly was because the rules change within whatever "1.1" means - all of the improvements you have in this PR for example. We didn't want to have those randomly forced into unrelated PRs just because someone happened to be freshly installing pyright and so got an update.

I think we could probably reconsider this now, because (a) uv.lock makes changing more deliberate, and (b) we have a much larger team so it's easier to throw in a small PR to bump pyright when needed.

ops/model.py
tonyandrewmeyer: I don't understand this. One version has a string that defaults to utf8 and one has None that defaults to None. Those are definitely distinguishable, and this definitely worked - if you pass in encoding=None type checkers (at least pyright and ty) know it's giving back bytes, and if not then it's string.

Also, I get that this is the overload, but we have an argument with no default after arguments with defaults, how can that be correct? These are kwargs so that doesn't have to be the case.

test/test_pebble.py
tonyandrewmeyer: Is this because the type checker isn't ok with a str subclass when only a small set of literals is acceptable? If the typed dict accepted a string or the enum would that fix it? That seems a safe change, and more correct. I think we should fix it now, even if not in this PR.

test/test_pebble.py
```diff
@@ -3494,8 +3494,9 @@ def test_wait_exit_nonzero(self, client: MockClient):
         self.add_responses(client, '456', 1)
 
         process = client.exec(['false'])
+        # https://github.com/microsoft/pyright/discussions/10774
         excinfo: pytest.ExceptionInfo[pebble.ExecError[str]]
-        with pytest.raises(pebble.ExecError) as excinfo:
+        with pytest.raises(pebble.ExecError) as excinfo:  # type: ignore
```
tonyandrewmeyer, line 3499: I've hit this a lot. I think I've talked about it with @james-garner-canonical before. There are also a few issues in the pytest repo about this, and about exception groups which have basically the same issue (but I think those specifically got fixed). I'm fine with the ignore for now.

ops/model.py
james-garner-canonical: I believe that dropping the default for this case is correct, since the real default is 'utf-8' -- so we only want this overload to be detected if `None` is explicitly provided. Normally having a parameter without a default following one or more parameters with a default is a syntax error, but apparently not with keyword only arguments.

test/test_pebble.py
james-garner-canonical: I wonder if adding a type annotation to `IdentityAccess.ADMIN` (`Literal['admin']`) etc would resolve this.

test/test_pebble.py
james-garner-canonical: This seems to be a specific problem with using `pytest.raises` with `Generic` exceptions.

I wonder if we *could* work around it by defining our own `AnyStr` `TypeVar` with a default. Requires importing `TypeVar` from `typing_exceptions` since defaults are added in 3.11, and would look like `AnyStr = TypeVar('AnyStr', str, bytes, default=bytes)` (or whatever the most sensible default is).

But I don't really think that would be worth it, I'm ok with the `ignore` for now too.

james-garner-canonical (approved):

james-garner-canonical:
Approved, but the one thing I do think we should look into further before merging is whether adding `Literal` type annotations to the fields of `pebble.IdentityAccess` (`ADMIN`, etc) improves the user experience.

IronCore864 (approved):

ops/model.py
dimaqq: > One version has a string that defaults to utf8 and one has None that defaults to None.

Then, which of the two overloads should be picked when no value is specified?

test/test_pebble.py
dimaqq: It's not possible to add type annotations to enum members.

+1 on making the user-facing change in a separate PR, with pros and cons discussion on that separate PR.

I've made #1977 for that.

pyproject.toml
dimaqq: Before uv, we pinned things because we didn't have a lock file.

Then we pinned `pyright==1.1.385` specifically to support Python 3.8.

Now we have a lock file, we could perhaps rely on the lock file for exact versions and show our intention in pyproject.toml, WDYT?

I'm with you to reconsider past practice because:
- We're now upgrading from 9-month-old pyright and the diff is only 20 LOC, the rest being the lock file, etc.
- Tools like dependabot can poke us to re-lock and the unit tests are run on those PR automatically, so there's no work for us unless code changes are needed and those PRs can be kept separate from human/feature PRs.

test/test_pebble.py
dimaqq: I've asked upstream too, https://github.com/microsoft/pyright/discussions/10774

pyproject.toml
benhoyt: Per discussion, this will only be re-locked if someone is adding/removing a dependency to Ops (which is a big deal), so we're going to go with the looser constraint `pyright~=1.1`.

We could do similar with Ruff and codespell (looser constraint but uv.lock'd) in a separate PR.

Issue #1783, opened 2025-06-04
Update docs: self.model.app --> self.app


dimaqq:
### Checklist
- [ ] rst and  md docs
- [ ] docstring in Python code
- [ ] `self.framework.model` -> `self.model`
- [ ] `self.model.app` -> `self.app`

### Context

https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$fWDmynlVnJzvETqnHPUk8JtuWRrz-UJ7NFZM8KQ7GRE?via=ubuntu.com&via=matrix.org&via=laquadrature.net
> 
> I've started reading the charm that I cobbled together, and I realised that I have a mixture of:
> 
> - `self.app` (in relations code)
> - `self.model.app` (in status code)
> 
> So I went down a rabbit hole of reading all those `@property` definitions in ops and realised that there's more~
> 
> - `self.framework.model.app`
> 
> Should we perhaps decide on one of these and update all the docs?

replies:

> I do think it would be good to replace all the non-shortcut ones (the ones via model) in the docs with the shortcut versions.

> Sounds like an excellent idea to me

tonyandrewmeyer:
When this is done, I'd suggest doing `self.config`, `self.unit`, `self.meta`, and `self.charm_dir` as well.

dwilding:
I searched for `framework.charm_dir`, `framework.meta`, and `framework.model` in the docs. The only result was this example in [How to manage libraries](https://documentation.ubuntu.com/ops/latest/howto/manage-libraries/):

```py
class DatabaseReadyEvent(ops.EventBase):
    """Event representing that the database is ready."""

    def __init__(self, handle: ops.Handle, *, credential_secret: ops.Secret):
        super().__init__(handle)
        self.credential_secret = credential_secret

    def snapshot(self) -> dict[str, str]:
        data = super().snapshot()
        data['credential_secret_id'] = self.credential_secret.id
        return data

    def restore(self, snapshot: dict[str, Any]):
        super().restore(snapshot)
        credential_secret_id = snapshot['credential_secret_id']
        self.credential_secret = self.framework.model.get_secret(id=credential_secret_id)
```

Per Dima's note in the issue description, I don't believe that `self.framework.model` should be changed to `self.model`. `DatabaseReadyEvent` is based on `ops.EventBase`, which doesn't have a `model` attribute.

So I don't plan to make any updates for `framework.charm_dir`, `framework.meta`, or `framework.model`.

PR #1968, opened 2025-08-08
docs: use classes from ops instead of ops.<submodule>

dwilding:
This PR finishes work on #1783. There are a few places in old docs where we're using classes from `ops.<submodule>` instead of directly from `ops`.

Drive-by: Pre-commit removed some whitespace. Best to hide whitespace when reviewing.

### DIFF
--- a/docs/howto/legacy/write-legacy-unit-tests-for-a-charm.md
+++ b/docs/howto/legacy/write-legacy-unit-tests-for-a-charm.md
@@ -66,7 +66,7 @@ When using `Harness.begin()` you are responsible for manually triggering events 
 ...
 # Add a relation and trigger relation-created.
 harness.add_relation('db', 'postgresql') # <relation-name>, <remote-app-name>
-# Add a peer relation and trigger relation-created 
+# Add a peer relation and trigger relation-created
 harness.add_relation('cluster', 'test-app') # <peer-relation-name>, <this-app-name>
 ```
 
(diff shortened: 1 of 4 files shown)


docs/howto/manage-libraries.md
```diff
@@ -79,7 +79,7 @@ class DatabaseReadyEvent(ops.EventBase):
 
 class DatabaseRequirerEvents(ops.ObjectEvents):
     """Container for Database Requirer events."""
-    ready = ops.charm.EventSource(DatabaseReadyEvent)
+    ready = ops.EventSource(DatabaseReadyEvent)
```
dwilding, line 82: Before my change, this was a bug, wasn't it? Should have been `ops.framework.EventSource` I think.

dimaqq:
```
>>> import ops
>>> ops.charm.EventSource
<class 'ops.framework.EventSource'>
```

There's a re-export apparently.
In any case, good to fix this.

docs/howto/legacy/write-legacy-unit-tests-for-a-charm.md
```diff
@@ -92,7 +92,7 @@ harness.disable_hooks()
 harness.update_config({"foo": "quux"})
 # Re-enable hooks
 harness.enable_hooks()
-# Set the status of the active unit. We'd need "from ops.model import BlockedStatus".
+# Set the status of the active unit. We'd need "from ops import BlockedStatus".
```
tonyandrewmeyer: Maybe we don't care because this is legacy, but we would generally recommend the "ops." be in the following line, and you just do "import ops".

tonyandrewmeyer (approved):
Thanks!

docs/howto/legacy/write-legacy-unit-tests-for-a-charm.md
dwilding: Ah thanks, I get what you mean. And since `ops` is already imported in the first example in the doc, I feel we don't need to mention it again. I've changed the code to this:

```python
# Set the status of the active unit
harness.charm.unit.status = ops.BlockedStatus("Testing")
```

PR #1971, opened 2025-08-11
docs: add Google Analyics integration and cookie consent banner

dwilding:
Now that the Ops docs are under documentation.ubuntu.com, we need to enable Google Analytics and the cookie consent banner. This PR matches https://github.com/canonical/pebble/pull/652. See that Pebble PR for additional context.


tonyandrewmeyer:
Both of these points may be things we're stuck with because everything is consistent across Canonical, but:

The layout isn't great here - couldn't the checkboxes be on the same vertical line as the headings? This is Chrome on Ubuntu.

image

I realise this goes against the preference I espoused to you last week, but I think it would be better if the cookie policy link opened in a new window/tab.

dwilding:
> I'm also guessing that the two UI elements I mentioned would have to be improved elsewhere.

Thanks for raising those - I've created https://github.com/canonical/RTD-cookie-banner-integration/issues/16 and https://github.com/canonical/RTD-cookie-banner-integration/issues/17 to track them.

PR #1972, opened 2025-08-11
chore: merge the static checks into the lint command

tonyandrewmeyer:
Adjust the tox commands to align with [OP061](https://docs.google.com/document/d/1GfOTT1Ir-pLAbILUrI4GS9T8AAI8Ni8gpF1Mh67Wx3E/edit?usp=sharing) by moving the static type checking into the general "lint" command.

* Updates the top-level `tox` so that pyright is run as part of `tox -e lint`, and the `static` env is removed.
* Similarly updates the `httpbin-demo` charm.
* Similarly updates the `ops-tracing` package.

Also re-locks the dependencies because there is no longer a `static` group (it's merged into `lint`).


examples/httpbin-demo/tox.ini
```diff
@@ -32,20 +32,13 @@ pass_env =
...
 
-[testenv:static]
-description = Run static type checks
-deps =
-    pyright
-    -r {tox_root}/requirements.txt
-commands =
-    pyright {posargs}
-
 [testenv:unit]
```
dwilding, line 51: I'm a bit confused why the static checks have moved into the `unit` env. Shouldn't they go in a dedicated `lint` env instead?

I see that `env_list` specifies `lint` and `format`, neither of which actually exist now. I think I removed those envs from tox.ini (when I migrated the charm from Charmcraft) because I wanted to inherit the formatting and linting rules from the Ops repo. But I must have forgotten to remove those envs from `env_list`.

If we keep the static checks, but move them to `lint`, I guess we should update the example-charm-tests workflow to run `tox -e lint` in addition to `tox -e unit`. Maybe as a separate PR?

.github/workflows/tiobe.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Install TIOBE and project dependencies
         run: |
-          uv export --no-emit-project --frozen --no-hashes --group unit --group static --group release > requirements.txt
+          uv export --no-emit-project --frozen --no-hashes --group unit --group release > requirements.txt
```
dwilding: I don't fully understand this part. Since the deps are moving from the `static` group to the `lint` group, doesn't the `lint` group need to be listed here?

.github/workflows/framework-tests.yaml
```diff
@@ -19,27 +19,12 @@ jobs:
...
       - run: uv tool install tox --with tox-uv
-      - name: Run linting
+      - name: Run linting and static checks
         run: tox -e lint
 
-      - name: Run tracing linter
+      - name: Run tracing linter and static checks
         run: cd tracing && tox -e lint
 
-  static:
```
dwilding, line 28: In the PR checks, I see one pending required check called `static`. I think we also need to remove `static` as a required check in the repo's rulesets.

examples/httpbin-demo/tox.ini
tonyandrewmeyer: Ah, that's a mix of me not looking carefully enough and the existing `tox.ini` saying that there is a `format` and `lint` environment but actually neither exist.

It's not obvious what the right thing is here. We could just leave it unchanged, but when the tutorials are updated to use the charmcraft init defaults, they'll have a `lint` that does the static checking, so we would want that to be kept (although they'll also have `format`).

I guess the best thing to do would be to add a `lint` environment.

.github/workflows/framework-tests.yaml
tonyandrewmeyer: Yes, I'll do that after the PR is merged (or at least after it's accepted, immediately prior to merge).

.github/workflows/tiobe.yaml
tonyandrewmeyer: Hmm, it might be yes. I think TIOBE complains if it can't find pyright, and that's why I had to add this in the first place.

.github/workflows/example-charm-tests.yaml
```diff
@@ -35,6 +35,10 @@ jobs:
           python-version: '${{ matrix.python-version }}'
       - name: Install tox
         run: pip install tox~=4.2
+      - name: Run linting and static checks
```
dwilding, line 38: Nice, thanks. I suppose the name `charm-unit-tests` on line 13 ought to change, now that we're running more tests

examples/k8s-1-minimal/tox.ini
```diff
@@ -32,6 +32,9 @@ pass_env =
     CHARM_BUILD_DIR
     MODEL_SETTINGS
 
+[testenv:lint]
+description = Dummy environment for now
```
dwilding, line 36: Are you thinking to leave this as a placeholder until we base the tutorial on the Charmcraft profile? I'd be OK with that

While we're thinking about linting, are you happy with what the profile provides?
https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/tox.ini.j2#L34-L45

dwilding (approved):
Thanks for doing this :slightly_smiling_face:

examples/k8s-1-minimal/tox.ini
tonyandrewmeyer: > Are you thinking to leave this as a placeholder until we base the tutorial on the Charmcraft profile? I'd be OK with that

Yes, that seemed simpler than putting in something conditional into the workflow.

> While we're thinking about linting, are you happy with what the profile provides? https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/tox.ini.j2#L34-L45

I have a [PR open there to do the same](https://github.com/canonical/charmcraft/pull/2405).

dwilding: Ah yes, a PR that I approved :sweat_smile:

.github/workflows/example-charm-tests.yaml
tonyandrewmeyer: Sure, will do.

dimaqq:
P.S. I recall the discussion about lint and static, but I missed the memo of them getting merged.

It's rather likely that charmers did too, so let's socialise this change (in the spec, in our repo, and in the recommended charming practice).

tonyandrewmeyer:
> P.S. I recall the discussion about lint and static, but I missed the memo of them getting merged.

Yeah, [looks like you were off that day, sorry](https://docs.google.com/document/d/1x4SW63T09iW6wcd5cNAACdmEw9AzL7qmial3lgnDVkI/edit?tab=t.0#heading=h.7sh2szkhn6z2). It was the most divisive aspect of the spec, and Jon pushed for combining, and with significantly faster type checking on the way, the strongest argument against it being separate was a lot weaker.

> It's rather likely that charmers did too, so let's socialise this change (in the spec, in our repo, and in the recommended charming practice).

Yes, this is in progress (CHARMTECH-815). I've done the PR to change charmcraft 4 to combine them, now I'm working through the Charm Tech tools, starting with charmlibs (thanks @james-garner-canonical), now operator, but also doing Jubilant and the other as well. Then working through some PRs for the wider community. Our [docs](https://documentation.ubuntu.com/ops/latest/howto/manage-charms/#develop-your-charm) avoid mentioning specific commands other than `list` and `-e integration`, so will update once charmcraft does, and the tutorial will update when it's updated to use `charmcraft init` as the starting point.

examples/httpbin-demo/tox.ini
```diff
@@ -32,12 +32,18 @@ pass_env =
...
 
-[testenv:static]
-description = Run static type checks
+[testenv:lint]
+description = Run linting and type checks
 deps =
+    ruff
+    codespell
     pyright
+    ops[testing]
```
dwilding, line 41: I _think_ this isn't actually achieving anything, because pyright is only configured to check the charm code. From `pyproject.toml`:

```toml
[tool.pyright]
include = ["src/*.py"]
```

Shall we have pyright check `tests/unit` too? I've verified that `tox -e lint` passes in this case.

If https://github.com/canonical/operator/pull/1987 merges before this PR, I think we'd be safe to have pyright check `tests/integration` too, after adding pytest and jubilant as deps of `lint`. (If I understand everything right...)

tonyandrewmeyer: (Discussed this in a call just now, and agreed that this is a good approach, and I'll wait until Dima's work is done.)

examples/httpbin-demo/pyproject.toml
```diff
@@ -1,3 +1,3 @@
 # Static checks configuration
 [tool.pyright]
-include = ["src/*.py"]
+include = ["src/*.py", "tests/unit/*.py", "tests/integration/*.py"]
```
dwilding: The Charmcraft profile uses:

```python
include = [
    "src/**/*.py",
    "tests/**/*.py",
]
```

See https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/pyproject.toml.j2#L73-L76

Do you have a preference for format? I figure it would be nice to use the same format in both places if possible

james-garner-canonical: Personally I like:
```toml
include = ["src", "tests"]
```

tonyandrewmeyer: I have no preference at all in this particular case, so someone propose a change and I'll go with whatever it is :)

dwilding: I vote for James's suggestion.

tonyandrewmeyer:
```suggestion
include = ["src", "tests"]
```

PR #1975, opened 2025-08-11
feat: support for generating charmcraft YAML from Python classes

tonyandrewmeyer:
A new package, `ops-tools` is added. This currently only provides a script to update `charmcraft.yaml` based on config and action Python classes, but is designed to also offer other tools in the future. It is released simultaneously with the other `ops-*` packages, but is not required by any of the others, or offered as an extra.

Charmers can use the exported functions (which take a Python class object and return a dictionary suitable for serialisation to YAML), but are expected to use the provided script (that works with a `charmcraft.yaml` file and Python modules).

Classes can provide a `to_juju_schema()` method if they need to provide YAML in a different way. The script will pass the generated config/action as a base to optionally work from.

The implementation is designed to work with the `load_config` and `load_params` functionality in ops. In particular, that means working with four types of class:

* Pydantic dataclasses

Only reference documentation is included in this PR ([Preview](https://canonical-ubuntu-documentation-library--1975.com.readthedocs.build/ops/1975/reference/ops-tools/)). The intention is that there will be a follow-up PR that includes at least how-to documentation.

More details are available in the [spec](https://docs.google.com/document/d/1vj2pTvVRXClCrYM7tUjft3ngISSsFdBQD1B-CzOG1Vs/edit?usp=sharing). This PR replaces #1702.


testing/src/scenario/state.py
```diff
@@ -1975,6 +1975,10 @@ def autoload(charm_type: type[CharmBase]) -> _CharmSpec[CharmType]:
             # try to load using legacy metadata.yaml/actions.yaml/config.yaml files
             meta, config, actions = _CharmSpec._load_metadata_legacy(charm_root)
 
+        # TODO: ideally, we would look for ConfigBase classes in the charm
+        # module and autoload from there at this point. Leaving this until the
+        # conversation about if & how the generation is done is resolved.
+
```
tonyandrewmeyer: My feeling at the moment is that this should be left for future work, rather than baking anything in to Scenario at this time. There is a simple example of using this functionality in tests for both config and actions in the Scenario tests - I think those show that it's not too arduous to use at the moment, without anything else.

tools/src/ops_tools/_attrdocs.py
```diff
@@ -0,0 +1,103 @@
...
+        last_attr = None
+        for child in node.body:
+            if isinstance(child, (ast.Assign, ast.AnnAssign)):
+                target = None  # Make the type checker happy.
+                if isinstance(child, ast.Assign):
+                    target = child.targets[0]
+                elif isinstance(child, ast.AnnAssign):
+                    target = child.target
+                assert isinstance(target, ast.Name)
+                last_attr = target.id
```
dimaqq, line 56: I would prefer to see unit tests for this class, the top-level function below in addition to the "functional" tests for YAML generation.

I've been bitten by AST before 🐍

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,102 @@
...
+            actions.update(action_to_juju_schema(cls))
+    actions = dict(sorted(actions.items()))  # Sort actions by name.
+
+    if config:
+        charmcraft_yaml['config'] = config
+    if actions:
+        charmcraft_yaml['actions'] = actions
+
+    with open(args.charmcraft_yaml, 'w') as raw:
+        yaml.safe_dump(charmcraft_yaml, raw)
```
james-garner-canonical: Do you think rewriting the `charmcraft.yaml` file (presumably stripping comments, etc) is the way to go? What about writing to `actions.yaml` and `config.yaml`?

tools/src/ops_tools/_attrdocs.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.debug('No source code found for %s', schema_class.__name__)
+            continue
+        try:
+            tree = ast.parse(source_code)
+        except (SyntaxError, IndentationError):
+            logger.debug('Failed to parse source code for %s', schema_class.__name__)
+            continue
+        extractor = AttributeDocstringExtractor()
+        extractor.visit(tree)
+        docs.update(extractor.attribute_docs)
```
james-garner-canonical: What's this part for, and is it exercised in tests? I can see that it's extracting attribute docstrings, but I'm surprised it clobbers both `model_fields` and `dataclasses.fields`. It also seems unusual that we update the `docs` dict in `mro` order, wouldn't that make docstrings in parent classes overwrite those in child classes?

tonyandrewmeyer: It's exercised in at least one test, because I added it to fix a test. It is only for a single inheritance layer (plus object). If your config/action (non-Pydantic) class inherits from another class, we need to get descriptions from the parent as well, or they'll be missing from the generated YAML.

It was clobbering `model_fields` and `dataclasses.fields` before the `mro` part was there, even in the original. I'm not sure what's correct here, although it seems unlikely that you would have a class with clobbering. I guess if there is a Pydantic description then it should win - should we *only* look for Pydantic descriptions if `model_fields` exists? And the same for Pydantic dataclasses? I could be convinced of that.

However, you're right about the order being backwards. In the test (and probably any real use) it's additive rather than replacing attributes, so I missed that.

tools/src/ops_tools/_update_charmcraft_yaml.py
tonyandrewmeyer: I'd like to do that, in a `charmcraft` part. However, that's not possible right now, because of when the separate YAMl files are created. The deal was that I would get this working without any charmcraft changes and then go back to them to advocate for a nicer UX with changes.

(I could double-check this - it's my recollection of the discussions but maybe there is hackery that would make it work).

I don't like the idea of use of this tool forcing charmers to go backwards and use individual config and action files rather than the unified one.

I don't like the invasive editing pyyaml does (I'm more sad about ordering than comments), but I couldn't find any way to avoid it. I could switch to a different yaml library that does this - I guess we don't need to be too strongly influenced by ops, since it's a separate tool, although keeping the team's dependency count down as a whole is desirable.

Another option that has been discussed is instead of editing charmcraft.yaml like this, it could diff the config and actions and pretty print the result and exit non-zero if there's a difference. So it's more of a linter for now, and could be used to manually copy over differences if desired. Would you prefer that?

dimaqq: If we want folks to use this feature, we could either:
- advocate that charmers abandon separate config/actions/metadata.yaml (my pref), or
- support updating both charmcraft.yaml and metadata.yaml, and autodetect which is present

P.S. I'm all for updating "in place", however we must use a round-trip safe yaml parser/serialiser and have tests to prove it, including YAML with comments in "inconvenient" places. I've run into that with the charm pin updater script, there are plenty of corner cases when key or content is interspersed with comments.

dimaqq:
> * Pydantic dataclasses
> * Pydantic 2.x BaseModel classes

IIRC these two actually differ in minutae when class is instantiated.

It's something along the lines of:
- pydantic dataclass can be instantiated with wrong types, doesn't automatically raise ValueError (?)
- pydantic BaseModel coerces types, for example either str --> int or int --> str, sorry I don't recall which way, I think each field applies own type as "constructor" with raw value as the parameter

A decent set of tests is needed here, even if it feels like we're testing Pydantic and now our code.

P.S.
Refs:
https://docs.pydantic.dev/latest/concepts/conversion_table/
https://docs.pydantic.dev/latest/concepts/strict_mode/
https://docs.pydantic.dev/latest/concepts/types/#strict-types
(the last feels like advice to the charmer, rather than something we can do on our end; but we do need to cover these with tests)

tonyandrewmeyer:
> IIRC these two actually differ in minutae when class is instantiated.

None of the classes are actually instantiated in the YAML generation.

> A decent set of tests is needed here, even if it feels like we're testing Pydantic and now our code.

There are tests for both pydantic dataclasses and BaseModel subclasses. For BaseModel with actions, I don't think we need much more than what we have, because the bulk of the work is done by pydantic rather than this code. For the others, if you'd like to suggest specific cases to check I'm happy to add additional tests.

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,427 @@
...
+    required_params.sort()
+    return params, required_params
+
+
+def config_to_juju_schema(cls: type[object]) -> dict[str, dict[str, OptionDict]]:
+    """Translate the class to YAML suitable for charmcraft.yaml.
+
+    For example, with the class::
+
+        class MyConfig(pydantic.BaseModel):
```
benhoyt: What would it take to make these doctests, that is, tested examples?

tools/README.md
```diff
@@ -0,0 +1,90 @@
...
+The definition of charm configuration options and actions should have a single source of truth. Our preference is that truth is in the charm code: this is in an expressive language (Python), that allows more complex type definitions and validation than possible in the Juju config schema language or action parameter JSONSchema. However, Juju and Charmcraft need to find these schema in the `charmcraft.yaml` file.
+
+The package provides tooling to generate the appropriate `charmcraft.yaml` sections for `config` and `actions` from config and action classes in the charm Python code.
+
+### Usage
+
+For example, if your charm contains a single config class in `src/charm.py` called `Config` and three actions, which all end with 'Action' (and no other classes have that name):
+
+```bash
+update-charmcraft --config-class Config --action-class '.+Action'
```
benhoyt: `update-charmcraft` seems like a very generic name. Should we include "schema", like `update-charmcraft-schema`?

tools/README.md
```diff
@@ -0,0 +1,90 @@
+# Ops Tools
```
benhoyt, line 1: Should this move to a proper docs page instead? I see we have a reference page for [`ops_tools`](https://canonical-ubuntu-documentation-library--1975.com.readthedocs.build/ops/1975/reference/ops-tools/#module-ops_tools), but not for the script.

I notice it's listed as `ops_tools` -- should it be namespaced as `ops.tools` like our other packages?

Also, do we need "how to install it"? I'd be fine if we just recommend `uvx ops.tools.update-charmcraft` or whatever, but I think we should at least mention it.

benhoyt (approved):
Couple of relatively minor comments, but approving optimistically. I haven't reviewed the code line-by-line (I did that in the previous incantation, and I'm also less concerned as it's an offline tool).

tools/src/ops_tools/_generate_yaml.py
tonyandrewmeyer: I love that idea. I haven't used doctest in ages, but still really like the concept. I'll make it so.

tools/README.md
tonyandrewmeyer: Yeah, my intention is to do a follow-up PR with actual documentation, and a bunch of what's in the README probably goes into a how-to guide I think. I'm not sure if the script should be in the reference section or not, but hammering that out in the doc PR would work I think.

james-garner-canonical: I was thinking about this as well. Currently, it's very different to the other packages, but rightfully so -- they're available under `ops.testing/tracing`, at testing/runtime, while these tools aren't intended to be.

Is a namespace package possible here? (Distribute as `ops-tools`, import as `ops.tools`, maybe do an extra ...).

I'm not sure it's worth shimming them in the same way we did `testing` and `tracing` (and then presumably distributing the package as `ops-tools` and making it available as an extra as `ops[tools]`).  But maybe it is? Consistency is nice, our users are familiar with this approach, and Tiexin's release script makes the version bumping and dependencies pretty painless for releases.

tonyandrewmeyer: > I notice it's listed as `ops_tools` -- should it be namespaced as `ops.tools` like our other packages?

`ops` can't be a namespace package, because it's a regular Python package. So if we want to put things in the `ops` namespace we have to do it manually like we do with tracing in `ops/__init__.py` and scenario in `ops/testing/__init__.py`. That didn't feel necessary here, particularly since I think people will just be running the script most of the time rather than actually doing an import.

> Also, do we need "how to install it"? I'd be fine if we just recommend `uvx ops.tools.update-charmcraft` or whatever, but I think we should at least mention it.

Yeah, fair point. I need to actually publish a version for `uvx` to nicely work, but I know what the syntax will be, so can add that in advance. I think in the README for now, and in the docs PR it can move/copy elsewhere.

tonyandrewmeyer: > Is a namespace package possible here? (Distribute as `ops-tools`, import as `ops.tools`, maybe do an extra ...).

It's possible I missed something, but as far as I know, neither `ops` nor `ops.testing` can be namespace packages, because they are regular Python packages. If we were doing a new ops from scratch that would probably be nice, but I don't think we can do it with the current ops.

I can add an extra if you think there's value in it. It depends a bit on how this is being run. If you're `uvx`'ing it then there's not much point. If you're `uv run`'ing it and it's in some step where you're using ops as well, maybe that is convenient?

> Tiexin's release script makes the version bumping and dependencies pretty painless for releases.

I already have it in the release script, so it gets released with the rest and shares the version number other than major. I don't think that really impacts namespaces or extras choices.

james-garner-canonical: You're right that we can't do a straightforward namespace package here, since `ops` can't import a package from the same namespace -- forget that I mentioned it.

benhoyt: > That didn't feel necessary here, particularly since I think people will just be running the script most of the time rather than actually doing an import.

Yeah, that's fine. I'm happy to leave that as is.

james-garner-canonical: My concerns are as follows:
* it's a bit ugly to have `ops-scenario` and `ops-tracing` on PyPI, but `ops_tools`
* it's a bit unexpected to import `ops.testing` and `ops.tracing`, but import/run `ops_tools`
However, I think you may have already made the right call here about consistency since users aren't expected to import it, and the intended ux is `uvx`. What would the `uvx` incantation be?

tonyandrewmeyer: > * it's a bit ugly to have `ops-scenario` and `ops-tracing` on PyPI, but `ops_tools`

It should be `ops-tools`, that's what's in the "name" field in pyproject.toml.

> * it's a bit unexpected to import `ops.testing` and `ops.tracing`, but import/run `ops_tools`
>   However, I think you may have already made the right call here about consistency since users aren't expected to import it,

You can import `ops_tracing` (but `scenario`, not `ops_scenario`), we just shove them into `ops` and `ops.testing` to make it feel more like it's baked in. My sense was that we wanted this to feel a little *less* baked in - these are tools to use with ops, not something that's part of ops itself.

I'm not against importing it into the `ops` namespace if it's installed (if we do that then I think we would want the extra as well). I think the most likely case for this is if you're adding in a `to_juju_schema()` method to customise what's there, and you want to use the `ops_tools.OptionDict` or `ops_tools.ActionDict` types rather than defining them yourself or using more `Any`s. But my guess was that would probably be in a `TYPE_CHECKING` block, which doesn't really go with the "inject into ops namespace" solution.

> and the intended ux is `uvx`. What would the `uvx` incantation be?

I think it's either `uvx` or `uv run`. I need to double-check this, but I think it's `uvx --from=ops-tools update-charmcraft-yaml`

benhoyt: > My sense was that we wanted this to feel a little less baked in - these are tools to use with ops, not something that's part of ops itself.

Yeah, I agree with that, and on reflection would prefer to avoid `ops.tools` for that reason.

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def main():
+    """Merge generated config and action sections into charmcraft.yaml."""
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    parser.add_argument(
+        '--charmcraft-yaml',
```
benhoyt: This is the path to the main artifact we're changing. What about calling this arg `--path` or similar? No need to be overly verbose.

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+        help='Path to the charmcraft.yaml file to update.',
+        default='charmcraft.yaml',
+    )
+    parser.add_argument(
+        '--config-class',
```
benhoyt: I realise this is more accurate, but it's a bit long. Would `--config` and `--action` be enough?

tools/README.md
```diff
@@ -0,0 +1,90 @@
...
+            compression:
+                type: string
+                description: The type of compression to use.
+                default: gzip
+                enum: [gzip, bzip2]
+        required: [filename]
+        additionalProperties: false
+```
+
+Type annotations for all classes should use the modern `a | b` and `a | None` form, rather than `Union[a, b]` or `Optional[a]`.
```
james-garner-canonical:
```suggestion
Type annotations for all classes must use the modern `a | b` and `a | None` form, rather than `Union[a, b]` or `Optional[a]`.
```
Is this actually a hard requirement of how we consume the annotations?

tools/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Software Development :: Libraries",
+]
+
+[project.scripts]
+update-charmcraft-schema = "ops_tools._update_charmcraft_yaml:main"
```
james-garner-canonical, line 49: This name is a bit confusing to me. I also wonder if subcommands might be a nicer way to handle this as we add more commands. How about if the script is `ops-tools`, and the functionality added in this PR is under `ops-tools generate-metadata` or similar?

tools/src/ops_tools/_attrdocs.py
```diff
@@ -0,0 +1,108 @@
...
+    if hasattr(cls, 'model_fields'):
+        for attr, field in cls.model_fields.items():  # type: ignore
+            if hasattr(field, 'description') and field.description:  # type: ignore
+                docs[attr] = field.description  # type: ignore
+        # If we have `model_fields` then we'll assume that this is all of the documentation.
+        return docs
+
+    # For Pydantic dataclasses, we expect to have the docstrings in field
+    # objects, but there's no "model_fields" attribute.
+    elif dataclasses.is_dataclass(cls):
```
james-garner-canonical:
```suggestion
    if dataclasses.is_dataclass(cls):
```

tools/src/ops_tools/_attrdocs.py
```diff
@@ -0,0 +1,108 @@
...
+        for field in dataclasses.fields(cls):
+            if (
+                hasattr(field, 'default')
+                and hasattr(field.default, 'description')
+                and field.default.description  # type: ignore
+            ):
+                docs[field.name] = field.default.description  # type: ignore
+        # This might be a standard library dataclass, where there aren't
+        # description fields, so we fall through to collect any docstrings from
+        # the code as well.
```
james-garner-canonical, line 92:
```suggestion
        if docs:
            return docs
        # This might be a standard library dataclass, where there aren't
        # description fields, so we fall through to collect any docstrings from
        # the code as well.
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+
+from . import _attrdocs
+
+logger = logging.getLogger(__name__)
+
+
+class OptionDict(TypedDict):
+    type: str
+    """The Juju option type."""
+    description: NotRequired[str]
```
james-garner-canonical, line 47: I kind of feel like blank lines are helpful when using attribute docstrings, wdyt?
```suggestion
    type: str
    """The Juju option type."""

    description: NotRequired[str]
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+class ActionDict(TypedDict, total=False):
+    description: str
+    params: dict[str, Any]
+    """A dictionary of parameters for the action."""
+    required: list[str]
+    """A list of required parameters for the action."""
+    additionalProperties: bool
+    """Whether additional properties are allowed in the action parameters."""
```
james-garner-canonical, line 60:
```suggestion
    description: str
    params: dict[str, Any]
    """A dictionary of parameters for the action."""

    required: list[str]
    """A list of required parameters for the action."""

    additionalProperties: bool
    """Whether additional properties are allowed in the action parameters."""
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+JUJU_TYPES: Final[Mapping[type, str]] = {
+    bool: 'boolean',
+    int: 'int',
+    float: 'float',
+    str: 'string',
+    ops.Secret: 'secret',
+}
+
+# We currently only handle the basic types that we expect to see in real charms.
+# lists and tuples (arrays) are handled without using this mapping.
```
james-garner-canonical: We do handle dicts too, right?
```suggestion
# JSON arrays and objects (lists/tuples and dicts) are handled without using this mapping.
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+JSON_TYPES: Final[Mapping[type, str]] = {
+    bool: 'boolean',
+    int: 'integer',
+    float: 'number',
+    str: 'string',
+}
+
+
+def attr_to_default(cls: type[object], name: str) -> Any:
```
james-garner-canonical:
```suggestion
def attr_to_default(cls: type[object], name: str) -> object:
```
We explicitly check the types after calling this, so we should be fine to use `object` instead of `Any` for better type safety, right?

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+    """Get the default value for the attribute."""
+    if not dataclasses.is_dataclass(cls):
+        return getattr(cls, name, None)
+    default = None
+    for field in dataclasses.fields(cls):
+        if field.name != name:
+            continue
+        break
```
james-garner-canonical:
```suggestion
        if field.name == name:
            break
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+    else:
+        return default
```
james-garner-canonical: This might be easier to follow if we do something like this:
```suggestion
    fields = {field.name: field for field in dataclasses.fields(cls)}
    if name not in fields:
        return None
    field = fields[name]
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+    )
+    # A hack to avoid importing Pydantic here.
+    if (
+        'PydanticUndefinedType' not in str(type(field_default))  # type: ignore
+        and field_default is not dataclasses.MISSING
+    ):
+        return field_default  # type: ignore
+    if field_default_factory is not dataclasses.MISSING:
+        return field_default_factory()  # type: ignore
+    return default
```
james-garner-canonical: We don't actually set this anywhere other than initially setting it to `None`, right? I suggest not setting it at all (in previous suggestion), and I think it's easier to follow the function without it (you don't have to wonder if it's being modified at any point).
```suggestion
    return None
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def _attr_to_yaml_type(cls: type[object], name: str, yaml_types: dict[type, str]) -> str:
+    try:
+        raw_hint = get_type_hints(cls)[name]
+    except KeyError:
+        pass
+    else:
```
james-garner-canonical:
```suggestion
        return 'string'
```
And dedent the rest of the function accordingly. Though here rather than having multiple occurences of the possible-to-type string value `'string'`, I would prefer using a constant, e.g. `default: Final[str] = yaml_type[str]`.

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+        # Collapse Optional[] and Union[] and so on to the simpler form.
+        origin = get_origin(raw_hint)
+        if origin in (list, tuple):
+            return yaml_types[origin]
+        elif origin:
```
james-garner-canonical:
```suggestion
        if origin:
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+def _attr_to_yaml_type(cls: type[object], name: str, yaml_types: dict[type, str]) -> str:
+    try:
+        raw_hint = get_type_hints(cls)[name]
+    except KeyError:
+        pass
+    else:
+        # Collapse Optional[] and Union[] and so on to the simpler form.
+        origin = get_origin(raw_hint)
+        if origin in (list, tuple):
+            return yaml_types[origin]
```
james-garner-canonical: We call this function with `yaml_types=JSON_TYPES` and `yaml_types=JUJU_TYPES`, neither of which contain `list` or `tuple`. I'm assuming that `list` and `tuple` are handled correctly elsewhere. Maybe the right thing to do here is raise an explicit error (instead of the `KeyError` we'd get in practice).

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+        # `int | str` -- then we can't determine the type, and we fall back to
+        # "string", even if `str` is not in the type hint, because our
+        # "we can't determine the type" choice is always "string".
+        if len(hints) > 1:
+            return 'string'
+        elif hints:
+            try:
+                return yaml_types[hints.pop()]
+            except KeyError:
+                pass
```
james-garner-canonical: I think this is a simpler way of writing the same logic -- with one difference: we now *always* return `'string'` if there are multiple hints, instead of making an exception for the case where exactly one of them is in `yaml_types`. To bring that logic back, filter the result of `get_args` based on `yaml_types` -- but I think the semantics are clearer this way.

```suggestion
        hints = set(get_args(raw_hint)) if origin else {raw_hint}
        # If there are multiple types -- for example, the type annotation is
        # `int | str` -- then we can't determine the type.
        # Likewise if there are no hints somehow (generic used without args?).
        # If we can't figure it out, then use "string", which should be the most
        # compatible, and most likely to be used for arbitrary types. Charms can
        # override `to_juju_schema` to adjust this if required.
        return yaml_types.get(hints.pop(), 'string') if len(hints) == 1 else 'string'
```

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,434 @@
...
+    """
+    return _attr_to_yaml_type(cls, name, JSON_TYPES)
+
+
+def juju_schema_from_model_fields(cls: type[object]) -> dict[str, OptionDict]:
+    """Generate a Juju schema from the model fields of a Pydantic model."""
+    # The many type: ignores are required because we don't want to import
+    # pydantic in this code.
+    options: dict[str, OptionDict] = {}
+    for name, field in cls.model_fields.items():  # type: ignore
```
james-garner-canonical, line 160: This is as far as I've gotten in my review so far. I'm wondering here if we can reuse any of the logic already implemented in ops for extracting fields from classes, e.g. [ops.charm._juju_fields](https://github.com/canonical/operator/blob/01740d4947512afc78bcc150a96f9998e3ab24e8/ops/charm.py#L1604).

james-garner-canonical (commented):
Partial review for now.

tools/README.md
tonyandrewmeyer: > Is this actually a hard requirement of how we consume the annotations?

It's the only one I'm explicitly testing now. It's possible that it still works the other way (when this was going to be in Ops 2.x it did, and most of the code is the same).

I'm fine saying that it must.

tools/pyproject.toml
tonyandrewmeyer: I'm not that keen on having things as sub-commands under a single `ops-tools` command. For one thing, I don't like the duplication:

```
uvx --from=ops-tools ops-tools generate-metadata
```

But also I feel like `ops-tools` is meant to be a collection of tools, not a single tool with multiple subcommands.

However, I haven't finished going through the CLI spec in detail, so maybe it has recommendations here that should be followed.

I would like to resolve the comment around how this should work (just a diff/lint, merging, replacing) before locking down what the command looks like and is named.

james-garner-canonical: Re: duplication, you shouldn't need any if the command name is the same as the package name, right? Just `uvx ops-tools generate-metadata`. Good call on figuring out the rest of the commandline interface first though, I agree that's more important, and I have some thoughts there too.

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+    )
+    parser.add_argument(
+        '--path',
+        help='Path to the charmcraft.yaml file to update.',
+        default='charmcraft.yaml',
+    )
+    parser.add_argument(
+        '--config',
+        action='append',
+        help='Python class with config classes (can be specified multiple times). '
```
james-garner-canonical:
```suggestion
        help='Python class with optional module path (can be specified multiple times). '
```

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+        'For example, "src.config:Config". The module defaults to "src.charm".'
+        'The class may be a regular expression.',
+        default=[],
+    )
+    parser.add_argument(
+        '--action',
+        action='append',
+        help='Python class with action classes (can be specified multiple times). '
```
james-garner-canonical:
```suggestion
        help='Python class with optional module path (can be specified multiple times). '
```

tools/pyproject.toml
james-garner-canonical: Leaving aside subcommands, since with uv that just means +/- the `--from` flag, I think the existing interface is nice overall. The main parts I find confusing are what will actually happen with the content we generate.

I'd suggest using mutually exclusive groups for the output options, and defaulting to a non-destructive output method like writing the generated content to stdout. How about:
- `--update` (current behaviour, write to existing file)
- `--diff` (current option)
- `--output=FILE` (default behaviour, with `-` for stdout)
I'd also be happy with a `--stdout` flag instead of `--output=` if that's simpler.

The `--merge` flag option had me a little confused too, I thought it was the option that turned on the `--update`  behaviour, but that's (currently) unconditional. Instead, it controls whether to keep actions/options that are in the `charmcraft.yaml` file but not present in our generated output, right? In that case, how about `--delete` / `--no-delete` flags for this (borrowing `rsync` semantics)?

Regarding the name, how about `update-charmcraft-yaml` instead? The term charmcraft schema seems a little off somehow. All together that gives us:
```
uvx {--from} ops-tools update-charmcraft-yaml
    --path=...  # This might be nice as an optional positional argument instead?
    --config=... ...
    --action=... ...
    [--delete / --no-delete (default)]
    [--update / --diff / --output='-' (default)]
```

james-garner-canonical: It also seems like it would be nice to try this out without consuming a `charmcraft.yaml` file at all. What do you think about additional entry points (scripts or subcommands) with much simpler interfaces that just write the actions/config options to stdout? 
```
uvx {--from} ops-tools generate-actions CLASS...
uvx {--from} ops-tools generate-config CLASS...
```
I'd suggest writing in the `actions.yaml` and `config.yaml` format, but there could always be a flag added to control that.

tools/src/ops_tools/_generate_yaml.py
```diff
@@ -0,0 +1,437 @@
...
+        field.default.default  # type: ignore
+        if hasattr(field.default, 'default')
+        else field.default
+    )
+    field_default_factory = (  # type: ignore
+        field.default.default_factory  # type: ignore
+        if hasattr(field.default, 'default_factory')
+        else field.default_factory
+    )
+    # A hack to avoid importing Pydantic here.
```
james-garner-canonical, line 102: What if `ops-tools` just depended on Pydantic? We don't want to add a heavy dependency to `ops` itself, but would this all be a lot easier if we included it in `ops-tools`?

james-garner-canonical (changes requested):
I've left some major and minor comments.

Major: I think the CLI interface is pretty nice, but what actually happens with the output could be made more clear. I'd like to see non-destructive default behaviour by default, and I'd also like to be able to dump a Python class to a Juju action/config definition without worrying about `charmcraft.yaml` at all.

More minor: while reading through the code, I've left a number of suggestions that I think might make some of the logic easier to follow. However ...

Big picture questions:
1. How much simpler would interacting with Pydantic model classes be if `ops-tools` could just import Pydantic?
    - Additionally, if we're importing a module that defines a Pydantic class, won't we have to have Pydantic installed anyway?
    - We could either add Pydantic to our own deps
    - Or if we're relying on the user doing `--with=pydantic` or something, then we could move our Pydantic handling logic to a module that imports `pydantic` and only import that module as needed. This would have the benefit of letting us fudge around the Pydantic version, but seems more complicated.
2. I may be wrong about this, but my impression is that we're doing similar class traversal logic in a few different places for different purposes. This is some of the trickier code to follow imo. Would it be worthwhile to centralise this logic and parse the classes we're consuming into an intermediate data structure that has everything we want represented nicely?
    - I also had kind of assumed that the core class parsing logic would be in `ops` (internals), and we'd use it in `ops-tools`, since we already parse classes for loading actions/config/relation data in `ops`. That wouldn't work if we went down the route of using Pydantic here, but perhaps something to keep in mind.

tools/src/ops_tools/_generate_yaml.py
tonyandrewmeyer: I was trying to avoid building up a dict just to get a single item out of it. With the fix from above, I think it's better than adjusting it this way.

(It's really annoying that there's no dataclasses.get_field, or that dataclasses.fields isn't a dict).

tools/src/ops_tools/_generate_yaml.py
tonyandrewmeyer: I'm pretty strongly opposed to pulling Pydantic in, even for ops-tools. It would be a bit easier, but not a lot, I feel.

james-garner-canonical: May I ask why -- is it the extra runtime overhead for users?

tools/src/ops_tools/_generate_yaml.py
tonyandrewmeyer: Nice catch. This is leftover from earlier code that was handling the list/tuple cases. That's done more cleanly (although with a bit of duplication, annoyingly) for actions now before this is called, so it'll never be triggered in that case.

For config, we don't have a way to say "this is an array" in the Juju form, so it should fall back to string like elsewhere.

tools/src/ops_tools/_attrdocs.py
dimaqq: Ouch I left this comment "Pending" on my side for three weeks!

tools/src/ops_tools/_generate_yaml.py
dimaqq: My 2c: if dataclasses doesn't provide a dict, a helper method that builds a dict is the most natural way to go.

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+from __future__ import annotations
+
+import argparse
+import difflib
+import importlib
+import re
+import sys
+from typing import Any, Generator
+
+import yaml
```
dimaqq, line 32: IMHO you really want `ruamel.yaml` here.

I'm guessing what you do is trying to figure out the boundaries of the existing section and then replacing it wholesale relying on the fact that it's YAML and not e.g. JSON, and that it's formatted with indentation.

In that case, how about including explicit delimiters, e.g.?

```yaml
storage:
  data:
    type: filesystem
    location: /tempo-data

# ==== BEGIN GENERATED ACTIONS ====

actions:
  list-receivers:
    description: |
      Returns a list of all enabled receiver endpoints.

# ==== END GENERATED ACTIONS ====

peers:
  tempo-peers:
    interface: tempo_peers
    description: |
      peer relation for internal coordination
```

WDYT?

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+            print('Action section differs from existing charmcraft.yaml:\n')
+            existing = yaml.safe_dump({'actions': existing_actions})
+            generated = yaml.safe_dump({'actions': actions})
+            differ = difflib.Differ()
+            result = differ.compare(existing.splitlines(), generated.splitlines())
+            print('\n'.join(result))
+            exit_code += 2
+        sys.exit(exit_code)
+
+    if args.merge and 'config' in charmcraft_yaml:
```
dimaqq: **question:** Is this for 12-factor charms?

Personally I would like to avoid merge like a plague of rats gorged on plagued infested locusts who ate all the plagued corn...

12F could be solved by a stub / base class for options.

tools/src/ops_tools/_update_charmcraft_yaml.py
```diff
@@ -0,0 +1,191 @@
...
+        config = charmcraft_yaml['config']['options'].update(config['options'])
+    raw_yaml = _insert_into_charmcraft_yaml(raw_yaml, 'config', {'config': config})
+    if actions:
+        if args.merge and 'actions' in charmcraft_yaml:
```
dimaqq: Are there unit tests and functional / test vector tests for the merge flag?

tools/tests_tools/test_generate_yaml_config.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+@pytest.mark.parametrize('config_class', _test_config_classes)
+def test_config_yaml_schema(config_class: type[object]):
+    generated_schema = ops_tools.config_to_juju_schema(config_class)
+    expected_schema = {
+        'options': {
+            'basic-bool': {
+                'type': 'boolean',
+            },
```
dimaqq, line 134: This is great, but omits YAML formatting.

There needs to be a test for verbatim YAML output.
I'm thinking a stability test perhaps (we may change the code, but the output keeps same "canonical" indent).

tools/README.md
```diff
@@ -0,0 +1,90 @@
...
+            filename:
+                type: string
+                description: The name of the backup file.
+            compression:
+                type: string
+                description: The type of compression to use.
+                default: gzip
+                enum: [gzip, bzip2]
+        required: [filename]
+        additionalProperties: false
```
dimaqq, line 113: I'd like to see a test that looks exactly like this.

dimaqq (commented):
First round of reviews.

PietroPasotti:
this brings back some memories: https://github.com/PietroPasotti/jinx

test/charms/test_main/src/charm.py
```diff
@@ -32,6 +32,7 @@
 # during unit tests, and test_main failures that subprocess out are often
 # difficult to debug. Uncomment this line to get more informative errors when
 # running the tests.
+# When uncommented the test_hook_and_dispatch_with_failing_hook test will fail.
```
dimaqq: **nit** could be perhaps written as "test expected to fail if uncommented".

testing/src/scenario/state.py
dimaqq: I'm not getting the point of this comment.
Specifically, if I were to read it a few years later, I would be missing the context.

Could future work be tracked in GitHub issues or Jira tickets or roadmap items instead?

testing/tests/test_e2e/test_config.py
```diff
@@ -101,3 +105,27 @@ def _on_event(self, event):
...
+        b: float
+        c: str
+
+    class Charm(ops.CharmBase):
+        def __init__(self, framework: ops.Framework):
+            super().__init__(framework)
+            framework.observe(self.on.config_changed, self._on_config_changed)
+
+        def _on_config_changed(self, event: ops.ConfigChangedEvent):
+            self.typed_config = self.load_config(Config, 10, c='foo')
```
dimaqq, line 127: **Question** this is unexpected to me.

I would have thought that default are specified in a model, not in the load config call site.

The defaults can be included in both dataclasses and pydantic models, so why add the extra arguments feature in the `load_config` method?

P.S. maybe I was lazy reading the spec, but somehow I don't recall this API 🙈 

P.P.S. I guess it's because Juju action parameters don't have defaults like Juju config does, isn't it?

That also means that we should be testing both `required` and non-required parameters somewhere... perhaps a negative test in Scenario?

tools/tests_tools/test_generate_yaml_config.py
```diff
@@ -0,0 +1,241 @@
...
+    my_int: int = 42
+    """A positive integer value."""
+
+    my_float: float = 3.14
+    """A floating point value."""
+
+    my_str: str = 'foo'
+    """A string value."""
+
+    my_secret: ops.Secret | None = None
```
dimaqq, line 54: **Issues**

What are the semantics for optional-Secret-typed fields?

I guess if a value is passed, the a secret with this id (not label) must exist, or load_config() fails.

And I guess on error, same `errors = raise|blocked` applies, doesn't it?

And if nothing is passed (None, not empty string), then field if None.

This needs to be documented somewhere.

The spec describes Secret fields, **but not optional Secret fields**.

dimaqq (commented):
Partial review

tools/tests_tools/test_generate_yaml_config.py
dimaqq: The implementation remains the same, doesn't it?

```py
            option_type = self.meta.config.get(key)
            # Convert secret IDs to secret objects. We create the object rather
            # that using model.get_secret so that it's entirely lazy, in the
            # same way that SecretEvent.secret is.
            if option_type and option_type.type == 'secret':
                assert isinstance(value, str)  # Juju will have made sure of this.
                value = model.Secret(
                    backend=self.model._backend,
                    id=value,
                    _secret_set_cache=self.model._cache._secret_set_cache,
                )
            config[attr] = value
```

The value seems mandatory?

benhoyt:
Tracking issue: https://github.com/canonical/operator/issues/2163

tonyandrewmeyer:
Opening this fresh as smaller PRs with some small spec'ing.

PR #1976, opened 2025-08-12
chore: remove the Concierge work-around

dimaqq:
I noticed that Tempo charms don't seem to be using the same work-around as we do.

They are also on Concierge 1.0.4.

This is trial and error to see if the work-around can be removed.

Edit: instead of removing the work-around, I'm restricting it to Canonical K8s.

This works: https://github.com/canonical/operator/actions/runs/16896879217?pr=1976

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '42 7 25 * *'
+  pull_request:
 
 permissions: {}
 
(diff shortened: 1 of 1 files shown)


dimaqq:
My gut feeling is that Tempo only tests against microk8s, not CK8s, in which case the work-around is probably still needed.

If that's the case, perhaps I can refactor the GitHub action to apply the work-around only for CK8s.

dimaqq:
As expected Canonical K8s still needs a work-around.

james-garner-canonical (approved):

PR #1978, opened 2025-08-12
ci: speed up integration test

IronCore864:
Use pydantic binary (`charm-binary-python-packages`) in the charm plugin when building the test tracing charm, saving around 10 minutes for integration tests.


dimaqq:
Copy-pasting chat discussion:
> let's merge it for now.
eventually we'll probably move this charm to uv but that would need someone's time.

PR #1987, opened 2025-08-13
docs: update the httpbin charm to jubilant

dimaqq:


IronCore864 (approved):

examples/httpbin-demo/charmcraft.yaml
```diff
@@ -42,4 +42,4 @@ resources:
     # The upstream-source field is ignored by Juju. It is included here as a reference
     # so the integration testing suite knows which image to deploy during testing. This field
     # is also used by the 'canonical/charming-actions' GitHub action for automated releasing.
-    upstream-source: kennethreitz/httpbin
+    upstream-source: docker.io/kennethreitz/httpbin:latest
```
dwilding, line 45: Is this change purely for clarification for whoever is reading the code?

examples/httpbin-demo/tests/integration/test_charm.py
```diff
@@ -12,34 +12,26 @@
...
 
     # Deploy the charm and wait for active/idle status
-    await asyncio.gather(
-        ops_test.model.deploy(charm, resources=resources, application_name=APP_NAME),
-        ops_test.model.wait_for_idle(
-            apps=[APP_NAME], status='active', raise_on_blocked=True, timeout=1000
-        ),
-    )
+    juju.deploy(f'./{charm}', app=APP_NAME, resources=resources)
+    juju.wait(jubilant.all_active)
```
benhoyt, line 37: This isn't a perfect translation, as it's missing the `raise_on_blocked=True` bit. Does that actually matter? Does this charm Go into blocked? If so, we could do:

```python
juju.wait(jubilant.all_active, error=jubilant.any_blocked)
```

examples/httpbin-demo/tests/integration/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+import subprocess
+
+import jubilant
+import pytest
+
+
+@pytest.fixture(scope='module')
+def juju(request: pytest.FixtureRequest):
+    with jubilant.temp_model() as juju:
+        juju.wait_timeout = 100
```
benhoyt: This is 1000 in the old wait_for_idle code. Is 100 going to be too short? (3*60=180 is the Jubilant default).

examples/httpbin-demo/charmcraft.yaml
dimaqq: Yes and no.

By convention, this line is a clarification for the admin who deploys the charm.

The test we recommend later in the doc pulls this value out and deploys with this resource.

Juju does allow to deploy with a shorthand, so in effect we'd recommend the charmer to run:
```sh
juju deploy foo.charm  --resource foo=kennethreitz/httpbin
```

Which does work, however the reference is translated:
```command
⋊> dima@bb ⋊> ~ sudo ctr --namespace k8s.io i list | ag ken                                                                                                                                                 09:12:06
docker.io/kennethreitz/httpbin:latest                                                                                                                 application/vnd.docker.distribution.manifest.v2+json      sha256:599fe5e5073102dbb0ee3dbb65f049dab44fa9fc251f6835c9990f8fb196a72b 204.3 MiB linux/amd64                                                                             io.cri-containerd.image=managed
docker.io/kennethreitz/httpbin@sha256:599fe5e5073102dbb0ee3dbb65f049dab44fa9fc251f6835c9990f8fb196a72b                                                application/vnd.docker.distribution.manifest.v2+json      sha256:599fe5e5073102dbb0ee3dbb65f049dab44fa9fc251f6835c9990f8fb196a72b 204.3 MiB linux/amd64                                                                             io.cri-containerd.image=managed
```

The counterpoint is that it's not possible by the shorthand:
- `sudo ctr --namespace k8s.io images pull kennethreitz/httpbin` ❌ 
- `sudo ctr --namespace k8s.io images pull docker.io/kennethreitz/httpbin:latest` ✅ 

The proposed change reduces the element of suprise.
Also explicit is better than implicit.

examples/httpbin-demo/tests/integration/test_charm.py
dimaqq: Given that it's the very beginning of the tutorial, I'd personally avoid the extra argument.

The next step, postgres integration, could not use the same, as we expect the application to be blocked temporarily.

examples/httpbin-demo/tests/integration/conftest.py
dimaqq: My bad.

I though I saw `100` and was confused, but then `100` was good enough in my setup.
Given the later postgres integration, `1000` from the get-go makes more sense.

Except... this is a separate tutorial, so maybe I'll use the default instead?

dimaqq: I'll remove custom timeout.
The default (3 minutes) is more than enough, here's a cold start:
```
================================================================================================ 1 passed in 51.73s ================================================================================================
  integration: OK (51.91=setup[0.01]+cmd[51.91] seconds)
  congratulations :) (51.96 seconds)
```

examples/httpbin-demo/tests/integration/test_charm.py
dimaqq: Sorry.. I confused the PRs too.

I'll use the short form.

The charm is capable of entering the blocked state, but only if the config is bad.

The test doesn't provide any config, thus the default ("info", declared in charmcraft.yaml) is used, the charm is happy, arguably a well-written charm should be happy with own default values.

IMO, `error=...` would be a distraction.

examples/httpbin-demo/tests/integration/conftest.py
dwilding: I agree. If the httpbin charm comes in over the default, something is wrong with the default ;)

examples/httpbin-demo/tests/integration/test_charm.py
dwilding: This charm (the httpbin charm) can block if the config value is bad. So perhaps it would be good to have an additional integration test for that. I think that could be left for a follow-up PR.

dwilding: (Oops, didn't see your message when was replying)

PR #1990, opened 2025-08-14
docs: add model-config best practice note

tonyandrewmeyer:
This is copied from a new addition to the Juju Charm Maturity page. That page is in the process of being removed, but we'd like to retain this note.

The note is shorter than the one in the Juju docs. In the Juju docs there is a link to a library that handles this: I've instead opened a [PR in charmlibs](https://github.com/canonical/charmlibs/pull/113) to list the library there, as that seems more appropriate than advertising it in a best practice note or the how-to more broadly. I've also added a note about the Juju proxy settings to the public listing review notes, since we don't document those in ops currently.

[Preview](https://canonical-ubuntu-documentation-library--1990.com.readthedocs.build/ops/1990/howto/manage-configuration/)

### DIFF
--- a/docs/howto/manage-configuration.md
+++ b/docs/howto/manage-configuration.md
@@ -12,11 +12,8 @@ In the `charmcraft.yaml` file of the charm, under `config.options`, add a config
 ```{admonition} Best practice
 :class: hint
 
-Don't duplicate model-level configuration options that are controlled by `juju model-config`.
+Don't duplicate model-level configuration options that are controlled by {external+juju:ref}`juju model-config <command-juju-model-config>`.
 ```
-
-> See more:
->  - {external+juju:ref}`Juju CLI | model-config <command-juju-model-config>`
 
 The example below shows how to define two configuration options, one called `name` of type `string` and default value `Wiki`, and one called `skin` with type `string` and default value `vector`:
 


docs/howto/manage-configuration.md
```diff
@@ -7,7 +7,14 @@
 
 ### Define a configuration option
 
-In the `charmcraft.yaml` file of the charm, under `config.options`, add a configuration definition, including a name, a description, the type, and the default value. The example below shows how to define two configuration options, one called `name` of type `string` and default value `Wiki`, and one called `skin` with type `string` and default value `vector`:
+In the `charmcraft.yaml` file of the charm, under `config.options`, add a configuration definition, including a name, a description, the type, and the default value.
+
+```{admonition} Best practice
```
dwilding, line 12: This will need the `hint` class to get the right styles:

```suggestion
```{admonition} Best practice
:class: hint
```

docs/howto/manage-configuration.md
```diff
@@ -7,7 +7,14 @@
...
+
+Avoid duplicating configuration options that are best controlled at a model level, through `juju model-config`.
```
dwilding: What do you think about this alternative wording? My justification: Strengthening "avoid" to "don't" to remove any wiggle room in the advice (if that's appropriate). And removing "best" because, as I understand it, there's a fixed-set of config options that `model-config` applies to. I'm also good if you prefer to stick with the original wording

```suggestion
Don't duplicate model-level configuration options that are controlled by `juju model-config`.
```

docs/howto/manage-configuration.md
```diff
@@ -7,7 +7,15 @@
 
 ### Define a configuration option
 
-In the `charmcraft.yaml` file of the charm, under `config.options`, add a configuration definition, including a name, a description, the type, and the default value. The example below shows how to define two configuration options, one called `name` of type `string` and default value `Wiki`, and one called `skin` with type `string` and default value `vector`:
+In the `charmcraft.yaml` file of the charm, under `config.options`, add a configuration definition, including a name, a description, the type, and the default value.
+
+```{admonition} Best practice
+:class: hint
+
+Don't duplicate model-level configuration options that are controlled by `juju model-config`.
```
james-garner-canonical: Should we link to what options these are? e.g. the list here https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/model-config/

james-garner-canonical (approved):
I wonder if the link to the model-level configuration options should go within the best practice box, since the content of that is automatically pulled into the charm listing review checklist, isn't it?

tonyandrewmeyer:
> I wonder if the link to the model-level configuration options should go within the best practice box, since the content of that is automatically pulled into the charm listing review checklist, isn't it?

It is, but both the public listing review checklist and the collated list of best practices are presented as bullet point lists. That doesn't work super well with extra paragraphs like a "see more" unless special-casing is added for that, which I'd rather avoid. A simple hyperlink in the text is fine, but didn't see to work well here.

The collated list links back to the page where it appears (and we encourage charmers to be reading those docs, rather than relying on the collated list). The checklist in the review issue tries to be fairly minimal, and I think a reviewer should know how to find the docs.

So my preference is to leave it as-is, but if you have a concrete suggestion that would work for the three places, I'm happy to consider it.

james-garner-canonical:
Ah if there's a link from the best practice item in the review checklist to here, then that seems fine to me.

james-garner-canonical:
Oh, sorry, you said the collated page links back to 'here' (the best practice location in the docs), not that the checklist does. I would just turn "model-level configuration options" into the link (even though it's linking to the docs for the command, that's where the options are listed). But if that doesn't seem right to you, then don't worry about it, I assume the review instructions link to at least the collated page link.

PR #1991, opened 2025-08-15
ci: enable doctests

tonyandrewmeyer:
In #1975, we'd like to include a couple of docstrings. This PR clears the way for that, as well as any other doctests we wish to use in the future.

The PR tweaks some docstrings that are in private methods, so not in in our reference documentation. In addition:

* `ActionEvent.set_results` changes from a block of `>>>` text to a set of bullet points. If this was doctest'ed then it'd actually call the method, but doing that with `set_results` itself would be complicated. This seems just as clear to me.
* `pebble.Client.exec` has a couple of lines to disable doctest. This could be a doctest, but that would require a real Pebble, which we currently split off into a different set of tests. This is a bit ugly, but putting `doctest: +SKIP` on most of the lines is uglier (in my opinion), and I don't think there's a cleaner way to skip a single docstring.

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2979,6 +2979,7 @@ class Client:
         below, however, input/output handling is a bit more complex. Some
         examples are shown below::
 
+            # These two lines are related to documentation testing, not exec()
             >>> import pytest
             >>> pytest.skip('these examples require a running Pebble')
 


ops/_private/harness.py
```diff
@@ -335,46 +335,47 @@ def _event_context(self, event_name: str):
 
         Example usage:
 
-        # this is how we test that attempting to write a remote app's
```
tonyandrewmeyer, line 338: This is private, so only someone developing Harness would see the change.

ops/framework.py
```diff
@@ -943,10 +943,16 @@ def _event_context(self, event_name: str):
         is completed.
 
         Usage:
+
```
tonyandrewmeyer, line 946: This is private, so only someone developing Ops would see it.

tox.ini
```diff
@@ -110,7 +113,8 @@ passenv =
 dependency_groups = unit, coverage
 commands =
     coverage run --source={[vars]src_path},{[vars]testing_src_path} \
-             --branch -m pytest --ignore={[vars]tst_path}smoke \
+             --branch -m pytest \
+             --ignore={[vars]tst_path}smoke \
```
tonyandrewmeyer, line 117: I originally had the doctests running for coverage and then changed my mind on that. However, I think it is nicer to keep the ignores on separate lines.

tracing/ops_tracing/vendor/charms/tempo_coordinator_k8s/v0/tracing.py
```diff
@@ -950,15 +950,17 @@ def charm_tracing_config(
...
+        from lib.charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
+        from lib.charms.tempo_coordinator_k8s.v0.tracing import charm_tracing_config
+        @trace_charm(tracing_endpoint="my_endpoint", cert_path="cert_path")
+        class MyCharm(...):
+            _cert_path = "/path/to/cert/on/charm/container.crt"
+            def __init__(self, ...):
+                self.tracing = TracingEndpointRequirer(...)
+                self.my_endpoint, self.cert_path = charm_tracing_config(
+                    self.tracing, self._cert_path
+                )
```
tonyandrewmeyer, line 963: Technically, I should provide a clean for this upstream, but we're considering reimplementing this library rather than vendoring it, and there's the "move all the interface libraries" project as well, so it doesn't seem worth doing.

tonyandrewmeyer:
> Can you show what a failing doctest looks like?

In the summary:

```
FAILED test/charms/test_main/lib/ops/pebble.py::lib.ops.pebble.Client.exec
```

And in the details:

```
/home/tameyer/w/config-schema/test/charms/test_main/lib/ops/framework.py:946: UnexpectedException
__________________________________________________________ [doctest] lib.ops.pebble.Client.exec ___________________________________________________________
[gw0] linux -- Python 3.13.0 /home/tameyer/w/config-schema/.tox/unit/bin/python3
2974 :class:`ExecProcess` that deals with strings if ``encoding`` is
2975 specified (the default ), or one that deals with bytes if ``encoding``
2976 is set to ``None``.l
2977 
2978 Most of the parameters are explained in the "Parameters" section
2979 below, however, input/output handling is a bit more complex. Some
2980 examples are shown below::
2981 
2982     # Simple command with no output; just check exit code
2983     >>> process = client.exec(['send-emails'])
UNEXPECTED EXCEPTION: NameError("name 'client' is not defined")
  File "/home/tameyer/.local/share/uv/python/cpython-3.13.0-linux-x86_64-gnu/lib/python3.13/doctest.py", line 1395, in __run
    exec(compile(example.source, filename, "single",
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 compileflags, True), test.globs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<doctest lib.ops.pebble.Client.exec[0]>", line 1, in <module>
NameError: name 'client' is not defined. Did you mean: 'Client'?
```

ops/pebble.py
```diff
@@ -2979,6 +2979,9 @@ def exec(
         below, however, input/output handling is a bit more complex. Some
         examples are shown below::
 
+            >>> import pytest
+            >>> pytest.skip('these examples require a running Pebble')
```
james-garner-canonical, line 2984: What do you think about changing the format of examples that we don't want to be doctests instead? For example, assuming it still renders nicely in the docs and actually disables doctests, a single leading `>` instead of `>>>`? That said, your approach here is nice and explicit, not too intrusive, and minimises diffs, so I'm happy to be overruled here.

ops/charm.py
```diff
@@ -168,16 +168,16 @@ def set_results(self, results: dict[str, Any]):
...
+        to flatten any more complex data structure like so:
 
-            >>> {'a': 'b'} # becomes: 'a'='b'
-            >>> {'a': {'b': 'c'}} # becomes: 'a.b'='c'
-            >>> {'a': {'b': 'c', 'd': 'e'}} # becomes: 'a.b'='c', 'a.d' = 'e'
-            >>> {'a.b': 'c', 'a.d': 'e'} # equivalent to previous
+        * ``{'a': 'b'}`` becomes: ``'a'='b'``
+        * ``{'a': {'b': 'c'}}`` becomes: ``'a.b'='c'``
+        * ``{'a': {'b': 'c', 'd': 'e'}}`` becomes: ``'a.b'='c', 'a.d' = 'e'``
+        * ``{'a.b': 'c', 'a.d': 'e'}`` equivalent to previous
```
dwilding: Very minor suggestions, to make things flow more smoothly now that we're using bullets and no `# comments`:

```suggestion
        * ``{'a': 'b'}`` becomes ``'a'='b'``
        * ``{'a': {'b': 'c'}}`` becomes ``'a.b'='c'``
        * ``{'a': {'b': 'c', 'd': 'e'}}`` becomes ``'a.b'='c', 'a.d' = 'e'``
        * ``{'a.b': 'c', 'a.d': 'e'}`` is equivalent to the previous example
```

ops/pebble.py
dwilding: I wasn't familiar with doctests until today, so I do think that

```
import pytest
pytest.skip('these examples require a running Pebble')
```

at the top of the example would have been a significant point of confusion.

I'm OK if this is the most practical solution. How about including a comment to make it even more explicit:

```suggestion
            # These two lines are related to documentation testing, not exec()
            >>> import pytest
            >>> pytest.skip('these examples require a running Pebble')
```

Is there a convenient way to disable doctests for pebble.py only? I'd prefer that if possible. (Sorry if we discussed that possibility during standup and I missed it)

Or inspired by James's suggestion, if we keep doctests for pebble.py, could we drop `>>>` entirely and use comments to explain the output of each example? This might work well, as we only need to show a small amount of output for each example.

dwilding (approved):

ops/pebble.py
tonyandrewmeyer: > Is there a convenient way to disable doctests for pebble.py only?

There's not for this specific docstring. For all of pebble.py, it depends on how you define "convenient" :smile:. For example, if we ran the doctests separately then we could only include the specific files, but I think it's better all in the single pytest command. I'm also not keen on excluding the whole module when we might want to use doctest on other docstrings in here in the future.

> could we drop >>> entirely and use comments to explain the output of each example? This might work well, as we only need to show a small amount of output for each example.

That looks like this for the simpler cases:

image

(I do prefer just dropping the `>>>` over using `>`).

It's not as nice for the more complicated ones, like multi-line output and exceptions:

image

And for completeness, with the comment it's like this:

image

Thoughts?

dwilding: I find the comments with output readable enough. But, yeah, I see that the format does have to vary a bit. Overall, it probably takes more work to read, compared with the standard `>>>` format.

With the "These two lines..." comment, at least I can easily ignore those lines as soon as I understand why they're there. So I vote for keeping `>>>` and pytest and the explanatory comment.

tracing/ops_tracing/vendor/charms/tempo_coordinator_k8s/v0/tracing.py
dimaqq: My 2c:
- let's not run doctests on vendored code
- testing charm libs is outside of ops scope (albeit great for community)
- doc-testing charm libs is even more so (albeit helps to keep charm lib doc strings in sync with charm lib code: https://github.com/canonical/haproxy-operator/pull/163)

tonyandrewmeyer: The challenge here is that pytest's doctest integration doesn't provide much in the way of configuration (pytest-doctestplus is better, but still doesn't have exactly what we want). There's no "ignore" option so to do this we'd have to have a lot of glob statements that match the files to be tested and don't match the files to not be tested.

PR #1993, opened 2025-08-18
ci: update actions/checkout to v5

dimaqq:
The dependabot is making its rounds... (charming repos I track, etc.)

So I figure we can bump actions/checkout now rather than wait until the end of the month.


dimaqq:
Our self-hosted runners are rebuilt several times a day, so should be fine.

dimaqq:
Yep, already there: https://github.com/canonical/github-runner-operator/actions/runs/16741545987/job/48268728063#step:1:1

PR #1996, opened 2025-08-19
feat: expose the Juju hook context in ops.JujuContext

tonyandrewmeyer:
Renames `ops.jujucontext._JujuContext` to `JujuContext` and exposes it in the top-level `ops` namespace.

The `from_dict` method is kept private (by renaming it to `_from_dict`). This is still used internally, and is the most forgiving method for instantiating a `JujuContext`. A new `from_environ` method is added, which strictly validates that the correct variables are in the environment (based on the [Juju documentation](https://documentation.ubuntu.com/juju/3.6/reference/hook/) and manual validation).

(The majority of the changes in terms of lines are from those two renames).

`JujuContext` now requires keyword arguments (presumably we would have done this originally, but were restricted to Python 3.8 at the time).

`JujuContext` has a few more attributes than `_JujuContext`:
* `availability_zone`
* `hook_name`
* `machine_id`
* `principal_unit`
* `storage_index` (this was overlooked in the spec)

The `JujuContext` reference docs ([Preview](https://canonical-ubuntu-documentation-library--1996.com.readthedocs.build/ops/1996/reference/ops/#ops.JujuContext)) are tweaked slightly to be aimed at charmers (or people writing charming APIs/frameworks) rather than developers working on Ops itself. There are no additional how-to guides or explanation docs, as we do not intend for this class to typically be used by charmers (everything charmers need should be available via the framework). The follow-up PR that adds `ops.hookcmds` might add a general 'writing your own charming API' doc (very much to be decided).

The spec has an `event_name` property being added to `JujuContext`, however, when implementing this I realised that it's essentially just `.hook_name`, so there doesn't seem any point in adding that.

Scenario is switched over to use `from_environ`, as we are in total control over the environment in that case, and this exercises the new code significantly.

There are two small adjustments to Scenario that arose from this work:

* A new consistency check for relation events, verifying that they provide the remote unit. There are existing warnings at (test) runtime if the remote unit is implicitly selected -- these now also produce consistency check warnings. At (test) runtime if there is no remote unit to be selected, the test would previously have continued after warning, but that scenario is not possible in Juju, so now raises a consistency error.
* The mocked `os.environ` that Scenario manages now correct uses dashes for `JUJU_DISPATCH_PATH` and `JUJU_HOOK_NAME`.

See [OP069](https://docs.google.com/document/d/1HJSFVxDSDVYatVcvv4k13nBMvohb9sdgWbRt2pGOiyk/edit?usp=sharing) for more details.


tonyandrewmeyer:
@dwilding no need to do a full review (I picked two others for that) but if you'd like to review the docs that would be appreciated.

tonyandrewmeyer:
> * A new consistency check for relation events, verifying that they provide the remote unit. There are existing warnings at (test) runtime if the remote unit is implicitly selected -- these now also produce consistency check warnings. At (test) runtime if there is no remote unit to be selected, the test would previously have continued after warning, but that scenario is not possible in Juju, so now raises a consistency error.

The observability test failure is due to this change. I'll check this against our big list of charms to see how widespread this is. Hopefully, it's just this one case and if so I'll open a PR to fix it upstream (since it is really a bug to test something in a way that can't happen in Juju).

uv.lock
```diff
@@ -405,87 +405,87 @@ wheels = [
...
-    { url = "https://files.pythonhosted.org/packages/72/9b/27fbf79451b1fac15c4bda6ec6e9deae27cf7c0648c1305aa21a3454f5c4/coverage-7.10.3-cp314-cp314t-win_amd64.whl", hash = "sha256:424ea93a323aa0f7f01174308ea78bde885c3089ec1bef7143a6d93c3e24ef64", size = 220898, upload-time = "2025-08-10T21:27:16.297Z" },
-    { url = "https://files.pythonhosted.org/packages/d1/cf/a32bbf92869cbf0b7c8b84325327bfc718ad4b6d2c63374fef3d58e39306/coverage-7.10.3-cp314-cp314t-win_arm64.whl", hash = "sha256:f5983c132a62d93d71c9ef896a0b9bf6e6828d8d2ea32611f58684fba60bba35", size = 218922, upload-time = "2025-08-10T21:27:18.22Z" },
-    { url = "https://files.pythonhosted.org/packages/84/19/e67f4ae24e232c7f713337f3f4f7c9c58afd0c02866fb07c7b9255a19ed7/coverage-7.10.3-py3-none-any.whl", hash = "sha256:416a8d74dc0adfd33944ba2f405897bab87b7e9e84a391e09d241956bd953ce1", size = 207921, upload-time = "2025-08-10T21:27:38.254Z" },
+version = "7.10.4"
```
dimaqq: **note:** this is an unrelated change, I'd rather avoid those.

Unless you needed newer coverage, in which case it should be bumped in pyproject.toml

tox.ini
```diff
@@ -76,7 +76,6 @@ commands =
 
 [testenv:static]
 description = Run static type checker
-allowlist_externals = cp
```
dimaqq: **praise:** nice fly-by

ops/jujucontext.py
```diff
@@ -12,36 +12,84 @@
...
 from typing import Any, Mapping
 
 from .jujuversion import JujuVersion
 
 
-@dataclasses.dataclass(frozen=True)
-class _JujuContext:
-    """_JujuContext collects information from environment variables named 'JUJU_*'.
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class JujuContext:
```
dimaqq, line 28: Given that it was private before, perhaps it's fine to break a single sunbeam test file?

ops/jujucontext.py
```diff
@@ -233,3 +290,101 @@ def from_dict(cls, env: Mapping[str, Any]) -> _JujuContext:
...
+            'stop',
+            'remove',
+            'config-changed',
+            'update-status',
+            'upgrade-charm',
+            'leader-elected',
+        ):
+            return cls._from_dict(environ)
+
+        if event in ('secret-changed', 'secret-rotate', 'secret-remove', 'secret-expired'):
```
dimaqq, line 334: **note to self:** Dima to review the new unit tests carefully

ops/jujucontext.py
```diff
@@ -12,36 +12,84 @@
...
     """
 
-    action_uuid: str | None = None
-    """The action's uuid.
+    hook_name: str
+    """The name of the hook.
 
-    For example '1' (from JUJU_ACTION_UUID).
+    For example, 'install' (from JUJU_HOOK_NAME). Note that this is the empty
+    string for actions, and for relation, storage, and workload hooks the name
```
dimaqq: **question:** what is it for actions, really?

the doc string claims it would be empty, but the unit test includes `"
action/act"` for that case.

ops/jujucontext.py
tonyandrewmeyer: Yes, if you choose to go outside our public API then in my opinion you've accepted that there may be breakage with any release. I can reach out to be nice, but we can't let people using private classes or methods stop us changing them, or there's no point having anything private.

ops/jujucontext.py
tonyandrewmeyer: > **question:** what is it for actions, really?

Do you mean in Juju? The empty string, at least in my testing.

> the doc string claims it would be empty, but the unit test includes `" action/act"` for that case.

`JUJU_DISPATCH_PATH` will be something like "actions/act" but `JUJU_HOOK_NAME` will be empty. I may have mixed them up in a test? I'll have a look.

uv.lock
tonyandrewmeyer: Ah, I had a conflict and resolved it by regenerating. I guess the right thing to do is keep the one from origin and re-sync in that situation? I'll find a way to clean this up.

ops/_main.py
```diff
@@ -276,12 +276,12 @@ def __init__(
...
         use_juju_for_storage: bool | None = None,
         charm_state_path: str = CHARM_STATE_FILE,
-        juju_context: _JujuContext | None = None,
+        juju_context: JujuContext | None = None,
     ):
         from . import tracing  # break circular import
 
         if juju_context is None:
-            juju_context = _JujuContext.from_dict(os.environ)
+            juju_context = JujuContext._from_dict(os.environ)
```
benhoyt, line 284: Should this (and similar lines) be changed to `JujuContext.from_environ()`? Or does it need to be more forgiving than that is?

test/test_jujucontext.py
```diff
@@ -123,3 +126,202 @@ def test_parsing_all_str_fields(self):
...
+    assert context.action_name == 'act'
+    assert context.action_uuid == '1'
+
+
+def test_from_environ_pebble_ready():
+    environ = _valid_minimal.copy()
+    environ['JUJU_HOOK_NAME'] = 'w-pebble-ready'
+    environ['JUJU_WORKLOAD_NAME'] = 'w'
+    context = JujuContext.from_environ(environ)
+    assert context.workload_name == 'w'
```
benhoyt: It might be good to use more than 1 char here just to test we're not picking off the first character only. In addition, would it be good to test that underscores/hyphens in the workload name come through as we expect them to? Similar elsewhere in this file.

ops/jujucontext.py
```diff
@@ -233,3 +290,101 @@ def from_dict(cls, env: Mapping[str, Any]) -> _JujuContext:
...
+        required = (
+            'DISPATCH_PATH',
+            'HOOK_NAME',
+            'MODEL_NAME',
+            'MODEL_UUID',
+            'UNIT_NAME',
+            'VERSION',
+        )
+        for var in required:
+            if f'JUJU_{var}' not in environ:
```
benhoyt: Nit: I think it would be consistent with the tests below, and slightly less confusion, to just include the `JUJU_` prefix in the *required* dict keys and avoid prepending it here.

ops/_main.py
tonyandrewmeyer: I had some back-and-forth about this with @james-garner-canonical in the spec (comments, and it didn't make it into the spec itself), and this is where we landed.

I feel we should err on the cautious/conservative side and stick with the current (very forgiving) method. I have checked `from_environ()` against the Juju documentation and against a real Juju (3.6, for the most part LXD but for a few things like the Pebble events MicroK8s). However, I worry that there is some edge case where the docs are wrong (or I misinterpreted something) and that didn't fall into my manual testing. For example, the recent case where the docs are wrong about `JUJU_VERSION` for two events that are not regularly tested. Even if the bug is on Juju's side, I feel we ought to continue supporting things until that's fixed.

I'm particularly of this mind because an exception here will cause the entire hook to fail without any real workaround other than up/down-grading the charm, which leaves the model in a particularly bad state.

For Scenario, we have control over the environment, and the stakes are obviously lower, so I think it's good to switch there. For people building new APIs or whatever with this I feel they can take up the responsibility of testing things.

Thoughts?

ops/jujucontext.py
```diff
@@ -233,3 +290,101 @@ def from_dict(cls, env: Mapping[str, Any]) -> _JujuContext:
...
+    def from_environ(cls, environ: dict[str, str] | None = None) -> JujuContext:
+        """Create a JujuContext from the environment.
+
+        If environ is ``None``, ``os.environ`` will be used.
+
+        Raises:
+            ValueError: If any required environment variables are missing.
+        """
+        if environ is None:
+            environ = os.environ  # type: ignore
```
james-garner-canonical: `type: ignore` probably goes away if we request `environ: Mapping[str, str] | None` instead of `dict` -- I haven't read ahead too far, but I assume we're not going to mutate `environ`.

ops/_main.py
james-garner-canonical: +1 on returning whatever `JujuContext` we currently would if we do get a weird environment in production to preserve our existing behaviour, rather than erroring out early if we got any of the new validation wrong. We can always switch to calling `from_environ` later if our confidence in the new logic increases.

But maybe it's worth running the new validation and logging a warning instead of raising an exception, to warn users of unexpected results, and so that we can catch edge cases we may have missed. WDYT?

benhoyt: That makes sense, Tony, thanks. I don't think it's worth running the new validation and giving a warning (warnings can be problematic anyway as they often don't show to the right "user").

ops/jujucontext.py
dimaqq: or maybe I mixed them up when reading the tests

ops/jujucontext.py
```diff
@@ -12,36 +12,84 @@
...
 
 from .jujuversion import JujuVersion
 
 
-@dataclasses.dataclass(frozen=True)
-class _JujuContext:
-    """_JujuContext collects information from environment variables named 'JUJU_*'.
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class JujuContext:
+    """JujuContext provides the Juju hook context.
```
dwilding: I think we can drop the class name, as for other classes. The last part of the docstring feels important to reiterate here. How about this:

```suggestion
    """Provides the Juju hook context, primarily for charming approaches outside of Ops.
```

ops/jujucontext.py
```diff
@@ -12,36 +12,84 @@
...
 
-    Source: https://documentation.ubuntu.com/juju/3.6/reference/hook/#hook-execution.
-    The HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398.
-    Only a subset of the above source, because these are what are used in ops.
+    Juju provides context for the hook in the form of environment variables.
+    Rather than directly accessing the environment, charms should use
+    :meth:`ops.JujuContext.from_environ` to access this information.
```
dwilding: I didn't fully get the point at first. How about this adjustment:

```suggestion
    :meth:`ops.JujuContext.from_environ` to create a ``JujuContext`` object that contains this
    information.
```

ops/jujucontext.py
```diff
@@ -12,36 +12,84 @@
...
+
+    Note that most of the information in ``JujuContext`` is exposed through the
+    framework, for example :attr:`ops.JujuContext.model_name` is
```
dwilding: Very minor suggestions:

```suggestion
    Most of the information in ``JujuContext`` is exposed through the
    framework. For example :attr:`ops.JujuContext.model_name` is
```

ops/jujucontext.py
```diff
@@ -67,127 +112,133 @@ class _JujuContext:
...
 
-    model_uuid: str = ''
-    """The uuid of the model.
+    # The remaining variables are all context for specific types of events.
 
-    For example 'cdac5656-2423-4388-8f30-41854b4cca7d' (from JUJU_MODEL_UUID).
+    action_name: str | None = None
+    """The action's name, for action events.
+
+    For example 'backup' (from JUJU_ACTION_NAME).
```
dwilding: Elsewhere in the docstrings we've (mostly) put Juju environment variables in code font. I think it would be best to use code font for these ones too

ops/jujucontext.py
```diff
@@ -50,9 +98,6 @@ class _JujuContext:
     """The root directory of the charm where it is running.
```
dwilding: We use this description for `Framework.charm_dir` - could use the same?

```suggestion
    """The charm project root directory.
```

ops/jujucontext.py
```diff
@@ -233,3 +290,101 @@ def from_dict(cls, env: Mapping[str, Any]) -> _JujuContext:
             version=JujuVersion(env.get('JUJU_VERSION', '0.0.0')),
             workload_name=env.get('JUJU_WORKLOAD_NAME') or None,
         )
+
+    @classmethod
+    def from_environ(cls, environ: dict[str, str] | None = None) -> JujuContext:
+        """Create a JujuContext from the environment.
```
dwilding:
```suggestion
        """Create a ``JujuContext`` object from the environment.
```

dwilding (approved):
Docs look good to me. A few small suggestions (some very minor)

ops/jujucontext.py
tonyandrewmeyer: Hmm, not sure about this one. Consistency with `Framework.charm_dir` would be good, but I'm not sure that's a great description.

For tests specifically, this will be (almost always) a temporary directory where the charm pieces have been copied. For Juju, this is a directory in the charm container or charm machine where the charm was unpacked. To me "project root directory" sounds like it's referring to the repo.

What about using this but dropping "project" both here and in the framework one?

dwilding: That makes sense. I think either of these would be good: "The directory where the charm is running" or "The charm's root directory".

test/test_jujucontext.py
```diff
@@ -123,3 +126,214 @@ def test_parsing_all_str_fields(self):
...
+        ('secret-rotate', {}),
+        ('secret-remove', {'JUJU_SECRET_ID': 'secret:1'}),
+        ('secret-expired', {'JUJU_SECRET_ID': 'secret:1'}),
+        ('a-workload-pebble-ready', {}),
+        ('a-workload-pebble-check-failed', {'JUJU_WORKLOAD_NAME': 'a-workload'}),
+        ('a-workload-pebble-check-recovered', {'JUJU_PEBBLE_CHECK_NAME': 'chk1'}),
+        ('a-workload-pebble-custom-notice', {}),
+        ('storage-1-storage-attached', {}),
+        ('storage-1-storage-detaching', {}),
+        ('the-act', {'JUJU_DISPATCH_PATH': 'actions/the-act', 'JUJU_HOOK_NAME': ''}),
```
dimaqq, line 189: I would have expected this way:

```suggestion
        ('', {'JUJU_DISPATCH_PATH': 'actions/the-act'}),
```

Is it the way it is so that the test name looks clean in pytest output only, or some other reason?

testing/src/scenario/_runtime.py
```diff
@@ -69,7 +68,8 @@ def _get_event_env(self, state: State, event: _Event, charm_root: Path):
             'JUJU_VERSION': self._juju_version,
             'JUJU_UNIT_NAME': f'{self._app_name}/{self._unit_id}',
             '_': './dispatch',
-            'JUJU_DISPATCH_PATH': f'hooks/{event.name}',
+            'JUJU_DISPATCH_PATH': f'hooks/{event._juju_name}',
+            'JUJU_HOOK_NAME': '' if event._is_action_event else event._juju_name,
```
dimaqq, line 72: **praise:** this looks nice and clean!

testing/src/scenario/_consistency_checker.py
```diff
@@ -543,6 +543,7 @@ def check_relation_consistency(
 ) -> Results:
     """Check the consistency of any relations in the :class:`scenario.State`."""
     errors: list[str] = []
+    warnings: list[str] = []
```
dimaqq, line 546: This looks like a fly-by, but I like it :)

ops/framework.py
```diff
@@ -595,7 +595,7 @@ class Framework(Object):
     """The charm's metadata."""
 
     charm_dir: pathlib.Path = None  # type: ignore
-    """The charm project root directory."""
+    """The directory where the charm is running."""
```
dimaqq, line 598: I guess that's the best we can do.

I was thinking "directory where Juju deployed charm to on the current unit", but that too can be misunderstood, like does that include trailing `/charm/` or not?

test/test_jujucontext.py
```diff
@@ -14,69 +14,72 @@
...
+import pytest
+
+from ops.jujucontext import JujuContext
 from ops.jujuversion import JujuVersion
 
 
 class TestJujuContext:
     def test_both_str_and_int_fields_default_to_none(self):
-        juju_context = _JujuContext.from_dict({'JUJU_VERSION': '0.0.0'})
+        juju_context = JujuContext._from_dict({'JUJU_VERSION': '0.0.0'})
```
dimaqq, line 28: It seems that `JujuContext.from_environ(some_dict)` is the new way to using the context, would it be better to test this new way of creating the context over the helper function?

dimaqq: Alternatively, maybe some of these tests can be retired if the same functionality is covered by new tests below?

dimaqq (approved):

test/test_jujucontext.py
tonyandrewmeyer: We're still using both at the moment (see the other thread with Ben and James). I think while they are both explicitly used, we should have some tests that explicitly test each of them, rather than relying on one being indirectly called.

test/test_jujucontext.py
tonyandrewmeyer: Mostly, it's aligning with the other tests, where the event name is used more than once, and in the asserts. It does help with the pytest output as well though, as you mention - if I have only the dictionary here I think I would need to wrap these in Params so that it's more understandable which one failed if one does.

tonyandrewmeyer:
> I'll check this against our big list of charms to see how widespread this is.

No additional failures with `unit` or `scenario`.

tonyandrewmeyer:
> The observability test failure is due to this change. I'll check this against our big list of charms to see how widespread this is. Hopefully, it's just this one case and if so I'll open a PR to fix it upstream (since it is really a bug to test something in a way that can't happen in Juju).

Opened https://github.com/canonical/alertmanager-k8s-operator/pull/357

ops/jujucontext.py
```diff
@@ -220,6 +275,9 @@ def from_dict(cls, env: Mapping[str, Any]) -> _JujuContext:
             secret_revision=(
                 int(env['JUJU_SECRET_REVISION']) if env.get('JUJU_SECRET_REVISION') else None
             ),
+            storage_index=(
+                int(env['JUJU_STORAGE_ID'].split('/')[-1]) if env.get('JUJU_STORAGE_ID') else None
+            ),
```
dimaqq, line 280: **minor:** I would prefer the storage name (below) and id (here) to be complementary.

For example:

```py
if val := env.get('JUJU_STORAGE_ID'):
    storage_name, _, index = val.rpartition("/")
    storage_index = int(index)
else:
    storage_name = storage_index = None
```

The way it's coded right now, if somehow the some future version of Juju sets the value to `foo/1/0` or `/bar/0`, we'd env up with the central element missing.

PR #1997, opened 2025-08-19
fix: correct magic equality methods

dimaqq:
I've noticed that we have a bunch of custom magic equality methods, but some of them felt a bit sus.

Here I'm fixing a few, and I'm hoping someone else can take over this PR and carry this to completion.

Specific fixes:
- `__eq__` should allow reflected operation: since there's no `__req__` in Python, that's achieved by returning `NotImeplemented`
- if a custom `__eq__` is declared, then a custom `__hash__` should be too, unless the class is a container
   - someone pls validate that it's the case for frozen dataclasses as well

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1264,10 +1264,13 @@ class _EntityStatus:
 
     _entity_statuses: ClassVar[dict[str, type[_EntityStatus]]] = {}
 
+    def __hash__(self) -> int:
+        return hash((self.name, self.message))
+
     def __eq__(self, other: Any):
         if isinstance(other, (StatusBase, _EntityStatus)):
             return (self.name, self.message) == (other.name, other.message)
-        return super().__eq__(other)
+        return NotImplemented
 
     def __repr__(self):
         status_type_name = self.name.title() + 'Status'
@@ -1430,10 +1433,13 @@ class Port(_max_posargs(1)):
                 'Port cannot be instantiated directly; please use TCPPort, UDPPort, or ICMPPort',
             )
 
+    def __hash__(self) -> int:
+        return hash((self.protocol, self.port))
+
     def __eq__(self, other: object) -> bool:
         if isinstance(other, (Port, ops.Port)):
             return (self.protocol, self.port) == (other.protocol, other.port)
-        return False
+        return NotImplemented
 
     def _to_ops(self) -> ops.Port:
         return ops.Port(port=self.port, protocol=self.protocol)
@@ -1533,10 +1539,13 @@ class Storage(_max_posargs(1)):
     For Kubernetes charms, this will always be 1. For machine charms, each new
     Storage instance gets a new index."""
 
+    def __hash__(self) -> int:
+        return hash((self.name, self.index))
+
     def __eq__(self, other: object) -> bool:
         if isinstance(other, (Storage, ops.Storage)):
             return (self.name, self.index) == (other.name, other.index)
-        return False
+        return NotImplemented
 
     def get_filesystem(self, ctx: Context) -> pathlib.Path:
         """Simulated filesystem root in this context."""
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -128,6 +128,8 @@ class Handle:
         return hash((self.parent, self.kind, self.key))
 
     def __eq__(self, other: Handle):
+        if not isinstance(other, Handle):
+            return NotImplemented
         return (self.parent, self.kind, self.key) == (other.parent, other.kind, other.key)
 
     def __str__(self):


dimaqq:
@james-garner-canonical would you be up to analyse this and if that's the right way go over all other classes where `__eq__` is defined?

testing/src/scenario/state.py
```diff
@@ -1430,10 +1433,13 @@ def __post_init__(self):
                 'Port cannot be instantiated directly; please use TCPPort, UDPPort, or ICMPPort',
             )
 
+    def __hash__(self) -> int:
+        return hash((self.protocol, self.port))
```
tonyandrewmeyer, line 1437: This is the same implementation as the version dataclasses provides for us, from what I can tell. What's the reason for adding this?

testing/src/scenario/state.py
```diff
@@ -1264,10 +1264,13 @@ class _EntityStatus:
...
     _entity_statuses: ClassVar[dict[str, type[_EntityStatus]]] = {}
 
+    def __hash__(self) -> int:
+        return hash((self.name, self.message))
+
     def __eq__(self, other: Any):
         if isinstance(other, (StatusBase, _EntityStatus)):
             return (self.name, self.message) == (other.name, other.message)
-        return super().__eq__(other)
+        return NotImplemented
```
tonyandrewmeyer, line 1273: Falling back to object doesn't make sense, since that's just checking `is` and it would have been caught above if true. However, I'm not sure why this wouldn't just be false - if the operands are swapped it'll be exactly the same check, so surely it's better to exit out quickly.

testing/src/scenario/state.py
```diff
@@ -1533,10 +1539,13 @@ class Storage(_max_posargs(1)):
...
     Storage instance gets a new index."""
 
+    def __hash__(self) -> int:
+        return hash((self.name, self.index))
+
     def __eq__(self, other: object) -> bool:
         if isinstance(other, (Storage, ops.Storage)):
             return (self.name, self.index) == (other.name, other.index)
-        return False
+        return NotImplemented
```
tonyandrewmeyer, line 1548: It seems like we should just remove this, it seems to be reimplementing what the dataclass provides. I think maybe this did something differently in the past with the index, and it was changed instead of removed.

testing/src/scenario/state.py
```diff
@@ -1430,10 +1433,13 @@ def __post_init__(self):
...
             )
 
+    def __hash__(self) -> int:
+        return hash((self.protocol, self.port))
+
     def __eq__(self, other: object) -> bool:
         if isinstance(other, (Port, ops.Port)):
             return (self.protocol, self.port) == (other.protocol, other.port)
-        return False
+        return NotImplemented
```
tonyandrewmeyer, line 1442: I believe this one exists so that the different types of port can be compared, but that doesn't actually make sense, because they'll never be the same. So I'd say we should remove this one too.

testing/src/scenario/state.py
dimaqq: Unless I'm mistaken, the moment custom `__eq__` is declared, dataclasses no longer provide `__hash__`, as there's no way to know how equality is implemented.

tonyandrewmeyer: Hmm, that's not what the docs say, but it could be the case, I'd have to check. It still seems like we want to remove eq, not add hash, though.

testing/src/scenario/state.py
james-garner-canonical: Indeed, as Tony says, this change should be a noop -- `object.__eq__` will just return `NotImplemented`. I doubt anyone has implemented a custom object that takes advantage of this equality loophole, but I'm not sure it's worth changing behavior here and returning False either.

testing/src/scenario/state.py
james-garner-canonical: Not the easiest docs to follow, but my read of the docs is that Tony is correct -- if eq and frozen are both True, we get hash for free. Probably worth checking in an interpreter to be sure.

testing/src/scenario/state.py
james-garner-canonical: Why will the two types of ports never have the same values?

While returning NotImplemented allows for greater flexibility here, returning False seems perfectly reasonable too. Perhaps worth checking the ops.Port implementation to see if there are any implications to this change.

testing/src/scenario/state.py
james-garner-canonical: This is like Port, where it supports comparisons with the ops equivalent as well, which seems reasonable at first glance?

testing/src/scenario/state.py
james-garner-canonical: I think this is true of the hash methods added to the other classes as well, isn't it?

testing/src/scenario/state.py
tonyandrewmeyer: > Why will the two types of ports never have the same values?

Ah, I misread this, I thought this was allowing comparisons of the different types of `testing.Port`, not `testing.Port` and `ops.Port`. We do need a custom `__eq__` here to handle that.

What flexibility does `NotImplemented` offer here? Doesn't that tell Python to keep looking for comparisons? Are you thinking of a case where someone else has implemented a class that can compare itself to `testing.Port` subclasses?

However, this is broken, because it looks like dataclasses replaces the inherited `__eq__`, so this `__eq__` is actually never used. I think we need to pass in `eq=False` for the `testing.Port` subclasses (or implement the custom one on each of them).

testing/src/scenario/state.py
tonyandrewmeyer: Yeah, I missed that here too. So we do need to keep it (and unlike `Port` it actually works). I'm unclear what benefit we get from returning `NotImplemented` over `False`.

testing/src/scenario/state.py
tonyandrewmeyer:
```python
>>> import dataclasses
>>> @dataclasses.dataclass(frozen=True)
... class D:
...     a: str
...     def __eq__(self, other):
...         return False
...         
>>> d1 = D('a')
>>> d2 = D('a')
>>> d1 == d2
False
>>> hash(d1)
-6386893542128907590
>>> hash(d2)
-6386893542128907590
```

testing/src/scenario/state.py
james-garner-canonical: > What flexibility does NotImplemented offer here? Doesn't that tell Python to keep looking for comparisons? Are you thinking of a case where someone else has implemented a class that can compare itself to testing.Port subclasses?

Yes, that's all I'm thinking of. Not an important use case here imo.

> However, this is broken, because it looks like dataclasses replaces the inherited __eq__, so this __eq__ is actually never used.

Ah yes, you're right. Does this mean we can/should just drop this method entirely?

testing/src/scenario/state.py
tonyandrewmeyer: > I think this is true of the hash methods added to the other classes as well, isn't it?

Yes, I was meaning it to apply to all of them, but didn't want to spam posting the message more than once. Particularly since I wasn't clear on the reasoning.

testing/src/scenario/state.py
james-garner-canonical: It does let the other party have a chance to do their own funky equality logic (e.g. always return True), rather than only when they're the left hand operand -- so I guess it's better practice over all? Nice for equality not to depend on operand ordering.

testing/src/scenario/state.py
tonyandrewmeyer: > Yes, that's all I'm thinking of. Not an important use case here imo.

Agreed, although I guess there's no reason to not make the change either.

> Ah yes, you're right. Does this mean we can/should just drop this method entirely?

Because clearly no-one is using it? I think it would be better to fix the bug. It does seem reasonable to compare testing and ops objects, and it works in most of the other cases. I think this wasn't found because (a) Tony must have missed adding tests and (b) unit tests for opening and closing ports aren't that common.

They really ought to hash to the same value as well. My recollection is that the default `__contains__` for sets and dicts uses first `__hash__` and then (if needed) `__eq__`. I thought dataclasses that had identical members (and values for those members) would hash the same, but that doesn't seem to be the case. So `ops.Port(port=80, protocol='tcp') in {testing.TCPPort(80)}` doesn't work, I believe, and that really should.

testing/src/scenario/state.py
james-garner-canonical: Having thought about this more, I think that returning `NotImplemented` is the right thing to do. It should only break people who (1) wrote custom classes that should compare equal to scenario classes, and (2) used them as the right hand side operand, and (3) expected them to return False when they should really have returned True -- which should hopefully be a very small set of people. Oh and I guess people that were calling `__eq__(...)` directly ...

testing/src/scenario/state.py
james-garner-canonical: Sounds reasonable to me.

testing/src/scenario/state.py
dimaqq: Why `NotImplemented`:

Suppose someone (we within ops, or downstream user) implements:

```py
class MyStat:
    ...
    def __eq__(self, other: object):
        if isinstance(other, ...): ...
        if isinstance(other, _EntityStatus): return self.xxx == other.yyy
        return NotImplemented
```

Then, there's code like:

```py
a = _EntityStatus(...)
b = MyStat(...)

assert a == b  # always False today, as MyState.__eq__ is not called
assert b == a  # may be True or False
```

Not having the comparison symmetric would be weird and hard to debug.

With the change though:

```py
assert a == b  # uses _EntityStatus.__eq__, fails, uses MyStat.__eq__ may succeed
assert b == a  # MyStat.__eq__ first, failing that _EntityStatus.__eq__
```

dimaqq:
I'm thinking to wash my hands off this.
Looking for a volunteer or, failing that, to close the PR.

benhoyt:
@james-garner-canonical Any interest in getting this over the goal line? Otherwise we can just close for now and someone can pick it back up at a later date if needed.

james-garner-canonical:
I've been hoping to get to this for a while now, but realistically it won't happen any time soon. Closing this for now, but would be good to revisit on a rainy day.

PR #1998, opened 2025-08-19
ci: ignore juju/4 timeouts

dimaqq:
Ignore Juju 4/beta timeouts on microk8s, as those seem to happen all the time.

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         juju-channel: ['2.9/stable', '3/stable', '4/beta']
         preset: ['machine', 'microk8s']
+      continue-on-error: ${{ matrix.juju-channel == '4/beta' && matrix.preset == 'microk8s' }}
 
     steps:
       - run: sudo snap install --classic concierge
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         juju-channel: ['2.9/stable', '3/stable', '4/beta']
         preset: ['machine', 'microk8s']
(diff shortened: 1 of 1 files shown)


dimaqq:
Example: https://github.com/dimaqq/operator/actions/runs/17060535353
Which looks green overall at https://github.com/dimaqq/operator/actions/workflows/smoke.yaml

.github/workflows/smoke.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
       matrix:
         juju-channel: ['2.9/stable', '3/stable', '4/beta']
         preset: ['machine', 'microk8s']
+    continue-on-error: ${{ matrix.juju-channel == '4/beta' && matrix.preset == 'microk8s' }}
```
tonyandrewmeyer, line 19: Am I understanding this correctly? Since we use fail-fast: false, this doesn't change whether other jobs run, but it will show Juju4+microk8s as successful even if there are errors.

james-garner-canonical: I *think* it means that Juju4 + microk8s will still be visible as a failure (good!), but the workflow doesn't care whether that particular job was a success or failure -- the workflow can still get a green tick even if Juju4 + microk8s fails.

tonyandrewmeyer: Do we still get an email about the failure? I do think that's nice, so that each month we can check to see what's up - if it was just an expected issue with Juju 4 that's fine (although worth knowing for when we talk about 4 with charmers, although if there are indeed no more betas then perhaps this is all moot), but if it fails for some other reason we'd want to look into that?

dimaqq: I don't know about the email.
Individual job is marked as failed.
The total job is marked as passed.

tonyandrewmeyer: Ok, happy to approve and figure it out next month if needed :)

tonyandrewmeyer (approved):

PR #1999, opened 2025-08-20
add first draft of new machine charm tutorial

dwilding:
This PR starts work on a new machine charm tutorial. The purpose of the new tutorial is to teach modern charming practices with a real (but not complex) workload.

**The new tutorial is not ready to be used by charmers yet.** But early feedback is extremely welcome. To avoid dropping a huge PR for the team to review, this is the first PR in a sequence. Subsequent PRs will fill in the TODOs and make any changes coming from this initial review. Also, the tutorial won't be ready to use until Charmcraft 4 and `charmlibs-apt` are released.

Focus for reviewing this PR:
- Overall structure and topics covered in the tutorial
- Code review, including unit tests

**[Preview doc](https://canonical-ubuntu-documentation-library--1999.com.readthedocs.build/ops/1999/tutorial/write-a-machine-charm-for-tinyproxy/)** (ultimately the tutorial will replace "Write your first machine charm" - it won't be a separate doc)

### DIFF
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -9,6 +9,7 @@ Writing a machine charm is not so different from writing a Kubernetes charm, but
 
 write-your-first-machine-charm
 from-zero-to-hero-write-your-first-kubernetes-charm/index
+write-a-machine-charm-for-tinyproxy
 ```
 
 To charm a 12-factor style application, you're able to build a charm by running a Charmcraft command and making a few modifications to the provided content. The Charmcraft tutorials are a great entry point for anyone new to making charms, particularly anyone building or managing applications that use these frameworks:
--- a/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
+++ b/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
@@ -148,9 +148,10 @@ def stop() -> None:
 def reload_config() -> None:
     """Ask tinyproxy to reload config."""
     pid = _get_pid()
-    if pid:
-        # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
-        os.kill(pid, signal.SIGUSR1)
+    if not pid:
+        raise RuntimeError("tinyproxy is not running")
+    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
+    os.kill(pid, signal.SIGUSR1)
 
 
 def _get_pid() -> int | None:
--- a/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
+++ b/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
@@ -215,7 +215,7 @@ class TinyproxyConfig:
 
 ### Write the charm code (placeholder title)
 
-TODO: Split this section up, building the code in logical steps, with commentary. Explain what "receiving an event" from Juju means in practice. Mention that the workload and charm code run on the same machine, called the "unit", but that the charm code is usually not running. Juju runs the charm code on demand, passing data in the enrivonment, which represents an event to be processed. Ops provides a higher-level framework for handling events and responding to Juju.
+TODO: Split this section up, building the code in logical steps, with commentary. Explain what "receiving an event" from Juju means in practice. Mention that the workload and charm code run on the same machine, called the "unit", but that the charm code is usually not running. Juju runs the charm code on demand, passing data in the environment, which represents an event to be processed. Ops provides a higher-level framework for handling events and responding to Juju.
 
 `src/charm.py`:
 
--- a/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
+++ b/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
@@ -135,7 +135,7 @@ def is_running() -> bool:
 
 def start() -> None:
     """Start tinyproxy."""
-    subprocess.run(["tinyproxy"], check=True)
+    subprocess.run(["tinyproxy"], capture_output=True, check=True)
 
 
 def stop() -> None:
--- a/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
+++ b/docs/tutorial/write-a-machine-charm-for-tinyproxy.md
@@ -1,0 +1,577 @@
+# Write a machine charm for tinyproxy
+
+```{important}
+This tutorial is a **work in progress**. Some steps don't work yet. For now, please use [](./write-your-first-machine-charm).
+```
+
+TODO:
+
+- What you'll need.
+- One/two sentence summary of what you'll do.
+- How to get help, including linking to the code in the Ops repo.
+
+## Study your application
+
+TODO:
+
+- Briefly introduce tinyproxy.
+- Briefly introduce how a machine charm works.
+- Summarise the experience the charm will provide - reverse proxy with a configurable slug.
+
+## Set up your environment
+
+### Create a virtual machine
+
+TODO: Multipass
+
+### Install Juju and charm development tools
+
+TODO: Charmcraft, Juju, LXD (all using Concierge)
+
+### Install Python development tools
+
+TODO: tox, uv
+
+## Create a charm project
+
+You'll need to work inside the virtual machine again. Open a terminal, then run:
+
+```text
+multipass shell juju-sandbox
+```
+
+The terminal prompt is now `ubuntu@juju-sandbox:~$`.
+
+Next, create a project directory and use Charmcraft to create the initial version of your charm:
+
+```text
+mkdir tinyproxy
+cd tinyproxy
+charmcraft init --profile machine
+```
+
+TODO:
+
+- Briefly summarise the most important files that Charmcraft creates.
+- Edit charmcraft.yaml to set the title, description, summary, and platform.
+- Suggest using nano to edit files.
+
+## Write your charm
+
+### Write a helper module
+
+Your charm will interact with tinyproxy, so it's a good idea to start by writing a helper module that wraps tinyproxy.
+Charmcraft created `src/tinyproxy.py` as a placeholder helper module.
+
+The helper module will be independent of the main logic of your charm. This will make it easier to test your charm. However, the helper module won't be a general-purpose wrapper for tinyproxy. The helper module will contain opinionated functions for managing tinyproxy on Ubuntu.
+
+The helper module will depend on some libraries that are useful when writing charms. Add the libraries to your charm's dependencies:
+
+```text
+uv add charmlibs-apt charmlibs-pathops
+```
+
+This has added the following Python packages to the `dependencies` list in `pyproject.toml`:
+
+- `charmlibs-apt` - A library for using APT to manage system packages. This is how your charm will install tinyproxy.
+- `charmlibs-pathops` - A file operations library, similar to `pathlib` from the standard library.
+
+Next, replace the contents of `src/tinyproxy.py` by:
+
+```python
+import logging
+import os
+import re
+import shutil
+import signal
+import subprocess
+
+from charmlibs import apt, pathops
+
+logger = logging.getLogger(__name__)
+
+CONFIG_FILE = "/etc/tinyproxy/tinyproxy.conf"
+PID_FILE = "/var/run/tinyproxy.pid"
+
+
+def check_slug(slug: str) -> None:
+    """Check that the URL slug is valid. Raise ValueError otherwise."""
+    if not re.fullmatch(r"[a-z0-9-]+", slug):
+        raise ValueError(f"Invalid slug: '{slug}'. Slug must match the regex [a-z0-9-]+")
+
+
+def ensure_config(port: int, slug: str) -> bool:
+    """Ensure that tinyproxy is configured. Return True if any changes were made."""
+    # For the config file format, see https://manpages.ubuntu.com/manpages/jammy/en/man5/tinyproxy.conf.5.html
+    config = f"""\
+PidFile "{PID_FILE}"
+Port {port}
+Timeout 600
+ReverseOnly Yes
+ReversePath "/{slug}/" "http://www.example.com/"
+"""
+    return pathops.ensure_contents(CONFIG_FILE, config)
+
+
+def install() -> None:
+    """Use APT to install the tinyproxy executable."""
+    apt.update()
+    try:
+        apt.add_package("tinyproxy-bin")
+    except apt.PackageError as e:
+        logger.error("Unable to install tinyproxy-bin. %s", e.message)
+        raise
+
+
+def is_installed() -> bool:
+    """Return whether the tinyproxy executable is available."""
+    return shutil.which("tinyproxy") is not None
+
+
+def is_running() -> bool:
+    """Return whether tinyproxy is running."""
+    return pathops.LocalPath(PID_FILE).exists()
+
+
+def start() -> None:
+    """Start tinyproxy."""
+    subprocess.run(["tinyproxy"], check=True)
+
+
+def stop() -> None:
+    """Stop tinyproxy."""
+    os.kill(_get_pid(), signal.SIGTERM)
+
+
+def reload_config() -> None:
+    """Ask tinyproxy to reload config."""
+    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
+    os.kill(_get_pid(), signal.SIGUSR1)
+
+
+def _get_pid() -> int:
+    """Return the PID of tinyproxy."""
+    return int(pathops.LocalPath(PID_FILE).read_text())
+```
+
+Notice that the helper module is stateless. In fact, your charm as a whole will be stateless. The main logic of your charm will:
+
+1. Receive an event from Juju. For example, the `start` event, meaning "it's time to start tinyproxy".
+2. Use the functions in the helper module to manage tinyproxy and check its status.
+3. Report the status back to Juju.
+
+```{tip}
+After adding code to your charm, run `tox -e lint` to check the code against coding style standards and run static checks. If the code has formatting issues, you can use `tox -e format` to format the code.
+```
+
+### Define a configuration option
+
+TODO: Add commentary to this section.
+
+`charmcraft.yaml`:
+
+```yaml
+config:
+  options:
+    slug:
+      description: |
+        Configures the path of the reverse proxy.
+
+        Must match the regex [a-z0-9-]+
+      default: example
+      type: string
+```
+
+`src/charm.py`:
+
+```python
+@dataclasses.dataclass(frozen=True)
+class TinyproxyConfig:
+    """Schema for the configuration of the tinyproxy charm."""
+
+    slug: str = "example"
+    """Configures the path of the reverse proxy."""
+
+    def __post_init__(self):
+        tinyproxy.check_slug(self.slug)  # Raises ValueError if slug is invalid.
+```
+
+### Write the charm code (placeholder title)
+
+TODO: Split this section up, building the code in logical steps, with commentary.
+
+`src/charm.py`:
+
+```python
+import dataclasses
+import logging
+
+import ops
+
+import tinyproxy
+
+logger = logging.getLogger(__name__)
+
+PORT = 8000
+
+
+@dataclasses.dataclass(frozen=True)
+class TinyproxyConfig:
+    """Schema for the charm's config options."""
+
+    slug: str = "example"
+    """Configures the path of the reverse proxy."""
+
+    def __post_init__(self):
+        tinyproxy.check_slug(self.slug)  # Raises ValueError if slug is invalid.
+
+
+class TinyproxyCharm(ops.CharmBase):
+    """Charm the application."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        framework.observe(self.on.collect_unit_status, self._on_collect_status)
+        framework.observe(self.on.install, self._on_install)
+        framework.observe(self.on.start, self._on_start)
+        framework.observe(self.on.config_changed, self._on_config_changed)
+        framework.observe(self.on.stop, self._on_stop)
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent):
+        """Report the status of tinyproxy (runs after each event)."""
+        try:
+            self.load_config(TinyproxyConfig)
+        except ValueError as e:
+            event.add_status(ops.BlockedStatus(str(e)))
+        if not tinyproxy.is_installed():
+            event.add_status(ops.MaintenanceStatus("Waiting for tinyproxy to be installed"))
+        if not tinyproxy.is_running():
+            event.add_status(ops.MaintenanceStatus("Waiting for tinyproxy to start"))
+        event.add_status(ops.ActiveStatus())
+
+    def _on_install(self, event: ops.InstallEvent) -> None:
+        """Install tinyproxy on the machine."""
+        if not tinyproxy.is_installed():
+            tinyproxy.install()
+
+    def _on_start(self, event: ops.StartEvent) -> None:
+        """Handle start event."""
+        self._configure_and_restart()
+
+    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
+        """Handle config-changed event."""
+        self._configure_and_restart()
+
+    def _on_stop(self, event: ops.StopEvent) -> None:
+        """Handle stop event."""
+        tinyproxy.stop()
+
+    def _configure_and_restart(self) -> None:
+        """Ensure that tinyproxy is running with the correct config."""
+        try:
+            config = self.load_config(TinyproxyConfig)
+        except ValueError:
+            return
+        if not tinyproxy.is_installed():
+            return
+        changed = tinyproxy.ensure_config(PORT, config.slug)
+        if not tinyproxy.is_running():
+            tinyproxy.start()
+        elif changed:
+            logger.info("Config changed while tinyproxy is running. Updating tinyproxy config")
+            tinyproxy.reload_config()
+
+
+if __name__ == "__main__":  # pragma: nocover
+    ops.main(TinyproxyCharm)
+```
+
+## Try your charm
+
+### Pack your charm
+
+TODO: Add commentary to this section.
+
+```text
+charmcraft pack
+```
+
+This created the charm:
+
+```text
+CONTRIBUTING.md  charmcraft.yaml  src                    tox.ini
+LICENSE          lib              tests                  uv.lock
+README.md        pyproject.toml   tinyproxy_amd64.charm
+```
+
+(The charm filename depends on architecture)
+
+### Deploy your charm
+
+TODO: Add commentary to this section.
+
+Open another terminal, using `multipass shell juju-sandbox`. Use this terminal to watch Juju status`. In the other terminal, deploy the charm:
+
+```text
+juju deploy ./tinyproxy_amd64.charm
+```
+
+(The charm filename depends on architecture)
+
+When the charm is ready, you'll see:
+
+```text
+Model    Controller     Cloud/Region         Version  SLA          Timestamp
+testing  concierge-lxd  localhost/localhost  3.6.8    unsupported  09:00:38+08:00
+
+App        Version  Status  Scale  Charm      Channel  Rev  Exposed  Message
+tinyproxy           active      1  tinyproxy             0  no
+
+Unit          Workload  Agent  Machine  Public address  Ports  Message
+tinyproxy/0*  active    idle   0        10.71.67.208
+
+Machine  State    Address       Inst id        Base          AZ  Message
+0        started  10.71.67.208  juju-8e7bd9-0  ubuntu@22.04      Running
+```
+
+Make a note of the IP address of the machine (10.71.67.208).
+
+### Try the proxy
+
+TODO: Add commentary to this section.
+
+```text
+curl <address>:8000/example/
+```
+
+Output:
+
+```text
+<!doctype html>
+<html>
+<head>
+    <title>Example Domain</title>
+    ...
+</head>
+
+<body>
+<div>
+    <h1>Example Domain</h1>
+    <p>This domain is for use in illustrative examples in documents. You may use this
+    domain in literature without prior coordination or asking for permission.</p>
+    <p><a href="https://www.iana.org/domains/example">More information...</a></p>
+</div>
+</body>
+</html>
+```
+
+### Change the configuration
+
+TODO: Add commentary to this section.
+
+```text
+juju config tinyproxy slug=foo
+curl <address>:8000/foo/  # Returns the same HTML as before.
+curl <address>:8000/example/  # Returns an error page.
+juju config tinyproxy slug=foo/bar  # Puts the charm into blocked status.
+juju config tinyproxy --reset slug  # Makes the charm active, with the original config.
+```
+
+## Write unit tests for your charm
+
+### Write tests for the helper module
+
+TODO: Add commentary to this section.
+
+`tests/unit/test_tinyproxy.py`:
+
+```python
+import pytest
+
+from charm import tinyproxy
+
+
+def test_slug_valid():
+    tinyproxy.check_slug("example")  # No error raised.
+
+
+def test_slug_invalid():
+    with pytest.raises(ValueError):
+        tinyproxy.check_slug("foo/bar")
+```
+
+### Write state-transition tests
+
+TODO: Add commentary to this section.
+
+`tests/unit/test_charm.py`:
+
+```python
+from dataclasses import dataclass
+
+import pytest
+from ops import testing
+
+from charm import PORT, TinyproxyCharm
+
+
+@dataclass
+class MockTinyproxy:
+    """Mock object that represents tinyproxy."""
+
+    config: None | tuple[int, str] = None
+    installed: bool = False
+    reloaded_config: bool = False
+    running: bool = False
+
+    def ensure_config(self, port: int, slug: str) -> bool:
+        old_config = self.config
+        self.config = (port, slug)
+        return self.config != old_config
+
+    def install(self) -> None:
+        self.installed = True
+
+    def is_installed(self) -> bool:
+        return self.installed
+
+    def is_running(self) -> bool:
+        return self.running
+
+    def start(self) -> None:
+        self.running = True
+
+    def stop(self) -> None:
+        self.running = False
+
+    def reload_config(self) -> None:
+        self.reloaded_config = True
+
+
+def patch_charm(monkeypatch: pytest.MonkeyPatch, tinyproxy: MockTinyproxy):
+    """Patch the helper module to use mock functions for interacting with tinyproxy."""
+    monkeypatch.setattr("charm.tinyproxy.ensure_config", tinyproxy.ensure_config)
+    monkeypatch.setattr("charm.tinyproxy.install", tinyproxy.install)
+    monkeypatch.setattr("charm.tinyproxy.is_installed", tinyproxy.is_installed)
+    monkeypatch.setattr("charm.tinyproxy.is_running", tinyproxy.is_running)
+    monkeypatch.setattr("charm.tinyproxy.start", tinyproxy.start)
+    monkeypatch.setattr("charm.tinyproxy.stop", tinyproxy.stop)
+    monkeypatch.setattr("charm.tinyproxy.reload_config", tinyproxy.reload_config)
+
+
+def test_install(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm correctly handles the install event."""
+    # Arrange:
+    tinyproxy = MockTinyproxy()
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    # Act:
+    ctx.run(ctx.on.install(), testing.State())
+    # Assert:
+    assert tinyproxy.is_installed()
+
+
+def test_start(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm correctly handles the start event."""
+    # Arrange:
+    tinyproxy = MockTinyproxy(installed=True)
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    # Act:
+    state_out = ctx.run(ctx.on.start(), testing.State())
+    # Assert:
+    assert state_out.unit_status == testing.ActiveStatus()
+    assert tinyproxy.is_running()
+    assert tinyproxy.config == (PORT, "example")
+
+
+def test_config_changed(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm correctly handles the config-changed event."""
+    # Arrange:
+    tinyproxy = MockTinyproxy(config=(PORT, "example"), installed=True, running=True)
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    state_in = testing.State(config={"slug": "foo"})
+    # Act:
+    state_out = ctx.run(ctx.on.config_changed(), state_in)
+    # Assert:
+    assert state_out.unit_status == testing.ActiveStatus()
+    assert tinyproxy.is_running()
+    assert tinyproxy.config == (PORT, "foo")
+    assert tinyproxy.reloaded_config
+
+
+def test_start_invalid_config(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm fails to start if the config is invalid."""
+    # Arrange:
+    tinyproxy = MockTinyproxy(installed=True)
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    state_in = testing.State(config={"slug": "foo/bar"})  # Invalid value!
+    # Act:
+    state_out = ctx.run(ctx.on.start(), state_in)
+    # Assert:
+    assert state_out.unit_status == testing.BlockedStatus(
+        "Invalid slug: 'foo/bar'. Slug must match the regex [a-z0-9-]+"
+    )
+    assert not tinyproxy.is_running()
+    assert tinyproxy.config is None
+
+
+def test_config_changed_invalid_config(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm fails to change config if the config is invalid."""
+    # Arrange:
+    tinyproxy = MockTinyproxy(config=(PORT, "example"), installed=True, running=True)
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    state_in = testing.State(config={"slug": "foo/bar"})  # Invalid value!
+    # Act:
+    state_out = ctx.run(ctx.on.config_changed(), state_in)
+    # Assert:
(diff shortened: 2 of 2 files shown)


docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+```python
+import logging
+import os
+import re
+import shutil
+import signal
+import subprocess
+
+from charmlibs import apt, pathops
```
dwilding, line 91: The `charmlibs-apt` package hasn't been published to PyPI, so importing `apt` will fail at the moment. My plan is to wait for `charmlibs-apt` to be published before finishing the tutorial. In the meantime, I've tested the code using `apt` from https://github.com/canonical/charmlibs/tree/main/apt

docs/tutorial/index.md
```diff
@@ -9,6 +9,7 @@ Writing a machine charm is not so different from writing a Kubernetes charm, but
 
 write-your-first-machine-charm
 from-zero-to-hero-write-your-first-kubernetes-charm/index
+write-a-machine-charm-for-tinyproxy
```
dimaqq, line 12: Let's group machine docs together.

P.S. I dunno if having the workload name in the doc title and file name is a good idea. Would maybe something generic could be used, like "machine charm with a workload?"

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+- One/two sentence summary of what you'll do.
+- How to get help, including linking to the code in the Ops repo.
+
+## Study your application
+
+TODO:
+
+- Briefly introduce tinyproxy.
+- Briefly introduce how a machine charm works.
+- Summarise the experience the charm will provide - reverse proxy with a configurable slug.
```
dimaqq, line 19: Is this going to be a principal or a subordinate charm?

docs/tutorial/index.md
dwilding: I want to replace the current machine charm tutorial with this new one, when it's finished. So at that point we'll be back to two: "Write your first machine charm" and "From zero to hero: Write your first Kubernetes charm". If you think it's confusing in the meantime, I can shuffle things around and rename this temporary tutorial doc

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Hmm, I hadn't thought about that. My understanding is a bit limited here, but principal, I guess? Because it's not going to be integrated with anything in the tutorial

docs/tutorial/index.md
dimaqq: Ah, I see, that makes sense.

Personally I'd rename the old one to "write-your-previous-machine-charm" or something clever, and name the new thing the correct name, but I think it's fine either way.

Or if this is going to be series of PRs, then... maybe it's fine?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dimaqq: I suppose the question is what is tinyproxy proxying in this scenario?

A related app: separate charm, principal<->principal relation
Some local workload shipped in a separate charm: probably makes tinyproxy a subordinate
Some local workload that's deployed by this very charm: a standalone principal

docs/tutorial/index.md
tonyandrewmeyer: I'm curious about this too - is the intention that PRs merge into this branch and only into main when it's ready? I don't love having an unfinished (maybe even partly unusable with charmcraft 4 not out) tutorial in the contents and index. If it's living in a branch until ready I'm good with that. If it's live but only findable if you know the URL I'm fine with that too.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+TODO: Multipass
+
+### Install Juju and charm development tools
+
+TODO: Charmcraft, Juju, LXD (all using Concierge)
+
+### Install Python development tools
+
+TODO: tox, uv
```
tonyandrewmeyer: I'd suggest simplifying by having Concierge install uv (pass it as an extra snap). For tox, can't that be run through uv? Or is this for doing a uv tool type command?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+The helper module will depend on some libraries that are useful when writing charms. Add the libraries to your charm's dependencies:
+
+```text
+uv add charmlibs-apt charmlibs-pathops
+```
+
+This has added the following Python packages to the `dependencies` list in `pyproject.toml`:
+
+- `charmlibs-apt` - A library for using APT to manage system packages. This is how your charm will install tinyproxy.
```
tonyandrewmeyer: I think it's worth linking to the relevant docs here and for pathops.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+- `charmlibs-pathops` - A file operations library, similar to `pathlib` from the standard library.
+
+Next, replace the contents of `src/tinyproxy.py` by:
```
tonyandrewmeyer: Should "by" be "with"?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+def ensure_config(port: int, slug: str) -> bool:
+    """Ensure that tinyproxy is configured. Return True if any changes were made."""
+    # For the config file format, see https://manpages.ubuntu.com/manpages/jammy/en/man5/tinyproxy.conf.5.html
+    config = f"""\
+PidFile "{PID_FILE}"
+Port {port}
+Timeout 600
+ReverseOnly Yes
+ReversePath "/{slug}/" "http://www.example.com/"
```
tonyandrewmeyer, line 113: I'd probably use https, if people try this out then it avoids the insecure warning (example.com doesn't redirect, as far as I remember).

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    return pathops.ensure_contents(CONFIG_FILE, config)
+
+
+def install() -> None:
+    """Use APT to install the tinyproxy executable."""
+    apt.update()
+    try:
+        apt.add_package("tinyproxy-bin")
+    except apt.PackageError as e:
+        logger.error("Unable to install tinyproxy-bin. %s", e.message)
```
tonyandrewmeyer: Is it worth logging this just to then re-raise? I haven't checked but would assume the traceback would have basically the same content (and more).

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+
+def is_running() -> bool:
+    """Return whether tinyproxy is running."""
+    return pathops.LocalPath(PID_FILE).exists()
+
+
+def start() -> None:
+    """Start tinyproxy."""
+    subprocess.run(["tinyproxy"], check=True)
```
tonyandrewmeyer: Does tinyproxy output anything when it daemonises? Even if it doesn't, I think for the tutorial to show good practice we should capture any output here, rather than let it fall through to Juju.

I feel like we probably also want to wait for it to finish daemonising before returning and check that was successful, unless you have some reason to not do that? For one thing, I'm not sure what the behaviour would be if the charm process exits before the tinyproxy one does.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+def reload_config() -> None:
+    """Ask tinyproxy to reload config."""
+    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
+    os.kill(_get_pid(), signal.SIGUSR1)
+
+
+def _get_pid() -> int:
+    """Return the PID of tinyproxy."""
+    return int(pathops.LocalPath(PID_FILE).read_text())
```
tonyandrewmeyer: I assume you've tested this works, but maybe it would be a little more resilient to add a strip?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+Notice that the helper module is stateless. In fact, your charm as a whole will be stateless. The main logic of your charm will:
+
+1. Receive an event from Juju. For example, the `start` event, meaning "it's time to start tinyproxy".
+2. Use the functions in the helper module to manage tinyproxy and check its status.
+3. Report the status back to Juju.
+
+```{tip}
+After adding code to your charm, run `tox -e lint` to check the code against coding style standards and run static checks. If the code has formatting issues, you can use `tox -e format` to format the code.
```
tonyandrewmeyer: I think this is purely a style choice, so feel free to ignore this, but what I would do is the other way around, run format to ensure it's formatted correctly, and then run lint to check (knowing the format is right).

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+`src/charm.py`:
+
+```python
+@dataclasses.dataclass(frozen=True)
+class TinyproxyConfig:
+    """Schema for the configuration of the tinyproxy charm."""
+
+    slug: str = "example"
+    """Configures the path of the reverse proxy."""
```
tonyandrewmeyer: It'd be nice if the docstring was exactly the same as the description (so include the regex bit) so switching this to general the YAML in the future is more seamless.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+        Must match the regex [a-z0-9-]+
+      default: example
+      type: string
+```
+
+`src/charm.py`:
+
+```python
+@dataclasses.dataclass(frozen=True)
```
tonyandrewmeyer, line 205: This is exactly the choice I would make (although I would add kw_only=True). *However*, I believe @benhoyt prefers that our docs and examples only use Pydantic classes. You'd also then need to explain adjusting the charmcraft YAML to support building from Rust and that packing might now take 20 minutes (unless you want to just assume that the uv plugin won't get fixed to build from source).

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
tonyandrewmeyer: For what it's worth, our docs on subordinates are almost non-existent, so that would be interesting. I don't think adding a relation would make the tutorial significantly more complicated, although you could start with it being some local thing and then in the future we adjust it.

At the moment it's being used as basically a short URL service, which maybe is fine (the workload is not the point, after all). But actually running as a reverse proxy would probably seem more realistic.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    """Charm the application."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        framework.observe(self.on.collect_unit_status, self._on_collect_status)
+        framework.observe(self.on.install, self._on_install)
+        framework.observe(self.on.start, self._on_start)
+        framework.observe(self.on.config_changed, self._on_config_changed)
+        framework.observe(self.on.stop, self._on_stop)
+
```
tonyandrewmeyer, line 257: I think for a straightforward case like tinyproxy it's worth having a remove event that uninstalls tinyproxy and maybe even cleans up the config file. This is a bit contentious since ideally the machine isn't shared and if it is maybe some other charm is using the executable, but it does seem the best recommendation for the tutorial.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    def _on_stop(self, event: ops.StopEvent) -> None:
+        """Handle stop event."""
+        tinyproxy.stop()
+
+    def _configure_and_restart(self) -> None:
+        """Ensure that tinyproxy is running with the correct config."""
+        try:
+            config = self.load_config(TinyproxyConfig)
+        except ValueError:
+            return
```
tonyandrewmeyer, line 296: For a tutorial I think it's worth having a comment here explaining that collect-status will be handling letting the user know about the problem.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
tonyandrewmeyer: Hmm, but then reading later where you test with curl, maybe it's more complicated in that case and since this is all in the VM maybe it doesn't matter?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+import pytest
+
+from charm import tinyproxy
+
+
+def test_slug_valid():
+    tinyproxy.check_slug("example")  # No error raised.
+
+
+def test_slug_invalid():
```
tonyandrewmeyer: It's only a tutorial, so maybe this doesn't matter, but I think it would be nice to have this parametrised, to show how that is done. Just 2-3 examples of bad slugs, for example, like the empty string, something with an underscore, ...

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+### Write state-transition tests
+
+TODO: Add commentary to this section.
+
+`tests/unit/test_charm.py`:
+
+```python
+from dataclasses import dataclass
```
tonyandrewmeyer: We should follow the Charm Tech style and do `import dataclasses` and `dataclasses.dataclass`.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+- How to uninstall Multipass.
+
+## Next steps
+
+TODO:
+
+- Links to more detail on the concepts covered in the tutorial.
+- Suggestions for real machine charms to look at.
+- Suggestions for what to learn next, e.g., learn about relations.
+- Challenge: Try modifying this tutorial charm to make the port configurable too.
```
tonyandrewmeyer: I like the challenge idea. Would you provide an example charm where it's done as well?

If you want a second challenge, maybe having an action that returns the tinyproxy stats? I think it records those in reverse proxy mode as well as regular mode.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+ReversePath "/{slug}/" "http://www.example.com/"
+"""
+    return pathops.ensure_contents(CONFIG_FILE, config)
+
+
+def install() -> None:
+    """Use APT to install the tinyproxy executable."""
+    apt.update()
+    try:
+        apt.add_package("tinyproxy-bin")
```
tonyandrewmeyer: I wonder if this should install a specific version? For the tutorial it's probably more convenient to not do that, and not needed. However, it does seem to be what most machine charms do, to make their upgrade and QA processes simpler.

(But not the config for installing different versions like the existing tutorial - that seems a strange thing to do).

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    """Stop tinyproxy."""
+    os.kill(_get_pid(), signal.SIGTERM)
+
+
+def reload_config() -> None:
+    """Ask tinyproxy to reload config."""
+    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
+    os.kill(_get_pid(), signal.SIGUSR1)
+
+
```
tonyandrewmeyer, line 163: I know we don't want to make the tutorial charm too large/complex, but maybe setting the workload version would be nice? I assume tinyproxy provides a nice way to get this. It's in the charmcraft profile with a placeholder, right? It's also a nice thing to test against, since it's in the Juju state.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+- How to remove the application from Juju.
+- How to save the code that you created inside the virtual machine.
+- How to stop the virtual machine.
+- How to delete the virtual machine.
+- How to uninstall Multipass.
+
+## Next steps
+
+TODO:
```
tonyandrewmeyer, line 644: Also a couple of integration tests, hopefully?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    state_out = ctx.run(ctx.on.start(), state_in)
+    # Assert:
+    assert state_out.unit_status == testing.BlockedStatus(
+        "Invalid slug: 'foo/bar'. Slug must match the regex [a-z0-9-]+"
+    )
+    assert not tinyproxy.is_running()
+    assert tinyproxy.config is None
+
+
+def test_config_changed_invalid_config(monkeypatch: pytest.MonkeyPatch):
```
tonyandrewmeyer: Alternatively/additionally, this could have the parameterisation of bad values.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    """Test that the charm correctly handles the install event."""
+    # Arrange:
+    tinyproxy = MockTinyproxy()
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    # Act:
+    ctx.run(ctx.on.install(), testing.State())
+    # Assert:
+    assert tinyproxy.is_installed()
+
```
tonyandrewmeyer, line 532: I think it's worth asserting on the unit status here too.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+from dataclasses import dataclass
+
+import pytest
+from ops import testing
+
+from charm import PORT, TinyproxyCharm
+
+
+@dataclass
+class MockTinyproxy:
```
tonyandrewmeyer, line 463: It seems a bit odd to use a dataclass here. You're only really using it for the init, right? Not the rest of what a dataclass gives, like eq, hash, easy dumping to JSON, etc. I feel like this would be more natural as just a plain class with an explicit init.

I do, however, like how clean having the workload interaction in a module makes this mocking.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+def test_slug_valid():
+    tinyproxy.check_slug("example")  # No error raised.
+
+
+def test_slug_invalid():
+    with pytest.raises(ValueError):
+        tinyproxy.check_slug("foo/bar")
+```
+
+### Write state-transition tests
```
tonyandrewmeyer, line 450: I think these tests are fine. The charm does so little interaction with Juju that there's not much to test other than the workload modules are called, which is a bit odd, but there isn't really anything else to do here, without making the charm more complicated.

tonyandrewmeyer (commented):
I really like the direction this is headed, and I think it's going to be much stronger than our existing machine tutorial when done.

A bunch of small comments, but nothing significant other than echoing Dima's two questions about how this gets rolled out and how the workload gets used.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+
+def start() -> None:
+    """Start tinyproxy."""
+    subprocess.run(["tinyproxy"], check=True)
+
+
+def stop() -> None:
+    """Stop tinyproxy."""
+    os.kill(_get_pid(), signal.SIGTERM)
```
dimaqq: Do we then wait for pidfile for disappear?

Or are stale pid files allowed, in which case, `is_running` needs to be changed?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dimaqq: Other errors cases:
- file doesn't exist
- file exists, but content is not numeric

The alternative would be to wrap this in try/except and return `int | None`.

Or maybe having stdlib raise exceptions here is fine... but then the call sites have to handle the errors.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+
+
+def is_installed() -> bool:
+    """Return whether the tinyproxy executable is available."""
+    return shutil.which("tinyproxy") is not None
+
+
+def is_running() -> bool:
+    """Return whether tinyproxy is running."""
+    return pathops.LocalPath(PID_FILE).exists()
```
dimaqq: How sure are you that a stale PID file is not left when tinyproxy is a) told to please quit and b) gets killed hard by the OS?

```suggestion
    try:
        os.kill(_get_pid(), 0)
        return True
    except ProcessLookupError:
        return False
```

Also see my note about _get_pid raising exceptions.

PTAL at https://github.com/dimaqq/tracing-fake-operator/blob/main/src/server.py for a more comprehensive example of the UNIX way.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+    )
+    assert tinyproxy.is_running()  # tinyproxy should still be running...
+    assert tinyproxy.config == (PORT, "example")  # ...with the original config.
+    assert not tinyproxy.reloaded_config
+
+
+def test_stop(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm correctly handles the stop event."""
+    # Arrange:
+    tinyproxy = MockTinyproxy(installed=True)
```
dimaqq: This seems to be reused a lot, consider making it a fixture.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,577 @@
...
+def test_start_invalid_config(monkeypatch: pytest.MonkeyPatch):
+    """Test that the charm fails to start if the config is invalid."""
+    # Arrange:
+    tinyproxy = MockTinyproxy(installed=True)
+    patch_charm(monkeypatch, tinyproxy)
+    ctx = testing.Context(TinyproxyCharm)
+    state_in = testing.State(config={"slug": "foo/bar"})  # Invalid value!
+    # Act:
+    state_out = ctx.run(ctx.on.start(), state_in)
+    # Assert:
```
dimaqq: I'm not a fan of **spelling out** arrange/act/assert trio.

Blank lines would do just as well, IMO.

docs/tutorial/index.md
dwilding: My original plan was to merge unfinished docs into main, which wasn't a great plan. New plan: I've retargeted this PR at a branch called new-machine-tutorial. I'll merge PRs into that branch until the doc is ready to publish, then I'll do a PR into main.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I had planned to install uv as a snap, so asking Concierge to install an extra snap sounds good. For tox, I was planning to use `uv tool install tox --with tox-uv`. I've left a TODO in the draft as a reminder:

```
TODO: uv (as an extra snap from Concierge), tox (with `uv tool`)
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I've added a comment:

```
# The collect-status handler will run next and will set status for the user to see.
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Done:

```
# Define a reusable fixture that provides invalid slugs.
@pytest.fixture(params=["", "foo_bar", "foo/bar"])
def invalid_slug(request):
    return request.param


def test_slug_invalid(invalid_slug: str):
    with pytest.raises(ValueError):
        tinyproxy.check_slug(invalid_slug)
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Yes :slightly_smiling_face: in a follow-up PR

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Done:

```
# For convenience, define a reusable fixture that provides a MockTinyproxy object
# and patches the helper module in the charm.
@pytest.fixture
def tinyproxy_installed(monkeypatch: pytest.MonkeyPatch):
    tinyproxy = MockTinyproxy(installed=True)
    patch_charm(monkeypatch, tinyproxy)
    return tinyproxy


def test_start(tinyproxy_installed: MockTinyproxy):
    """Test that the charm correctly handles the start event."""
    ctx = testing.Context(TinyproxyCharm)

    state_out = ctx.run(ctx.on.start(), testing.State())

    assert state_out.unit_status == testing.ActiveStatus()
    assert tinyproxy_installed.is_running()
    assert tinyproxy_installed.config == (PORT, "example")
```

Along with a similar fixture that provides `MockTinyproxy(config=(PORT, "example"), installed=True, running=True)`.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I agree with this. For the first unit test, I've retained comments to explain the structure of a test:

```
def test_install(monkeypatch: pytest.MonkeyPatch):
    """Test that the charm correctly handles the install event."""
    # A state-transition test has three broad steps:
    # Step 1. Arrange the input state.
    tinyproxy = MockTinyproxy()
    patch_charm(monkeypatch, tinyproxy)
    ctx = testing.Context(TinyproxyCharm)
    # Step 2. Simulate an event, in this case an install event.
    state_out = ctx.run(ctx.on.install(), testing.State())
    # Step 3. Verify the output state.
    assert state_out.unit_status == testing.MaintenanceStatus("Waiting for tinyproxy to start")
    assert tinyproxy.is_installed()
```

Then in subsequent tests I've used blank lines.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I've been thinking about this some more. I'd prefer the tutorial to have a single charm, to be more approachable to first-time charmers. Here's the approach I have in mind:

- The tutorial charm will be a principal charm.

- When introducing the workload, we mention that we've chosen a workload that has few requirements. For the tutorial, we focus on the structure of the charm code, rather than the workload.

- Mention that in reality, a tinyproxy charm might support a more complex workload, perhaps helping to expose part of an internal service that's running as a separate charm. (Can briefly mention the terms principal & subordinate).

I do think it would be nice to cover subordinate charms in more detail, including how to structure a relation between subordinate and principal. I've added that to our internal sheet as an idea for next cycle.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I like your challenge too! I was thinking through options:

1. Implement the challenges in the example charm, but not in the tutorial
2. Implement the challenges in the example charm. In the tutorial, briefly mention how to implement and use them (e.g., how to run the action)
3. Have an "extended" example charm that implements the challenges

I don't have a preference at the moment. For now, I've tracked the challenges as a TODO for after the first version of the tutorial is published. I'll remove my note from the draft

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Good point. I've replaced the catch & re-raise by a comment:

```
apt.add_package("tinyproxy-bin")
# If this call fails, the charm will go into error status. The Juju logs will show the error:
# charmlibs.apt.PackageError: Failed to install packages: tinyproxy-bin
```

Here's my general thinking about how to approach error handling in the tutorial charm:

- Make the charm as resilient as reasonably possible. Catch errors if they can be recovered from
- Also catch errors if the user would be able to unblock the charm
- Otherwise, let errors bubble up for a charmer to investigate. Add a comment in the source to call attention to points where errors might happen

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: > you test with curl, maybe it's more complicated in that case and since this is all in the VM maybe it doesn't matter

That was my thinking too, so I'll leave this as-is

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Thanks for reminding me about this best practice. tinyproxy doesn't output anything meaningful, but I've added `capture_output=True` to make sure that nothing falls though to Juju.

Good idea to wait for tinyproxy to run. I've added a `wait_for_running()` method on the charm class, similar to what we do in the Charmcraft K8s profile:

```python
def wait_for_running(self) -> None:
    """Wait for tinyproxy to be running."""
    for _ in range(3):
        if tinyproxy.is_running():
            return
        time.sleep(1)
    raise RuntimeError("tinyproxy was not running within the expected time")
    # Raising a runtime error will put the charm into error status. The error message is for
    # you (the charm author) to see in the Juju logs, not for the user of the charm.
```

I'm calling this method immediately after `tinyproxy.start()`.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Good point, thanks. I've checked that the pidfile disappears when tinyproxy stops. So it looks to be a robust indicator of whether tinyproxy is running.

In addressing one of Tony's comments, I added a method to wait for tinyproxy to start (i.e., waiting for the pidfile). I've just added a corresponding method that waits for tinyproxy to stop:

```python
def wait_for_not_running(self) -> None:
    """Wait for tinyproxy to not be running."""
    for _ in range(3):
        if not tinyproxy.is_running():
            return
        time.sleep(1)
    raise RuntimeError("tinyproxy was still running after the expected time")
```

I'm calling this right after `tinyproxy.stop()`.

dwilding: Now I'm reading your other comment that says "How sure are you that a stale PID file is not left when tinyproxy [...] gets killed hard by the OS?" This probably makes it less robust than I thought. I'm looking into it

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Thanks for the reference and raising this point, both were very helpful. I'll reply in [this thread](https://github.com/canonical/operator/pull/1999/files#r2292818625)

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I tested `int(" 42 \n") == 42`, so I think we're good re strip. But I can add strip if you think it's clearer that way

Re Dima's points: I've gone with returning `int | None` instead. I'll reply in more detail in [this thread](https://github.com/canonical/operator/pull/1999/files#r2292818625)

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: OK, I made some changes to hopefully make this more robust. Thanks again Dima for pointing me in a better direction. Please let me know if this still looks off.

I changed `_get_pid()` to return `int | None`:

```python
def _get_pid() -> int | None:
    """Return the PID of the tinyproxy process, or None if the process can't be found."""
    if not pathops.LocalPath(PID_FILE).exists():
        return None
    pid = int(pathops.LocalPath(PID_FILE).read_text())
    # Check that we can send signals to the process.
    try:
        os.kill(pid, 0)
    except ProcessLookupError:
        return None
    return pid
```

I also changed `is_running()`:

```python
def is_running() -> bool:
    """Return whether tinyproxy is running."""
    return bool(_get_pid())
```

Previously, `is_running()` only checked whether the pidfile existed. As Dima pointed out, this would have erroneously returned True in case of a stale pidfile.

The `stop()` and `reload_config()` functions now check for a truthy pid before trying to send a signal to the process:

```python
def stop() -> None:
    """Stop tinyproxy."""
    pid = _get_pid()
    if pid:
        os.kill(pid, signal.SIGTERM)

def reload_config() -> None:
    """Ask tinyproxy to reload config."""
    pid = _get_pid()
    if not pid:
        raise RuntimeError("tinyproxy is not running")
    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
    os.kill(pid, signal.SIGUSR1)
```

Here's how I stress-tested the revised implementation:

1. Deploy the charm, check that it goes active, then check that curl works.
2. SSH into the unit, read the pid from the pidfile, then `kill -9 <pid>`.
3. Check that the pidfile still exists (it does) and that curl doesn't work (it doesn't).
4. Use `jhack utils ffwd` to speed up Juju noticing that tinyproxy is not running. The charm goes into maintenance status.
5. Use `juju config` to change the charm's configuration, which nudges the charm into restarting tinyproxy. The charm goes active again.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Good idea, this only required a little bit extra code

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Ah, I thought handling the stop event would be enough. We discussed this in Charm Tech daily yesterday. The consensus was to handle stop and remove:

- On stop, stop the reverse proxy
- On remove, uninstall tinyproxy and clean up files

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: We discussed this in Charm Tech daily yesterday and agreed to use Pydantic.

dwilding:
Thank you again @dimaqq & @tonyandrewmeyer for the thorough review! I've worked through your feedback and have made changes for most points.

To keep the tutorial on the simpler side, I don't plan to implement these things:
- Install a specific version of tinyproxy
- Configure the reverse proxy to use HTTPS. I've kept HTTP
- Introduce a subordinate charm or change to a subordinate charm. I've kept it as a principal charm

Rough sequence of next steps:
1. If we're good with this PR, I'll merge it into the WIP branch
2. Add all the code to the WIP branch (as a charm in the examples dir)
3. Handle the remove event: uninstall tinyproxy and clean up files
4. Switch to Pydantic for config validation
5. Add integration tests
6. Complete the tutorial doc, then merge the WIP branch into main
7. Think about follow-up work, e.g., challenges for the reader

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -140,18 +140,30 @@ def start() -> None:
...
     """Ask tinyproxy to reload config."""
-    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
-    os.kill(_get_pid(), signal.SIGUSR1)
-
-
-def _get_pid() -> int:
-    """Return the PID of tinyproxy."""
-    return int(pathops.LocalPath(PID_FILE).read_text())
+    pid = _get_pid()
+    if pid:
```
dimaqq: Arguably, the call should then `.wait(pid)` to make sure that the tiny proxy has finished quitting.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -291,6 +304,14 @@ class TinyproxyCharm(ops.CharmBase):
...
+    def wait_for_not_running(self) -> None:
+        """Wait for tinyproxy to not be running."""
+        for _ in range(3):
+            if not tinyproxy.is_running():
+                return
+            time.sleep(1)
+        raise RuntimeError("tinyproxy was still running after the expected time")
+
 
 if __name__ == "__main__":  # pragma: nocover
```
dimaqq, line 326: This function is only used once, why not inline it in `stop()`?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -140,18 +140,30 @@ def start() -> None:
...
+    if not pathops.LocalPath(PID_FILE).exists():
+        return None
+    pid = int(pathops.LocalPath(PID_FILE).read_text())
+    # Check that we can send signals to the process.
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return None
+    return pid
 ```
```
dimaqq, line 176: **nit:** the comment could be better.

UNIX folks don't need the comment. The new gen needs a better comment.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,647 @@
...
+
+
+def _get_pid() -> int | None:
+    """Return the PID of the tinyproxy process, or None if the process can't be found."""
+    if not pathops.LocalPath(PID_FILE).exists():
+        return None
+    pid = int(pathops.LocalPath(PID_FILE).read_text())
+    # Check that we can send signals to the process.
+    try:
+        os.kill(pid, 0)
```
dimaqq: How about this instead?

```suggestion
    try:
        # Signal 0 doesn’t terminate the process; it just checks if the PID exists.
        os.kill(pid, 0)
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,647 @@
...
+    except ProcessLookupError:
+        return None
+    return pid
+```
+
+Notice that the helper module is stateless. In fact, your charm as a whole will be stateless. The main logic of your charm will:
+
+1. Receive an event from Juju. For example, the `start` event, meaning "it's time to start tinyproxy".
```
dimaqq: This is sadly incorrect.
"start" means this unit has started.
it may or may not be able to start the workload... because the storage is not attached yet.

See https://discourse.charmhub.io/t/charm-lifecycle/5938

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: This part is about sending a signal to tinyproxy to reload the config file, without quitting the process. So I don't think we can wait for anything here. In the part that really sends a signal to quit tinyproxy, the charm code waits to make sure it's stopped

I realise that I should add a clarifying comment here, to say that `kill` is not stopping the process. Similar to your suggestion for `kill(pid, 0)`, which I think is an excellent idea

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Thanks a lot, I wasn't sure what to write here. You've put it very nicely. Updated a couple of places:

```py
def reload_config() -> None:
    """Ask tinyproxy to reload config."""
    pid = _get_pid()
    if not pid:
        raise RuntimeError("tinyproxy is not running")
    # Sending signal SIGUSR1 doesn't terminate the process. It asks the process to reload config.
    # See https://manpages.ubuntu.com/manpages/jammy/en/man8/tinyproxy.8.html#signals
    os.kill(pid, signal.SIGUSR1)

def _get_pid() -> int | None:
    """Return the PID of the tinyproxy process, or None if the process can't be found."""
    if not pathops.LocalPath(PID_FILE).exists():
        return None
    pid = int(pathops.LocalPath(PID_FILE).read_text())
    try:
        # Sending signal 0 doesn't terminate the process. It just checks whether the PID exists.
        os.kill(pid, 0)
    except ProcessLookupError:
        return None
    return pid
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Mainly because I find `return` in the loop clearer than

```
for _ in range(3):
    if tinyproxy.is_running():
        break
    time.sleep(1)
else:
    raise RuntimeError(...)
```

But I admit it's not a huge difference

dwilding: Sorry, this doesn't properly answer your question. I could inline it in `_on_stop()` in charm.py or `stop()` in tinyproxy.py without having to change `return` - my oversight.

To answer your actual question, "why not inline it in `stop()`"? The approach I was taking is that the code in charm.py is responsible for deciding how long to wait for stuff to happen. Somehow aligning with a K8s-based approach where charm.py asks Pebble to do something and then waits to check it's been done.

If you think that's misguided here, I can inline in `stop()` in tinyproxy.py

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: [This doc](https://documentation.ubuntu.com/juju/3.6/reference/hook/#installation-phase) says that for machine charms, storage-attached will always happen before start (and for K8s charms storage-attached happens sometime after start). So maybe it's OK?

In any case, I don't want to mislead people. I've changed the wording to this:

> 1. Receive an event from Juju. Later in the tutorial, we'll explain what this means in practice.

I've also added a TODO for the later part. I'll flesh this out in a follow-up PR:

> TODO: [...] Explain what "receiving an event" from Juju means in practice. Mention that the workload and charm code run on the same machine, called the "unit", but that the charm code is usually not running. Juju runs the charm code on demand, passing data in the environment, which represents an event to be processed. Ops provides a higher-level framework for handling events and responding to Juju.

dwilding:
I'm going to merge this into the WIP branch. I've tracked the [inlining question](https://github.com/canonical/operator/pull/1999#discussion_r2323894733) to come back to before finalising the WIP branch

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dimaqq: Up to you, really.

dimaqq:
Looks good, thanks!

PR #2001, opened 2025-08-20
docs: change some best practices to tips

tonyandrewmeyer:
We're using "best practice" blocks to mean "required for public listing on charmhub". That implies that they should be items that are reasonable to review & resolve at that time, rather than earlier in the charm development process. We're generally using "tip" blocks for suggestions that occur earlier.

This PR changes some "best practice" blocks to tips:

* `logger.info(f"foo {bar}")` rather than `logger.info("foo %s", bar)` (in practice this also doesn't tend to make a lot of difference in the charming case, although it is more correct)
* Absolute paths in `subprocess` calls (this is also less relevant in the charming context than in general Python code)
* Avoid running subprocesses via the shell.

[More context](https://github.com/tonyandrewmeyer/charmhub-listing-review/issues/3).

### DIFF
--- a/docs/howto/run-workloads-with-a-charm-machines.md
+++ b/docs/howto/run-workloads-with-a-charm-machines.md
@@ -78,7 +78,6 @@ Use absolute paths in subprocesses to prevent security issues.
 ```
 
 ```{tip}
-
 Execute processes directly rather than via the shell.
 ```
 
--- a/docs/howto/run-workloads-with-a-charm-machines.md
+++ b/docs/howto/run-workloads-with-a-charm-machines.md
@@ -69,19 +69,15 @@ class MachineCharm(ops.CharmBase):
 ```
 
 ```{tip}
-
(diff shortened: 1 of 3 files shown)


benhoyt (approved):
Sounds good to me, thanks.

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -69,19 +69,15 @@ class MachineCharm(ops.CharmBase):
...
-```{admonition} Best practice
-:class: hint
-
+```{tip}
 Use absolute paths in subprocesses to prevent security issues.
 ```
 
-```{admonition} Best practice
-:class: hint
+```{tip}
```
dwilding: Since you're removing blank lines from other tips

```suggestion
```

dwilding (approved):
I agree with this approach, thanks!

PR #2003, opened 2025-08-22
docs: make a `Layer` instead of a `LayerDict` in the httpbin charm

dwilding:
This PR resolves https://github.com/canonical/charmcraft/issues/1116 by changing the return type of the `_make_pebble_layer` function in the httpbin demo charm. Our preference is to pass around `Layer` objects rather than `LayerDict` objects. See the issue for detailed discussion.

Drive-by: extract the environment variables from the layer definition, into a separate variable.

### DIFF
--- a/examples/httpbin-demo/src/charm.py
+++ b/examples/httpbin-demo/src/charm.py
@@ -129,6 +129,7 @@ class HttpbinDemoCharm(ops.CharmBase):
 
     def _make_pebble_layer(self, log_level: str) -> ops.pebble.Layer:
         """Return a Pebble layer that starts gunicorn with the specified log level."""
+        environment = {'GUNICORN_CMD_ARGS': f'--log-level {log_level}'}
         layer: ops.pebble.LayerDict = {  # Typing as a LayerDict hints the layer's keys.
             'services': {
                 SERVICE_NAME: {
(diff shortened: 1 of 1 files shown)


examples/httpbin-demo/src/charm.py
```diff
@@ -127,21 +127,21 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent):
...
                 SERVICE_NAME: {
                     'override': 'replace',
                     'summary': 'httpbin',
                     'command': 'gunicorn -b 0.0.0.0:80 httpbin:app -k gevent',
                     'startup': 'enabled',
-                    'environment': {'GUNICORN_CMD_ARGS': f'--log-level {log_level}'},
+                    'environment': environment,
                 }
-            },
+            }
```
dwilding: Ruff doesn't seem to care whether there's a comma here. Do we have a preference?

tonyandrewmeyer: I think this is because of [the magic trailing comma behaviour](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma). Basically, if you put a comma there it tells Ruff to format it to a separate line, and if there's no comma that it doesn't have to do that.

I'd add the trailing comma here, personally.

tonyandrewmeyer (approved):
LGTM! Thanks :)

examples/httpbin-demo/src/charm.py
dwilding: Ah, thanks. I suppose there should be trailing commas in two places in that case. After `'services': {},` and after `SERVICE_NAME: {},`

dimaqq (approved):

benhoyt (approved):

PR #2006, opened 2025-08-22
docs: update broken link to HookVars source code

niwamo:


ops/jujucontext.py
```diff
@@ -28,7 +28,7 @@ class _JujuContext:
     """_JujuContext collects information from environment variables named 'JUJU_*'.
 
     Source: https://documentation.ubuntu.com/juju/3.6/reference/hook/#hook-execution.
-    The HookVars function: https://github.com/juju/juju/blob/3.6/worker/uniter/runner/context/context.go#L1398.
+    The HookVars function: https://github.com/juju/juju/blob/3.6/internal/worker/uniter/runner/context/context.go#L1398.
```
benhoyt: If we're going to have a line number we also probably need a permalink (commit hash/tag) rather than branch. Pressing `y` in the GitHub UI gives me:

```suggestion
    The HookVars function: https://github.com/juju/juju/blob/be9dce813b631a469eb8ca8b5c6bc3c0fe82b954/internal/worker/uniter/runner/context/context.go#L1398
```

dimaqq: Btw., short hash also works, e.g. https://github.com/juju/juju/blob/be9dce8/internal/worker/uniter/runner/context/context.go#L1398

GitHub is smart enough, they do like 7-hex git convention :)

dimaqq:
I've renamed the PR, as `fix:` by convention means user-facing functionality fix, not docstrings.

Issue #1994, opened 2025-08-18
Respect style of example charms


dwilding:
Eventually, our [example charms](https://github.com/canonical/operator) will all use the same coding style and linting rules - they'll match what `charmcraft init` produces. This will cause some conflicts with Ops style & rules:
- Single quotes vs double quotes
- Ordering of imports (in example charms, `ops` shouldn't be treated as a local import)

During team discussion, we agreed that it's important to promote consistency among charms, and that charms don't need to 100% match what makes sense for Ops itself.

### Proposed solution

- Exclude example charms from `tox -e format`, `tox -e lint`, and `tox -e static` at the repo-level. We already partially do this. In pyproject.toml, we exclude fetched libs from formatting and linting, and we relax the linting rules for tests.

- Add dedicated environments to the repo's tox.ini file:
    - `format-examples` - Runs `tox -e format` for each example charm
    - `lint-examples` - Runs `tox -e lint` for each example charm (which will include static checks, per [#1972](https://github.com/canonical/operator/pull/1972) and [charmcraft#2405](https://github.com/canonical/charmcraft/pull/2405))

- Update the [Example Charm Tests](https://github.com/canonical/operator/blob/main/.github/workflows/example-charm-tests.yaml) workflow to run `tox -e lint` for each example charm (currently it runs `tox -e unit` for each charm). Tony's already doing this as part of [#1972](https://github.com/canonical/operator/pull/1972)

PR #2008, opened 2025-08-25
docs: use Charmcraft-style format and lint for example charms, not Ops-style

dwilding:
This PR addresses most of #1994. Key changes:

- Exclude the example charms from repo-level formatting and linting.

- Add/update `format` and `lint` environments to the example charms, using tool config that matches what comes from `charmcraft init`.

- Reformat the example charms to match the new environments:

    - For the httpbin charm, the code has switched to double quotes. Apart from this, the only difference is the order of imports. We're no longer treating Ops as a local import.

    - For the K8s tutorial charms, I didn't want to change all the code in the tutorial, so I've asked ruff to use single quotes. For the time being. At some point in the future we'll make the tutorial start from `charmcraft init` - I figure we can switch to double quotes at that point.

        After reformatting, there were a few differences in addition to the new order of imports. I've reflected those in the tutorial too.

- Pre-commit was trying to format code using `ruff --preview`, which caused a conflict with some of the new formatting in the tutorial charms. So I've excluded the `examples` directory from the relevant pre-commit hooks.

---

**Not implemented:** In #1994, I'd mentioned creating tox environments `format-examples` and `lint-examples` for running `tox -e format` and `tox -e lint` in all the example charm directories. I experimented with this and it turned out to be more complicated than I thought. It worked, but there seemed to be conflicts from running sub-environments with the same name as a top-level environment. Whenever I ran `tox -e format` or `tox -e format-examples`, I got messages like:

```
format: recreate env because env type changed from {'name': 'format', 'type': 'UvVenvLockRunner'} to {'name': 'format', 'type': 'UvVenvRunner'}
format: remove tox env folder .../operator/.tox/format
```

I don't think it's worth the time/complexity of trying to set this up in a robust way. We already have a workflow that runs `tox -e lint` (and `tox -e unit`) for all the example charms.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,9 +21,11 @@ repos:
     rev: v0.11.2
     hooks:
       - id: ruff
-        args: [ --config, pyproject.toml ]
+        args: [ --preview ]
+        exclude: '^examples/'
       - id: ruff-format
-        args: [ --config, pyproject.toml ]
+        args: [ --preview ]
+        exclude: '^examples/'
   # Spellcheck the code.
   - repo: https://github.com/codespell-project/codespell
     rev: v2.4.1


tox.ini
```diff
@@ -72,7 +72,7 @@ dependency_groups = lint, unit
 commands =
     ruff check --preview
     ruff format --preview --check
-    codespell --skip ./docs/_build {posargs}
+    codespell {posargs}
```
dwilding, line 75: Drive-by: `./docs/_build` was excluded twice - with `--skip` here and with `skip` in `pyproject.toml` - so I removed this one.

.github/workflows/example-charm-tests.yaml
```diff
@@ -47,6 +41,12 @@ jobs:
           if grep -Eq "^charm-libs:" charmcraft.yaml; then
             charmcraft fetch-libs
           fi
+      - name: Install tox
+        run: pip install tox~=4.2
+      - name: Run linting and static checks
+        run: |
+          cd ${{ matrix.dir }}
+          tox -e lint
```
dwilding, line 49: Now that the K8s tutorial charms have real `lint` environments that include static checks, we need to run `tox -e lint` after fetching any libs.

examples/httpbin-demo/pyproject.toml
```diff
@@ -1,3 +1,41 @@
-# Static checks configuration
+# Testing tools configuration
```
dwilding, line 1: This config is based on what comes from `charmcraft init`. If there's anything objectionable, probably we ought to remove from the profiles too.

tonyandrewmeyer:
> I predict Tony may want to remove McCabe from charmcraft init template, which I'm on board with -- but I guess that would be separate from this PR anyway.

Correct on both counts. I would like Charm Tech to be more opinionated about linting settings (people can always change them later, but I think we should have a better default set in the profiles). But not in this PR, probably not in this cycle.

.pre-commit-config.yaml
```diff
@@ -21,9 +21,9 @@ repos:
     rev: v0.11.2
     hooks:
       - id: ruff
-        args: [ --preview ]
+        args: [ --config, pyproject.toml ]
       - id: ruff-format
-        args: [ --preview ]
+        args: [ --config, pyproject.toml ]
```
tonyandrewmeyer: This probably makes pre-commit consistent with the example charms, but makes it inconsistent with the ops[...] packages, doesn't it? We could discuss dropping `--preview` for ops[...] but I still think it's nice to have the latest checks.

More generally, if `tox -e format` and `tox -e lint` at the top level don't include the example charms, then I don't think pre-commit should either.

examples/httpbin-demo/pyproject.toml
tonyandrewmeyer: I do think there are improvements to be made, but I think those should be done separately (and then copied here).

.pre-commit-config.yaml
dwilding: Ah, thanks. I was getting confused when I made this change.

What I wanted to achieve was: Exclude the example charms from pre-commit, by passing `--config pyproject.toml`. In pre-commit, ruff doesn't seem to auto-discover this, I don't know why...

I did achieve that. But, I mistakenly thought that `--preview` was specified in `pyproject.toml`. When actually it comes from `tox.ini`, doesn't it.

So I think this works - what do you think?

```
      - id: ruff
        args: [ --preview, --config, pyproject.toml ]
      - id: ruff-format
        args: [ --preview, --config, pyproject.toml ]
```

tonyandrewmeyer: What about just having an `exclude: [ '^examples/' ]` section instead?

dwilding: That works :slightly_smiling_face: I've just updated the hooks config to:

```
      - id: ruff
        args: [ --preview ]
        exclude: '^examples/'
      - id: ruff-format
        args: [ --preview ]
        exclude: '^examples/'
```

PR #2011, opened 2025-08-27
docs: use uv for testing and packing the httpbin charm

dwilding:
This PR completes #1805. Changes:

- Put charm dependencies in `pyproject.toml` instead of `requirements.txt`
- Put testing dependencies `pyproject.toml` instead of `tox.ini`
- Add `uv.lock`
- Switch `charmcraft.yaml` to use the uv plugin
- Update CI to support/expect a uv-based charm:
    - Example Charm Tests
    - Charmcraft Pack Test

For both `pyproject.toml` and `tox.ini`, I'm using the setup that will come from `charmcraft init` in Charmcraft 4. If anything needs adjusting, we should consider adjusting the templates too.

:heavy_check_mark: all charm tests have passed, including integration tests

### DIFF
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -17,20 +17,23 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Set up uv
+        uses: astral-sh/setup-uv@e92bafb6253dcd438e0484186d7669ea7a8ca1cc  # v6.4.3
+
       - name: Update 'ops' dependency in test charm to latest
         run: |
-          cd examples/httpbin-demo
-          sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" requirements.txt
           if [ -z "$CLONE_SHA" ]
           then
-            echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
+            ops_source="ops @ git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops"
           else
             # If on a PR, we need to reference the PR branch's repo and commit (not the GITHUB_SHA
             # temporary merge commit), because charmcraft pack does a git checkout which
             # can't see the temporary merge commit.
-            echo -e "\ngit+$CLONE_URL@$CLONE_SHA#egg=ops" >> requirements.txt
+            ops_source="ops @ git+$CLONE_URL@$CLONE_SHA#egg=ops"
           fi
-          cat requirements.txt
+          cd examples/httpbin-demo
+          uv add "$ops_source"
+          cat pyproject.toml
         env:
           CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
           CLONE_SHA: ${{ github.event.pull_request.head.sha }}
@@ -41,8 +44,8 @@ jobs:
       - name: Add 'git' as a build package
         run: |
           cd examples/httpbin-demo
-          if yq --exit-status 'has("parts")' charmcraft.yaml; then
-            echo "'parts' already exists in charmcraft.yaml"
+          if yq --exit-status '.parts.charm | has("build-packages")' charmcraft.yaml; then
+            echo "'parts.charm.build-packages' already exists in charmcraft.yaml"
             exit 1
           fi
           yq --inplace '.parts.charm.build-packages = ["git"]' charmcraft.yaml
--- a/examples/httpbin-demo/tox.ini
+++ b/examples/httpbin-demo/tox.ini
@@ -15,7 +15,7 @@
 [tox]
 no_package = True
 skip_missing_interpreters = True
-env_list = lint, unit
+env_list = format, lint, unit
 min_version = 4.0.0
 
 [vars]
@@ -42,15 +42,12 @@ commands =
     ruff check --fix {[vars]all_path}
 
 [testenv:lint]
-description = Run linting and type checks
-deps =
-    ruff
-    codespell
-    pyright
-    ops[testing]
-    -r {tox_root}/requirements.txt
-    pytest
-    jubilant
+description = Check code against coding style standards, and static checks
+runner = uv-venv-lock-runner
+dependency_groups =
+    lint
+    unit
+    integration
 commands =
     codespell {tox_root}
     ruff check {[vars]all_path}
@@ -59,11 +56,9 @@ commands =
 
 [testenv:unit]
 description = Run unit tests
-deps =
-    pytest
-    coverage[toml]
-    ops[testing]
-    -r {tox_root}/requirements.txt
+runner = uv-venv-lock-runner
+dependency_groups =
+    unit
 commands =
     coverage run --source={[vars]src_path} -m pytest \
         -v \
@@ -75,10 +70,9 @@ commands =
 
 [testenv:integration]
 description = Run integration tests
-deps =
-    pytest
-    jubilant
-    -r {tox_root}/requirements.txt
+runner = uv-venv-lock-runner
+dependency_groups =
+    integration
 commands =
     pytest \
         -v \
--- a/examples/httpbin-demo/uv.lock
+++ b/examples/httpbin-demo/uv.lock
@@ -1,0 +1,450 @@
+version = 1
+revision = 3
+requires-python = ">=3.10"
+
+[[package]]
+name = "codespell"
+version = "2.4.1"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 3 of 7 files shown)


dwilding:
Putting this in draft status while I fix CI

examples/httpbin-demo/tox.ini
```diff
@@ -15,7 +15,7 @@
 [tox]
 no_package = True
 skip_missing_interpreters = True
-env_list = lint, unit
+env_list = format, lint, unit
```
dwilding, line 18: This is an unrelated change. We're inheriting it from the upcoming Charmcraft profile

.github/workflows/charmcraft-pack.yaml
```diff
@@ -17,20 +17,23 @@ jobs:
...
           else
             # If on a PR, we need to reference the PR branch's repo and commit (not the GITHUB_SHA
             # temporary merge commit), because charmcraft pack does a git checkout which
             # can't see the temporary merge commit.
-            echo -e "\ngit+$CLONE_URL@$CLONE_SHA#egg=ops" >> requirements.txt
+            ops_source="ops @ git+$CLONE_URL@$CLONE_SHA#egg=ops"
           fi
-          cat requirements.txt
+          cd examples/httpbin-demo
+          uv add "$ops_source"
```
dwilding, line 35: This adds a `tool.uv.sources` entry in `pyproject.toml` for Ops. I did a local pack and inspected venv inside the charm to verify that Charmcraft will get Ops from GitHub instead of PyPI

james-garner-canonical (approved):
Look good!

examples/httpbin-demo/uv.lock
```diff
@@ -0,0 +1,450 @@
...
+dependencies = [
+    { name = "importlib-metadata" },
+    { name = "opentelemetry-api" },
+    { name = "pyyaml" },
+    { name = "websocket-client" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/e0/fb/7ec08c56511b108cedd92b2eb978419d982b0b87776d8e860cd981e10b9e/ops-3.1.0.tar.gz", hash = "sha256:49bc8ce4f24815dae57aec752d7d0c9cd34e7942c29966f29a9f9774d1b97512", size = 531550, upload-time = "2025-07-30T02:26:49.875Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/67/c4/db1450da4cb5ac49da1a814bf6b6ffcff873dfd15abd39520370d85848ab/ops-3.1.0-py3-none-any.whl", hash = "sha256:4553052c881c9e2ba76209dc01be48915330e4d05287fca0fc5ae05ae0d20f0a", size = 188440, upload-time = "2025-07-30T02:26:44.945Z" },
+]
```
dimaqq, line 238: I guess that's fair, lock current ops from PYPI, then switch to local ops.

examples/httpbin-demo/tox.ini
```diff
@@ -75,10 +70,9 @@ commands =
 
 [testenv:integration]
 description = Run integration tests
-deps =
-    pytest
-    jubilant
-    -r {tox_root}/requirements.txt
+runner = uv-venv-lock-runner
```
dimaqq, line 73: IMO, it would make sense to specify base_python here or in `testenv` (it's inherited, but only level, see tox.ini at top level for ops).

examples/httpbin-demo/tox.ini
```diff
@@ -59,11 +56,9 @@ commands =
 
 [testenv:unit]
 description = Run unit tests
-deps =
-    pytest
-    coverage[toml]
-    ops[testing]
-    -r {tox_root}/requirements.txt
+runner = uv-venv-lock-runner
```
dimaqq, line 59: Given two bases in charmcraft.yaml it would make sense to test against two versions of Python.
Then again, we didn't do that previously, so maybe leave that out for now.

.github/workflows/charmcraft-pack.yaml
dimaqq: How about `ops[testing]` for the unit tests?

examples/httpbin-demo/tox.ini
dwilding: We have `requires-python = ">=3.10"` [in pyproject.toml](https://github.com/canonical/operator/blob/13e736b320622073aa512ebbffd450b72aadfc4e/examples/httpbin-demo/pyproject.toml#L18) - does that cover it?

examples/httpbin-demo/tox.ini
dwilding: This part of `charmcraft.yaml`?

```
bases:
  - build-on:
    - name: ubuntu
      channel: "22.04"
    run-on:
    - name: ubuntu
      channel: "22.04"
```

Doesn't this mean we only have to worry about a single base, 22.04? Unless I misunderstood

In the "Example Charm Tests" workflow we already have:

```
matrix:
  python-version: ["3.10", "3.12"]
```

We don't do this in the "Charmcraft Pack Test" workflow, though. Maybe we should matrix Python versions there too? As I understand it, that workflow is more about testing that a charm (any charm) can be packed with pre-release Ops, rather than testing the charm itself, so I'm not sure what would be appropriate. Separate PR?

.github/workflows/charmcraft-pack.yaml
dwilding: As I understand it, the purpose of this workflow is to check that Charmcraft is able to pack some charm using pre-release Ops. So we're not testing the charm itself here. For the httpbin charm, the "Example Charm Tests" workflow runs the charm's unit tests.

I seem to remember that we discussed whether "Example Charm Tests" should use `ops` and `ops[testing]` from GitHub instead of PyPI (belt and braces...), but I think we decided it wasn't worth worrying about.

examples/httpbin-demo/tox.ini
dimaqq: I don't know for sure (you can test with `requires-python = ">3.15"` maybe), but I don't think that pyproject.toml covers it.

I think that tox decides on Python version first, then create a venv, then instructs `uv` to fill that venv with packages.

examples/httpbin-demo/tox.ini
dimaqq: totally fine to defer to a separate PR, or even never :)

examples/httpbin-demo/tox.ini
dwilding: I just tried it:

With `requires-python = ">3.15"` in `pyproject.toml` I get:

```
$ tox -e unit
unit: uv-sync> uv sync --locked --python-preference system --no-default-groups --group unit -p /home/david.wilding@canonical.com/.local/share/uv/tools/tox/bin/python3
Using CPython 3.12.3 interpreter at: /home/david.wilding@canonical.com/.local/share/uv/tools/tox/bin/python3
error: The requested interpreter resolved to Python 3.12.3, which is incompatible with the project's Python requirement: `>3.15`
unit: exit 2 (0.01 seconds) /home/david.wilding@canonical.com/p1/operator/dwilding-httpbin-uv/examples/httpbin-demo> uv sync --locked --python-preference system --no-default-groups --group unit -p /home/david.wilding@canonical.com/.local/share/uv/tools/tox/bin/python3 pid=2062634
  unit: FAIL code 2 (0.01 seconds)
  evaluation failed :( (0.05 seconds)
```

Then I reverted to `requires-python = ">=3.10"` and tried this:

```
$ uv run --python 3.8 tox -e unit
Using CPython 3.8.20
error: The requested interpreter resolved to Python 3.8.20, which is incompatible with the project's Python requirement: `>=3.10` (from `project.requires-python`)
```

So it looks like we're getting everything we need from `pyproject.toml`?

dimaqq: What this means is that:
- Tox picks a Python version by its own rules (probably system default Python)
- Tox creates venv with that Python
- then uv is asked to install the packages into that venv
- uv agrees (or rejects) the venv depending on pyproject.toml setting

The good thing about this is that ultimately, either the command fails or the unit tests are run with some compatible Python.

The bad part is that the Python chosen depends on the user environment. We can easily end up in a situation where it's 3.14rc2 for me and 3.12.11 in GitHub actions.

Imagine that some Python 3.12 feature in the code, the unit tests pass, but when the charm is built for 22.04 base, that feature is not available in the system Python 3.10. 

The GitHub action matrix will pick that up.
(More pragmatically, `ruff` and `pyright` will pick up, albeit only 99% of the time, as static analysis tools are never perfect).

I dunno if this is a pattern we want to recommend to charmers, but I guess it's good enough for the example charms.

examples/httpbin-demo/tox.ini
dimaqq: Snapcraft pack should have a matrix of bases, not Python versions.
Today it does only this: `Packed httpbin-demo_ubuntu-22.04-amd64.charm`
Then again, it's an example charm, we can just let it be.

.github/workflows/charmcraft-pack.yaml
dimaqq: Copy-pasting offline discussion:

> Thing is, we're using the httpbin charm in two different CI tests. One to check that charmcraft pack works with the latest ops (nothing specific to this charm) and one to actually check the charm itself
> For checking the charm itself, we previously decided that it wasn't worth using local ops

dimaqq (approved):

PR #2014, opened 2025-08-27
ci: Use a consistent format for release notes and changelog lines

james-garner-canonical:
This PR updates the release script to use the same format for changelog lines as the release notes. This will make it easy to copy improvements and corrections to the changelog from the release PR review to the published release notes. Presumably the script could actually be simplified a bit since we're using the same format in both places, but this PR currently optimises for simplicity in the change itself.

### DIFF
--- a/release.py
+++ b/release.py
@@ -292,7 +292,7 @@ def format_changes(categories: dict[str, list[tuple[str, str]]], tag: str) -> st
                 match = re.match(r'https?://[^ ]+/pull/(\d+)', pr_link)
                 if match:
                     pr_num = match.group(1)
-                lines.append(f'* {description} (#{pr_num})')
+                lines.append(f'* {description} in {pr_link}')
             lines.append('')
     return '\n'.join(lines) + '\n'
 
--- a/release.py
+++ b/release.py
@@ -288,10 +288,6 @@ def format_changes(categories: dict[str, list[tuple[str, str]]], tag: str) -> st
         if items:
             lines.append(f'## {commit_type_to_category(commit_type)}\n')
             for description, pr_link in items:
-                pr_num = '?'
-                match = re.match(r'https?://[^ ]+/pull/(\d+)', pr_link)
-                if match:
-                    pr_num = match.group(1)
                 lines.append(f'* {description} in {pr_link}')
             lines.append('')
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -288,11 +288,7 @@ def format_changes(categories: dict[str, list[tuple[str, str]]], tag: str) -> st
         if items:
             lines.append(f'## {commit_type_to_category(commit_type)}\n')
             for description, pr_link in items:
-                pr_num = '?'
-                match = re.match(r'https?://[^ ]+/pull/(\d+)', pr_link)
-                if match:
-                    pr_num = match.group(1)
-                lines.append(f'* {description} (#{pr_num})')
+                lines.append(f'* {description} in {pr_link}')
```
tonyandrewmeyer, line 291: What about this?

```suggestion
                lines.append(f'* {description} in [#{pr_num}]({pr_link})')
```

This isn't *textually* the same as the release notes, but it's *effectively* the same, because of the auto-linking that GitHub does. It'll look a little worse as plain text (a lot worse than just the number), but nicer in rendered Markdown (also compared to just the number).

tonyandrewmeyer (approved):
In terms of `#num` versus `http://github.com/canonical/operator/issues/num` I think we had it backwards this whole time (I think I've said that somewhere already, and I think it was originally my mistake). GitHub will automatically expand `#num` into the full link, so it's fine to use the shorthand there -- but it will also automatically shorten the full URL as well, so there's not a lot of point. However, the changelog is plain Markdown so the full URL makes sense.

In terms of "in", I think `(#num)` is much cleaner than `in {link}` but I can get behind a simplification argument based on using the same format as GitHub's draft.

release.py
james-garner-canonical: The developer experience benefits (easily sync release notes and changelog when editing after automatic generation) come from the content being textually the same. This does look nice though, and we could get textual equality by making the same change to the release notes. Since we'd be making a change to the generated release notes anyway, I'd be on board with using `()` instead of `in` as well, so the line would this in both cases:
```py
                lines.append(f'* {description} ([#{pr_num}]({pr_link}))')
```

james-garner-canonical:
Closing in favour of the PR that may come out of discussion in daily sync.

PR #2017, opened 2025-08-28
ci: nicer logging output in the release script using rich

james-garner-canonical:
This PR adds coloured output for the logging in the `release.py` script, using `rich.logging`. This will be especially nice for the draft release, as it will more clearly separate the logging output visually from the release notes being printed. I tested this change by generating the post-release PR using this version of the script. I also tested that it runs cleanly for the draft release, bailing out with a keyboard interrupt on user input.

Before:
image

After:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,7 @@ integration = [
 ]
 release = [
     "pygithub~=2.6",
+    "rich~=14.1",
 ]
 
 [project.urls]
--- a/release.py
+++ b/release.py
@@ -28,14 +28,12 @@ import sys
 import github
 import github.GitRelease
 import github.Repository
+import rich.logging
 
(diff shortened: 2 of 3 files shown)


benhoyt (approved):
Fine with me. Can you show a before and after screenshot, just to refresh our memories of how it looked previously?

james-garner-canonical:
Here's an example of some logging from the `release.py` script without `rich.logging`.
image

james-garner-canonical:
Oh actually, the image I included in the PR body is not a good one, that was an intermediate version -- here's the actual output with `rich.logging`.
image

james-garner-canonical:
I've amended the PR description with accurate before and after screenshots.

dimaqq (approved):

PR #2018, opened 2025-08-28
docs: in httpbin charm integration tests, add env var for charm file to deploy

dwilding:
This PR updates the integration tests of the httpbin charm to include an environment variable `CHARM_PATH`:

- If `CHARM_PATH` is not set, `tox -e integration` looks for a `.charm` file, deploys that charm, then runs the tests.

- If `CHARM_PATH` is set, `tox -e integration` deploys the specified charm file then runs the tests. For example:

    ```
    CHARM_PATH=mycharm.charm tox -e integration
    ```

In either case, `charmcraft pack` needs to have been run before `tox -e integration`.

### DIFF
--- a/examples/httpbin-demo/tests/integration/conftest.py
+++ b/examples/httpbin-demo/tests/integration/conftest.py
@@ -34,9 +34,9 @@ def juju(request: pytest.FixtureRequest):
 def charm():
     charm_path = os.getenv("CHARM_PATH")
     if charm_path is not None:
+        if not pathlib.Path(charm_path).exists():
+            raise FileNotFoundError(f"Charm does not exist: {charm_path}")
         return pathlib.Path(charm_path)
-        # TODO: Raise an error if the charm doesn't exist?
-        #       Will juju.deploy() raise for us?
 
     subprocess.check_call(["charmcraft", "pack"])
     return next(pathlib.Path(".").glob("*.charm"))
--- a/examples/httpbin-demo/tests/integration/conftest.py
+++ b/examples/httpbin-demo/tests/integration/conftest.py
@@ -40,5 +40,6 @@ def charm():
     if not charm_paths:
         raise FileNotFoundError("No .charm file in current directory")
     if len(charm_paths) > 1:
-        raise ValueError("More than one .charm file in current directory")
+        path_list = ", ".join(str(path) for path in charm_paths)
+        raise ValueError(f"More than one .charm file in current directory: {path_list}")
     return charm_paths[0]
--- a/examples/httpbin-demo/tests/integration/conftest.py
+++ b/examples/httpbin-demo/tests/integration/conftest.py
@@ -32,5 +32,11 @@ def juju(request: pytest.FixtureRequest):
 
 @pytest.fixture(scope="session")
 def charm():
+    charm_path = os.getenv("CHARM_PATH")
+    if charm_path is not None:
+        return pathlib.Path(charm_path)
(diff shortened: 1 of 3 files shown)


examples/httpbin-demo/tests/integration/test_charm.py
```diff
@@ -20,7 +20,7 @@
 
 logger = logging.getLogger(__name__)
 
-METADATA = yaml.safe_load(Path("./charmcraft.yaml").read_text())
+METADATA = yaml.safe_load(Path("charmcraft.yaml").read_text())
```
dwilding: I figure we might as well change this while we're here

examples/httpbin-demo/tox.ini
```diff
@@ -79,6 +79,8 @@ deps =
     pytest
     jubilant
     -r {tox_root}/requirements.txt
+pass_env =
```
dwilding, line 82: To support tox 4, must be `pass_env` **not** `passenv`. I can't find this documented anywhere

examples/httpbin-demo/tests/integration/conftest.py
```diff
@@ -31,5 +32,11 @@ def juju(request: pytest.FixtureRequest):
 
 @pytest.fixture(scope="session")
 def charm():
-    subprocess.check_call(["charmcraft", "pack"])  # noqa
```
dwilding, line 34: `#noqa` was previously needed because of Ops linting rules, but we've decoupled the charm from Ops rules now

examples/httpbin-demo/tox.ini
```diff
@@ -79,6 +79,10 @@ deps =
...
+    # The integration tests pack the charm before deploying it.
+    # To skip the packing step, set CHARM_PATH to the location of a .charm file.
+    CHARM_PATH
```
dwilding: I'm using `pass_env` rather than `passenv` for consistency with earlier in the file. If the names don't match, `CHARM_PATH` won't be passed through

examples/httpbin-demo/tests/integration/conftest.py
```diff
@@ -31,5 +31,14 @@ def juju(request: pytest.FixtureRequest):
...
+    if "CHARM_PATH" in os.environ:
+        charm_path = pathlib.Path(os.environ["CHARM_PATH"])
+        if not charm_path.exists():
+            raise FileNotFoundError(f"Charm does not exist: {charm_path}")
+        return charm_path
+    charm_paths = list(pathlib.Path(".").glob("*.charm"))
+    if not charm_paths:
+        raise FileNotFoundError("No .charm file in current directory")
+    if len(charm_paths) > 1:
+        raise ValueError("More than one .charm file in current directory")
```
benhoyt: Nit: might be helpful to show what they are, something like:
```suggestion
        raise ValueError("More than one .charm file in current directory: " + ', '.join(str(p) for p in charm_paths))
```

benhoyt (approved):
Looks good, one nit comment.

examples/httpbin-demo/tests/integration/conftest.py
dwilding: Good idea, thanks. I've gone with this (easier to read after ruffing):

```py
path_list = ", ".join(str(path) for path in charm_paths)
raise ValueError(f"More than one .charm file in current directory: {path_list}")
```

tonyandrewmeyer (approved):
Looks good to me - I think things are definitely heading towards packing happening outside of the tests as a prep step, so this aligns well with that.

PR #2023, opened 2025-08-31
docs: improve intro to ops.testing reference

### LINKED ISSUE #1826: Improve discoverability of how to test relations
The how-to guide for relations has a section called [Write unit tests](https://ops.readthedocs.io/en/latest/howto/manage-relations.html#test-the-feature). But if I'm starting at the [ops.testing reference](https://ops.readthedocs.io/en/latest/reference/ops-testing.html), I wouldn't know that.

One solution would be to add a link from [Relation](https://ops.readthedocs.io/en/latest/reference/ops-testing.html#ops.testing.Relation) to the how-to guide. (That link would need to be a full URL, not an intersphinx link, so that it is usable in IDE help).

More generally, Relation probably isn't the only place to do this. Do we need something obvious at the very beginning of the `ops.testing` reference, which explains that the various how-to guides explain how to test different features?

dwilding:
Resolves #1826. This PR makes a few small improvements to the intro of the `ops.testing` reference:

- Mention that how-to guides have info about testing each feature
- Link to the unit testing how-to guide and the testing explanation (moved from the end of the intro)
- Mention the term "unit tests" much earlier
- Move the "previously known as 'Scenario'" part to the end of the intro, as Scenario is already mentioned in the title

**[Preview doc](https://canonical-ubuntu-documentation-library--2023.com.readthedocs.build/ops/2023/reference/ops-testing/)**

---

Not implemented:

- In [#1826](https://github.com/canonical/operator/issues/1826), I mentioned possibly adding a link from [`Relation`](https://documentation.ubuntu.com/ops/latest/reference/ops-testing/#ops.testing.Relation) to the how-to guide. But why only from `Relation`? I'm adding the new info right at the beginning of the intro, so let's first see how well that works.

- I also mentioned using full URLs for links, so that the links work in docstrings. We don't actually need to do that here. The docstring of `ops.testing` comes from `ops/testing.py` not the intro of `docs/reference/ops-testing.rst` - that intro is purely for the docs.

### DIFF
--- a/docs/reference/ops-testing.rst
+++ b/docs/reference/ops-testing.rst
@@ -14,10 +14,14 @@ See also:
 - :doc:`/howto/write-unit-tests-for-a-charm`
 - :doc:`/explanation/testing` - A summary of types of charm tests
 
-State-transition tests, also called 'unit' tests, expect you to define the
+State-transition tests expect you to define the
 Juju state all at once, define the Juju context against which to test the charm,
 and fire a single event on the charm to execute its logic. The tests can then
 assert that the Juju state has changed as expected.
+
+Unlike integration tests, state-transition tests do not deploy your charm with a real Juju
+controller and model. Instead, your charm code is run using the 'Scenario' testing framework.
+These are the primary form of unit tests that you should write for a charm.
 
 A very simple test, where the charm has no config, no relations, the unit
 is the leader, and has a `start` handler that sets the status to active might
@@ -38,8 +42,7 @@ how the state changes in reaction to events. They are not
 necessarily tests of individual methods or functions;
 they are testing the 'contract' of the charm: given
 a certain state, when a certain event happens, the charm should transition to
-another state. Unlike integration tests, they do not test using a real Juju
-controller and model, and focus on a single Juju unit.
+another state.
 
 Writing these tests should nudge you into thinking of a charm as a black-box
 'input to output' function. The inputs are:
--- a/docs/reference/ops-testing.rst
+++ b/docs/reference/ops-testing.rst
@@ -72,10 +72,6 @@ A test consists of three broad steps:
     - verify that the output state is what you expect it to be
     - verify that the charm has seen a certain sequence of statuses, events, and `juju-log` calls
 
-.. note::
-    Unit testing is only one aspect of a comprehensive testing strategy. For more
-    on testing charms, see :doc:`/explanation/testing`.
-
 
 ..
    _The list here is manually maintained, because the `automodule` directive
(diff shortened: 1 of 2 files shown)


docs/reference/ops-testing.rst
```diff
@@ -6,7 +6,15 @@
...
-State-transition tests, previously known as 'Scenario', expect you to define the
+To learn how to test a particular feature, such as relations, see the relevant how-to guide.
+For example :doc:`/howto/manage-relations` > Test the feature.
+
+See also:
+
+- :doc:`/howto/write-unit-tests-for-a-charm`
+- :doc:`/explanation/testing` - A summary of types of charm tests
+
+State-transition tests, also called 'unit' tests, expect you to define the
```
james-garner-canonical: This reads as defining 'unit tests' === 'state-transition tests'. The intent of the previous version of this line was to say 'scenario' === 'state-transition tests'. If the intent is to remove references to 'scenario', how about 'state-transition unit tests'? Otherwise perhaps mention `ops-scenario`/`ops[testing]`/etc?

IronCore864:
It would be nice if the "Test the feature" part worked as a link, but I played with it a little bit, and it doesn't...

docs/reference/ops-testing.rst
dwilding: Good point, the terms are getting a bit confused here. How about if I modify the first paragraph about state-transition tests and include a second paragraph:

> State-transition tests expect you to define the Juju state all at once, define the Juju context against which to test the charm, and fire a single event on the charm to execute its logic. The tests can then assert that the Juju state has changed as expected.
>
> Unlike integration tests, state-transition tests do not test using a real Juju controller and model. They focus on a single Juju unit. For this reason, state-transition tests are a form of 'unit' tests. In fact, they are the primary form of unit tests that you should write for a charm.

For the second paragraph, I'm moving & adapting some wording from earlier in the intro.

james-garner-canonical: First paragraph looks great. Nice changes overall, but I'm not sure about the term 'a single Juju unit'. What's a Juju unit?  A charm? How about something like this for the second paragraph.

> Unlike integration tests, state-transition tests do not deploy your charm with a real Juju controller and model. Instead, your charm code is run using the 'scenario' testing framework. This is the primary form of unit tests that you should write for a charm.

dwilding: Nice, I like this, thanks. Updated: https://canonical-ubuntu-documentation-library--2023.com.readthedocs.build/ops/2023/reference/ops-testing/

dwilding:
> It would be nice if the "Test the feature" part worked as a link

Good idea, I've added a cross-reference target so that it can become a link. Updated: https://canonical-ubuntu-documentation-library--2023.com.readthedocs.build/ops/2023/reference/ops-testing/

james-garner-canonical (approved):

PR #2029, opened 2025-09-03
docs: update the holistic/delta explanation with the reconciler pattern

dimaqq:
Update the holistic/delta explanation doc to:

- unify "holistic" and "reconciler" terminology
- update the holistic example to a true reconciler
- express why holistic charms are more robust
- express when delta charms are suitable
- update examples
- update links to existing charms

## New TOC

<picture>
  <source media="(prefers-color-scheme: light)" srcset="https://github.com/user-attachments/assets/9711c23a-a6c1-4aa5-bb5f-67faa9e0c9b2">
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/d2504d3b-16a1-4855-9e84-c51523d116c6">
  TOC preview
</picture>

## Live link

https://canonical-ubuntu-documentation-library--2029.com.readthedocs.build/ops/2029/explanation/holistic-vs-delta-charms/

### DIFF
--- a/docs/explanation/holistic-vs-delta-charms.md
+++ b/docs/explanation/holistic-vs-delta-charms.md
@@ -1,7 +1,7 @@
 (holistic-vs-delta-charms)=
 # Holistic vs delta charms
 
-The charming community developed two approaches to writing robust, stateless charms: "holistic" and "delta".
+The charming community has developed two approaches to writing robust, stateless charms: "holistic" and "delta".
 
 - Holistic charms reconcile towards a goal state on every event.
 - Delta charms handle each Juju event individually.
@@ -33,23 +33,23 @@ Conceptually, a Juju event informs the unit that a specific thing has changed in
   - Consumer requirements are defined
 ```
 
-Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed.
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that one or more configuration values have changed, `relation-changed` says that relation data has changed.
 
 However, this only goes so far.
-The `config-changed` event doesn't tell the charm which configuration keys have changed or whether configuration has become valid.
-The `relation-changed` event doesn't tell the charm how the relation data has changed or whether the data can be parsed.
+The `config-changed` event doesn't tell the charm which configuration keys have changed.
+The `relation-changed` event doesn't tell the charm how the relation data has changed.
 
 In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready.
-In such cases, a delta charm may defer the event, effectively storing a small amount of state.
+In such cases, a delta charm may defer the event, effectively storing a small amount of state, or have both `config-changed` and `pebble-ready` check if everything is ready and then run the common code, which is the first step towards a holistic approach.
 
 The holistic approach side-steps this disparity and applies a single code path, `_reconcile`, to all events of interest.
 The code path reads all state it needs to perform computation, and updates all writable state.
 
 ### Example
 
-The reconciler method above has been reduced for clarity, but is representative of the common reconciler pattern.
+The reconciler method below has been reduced for clarity, but is representative of the common reconciler pattern.
 
-The body of the reconciler loop comprises three main parts:
+The body of the reconciler method comprises three main parts:
 
 - reading the inputs (configuration, workload, and Juju environment state)
 - computing the new state
@@ -107,7 +107,7 @@ def _reconcile(self, _: ops.EventBase):
     workload_ready = self.workload.is_ready
     foo_ready = self.foo_requirer.is_ready
     bar_ready = self.bar_provider.is_ready
-    
+
     if not workload_ready or not foo_ready or not bar_ready:
         # Status will be set in `_on_collect_unit_status`
         return
@@ -131,7 +131,7 @@ Rarely, the goal state may change:
 - unit lifecycle end, that is, graceful shutdown
 - failover to another unit, leaving this unit on stand-by
 
-Complex charms often split the reconciler method into functional constituents. See [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
+Complex charms often split the reconciler method into functional constituents. See the [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
 
 Charm libraries too can be written using the reconciler pattern. See [the implementation of `ops.tracing.Tracing`](https://github.com/canonical/operator/blob/af6764fbfdcf8ed42a0edb331870dd9d37dc804b/tracing/ops_tracing/_api.py#L128-L163) for an example.
 
@@ -139,14 +139,14 @@ Some applications, like database or observability providers, are expected to be 
 It may become expensive to scrape all the data from that relation on every hook.
 
 If you're developing such an application, you should profile the application. Based on the results, you might cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or reconsider whether this key relation processing could be moved out of the reconciliation loop.
-Similar [performance concern](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) applies to peer relation processing for applications where hundreds of units may be expected.
+Similar [performance concerns](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) apply to peer relation processing for applications where hundreds of units may be expected.
 
 ## Handling events individually
 
 A delta-based charm is when the charm handles each kind of Juju hook with a separate handler function, which does the minimum necessary to process that kind of event.
 
 Juju itself nudges charm authors in this direction.
-Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble in the workload container has become ready, and so on.
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that one or more configuration values have changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble in the workload container has become ready, and so on.
 
 ### Example
 
@@ -193,7 +193,7 @@ There's less boilerplate and flow control is predictable.
 
 ### Developing your charm
 
-The charm code and tests can get messy when event interdependencies are accounted for:
+However, the charm code and tests can get messy when event interdependencies are accounted for:
 
 - creating a resource is only possible after the workload has been installed
 - the available data typically needs to be mixed with application's configuration
@@ -212,11 +212,12 @@ The reconciler pattern is especially suitable for mature, feature-rich charms th
 Some events are special, and are typically processed outside the reconciler pattern in holistic charms.
 In both styles of charms a dedicated event handler is used for:
 
-- unit lifecycle end events, because the goal of reconciliation is different
-- action event, as it is synchronous and the event object holds arguments and results
-- secret rotation event, where secret must be modified synchronously
-- secret remove event, because the revision that needs removal is not available otherwise
-- Ops lifecycle events, like `on_collect_unit_status`, which accompany the Juju event
+- unit lifecycle end events, `stop`, and `remove`, because the goal of reconciliation is different
+- action events, as these are synchronous, and the [](ops.ActionEvent) object holds arguments and results
+- the `secret-rotate` event, where a secret must be modified synchronously
+- the `secret-remove` event, because the revision that needs removal is not available otherwise
+- the `secret-expired` event, as the expired secret must be retired or deleted
+- Ops lifecycle events, like `on.collect_unit_status`, which accompany the Juju event
 - some custom events, if a specific charm library API is not suited for reconciliation
 
 A good rule of thumb is this: if an event [cannot be deferred](#how-and-when-to-defer-events), it needs a dedicated handler.
@@ -226,16 +227,19 @@ A good rule of thumb is this: if an event [cannot be deferred](#how-and-when-to-
 A holistic charm attaches the same observer to a group of events.
 In a charm following the reconciler pattern, the `_reconcile` method is attached to every interesting Juju event:
 
-- unit lifecycle start events
-- config change event
+- the `install` and `start` events
+- the `config-changed` event
 - most or all relation events
 - Pebble events
 - storage events
+- the `secret-changed` event
+- the `upgrade-charm` event
+- the `update-status` event
 
 Consider relations: if a charm declares a relation, it's natural to expect that this unit's behaviour or the remote app behaviour depends on the databag content of the relation.
 Therefore, the reconcile method has to read the relation data to apply the latest data, or overwrite its own databag with the latest computed data.
 Thus, the reconciler loop ought to run for every event on this relation.
-Following the same logic, the reconciler loop ought to run for every configuration, storage, and containers event.
+Following the same logic, the reconciler loop ought to run for every configuration, storage, and container event.
 
 ### When to use the holistic approach
 


dimaqq:
Poking @dwilding for early high-level comments: things like "this doesn't belong" here or "you forgot this other thing".

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -38,19 +38,136 @@ class MyCharm(ops.CharmBase):
...
+        self.bar.update_foo_port(foo_port)
+        self.foo.update_bar_name(bar_name)
+
+    else:
+        ...
+```
+
+Expert reconcilers
+
+Reconcilliation happens towards a certain goal state. For many charms, and for most of the charm's lifecycle, the goal state is the same: a running workload with correct configuration. Sometimes, the goal state may change:
```
dwilding: > Reconcilliation happens towards a certain goal state. For many charms, and for most of the charm's lifecycle, the goal state is the same: a running workload with correct configuration.

I like this explanation. I think it gets the point across nicely

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -64,3 +181,5 @@ Similarly, events like `secret-expired` and `secret-rotate` don't make sense to
 This is very closely related to [which events can be deferred](#how-and-when-to-defer-events). A good rule of thumb is this: if an event can be deferred, it may make sense to handle it holistically.
 
 On the other hand, if an event cannot be deferred, the charm cannot handle it holistically. This applies to action "events", `stop`, `remove`, `secret-expired`, `secret-rotate`, and Ops-emitted events such as `collect-status`.
+
+FIXME: since the doc is part of explanation, should I explain briefly why each class of events doesn't belong to the reconciler?
```
dwilding: Personally, "because the charm must do something specific in response to the event" was enough justification for me

dwilding (commented):
I read through and had a think about the overall arrangement of content within the doc. Here's how I might approach it - if this helps!

---

For the intro:

- What do we mean by "delta" charm? Handle events individually
- High-level summary of when holistic event handling is appropriate
- Introduce reconciler charms: do everything holistically, as far as reasonably possible

**Handling events individually**

- Discussion of this is what Juju nudges you towards
- An example of a delta charm

**Handling events holistically**

- Discussion of how individual handling can be limited (config-changed doesn't tell the charm which config keys changed, etc)
- An example of holistic handling - maybe even the same example as the previous section, but made holistic

**Which events can be handled holistically?**

- List of events that can be handled holistically
- List of events that shouldn't be handled holistically

**Reconciler charms**

- Example of what a reconciler charm typically looks like
- Discussion of different goals of reconciliation
- Considerations for implementing a reconciler charm (e.g., I agree that it will be good to explain about collecting status)

dimaqq:
Thank you, David, the outline you've suggested looks really nice.
I'm borrowing at least the first half of it.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
 
+Stateless charms are more robust. The charming community developed two approaches to keep charms mostly stateless: "holistic" and "delta".
 
-Charm developers have had many discussion about "holistic" charms compared to "delta" charms, and which approach is better. First, let's define those terms:
+- Holistic charms reconcile towards a goal state on every event.
+- Delta charms handle each Juju event individually.
 
-* A *delta-based* charm is when the charm handles each kind of Juju hook with a separate handler function, which does the minimum necessary to process that kind of event.
-* A *holistic* charm handles some or all Juju hooks using a common code path such as `_update_charm`, which queries the charm config and relation data and "rewrites the world", that is, rewrites application configuration and restarts necessary services.
+Note that processing of some Juju events doesn't lend to this distinction.
```
benhoyt: What about a single example here:

```suggestion
Note that some Juju events, like `secret-rotate`, can only be processed in a "delta" way.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
 
-Juju itself nudges charm authors in the direction of delta-based charms, because it provides specific event kinds that signal that one "thing" changed: `config-changed` says that a config value changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble container is ready, and so on.
+## Handling events holistically
 
-However, this only goes so far: `config-changed` doesn't tell the charm which config keys changed, and `relation-changed` doesn't tell the charm how the relation data changed.
+A typical holistic charm subscribes the same observer method, often called `_reconcile` to all interesting Juju events.
```
benhoyt:
```suggestion
A typical holistic charm subscribes the same observer method, often called `_reconcile`, to all interesting Juju events.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+The event payload is ignored, and reconciliation progresses towards a defined goal state: typically a functional workload configured appropriately and able to interact with the related applications.
+This implies that the observer has to scrape the entirety of the Juju state available to the current unit, as well as workload state.
```
benhoyt: This paragraph is quite long. Maybe:

```suggestion

This implies that the observer has to scrape the entirety of the Juju state available to the current unit, as well as workload state.
```

benhoyt: "scrape" sounds hacky. Also, "the entirety of the Juju state available" is not quite right. What about "This implies the charm has to fetch all the state it needs: the state of the current unit, and any workload state."

docs/explanation/holistic-vs-delta-charms.md
benhoyt: Or maybe we drop this whole sentence here, as it's clarified further down.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+The observer then rewrite the world: updates all the Juju state it can, overwrites the workload configuration and instructs the workload to reload the configuration.
```
benhoyt:
```suggestion
The observer then "rewrites the world": it updates all the Juju state it needs to, overwrites the workload configuration, and instructs the workload to reload its configuration.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
 
-In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready (`pebble-ready` has not yet been triggered). In such cases, charms could try to wait for both events to occur, possibly storing state to track which events have occurred -- but that is error-prone.
+### Why reconcile?
 
-Alternatively, a charm can use a holistic approach and handle both `config-changed` and `pebble-ready` with a single code path, as in this example:
+Conceptually, a Juju event informs the unit that a specific thing has changed at a pre-defined level in the Juju event model, while the charm author wants the charm to process unit's logical state at workload's own granularity.
```
benhoyt:
```suggestion
Conceptually, a Juju event informs the unit that a specific thing has changed at a predefined level in the Juju event model, while the charm author wants the charm to process a unit's logical state at the workload's own granularity.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
-            # pebble-ready happened first, wait for config-changed
-            return
+* - Juju event
+  - Unit logical state
+* - `config-changed`
+  - Configuration is valid and complete
+* - `pebble-ready`
+  - Workload container is reachable
+* - `relation-joined`
+  - N+1 consumers expect resources
```
benhoyt: Can we restate this? Readers might be confused as to what N+1 means.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+* - `relation-changed`
+  - Consumer requirements are defined
+```
+
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed.
+
+However, this only goes so far: `config-changed` doesn't tell the charm which configuration keys have changed or whether configuration has become valid; `relation-changed` doesn't tell the charm how the relation data have changed or whether the data can be parsed.
```
benhoyt:
```suggestion
However, this only goes so far: `config-changed` doesn't tell the charm which configuration keys have changed or whether configuration has become valid; `relation-changed` doesn't tell the charm how the relation data has changed or whether the data can be parsed.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+
+In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready. In such cases, a delta charm may defer the event (effectively storing a small amount of state) or could try to wait for both events to occur using intricate, error-prone custom logic.
+
+The holistic approach side-steps this disparity and applies a single code path, `_reconcile` to all events of interest, which scrapes all of the visible state to perform computation, and updates all writable state.
```
benhoyt:
```suggestion
The holistic approach side-steps this disparity and applies a single code path, `_reconcile`, to all events of interest, which scrapes all of the visible state to perform computation, and updates all writable state.
```

Or just drop the `_reconcile` mention as we've mentioned that above already.

Also, the second part, "which scrapes" is just restating what we've already said above. I think we can drop from "which scrapes" on.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+    self.workload = Workload()
+    self.foo = FooRequirer()
+    self.bar = BarProvider()
+
+    events = [self.on.start, self.on.config_changed, self.on.foo_relation_changed, ...]
+
+    for event in events:
+        framework.observe(event, self._reconcile)
+
+def _reconcile(self, event: Any):
```
benhoyt: Would this be better, to show `event` is unused (and always an `EventBase` subclass)?
```suggestion
def _reconcile(self, _: ops.EventBase):
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+    # Early checks
+    workload_ready = self.workload.is_ready
```
benhoyt, line 107: Previously we've advocated EAFP (easier to ask forgiveness than permission) instead of workload is-ready checks. This is because a workload could become unready between the time you say `is_ready` and when you actually hit Pebble. Should we follow that here? What do you think, @tonyandrewmeyer?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
-        # If both the Pebble container and config are ready, rewrite the
-        # container's config file and restart Redis if needed.
-        container = self.unit.get_container('redis')
-        try:
-            self._update_redis_config(container, redis_port)
-        except ops.pebble.ConnectionError:
-            # config-changed happened first, wait for pebble-ready
-            return
+    else:
+        # Error handling
```
benhoyt, line 121: What are you envisioning would go here? This isn't really error handling, right, but just "not ready yet" handling?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+def _reconcile(self, event: Any):
+    # Early checks
+    workload_ready = self.workload.is_ready
+    foo_ready = self.foo.is_ready
+    bar_ready = self.bar.is_ready
+    
+    if workload_ready and foo_ready and bar_ready:
+        # Read the inputs: configuration, libraries and the workload
+        path = self.typed_config.some_path
+        foo_port = self.foo.service_port
```
benhoyt: It'd probably be nice if this showed reading config and relation data concretely, rather than `typed_config` (unclear where that comes from) and `self.foo` / `self.bar`.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+        current_config = self.workload.config
+        workload_path = self.workload.special_path
+
+        # Compute the new state
+        workload_config = self.render_config(path, foo_port, bar_name)
+
+        # Write the outputs to the libraries and the workload
+        if workload_config != current_config:
+            self.workload.update_config_and_restart(workload_config)
+        self.bar.update_foo_port(foo_port)
```
benhoyt: What's the intention here -- what would these update foo/bar calls do?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+It consists of three main parts in addition to early checks and error handling:
+
+- reading the inputs (configuration, workload and Juju state)
+- computing the new state
+- writing the output (updating the workload and writing out the Juju state)
+
+Well-written charms include helper methods or classes for the workload and use charm libraries to read and write Juju state on charm's relations.
+
+You may notice that the role of a complex charm is to cross-connect configuration, workload and libraries.
+In fact, complex charms often use a dozen charm libraries, moving the emphasis towards shovelling data from a set of charm libraries to other charm libraries.
```
benhoyt:
```suggestion
In fact, complex charms often use many charm libraries, moving the emphasis towards shovelling data from a set of charm libraries to other charm libraries.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+### Expert reconcilers
+
+Reconciliation happens towards a certain goal state.
+For many charms, and for most of the charm's lifecycle, the goal state is the same: a running workload with correct configuration.
+Rarely, the goal state may change:
+
+- unit lifecycle end, that is, graceful shutdown
+- failover to another unit, leaving this unit on stand-by
+
+Complex charms often split the reconciler method in functional constituents, see [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
```
benhoyt:
```suggestion
Complex charms often split the reconciler method in functional constituents. See [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+Some applications, like database or observability providers, are expected to be related to an unbounded number of other applications.
+It may become expensive to scrape all the data from that relation on every hook.
+
+Such applications should be profiled, and, based on the results, the charm developer may cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or perhaps reconsider whether this key relation processing could be moved out of the reconciliation loop.
+Similar [performance concern](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) applies to peer relation processing for applications where hundreds of units may be expected.
+
+## Which approach to use?
 
-## When to use the holistic approach
+At a high level, simple workloads are perfectly served by delta charms.
```
benhoyt:
```suggestion
At a high level, simple workloads are served well by delta charms.
```

benhoyt: Is this true, though? Isn't it more a case of these types of workloads (typical server or application) are handled better via reconciler, whereas charms with lots of complex interactions (like the database charms) may benefit from delta handling. Not sure on this.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+
+```py
+def __init__(self, framework: ops.Framework):
+    super().__init__(framework)
+    self.typed_config = self.load_config(ConfigClass, errors='blocked')
+    self.workload = Workload()
+    self.foo = FooRequirer()
+    self.bar = BarProvider()
+
+    events = [self.on.start, self.on.config_changed, self.on.foo_relation_changed, ...]
```
benhoyt: Might be better to list these out one per line. They start to wrap on my screen at least.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+In a charm following the reconciler pattern, the `_reconcile` method is attached to every interesting Juju event:
+
+- unit lifecycle start events
+- config change event
+- most or all relation events
+- pebble events
+- storage events
+
+Pragmatically, most Juju events are opportunities for reconciliation.
+Consider relations: if a charm declares a relation, then it's natural to expect that this unit's behaviour and/or remote app behaviour depend on the databag content in this relation.
```
benhoyt:
```suggestion
Consider relations: if a charm declares a relation, it's natural to expect that this unit's behaviour or remote app behaviour depend on the databag content of the relation.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+Therefore, the reconcile method most likely has to read the databag content to apply the latest data, and/or overwrite the databag with the latest computed data.
```
benhoyt:
```suggestion
Therefore, the reconcile method has to read the relation data to apply the latest data, or overwrite the databag with the latest computed data.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+Thus, any event on this relation is subject to running the reconciler loop.
+Following the same logic, all configuration fields, storage objects, and containers are subject to reconciliation.
+
+### When to use the holistic approach
+
+Charms that implement the reconciler pattern have been proven more robust.
```
benhoyt: What about making this claim more specific:
```suggestion
Several charming teams at Canonical have found that using the reconciler pattern makes a charm more robust.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+This comes down to two observations:
+
+- The charm author is thinking about their workload more than about minute Juju semantics.
+- When the Juju event is taken out of the equation, same number of unit tests covers a larger portion of (state, event) space.
```
benhoyt:
```suggestion
- When the Juju event is taken out of the equation, the same number of unit tests cover a larger portion of state and event space.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+
+At the same time, simple operators can be trivially written as delta charms.
+Either of the following hints at suitability of such approach:
+
+- The Juju event model can be mapped directly to the workload semantics.
+- The charm is trivial.
+
+See [SSSD Operator](https://github.com/canonical/sssd-operator/blob/9118ecec6e45820f79dda97f6f7dd287a20b39ac/src/charm.py#L44-L69) and [Apache Kafka Rack Awareness Operator](https://github.com/canonical/kafka-broker-rack-awareness-operator/blob/980781a49b6d65e6d4356819c1f3a2c57a0e3625/src/charm.py#L27-L30) for examples of workloads where a delta charm is suitable.
+
+## Handling events individually
```
benhoyt, line 144: To me it seems a little weird that we have this structure:

- holistic
- which approach to use
- delta

Would it be better as:

- holistic
- delta
- which approach to use

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+There's less boilerplate and flow control is predictable.
 
-Only some events make sense to handle holistically. For example, `remove` is triggered when a unit is about to be terminated, so it doesn't make sense to handle it holistically.
+### Developing your charm
 
-Similarly, events like `secret-expired` and `secret-rotate` don't make sense to handle holistically, because the charm must do something specific in response to the event. For example, Juju will keep triggering `secret-expired` until the charm creates a new secret revision by calling [`event.secret.set_content()`](ops.Secret.set_content).
+Unit testing is initially straightforward, however can get messy when interdependencies are accounted for:
 
-This is very closely related to [which events can be deferred](#how-and-when-to-defer-events). A good rule of thumb is this: if an event can be deferred, it may make sense to handle it holistically.
+- creating a logical bar is only possible after the workload has been installed
```
benhoyt: "logical bar" reads oddly. Should `bar` be backticked? Or maybe we should use a more concrete example than "bar"?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
+- the available data typically needs to be mixed with application's configuration
+- sequences like `upgrade-charm -> config-changed -> start` due to [Juju being helpful](https://discourse.charmhub.io/t/charm-lifecycle/5938)
```
benhoyt: I'm not sure what this last bullet point means, or if the link is helpful. Can you clarify?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,226 @@
...
 
-On the other hand, if an event cannot be deferred, the charm cannot handle it holistically. This applies to action "events", `stop`, `remove`, `secret-expired`, `secret-rotate`, and Ops-emitted events such as `collect-status`.
+A good rule of thumb is this: if you're tempted to defer an event, consider if it's time to rewrite the charm using the reconciler pattern.
```
benhoyt: Perhaps slightly weakening this:
```suggestion
A good rule of thumb is this: if you're starting to use `defer` in various places, consider whether it's time to rewrite the charm using the reconciler pattern.
```

docs/conf.py
```diff
@@ -449,15 +449,14 @@
...
-    ('py:class', '_Readable'),
     ('py:class', '_RelationMetaDict'),
     ('py:class', '_ResourceMetaDict'),
     ('py:class', '_StateKwargs'),
     ('py:class', '_StorageMetaDict'),
     ('py:class', '_TaskDict'),
     ('py:class', '_TextOrBinaryIO'),
     ('py:class', '_WarningDict'),
-    ('py:class', '_Writeable'),
+    #('py:class', '_Writeable'),
```
benhoyt: Can we just remove this if it's not needed anymore?

dimaqq: Sorry I've forgotten to remove it after making sure it was no longer needed.

dimaqq: I'v moved this fly-by into a separate https://github.com/canonical/operator/pull/2044

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I'll drop it; it's in the TOC as well as below anyway.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I actually wanted to make a point that the reconcile loop should fetch everything the given unit is allowed to see:

- if a charm declares a relation, the reconcile loop should read all relations on that endpoint...
   - arguably, only remote app or remote unit data, probably not both, probably not own unit data
   - arguably only the fields that charm understands
- if there's a peer relation, it should read that
   - arguably either app (leader) databag or all peer unit databags
- if there are configuration options, each config field value

If something is not read in the reconcile loop, when would it be read?

That being said, this part is kinda intro, so maybe I can be concise here.

Let me think about this.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: That's true.

I'm not sure what to put there, because the reconciler community seems to be split between the two camps:
- those who set unit status in the reconciler (with ad hoc priorities)
- and those who observe `collect_unit_status` (using ops, but also repeating error checking that `_reconcile` performs)

So it's not like I can recommend one or the other.

I could also write it this way:

```py
def _reconcile(self, _: ops.EventBase):
    # Early checks
    workload_ready = self.workload.is_ready
    foo_ready = self.foo.is_ready
    bar_ready = self.bar.is_ready

    if not workload_ready or not foo_ready or not bar_ready:
        self.unit.status = ...
        return

    try:
        # Read the inputs: configuration, libraries and the workload
        path = self.typed_config.some_path
        foo_port = self.foo.service_port
        bar_name = self.bar.remote_name
        current_config = self.workload.config
        workload_path = self.workload.special_path

        # Compute the new state
        workload_config = self.render_config(path, foo_port, bar_name)

        # Write the outputs to the libraries and the workload
        if workload_config != current_config:
            self.workload.update_config_and_restart(workload_config)
        self.bar.update_foo_port(foo_port)
        self.foo.update_bar_name(bar_name)

    except (WorkloadError, FooError, BarError, ops.ModelError, ...):
        # Error handling
        ...
```

wdyt?

docs/explanation/holistic-vs-delta-charms.md
dimaqq: `self.typed_config` is documented here https://documentation.ubuntu.com/ops/latest/howto/manage-configuration/#observe-the-config-changed-event-and-define-the-event-handler

Now that Tony's typed config has landed, I figured it's better to use this idiom.

`self.typed_config`, `self.foo` and `self.bar` are initialised in charm's `__init__` in the same code snippet, 18 lines above.

Am I perhaps missing the concern?

docs/explanation/holistic-vs-delta-charms.md
dimaqq: My intention was to show that charm's job is to cross-connect multiple charm libs.

Here, something is taken from Foo and stuffed into Bar.
The implication is that bar library will update the bar relation databag.

I'd love to hear ideas how to make this clearer.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I personally see two cases where delta processing works:

- very simple charms, like ubuntu man pages
- key relation processing in database charms

I'm not sure if we should recommend the latter though, because it can get messy very quickly (e.g. pebble not ready on k8s).

docs/explanation/holistic-vs-delta-charms.md
dimaqq: Rewritten... though not a link could be better.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: done

docs/explanation/holistic-vs-delta-charms.md
dimaqq: updated with the suggestion

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+  - The set of consumers requires provisioning
+* - `relation-changed`
+  - Consumer requirements are defined
+```
+
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed.
+
+However, this only goes so far: `config-changed` doesn't tell the charm which configuration keys have changed or whether configuration has become valid; `relation-changed` doesn't tell the charm how the relation data has changed or whether the data can be parsed.
+
+In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready. In such cases, a delta charm may defer the event (effectively storing a small amount of state) or could try to wait for both events to occur using intricate, error-prone custom logic.
```
tonyandrewmeyer: "intricate, error-prone logic" feels too far to me, personally. I believe we're maintaining a position that both delta and holistic are reasonable approaches, rather than pushing towards either one.

For the config-change and pebble-ready case, you don't need intricate custom logic or defer, you just return (with an appropriate status) if both haven't occurred, because you know that the other one will happen (when the container is ready or the Juju user sets the required config). This is true even if this isn't a pebble-ready in startup but from some pod/container recreation.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+    self.typed_config = self.load_config(ConfigClass, errors='blocked')
+    self.workload = Workload()
+    self.foo = FooRequirer()
+    self.bar = BarProvider()
+
+    events = [
+        self.on.start,
+        self.on.config_changed,
+        self.on.foo_relation_changed,
+        self.on.bar_relation_changed,
```
tonyandrewmeyer: I think we're trying to move towards the other style for these.

```suggestion
        self.on['foo'].relation_changed,
        self.on['bar'].relation_changed,
```

docs/explanation/holistic-vs-delta-charms.md
tonyandrewmeyer: Yeah, my feelings aren't quite as strong as they once were, but I do still feel that it's better to avoid introducing race conditions, even if they're extremely unlikely.

Here, I guess it depends what the expectation for `.is_ready`. If it's `.can_connect()` then I think it is better to avoid that and just try to update the container, retry a small number of times, and then move into error otherwise. If it's assuming the container is ready and checking the *workload* is ready in some way, then I'd still expect that EAFP is better, since as Ben says the workload itself might become unready too.

I think the following two lines, which check that a relation is "ready" are a bit different. I assume that's meant to be "has required data" as well as "is integrated" and those can't change during the hook, so LBYL is pretty safe there.

docs/explanation/holistic-vs-delta-charms.md
tonyandrewmeyer: I'm +1 (admittedly not unbiased here :laughing:) on using `load_config` and an `__init__` time load into `typed_config` is a pattern we like, I think, rather than loading in multiple methods (particularly if `errors="blocked"` is used, like here).

The `.foo` and `.bar` are definitely an established pattern, and if they have observers (which they don't here since `self` wasn't passed in, but often would) they need to be attributes of the charm to avoid the handler being garbage collected so you're almost forced into this anyway. Maybe the issue is just that it's not clear for the generic names that these are interface library classes?

docs/explanation/holistic-vs-delta-charms.md
tonyandrewmeyer: I think we can recommend using collect-status, without saying that charms *have* to use it. If you don't have custom priority needs then I think our position is that you should use collect-status.

Repeating some of the checks is not ideal, but that's how the feature works. If you really don't want that, then I think we're ok with the approach where the reconcile function collects status objects and then collect-status does something like:

```python
for s in self._statuses:
    event.add_status(s)
````

I think early-exit guards that just return (or add a status and return) is a cleaner reconcile than wrapping the whole thing and exiting at the end.

File: docs/explanation/holistic-vs-delta-charms.md [hunk 31]
```diff
@@ -1,66 +1,230 @@
...
+    hostname = socket.getfqdn()
+    self.foo = FooRequires(self, "foo-relation", address=hostname)
+    self.framework.observe(self.foo.on.data_available, self._on_data_available)
+
+    self.bar = BarProvider(self, "bar-relation")
+    self.framework.observe(self.bar.on.create_bar, self._on_create_bar)
+
+def _on_install(self, event: ops.InstallEvent):
+    # Install the workload binary
+    apt.add_package("some-workload")
```
tonyandrewmeyer: I think we should probably be consistent here with our guidance elsewhere and in the profiles and call out to a workload module here, like `workload.install_packages()`.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+    if self.unit.is_leader():
+        ...
+
+def _on_data_available(self, event: DataAvailableEvent):
+    # Update the workload with event's data
+    self.workload.reconfigure(some_key=event.some_value)
+
+def _on_create_bar(self, event: CreateBarEvent):
+    # Provision a `Bar` resourse in the workload
+    self.workload.create_foo(event.some_field)
```
tonyandrewmeyer: Probably no "self" here, since there's no `self.workload = ` in `__init__`, and we're generally suggesting that the workload module has these sorts of things, without needing a class/object.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
 (holistic-vs-delta-charms)=
 # Holistic vs delta charms
 
+Stateless charms are more robust. The charming community developed two approaches to keep charms mostly stateless: "holistic" and "delta".
```
dwilding: When I read the first sentence, I wonder "More robust than what? Why?" I think that's not a discussion we want to get into here. How about this instead:

```suggestion
The charming community developed two approaches to writing robust, stateless charms: "holistic" and "delta".
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+At a high level, simple workloads are served well by delta charms.
+Complex workloads and especially mature feature-rich charms utilising many charm libraries are more robust if the reconciler pattern is followed.
+
+### Dedicated handlers for specific Juju events
+
+Some events are special, and are typically processed outside the reconciler pattern in holistic charms.
+In both styles of charms a dedicated event handler is used for:
+
+- unit lifecycle end events, because the goal of reconciliation is different
+- action event, as it is synchronous and the event object holds arguments and result sink
```
tonyandrewmeyer: "result sink" feels odd to me.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+- secret rotation event, where secret must be modified synchronously
+- secret remove event, because the revision that needs removal is not available otherwise
+- Ops lifecycle events, like `update-unit-status`, which accompany the Juju event
```
tonyandrewmeyer: Probably meant to be `collect-unit-status`?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
 
-In other words, when writing a charm, it's not so much "should the *charm* be holistic?" as "does it make sense for *these events* to be handled holistically?"
+A holistic charm attaches the same observer to a group of events.
+In a charm following the reconciler pattern, the `_reconcile` method is attached to every interesting Juju event:
 
-Using the holistic approach is normally centred around configuring an application. Various events that affect configuration use a common handler, to simplify writing an application config file and restarting the application.  This is common for events like `config-changed`, `relation-changed`, `secret-changed`, and `pebble-ready`.
+- unit lifecycle start events
+- config change event
+- most or all relation events
+- pebble events
```
tonyandrewmeyer: Nit:

```suggestion
- Pebble events
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
-* A *holistic* charm handles some or all Juju hooks using a common code path such as `_update_charm`, which queries the charm config and relation data and "rewrites the world", that is, rewrites application configuration and restarts necessary services.
+## Handling events holistically
 
-Juju itself nudges charm authors in the direction of delta-based charms, because it provides specific event kinds that signal that one "thing" changed: `config-changed` says that a config value changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble container is ready, and so on.
+A typical holistic charm subscribes the same observer method, often called `_reconcile`, to all interesting Juju events.
+The event payload is ignored, and reconciliation progresses towards a defined goal state: typically a functional workload configured appropriately and able to interact with the related applications.
+This implies that a holistic charm has to fetch all the state it needs: the state of the current unit, and any workload state.
```
dwilding, line 14: Very minor - I'd probably break here for readability.

```suggestion

This implies that a holistic charm has to fetch all the state it needs: the state of the current unit, and any workload state.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+The observer then "rewrites the world": updates all the Juju state it needs to, overwrites the workload configuration, and instructs the workload to reload its configuration.
 
-However, this only goes so far: `config-changed` doesn't tell the charm which config keys changed, and `relation-changed` doesn't tell the charm how the relation data changed.
+### Why reconcile?
 
-In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready (`pebble-ready` has not yet been triggered). In such cases, charms could try to wait for both events to occur, possibly storing state to track which events have occurred -- but that is error-prone.
+Conceptually, a Juju event informs the unit that a specific thing has changed at a predefined level in the Juju event model, while the charm author wants the charm to process a unit's logical state at the workload's own granularity.
```
dwilding: A couple of minor suggestions (removing "at a predefined level" because I don't think it adds significantly to the explanation):

```suggestion
Conceptually, a Juju event informs the unit that a specific thing has changed in the Juju event model, while the charm author wants the charm to process a unit's logical state at the workload's own granularity. For example:
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+  - Workload container is reachable
+* - `relation-joined`
+  - The set of consumers requires provisioning
+* - `relation-changed`
+  - Consumer requirements are defined
+```
+
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed.
+
+However, this only goes so far: `config-changed` doesn't tell the charm which configuration keys have changed or whether configuration has become valid; `relation-changed` doesn't tell the charm how the relation data has changed or whether the data can be parsed.
```
dwilding: Very minor - I'd go with this style:

```suggestion
However, this only goes so far. `config-changed` doesn't tell the charm which configuration keys have changed or whether configuration has become valid. `relation-changed` doesn't tell the charm how the relation data has changed or whether the data can be parsed.
```

docs/explanation/holistic-vs-delta-charms.md
dwilding: I was reviewing while Tony's comment came in, and was going to say that "try to wait for both events to occur using intricate, error-prone custom logic" seems like a more involved way to say "defer the event". So I was going to suggest this:

```suggestion
In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready. In such cases, a delta charm may defer the event, effectively storing a small amount of state.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready. In such cases, a delta charm may defer the event (effectively storing a small amount of state) or could try to wait for both events to occur using intricate, error-prone custom logic.
+
+The holistic approach side-steps this disparity and applies a single code path, `_reconcile`, to all events of interest, which scrapes all state it needs to perform computation, and updates all writable state.
```
dwilding: I think there was an earlier comment about "scrapes" but I can't find it now. I might use a different word and split the long sentence:

```suggestion
The holistic approach side-steps this disparity and applies a single code path, `_reconcile`, to all events of interest. This code path reads all state it needs to perform computation, and updates all writable state.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+```
+
+The reconciler method above has been reduced for clarity, but is representative of the common reconciler pattern.
+It consists of three main parts in addition to early checks and error handling:
+
+- reading the inputs (configuration, workload and Juju state)
+- computing the new state
+- writing the output (updating the workload and writing out the Juju state)
+
+Well-written charms include helper methods or classes for the workload and use charm libraries to read and write Juju state on charm's relations.
```
dwilding: Does "Juju state" here mean something like "the state of other applications in the same Juju model"? If not, I don't fully understand this part. Would this be correct?

```suggestion
Well-written charms include helper methods or classes for the workload and use charm libraries to read and write relation data.
```

docs/explanation/holistic-vs-delta-charms.md
dwilding: Also "into functional constituents" rather than "in functional constituents"?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+Reconciliation happens towards a certain goal state.
+For many charms, and for most of the charm's lifecycle, the goal state is the same: a running workload with correct configuration.
+Rarely, the goal state may change:
+
+- unit lifecycle end, that is, graceful shutdown
+- failover to another unit, leaving this unit on stand-by
+
+Complex charms often split the reconciler method in functional constituents, see [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
+
+Charm libraries too can be written using the reconciler pattern, see [the implementation of `ops.tracing.Tracing`](https://github.com/canonical/operator/blob/af6764fbfdcf8ed42a0edb331870dd9d37dc804b/tracing/ops_tracing/_api.py#L128-L163) for an example.
```
dwilding:
```suggestion
Charm libraries too can be written using the reconciler pattern. See [the implementation of `ops.tracing.Tracing`](https://github.com/canonical/operator/blob/af6764fbfdcf8ed42a0edb331870dd9d37dc804b/tracing/ops_tracing/_api.py#L128-L163) for an example.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+Some applications, like database or observability providers, are expected to be related to an unbounded number of other applications.
+It may become expensive to scrape all the data from that relation on every hook.
+
+Such applications should be profiled, and, based on the results, the charm developer may cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or perhaps reconsider whether this key relation processing could be moved out of the reconciliation loop.
```
dwilding: I'd split the long sentence and make it more active. Something like:

```suggestion
If you're developing such an application, you should profile the application. Based on the results, you might cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or reconsider whether this key relation processing could be moved out of the reconciliation loop.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+Juju itself nudges charm authors in this direction.
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble in the workload container has become ready, and so on.
 
-    def _update_charm(self, _: ops.EventBase):  # event parameter isn't used
-        redis_port = self.config.get('redis-port')
-        if not redis_port:
-            # pebble-ready happened first, wait for config-changed
-            return
+A delta charm may look like this:
```
dwilding: I think we can go straight into the example

```suggestion
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+Some applications, like database or observability providers, are expected to be related to an unbounded number of other applications.
+It may become expensive to scrape all the data from that relation on every hook.
+
+Such applications should be profiled, and, based on the results, the charm developer may cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or perhaps reconsider whether this key relation processing could be moved out of the reconciliation loop.
+Similar [performance concern](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) applies to peer relation processing for applications where hundreds of units may be expected.
+
+## Handling events individually
+
+A *delta-based* charm is when the charm handles each kind of Juju hook with a separate handler function, which does the minimum necessary to process that kind of event.
```
dwilding: I think we can drop the italics, as we seem to have dropped them elsewhere in the doc

```suggestion
A delta-based charm is when the charm handles each kind of Juju hook with a separate handler function, which does the minimum necessary to process that kind of event.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+    # Provision a `Bar` resourse in the workload
+    self.workload.create_foo(event.some_field)
 ```
 
+Notice how the handler methods map to Juju and custom events directly.
+There's less boilerplate and flow control is predictable.
+
+### Developing your charm
+
+Unit testing is initially straightforward, however both charm and tests get messy when event interdependencies are accounted for:
```
dwilding: I found this slightly difficult to parse. How about this? I don't think it adds much to mention unit testing first.

```suggestion
The charm code and tests can get messy when event interdependencies are accounted for:
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+- creating a resource is only possible after the workload has been installed
+- the available data typically needs to be mixed with application's configuration
+- cases where something is inferred from sequence of events [Juju being helpful](https://discourse.charmhub.io/t/charm-lifecycle/5938)
```
dwilding: I don't think we should like to the old Discourse docs, if possible. How about:

```suggestion
- cases where something is inferred from the sequence of events (see {external+juju:ref}`Juju | Hook ordering <hook>`)
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+A good rule of thumb is this: if you're starting to use `defer` in various places, consider whether it's time to rewrite the charm using the reconciler pattern.
+
+## Which approach to use?
+
+At a high level, simple workloads are served well by delta charms.
+Complex workloads and especially mature feature-rich charms utilising many charm libraries are more robust if the reconciler pattern is followed.
```
dwilding: I found this sentence a bit tough. How about:

```suggestion
Complex workloads are more robust if the reconciler pattern is followed. The reconciler pattern is especially suitable for mature, feature-rich charms that use several charm libraries.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+
+Some events are special, and are typically processed outside the reconciler pattern in holistic charms.
+In both styles of charms a dedicated event handler is used for:
+
+- unit lifecycle end events, because the goal of reconciliation is different
+- action event, as it is synchronous and the event object holds arguments and result sink
+- secret rotation event, where secret must be modified synchronously
+- secret remove event, because the revision that needs removal is not available otherwise
+- Ops lifecycle events, like `update-unit-status`, which accompany the Juju event
+- some custom events, if specific charm library API is not suited for reconciliation
```
dwilding:
```suggestion
- some custom events, if a specific charm library API is not suited for reconciliation
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+- unit lifecycle start events
+- config change event
+- most or all relation events
+- pebble events
+- storage events
 
-Many existing charms use holistic event handling. A few examples are:
+Pragmatically, most Juju events are opportunities for reconciliation.
+Consider relations: if a charm declares a relation, it's natural to expect that this unit's behaviour or the remote app behaviour depends on the databag content of the relation.
+Therefore, the reconcile method has to read the relation data to apply the latest data, or overwrite own databag with the latest computed data.
```
dwilding:
```suggestion
Therefore, the reconcile method has to read the relation data to apply the latest data, or overwrite its own databag with the latest computed data.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
 
-Using the holistic approach is normally centred around configuring an application. Various events that affect configuration use a common handler, to simplify writing an application config file and restarting the application.  This is common for events like `config-changed`, `relation-changed`, `secret-changed`, and `pebble-ready`.
+- unit lifecycle start events
+- config change event
+- most or all relation events
+- pebble events
+- storage events
 
-Many existing charms use holistic event handling. A few examples are:
+Pragmatically, most Juju events are opportunities for reconciliation.
```
dwilding: This paragraph seems to be restating the above bullets, but with more explanation. I'd drop the first sentence and get right into things.

```suggestion
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+Consider relations: if a charm declares a relation, it's natural to expect that this unit's behaviour or the remote app behaviour depends on the databag content of the relation.
+Therefore, the reconcile method has to read the relation data to apply the latest data, or overwrite own databag with the latest computed data.
+Thus, any event on this relation is subject to running the reconciler loop.
+Following the same logic, all configuration fields, storage objects, and containers are subject to reconciliation.
```
dwilding: I'm not sure how to interpret "subject to reconciliation". How about:

```suggestion
Thus, the reconciler loop ought to run for every event on this relation.
Following the same logic, the reconciler loop ought to run for every configuration, storage, and container event.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
-Only some events make sense to handle holistically. For example, `remove` is triggered when a unit is about to be terminated, so it doesn't make sense to handle it holistically.
+### When to use the delta approach
 
-Similarly, events like `secret-expired` and `secret-rotate` don't make sense to handle holistically, because the charm must do something specific in response to the event. For example, Juju will keep triggering `secret-expired` until the charm creates a new secret revision by calling [`event.secret.set_content()`](ops.Secret.set_content).
+At the same time, simple operators can be trivially written as delta charms.
+Either of the following hints at suitability of such approach:
 
-This is very closely related to [which events can be deferred](#how-and-when-to-defer-events). A good rule of thumb is this: if an event can be deferred, it may make sense to handle it holistically.
+- The Juju event model can be mapped directly to the workload semantics.
+- The charm is trivial.
```
dwilding: I think this can be condensed. How about:

```
A delta approach is most appropriate when the Juju event model can be mapped directly to the workload semantics. Simple operators can typically be written as delta charms.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,230 @@
...
+    # This unit has been started, peer relations are usable
+    if self.unit.is_leader():
+        ...
+
+def _on_data_available(self, event: DataAvailableEvent):
+    # Update the workload with event's data
+    self.workload.reconfigure(some_key=event.some_value)
+
+def _on_create_bar(self, event: CreateBarEvent):
+    # Provision a `Bar` resourse in the workload
```
dwilding:
```suggestion
    # Provision a `Bar` resource in the workload
```

(This typo is what's causing the linting workflow to fail)

dwilding (approved):
Some suggestions, mostly minor. Nice work!

docs/explanation/holistic-vs-delta-charms.md
benhoyt: Okay, I'm sold on `self.typed_config` then. But yeah, it's not clear from the generic names `foo` and `bar` what they are intended to represent.

docs/explanation/holistic-vs-delta-charms.md
benhoyt: Okay, fair enough. This seems like a reasonable place to start.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: maybe `foo_value = self.foo.some_relation_property`

docs/explanation/holistic-vs-delta-charms.md
dimaqq: Discussion notes:
- vm charms are more delta-friendly, because the workload state is durable
- maintaining existing delta charm: continue with delta probably
- creating a wkld resource when relation is joined, removing when it's departed?
- maybe when enumerating all relations or all workload things is hopeless slow
   - maybe an S3 adaptor: only uses one relation and on cloud thing... a proxy for something external???
   - maybe a integration test double (local minio instead of real s3: one relation only, predefined config, no secrets) ✨

Probably not:
- adaptor charms (workload-less, e.g. haproxy to ingress, relation limit), still a mini-reconciler
- action-only charms?
- toy charms, maybe??
- not simple charms though, because if it's a charm, it ought to have at least two inputs (e.g. config and relation)

docs/explanation/holistic-vs-delta-charms.md
dimaqq: David's suggesting is good considering the next paragraph

docs/explanation/holistic-vs-delta-charms.md
dimaqq: "holds arguments and results"

docs/explanation/holistic-vs-delta-charms.md
dimaqq: like `on_collect_unit_status`

docs/explanation/holistic-vs-delta-charms.md
dimaqq: No, what I'm trying to refer to is "the subset of Juju model state that's visible to the given unit" which is roughly the same as "all the Juju state that unit can scrape using the hook tools".

I think I've rewritten this bit twice already, and it's still unclear.
What would you call it?

dwilding: Ah, thanks. I now realise that I'd also misinterpreted the bullets that come before this part. How about saying "Juju environment state" (dropping Juju after the first use, for brevity). Something like this:

```
- reading the inputs (configuration, workload, and Juju environment state)
- computing the new state
- writing the output (updating the workload and writing environment state)

Well-written charms include helper methods or classes for the workload and use charm libraries to read and write environment state on relations.
```

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I've updates this to:

```
    self.foo = FooRequirer()

        self.on['foo'].relation_changed,

    foo_ready = self.foo.is_ready

        foo_value = self.foo.some_relation_property

        self.foo.update_some_relation_field(bar_value)
```

I waas trying to abstract away from any specific provider and requirer, but this is about a charm library that wraps a relation.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: rewritten

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I'll add `self.workload = ...` for parity with the holistic example.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I've reworked this, ptal.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: done and reworked on top of that.
I'm not going to cover status collection in this doc.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: Renamed to `foo_requirer` and `bar_provider`.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: done

docs/explanation/holistic-vs-delta-charms.md
dimaqq: done.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I've expanded this section.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
 (holistic-vs-delta-charms)=
 # Holistic vs delta charms
 
+The charming community developed two approaches to writing robust, stateless charms: "holistic" and "delta".
```
benhoyt: Reads slightly weird now, as if it's all in the past. Perhaps:
```suggestion
The charming community has developed two approaches to writing robust, stateless charms: "holistic" and "delta".
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+
+In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready.
+In such cases, a delta charm may defer the event, effectively storing a small amount of state.
+
+The holistic approach side-steps this disparity and applies a single code path, `_reconcile`, to all events of interest.
+The code path reads all state it needs to perform computation, and updates all writable state.
+
+### Example
+
+The reconciler method above has been reduced for clarity, but is representative of the common reconciler pattern.
```
benhoyt: What does "reconciler method above" refer to here? Should it be "below"?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+
+The body of the reconciler loop comprises three main parts:
```
benhoyt: "loop" might be confusing, and we've called it a "method" above, so:
```suggestion
The body of the reconciler method comprises three main parts:
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+### Expert reconcilers
+
+Reconciliation happens towards a certain goal state.
+For many charms, and for most of the charm's lifecycle, the goal state is the same: a running workload with correct configuration.
+Rarely, the goal state may change:
+
+- unit lifecycle end, that is, graceful shutdown
+- failover to another unit, leaving this unit on stand-by
+
+Complex charms often split the reconciler method into functional constituents. See [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
```
benhoyt:
```suggestion
Complex charms often split the reconciler method into functional constituents. See the [Parca Kubernetes Operator](https://github.com/canonical/parca-k8s-operator/blob/6fda0c9844bc7a45b93ac12806a6ef04036c50c4/src/charm.py#L188-L206) for an example.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
-        try:
-            self._update_redis_config(container, redis_port)
-        except ops.pebble.ConnectionError:
-            # config-changed happened first, wait for pebble-ready
-            return
+Some applications, like database or observability providers, are expected to be related to an unbounded number of other applications.
+It may become expensive to scrape all the data from that relation on every hook.
+
+If you're developing such an application, you should profile the application. Based on the results, you might cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or reconsider whether this key relation processing could be moved out of the reconciliation loop.
+Similar [performance concern](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) applies to peer relation processing for applications where hundreds of units may be expected.
```
benhoyt:
```suggestion
Similar [performance concerns](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) apply to peer relation processing for applications where hundreds of units may be expected.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+    # Provision a `Bar` resource in the workload
+    self.workload.create_bar(event.some_field)
 ```
 
+Notice how the handler methods map to Juju and custom events directly.
+There's less boilerplate and flow control is predictable.
+
+### Developing your charm
+
+The charm code and tests can get messy when event interdependencies are accounted for:
```
benhoyt: This seems to need a transition word, like:
```suggestion
However, the charm code and tests can get messy when event interdependencies are accounted for:
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+Complex workloads are more robust if the reconciler pattern is followed.
+The reconciler pattern is especially suitable for mature, feature-rich charms that use several charm libraries.
+
+### Dedicated handlers for specific Juju events
+
+Some events are special, and are typically processed outside the reconciler pattern in holistic charms.
+In both styles of charms a dedicated event handler is used for:
 
-## When to use the holistic approach
+- unit lifecycle end events, because the goal of reconciliation is different
```
benhoyt: What does "unit lifecycle and events" refer to -- can you rephrase or clarify?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+- action event, as it is synchronous and the event object holds arguments and results
```
benhoyt:
```suggestion
- action events, as these are synchronous, and the event object holds arguments and results
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+- secret rotation event, where secret must be modified synchronously
```
benhoyt:
```suggestion
- the `secret-rotate` event, where a secret must be modified synchronously
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+- secret remove event, because the revision that needs removal is not available otherwise
```
benhoyt:
```suggestion
- the `secret-remove` event, because the revision that needs removal is not available otherwise
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
 
-In other words, when writing a charm, it's not so much "should the *charm* be holistic?" as "does it make sense for *these events* to be handled holistically?"
+### Events observed by the reconcile method
 
-Using the holistic approach is normally centred around configuring an application. Various events that affect configuration use a common handler, to simplify writing an application config file and restarting the application.  This is common for events like `config-changed`, `relation-changed`, `secret-changed`, and `pebble-ready`.
+A holistic charm attaches the same observer to a group of events.
+In a charm following the reconciler pattern, the `_reconcile` method is attached to every interesting Juju event:
 
-Many existing charms use holistic event handling. A few examples are:
+- unit lifecycle start events
```
benhoyt: The word "start" here helps, but similar to above, can you be more specific here?

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+- config change event
```
benhoyt:
```suggestion
- the `config-changed` event
```

benhoyt (changes requested):
Looking good -- thanks for the updates. Requesting some (minor) changes.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+  - Configuration is valid and complete
+* - `pebble-ready`
+  - Workload container is reachable
+* - `relation-joined`
+  - The set of consumers requires provisioning
+* - `relation-changed`
+  - Consumer requirements are defined
+```
+
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed.
```
tonyandrewmeyer:
```suggestion
Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that one or more configuration values have changed, `relation-changed` says that relation data has changed.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+
+However, this only goes so far.
+The `config-changed` event doesn't tell the charm which configuration keys have changed or whether configuration has become valid.
```
tonyandrewmeyer: I think we should leave off the bit about validity. From Juju's point of view, you only get config-changed if the config *is* valid, because it prevents you setting anything invalid (like setting a numeric config to a string). From the charm's point of view, there's more validation that's required than what Juju offers, so often there's additional validation code in the charm, but that's really unrelated to anything holistic/delta.

```suggestion
The `config-changed` event doesn't tell the charm which configuration keys have changed.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+The `relation-changed` event doesn't tell the charm how the relation data has changed or whether the data can be parsed.
```
tonyandrewmeyer: Similarly, here, from Juju's point of view, you get a string and that's it. Charms (typically) add more structure on top of that, but (until there's scriptlets or whatever that might solve this) it's not really related to delta/holistic.

```suggestion
The `relation-changed` event doesn't tell the charm how the relation data has changed.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+
+In addition, the charm may receive an event like `config-changed` before it's ready to handle it, for example, if the container is not yet ready.
+In such cases, a delta charm may defer the event, effectively storing a small amount of state.
```
tonyandrewmeyer: I think it's worth pointing out here how this tends to lead towards a reconcile approach. Something like:

```suggestion
In such cases, a delta charm may defer the event, effectively storing a small amount of state, or have both `config-changed` and `pebble-ready` check if everything is ready and then run common code, which is the first step towards a holistic approach.
```

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+
+If you're developing such an application, you should profile the application. Based on the results, you might cache the Juju state, communicate less on the relation and [more out of band](https://github.com/charmed-hpc/slurm-charms/blob/edc47369560845fa54f81a2f02e0a3870b14302a/charms/slurmrestd/src/charm.py#L93-L98), or reconsider whether this key relation processing could be moved out of the reconciliation loop.
+Similar [performance concern](https://discourse.charmhub.io/t/performance-limitations-of-peer-relations-and-a-solution/18144) applies to peer relation processing for applications where hundreds of units may be expected.
+
+## Handling events individually
+
+A delta-based charm is when the charm handles each kind of Juju hook with a separate handler function, which does the minimum necessary to process that kind of event.
+
+Juju itself nudges charm authors in this direction.
+Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that a configuration value has changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble in the workload container has become ready, and so on.
```
tonyandrewmeyer:
```suggestion
Juju provides specific event kinds that signal that one "thing" changed: `config-changed` says that one or more configuration values have changed, `relation-changed` says that relation data has changed, `pebble-ready` signals that the Pebble in a workload container has become ready, and so on.
```

docs/explanation/holistic-vs-delta-charms.md
tonyandrewmeyer: I think it's `stop` and `remove`, and if that's the case, I think we should just explicitly name them since there are only two.

```suggestion
- unit lifecycle end events, `stop`, and `remove`, because the goal of reconciliation is different
```

docs/explanation/holistic-vs-delta-charms.md
tonyandrewmeyer: What do you think about also having `secret-expired` here? In theory you could have the reconcile function loop through all the owned secrets getting the metadata and seeing if it's expired, but that seems quite inefficient when there's already Juju code doing that for you. I think having a specific `secret-expired` observer and then having the resulting `secret-changed` (assuming that as well as removing the expired revision you create a new one) trigger the reconcile is a cleaner pattern.

I think the three secret events that repeat until an action is taken are all good candidates for specific observers, even though you could (other than remove unless you're also tracking used revisions) implement them in a holistic handler.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
+
+Some events are special, and are typically processed outside the reconciler pattern in holistic charms.
+In both styles of charms a dedicated event handler is used for:
 
-## When to use the holistic approach
+- unit lifecycle end events, because the goal of reconciliation is different
+- action event, as it is synchronous and the event object holds arguments and results
+- secret rotation event, where secret must be modified synchronously
+- secret remove event, because the revision that needs removal is not available otherwise
+- Ops lifecycle events, like `on_collect_unit_status`, which accompany the Juju event
```
tonyandrewmeyer: I think this should either be `on.collect_unit_status` or `collect-status` (or maybe `collect-unit-status`).

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
-Using the holistic approach is normally centred around configuring an application. Various events that affect configuration use a common handler, to simplify writing an application config file and restarting the application.  This is common for events like `config-changed`, `relation-changed`, `secret-changed`, and `pebble-ready`.
+A holistic charm attaches the same observer to a group of events.
+In a charm following the reconciler pattern, the `_reconcile` method is attached to every interesting Juju event:
 
-Many existing charms use holistic event handling. A few examples are:
+- unit lifecycle start events
+- config change event
+- most or all relation events
+- Pebble events
+- storage events
```
tonyandrewmeyer, line 234: I think `upgrade-charm` should be here, although perhaps with a mention that's it's quite complicated to handle for stateful charms.

`secret-changed` should be here.

I wonder if it's worth mentioning `update-status`. It doesn't really fit "interesting Juju event", but it is usually included. I think there's some justification that it's to capture updates to the workload status that have occurred outside of Juju events (although for K8s, Pebble events ought to be doing that). Realistically, it's the "just in case something was missed, hopefully this will trigger on a fairly short interval" case.

docs/explanation/holistic-vs-delta-charms.md
```diff
@@ -1,66 +1,252 @@
...
-- [`hello-kubecon` is a simple charm that handles `config-changed` and `pebble-cready` holistically](https://github.com/jnsgruk/hello-kubecon/blob/dbd133466dde59ee64f20a732a8f3d2e560ec3b8/src/charm.py#L32-L33)
-- [`prometheus-k8s` uses a common `_configure` method to handle various events](https://github.com/canonical/prometheus-k8s-operator/blob/84c6a406ed585cdb7ba40e01a258864987d6f67f/src/charm.py#L221-L230)
-- [`sdcore-gnbsim-k8s` also uses a common `_configure` method](https://github.com/canonical/sdcore-gnbsim-k8s-operator/blob/ea2afe069346757b1eb6c02de5b4f50f90e81698/src/charm.py#L84-L92)
+Consider relations: if a charm declares a relation, it's natural to expect that this unit's behaviour or the remote app behaviour depends on the databag content of the relation.
+Therefore, the reconcile method has to read the relation data to apply the latest data, or overwrite its own databag with the latest computed data.
+Thus, the reconciler loop ought to run for every event on this relation.
+Following the same logic, the reconciler loop ought to run for every configuration, storage, and containers event.
```
tonyandrewmeyer:
```suggestion
Following the same logic, the reconciler loop ought to run for every configuration, storage, and container event.
```

tonyandrewmeyer (commented):
Looking really thorough, and much closer to what charmers are doing than the current doc, thanks!

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I'll try "... logically consistent", because that's kind of the point: advanced charms process inputs at different, application level than raw Juju state.

dimaqq: On the second thought, maybe it's better to cut this, keep the full docs  a bit shorter.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I think I'll add these without explanation...

Explanations would be great, but probably not in this doc.

docs/explanation/holistic-vs-delta-charms.md
dimaqq: I've added the `secret-expire` event to the list.

However, what if the consumer reconciler always pulled all "named" secrets (that is secret uris from config and relations), always got the content with `--refresh` and applied it to the workload.

Wouldn't that solve the `secret-expired` event concern?
Or is the concern about expiration without a new revision being available?

dimaqq: Sorry, I've confused secret-expired and secret-changed yet again.

PR #2031, opened 2025-09-04
docs: update get_cloud_spec doc now that credential-get works on K8s

benhoyt:
Update the docstring for `Model.get_cloud_spec` now that the Juju fix for credential-get to work on K8s is merged (https://github.com/juju/juju/pull/20428).

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -335,8 +335,9 @@ class Model:
     def get_cloud_spec(self) -> CloudSpec:
         """Get details of the cloud in which the model is deployed.
 
-        Note: This information is only available for machine charms,
-        not Kubernetes sidecar charms.
+        Since Juju 3.6.10, this information is available on both machine
+        charms and Kubernetes charms. With earlier Juju versions, it was only
+        available on machine charms.
 
         Returns:
             a specification for the cloud in which the model is deployed,


tonyandrewmeyer (approved):
Thanks!

IronCore864 (approved):

ops/model.py
```diff
@@ -335,8 +335,9 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
     def get_cloud_spec(self) -> CloudSpec:
         """Get details of the cloud in which the model is deployed.
 
-        Note: This information is only available for machine charms,
-        not Kubernetes sidecar charms.
+        Since Juju 3.6.10, this information is available on both machine
+        charms and Kubernetes charms. With earlier Juju versions, it was only
+        available on machine charms.
```
dimaqq: There's a special Sphinx pragma that Tony coded up that could be used here:

```suggestion
        .. jujuchanged:: 3.6.10
            This information is available on both machine charms and Kubernetes
            charms. With earlier Juju versions, it was only available on machine charms.
```

benhoyt: I'll try that.

tonyandrewmeyer: Note that our docs [say](https://github.com/canonical/operator/blob/e33459f28197eeecff407c42cf731d9d8f6d9a7c/CONTRIBUTING.md?plain=1#L103):

> Use `.. jujuchanged:: x.y` when the feature's behaviour _in Ops_ changes.

And that doesn't really fit here - the Juju behaviour has changed but the Ops behaviour still just reflects Juju.

I don't recall why we defined it this way, but if we stick to it I think the original is better.

benhoyt: Hmm, strange. Here are the bullet points:

- Use `.. jujuadded:: x.y` to indicate that the feature is only available when using version x.y (or higher) of Juju.
- Use `.. jujuchanged:: x.y` when the feature's behaviour _in Ops_ changes.
- Use `.. jujuremoved:: x.y` when the feature will be available in Ops but not in that version (or later) of Juju.

What about we change the second and third to this (it seems to me that's what it should be and how it's used):

- Use `.. jujuadded:: x.y` to indicate that the feature is only available when using version x.y (or higher) of Juju.
- Use `.. jujuchanged:: x.y` when the feature's behaviour changed in version x.y of Juju.
- Use `.. jujuremoved:: x.y` when the feature was removed in version x.y of Juju.

benhoyt: @tonyandrewmeyer Ping on the above -- what do you think about this change to that text?

tonyandrewmeyer: I think the `jujuremoved` definition is definitely better. I think the current one should really say "may be" rather than "will be", and if that change is made then you version is a more concise way to say the same thing.

For `jujuchanged` it seems like there must have been a clear reason for it being so defined, but it was added in a drive-by and I can't find any records of the discussions that led to it. We only use it once currently:

```
        .. jujuchanged:: 3.6
            A new secret revision will *not* be created if the content being set
            is identical to the latest revision.
```

This is not at all behaviour "in Ops" changing, it's the behaviour in Juju changing. If the behaviour in Ops was changing, then it seems weird to pair that with a *Juju* release.

The only thing I can think of is that we were trying to avoid duplicating too much of the Juju changelog in the Ops docs. However, just like some `jujuadded` and `jujuremoved` makes sense (to me, at least), I think judicious (jujuicious :wink:?) `jujuchanged` makes sense too.

So I'm happy to changing our internal docs to what you're proposing, and then would be fine with using `jujuchanged` here too.

ops/model.py
```diff
@@ -335,8 +335,9 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
...
         """Get details of the cloud in which the model is deployed.
 
-        Note: This information is only available for machine charms,
-        not Kubernetes sidecar charms.
+        .. jujuchanged:: 3.6.10
+            This information is available on both machine charms and Kubernetes
+            charms. With earlier Juju versions, it was only available on machine charms.
 
         Returns:
             a specification for the cloud in which the model is deployed,
```
tonyandrewmeyer, line 343: Noticed while looking through `ModelError` raising docs: there's a "Raises" note a few lines below this that should also be removed or updated ("if called in a Kubernetes model" should probably be "if called without trust").

benhoyt: Done, thanks.

ops/model.py
benhoyt: Done, thanks.

PR #2033, opened 2025-09-04
fix: don't cache secret values

### LINKED ISSUE #1536: Don't cache secret content in Ops
With this charm:

```python
class DifferentSecretRefreshesCharm(ops.CharmBase):
        super().__init__(framework)
        framework.observe(self.on.run_action, self._on_run)

        try:
            self.model.get_secret(label="my-secret")
        except ops.SecretNotFoundError:
            self.unit.add_secret({"foo": "bar"}, label="my-secret")

    def _on_run(self, event: ops.ActionEvent):
        secret1 = self.model.get_secret(label="my-secret")
        secret1.set_content({"foo": "baz"})
        secret2 = self.model.get_secret(label="my-secret")
        content1 = secret1.get_content(refresh=True)
        content2 = secret2.get_content()
        event.set_results({"content1": content1, "content2": content2})
```

What is the expect output?

Answer 1: the two Secret objects are distinct, and should have the same content unless I call `get_content(refresh=True)` on that specific object.

Answer 2: the two Secret objects are wrappers on the same Juju secret, and `get_content()` should give me the content that the Juju `secret-get` command provides.

At the moment, ops gives the former:

```
$ juju run different-secret-refreshes/0 run
Running operation 472 with 1 task
  - task 473 on unit-different-secret-refreshes-0

Waiting for task 473...
content1:
  foo: baz
content2:
  foo: bar
```

I have some sympathy for the idea that the content is part of the Python Secret object and so shouldn't change between `get_content()` calls if *that specific object* hasn't told it to. However, I think this is confusing, because charmers don't think of these as Python object, they think of them as the Juju secret, and that has changed.

dimaqq:
Fixes #1536

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -323,12 +323,10 @@ class Model:
         if id is not None:
             # Canonicalize to "secret:<id>" form for consistency in backend calls.
             id = Secret._canonicalize_id(id, self.uuid)
-        content = self._backend.secret_get(id=id, label=label)
         return Secret(
             self._backend,
             id=id,
             label=label,
-            content=content,
             _secret_set_cache=self._cache._secret_set_cache,
         )
 
@@ -544,7 +542,6 @@ class Application:
             self._backend,
             id=id,
             label=label,
-            content=content,
             _secret_set_cache=self._cache._secret_set_cache,
         )
 
@@ -736,7 +733,6 @@ class Unit:
             self._backend,
             id=id,
             label=label,
-            content=content,
             _secret_set_cache=self._cache._secret_set_cache,
         )
 
@@ -1352,7 +1348,8 @@ class Secret:
         self._backend = backend
         self._id = id
         self._label = label
-        self._content = content
+        if content is not None:
+            warnings.warn('Secret(content=???) is deprecated', DeprecationWarning, stacklevel=2)
         self._secret_set_cache: collections.defaultdict[str, dict[str, Any]] = (
             collections.defaultdict(dict) if _secret_set_cache is None else _secret_set_cache
         )
@@ -1492,26 +1489,22 @@ class Secret:
     def get_content(self, *, refresh: bool = False) -> dict[str, str]:
         """Get the secret's content.
 
-        The content of the secret is cached on the :class:`Secret` object, so
-        subsequent calls do not require querying the secret storage again,
-        unless ``refresh=True`` is used, or :meth:`set_content` is called.
-
         Returns:
             A copy of the secret's content dictionary.
 
         Args:
-            refresh: If true, fetch the latest revision's content and tell
-                Juju to update to tracking that revision. The default is to
-                get the content of the currently-tracked revision.
+            refresh: ignored. Retained for backwards compatibility.
 
         Raises:
             SecretNotFoundError: if the secret no longer exists.
             ModelError: if the charm does not have permission to access the
                 secret.
+
+        .. versionchanged:: 3.3.0
+          Secret content is no longer cached in Ops. Each ``get_content()`` call
+          queries the secret storage again. The ``refresh`` argument is ignored.
         """
-        if refresh or self._content is None:
-            self._content = self._backend.secret_get(id=self.id, label=self.label, refresh=refresh)
-        return self._content.copy()
+        return self._backend.secret_get(id=self.id, label=self.label, refresh=refresh)
 
     def peek_content(self) -> dict[str, str]:
         """Get the content of the latest revision of this secret.
--- a/ops/model.py
+++ b/ops/model.py
@@ -1307,6 +1307,19 @@ class SecretInfo:
             f'description={self.description!r})'
         )
 
+    def __eq__(self, other: object):
+        if not isinstance(other, SecretInfo):
+            return NotImplemented
+        return (
+            self.id == other.id
+            and self.label == other.label
+            and self.revision == other.revision
+            and self.expires == other.expires
+            and self.rotation == other.rotation
+            and self.rotates == other.rotates
+            and self.description == other.description
+        )
+
 
 class Secret:
     """Represents a single secret in the model.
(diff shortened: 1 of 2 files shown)


dimaqq:
@tonyandrewmeyer I think it's a reviewable draft :)

- secret content is no longer cached
- secret metadata cache machinery is still there, and metadata is reused
- a deprecation warning if someone tries to create a `Secret` object with content
- doc update

A few open questions:
- do we actually want to keep secret metadata machinery?
- did I forget anything major?

To do:
- we're not loading secret content immediately any more, which means a bunch of tests need to be updated (the model errors is delayed, test needs to `.get_content()`)
- some e2e tests fail, I haven't drilled into those yet.

ops/model.py
```diff
@@ -1479,26 +1489,22 @@ def _on_secret_changed(self, event):
...
-        unless ``refresh=True`` is used, or :meth:`set_content` is called.
-
         Returns:
             A copy of the secret's content dictionary.
 
         Args:
-            refresh: If true, fetch the latest revision's content and tell
-                Juju to update to tracking that revision. The default is to
-                get the content of the currently-tracked revision.
+            refresh: ignored. Retained for backwards compatibility.
```
tonyandrewmeyer: This isn't part of caching, this is the way to say to Juju "move the pointer to the latest revision of the Secret", so it needs to stay as-is.

ops/model.py
```diff
@@ -1479,26 +1489,22 @@ def _on_secret_changed(self, event):
...
 
         Raises:
             SecretNotFoundError: if the secret no longer exists.
             ModelError: if the charm does not have permission to access the
                 secret.
+
+        .. versionchanged:: 3.3.0
+          Secret content is no longer cached in Ops. Each ``get_content()`` call
+          queries the secret storage again. The ``refresh`` argument is ignored.
```
```suggestion
          queries the secret storage again.
```

ops/model.py
```diff
@@ -1339,7 +1348,8 @@ def __init__(
         self._backend = backend
         self._id = id
         self._label = label
-        self._content = content
+        if content is not None:
+            warnings.warn('Secret(content=???) is deprecated', DeprecationWarning, stacklevel=2)
         self._secret_set_cache: collections.defaultdict[str, dict[str, Any]] = (
             collections.defaultdict(dict) if _secret_set_cache is None else _secret_set_cache
         )
```
tonyandrewmeyer: I think this thing can be removed too, no?

ops/model.py
```diff
@@ -323,12 +323,10 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
         if id is not None:
```
tonyandrewmeyer, line 316: The docstring above here (original line 304 onwards) needs updating.

ops/model.py
```diff
@@ -1307,6 +1303,19 @@ def __repr__(self):
...
+        return (
+            self.id == other.id
+            and self.label == other.label
+            and self.revision == other.revision
+            and self.expires == other.expires
+            and self.rotation == other.rotation
+            and self.rotates == other.rotates
+            and self.description == other.description
+        )
+
```
tonyandrewmeyer: Is this required, or just a drive-by?

I feel like `==` for two `ops.SecretInfo` objects is a bit tricky, because it's not clear to the user if this is comparing the underlying metadata or just what's in memory. If it's the underlying metadata, then just `id` or `label` (or both) matching would be equal. If it's what's in memory then it's what you have but it could actually compare false but still be the same secret.

tonyandrewmeyer:
> * do we actually want to keep secret metadata machinery?

Sorry, not sure what this is referring to. Do you mean `SecretInfo`? Or something else here?

> To do:

To add to the list (sorry!) I do think we need to do some sort of review to get an idea of how many charms will be impacted here. I guess mostly that's how often per hook `get_content()` is called?

Probably also a heads-up in Charm Development (and maybe Accendo?) that this is coming.

ops/model.py
dimaqq: Indeed I misunderstood what refresh does.
Will revert the doc change and probably add more tests too!

ops/model.py
dimaqq: It's mostly drive-by.

I tried comparing `secret1.get_info() == secret2.get_info()` in the unit test and that failed because the info objects are instantiated on the fly, even though the underlying data comes from a shared cache.

I followed `__repr__` for the fields. I figured that objects should be equal roughly when their representations are equal.

The alternative is for the unit test to be more like:

```py
assert secret1.get_info().id == secret2.get_info().id
assert secret1.get_info().description == secret2.get_info().description

assert secret1.get_info().revision == secret2.get_info().revision
```

This test is actually more explicit too, as both pure metadata are compared as well as revision that comes from Juju (or Scenario).

I'm probably leaning towards removing equality operator and making the unit test more explicit.

ops/model.py
dimaqq: This is the metadata cache, that is correctly shared between multiple instances of `Secret`'s. It's used in `.get_info()`.

It's independent from the secret value cache that's being removed.

That's what my top-level question is about, if we mean to whack this cache too.

tonyandrewmeyer: > That's what my top-level question is about, if we mean to whack this cache too.

This was added because Juju sets both metadata and content with `secret-set` even though ops splits those out into two different methods. Before this, there was buggy behaviour, particularly around caching, if charms used both ops methods, because they would clobber each other at the hook command level.

So it seems to me that there's no point in keeping it with the content caching removed.

tonyandrewmeyer: Actually, I might be wrong about that - if subsequent `secret-set` calls wipe previous ones, then we need to either keep the cache or get the content to re-set it. I can't remember what the behaviour was, but if you follow the blame to the PR that added that caching, the issue or the PR has details on the Juju behaviour I believe.

ops/model.py
dimaqq: reverted

ops/model.py
dimaqq: reverted.

dimaqq:
will update and reopen when ready.

PR #2039, opened 2025-09-10
docs: clean up Resources.fetch docstring, add ModelError exception

### LINKED ISSUE #2036: Improve docstring for fetch method of model resource
The `fetch` method 
https://github.com/canonical/operator/blob/e33459f28197eeecff407c42cf731d9d8f6d9a7c/ops/model.py#L2413
can also raise `ModelError` from the call to https://github.com/canonical/operator/blob/e33459f28197eeecff407c42cf731d9d8f6d9a7c/ops/model.py#L3619

but the current docstring for this method only mentions about `NameError` raise.

benhoyt:
This adds the `ModelError` to the docstring, and adjust the docstring to use "Returns" properly.

[**Docs preview**](https://canonical-ubuntu-documentation-library--2039.com.readthedocs.build/ops/2039/reference/ops/#ops.Resources.fetch)

Fixes #2036.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -2419,7 +2419,7 @@ class Resources:
 
         Raises:
             NameError: if the resource name is not in the charm metadata.
-            ModelError: if the controller is unable to fetch the resource, for
+            ModelError: if the controller is unable to fetch the resource; for
                 example, if you ``juju deploy`` from a local charm file and
                 forget the appropriate ``--resource``.
         """


ops/model.py
```diff
@@ -2412,13 +2412,16 @@ def __init__(self, names: Iterable[str], backend: _ModelBackend):
...
 
-        If successfully fetched, this returns the path where the resource is stored
-        on disk, otherwise it raises a :class:`NameError`.
+        Returns:
+            The path where the resource is stored on disk.
 
         Raises:
-            NameError: if the resource's path cannot be fetched.
+            NameError: if the resource name is not in the charm metadata.
+            ModelError: if the controller is unable to fetch the resource, for
```
tonyandrewmeyer: Very much a nit, but I think this technically should be:

```suggestion
            ModelError: if the controller is unable to fetch the resource; for
```

tonyandrewmeyer (approved):
Nice - the extra improvements other than the extra raises make it much better.

PR #2051, opened 2025-09-19
docs: note that the peer databag isn't usable during the install event

### LINKED ISSUE #1452: Cannot do peer data IO during install
I learned last week that you cannot read/write to the peer relation databag during install.

[See juju bug](https://bugs.launchpad.net/juju/+bug/2086837)

[See thread in matrix](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$mG2TmMtGkEkEqudrHda-iI8a5iojGbSVKhNik4lCiAE?via=ubuntu.com&via=matrix.org&via=xentonix.net)

At the moment there is no good way to test this behaviour since scenario.State doesn't keep track of what events you've seen in the past.
Also, ops doesn't give you nice errors. I have encountered several times a 'permission denied' ModelError caused by attempting IO on peer data during `install`. Not sure how `ops` can surface this better though, because ops also doesn't keep track of the event history.

Perhaps we could add a `NotAccessible` singleton to allow the user to do:

`state = State(relations={Relation(remote_app_data=NotAccessible(), local_app_data=NotAccessible())})`
 
to signal that the relation is there, yet the databags can't be written or read.

tonyandrewmeyer:
Add a note to the `Relation.data` docs that points out that peer relation data cannot be used during the `install` event.

Fixes #1452

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1736,6 +1736,9 @@ class Relation:
     """Holds the data buckets for each entity of a relation.
 
     This is accessed using, for example, ``Relation.data[unit]['foo']``.
+
+    Note that peer relation data is not readable or writable during the Juju ``install``
+    event, even though the relation exists.
     """
 
     active: bool
--- a/ops/model.py
+++ b/ops/model.py
@@ -1738,7 +1738,7 @@ class Relation:
     This is accessed using, for example, ``Relation.data[unit]['foo']``.
 
     Note that peer relation data is not readable or writable during the Juju ``install``
-    event, even though the relation exists.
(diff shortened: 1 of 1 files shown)


ops/model.py
```diff
@@ -1736,6 +1736,9 @@ class Relation:
     """Holds the data buckets for each entity of a relation.
 
     This is accessed using, for example, ``Relation.data[unit]['foo']``.
+
+    Note that peer relation data is not readable or writable during the Juju ``install``
+    event, even though the relation exists.
```
```suggestion
    event, even though the relation exists. :class:`ModelError` will be raised in that case.
```

dwilding (approved):

PR #2053, opened 2025-09-19
chore: secure permission bits on the tracing database file

### LINKED ISSUE #1950: Restrict the permissions of the tracing database
We should be able to restrict (to 0600) the permissions of the tracing database in the same way that we do the unit state database.

dimaqq:
This change ensures that the tracing data file has same minimal permissions as the unit state file.

fly-by: `path|str --> path` type change in the private api

not included: custom exceptions, I figured if sqlite can connect to the path, then chmod should work as well.

note that we don't care about `:memory:` paths because tracing is mocked out at a different level in `ops[testing]` and is disabled completely in legacy unit testing framework.

Fixes #1950

### DIFF
--- a/tracing/ops_tracing/_buffer.py
+++ b/tracing/ops_tracing/_buffer.py
@@ -19,6 +19,7 @@ import dataclasses
 import functools
 import pathlib
 import sqlite3
+import stat
 from typing import TYPE_CHECKING, Callable
 
 if TYPE_CHECKING:
@@ -89,6 +90,8 @@ class Buffer:
     Access to the buffer attributes is effectively protected by an sqlite transaction.
     """
 
+    path: pathlib.Path
+    """Location of the buffer on disk."""
     ids: set[int]
     """Trace data ids buffered during this dispatch invocation."""
     observed = False
@@ -96,14 +99,17 @@ class Buffer:
     stored: int | None = None
     """Estimate of stored data size in bytes if known, or None if not yet known."""
 
-    def __init__(self, path: pathlib.Path | str):
-        self.path = str(path)
+    def __init__(self, path: pathlib.Path):
+        self.path = path
         self.ids = set()
         self._set_db_schema()
 
     @retry
     def _set_db_schema(self) -> None:
         with self.tx(timeout=LONG_DB_TIMEOUT) as conn:
+            mode = stat.S_IRUSR | stat.S_IWUSR
+            if stat.S_IMODE(self.path.stat().st_mode) != mode:
+                self.path.chmod(mode)
             conn.execute(
                 """
                 CREATE TABLE IF NOT EXISTS tracing (
--- a/tracing/ops_tracing/_buffer.py
+++ b/tracing/ops_tracing/_buffer.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import contextlib
 import dataclasses
 import functools
+import os
 import pathlib
 import sqlite3
 import stat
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
Looks good, thanks!

I didn't test this in a real Juju environment - I'm assuming that you've done that and everything was fine. It seems like the upgrade case should also be fine - in K8s it'll be a new file (so it doesn't matter if charm-user has changed) and on machine there's no charm-user so it'll always be `root` owning this file.

tracing/ops_tracing/_buffer.py
```diff
@@ -89,21 +90,26 @@ class Buffer:
...
+        self.path = path
         self.ids = set()
         self._set_db_schema()
 
     @retry
     def _set_db_schema(self) -> None:
         with self.tx(timeout=LONG_DB_TIMEOUT) as conn:
+            mode = stat.S_IRUSR | stat.S_IWUSR
+            if stat.S_IMODE(self.path.stat().st_mode) != mode:
+                self.path.chmod(mode)
```
benhoyt, line 115: This is a somewhat different way than we [do it for the unit state file](https://github.com/canonical/operator/blob/24f13813f3a44175fa42c095ed74750ca7e3a885/ops/storage.py#L76). It looks like there we create it *before* SQLite creates the file (and chmod if the file exists), whereas here we do it after SQL creates it (with presumably slightly looser permissions). I think this is a tiny race condition, and perhaps we should change it to both use the other approach.

benhoyt (commented):
One comment about the approach.

And yes, let's ensure we test manually on both k8s and machine (or maybe we have integration tests that cover this?).

tracing/ops_tracing/_buffer.py
james-garner-canonical: If I understand you correctly, @benhoyt, your concern is that there's potentially a window where the file exists on disk with the looser permissions than desired between the creation of the file by SQLite and our chmod (and I guess potentially sticks around with those permissions if we were to crash here?).

I'm not sure how big a concern this is, but it does seem reasonable to me to follow the approach we use already -- chmod existing file or atomically create with the correct mode before connecting to the database.

benhoyt: Yeah, that's right -- it's a race condition or a kind of [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue. I'm not actually too worried about it being exploited here, but it seems better to do the right thing (as we've done elsewhere in the codebase).

dimaqq: Ah, I misread the unit state db code.

If the file doesn't exist, a new one with specific permission flags is created.
That makes sense. I should probably do the same.

P.S. I didn't realise that sqlite allows to connect to an empty (size 0) file.

james-garner-canonical (approved):

james-garner-canonical:
Small whitespace formatting issue to fix, but looks good to me. As the others have said, assuming manual verification was fine.

dimaqq:
Sorry I forgot to push the last commit 🙈

tracing/ops_tracing/_buffer.py
dimaqq: done

dimaqq:
> Looks good to me, thanks! Can you please that you've manually tested end-to-end before merging?

Tested manually by patching a running charm with locally build `ops_tracing`:
- k8s; existing db file, old/wrong mode --> gets reset to `-rw-------`, no errors in the log
- k8s; missing db file --> file appears with mode `-rw-------`, no errors in the log

I think our integration and 3p charm tests should actually catch it, if there's some major problem.

PR #2058, opened 2025-09-26
ci: clean up PYTHONPATH in tox.ini

### LINKED ISSUE #1825: Clean up tox.ini
For example there's this line:
```ini
  PYTHONPATH = {toxinidir}:{toxinidir}/lib:{[vars]src_path}
```

Which appears to have been in the repo since tox.ini was added many years ago, which could have been a copy-paste from some charm repo.

This project doesn't have `./lib`.

Likewise, `src_path` (which is `./ops`) should not be in PYTHOPATH, as ops is a package.

Tentative idea: `git blame` and go through all old lines with a haircomb.

benhoyt:
Per #1825, we can simplify the PYTHONPATH here. In fact, it seems like we don't need it at all, so just drop it entirely.

I've skimmed over the rest of tox.ini as well, and it doesn't look like there's anything else egregious.

Fixes #1825

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 [tox]
-skipsdist=True
+skipsdist = True
 skip_missing_interpreters = True
 envlist = lint, unit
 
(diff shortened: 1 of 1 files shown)


dimaqq (approved):

dimaqq:
P.S. I figure `skip_missing_interpreters` need not be set; but that's minor.

PR #2062, opened 2025-09-28
docs: add a how-to guide for migrating away from Harness

dwilding:
This PR adds a how-to guide for migrating unit tests away from Harness. I've not linked the new guide from any other docs yet, or removed [How to write legacy unit tests for a charm](https://documentation.ubuntu.com/ops/latest/howto/legacy/write-legacy-unit-tests-for-a-charm/). I'll leave those tasks for a follow-up PR.

### DIFF
--- a/docs/howto/legacy/migrate-unit-tests-from-harness.md
+++ b/docs/howto/legacy/migrate-unit-tests-from-harness.md
@@ -1,0 +1,669 @@
+(harness-migration)=
+# How to migrate unit tests from Harness
+
+TODO: Introduction, including links to other resources.
+
+(harness-migration-action)=
+## Test a minimal action
+
+Suppose that we have the following charm code:
+
+```python
+"""A demo charm."""
+
+import ops
+
+
+class DemoCharm(ops.CharmBase):
+    """Manage the workload."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        framework.observe(self.on.get_value_action, self._on_get_value_action)
+        ...  # Observe other events.
+
+    def _on_get_value_action(self, event: ops.ActionEvent) -> None:
+        """Handle the get-value action."""
+        if event.params["value"] == "please fail":
+            event.fail("Action failed, as requested")
+        else:
+            event.set_results({"out-value": event.params["value"]})
+
+    ...  # Handle other events.
+
+
+if __name__ == "__main__":
+    ops.main(DemoCharm)
+```
+
+Also suppose that we have the following testing code, written using Harness:
+
+```python
+from ops import testing
+
+from charm import DemoCharm
+
+
+def test_action():
+    harness = testing.Harness(DemoCharm)
+
+    harness.begin()
+    output = harness.run_action("get-value", {"value": "foo"})
+    assert output.results == {"out-value": "foo"}
+
+    harness.cleanup()
+```
+
+Let's implement this test as a state-transition test.
+
+We'll structure the test around [`Context.on.action`](ops.testing.CharmEvents.action). Our expected input and output states are:
+
+- Before `action` runs -- A generic state. The action doesn't use data from the charm or workload.
+- After `action` runs -- A successful action result.
+
+Here's the test:
+
+```python
+from ops import testing
+
+from charm import DemoCharm
+
+
+def test_get_value_action():
+    ctx = testing.Context(DemoCharm)
+
+    state_in = testing.State()
+    ctx.run(ctx.on.action("get-value", params={"value": "foo"}), state_in)
+    assert ctx.action_results == {"out-value": "foo"}
+```
+
+While we're here, let's write a test for the action's failure case:
+
+```python
+import pytest
+from ops import testing
+
+from charm import DemoCharm
+
+
+def test_get_value_action_failed():
+    ctx = testing.Context(DemoCharm)
+
+    state_in = testing.State()
+    with pytest.raises(testing.ActionFailed) as exc_info:
+        ctx.run(ctx.on.action("get-value", params={"value": "please fail"}), state_in)
+    assert exc_info.value.message == "Action failed, as requested"
+```
+
+In a more realistic charm, the action will use data from the charm or workload. For examples, see [](#harness-migration-relation) and [](#harness-migration-container). When writing state-transition tests for a real action, we also need to consider collect-status.
+
+(harness-migration-collect-status)=
+## Consider collect-status
+
+A significant difference between Harness tests and state-transition tests is how the testing framework handles `collect_unit_status` and `collect_app_status`.
+
+- In a Harness test, you need to trigger these events using [`evaluate_status`](ops.testing.Harness.evaluate_status).
+- In a state-transition test, the framework automatically triggers these events. This matches what happens when the charm is deployed.
+
+To demonstrate the difference, let's modify the charm code from [](#harness-migration-action) to include a workload container and observe `collect_unit_status`:
+
+```python
+"""A demo charm that has a workload container."""
+
+import ops
+
+
+class DemoCharm(ops.CharmBase):
+    """Manage the workload."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        self.container = self.unit.get_container("my-container")
+        framework.observe(self.on.collect_unit_status, self._on_collect_status)
+        framework.observe(self.on.get_value_action, self._on_get_value_action)
+        ...  # Observe other events, including pebble-ready.
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
+        """Report the status of the workload."""
+        try:
+            service = self.container.get_service("workload")
+            if not service.is_running():
+                event.add_status(ops.MaintenanceStatus("waiting for workload"))
+        except (ops.ModelError, ops.pebble.ConnectionError):
+            # ModelError: The service wasn't found, likely because we define
+            # the service on pebble-ready and that event hasn't happened yet.
+            # pebble.ConnectionError: Pebble isn't ready yet, or perhaps the
+            # connection to the container is temporarily down.
+            event.add_status(ops.MaintenanceStatus("waiting for container"))
+        event.add_status(ops.ActiveStatus())
+
+    def _on_get_value_action(self, event: ops.ActionEvent) -> None:
+        """Handle the get-value action."""
+        if event.params["value"] == "please fail":
+            event.fail("Action failed, as requested")
+        else:
+            event.set_results({"out-value": event.params["value"]})
+
+    ...  # Handle other events, including pebble-ready.
+
+
+if __name__ == "__main__":
+    ops.main(DemoCharm)
+```
+
+Our Harness test `test_action` still works (although it doesn't exercise `_on_collect_status`). However, our state-transition test `test_get_value_action` now fails with an error:
+
+```text
+FAILED tests/unit/test_charm.py::test_get_value_action -
+scenario.errors.UncaughtCharmError: Uncaught RuntimeError in charm, ...
+```
+
+When `test_get_value_action` uses `Context.on.action` to simulate running the action, the testing framework also simulates running `_on_collect_status`. Then, when `_on_collect_status` tries to determine the status of the workload, `self.container.get_service("workload")` produces an error because we didn't provide a container to the testing framework.
+
+As a reminder, here's our definition of `test_get_value_action`:
+
+```python
+def test_get_value_action():
+    ctx = testing.Context(DemoCharm)
+
+    state_in = testing.State()
+    ctx.run(ctx.on.action("get-value", params={"value": "foo"}), state_in)
+    assert ctx.action_results == {"out-value": "foo"}
+```
+
+To fix the test, we need to add a mock container to the input state:
+
+```python
+def test_get_value_action():
+    ctx = testing.Context(DemoCharm)
+
+    container = testing.Container("my-container", can_connect=True)
+    state_in = testing.State(containers={container})
+    ctx.run(ctx.on.action("get-value", params={"value": "foo"}), state_in)
+    assert ctx.action_results == {"out-value": "foo"}
+```
+
+> See more:
+>
+> - [`testing.Container`](ops.testing.Container)
+> - [`testing.State`](ops.testing.State)
+
+In [](#harness-migration-container), we'll work through a more realistic example that shows:
+
+- How to test the pebble-ready event handler.
+- How to test the status reporting logic in `_on_collect_status`.
+- How to test an action that interacts with the workload container.
+
+(harness-migration-relation)=
+## Test how a relation is handled
+
+### Charm with Harness tests
+
+Suppose that we have the following charm code:
+
+```python
+"""A demo charm that requires a database."""
+
+import ops
+
+# Use the 'data_interfaces' library to help integrate with a database provider.
+from charms.data_platform_libs.v0.data_interfaces import (
+    DatabaseCreatedEvent,
+    DatabaseEndpointsChangedEvent,
+    DatabaseRequires,
+)
+
+
+class DemoCharm(ops.CharmBase):
+    """Manage the workload."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        self.database = DatabaseRequires(self, relation_name="database", database_name="my-db")
+        framework.observe(self.database.on.database_created, self._on_database_available)
+        framework.observe(self.database.on.endpoints_changed, self._on_database_available)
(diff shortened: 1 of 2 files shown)


docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+                event.add_status(ops.MaintenanceStatus("waiting for workload"))
+        except (ops.ModelError, ops.pebble.ConnectionError):
+            # ModelError: The service wasn't found, likely because we define
+            # the service on pebble-ready and that event hasn't happened yet.
+            # pebble.ConnectionError: Pebble isn't ready yet, or perhaps the
+            # connection to the container is temporarily down.
+            event.add_status(ops.MaintenanceStatus("waiting for container"))
+        event.add_status(ops.ActiveStatus())
+
+    ...  # Handle other events.
```
dwilding: I was originally going to include an action that executes a command in the workload container. I removed that because I wasn't confident in the approach, and this PR was getting big enough already. I might add that back in a follow-up PR

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
+(harness-migration)=
+# How to migrate unit tests from Harness
+
+This guide is a starting point for how to migrate [Harness](/reference/ops-testing-harness) tests to state-transition tests. Both approaches to unit testing use the `ops.testing` framework, but Harness tests are deprecated and won't be supported by default in a future Ops release.
```
tonyandrewmeyer:
```suggestion
This guide is a starting point for how to migrate [Harness](/reference/ops-testing-harness) tests to state-transition tests. Both approaches to unit testing use the `ops.testing` namespace, but Harness tests are deprecated and won't be supported by default in a future Ops release.
```

They're really two different frameworks that are (for better or worse) sharing a namespace.

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+Each state-transition test is an isolated test of a particular event handler. This matches how the charm runs when deployed; each time Juju triggers an event, the charm code is run from scratch and handles that event.
+
+In contrast, Harness tests typically build up a state event-by-event, using the same instantiation of the charm class for each event. This introduces a danger of the charm getting into an unrealistic state.
+
+To help focus on the differences between the two approaches, we don't use fixtures in this guide.
+
+For more information about state-transition testing, see:
+
+- [](#write-unit-tests-for-a-charm)
+- [The reference docs](/reference/ops-testing), especially [](ops.testing.State) and [](ops.testing.CharmEvents)
```
tonyandrewmeyer: Interesting to choose `CharmEvents` here. I would have thought that `Context` and `State` were the two you'd most what to see. I guess `CharmEvents` has all of the events and they aren't 100% the same as in ops, although very similar.

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+
+import ops
+
+
+class DemoCharm(ops.CharmBase):
+    """Manage the workload."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        framework.observe(self.on.get_value_action, self._on_get_value_action)
```
tonyandrewmeyer: I think we're trying to stick to this style now:

```suggestion
        framework.observe(self.on['get-value'].action, self._on_get_value_action)
```

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+
+```python
+from ops import testing
+
+from charm import DemoCharm
+
+
+def test_get_value_action():
+    ctx = testing.Context(DemoCharm)
+
```
tonyandrewmeyer: A nit, but I would not have this blank line (also also throughout all the other examples). I think if you're going to have blank lines in Scenario tests, they'd be between arrange/act/assert, not here.

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+class DemoCharm(ops.CharmBase):
+    """Manage the workload."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        self.database = DatabaseRequires(self, relation_name="database", database_name="my-db")
+        framework.observe(self.database.on.database_created, self._on_database_available)
+        framework.observe(self.database.on.endpoints_changed, self._on_database_available)
+        framework.observe(self.on.get_db_endpoint_action, self._on_get_db_endpoint_action)
```
tonyandrewmeyer: And like above:

```suggestion
        framework.observe(self.on['get_db_endpoint'].action, self._on_get_db_endpoint_action)
```

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+    container_in = testing.Container("my-container", can_connect=True)
+    state_in = testing.State(containers={container_in})
+    state_out = ctx.run(ctx.on.pebble_ready(container_in), state_in)
+    container_out = state_out.get_container(container_in.name)
+    assert "workload" in container_out.plan.services
+    assert container_out.plan.services["workload"].command == "run-workload"
+```
+
+```{note}
+When writing a state-transition test, it's best to treat objects as immutable. We shouldn't expect `ctx.run` to modify the mock container that's passed to `ctx.on.pebble_ready`. In other words, the test would break if we replaced the line `assert "workload" ...` with `assert "workload" in container_in.plan.services`.
```
tonyandrewmeyer: I think we should be more assertive here, something like:

```suggestion
In state-transition tests, the objects in the `State` are immutable. Calling `ctx.run` will modify the mock container in the output state, not the one that's passed to `ctx.on.pebble_ready`. In other words, the test would break if we replaced the line `assert "workload" ...` with `assert "workload" in container_in.plan.services`.
```

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+import ops
+
+
+class DemoCharm(ops.CharmBase):
+    """Manage the workload."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        framework.observe(self.on.get_value_action, self._on_get_value_action)
+        ...  # Observe other events.
```
dimaqq:
```suggestion
        ...
```

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+
+    def _on_get_value_action(self, event: ops.ActionEvent) -> None:
+        """Handle the get-value action."""
+        if event.params["value"] == "please fail":
+            event.fail("Action failed, as requested")
+        else:
+            event.set_results({"out-value": event.params["value"]})
+
+    ...  # Handle other events.
```
dimaqq:
```suggestion
```

I would drop the "other events" concept entirely.
It's pretty clear that charms handle multiple events.

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+        """Report the status of the workload."""
+        try:
+            service = self.container.get_service("workload")
+            if not service.is_running():
+                event.add_status(ops.MaintenanceStatus("waiting for workload"))
+        except (ops.ModelError, ops.pebble.ConnectionError):
+            # ModelError: The service wasn't found, likely because we define
+            # the service on pebble-ready and that event hasn't happened yet.
+            # pebble.ConnectionError: Pebble isn't ready yet, or perhaps the
+            # connection to the container is temporarily down.
```
dimaqq: I'd drop this entirely.

Or, if needed cover this in one line, like "service wasn't defined; container not ready"

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+            event.add_status(ops.MaintenanceStatus("waiting for container"))
+        event.add_status(ops.ActiveStatus())
+
+    def _on_get_value_action(self, event: ops.ActionEvent) -> None:
+        """Handle the get-value action."""
+        if event.params["value"] == "please fail":
+            event.fail("Action failed, as requested")
+        else:
+            event.set_results({"out-value": event.params["value"]})
```
dimaqq: I'd drop this entirely, or if it's really needed, then:


```suggestion
    def _on_get_value_action(self, event: ops.ActionEvent) -> None:
        # See above
```

dimaqq (commented):
Only a partial review.

I feel that as the examples get longer and longer, at some point the doc becomes tedious.

Could the examples be trimmed to testing one thing at a time, with the mocking step, but without the actual mocks?

A tutorial would be self-contained, while the how-to guide could be more focused.

WDYT?

Rule of thumb: the doc renders 8 pages long for me (reasonable font, large monitor). At this point it could be split into 4~6 separate docs.

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+1. When the relation data changes, the charm writes new workload config.
+2. The action returns the correct value from the current relation data.
+
+Let's implement these tests as independent state-transition tests.
+
+### Test a data change
+
+To test what happens when relation data changes, we'll structure a test around [`Context.on.relation_changed`](ops.testing.CharmEvents.relation_changed). Our expected input and output states are:
+
+- Before `relation_changed` runs -- A mock workload object configured with `foo.local:1234`. Relation data `bar.local:5678`.
```
tonyandrewmeyer, line 267: I wonder if it's worth adding a little something about the relation data. With Harness, you have to set up the harness with the old value (`foo.local:1234`) and then you do an update operation that has the new value. With Scenario you're starting with the Juju state as it will be at the start of the `relation-changed` event, so the `State` already has the *new* value `bar.local:5678`, and the old value is only in the (mock) workload state.

I think this is one of the differences that is harder to get used to, when moving from Harness to Scenario, because you have to think about different points in the process.

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+    assert "workload" in container_out.plan.services
+    assert container_out.plan.services["workload"].command == "run-workload"
+    assert state_out.unit_status == testing.ActiveStatus()
+```
+
+### Test the status reporting
+
+We already have a test (`test_pebble_ready`) that partially exercises the charm's status reporting, but it's a good idea to independently cover each situation that the status reporting logic accounts for.
+
+Our Harness test uses [`evaluate_status`](ops.testing.Harness.evaluate_status) to trigger `collect_unit_status`, which causes the charm's `_on_collect_status` method to run. In a state-transition test, the best approach is to simulate the update-status event instead.
```
tonyandrewmeyer: I think it would be good to explain why this event is chosen, to avoid confusion that links "update-status" and "collect-status". Maybe something like:

```suggestion
Our Harness test uses [`evaluate_status`](ops.testing.Harness.evaluate_status) to trigger `collect_unit_status`, which causes the charm's `_on_collect_status` method to run. In a state-transition test, the best approach is to simulate a Juju event that isn't observed instead. The typical choice is update-status.
```

docs/howto/legacy/migrate-unit-tests-from-harness.md
james-garner-canonical: Though would that be `'get-value'` then?

docs/howto/legacy/migrate-unit-tests-from-harness.md
```diff
@@ -0,0 +1,626 @@
...
+        try:
+            service = self.container.get_service("workload")
+            if not service.is_running():
+                event.add_status(ops.MaintenanceStatus("waiting for workload"))
+        except (ops.ModelError, ops.pebble.ConnectionError):
+            # ModelError: The service wasn't found, likely because we define
+            # the service on pebble-ready and that event hasn't happened yet.
+            # pebble.ConnectionError: Pebble isn't ready yet, or perhaps the
+            # connection to the container is temporarily down.
+            event.add_status(ops.MaintenanceStatus("waiting for container"))
```
james-garner-canonical, line 123:
```suggestion
        try:
            service = self.container.get_service("workload")
        except (ops.ModelError, ops.pebble.ConnectionError):
            # Service not defined yet or can't connect to Pebble.
            event.add_status(ops.MaintenanceStatus("waiting for container"))
        else:
            if not service.is_running():
                event.add_status(ops.MaintenanceStatus("waiting for workload"))
```

james-garner-canonical: I prefer `try ... except ... else` to make it clear where the error is coming from. Also a suggestion on how to make the comments shorter -- I like the detail in the original comments, they make it clear why we're doing what we are here, but the detail to some extent isn't relevant to the goal of the guide. Do we document this pattern anywhere else? I wonder if it's worth capturing this demo charm somewhere (e.g. in our examples?) or making sure this pattern is included in one of our existing examples (we could then link to it from here for details if needed).

james-garner-canonical (approved):
This is looking great to me, modulo the existing suggestions.

docs/howto/legacy/migrate-unit-tests-from-harness.md
dwilding: I've reworked this bit, partly after your feedback. To make the links easier to find (and to make the intro shorter) I've added a dedicated [See more](https://canonical-ubuntu-documentation-library--2062.com.readthedocs.build/ops/2062/howto/legacy/migrate-unit-tests-from-harness/#see-more) section at the end of the doc. In that, I've linked to `Context`, `State`, and `CharmEvents`.

docs/howto/legacy/migrate-unit-tests-from-harness.md
dwilding: I've switched to `try ... except ... else` - thanks.

Per Dima's suggestion, I've dropped the long comments about error handling. Instead, I've linked to some examples in the new [See more](https://canonical-ubuntu-documentation-library--2062.com.readthedocs.build/ops/2062/howto/legacy/migrate-unit-tests-from-harness/#see-more) section. The examples are a better place to explain what's going on.

I've also updated the part of the doc that talks about "two more situations" that aren't covered by the unit tests. I'm now mentioning the relevant bits of charm code:

> - The container has a Pebble plan, but the service isn't running. This corresponds to the `not service.is_running()` part of the logic. To cover this situation, we could add a variant of `test_status_active` that sets the service status to `INACTIVE`.
> - The container is available, but doesn't have a Pebble plan. This corresponds to the `ops.ModelError` part of the logic. To cover this situation, we could add a variant of `test_status_container_down` that sets `can_connect=True`.

james-garner-canonical: Sounds good to me, thanks.

docs/howto/legacy/migrate-unit-tests-from-harness.md
dwilding: Good idea - I've added some extra explanation:

> It might be surprising that the input state has relation data `bar.local:5678`, not `foo.local:1234`. This highlights a significant difference between Harness tests and state-transition tests:
>
> - Our Harness test uses `update_relation_data` to simulate Juju *telling* the charm about new relation data, so we start by setting the old relation data `foo.local:1234` in the mock Juju state.
> - Our state-transition test will simulate the charm *learning* about new relation data, so we'll start with new relation data and check that the charm changes the configuration of the mock workload object.

docs/howto/legacy/migrate-unit-tests-from-harness.md
dwilding: Yeah, this makes sense, thanks. I've replaced the whole method by:

```
...  # _on_get_value_action is unchanged.
```

dwilding:
Thanks all for the feedback! I've addressed everything, but won't merge until next week (when Tony is back).

While applying the feedback, I made a couple of changes to reduce the cognitive load of things that aren't relevant to the core point. Thanks Dima for your comments around this.

- Removed the definition of the `MockWorkload` class, which was duplicated between the Harness and state-transition tests. Instead, I've left a comment:

    > Prepare a mock workload object with matching config, assuming we've defined a MockWorkload class with suitable attributes and methods.

- Reduced charm code to the charm class only, since that's all we really care about.

PR #2073, opened 2025-09-30
docs: switch to makefile for building the docs

dwilding:


docs/Makefile
```diff
@@ -0,0 +1,182 @@
...
+	@echo "Run 'make help' to see supported targets."
+
+# This is a custom target for Ops docs, not the target that comes with the starter pack.
+# Our conf.py checks that this target hasn't been overwritten (e.g., during a starter pack upgrade).
+$(VENVDIR): requirements.txt ../uv.lock
+	@which uv || (echo "Error: uv is not installed. To install uv, run 'sudo snap install astral-uv --classic' or see https://docs.astral.sh/uv/"; exit 1)
+	@test -d $(VENVDIR) || uv venv --python 3.11 $(VENVDIR)
+	uv pip install --directory $(VENVDIR) --requirements ../../requirements.txt
+	uv pip install --directory $(VENVDIR) "ops[testing,tracing] @ ../../.."
+	@touch $(VENVDIR)
```
dwilding, line 63: Apart from the `$(VENVDIR)` target, everything else in the makefile matches the starter pack.

tox.ini
dimaqq, line 1: **Praise** I actually like seeing all these hacks gone from tox.ini

docs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+canonical-sphinx[full]
+packaging
+sphinxcontrib-svg2pdfconverter[CairoSVG]
+sphinx-last-updated-by-git
+sphinx-sitemap
```
dimaqq, line 5: **Thinking** this is very KISS, but the versions are no longer pinned. Pros and cons.

docs/Makefile
dimaqq: **Suggestion:** please add a comment above or on this line to this effect pls. Also mention why it's needed.

That way, if I were to update the Makefile a year from now with the newest from the starter pack, the PR diff would scream at me on this very line and I would know what to do about it.

dimaqq (commented):
First pass review

docs/Makefile
dimaqq: Ugh, I'm blind!

CONTRIBUTING.md
```diff
@@ -76,17 +76,30 @@ Recommended tone:
...
+To build the docs:
+
+```sh
+make -C docs html
+```
+
+This generates HTML docs in the `docs/_build` directory. To view the docs, run a web server in the `docs/_build` directory:
+
+```sh
+uvx --directory docs/_build python -m http.server
```
benhoyt: Do we need to present this option, or should we just recommend `make -C docs run` as the one right way?

benhoyt (approved):
Looks good to me, thanks. One nit comment.

CONTRIBUTING.md
dwilding: I don't personally use `make -C docs run` because I find it builds too frequently for my liking (I know, weird complaint...) However, I don't use `python -m http.server` either - I have a general-purpose `serve` command installed on my system.

TL;DR: I'm OK with removing this option and having `make -C docs run` as the one right way. People who prefer a different approach are free to use that approach.

docs/requirements.txt
dwilding: Capturing internal discussion: We decided to leave versions unpinned, to match the starter pack. If this causes problems down the line, we can revisit the decision.

PR #2074, opened 2025-09-30
docs: update info about release docs, mark testing changelog as not maintained

dwilding:
This PR updates the "release documentation" part of our development docs. I was planning to only update the parts about CHANGES.md, but once I got started I reworked the whole of the release documentation section. (Now it's 7pm before a holiday and time to stop :laughing:)

I also reshuffled the order a bit, so that the info about release documentation comes before the release process. Since the release process requires a release summary, I figure it's useful to learn about that in advance.

**[Preview of HACKING.md](https://github.com/dwilding/operator/blob/update-changes-md-info/HACKING.md#releases)**

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -235,8 +235,6 @@ the build frontend is [build](https://pypi.org/project/build/).
 
 ## Release documentation
 
-We produce several pieces of documentation for Ops releases.
-
 As part of the release process, you'll write a summary of the release.
 The summary appears in the GitHub release notes and in Discourse and Matrix.
 
@@ -253,7 +251,7 @@ transition testing".
 
 [CHANGES.md](CHANGES.md) lists the changes in each release. The changelog is
 kept up-to-date by the PR that's created when you run `tox -e draft-release`
-duing the release process. You only need to manually edit the changelog if a
+during the release process. You only need to manually edit the changelog if a
 commit message needs adjusting (we try to avoid doing this).
 
 There's also a changelog for `ops-scenario`:
@@ -270,13 +268,13 @@ edit the draft release after a review.
 ### Discourse and Matrix
 
 After completing the release process, post to
-[framework in Discourse](https://discourse.charmhub.io/c/framework/42) and
+[the 'framework' category in Discourse](https://discourse.charmhub.io/c/framework/42) and
 [Charm Development in Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com).
 
 The Discourse post title should be:
 
 ```
-Ops x.y.z (and ops-scenario w.y.z and ops-tracing x.y.z) released
+Ops x.y.z released
 ```
 
 And the post should resemble this:


HACKING.md
```diff
@@ -231,7 +231,65 @@ dependencies can be found in the relevant `tox.ini` environment `deps` field.
...
+### Discourse and Matrix
+
+After completing the release process, post to
+[framework in Discourse](https://discourse.charmhub.io/c/framework/42) and
+[Charm Development in Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com).
+
+The Discourse post title should be:
+
+```
+Ops x.y.z (and ops-scenario w.y.z and ops-tracing x.y.z) released
```
dimaqq: My 2c:


```suggestion
Ops x.y.z, ops-scenario w.y.z, and ops-tracing x.y.z released
```

Somehow the parentheses irk me.
My opinion could be swayed though.

HACKING.md
```diff
@@ -231,7 +231,65 @@ dependencies can be found in the relevant `tox.ini` environment `deps` field.
...
+
+The GitHub release notes include the summary of the release and
+the list of changes found in the changelog. A draft release is created when
+you run `tox -e draft-release` duing the release process. You might need to
+edit the draft release after a review.
+
+### Discourse and Matrix
+
+After completing the release process, post to
+[framework in Discourse](https://discourse.charmhub.io/c/framework/42) and
```
dimaqq: **Question:**
- https://discourse.charmhub.io/c/framework/42
- https://discourse.charmhub.io/c/framework/

Both work. What does 42 really do?
(other numeric values don't work, so it's not a page number, maybe internal category id?)

tonyandrewmeyer: Discourse does links in a way that's designed to bridge being human-readable and human-writable, being short, and being resilient to change.

The "42" here is the category ID. It's the first choice for identifying which category you're after - you can do `/c/42` and it'll get to you the right place, and you can do `/c/anytexthere/42` (which handles the category being renamed). If you leave off the category ID, then the current description ("framework" in this case) is used to identify the category you're after.

The best way to share Discourse links is to include both, because you get the resilient link to the right thing based on the unchanging ID but also the human-readable name of the category (at least as it was when the link was created).

HACKING.md
tonyandrewmeyer: We dropped the ops-scenario and ops-tracing mentions in the GitHub release title. What about dropping them here as well, do it's just "Ops x.y.z"?

HACKING.md
tonyandrewmeyer: I think "post to framework in Discourse" reads oddly. Can we keep the "category" word?

```suggestion
[the 'framework' category in Discourse](https://discourse.charmhub.io/c/framework/42) and
```

HACKING.md
```diff
@@ -273,118 +331,13 @@ Then, check out the main branch of your forked operator repo and pull upstream t
...
-
-We produce several pieces of documentation for `ops` and `ops-scenario`
-releases, each serving a separate purpose and covering a different level.
-
-Avoid using the word "Scenario", preferring "unit testing API" or "state
-transition testing". Users should install `ops-scenario` with
-`pip install ops[testing]` rather than using the `ops-scenario` package name
-directly.
-
-### `git log`
```
tonyandrewmeyer, line 293: For what it's worth, I think it's fine to drop this. When I originally added this section, some of it was really 'explanation' of the different types of documentation that describe releases/changes/history and how they each serve different purposes and audiences. Basically, I was not just documenting what we have, but trying to convince people that we should have more than one type. That doesn't need to be in HACKING.md.

HACKING.md
```diff
@@ -231,7 +231,65 @@ dependencies can be found in the relevant `tox.ini` environment `deps` field.
 The build backend is [setuptools](https://pypi.org/project/setuptools/), and
 the build frontend is [build](https://pypi.org/project/build/).
 
-# Publishing a Release
+# Releases
+
+## Release documentation
+
+We produce several pieces of documentation for Ops releases.
+
```
tonyandrewmeyer: I think we could drop this paragraph.

```suggestion
```

HACKING.md
```diff
@@ -231,7 +231,65 @@ dependencies can be found in the relevant `tox.ini` environment `deps` field.
...
+There's also a changelog for `ops-scenario`:
+[testing/CHANGES.md](testing/CHANGES.md). Don't add new entries to this file.
+We've kept it for historical reference, but we no longer maintain it.
+
+### GitHub release notes
+
+The GitHub release notes include the summary of the release and
+the list of changes found in the changelog. A draft release is created when
+you run `tox -e draft-release` duing the release process. You might need to
```
tonyandrewmeyer, line 265:
```suggestion
you run `tox -e draft-release` during the release process. You might need to
```

tonyandrewmeyer (approved):
Thanks for taking care of this!

HACKING.md
dwilding: I took this to Matrix and the consensus to simplify to "Ops x.y.z released"

PR #2078, opened 2025-10-01
ci: add integration and state transition tests for the secrets API

dimaqq:
### High level idea

Run the same series of Ops API calls in unit and integration tests.

- Scaffolding
- A set of test cases
- Run each against Scenario
- Run each against Juju (several versions) with Jubilant

### This PR

- Enough scaffolding to get Scenario to work
- One sample test case
- [wip] Juju/Jubilant

### DIFF
--- a/test/charms/test_secrets/.gitignore
+++ b/test/charms/test_secrets/.gitignore
@@ -1,0 +1,3 @@
+requirements.txt
+*.tar.gz
+*.charm
--- a/test/charms/test_secrets/src/charm.py
+++ b/test/charms/test_secrets/src/charm.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import json
+
+import ops
+
+
+class TestSecretsCharm(ops.CharmBase):
+    _stored = ops.StoredState()
+
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        self._stored.set_default(secret_id=None)
+        framework.observe(self.on['exec'].action, self._on_exec)
+
+    def _on_exec(self, event: ops.ActionEvent):
+        """Action to execute arbitrary Python code in specific context."""
+        assert event.params['code']
+        rv = {}
+
+        rv['_before'] = self._snapshot()
+
+        exec(event.params['code'], globals(), locals())  # noqa
+
+        rv['_after'] = self._snapshot()
+
+        event.set_results({'rv': json.dumps(rv)})
+
+    def _snapshot(self):
+        secret_id = self._stored.secret_id
+        if not secret_id:
+            return None
+        secret = self.model.get_secret(id=secret_id)
+        return {
+            'info': _to_dict(secret_info=secret.get_info()),
+            'tracked': secret.get_content(),
+            'latest': secret.peek_content(),
+        }
+
+
+def _to_dict(secret_info: ops.SecretInfo):
+    return {
+        'id': secret_info.id,
+        'label': secret_info.label,
+        'revision': secret_info.revision,
+        'expires': secret_info.expires,
+        'rotation': secret_info.rotation,
+        'rotates': secret_info.rotates,
+        'description': secret_info.description,
+    }
+
+
+if __name__ == '__main__':
+    ops.main(TestSecretsCharm)
--- a/test/integration/test_secrets.py
+++ b/test/integration/test_secrets.py
@@ -1,0 +1,31 @@
+# Copyright 2025 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import Callable
+
+import jubilant
+
+
+def test_foo(build_secret_charm: Callable[[], str], juju: jubilant.Juju):
+    charm_path = build_secret_charm()
+    juju.deploy(charm_path)
+    status = juju.wait(jubilant.all_active)
+
+    unit, unit_obj = next(iter(status.apps['test-secrets'].units.items()))
+    assert unit_obj.leader
+
+    rv = juju.run(unit, 'foo', params={'a': 42})
+    assert rv.results['foo'] == 'bar'
--- a/testing/tests/test_e2e/test_secrets.py
+++ b/testing/tests/test_e2e/test_secrets.py
@@ -1,10 +1,16 @@
 from __future__ import annotations
 
+import ast
 import datetime
+import inspect
+import json
+import textwrap
+from typing import Any, Callable, NamedTuple
+from unittest.mock import ANY
 
 import pytest
 from ops.charm import CharmBase
-from ops.framework import Framework
+from ops.framework import Framework, StoredState
 from ops.model import ModelError
 from ops.model import Secret as ops_Secret
 from ops.model import SecretNotFoundError, SecretRotate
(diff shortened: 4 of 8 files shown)


test/charms/test_secrets/src/__init__.py
dimaqq, line 1: Here I make the test charm a package, so that the charm class can be imported into Scenario tests without hacks.

It feels a bit ugly, but that's the best way I could come up with.

.github/workflows/integration.yaml
```diff
@@ -32,6 +33,45 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        juju-channel:
+          - 3.1/stable
+          - 3.2/stable
+          - 3.3/stable
+          - 3.4/stable
+          - 3.5/stable
+          - 3.6/stable
```
benhoyt, line 52: Do you think we need to run the tests on *all* of these tracks?

test/charms/test_secrets/src/charm.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def _on_exec(self, event: ops.ActionEvent):
+        """Action to execute arbitrary Python code in specific context."""
+        assert event.params['code']
+        rv: dict[str, Any] = {}
+
+        rv['_before'] = self._snapshot()
+
+        try:
+            exec(event.params['code'], globals(), locals())  # noqa
```
benhoyt: Per voice discussion, I'd much prefer to avoid the super-generic `exec` approach and include specific actions that we need for the tests.

test/charms/test_secrets/src/charm.py
```diff
@@ -0,0 +1,78 @@
...
+
+        event.set_results({'rv': json.dumps(rv)})
+
+    def _snapshot(self):
+        secret_id = self._stored.secret_id
+        if not secret_id:
+            return None
+        secret = self.model.get_secret(id=secret_id)
+        return {
+            'info': _to_dict(secret_info=secret.get_info()),
```
benhoyt: Could potentially use `secret.get_info().__dict__`?

.github/workflows/integration.yaml
```diff
@@ -32,6 +33,45 @@ jobs:
...
+          sudo concierge prepare
+          --juju-channel=${{ matrix.juju-channel }}
+          --charmcraft-channel=3.x/stable
+          -p microk8s
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@557e51de59eb14aaaba2ed9621916900a91d50c6  # v6.6.1
+      - run: uv tool install tox --with tox-uv
+      - run: tox -e integration -- --log-cli-level=INFO -s test/integration/test_secrets.py
```
benhoyt, line 77: I guess `test_secrets` will be run once by the `test` job above, and again here for the 3/stable (3.6/stable) track. Do we care?

test/integration/secrets_test_cases.py
```diff
@@ -0,0 +1,157 @@
...
+    # - jubilant_assertions
+    # - setup?
+    # - cleanup?
+
+    @property
+    def code(self) -> str:
+        func_src = inspect.getsource(self.func)
+        func_def = ast.parse(func_src).body[0]
+        assert isinstance(func_def, ast.FunctionDef)
+        return '\n'.join(textwrap.dedent(ast.unparse(stmt)) for stmt in func_def.body)
```
benhoyt: Per voice discussion, I think everything above here in this file is too clever (and hard for the reader to follow the thread of what's going on), and we should instead use simple tests and static test cases.

test/integration/secrets_test_cases.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def add_secret(self: TestSecretsCharm, rv: dict[str, Any]):
+    """Add secret with content."""
+    secret: ops.Secret = self.app.add_secret({'foo': 'bar'})
```
benhoyt: It seems to this entire complicated setup exists to reuse this one line of a code, a simple `add_secret` call. I know we may add test cases in future, but I think we should revert to a KISS approach, with completely separate tests in `test_e2e` (unit) and `test_secrets` (integration), with no shared code. If later on it gets to be a *lot* of duplication we can evaluate the best way to share.

benhoyt (changes requested):
Per voice discussion (and comments below), let's aim for a much simpler approach with static tests in their regular place. If in future we end up having a lot of duplication, we can figure out the best way to share the common code.

dimaqq:
Alternative PR: https://github.com/dimaqq/operator/pull/74

.github/workflows/integration.yaml
dimaqq: Today, I'm aware of three version ranges that we must test:
- pre-3.5 (possibly any between 3.1 and 3.4)
- 3.5.5+
- 3.6.0+

I'm thinking to test on a wider variety of Jujus until we've got good secrets API coverage.
At the point, individual tests and their caveats (e.g. `juju_version > 3.3.7: assert foo`) would allows us to select a more precise set of Juju versions.
WDYT?

.github/workflows/integration.yaml
dimaqq: Note that today generic integration tests must be listed in `jobs.test.strategy.matrix.test: ['test_foo', ...]`.

These new tests are not listed there.

test/integration/secrets_test_cases.py
dimaqq: I've added another test; also reverted to a simpler approach.

test/integration/secrets_test_cases.py
dimaqq: done

test/charms/test_secrets/.gitignore
```diff
@@ -0,0 +1,3 @@
+requirements.txt
```
tonyandrewmeyer, line 1: Seems strange that we have top-level ignores here and also individual ones, and that overlap. But we do already have this pattern. Added this to discuss at daily tomorrow.

pyproject.toml
```diff
@@ -203,6 +203,9 @@ ignore = [
     # Manual dict comprehension.
     "PERF403",
 
+    # Quoted types in casts.
+    "TC006",
+
```
tonyandrewmeyer: I'm not an expert on this, but I thought we wanted to have the quotes (and did that deliberately earlier this year). Why are we adding it to the ignore list?

test/integration/test_secrets.py
```diff
@@ -0,0 +1,139 @@
...
+    if secret:
+        assert secret.owner == 'test-secrets'
+        short_juju_id = short_id(secret.uri)
+        assert result['secretid']
+        short_ops_id = short_id(result['secretid'])
+        short_ops_info_id = short_id(result['after']['info']['id'])
+        assert short_juju_id == short_ops_id == short_ops_info_id
+        assert secret.content == result['after']['latest']
+
+    assert secret
```
tonyandrewmeyer: Nit: seems like this could be before the "if secret" and then the if statement could go away.

test/integration/test_secrets.py
```diff
@@ -0,0 +1,139 @@
...
+    assert secret.revision == 1
+    assert secret.content == {'foo': 'bar'}
```
tonyandrewmeyer, line 41: Nit: couldn't this be on line 53, `assert secret.content == result['after']['latest'] == {'foo: 'bar'}`?

test/charms/test_secrets/src/charm.py
```diff
@@ -0,0 +1,104 @@
...
+            'secretid': secret.id,
+            'after': self._snapshot(secret.id),
+        }
+        event.set_results(cast(dict[str, Any], result))
+
+    def add_with_meta(self, event: ops.ActionEvent):
+        field_keys = event.params['fields'].split(',')
+        full_meta = {
+            'label': 'label',
+            'description': 'description',
```
tonyandrewmeyer: This is maybe just my paranoia, but I'm always a bit wary of having values that are the same as the keys, in case there's some weird bug somewhere where you get the label back as the value. Feel free to ignore, but what about `lbl` and `desc` or something?

test/integration/test_secrets.py
```diff
@@ -0,0 +1,139 @@
...
+        assert info['description'] == 'description'
+    if 'expire' in fields:
+        assert secret.expires == '2020-01-01T00:00:00Z'
+        assert info['expires'] == '2020-01-01 00:00:00+00:00'
+    if 'rotate' in fields:
+        assert secret.rotation == 'daily'
+        assert info['rotation'] == 'SecretRotate.DAILY'
+        assert secret.rotates  # approx 24h from now
+        # It seems that Juju coerces the result values to strings.
+        assert info['rotates'] == 'None'
```
tonyandrewmeyer, line 90: I think it'd be worth having a comment here that explains why this is None rather than approx 24 hours from now (that ops doesn't read it or Juju doesn't provide it, whichever is the case).

test/integration/test_secrets.py
```diff
@@ -0,0 +1,139 @@
...
+            juju.remove_secret(secret.uri)
+
+
+@pytest.fixture
+def leader(juju: jubilant.Juju) -> str:
+    status = juju.status()
+    for name, unit in status.apps['test-secrets'].units.items():
+        if unit.leader:
+            return name
+    raise Exception(f'no leader in {status}')
```
tonyandrewmeyer, line 110: Does this mean that if things run really quickly (super unlikely, for sure) then the tests might fail? Would it be better to have a short sleep and retry and then if that fails to raise?

test/integration/test_secrets.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@pytest.fixture
+def follower(juju: jubilant.Juju) -> str:
```
tonyandrewmeyer, line 114: Is there here because you expect to use it in later tests? I don't see it used at the moment, unless I'm missing it.

testing/tests/test_e2e/test_secrets.py
```diff
@@ -590,3 +593,96 @@ def test_default_values():
     assert secret.rotate is None
     assert secret.expire is None
     assert secret.remote_grants == {}
+
+
+def test_add_secret(secrets_context: Context[CharmBase]):
```
tonyandrewmeyer: Nit: presumably this could be `Context[SecretsCharm]`.

testing/tests/test_e2e/test_secrets.py
```diff
@@ -590,3 +593,96 @@ def test_default_values():
...
+    state = State(leader=True)
+    state = secrets_context.run(secrets_context.on.action('add-secret'), state)
+
+    result = cast(Result, secrets_context.action_results)
+    assert result is not None
+    charm_secret_id = result.get('secretid')
+    scenario_secret = next(iter(state.secrets)) if state.secrets else None
+
+    assert bool(scenario_secret) == bool(charm_secret_id)
```
tonyandrewmeyer: What's this testing? Won't `bool()` on a `testing.Secret()` always be `True`?

testing/tests/test_e2e/test_secrets.py
```diff
@@ -590,3 +593,96 @@ def test_default_values():
...
+        assert scenario_secret.description == 'description'
+        # https://github.com/canonical/operator/issues/2037
+        # assert info['description'] == 'description'
+    if 'expire' in fields:
+        assert scenario_secret.expire == datetime.datetime(2020, 1, 1, 0, 0, 0)
+        assert info['expires'] == datetime.datetime(2020, 1, 1, 0, 0, 0)
+    if 'rotate' in fields:
+        assert scenario_secret.rotate == SecretRotate.DAILY
+        assert info['rotation'] == SecretRotate.DAILY
+        # FIXME is this a Scenario limitation?
```
tonyandrewmeyer: If it's not in Ops, then presumably it shouldn't be in Scenario. But I wonder if it ought to be in both?

testing/tests/test_e2e/test_secrets.py
```diff
@@ -590,3 +593,96 @@ def test_default_values():
...
+        assert info['rotates'] is None
+
+
+def common_assertions(scenario_secret: Secret | None, result: Result):
+    if scenario_secret:
+        assert scenario_secret
```
tonyandrewmeyer: It doesn't seem to make sense to have both of these statements. If `scenario_secret` can be false-y then let's switch it around and have an early exit in that case so we can dedent the rest of the code, and then we don't need the assert.

test/charms/test_secrets/.gitignore
dimaqq: Let's clean that up in a separate PR.

pyproject.toml
dimaqq: I'm happy to revert this bit and bring this up for discussion.

I can't recall our last decision, I was just surprised that we've dropped py3.8 but our tooling was configured to use quotes in casts specifically.

The rationale for this rule is a fun read: https://docs.astral.sh/ruff/rules/runtime-cast-value/

TL;DR: some types may be conditionally imported `if TYPE_CHECKING: ...`, and those must be quoted, so let's force-quote everywhere for consistency.

(hobgoblins come to mind)

testing/tests/test_e2e/test_secrets.py
dimaqq: Juju would compute `rotates = new + 24h` in this case (freshly created secret).
Scenario doesn't.
I suspect because Scenario in general avoid time-dependent stuff.
I'll open a separate issue for future work and link it here.

testing/tests/test_e2e/test_secrets.py
dimaqq: It's testing that either both are set or both are reset.

dimaqq: I suppose that in this particular case, I know that both ought to be set 🤔

testing/tests/test_e2e/test_secrets.py
dimaqq: https://github.com/canonical/operator/issues/2104

test/integration/test_secrets.py
dimaqq: Yes, I'm expecting to use it later.

test/integration/test_secrets.py
dimaqq: Thanks, the value is not visible in the same dispatch call. It will be later.
I've added a comment.

test/integration/test_secrets.py
dimaqq: No, because `test_setup` above includes `juju.wait(jubilant.all_active)`.

No leader is possible when the app is freshly deployed (e.g. lxd starting up).
Here, the app has been deployed, and both units' stati had been OK for Jubilant's default of three status checks with one second delays in between.

test/charms/test_secrets/src/charm.py
dimaqq: Done.

test/integration/test_secrets.py
dimaqq: Thanks for pointing this out, I was supposed to have "common_checks" like in unit tests.

pyproject.toml
dimaqq: done

testing/tests/test_e2e/test_secrets.py
dimaqq: done

Issue #2072, opened 2025-09-30
docs: How to test a small part of a charm


yanksyoon:
*Please describe the question or issue you're facing with "How to write unit tests for a charm - Ops documentation".*

Hello team!

When writing a charm that has multiple components, we segment the component in the code by a paradigm called an "Observer".

For example, I would have a small component that manages storage of a charm. This component receives the parent charm in the init function and provides the parent charm scoped functions that are relevant to storage.

```
import ops

class MyStorageOperator(ops.Object):
  def __init__(self, charm: ops.CharmBase):
    super().__init__(charm, "my-storage-operator")

  def is_storage_ready(self):
    return "my-storage" in self.model.storages
```

Before with Harness testing, I was able to inject the charm from the harness to the `MyStorageOperator`, i.e. `MyStorageOperator(harness.charm)`. Then, I could scope the unit tests limited to the storage functionality.

```
def _parametrize_my_storage_operator():
  ...


@pytest.mark.parametrize("charm, expected", _parametrize_my_storage_operator())
def test_is_storage_ready(charm: ops.CharmBase, expected: bool):
  storage = MyStorageOperator(charm)

  assert storage.is_storage_ready() == expected
```

Now, I'm not sure how I could achieve the same test (with scoped responsibilities of a charm component) using ops.testing framework - given that it initializes the `ops.main` to test the charm and no longer exposes the charm with testing framework backend.

Would you have any guidance on it?

Thank you :)

---
*Reported from: https://documentation.ubuntu.com/ops/latest/howto/write-unit-tests-for-a-charm/*

yanksyoon:
From Matrix: https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$zryvHXvHEQbhwSQFLxcvRt5jvXzPlvHAiTaBp6q1GNA?via=ubuntu.com&via=matrix.org&via=xentonix.net
```
with ctx(ctx.on.update_status(), state_in) as mgr:
    result = mgr.charm.holistic_method()
    assert result == ...
```

benhoyt:
@yanksyoon Do you need anything further here?

@dwilding Do you see a good place to give an example like this in our docs? Potentially in the Harness migration guide you're writing, or maybe in a small section at the bottom of https://documentation.ubuntu.com/ops/latest/howto/write-unit-tests-for-a-charm/?

yanksyoon:
@benhoyt This can be resolved, I'll contribute to the said section above, thank you!

PR #2084, opened 2025-10-02
docs: extract charm from testing context

yanksyoon:
Addresses #2072

Adds an example for extracting charm from testing context.

### DIFF
--- a/docs/howto/write-unit-tests-for-a-charm.md
+++ b/docs/howto/write-unit-tests-for-a-charm.md
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
 
 If you use pytest, you should put the `my_charm` fixture in a top level `conftest.py`, as it will likely be shared between all your unit tests.
 ```
+
+## Test a function that receives charm as an argument
+
+A function might take in charm as an argument to extract charm state or charm relation information. To test such functions, you can use
+the following method to test the function.
+
+
+```python
+# Example function to test that takes the charm as an argument
+def get_num_relation(my_charm):
+    return len(self.model.relations.get("my-relation", []))
+
+# Test for the function that takes the charm as an argument
+def test_get_num_relation():
+    ctx = testing.Context(my_charm)
+    state_in = testing.State(relations=[])
+    with ctx(ctx.on.update_status(), state_in) as mgr: # use any hook event here
+        result = takes_charm_as_argument(my_charm=mgr.charm)
+        assert result == 0
+```


docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+## Test a function that receives charm as an argument
+
+A function might take in charm as an argument to extract charm state or charm relation information. To test such functions, you can use
+the following method to test the function.
+
+
+```python
+# Example function to test that takes the charm as an argument
+def get_num_relation(my_charm):
+    return len(self.model.relations.get("my-relation", []))
```
dimaqq: Let's come up with an example where the `my_charm` argument is actually used.

It could be a real example, or maybe an abstract one, like `foo(my_charm.DEFAULT_PORT)`.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+
+# Test for the function that takes the charm as an argument
+def test_get_num_relation():
+    ctx = testing.Context(my_charm)
+    state_in = testing.State(relations=[])
+    with ctx(ctx.on.update_status(), state_in) as mgr: # use any hook event here
+        result = takes_charm_as_argument(my_charm=mgr.charm)
```
dimaqq: Here, it's unclear why the reference to the charm class is being gotten from `mgr.charm` when it's available as `my_charm` at the start of test function body.

How about this example is split into two functions:
- a pytest.fixture that internally uses the charm type and returns a Context ready to be used
- a test function that only receives the Context and thus has a reason to look up the charm type through the Context or Manager.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+
+## Test a function that receives charm as an argument
+
+A function might take in charm as an argument to extract charm state or charm relation information. To test such functions, you can use
+the following method to test the function.
+
+
+```python
+# Example function to test that takes the charm as an argument
+def get_num_relation(my_charm):
```
dimaqq:
```suggestion
def get_num_relation(my_charm: type[MyCharm]):
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+    return len(self.model.relations.get("my-relation", []))
+
+# Test for the function that takes the charm as an argument
+def test_get_num_relation():
```
dimaqq:
```suggestion
@pytest.fixture
def context(...) -> ops.testing.Context:
    ...
    return Context(MyCharm, ...)

def test_get_num_relation(context):
```

dimaqq:
Adding Ben for:
- quick take whether this recipe belongs in this specific doc page
- if so, to review or delegate to someone to review

docs/howto/write-unit-tests-for-a-charm.md
dimaqq: I guess I have misunderstood this.

`Manager.charm` is a property that returns the charm object (instance), not the charm class.

Which also means we'd have to document the exact "point in time" or charm object state that the test would find `mgr.charm` in.

docs/howto/write-unit-tests-for-a-charm.md
dimaqq:
```suggestion
def get_relations_count(charm_instance: MyCharm):
    return len(charm_instance.model.relations.get("my-relation", []))
```

docs/howto/write-unit-tests-for-a-charm.md
dimaqq: Maybe this is not actually relevant.

docs/howto/write-unit-tests-for-a-charm.md
yanksyoon: > Which also means we'd have to document the exact "point in time" or charm object state that the test would find mgr.charm in.

That would be great - the purpose of this recipe is to just take the "state" of the charm (integrations, storages, configurations, ...etc) and see if a smaller handler for a particular "state" is able to pick up the state as intended and is able to operate on it.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
 
 If you use pytest, you should put the `my_charm` fixture in a top level `conftest.py`, as it will likely be shared between all your unit tests.
 ```
+
+## Test a function that receives charm as an argument
```
benhoyt: I'd shorten to:
```suggestion
## Test a function that takes a charm instance
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+
+A function might take in charm as an argument to extract charm state or charm relation information. To test such functions, you can use
```
benhoyt:
```suggestion
A function might take the charm as a parameter to extract charm state or charm relation information. To test such functions, you can use
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+the following method to test the function.
```
benhoyt:
```suggestion
the `testing.Context` instance as a context manager, and then access `mgr.charm`. You should generally use an event the charm doesn't observe, for example, `update_status`:
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+
+
+```python
+# Example function to test that takes the charm as an argument
+def get_relations_count(charm_instance: MyCharm):
+    return len(charm_instance.model.relations.get("my-relation", []))
```
benhoyt: I don't see a good reason for the `_instance` suffix. How about:
```suggestion
def get_relations_count(charm: MyCharm):
    return len(charm.model.relations.get("my-relation", []))
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+
+# Test for the function that takes the charm as an argument
+def test_get_num_relation():
```
benhoyt: Consistency with naming above:
```suggestion
def test_get_relations_count():
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,23 @@ def test_charm_runs(my_charm):
...
+    ctx = testing.Context(my_charm)
+    state_in = testing.State(relations=[])
+    with ctx(ctx.on.update_status(), state_in) as mgr: # use any hook event here
+        result = takes_charm_as_argument(my_charm=mgr.charm)
```
benhoyt: Shouldn't this be `get_relations_count` to match the above? Also let's drop the keyword arg (it's wrong in any case):
```suggestion
        result = get_relations_count(mgr.charm)
```

benhoyt (commented):
I think this is a reasonable place for this (though let's wait till @dwilding is back to check and finalise). However, I've added some minor comments for style and consistency.

docs/howto/write-unit-tests-for-a-charm.md
yanksyoon: Thanks for the feedback, reflected all the comments!

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -161,3 +161,24 @@ def test_charm_runs(my_charm):
...
+def get_relations_count(charm: MyCharm):
+    return len(charm.model.relations.get("my-relation", []))
+
+# Test for the function that takes the charm as an argument
+def test_get_relations_count():
+    ctx = testing.Context(my_charm)
+    state_in = testing.State(relations=[])
+    with ctx(ctx.on.update_status(), state_in) as mgr: # use any hook event here
+        result = get_relations_count(mgr.charm)
+        assert result == 0
```
dimaqq: **minor**: while technically correct, the reader may ask whether this snippet tests the state transition testing library itself or the charm.

I wonder how to express this clearer.

yanksyoon: I see, how about I separate out the charm from the context using a fixture like so?
```
@pytest.fixture(name="charm")
def charm_fixture():
  """The charm from the update_status context."""
  with ctx(ctx.on.update_status(), state_in) as mgr: 
    yield mgr.charm
```

However, this means that the `state_in` is dependent on the fixture itself. In this scenario, we could write a helper function that extracts the charm like so:
```
def _generate_charm_from_context(state):
  with ctx(ctx.on.update_status(), state) as mgr:
    yield mgr.charm
```

WDYT?

dwilding:
Preview of new section in the docs: https://canonical-ubuntu-documentation-library--2084.com.readthedocs.build/ops/2084/howto/write-unit-tests-for-a-charm/#accessing-the-charm-instance

yanksyoon:
@dwilding Thank you for the changes, LGTM :)

PR #2087, opened 2025-10-03
fix: move `_FilesParser.remove_files()` call to `finally` block

james-garner-canonical:


dimaqq:
So, now the function returns a Python `file` object that's connected to a file descriptor that's unlinked in the file system.

Did I get that right?

That should work fine on Linux and macOS, but not Windows, IIRC.
How lucky that Windows support was dropped ages ago (technically Juju 3.0.0).

I suppose it's a solid change. How is it tested?

dimaqq:
> If we do end up preferring the context manager route, ...

Having looked at the code again, I figure there's no point.

If we wanted to rehash this a bit more, we'd propagate the pulled path into the parser, and skip the content for any other files (or in fact raise, because there shouldn't be other files). I'm happy to leave this idea for a few years :P

james-garner-canonical:
> So, now the function returns a Python file object that's connected to a file descriptor that's unlinked in the file system.

I believe that's what it returned originally too, since `parser.remove_files()` was called before returning.

> I suppose it's a solid change. How is it tested?

I haven't added tests that the temporary files are cleaned up if there's an exception in the `try ... finally`. Presumably this wasn't considered worth testing for originally, which turned out to be a mistake, but surely *this* time we've got it right xD.

Happy to look into adding some tests.

At the very least I should test this manually before merging by hardcoding the temporary file path.

dimaqq:
> At the very least I should test this manually before merging by hardcoding the temporary file path.

Or by pulling a large file in the real Pebble tests. WDYT?

benhoyt (approved):
Looks good to me, thanks. Yes, I agree about doing a manual test, and Dima's suggestion about adding a test to the "real Pebble" tests is a good one.

dimaqq (approved):

james-garner-canonical:
I've added a test, would be good to get at least one re-review before merging (@benhoyt, @dimaqq).

test/test_real_pebble.py
```diff
@@ -362,6 +363,20 @@ def pebble_dir(self):
         yield pebble_dir
         shutil.rmtree(pebble_dir)
 
+    def test_temp_files_cleaned_up_on_failed_pull(
```
benhoyt, line 352: Should this live under `TestRealPebble` rather than `TestPebbleStorageAPIsUsingRealPebble`?

test/test_real_pebble.py
```diff
@@ -362,6 +363,20 @@ def pebble_dir(self):
...
+        self,
+        pebble_dir: str,
+        client: pebble.Client,
+        monkeypatch: pytest.MonkeyPatch,
+    ):
+        client.push(f'{pebble_dir}/test', os.urandom(1024 * 1024))  # chunk size is 16 * 2014
+        tf = tempfile.NamedTemporaryFile(delete=False)  # noqa: SIM115
+        monkeypatch.setattr(pebble._FilesParser, 'get_response', lambda self: None)  # type: ignore
```
benhoyt, line 361: Perhaps add a comment like `# Patch get_response to force a ProtocolError exception`.

test/test_real_pebble.py
```diff
@@ -362,6 +363,20 @@ def pebble_dir(self):
...
+        monkeypatch.setattr(tempfile, 'NamedTemporaryFile', lambda *args, **kwargs: tf)  # type: ignore
```
benhoyt, line 363: I don't quite follow why `NamedTemporaryFile` needs to be patched. Perhaps a comment on that one too?

test/test_real_pebble.py
james-garner-canonical: The test is based off those in the storage API tests and fits in pretty nicely with them, but can't go in the mixin since it will only work with real pebble (due to mocking stuff specific to the real pebble client that the mock client doesn't implement). IMO either place would be fine, so I'll go with your suggestion.

test/test_real_pebble.py
james-garner-canonical:
```suggestion
        # Patch get_response to force a ProtocolError exception.
        monkeypatch.setattr(pebble._FilesParser, 'get_response', lambda self: None)  # type: ignore
```

test/test_real_pebble.py
james-garner-canonical:
```suggestion
        # Use our previously created temp dir so we can verify that it gets cleaned up.
        monkeypatch.setattr(tempfile, 'NamedTemporaryFile', lambda *args, **kwargs: tf)  # type: ignore
```

PR #2092, opened 2025-10-08
ci: add pytest.warns to note an expected warning

tonyandrewmeyer:
Doing `stop_check` triggers a security event log, but in the Harness tests we haven't fully configured logging, which means that a `RuntimeWarning` is triggered. Use `pytest.warns` in the test to avoid this polluting the test output (and also to ensure that the warning does occur).

In addition, there's some extra cleanup of the logging. The most significant of these is in Scenario - when the manager finishes, it removes the JujuHandlers that were added when it was created. This seems generally cleaner (as well as avoiding the test inconsistency) and means that adding in the next test run is cleaner too.

### DIFF
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -51,7 +51,7 @@ def backend():
 @pytest.fixture()
 def logger():
     logger = logging.getLogger()
-    orig_showwarning = warnings.showwarning
+    orig_showwarning = warnings.showwarning  # Modified when setting up logging in ops.log.
     yield logger
     logging.getLogger().handlers.clear()
     sys.excepthook = sys.__excepthook__
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -87,6 +87,7 @@ def root_logging():
 class TestModel:
     @pytest.fixture
     def harness(self):
+        # Clear app ID cached when logging security events.
         _get_juju_log_and_app_id.cache_clear()
         harness = ops.testing.Harness(
             ops.CharmBase,
@@ -120,6 +121,7 @@ class TestModel:
         """,
         )
         yield harness
+        # Clear app ID cached when logging security events.
         _get_juju_log_and_app_id.cache_clear()
         harness.cleanup()
 
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -7137,7 +7137,7 @@ class TestChecks:
     def test_stop_checks(self, request: pytest.FixtureRequest):
         container = self._container_with_layer(request)
         # This generates a warning because there's a security event
-        # logged, but we haven't set up logging properly.
+        # logged, but we haven't set up logging for Harness in these tests.
         with pytest.warns(RuntimeWarning):
             changed = container.stop_checks('chk1', 'chk2', 'chk3')
         assert changed == ['chk1', 'chk2']
--- a/testing/src/scenario/_ops_main_mock.py
+++ b/testing/src/scenario/_ops_main_mock.py
@@ -279,6 +279,7 @@ class Ops(_Manager, Generic[CharmType]):
         for handler in logger.handlers[:]:
             if isinstance(handler, JujuLogHandler):
                 logger.removeHandler(handler)
+        # Clear app ID cached when logging security events.
         _get_juju_log_and_app_id.cache_clear()
 
 


dimaqq (approved):

dimaqq:
`Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.`

On some version combinations in the test matrix.

james-garner-canonical (approved):
Looks good to me, I assume that setting up logging properly in the container would be more work and lines of code to read than is worthwhile here.

dimaqq:
Screenshot 2025-10-08 at 10 44 18

Screenshot 2025-10-08 at 10 46 23



this is weird... I wonder if the test is flaky.

tonyandrewmeyer:
> Screenshot 2025-10-08 at 10 44 18
> this is weird... I wonder if the test is flaky.

Yeah, very odd, it should always be triggered because we aren't setting up logging and it always results in a warning. Maybe there's a test that leaks setting up logging. I'll look into it.

tonyandrewmeyer:
> Looks good to me, I assume that setting up logging properly in the container would be more work and lines of code to read than is worthwhile here.

Yeah, Harness doesn't do the `setup_root_logging()` call and I don't think we want to introduce that at this point. I suppose this specific test could do that, but I think the `pytest.warns()` is just as useful.

benhoyt:
@tonyandrewmeyer to pick this up at some point

tonyandrewmeyer:
This reliably passed for me now, but I assume the macOS failures mean there is still an issue.

tonyandrewmeyer:
@dimaqq, @james-garner-canonical this is ready for review again, I believe. Thanks!

test/test_log.py
```diff
@@ -50,9 +51,11 @@ def backend():
 @pytest.fixture()
 def logger():
     logger = logging.getLogger()
+    orig_showwarning = warnings.showwarning
```
james-garner-canonical:
```suggestion
    orig_showwarning = warnings.showwarning  # Modified when setting up logging in ops.log.
```

test/test_model.py
```diff
@@ -87,6 +87,7 @@ def root_logging():
 class TestModel:
     @pytest.fixture
     def harness(self):
+        _get_juju_log_and_app_id.cache_clear()
```
james-garner-canonical, line 91:
```suggestion
        # Clear app ID cached when logging security events.
        _get_juju_log_and_app_id.cache_clear()
```

test/test_model.py
```diff
@@ -119,6 +120,7 @@ def harness(self):
         """,
         )
         yield harness
+        _get_juju_log_and_app_id.cache_clear()
```
james-garner-canonical, line 125:
```suggestion
        # Clear app ID cached when logging security events.
        _get_juju_log_and_app_id.cache_clear()
```

test/test_testing.py
```diff
@@ -7136,7 +7136,10 @@ def test_start_checks(self, request: pytest.FixtureRequest):
 
     def test_stop_checks(self, request: pytest.FixtureRequest):
         container = self._container_with_layer(request)
-        changed = container.stop_checks('chk1', 'chk2', 'chk3')
+        # This generates a warning because there's a security event
+        # logged, but we haven't set up logging properly.
```
james-garner-canonical:
```suggestion
        # logged, but we haven't set up logging for Harness in these tests.
```

testing/src/scenario/_ops_main_mock.py
```diff
@@ -273,6 +274,13 @@ def destroy(self):
             self.trace_data = self._tracing_exporter.get_finished_spans()
             self._tracing_mock.__exit__(None, None, None)
 
+        # Clean up logging handlers to avoid test pollution.
+        logger = logging.getLogger()
+        for handler in logger.handlers[:]:
+            if isinstance(handler, JujuLogHandler):
+                logger.removeHandler(handler)
+        _get_juju_log_and_app_id.cache_clear()
```
james-garner-canonical, line 283:
```suggestion
        # Clear app ID cached when logging security events.
        _get_juju_log_and_app_id.cache_clear()
```

PR #2096, opened 2025-10-09
fix: `get_deferred_events` breaks if `use_juju_for_storage=True`

PietroPasotti:
I'm using `_ops_main_mock` in `jhack` and I found out that, if `use_juju_for_storage` is `True`, the `_db` attribute holds a variable of a different type.

This updates `_ops_main_mock` to support both types. 
I'd appreciate some pointers on: 
- how to (regression) test this
- how to fix the similar issue with `get_stored_states`, because `list_snapshots` is undefined on `JujuStorage`

This probably isn't a bug, because `_ops_main_mock` is only used when unittesting, and when unittesting we're never 'using juju for storage', that's simply not applicable. But it limits its usability by external scripts. But it's a private module so you could also say "meh". 
I can patch it on my side so it's not a blocker for me.

### DIFF
--- a/testing/src/scenario/_ops_main_mock.py
+++ b/testing/src/scenario/_ops_main_mock.py
@@ -15,8 +15,7 @@ from typing import (
     Any,
     Generic,
     Sequence,
-    cast,
-    Union,
+    cast
 )
 
 import ops


testing/src/scenario/_ops_main_mock.py
```diff
@@ -15,7 +15,7 @@
     Any,
     Generic,
     Sequence,
-    cast,
+    cast
```
dimaqq, line 18: This change isn't right.

```suggestion
    cast,
```

Please run `tox -e format` and `tox -e lint`.

dimaqq:
> I'm using `_ops_main_mock` in `jhack`...

In production or in unit tests for jhack?

dimaqq:
Pointer for someone trying to find the call site: 
https://github.com/canonical/jhack/blob/99b53ed86b11257a2d807bfe55bdeb612ce4992d/jhack/scenario/snapshot.py#L966

testing/src/scenario/_ops_main_mock.py
```diff
@@ -78,22 +78,20 @@ def get_deferred_events(self) -> list[DeferredEvent]:
...
-                        snapshot_data: dict[str, Any] = {}
-
-                    event = DeferredEvent(
-                        handle_path=handle,
-                        owner=owner,
-                        observer=observer,
-                        snapshot_data=snapshot_data,
-                    )
-                    deferred.append(event)
+        for handle, owner, observer in sorted(db.notices(), key=lambda n: n[0]):
```
dimaqq, line 81: I'm rather confused by this change. Snapshots and notices are different tables, though related. I'll leave this to Tony to investigate.

Sorting snapshots/notices/deferred events by something doesn't seem right to me.
If I recall correctly, there are very specific semantics for the order of the deferred events.

This particular function is only called once in Scenario where the "state_out" is stamped with a list of deferred events. Which matters in two cases:
- assertions in tests that examine deferred events after running a real event
- charm code in tests that pass state around to run multiple real events

tonyandrewmeyer:
> * how to fix the similar issue with `get_stored_states`, because `list_snapshots` is undefined on `JujuStorage`

This is only used by `Framework.remove_unreferenced_events()`, which is public but not used by Ops itself, or any charms I can find. I'm guessing that it was meant to be used by charms to do a one-off clean-up in upgrade or something like that. I don't think we'd put something like this in Ops itself these days. However, it's public so we're stuck with it.

In theory, the right thing to do would be to add a `list_snapshots` method to `JujuStorage`, which would mean adding a `list` method to `_JujuStorageBackend`, but there's no `state-list` hook command. `state-get` with no args gives all the keys and values, but we'd then have to make assumptions about which ones are relevant here.

So, I think the most appropriate thing to do here would be to adjust `remove_unreferenced_events()` to fail loudly if the storage doesn't have a `list_snapshots` method.

benhoyt:
Per team discussion, we'd rather not do this in Ops:

> I can patch it on my side so it's not a blocker for me.

PR #2097, opened 2025-10-10
feat: make PebbleClient file methods also accept pathlib.PurePath

james-garner-canonical:
I got tripped up while working on #2087 with methods like `ops.pebble.PebbleClient.push` only accepting a `str` for their `path` argument, while the wrapper methods like `ops.Container.push` accept `str | pathlib.PurePath`. Nowadays pretty much everything you're likely to run into in the `stdlib` that operates on file paths will accept path-like objects, so it seems nice to have the same consistency here.

- This PR updates the `PebbleClient` methods to also accept `pathlib.PurePath`, immediately converting the argument to `str` to avoid the signature change affecting their implementation.
- The equivalent methods on `_TestingPebbleClient` have been updated the same way.
- The `ops.Container` calls to these methods have been updated to not convert the `path` argument to `str` since it will be converted after calling anyway.


ops/_private/harness.py
```diff
@@ -3532,7 +3532,10 @@ def _check_absolute_path(path: str):
         if not path.startswith('/'):
             raise pebble.PathError('generic-file-error', f'paths must be absolute, got {path!r}')
 
-    def pull(self, path: str, *, encoding: str | None = 'utf-8') -> BinaryIO | TextIO:
+    def pull(
+        self, path: str | pathlib.PurePath, *, encoding: str | None = 'utf-8'
+    ) -> BinaryIO | TextIO:
+        path = str(path)
```
benhoyt, line 3538: I think this type and implementation makes sense given the existing types.

However, I'm just curious: for new code, do you know if using the type `os.PathLike` and `path = os.fspath(path)` be the more general approach?

james-garner-canonical: Yes, that would be the 'right' way to do it, though annoyingly `os.PathLike` is generic because `__fspath__` can return `str` or `bytes`, so we'd probably want to use `os.PathLike[str]`.

Also for new code I'd tend towards using `path = pathlib.Path(path)` internally where possible.

james-garner-canonical:
>  We should probably have unit tests that exercise this with a pathlib.Path instance, at least for the pebble.py methods.

EDIT: I guess this opening paragraph is wrong, since the `test_pebble` tests which skip the Harness backend use `str` arguments.

~These actually get exercised (as much as they were previously with strings) with paths already now, due to how the change is implemented.~ Previously the `ops.Container` methods were tested with `pathlib.Path`s, which they converted into strings and passed to the `pebble` methods. Now they pass them without converting them, since they're converted internally.

However, the [current tests exercising this](https://github.com/canonical/operator/blob/8e355c0c6db07e9dc4e509d7d6394834ddeb3919/test/test_model.py#L1202) seem to be Harness based, so they're using Harness's mock pebble client instead of the real one. 

Would it be worth perhaps parametrizing (or just duplicating) a number of the `test_pebble` tests to use `pathlib.Path` arguments instead of `str`?

ops/_private/harness.py
```diff
@@ -3532,7 +3532,10 @@ def _check_absolute_path(path: str):
         if not path.startswith('/'):
             raise pebble.PathError('generic-file-error', f'paths must be absolute, got {path!r}')
 
-    def pull(self, path: str, *, encoding: str | None = 'utf-8') -> BinaryIO | TextIO:
+    def pull(
+        self, path: str | pathlib.PurePath, *, encoding: str | None = 'utf-8'
```
tonyandrewmeyer, line 3536: Given that Ops, Charmcraft, and Juju don't support Windows, would it be better for this to be `PurePosixPath`? (And throughout the PR, obviously). Given that we're not using `os.PathLike[str]`.

ops/_private/harness.py
tonyandrewmeyer: > Also for new code I'd tend towards using `path = pathlib.Path(path)` internally where possible.

In that case, would it not be better to make that change here too? It's more invasive, but I doubt significantly so, and if it's the better choice, wouldn't we rather do that now? It seems unlikely it would happen in the future.

ops/_private/harness.py
james-garner-canonical: That's an interesting question. Using `str` to convert the path *does* assume that it's a `PosixPath` to get sensible results (as does calling `os.fspath` for that matter). The downside of `PurePosixPath` is that the readability is worse (imo) -- and it only helps prevent errors for people who explicitly pass a `WindowsPath` (why?), or who are running this on Windows, which we already don't support. So I'd argue for keeping the slightly shorter and more readable annotation.

ops/_private/harness.py
james-garner-canonical: For these `pebble` methods specifically, I don't think there's a benefit in working with `Path` objects for the arguments, since they're typically immediately just used as values in a `dict` for a request.

james-garner-canonical:
> For consistency, do you think exec's working_dir should also have this change?

This would be more consistent overall. The reason I didn't cover this parameter is because `ops.Container.exec` doesn't accept a `Path` here, only `str | None`.

I think it would be fine to merge this PR without adding this additional feature. However, since we won't be merging without adding additional tests, it shouldn't be much work to also add `Path` support to `working_dir`, so I'll do that.

`working_dir` being `None`-able makes me think  we should use `os.fspath` instead of `str` for conversion, since that will ensure we handle `None` correctly (instead of converting it to `'None'`).

james-garner-canonical:
I've opted not to bundle switching from `str` to `os.fspath` with this PR.

It's technically not backwards compatible.

Previously if someone used `None` or some other strange object as the `path` argument to an `ops.Container` method, it would get stringified and probably just work, writing to some strangely named file. It would also work (though it probably shouldn't have) with `pathops.ContainerPath` objects, which have a path-like `str`, but are intentionally not `os.PathLike`.

If we want to switch, I think it should be a separate PR, so it can be more easily reverted if it causes trouble.

PR #2083, opened 2025-10-02
docs: add holistic way of managing storage

yanksyoon:
Add documentation on managing storage in a holistic reconcile function.

Resolves: #2069

PR #2098, opened 2025-10-10
docs: fix and clarify holistic example of handling storage

dwilding:
This PR significantly reworks [How to manage storage](https://documentation.ubuntu.com/ops/latest/howto/manage-storage/). The current doc has several problems:

- It's not clear that `_update_configuration` is supposed to be an example of a holistic handler.

- The code incorrectly tries to access `self.model.storages["cache"]` as an object (it's a list).

- The doc mixes up machine and Kubernetes charms, which require different approaches because of (1) multiple vs singular storage and (2) how & whether to configure the workload with storage instance paths.

- The doc shows how to specify a location in a storage definition and explains the default location that Juju determines. This can mislead charmers into thinking that locations are predictable and could be hard-coded, when actually it's a better practice to transparently accept the locations that Juju determines.

Since this PR is substantial enough already, I've tried to leave the parts about detaching and tests largely unchanged.

**[Preview of updated doc](https://canonical-ubuntu-documentation-library--2098.com.readthedocs.build/ops/2098/howto/manage-storage/)**

See related discussion in https://github.com/canonical/operator/pull/2083

### DIFF
--- a/docs/howto/manage-storage.md
+++ b/docs/howto/manage-storage.md
@@ -162,7 +162,7 @@ To access the storage instance in charm code, use {external+charmlibs:ref}`patho
     (charm_cache_root / "processed-data").mkdir(exist_ok=True)
 ```
 
-Alternatively, use {external+charmlibs:class}`pathops.ContainerPath` to access `web_cache_path` in the workload container. This approach is more appropriate if you need to manipulate lots of data in the workload container.
+Alternatively, use {external+charmlibs:class}`pathops.ContainerPath` to access `web_cache_path` in the workload container. This approach is more appropriate if you need to reference additional data in the workload container.
 
 ## Handle storage detaching
 


docs/howto/manage-storage.md
```diff
@@ -90,13 +90,15 @@ Now, in the body of the charm definition, define the event handler, or adjust an
...
     cache = self.model.storages["cache"]
-    if cache.location is None:
-        # This must be one of the other events. Return and wait for the storage-attached event.
+    if not cache:
+        # The storage-attached event hasn't happened yet.
         logger.info("Storage is not yet ready.")
         return
     try:
-        self.push_configuration(cache_dir=cache.location)
+        self.push_configuration(cache_dir=cache[0].location)
```
tonyandrewmeyer: Unfortunately, we have more problems here than just the type of `cache`. Partly, this is because the behaviour differs between machine and K8s and the doc is mixing things up here.

In K8s, the `cache` value is going to be a single-item list of a *charm container* path. That path is a shared volume that is also mounted in the *workload container* (in the location specified in the `mounts` field of the container). So we shouldn't be using that location in a workload config, because the path in the workload container may be somewhere completely different. An `ops.Container` is mostly a wrapper around Pebble rather than Juju so it doesn't know anything about mount locations, so you'd need to get the location that the *workload* should use from the metadata. We could do this here but then it looks like it's going against what the text is telling you to do.

With machines, the `cache` value will be a list of one or more paths and everything is in the same filesystem so it's much more straightforward.

I think we should either make this example something suitable for a machine charm, or make more substantial changes so it makes sense for a K8s charm, or have both.

Happy to talk this over together if that's best.

dwilding:
Thanks @tonyandrewmeyer for reviewing and explaining where the doc is mixing things up. I worked through the doc more carefully and totally agree with you. Also, I find the structure of the doc quite confusing.

I've tried a partial refactoring of the doc, to make it easier to follow and clearly separate the machine and K8s info. There's more that really ought to be improved, but this is enough for one PR I think.

**[Preview doc](https://canonical-ubuntu-documentation-library--2098.com.readthedocs.build/ops/2098/howto/manage-storage/)**

New/reworked sections:

- Manage storage for a machine charm >
    - Define the storage
    - Access the storage
- Manage storage for a Kubernetes charm >
    - Define the storage
    - Access the storage

Other sections are largely untouched.

docs/howto/manage-storage.md
```diff
@@ -164,7 +96,107 @@ The storage will not be available immediately after that call - the charm should
...
+    cache_root = pathops.ContainerPath(
+        cache_in_container.location,
+        container=self.unit.get_container("web-service"),
+    )
+    try:
+        (cache_root / "data.json").write_text("...")
+    except ops.pebble.ConnectionError:
+        # Pebble isn't ready yet. Return and wait for the pebble-ready event.
+        logger.info("Pebble is not yet ready.")
+        return
```
dwilding: If the charm wants to write data to the storage, it can access the storage in the charm container. There's no point going through the workload container. So maybe we should drop this example and only have the (following) example of putting the storage location in a config file in the workload container.

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
+    type: filesystem
+    description: Somewhere to cache files locally.
+    location: /cache/
+    minimum-size: 1G
+    properties:
+      - transient
 ```
 
-For Kubernetes charms, you also need to define where on the workload container the volume will be mounted. For example, to mount a similar cache filesystem in `/var/cache/`:
+If you don't specify a location for the storage, Juju mounts the storage at the default location:
```
dimaqq: **minor:** my ears perked up, "for storage" or "for the storage"?

I think it comes down to "storage" being vague.
- Is it a unit of storage, like `name/0`? --> in plural, maybe storage units
- Is it a defined name, like `name`? --> with arcticle, maybe disambiguate
- Is it an abstract concept? --> without the article, which concept though...

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
-      type: filesystem
-      description: Somewhere to cache files locally.
-      location: /cache/
-      minimum-size: 1G
-      properties:
-          - transient
+  cache:
+    type: filesystem
+    description: Somewhere to cache files locally.
+    location: /cache/
```
dimaqq: Would that actually be `/cache` when mounted or `/cache/0`?

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
         return
-    try:
-        self.push_configuration(cache_dir=cache.location)
-    except ops.pebble.ConnectionError:
-        # Pebble isn't ready yet. Return and wait for the pebble-ready event.
-        logger.info("Pebble is not yet ready.")
-        return
+    # Storage is available. Write some data to the cache.
+    cache_root = pathops.LocalPath(cache[0].location)
+    (cache_root / "data.json").write_text("...")
```
dimaqq: Would be nice to show what "cache_root" may look like here.

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
 ```
 
-> Examples: [ZooKeeper ensuring that permission and ownership is correct](https://github.com/canonical/zookeeper-operator/blob/106f9c2cd9408a172b0e93f741d8c9f860c4c38e/src/charm.py#L247), [Kafka configuring additional storage](https://github.com/canonical/kafka-k8s-operator/blob/25cc5dd87bc2246c38fc511ac9c52f35f75f6513/src/charm.py#L298)
+This example uses {external+charmlibs:ref}`charmlibs-pathops <charmlibs-pathops>` to write data to the storage.
```
dimaqq: "the storage" --> "a storage unit" perhaps?

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
 
-```
+```python
 def _update_configuration(self, event: ops.EventBase):
     """Update the workload configuration."""
     cache = self.model.storages["cache"]
-    if cache.location is None:
-        # This must be one of the other events. Return and wait for the storage-attached event.
+    if not cache:
+        # The storage-attached event hasn't happened yet.
```
dimaqq: If we're talking holistic, maybe the charmer doesn't care about specific event, rather that the unit is in a state where storage has been declared, but not yet attached?

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
         logger.info("Storage is not yet ready.")
         return
-    try:
-        self.push_configuration(cache_dir=cache.location)
-    except ops.pebble.ConnectionError:
-        # Pebble isn't ready yet. Return and wait for the pebble-ready event.
-        logger.info("Pebble is not yet ready.")
-        return
+    # Storage is available. Write some data to the cache.
+    cache_root = pathops.LocalPath(cache[0].location)
```
dimaqq: I find it a little weird that we check `if cache`, but then access `cache[0]`.

That's valid for unitary storage (actually very common in charms), but what about multiple storage units?

dimaqq: I suppose what we're not saying here is that if ops returns a Storage object to the charm, then the storage unit is ready, and if it doesn't, charm doesn't get to know the reason for unit not being ready:
- neither charm requested it, nor admin attached it
- charm requested it, but juju didn't provide it yet 🧨
- juju did provide it, but we're behind events and it's not visible yet
- storage being detached (admin only)

The second case may be particularly thorny for reconciler charms that request more storage: there's a data race that could lead to the charm repeatedly requesting more storage and ending up with much more storage than needed.

docs/howto/manage-storage.md
dwilding: Yeah, not obvious what the right thing is here... We haven't introduced the concept of additional storage instances yet. So perhaps the best wording at this point is:

> If you don't specify a location for the storage volume, ...

Matching the wording at the very beginning of the doc

docs/howto/manage-storage.md
dwilding: It would be `/cache/0`, I think. Before my changes, there was a part that explained what the actual path would be, and why there would be numbers. I felt it was too distracting at this point, so removed it. (Why do I really care about the exact structure under`/cache/` if I'm not going to hardcode the path in my charm?)

But I'd be happy to clarify/reinstate the details

docs/howto/manage-storage.md
dwilding: That would probably make sense. Storage (= "storage volume"?) is the abstract concept, "storage unit" is the directory on disk that you actually write data to

docs/howto/manage-storage.md
dimaqq: I care because the charmer would copy-paste this declaration and code their charm to write to `/cache/mydb.sqlite`, which would fail.

We advise them to get the location from ops, but when it's this straightforward, they won't read or heed the advice.

dwilding: That's a fair point. I suspect you're right that, especially if I'm in a hurry, I'd copy the path and press on with using the path verbatim. You've convinced me that we need to add more context to this part

docs/howto/manage-storage.md
```diff
@@ -164,7 +96,107 @@ The storage will not be available immediately after that call - the charm should
...
+        ...
+    # Ask the workload to reload its configuration.
+```
+
+## Handle storage detaching
+
+In the `src/charm.py` file, in the `__init__` function of your charm, set up an observer for the detaching event associated with your storage and pair that with an event handler. For example:
+
+```
+self.framework.observe(self.on.cache_storage_detaching, self._on_storage_detaching)
```
tonyandrewmeyer:
```suggestion
self.framework.observe(self.on["cache"].storage_detaching, self._on_storage_detaching)
```

docs/howto/manage-storage.md
```diff
@@ -164,7 +96,107 @@ The storage will not be available immediately after that call - the charm should
...
+
+In your charm's `__init__` method, observe the [storage-attached](ops.StorageAttachedEvent) event:
+
+```python
+framework.observe(self.on["cache"].storage_attached, self._update_configuration)
+```
+
+In this example, we use a holistic event handler called `_update_configuration`. Alternatively, you could use a dedicated handler for the storage-attached event. To learn more about the different approaches, see [](#holistic-vs-delta-charms).
+
+Next, access the storage in `_update_configuration`. Don't hard-code the location of the storage. Instead, get the location from [`ContainerMeta.mounts`](ops.ContainerMeta.mounts) for the workload container:
```
tonyandrewmeyer: I think we should emphasise that the location shouldn't be hard-coded. I don't think it needs to be an admonition box, but perhaps:

```suggestion
Next, access the storage in `_update_configuration`. **Don't hard-code the location of the storage.** Instead, get the location from [`ContainerMeta.mounts`](ops.ContainerMeta.mounts) for the workload container:
```

docs/howto/manage-storage.md
tonyandrewmeyer: I think we should probably remove this line. Juju might change where the storage is ([and are looking at that for Juju 4](https://github.com/juju/juju/pull/21008)), and the charm really shouldn't care. I think by mentioning it, we're pushing them in the wrong direction.

docs/howto/manage-storage.md
tonyandrewmeyer: `/cache/<something>` where the 'something' is really up to Juju, although in current 3.6 it will be `0`, `1`, etc, yes.

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
 > See first: {external+juju:ref}`Juju | Storage <storage>`, {external+juju:ref}`Juju | Manage storage <manage-storage>`, {external+charmcraft:ref}`Charmcraft | Manage storage <manage-storage>`
 
-## Implement the feature
+## Manage storage for a machine charm
 
-### Declare the storage
+### Define the storage
 
-To define the storage that can be provided to the charm, define a `storage` section in `charmcraft.yaml` that lists the storage volumes and information about each storage. For example, for a transient filesystem storage mounted to `/cache/` that is at least 1GB in size:
+Define a `storage` section in `charmcraft.yaml` that lists the storage volumes and information about each storage. For example, for a transient filesystem storage mounted to `/cache/` that is at least 1GB in size:
```
tonyandrewmeyer: I wonder if we should leave out the location entirely. It's not particularly useful to the charm, because the charm should always get the location from Juju, and it leads to bad behaviour where the charm hard-codes paths based on this instead of asking Juju.

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
-self.framework.observe(self.on.cache_storage_attached, self._update_configuration)
-```
-
-> See more: [](ops.StorageAttachedEvent), [](/explanation/holistic-vs-delta-charms)
-
-Storage volumes will be automatically mounted into the charm container at either the path specified in the `location` field in the metadata, or the default location `/var/lib/juju/storage/<storage-name>`. However, your charm code should not hard-code the location, and should instead use the `.location` property of the storage object.
+In this example, we use a holistic event handler called `_update_configuration`. Alternatively, you could use a dedicated handler for the storage-attached event. To learn more about the different approaches, see [](#holistic-vs-delta-charms).
 
-Now, in the body of the charm definition, define the event handler, or adjust an existing holistic one. For example, to provide the location of the attached storage to the workload configuration:
+Next, access the storage in `_update_configuration`. Don't hard-code the location of the storage. Instead, get the location from [`self.model.storages`](ops.Model.storages):
```
tonyandrewmeyer: Similarly here, I think we should emphasise not hard-coding the location.

```suggestion
Next, access the storage in `_update_configuration`. **Don't hard-code the location of the storage.** Instead, get the location from [`self.model.storages`](ops.Model.storages):
```

docs/howto/manage-storage.md
tonyandrewmeyer:
```suggestion
        # The charm doesn't currently have any "cache" storage.
```

docs/howto/manage-storage.md
tonyandrewmeyer: Perhaps instead of writing data to the cache storage, we could set it up in some way, and ensure we do that for all the "cache" storage attached. Something like:

```python
# Ensure that the storages are set up.
for storage in cache:
    cache_root = pathops.LocalPath(storage.location)
    (cache_root / "images").mkdir(exist_ok=True)
    (cache_root / "videos").mkdir(exist_ok=True)
```

docs/howto/manage-storage.md
tonyandrewmeyer: I think we should *not* do that, because the charm shouldn't know/care where the storage has been mounted, it should just trust that the model provides the location. This isolates the charm from any changes Juju makes in terms of where/how storage gets mounted.

docs/howto/manage-storage.md
```diff
@@ -2,136 +2,68 @@
...
-
-```
-def _on_storage_detaching(self, event: ops.StorageDetachingEvent):
-    """Handle the storage being detached."""
-    self.unit.status = ops.ActiveStatus("Caching disabled; provide storage to boost performance)
+```python
+    ...
+    # Storage is available. Pass the storage location to the workload.
+    config_path = pathops.LocalPath("/etc/my-app/cache-path.config")
+    config_path.write_text(cache[0].location)
```
tonyandrewmeyer: And perhaps we could use all the storages here too, to make it clear that it's not always `[0]`. Something like:

```suggestion
    config_path = pathops.LocalPath("/etc/my-app/cache.config")
    config_path.write_text(f"paths = { ','.join(s.location for s in cache) }")
```

docs/howto/manage-storage.md
```diff
@@ -164,7 +96,107 @@ The storage will not be available immediately after that call - the charm should
 observe the `storage-attached` event and handle any remaining setup once Juju
```
tonyandrewmeyer: I can't attach this to the right line because it hasn't changed, but I believe there should be a "range" for the "scratch" storage, like the one above it. "multiple: range: 1-10" rather than "multiple: 1-10". I would probably again remove the "location" field as well.

Just above that, there's also a "storage-attached" that should probably be "`storage-attached`".

docs/howto/manage-storage.md
tonyandrewmeyer: Yeah, I think we should assume that if the charm wants to access the storage directly, it should do it similarly to the machine charm example, and then show the difference for the workload accessing the storage.

What about we have a first example that says something like "The storage is mounted locally in the charm container with the location provided by the storage.location attribute. For example, to have the charm set up the storage, you could ...". And then basically the machine example, except that there's only one storage.

And then say something about "the storage is also mounted on the workload container. If you need to provide the location to the workload, get it from meta like ...".

docs/howto/manage-storage.md
```diff
@@ -164,7 +96,107 @@ The storage will not be available immediately after that call - the charm should
...
+
+```
+def _on_storage_detaching(self, event: ops.StorageDetachingEvent):
+    """Handle the storage being detached."""
+    self.unit.status = ops.ActiveStatus("Caching disabled; provide storage to boost performance)
+```
+
+> Examples: [MySQL handling cluster management](https://github.com/canonical/mysql-k8s-operator/blob/4c575b478b7ae2a28b09dde9cade2d3370dd4db6/src/charm.py#L823), [MongoDB updating the set before storage is removed](https://github.com/canonical/mongodb-operator/blob/b33d036173f47c68823e08a9f03189dc534d38dc/src/charm.py#L596)
+
+## Write unit tests
```
tonyandrewmeyer, line 187: I think we could have a better test example here, where the interaction with the storage is in the charm (maybe actually the content from the examples above) rather than faking it all in the context manager. But that could also easily be left for a future improvement.

tonyandrewmeyer (commented):
Thanks for working on this - it's going to be much better (and actually correct) once we've done this. A few suggestions, mostly that we should push harder on always getting the location from Juju and not say where that probably will be.

benhoyt:
We'll wait to progress this one till David is back as it's not urgent.

docs/howto/manage-storage.md
dwilding: In the latest revision, I've gone with "storage instance"

docs/howto/manage-storage.md
dwilding: In the latest revision, I've removed the comment in favour of logging, which I feel should be enough explanation:

```python
logger.info("No instances available for storage 'cache'.")
```

docs/howto/manage-storage.md
dwilding: In the latest revision, I've gone with essentially what Tony suggested. I've also refactored the example to clearly be about multiple storage instances, with an explanation that `cache` gives you a singleton list in the case of a singular storage.

docs/howto/manage-storage.md
dwilding: I'm going to leave that for the future, as this PR already has a lot of rework

docs/howto/manage-storage.md
dwilding: In the latest revision, I've removed the code that tells the workload about the storage instance paths, as I felt it risks distracting from the more important points. Happy to reverse that decision if needed, though

docs/howto/manage-storage.md
dwilding: In the latest revision, I've refactored the example to not specify a location for the storage. I mention that you can, but explain why we don't recommend that you do.

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
+    type: filesystem
+    properties:
+      - transient
+    minimum-size: 1G
+    multiple:
+      range: 1-10
 ```
 
-For Kubernetes charms, you also need to define where on the workload container the volume will be mounted. For example, to mount a similar cache filesystem in `/var/cache/`:
+When your charm is deployed, Juju attaches one storage instance to each unit - the minimum of the range 1-10. The instance is at least 1GB in size. Each additional instance that's attached will also be at least 1GB in size.
```
dwilding: > Each additional instance that’s attached will also be at least 1GB in size.

I've not verified this. We'll need to verify or confirm with the Juju team.

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
+    web_cache_path = self.meta.containers["web"].mounts["cache"].location
+    # Configure the workload to use the storage instance path.
+    ...
 ```
 
-> Examples: [MySQL handling cluster management](https://github.com/canonical/mysql-k8s-operator/blob/4c575b478b7ae2a28b09dde9cade2d3370dd4db6/src/charm.py#L823), [MongoDB updating the set before storage is removed](https://github.com/canonical/mongodb-operator/blob/b33d036173f47c68823e08a9f03189dc534d38dc/src/charm.py#L596)
+> See more: [](ops.Model.storages), [](ops.ContainerMeta.mounts)
 
-### Request additional storage
+To access the storage instance in charm code, use one of the following approaches. The approaches are essentially equivalent - choose whichever you prefer.
```
dwilding: > The approaches are essentially equivalent - choose whichever you prefer.

Can we give better guidance than this? Maybe it's OK. I figured it's good to show both approaches for the sake of clarity about what's going on.

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
-...
-storage = self.model.storages['my-storage'][0]
-root = storage.location
-
-fname = 'foo.txt'
-fpath = root / fname
-with fpath.open('w') as f:
-    f.write('config info')
-...
+    framework.observe(self.on["cache"].storage_attached, self._update_configuration)
```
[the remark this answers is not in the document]
dwilding: No, we're assuming that `framework` is an argument of the `__init__` method, which is our usual convention.

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
 
-In the `src/charm.py` file, in the `__init__` function of your charm, set up an observer for the detaching event associated with your storage and pair that with an event handler. For example:
+### Configure the workload or access the storage
 
-```
-self.framework.observe(self.on.cache_storage_detaching, self._on_storage_detaching)
+In your charm's `__init__` method, observe the [storage-attached](ops.StorageAttachedEvent) event:
+
+```python
+    framework.observe(self.on["cache"].storage_attached, self._update_configuration)
```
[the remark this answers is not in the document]
dwilding: No, we're assuming that `framework` is an argument of the `__init__` method, which is our usual convention.

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
+In the `src/charm.py` file, in the `__init__` function of your charm, set up an observer for the detaching event associated with your storage and pair that with an event handler. For example:
 
-```yaml
-storage:
-    scratch:
-        type: filesystem
-        location: /scratch
-        multiple: 1-10
+```python
+    framework.observe(self.on["cache"].storage_detaching, self._on_storage_detaching)
```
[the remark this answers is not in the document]
dwilding: No, we're assuming that `framework` is an argument of the `__init__` method, which is our usual convention.

dwilding:
@dimaqq & @tonyandrewmeyer would you mind re-reviewing this one? I've significantly reworked the doc to address fundamental issues, so it would be best to review from scratch if possible. (I've updated the PR description to reflect the new approach)

Thanks a lot, and please take your time - this is not high priority

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
 > See first: {external+juju:ref}`Juju | Storage <storage>`, {external+juju:ref}`Juju | Manage storage <manage-storage>`, {external+charmcraft:ref}`Charmcraft | Manage storage <manage-storage>`
 
-## Implement the feature
+## Manage storage for a machine charm
 
-### Declare the storage
+### Define the storage
 
-To define the storage that can be provided to the charm, define a `storage` section in `charmcraft.yaml` that lists the storage volumes and information about each storage. For example, for a transient filesystem storage mounted to `/cache/` that is at least 1GB in size:
+Each storage can be defined as supporting a single or multiple storage instances. If you define a storage as supporting multiple instances, your charm's users can use `juju add-storage` to increase the number of instances attached to your charm. (This command doesn't 'grow' existing instances).
```
dimaqq:
```suggestion
Each storage can be defined as supporting a single or multiple storage instances. If you define a storage as supporting multiple instances, your charm's users can use `juju add-storage` to increase the number of instances attached to the current unit. (Note that command doesn't 'grow' existing instances).
```

docs/howto/manage-storage.md
dimaqq:
```suggestion
When your charm is deployed, by default, Juju attaches one storage instance to each unit - the minimum of the range 1-10. The instance is at least 1GB in size. Each additional instance that's attached will also be at least 1GB in size.
```

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
+    cache = self.model.storages["cache"]
+    if not cache:
+        logger.info("No instances available for storage 'cache'.")
+        return
+    cache_paths = [instance.location for instance in cache]
+    # Configure the workload to use the storage instance paths.
+    ...
+```
+
+The length of `cache_paths` matches the number of storage instances attached to the unit.
```
dimaqq:
```suggestion
The length of `cache_paths` matches the number of storage instances currently attached to the unit.
```

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
-```
-self.framework.observe(self.on.cache_storage_detaching, self._on_storage_detaching)
+In your charm's `__init__` method, observe the [storage-attached](ops.StorageAttachedEvent) event:
+
+```python
+    framework.observe(self.on["cache"].storage_attached, self._update_configuration)
 ```
 
-> See more: [](ops.StorageDetachingEvent)
+In this example, we use a holistic event handler called `_update_configuration`. Alternatively, you could use a dedicated handler for the storage-attached event. To learn more about the different approaches, see [](#holistic-vs-delta-charms).
```
dimaqq, line 128: totally optional: we could validate if pod comes up with storage already.

in which case workload container either has storage from the get-go or never has storage (e.g. `multiple: 0-1` deployed without)

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
+```python
+def _update_configuration(self, event: ops.EventBase):
+    """Update the workload configuration."""
+    cache = self.model.storages["cache"]
+    if not cache:
+        logger.info("No instance available for storage 'cache'.")
+        return
+    web_cache_path = self.meta.containers["web"].mounts["cache"].location
+    # Configure the workload to use the storage instance path.
+    ...
```
dimaqq: WDYT about a recommendation like this:

```py
if not cache: return
if not pebble-ready: return

container.replan(plan-with-storage-location)
```

However, this point is moot if the pod always comes up with storage.

If the storage is guaranteed to be there, and we hard-code the storage location, we could as well have the workload container image to expect the storage at that location, so the charm need not do anything (apart from kicking off the service on pebble ready)

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
-```
+    ```python
+        # Prepare the storage instance for use by the workload.
+        web_container = self.unit.get_container("web")
+        web_cache_root = pathops.ContainerPath(web_cache_path, container=web_container)
+        try:
+            (web_cache_root / "uploaded-data").mkdir(exist_ok=True)
+            (web_cache_root / "processed-data").mkdir(exist_ok=True)
+        except ops.pebble.ConnectionError:
+            logger.info("Unable to access container. Waiting for pebble-ready.")
```
dimaqq: Maybe the error can be simplified:
- this is holistic approach, we shouldn't care what event may come later
- there are other causes for pebble not being reachable (e.g. container restart)

docs/howto/manage-storage.md
```diff
@@ -2,169 +2,194 @@
...
+        charm_cache_root = pathops.LocalPath(charm_cache_path)
+        (charm_cache_root / "uploaded-data").mkdir(exist_ok=True)
+        (charm_cache_root / "processed-data").mkdir(exist_ok=True)
+    ```
 
-Juju will deploy the application with the minimum of the range (1 storage instance). Running `juju add-storage <unit> my-storage=32G,2` will add two additional instances to this storage. Adding storage does not modify existing storage mounts. This would generate two separate storage-attached events that should be handled.
+## Handle storage detaching
 
-If the *charm* needs additional units of a storage, it can request that with the `storages.request` method. The storage must be defined in the metadata as allowing multiple, for example:
+In the `src/charm.py` file, in the `__init__` function of your charm, set up an observer for the detaching event associated with your storage and pair that with an event handler. For example:
```
dimaqq, line 169: mention that it's only for machines... if that's only for machines

tonyandrewmeyer:
@dwilding sorry I've been slow on this - I meant to review before our doc session today, but that's looking unlikely now. I'll definitely do so this week before going into my break.

docs/howto/manage-storage.md
tonyandrewmeyer: They feel extra similar because `pathops` is hiding away that one is using Pebble (other than the error). I think if we're going to use `pathops` then we should just stick with one of them (and the local one probably makes more sense, because you don't need to worry about Pebble issues).

tonyandrewmeyer (approved):
Looks great! Thanks for doing all the re-work after the previous round.

docs/howto/manage-storage.md
dwilding: This is intended to apply to both machine and K8s. While I'm checking other parts in a K8s context, I'll check this too.

docs/howto/manage-storage.md
dwilding: Good suggestion, thanks. I've gone with this, to also capture something that Dima and I talked about:

> To access the storage instance in charm code, use [pathops](https://documentation.ubuntu.com/charmlibs/reference/charmlibs/pathops/#charmlibs-pathops) or standard file operations in the charm container. For example:
>
> ```python
>     # Prepare the storage instance for use by the workload.
>     charm_cache_path = cache[0].location  # Always index 0 in a K8s charm.
>     charm_cache_root = pathops.LocalPath(charm_cache_path)
>     (charm_cache_root / "uploaded-data").mkdir(exist_ok=True)
>     (charm_cache_root / "processed-data").mkdir(exist_ok=True)
> ```
>
> Alternatively, use [pathops.ContainerPath](https://documentation.ubuntu.com/charmlibs/reference/charmlibs/pathops/#pathops.ContainerPath) to access `web_cache_path` in the workload container. This approach is more appropriate if you need to manipulate lots of data in the workload container.

docs/howto/manage-storage.md
dwilding: In my simple test charm, storage was always available when my holistic handler looked for it earlier than storage-attached. However, I don't think this is guaranteed. I can't find any Juju docs that directly address this, [but there's a comment in the code](https://github.com/juju/juju/blob/3.6/internal/worker/uniter/storage/resolver.go):

> For CAAS models, we wait until after the start hook has run before running storage-attached hooks since storage is provisioned only after the pod has been created. [...] any charm logic that needs storage cannot be in install or start and if in config-changed, needs to be deferred until storage is available.

I've taken your suggestions and reworked the example like this:

```python
def _update_configuration(self, event: ops.EventBase):
    """Update the workload configuration."""
    cache = self.model.storages["cache"]
    if not cache:
        logger.info("No instance available for storage 'cache'.")
        return
    web_cache_path = self.meta.containers["web"].mounts["cache"].location
    # Configure the workload to use the storage instance path (assuming that
    # the workload container image isn't preconfigured to expect storage at
    # the location specified in charmcraft.yaml).
    # For example, provide the storage instance path in the Pebble layer.
    web_container = self.unit.get_container("web")
    try:
        web_container.add_layer(...)
    except ops.pebble.ConnectionError:
        logger.info("Workload container is not available.")
        return
    web_container.replan()
```

docs/howto/manage-storage.md
dwilding: Copilot and I did some digging. The [`add-storage` docs](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/add-storage/) say:

> `<size>` is the number of bytes to provision per storage instance.

Which, by default, is populated based on `minimum-size` in the metadata ([code](https://github.com/juju/juju/blob/8b588fd9f87eba537c275a38504be396e8076b0e/state/storage.go#L1988-L1989)).

docs/howto/manage-storage.md
dwilding: On reflecting, this is less about the amount of data. The point is more that there might be other data in the workload container that you need to put into the cache (for example). In which case it makes more sense to do the file operations in the container. I've changed the wording to:

> This approach is more appropriate if you need to reference additional data in the workload container.

docs/howto/manage-storage.md
dwilding: I looked into this further. Detaching storage manually is only possible for machine charms. But a K8s charm might still want to observe the event. I've created a separate issue for improving this section of the doc: https://github.com/canonical/operator/issues/2257

Issue #456, opened 2020-11-23
Test harness relations missing Juju managed data


stub42:
The `ingress-address` and `egress-subnets` relation data is maintained by Juju, and often required. The Operator Framework test harness does not initialize this information when doing harness.add_relation_unit(), so relation-joined handlers can fail due to missing relation data that should always be there.

chipaca:
Is there any reasonable default that could be populated by the harness itself?

stub42:
Dummy values could be set, random-but-stable IP addresses or just completely random would work.

Alternatively, a way for the charm author to initialize them via add_relation_unit (to ensure they are set during relation-joined hooks) would be enough.

Or defer until we know how the harness is handling other network information, since the relation data really should match what is returned by network-get.

jameinel:
I think we have an open issue for modeling networking generally, but we can
take this as a focused piece of that puzzle.

dstathis:
I'd like to bump this. I am currently using this data in one of my charm libraries and would love not to have to set it automatically.

PR #2099, opened 2025-10-11
ci: add automated doc checks (and related starter pack updates)

dwilding:
This PR does a few things related to the [documentation starter pack](https://github.com/canonical/sphinx-docs-starter-pack):

- Makes sure that the spell checker can be run with `make -C docs spelling`, makes sure that the spell checker _passes_, and adds a workflow to run the spell checker on PRs & pushes.

- Configures the link checker with URLs that are known to fail, and adds a manual workflow to run the link checker. Manual for now because the link checker takes a long time and produces unexpected errors (see my comments for more detail).

- Adds the update script from the starter pack.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -253,9 +253,8 @@ redirects = {}
 # Link checker exceptions #
 ###########################
 
-# A regex list of URLs that are ignored by 'make linkcheck'
-#
-# TODO: Remove or adjust the ACME entry after you update the contributing guide
+# During linkcheck, if a target URL matches any of these regexes (using re.match)
+# then the URL is skipped.
 
 linkcheck_ignore = [
     # Excluded because the pages don't contain elements with an ID matching the URL fragment.


.github/workflows/docs-linkcheck.yaml
```diff
@@ -0,0 +1,24 @@
+name: Docs links
+
+on:
+  # The docs have a lot of external links, so the link checker takes a long time to run.
+  # As of Oct 2025, the link checker fails on many Read the Docs links because RTD's
+  # servers respond with 429 (Too Many Requests) without allowing retry after delay.
+  # Given this, it's best to run the workflow manually and inspect the results.
```
dwilding, line 7: I believe we're talking to RTD about this

docs/conf.py
```diff
@@ -258,7 +258,9 @@
 # TODO: Remove or adjust the ACME entry after you update the contributing guide
 
 linkcheck_ignore = [
-    "http://127.0.0.1:8000"
+    r"https://matrix\.to/#/",
+    r"https://documentation\.ubuntu\.com/pebble/reference/api/#/",
+    r"https://documentation\.ubuntu\.com/juju/3\.6/reference/hook-command/list-of-hook-commands/#list-of-hook-commands",
```
dwilding, line 263: These URLs need to be excluded because the pages don't contain elements with an ID matching the URL fragment

docs/howto/manage-storage.md
```diff
@@ -43,7 +43,7 @@ containers:
 When you use storage mounts with Juju, they will be automatically mounted into the charm container at either:
 
 * the specified location based on the storage section of `charmcraft.yaml`, or
-* the default location `/var/lib/juju/storage/<storage-name>/<num>`, where num is zero for 'normal'/singular storages or an integer ID for storages that support multiple attachments.
+* the default location `/var/lib/juju/storage/<storage-name>/<num>`, where `<num>` is zero for 'normal'/singular storages or an integer ID for storages that support multiple attachments.
```
dwilding, line 46: Minor doc change to avoid adding a spelling exception for "num"

docs/Makefile
```diff
@@ -60,6 +60,7 @@ $(VENVDIR): requirements.txt ../uv.lock
 	@test -d $(VENVDIR) || uv venv --python 3.11 $(VENVDIR)
 	uv pip install --directory $(VENVDIR) --requirements ../../requirements.txt
 	uv pip install --directory $(VENVDIR) "ops[testing,tracing] @ ../../.."
+	uv pip install --directory $(VENVDIR) rst2html vale
```
dwilding: These dependencies are required by the spellchecker. I suppose we could add them to `requirements.txt`, but then that file would no longer match the starter pack. This part of the Makefile already doesn't match the starter pack

docs/.sphinx/update_sp.py
```diff
@@ -0,0 +1,269 @@
+#! /usr/bin/env python
```
dwilding, line 1: File added from the starter pack, unchanged

docs/.sphinx/get_vale_conf.py
```diff
@@ -0,0 +1,151 @@
+#! /usr/bin/env python
```
dwilding, line 1: File added from the starter pack, unchanged except for cleaning up whitespace (which I'm fixing in the starter pack: [#448](https://github.com/canonical/sphinx-docs-starter-pack/pull/448))

docs/conf.py
james-garner-canonical: Worth adding this as a comment?

docs/Makefile
james-garner-canonical: I'm curious why the starter pack doesn't include the spellchecker requirements?

dwilding: It does, but it [hardcodes them in the Makefile](https://github.com/canonical/sphinx-docs-starter-pack/blob/main/docs/Makefile#L115). Actually, now I think about it, why aren't they listed in `requirements.txt` instead...? I'll investigate that before finishing this PR

dwilding:
Thanks!

> I guess we weren't running the docs workflows in CI in `ops` previously, since these are new workflows.

Yes, that's right

docs/.sphinx/version
```diff
@@ -0,0 +1 @@
+1.2.0
```
dimaqq, line 1: What is this version of?

docs/conf.py
dimaqq: Are these URLs or regular expressions to match URLs against? prefixes? re.search or re.match patterns?
- If these are exact URLs, why are dot escaped?
- And these are patterns, please update the comment.

docs/Makefile
dwilding: I'm updating the starter pack to put the dependencies in `requirements.txt` ([#456](https://github.com/canonical/sphinx-docs-starter-pack/pull/456)). I've preemptively included the same changes here.

docs/.sphinx/version
dwilding: This is the version of the starter pack that was used. The file is managed by `docs/.sphinx/update_sp.py` which can do a (semi automated) in-place upgrade of the starter pack.

docs/conf.py
dwilding: Thanks for raising this. I agree, it's poorly explained. I've improved the comment:

> During linkcheck, if a target URL matches any of these regexes (using re.match) then the URL is skipped.

dimaqq: Can't be `re.match` though.
What brought me here was this logic:

Surely, `r"https://matrix\.to/#/"` is not a full URL, some channel is needed to make it sensical.

But then, why don't I see what is variable?

If it's a search pattern, maybe it should look like `^http://xxx`?

If it's a match pattern, then maybe smth like `xxx/#/.*`?

dimaqq: Apparently they use `re.match` which "matches at the start of the string" and not `re.fullmatch` which would require both start and end to be exact.

PR #2100, opened 2025-10-13
ci: support monorepos in ops charm compatibility testing

james-garner-canonical:
This PR extends the `published-charms-tests` workflow to support running unit tests for charms that are in sub-directories of their source repo, including charms in monorepos.

## Changes

The `update-published-charms-tests-workflow.py` script has been updated to move the charms that were included in the `SKIP` list due to being in sub-directories to a dictionary mapping these repo names to their charm sub-directory(s). The output of the script has been updated to include the `charm-root` key, and the workflow has been updated to set this as the working directory.

### Changes supporting local development and testing

Additionally, the script now accepts CLI arguments:
- `--packages` to load a `packages.json` (avoiding a request to `charmhub.io` each run that may time out -- I needed this locally to run the script at all)
- `--charms` to load a `charms.json` file, avoiding an additional request to `charmhub.io` for each charm, which otherwise takes up the vast majority of the scripts execution time.
- 
## Testing

I've tested these changes [by running the workflow in my fork](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918558). We test eight new charms due to the changes introduced here, four of which pass their unit tests with the latest copy of `ops`, three of which fail due to requiring `ops` < 3, and one of which fails due to a [particularly unique charm layout](https://github.com/canonical/k8s-operator/blob/main/charms/CONTRIBUTION.md) (however the other charm in that repo passes).

* ❌ [canonical/argo-operators:charms/argo-controller](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918538) requires `ops` < 3
* ✅ [canonical/jimm:charms/jimm](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918558)
* ❌ [canonical/k8s-operator:charms/worker](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918560) No `requirements.txt` etc (`pyproject.toml` is in a subfile, because this monorepo [uses a nested charm layout instead of side-by-side](https://github.com/canonical/k8s-operator/blob/main/charms/CONTRIBUTION.md))
* ✅ [canonical/k8s-operator:charms/worker/k8s](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918563)
* ❌ [canonical/notebook-operators:charms/jupyter-controller](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918586) requires `ops` < 3
* ❌ [canonical/notebook-operators:charms/jupyter-ui](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918584#logs) requires `ops` < 3
* ✅ [canonical/vault-k8s-operator:k8s](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918648)
* ✅ [canonical/vault-k8s-operator:machine](https://github.com/james-garner-canonical/operator/actions/runs/18451595327/job/52565918655)


.github/workflows/published-charms-tests.yaml
```diff
@@ -17,52 +17,162 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - charm-repo: canonical/admission-webhook-operator
```
tonyandrewmeyer: What do you think about having a separate PR that refreshes this file? It would then be clearer what's just a newly public charm (or in one case a un-public'd one) and what's being done in this PR.

.github/update-published-charms-tests-workflow.py
```diff
@@ -51,29 +51,32 @@
...
+    'k8s-operator': ['charms/worker', 'charms/worker/k8s'],
+    'notebook-operators': ['charms/jupyter-controller', 'charms/jupyter-ui'],
+    'vault-k8s-operator': ['k8s', 'machine'],
+}
 
 
 def packages():
     """Get the list of published charms from Charmhub."""
+    if (packages_file := pathlib.Path('packages.json')).exists():
+        return json.loads(packages_file.read_text())['packages']
```
tonyandrewmeyer: I don't (right now) have any issue downloading the list of packages (it takes 2-9s for me right now), but I'm open to having a way to provide a file locally for development purposes.

However, rather than this implicit "if there's a file of the right name locally" approach, I would rather that we added an explicit option for running the script to provide a (path to the) file.

.github/update-published-charms-tests-workflow.py
```diff
@@ -51,29 +51,32 @@
...
     logger.info('Fetching the list of published charms')
     url = 'https://charmhub.io/packages.json'
     with urllib.request.urlopen(url, timeout=120) as response:  # noqa: S310 (unsafe URL)
         data = response.read().decode()
-        packages = json.loads(data)['packages']
-    return packages
+    packages_file.write_text(data)
+    return json.loads(data)['packages']
```
tonyandrewmeyer: Is the purpose of this change to cache the file? I don't like the idea that we cache it by default. I guess leaving it in place could be an option, but a plain `curl` or `wget` seems a simpler way to download it if you want to do that.

If we do cache it, I would much rather it was somewhere else, like the temporary directory or a cache directory, rather than in the ops clone.

.github/update-published-charms-tests-workflow.py
```diff
@@ -85,7 +88,7 @@ def get_source_url(charm: str):
         ) as response:
             data = json.loads(response.read().decode())
             return data['result']['links']['source'][0]
-    except (urllib.error.HTTPError, KeyError):
+    except (urllib.error.HTTPError, KeyError, IndexError):
```
tonyandrewmeyer, line 97: I'm assuming that this means that someone has an empty links:source list? It seems reasonable to handle that, although I'd also consider suggesting upstream that it's a poor decision.

.github/update-published-charms-tests-workflow.py
```diff
@@ -51,29 +51,32 @@
...
     # Maintenance mode / archived.
     'charm-sysconfig',
 }
+CHARM_ROOTS = {
+    'argo-operators': ['charms/argo-controller'],
+    'jimm': ['charms/jimm'],
+    'k8s-operator': ['charms/worker', 'charms/worker/k8s'],
+    'notebook-operators': ['charms/jupyter-controller', 'charms/jupyter-ui'],
+    'vault-k8s-operator': ['k8s', 'machine'],
+}
```
tonyandrewmeyer, line 75: This can definitely be left for later work, either as part of this project or another time, but I think we should be able to figure this out rather than hard code it, particularly while we're only handling GitHub repositories.

Something like "look at the top level for a metadata/charmcraft.yaml file, and if not look at each folder in the top level" would find it, right? That's potentially quite a lot of extra https requests, so maybe it ought to be cached, though, or only done when adding to the file.

.github/update-published-charms-tests-workflow.py
```diff
@@ -123,13 +126,27 @@ def url_to_charm_name(url: str | None):
 def main():
     """Update the workflow file."""
     logging.basicConfig(level=logging.INFO)
-    charms = [url_to_charm_name(get_source_url(package['name'])) for package in packages()]
-    charms = [charm for charm in charms if charm and charm not in SKIP]
-    charms.sort()
+    if (charms_file := pathlib.Path('charms.json')).exists():
```
tonyandrewmeyer: I'm less keen on caching this, if it's purely "make it faster while developing" rather than "make it work when developing because the backend is too slow sometimes". But if we do, then again I would rather it was an option where you provide some location for the file, not defaulting to be next to the script.

If they do stay where they are, surely they have to be in .gitignore, otherwise there's this:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	charms.json
	packages.json
```

I'm curious here why the skip check was moved to later - what's the advantage of having the skipped charms in the cache file?

.github/workflows/published-charms-tests.yaml
```diff
@@ -17,52 +17,162 @@ jobs:
...
+          - charm-repo: canonical/kafka-k8s-operator
+            charm-root: .
           - charm-repo: canonical/kafka-operator
+            charm-root: .
+          - charm-repo: canonical/katib-operators
+            charm-root: .
+          - charm-repo: canonical/kfp-operators
+            charm-root: .
+          - charm-repo: canonical/kfp-operators
+            charm-root: .
```
tonyandrewmeyer: Odd that this is there twice. Also, when I run the script locally, one goes away. Maybe a glitch on the charmhub side when you ran it?

.github/workflows/published-charms-tests.yaml
```diff
@@ -17,52 +17,162 @@ jobs:
...
           - charm-repo: canonical/nginx-ingress-integrator-operator
+            charm-root: .
+          - charm-repo: canonical/notebook-operators
+            charm-root: charms/jupyter-controller
+          - charm-repo: canonical/notebook-operators
+            charm-root: charms/jupyter-ui
+          - charm-repo: canonical/notebook-operators
+            charm-root: charms/jupyter-controller
+          - charm-repo: canonical/notebook-operators
+            charm-root: charms/jupyter-ui
```
tonyandrewmeyer: Similarly here, it's duplicated, but when I run it locally the duplicate disappears.

.github/workflows/published-charms-tests.yaml
james-garner-canonical: No objections to separating it out.

Would you want to include the newly-not-skipped monorepo charms in the workflow in this PR, or have that wait till the workflow file is refreshed too?

.github/update-published-charms-tests-workflow.py
james-garner-canonical: I'm totally happy with an opt-in cli args based approach.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: I think you'd have the monorepo ones in this one.

.github/update-published-charms-tests-workflow.py
james-garner-canonical: If we add a cli argument, then dropping this caching like a good idea to me.

.github/update-published-charms-tests-workflow.py
james-garner-canonical: Agreed on leaving this as-is for now.

---

> That's potentially quite a lot of extra https requests, so maybe it ought to be cached, though, or only done when adding to the file.

I guess (shallowly) cloning the repo might be the way to go if we did this.

> "look at the top level for a metadata/charmcraft.yaml file, and if not look at each folder in the top level"

This recipe wouldn't find all these charms (it would only find the vault ones) -- we'd need to go at least two levels deep. I guess we'd stop recursing after finding a charm, to avoid collecting any charms defined in its tests, though that would mean that the 'inner' `k8s-operator` charm wouldn't be collected.

---

My favoured solution to avoid hard-coding it on our end would be for charm monorepos to have a top-level `charms.yaml` file with this information.

.github/update-published-charms-tests-workflow.py
james-garner-canonical: Aside from the timeout issues I faced with downloading `packages.json` via the script (I pulled it via my browser instead), building the charm URLs list is actually by far the slowest part of running this script, because of all the requests it has to make to Charmhub.

I'm fine with having the caching / loading controlled by CLI args.

> I'm curious here why the skip check was moved to later - what's the advantage of having the skipped charms in the cache file?

Caching the skipped charms allows you to make changes to the skip list without needing to rebuild the charms list, which was useful when working on this PR, as I moved charms from `SKIP` to `REPO_ROOTS` incrementally, and seems generally useful.

.github/workflows/published-charms-tests.yaml
james-garner-canonical: I'm not sure how this happened, but re-running the script locally for me now removes the duplicates, so I've fixed this.

.github/workflows/published-charms-tests.yaml
james-garner-canonical: Same, resolved.

.github/workflows/published-charms-tests.yaml
james-garner-canonical: Done

.github/update-published-charms-tests-workflow.py
james-garner-canonical: I've added CLI args and dropped the automatic caching and loading.

.github/update-published-charms-tests-workflow.py
james-garner-canonical: Indeed, the response for [coredns](https://charmhub.io/coredns) looks like this:
```json
{
  "id": "0gorpaPu7KPe5la0wHtT9B25C0FqlzHw",
  "name": "coredns",
  "result": {
    "links": {
      "contact": [
        "mailto:Cory Johns <cory.johns@canonical.com>"
      ],
      "docs": [
        "https://discourse.charmhub.io/t/coredns-docs-index/6322"
      ],
      "issues": [],
      "source": [],
      "website": []
    }
  },
  "type": "charm"
}
```

.github/update-published-charms-tests-workflow.py
tonyandrewmeyer: > I guess (shallowly) cloning the repo might be the way to go if we did this.

Yes, I wondered about that too. Particularly if that is optionally cached - in my experience doing a large number of `git pull` operations concurrently is very fast.

> This recipe wouldn't find all these charms (it would only find the vault ones) -- we'd need to go at least two levels deep.

Ah, ok. Even more motivation to just clone it I think. Definitely something for another day.

> My favoured solution to avoid hard-coding it on our end would be for charm monorepos to have a top-level `charms.yaml` file with this information.

Yeah, I should have mentioned above that another reason to not do it now is that maybe it gets solved by something like that in the near-ish future.

.github/update-published-charms-tests-workflow.py
```diff
@@ -123,13 +133,44 @@ def url_to_charm_name(url: str | None):
...
     logging.basicConfig(level=logging.INFO)
-    charms = [url_to_charm_name(get_source_url(package['name'])) for package in packages()]
-    charms = [charm for charm in charms if charm and charm not in SKIP]
-    charms.sort()
+    parser = argparse.ArgumentParser()
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        '--packages',
+        type=pathlib.Path,
+        help='A json file matching the format served by Charmhub.',
```
tonyandrewmeyer:
```suggestion
        help='A JSON file matching the format served by Charmhub.',
```

.github/update-published-charms-tests-workflow.py
```diff
@@ -123,13 +133,44 @@ def url_to_charm_name(url: str | None):
...
+    )
+    group.add_argument(
+        '--charms',
+        type=pathlib.Path,
+        help='A json file matching the format output by this script.',
```
tonyandrewmeyer:
```suggestion
        help='A JSON file matching the format output by this script.',
```

.github/update-published-charms-tests-workflow.py
```diff
@@ -123,13 +133,44 @@ def url_to_charm_name(url: str | None):
...
+    )
+    args = parser.parse_args()
+    if args.charms:
+        charms = json.loads(args.charms.read_text())
```
tonyandrewmeyer, line 150: We had this discussion elsewhere recently, but just in case you changed your mind, what about:

```suggestion
        with args.charms.open() as f:
            charms = json.load(f)
```

.github/update-published-charms-tests-workflow.py
```diff
@@ -123,13 +133,44 @@ def url_to_charm_name(url: str | None):
...
+    else:
+        if args.packages:
+            packages = json.loads(args.packages.read_text())['packages']
+        else:
+            packages = get_packages_from_charmhub()
+        charms = [url_to_charm_name(get_source_url(pkg['name'])) for pkg in packages]
+        charms = sorted({c for c in charms if c})
+    print(json.dumps(charms))
```
tonyandrewmeyer, line 158: To be fair, I guess I'm not suggesting `json.dump(charms, sys.stdout)` here...

.github/update-published-charms-tests-workflow.py
james-garner-canonical: Funnily enough, my initial work on this script used the `with` form.

I had a painful time noticing that I had the `json.dump` arguments the wrong way around, resulting in several costly runs of generating all the charm URLs only to crash when trying to write them to a file.

I know this suggestion is for `json.load`, but that experience (combined with `tomllib.load` requiring the file to be open in binary mode) has cemented my preference for the `{toml,json,yaml}.{load,dump}s(path.read_text())` form.

.github/update-published-charms-tests-workflow.py
```diff
@@ -123,13 +133,44 @@ def url_to_charm_name(url: str | None):
...
+            continue
+        for root in CHARM_ROOTS.get(charm, ['.']):
+            item = """
+          - charm-repo: canonical/{repo}
+            charm-root: {root}
+""".strip('\n').format(repo=charm, root=root)
+            items.append(item)
+    target = r'(\s{10}- charm-repo: \S+\n\s{10}  charm-root: \S+\n)+'
+    replacement = '\n'.join(items) + '\n'
+    workflow = re.sub(target, replacement, workflow, count=1)
```
dimaqq: Discussed during the meeting:

what if we always overwrote the YAML file. e.g.:

```yaml
on:
  foo:

### BEGIN OPS STUFF ###
  - this:
    get:
  - overwritten:
    entirely:  # with fresh comments if needed
### END OPS STUFF ###

james-garner-canonical:
Per @dimaqq's suggestion, I've replaced the regex use with begin and end markers.

I've verified locally that the only difference in the workflow file when running the regex based implementation vs running the new implementation is the presence of the begin and end markers.

dimaqq (approved):

tonyandrewmeyer (approved):
The simpler/non-regex approach works for me.

PR #2103, opened 2025-10-14
ci: test both Charmcraft 3 and Charmcraft 4 profiles

tonyandrewmeyer:
Add support for Charmcraft 4 in preparation for it becoming "latest/stable".

* The "simple" profile is not present in Charmcraft 3, so we shouldn't try to run tests against it.
* The "static" tox environment has been combined with the "lint" environment in the Charmcraft 4 profiles.

It seems like for a while it's worth testing both Charmcraft 3 and Charmcraft 4, so rather than replacing 3, add 4 into a matrix.

[Example run in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/18576943297/job/52964083743).

### DIFF
--- a/.github/workflows/published-charms-tests.yaml
+++ b/.github/workflows/published-charms-tests.yaml
@@ -75,6 +75,7 @@ jobs:
 
       - name: Update 'ops' and 'ops-scenario' dependencies in test charm to latest
         run: |
+          set -xueo pipefail
           if [ -e "test-requirements.txt" ]; then
             sed -i -e "/^ops[ ><=]/d" -e "/canonical\/operator/d" -e "/#egg=ops/d" test-requirements.txt
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> test-requirements.txt
@@ -118,6 +119,7 @@ jobs:
         id: has-static
         if: ${{ !cancelled() && steps.deps.outcome == 'success' }}
         run: |
+          set -xueo pipefail
           if tox list --no-desc | grep '^static$'; then
             echo 'static=true' >> "$GITHUB_OUTPUT"
           else


dimaqq:
The PR title doesn't seem to match the first paragraph of the description.

tonyandrewmeyer:
> The PR title doesn't seem to match the first paragraph of the description.

I'm not sure what you mean here. Did you mis-read "only test 'simple'" as "test only 'simple'" maybe?

In any case, the scope expanded slightly, so I'll update the summary.

.github/workflows/published-charms-tests.yaml
```diff
@@ -155,10 +159,26 @@ jobs:
...
+
+      # Charmcraft 4 doesn't have a separate 'static' environment.
+      - name: Check for tox static environment
+        id: has-static
+        if: ${{ !cancelled() && steps.deps.outcome == 'success' }}
+        run: |
+          if tox list --no-desc | grep '^static$'; then
+            echo 'static_cc=true' >> "$GITHUB_OUTPUT"
+          else
+            echo 'static_cc=false' >> "$GITHUB_OUTPUT"
```
james-garner-canonical: The meaning of the `cc` suffix isn't clear to me. WDYT about just `static` or `result`?

.github/workflows/published-charms-tests.yaml
```diff
@@ -155,10 +159,26 @@ jobs:
...
 
       - name: Install dependencies
+        id: deps
         run: pip install tox~=4.2
 
       - name: Run the charm's unit tests
         run: tox -vve unit
 
+      - name: Run the charm's lint tests
+        run: tox -vve lint
```
james-garner-canonical, line 176: Should this also run if unit tests fail? WDYT about adding `if: ${{ !cancelled() && steps.deps.outcome == 'success' }}` to both the unit step and this step?

.github/workflows/published-charms-tests.yaml
```diff
@@ -155,10 +159,26 @@ jobs:
           fi
 
       - name: Install dependencies
+        id: deps
         run: pip install tox~=4.2
```
james-garner-canonical: Won't tox already be installed at this point because it's a dependency of `tox-uv`? If we don't need a `deps` step after updating the charm's dependencies, I guess the subsequent steps can depend on the 'Update dependencies' step succeeding instead.

.github/workflows/published-charms-tests.yaml
```diff
@@ -134,17 +134,21 @@ jobs:
...
       - name: Install charmcraft
-        run: sudo snap install charmcraft --classic
+        run: sudo snap install charmcraft --classic --channel=${{ matrix.charmcraft_channel }}
 
       - name: Charmcraft init
         run: charmcraft init --profile=${{ matrix.profile }} --author=charm-tech
 
+      - name: Install patch dependencies
+        run: pip install uv~=0.9 tox-uv~=1.29
+
```
james-garner-canonical: Do we actually need these for patching deps? Looks like we only handle the `requrements.txt` case which is pure `sed`. But I guess the actual answer is that we do need these, we just need to also copy paste the `uv.lock` branch from earlier in the workflow as well, since Charmcraft 4 profiles don't use `requirements.txt`.

I'm guessing the workflow passes because it just doesn't patch the dependency on the new profiles. WDYT about adding a final else block to error out if we fall through without patching?

.github/workflows/published-charms-tests.yaml
```diff
@@ -134,17 +134,21 @@ jobs:
...
       fail-fast: false
       matrix:
+        profile: [machine, kubernetes]
+        charmcraft_channel: ["3.x/stable", "latest/edge"]
         include:
-          - profile: machine
-          - profile: kubernetes
-          - profile: simple
+          - charmcraft_channel: "3.x/stable"  # Last version to include "simple"
+            profile: simple
```
james-garner-canonical, line 143: Minor: how about matching the order of the matrix arguments so the job names are consistent? (Either swap the order of the two matrix lines, or these two.)

james-garner-canonical (changes requested):
There are a few things I think need addressing. Most significantly, I don't think we're actually patching the dependencies for the Charmcraft 4 profiles currently. The overall approach looks good though, thanks for doing this.

.github/workflows/published-charms-tests.yaml
```diff
@@ -134,17 +134,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        profile: [machine, kubernetes]
+        charmcraft_channel: ["3.x/stable", "latest/edge"]
```
tonyandrewmeyer:
```suggestion
        charmcraft_channel: ["3.x/stable", "latest/edge"]
        profile: [machine, kubernetes]
```

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: We need tox-uv because the tox commands used in the Charmcraft 4 profiles won't run without it (the runner keyerror thing). The "patch" word is a bad copy & paste.

.github/workflows/published-charms-tests.yaml
```diff
@@ -134,17 +134,21 @@ jobs:
...
+            profile: simple
     steps:
       - name: Install charmcraft
-        run: sudo snap install charmcraft --classic
+        run: sudo snap install charmcraft --classic --channel=${{ matrix.charmcraft_channel }}
 
       - name: Charmcraft init
         run: charmcraft init --profile=${{ matrix.profile }} --author=charm-tech
 
+      - name: Install patch dependencies
```
tonyandrewmeyer:
```suggestion
      - name: Install dependencies
```

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: "cc" is charmcraft. I wanted to avoid using the same name as in the charm job. I can change it to something else.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: Pushing the scope of this PR a bit, but ok.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: Yes, we don't need two dependency installs, merged.

.github/workflows/published-charms-tests.yaml
```diff
@@ -154,11 +159,25 @@ jobs:
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops-scenario&subdirectory=testing" >> requirements.txt
           fi
 
-      - name: Install dependencies
-        run: pip install tox~=4.2
-
       - name: Run the charm's unit tests
+        if: ${{ !cancelled() && steps.deps.outcome == 'success' }}
```
james-garner-canonical: Now that deps has moved before the patch step, I think it would be better to depend on the patch step instead.

You're probably in the middle of addressing it, but the comment has disappeared from the file view now, so just to note that I think the patch step needs to handle the `uv.lock` case, and should have a fallthrough to fail if we don't patch.

tonyandrewmeyer: Yeah, I accidentally assigned reviewers before moving out of draft, sorry. I'll re-request review when it's ready.

james-garner-canonical: I actually didn't notice it was in draft till after reviewing. I'm surprised we got notified before you hit 'ready for review'.

tonyandrewmeyer: Yeah, I wish GitHub didn't do that, and if you do want review while something is in draft, you could just @ people in a comment.

.github/workflows/published-charms-tests.yaml
```diff
@@ -158,19 +157,26 @@ jobs:
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops" >> requirements.txt
             echo -e "\ngit+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA#egg=ops-scenario&subdirectory=testing" >> requirements.txt
           fi
+          if [ -e "uv.lock" ]; then
```
james-garner-canonical: Should be `elif` with the previous line. Sorry for backseat coding haha.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: The uv updating is in the script now.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: I got rid of these instead - I don't think they're really needed in the charmcraft profile tests. If installing tox or tox-uv or uv fails then the job can fail at that point, or when we try to use them.

.github/workflows/published-charms-tests.yaml
james-garner-canonical: Nice, LGTM.

.github/workflows/published-charms-tests.yaml
```diff
@@ -134,31 +134,54 @@ jobs:
...
+        run: sudo snap install charmcraft --classic --channel=${{ matrix.charmcraft_channel }}
 
       - name: Charmcraft init
         run: charmcraft init --profile=${{ matrix.profile }} --author=charm-tech
 
+      - name: Install dependencies
+        run: pip install tox~=4.2 uv~=0.9 tox-uv~=1.29
+
       - name: Update 'ops' and 'ops-scenario' dependencies in test charm to latest
         run: |
```
james-garner-canonical, line 155: WDYT about setting some standard defensive bash flags here?
```suggestion
        run: |
          set -xueo pipefail
```
- `-x` Will print lines before executing them, with variables resolved, very helpful for debugging -- if we only used one flag, I would pick this one.
- `-u` Will fail if variables are unassigned -- not expected here, but handy to avoid typos in github env vars etc.
- `-e` Fail if any of the commands fail. Might require adding `|| true` or similar if some of these are expected to fail through.
- `-o pipefail` Not actually relevant here since we don't use any pipelines, but a sensible default that I use.

.github/workflows/published-charms-tests.yaml
```diff
@@ -134,31 +134,54 @@ jobs:
...
       - name: Run the charm's unit tests
         run: tox -vve unit
 
+      - name: Run the charm's lint tests
+        run: tox -vve lint
+
+      # Charmcraft 4 doesn't have a separate 'static' environment.
+      - name: Check for tox static environment
+        id: has-static
+        run: |
```
james-garner-canonical, line 181:
```suggestion
        run: |
          set -xueo pipefail
```

I'd be on board with adding these to the other non-trivial bash scripts in this file too while we're at it, like the ones in the published charm tests. Even `set -x` alone would be helpful and not change the failure conditions at all.

james-garner-canonical (approved):
Nice, I see this is working as expected in [the workflow run](https://github.com/tonyandrewmeyer/operator/actions/runs/18576943297/job/52964083739) now. A small suggestion to make the bash scripts easier to debug, but LGTM.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: This feels like more scope creep, but ok.

PR #1795, opened 2025-06-06
docs: improve testing.errors.UncaughtCharmError message

benhoyt:
This changes the UncaughtCharmError message and puts the message on the same source line as the "raise", so that when using "pytest --pdb" to debug a charm exception, it shows more clearly what to do.

In Python 3.13 the [`exceptions [n]`](https://docs.python.org/3/library/pdb.html#pdbcommand-exceptions) command was added to pdb, which allows you to go back up the exception cause stack. For example:

```
$ pytest --pdb
...
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/ben/.cache/uv/archive-v0/gz1Vrbxq-976wQzrfehra/lib/python3.13/site-packages/scenario/_runtime.py(345)exec()
-> raise UncaughtCharmError(f'Uncaught {type(e).__name__} in charm, try "exceptions [n]" if using pdb on Python 3.13+. Details: {e!r}') from e  # fmt: skip
(Pdb) exceptions
    0 ValueError('a value error')
>   1 UncaughtCharmError('Uncaught ValueError in charm, try "exceptions [n]" if usi...
(Pdb) exceptions 0
> /home/ben/w/test-charms/database/src/charm.py(75)_on_debug_action()
-> raise ValueError("a value error")
(Pdb) event
<ActionEvent self.id='2' via DatabaseCharm/on/debug_action[1]>
```

Previously it would just show the "raise UncaughtCharmError" part:

```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /code/hexanator/.tox/unit/lib/python3.12/site-packages/scenario/runtime.py(469)exec()
-> raise UncaughtCharmError(
(Pdb)
```

Note that the above only shows the last line -- the full output from pdb did already include more context:

```
...
>               raise UncaughtCharmError(f'Uncaught {type(e).__name__} in charm, try "exceptions [n]" if using pdb on Python 3.13+. Details: {e!r}') from e  # fmt: skip
[...]

Issue #1428, opened 2024-10-10
Better Scenario error message surfacing


tonyandrewmeyer:
Wrapping operator-raised errors with UncaughtCharmError can be misleading. Reconsider that choice and possibly let the charm errors bubble up.

Given this code:

```py
class SomeCharm(ops.CharmBase):
    ...
    def _on_relation(self, event):
        assert event.something == some_sanity_check
```

I'd expect to run `pytest --pdb` and get dropped into this failing assertion line.

Instead, I'm getting:

```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /code/hexanator/.tox/unit/lib/python3.12/site-packages/scenario/runtime.py(469)exec()
-> raise UncaughtCharmError(
(Pdb)
```

And I can't get to the original line that triggered the exception.

Here's a cleaned up back trace for reference:

```
_pytest/python.py(162)pytest_pyfunc_call()
-> result = testfunction(**testargs)

tests/unit/test_scenario.py(56)test_rate_limit()
-> ctx.run(ctx.on.relation_created(relation), initial_state)

scenario/context.py(613)run()
-> with self._run_event(event=event, state=state) as ops:

contextlib.py(158)__exit__()
-> self.gen.throw(value)

scenario/context.py(598)_run_event()
-> with self._run(event=event, state=state) as ops:

contextlib.py(158)__exit__()
-> self.gen.throw(value)

scenario/context.py(660)_run()
-> with runtime.exec(

contextlib.py(158)__exit__()
-> self.gen.throw(value)

scenario/runtime.py(469)exec()
-> raise UncaughtCharmError(
```

Moved from https://github.com/canonical/ops-scenario/issues/76 and https://github.com/canonical/ops-scenario/issues/154

benhoyt:
@tonyandrewmeyer's proposal is just to drop UncaughtCharmError (after determining that it's not really used). If so, this can be a rainy-day (< half day) issue for someone.

[...]

PR #2106, opened 2025-10-15
feat(testing): don't inject UncaughtCharmError exception

TheJJ:
this exception provides no additional value, instead it just creates a intermediate exception referencing the original one.

to simplify debug output, just propagate all exceptions - so we have a single backtrace instead of two.

why is this useful? when running test with `pytest --pdb` you can directly inspect the real problem, instead of being dropped to the point where `raise UncaughtCharmError()` is called.

### DIFF
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -331,13 +331,6 @@ class Runtime:
                     if e.exit_code != 0:
                         raise
 
-            except (NoObserverError, ActionFailed):
-                raise  # propagate along
-            except Exception as e:
-                # The following is intentionally on one long line, so that the last line of pdb
-                # output shows the error message (pdb shows the "raise" line).
-                raise UncaughtCharmError(f'Uncaught {type(e).__name__} in charm, try "exceptions [n]" if using pdb on Python 3.13+. Details: {e!r}') from e  # fmt: skip
-
             finally:
                 if ops:
                     ops.destroy()


benhoyt:
Thanks for the contribution. We [wanted to remove this](https://github.com/canonical/operator/issues/1428), but unfortunately it's not a backwards compatible change (and several charm tests are expecting `UncaughtCharmError` to be raised). We tried a couple of different approaches, but in the end decided to just [improve the error message](https://github.com/canonical/operator/pull/1795) to mention pdb's `exceptions [n]` feature. Are you aware of that feature?

TheJJ:
Thanks, I see. I think that's quite inconvenient and can be done better :)

Another idea would be to add a new parameter (maybe `State.bare_exceptions=True` fits), with a default being set to False. Then we can deprecate the default value, and get rid of this behavior slowly and surely in the next bigger release.

What do you think about that? I've quickly implemented this and updated the pull request. Feel free to continue and adjust this idea as you like.

TheJJ:
Hm, the commit doesn't show up here since the pullreq is closed, i suppose. I've pushed it to the same branch, anyway.

dimaqq:
Discussed at the standup:
- if it's a kwarg, it belongs to Context, not State
- if it's a kwarg, we could make it an experimental feature, put this out there, and a few months later see how many charmers chose to use it
- an environment variable is an alternative for the feature flag, easier to deprecate

PR #2102, opened 2025-10-14
ci: temporarily disable tracing integration tests

dimaqq:
Current status is that:
- we can't force specific version of Juju (like 3.6.8), only select a track (which yields 3.6.9)
- Tempo 1/ track is broken on Juju 3.6.9
- Tempo 2/ track is only available as 2/edge

Ref: https://github.com/canonical/observability-stack/issues/110

This PR disables the tracing integration tests until a fix is released on 1/ channel of Tempo.

Discussion: https://matrix.to/#/!nHXpRkcSNJHlHdUGbQ:ubuntu.com/$Nhj8XxLNQXENt7I72FdB2IiVhUtRGGL6UQ0qNttRsOk?via=ubuntu.com&via=matrix.org&via=laquadrature.net

PR #2108, opened 2025-10-15
fix!: change JujuContext.machine_id from int to str

### LINKED ISSUE #2107: JujuContext creation fails when loading JUJU_MACHINE_ID
From [Matrix](https://matrix.to/#/!xdClnUGkurzjxqiQcN:ubuntu.com/$Bss03fOsKh7m42_eJZWvaGL5SWK3h2Kve02BGQdg3Zw?via=ubuntu.com&via=matrix.org&via=sft.lol):

I have an issue with one charm that I've already written, avaialble here https://github.com/dell-esg/powermax-cinder-charm
When deploying it, the following error shows up in the logs

```
'unit-cinder-powermax1646-2: 16:10:03 WARNING unit.cinder-powermax1646/2.install   File "/var/lib/juju/agents/unit-cinder-powermax1646-2/charm/venv/ops/jujucontext.py", line 256, in _from_dict
unit-cinder-powermax1646-2: 16:10:03 WARNING unit.cinder-powermax1646/2.install     machine_id=int(env['JUJU_MACHINE_ID']) if env.get('JUJU_MACHINE_ID') else None,
unit-cinder-powermax1646-2: 16:10:03 WARNING unit.cinder-powermax1646/2.install ValueError: invalid literal for int() with base 10: '0/lxd/4'
```

tonyandrewmeyer:
The `JUJU_MACHINE_ID` context variable sometimes looks like an integer, but is actually typed as a string, and can be a non-integer string in some cases.

The field in `JujuContext` is changed, as well as the corresponding field in `testing.Context`.

Fixes #2107

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '42 7 25 * *'
+  pull_request:  # temporary
 
 permissions: {}
 
@@ -15,7 +16,11 @@ jobs:
       fail-fast: false
       matrix:
         preset: ['k8s', 'microk8s']
-        test: ['test_direct_connection', 'test_with_tls', 'test_relation_units']
+        test:
+          # https://github.com/canonical/observability-stack/issues/110
+          # - test_direct_connection
+          # - test_with_tls
+          - test_relation_units
 
     steps:
       - run: |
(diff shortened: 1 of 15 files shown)


ops/jujucontext.py
```diff
@@ -253,7 +253,7 @@ def _from_dict(cls, env: Mapping[str, Any]) -> JujuContext:
             ),
             dispatch_path=env.get('JUJU_DISPATCH_PATH', ''),
             hook_name=env.get('JUJU_HOOK_NAME', ''),
-            machine_id=int(env['JUJU_MACHINE_ID']) if env.get('JUJU_MACHINE_ID') else None,
+            machine_id=env['JUJU_MACHINE_ID'] if env.get('JUJU_MACHINE_ID') else None,
```
benhoyt: I think this could be shortened to `env.get('JUJU_MACHINE_ID') or None` like we have with `JUJU_NOTICE_ID` below.

tonyandrewmeyer:
```suggestion
            machine_id=env.get('JUJU_MACHINE_ID') or None,
```

james-garner-canonical (approved):
This is technically a breaking change for anyone that was only running on machines with `int` IDs, and started relying on `machine_id` being an `int`, but it's pretty new surface area so hopefully it doesn't break anyone.

I guess a super tox run might not catch any breakages, but might be worth a try.

I assuming we're going to do a hotfix release when this lands?

ops/jujucontext.py
tonyandrewmeyer: I'm unsure about the `or None`. If it's not in the dictionary, then it'll be `None` via the `get` default, so the `or None` would be used if it was in the environment as an empty string. If Juju provides the ID as an empty string, perhaps we should be leaving that for the user?

I'm not sure why we do this with `JUJU_NOTICE_ID` - do we expect it to be present when it doesn't have a value?

benhoyt: I'm fine with or without the `or None` here. It's kind of a defensive programming thing: just in case Juju sends it as empty string instead of not present.

tonyandrewmeyer:
I believe the test failures will be fixed by #2102.

benhoyt:
I just approved and merged #2102

PR #2116, opened 2025-10-17
refactor: use ops.hookcmds in _ModelBackend

tonyandrewmeyer:
This PR removes the subprocess calls from model.py's `_ModelBackend`, replacing them with using the new ops.hookcmds package.

Other changes:
* Fix the `_NetworkDict` definition (several fields were missing). This introduces use of `NotRequired` so it also moves into the `TYPE_CHECKING` block.
* `network-get` has the `--format=json` argument first, as the other commands do, and as the documentation suggests.
* `opened-ports` no longer raises if there's an unexpected protocol, because model only warns about this, so we need the lower-level to accept other protocols for backwards compatibility.
* `relation-list` returns a `str` when `--app` is used, and a `list[str] `otherwise.
* Adjust the order of `--app` in `relation-set` to match `relation-get` (this is just for more consistent looking commands).
* Fix a comment typo that @benhoyt noticed in the previous PR.
* Adds a missing default argument in `secret-revoke`.
* A lot of the network fields have tests for them being missing, so I'm assuming that might happen (I think it's actually in very old Juju, but it seems good to be conservative) and loading the data class is adjusted to handle that.
* Drop support for testing on Windows (this was simpler than adjusting the batch scripts, and we already gave up on Windows).
* Drops support for using `relation-get` on Juju 2.7 and 2.8 where there's no `--app` argument (and the corresponding tests). I believe our minimum supported Juju is 2.9 with Ops 3.x.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -4245,7 +4245,7 @@ class TestPorts:
         ]
 
     def test_opened_ports(self, fake_script: FakeScript, unit: ops.Unit):
-        fake_script.write('opened-ports', """echo [\\"8080/tcp\\", \\"icmp\\"]""")
+        fake_script.write('opened-ports', """echo '["8080/tcp", "icmp"]'""")
 
         ports_set = unit.opened_ports()
         assert isinstance(ports_set, set)
@@ -4265,9 +4265,7 @@ class TestPorts:
     def test_opened_ports_warnings(
         self, caplog: pytest.LogCaptureFixture, fake_script: FakeScript, unit: ops.Unit
     ):
-        fake_script.write(
-            'opened-ports', """echo [\\"8080/tcp\\", \\"1234/ftp\\", \\"1000-2000/udp\\"]"""
-        )
+        fake_script.write('opened-ports', """echo '["8080/tcp", "1234/ftp", "1000-2000/udp"]'""")
 
         with caplog.at_level(level='WARNING', logger='ops.model'):
             ports_set = unit.opened_ports()
@@ -4306,7 +4304,7 @@ class TestPorts:
         # Two open ports, leave one alone and open another one.
         fake_script.write('open-port', 'exit 0')
         fake_script.write('close-port', 'exit 0')
-        fake_script.write('opened-ports', 'echo [\\"8025/tcp\\", \\"8028/tcp\\"]')
+        fake_script.write('opened-ports', """echo '["8025/tcp", "8028/tcp"]'""")
         unit.set_ports(ops.Port('udp', 8022), 8028)
         assert fake_script.calls(clear=True) == [
             ['opened-ports', '--format=json'],
@@ -4317,7 +4315,7 @@ class TestPorts:
     def test_set_ports_replace(self, fake_script: FakeScript, unit: ops.Unit):
         fake_script.write('open-port', 'exit 0')
         fake_script.write('close-port', 'exit 0')
-        fake_script.write('opened-ports', 'echo [\\"8025/tcp\\", \\"8028/tcp\\"]')
+        fake_script.write('opened-ports', """echo '["8025/tcp", "8028/tcp"]'""")
         unit.set_ports(8001, 8002)
         calls = fake_script.calls(clear=True)
         assert calls.pop(0) == ['opened-ports', '--format=json']
@@ -4332,7 +4330,7 @@ class TestPorts:
     def test_set_ports_close_all(self, fake_script: FakeScript, unit: ops.Unit):
         fake_script.write('open-port', 'exit 0')
         fake_script.write('close-port', 'exit 0')
-        fake_script.write('opened-ports', 'echo [\\"8022/udp\\"]')
+        fake_script.write('opened-ports', """echo '["8022/udp"]'""")
         unit.set_ports()
         assert fake_script.calls(clear=True) == [
             ['opened-ports', '--format=json'],
@@ -4342,7 +4340,7 @@ class TestPorts:
     def test_set_ports_noop(self, fake_script: FakeScript, unit: ops.Unit):
         fake_script.write('open-port', 'exit 0')
         fake_script.write('close-port', 'exit 0')
-        fake_script.write('opened-ports', 'echo [\\"8000/tcp\\"]')
+        fake_script.write('opened-ports', """echo '["8000/tcp"]'""")
         unit.set_ports(ops.Port('tcp', 8000))
         assert fake_script.calls(clear=True) == [
             ['opened-ports', '--format=json'],


ops/hookcmds/_port.py
```diff
@@ -171,7 +171,8 @@ def opened_ports(*, endpoints: bool = False) -> list[Port]:
         else:
             protocol = protocol or 'tcp'
             port = int(port)
-        if protocol not in ('tcp', 'udp', 'icmp'):
-            raise RuntimeError(f'Unexpected protocol from Juju: {protocol}')
-        ports.append(Port(protocol=protocol, port=port, to_port=to_port, endpoints=port_endpoints))
+        # The type: ignore is required because we know that protocol will be tcp, udp, or icmp
+        # but we can't raise if not, because model.py only emits a warning in that case, and
+        # we need to maintain backwards-compatibility.
```
tonyandrewmeyer:
```suggestion
        # we need to maintain backwards compatibility.
```

ops/hookcmds/_other.py
```diff
@@ -164,11 +164,11 @@ def network_get(binding_name: str, *, relation_id: int | None = None) -> Network
         binding_name: A name of a binding (relation name or extra-binding name).
         relation_id: An optional relation id to get network info for.
     """
-    args: list[str] = []
+    args: list[str] = ['--format=json']
```
benhoyt: Nit: you can probably drop the `: list[str]` now.

ops/_private/harness.py
```diff
@@ -75,6 +75,8 @@
         class State:
             pass
 
+    from ..model import _NetworkDict
+
```
james-garner-canonical: Nit: move this to the top of the `typing.TYPE_CHECKING` block?

ops/hookcmds/_relation.py
```diff
@@ -124,7 +129,7 @@ def relation_list(id: int | None = None, *, app: bool = False) -> list[str]:
     if id is not None:
         args.extend(['-r', str(id)])
     stdout = run('relation-list', *args)
-    result = cast('list[str]', json.loads(stdout))
+    result = cast('str', json.loads(stdout)) if app else cast('list[str]', json.loads(stdout))
     return result
```
james-garner-canonical, line 133: I'd drop these `cast`s entirely, the result of `json.loads` is typed as `Any` so it shouldn't be needed to satisfy the type checker. So the `cast`s are just for the maintainers, and I think it's more maintainable without them.
```suggestion
    return json.loads(stdout)
```

ops/hookcmds/_types.py
```diff
@@ -54,9 +54,9 @@ class Address:
     @classmethod
     def _from_dict(cls, d: AddressDict) -> Address:
         return cls(
-            hostname=d['hostname'],
-            value=d['value'],
-            cidr=d['cidr'],
+            hostname=d.get('hostname', ''),
+            value=d.get('value', ''),
+            cidr=d.get('cidr', ''),
```
james-garner-canonical, line 72: Does this mean that `AddressDict` should be `total=False`?

ops/hookcmds/_port.py
```diff
@@ -171,7 +171,8 @@ def opened_ports(*, endpoints: bool = False) -> list[Port]:
...
+        ports.append(Port(protocol=protocol, port=port, to_port=to_port, endpoints=port_endpoints))  # type: ignore
```
james-garner-canonical:
```suggestion
        port = Port(
            protocol=protocol,  # type: ignore
            port=port,
            to_port=to_port,
            endpoints=port_endpoints,
        )
        ports.append(port)
```

ops/hookcmds/_types.py
```diff
@@ -54,14 +54,14 @@ class Address:
...
         )
 
 
 BindAddressDict = TypedDict(
-    '_BindAddressDict',
+    'BindAddressDict',
     {'mac-address': str, 'interface-name': str, 'addresses': list[dict[str, str]]},
```
james-garner-canonical: This seems like it would avert the need to `cast` on line 80.
```suggestion
    {'mac-address': str, 'interface-name': str, 'addresses': list[AddressDict]},
```

ops/hookcmds/_types.py
```diff
@@ -77,11 +77,11 @@ class BindAddress:
     @classmethod
     def _from_dict(cls, d: BindAddressDict) -> BindAddress:
         addresses = [
-            Address._from_dict(cast('AddressDict', addr)) for addr in d.get('addresses', [])
+            Address._from_dict(cast('AddressDict', addr)) for addr in d.get('addresses') or []
```
james-garner-canonical: `d.get('addresses') or []` suggests:
1. `BindAddressDict` should be `total=False`
2. We need a comment as to why `d.get('addresses', [])` wouldn't work -- what other Falsey values can we get here, and should the type annotations be updated to document them?

ops/hookcmds/_relation.py
```diff
@@ -103,7 +104,11 @@ def relation_ids(name: str) -> list[str]:
...
 
 
-def relation_list(id: int | None = None, *, app: bool = False) -> list[str]:
+@overload
+def relation_list(id: int | None = None, *, app: Literal[True]) -> str: ...
+@overload
+def relation_list(id: int | None = None, *, app: Literal[False] = False) -> list[str]: ...
+def relation_list(id: int | None = None, *, app: bool = False) -> str | list[str]:
```
dimaqq, line 111: Maybe @james-garner-canonical  can weigh in on this, is an aiddtional overload needed for "app kwarg not included in the call"?

ops/hookcmds/_types.py
```diff
@@ -54,14 +54,14 @@ class Address:
     @classmethod
     def _from_dict(cls, d: AddressDict) -> Address:
         return cls(
-            hostname=d['hostname'],
-            value=d['value'],
-            cidr=d['cidr'],
+            hostname=d.get('hostname', ''),
+            value=d.get('value', ''),
+            cidr=d.get('cidr', ''),
```
dimaqq, line 72: Is this for compatibility / not breaking the existing users?
Or should the fields be typed as `str | None` instead?

ops/hookcmds/_relation.py
james-garner-canonical: That's covered by this one, because it has a default:
```py
def relation_list(id: int | None = None, *, app: Literal[False] = False) -> list[str]: ...
```

james-garner-canonical (commented):
Posting a partial review so it will let me reply to @dimaqq's comment.

ops/hookcmds/_types.py
```diff
@@ -232,7 +232,7 @@ class Network:
     def _from_dict(cls, d: dict[str, Any]) -> Network:
         bind: list[BindAddress] = [
             BindAddress._from_dict(bind_data)
-            for bind_data in cast('list[BindAddressDict]', d['bind-addresses'])
+            for bind_data in cast('list[BindAddressDict]', d.get('bind-addresses', []))
         ]
```
james-garner-canonical: How about:
```python
bind_dicts: list[BindAddressDict] = d.get('bind-addresses', [])
bind = [BindAddress._from_dict(bind_dict) for bind_dict in bind_dicts]
```

test/test_main.py
```diff
@@ -976,7 +976,9 @@ def test_sets_model_name(self, fake_script: FakeScript):
     @pytest.mark.usefixtures('setup_charm')
     def test_has_valid_status(self, fake_script: FakeScript):
         fake_script.write('action-get', "echo '{}'")
-        fake_script.write('status-get', """echo '{"status": "unknown", "message": ""}'""")
+        fake_script.write(
+            'status-get', """echo '{"status": "unknown", "message": "", "status-data": {}}'"""
+        )
```
james-garner-canonical, line 981: Is the new data needed for the test to pass? If not, WDYT about parametrizing the test instead?
Or is it guaranteed that `status-data` is present?

test/test_model.py
```diff
@@ -1094,7 +1094,7 @@ def test_storage(self, fake_script: FakeScript, fake_juju_version: None):
         fake_script.write(
             'storage-list',
             """
-            if [ "$1" = disks ]; then
+            if [ "$2" = disks ]; then
```
james-garner-canonical, line 1097: It looks like most of the changes to this file are due to arguments being in different orders, which seems fine.

test/test_model.py
```diff
@@ -2906,65 +2906,6 @@ def test_relation_tool_errors(self, fake_script: FakeScript, monkeypatch: pytest
                 run()
             assert fake_script.calls(clear=True) == calls
 
-    @pytest.mark.parametrize('version', ['2.8.0', '2.7.0'])
-    def test_relation_get_juju_version_quirks(
```
james-garner-canonical, line 2912: Why do these tests go away?

tonyandrewmeyer:
I added code like:

```python
    with real_open('run.log', 'a', encoding='utf-8') as f:
        print(json.dumps({'args': args, 'input': input}), file=f)
```

To hookcmds/_utils.py's `run()` and model.py's `ModelBackend.run()` in this branch and in main@HEAD.

There are some expected changes, so I post-processed the main@HEAD one with:

```python
    for line in content:
        if line['args'][0] in ('action-get', 'is-leader', 'relation-list', 'relation-ids', 'storage-get', 'storage-list', 'secret-get'):
            line['args'].insert(1, '--format=json')
        if line['args'][0] in ('status-set', 'action-log', 'action-fail'):
            line['args'].insert(-1, '--')
        if line['args'][0] == 'storage-get':
            line['args'].pop(-1)  # We always get all attributes.
        if line['args'][0] == 'secret-set' and '--owner' not in line['args']:
            line['args'].insert(1, 'application')
            line['args'].insert(1, '--owner')
```

The resulting changes are:

```diff
$ diff -c run.log ../record-run-calls/run-adjusted.log 
*** run.log	2025-10-17 14:43:40.400268670 +1300
--- ../record-run-calls/run-adjusted.log	2025-10-17 15:18:30.549980805 +1300
***************
*** 1,4 ****
! {"args": ["relation-ids", "req1", "--format=json"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
--- 1,4 ----
! {"args": ["relation-ids", "--format=json", "req1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
***************
*** 29,39 ****
  {"args": ["action-get", "--format=json"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed", "key#file=/tmp/tmpyxt0hlub/key"], "input": null}
  {"args": ["secret-get", "--format=json", "secret:changed", "--peek"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed", "key#file=/tmp/tmprw_6etsx/key"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed", "key#file=/tmp/tmpssr2evkg/key"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
  {"args": ["status-set", "--application=True", "blocked", "--", "first"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
--- 29,39 ----
  {"args": ["action-get", "--format=json"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc", "key#file=/tmp/tmpeiaq076y/key"], "input": null}
  {"args": ["secret-get", "--format=json", "secret:changed", "--peek"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc", "key#file=/tmp/tmpzl_0gnso/key"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc", "key#file=/tmp/tmpeomc4x7m/key"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
  {"args": ["status-set", "--application=True", "blocked", "--", "first"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
```

Those boil down to two things, which are simple to visually inspect rather than handle in the script:
* `relation-ids` has the name and `--format=json` in a different order (`hookcmds` matches the docs)
* `secret-set` has the description in a different order in the args, and the file is different (because it's a tempfile specific to the run).

So I'm confident that for the use covered by all our unit tests there aren't any unexpected changes in the subprocess calls.

ops/model.py
```diff
@@ -3613,69 +3619,16 @@ def __init__(
...
-                else:
-                    text: str = result.stdout  # type: ignore
-                    if use_json:
-                        return json.loads(text)  # type: ignore
-                    else:
-                        return text  # type: ignore
+                span.set_attribute('argv', args)
+                yield span
+            else:
+                yield contextlib.nullcontext()
```
dimaqq: **issue:** 

The only case for not having a span, is when we're calling `juju-log` in which case current code in main uses a context manager that doesn't start any span and returns `None`.

ops/_private/harness.py
tonyandrewmeyer: Why? The only thing above it is also an import statement, also from within ops.

ops/hookcmds/_types.py
tonyandrewmeyer: It's for compatibility with old Juju. I'm fairly sure that Juju 2.9+ will have these values, but we have tests for them being missing (which is why they've changed here), and it seems overly risky crash if they are missing. However, in terms of usability, a ` | None` is very annoying, because you need to put in a bunch of `assert x is not None` code, when you really know that with your modern Juju (covered by your integration tests) you will always have something there.

ops/hookcmds/_types.py
tonyandrewmeyer: There's an existing test that provides a `None` value in `addresses`. Similar to the above, I don't see how this can be the case in Juju 2.9, but don't want to risk some missed case and breaking things, so I would rather be cautious and allow it. I can add a comment, I guess.

ops/hookcmds/_relation.py
tonyandrewmeyer: I don't agree here (and it's really outside of the scope of this PR as well, because in the previous PR all of those casts were added, and (through the review process) made consistent across all of the commands). I think it's much clearer for the maintainer to know what shape the data from Juju is coming as.

If you feel strongly about this, please open an issue to have it changed consistently across the entire package, and the team can discuss that in that other issue.

ops/_private/harness.py
james-garner-canonical: This is what I had in mind:

1. We typically alphabetize imports, though I guess the linter doesn't care here
2. We typically start with our unconditional imports (if possible), and then do conditional stuff later
3. It's easier to see the scope/indentation level of a statement when it's near the top of an indented block. Putting the short statement at the top makes it easy for both imports, while having it after some more indented blocks makes it harder.

ops/hookcmds/_types.py
james-garner-canonical: Rather than a comment, would it make sense to change the type of `BindAddressDict.addresses` to be `None`-able, so we always catch this?

ops/hookcmds/_types.py
```diff
@@ -54,16 +64,14 @@ class Address:
...
+            cidr=d.get('cidr', ''),
         )
 
-
-BindAddressDict = TypedDict(
-    '_BindAddressDict',
-    {'mac-address': str, 'interface-name': str, 'addresses': list[dict[str, str]]},
-)
+    @property
+    def address(self) -> str:
```
benhoyt: Why is this alias being added? It seems to me we should we call it either `value` or `address` (the latter seems clearer?) and avoid having an alias for new code/types.

test/test_charm.py
```diff
@@ -313,10 +313,10 @@ def _on_stor4_attach(self, event: ops.StorageAttachedEvent):
...
         'storage-get',
         """
-        if [ "$1" = "-s" ]; then
-            id=${2#*/}
-            key=${2%/*}
-            echo "\\"/var/srv/${key}/${id}\\"" # NOQA: test_quote_backslashes
+        if [ "$2" = "-s" ]; then
+            id=${3#*/}
+            key=${3%/*}
+            echo "{\\"kind\\": \\"filesystem\\", \\"location\\": \\"/var/srv/${key}/${id}\\"}"
```
benhoyt, line 319: Why does this output change?

test/test_main.py
```diff
@@ -976,7 +976,9 @@ def test_sets_model_name(self, fake_script: FakeScript):
     @pytest.mark.usefixtures('setup_charm')
     def test_has_valid_status(self, fake_script: FakeScript):
         fake_script.write('action-get', "echo '{}'")
-        fake_script.write('status-get', """echo '{"status": "unknown", "message": ""}'""")
+        fake_script.write(
+            'status-get', """echo '{"status": "unknown", "message": "", "status-data": {}}'"""
```
benhoyt, line 980: Is the new `status-data` field because we always specify `--include-data` now?

test/test_model.py
```diff
@@ -1104,10 +1104,10 @@ def test_storage(self, fake_script: FakeScript, fake_juju_version: None):
         fake_script.write(
             'storage-get',
             """
-            if [ "$2" = disks/0 ]; then
-                echo '"/var/srv/disks/0"'
-            elif [ "$2" = disks/1 ]; then
-                echo '"/var/srv/disks/1"'
+            if [ "$3" = disks/0 ]; then
+                echo '{"kind": "filesystem", "location": "/var/srv/disks/0"}'
```
benhoyt, line 1108: Why does this output change?

ops/model.py
```diff
@@ -3613,69 +3602,16 @@ def __init__(
...
-            }
-            if input_stream:
-                kwargs.update({'input': input_stream})
-            which_cmd = shutil.which(args[0])
-            if which_cmd is None:
-                raise RuntimeError(f'command not found: {args[0]}')
-            args = (which_cmd,) + args[1:]
-            if use_json:
-                args += ('--format=json',)
+    def _tracing_span(self, cmd: str, *args: Any):
```
benhoyt: So the need for the recursion prevention goes away completely, because hookcmds never logs? That's a nice side simplification!

ops/model.py
```diff
@@ -3704,20 +3640,24 @@ def _is_relation_not_found(model_error: Exception) -> bool:
...
-            )
-            return typing.cast('list[str]', rel_list)
-        except ModelError as e:
-            if self._is_relation_not_found(e):
+            with self._tracing_span('relation-list', relation_id):
+                return hookcmds.relation_list(relation_id)
+        except hookcmds.Error as e:
+            self._check_for_security_event('relation-list', e.returncode, e.stderr)
+            err = ModelError(e.stderr)
+            if self._is_relation_not_found(err):
```
benhoyt: This is a bit awkward. Could we change the `_is_relation_not_found` and similar helpers just to take the string and pass `e.stderr` directly, and then `raise ModelError` below?

ops/model.py
```diff
@@ -3733,20 +3673,14 @@ def relation_remote_app_name(self, relation_id: int) -> str | None:
...
         try:
-            rel_id = self._run(
-                'relation-list', '-r', str(relation_id), '--app', return_output=True, use_json=True
-            )
-            # if it returned anything at all, it's a str.
-            return typing.cast('str', rel_id)
-
-        except ModelError as e:
-            if self._is_relation_not_found(e):
+            with self._tracing_span('relation-list', relation_id, True):
```
benhoyt: Does this mean the names of the arguments will be lost? I'm not saying we should try to put `-r` in here just for tracing, but maybe we should use a kwargs or a dict if that's possible, like `relation_id=relation_id, app=True` so they appear more human-readable in the tracing output? Or maybe positional args where the hookcmds method has positional, and kwarg otherwise, like `relation_id, app=True` here.

CC: @dimaqq

ops/model.py
```diff
@@ -3784,32 +3716,35 @@ def relation_set(self, relation_id: int, data: Mapping[str, str], is_app: bool)
...
-            raise
+            raise err from e
+        return dataclasses.asdict(raw)
 
     def config_get(self) -> dict[str, bool | int | float | str]:
-        out = self._run('config-get', return_output=True, use_json=True)
-        return typing.cast('dict[str, bool | int | float | str]', out)
+        try:
+            with self._tracing_span('config-get'):
+                return dict(hookcmds.config_get())
```
benhoyt: The `dict(...)` is awkward, when it's actually just a dict already. I wonder if we should simplify hookcmds return values and just have them return plain `dict` instead of mapping. Seems better to "take abstract types as parameters, use concrete types as return values". It's never going to be anything but a `dict` anyway.

ops/model.py
```diff
@@ -3907,16 +3837,31 @@ def status_set(
             raise TypeError('message parameter must be a string')
         if status not in _SETTABLE_STATUS_NAMES:
             raise InvalidStatusError(f'status must be in {_SETTABLE_STATUS_NAMES}, not {status!r}')
-        self._run('status-set', f'--application={is_app}', status, message)
+        try:
+            with self._tracing_span('status-set', status, message, is_app):
+                hookcmds.status_set(status, message, app=is_app)
```
benhoyt: So many of these are basically the same: with tracing-span, run hookcmd method, try/except hookcmds.Error, check for security event, convert hookcmds.Error to ModelError. What about something more like this:

```python
with self._trace_and_wrap_errors('status-set, status, message, app=is_app):
    hookcmds.status_set(status, message, app=is_app)
```

Perhaps with an `_is_secret=True` option to handle the `SecretNotFound` case.

benhoyt:
Also, thanks for the diff tests @tonyandrewmeyer. That really helps!

ops/hookcmds/_other.py
tonyandrewmeyer:
```suggestion
    args = ['--format=json']
```

ops/hookcmds/_relation.py
james-garner-canonical: Per conversation, I'm now convinced that we should keep this as-is, since we use the pattern everywhere; and sometimes the value isn't immediately returned, so it's useful documentation of the Juju return type; and it's nice for the functions to all follow the same pattern.

james-garner-canonical (commented):
Another partial review so I can reply.

ops/hookcmds/_types.py
tonyandrewmeyer: Done as part of unifying the typeddicts.

test/test_charm.py
tonyandrewmeyer: We don't provide a `storage_get()` that only gets a single attribute, because there are only two, so it seemed more sensible (discussed in the spec and the previous PR) to always get both and then use `.location` or `.kind`. So we (a) always use `--format=json` now, and (b) never ask for a specific key. That means that the mocked response from Juju is a dict with both keys, not a single string with just the location, even though that's all that's used in the test.

test/test_main.py
tonyandrewmeyer: We always specify `--include-data` already, but don't do anything with the `status-data` in the response (a strange choice and I haven't gone back through the history to try to understand why we do that).

The `status-data` field is in the mocked response because now we actually read it in to the dataclass that's returned (even though `_ModelBackend` still doesn't do anything with it).

test/test_model.py
tonyandrewmeyer: This is the same as the storage example above. We (a) always use `--format-json` now, and (b) always get all the attributes and let callers access the one(s) they care about.

ops/model.py
tonyandrewmeyer: This feels a bit of a scope creep. What about doing this in a follow-up issue/PR instead? We don't get the names of the arguments now, just the values.

test/test_main.py
tonyandrewmeyer: With both main@HEAD and in the branch, the actual Juju command would return the new data. With main, it would be ignored, so the test could be a bit sloppy and not include it. With the branch, it's always used, so needs to be here, and parametrising would not work.

test/test_model.py
tonyandrewmeyer: I should have put that in the description, sorry - I'll add it there now.

I dropped a small amount of code, and these tests, that were only there to support Juju older than 2.9, when relation-get did not have an `--app` argument.

ops/hookcmds/_types.py
tonyandrewmeyer: Yes, that was the wrong fix, it was meant to be in the loading. Changed.

test/test_model.py
tonyandrewmeyer: Yes, mostly the addition of `--format=json`.

tonyandrewmeyer:
> I think it might be better to move the order of `--format=json` and `--app` out of this PR. It causes quite a bit of unrelated noise here.

I think you may be overestimating how much of it is the couple of minor changes in this PR and how much is simply changing to using hookcmds, where the order was set in the previous PR (and also a lot of the noise is actually just because we are *using* `--format=json` in cases where we were not before). We could perhaps have another PR before this one that changed the hookcmds ordering to be more similar to what `_ModelBackend` does, but I don't think that's the right thing to do.

> We may or may not do it in a separate PR (though I'd argue it's probably not worth it, as the placement of those args doesn't matter).

Well, that's true in practice now, but it's not following the documented usage of the hook commands (which is why they are ordered the way they are in the previous PR). For example, [relation-ids](https://documentation.ubuntu.com/juju/3.6/reference/hook-command/list-of-hook-commands/relation-ids/):

> `relation-ids [options] <name>`

That doesn't say that `relation-ids <name> [options]` is valid, although it is at the moment, and I think it's likely that changing that in a minor release wouldn't be considered acceptable. However, I do pretty strongly believe that when adding our new public interface to the hook commands we should follow the documented order, not hope that any order continues to work.

ops/model.py
benhoyt: I believe previously we would have got the raw arguments, including `-r` and `--app`: `['relation-list', '-r', '1234', '--app']`, which seems a lot clearer than `['relation-list', '1234', True]`.

benhoyt:
> I think you may be overestimating how much of it is the couple of minor changes in this PR and how much is simply changing to using hookcmds, where the order was set in the previous PR (and also a lot of the noise is actually just because we are using --format=json in cases where we were not before). We could perhaps have another PR before this one that changed the hookcmds ordering to be more similar to what _ModelBackend does, but I don't think that's the right thing to do.

Okay, fair enough -- I'm happy to continue with this as you have it now.

ops/model.py
tonyandrewmeyer: Ah, that's true, yes. I was thinking of the arguments to `_run()`, rather than the command argument to `subprocess.run`.

I'd still rather do this in a follow-up PR, but could do one immediately after this.

ops/model.py
```diff
@@ -95,27 +96,15 @@
...
     '_NetworkDict',
     {
-        'bind-addresses': List[_BindAddressDict],
-        'ingress-addresses': List[str],
-        'egress-subnets': List[str],
+        'bind-addresses': list['BindAddressDict'],
+        'ingress-addresses': list[str],
+        'egress-subnets': list[str],
     },
 )
```
james-garner-canonical, line 108: I wonder if this should move to `hookcmds._types` as well.

ops/model.py
```diff
@@ -69,6 +67,9 @@
 if typing.TYPE_CHECKING:
     from typing_extensions import TypeAlias
 
+    from .hookcmds._types import AddressDict, BindAddressDict
```
james-garner-canonical: I wonder if these names should be/stay private to avoid expanding the public surface of this module.

We're importing them from a private module anyway, so making the names private too seems like it doesn't introduce any new complications on that front.

ops/model.py
james-garner-canonical: Maybe `self._raise_if_relation_not_found(msg=e.stderr)`?

ops/model.py
```diff
@@ -3784,32 +3716,35 @@ def relation_set(self, relation_id: int, data: Mapping[str, str], is_app: bool)
...
+            with self._tracing_span('relation-model-get', relation_id):
+                raw = hookcmds.relation_model_get(relation_id)
+        except hookcmds.Error as e:
+            self._check_for_security_event('relation-model-get', e.returncode, e.stderr)
+            err = ModelError(e.stderr)
+            if self._is_relation_not_found(err):
                 raise RelationNotFoundError() from e
-            raise
+            raise err from e
+        return dataclasses.asdict(raw)
```
james-garner-canonical, line 3730: Out of curiosity, what's the philosophy for which hook commands return a data class, and which return a dict?

ops/model.py
james-garner-canonical: I think this would be quite a nice pattern even if it didn't grow extra arguments to handle the slightly different cases, and those cases just stayed as is (e.g. secrets, relations).

Another alternative to avoid extra arguments would be specific context managers like `_trace_and_wrap_for_secret` which just wraps `_trace_and_wrap_errors` and adds the extra handling.

ops/model.py
```diff
@@ -3990,16 +3975,30 @@ def network_get(self, binding_name: str, relation_id: int | None = None) -> _Net
...
+            'bind-addresses': [
+                {
+                    'mac-address': b_addr.mac_address,
+                    'interface-name': b_addr.interface_name,
+                    'addresses': [
+                        {'value': addr.value, 'cidr': addr.cidr, 'hostname': addr.hostname}
+                        for addr in b_addr.addresses
+                    ],
+                }
+                for b_addr in raw.bind_addresses
```
james-garner-canonical, line 3931: I wonder if this particular chunk of logic belongs on the `BindAddress` class.

Actually, wouldn't `dataclasses.asdict` just work here, since [it recurses](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) into attributes that are also `dataclasses`?

ops/model.py
tonyandrewmeyer: Sure. I'd had some discussions with @james-garner-canonical previously around how to handle this - in particular, without losing the typing information, and I wanted to avoid using generics if possible. I left it very explict and duplicative to see what reviewers thought, but I'm happy to cut it down.

ops/model.py
```diff
@@ -4059,47 +4066,39 @@ def secret_get(
...
-            result = self._run('secret-get', *args, return_output=True, use_json=True)
-        except ModelError as e:
-            if 'not found' in str(e):
-                raise SecretNotFoundError() from e
-            raise
-        return typing.cast('dict[str, str]', result)
+            # The type: ignore here is because the type checker can't tell that
+            # we will always have refresh or peek but not both.
+            with self._tracing_span('secret-get', id, label, refresh, peek):
+                return hookcmds.secret_get(id=id, label=label, refresh=refresh, peek=peek)  # type: ignore
```
james-garner-canonical: How do we enforce only having one of refresh of peek be true?
That is, why is the type checker wrong?

Important: we should make sure that we raise the same error as we would have previously if both are set to true.

Minor: would this be clearer with the ignores more tightly scoped?
```suggestion
            with self._tracing_span('secret-get', id, label, refresh, peek):
                return hookcmds.secret_get(
                    id=id,
                    label=label,
                    # secret_get will raise if both of these are True
                    refresh=refresh,  # type: ignore
                    peek=peek,   # type: ignore
                )
```

ops/model.py
```diff
@@ -4059,47 +4066,39 @@ def secret_get(
...
+            if 'not found' in str(err):
+                raise SecretNotFoundError() from err
+            raise err from e
 
-    def _run_for_secret(
-        self, *args: str, return_output: bool = False, use_json: bool = False
-    ) -> str | Any | None:
+    def secret_info_get(self, *, id: str | None = None, label: str | None = None) -> SecretInfo:
+        # The type: ignore here is because the type checker can't tell, even
+        # with local overloads, that either id or label must be provided.
```
james-garner-canonical, line 4008: The reason the overloads don't help here is because they only apply to callers of the overloaded function -- the type checker doesn't make use of them when type checking the definition of the overloaded function.

ops/model.py
tonyandrewmeyer: > I'd still rather do this in a follow-up PR, but could do one immediately after this.

I changed my mind and did this as part of moving the error wrapping code.

ops/model.py
tonyandrewmeyer: No longer relevant with the error transformation moved.

ops/model.py
james-garner-canonical: I guess we enforce this where it's called internally?

ops/hookcmds/_types.py
tonyandrewmeyer: I feel like that's a much worse experience for the user, because they need to then handle the `None` case in their code with a cast or assert, even though in reality we expect this to *always* be present.

I also feel like a charm is not going to handle "empty list" any differently than "addresses weren't provided", so an empty list is a reasonable response.

ops/model.py
tonyandrewmeyer: > I wonder if this should move to `hookcmds._types` as well.

Unused, just for consistency? I'm not a fan of that. Or do you think I should be using it in `hookcmds` somewhere?

ops/model.py
tonyandrewmeyer: > I wonder if these names should be/stay private to avoid expanding the public surface of this module.

Yes, I missed that, thanks.

ops/model.py
james-garner-canonical: It seems odd that all the other dict equivalents to the dataclasses are defined there, but this one isn't. As you say, I guess it's because this one doesn't happen to be used internally? It just seemed tidier to have them all defined in the same place, but they're internal and this type is internal, so I suppose it's not very important either way.

ops/model.py
tonyandrewmeyer: > Out of curiosity, what's the philosophy for which hook commands return a data class, and which return a dict?

If we know the structure, then a data class. If the structure is user-defined (relation-get, config-get), then a dict.

james-garner-canonical: Makes sense, thanks for explaining.

ops/hookcmds/_types.py
james-garner-canonical: I agree that we shouldn't expose the `None`-able-ness to users. But isn't `BindAddressDict` only used internally? Typing `BindAddressDict.addresses` with `| None` would just ensure that we get an error if someone tries to change this to `d.get('addresses', [])` someday, and they'll have to explicitly decide to change the type if we believe it will never be `None`.

ops/model.py
tonyandrewmeyer: > How do we enforce only having one of refresh of peek be true? That is, why is the type checker wrong?

Juju enforces it. The `hookcmds` philosophy is that we leave error checking to Juju, rather than duplicating checks. In `hookcmds` we provide overloads that will show an error if you provide both=True. I believe that doesn't work here because `_ModelBackend.secret_get` doesn't also have those overloads, so the type checker can't tell that you aren't passing both=True.

I could fix this by adding the overloads to `_ModelBackend.secret_get`. I guess that would be just as tidy as the comment.

> Important: we should make sure that we raise the same error as we would have previously if both are set to true.

Yes, that's the case (at runtime, not necessarily at type checking time) - in main we also let the error fall through to Juju and both in main and now we would raise `ModelError`. It's also not that important, actually, because this is in a private class, and the only use of this method only provides `peek` or `refresh` or neither.

> Minor: would this be clearer with the ignores more tightly scoped?

I'm not sure what our philosophy is there - would we rather have more lines so that we do this? Probably an interesting daily question to add to our style guide?

james-garner-canonical: Good to hear the error will be the same, that's the important bit.

Using `overload` instead of a comment sounds good, though that could always be a separate PR.

I've added an item to the daily discussion doc for this. I don't think this needs to block this PR at all if more tightly scoped ignores are controversial.

tonyandrewmeyer: > I could fix this by adding the overloads to `_ModelBackend.secret_get`. I guess that would be just as tidy as the comment.

That didn't fix it, and I'm not sure why. I'm not particularly convinced that it's worth fixing `_ModelBackend`, but if you are, I can try to dig into this more.

ops/model.py
tonyandrewmeyer: I don't want to put things into `hookcmds` if they aren't used. `model.py` is expanding on `hookcmds`, so it's fine (in my opinion) for it to build on it, including adding types that build on the `hookcmds` types.

ops/model.py
james-garner-canonical: Ah, sorry, I misunderstood your suggestion. `overload` would remove the need to *document* that we expect it to be one or the other. But we'd still need the `# type: ignore`. I'm totally happy with not adding the `overload`.

ops/model.py
james-garner-canonical: Sounds reasonable.

ops/model.py
tonyandrewmeyer: If I understand you correctly, you're saying that `start_as_current_span` will always return a span (checking the [docs](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.Tracer.start_as_current_span) that does seem to be the case) so I don't need the `else` here. Changed.

ops/model.py
tonyandrewmeyer: > So the need for the recursion prevention goes away completely, because hookcmds never logs? That's a nice side simplification!

Agreed. It's mostly going away because we're explicitly wrapping each call now (so just skip it for the special-cases) but it is indeed safe because `hookcmds` doesn't know anything about `model` so can't re-trigger anything there.

ops/model.py
tonyandrewmeyer: > I wonder if this particular chunk of logic belongs on the `BindAddress` class.

If `BindAddress` was defined in `model.py` then I'd say yes. I don't really want it in `hookcmds`, though.

> Actually, wouldn't `dataclasses.asdict` just work here, since [it recurses](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) into attributes that are also `dataclasses`?

That would then require a `cast` to `_NetworkDict` from what I can tell (because `asdict` returns `dict[str, Any]`, so you lose the structure that `Network` has). I think it's better to be explicit - the explicitness is how I found that we (in main) are missing a bunch of data here.

`network-get` is a bit weird compared to the rest, because of the shared typed dicts, because we're getting JSON from Juju, casting that to the typed dict, converting it to a nice data class, then breaking that data class object down into mostly the same typed dicts that we had before (and then converting that into an object of a class that resembles but isn't a dataclass).

If we were doing this from scratch, I think we'd have the `_ModelBackend` methods return the `hookcmd` dataclass objects (or even avoid wrapping `hookcmds` in some cases, where the state in `_ModelBackend` isn't required). It's private, so we could do that now, but in the spec I said that I'd leave `model.py` alone other than changing to call `hookcmds` and I'd prefer to stick with that for now.

ops/model.py
```diff
@@ -3613,69 +3603,25 @@ def __init__(
...
-            if return_output:
-                if result.stdout is None:  # type: ignore
-                    return ''
-                else:
-                    text: str = result.stdout  # type: ignore
-                    if use_json:
-                        return json.loads(text)  # type: ignore
-                    else:
-                        return text  # type: ignore
+                span.set_attribute('argv', args)
```
benhoyt: This isn't strictly "argv" anymore, so maybe rename the attribute to "args"?

ops/model.py
```diff
@@ -3613,69 +3603,25 @@ def __init__(
         self._is_leader: bool | None = None
         self._leader_check_time = None
         self._hook_is_running = ''
-        self._is_recursive = contextvars.ContextVar('_is_recursive', default=False)
 
     @contextlib.contextmanager
-    def _prevent_recursion(self):
-        token = self._is_recursive.set(True)
+    def _trace_and_wrap_errors(self, cmd: str, *args: Any, **kwargs: Any):
```
benhoyt: I really like how this turned out -- makes the usages so much simpler. However, (my suggestion) of the name is a big of a mouthful. How about just renaming to just `_run` or maybe `_run_hookcmd`? :-)

ops/model.py
```diff
@@ -3613,69 +3603,25 @@ def __init__(
...
-                        return json.loads(text)  # type: ignore
-                    else:
-                        return text  # type: ignore
+                span.set_attribute('argv', args)
+                span.set_attribute('kwargs', [f'{k}={v}' for k, v in kwargs.items()])
+                yield span
+        except hookcmds.Error as e:
+            self._check_for_security_event(e.cmd[0], e.returncode, e.stderr)
+            if (
+                cmd.startswith(('relation-', 'network-'))
```
benhoyt, line 3637: Ah, nice -- I like how you've done this with specific tests here, rather than special modes/keyword args.

ops/model.py
```diff
@@ -3733,19 +3668,11 @@ def relation_remote_app_name(self, relation_id: int) -> str | None:
...
-                return None
-            if 'option provided but not defined: --app' in str(e):
-                # "--app" was introduced to relation-list in Juju 2.8.1, so
-                # handle previous versions of Juju gracefully
-                return None
+            with self._trace_and_wrap_errors('relation-list', relation_id=relation_id, app=True):
+                return hookcmds.relation_list(relation_id, app=True)
+        except RelationNotFoundError:
+            return None
+        except ModelError:
```
benhoyt: I guess we could just drop this `except ModelError` now?

ops/model.py
```diff
@@ -3861,34 +3765,12 @@ def status_get(self, *, is_app: bool = False) -> _StatusDict:
...
-            app_status = content['application-status']
-            return {'status': app_status['status'], 'message': app_status['message']}
-        else:
-            return typing.cast('_StatusDict', content)
+        with self._run_hookcmd('status-get', app=is_app):
+            content = hookcmds.status_get(app=is_app)
+
+        # hookcmds doesn't constrain the status to the five that _StatusDict expects,
+        # but we know that will be the case, so we type: ignore.
+        return {'status': content.status, 'message': content.message}  # type: ignore
```
james-garner-canonical:
```suggestion
        return {
            'status': content.status,  # type: ignore
            'message': content.message,
        }
```

ops/model.py
```diff
@@ -3861,34 +3765,12 @@ def status_get(self, *, is_app: bool = False) -> _StatusDict:
...
-
-        if is_app:
-            content = typing.cast('dict[str, _StatusDict]', content)
-            app_status = content['application-status']
-            return {'status': app_status['status'], 'message': app_status['message']}
-        else:
-            return typing.cast('_StatusDict', content)
+        with self._run_hookcmd('status-get', app=is_app):
+            content = hookcmds.status_get(app=is_app)
+
```
james-garner-canonical, line 3789:
```suggestion
```

ops/model.py
```diff
@@ -3613,69 +3603,25 @@ def __init__(
         self._is_leader: bool | None = None
         self._leader_check_time = None
         self._hook_is_running = ''
-        self._is_recursive = contextvars.ContextVar('_is_recursive', default=False)
 
     @contextlib.contextmanager
-    def _prevent_recursion(self):
-        token = self._is_recursive.set(True)
+    def _run_hookcmd(self, cmd: str, *args: Any, **kwargs: Any):
```
james-garner-canonical: How about `_wrap_hookcmd` instead? `_run_hookcmd` made me think that we would call `hookcmds._utils.run` in here.
```suggestion
    def _wrap_hookcmd(self, cmd: str, *args: Any, **kwargs: Any):
```

ops/model.py
```diff
@@ -3980,8 +3871,14 @@ def log_split(
 
     def juju_log(self, level: str, message: str) -> None:
         """Pass a log message on to the juju logger."""
+        # We do not trace this call, to avoid an infinite recursion loop.
         for line in self.log_split(message):
-            self._run('juju-log', '--log-level', level, '--', line)
+            # For backwards compatibility we allow arbitrary level strings.
+            try:
+                hookcmds.juju_log(line, level=level)  # type: ignore
```
james-garner-canonical:
```suggestion
                hookcmds.juju_log(
                    line,
                    level=level,  # type: ignore
                )
```

ops/model.py
james-garner-canonical: Thanks for explaining why you wouldn't want this logic in `hookcmds`. I'm still on the fence about whether that's the right call, but since both locations are internal I don't think it matters too much.

I've also just realised that we cant just `return dataclasses.asdict(raw)  # type: ignore` here, since the dictionary requires kebab-case keys, so I think this is best as-is.

ops/hookcmds/_types.py
```diff
@@ -34,10 +34,19 @@
...
+
+    class AddressDict(TypedDict, total=False):
+        hostname: str
+        address: str  # Juju < 2.9
+        value: str  # Juju >= 2.9
+        cidr: str
+
+    BindAddressDict = TypedDict(
+        'BindAddressDict',
+        {'mac-address': NotRequired[str], 'interface-name': str, 'addresses': list[AddressDict]},
```
james-garner-canonical:
```suggestion
        {'mac-address': NotRequired[str], 'interface-name': str, 'addresses': list[AddressDict] | None},
```
My other comment:

> I agree that we shouldn't expose the `None`-able-ness to users. But isn't `BindAddressDict` only used internally? Typing `BindAddressDict.addresses` with `| None` would just ensure that we get an error if someone tries to change `d.get('addresses') or []` to `d.get('addresses', [])` someday, and they'll have to explicitly decide to change the type if we believe it will never be `None`.

This would also make sure that we're consistent about expecting that `addresses` might be `None`

ops/model.py
```diff
@@ -4059,47 +3963,27 @@ def secret_get(
...
-            return self._run(*args, return_output=return_output, use_json=use_json)
-        except ModelError as e:
-            if 'not found' in str(e):
-                raise SecretNotFoundError() from e
-            raise
+        # The type: ignore here is because the type checker can't tell that
+        # we will always have refresh or peek but not both, and either id or
+        # label.
+        with self._run_hookcmd('secret-get', id=id, label=label, refresh=refresh, peek=peek):
+            return hookcmds.secret_get(id=id, label=label, refresh=refresh, peek=peek)  # type: ignore
```
james-garner-canonical:
```suggestion
            return hookcmds.secret_get(
                id=id, label=label, refresh=refresh, peek=peek  # type: ignore
            )
```

ops/model.py
```diff
@@ -4059,47 +3963,27 @@ def secret_get(
...
-            args.extend(['--label', label])
-        result = self._run_for_secret('secret-info-get', *args, return_output=True, use_json=True)
-        info_dicts = typing.cast('dict[str, Any]', result)
-        id = next(iter(info_dicts))  # Juju returns dict of {secret_id: {info}}
-        return SecretInfo.from_dict(
-            id, typing.cast('dict[str, Any]', info_dicts[id]), model_uuid=self.model_uuid
+        # The type: ignore here is because the type checker can't tell, even
+        # with local overloads, that either id or label must be provided.
+        with self._run_hookcmd('secret-info-get', id=id, label=label):
+            raw = hookcmds.secret_info_get(id=id, label=label)  # type: ignore
```
james-garner-canonical:
```suggestion
            raw = hookcmds.secret_info_get(
                id=id, label=label  # type: ignore
            )
```

ops/model.py
dimaqq: You are correct about span always being truthy, and that's why I got here in the first place.

The larger issue I probably need to read the changes carefully.
In short:
- we hook up a custom logger
- that logger is going to call `juju-log`
- we [probably] want that exec to show up in tracing (that's what happens in main)
- however
- when we're exporting buffered traces, an exception could be raised
- which is logged
- which calls `juju-log`
- we don't want that exec to touch tracing (if it did, we'd trigger another export, another `juju-log`, another trace, another export...)

dimaqq: Or maybe that's achieved differently.
I don't recall off the top of my head, see my other comment for another scenario.

ops/model.py
dimaqq: I wonder about that.
I'll have to refresh my reading of my own changes from just 6 months ago, but here's the other concern (separate from my comment about spans above):

- something is logged
   - we call `juju-log`
      - which fails, and that's logged
         - we call `juju-log`
            - ...

ops/model.py
tonyandrewmeyer: Ah, I missed that `juju-log` does get a span in main - I thought we were skipping it not only because of the recursion but because it felt duplicative to have logging in the tracing. I'm not sure how in main it happens, though - if the command is `juju-log` doesn't it always end up with a `None` span?

Loops from exceptions are an issue, though. I think the cleanest way to do that is to have the logging use the `hookcmds` method (the `_ModelBackend` version only provides wrapping to multiple lines, which the logger could also do, and re-raising as `ModelError`, which the logger could do but really shouldn't happen in the logging anyway).

It's a pretty small change, and nicer code anyway in my opinion:

```diff
diff --git a/ops/log.py b/ops/log.py
index 0df47c02..0416ecf1 100644
--- a/ops/log.py
+++ b/ops/log.py
@@ -29,6 +29,8 @@ import warnings
 if typing.TYPE_CHECKING:
     from .model import _ModelBackend
 
+from .hookcmds import juju_log
+
 
 class JujuLogHandler(logging.Handler):
     """A handler for sending logs and warnings to Juju via juju-log."""
@@ -43,7 +45,10 @@ class JujuLogHandler(logging.Handler):
         This method is not used directly by the ops library, but by
         :class:`logging.Handler` itself as part of the logging machinery.
         """
-        self.model_backend.juju_log(record.levelname, self.format(record))
+        juju_log(
+            self.format(record),
+            level=record.levelname,  # type: ignore  # trust that this will be a valid level name
+        )
 
 
 def setup_root_logging(
@@ -129,26 +134,16 @@ class _SecurityEventLevel(enum.Enum):
 
 
 @functools.cache
-def _get_juju_log_and_app_id():
+def _get_app_id():
     logger = logging.getLogger()
     for juju_handler in logger.handlers:
         if isinstance(juju_handler, JujuLogHandler):
             model_backend = juju_handler.model_backend
             juju_context = model_backend._juju_context
             app_id = f'{juju_context.model_uuid}-{juju_context.unit_name}'
-            juju_log = model_backend.juju_log
-            return juju_log, app_id
-
-    warnings.warn(
-        'JujuLogHandler is not set up for the logger. '
-        'Call setup_root_logging() before logging security events.',
-        RuntimeWarning,
-    )
-
-    def juju_log(level: str, message: str):
-        logger.debug(message)
+            return app_id
 
-    return juju_log, 'charm'
+    return 'charm'
 
 
 def _log_security_event(
@@ -169,7 +164,7 @@ def _log_security_event(
           consumption. Includes additional details of the event that do not
           fit in the event name.
     """
-    juju_log, app_id = _get_juju_log_and_app_id()
+    app_id = _get_app_id()
     type = event_type if isinstance(event_type, str) else event_type.value
     data: dict[str, typing.Any] = {
         # This duplicates the timestamp that will be in the Juju log, but is
@@ -183,4 +178,4 @@ def _log_security_event(
         'event': f'{type}:{event_data}',
         'description': description,
     }
-    juju_log('TRACE', json.dumps(data))
+    juju_log(json.dumps(data), level='TRACE')
```

There would need to be a bunch of test adjustments, though. So perhaps as a follow-up PR to avoid making this long one that has partial approval more complicated?

Currently, `hookcmds` has no logging. In theory, some could be added in the future, and `setup_root_logging` could grab that logger and also call `juju-log` from it, which could introduce a loop. So it's slightly less safe in that regard, although there could be a test and a comment about that.

I feel like it's a bit simpler to cut off the place where it could loop than to keep the recursion-checking system. Would you prefer we kept the system that's in main now?

ops/model.py
tonyandrewmeyer: If the failure is an exception, then this is the same as the situation above, it doesn't really matter whether `juju-log` or something else is raising the exception.

Other than the exception case, there's no logging in `hookcmds`, so it can't trigger any `_ModelBackend.juju_log` usage.

tonyandrewmeyer: Covered in the other thread.

ops/model.py
tonyandrewmeyer: I like this blank space, I think it makes the comment more readable.

ops/model.py
tonyandrewmeyer: I've put back the system from main, as per the discussion earlier today.

tonyandrewmeyer:
Here's the diff (it's exactly the same as the one above other than the temporary file names):

```diff
$ diff -c ../record-run-calls/run-edited.log run.log 
*** ../record-run-calls/run-edited.log	2025-10-20 16:12:35.458508140 +1300
--- run.log	2025-10-20 16:08:24.262596338 +1300
***************
*** 1,4 ****
! {"args": ["relation-ids", "--format=json", "req1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
--- 1,4 ----
! {"args": ["relation-ids", "req1", "--format=json"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
  {"args": ["relation-list", "--format=json", "-r", "1"], "input": null}
***************
*** 29,39 ****
  {"args": ["action-get", "--format=json"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc", "key#file=/tmp/tmpqvi3e_o5/key"], "input": null}
  {"args": ["secret-get", "--format=json", "secret:changed", "--peek"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc", "key#file=/tmp/tmpk0scd1av/key"], "input": null}
! {"args": ["secret-set", "--owner", "application", "secret:changed", "--description", "desc", "key#file=/tmp/tmpf9ukvpiq/key"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
  {"args": ["status-set", "--application=True", "blocked", "--", "first"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
--- 29,39 ----
  {"args": ["action-get", "--format=json"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
  {"args": ["secret-remove", "secret:remove", "--revision", "28"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed", "key#file=/tmp/tmpxoacd03q/key"], "input": null}
  {"args": ["secret-get", "--format=json", "secret:changed", "--peek"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed", "key#file=/tmp/tmpxjomozai/key"], "input": null}
! {"args": ["secret-set", "--description", "desc", "--owner", "application", "secret:changed", "key#file=/tmp/tmpwalwfxnv/key"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
  {"args": ["status-set", "--application=True", "blocked", "--", "first"], "input": null}
  {"args": ["is-leader", "--format=json"], "input": null}
```

ops/model.py
james-garner-canonical: Feels odd to me in such a short function, but not a big deal either way.

test/test_model.py
```diff
@@ -4350,20 +4332,20 @@ def test_set_ports_replace(self, fake_script: FakeScript, unit: ops.Unit):
     def test_set_ports_close_all(self, fake_script: FakeScript, unit: ops.Unit):
         fake_script.write('open-port', 'exit 0')
         fake_script.write('close-port', 'exit 0')
-        fake_script.write('opened-ports', 'echo 8022/udp')
+        fake_script.write('opened-ports', 'echo [\\"8022/udp\\"]')
```
dimaqq: Is this change because we used to parse text and now we force JSON output?

test/test_model.py
```diff
@@ -4279,13 +4259,15 @@ def test_opened_ports(self, fake_script: FakeScript, unit: ops.Unit):
...
-            ['opened-ports', ''],
+            ['opened-ports', '--format=json'],
         ]
 
     def test_opened_ports_warnings(
         self, caplog: pytest.LogCaptureFixture, fake_script: FakeScript, unit: ops.Unit
     ):
-        fake_script.write('opened-ports', """echo 8080/tcp; echo 1234/ftp; echo 1000-2000/udp""")
+        fake_script.write(
+            'opened-ports', """echo [\\"8080/tcp\\", \\"1234/ftp\\", \\"1000-2000/udp\\"]"""
```
dimaqq: WDYT about this for readability?

```suggestion
            'opened-ports', """echo '["8080/tcp", "1234/ftp", "1000-2000/udp"]'"""
```

dimaqq: If you like this, consider applying this style to a bunch more places in this file.

ops/model.py
```diff
@@ -3980,8 +3891,17 @@ def log_split(
 
     def juju_log(self, level: str, message: str) -> None:
         """Pass a log message on to the juju logger."""
+        # We do not trace this call, to avoid an infinite recursion loop.
```
dimaqq: This is not quite accurate.

"prevent_recrusion" prevents the recursion, which is out of scope here.

additionally, `juju-log` is exempted from tracing because touching logging in the trace exporter would lead to a feedback loop. IIRC it's both in case of error, but also non-error `logging.debug` that happens to be on the path somewhere.

(we could do the above selectively: trace main thread log but not the tracing transport thread, for example).

```suggestion
        # Note that this hook command is exempted from tracing, because trace exporter touches logging, which may call our custom logging handler, which would trigger this hook command, resulting in a feedback loop.
```

(feel free to hack this up for readability)

dimaqq (approved):

ops/model.py
tonyandrewmeyer: Ah, yes, I missed updating that comment when I put the anti-recursion system back, sorry. Will fix.

test/test_model.py
tonyandrewmeyer: Yes.

PR #2117, opened 2025-10-17
fix!: ensure that the testing context manager is exited when an exception occurs

### LINKED ISSUE #1926: OPERATOR_DISPATCH should be reset after every Context.run
From: https://github.com/canonical/operator/pull/1925#issuecomment-3130188546

> > `ops/_main.py", line 165, in __init__    raise _Abort(0)` -- this is weird, that's the OPERATOR_DISPATCH trick to prevent recursion... but why does this happen in the same process, under Scenario? https://github.com/canonical/operator/blob/main/ops/_main.py#L163-L165

> This is happening when a test crashes, for subsequent tests. Fixing the crashing test stops it happening. It seems like the environment isn't being reset, maybe related to the recent change in handling _Abort.

tonyandrewmeyer:
Currently, if `testing.Context` is used as a context manager, and an exception occurs during the event, proper cleanup is not done. This is noticeable in two ways in particular:
* Exceptions aren't wrapped in `UncaughtCharmError`.
* The environment is not cleaned up. This can manifest in a cascading failure of tests if `OPERATOR_DISPATCH` is left set.

We currently manually call the `__exit__` function of one of the context managers when wrapping up. However, if an exception occurs, we never reach that point, and the context manager is left un-exited (you'll also often see errors when exiting Python, because it doesn't like being left in that state). The PR fixes this by also calling `__exit__` if exiting the `Context` context manager, and it's handling an exception.

Many thanks to @james-garner-canonical for helping figure out where things were wrong wrong and a good way to fix this.

Fixes #1926

### DIFF
--- a/testing/tests/test_context.py
+++ b/testing/tests/test_context.py
@@ -128,3 +128,34 @@ def test_run_uncaught_error():
     with pytest.raises(UncaughtCharmError):
         ctx.run(ctx.on.start(), State())
     assert 'TEST_ENV_VAR' not in os.environ
+
+
+def test_context_manager_env_cleared():
+    class GoodCharm(CharmBase):
+        def __init__(self, framework):
+            super().__init__(framework)
+            self.framework.observe(self.on.start, self._on_start)
+            os.environ['TEST_ENV_VAR'] = '1'
+
+        def _on_start(self, event):
+            os.environ['TEST_ENV_VAR'] = '2'
+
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
> Do you think it's worth adding the 'environment gets cleaned up' tests for non-crashy charms too?

I thought there was already one for this, but can't find it, so yes I've added two more.

benhoyt (approved):
Thanks guys for the analysis and fix!

PR #2119, opened 2025-10-20
docs: replace machine charm tutorial by an improved tutorial

dwilding:
This PR replaces [Write your first machine charm](https://documentation.ubuntu.com/ops/latest/tutorial/write-your-first-machine-charm/) by new & improved tutorial based on tinyproxy.

**[Preview of tutorial](https://canonical-ubuntu-documentation-library--2119.com.readthedocs.build/ops/2119/tutorial/write-your-first-machine-charm/)**

### DIFF
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -567,7 +567,7 @@ You should now have two terminals:
 
     ```text
     Model    Controller     Cloud/Region         Version  SLA          Timestamp
-    testing  concierge-lxd  localhost/localhost  3.6.8    unsupported  09:00:00+08:00
+    testing  concierge-lxd  localhost/localhost  3.6.11   unsupported  09:00:00+08:00
     ```
 
 You're now ready to deploy your charm.
@@ -584,7 +584,7 @@ When your charm has started tinyproxy, the application will go into "active" sta
 
 ```text
 Model    Controller     Cloud/Region         Version  SLA          Timestamp
-testing  concierge-lxd  localhost/localhost  3.6.8    unsupported  09:01:38+08:00
+testing  concierge-lxd  localhost/localhost  3.6.11   unsupported  09:01:38+08:00
 
 App        Version  Status  Scale  Charm      Channel  Rev  Exposed  Message
 tinyproxy  1.11.0   active      1  tinyproxy             0  no
@@ -592,8 +592,8 @@ tinyproxy  1.11.0   active      1  tinyproxy             0  no
 Unit          Workload  Agent  Machine  Public address  Ports  Message
 tinyproxy/0*  active    idle   0        10.71.67.208
 
-Machine  State    Address       Inst id        Base          AZ  Message
-0        started  10.71.67.208  juju-8e7bd9-0  ubuntu@22.04      Running
+Machine  State    Address       Inst id        Base          AZ            Message
+0        started  10.71.67.208  juju-8e7bd9-0  ubuntu@22.04  juju-sandbox  Running
 ```
 
 ```{tip}
@@ -873,10 +873,10 @@ The output should be similar to:
 unit: commands[1]> coverage report
 Name               Stmts   Miss Branch BrPart  Cover   Missing
 --------------------------------------------------------------
-src/charm.py          72      5     20      7    87%   67->exit, 97, 102->exit, 111-112, 121-122
-src/tinyproxy.py      47     26      6      0    40%   34-41, 52-53, 60, 65, 70-75, 80, 85-87, 92-95, 100-108
+src/charm.py          71      5     20      7    87%   71->exit, 101, 106->exit, 115-116, 125-126
+src/tinyproxy.py      47     26      6      0    40%   34-41, 52-56, 63, 68, 73-78, 83, 88-90, 95-98, 103-111
 --------------------------------------------------------------
-TOTAL                119     31     26      7    70%
+TOTAL                118     31     26      7    69%
   unit: OK (1.21=setup[0.05]+cmd[1.03,0.13] seconds)
   congratulations :) (1.30 seconds)
 ```
@@ -890,7 +890,7 @@ When you created the initial version of your charm, Charmcraft included integrat
 In `tests/integration/test_charm.py`, change `juju.wait(jubilant.all_active)` to:
 
 ```python
-juju.wait(jubilant.all_active, timeout=600)
+    juju.wait(jubilant.all_active, timeout=600)
 ```
 
 This extends the duration that Jubilant waits for your charm to deploy, in case the integration tests run slowly in your virtual machine. The default duration would be sufficient if the integration tests were running in a continuous integration environment.
@@ -953,8 +953,8 @@ Congratulations on reaching the end of the tutorial!
 
 You can keep things running, to explore further, or you can remove what you created:
 
-- If you're still watching Juju status, press <kbd>Ctrl</kbd> + <kbd>C</kbd> to stop watching.
 - To remove your charm from Juju, run `juju remove-application tinyproxy`. You don't need to do this if you plan to remove your virtual machine.
+- If you're still watching Juju status, press <kbd>Ctrl</kbd> + <kbd>C</kbd> to stop watching.
 - To exit your virtual machine, run `exit`. The terminal switches back to your host machine.
 - To stop your virtual machine, run `multipass stop juju-sandbox`.
 - To remove your virtual machine, run `multipass delete juju-sandbox`.


docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,958 @@
...
+To add the libraries to your charm's dependencies, run:
+
+```text
+uv add charmlibs-apt charmlibs-pathops
+```
+
+This has added the following Python packages to the `dependencies` list in `pyproject.toml`:
+
+- [`charmlibs-apt`](https://documentation.ubuntu.com/charmlibs/reference/charmlibs/apt/) - A library for using APT to manage system packages. This is how your charm will install tinyproxy.
+- [`charmlibs-pathops`](https://documentation.ubuntu.com/charmlibs/reference/charmlibs/pathops/) - A file operations library, similar to `pathlib` from the standard library.
```
dwilding: @james-garner-canonical do you know what the intersphinx targets would be for these? I had a quick try, but couldn't figure out the right targets.

examples/machine-tinyproxy/tests/integration/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="module")
+def juju(request: pytest.FixtureRequest):
+    """Create a temporary Juju model for running tests."""
+    with jubilant.temp_model() as juju:
+        juju.wait_timeout = 10 * 60
```
dwilding: I'll double check whether this needs to be as high as 10 mins

docs/tutorial/index.md
```diff
@@ -9,6 +9,7 @@ Writing a machine charm is not so different from writing a Kubernetes charm, but
 
 write-your-first-machine-charm
 from-zero-to-hero-write-your-first-kubernetes-charm/index
+write-a-machine-charm-for-tinyproxy
```
dwilding: This won't be the final state. Before merging the PR, I'll remove `write-a-machine-charm-for-tinyproxy.md` and overwrite `write-your-first-machine-charm.md` with the new tutorial.

examples/machine-tinyproxy/tests/integration/conftest.py
dwilding: I just realised that `juju.wait_timeout = 10 * 60` isn't mentioned in the tutorial itself. For some reason, I thought that the tutorial wouldn't need to mention it, but of course it would... :man_facepalming: I'll think about how to include this without teaching people to blindly apply a long timeout.

dwilding: Ben and I just discussed this. Instead of modifying `conftest.py`, I should be able to add a timeout to `test_deploy` in `test_charm.py`. This should be easier to explain in the tutorial, as we're already modifying that file. It lets us keep `conftest.py` as generated by Charmcraft.

dwilding: I've updated the example charm and the tutorial, to set a timeout in `test_deploy`:

```python
juju.wait(jubilant.all_active, timeout=600)
```

I've also explained why this is needed. One of my test runs took > 300 seconds to complete, so I stuck with 600 to be on the safe side.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: James is adding targets in https://github.com/canonical/charmlibs/pull/233. I'll switch to intersphinx links after that PR merges

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+In your virtual machine, run:
+
+```text
+sudo snap install --classic concierge
+sudo concierge prepare -p machine --extra-snaps astral-uv --charmcraft-channel latest/candidate
+```
+
+This first installs Concierge, then uses Concierge to install and configure the other tools (except tox). The option `-p machine` tells Concierge that we want tools for developing machine charms.
+
+It will take up to 15 minutes to install the tools, depending on your computer and network. When the tools have been installed, you'll see a message that ends with:
```
tonyandrewmeyer: I wonder if we avoid a maximum, since if your network is *really* bad then I'm pretty sure it could be more than 15 minutes. Maybe something like "The time to install the tools depends on your computer and network, but is typically less than 15 minutes".

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+```
+
+Notice that the helper module is stateless. In fact, your charm as a whole will be stateless. The main logic of your charm will:
+
+1. Receive an event from Juju.
+2. Use the functions in the helper module to manage tinyproxy and check its status.
+3. Report the status back to Juju.
+
+```{tip}
+After adding code to your charm, run `tox -e format` to format the code. Then run `tox -e lint` to check the code against coding style standards and run static checks.
```
tonyandrewmeyer: It's a bit unfortunate that if you run `tox -e format` at this point you get an error, because there's no docstring. I know we want to keep the code snippets small, but maybe it would be worth having a one-liner version just so that the format check exits without errors?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+```python
+    def _on_install(self, event: ops.InstallEvent) -> None:
+        """Install tinyproxy on the machine."""
+        if not tinyproxy.is_installed():
+            tinyproxy.install()
+            version = tinyproxy.get_version()
+            self.unit.set_workload_version(version)
+```
+
+When your charm receives the "install" event from Juju, Ops runs this method and responds to Juju with the version of tinyproxy that's installed on the machine. Juju shows the tinyproxy version in its status output.
```
tonyandrewmeyer: "responds to Juju" doesn't seem right to me - it's not the "response" of the install event. What about something like "and tells Juju the version", or "responds by telling Juju the version".

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+
+    def wait_for_running(self) -> None:
+        """Wait for tinyproxy to be running."""
+        for _ in range(3):
+            if tinyproxy.is_running():
+                return
+            time.sleep(1)
+        raise RuntimeError("tinyproxy was not running within the expected time")
+        # Raising a runtime error will put the charm into error status. The error message is for
+        # you (the charm author) to see in the Juju logs, not for the user of the charm.
```
tonyandrewmeyer: I feel like the second sentence is only partly true. Juju users are definitely looking at the log and it will provide more information in this case than the error status, I think.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+
+Before you can try your charm, you need to "pack" it. Packing combines the charm code and metadata into a single file that can be deployed to Juju.
+
+In your virtual machine, make sure that the working directory is `~/tinyproxy`. Then run:
+
+```text
+charmcraft pack
+```
+
+Charmcraft will take about 20 minutes to pack your charm, depending on your computer and network.
```
tonyandrewmeyer: Since we're using the `uv` plugin and exploiting the bug to avoid building `pydantic`, this seems excessive. I do find that it's regularly about 20 minutes if you're building everything from source, but this took 8m35s for me because we skipped that. Maybe we should be less certain about this and say it could take as much as 20 minutes, or something like that?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+    """Patch the helper module to use mock functions for interacting with tinyproxy."""
+    monkeypatch.setattr("charm.tinyproxy.ensure_config", tinyproxy.ensure_config)
+    monkeypatch.setattr("charm.tinyproxy.get_version", tinyproxy.get_version)
+    monkeypatch.setattr("charm.tinyproxy.install", tinyproxy.install)
+    monkeypatch.setattr("charm.tinyproxy.is_installed", tinyproxy.is_installed)
+    monkeypatch.setattr("charm.tinyproxy.is_running", tinyproxy.is_running)
+    monkeypatch.setattr("charm.tinyproxy.start", tinyproxy.start)
+    monkeypatch.setattr("charm.tinyproxy.stop", tinyproxy.stop)
+    monkeypatch.setattr("charm.tinyproxy.reload_config", tinyproxy.reload_config)
+    monkeypatch.setattr("charm.tinyproxy.uninstall", tinyproxy.uninstall)
```
tonyandrewmeyer: I didn't test this, so maybe it doesn't work, but could you just monkeypatch `charm.tinyproxy` to be `MockTinyproxy` here instead of individually doing each method?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+    assert state_out.unit_status == testing.ActiveStatus()
+    assert tinyproxy_configured.is_running()
+    assert tinyproxy_configured.config == (PORT, "foo")
+    assert tinyproxy_configured.reloaded_config
+
+
+# Define a reusable fixture that provides invalid slugs.
+@pytest.fixture(params=["", "foo_bar", "foo/bar"])
+def invalid_slug(request):
+    return request.param
```
tonyandrewmeyer: TIL this fixture technique :)

Is there any benefit of this over doing this?

```python
@pytest.mark.parametrize("invalid_slug", ["", "foo_bar", "foo/bar"])
```

Perhaps not having to repeat that line twice?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+
+When you created the initial version of your charm, Charmcraft included integration tests. The tests use {external+jubilant:doc}`Jubilant <index>` to interact with Juju. We'll expand the tests to cover more of your charm's functionality.
+
+In `tests/integration/test_charm.py`, change `juju.wait(jubilant.all_active)` to:
+
+```python
+juju.wait(jubilant.all_active, timeout=600)
+```
+
+This extends the duration that Jubilant waits for your charm to deploy, in case the integration tests run slowly in your virtual machine. The default duration would be sufficient if the integration tests were running in a continuous integration environment.
```
tonyandrewmeyer: I'm happy with this explanation to extend the timeout, and how it doesn't suggest that you should always do this. As a data point, the tests took 2m2.3s on my laptop.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+
+It's important to exit your virtual machine at this point. If you don't exit your virtual machine, you won't be able to use tox later in the tutorial.
+
+To exit your virtual machine, run:
+
+```text
+exit
+```
+
+The terminal switches back to your usual operating system. Your virtual machine is still running in the background.
```
james-garner-canonical: Does `uv tool update-shell` not eliminate the need for this?

I wanted to test it myself, but I can't download the `multipass` image currently.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,980 @@
...
+
+The terminal prompt is now `ubuntu@juju-sandbox:~$`.
+
+Next, create a project directory and use Charmcraft to create the initial version of your charm:
+
+```text
+mkdir tinyproxy
+cd tinyproxy
+charmcraft init --profile machine
+```
```
james-garner-canonical: Are we sure we want to do this inside the VM? I recall an older tutorial (the previous machine one?) had you create files locally and mount the directory in the VM. This was nice as it let you trivially blow the VM away if it got into a bad state without losing any of your work following the tutorial.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: From my memory, it updates one of your dotfiles, but you need to restart your shell for that to take effect. I figured it's easiest to exit the VM and go back in, but am very open to other suggestions for handling this

tonyandrewmeyer: > Does `uv tool update-shell` not eliminate the need for this?

No, I wondered about that too. `update-shell` updates things so that when you re-start the shell everything is available. You could do it in other ways, like `exec bash` or `source (the file that changed, probably bashrc?)` or something, but I guess this is probably simplest even if not most convenient.

james-garner-canonical: Thanks for confirming, Tony.

Since it's in a VM image that we know the contents of (the shell, the config files), it seems like we should be able to provide a nice one line command the user can run instead of the more verbose exit and come back instructions. That seems both more convenient and simpler in a sense -- less for the user to read and do.

Unfortunately I can't test this locally. With `tox` and `tox-uv` uninstalled, I can run `uv tool install tox --with tox-uv` in a fresh login tty and then run `tox` without any issues.

Would someone with a VM setup be willing to look into this?

james-garner-canonical: Is it because `~/.local/bin` isn't in `PATH`? If so, the command could be as simple as:
```bash
export PATH="$HOME/.local/bin:$PATH"
```

benhoyt: Yeah, this works (makes `tox` available at least).

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,956 @@
...
+
+This has added the following Python packages to the `dependencies` list in `pyproject.toml`:
+
+- [`charmlibs-apt`](https://documentation.ubuntu.com/charmlibs/reference/charmlibs/apt/) - A library for using APT to manage system packages. This is how your charm will install tinyproxy.
+- [`charmlibs-pathops`](https://documentation.ubuntu.com/charmlibs/reference/charmlibs/pathops/) - A file operations library, similar to `pathlib` from the standard library.
+
+Next, replace the contents of `src/tinyproxy.py` with:
+
+```python
+import logging
```
benhoyt: If you don't have a module docstring, `tox -e format` complains "Missing docstring in public module":

```suggestion
"""Functions to interact with tinyproxy."""

import logging
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
benhoyt: Yeah, this took 2m29s for me.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,956 @@
...
+msg="Bootstrapped Juju" provider=lxd
+```
+
+To install tox, run:
+
+```text
+uv tool install tox --with tox-uv
+export PATH="$HOME/.local/bin:$PATH"
+```
+
```
james-garner-canonical:
`````suggestion

````{tip}
Take a snapshot of your VM now so that you can easily restart the tutorial steps from scratch without waiting for the tools to be reinstalled.
```text
juju-sandbox juju-sandbox --name=juju-sandbox-with-tools
```
````

`````

File: docs/tutorial/write-a-machine-charm-for-tinyproxy.md [hunk 16]
```diff
@@ -0,0 +1,956 @@
...
+### Change the configuration
+
+Let's see what happens to the reverse proxy if we change the `slug` configuration option. Run:
+
+```text
+juju config tinyproxy slug=foo
+```
+
+Where `<address>` is the IP address of machine 0 from Juju status.
```
benhoyt: This doesn't seem to make sense here (just after the `juju config` command). Probably don't need it again after the above.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,956 @@
...
+
+
+def test_version(monkeypatch: pytest.MonkeyPatch):
+    """Test that the helper module correctly returns the version of tinyproxy."""
+    version_process = MockVersionProcess("1.11.0")
+    monkeypatch.setattr("subprocess.run", lambda *args, **kwargs: version_process)
+    assert tinyproxy.get_version() == "1.11.0"
+```
+
+We'll run all the tests later in the tutorial. But if you'd like to see whether this test passes, you can run `tox -e unit -- tests/unit/test_tinyproxy.py`.
```
benhoyt: Maybe put this command in a command block? It's a bit hard to copy-paste here, especially with the trailing `.`.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
james-garner-canonical: Could also note that this is 'to pack your charm for the first time', since it will cache the built/cheated deps after that, right?

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,956 @@
...
+```python
+    assert version == "1.11.0"
+```
+
+You should now have the following tests:
+
+- `test_deploy` - Deploys your charm and checks that it goes into active status.
+- `test_workload_version_is_set` - Checks that your charm reports the correct version of tinyproxy to Juju.
+
+You might have a question at this point: what makes 1.11.0 the "correct" version of tinyproxy? The answer is, we're cheating slightly because we happen to know that APT will install version 1.11.0 on the machine. In general, it's good practice for charms to pin workload versions, so the charm code ought to ask APT to install a specific version of tinyproxy.
```
benhoyt: This will break if the apt package gets updated, right? Can we avoid "cheating" (and potentially having the tutorial break) and pin this now?

benhoyt (approved):
Very nice! Took 45 minutes for me, including installing Multipass, but I rushed through it and somewhat skimmed the text. So 2 hours seems reasonable.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
benhoyt: Just for clarity, the 2m29s was the first time.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
james-garner-canonical: This is merged now btw. The label is the canonical distribution package name: `charmlibs-apt` and `charmlibs-pathops`.

dwilding: Great, thanks! I've just switched to intersphinx links

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Ah, thanks for catching this. The full module in the code has a docstring, but I'd missed that I need to show that in the tutorial too. Updated.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Resolved, thanks!

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Wow, this makes things much nicer! Why didn't I think of that :laughing: 

I've been able to replace the whole `patch_charm` function by:

```py
monkeypatch.setattr("charm.tinyproxy", tinyproxy)
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Yeah, I was previously using the fixture in three places (before I switched to Pydantic for validating the slug). I agree that the repetition isn't bad, so I've switched to `@pytest.mark.parametrize`.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Got you - I've changed to this:

```python
# Raising a runtime error will put the charm into error status.
# The Juju logs will show the error message, to help you debug the error.
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I like "and tells Juju the version" - I've gone with that

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Yep, `export PATH="$HOME/.local/bin:$PATH"` works. We don't even need `uv tool update-shell`, because `.profile` in the VM already has this:

```sh
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

I've got rid of the steps that have you restart a terminal in the VM. The step for installing tox is now:

> To install tox, run:
>
> ```
> export PATH="$HOME/.local/bin:$PATH"
> uv tool install tox --with tox-uv
> ```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I agree, that works better - done

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Thanks for raising this. I agree it's better to show the best practice.

I've made a few changes:

- Updated the function that installs tinyproxy:

    ```py
    # Install a specific package from ubuntu@22.04
    # See https://packages.ubuntu.com/jammy/tinyproxy-bin
    # In general, it's good practice for charms to pin workload versions.
    apt.add_package("tinyproxy-bin", "1.11.0-1")
    ```

- Removed the paragraph that talks about "cheating"

- In the unit tests that use a mock version number, deliberately used "1.0.0" instead of the real version number, so that it's clearer where the pinned version number needs to go.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: Good feedback and suggestions, thanks. I've gone with this explanation:

> Charmcraft will take up to 20 minutes to pack your charm, depending on your computer and network. If you modify the charm code after completing the tutorial, packing will be faster the second time because Chamcraft has cached the packing environment.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I've gone with this wording (and similarly for the step where the VM is created):

> This step should take less than 15 minutes, but the time depends on your computer and network.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: It turns out that snapshot only works for stopped VMs. Since this isn't critical for the tutorial, I'll leave it out

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I agree. We also had internal discussion that it's nicer for people to be able to use their familiar editor/IDE. We don't really want to teach people to use nano to work on real charms.

I've added a `multipass mount` step in [Create a virtual machine](https://canonical-ubuntu-documentation-library--2119.com.readthedocs.build/ops/2119/tutorial/write-a-machine-charm-for-tinyproxy/#create-a-virtual-machine). Then in [Edit the metadata](https://canonical-ubuntu-documentation-library--2119.com.readthedocs.build/ops/2119/tutorial/write-a-machine-charm-for-tinyproxy/#edit-the-metadata) (the first section that edits a file) I've removed nano and left a single instruction for editing the file locally. Later parts of the tutorial don't specify how to edit files.

During testing, I discovered that tox complains if you try to run the same command from inside then outside the VM (and vice versa). I've included troubleshooting advice in the note that first mentions running tox:

> After adding code to your charm, run `tox -e format` to format the code. Then run `tox -e lint` to check the code against coding style standards and run static checks. You can run these commands from anywhere in the `~/tinyproxy` directory in your virtual machine.
>
> You can also run these commands in `~/tinyproxy-tutorial` if uv and tox are available on your usual operating system. However, be careful when running the same tox command inside and outside your virtual machine. If tox fails with an error related to the `.tox` directory, use `-re` instead of `-e` in the commands. This recreates the tox environment.

Thanks for the suggestions about all of this.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
+# Write a machine charm for tinyproxy
+
+In this tutorial, you'll write a {external+juju:ref}`machine charm <machine-charm>` for Juju using Ops and other charm development tools.
+
+It will take about 2 hours for you to complete the tutorial.
+
+What you'll need:
+
+- A workstation. For example, a laptop with an amd64 architecture. You'll need sufficient resources to launch a virtual machine with 4 CPUs, 8 GB RAM, and 50 GB disk space.
```
james-garner-canonical:
```suggestion
- A workstation. For example, a laptop with an amd64 architecture. To test the charm in a virtual machine, you'll need sufficient resources to launch a virtual machine with 4 CPUs, 8 GB RAM, and 50 GB disk space.
```
I wonder if it's worth scoping this recommendation more narrowly -- you could in principle follow the tutorial without the overhead of a VM. Just my 2c as a low-end laptop user.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+
+Where `<address>` is the IP address of the machine that tinyproxy is running on. You'll be able to use the `juju config` command to change `/example/` to a custom path.
+
+This application isn't especially realistic in isolation. But it's a good way to illustrate typical interactions between Juju, a charm, a machine, and a workload.
+
+## Set up your environment
+
+### Create a virtual machine
+
+You'll test and deploy your charm inside an Ubuntu virtual machine that's running on your computer. Your virtual machine will provide an isolated environment that's safe for you to experiment in, without affecting your usual operating system.
```
james-garner-canonical:
```suggestion
You'll test and deploy your charm inside an Ubuntu virtual machine that's running on your computer. Your virtual machine will provide an isolated environment that's safe for you to experiment in, without affecting your usual operating system. This is especially helpful for the charm's integration tests, which require a local Juju controller and LXD cloud.
```
Again, I wonder if it's worth tightening the scope of the recommendation -- most charm development is perfectly fine to do directly on your local machine, and we don't really want to scare people away from that.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+
+The helper module will depend on some libraries that are useful when writing charms.
+
+To add the libraries to your charm's dependencies, run:
+
+```text
+uv add charmlibs-apt charmlibs-pathops
+```
+
+This has added the following Python packages to the `dependencies` list in `pyproject.toml`:
```
james-garner-canonical:
```suggestion
This adds the following Python packages to the `dependencies` list in `pyproject.toml`:
```
Not sure what tense is best here, maybe you have it right already.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+
+
+def get_version() -> str:
+    """Get the version of tinyproxy that is installed."""
+    result = subprocess.run(["tinyproxy", "-v"], check=True, capture_output=True, text=True)
+    return result.stdout.removeprefix("tinyproxy").strip()
+
+
+def install() -> None:
+    """Use APT to install the tinyproxy executable."""
```
james-garner-canonical: Feel odd to see `apt` styled this way, but I guess it's correct --  actually I think this is my first time learning that it's an acronym.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+    if pid:
+        os.kill(pid, signal.SIGTERM)
+
+
+def uninstall() -> None:
+    """Uninstall the tinyproxy executable and remove files."""
+    apt.remove_package("tinyproxy-bin")
+    pathops.LocalPath(PID_FILE).unlink(missing_ok=True)
+    pathops.LocalPath(CONFIG_FILE).unlink(missing_ok=True)
+    pathops.LocalPath(CONFIG_FILE).parent.rmdir()
```
james-garner-canonical: Would it make sense to have these constants be defined as `pathops.LocalPath` objects instead of `str`?
Because their `str` representation is just the file path, they can still be used in the `config` as they are currently.
But all the call sites would look like this instead:
```suggestion
    PID_FILE.unlink(missing_ok=True)
    CONFIG_FILE.unlink(missing_ok=True)
    CONFIG_FILE.parent.rmdir()
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+
+Then add the following method to the charm class:
+
+```python
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
+        """Report the status of tinyproxy (runs after each event)."""
+        try:
+            self.load_config(TinyproxyConfig)
+        except pydantic.ValidationError as e:
+            slug_error = e.errors()[0]  # Index 0 because 'slug' is the only option validated.
```
james-garner-canonical: Very minor style thing, perhaps just my personal taste:
```suggestion
            slug_error, *_ = e.errors()  # 'slug' is the first and only option validated.
```
Or if this is correct:
```suggestion
            (slug_error,) = e.errors()  # 'slug' is the first and only option validated.
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+Then add the following class to `src/charm.py`:
+
+```python
+class TinyproxyConfig(pydantic.BaseModel):
+    """Schema for the charm's config options."""
+
+    slug: str = pydantic.Field(
+        "example",
+        pattern=r"^[a-z0-9-]+$",
+        description="Configures the path of the reverse proxy. Must match the regex [a-z0-9-]+",
```
james-garner-canonical: In this specific place, the 'must match the regex' message feels redundant with `pattern` defined on the line above. But if you think it's helpful, feel free to keep this.
```suggestion
        description="Configures the path of the reverse proxy.",
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+
+You're now ready to deploy your charm.
+
+In the `~/tinyproxy` directory, run `juju deploy ./<charm-file>`, where `<charm-file`> is the name of the file created by `charmcraft pack`. For example:
+
+```text
+juju deploy ./tinyproxy_amd64.charm
+```
+
+Juju creates an "application" from your charm. For each unit in the application, Juju starts a machine and installs your charm on the machine. We didn't tell Juju how many units we want, so Juju assumes one unit and starts one machine. After Juju has installed your charm on the machine, Juju starts sending events to your charm so that your charm can install and start tinyproxy.
```
james-garner-canonical:
```suggestion
Juju creates an "application" from your charm. For each unit in the application, Juju starts a LXD virtual machine and installs your charm on the machine. We didn't tell Juju how many units we want, so Juju assumes one unit and starts one machine. After Juju has installed your charm on the machine, Juju starts sending events to your charm so that your charm can install and start tinyproxy.
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: James and I just had further discussion that we might be able to reorder the VM steps to support snapshotting more easily (since we have to stop the VM before the mount). I'll take a look at this tomorrow

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+```python
+juju.wait(jubilant.all_active, timeout=600)
+```
+
+This extends the duration that Jubilant waits for your charm to deploy, in case the integration tests run slowly in your virtual machine. The default duration would be sufficient if the integration tests were running in a continuous integration environment.
+
+Next, remove the `@pytest.mark.skip` decorator from `test_workload_version_is_set`. Then change `assert version == ...` to:
+
+```python
+    assert version == "1.11.0"
```
james-garner-canonical:
```suggestion
    assert version == "1.11.0"  # the version installed by tinyproxy.install
```

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+```
+
+## Tear things down
+
+Congratulations on reaching the end of the tutorial!
+
+You can keep things running, to explore further, or you can remove what you created:
+
+- If you're still watching Juju status, press <kbd>Ctrl</kbd> + <kbd>C</kbd> to stop watching.
+- To remove your charm from Juju, run `juju remove-application tinyproxy`.
```
james-garner-canonical: I wonder if it's worth mentioning that it doesn't matter if these steps go wrong if you're going to delete the VM anyway?

james-garner-canonical: I'm also wondering if we're going to explain how to unsetup the VM, if we should include cleanup instructions for removing `tox`, `charmcraft`, `juju` (including removing the juju cloud and controller -- perhaps via concierge?). But I think maybe that's out of scope for this tutorial since we're having the reader follow it in a VM anyway.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
```diff
@@ -0,0 +1,926 @@
...
+Launched: juju-sandbox
+```
+
+Although you'll test and deploy your charm inside your virtual machine, you'll probably find it more convenient to write your charm using your usual text editor or IDE. Let's use the Multipass {external+multipass:ref}`mount <reference-command-line-interface-mount>` command to make a directory available inside your virtual machine:
+
+```text
+multipass stop juju-sandbox
+mkdir ~/tinyproxy-tutorial  # You'll write your charm in this directory.
+multipass mount --type native ~/tinyproxy-tutorial juju-sandbox:~/tinyproxy
+```
```
james-garner-canonical: Per discussion with @dwilding elsewhere, we're thinking about changing the order of operations to:
1. create VM
2. shell
3. concierge and friends
4. stop
5. snapshot
6. mount

This lets us take a snapshot at a useful point, and removes the need to wrangle `PATH` for `uv install tox` (as @dwilding noticed) -- and we need to stop the VM before mounting anyway.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I dunno, I think what I have is not wrong, but when I read it I prefer your version

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I'm going to keep this, as it matches the description in `charmcraft.yaml`. From discussion with Tony, in the future it should be possible to generate the `charmcraft.yaml` config from this.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I've added a sentence to say that you don't need to remove the charm if you're going to remove the VM.

I don't think it's worth explaining how to unsetup the VM - out of scope, I'd say.

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: I find `(slug_error,)` more intuitive - going with that version

docs/tutorial/write-a-machine-charm-for-tinyproxy.md
dwilding: This approach works, thanks for the suggestion! I've updated the [preview tutorial](https://canonical-ubuntu-documentation-library--2119.com.readthedocs.build/ops/2119/tutorial/write-a-machine-charm-for-tinyproxy/). Would you mind reading through from "To install tox" to the "Create a charm project" section?

I've addressed your other feedback too, BTW

james-garner-canonical: The changes look great, thanks for addressing all my feedback.

PR #2120, opened 2025-10-20
chore: ruff for Python 3.10

dimaqq:
We've left ruff at py38 when we dropped Python 3.8 support because the PR would've gotten too big.
Now I want to use `match/case` in another PR, so I figured I'd ruff up the codebase with Python 3.10 setting.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -88,7 +88,7 @@ _ContainerMeta_Raw: TypeAlias = 'dict[str, _charm.ContainerMeta]'  # prevent imp
 
 # relation data is a string key: string value mapping so far as the
 # controller is concerned
-_RelationDataContent_Raw = dict[str, str]
+_RelationDataContent_Raw: TypeAlias = dict[str, str]
 UnitOrApplicationType: TypeAlias = 'type[Unit] | type[Application]'
 
 _NetworkDict = TypedDict(


ops/pebble.py
```diff
@@ -2484,7 +2476,7 @@ def _wait_change_using_wait(self, change_id: ChangeID, timeout: float | None):
 
             try:
                 return self._wait_change(change_id, this_timeout)
-            except (socket.timeout, builtins.TimeoutError):
+            except builtins.TimeoutError:
```
dimaqq, line 2479: This is because `socket.timeout is TimeoutError` in Py 3.10+

testing/src/scenario/_runtime.py
```diff
@@ -14,7 +14,6 @@
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
-    Type,
 )
```
james-garner-canonical:
```py
from typing import TYPE_CHECKING
```

james-garner-canonical (commented):
This all seems good to me, but I suggest we hold off on merging this until #2116 lands to avoid requiring any further changes there.

testing/src/scenario/_runtime.py
dimaqq: good call, done here and a few other places.

tonyandrewmeyer:
> I could use maybe... Tony's? opinion on test_ops_testin_doc test that now fails:

You'll need to change [this line](https://github.com/canonical/operator/blob/3ac3706b0dda0c0139e9d565c376bde8cb5c4b54/test/test_infra.py#L65) to handle some of the type aliases coming from `collections.abc` I would think.

benhoyt (commented):
Thanks for this, looks good to me. I'd forgotten we hadn't done this yet!

dimaqq:
Ok, this was a deep typed rabbit hole, but I've come out the other end!

tonyandrewmeyer (commented):
Looks good, just needs the conflicts resolved.

dimaqq:
Thanks, conflicts are resolved; will merge after the tests pass.

ops/hookcmds/_types.py
```diff
@@ -31,7 +30,7 @@
 SecretRotate = Literal['never', 'hourly', 'daily', 'weekly', 'monthly', 'quarterly', 'yearly']
 SettableStatusName = Literal['active', 'blocked', 'maintenance', 'waiting']
 ReadOnlyStatusName = Literal['error', 'unknown']
-StatusName: TypeAlias = SettableStatusName | ReadOnlyStatusName
+StatusName = SettableStatusName | ReadOnlyStatusName
```
james-garner-canonical: Does the tooling want to get rid of `TypeAlias`, or was this a manual choice?

I see it was retained with a comment in one case where the name used in the alias isn't actually available at runtime properly, so I'm guessing it's manual.

I think we should keep using `TypeAlias` until our minimum version is 3.12, and then replace it with the `type` keyword since 3.12 deprecates `TypeAlias` (no planned removal yet).

test/test_infra.py
```diff
@@ -62,7 +65,17 @@ def test_ops_testing_doc():
         for name in ops.testing.__all__
         if name != 'errors'
         and name not in ops.testing._compatibility_names
-        and getattr(ops.testing, name).__class__.__module__ != 'typing'
+        and typing.get_origin(getattr(ops.testing, name))
+        not in (
```
james-garner-canonical: Is this formatting preferred over:
```suggestion
        and typing.get_origin(getattr(ops.testing, name)) not in (
```

test/test_infra.py
```diff
@@ -62,7 +65,17 @@ def test_ops_testing_doc():
...
+            # ReadableBuffer = bytes | ...
+            types.UnionType,
+            # ExecHandler = Callable[...]
+            collections.abc.Callable,
+            # RawDataBagContents = dict[...]  # Py310
+            dict,
+        )
```
james-garner-canonical: Is this simplified if we use `TypeAlias`?

test/test_infra.py
```diff
@@ -62,7 +65,17 @@ def test_ops_testing_doc():
...
+        # CharmType = TypeVar(...)
+        and getattr(ops.testing, name).__class__ is not typing.TypeVar
```
james-garner-canonical: This one would be caught by the more general (removed) original check, wouldn't it?
```py
        and getattr(ops.testing, name).__class__.__module__ != 'typing'
```

test/test_infra.py
james-garner-canonical: That is, annotate all our type aliases with `TypeAlias` and exclude any such annotated items from the docs check instead of special casing them individually.

dimaqq: I think not, because annotations are stored separately (in some module dict?).

If you can think of something simpler, please raise a PR.

From my experimentation, 3.11~3.14 is relatively simple, but 3.10 is harder.

test/test_infra.py
dimaqq: that's correct.

test/test_infra.py
james-garner-canonical: In lieu of a PR, how about this:
```python
    expected_names = set(
        name
        for name in ops.testing.__all__
        if name not in ops.testing._compatibility_names
        and type(getattr(ops.testing, name)) in (type, types.FunctionType)
    )
```

test/test_infra.py
james-garner-canonical: Resolved if we go with [this](https://github.com/canonical/operator/pull/2120/files/6dbd28b3a1cb2caac6ce6225ba6f090f5b1f52a8#r2446536587).

test/test_infra.py
dimaqq: I'm not sure how that works, off the top of my head.

How does that distinguish aliases and callables from real (public) declarations?

dimaqq: The latest counter-proposal is unlikely to work because regular public classes would be excluded:
```py
>>> class Foo: pass
...
>>> type(Foo)
<class 'type'>

Comment (dimaqq) [reply test/test_infra.py]: Oh I misread your counter-proposal before.

Now I see it as:

```py
expected_names = set(
        name
        for name in ops.testing.__all__
        if name not in ops.testing._compatibility_names
        and type(getattr(ops.testing, name)) in (
            type,  # public classes
            types.FunctionType,  # public functions
        )
    )
```

Which is elegant.
An allow-list is more concise that filtering out the undesirables.

dimaqq: Updated.

test/test_infra.py
dimaqq: code updated

test/test_infra.py
dimaqq: no longer relevant as code was updated.

ops/hookcmds/_types.py
dimaqq: My take is that TypeAlias annotation was needed before Python 3.10.

Starting from Python 3.10, it's only needed in case of forward references, as those cause the entire type to be quoted:
```py
FooType = str | int
BarType: TypeAlias = 'str | SomeClassLater'
```

Is there some other reason to keep TypeAlias annotation in type aliases?

james-garner-canonical: The way I see it is that you can, in practice, get away without using `TypeAlias` if you don't have forward refs, but you should use it anyway, because that's how you make it clear that this is a type alias, not a variable for use at runtime, and it's best to be explicit.

The [docs](https://typing.python.org/en/latest/spec/aliases.html#typealias) give a nice example of the ambiguity that the `TypeAlias` annotation resolves (edited by me here):
```py
x: TypeAlias = int  # type alias
x: type[int] = int  # typed global expression
x = int  # untyped global expression (typically treated as a type alias)
```

I think type checking in Python is cleanest when the boundary between type checking specific constructs and runtime constructs are kept clear (for example, annotations vs most other things), even though the implementation means that there's some overlap like `typing.cast`.

Perhaps something to discuss at daily?

james-garner-canonical: PS: It wasn't always *needed* pre 3.10 either -- as today, you could get away without using it except for cases where you had to use stringified annotations (which now I guess is only forward refs).

dimaqq: discussed at the stand-up: explicit annotation may be helpful in the future when we move to Python 3.12 and drop start using the `type` keyword/syntax.

I've put the two TypeAlias annotations back, but I'm not adding any additional ones in this PR.

I want to keep this PR concise.

james-garner-canonical (approved):
Looks good to me, thanks for doing this.

PR #2121, opened 2025-10-20
fix: raise ActionFailed when using Context as a context manager

tonyandrewmeyer:
This PR ensures that `ActionFailed` is raised both when using `Context.run()` directly, and when it is used as a context manager, by moving the relevant code to a method common to both techniques.

To reproduce the issue, start with this setup:

```python
>>> from ops import testing
>>> import ops
>>> class C(ops.CharmBase):
...     def __init__(self, framework):
...         super().__init__(framework)
...         framework.observe(self.on['act'].action, self._on_act)
...     def _on_act(self, event):
...         event.fail("Failed")
...         
```

With a regular run, when an action fails, `ActionFailed` is raised. This is the expected behaviour:

```python
>>> ctx = testing.Context(C, meta={'name': 'foo'}, actions={'act': {}})
>>> ctx.run(ctx.on.action('act'), testing.State())
  File "<python-input-5>", line 1, in <module>
    ctx.run(ctx.on.action('act'), testing.State())
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tameyer/w/operator/testing/src/scenario/context.py", line 830, in run
    raise ActionFailed(
    ...<2 lines>...
    )
ops._private.harness.ActionFailed: Failed
```

However, when using `Context` as a context manager, it is not:

```python
>>> with ctx(ctx.on.action('act'), testing.State()) as mgr:
...     mgr.run()
...     
State(config={}, relations=frozenset(), networks=frozenset(), containers=frozenset(), storages=frozenset(), opened_ports=frozenset(), leader=False, model=Model(name='dETzC56WsRPvSb3pDPY8', uuid='912539dd-57d4-453d-9e90-166d04fef696', type='kubernetes', cloud_spec=None), secrets=frozenset(), resources=frozenset(), planned_units=1, deferred=[], stored_states=frozenset({StoredState(name='_stored', owner_path=None, content={'event_count': 4}, _data_type_name='StoredStateData')}), app_status=UnknownStatus(), unit_status=UnknownStatus(), workload_version='')
```

The issue here is that `Manager.run()` and `Context.run()` use `Context._run()` to call `ops.run()` (too many "run"s!), but the code that handles raising `ActionFailed` (and setting the logs and action messages) is only in `Context.run()`.

### DIFF
--- a/testing/tests/test_e2e/test_actions.py
+++ b/testing/tests/test_e2e/test_actions.py
@@ -134,7 +134,7 @@ def test_action_event_fail_context_manager(mycharm):
     ctx = Context(mycharm, meta={'name': 'foo'}, actions={'foo': {}})
     with pytest.raises(ActionFailed) as exc_info:
         with ctx(ctx.on.action('foo'), State()):
-            assert False, "ActionFailed should be raised in the context manager."
+            assert False, 'ActionFailed should be raised in the context manager.'
     assert exc_info.value.message == 'action failed!'
 
 


testing/src/scenario/context.py
```diff
@@ -843,9 +830,24 @@ def _run(self, event: _Event, state: State):
...
+            self.action_logs.clear()
+            if self.action_results is not None:
+                self.action_results.clear()
+            self._action_failure_message = None
+
         with runtime.exec(
             state=state,
             event=event,
-            context=self,  # type: ignore
+            context=self,
```
tonyandrewmeyer, line 844: The type checker now tells me this `type: ignore` is not required. I'm not sure how this PR changes anything with that, but removing it is nice anyway.

testing/src/scenario/context.py
```diff
@@ -843,9 +830,24 @@ def _run(self, event: _Event, state: State):
...
         ) as ops:
             yield ops
+
+        if event.action:
+            if self._action_failure_message is not None:
```
benhoyt: I know you just moved this, but I think this would be better as `if event.action and self._action_failure_message is not None: ...`

testing/tests/test_e2e/test_actions.py
```diff
@@ -109,6 +109,35 @@ def handle_evt(_: CharmBase, evt: ActionEvent):
...
+        if not isinstance(evt, ActionEvent):
+            return
+        evt.fail('action failed!')
+
+    mycharm._evt_handler = handle_evt
+
+    ctx = Context(mycharm, meta={'name': 'foo'}, actions={'foo': {}})
+    with pytest.raises(ActionFailed) as exc_info:
+        with ctx(ctx.on.action('foo'), State()):
+            pass  # should not reach here
```
james-garner-canonical:
```suggestion
            assert False, "ActionFailed should be raised in the context manager."
```

testing/src/scenario/context.py
```diff
@@ -843,9 +830,23 @@ def _run(self, event: _Event, state: State):
...
-            context=self,  # type: ignore
+            context=self,
         ) as ops:
             yield ops
+
+        if event.action and self._action_failure_message is not None:
+            raise ActionFailed(
+                self._action_failure_message,
+                state=self._output_state,  # type: ignore
+            )
```
james-garner-canonical, line 852: We should be able to `assert self._output_state is not None` here, since `run` does so immediately after we exit from `_run`.

If we add that, does this `# type: ignore` also go away? (I know it was present in the original).

testing/tests/test_e2e/test_actions.py
tonyandrewmeyer: I didn't have this because it's duplicating the functionality of the `pytest.raises`, but I guess the duplication doesn't hurt.

james-garner-canonical: Good point about the duplication, but yeah agreed that it doesn't hurt.

testing/src/scenario/context.py
tonyandrewmeyer: > If we add that, does this `# type: ignore` also go away? (I know it was present in the original).

No. That `type: ignore` is an annoying byproduct of merging Harness and Scenario into `ops.testing`. Both Harness and Scenario have an `ActionFailed` class, so when we put them both in the same namespace one had to "win", and for backwards compatibility that was the Harness one. But they're not quite identical - in Scenario there's the `state` attribute. So Harness gained that attribute, meaningless in Harness, and it's typed as a Harness `State` class, because Harness knows nothing about Scenario.

It's all a bit messy, although it does work. Cleaning it up is too much for this PR, I think.

james-garner-canonical: Thanks for the explanation. Agreed that it's out of scope for this PR. Maybe still worth adding the `assert` so that whoever does clean it up someday at least knows that it's not `None` at this point.

PR #2123, opened 2025-10-21
docs: Clarify that Context is the testing context not only the Juju context

tonyandrewmeyer:
Small tweak to the `Context` docstring to clarify that it contains the broad testing context, and that the Juju context is only part of that.

Consistently use `mgr` rather than `manager` (without re-opening the debate about whether this should be `resource` or `event` etc).

Add the missing attributes to the documentation:
* `app_trusted`
* `juju_version`
* `exec_history`

`charm_spec` is still not included, but, although it's public, the type is not, and I don't think we want to encourage its use (I think we would make this private if we could). Similarly, the two capture settings are also not included, and again don't seem to be meant for public use.


testing/src/scenario/context.py
```diff
@@ -429,8 +430,8 @@ class Context(Generic[CharmType]):
     Each ``Context`` instance is in principle designed to be single-use:
     ``Context`` is not cleaned up automatically between charm runs.
 
-    Any side effects generated by executing the charm, that are not rightful part of the
-    ``State``, are in fact stored in the ``Context``:
+    Any side effects generated by executing the charm, that are not rightfully part of the
```
dwilding: I think it would be nice to clarify what "not rightfully part of the state" means. I struggled to come up with a neat explanation, but how about something like this:

```
If the charm generates data that doesn't belong in :class:``State``, this data is stored in ``Context``. Roughly speaking, ``State`` is for data that charms can normally access. ``Context`` stores any other generated data:
```

tonyandrewmeyer: That definition is correct, I think, but it doesn't feel right to me, because typically the charm caused the side effects, so, while it can't get access to what has happened, it does know about it. For example, if the version is changed twice, the charm knows what versions were set, but can't read that, or for logs, the charm knows (mostly) what was logged, even though it can't read it back.

I'm also a bit reluctant to add a new definition for `State`.

What about this?

```suggestion
    Any side effects generated by executing the charm, that are not part of the mocked Juju
```

dwilding: Your suggestion sounds good, let's go with that

tonyandrewmeyer:
> Idea: I wonder if it's worth deprecating the public `charm_spec` attribute in a separate PR.

Interesting thought. I can only find two uses:

* [grafana-agent-operator](https://github.com/canonical/grafana-agent-operator/blob/8ef089753a0e5e58d9ed26becec753fe67db6608/tests/unit/test_cos_agent_e2e.py#L207) - this is using `charm_spec` to get the charm Class and adjust a private attribute on it. I would have thought the test could just import the charm class and do it directly - the `charm_type` isn't a copy, it's the same class object.
* [charmcraft](https://github.com/canonical/charmcraft/blob/main/tests/spread/commands/init-extensions/tests/unit/test_charm.py) (!) - I guess this is a legitimate use (so perhaps we shouldn't deprecate and maybe should document), pulling the container name from the metadata rather than hard-coding it in the test. The metadata isn't available elsewhere (you can get it from the model, but not until you've already called `run`). However, [from_context()](https://documentation.ubuntu.com/ops/latest/reference/ops-testing/#ops.testing.State.from_context) would do this for you in a nicer way, so maybe the right thing is to encourage that. (The test is using `scenario` as the namespace, so maybe it's quite old?).

PR #2126, opened 2025-10-21
docs: update the Juju release the metrics functionality is removed from 4.0 to 3.6.11

tonyandrewmeyer:
The metrics events (of which we only have `CollectMetrics`) and `add-metrics` hook tool were scheduled for removal in Juju 4.0, but were removed in 3.6.11. Update the documentation to reflect that.

### DIFF
--- a/testing/src/scenario/mocking.py
+++ b/testing/src/scenario/mocking.py
@@ -727,8 +727,8 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
         labels: Mapping[str, str] | None = None,
     ) -> NoReturn:
         raise NotImplementedError(
-            'add-metrics is not implemented in Scenario (and never will be: '
-            'it was removed in Juju 3.6.11)',
+            'add-metrics is not implemented in Scenario '
+            '(and never will be: it was removed in Juju 3.6.11)'
         )
 
     def resource_get(self, resource_name: str) -> str:


dimaqq (approved):

testing/src/scenario/mocking.py
```diff
@@ -727,8 +727,8 @@ def add_metrics(
         labels: Mapping[str, str] | None = None,
     ) -> NoReturn:
         raise NotImplementedError(
-            'add-metrics is not implemented in Scenario (and probably never will be: '
-            "it's deprecated API)",
+            'add-metrics is not implemented in Scenario (and never will be: '
+            'it was removed in Juju 3.6.11)',
```
james-garner-canonical: Scope creep: nicer (imo) linebreak, and remove (possibly error prone) trailing comma:
```suggestion
            'add-metrics is not implemented in Scenario '
            '(and never will be: it was removed in Juju 3.6.11)'
```

tonyandrewmeyer: I dislike the leading space, but happy to do the other adjustments.

dwilding (approved):

testing/src/scenario/mocking.py
james-garner-canonical: Thanks. I like leading spaces when using implicit string concatenation like this, as it lets subsequent lines be added or removed without modifying preceding lines, but not a big deal either way.

Issue #2125, opened 2025-10-21
ops[testing] Secret.owner inconsistency, 'application' vs 'app'


dimaqq:
When a charm creates a secret, the Scenario (output state) Secret owner reads `'application'`.

It's set along this path:
```py
  /code/operator/testing/tests/test_e2e/test_secrets.py(158)test_add()
-> charm.app.add_secret({'foo': 'bar'}, label='mylabel')
  /code/operator/ops/model.py(520)add_secret()
-> id = self._backend.secret_add(
  /code/operator/testing/src/scenario/mocking.py(431)secret_add()
-> secret = Secret(
  <string>(14)__init__()
> /code/operator/testing/src/scenario/state.py(379)__post_init__()
# self.owner == 'application'
```

The production code in question is `ops.Model.add_secret`:
https://github.com/canonical/operator/blob/159aaf36ca382479c792cda4979a0cdaa2cdb9da/ops/model.py#L527-L534

Meanwhile, the declared type for this field should be `'app'` instead:
https://github.com/canonical/operator/blob/159aaf36ca382479c792cda4979a0cdaa2cdb9da/testing/src/scenario/state.py#L351-L357

This was introduced or fixed incompletely here: https://github.com/canonical/operator/commit/a8bb04d98178534cc651e91beb06c4e30ae3f438

The docs were updated in 1da9684e5

dimaqq:
@tonyandrewmeyer wdyt, is it possible to fix this without breaking our users?
- `application` -> `app` may break unit tests if the owner is validated
- `app` -> `application` may break static analysis if/when the test secret is created

dimaqq:
My cursory search (charm-analysis and github) doesn't show a single unit test that verifies Scenario secret owner except for the one test that I wrote.

dimaqq:
I think there are three ways this could be fixed...
- rename Scenario's 'app' back to 'application' to be consistent with Juju --> PR TBD
   - add a backwards-compatibility rename app->application for secrets created in the test setup
- translate 'application' to 'app' in mock backend --> #2127
   - no public API changes, small internal inconsistency between ops and scenario
- update Ops to use 'app' throughout, translate hook command 'app' to 'application' in hookcmd --> https://github.com/canonical/operator/pull/2128
   - would make the hook command API inconsistent with Juju
- update Ops to use 'app' in the framework, translate 'app' to 'application' in the backend --> PR TBD
[...]

PR #2127, opened 2025-10-21
fix: normalise Secret.owner to 'app' for ops[testing] output state

dimaqq:
`ops.Model.add_secret()` currently sets `owner='application'`, while Scenario’s declared type for `Secret.owner` is `Literal['unit', 'app', None]`, causing an inconsistency in test (resulting state) output.

This PR implements option 1a: translate 'application' → 'app' in the Scenario backend when constructing the output state, with no public API changes and no changes to the model or hookcmd call path.

- keeps Ops (runtime path) as-is
- preserves Juju-like semantics in hookcmds
- aligns Scenario with its documented 'app' literal

The API stays the same, and most or majority of unit tests don't get broken.
Output state secret owner string could be validated in tests, but so far I didn't find a test like this across GitHub or the charms that we track the source code for.

### DIFF
--- a/testing/src/scenario/mocking.py
+++ b/testing/src/scenario/mocking.py
@@ -424,7 +424,7 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
         description: str | None = None,
         expire: datetime.datetime | None = None,
         rotate: SecretRotate | None = None,
-        owner: Literal['unit', 'app'] | None = None,
+        owner: Literal['unit', 'application'] | None = None,
     ) -> str:
         from .state import Secret
 
@@ -434,7 +434,8 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
             description=description,
             expire=expire,
             rotate=rotate,
-            owner=owner,
+            # It's called 'application' in Ops, but 'app' in Scenario.
+            owner='app' if owner == 'application' else owner,
         )
         secrets = set(self._state.secrets)
         secrets.add(secret)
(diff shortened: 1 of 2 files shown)


dimaqq:
We've got two alternatives here:

opt 1a, as in this PR:
- no change in model
- translate application to app in Scenario backend
- no change in backend calling hookcmd

opt 1b. have Ops use 'app' internally:
- model would call `backend.foo('app')`
- no change needed in Scenario's backend
- prod backend would translate app -> application before calling hookcmd

james-garner-canonical:
I don't fully understand the context for this PR.

dimaqq:
James, https://github.com/canonical/operator/issues/2125 is mandatory reading here I'm afraid.
Meanwhile, I'll write up a better summary and title.

dimaqq:
I've updated the title and descirption. PTAL.

tonyandrewmeyer:
> Looks good to me, though I guess this should be `fix!` so it gets called out properly in the next release notes, even if we suspect no one is using it.

I feel like `!` is for cases when the documented behaviour changes. Otherwise, if it's for any breaking change, it seems like almost all bug fixes would need a `!`. For our two most recent ones, the machine ID was documented as being `int` typed and that changed to `str`, but the `ctx` one not raising `UncaughtCharmError` was much more borderline, because we don't actually have almost any docs on that behaviour. Arguably, the `!` was wrong there, although calling out the change is the right thing to do, I think.

Here, we already documented that the value would be `app` and the PR fixes things so we do what the documentation says.

PR #2128, opened 2025-10-21
fix: scenario secret owner value (option 2)

dimaqq:


dimaqq:
This approach was rejected in https://github.com/canonical/operator/issues/2125

PR #2130, opened 2025-10-21
test: expand secrets integration and state transition tests

dimaqq:
In preparation for secrets refactor, add a good dose of secrets tests.

This PR adds one more action to the test charm, state transition and integration tests to exercise it.
The action is parametrised with a sequence of ops secret API operations that charm performs during a single hook.
That's designed to validate the "collapsing" or "coalescing" or secret operations.

Done in preparation for refactoring secrets code in Ops: #1536

### DIFF
--- a/test/integration/test_secrets.py
+++ b/test/integration/test_secrets.py
@@ -106,9 +106,10 @@ def test_set_secret(
     flow: str,
     lookup_by: Literal['id', 'label'],
 ):
-    label_count = collections.Counter(flow.split(','))['label']
-    if label_count >= 2 and juju.status().model.version.startswith('3.2.'):
+    counts = collections.Counter(flow.split(','))
+    if counts['label'] >= 2 and juju.status().model.version.startswith('3.2.'):
         pytest.skip('Label is sticky on Juju 3.2 in this case')
+
     if lookup_by == 'id':
         params = {'flow': flow, 'secretid': fresh_secret}
     else:
@@ -127,7 +128,6 @@ def test_set_secret(
     secret = juju.show_secret(secrets[0].uri, reveal=True)
     common_checks(secret, result)
 
-    counts = collections.Counter(flow.split(','))
     if counts['content']:
         assert result['after']['latest'] == {'val': str(counts['content'])}
     if counts['label']:


dimaqq:
2120 merged into main and main merged into this branch.
should be reviewable now!

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +82,53 @@ def add_with_meta(self, event: ops.ActionEvent):
         }
         event.set_results(cast('dict[str, Any]', result))
 
+    def set_secret_flow(self, event: ops.ActionEvent):
+        result: Result = {}
+        contentiter = iter(contents())
+        labeliter = iter(labels())
+        descriptioniter = iter(descriptions())
+        expireiter = iter(expires())
+        rotateiter = iter(rotates())
```
james-garner-canonical: Isn't calling `iter` here unnecessary? They're already generators.
```suggestion
        contentiter = contents()
        labeliter = labels()
        descriptioniter = descriptions()
        expireiter = expires()
        rotateiter = rotates()
```

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +82,53 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+
+                match field:
+                    case 'content':
+                        secret.set_content(next(contentiter))
+                    case 'label':
+                        new_label = next(labeliter)
+                        secret.set_info(label=new_label)
+                        if secretlabel:
+                            # So that we can find the secret again
+                            secretlabel = new_label
```
james-garner-canonical: I think this might be significantly easier to follow if we had a separate action and handler for the `id` and `label` cases.

It should be possible to have two handlers without significant code duplication -- for example, the iterators could be attributes of the charm instance, and the `match` logic can then be abstracted away into a function taking a secret and a case and returning a label.

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +82,53 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+                        secret.set_info(rotate=next(rotateiter))
+                    case _:
+                        raise ValueError(f'Unsupported {field=}')
+
+            if not secretid:
+                secretid = self.model.get_secret(label=secretlabel).get_info().id
+            result['secretid'] = secretid
+            result['after'] = self._snapshot(secretid)
+        except Exception as e:
+            result['exception'] = str(e)
```
james-garner-canonical: How about using [event.fail](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.ActionEvent.fail) here, since the tests first `assert` that we don't have an exception?

test/charms/test_secrets/src/charm.py
```diff
@@ -91,5 +140,34 @@ def _snapshot(self, secret_id: str) -> SecretSnapshot:
...
+    """Generate predictable, but different content values every time."""
+    for i in range(1, 10):
+        yield datetime.datetime(2010 + i, 1, 1, 0, 0, 0)
+
+
+def rotates() -> Generator[ops.SecretRotate]:
+    """Generate predictable, but different `rotate` values every time."""
+    yield from ops.SecretRotate.__members__.values()
+
+
```
james-garner-canonical: How about defining these as generator expressions in the charm instead for less boilerplate?

They could even be attributes of the charm itself to easily make them available in the different event handlers if we go with separate ID and label based handlers.

e.g.
```python
        self.secret_contents = ({'val': f'{i}'} for i in range(1, 10))
        self.secret_labels = (f'label{i}' for i in range(1, 10))
        self.secret_descriptions = (f'description{i}' for i in range(1, 10))
        self.secret_expures = (datetime.datetime(2010 + i, 1, 1, 0, 0, 0) for i in range(1, 10))
        self.secret_rotates = iter(ops.SecretRotate.__members__.values())
```

Could also use `itertools.count(start=1)` instead of `range(1, 10)` to avoid the arbitrary limit of 9 times, but I guess it's not that relevant since the 'rotate' case is going to be limited to 7 times.

test/integration/test_secrets.py
```diff
@@ -102,6 +164,17 @@ def cleanup(juju: jubilant.Juju, leader: str) -> None:
             juju.remove_secret(secret.uri)
 
 
+@pytest.fixture
+def good_secret(juju: jubilant.Juju, leader: str, cleanup: None) -> str:
+    """Remove all old secrets and add a new secret owned by the test app."""
+    juju.exec('secret-add --label thelabel some=content', unit=leader)
+    secrets = juju.secrets()
+    assert secrets
+    assert secrets[0].owner == 'test-secrets'
```
james-garner-canonical: Looks like the docstring is out of date, as we don't seem to remove old secrets here.

Also I'm not sure why we are `secrets[0]` here?

test/integration/test_secrets.py
```diff
@@ -91,8 +92,69 @@ def test_add_with_meta(juju: jubilant.Juju, cleanup: None, leader: str, fields:
...
+    assert 'after' in result
+    assert result['after']
+    assert result['after']['info']
+    info = result['after']['info']
+
+    secrets = juju.secrets()
+    secret = juju.show_secret(secrets[0].uri, reveal=True)
+    common_checks(secret, result)
+
+    counts = collections.Counter(flow.split(','))
```
james-garner-canonical: This is a clever way of making the result predictable, since each fields value just depends on the number of times it was called.

testing/tests/test_e2e/test_secrets.py
```diff
@@ -661,17 +662,65 @@ def test_add_secret_with_metadata(secrets_context: Context[SecretsCharm], fields
...
+    if counts['rotate']:
+        rotation_values = ['sentinel', *SecretRotate.__members__.values()]
+        assert info['rotation'] == rotation_values[counts['rotate']]
+
+
 def common_assertions(scenario_secret: Secret | None, result: Result):
     if scenario_secret:
-        assert scenario_secret.owner == 'application'
+        # https://github.com/canonical/operator/issues/2125
+        assert scenario_secret.owner in ('app', 'application')
```
tonyandrewmeyer: Rather than have the test pass either way, I think we should have one value here. Either 'app' and get the other issue fixed first, or 'application' and adjust this in the other PR.

testing/tests/test_e2e/test_secrets.py
```diff
@@ -661,17 +662,65 @@ def test_add_secret_with_metadata(secrets_context: Context[SecretsCharm], fields
...
+        'label,content,label,content',
+    ],
+)
+def test_set_secret(
+    secrets_context: Context[SecretsCharm], flow: str, lookup_by: Literal['id', 'label']
+):
+    secret = Secret({'some': 'content'}, owner='app', id='theid', label='thelabel')
+    state = State(leader=True, secrets={secret})
+    params = {'flow': flow}
+    match lookup_by:
```
tonyandrewmeyer: It doesn't seem like `match` is doing anything here than a simple if/else would provide. Or even `params[f'secret{lookup_by}'] = f'the{lookup_by'}'`.

testing/tests/test_e2e/test_secrets.py
```diff
@@ -661,17 +662,65 @@ def test_add_secret_with_metadata(secrets_context: Context[SecretsCharm], fields
...
+        case 'id':
+            params['secretid'] = 'theid'
+        case 'label':
+            params['secretlabel'] = 'thelabel'
+    state = secrets_context.run(secrets_context.on.action('set-secret-flow', params=params), state)
+    scenario_secret = next(iter(state.secrets))
```
tonyandrewmeyer: I think generally it's better to explicitly ask for the secret, rather than the "convert to Iterator and get first item" paradigm. Something like:

```suggestion
    scenario_secret = state.get_secret(id='theid') if lookup_by == 'id' else state.get_secret(label='thelabel')
```

testing/tests/test_e2e/test_secrets.py
```diff
@@ -661,17 +662,65 @@ def test_add_secret_with_metadata(secrets_context: Context[SecretsCharm], fields
...
+    result = cast('Result', secrets_context.action_results)
```
tonyandrewmeyer, line 687: I have a slight preference for an assert is not None rather than this, but fine either way.

test/integration/test_secrets.py
tonyandrewmeyer: The `cleanup` fixture removes all the application and user secrets.

test/integration/test_secrets.py
```diff
@@ -91,8 +92,72 @@ def test_add_with_meta(juju: jubilant.Juju, cleanup: None, leader: str, fields:
...
+        pytest.skip("Label is sticky on Juju 3.2 in this case")
+    params = {'flow': flow}
+    match lookup_by:
+        case 'id':
+            params['secretid'] = good_secret
+        case 'label':
+            params['secretlabel'] = 'thelabel'
+
+    rv = juju.run(leader, 'set-secret-flow', params=params)
+    result = cast('Result', rv.results)
```
tonyandrewmeyer, line 119: Similarly here, a small preference for an assert instead.

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +82,53 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+        rotateiter = iter(rotates())
+
+        try:
+            secretid = event.params.get('secretid')
+            secretlabel = event.params.get('secretlabel')
+
+            for field in event.params['flow'].split(','):
+                assert secretid or secretlabel
+                assert not (secretid and secretlabel)
+                secret = self.model.get_secret(
```
tonyandrewmeyer: This seems a complicated way to do this. Wouldn't it be easier to read if kwargs was defined outside the call? Like:

```python
kwargs = {}
if secretid:
    kwargs['id'] = secretid
if secretlabel:
    kwargs['label'] = secretlabel
```

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +82,53 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+                    **({'id': secretid} if secretid else {}),
+                    **({'label': secretlabel} if secretlabel else {}),
+                )
+
+                match field:
```
tonyandrewmeyer, line 104: Is `match` providing anything here that a simple if/elif/else wouldn't?

test/charms/test_secrets/src/charm.py
tonyandrewmeyer: Do we need the try/except at all? It seems simpler to let it fail, which for Scenario would bubble up the exception wrapped in UncaughtCharmError and the test would (correctly) fail, and for Jubilant the charm would go to error state (which I assume means again the test correctly fails).

+1 on event.fail if we *expect* a failure, but I don't think we should catch unexpected ones.

test/charms/test_secrets/src/charm.py
tonyandrewmeyer: I agree we should get rid of the arbitrary 9. You shouldn't need `iterations.count`, just an `i=1` at the start, a `while True:` and an `i+=1`. Unless you prefer using the itertools version, but I think it's easier for people less familiar with itertools to have it explicit.

And a basic `while True` for the rotate, or itertools.cycle if you prefer that style.

tonyandrewmeyer:
Oh, and I think a more detailed PR description would be good, with a short summary of what tests are being added here, maybe also why. And avoid "scenario" in the PR title since it's presumably going to be the commit message and therefore the default change log entry.

test/charms/test_secrets/src/charm.py
james-garner-canonical: This was my first thought too, but the problem is that the label may be updated in this loop, and then we want to use the new label, though only if we were using the label initially, otherwise we keep using the id (I guess it could update the kwargs dict though).

That's what lead me to suggest a separate handler for the `id` and `label` cases.

test/integration/test_secrets.py
james-garner-canonical: Ah, thanks for the explanation.

test/charms/test_secrets/src/charm.py
dimaqq: This is a great idea.
I'll start with locals rather than instance members for now.

test/charms/test_secrets/src/charm.py
dimaqq: thank you, I'll convert this to `event.fail(); return`

test/charms/test_secrets/src/charm.py
dimaqq: I've refactored this a bit, PTAL when I push the changes.

dimaqq:
Thank you both for thorough reviews.
I've addressed about 1/2 of the comments so far.

testing/tests/test_e2e/test_secrets.py
dimaqq: I'll see what I can do here.
The label may have been changed by the action...
But I can always find it by the original (state_in) id. 💡

test/charms/test_secrets/src/charm.py
dimaqq: refactored. yes label may be updated.

testing/tests/test_e2e/test_secrets.py
dimaqq: I'll keep it like this for a moment, hoping that #2127 is merged first, whereupon I can replace this with `owner == 'app'`.

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +81,54 @@ def add_with_meta(self, event: ops.ActionEvent):
         }
         event.set_results(cast('dict[str, Any]', result))
 
+    def set_secret_flow(self, event: ops.ActionEvent):
+        secretid = event.params.get('secretid')
+        secretlabel = event.params.get('secretlabel')
+        contentses = ({'val': f'{i}'} for i in itertools.count(1))
+        labels = (f'label{i}' for i in itertools.count(1))
+        descriptions = (f'description{i}' for i in itertools.count(1))
+        expires = (datetime.datetime(2010 + i, 1, 1, 0, 0, 0) for i in itertools.count(1))
```
james-garner-canonical, line 90:
```suggestion
        contentses = ({'val': f'{i}'} for i in itertools.count(start=1))
        labels = (f'label{i}' for i in itertools.count(start=1))
        descriptions = (f'description{i}' for i in itertools.count(start=1))
        expires = (datetime.datetime(2010 + i, 1, 1, 0, 0, 0) for i in itertools.count(start=1))
```
Might as well make the argument's meaning explicit, since as @tonyandrewmeyer noted, readers are likely less familiar with `itertools.count` than `range`.

test/integration/test_secrets.py
```diff
@@ -91,8 +90,71 @@ def test_add_with_meta(juju: jubilant.Juju, cleanup: None, leader: str, fields:
...
+)
+def test_set_secret(
+    juju: jubilant.Juju,
+    fresh_secret: str,
+    leader: str,
+    flow: str,
+    lookup_by: Literal['id', 'label'],
+):
+    label_count = collections.Counter(flow.split(','))['label']
+    if label_count >= 2 and juju.status().model.version.startswith('3.2.'):
```
james-garner-canonical: Move `counts` construction to the start of the test so it's not duplicated, and to make it clear that it depends only on the test inputs, not the outputs?
```suggestion
    counts = collections.Counter(flow.split(','))
    if counts['label'] >= 2 and juju.status().model.version.startswith('3.2.'):
```

test/integration/test_secrets.py
james-garner-canonical: I don't think it can be an `assert` because `Result` is a `TypedDict` -- but maybe that indicates it should be named `ResultDict`?

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +81,54 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+
+            if not secretid:
+                secretid = self.model.get_secret(label=secretlabel).get_info().id
+            result['secretid'] = secretid
+            result['after'] = self._snapshot(secretid)
+        except Exception as e:
+            event.fail(str(e))
+            return
+
+        event.set_results(cast('dict[str, Any]', result))
```
james-garner-canonical, line 130:
```suggestion
        else:
            event.set_results(cast('dict[str, Any]', result))
```

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +81,54 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+
+        try:
+            for field in event.params['flow'].split(','):
+                if secretid:
+                    secret = self.model.get_secret(id=secretid)
+                elif secretlabel:
+                    secret = self.model.get_secret(label=secretlabel)
+                else:
+                    event.fail('Must provide secretid or secretlabel')
+                    return
```
james-garner-canonical, line 102: I still think this would be much easier to follow with separate actions.

test/integration/test_secrets.py
```diff
@@ -91,8 +90,71 @@ def test_add_with_meta(juju: jubilant.Juju, cleanup: None, leader: str, fields:
...
+        pytest.skip('Label is sticky on Juju 3.2 in this case')
+    params = {'flow': flow}
+    match lookup_by:
+        case 'id':
+            params['secretid'] = fresh_secret
+        case 'label':
+            params['secretlabel'] = 'thelabel'
```
james-garner-canonical: This also makes me think separate tests with parametrized actions would be nicer, I think it's good to avoid conditional control flow inside the test itself where possible.

test/integration/test_secrets.py
james-garner-canonical: Does this work?
```py
result: ResultDict = juju.run(leader, 'set-secret-flow', params=params)
```

testing/tests/test_e2e/test_secrets.py
james-garner-canonical: Echoing my comments on the other instance:
- can't be `assert` because `TypedDict`
- maybe should be `ResultDict` instead of `Result` for clarity
- can we use an annotation instead of a `cast` to narrow the type?

test/charms/test_secrets/src/charm.py
james-garner-canonical: Also `contentses` is probably better as `contents`, despite the [humour value](https://en.wiktionary.org/wiki/hobbitses).

testing/tests/test_e2e/test_secrets.py
```diff
@@ -661,17 +662,60 @@ def test_add_secret_with_metadata(secrets_context: Context[SecretsCharm], fields
...
         assert result.get('after')
         info = result['after']['info']
         # Verify that the unit and the scaffolding see the same data
-        assert scenario_secret.id == info['id']
+        #
+        # Scenario presents a secret with a full secret URI to the charm
+        # however, the id on the scenario Secret object is a plain id
+        assert scenario_secret.id.split('/')[-1] == info['id'].split('/')[-1]
         assert scenario_secret.label == info['label']
-        assert scenario_secret._tracked_revision == info['revision']
```
tonyandrewmeyer, line 674: Since this is being removed, as it's not always going to be the case now, should there be some `assert scenario_secret...` type assertions in the individual tests to replace it? At the moment we have tests in this common function that the Scenario object matches the data loaded in the charm code, and in the individual test checks that the data loaded in the charm code matches the initial state. Maybe that's all you're interested in, in which case you can just ignore this.

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +81,54 @@ def add_with_meta(self, event: ops.ActionEvent):
...
         event.set_results(cast('dict[str, Any]', result))
 
+    def set_secret_flow(self, event: ops.ActionEvent):
+        secretid = event.params.get('secretid')
+        secretlabel = event.params.get('secretlabel')
+        contentses = ({'val': f'{i}'} for i in itertools.count(1))
+        labels = (f'label{i}' for i in itertools.count(1))
+        descriptions = (f'description{i}' for i in itertools.count(1))
+        expires = (datetime.datetime(2010 + i, 1, 1, 0, 0, 0) for i in itertools.count(1))
+        rotates = iter(ops.SecretRotate.__members__.values())
```
tonyandrewmeyer: Since all the others are now infinite, I'd make this one infinite as well, via `itertools.cycle`. The assert should just need a `% length`, I would think.

test/charms/test_secrets/src/charm.py
tonyandrewmeyer: For what it's worth, I think the existing style where there's a `return` and you get to dedent, is a better choice.

tonyandrewmeyer: For what it's worth, I think the existing style where there's a `return` and you get to dedent, is a better choice. It also aligns with our recommendation to always explicitly return after fail.

test/charms/test_secrets/src/charm.py
```diff
@@ -80,6 +81,54 @@ def add_with_meta(self, event: ops.ActionEvent):
...
+                        secret.set_info(rotate=next(rotates))
+                    case _:
+                        raise ValueError(f'Unsupported {field=}')
+
+            if not secretid:
+                secretid = self.model.get_secret(label=secretlabel).get_info().id
+            result['secretid'] = secretid
+            result['after'] = self._snapshot(secretid)
+        except Exception as e:
+            event.fail(str(e))
```
tonyandrewmeyer, line 127: I still feel like it would be cleaner to have the actual exception bubble up in Scenario. The integration test is going to raise `TaskError`, which will show the failure message either way (my comment in an earlier thread about going to error status was wrong - I forgot that wouldn't happen with an action). But if you strongly prefer this, then leave it as-is.

tonyandrewmeyer (approved):
I think the commit category should be "tests" (or "test"? I never remember if it's plural) not "chore", right? This seems like something worth mentioning in our change log and release notes, given that we mention all the CI items.

A couple of small comments/suggestions, including wondering why we're using match/case when a perfectly good if/else would do the same job. But happy to approve optimistically.

test/charms/test_secrets/src/charm.py
dimaqq: nice trick :)
I reckon the assert doesn't need to be modified, because the number of updates is hardcoded when the test is parametrised.

dimaqq:
> can we use an annotation instead of a cast to narrow the type?

yes, but then I have to slap `# type: ignore` which could shadow some other typing nit.

test/charms/test_secrets/src/charm.py
dimaqq: When run against Juju, I'd want to see and be able to assert on the text representation of the exception.
For example in tests like "can't get info when not a leader".
Unhandled exception can be read, but I feel that's more finicky.

```
stderr='Uncaught ZeroDivisionError in charm code: division by zero\nUse `juju debug-log` to see the full traceback.\n'
```

test/integration/test_secrets.py
dimaqq: `Type "dict[str, Any]" is not assignable to declared type "Result"`

test/charms/test_secrets/src/charm.py
dimaqq: Not really, it's pretty similar.

testing/tests/test_e2e/test_secrets.py
dimaqq: done now

dimaqq:
Let me think about the case for/against separate actions for id/label.

test/charms/test_secrets/src/charm.py
james-garner-canonical: Return after fail is a great point, thanks.

PR #2132, opened 2025-10-22
fix: detect categories with an explanation mark indicating breaking changes

### LINKED ISSUE #2111: Release script doesn't detect breaking changes
When creating the changelog, the release script organises the changes by the conventional commit category (fix, feat, etc). However, if the category has the breaking-change `!` marker, it doesn't detect it.

We should at minimum include these - but perhaps also do something with that information like call it out in the changes.

tonyandrewmeyer:
Currently, the release script fails to recognise lines with the `!` breaking change indicator in the category, meaning they are not included in the change log or release notes (and yet, are probably the most important to include).

The PR:
* Changes the regular expression to use named groups for better readability.
* Changes the regular expression to capture the `!` if present.
* Puts any breaking changes into a separate category (keeping the main category in the description).
* In the changelog, puts the changes into a "Breaking Changes" group.
* In the release notes, adds a top-level section that calls out that there are breaking changes, and lists them, and notifies the user about this, because it's likely that some hand-crafted text will be required to cover these.

Also, to handle the upcoming release, adds support for alpha, beta, and release candidate numbers in the version (in a fairly minimal way - the tag, the title, and generally not crashing).

Fixes #2111

### DIFF
--- a/release.py
+++ b/release.py
@@ -205,7 +205,10 @@ def parse_release_notes(release_notes: str) -> tuple[dict[str, list[tuple[str, s
                 description = description[0].upper() + description[1:]
                 pr_link = match.group('pr').strip()
                 if match.group('breaking') == '!':
-                    categories['breaking'].append((f'{category}: {description}', pr_link))
+                    categories['breaking'].append((
+                        f'{category.capitalize()}: {description}',
+                        pr_link,
+                    ))
                 else:
                     categories[category].append((description, pr_link))
 


release.py
```diff
@@ -189,16 +192,19 @@ def parse_release_notes(release_notes: str) -> tuple[dict[str, list[tuple[str, s
...
     }
     full_changelog_line = None
 
     for line in release_notes.splitlines():
-        if match := re.match(r'^\* (\w+): (.*) by [^ ]+ in (.*)', line.strip()):
-            category = match.group(1)
+        if match := re.match(CHANGE_LINE_REGEX, line.strip()):
+            if match.group('breaking') == '!':
+                category = 'breaking'
+            category = match.group('category').strip()
```
james-garner-canonical: It seems like this won't work as-is, since we immediately overwrite `category`.

But I wonder if we should do something like this instead:
```py
            category = match.group('category').strip()
            if category in categories:
                description = match.group('summary').strip()
                description = description[0].upper() + description[1:]
                pr_link = match.group('pr').strip()
                if match.group('breaking') == '!':
                    categories['breaking'].append((f'{category}: {description}', pr_link))
                else:
                    categories[category].append((description, pr_link))
```

So the original category is retained in the messages for the breaking changes.

release.py
```diff
@@ -216,7 +222,19 @@ def format_release_notes(
     If `full_changelog` is provided, it is appended at the end.
     """
     lines = ["## What's Changed", '']
+    if categories['breaking']:
```
james-garner-canonical, line 230: Probably `categories` should be typed as `Mapping` to make it clear you can't safely mutate it here (which you don't), as we pass it to `format_changes` as well -- otherwise it's not clear locally. Not necessarily in scope for this PR.

1. If we have a mutatable copy here (we don't, but could take a copy), we could `categories.pop` and avoid the need for the continue in the loop over `categories` later.
2. Does `format_changes` need to be updated to correctly format breaking changes too? Or does that 'just work' by putting them in their own category?

release.py
```diff
@@ -189,16 +192,19 @@ def parse_release_notes(release_notes: str) -> tuple[dict[str, list[tuple[str, s
         'refactor': [],
         'perf': [],
         'revert': [],
+        'breaking': [],  # a meta category for breaking changes
```
james-garner-canonical: How about put this one first, so in `format_changes` the `breaking` category is at the top?

james-garner-canonical (changes requested):
I think there's a bug. The approach looks good overall though.

Would it be possible to see the output for an example release (e.g. this planned one) with both versions before merging?

tonyandrewmeyer:
> I think there's a bug.

Oops, yes. I broke things moving some stuff around just before committing.

> Would it be possible to see the output for an example release (e.g. this planned one) with both versions before merging?

Yeah, I was being a bit lazy thinking I would test it with the release for today rather than go to the effort of organising a separate release. But that would be better I guess. Will do in a moment after addressing the other comments.

release.py
tonyandrewmeyer: Ah, I'm too old :laughing:. I was thinking it was a `dict` so order didn't matter, and it wasn't worth complicating the changes code to put it first. But indeed, since dicts preserve order these days I can just put it at the top. Thanks!

release.py
tonyandrewmeyer: > Probably `categories` should be typed as `Mapping` to make it clear you can't safely mutate it here (which you don't), as we pass it to `format_changes` as well -- otherwise it's not clear locally. Not necessarily in scope for this PR.

Not really, but simple enough to do, so I'll adjust that.

> 1. If we have a mutatable copy here (we don't, but could take a copy), we could `categories.pop` and avoid the need for the continue in the loop over `categories` later.

It doesn't feel significantly better to me to `pop` rather than have a simple guard.

> 2. Does `format_changes` need to be updated to correctly format breaking changes too? Or does that 'just work' by putting them in their own category?

Yeah, my thought for the changelog was to keep it simple by just using the same formatting but break them out into a separate group (your suggestion of keeping the original category is probably good too - I also wondered about that but wasn't sure if it would be worth doing or not).

release.py
tonyandrewmeyer: Yeah, I'm happy to do that. I wasn't sure whether we should or not, but I think the extra info is better than the extra consistency.

benhoyt (approved):
I'm happy to just go with James' review on this one, seeing it's the release script. Thanks!

release.py
james-garner-canonical: I like `pop` because then you don't have to "remember" about guarding later, because it's all taken care of by:
```py
if breaking_changes := categories.pop('breaking', None)
```
But the current form is indeed fine.

release.py
```diff
@@ -123,11 +127,11 @@ def get_new_tag_for_release(
...
 
         if not new_tag:
-            logger.error('Error: No tag specified and no suggestion available')
+            logger.error('No tag specified and no suggestion available')
             continue
 
-        if not re.match(r'^\d+\.\d+\.\d+$', new_tag):
-            logger.error('Error: Tag must be in format X.Y.Z')
+        if not re.match(r'^\d+\.\d+\.\d+(?:(?:a|b|rc)\d+)?$', new_tag):
+            logger.error('Tag must be in format X.Y.Z')
```
james-garner-canonical, line 134: I guess this doesn't handle pre and post releases either. Perhaps saner to parse with `packing.version`, but probably fine for now.

tonyandrewmeyer:
Release notes look like this:

image

Changelog looks like this:

image

release.py
tonyandrewmeyer: Yeah, since this wasn't the main point of the PR and I just needed it to get the script to run, I did the minimal change. I agree using the official parser would be better, but I think that's a separate PR and more thought should be put into how the pieces are used.

Issue #1875, opened 2025-07-02
Why did real pebble tests not fail?


dimaqq:
The 3.0 PR forgot to bump the ops version in operator/version.py

lint, static and unit tests all failed.
but (real) pebble tests did not.

```
Run tox -e py3.10-pebble
py3.10-pebble: venv> /home/runner/.local/share/uv/tools/tox/bin/uv venv -p cpython3.10 --allow-existing --python-preference system /home/runner/work/operator/operator/.tox/py3.10-pebble
py3.10-pebble: uv-sync> uv sync --locked --python-preference system --no-default-groups --group unit -p cpython3.10
py3.10-pebble: freeze> /home/runner/.local/share/uv/tools/tox/bin/uv --color never pip freeze
py3.10-pebble: annotated-types==0.7.0,exceptiongroup==1.3.0,importlib-metadata==8.7.0,iniconfig==2.1.0,jsonpatch==1.33,jsonpointer==3.0.0,opentelemetry-api==1.34.1,opentelemetry-sdk==1.34.1,opentelemetry-semantic-conventions==0.55b1,-e file:///home/runner/work/operator/operator,-e file:///home/runner/work/operator/operator/testing,-e file:///home/runner/work/operator/operator/tracing,packaging==25.0,pluggy==1.6.0,pydantic==2.11.5,pydantic-core==2.33.2,pytest==8.3.5,pyyaml==6.0.2,tomli==2.2.1,typing-extensions==4.14.0,typing-inspection==0.4.1,websocket-client==1.8.0,zipp==3.23.0
py3.10-pebble: commands_pre[0]> bash -c 'umask 0; pebble run --http='"'"':4000'"'"' --create-dirs &>/dev/null& sleep 1'
py3.10-pebble: commands[0]> pytest -v --tb native test/test_real_pebble.py
```

Note the editable install.
Other tox envs don't do that.
Maybe it's an issue with tox envs not inheriting all the settings all the way up.
Or something else.

benhoyt:
Begs the question: are we testing against the correct ops? Needs investigation.

james-garner-canonical:
I tried to reproduce this locally by down-bumping the versions in Ops's `version.py` (leaving all the other version specifiers unchanged), but lint, unit, and real pebble all pass for me.

[...]

PR #2139, opened 2025-10-23
ci: test if package versions match dependency versions before publishing

james-garner-canonical:
This PR adds steps to the `build-n-publish` job of the `publish` workflow to validate that the versions of the packages are compatible with each others dependencies (on each other), including the extras specified by `ops`.

The new steps run after building and before attesting and publishing.
1. ~Use `python` to extract the extras to a `constraints.txt` file.~
2. Use `uvx` to install the three built packages together, respecting the `constraints.txt` file.

Fixes: #1875

## Testing

I've tested the steps locally, by running `uv build --all`, and then running the code from the two steps sequentially. The steps all pass.

I repeated the testing steps twice more:
1. Changed the `ops` version to `1`. The `uvx` step fails with:
```
  × No solution found when resolving tool dependencies:
  ╰─▶ Because there is no version of ops==3.4.0.dev0 and ops-scenario==8.4.0.dev0 depends on ops==3.4.0.dev0, we can conclude that ops-scenario==8.4.0.dev0 cannot be used.
      And because you require ops-scenario==8.4.0.dev0, we can conclude that your requirements are unsatisfiable.
```
2. Reverted (1) and changed the version in `testing/pyproject.toml` to `0.0`. The `uvx` step fails with:
```
  × No solution found when resolving tool dependencies:
  ╰─▶ Because there is no version of ops-scenario==8.4.0.dev0 and you require ops-scenario==8.4.0.dev0, we can conclude that your requirements are unsatisfiable.
```

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -24,17 +24,10 @@ jobs:
         with:
           enable-cache: false
       - run: uv build --all
-      - name: Create constraints file for test
-        shell: python
+      - name: Test if built package versions and dependency versions match
         run: |
-          import itertools, pathlib, tomllib
-          toml = tomllib.loads(pathlib.Path('pyproject.toml').read_text())
-          extras = toml['project']['optional-dependencies']
-          constraints = '\n'.join(itertools.chain(extras['testing'], extras['tracing']))
-          print(constraints)
-          pathlib.Path('constraints.txt').write_text(constraints)
-      - name: Test if package versions match dependency versions
-        run: uvx --with dist/ops-*.whl --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl --constraints constraints.txt python -c 'print("OK")'
+          set -xueo pipefail
+          uvx --with "$(echo -n dist/ops-*.whl)[testing,tracing]" --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl python -c 'import ops, scenario, ops_tracing; print("OK")'
       - uses: actions/attest-build-provenance@v3
         with:
           subject-path: 'dist/*'


.github/workflows/publish.yaml
```diff
@@ -24,6 +24,17 @@ jobs:
         with:
           enable-cache: false
       - run: uv build --all
+      - name: Create constraints file for test
+        shell: python
+        run: |
+          import itertools, pathlib, tomllib
+          toml = tomllib.loads(pathlib.Path('pyproject.toml').read_text())
+          extras = toml['project']['optional-dependencies']
+          constraints = '\n'.join(itertools.chain(extras['testing'], extras['tracing']))
```
benhoyt: These are plain lists, so I'd prefer the simpler version (and drop `import itertools`):
```suggestion
          constraints = '\n'.join(extras['testing'] + extras['tracing'])
```

.github/workflows/publish.yaml
```diff
@@ -24,6 +24,17 @@ jobs:
         with:
           enable-cache: false
       - run: uv build --all
+      - name: Create constraints file for test
+        shell: python
+        run: |
+          import itertools, pathlib, tomllib
```
james-garner-canonical:
```suggestion
          import pathlib, tomllib
```

.github/workflows/publish.yaml
james-garner-canonical: Good call!

.github/workflows/publish.yaml
```diff
@@ -24,6 +24,17 @@ jobs:
...
+        shell: python
+        run: |
+          import pathlib, tomllib
+          toml = tomllib.loads(pathlib.Path('pyproject.toml').read_text())
+          extras = toml['project']['optional-dependencies']
+          constraints = '\n'.join(extras['testing'] + extras['tracing'])
+          print(constraints)
+          pathlib.Path('constraints.txt').write_text(constraints)
+      - name: Test if package versions match dependency versions
+        run: uvx --with dist/ops-*.whl --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl --constraints constraints.txt python -c 'print("OK")'
```
dimaqq: **Praise:** Wow this is very clever. And short. Elegant.

I wonder what corner cases we may have missed.
The constraints file lists ops-tracing and ops-scenario but not ops. How come?

My spot testing seems to suggest that an empty constraints file already catches some (or all?) mismatches, for example:

```diff
--- a/tracing/pyproject.toml
+++ b/tracing/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 dependencies = [
     "opentelemetry-api~=1.0",
     "opentelemetry-sdk~=1.30",
-    "ops==3.4.0.dev0",
+    "ops==3.4.3.dev0",
     "pydantic",
 ]
```

```command
> echo > constraints.txt
> uvx --with dist/ops-*.whl --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl --constraints constraints.txt python -c 'print("OK")'
warning: Requirements file `constraints.txt` does not contain any dependencies
  × No solution found when resolving tool dependencies:
  ╰─▶ Because there is no version of ops==3.4.3.dev0 and ops-tracing==3.4.0.dev0 depends on ops==3.4.3.dev0, we can conclude that
      ops-tracing==3.4.0.dev0 cannot be used.
      And because only ops-tracing==3.4.0.dev0 is available and you require ops-tracing, we can conclude that your requirements are unsatisfiable.
```

Let's clarify what the constraints file is needed for in a comment.

james-garner-canonical: The constraints file captures the `ops` requirements for `testing` and `tracing` because they're in `extras` -- so unless the extra is specified at install time (which we're blocked from using here due to it breaking the glob), then that requirement isn't checked.

The `testing` and `tracing` requirements for `ops` are specified in their regular dependencies, so a bad version of `ops` or a bad dependency specification for `ops` will be caught without the `constraints.txt`.

---

Actually I think we could get away without `constraints.txt` by doing this:

```suggestion
        run: |
          set -xueo pipefail
          shopt -s nullglob
          OPS=dist/ops-*.whl
          uvx --with "$OPS[testing,tracing]"  --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl python -c 'print("OK")'
```

dimaqq: The last can be expressed simpler as:

```sh
uvx --with dist/ops-*.whl'[testing,tracing]' ...
```

P.S. I wonder if the python command should be something like `import ops, ops_tracing, scenario; print("OK")` 🤔

james-garner-canonical: Ah, good use of single quotes there, thanks! Good call on testing library imports too, will do.

james-garner-canonical:
I've dropped the separate constraints building step in favour of using `--with ...[testing,tracing]`. The single quotes suggestion didn't work directly, so I've used a subshell for concise in-lining.

I've tested this new change in both directions: changing the `ops` version to `1`, then running `uv build --all` and:
```bash
> uvx --with "$(echo -n dist/ops-*.whl)[testing,tracing]" --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl python -c 'import ops, scenario, ops_tracing; print("OK")'
++ echo -n dist/ops-1-py3-none-any.whl
+ uvx --with 'dist/ops-1-py3-none-any.whl[testing,tracing]' --with dist/ops_scenario-8.4.0.dev0-py3-none-any.whl --with dist/ops_tracing-3.4.0.dev0-py3-none-any.whl python -c 'import ops, scenario, ops_tracing; print("OK")'
  × No solution found when resolving tool dependencies:
  ╰─▶ Because ops[testing]==1 depends on ops-scenario==8.4.0.dev0 and ops-scenario==8.4.0.dev0 depends on ops==3.4.0.dev0, we can conclude that ops[testing]==1 depends on
      ops==3.4.0.dev0.
      And because only ops[testing]==1 is available and you require ops[testing], we can conclude that your requirements are unsatisfiable.
```

Reverting this version change, then changing the `testing` version to `1`, rebuilding, and:
```bash
> uvx --with "$(echo -n dist/ops-*.whl)[testing,tracing]" --with dist/ops_scenario-*.whl --with dist/ops_tracing-*.whl python -c 'import ops, scenario, ops_tracing; print("OK")'
++ echo -n dist/ops-3.4.0.dev0-py3-none-any.whl
+ uvx --with 'dist/ops-3.4.0.dev0-py3-none-any.whl[testing,tracing]' --with dist/ops_scenario-1-py3-none-any.whl --with dist/ops_tracing-3.4.0.dev0-py3-none-any.whl python -c 'import ops, scenario, ops_tracing; print("OK")'
  × No solution found when resolving tool dependencies:
  ╰─▶ Because only ops[testing]==3.4.0.dev0 is available and ops[testing]==3.4.0.dev0 depends on ops-scenario==8.4.0.dev0, we can conclude that all versions of ops[testing] depend
      on ops-scenario==8.4.0.dev0.
      And because there is no version of ops-scenario==8.4.0.dev0 and you require ops[testing], we can conclude that your requirements are unsatisfiable.
```

And when the versions match, everything is OK.

benhoyt:
@james-garner-canonical Can this one be merged?

.github/workflows/publish.yaml
```diff
@@ -24,6 +24,10 @@ jobs:
         with:
           enable-cache: false
       - run: uv build --all
+      - name: Test if built package versions and dependency versions match
+        run: |
+          set -xueo pipefail
```
dimaqq, line 29: Our AI friend tells me that:

> On Ubuntu runners, GitHub Actions runs your run: | steps with:
> `bash --noprofile --norc -e -o pipefail {0}`

So it looks like this line may not be needed?

Apparently that's when `shell: bash` but not the default...

james-garner-canonical: Ah yeah, weird stuff -- here are the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell).

If you don't specify a shell, it's `bash -e {0}` on Linux or MacOS, or `sh` if `bash` isn't found (while Windows has a different default, `pwsh`). On the other hand, if you explicitly request `bash` on any platform, you get `bash --noprofile --norc -eo pipefail {0}`, like you said.

This seems odd enough and easy enough to get wrong that I'd be inclined to keep using `set -xueo pipefail` explicitly instead of relying on `shell: bash`.

PR #2141, opened 2025-10-23
test: Re-enable scenario consistency checks after disabling them

james-garner-canonical:
This PR corrects one of the `testing` tests to correctly re-enable consistency checks after turning them off. They default to on, and we run our own tests with them on.

Before this PR, they would be turned off, as we check `if os.getenv(...)`, which will be `True` when the variable is set to the string `'false'`.

### DIFF
--- a/testing/tests/test_e2e/test_rubbish_events.py
+++ b/testing/tests/test_e2e/test_rubbish_events.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import os
-
 import pytest
 from ops.charm import CharmBase, CharmEvents
 from ops.framework import EventBase, EventSource, Framework, Object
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):
Interesting that nothing else had to be fixed. I guess we normally check consistency more directly, or have things consistent.

tonyandrewmeyer:
> Would this be easier to follow in future if `test_rubbish_event_raises` became two separate tests? One test that checks for `AttributeError` on the non-pebble-ready events, and another that checks for `AttributeError` on `kazoo_pebble_ready` (then consistency checks could be set & unset unconditionally)

It'd probably be easier to use `monkeypatch.setenv` then too, which seems like it would be the typical way to do this and ensure the cleanup. But good with me either way.

james-garner-canonical:
I've refactored to use two tests as you suggested, @dwilding. Would you mind having another quick look before merge?

dwilding (approved):
Looks really clear to me, thanks!

PR #2142, opened 2025-10-23
feat: env var to control exception wrapping in tests

### LINKED ISSUE #1428: Better Scenario error message surfacing

james-garner-canonical:
This PR adds a new environment variable, `SCENARIO_BARE_CHARM_ERRORS`. If this is set to `1` or `true` (case insensitive), then exceptions raised in charm code in state-transition tests are re-raised as-is. If unset or set to any other value, then the default behaviour matches the current behaviour: charm errors are wrapped in `UncaughtCharmError` and that is raised instead.

The idea would be to make `SCENARIO_BARE_CHARM_ERRORS=1` the default behaviour in a future major release. In the meantime, the environment variable allows charms to explicitly opt into the new behaviour now, for a better debugging experience, or explicitly lock in the current behaviour to avoid future breakages.

The name is up for debate, but was chosen to follow the format of `SCENARIO_SKIP_CONSISTENCY_CHECKS`.

The `testing` tests have been updated to exercise both behaviours, while the `ops` tests continue to assume the existing behaviour, and will require updates when we eventually change the default.

`tox.ini` has been updated to always pass the scenario environment variables along.

Resolves #1428.

### DIFF
--- a/testing/src/scenario/_runtime.py
+++ b/testing/src/scenario/_runtime.py
@@ -341,7 +341,8 @@ class Runtime:
             except (NoObserverError, ActionFailed):
                 raise  # propagate along
             except Exception as e:
-                if os.getenv('SCENARIO_BARE_CHARM_ERRORS', 'false').lower() in {'1', 'true'}:
+                bare = os.getenv('SCENARIO_BARE_CHARM_ERRORS', 'false')
+                if bare.lower() == 'true' or (bare.isdigit() and int(bare)):
                     raise
                 # The following is intentionally on one long line, so that the last line of pdb
                 # output shows the error message (pdb shows the "raise" line).
--- a/testing/tests/test_runtime.py
+++ b/testing/tests/test_runtime.py
@@ -179,15 +179,19 @@ class ValueErrorCharm(ops.CharmBase):
 @pytest.mark.parametrize(
     ('expected_error', 'bare_charm_errors'),
     (
-        # '1' and 'true' (case-insensitive) are treated as logically true
+        # any non-zero string that `isdigit`
         (ValueError, '1'),
+        (ValueError, '007'),
+        # 'true' (case-insensitive)
         (ValueError, 'true'),
         (ValueError, 'True'),
         (ValueError, 'tRuE'),
-        # any other value is treated as logically false
+        # falsey digits
         (UncaughtCharmError, '0'),
+        (UncaughtCharmError, '00'),
+        # any other value
+        (UncaughtCharmError, '-1'),
         (UncaughtCharmError, 'false'),
-        (UncaughtCharmError, '11'),
         (UncaughtCharmError, 'yes'),
         (UncaughtCharmError, '✩ anything ✨ else ✧'),
         # the actually unset case is tested separately


testing/src/scenario/_environ.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import os
+
+
+def wrap_charm_errors() -> bool:
```
benhoyt: I think it's quite unusual for a boolean environment variable to be true by default (in fact, often setting it means true regardless of value). So I'd prefer we invert the name and sense, perhaps `SCENARIO_BARE_CHARM_ERRORS`.

testing/src/scenario/_environ.py
```diff
@@ -0,0 +1,10 @@
...
+    """Return whether scenario should wrap charm errors with ``UncaughtCharmError``."""
+    default = 'True'
+    return os.getenv('SCENARIO_WRAP_CHARM_ERRORS', default).capitalize() != 'False'
```
benhoyt: In my experience it's more typical to use `=1` rather than `=true` (or `=false`) for environment variables. Also, nit, but I don't think we need to name the *default* variable. Let's do this:

```python
return os.getenv('SCENARIO_BARE_CHARM_ERRORS', 'false') in ['1', 'true']
```

testing/tests/test_context.py
```diff
@@ -9,6 +9,7 @@
 from scenario import Context, State
 from scenario.errors import UncaughtCharmError
 from scenario.state import _Event, _next_action_id
+from scenario._environ import wrap_charm_errors
```
benhoyt: I don't think we should be pulling in this private code in tests. We could just have the test read the environment variable directly. This is too dependent on the exact structure of the implementation.

However, I think we should do it an alternative way: keep the tests which test the existing behaviour, and add a test or two that sets `SCENARIO_BARE_CHARM_ERRORS=1`  and tests that behaviour too. This would also simplify running the tests.

In addition, I don't think we should have a whole separate `_environ` module  just for this one-liner helper. I reckon just put it in `_runtime.py`.

.github/workflows/framework-tests.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: ["3.10", "3.12", "3.14"]
+        wrap-charm-errors: ["True", "False"]
```
benhoyt: Per comment further down, I'd prefer we keep this as is, and just add a test that sets the env var and tests the bare exception is raised. This would simplify considerably, and make it so running the unit tests locally would also test both behaviours.

testing/src/scenario/_environ.py
james-garner-canonical: This is forward-looking naming, since the intention is that the default will eventually be false, but I don't think it's a big deal either way. Happy for it to start out named such that the default is false, and eventually switch to being true by default.

testing/src/scenario/_environ.py
james-garner-canonical: Good idea to support additional booly values like `0` and `1`.

I named the default to call attention to it, since that's the line that will change in future -- but I don't mind inlining it.

testing/tests/test_context.py
james-garner-canonical: > I don't think we should be pulling in this private code in tests.

I figured it was fine since scenario's tests already use its private, internal things (e.g. the line above).

> We could just have the test read the environment variable directly. This is too dependent on the exact structure of the implementation.

The intention was to make the tests *less* dependent on the environment variable name, default, and semantics.  I dislike having the tests ready the environment directly for that reason. But if we switch to them only writing to the environment as you suggest, then the usefulness of the abstraction goes away.

> However, I think we should do it an alternative way: keep the tests which test the existing behaviour, and add a test or two that sets `SCENARIO_BARE_CHARM_ERRORS=1` and tests that behaviour too. This would also simplify running the tests.

Good idea. That just means we'll be modifying all the tests that assume `UncaughtCharmError` later when we change the default value of the environment variable, but that's not too bad.

> In addition, I don't think we should have a whole separate _environ module just for this one-liner helper. I reckon just put it in _runtime.py.

If we're not using it in tests, then this makes sense, it's where I had the logic initially -- we don't even need the helper if the tests don't need to use the logic, so I'll just in-line it all.

.github/workflows/framework-tests.yaml
james-garner-canonical: Sounds reasonable. This means it will be more work to change the default, since we'll need to update all the tests to reflect the behaviour change instead of pre-emptively testing both settings, but keeps this PR simpler.

james-garner-canonical: Thinking about using parametrized tests for this, avoids CI changes, and makes it so local unit tests test both behaviours, while also exercising our existing test logic with both settings.

testing/src/scenario/_environ.py
james-garner-canonical: I've changed the name and default as suggested.

testing/src/scenario/_environ.py
james-garner-canonical: This is now in-lined at the point where we raise `UncaughtCharmError`.

testing/tests/test_context.py
james-garner-canonical: I've dropped the helper and updated the tests to just set the environment variable (not read it), using `monkeypatch.setenv`.

.github/workflows/framework-tests.yaml
james-garner-canonical: I've reverted the changes to this file, so there are no CI changes now.

james-garner-canonical:
I've refactored as per @benhoyt's suggestions:
- Dropped the CI changes
- Inverted the name and default (now `SCENARIO_BARE_CHARM_ERRORS` defaulting to false)
- Dropped the helper function and in-lined the logic at the place we raise `UncaughtCharmError`

I've parametrized the existing tests that caught `UncaughtCharmError` to set the environment variable using `monkeypatch.setenv`, which both exercises the environment variable logic and ensures that our expectations about the shape and type of exceptions hold in both cases. I think this is the right approach for the internal `testing` tests, but I'm less sure about the noise it adds to `test/test_model_relation_data_class.py` -- maybe the `ops` tests should just assume the current default.

dimaqq (commented):
How will this env var be documented?

The other env var used to be documented: https://pypi.org/project/ops-scenario/4.0.4.1/
It's no longer documented today.

TheJJ:
I've implemented the alternative approach that avoids environment variables, and instead controls this through `Context` in #2145.

TheJJ:
Thanks for reconsidering. I'm convinced solving this via environment variable is the wrong place to solve this - the test execution becomes dependent on an environment variable, instead of being self-contained and self-descriptive.

Just executing the tests directly without the "correct env" will lead to failures that are not obvious to resolve (and then in the wrong layer: env instead of test configuration).

Instead this exception handling should be configurable per-test (imagine having hundreds to migrate), and be self-contained from within the tests.

If you'd like, we can discuss this in person during the engineering sprint :)

james-garner-canonical:
I lean towards the environment variable as I think it will give us a better pathway to changing the default behavior in future, but I'm definitely open to discussing alternatives.


Regarding the developer experience, it's worth noting that you can still define the behavior in your tests with an environment variable, for example with an autouse fixture. To control the behavior per-test with relatively low boilerplate you could instead use pytest.mark.usefixtures.

benhoyt:
@james-garner-canonical For the record, decision is to use this (env var) approach.

james-garner-canonical:
I've updated this PR for review. The PR description reflects the current implementation.

@dimaqq raised a good question -- thoughts?

> How will this env var be documented?
> 
> The other env var used to be documented: https://pypi.org/project/ops-scenario/4.0.4.1/ It's no longer documented today.

testing/tests/test_context.py
```diff
@@ -96,7 +96,8 @@ def test_app_name_and_unit_id():
     assert ctx.unit_id == 42
 
 
-def test_context_manager_uncaught_error():
+@pytest.mark.parametrize('bare_charm_errors', ('1', ''))
```
tonyandrewmeyer: I feel like these should all have `''`, `'0'`, `'1'`, `'false'`, `'true'`, `'FALSE'`, `'TRUE'`. Maybe also a `None` that skips putting anything in the env at all?

That will break the `error = ` line - the cleanest thing would be to have the expected error as part of a paired param.

testing/src/scenario/_runtime.py
```diff
@@ -331,6 +331,8 @@ def exec(
             except (NoObserverError, ActionFailed):
                 raise  # propagate along
             except Exception as e:
+                if os.getenv('SCENARIO_BARE_CHARM_ERRORS', 'false').lower() in {'1', 'true'}:
```
tonyandrewmeyer: I vaguely remember that there were subtle differences between `os.getenv`/`os.setenv` and `os.environ`, but don't remember exactly what they were (something to do with whether it just changed the Python dictionary or the actual env?) or whether that's still true in modern Python.

Is this the same as `os.environ.get('SCENARIO_BARE_CHARM_ERRORS', 'false')`? If so, is there a reason to use `getenv` over it?

tonyandrewmeyer (approved):
Thanks! I think this will be quite nice for many people.

Everything seems good to me, but I think we should expand the tested cases. Approving optimistically.

tonyandrewmeyer:
> How will this env var be documented?

I'd just mention it in [`UncaughtCharmError`](https://documentation.ubuntu.com/ops/latest/reference/ops-testing/#ops.testing.errors.UncaughtCharmError), which it's most closely related to, to start with, and if people still complain about it then we could put it in the how-to perhaps.

testing/src/scenario/_runtime.py
james-garner-canonical: I suspect the subtle issue you're remembering is that [os.environ](https://docs.python.org/3/library/os.html#os.environ) is read once and then cached, and may not reflect updates made to the environment since then.

To avoid issues with this when *writing*, it seems best to use `os.environ[...] =` over [os.putenv](https://docs.python.org/3/library/os.html#os.putenv). `os.environ` uses `os.putenv` under the hood, but direct calls to `os.putenv` will bypass updating `os.environ`.

[os.getenv](https://docs.python.org/3/library/os.html#os.getenv) however *does* read from `os.environ` internally, so I believe there's no difference between using `getenv` over `environ.get`. My inclination would be to use the more specific function, but I don't object to changing to `os.environ.get`.

testing/tests/test_context.py
james-garner-canonical: I'm OK with making this change, but I'll wait to see what @benhoyt thinks. I also wonder if we could just have one function that explicitly tests all those variants, and leave the rest as-is.

testing/src/scenario/_runtime.py
tonyandrewmeyer: Ah, thanks, that was it I think. Happy to leave it to you to choose.

testing/tests/test_context.py
tonyandrewmeyer: One test that explicitly tests all of those would be fine with me.

james-garner-canonical: I'll do that then :)

benhoyt: I agree it'd be good to have a few more variants. However, I don't love if statements in parametrized code, and don't love repeating code logic in the tests (eg: `val.lower() in {'1', 'true'}`). I reckon having two different tests would be simplest (the falses and the trues). Alternatively use tuples for the parametrization with `bare_charm_errors,expect_bare` with for example `('1', True)`.

testing/tests/test_e2e/test_relations.py
```diff
@@ -623,7 +623,8 @@ def test_broken_relation_not_in_model_relations(mycharm):
         assert charm.model.relations['foo'] == []
 
 
-def test_get_relation_when_missing():
+@pytest.mark.parametrize('bare_charm_errors', ('1', ''))
```
benhoyt: Do we need the testing of this logic repeated in all these places? Or could we just have it once in `test_context.py` or wherever, knowing that the higher-level code uses the context manager? (I assume it does.) Seems a bit unnecessary to test the env var logic in multiple places.

testing/src/scenario/_runtime.py
james-garner-canonical: Will leave it as is.

testing/tests/test_context.py
james-garner-canonical: I've added the variant tests [here](https://github.com/canonical/operator/pull/2142/commits/cab78113edcc3875b8ad8f80987005217189d0d1), PTAL.

testing/tests/test_e2e/test_relations.py
james-garner-canonical: Yeah, I guess doing this two all the tests had two purposes, one being exercising the code paths, which I agree could be confined to just `test_context`, and the other being making sure that our tests handle both states of the environment variable.

It seemed undesirable to me that setting `SCENARIO_BARE_CHARM_ERRORS=1` would break lots (or any) of our tests. Currently it'll break some `ops` tests (since I removed the conditional logic from there), but all the `testing` tests will pass. But maybe that's OK actually?

james-garner-canonical: An alternative option for env var proofing would be if these tests just became a bit looser, they could all use something like this instead of parametrizing:
```py
with pytest.raises((UncaughtCharmError, TheActualError)):
```
Though I guess you'd still need to get conditional to check any of the error contents ...

benhoyt: I like that idea -- do the parametrized thing in one place, then the `with pytest.raises((UncaughtCharmError, TheActualError))` trick everywhere else.

james-garner-canonical: I'm now wondering if it would be even cleaner to just `monkeypatch.setenv('SCENARIO_BARE_CHARM_ERRORS', 'false')`  in the tests where we don't care about testing both behaviours to make explicit which behaviour we're testing. I'll iterate on this next week.

testing/tests/test_runtime.py
```diff
@@ -118,7 +121,8 @@ def test_env_clean_on_charm_error():
             assert 'JUJU_REMOTE_APP' in os.environ
             _ = 1 / 0  # raise some error
     # Ensure that some other error didn't occur (like AssertionError!).
-    assert 'ZeroDivisionError' in str(exc.value)
+    exc = exc_info.value if bare_charm_errors else exc_info.value.__cause__
+    assert isinstance(exc, ZeroDivisionError)
```
dimaqq: +1 on the discussion in the other comment that the code flow should not depend on the value of the test parameter.

Then most tests could read something like:

```suggestion
    with pytest.raises((Uncaught, ZeroDivisionError), match="foo"):
        ...
        
    # if really needed
    exc = exc_info.value
    assert isinstance(exc, ZeroDivisionError) or isinstance(exc.__cause__, ZeroDivisionError)
```

testing/tests/test_runtime.py
```diff
@@ -153,13 +158,50 @@ def _on_start(self, _: ops.StartEvent):
...
+        raise ValueError()
+
+
+@pytest.mark.parametrize(
+    ('expected_error', 'bare_charm_errors'),
+    (
+        (ValueError, '1'),
+        (ValueError, 'tRuE'),  # case insensitive
+        (UncaughtCharmError, ''),
+        (UncaughtCharmError, '✨ aNy ✩ sTrInG ✧ ReAlLy ...'),
```
dimaqq: Nice, but maybe too clever.

How about:

```suggestion
        (ValueError, '1'),
        (ValueError, 'True'),
        (ValueError, 'true'),
        (UncaughtCharmError, '0'),
        (UncaughtCharmError, 'False'),
        (UncaughtCharmError, 'false'),
        (UncaughtCharmError, ''),
        (UncaughtCharmError, 'anything else'),
```

I would also swap them around (setting, observable_effect).

Aside: this doesn't test the missing env var case, though I imagine that's covered by regular tests.

james-garner-canonical: > Nice, but maybe too clever.

I figured this made the underlying logic explicit while avoiding a proliferation of test cases, but explicitly testing that `0` and the various `FaLsE` values are handled correctly sounds like a good idea, will do.

> I would also swap them around (setting, observable_effect).

I agree that that's a more natural *ordering*, but I found the (error, input) ordering to be more readable due to alignment and natural *grouping*.

> Aside: this doesn't test the missing env var case, though I imagine that's covered by regular tests.

This is tested in `test_bare_charm_errors_not_set` below.

testing/tests/test_runtime.py
james-garner-canonical: Hiding the conditionality with an `or` is a good idea if needed. I think this problem will go away if we explicitly `monkeypatch` in the environment state we're assuming, so I'll try that.

james-garner-canonical:
I've further minimised the diff in the unit tests compared to `main` via the techniques we discussed, namely:
```py
with pytest.raises((UncaughtCharmError, TheRealError)):
```
And where we do additionaly validation on the `pytest.raises` context object:
```py
monkeypatch.setattr('SCENARIO_BARE_CHARM_ERRORS', 'false')
```

The tests in `testing`'s `test_context` are lightly parametrized `('1', '0')` to test both cases, as these tests explicitly check that the context handles charm errors correctly. Additional tests have been added in `testing`'s `test_runtime` to more exhaustively test the different possible environment variable values.

All `testing` tests pass when running `env SCENARIO_WRAP_CHARM_ERRORS=true tox -e unit` but a number of `ops` tests fail. I believe is what we wanted -- these tests will need to be updated to expect the appropriate non-wrapped error types when we change the default value of this environment variable in a future major release.

james-garner-canonical:
@dimaqq do you want to review this PR further, or are you happy for it to be merged?

testing/tests/test_runtime.py
```diff
@@ -163,3 +165,47 @@ def _on_start(self, _: ops.StartEvent):
...
+    (
+        # '1' and 'true' (case-insensitive) are treated as logically true
+        (ValueError, '1'),
+        (ValueError, 'true'),
+        (ValueError, 'True'),
+        (ValueError, 'tRuE'),
+        # any other value is treated as logically false
+        (UncaughtCharmError, '0'),
+        (UncaughtCharmError, 'false'),
+        (UncaughtCharmError, '11'),
```
dimaqq: I would expect `"11"` to be truthy.

dimaqq (changes requested):

testing/tests/test_runtime.py
dimaqq: That being said, I'm happy to go with a majority vote.

james-garner-canonical: @benhoyt I went with your initial suggestion of testing for 1 and true, but do you think it's worth expanding this at all?

james-garner-canonical: @benhoyt I went with your initial suggestion of testing for 1 and true, but do you think it's worth expanding this at all?

benhoyt: I don't have strong feelings, but I think what you have is fine ("1" or "true"). Dima's right that normally any nonzero integer would be truthy, but no one's going to set it to "11", so I don't think it matters, and I'd rather avoid more churn on this.

james-garner-canonical: Per discussion with @benhoyt on Monday, I've updated the implementation to treat positive integers as true as well, and updated the test cases accordingly.

PR #2143, opened 2025-10-23
fix: don't cache secrets in Ops

dimaqq:
This PR removes the "shared secret metadata cache".

Specific changes:
- direct hook command calls on Juju 3.6 and later
- read-modify-write sequence for Juju 3.5 and earlier
- the logic moved from model to the model backend
- unit tests updated
- integration tests pass as is

Open questions and comments:
- what other areas should be tested?
- what kinds of additional integration tests could be helpful?
- please check Secret object instantiation carefully

Ref #1536

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -3981,14 +3981,17 @@ class _ModelBackend:
         if self._juju_context.version < '3.6':
             # Older Juju series don't coalesce multiple secret updates within a hook.
             # To work around that, we perform a smart read-modify-write cycle.
+            # See https://bugs.launchpad.net/juju/+bug/2081034 for details.
             if content is None:
                 content = self.secret_get(id=id, peek=True)
             if description is None or expire is None or rotate is None or label is None:
+                # Metadata fix is needed for Juju < 3.6 or < 3.5.5
                 info = self.secret_info_get(id=id)
                 description = description or info.description
                 expire = expire or info.expires
                 rotate = rotate or info.rotation
-                label = label or info.label  # The label fix needed for Juju < 3.5
+                # The label fix is needed for Juju < 3.5
+                label = label or info.label
         with self._wrap_hookcmd(
             'secret-set',
             id=id,


ops/model.py
```diff
@@ -1533,24 +1522,13 @@ def set_content(self, content: dict[str, str]):
...
             typing.cast('str', self.id),
             content=content,
-            label=cached_data.get('label'),
-            description=cached_data.get('description'),
-            expire=expire,
-            rotate=cached_data.get('rotate'),
+            label=None,
+            description=None,
+            expire=None,
+            rotate=None,
```
tonyandrewmeyer: `None` is the default, so we could just leave out these lines, no?

ops/model.py
```diff
@@ -1586,28 +1564,9 @@ def set_info(
...
-        cached_data['rotate'] = rotate
-        # Get the previous content from the unit agent's cache.
-        content = (
-            self._backend.secret_get(id=self._id, peek=True) if 'content' in cached_data else None
-        )
-
         self._backend.secret_set(
             typing.cast('str', self.id),
-            content=content,
+            content=None,
```
tonyandrewmeyer: Similarly here, just removing this line seems simpler.

ops/model.py
```diff
@@ -4027,6 +3986,17 @@ def secret_set(
         rotate: SecretRotate | None = None,
     ):
         # The content is None or has already been validated with Secret._validate_content
+        if self._juju_context.version < '3.6':
+            # Older Juju series don't coalesce multiple secret updates within a hook.
+            # To work around that, we perform a smart read-modify-write cycle.
```
tonyandrewmeyer, line 3983: Is there a GitHub issue or a changelog line that we could link to here to make it easier to dig into these changes for anyone that works on this in the future?

ops/model.py
```diff
@@ -4027,6 +3986,17 @@ def secret_set(
...
+            if content is None:
+                content = self.secret_get(id=id, peek=True)
+            if description is None or expire is None or rotate is None or label is None:
+                info = self.secret_info_get(id=id)
+                description = description or info.description
+                expire = expire or info.expires
+                rotate = rotate or info.rotation
+                label = label or info.label  # The label fix needed for Juju < 3.5
```
tonyandrewmeyer: Similarly, if there was a GitHub (launchpad?) issue to link to for this one, that would likely help any future people.

ops/model.py
benhoyt: I think we should just drop these -- they explicitly have `=None` defaults:
```suggestion
```

ops/model.py
benhoyt: This also has an explicit `=None` default, so I think we just drop this too:
```suggestion
```

test/test_model.py
```diff
@@ -3787,7 +3787,18 @@ def test_from_dict(self):
...
+            ops.CharmMeta(),
+            _ModelBackend(
+                'myapp/0',
+                model_uuid='abcd',
+                juju_context=JujuContext._from_dict({'JUJU_VERSION': '3.6.0'}),
+            ),
+        )
+
+    @pytest.fixture
+    def model_pre36(self, fake_juju_version: None):
```
benhoyt, line 3801: Maybe for clarity we should be explicit about the JUJU_VERSION (3.5.0) here.

dimaqq:
We want to get rid of `self._content = self._backend.secret_get(...)` as well, in a separate PR.

dimaqq:
Discussed, add `Model.get_secret(..., validate=True): ...`

benhoyt:
> We want to get rid of self._content = self._backend.secret_get(...) as well, in a separate PR.

We also discussed the awkwardness of `Model.get_secret` now having to call `secret-get` only to throw the content away. We'll add a `validate=True` argument to preserve the default exception-raising behaviour, but for a performance-sensitive charm, charms could set `validate=False`.

ops/model.py
dimaqq: Done, thanks

ops/model.py
dimaqq: done

test/test_model.py
dimaqq: Added. I've picked 3.5.7 the last in 3.5 series.

ops/model.py
dimaqq: I didn't find the issue for this one specifically.
I used to think that labels were separate from metadata, but alas!

ops/model.py
dimaqq: done

ops/model.py
dimaqq: done

PR #2149, opened 2025-10-27
feat: Model.get_secret(..., validate=...) keyword argument; don't cache content on the Secret object

dimaqq:
User-facing API changes:

`Model.get_secret()` gains a new keyword argument, `validate: bool`, which is `True` by default.

With the default setting, `data = Model.get_secret(...).get_content()` makes invocations of the hook command and therefore two calls to the Juju secret backend.

Performance-sensitive applications should use `validate=False`, in which case, `Model.get_secret()` doesn't raise exceptions in simple cases, and only `Secret.get_content()` does.

Note that if both `id` and `label` parameters are supplied, the hook command is always invoked in order to set the secret consumer label.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -286,15 +286,20 @@ class Model:
         owners set a label using ``add_secret``, whereas secret observers set
         a label using ``get_secret`` (see an example at :attr:`Secret.label`).
 
+        Secret content is not cached. If the hook command is run, the result is
+        discarded.
+
         .. jujuadded:: 3.0
             Charm secrets added in Juju 3.0, user secrets added in Juju 3.3
 
         Args:
             id: Secret ID if fetching by ID.
             label: Secret label if fetching by label (or updating it).
-            validate: Try to fetch the secret content from Juju secret backend,
+            validate: Try to fetch the secret content from the Juju secret backend,
                 so that an exception is raised early if this unit doesn't have
                 permission to access the secret or the secret doesn't exist.
+                If `False`, the hook command will only be run when both `id` and
+                `label` arguments are specified.
 
         Raises:
             SecretNotFoundError: If a secret with this ID or label doesn't exist.
@@ -306,6 +311,10 @@ class Model:
         if id is not None:
             # Canonicalize to "secret:<id>" form for consistency in backend calls.
             id = Secret._canonicalize_id(id, self.uuid)
+        # The `validate` keyword is used for backwards compatibility, where
+        # charms expect any errors to be raised in the `get_secret()` call.
+        # If both `id` and `label` are supplied, the charm is trying to set
+        # the consumer label and we must call the hook command.
         if validate or (id and label):
             self._backend.secret_get(id=id, label=label)
         return Secret(self._backend, id=id, label=label)
@@ -1437,7 +1446,7 @@ class Secret:
     def get_content(self, *, refresh: bool = False) -> dict[str, str]:
         """Get the secret's content.
 
-        The content of the secret is no longer cached on the :class:`Secret`,
+        The content of the secret is not cached on the :class:`Secret`;
         every call will get the secret content from the Juju secret backend.
 
         Returns:
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -3848,30 +3848,6 @@ class TestSecretClass:
         assert secret.id is None
         assert secret.label == 'y'
 
-    def test_get_content_cached(self, model: ops.Model, fake_script: FakeScript):
-        fake_script.write('secret-get', """exit 1""")
-        return
-        # FIXME
-
-        secret = self.make_secret(model, id='x', label='y')
-        content = secret.get_content()  # will use cached content, not run secret-get
-        assert content == {'foo': 'bar'}
-
-        assert fake_script.calls(clear=True) == []
-
-    def test_get_content_refresh(self, model: ops.Model, fake_script: FakeScript):
-        fake_script.write('secret-get', """echo '{"foo": "refreshed"}'""")
-
-        return
-        # FIXME
-        secret = self.make_secret(model, id='y')
-        content = secret.get_content(refresh=True)
-        assert content == {'foo': 'refreshed'}
-
-        assert fake_script.calls(clear=True) == [
-            ['secret-get', '--format=json', f'secret://{model._backend.model_uuid}/y', '--refresh']
-        ]
-
     def test_get_content_uncached(self, model: ops.Model, fake_script: FakeScript):
         fake_script.write('secret-get', """echo '{"foo": "notcached"}'""")
 


ops/model.py
```diff
@@ -284,17 +286,15 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
         owners set a label using ``add_secret``, whereas secret observers set
         a label using ``get_secret`` (see an example at :attr:`Secret.label`).
 
-        The content of the secret is retrieved, so calls to
-        :meth:`Secret.get_content` do not require querying the secret storage
-        again, unless ``refresh=True`` is used, or :meth:`Secret.set_content`
-        has been called.
-
```
tonyandrewmeyer, line 291: I think we would be better off keeping (but rewriting) this - it was added because charmers weren't clear on when caching was done. It's a much simpler answer now, but I think we still want it somewhere.

Something like "The content of the secret is not stored, even if `validate` is set to `True`" would probably be enough.

ops/model.py
```diff
@@ -284,17 +286,15 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
...
         .. jujuadded:: 3.0
             Charm secrets added in Juju 3.0, user secrets added in Juju 3.3
 
         Args:
             id: Secret ID if fetching by ID.
             label: Secret label if fetching by label (or updating it).
+            validate: Try to fetch the secret content from Juju secret backend,
```
```suggestion
            validate: Try to fetch the secret content from the Juju secret backend,
```

dimaqq:
I suppose when this gets closer to ops being released (beta or final), I should write up a post on charmhub discourse, outlining our idea of possibly deprecating `validate=True`:
- to set the consumer label, call `Model.get_secret(id=..., label=...)`
- to get the content, call `Model.get_secret(...).get_content()`

The specifics of the future internal change will be up for discussion, we could
- either change the default to a sentinel, like `validate: bool = UNSET`
- or require `validate = False` unless both `id` and `label` are passed

ops/model.py
```diff
@@ -284,17 +286,15 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
...
+                so that an exception is raised early if this unit doesn't have
+                permission to access the secret or the secret doesn't exist.
```
tonyandrewmeyer, line 300: Perhaps also something confirming the `False` case, like "Otherwise, the secret backend is not accessed"? But also covering the "id and label" bit.

ops/model.py
```diff
@@ -1307,7 +1293,6 @@ def __init__(
         backend: _ModelBackend,
         id: str | None = None,
         label: str | None = None,
-        content: dict[str, str] | None = None,
```
tonyandrewmeyer, line 1310: I wonder if we have to keep this here (and not use it, or warn `DeprecationWarning` if it's not None). This is a public class, even though we say to charms that they shouldn't instantiate it directly.

ops/model.py
```diff
@@ -306,13 +306,9 @@ def get_secret(self, *, id: str | None = None, label: str | None = None) -> Secr
...
             # Canonicalize to "secret:<id>" form for consistency in backend calls.
             id = Secret._canonicalize_id(id, self.uuid)
-        content = self._backend.secret_get(id=id, label=label)
-        return Secret(
-            self._backend,
-            id=id,
-            label=label,
-            content=content,
-        )
+        if validate or (id and label):
```
benhoyt, line 318: The "id or label" thing is a good catch. However, now that I realise we need it for that case, maybe it's not worth adding the new `validate` arg, as presumably people will want to apply a label most of the time -- so maybe we should just do it all the time? Or is my assumption not true?

ops/model.py
```diff
@@ -1453,9 +1437,8 @@ def _on_secret_changed(self, event):
     def get_content(self, *, refresh: bool = False) -> dict[str, str]:
         """Get the secret's content.
 
-        The content of the secret is cached on the :class:`Secret` object, so
-        subsequent calls do not require querying the secret storage again,
-        unless ``refresh=True`` is used, or :meth:`set_content` is called.
+        The content of the secret is no longer cached on the :class:`Secret`,
```
tonyandrewmeyer: I think "no longer" will age badly.

```suggestion
        The content of the secret is not cached on the :class:`Secret`;
```

ops/model.py
tonyandrewmeyer, line 318: Maybe a small "# If both `id` and `label` are provided, the charm is setting the label, so we must run the hook command" type comment here to help future people out?

test/test_model.py
```diff
@@ -3832,8 +3850,10 @@ def test_id_and_label(self, model: ops.Model):
 
     def test_get_content_cached(self, model: ops.Model, fake_script: FakeScript):
         fake_script.write('secret-get', """exit 1""")
+        return
+        # FIXME
```
benhoyt: What are these FIXMEs for -- can we fix them now?

tonyandrewmeyer: What's the story here?

test/test_model.py
```diff
@@ -3842,7 +3862,9 @@ def test_get_content_cached(self, model: ops.Model, fake_script: FakeScript):
     def test_get_content_refresh(self, model: ops.Model, fake_script: FakeScript):
         fake_script.write('secret-get', """echo '{"foo": "refreshed"}'""")
 
-        secret = self.make_secret(model, id='y', content={'foo': 'bar'})
+        return
+        # FIXME
```
tonyandrewmeyer: And here?

test/test_model.py
dimaqq: oops, forgot!

ops/model.py
dimaqq: Good point, adding!

ops/model.py
dimaqq: My slight preference would be to remove it now, because:
- no charm seems to use it in supertox set
- we've told charmers not to instantiate directly
- pragmatically, it's hard for a charm to instantiate directly (they'd have to find a way to get the model backend†)
- the arguments are not documented

I'm happy to defer to @benhoyt on the decision.

† some charms do reach the model backend, but for other reasons:
- to run `goal-state` or `unit-get` hook commands directly
- to get storage event guts
See https://github.com/search?q=org%3Acanonical%20%22model._backend%22&type=code

ops/model.py
dimaqq: There (at least) three cases to consider:
- instantiate secret by id, e.g. when secret id is shared over a relation
- instantiate secret by label, e.g. this app owns the secret and non-leader unit reads the content to apply to the workload
- calling `get_conent(id=..., label=...)` to bind a consumed secret

The code we've looked at today:
https://github.com/canonical/data-platform-libs/blob/92cb654cfbaaad9bf17293a8e72587e9f0530e0b/lib/charms/data_platform_libs/v1/data_interfaces.py#L535-L550

Uses the API in two distinct ways, although both relying on exceptions raised early.

dimaqq: In general, applying a label is tricky:
- secret owner would set the label with `secret-set`
- secret grantee's leader would set the label with `secret-get`
- non-leader consumer can't quite apply labels (except see https://github.com/juju/juju/pull/16997)

In my opinion, the label must be set:
- when owner creates the secret
- when grantee (leader) first receives the secret

After that, the label should not be changed; rather charm should look up the secret by either id or label.

> as presumably people will want to apply a label most of the time ... Or is my assumption not true?

I don't find this assumption true.

tonyandrewmeyer: I also feel like we ought to be more cautious, even though it makes the API messier than we’d like, and offer an explicit choice whether to do the call at this time, separate from whether or not a label is used.

dimaqq:
Discussed in person:
- the `validate=` keyword addition feels clunky
- we're introducing a performance hit
- to get out of the performance hit, charmers have to update their code
- which introduces churn into the ecosystem

We need to decide if the benefits are worth it:
- solving multiple Secret instances getting out of sync
- being able to deprecate specific invocation styles to move charmers to clearer API

We're going to let this PR sit around for a bit, and talk this over during the week.

jameinel:
I'm not sure how multiple Secret instances would get out of sync. Is that if you set the content with one of them, and then read it from the other one?

I don't know whether the Juju agent actually caches the secret content. I wouldn't make guarantees there because of the sensitivity of secret content.  So it is reasonably likely .get_content() will be a round-trip to the Secret Backend each time. (now, I know that `set_content` doesn't synchronize until the hook exits, which means we do clearly have to keep the content in memory for some period of time)

I do think being clear with the charms that they might want to cache the contents is relevant, but that they should be careful about it.

tonyandrewmeyer:
> I'm not sure how multiple Secret instances would get out of sync. Is that if you set the content with one of them, and then read it from the other one?

Yes. Or set on both with different content (which is essentially the same).

dimaqq:
This PR may need a change for lazy secrets (from event and config)

dimaqq:
Decided not to this now.
Ref https://github.com/canonical/operator/issues/1536

PR #2150, opened 2025-10-28
docs: in integration tests, use consistent approach to logging and packing

dwilding:
This PR updates [How to write integration tests for a charm](https://documentation.ubuntu.com/ops/latest/howto/write-integration-tests-for-a-charm/) and the [httpbin-demo charm](https://github.com/canonical/operator/tree/main/examples/httpbin-demo) to be consistent with the Charmcraft 4 profiles and our new machine tutorial:

- **How to write integration tests for a charm**
    - Use `sleep` to wait for Juju to process logs, then output the logs to `stderr`.
    - Change `conftest.py` to check the `CHARM_PATH` environment variable instead of packing the charm.
    - Update surrounding docs to clarify that packing should be separate from running integration tests, and to explain how to specify which `.charm` file to deploy in the tests.
    - Make all tests depend on the `charm` fixture, for a better failure experience if there's no `.charm` file available.

    [Preview of updated doc](https://canonical-ubuntu-documentation-library--2150.com.readthedocs.build/ops/2150/howto/write-integration-tests-for-a-charm/)

- **httpbin-demo charm** - Use `sleep` to wait for Juju to process logs, then output the logs to `stderr`.

I haven't updated the integration tests in the Kubernetes charm tutorial. We plan to switch the tutorial to use the Charmcraft 4 profile at a later date.

### DIFF
--- a/docs/howto/write-integration-tests-for-a-charm.md
+++ b/docs/howto/write-integration-tests-for-a-charm.md
@@ -407,9 +407,10 @@ Your tests will detect a `.charm` file and deploy it. To specify which `.charm` 
 CHARM_PATH=/path/to/foo.charm tox -e integration
 ```
 
-Your tests will use the context of the current controller in Juju, and by default will create a new model per module, that will be destroyed when the test is finished. The cloud, controller and model name can be specified with the parameters `--cloud`, `--controller` and `--model` parameters.
+Your tests will use the current Juju controller. By default, a new model will be created for each test module. The model will be destroyed when the test is finished. This is determined by the scope of the `juju` fixture.
 
-If you specify the model name and do not delete the model on test tear down with the parameter `--keep-models`, you can reuse a model from a previous test run, as in the next example:
+Use the `--cloud`, `--controller`, and `--model` parameters to specify the cloud, controller, and model name. If you specify the model name and include the `--keep-models` parameter, you can reuse a model from a previous test run. For example:
+
 ```text
 # in the initial execution, the new model will be created
 tox -e integration -- --keep-models --model test-example-model


docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -239,14 +262,16 @@ def test_my_integration(juju: jubilant.Juju):
 
 > See more: [](jubilant.Juju.integrate)
 
+This test (and subsequent tests) don't need to depend on the `charm` fixture. However, it's helpful for each test to depend on `charm`, so that each test fails immediately if a `.charm` file isn't available.
```
james-garner-canonical: Maybe this suggests that `charm` should be an `autouse` fixture? Or maybe that could just be added to this tip.
```suggestion
This test (and subsequent tests) don't need to depend on the `charm` fixture. However, it's helpful for each test to depend on `charm`, so that each test fails immediately if a `.charm` file isn't available. If all your tests depend on the same charm being deployed, you could make `charm` an `autouse` fixture.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -376,7 +401,13 @@ tox -e integration
...
 ```
 
-These tests will use the context of the current controller in Juju, and by default will create a new model per module, that will be destroyed when the test is finished. The cloud, controller and model name can be specified with the parameters `--cloud`, `--controller` and `--model` parameters.
+Your tests will detect a `.charm` file and deploy it. To specify which `.charm` file to deploy:
+
+```text
+CHARM_PATH=/path/to/foo.charm tox -e integration
+```
+
+Your tests will use the context of the current controller in Juju, and by default will create a new model per module, that will be destroyed when the test is finished. The cloud, controller and model name can be specified with the parameters `--cloud`, `--controller` and `--model` parameters.
```
james-garner-canonical:
```suggestion
Your tests will use the context of the current controller in Juju. By default a new model  is created for each test module, that will be destroyed when the test is finished -- this is determined by the scope of the `juju` fixture. The cloud, controller and model name can be specified with the parameters `--cloud`, `--controller` and `--model`.
```

james-garner-canonical (approved):
Some small suggestions, but looks good to me overall -- and we'll need to update this next cycle when we work on `pytest-jubilant` anyway.

docs/howto/write-integration-tests-for-a-charm.md
tonyandrewmeyer: I'd rather we add the comment as suggested rather than actually make it autouse ourselves.

docs/howto/write-integration-tests-for-a-charm.md
tonyandrewmeyer: Building on James's:

```suggestion
Your tests will use the current Juju controller. By default a new model will be created  for each test module, which will be destroyed when the test is finished -- this is determined by the scope of the `juju` fixture. The cloud, controller and model name can be specified with the parameters `--cloud`, `--controller` and `--model` parameters.
```

dwilding: Good suggestions, thanks. I adjusted slightly for readability, as I feel the sentence about parameters fits better with the following paragraph. Here's what I've gone with:

> Your tests will use the current Juju controller. By default, a new model will be created for each test module. The model will be destroyed when the test is finished. This is determined by the scope of the `juju` fixture.
>
> Use the `--cloud`, `--controller`, and `--model` parameters to specify the cloud, controller, and model name. If you specify the model name and include the `--keep-models` parameter, you can reuse a model from a previous test run. For example:

james-garner-canonical: Just noticed that 'the model will be destroyed when the test is finished' is a bit inaccurate, since it's not till the module of tests has finished

dwilding: Well spotted! I've opened https://github.com/canonical/operator/pull/2154 to fix

PR #2151, opened 2025-10-28
docs: change HACKING.md instructions for maintaining Charmcraft profiles

dwilding:
This PR changes "Updating the Ops versions in the Charmcraft profiles" in HACKING.md to use [charmcraft-profile-tools](https://github.com/canonical/charmcraft-profile-tools) for initialising charms and generating lockfile templates.

_Why not keep the manual process too? Doesn't charmcraft-profile-tools make the steps less transparent?_ It does - but the manual process wasn't quite right any more. The lockfiles for the K8s and machine profiles aren't exactly the same, because the K8s profile has PyYAML as an integration dependency. So we can't say "repeat the full process for the machine profile" any more. Having tooling is more practical now.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -370,7 +370,7 @@ The Charmcraft `kubernetes` and `machine` profiles specify a minimum Ops version
 
     If you don't have [just](https://just.systems/man/en/) installed, use `uvx --from rust-just just` instead.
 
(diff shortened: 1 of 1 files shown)


dwilding:
I'm happy for this to be merged - just needs another tick

Issue #1671, opened 2025-04-08
Scenario relation databag out of sync


skourta:
Updating or reading updated data bags is no longer in sync
For example:
```
    ctx = testing.Context(TestCharmCharm)
    peer_relation = testing.PeerRelation(
        id=1,
        endpoint="peers",
        local_app_data={
            "foo": "bar",
        },
    )
    state_in = testing.State(
        relations=[peer_relation],
    )
    # Act:
    state_out = ctx.run(ctx.on.start(), state_in)
    # Assert:
    assert state_out.unit_status == testing.ActiveStatus()
    peer_relation.local_app_data["fooo"] = "bar"

    state_out = ctx.run(ctx.on.update_status(), state_out)
    assert next(iter(state_out.relations)).local_app_data["fooo"] == "bar"
```
The last assertion is false as of scenario `7.20`
This used to work in scenario < `7.20` but no longer works as of `7.20`. 
The change does not get propagated to the relation databag. The other way around also does not work, aka If your charm updates its databag through the code, the databag you see through the scenario relation is also outdated.

[Minimal reproduction repository](https://github.com/skourta/scenario-databag-bug-reproduction)

Updating ops to `7.20` on the reproduction repository will make the test fail.

tonyandrewmeyer:
>     peer_relation.local_app_data["fooo"] = "bar"

All the Scenario objects should be treated as immutable (we recently fixed some bugs around that, which were probably what was allowing the tests to pass). You need to do something like this (you could shrink it down):

```python
# Create a new relation with the updated data
updated_data = peer_relation.local_app_data.copy()
updated_data["fooo"] = "bar"
updated_peer_relation = dataclasses.replace(peer_relation, local_app_data=updated_data)

# Create a new state with the new relation
adjusted_state_out = dataclasses.replace(state_out, relations={updated_peer_relation})
```

>     assert next(iter(state_out.relations)).local_app_data["fooo"] == "bar"

Unrelated, but this would be simpler as:

```python
assert state_out.get_relation(peer_relation.id).local_app_data["fooo"] == "bar"
```

> The other way around also does not work, aka If your charm updates its databag through the code, the databag you see through the scenario relation is also outdated.

Do you have an example of this? My guess is that your test code is looking at the *input* relation, rather than the output one.

skourta:
Thank you, @tonyandrewmeyer .
Your guess is correct. I have updated all tests to match what you mentioned. 
It would be nice to have this well documented in the scenario docs.
[...]

PR #2153, opened 2025-10-29
docs: for state-transition tests, clarify about immutability and reusing state

### LINKED ISSUE #1676: Make sure that immutability in unit testing (Scenario) is documented
Per https://github.com/canonical/operator/issues/1671, we should make sure the docs are clear about:

- Objects should be treated as immutable
- How to create adjusted objects during a unit test (see [Tony's comment](https://github.com/canonical/operator/issues/1671#issuecomment-2787745918) for examples)
- Why you might want to run more than one event as part of the same unit test, along with mutability considerations (referring back to the above two points)

dwilding:
This PR updates the state-transition testing docs to clarify that objects in `State` are immutable. I'm also documenting how to modify a `State` for reuse: copy & adjust the relevant data, then create a `State`.

Previews of updated docs:

- [How to write unit tests for a charm](https://canonical-ubuntu-documentation-library--2153.com.readthedocs.build/ops/2153/howto/write-unit-tests-for-a-charm/#reusing-state) - New section called "Reuse state"
- [ops.testing (was: Scenario)](https://canonical-ubuntu-documentation-library--2153.com.readthedocs.build/ops/2153/reference/ops-testing/) - Mentioned immutability after the image. Included a drive-by to make class names links.
- [State-transition testing](https://canonical-ubuntu-documentation-library--2153.com.readthedocs.build/ops/2153/explanation/state-transition-testing/) - Mentioned immutability after the image.

Fixes #1676.

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -380,7 +380,7 @@ class ConfigChangedEvent(HookEvent):
     - Right after the unit starts up for the first time.
       This event notifies the charm of its initial configuration.
       Typically, this event will fire between an :class:`~ops.InstallEvent`
-      and a :class:~`ops.StartEvent` during the startup sequence
+      and a :class:`~ops.StartEvent` during the startup sequence
       (when a unit is first deployed), but in general it will fire whenever
       the unit is (re)started, for example after pod churn on Kubernetes, on unit
       rescheduling, on unit upgrade or refresh, and so on.


docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,25 @@ def test_charm_runs(my_charm):
 If you use pytest, you should put the `my_charm` fixture in a top level `conftest.py`, as it will likely be shared between all your unit tests.
 ```
 
+## Reusing state
+
+Try to make each test an isolated test of a particular event handler. After checking the `State` that `ctx.run` returns, you shouldn't need to reuse the `State` as input to another event handler.
+
+However, if you do need to reuse a `State`, don't try to modify the `State`. In general, the objects in a `State` are immutable. To reuse a `State`, create a `State` with the required data. For example:
```
dwilding: I'm not explaining why you'd want to reuse a State - ie why would you want to run more than one event? I don't know what's best to write here. We can add more explanation if needed.

docs/explanation/state-transition-testing.md
```diff
@@ -35,6 +35,8 @@ event, and based on whether it has leadership or not (according to its input
 state), it will decide to set `active` or `blocked` status (which will be
 reflected in the output state).
 
+When the testing framework runs the event, the input state is not modified. In general, the objects in a `State` are immutable.
```
james-garner-canonical:
```suggestion
When the testing framework runs the event, the input state is not modified.
Instead, a new state object is returned.
`State` objects are generally immutable, but be careful when working with `dict` attributes, as they do not enforce immutability.
```
Worth mentioning here that a new state object is returned instead?
Worth describing in more detail what generally immutable means?

I notice that the same text is used in two other docs, up to you if/where this extra content should go.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,25 @@ def test_charm_runs(my_charm):
...
+
+```python
+state_out = ctx.run(...)  # The State we want to reuse.
+relation = state_out.get_relation(...)  # A relation we want to modify.
+
+# Copy and modify the relation data.
+new_local_app_data = local_app_data.copy()
```
james-garner-canonical: I know it's just an example, but it seems like it would be nice to show where `local_all_data` comes from.

docs/howto/write-unit-tests-for-a-charm.md
james-garner-canonical:
```suggestion
However, if you do need to reuse a state, don't try to modify the original `State` object. Instead, create a new `State` object with the required data. For example:
```

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,25 @@ def test_charm_runs(my_charm):
...
+new_local_app_data["foo"] = "bar"
+
+# Create a State.
```
james-garner-canonical:
```suggestion
# Create a new State.
```

docs/howto/write-unit-tests-for-a-charm.md
dwilding: Whoops, thanks for catching! I've corrected the line:

```
new_local_app_data = relation.local_app_data.copy()
```

dimaqq (commented):
Bird-eye view LGTM

docs/explanation/state-transition-testing.md
dwilding: Good suggestions, thanks. Looking at this again, I think the information flows better if we shift the new part slightly further down. In context, the addition now looks like:

> Using the state-transition framework for charm unit tests, then, means verifying that:
>
> - the charm does not raise uncaught exceptions while handling the event
> - the output state (or the diff with the input state) is as expected.
>
> When the testing framework runs the event, the input state isn't modified. Instead, the output state is a new `State` object. `State` objects are generally immutable - but be careful when working with `dict` attributes, as they don't enforce immutability.

I've done the same thing in the other two docs too.

dwilding:
I'm happy for this to be merged - just needs another tick

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,25 @@ def test_charm_runs(my_charm):
 If you use pytest, you should put the `my_charm` fixture in a top level `conftest.py`, as it will likely be shared between all your unit tests.
 ```
 
+## Reusing state
+
+Try to make each test an isolated test of a particular event handler. After checking the `State` that `ctx.run` returns, you shouldn't need to reuse the `State` as input to another event handler.
```
tonyandrewmeyer: I'm not sure about this paragraph. It's not inherently wrong to chain, like:

```python
st1 = ctx.run(ctx.on.config_changed(), st0)
assert s1...
st2 = ctx.run(ctx.on.start(), st1)
assert s2...
```

We suggest that typically each test is a testing a single event, but it doesn't have to.

It's also fine to have a state and re-use it to evaluate what happens with different events. For example, a State that's an expected stable state (maybe a fixture) and asserting that you end up with an output state regardless of event order.

dimaqq:
I wonder about the overall tone of this do.
There's a lot of "you this" and "you that", and now we're adding "you shouldn't".

How about something like this:

Ideally the test setup and validation are precise:
```py
state_in = ...
state_out = ctx.run(..., state_in)
assert start_out.foo == ...
```

If you find the input state being reused within a test, a better alternative is a fixture that's shared across tests, for example:
```py
# instead of this
def test_foo():
    state_in = ...
    state_out1 = ctx.run(thing1, state_in)
    state_out2 = ctx.run(thing2, state_in)
    assert state_out1.foo = ...
    assert state_out2.bar = ...

# consider this
@pytest.fixture
def state_in():
    return ...

def test_one(state_in: ops.testing.State):
    state_out = ctx.run(thing1, state_in)
    assert state_out.foo = ...

def test_two(state_in: ops.testing.State):
    state_out = ctx.run(thing2, state_in)
    assert state_out.bar = ...
```

If you find a pipeline of operations (terminology?), a better alternative is (I don't know, smth better feels case-by-case to me):
```py
state_init = ...
state_one = ctx.run(..., state_init)
state_two = ctx.run(..., state_one)
state_three = ctx.run(..., state_two)
assert state_three.foo = ...
```

In fact I often start with the above pattern, but rarely end up with it, though at a cost of "duplicatey-ish" test setup: it feels repeated, but it's not. I think I usually extra parts of initial state into fixtures (e.g. a key relation, config), and then instantiate the state in each test.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,25 @@ def test_charm_runs(my_charm):
...
+state_out = ctx.run(...)  # The State we want to reuse.
+relation = state_out.get_relation(...)  # A relation we want to modify.
+
+# Copy and modify the relation data.
+new_local_app_data = relation.local_app_data.copy()
+new_local_app_data["foo"] = "bar"
+
+# Create a new State.
+new_relation = dataclasses.replace(relation, local_app_data=new_local_app_data)
+new_state = dataclasses.replace(state_out, relations={new_relation})
```
dimaqq, line 183: I wonder about `import dataclasses; dataclasses.replace(...)` vs. `from dataclasses import replace; replace(...)`.

Our style guide prefers the earlier, and it's fine when replacing one or two things.

But if we recommend this, then I guess there would be many replacements in a give test file, making it pretty verbose, unnecessarily explicit.

dwilding: I see what you mean. In this instance, I'm OK with keeping `dataclasses.replace`, as we're not showing any imports, so I think it's clearer this way.

dwilding:
Thanks for your feedback @tonyandrewmeyer & @dimaqq - I've rewritten [Reusing state](https://canonical-ubuntu-documentation-library--2153.com.readthedocs.build/ops/2153/howto/write-unit-tests-for-a-charm/#reusing-state) to be less scary and give better guidance on different approaches. Would you mind taking another look when you get a chance?

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,50 @@ def test_charm_runs(my_charm):
 If you use pytest, you should put the `my_charm` fixture in a top level `conftest.py`, as it will likely be shared between all your unit tests.
 ```
 
+## Reusing state
+
+Each test is typically an isolated test of a single event handler. If you need to use the same input state for several events, we recommend using a fixture that produces the input state. For example:
```
dimaqq:
```suggestion
Each test is typically an isolated test of a single state transition. If you need to use the same input state for several events, we recommend using a fixture that produces the input state. For example:
```

"event handler" is a thing in Ops, and a charm can and does have multiple attached to a single Juju event or Ops lifecycle event.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,50 @@ def test_charm_runs(my_charm):
...
+
+def test_status(state_normal):
+    ctx = testing.Context(Charm)
+    state_out = ctx.run(ctx.on.update_status(), state_normal)
+    # Check the output state.
+```
+
+Sometimes it's more convenient to test several event handlers in the same function. For example, to check what happens when your charm receives events in a particular order.
+
+After checking a `State` object that `ctx.run` returns, you can provide the same state as input to another event handler. If you need to modify the state between the event handlers, create a new `State` object instead of modifying the original `State` object. For example:
```
dimaqq: I'm not sure what to change "event handlers" to, but change we must.

dwilding: Thanks for pushing to make this clearer. How about this (emphasis added for easier review):

> Each test is typically an isolated test of **how your charm responds to a single event**. If you need to use the same input state for several events, we recommend using a fixture that produces the input state. For example:
>
> [snip]
>
> Sometimes it's more convenient to **simulate several events** in the same **test**. For example, to check what happens when your charm receives events in a particular order.
>
> After checking a `State` object that `ctx.run` returns, you can provide the same state as input to another **simulated event**. If you need to modify the state between the events, create a new `State` object instead of modifying the original `State` object. For example:

dimaqq: that works

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,50 @@ def test_charm_runs(my_charm):
 If you use pytest, you should put the `my_charm` fixture in a top level `conftest.py`, as it will likely be shared between all your unit tests.
 ```
 
+## Reusing state
+
+Each test is typically an isolated test of how your charm responds to a single event. If you need to use the same input state for several events, we recommend using a fixture that produces the input state. For example:
+
+```python
+@pytest.fixture
+def state_normal():
```
dimaqq: Minor: normal_state would read more naturally, though still makes someone wonder what normal means. I’d personally call it good state or quiescent state or perhaps something precise like leader_configured_state or workload_running_state…

That being said, I’m ok with merging in the current form too.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,50 @@ def test_charm_runs(my_charm):
...
+    container = testing.Container("some-container", can_connect=True)
+    return testing.State(
+        containers={container},
+        config={"server-port": 8080},
+        leader=True,
+    )
+
+
+def test_config_changed(state_normal):
```
dimaqq: This is an odd test, because we wouldn’t expect the state to change unless config is different. 

It can be readily applied to neither Delta nor holistic charms.

What would make this better is either new configuration value or a comment stating that we don’t expect any change.

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -162,6 +162,50 @@ def test_charm_runs(my_charm):
...
+    ctx = testing.Context(Charm)
+    state_out = ctx.run(ctx.on.config_changed(), state_normal)
+    # Check the output state.
+
+
+def test_status(state_normal):
```
dimaqq: Update status is a very special event. If we use it, we have to be clear that this is used as a kludge for Delta charms not being able to process the actual juju event previously at that point in time.

What bothers me is that we are suggesting our users to cover all of events that they subscribe to. Which would be a form of useless coverage in unit testing. What we should be doing instead is to steer them towards testing interesting state transitions.

docs/reference/ops-testing.rst
```diff
@@ -47,11 +47,11 @@ another state.
 Writing these tests should nudge you into thinking of a charm as a black-box
 'input to output' function. The inputs are:
 
-- Event: why am I, the charm, being executed
-- State: am I the leader? what is my relation data? what is my config?
-- Context: what relations can I have? what containers can I have?
+- Event (:class:`CharmEvents <ops.testing.CharmEvents>`): why am I, the charm, being executed
```
dimaqq: The ? Is missing at the end.

dimaqq (approved):

benhoyt:
@dimaqq Could you please get this over the goal line?

dwilding:
Reassigning as I'll have another look at this one this week

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -164,6 +164,50 @@ def test_charm_runs(my_charm):
...
+state_out = ctx.run(...)  # The State we want to reuse.
+relation = state_out.get_relation(...)  # A relation we want to modify.
+
+# Copy and modify the relation data.
+new_local_app_data = relation.local_app_data.copy()
+new_local_app_data["foo"] = "bar"
+
+# Create a new State.
+new_relation = dataclasses.replace(relation, local_app_data=new_local_app_data)
+new_state = dataclasses.replace(state_out, relations={new_relation})
```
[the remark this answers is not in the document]
dwilding: I'm OK as-is, as it's a snippet of code, and `dataclasses` should be familiar enough for people.

dwilding:
I agree with @dimaqq's latest concerns. To paraphrase, how common/realistic is it to use _exactly the same_ input state in multiple tests? We do have [an instance in the tinyproxy charm](https://github.com/canonical/operator/blob/main/examples/machine-tinyproxy/tests/unit/test_charm.py#L121-L145), but I feel this needs more discussion before going into the how-to docs.

In the interest of progress, I'll merge this PR without the part about a state fixture. That aspect didn't come up in the original issue https://github.com/canonical/operator/issues/1671.

docs/reference/ops-testing.rst
```diff
@@ -47,11 +47,11 @@ another state.
 Writing these tests should nudge you into thinking of a charm as a black-box
 'input to output' function. The inputs are:
 
-- Event: why am I, the charm, being executed
-- State: am I the leader? what is my relation data? what is my config?
-- Context: what relations can I have? what containers can I have?
+- Event (:class:`CharmEvents <ops.testing.CharmEvents>`): why am I, the charm, being executed?
```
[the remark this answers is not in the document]
dwilding: This is not about a specific parameter called `Event`, it's a heuristic for how a state-transition test should be structured. In that context, I'm OK with linking to `CharmEvents`

docs/howto/write-unit-tests-for-a-charm.md
```diff
@@ -164,6 +164,25 @@ def test_charm_runs(my_charm):
...
```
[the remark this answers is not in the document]
dwilding: Why not a set with one element? That's still a set of relations. Copilot is wrong here.

PR #2161, opened 2025-11-02
feat: log the total number of deferred events

### LINKED ISSUE #2101: Add ability to 1) see deferred queue length (in juju debug-log?) + 2) de-duplicate deferred queue
Hi,

The recent PS6 incidents showed us a need to see the amount of deferred events in queue to estimate the charm catch-up.
In this case, the entire events queue took about 10 hours to clean (the outage took a ~day(s)). 
There were more then 200 events in the queue (3 different events have been deferred by charm):
```
dbaas-secret-system@ps6:$ sqlite3 .unit-state.db-backup-20251009
SQLite version 3.37.2 2022-01-06 13:25:41
...
sqlite> select count(*) from notice;
240

sqlite> select count(*) from snapshot;
241
```

Also the charm event process was slow as well (charm refactoring is expected).

Anyway, it is a feature request to:
 * add ability for users to see the size of event queue (for deferred events). It will allow to provide customer estimates and plan hardware usage better, as differed event execution takes CPU). E.g in juju debug-log:
```
Re-emitting deferred event < ... /config_changed[169]> of total 46
```
 * de-duplicate deferred events (e.g. we know the system has 100 of identical events). Consider to execute it once/twice (number 1 and number 100 in queue, and skip all in between).
 * add ability to clean unnecessary events somehow (e.g. uning jhack). At the moment users have to go into sqlite on unit to clean the queue => the dangerous and risky way. Otherwise wait for hours/days wasting CPU/time.

AFAIK, Ops stores the last 100 deferred events, but many events will be deferred for long outage.
The popular deferred events for PostgreSQL VM charm are: `secret-changed`/`secret-remove`/`*-peers-relation-changed`.
Those events have to be processed, but once only, no need to repeat them 100 times with the same content.

Tnx!

tonyandrewmeyer:
When logging that a deferred notice is being processed, also log the total number in the queue (as of the start of the hook). This allows anyone examining the debug-level Juju log to see that there is a large backlog, presumably indicating problems (see #2101 for an example).

There is a small cost here, in that we are collapsing a generator into a tuple, although we expect it to be small, and we're already getting it as a list from the storage.

Fixes #2101

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -973,7 +973,7 @@ class Framework(Object):
         last_event_path = None
         deferred = True
         notices = tuple(self._storage.notices(single_event_path))
-        for event_path, observer_path, method_name in notices:
+        for i, (event_path, observer_path, method_name) in enumerate(notices):
             event_handle = Handle.from_path(event_path)
 
             if last_event_path != event_path:
@@ -995,7 +995,10 @@ class Framework(Object):
             if observer:
                 if single_event_path is None:
                     logger.debug(
-                        'Re-emitting deferred event %s (%d in queue).', event, len(notices)
+                        'Re-emitting deferred event %s (%d of %d in queue).',
+                        event,
+                        i + 1,
+                        len(notices),
                     )
                 elif isinstance(event, LifecycleEvent):
                     # Ignore Lifecycle events: they are "private" and not interesting.


ops/framework.py
```diff
@@ -993,7 +994,7 @@ def _reemit(self, single_event_path: str | None = None):
 
             if observer:
                 if single_event_path is None:
-                    logger.debug('Re-emitting deferred event %s.', event)
+                    logger.debug('Re-emitting deferred event %s of total %d.', event, len(notices))
```
benhoyt: The language "of total %d" seems a bit weird, because it's not "%d of total %d". What about this or similar:

```suggestion
                    logger.debug('Re-emitting deferred event %s (%d in queue).', event, len(notices))
```

ops/framework.py
```diff
@@ -972,7 +972,8 @@ def _event_context(self, event_name: str):
     def _reemit(self, single_event_path: str | None = None):
         last_event_path = None
         deferred = True
-        for event_path, observer_path, method_name in self._storage.notices(single_event_path):
+        notices = tuple(self._storage.notices(single_event_path))
```
james-garner-canonical, line 975: It seems like this is the only place that `notices` is actually called internally, is that right?

Optimization: How about modifying `notices` to return `c.fetchall()`, if we've decided that we don't need to worry about the time and memory overhead of fetching the rows greedily?

Concern: On the other hand, if we do need to worry about the overhead, I guess we could get the count with a `SELECT COUNT` first?

---

On the backwards compatibility of modifying `notices` to return a list instead of a generator, since I guess it's technically a public method of a public class ... the instance used in the framework is in a private attribute (`_storage`) -- and I had a quick look and it doesn't look like it's being called in charms aside from these two instances in my charms snapshot, which are accessing via the private attribute and so asking for trouble, and returning a list wouldn't break their use case anyway.
```
hardware-observer/tests/unit/test_charm.py
46:        for event_path, _, _ in self.harness.charm.framework._storage.notices(None):

bootstack-charmers-storage-connector/src/charm.py
440:        for event_path, _, _ in self.framework._storage.notices(None):  # pylint: disable=W0212
```

ops/framework.py
```diff
@@ -972,7 +972,8 @@ def _event_context(self, event_name: str):
...
+        for event_path, observer_path, method_name in notices:
```
james-garner-canonical: Minor enhancement: Could log the position in the queue with something like this:
```suggestion
        for i, (event_path, observer_path, method_name) in enumerate(notices):
```
Doing something like:
```py
logger.debug('Re-emitting deferred event %s (%d of %d in queue).', event, i+1, len(notices))
```

james-garner-canonical (approved):
Approving so as not to block this, but I do think we should either (a) make the greedy fetching explicit earlier if we've decided we don't need to worry about memory overhead here, or (b) worry about memory overhead if we do need to. I'm happy for (a) to be captured in an issue for later and not addressed directly in this PR.

james-garner-canonical:
> we're already getting it as a list from the storage.

PS: This doesn't seem to be quite true in general, since we currently repeatedly yield from calling `c.fetchmany`, rather than calling `c.fetchall`. I assume if there are ever a memory prohibitive amount of deferred events, things have gone badly wrong. The current behaviour would have the charm spend a long time processing them one by one. I guess that's better than eating all the memory immediately.

ops/framework.py
benhoyt: I don't think we should overthink this. Allocating a few bytes (or even a few hundred if there are heaps of defers) of memory to convert the generator to a tuple is going to be super cheap in this context. We could make the greedy fetching explicit earlier, that is, have `notices()` return a list ... but I'm not sure it's worth the churn to change that (admittedly internal) method signature on both interfaces.

james-garner-canonical: > Allocating a few bytes (or even a few hundred if there are heaps of defers) of memory to convert the generator to a tuple is going to be super cheap in this context.

:+1: 

> We could make the greedy fetching explicit earlier, that is, have notices() return a list ... but I'm not sure it's worth the churn to change that (admittedly internal) method signature on both interfaces.

It feels like a bit of an antipattern to have this lazy loading generator logic but immediately greedily consume the generator. Dropping the lazy loading logic seems like a nice little readability improvement for future maintainers.
```diff
-        while True:
-            rows = c.fetchmany()
-            if not rows:
-                break
-            for row in rows:
-                yield cast('_Notice', tuple(row))
+        return c.fetchall()
```
I'm happy to take care of it in a follow-up PR if we think it's worth it.

benhoyt: Tony would rather merge the PR in the current state (Ben & Dima agree), so we're going to merge this and can consider the other change separately if desired.

ops/framework.py
benhoyt: @tonyandrewmeyer will update this

ops/framework.py
james-garner-canonical: Sounds good, should I make a PR for discussion later,  or leave it?

benhoyt: I'd leave it; it just doesn't seem worth it to me.

PR #2167, opened 2025-11-13
ci: update spelling

tonyandrewmeyer:
The docs spellcheck seems to have updated underneath us, so make adjustments so that the checks pass again.

For the most part, I've added words to the custom word list, but I've made a few adjustments to the docs in cases where it seemed better than adding the word.

The PR also corrects the description of Scenario's behaviour when creating a temporary charm root: only the metadata files are written - the `src` and `lib` folders are not also copied there.

(See also https://github.com/canonical/jubilant/pull/225)

### DIFF
--- a/testing/src/scenario/context.py
+++ b/testing/src/scenario/context.py
@@ -600,7 +600,7 @@ class Context(Generic[CharmType]):
     charm_root: str | pathlib.Path | None
     """The charm root directory to use when executing the charm.
 
-    Before running the event, the charm's metadata, config, and action YAML is
+    Before running the event, the charm's metadata, config, and action YAML files are
     copied to the charm root. If ``charm_root`` is None, then a temporary directory is
     created and used. To set up the charm root directory with other files, pass in the location of
     the directory to use. If a ``charm_root`` is provided, the tests are also able to inspect the


docs/explanation/state-transition-testing.md
```diff
@@ -192,9 +192,9 @@ emit multiple events in a single charm execution.
 
 ## The virtual charm root
 
-Before executing the charm, the framework copies the charm's `/src`, any libs,
-and the metadata, config, and actions YAML to a temporary directory. The charm
-will see that temporary directory as its 'root'. This allows us to keep things
+Before executing the charm, the framework copies the charm's `/src`, any Charmhub
+charm libraries, and the metadata, config, and actions YAML to a temporary directory.
```
james-garner-canonical: Does it actually just copy the whole `/lib` directory? I know some charms have included non-Charmhub library content in there at times, so that would reflect `charmcraft`'s behaviour. If so, this might be more accurate.
```suggestion
Before executing the charm, the framework copies the charm's `/src`, `/lib` (where Charmhub
libraries are defined), and the metadata, config, and actions YAML to a temporary directory.
```
Or more concisely:
```suggestion
Before executing the charm, the framework copies the charm's `/src`, `/lib`,
and the metadata, config, and actions YAML to a temporary directory.
```
However I'm ok with what you have currently if you prefer that.

docs/explanation/tracing.md
```diff
@@ -9,7 +9,7 @@ Instrument generously, so that more context is captured than is immediately need
 
 **Field engineers** can inspect the trace data from live deployments, ideally avoiding the need to reproduce failures in staging.
 
-**SREs** may want to set latency and error rate alerts on specific elements of trace data preemptively. They can also get more insights to help root-cause analysis across services.
+**Site Reliability Engineers** may want to set latency and error rate alerts on specific elements of trace data preemptively. They can also get more insights to help root-cause analysis across services.
```
tonyandrewmeyer, line 12: This is the one I'm least sure about. I'm really not sure how widely SRE is known as a term. I'm open to this going into the word list instead, or going the other way and making this more generic, like "Operations teams".

james-garner-canonical: 'SREs', 'Site Reliability Engineers', and 'Operations teams' all seem reasonable to me.

docs/explanation/state-transition-testing.md
tonyandrewmeyer: Ah, I was just looking at the supposedly mis-spelt words rather than what they were actually saying, nice catch.

Scenario doesn't copy `src` or `lib` (there's #1673 to figure out some sort of copying). Maybe it did once, which is why the docs say it should.

I guess fixing that in this PR is reasonable.

benhoyt:
Ah, thanks for this. Yeah, I noticed this [over in Jubilant-land too](https://github.com/canonical/jubilant/pull/225). It's another thing that makes me wonder about the worthwhileness of spelling checks. But we can get David's opinion on all that when he's back.

testing/src/scenario/context.py
```diff
@@ -600,8 +600,8 @@ def test_emitted():
     charm_root: str | pathlib.Path | None
     """The charm root directory to use when executing the charm.
 
-    Before running the event, the charm's ``/src``, any libs, and the metadata, config, and action
-    YAML is copied to the charm root. If ``charm_root`` is None, then a temporary directory is
+    Before running the event, the charm's metadata, config, and action YAML is
```
dimaqq: my 2c: either "YAML file", or list the actual file names explicitly, like `metadat.yaml`.

docs/.custom_wordlist.txt
```diff
@@ -1,17 +1,35 @@
...
+snappass
 storages
+systemd
+subprocess
 tinyproxy
 traceback
+untyped
+unpatched
 uv
+VM's
```
dimaqq: I wonder if adding "VM" alone would suffice.

docs/explanation/state-transition-testing.md
```diff
@@ -192,13 +192,12 @@ emit multiple events in a single charm execution.
...
 
-Before executing the charm, the framework copies the charm's `/src`, any libs,
-and the metadata, config, and actions YAML to a temporary directory. The charm
-will see that temporary directory as its 'root'. This allows us to keep things
-simple when dealing with metadata that can be either inferred from the charm
-type being passed to `Context` or be passed to it as an argument, thereby
-overriding the inferred one. This also allows you to test charms defined on the
-fly, as in:
+Before executing the charm, the framework writes the metadata, config, and actions
+YAML to a temporary directory. The charm will see that temporary directory as its
```
dimaqq: my 2c: either "YAML files" or list the file names with the extension.

tonyandrewmeyer:
```suggestion
YAML files to a temporary directory. The charm will see that temporary directory as its
```

testing/src/scenario/context.py
tonyandrewmeyer:
```suggestion
    Before running the event, the charm's metadata, config, and action YAML files are
```

dwilding:
Thanks for doing this @tonyandrewmeyer. The extra spelling exceptions were needed because the spelling check previously wasn't checking all text. I was tracking the fix in https://github.com/canonical/documentation-style-guide/pull/138 (sorry, I think I forgot to mention that in my notes before I went away).

I don't expect major changes to the spelling setup from now on, now that (as far as I could test previously) it's checking everything that it should be checking.

PR #2168, opened 2025-11-13
docs: Add AI contribution note and style guideline for type annotation of return values

tonyandrewmeyer:
Two small updates to our contribution and style guidelines:

* A statement on use of AI tools, summarising our discussion in Sweden. Basically: leave the Copilot reviews to us, but you're welcome to submit PRs that were (co-)written with AI, as long as you're a human taking responsibility for it.
* An update to our style guide recording our decision on when a type annotation for the return value of a function or method should be provided.


STYLE.md
```diff
@@ -128,6 +128,27 @@ if status is not pebble.ServiceStatus.ACTIVE:
     print('Stopped')
 ```
 
+### Provide return type annotations
+
+Function and method signatures should include a type annotation for the returned value, other than for `__init__` or in test code.
```
james-garner-canonical, line 133: This is accurate, but feels oddly specific without any broader context of our approach to type annotations.

Maybe all it needs is a line saying something like "All code, including tests, should be fully type annotated."

Our only other specific type annotation policy might be something like "Use `typing.cast` and `type: ignore` sparingly where satisfying the type checker would otherwise make the code harder to maintain."

I'm happy for it to land as-is though.

tonyandrewmeyer: I think the challenge here is that currently STYLE.md is mostly a record of decisions (plus a general "do what the tools tell you to do"), not an actual full style guide.

For this specific suggestion, I think it's covered by the "do what the tools tell you" bit, because they'll complain if you don't add appropriate annotations. For the typing.cast/type:ignore suggestion, I'd be ok adding that, but I think it should get the team approval before being added to the doc.

PR #2169, opened 2025-11-13
chore: add instructions for AI agents developing on the Ops codebase

tonyandrewmeyer:
*These are just an initial attempt*. We'll certainly want to continually iterate and tweak these to get the best results, both as a result of the value we find (or don't) and to handle adjustments in the tools themselves.

This is a mixture of instructions generated by the tools themselves and some manual tweaking based on my experience.

As requested, I've merged, where feasible, instructions for Copilot, Claude, and Cursor into a generic AGENTS.md.

For Claude, there's also:
* A post-edit hook that will ensure that `tox -e format` is run after editing Python files.
* A few settings, particularly ones that allow access to the main documentation sites and some reasonably safe commands.
* A couple of subagents to experiment with (subagents are relatively new, it's not clear if these are the right ones to have or whether they are generally useful).

For similar settings for Copilot, we'll need to adjust the settings in the repository.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,4 +158,5 @@ The project does not use conventional commit "scopes".
 5. Run `tox -e lint` to check linting and types
 6. Run `tox -e unit` to verify unit tests pass - avoid drops in coverage (`tox -e coverage`)
 7. Run `make html` in the `docs` folder to ensure that the documentation can be generated
-8. If a virtual environment or sandbox is available, run `tox -e pebble` and `tox -e integration`
+8. Search the explanation, how-to, and tutorial documentation in the `docs` folder for topics related to the changes, then suggest places that might need expanding/altering
+9. If a virtual environment or sandbox is available, run `tox -e pebble` and `tox -e integration`


.claude/hooks/after-edit.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
james-garner-canonical, line 1: How about just run `tox -e format` unconditionally?

.claude/README.md
```diff
@@ -0,0 +1,43 @@
+# Claude Code Configuration
```
james-garner-canonical, line 1: This file is for human consumption, right? I feel like it could be a lot shorter, or dropped entirely. Some things, like documenting what the `after-edit.sh` hook does or when it runs feel like they'll just get out of sync.

.claude/CLAUDE.md
```diff
@@ -0,0 +1,62 @@
+# Claude-Specific Instructions for Ops Library
```
james-garner-canonical: There doesn't seem to be any Claude specific information here. Could it just be a symlink or textual pointer to `AGENTS.md`?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,395 @@
+# Copilot Instructions for Ops
```
james-garner-canonical: Likewise, none of this really seems Copilot specific. It seems like any information not currently present in `AGENTS.md` could be folded into it, and this file could be replaced with a symlink or textual redirect for LLMs.

james-garner-canonical (commented):
Personally, I dislike the duplication of content across `CLAUDE.md`, `copilot-instructions.md` and `AGENTS.md`. Maybe I just haven't gotten past the idea of human-looking prose being auto-generated, throwaway script output, but it seems like it would be nice to consolidate this into a single source of truth for LLMs, presumably in `AGENTS.md`. 

If we want to keep `AGENTS.md` short, and separate out (e.g.) testing instructions, I don't see why we couldn't put that in something like `AGENTS-TESTING.md` and mention it. But context windows are pretty big now, so a longer `AGENTS.md` is probably fine.

I'm not opposed to having some separate instructions for Claude and Copilot, but I think they should be specific to those agents.

benhoyt (commented):
I also dislike the duplication and number of files here. Won't all the tools look at AGENTS.md, and can't we just have that one?

.claude/hooks/after-edit.sh
tonyandrewmeyer: Agents tend to edit files pretty frequently (more than a human would) and that includes the docs and configuration files and so on. I feel it's cleaner to only do it if it is likely that formatting will change anything.

james-garner-canonical: Cleaner in what sense?

tonyandrewmeyer: In the sense that we're not wasting time running something that we know won't do anything.

tonyandrewmeyer:
> I've started with dedicated instructions for Copilot and Claude, as well as a generic AGENTS.md (which Cursor will automatically use). We may find that we want to consolidate these in the future, but I think tweaking individually at first may be interesting.

However, everyone else disagrees, so I'll reverse this and we can go the other way. I do strongly feel that the instructions mix differently with each agent's built-in instructions (that all then potentially go to the same model) and so separate instructions make sense even if you're not referring to capabilities of the specific agent, and so on. But we can validate that the other way around as well.

Copilot and Claude (possibly also others) have a layered approach to instructions, where you have local instructions that augment or override the project ones (and you don't commit the local ones, but do commit the project ones). I'm not totally sure how that works if the project ones are in the generic `AGENTS.md` rather than in the proper project file, but I'll assume that it works for now. The UI for updating the project level instructions won't be what it could be - I think it'll have to be done manually.

.claude/hooks/after-edit.sh
james-garner-canonical: `tox -e format` runs for me in under 500ms after the first time (which creates the virtual environment). IDK, if this is something that you find makes your developer experience with Claude smoother, then it seems fine, but if it's just theoretical I'd be inclined to drop it.

tonyandrewmeyer: Everything the agent does involves not just that thing, but a bunch of tokens being sent to and from the model. We don't need to burn the planet more than necessary. It's also in the UI, purely as noise, but that's less of a concern than wasting tokens.

james-garner-canonical: Not sending the `tox -e format` output to the LLM unnecessarily sounds like a worthy goal to me, I hadn't thought of that. Presumably it's only a small part of the data that's exchanged on every interaction, but that makes sense.

.claude/hooks/after-edit.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ "$python_files_edited" = true ]; then
+    echo "Python files modified, running formatter..."
+    tox -e format
+
+    if [ $? -eq 0 ]; then
+        echo "✓ Code formatted successfully"
+    else
+        echo "⚠ Formatting failed - please check output above"
+        exit 1
+    fi
```
james-garner-canonical, line 32:
```suggestion
    exit $?
```
Worth minimising token use like this?

dimaqq: side-note: that `exit $?` is not needed.

james-garner-canonical: True, as long as `tox -e format` is the last command run in the script.

.claude/settings.json
```diff
@@ -0,0 +1,38 @@
...
+  "permissions": {
+    "allow": [
+      "WebFetch(domain:documentation.ubuntu.com)",
+      "WebFetch(domain:docs.zizmor.sh)",
+      "WebFetch(domain:charmhub.io)",
+      "WebFetch(domain:github.com)",
+      "WebFetch(domain:diataxis.fr)",
+      "Bash(ls:*)",
+      "Bash(grep:*)",
+      "Bash(tox:*)"
```
tonyandrewmeyer, line 35: There's a little bit of risk with these three, but in my experience it's generally fine, and it's certainly fine if you're using it in some sort of sandboxed environment. I find that I auto-approve these three anyway, so it's a reasonable starting point. However, if people would rather be more cautious we could remove any or all of them.

.claude/settings.json
```diff
@@ -0,0 +1,38 @@
...
+    "typeChecker": "pyright",
+    "docstringStyle": "google",
+    "formatter": "ruff"
+  },
+  "permissions": {
+    "allow": [
+      "WebFetch(domain:documentation.ubuntu.com)",
+      "WebFetch(domain:docs.zizmor.sh)",
+      "WebFetch(domain:charmhub.io)",
+      "WebFetch(domain:github.com)",
```
tonyandrewmeyer, line 31: There's a bit of risk here, in that both of these allow anyone to add content. That means that there could be hidden instructions on these sites, and GitHub at least (maybe also Charmhub) can be used for exfiltration. I feel that for our use, this should be safe enough, and that you shouldn't let the agent have access to any data you wouldn't want to leave your system anyway.

However, if we want safer defaults, we could drop these and individually approve them as needed.

.claude/README.md
tonyandrewmeyer: I've trimmed it a bit, but I feel like it is useful to have a bit of an explainer for these extra Claude-specific things so you don't have to immediately go off to the Claude docs.

.claude/hooks/after-edit.sh
tonyandrewmeyer: I'm not 100% sure, but I think the agent reads the output, so saying "please check output above" ends up circling back into an instruction, whereas just the plain output may not be as clear.

AGENTS.md
```diff
@@ -0,0 +1,159 @@
+# Agent Instructions for Ops
+
+## Project Overview
+Ops is the official Python framework for developing Kubernetes and machine charms within the Juju ecosystem. It provides an event-driven abstraction layer for charm developers.
+
+This workspace is a monorepo containing:
+- **`ops/`** - Core framework providing the event system, charm base classes, and model abstractions. Note that the tests are in a top-level folder `tests`
+- **`testing/`** - The `ops-scenario` state transition testing framework (accessed as `ops.testing`). This uses src-layout.
+- **`tracing/`** - The `ops-tracing` observability integration. The tests are in a subfolder called `test`.
+- **`docs/`** - Common documentation, predominately Markdown.
```
tonyandrewmeyer:
```suggestion
- **`docs/`** - Common documentation, predominantly Markdown.
```

AGENTS.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Before Submitting
+1. Add tests for any new functionality
+2. Update docstrings for any API changes
+3. Ensure backward compatibility
+4. Run `tox -e format` to format code
+5. Run `tox -e lint` to check linting and types
+6. Run `tox -e unit` to verify unit tests pass - avoid drops in coverage (`tox -e coverage`)
+7. Run `make html` in the `docs` folder to ensure that the documentation can be generated
+8. If a virtual environment or sandbox is available, run `tox -e pebble` and `tox -e integration`
```
dwilding: How about adding something like:

```suggestion
7. Run `make html` in the `docs` folder to ensure that the documentation can be generated
8. Search the explanation, how-to, and tutorial documentation in the `docs` folder for topics related to the changes, then suggest places that might need expanding/altering
9. If a virtual environment or sandbox is available, run `tox -e pebble` and `tox -e integration`
```

AGENTS.md
```diff
@@ -0,0 +1,159 @@
...
+    Example:
+        >>> my_function("hello", 2)
+        ['hello', 'hello']
+    """
+```
+
+**Documentation writing tips:**
+- Use **active voice**: "Create a check" not "A check is created"
+- Be **objective**: Avoid "simply", "easily", "just"
+- Be **concise**: No long introductions, get to the point
```
dwilding, line 110:
```suggestion
- Be **concise**: No long introductions, get to the point
- Use short sentences and simple phrasing
- Be consistent with choice of words
```

AGENTS.md
```diff
@@ -0,0 +1,159 @@
+# Agent Instructions for Ops
+
+## Project Overview
+Ops is the official Python framework for developing Kubernetes and machine charms within the Juju ecosystem. It provides an event-driven abstraction layer for charm developers.
+
+This workspace is a monorepo containing:
+- **`ops/`** - Core framework providing the event system, charm base classes, and model abstractions. Note that the tests are in a top-level folder `tests`
```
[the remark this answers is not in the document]
dimaqq: Indeed we've got `test` at the top level, but `tests` under `testing`.

tonyandrewmeyer:
> Though I thought we were going to go with the single `AGENTS.md` and not have separate stuff for each type (eg: `.claude/*`)?

I removed the separate CLUADE.md, and copilot-instructions.md and merged those into a single AGENTS.md, since there was so much pushback about that. There are some Claude-specific extra files still, but those cannot be put into AGENTS.md (and for Copilot must be done outside of files in the repository).

> Long term I hope the agents improve so we can just have this information in human-first files like STYLE.md, and point the agents there (if it doesn't pick it up automatically). 

AGENTS.md already refers to STYLE.md, CONTRIBUTING.md, HACKING.md, and so on. The agents file is loaded (or with Claude, potentially loaded) for every request, which uses up context space. Loading all of the files would be too wasteful (and even as context windows get larger, trying to put everything into every request is not just inefficient and expensive but also seems to hurt performance), so it's helpful to have a file that points the agent to read the right files at the right time, and extract out the most critical parts to avoid needing to do that all the time.

PR #2172, opened 2025-11-14
refactor: don't get the storage details from --help

### LINKED ISSUE #2109: Don't extract the storage key from `storage-get --help`
model.py has this code (it will have changed slightly by the time someone gets to this ticket, because of the hookcmds PRs):

```python
    def _storage_event_details(self) -> tuple[int, str]:
        output = self._run('storage-get', '--help', return_output=True)
        output = typing.cast('str', output)
        # Match the entire string at once instead of going line by line
        match = self._STORAGE_KEY_RE.match(output)
        if match is None:
            raise RuntimeError(f'unable to find storage key in {output!r}')
        key = match.groupdict()['storage_key']

        index = int(key.split('/')[1])
        location = self.storage_get(key, 'location')
        return index, location
```

This private method on `_ModelBackend` is used in `_main.py` to get the storage object to attach to the storage event object (if it's a storage hook).

It runs `storage-get --help` to find the storage name and index. This works because in the help output it shows the default value for `-s`, and that's the storage name/index *if run inside a storage hook* (otherwise it's the empty string). The code extracts that with a regular expression.

This is fragile (since it's relying on the help output) and also completely unnecessary, twice over:
* The storage key is provided in the `JUJU_STORAGE_ID` environment variable in storage events. We make the name and index available in `JujuContext`.
* The storage name (the `-s` argument) isn't required in a storage event, so we don't need to provide it when getting the location. This is somewhat obviously the case, because we're pulling the default value out of the help text, in order to provide the default value...

We could do:

```python
    def _storage_event_details(self) -> tuple[int, str]:
        assert self._juju_context.storage_name is not None
        key = self._juju_context.storage_name
        assert self._juju_context.storage_index is not None
        index = self._juju_context.storage_index
        location = self.storage_get(key, 'location')
        return index, location
```

Or perhaps:

```python
    def _storage_event_details(self) -> tuple[int, str]:
        assert self._juju_context.storage_index is not None
        index = self._juju_context.storage_index
        try:
            with self._tracing_span('storage-get', 'location'):
                stor = hookcmds.storage_get()
        except hookcmds.Error as e:
            self._check_for_security_event('storage-get', e.returncode, e.stderr)
        return index, stor.location
```

tonyandrewmeyer:
Remove the private `_ModelBackend._storage_event_details` method (and the `_ModelBackend.STORAGE_KEY_RE` only used in that method), which was only used in `_main`, and retrieved the storage index and location by parsing `storage-get --help`.

In `_main`, get the storage index from the `JujuContext` objects (so from the environment variable provided by Juju), and call `_ModelBackend.storage_get` directly to get the storage location.

Fixes #2109


james-garner-canonical (approved):
Looks good to me, nice cleanup.

Do you know how well covered this code path is in our tests?

tonyandrewmeyer:
> Do you know how well covered this code path is in our tests?

This path is exercised for any storage events, so reasonably well. Or in a more data-driven answer:

image

I don't believe the missed line would ever be reached with current Juju, but I'd rather leave it there for now.

james-garner-canonical:
Thanks for checking that!

PR #2173, opened 2025-11-14
refactor: drop 3.8 and 3.9 compatibility code

### LINKED ISSUE #2056: Clean up Python 3.8 hacks and comments
ops/pebble.py:

```
2486-            except (socket.timeout, builtins.TimeoutError):
2487-                # NOTE: in Python 3.10 socket.timeout is an alias of TimeoutError,
2488:                # but we still need to support Python 3.8, so catch both.
2489-                # Catch timeout from wait endpoint and loop to check deadline.
2490-                pass
```

ops/model.py:

```
1926-            except TypeError:
1927:                # Most likely Python 3.8. It's not as good, but use __annotations__.
1928-                fields = {k: k for k in obj.__class__.__annotations__}
1929-            values = {field: getattr(obj, field) for field in fields}
```

testing/src/scenario/state.py:

```
2068:        # TODO: when we drop Python 3.8, we can change the whole if-else below to
2069-        # instance.prefix = string.removesuffix(suffix)
2070-        if suffix:
2071-            instance.prefix, _ = string.rsplit(suffix, maxsplit=1)
2072-        else:
2073-            instance.prefix = string
```

test/test_pebble.py:

references Python 3.8.10 a lot... as if it's being tested against that version maybe?

test/test_charm.py:

```
1894:    # The latest version of Pydantic available for Python 3.8 does not support
1895-    # the `alias` metadata, so we need to skip the test for that case.
1896-    if pydantic is not None and action_class is _PydanticDataclassesAlias:
1897-        pytest.skip('Pydantic does not support dataclasses alias metadata in this version.')
```

```
1671-# Note that we would really like to have kw_only=True here as well, but that's
1672:# not available in Python 3.8.
1673-@dataclasses.dataclass(frozen=True)
1674-class MyDataclassAction:
1675-    my_str: str
1676-    my_bool: bool = False
1677-    my_int: int = 42
```

test/test_model_relation_data_class.py:

```
326:    # The latest version of Pydantic available for Python 3.8 does not support
327-    # the `alias` metadata, so we need to skip the test for that case.
328-    if pydantic is not None and relation_data_class is _PydanticDataclassesAlias:
329-        pytest.skip('Pydantic does not support dataclasses alias metadata in this version.')
```

tracing/ops_tracing/_buffer.py:

```
288-            # The RETURNING clause would be ideal here, but it can't be used.
289:            # Sqlite shipped with Python 3.8 is too old.
290-            row = conn.execute(
291-                """
292-                SELECT (length(data)+4095)/4096*4096
293-                FROM tracing
294-                WHERE id = ?
```

tonyandrewmeyer:
Remove code that provides compatibility with Python 3.8 or Python 3.9:

* In `Relation.save`, `get_type_hints` will always work without needing to fall back to `__annotations__`
* In our test dataclasses we can use `kw_only=True`
* We don't need to skip tests that don't work with the 3.8-compatible Pydantic
* Remove a comment that refers to unittest in 3.10, since we use pytest now
* Use `type[x]` rather than `Type[x]` in ops-tracing
* Drop the custom implementation of `kw_only=True` and `KW_ONLY` from the Scenario state objects in favour of the native dataclasses functionality
* Add one use of `removesuffix`

There's a 3.8 comment in `harness.py` that I've left as-is. I think we might as well maintain 3.8 compatibility for Harness, or at least not bother updating it.

<details>
<summary>Benchmarking results</summary>
We expect an improvement in the testing performance as a result of dropping the custom keyword-only code in state.

0001 is with main, NOW is with the branch. All of the median times go down.

```
-------------------------------------------------------------------------------------------------------- benchmark: 20 tests ---------------------------------------------------------------------------------------------------------
Name (time in ms)                                                 Min                 Max                Mean             StdDev              Median                IQR            Outliers          OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_context_autoload_meta (0001_5e752be)                      2.4994 (49.44)      4.9236 (19.15)      3.1715 (48.06)     0.4406 (36.86)      3.2131 (52.81)     0.6879 (126.04)       79;1     315.3127 (0.02)        222           1
test_context_autoload_meta (NOW)                               2.3794 (47.06)      5.6518 (21.98)      2.7528 (41.71)     0.4065 (34.01)      2.6396 (43.38)     0.3319 (60.80)       24;10     363.2701 (0.02)        258           1
test_context_explicit_meta (0001_5e752be)                      0.0633 (1.25)       0.4497 (1.75)       0.0768 (1.16)      0.0189 (1.58)       0.0724 (1.19)      0.0055 (1.0)        99;149  13,019.3341 (0.86)       1638           1
test_context_explicit_meta (NOW)                               0.0506 (1.0)        0.2571 (1.0)        0.0660 (1.0)       0.0163 (1.36)       0.0608 (1.0)       0.0098 (1.79)      169;165  15,153.7272 (1.0)        2294           1
test_context_explicit_meta_config_actions (0001_5e752be)       0.0703 (1.39)       2.8595 (11.12)      0.0863 (1.31)      0.0624 (5.22)       0.0788 (1.30)      0.0103 (1.89)       80;256  11,581.2232 (0.76)       4338           1
test_context_explicit_meta_config_actions (NOW)                0.0554 (1.10)       0.3183 (1.24)       0.0688 (1.04)      0.0120 (1.0)        0.0674 (1.11)      0.0078 (1.43)      301;246  14,532.6029 (0.96)       4378           1
test_deferred_events (0001_5e752be)                           13.6176 (269.35)    21.5901 (83.97)     17.4270 (264.08)    2.1103 (176.53)    17.3957 (285.91)    3.3594 (615.51)       16;0      57.3821 (0.00)         50           1
test_deferred_events (NOW)                                    12.1701 (240.72)   188.6504 (733.69)    16.8910 (255.96)   23.6401 (>1000.0)   13.1619 (216.32)    1.9912 (364.82)        1;3      59.2031 (0.00)         55           1
test_full_state (0001_5e752be)                                 0.1276 (2.52)       0.8470 (3.29)       0.1794 (2.72)      0.0790 (6.61)       0.1706 (2.80)      0.0313 (5.73)        37;51   5,574.1843 (0.37)       1212           1
test_full_state (NOW)                                          0.1052 (2.08)       0.5225 (2.03)       0.1257 (1.90)      0.0193 (1.62)       0.1204 (1.98)      0.0137 (2.51)      305;201   7,956.9504 (0.53)       3252           1
test_lots_of_logs (0001_5e752be)                             189.5452 (>1000.0)  357.2306 (>1000.0)  233.8104 (>1000.0)  70.0612 (>1000.0)  209.5835 (>1000.0)  61.6745 (>1000.0)       1;1       4.2770 (0.00)          5           1
test_lots_of_logs (NOW)                                      167.7643 (>1000.0)  312.8763 (>1000.0)  200.6293 (>1000.0)  63.0276 (>1000.0)  171.3189 (>1000.0)  46.7636 (>1000.0)       1;1       4.9843 (0.00)          5           1
test_many_tests_autoload_meta (0001_5e752be)                  42.9857 (850.23)    54.4689 (211.84)    46.1764 (699.74)    2.7820 (232.73)    45.2766 (744.14)    3.5451 (649.51)        6;1      21.6561 (0.00)         21           1
test_many_tests_autoload_meta (NOW)                           37.2540 (736.86)    54.6456 (212.53)    44.8380 (679.46)    5.9808 (500.31)    41.4991 (682.06)    9.9863 (>1000.0)       7;0      22.3025 (0.00)         21           1
test_many_tests_explicit_meta (0001_5e752be)                  21.3201 (421.70)    97.6800 (379.89)    28.5463 (432.58)   12.1429 (>1000.0)   26.7984 (440.44)    4.2433 (777.45)        1;1      35.0308 (0.00)         36           1
test_many_tests_explicit_meta (NOW)                           23.9549 (473.81)    71.4429 (277.85)    28.0195 (424.60)    8.6270 (721.67)    25.9442 (426.41)    2.6933 (493.45)        1;2      35.6894 (0.00)         29           1
test_run_no_observer (0001_5e752be)                            6.4473 (127.52)    12.1620 (47.30)      7.3986 (112.12)    0.9792 (81.91)      7.0693 (116.19)    0.7906 (144.85)        5;3     135.1615 (0.01)         50           1
test_run_no_observer (NOW)                                     5.8797 (116.30)    11.7443 (45.68)      7.1829 (108.85)    1.2339 (103.22)     6.7236 (110.50)    1.2222 (223.92)        7;4     139.2189 (0.01)         57           1
test_run_observed (0001_5e752be)                               7.1957 (142.33)    15.5289 (60.39)      8.7758 (132.99)    1.2629 (105.65)     8.5763 (140.96)    1.0103 (185.10)       19;9     113.9503 (0.01)         98           1
test_run_observed (NOW)                                        6.7909 (134.32)    12.8278 (49.89)      7.9440 (120.38)    1.0517 (87.98)      7.6466 (125.67)    0.9699 (177.70)       17;6     125.8808 (0.01)        102           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```
</details>

Fixes #2056

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -566,30 +566,9 @@ _DEFAULT_JUJU_DATABAG = {
 }
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, kw_only=True)
 class Relation(RelationBase):
     """A relation between the charm and another application."""
-
-    endpoint: str
-    """Relation endpoint name. Must match some endpoint name defined in the metadata."""
-
-    interface: str | None = None
-    """Interface name. Must match the interface name attached to this endpoint in the metadata.
-    If left empty, it will be automatically derived from the metadata."""
-
-    _: dataclasses.KW_ONLY
-
-    id: int = dataclasses.field(default_factory=_next_relation_id)
-    """Juju relation ID. Every new Relation instance gets a unique one,
-    if there's trouble, override."""
-
-    local_app_data: RawDataBagContents = dataclasses.field(default_factory=dict)
-    """This application's databag for this relation."""
-
-    local_unit_data: RawDataBagContents = dataclasses.field(
-        default_factory=lambda: _DEFAULT_JUJU_DATABAG.copy(),
-    )
-    """This unit's databag for this relation."""
 
     remote_app_name: str = 'remote'
     """The name of the remote application, as in the charm's metadata."""
@@ -634,30 +613,9 @@ class Relation(RelationBase):
         yield from self.remote_units_data.values()
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, kw_only=True)
 class SubordinateRelation(RelationBase):
     """A relation to share data between a subordinate and a principal charm."""
-
-    endpoint: str
-    """Relation endpoint name. Must match some endpoint name defined in the metadata."""
-
-    interface: str | None = None
-    """Interface name. Must match the interface name attached to this endpoint in the metadata.
-    If left empty, it will be automatically derived from the metadata."""
-
-    _: dataclasses.KW_ONLY
-
-    id: int = dataclasses.field(default_factory=_next_relation_id)
-    """Juju relation ID. Every new Relation instance gets a unique one,
-    if there's trouble, override."""
-
-    local_app_data: RawDataBagContents = dataclasses.field(default_factory=dict)
-    """This application's databag for this relation."""
-
-    local_unit_data: RawDataBagContents = dataclasses.field(
-        default_factory=lambda: _DEFAULT_JUJU_DATABAG.copy(),
-    )
-    """This unit's databag for this relation."""
 
     remote_app_data: RawDataBagContents = dataclasses.field(default_factory=dict)
     """The current content of the remote application databag."""
@@ -702,30 +660,9 @@ class SubordinateRelation(RelationBase):
         return f'{self.remote_app_name}/{self.remote_unit_id}'
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, kw_only=True)
 class PeerRelation(RelationBase):
     """A relation to share data between units of the charm."""
-
-    endpoint: str
-    """Relation endpoint name. Must match some endpoint name defined in the metadata."""
-
-    interface: str | None = None
-    """Interface name. Must match the interface name attached to this endpoint in the metadata.
-    If left empty, it will be automatically derived from the metadata."""
-
-    _: dataclasses.KW_ONLY
-
-    id: int = dataclasses.field(default_factory=_next_relation_id)
-    """Juju relation ID. Every new Relation instance gets a unique one,
-    if there's trouble, override."""
-
-    local_app_data: RawDataBagContents = dataclasses.field(default_factory=dict)
-    """This application's databag for this relation."""
-
-    local_unit_data: RawDataBagContents = dataclasses.field(
-        default_factory=lambda: _DEFAULT_JUJU_DATABAG.copy(),
-    )
-    """This unit's databag for this relation."""
 
     peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(default_factory=dict)
     """Current contents of the peer databags.
@@ -2046,12 +1983,10 @@ class _EventPath(str):
         instance.name = name = string.split('.')[-1]
         instance.owner_path = string.split('.')[:-1] or ['on']
 
-        instance.suffix, instance.type = suffix, _ = _EventPath._get_suffix_and_type(
-            name,
-        )
-        instance.prefix = string.removesuffix(suffix)
+        instance.suffix, instance.type = _EventPath._get_suffix_and_type(name)
+        instance.prefix = string.removesuffix(instance.suffix)
+        instance._is_custom = instance.suffix == ''
 
-        instance.is_custom = suffix == ''
         return instance
 
     @staticmethod


tonyandrewmeyer:
Moving back to draft while I figure out why the observability tests are failing. Will re-ping you when that's done.

tonyandrewmeyer:
@james-garner-canonical , @dimaqq this can be reviewed now. The issue was that the downstream settings for pyright were requiring compatibility with 3.8.

testing/src/scenario/state.py
```diff
@@ -633,6 +570,27 @@ def _validate_databag(self, databag: dict[str, str]):
...
+    if there's trouble, override."""
+
+    local_app_data: RawDataBagContents = dataclasses.field(default_factory=dict)
+    """This application's databag for this relation."""
+
+    local_unit_data: RawDataBagContents = dataclasses.field(
+        default_factory=lambda: _DEFAULT_JUJU_DATABAG.copy(),
+    )
+    """This unit's databag for this relation."""
+
```
james-garner-canonical: It shouldn't be necessary to repeat the fields from the parent class here. Since the subclass only adds `kw_only` fields, we should be able to just use `kw_only=True` in the subclass decorator.

testing/src/scenario/state.py
```diff
@@ -680,6 +638,27 @@ def _databags(self):  # type: ignore
...
```
james-garner-canonical: Likewise here, we should be able to replace this with:
```py
```

testing/src/scenario/state.py
```diff
@@ -727,6 +706,27 @@ def remote_unit_name(self) -> str:
...
```
james-garner-canonical: Ditto.

testing/src/scenario/state.py
```diff
@@ -2059,12 +2049,7 @@ def __new__(cls, string: str):
...
         )
-        # TODO: when we drop Python 3.8, we can change the whole if-else below to
-        # instance.prefix = string.removesuffix(suffix)
-        if suffix:
-            instance.prefix, _ = string.rsplit(suffix, maxsplit=1)
-        else:
-            instance.prefix = string
+        instance.prefix = string.removesuffix(suffix)
 
         instance.is_custom = suffix == ''
```
james-garner-canonical: Style comment, non-blocking: Do you think this might be more readable than the current chain assignment while retaining the same optimization of using a local variable over an instance lookup?
```python
suffix, instance.type = _EventPath._get_suffix_and_type(name)
instance.prefix = string.removesuffix(suffix)
instance.suffix = suffix
instance._is_custom = suffix == ''
```
Alternatively, if the current optimization isn't worth it (which I don't really think it is):
```python
```
I think this would be in scope for this PR, since we now look up `suffix` one less time due to dropping the `if`.

james-garner-canonical (changes requested):
Thanks for taking care of cleaning up this now unneeded code! I've requested changes because I think the relation subclasses could be handled without repeating the attributes. I also have a readability suggestion for the constructor where we can now use `removesuffix`.

testing/src/scenario/state.py
james-garner-canonical: PS: Maybe `instance._is_custom = not instance.suffix` would be nicer readability too. It handles non-string values differently, but presumably we're guaranteed a string by this point since we `removesuffix` with it.

testing/src/scenario/state.py
tonyandrewmeyer: Thanks - I wasn't aware that dataclasses worked like that.

testing/src/scenario/state.py
tonyandrewmeyer: I've gone with the first suggestion (I agree that the optimisation of having a local variable is not worth it). I think comparing to the empty string explicitly is clearer in the intent, so left that one as-is.

james-garner-canonical (approved):

dimaqq (approved):
Thank you!

A question: how sure are we that every runtime code path subject to these changes is already covered by existing unit tests?

For example, I wouldn't know how to use `KW_ONLY` to demarcate regular from keywordey argument off the top of my head.

tonyandrewmeyer:
> A question: how sure are we that every runtime code path subject to these changes is already covered by existing unit tests?

Pretty sure.

> In Relation.save, get_type_hints will always work without needing to fall back to __annotations__

This is removing a code path, so not relevant.

> In our test dataclasses we can use kw_only=True

These are only in tests, so they are definitely tested :)

> We don't need to skip tests that don't work with the 3.8-compatible Pydantic

These were simply not running any more now that we don't test against 3.8.

> Remove a comment that refers to unittest in 3.10, since we use pytest now

Just a comment, not relevant.

> Use type[x] rather than Type[x] in ops-tracing

Only relevant for the type checker.

> Drop the custom implementation of kw_only=True and KW_ONLY from the Scenario state objects in favour of the native dataclasses functionality

When the 3.8 version was added that included tests for all the state objects (that they accepted the correct number of positional arguments, with the rest needing to be keyword). Those tests are all still in place and all passing.

> Add one use of removesuffix

This is in very core Scenario, so is exercised in most Scenario tests.

PR #2174, opened 2025-11-14
refactor: use json.dumps to produce the YAML in relation-set and state-set

tonyandrewmeyer:
`relation-set` and `state-set` both produce YAML of a simple `str:str` dictionary that is then loaded by the Juju Go YAML library.

Rather than using `yaml.safe_dump` for this, use `json.dumps`, which will produce JSON that is compatible YAML for this simple case. This avoids a dependency on `ops._private.yaml`, as well as the YAML library and the extra surface area of dumping YAML.

This is also much faster, although no-one will actually notice that of course:

```
$ uvx --with=pyyaml --python=3.14 python -m timeit -n 50000 -s 'import yaml;a={"a":"1","b":"2","c":"3"}' 'yaml.dump(a, Dumper=yaml.CSafeDumper)'
50000 loops, best of 5: 22.1 usec per loop
$ uvx --python=3.14 python -m timeit -n 50000 -s 'import json;a={"a":"1","b":"2","c":"3"}' 'json.dumps(a)'
50000 loops, best of 5: 1.65 usec per loop
```

Suggested by @carlcsaposs-canonical , thanks!


dimaqq:
I have two questions, albeit related, about this:

1. binary compatibility
2. juju compatibility

Suppose the charm saves some binary (or otherwise weird) strings.
YAML has a few ways to escape the data.
JSON doesn't allow random bytes IIRC, only a subset, which is superset of good unicode (or something like that).

---

### bytes

Practical example:
```py
In [18]: yaml.safe_dump({"a": b"bbb"})
Out[18]: 'a: !!binary |\n  YmJi\n'

In [19]: import json

In [20]: json.dumps({"a": b"bbb"})
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
```

### unicode escape

The unit agent will parse our data as YAML, and JSON should be the subset of YAML, however I suspect that treatment of weird unicode (e.g. unpaired surrogates) would vary.

---

At the same time, if we are absolutely sure that values are well-formed strings, there should be no issue.

tonyandrewmeyer:
> At the same time, if we are absolutely sure that values are well-formed strings, there should be no issue.

`RelationData` is defined as `MutableMapping[str, str]`. I believe that means that we do not support anything other than `str`:`str` content, so if you're using the model backend, then you already need to have correctly encoded strings (and if you're wanting to use `hookcmds` directly for this, that isn't something we are adding support for).

If the charm is saving binary data (for example, it's compressing the content), then it needs to encode that as a string before saving it to the relation data.

On the Juju side, if you write binary `bbb` to the data, then when you read it back, you get the *string* `bbb`. I can't find any documentation that definitely states this, but I don't think Juju supports binary relation data, only strings. It seems likely to me that invalid Unicode would fail when loaded in Go as well as Python, and, as above, I don't believe we support that as a use-case.

PR #2175, opened 2025-11-14
fix: minor hookcmds fixes

tonyandrewmeyer:
Small fixes for the `hookcmds` package:

* When flattening action parameters, recurse into any mapping, not just mutable mappings (this leaves the `model` version unchanged, but that's more specifically expecting a dict)
* Correct the return type of `action-get`, which can be any action parameter type, which means essentially `Any`
* Correct the docstring of `relation-get`'s `unit` argument

These were reported by @carlcsaposs-canonical , thanks!

### DIFF
--- a/ops/hookcmds/_action.py
+++ b/ops/hookcmds/_action.py
@@ -63,7 +63,6 @@ def format_result_dict(
             key = f'{parent_key}.{key}'
 
         if isinstance(value, Mapping):
-            value = cast('dict[str, Any]', value)
             output_ = format_result_dict(value, key, output_)
         elif key in output_:
             raise ValueError(


ops/hookcmds/_relation.py
```diff
@@ -59,7 +59,8 @@ def relation_get(
         id: The ID of the relation to get data for, or ``None`` to get data for
             the relation that triggered the current hook.
         key: The specific key to get data for, or ``None`` to get all data.
-        unit: The unit to get data for, or ``None`` to get data for all units.
+        unit: The unit to get data for, or ``None`` to get data for the unit
+            that triggered the current hook.
```
carlcsaposs-canonical, line 63: ~~I believe `None` will get data for the unit the charm code is running on, not necessarily the unit that triggered the current event~~

carlcsaposs-canonical: nvm I was mistaken

(think I got mixed up by units having the same data in the databag)

ops/hookcmds/_action.py
```diff
@@ -114,11 +114,7 @@ def action_get(key: str | None = None) -> dict[str, Any] | str:
     if key is not None:
         args.append(key)
     stdout = run('action-get', *args)
-    result = (
-        cast('dict[str, Any]', json.loads(stdout))
-        if key is None
-        else cast('str', json.loads(stdout))
-    )
+    result = cast('dict[str, Any]', json.loads(stdout)) if key is None else json.loads(stdout)
     return result
```
james-garner-canonical, line 117: Sorry to revisit a discussion we've had before, but I really don't think the '`cast` from `json.loads` then `return`' pattern is beneficial here:
- We're now returning an `Any` typed object in one branch (the result of `json.loads(stdout)`).
- All other types, including `dict[str, Any]` are logically subtypes of it.
- The previous `cast` to two disjoint types was unnecessary due to the immediate return, but casting to a subtype feels somehow even more unnecessary.

Additionally:
- The overloads already capture the return type logic of this function clearly and concisely, and that's all the user will see.
- Most importantly, the conditional cast complexity for readers and maintainers to an otherwise very short and clear function -- for example the changes in this PR are really only to the return type annotation, but require touching the internal `if ... else` expression too without making any actual runtime changes.

```suggestion
    return json.loads(stdout)
```
If we're really committed to using cast internally to re-document return types, how about lifting the actual evaluation of the result out of the cast logic:
```python
    result = json.loads(stdout)
    if key is None:
        return cast('dict[str, Any]', result)
    return result
```

ops/hookcmds/_action.py
```diff
@@ -62,7 +62,7 @@ def format_result_dict(
         if parent_key:
             key = f'{parent_key}.{key}'
 
-        if isinstance(value, MutableMapping):
+        if isinstance(value, Mapping):
             value = cast('dict[str, Any]', value)
```
james-garner-canonical: Should this be:
```suggestion
            value = cast('Mapping[str, Any]', value)
```
?

ops/hookcmds/_action.py
tonyandrewmeyer: > Sorry to revisit a discussion we've had before

I do feel like this is shoe-horning in a previously-resolved discussion into a PR that is mostly unrelated.

I'm ok with re-opening this discussion, but I'd rather that was done through opening an issue or PR that addresses this throughout all of `hookcmds` (or alternatively, adding to the Charm Tech daily discussions).

ops/hookcmds/_action.py
tonyandrewmeyer: I think this is actually unnecessary because `input` is a `Mapping`, unlike in the original in `model.py` where it is a `dict`.

ops/hookcmds/_action.py
james-garner-canonical: My rationale for revisiting it here is that *even if we think it's the right choice everywhere else in hookcmds* I think the changes introduced in this PR make it the wrong choice here, and I'd prefer it not to be committed as-is. But I approved because I do think it's valid call to want to keep it this way for consistency and a smaller diff.

ops/hookcmds/_action.py
james-garner-canonical: Ah that makes sense, I had assumed it was here because we wanted to narrow the content type of the mapping, but I guess it's narrowed to `Mapping[Any, Any]` here which satisfies the type checker.

ops/hookcmds/_action.py
tonyandrewmeyer: Yes, which is why I'm leaving it in place. One of the primary reasons that `hookcmds` is like this is because there was a desire / request to have all of the methods consistent in this area, so I don't want to break that consistency in this PR.

PR #2176, opened 2025-11-14
docs: Add links to the Juju hook from each event class

### LINKED ISSUE #2047: Add links from each event to the Juju hook documentation
We should link to the appropriate section in [the Juju hook doc](https://documentation.ubuntu.com/juju/3.6/reference/hook/#list-of-hooks) from either the `CharmEvent` (e.g. "on.config_changed") or various event (e.g. "ConfigChangedEvent") reference docs.

We planned to do this a while ago, but Dora felt at the time that the Juju docs need work before they should be linked to, but that work has been done now. This helps with the connectivity of the various charming ecosystem docs, and avoids having to duplicate much information in the Ops docs, as charmers can get the information from the Juju page instead.

tonyandrewmeyer:
Link to the appropriate section of the Juju `Hook` from each event class, so that all of the detailed information about the Juju event can live in the Juju documentation, but be easily found from the Ops docs.

Also a drive-by fix to update a Charmcraft docs link from the ReadTheDocs site to the documentation.ubuntu.com one.

Fixes #2047


james-garner-canonical (approved):
A very nice docs improvement!

PR #2179, opened 2025-11-17
refactor: rely on type annotations instead of casts in hookcmds

james-garner-canonical:
`hookcmds` functions that return values generally work by calling `json.loads` on the result of a hook command CLI invocation. Values returned by `json.loads` are typed as `Any`, but we know more about the values returned by different hook commands than this.

Currently, this is expressed by calling `typing.cast` when calling `json.loads`. This PR changes this to be expressed with type annotations, for improved readability and lower runtime overhead.

### DIFF
--- a/ops/hookcmds/_action.py
+++ b/ops/hookcmds/_action.py
@@ -109,7 +109,10 @@ def action_get(key: str | None = None) -> dict[str, Any] | Any:
     if key is not None:
         args.append(key)
     stdout = run('action-get', *args)
-    return json.loads(stdout)
+    if key is not None:
+        return json.loads(stdout)
+    result: dict[str, Any] = json.loads(stdout)
+    return result
 
 
 def action_log(message: str):
--- a/ops/hookcmds/_other.py
+++ b/ops/hookcmds/_other.py
@@ -68,7 +68,11 @@ def config_get(
     if key:
         args.append(key)
     stdout = run('config-get', *args)
-    return json.loads(stdout)
+    if key:
+        key_result: bool | int | float | str = json.loads(stdout)
+        return key_result
+    result: dict[str, bool | int | float | str] = json.loads(stdout)
+    return result
 
 
 def credential_get() -> CloudSpec:
@@ -103,7 +107,8 @@ def is_leader() -> bool:
     `Juju | Hook commands | is-leader <https://documentation.ubuntu.com/juju/3.6/reference/hook-command/list-of-hook-commands/is-leader/>`_
     """
     stdout = run('is-leader', '--format=json')
-    return json.loads(stdout)
+    result: bool = json.loads(stdout)
+    return result
 
 
 def juju_log(
--- a/ops/hookcmds/_relation.py
+++ b/ops/hookcmds/_relation.py
@@ -79,7 +79,11 @@ def relation_get(
         # The unit is not required when inside a relation hook other than relation-broken.
         args.append(key)
     stdout = run('relation-get', *args)
-    return json.loads(stdout)
+    if key is not None:
+        key_result: str = json.loads(stdout)
+        return key_result
+    result: dict[str, str] = json.loads(stdout)
+    return result
 
 
 def relation_ids(name: str) -> list[str]:
@@ -92,7 +96,8 @@ def relation_ids(name: str) -> list[str]:
         name: the endpoint name.
     """
     stdout = run('relation-ids', name, '--format=json')
-    return json.loads(stdout)
+    result: list[str] = json.loads(stdout)
+    return result
 
 
 @overload
@@ -120,7 +125,11 @@ def relation_list(id: int | None = None, *, app: bool = False) -> str | list[str
     if id is not None:
         args.extend(['-r', str(id)])
     stdout = run('relation-list', *args)
-    return json.loads(stdout)
+    if app:
+        app_result: list[str] = json.loads(stdout)
+        return app_result
+    result: str = json.loads(stdout)
+    return result
 
 
 def relation_model_get(id: int | None = None) -> RelationModel:
--- a/ops/hookcmds/_state.py
+++ b/ops/hookcmds/_state.py
@@ -51,7 +51,11 @@ def state_get(key: str | None) -> dict[str, str] | str:
     if key is not None:
         args.append(key)
     stdout = run('state-get', *args)
-    return json.loads(stdout)
+    if key is not None:
+        key_result: str = json.loads(stdout)
+        return key_result
+    result: dict[str, str] = json.loads(stdout)
+    return result
 
 
 # We don't offer a `file` argument here as we expect that charms will generally
--- a/ops/hookcmds/_secret.py
+++ b/ops/hookcmds/_secret.py
@@ -151,7 +151,8 @@ def secret_get(
     if peek:
         args.append('--peek')
     stdout = run('secret-get', '--format=json', *args)
-    return json.loads(stdout)
+    result: dict[str, str] = json.loads(stdout)
+    return result
 
 
 def secret_grant(id: str, relation_id: int, *, unit: str | None = None):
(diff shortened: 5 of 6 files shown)


ops/hookcmds/_action.py
```diff
@@ -113,8 +109,7 @@ def action_get(key: str | None = None) -> dict[str, Any] | Any:
     if key is not None:
         args.append(key)
     stdout = run('action-get', *args)
-    result = cast('dict[str, Any]', json.loads(stdout)) if key is None else json.loads(stdout)
-    return result
+    return json.loads(stdout)
```
james-garner-canonical: This logic is captured by the overloads.

ops/hookcmds/_other.py
```diff
@@ -68,11 +68,7 @@ def config_get(
     if key:
         args.append(key)
     stdout = run('config-get', *args)
-    if key:
-        result = cast('bool | int | float | str', json.loads(stdout))
-    else:
-        result = cast('dict[str, bool | int | float | str]', json.loads(stdout))
-    return result
+    return json.loads(stdout)
```
james-garner-canonical: This logic is captured by the overloads.

ops/hookcmds/_relation.py
```diff
@@ -80,11 +80,7 @@ def relation_get(
         # The unit is not required when inside a relation hook other than relation-broken.
         args.append(key)
     stdout = run('relation-get', *args)
-    if key is not None:
-        result = cast('str', json.loads(stdout))
-    else:
-        result = cast('dict[str, str]', json.loads(stdout))
-    return result
+    return json.loads(stdout)
```
james-garner-canonical: This logic is captured by the overloads.

ops/hookcmds/_state.py
```diff
@@ -55,12 +52,7 @@ def state_get(key: str | None) -> dict[str, str] | str:
     if key is not None:
         args.append(key)
     stdout = run('state-get', *args)
-    result = (
-        cast('dict[str, str]', json.loads(stdout))
-        if key is None
-        else cast('str', json.loads(stdout))
-    )
-    return result
+    return json.loads(stdout)
```
james-garner-canonical: This logic is captured by the overloads.

ops/hookcmds/_status.py
```diff
@@ -44,10 +44,11 @@ def status_get(*, app: bool = False) -> AppStatus | UnitStatus:
...
     args = ['--include-data', '--format=json', f'--application={str(app).lower()}']
     stdout = run('status-get', *args)
+    result = json.loads(stdout)
     if app:
-        result = cast('AppStatusDict', json.loads(stdout))
+        result = cast('AppStatusDict', result)
         return AppStatus._from_dict(result)
-    result = cast('StatusDict', json.loads(stdout))
+    result = cast('StatusDict', result)
     return UnitStatus._from_dict(result)
```
james-garner-canonical: This adds an additional line, but I think it makes sense to lift the common call (`json.loads`) out of the conditional return logic.

benhoyt: I'm ambivalent about this one, but not opposed to it.

james-garner-canonical: A terser version that does the same thing would be this, which I'm also ok with.
```python
    result = json.loads(stdout)
    if app:
        return AppStatus._from_dict(cast('AppStatusDict', result))
    return UnitStatus._from_dict(cast('StatusDict', result))
```

james-garner-canonical: Actually there are a few versions that have their own tradeoffs.

Using an annotation instead of a `cast`, with the downside of requiring more variable names.
```python
    result = json.loads(stdout)
    if app:
        app_result: AppStatusDict = result
        return AppStatus._from_dict(app_result)
    unit_result: StatusDict = result
    return UnitStatus._from_dict(unit_result)
```

A terser version with annotations, with the downside of putting `json.loads` in the branches.
```python
    if app:
        app_result: AppStatusDict = json.loads(stdout)
        return AppStatus._from_dict(app_result)
    result: StatusDict = json.loads(stdout)
    return UnitStatus._from_dict(result)
```

benhoyt: That last version seems most consistent with the other code (same shape). The `json.loads` in each branch doesn't bother me at all.

james-garner-canonical: Cool, I've committed that version.

ops/hookcmds/_action.py
dimaqq: Curious, I would have imagined that the cast was there specifically to assuage pyright fears wrt the overloads.

tonyandrewmeyer:
> 1. The type narrowing from `Any` to one specific type can be expressed by a type annotation instead of a `cast`, which is more idiomatic and has less runtime overhead.

I have no objection to this change.

> 2. Most of the time when we conditionally narrow the type based on the arguments, we immediately return the result. In this case, the return type annotation already expresses the type narrowing via overloads, so we can directly `return json.loads(stdout)`.

I feel like this goes in the opposite direction from the decision we just made to always have a return type annotation. The argument then was that it's protection against the code and the return value type accidentally diverging, because you get the inferred type and the explicit type. Here, the suggestion is that we go away from the explicit type in the return and explicit casting (or annotating) in the body to only having one of those.

I'm also unsure whether we should be relying on overloads to be the documentation for what Juju returns (having this documentation is the critical point here, in my opinion, because it's not anywhere else other than the Juju code, and it was a lot of work to figure it all out, and we don't want to make it hard for people to work with this in the future). For example, this is a common pattern:

```python
@overload
def tool(key: None = None) -> dict[str, str]: ...
def tool(key: str) -> str: ...
def tool(key: str | None = None) -> str | dict[str, str]:
    ...
    if key is not None:
        result: str = json.loads(stdout)
    else:
        result: dict[str, str] = json.loads(stdout)
    return result
```

If that is instead:

```python
@overload
def tool(key: None = None) -> dict[str, str]: ...
def tool(key: str) -> str: ...
def tool(key: str | None = None) -> str | dict[str, str]:
    ...
    return json.loads(stdout)
```

It's shorter and simpler and will pass the type checks, and the type checker will be able to figure out whether the caller is getting a `str` or `dict` based on the input. However, as a *human* reading this code, I have to use the overloads to understand that passing in a key means getting a string back, and otherwise I get a dict. I think that's less clear than the simple if statement, especially if there are actually many combinations of overloads.

@benhoyt asked for all of `hookcmds` to use the same pattern for loading the content from the hook command, annotating it appropriately, and then returning the value. That's why they all have exactly the same layout even when there's simpler ways of handling some of them. I agree that this consistency has value, particularly if you're adding or editing `hookcmds` methods. What are we gaining in exchange for losing the consistency?

> 3. The current pattern involves different runtime code execution paths when the actual difference is at type checking time only

I have some sympathy for this argument. However, it's typically just going to be a cast or annotation, so I don't think there's much risk of, for example, code wrongly diverging. From a coverage point of view, I feel it's actually nicer, since you can tell if there are tests exercising each type, which I don't think you can get otherwise (there's no "typing coverage", right?).

I suspect that this is best discussed in real-time rather than solely in a PR.

ops/hookcmds/_action.py
tonyandrewmeyer: The cast was added to explicitly document what type we get back from Juju (consistently, across the entire package. Since `json.loads` returns `Any`, the awful "ignore all typing" trick, the casting isn't needed if you're willing to trust that the combination of overloads (if any) and primary method signature are enough.

tonyandrewmeyer (changes requested):
Since this approach was specifically requested in a PR review only a couple of months ago, and since it breaks the consistency that was particularly wanted, and since I'm not convinced that we gain enough from it, or that we are left with enough documentation for *humans*, I would prefer that this is discussed in real-time before giving this an "approve". Please organise a call (or use the daily Charm Tech sync) for that.

ops/hookcmds/_action.py
james-garner-canonical: Note also that any types declared in the function body don't directly interact with the overload signatures. Pyright only checks that the implementation signature is compatible with the overload signatures, and checks that what it knows about the real return type matches the implementation signature.

PR #2182, opened 2025-11-18
chore: prepare a 3.4.0b3

tonyandrewmeyer:
The `[testing]` extra had the wrong `ops` version :(

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # 3.4.0b3 - 18 November 2025
 
-No changes to the packages themselves, just fixes the testing optional extra version.
+3.4.0b2 was released on GitHub, but not on PyPI, because the `testing` optional extra had the wrong version for ops. 3.4.0b3 fixes this but has no other changes from 3.4.0b2.
 
 # 3.4.0b2 - 18 November 2025
 


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# 3.4.0b3 - 18 November 2025
+
+No changes to the packages themselves, just fixes the testing optional extra version.
+
```
dimaqq, line 4: I'm OK keeping this, however I think it would be better to hoist the changes under the b3 heading:

```
# 3.4.0b3 - 18 November 2025

- change (deferred event count)
- change 
- change

# 3.4.0b2 - never released

# 3.4.0b1 - ...

 - change
 - change
 - change
```

tonyandrewmeyer: It was released on GitHub, but not PyPI. I'll try to make that clearer.

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# 3.4.0b3 - 18 November 2025
+
+No changes to the packages themselves, just fixes the testing optional extra version.
```
tonyandrewmeyer:
```suggestion
3.4.0b2 was released on GitHub, but not on PyPI, because the `testing` optional extra had the wrong version for ops. 3.4.0b3 fixes this but has no other changes from 3.4.0b2.
```

PR #2184, opened 2025-11-19
docs: add a short umask note

setharnold:
Charm authors are responsible for their umask(2) setting. https://github.com/canonical/operator/issues/2181

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -76,7 +76,7 @@ Otherwise, Ops doesn't introduce any new security risks. Ops does expand the imp
 * Use {external+juju:ref}`Juju secrets <secret>` for storing and sharing sensitive data.
 * Juju users that integrate a charm with a tracing receiver should also integrate with a certificate authority provider, to ensure all traces are sent via HTTPS.
 * Charms should follow best practices for writing secure Python code.
-* Charms should manage Unix permissions, including [`umask(2)`](https://manpages.ubuntu.com/manpages/noble/man2/umask.2.html) to respect the [principle of least authority](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
+* Machine charms are responsible for setting appropriate ownership and permissions on the files and directories they create, for example using {py:func}`os.umask`.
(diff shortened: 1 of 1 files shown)


docs/explanation/security.md
```diff
@@ -76,6 +76,7 @@ Otherwise, Ops doesn't introduce any new security risks. Ops does expand the imp
 * Use {external+juju:ref}`Juju secrets <secret>` for storing and sharing sensitive data.
 * Juju users that integrate a charm with a tracing receiver should also integrate with a certificate authority provider, to ensure all traces are sent via HTTPS.
 * Charms should follow best practices for writing secure Python code.
+* Charms should manage Unix permissions, including [`umask(2)`](https://manpages.ubuntu.com/manpages/noble/man2/umask.2.html) to respect the [principle of least authority](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
```
tonyandrewmeyer: Since this is a framework for writing charms in Python, I wonder if it would be better to link to [the Python docs](https://docs.python.org/3/library/os.html#os.umask) rather than manpages. I think that would be `:py:func:`python:os.umask`.

I wonder about "Unix" too, whether it should really be POSIX, or just plain "file permissions" perhaps?

james-garner-canonical: This is a good addition to the docs, thanks. Linking to the Python docs is a good idea, since that's presumably how charms will need to set it, but the manpage docs are much more helpful, so I'd be in favour of having both. I don't love the Wikipedia link, but maybe it's fine.

dimaqq:
```suggestion
* Machine charms are responsible for setting appropriate ownership and permissions on the files and directories they create, for example using {py:func}`os.umask`.
```

Reasons for rewriting it so:
- certainly an issue for machine charms, not so much for k8s charms where pods are isolated, and some barriers exist between the containers; specifically everything in the charm container is run as the same user.
- "should" felt too prescriptive, I wanted to avoid the wall of "shoulds" and keep it more neutral/factual.

james-garner-canonical: I like this suggestion.

dimaqq: I think I resolved this in my suggestion.

tonyandrewmeyer (approved):

PR #2192, opened 2025-11-25
ci: store charmcraft logs if smoke tests fail

tonyandrewmeyer:
Sometimes the smoke test fails not in the interaction between the charm and Juju, but before that when the charm is being packed. The charmcraft output is almost always useless in this case, with the actual failure buried in the charmcraft log file.

This PR updates the smoke workflow to upload the charmcraft log files if the smoke tests fails, so that it's easier to investigate in such situations.

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -37,3 +37,10 @@ jobs:
       - name: Run smoke tests
         run: tox -e smoke
         timeout-minutes: 20  # Juju 4 gets stuck while destroying the test model
+
+      - name: Archive charmcraft logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: charmcraft-logs
+          path: /home/runner/.local/state/charmcraft/log/*.log


tonyandrewmeyer:
> I haven't verified that this works as intended, but it's a very nice improvement, and I figure we'll quickly notice if anything is wrong and can follow up with fixes if needed.

For what it's worth, it's a direct copy from the same step I added to the Dell charms, and it worked there.

.github/workflows/smoke.yaml
```diff
@@ -37,3 +37,10 @@ jobs:
       - name: Run smoke tests
         run: tox -e smoke
         timeout-minutes: 20  # Juju 4 gets stuck while destroying the test model
+
+      - name: Archive charmcraft logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: charmcraft-logs
+          path: /home/runner/.local/state/charmcraft/log/*.log
```
dimaqq, line 46: I think I saw this step occasionally failing with "permission denied" somewhere else.

At the same time, the test has already failed, so this won't hurt.

dimaqq (approved):

.github/workflows/smoke.yaml
tonyandrewmeyer: Hmm, I wonder what would cause that, and if the permission was reading the log file or uploading the artefact. Probably easiest to investigate if we see it happen. As you say, it won't make things worse.

PR #2193, opened 2025-11-26
ci: avoid jitter in the best practice doc PRs

tonyandrewmeyer:
* Sort the collected best practices, so that they're in a consistent order (the first PR may have some noise as a result of this).
* Adjust the operator clone so that the operator and charmcraft directories are siblings, which is what the default layout expects.

### DIFF
--- a/.github/workflows/update-best-practice-doc.yaml
+++ b/.github/workflows/update-best-practice-doc.yaml
@@ -14,6 +14,7 @@ jobs:
       - uses: actions/checkout@v5
         with:
           token: ${{ secrets.UPDATE_DOCS_ACCESS_TOKEN }}
+          path: operator
           persist-credentials: true
 
       - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/update-best-practice-doc.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
-      - run: python .github/update-best-practice-table.py > docs/reuse/best-practices.txt
+      - run: python .github/update-best-practice-table.py --path-to-charmcraft=./charmcraft > docs/reuse/best-practices.txt
```
james-garner-canonical: Should we make the default accurate instead?

james-garner-canonical (approved):

.github/workflows/update-best-practice-doc.yaml
tonyandrewmeyer: The default is (intended to be) ideal when you're running this locally. I think that's generally nicer than making the automated one simplest, since in one someone is manually typing stuff and in the other it's all encoded in this file.

However, maybe the better solution is to actually make them both the same? I think I can do that by altering the checkout action instead.

james-garner-canonical: That does make sense. What assumptions is it making for the local setup? `charmcraft` as a sibling to current git root? Wouldn't work out of the box for me due to worktrees, but that doesn't really matter. Current fix seems fine.

tonyandrewmeyer: `pathlib.Path(__file__).parent.parent.parent.parent / 'charmcraft'`, so yes. I should indeed have said "running this locally when you have a layout like me" 😆.

PR #2200, opened 2025-11-27
chore: linting fixes required to remove all the excludes for testing/*

### LINKED ISSUE #1693: [chore] scenario: cleanup the per-file-ignores introduced when including testing in ops linting
In #1694, we removed the `ruff` config from `testing/pyproject.toml`, allowing the `ruff` settings in the `pyproject.toml` in the project directory to apply to `testing/` (note: the actual formatting changes were made separately from the config  changes in #1691, which has been added to the git blame ignore list).

The purpose of this PR was to unify on single-quotes as the quoting style, however it also revealed that many other `ruff` linting rules were not being checked for `testing/`. These rules have been added to the `per-file-ignores` in the project level `pyproject.toml` for `testing/*` and `testing/tests/*`.

We should clean up these ignores as much as possible.

tonyandrewmeyer:
Removes all the entries in the ruff excludes that apply specifically to `testing/*` and were left over from the unification of the ruff settings for ops and Scenario.

In order to not have any warnings, various minor cleanup:

* Wrap some long lines.
* Add noqa's for classes and methods that copy docstrings.
* A few import order tidy-ups.
* Inside of list comps/gen-exps use `exe` rather than `exec` (the shadowing causes no real issue, but this seems simpler than adding the noqa).
* Add noqa's for a few exceptions inside of loops.
* In dataclasses, use `field` with `default_factory` rather than instantiating an actual object.
* Add noqa's for a couple of uses of random, where security is not relevant.
* Add `from None` when raising new exceptions in an exception handler.
* A few other minor adjustments.

Fixes #1693

### DIFF
--- a/testing/src/scenario/_consistency_checker.py
+++ b/testing/src/scenario/_consistency_checker.py
@@ -550,12 +550,13 @@ def check_relation_consistency(
             return ()
 
     # check relation types
-    for relation in _get_relations(endpoint):
-        if not isinstance(relation, PeerRelation):
-            errors.append(
-                f'endpoint {endpoint} is a peer relation; '
-                f'expecting relation to be of type PeerRelation, got {type(relation)}',
-            )
+    for endpoint, _ in peer_relations_meta:
+        for relation in _get_relations(endpoint):
+            if not isinstance(relation, PeerRelation):
+                errors.append(
+                    f'endpoint {endpoint} is a peer relation; '
+                    f'expecting relation to be of type PeerRelation, got {type(relation)}',
+                )
 
     known_endpoints = [a[0] for a in all_relations_meta]
     for relation in state.relations:
--- a/testing/src/scenario/_ops_main_mock.py
+++ b/testing/src/scenario/_ops_main_mock.py
@@ -247,7 +247,7 @@ class Ops(_Manager, Generic[CharmType]):
         try:
             for step in path:
                 obj = getattr(obj, step)
-        except AttributeError:  # noqa: PERF203
+        except AttributeError:
             raise BadOwnerPath(
                 f'event_owner_path {path!r} invalid: {step!r} leads to nowhere.',
             ) from None


testing/src/scenario/_consistency_checker.py
```diff
@@ -551,17 +551,19 @@ def _get_relations(r: str):
...
 
     known_endpoints = [a[0] for a in all_relations_meta]
-    for relation in state.relations:
-        if (ep := relation.endpoint) not in known_endpoints:
-            errors.append(f'relation endpoint {ep} is not declared in metadata.')
+    errors.extend(
+        f'relation endpoint {ep} is not declared in metadata.'
+        for relation in state.relations
+        if (ep := relation.endpoint) not in known_endpoints
+    )
```
james-garner-canonical: I wonder if we should disable this rule broadly. I know @benhoyt isn't a big fan of it, and personally I find the originals a lot more readable for both of these.

testing/src/scenario/_ops_main_mock.py
```diff
@@ -254,10 +247,10 @@ def _get_owner(root: Any, path: Sequence[str]) -> ops.ObjectEvents:
         for step in path:
             try:
                 obj = getattr(obj, step)
-            except AttributeError:
+            except AttributeError:  # noqa: PERF203
```
james-garner-canonical: Since we raise immediately, maybe the right call here is to lift the `try ... except` out of the loop like `PERF203` wants? But this fix is simpler, so I'm fine with leaving it as-is.

testing/src/scenario/state.py
```diff
@@ -1178,9 +1177,13 @@ def layer_from_rockcraft(path: pathlib.Path | str) -> pebble.Layer:
...
     with path.open('r') as f:
         rockcraft = yaml.safe_load(f)
+    description = (
+        f'{rockcraft.get("description", "(no description)")} '
+        f'(built from the rockcraft.yaml at {path})'
+    )
     layer_dict: pebble.LayerDict = {
         'summary': rockcraft['summary'],
-        'description': f'{rockcraft.get("description", "(no description)")} (built from the rockcraft.yaml at {path})',
+        'description': description,
```
james-garner-canonical, line 1186: Could be inlined? But OK as-is if you prefer this.

testing/src/scenario/state.py
```diff
@@ -1547,7 +1550,7 @@ class State:
     """Ports opened by Juju on this charm."""
     leader: bool = False
     """Whether this charm has leadership."""
-    model: Model = Model()
+    model: Model = dataclasses.field(default_factory=Model)
```
james-garner-canonical, line 1553: Technically a breaking change, but presumably no one was relying on the mutable `State.model` default for information sharing ... and likewise for the other changes like this.

james-garner-canonical (approved):
A couple of minor questions/suggestions, but nothing blocking, thanks for doing this.

testing/src/scenario/_consistency_checker.py
benhoyt: I agree this is not an improvement, and I'd be in favour of that. Though I don't feel too strongly.

dimaqq: Personally, I'd probably refactor this "down" a bit, e.g.:

```py
other_relations = [r for r in _get_relations(endpoint) if not isinstance(r, PeerRelation)]
other_errors = [f"...{xx}..." for ...]
errors.extend(other_errors)
```

At the same time, the original code was quite readable, and IMO wins over the one-big-extend-statement version.

I feel the linter rule is valid for tight inner loops, but too prescriptive for coarse-grained code, kinda like "use functional programming over imperative".

For that reason, I am too OK with the rule being disabled.

testing/src/scenario/state.py
tonyandrewmeyer: Yeah, if it's inlined you get the thing where this line doesn't end with a comma, because it's concatenation with the next, but it somewhat looks like the other lines, so it's a little less readable in that way. I think for these it's nicer to pull the string out, so the fields are consistently one per line.

testing/src/scenario/state.py
tonyandrewmeyer: Yeah, I should have called that out, thanks. I think the existing way is a bug, really, even if someone was relying on it, they shouldn't be.

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: I feel this is worse because the string is split over two lines, and if it wasn't the extend rather than loop+append is either a slight improvement or neutral to my eyes.

I'd be fine having a team decision to exclude the rule globally (which would be implemented before merging this and then this change could be dropped). Separate PR, though.

tonyandrewmeyer: https://github.com/canonical/operator/pull/2201

testing/src/scenario/_consistency_checker.py
```diff
@@ -551,17 +551,19 @@ def _get_relations(r: str):
...
-                errors.append(
-                    f'endpoint {endpoint} is a peer relation; '
-                    f'expecting relation to be of type PeerRelation, got {type(relation)}',
-                )
+        errors.extend(
+            f'endpoint {endpoint} is a peer relation; '
+            f'expecting relation to be of type PeerRelation, got {type(relation)}'
+            for relation in _get_relations(endpoint)
+            if not isinstance(relation, PeerRelation)
+        )
```
tonyandrewmeyer:
```suggestion
        for relation in _get_relations(endpoint):
            if not isinstance(relation, PeerRelation):
                errors.append(
                    f'endpoint {endpoint} is a peer relation; '
                    f'expecting relation to be of type PeerRelation, got {type(relation)}',
                )
```

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer:
```suggestion
    for relation in state.relations:
        if (ep := relation.endpoint) not in known_endpoints:
            errors.append(f'relation endpoint {ep} is not declared in metadata.')
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -709,13 +710,13 @@ def check_containers_consistency(
...
-                        f'different {attr!r} ({getattr(check, attr)}) '
-                        f'than the plan ({getattr(plan_check, attr)}).',
-                    )
+            errors.extend(
+                f'container {container.name!r} has a check {check.name!r} with a '
+                f'different {attr!r} ({getattr(check, attr)}) '
+                f'than the plan ({getattr(plan_check, attr)}).'
+                for attr in ('level', 'startup', 'threshold')
+                if getattr(check, attr) != getattr(plan_check, attr)
+            )
```
tonyandrewmeyer:
```suggestion
            for attr in ('level', 'startup', 'threshold'):
                if getattr(check, attr) != getattr(plan_check, attr):
                    errors.append(
                        f'container {container.name!r} has a check {check.name!r} with a '
                        f'different {attr!r} ({getattr(check, attr)}) '
                        f'than the plan ({getattr(plan_check, attr)}).',
                    )
```

testing/src/scenario/_ops_main_mock.py
```diff
@@ -254,10 +247,10 @@ def _get_owner(root: Any, path: Sequence[str]) -> ops.ObjectEvents:
...
                 raise BadOwnerPath(
                     f'event_owner_path {path!r} invalid: {step!r} leads to nowhere.',
-                )
+                ) from None
```
tonyandrewmeyer:
```suggestion
        try:
            for step in path:
                obj = getattr(obj, step)
        except AttributeError:
            raise BadOwnerPath(
                f'event_owner_path {path!r} invalid: {step!r} leads to nowhere.',
            ) from None
```

testing/src/scenario/state.py
```diff
@@ -1573,9 +1576,9 @@ class State:
     """Contents of a charm's stored state."""
 
     # the current statuses.
-    app_status: _EntityStatus = UnknownStatus()
+    app_status: _EntityStatus = dataclasses.field(default_factory=UnknownStatus)
```
dimaqq, line 1579: **mostly homour*:
I hope that none relies on this, e.g. has tests like:
```py
assert state_in.app_status is state_out.app-status
```

dimaqq (commented):
A partial review so far.
I like where this is going!

testing/src/scenario/_consistency_checker.py
dimaqq: How about spelling these out individually instead?

```py
if check.level != plan_check.level:
    errors.append(...)
if check.startup != plan_check.startup:
    ...
...
```

or perhaps

```py
if (a := check.level) != (b := plan_check.level):
    errors.append(f"... {a} {b}")
...

if (actual := check.level) != (planned := plan_check.level):
    errors.append(f"... {actual=} {planned=}")
```

I feel that it would be about as long, and much easier to read.

dimaqq (approved):

Issue #2032, opened 2025-09-04
Clean up ops unit tests wrt optional ops[testing]


dimaqq:
@dimaqq 
Whom do ops tests belong to?

Here's an example test in `canonical/operator/test/test_model.py`:

```py
@pytest.mark.skipif(
    not hasattr(ops.testing, 'Context'), reason='requires optional ops[testing] install'
)
def test_scenario_available():
    ctx = ops.testing.Context(ops.CharmBase, meta={'name': 'foo'})
    state = ctx.run(ctx.on.start(), ops.testing.State())
    assert isinstance(state.unit_status, ops.testing.UnknownStatus)
```

Now that operator is a monorepo, if someone checks out the code, they pretty much always have ops[testing].

However, if someone downloads an sdist, they don't. (I dunno who does that, but still)

I'm leaning towards removing `pytest.mark.skipif` like these.
Do you?

@tonyandrewmeyer 
There are only two of these, from memory, and elsewhere we've decided that ops unit tests will always be run with ops-scenario installed.

I think this specific one the whole test could be removed. It's just testing that the magic extension of ops.testing works when ops-scenario is available, but if that broke all sorts of things would break, so I don't think we need a specific test for it any more.

The other one is making sure that we have all the expected docs. I'd keep that one, I think. Removing the mark is fine if you'd prefer that.

Ref: https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$lvcovmciOio1yqOQcQotXUfD8BmD1leQDmYpG3BJHDU?via=ubuntu.com&via=matrix.org&via=laquadrature.net

PR #2215, opened 2025-12-04
test: don't skip tests if ops[testing] isn't installed

james-garner-canonical:
This PR removes the `skipif` from the testing docs test, and drops the 'scenario is installed' test entirely (per discussion), since we rely on `ops[testing]` in other tests now, so it would always need to be installed.

Resolves: #2032


tonyandrewmeyer (approved):
Thanks!

PR #2216, opened 2025-12-04
refactor!: drop now-unnecessary generator logic

james-garner-canonical:
In #2161, we switched switched to greedily consuming stored notices. This PR updates the internal storage API to directly return a list of rows (tuples), dropping the now unnecessary generator logic.

The Juju storage API's `notices` method is updated to match, as it was already greedy loading from YAML, and then converting the list to a generator to match the storage API. Conversion of rows loaded from YAML to tuples (which would be lists by default) is now handled in the loading method, rather than the `notices` method for more accurate typing.

The primary intended benefit of this PR is code cleanup: now that we're happy to load the notices greedily into memory on consumption, the generator logic is unnecessary. I'd expect this to also be slightly more performant, but that's not a major consideration at all.

I'm open for suggestions on how to exercise these code changes in tests. The `test_storage` suite caught a logic error I made in this PR, and passes now that it's fixed, so maybe we have sufficient coverage already.

### DIFF
--- a/ops/storage.py
+++ b/ops/storage.py
@@ -294,7 +294,9 @@ class JujuStorage:
             Iterable of (event_path, observer_path, method_name) tuples
         """
         notice_list = self._load_notice_list()
-        return [row for row in notice_list if not (event_path and row[0] != event_path)]
+        if not event_path:
+            return notice_list
+        return [row for row in notice_list if row[0] == event_path]
 
     def _load_notice_list(self) -> _Notices:
         """Load a notice list from current key.


ops/storage.py
```diff
@@ -187,15 +186,15 @@ def drop_notice(self, event_path: str, observer_path: str, method_name: str) ->
             (event_path, observer_path, method_name),
         )
 
-    def notices(self, event_path: str | None = None) -> _NoticeGenerator:
+    def notices(self, event_path: str | None = None) -> _Notices:
```
benhoyt, line 189: I think this type already exists, but `list[_Notice]` seems clearly to me. Can we change to that?

ops/storage.py
```diff
@@ -317,7 +310,7 @@ def _load_notice_list(self) -> _Notices:
             return []
         if notice_list is None:
             return []
-        return notice_list
+        return [tuple(notice) for notice in notice_list]
```
benhoyt, line 313: Why is tuple() needed here?

benhoyt (approved):
Fine with me -- just a couple of minor questions.

ops/storage.py
james-garner-canonical: Good call. I prefer `list[_Notice]` as well. I'll replace our uses of `_Notices` with it.

ops/storage.py
james-garner-canonical: Because `notice_list` is loaded from YAML, the rows will be lists. Previously we called `tuple` on each row in the `notices` generator before yielding them.

dimaqq:
Looks good on a cursory examination.

ops/storage.py
```diff
@@ -187,15 +186,15 @@ def drop_notice(self, event_path: str, observer_path: str, method_name: str) ->
...
+    def notices(self, event_path: str | None = None) -> _Notices:
         """Part of the Storage API, return all notices that begin with event_path.
 
         Args:
             event_path: If supplied, will only yield events that match event_path. If not
                 supplied (or None/'') will return all events.
 
         Returns:
-            Iterable of (event_path, observer_path, method_name) tuples
+            List of (event_path, observer_path, method_name) tuples
```
tonyandrewmeyer, line 197: Strictly speaking, this breaks backwards compatibility, doesn't it? For example, if a caller was relying on the performance aspect or on the return value being consumed in a single pass.

I think it's probably fine, since it seems extremely unlikely that anyone is both using this class directly and also doing something where the difference would matter.

However, I think that (a) you should check if jhack uses this, since it seems like the only obvious possible case, and (b) this shouldn't be "chore", it should be "refactor!".

james-garner-canonical: I had in my head that this was ultimately a private change, but I see now that not only is the method public, but the class and the module too. So this really is a backwards-incompatible change -- you can't call `next` on the return value anymore, for example. I probably wouldn't have pushed this PR up if I'd realised this method was completely public.

benhoyt:
Per discussion, we're going to close this to avoid changing the public interface.

PR #2218, opened 2025-12-05
test: switch the integration test charms to use the uv plugin

### LINKED ISSUE #1986: Use uv plugin for integration test
If we change the current charm plugin (without the `pydantic` binary) to uv plugin, our integration tests would be ~13 minutes faster. Even compared to the charm plugin with the `pydantic` binary, there is still a ~3-minute benefit.

We need a mechanism similar to [this](https://github.com/canonical/operator/pull/1966) to generate `pyproject.toml` and `uv.lock` on the fly.

james-garner-canonical:
This PR updates our integration test setup code and the corresponding charm files to use the `uv` plugin.

Resolves #1986

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '42 7 25 * *'
-  pull_request:  # FIXME: don't merge!
 
 permissions: {}
 


tonyandrewmeyer (approved):
Nice, thanks! Looks very clean.

.github/workflows/integration.yaml
```diff
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '42 7 25 * *'
+  pull_request:  # FIXME: don't merge!
```
dimaqq:
```suggestion
```

dimaqq (approved):

PR #2219, opened 2025-12-05
feat: deprecate testing.Context.charm_spec

### LINKED ISSUE #2131: Deprecate or document testing.Context.charm_spec
In #2123 we excluded the `charm_spec` attribute from the `ops.testing.Context because:

> ... although it's public, the type is not, and I don't think we want to encourage its use (I think we would make this private if we could).

I wonder if it's worth deprecating the public `charm_spec` attribute in a separate PR. You convinced me in the other, now-closed PR that it's worth doing this kind of thing (especially in testing) to steer users in the right direction even if we don't have a timeline for the next major release -- and things that we've prepped with a sufficient deprecation warning period can probably be removed then, regardless of why we're doing the major release.

@tonyandrewmeyer said:

> Interesting thought. I can only find two uses:
> * [grafana-agent-operator](https://github.com/canonical/grafana-agent-operator/blob/8ef089753a0e5e58d9ed26becec753fe67db6608/tests/unit/test_cos_agent_e2e.py#L207) - this is using `charm_spec` to get the charm Class and adjust a private attribute on it. I would have thought the test could just import the charm class and do it directly - the `charm_type` isn't a copy, it's the same class object.
> * [charmcraft](https://github.com/canonical/charmcraft/blob/main/tests/spread/commands/init-extensions/tests/unit/test_charm.py) (!) - I guess this is a legitimate use (so perhaps we shouldn't deprecate and maybe should document), pulling the container name from the metadata rather than hard-coding it in the test. The metadata isn't available elsewhere (you can get it from the model, but not until you've already called `run`). However, [from_context()](https://documentation.ubuntu.com/ops/latest/reference/ops-testing/#ops.testing.State.from_context) would do this for you in a nicer way, so maybe the right thing is to encourage that. (The test is using `scenario` as the namespace, so maybe it's quite old?).

We should either plan and go ahead with deprecating this attribute:
- starting with at the very least documenting that it is private
- then moving on to adding a deprecation warning to the public attribute
- and helping the existing users move away.

Alternatively, we should bite the bullet and decide that it's part of the public API and document it as such.

james-garner-canonical:
This PR deprecates accessing `testing.Context.charm_spec`, because the `_CharmSpec` type is private, and was never intended to be exposed publicly.

I suggest we open PRs for the two known users of this attribute before the next release, or even before merging this PR, per @tonyandrewmeyer's research and suggested fixes:
> * [grafana-agent-operator](https://github.com/canonical/grafana-agent-operator/blob/8ef089753a0e5e58d9ed26becec753fe67db6608/tests/unit/test_cos_agent_e2e.py#L207) - this is using `charm_spec` to get the charm Class and adjust a private attribute on it. I would have thought the test could just import the charm class and do it directly - the `charm_type` isn't a copy, it's the same class object.
> * [charmcraft](https://github.com/canonical/charmcraft/blob/main/tests/spread/commands/init-extensions/tests/unit/test_charm.py) (!) - I guess this is a legitimate use (so perhaps we shouldn't deprecate and maybe should document), pulling the container name from the metadata rather than hard-coding it in the test. The metadata isn't available elsewhere (you can get it from the model, but not until you've already called `run`). However, [from_context()](https://documentation.ubuntu.com/ops/latest/reference/ops-testing/#ops.testing.State.from_context) would do this for you in a nicer way, so maybe the right thing is to encourage that. (The test is using `scenario` as the namespace, so maybe it's quite old?).

Resolves #2131

---

Requesting review from @benhoyt, since his instinct was to avoid deprecation noise and publicly document this instead; and from @tonyandrewmeyer since he's the most familiar with this and what charms should be using instead.

### DIFF
--- a/testing/src/scenario/context.py
+++ b/testing/src/scenario/context.py
@@ -741,13 +741,13 @@ class Context(Generic[CharmType]):
     @property
     @deprecated('Use State.from_context to generate a State from charm metadata.', stacklevel=2)
     def charm_spec(self):
-        """Deprecated property for accessing the Context's charm spec.
+        """Deprecated property for accessing the Context's charm specification.
 
-        The _CharmSpec class is private and intended for internal use only.
-        Context.charm_spec will be removed in a future major version.
+        The ``_CharmSpec`` class is private and intended for internal use only.
+        ``Context.charm_spec`` will be removed in a future major version.
 
-        Consider accessing the charm class or metadata directly, or using State.from_context to
-        generate a State from your charm's metadata.
+        Consider accessing the charm class or metadata directly, or using
+        :meth:`State.from_context` to generate a ``State`` from your charm's metadata.
         """
         return self._charm_spec
 


tonyandrewmeyer:
`CharmSpec` has:
 * the charm class - I feel like you should know what this is in your tests and just use it directly rather than via this object
 * the `meta` dict, which is roughly `metadata.yaml`
 * the `actions` dict, which is roughly `actions.yaml`
 * the `config` dict, which is roughly `config.yml`
 * an `autoloaded` Boolean - I feel this is definitely for internal use, and not something you'd use directly in tests.
 * a helper to get all the relations - it seems like no-one is using this, and it does seem like something that would make more sense exposed in a different way (as a general helper function or on the `Context` object) if people did want it

For the metadata, it seems like a `meta` property on `Context` that returned an `ops.CharmMeta` would be a cleaner way to provide this rather than the raw dictionaries. In my earlier (very quick) research it did seem like only charmcraft was actually using this. It does seem reasonable to want this information, but it also seems like in a lot of tests it'd be hard-coded  and that isn't really an issue (and using `from_context()` is an alternative).

So I'm +0 on deprecating (and also opening a PR for grafana-agent-operator - maybe also charmcraft, although I think it's pinned to an old ops-scenario version, so there might be a need for a larger PR that modernises other things).

I'm interested in @PietroPasotti's take here too. Essentially: do you think we should document `Context.charm_spec` or deprecate it?

testing/src/scenario/context.py
```diff
@@ -736,6 +737,12 @@ def __init__(
 
         self.on = CharmEvents()
 
+    @property
+    def charm_spec(self):
+        """Deprecated property for accessing the Context's charm spec."""
+        warnings.warn('Accessing Context.charm_spec is deprecated.', DeprecationWarning)
```
tonyandrewmeyer: It's nice to provide alternatives in these sort of things. Tricky when we don't have a suggestion, which perhaps indicates that it's not ideal to deprecate.

Maybe we could suggest just for the metadata, like "... is deprecated. To generate a State based on the metadata, consider using State.from_context().". And assume people will figure something out for the other attributes.

tonyandrewmeyer (commented):
Assuming that we do deprecate, the code change looks good to me.

I think we should add a test that validates that there's a warning (pytest.warns) if you access `.charm_spec`. We probably don't need other tests, assuming that none of them are using `.charm_spec` directly (if so, we should pytest.warns there to silence the warning, but also see if that's actually a use-case).

dimaqq (approved):

testing/src/scenario/context.py
dimaqq: Aside: I suspect that `stacklevel=2,` is needed here.

benhoyt: It would be nice to make this surface clearly in people's IDEs as deprecated (I'm assuming emitting a deprecation warning at runtime doesn't do this). Either with a fancy type annotation or some other annotation that IDEs understand.

james-garner-canonical:
I've opened PRs to migrate the two users off `Context.charm_spec`.

- https://github.com/canonical/grafana-agent-operator/pull/355
- https://github.com/canonical/charmcraft/pull/2527

I'll update this PR to use [@warnings.deprecated](https://docs.python.org/3/library/warnings.html#warnings.deprecated) and to mention alternatives in the docstring before requesting further review.

testing/tests/test_context.py
```diff
@@ -179,3 +179,9 @@ def _on_start(self, event):
     ctx = Context(GoodCharm, meta={'name': 'crashy'})
     ctx.run(ctx.on.start(), State())
     assert 'TEST_ENV_VAR' not in os.environ
+
+
+def test_charm_spec_is_deprecated():
+    ctx = Context(CharmBase, meta={'name': 'some-name'})
+    with pytest.warns(DeprecationWarning):
+        ctx.charm_spec  # type: ignore
```
james-garner-canonical, line 187: We need the type ignore because it's deprecated -- note that linting would complain if the type ignore was unnecessary. An alternative would be `pyright: ignore[reportDeprecated]`, but we seem to prefer the conciseness of `type: ignore`.

testing/pyproject.toml
```diff
@@ -23,6 +23,7 @@ keywords = ["juju", "test"]
 dependencies = [
     "ops==3.5.0.dev0",
     "PyYAML>=6.0.1",
+    "typing_extensions>=4.9.0",  # for `deprecated` decorator, replace with stdlib `warnings.deprecated` when we're Python 3.13+
```
james-garner-canonical, line 26: `warnings.deprecated` is Python 3.13+. Luckily, `typing_extensions` has backported it, and type checkers respect the backport. This introduces a testing-only runtime dependency on `typing_extensions` (currently it's type checking only in `ops`), which I think is OK.

james-garner-canonical:
Turns out it needs to be [typing_extensions.deprecated](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated) instead since `warnings.deprecated` is Python 3.13+. I've added a test to assert the warning is raised at runtime. The docstring and warning show up nicely in the IDE -- in VS Code, `ctx.charm_spec` is highlighted as an error, and the hover contents show:

Screenshot from 2025-12-15 10-22-02

testing/src/scenario/context.py
```diff
@@ -736,6 +738,19 @@ def __init__(
...
 
+    @property
+    @deprecated('Use State.from_context to generate a State from charm metadata.', stacklevel=2)
+    def charm_spec(self):
+        """Deprecated property for accessing the Context's charm spec.
+
+        The _CharmSpec class is private and intended for internal use only.
+        Context.charm_spec will be removed in a future major version.
+
+        Consider accessing the charm class or metadata directly, or using State.from_context to
```
tonyandrewmeyer: The `State.from_context` can be something like `:attr:State.from_context` (with the extra backticks) I think?

james-garner-canonical: Good call, done (plus some extra backticks around code things too).

PR #2220, opened 2025-12-05
docs: match Charmcraft profiles in "Write and structure charm code"

dwilding:
This PR updates [How to write and structure charm code](https://documentation.ubuntu.com/ops/latest/howto/write-and-structure-charm-code/) to:
- Show the charm files that are generated by the latest machine and Kubernetes profiles
- Show the dependency groups in `pyproject.toml` that are generated by the latest profiles
- Explain about `uv.lock` and recommend `uv add` & `uv remove` for managing deps

**[Preview doc](https://canonical-ubuntu-documentation-library--2220.com.readthedocs.build/ops/2220/howto/write-and-structure-charm-code/)** - See "Add Python dependencies to pyproject.toml and update the lock file" for bulk of the updates

### DIFF
--- a/docs/howto/write-and-structure-charm-code.md
+++ b/docs/howto/write-and-structure-charm-code.md
@@ -134,23 +134,11 @@ Ensure that the `pyproject.toml` *and* the lock file are committed to version
 control, so that exact versions of charms can be reproduced.
 ```
 
-If you prefer to use Poetry to manage your charm's dependencies, specify the {external+charmcraft:ref}`poetry plugin <craft_parts_poetry_plugin>` in `charmcraft.yaml`. In this case, your charm's lock file will be `requirements.txt`. The `poetry` plugin generates `requirements.txt` from `pyproject.toml`.
-
-If your charm uses Charmcraft's `charm` plugin, you'll need to generate `requirements.txt` from `pyproject.toml`. You can do this using a tool such as [pip-compile](https://pip-tools.readthedocs.io/en/latest/).
-
-````{admonition} Best practice
+```{admonition} Best practice
 :class: hint
 
-If your charm uses Charmcraft's `charm` plugin, ensure that you set strict
-dependencies to true. For example:
-
-```yaml
-parts:
-  my-charm:
-    plugin: charm
-    charm-strict-dependencies: true
+Avoid using Charmcraft's `charm` plugin if possible. Instead, {external+charmcraft:ref}`migrate to the uv plugin <howto-migrate-to-uv>` or the {external+charmcraft:ref}`poetry plugin <howto-migrate-to-poetry>`.
 ```
-````
 
 (design-your-python-modules)=
 ## Design your Python modules


docs/howto/write-and-structure-charm-code.md
```diff
@@ -38,12 +38,13 @@ my-charm-set-operators/
 │   │   ├── charmcraft.yaml
 │   │   ├── pyproject.toml
 │   │   ├── README.md
-│   │   ├── requirements.txt
 │   │   ├── src
-│   │   │   └── charm.py
+│   │   │   ├── charm.py
+│   │   │   └── core.py
```
dwilding, line 43: Adding a workload module based on the charm directory name. I should have done this earlier, after the relevant Charmcraft 3.x release.

docs/howto/write-and-structure-charm-code.md
```diff
@@ -73,59 +74,51 @@ version in your `pyproject.toml` so that tooling will detect any use of Python
...
 
 Specify all the direct dependencies of your charm in your `pyproject.toml`
 file in the top-level charm folder. For example:
 
 ```toml
-# Required group: these are all dependencies required to run the charm.
+# All dependencies required to run the charm
 dependencies = [
-    "ops~=2.19",
+    "ops>=3,<4",
```
dimaqq, line 85: **minor:** I'd personally write this as `~=3.0`

docs/howto/write-and-structure-charm-code.md
```diff
@@ -134,23 +127,31 @@ Ensure that tooling is configured to automatically detect new versions,
...
 ```{admonition} Best practice
 :class: hint
 
 Ensure that the `pyproject.toml` *and* the lock file are committed to version
 control, so that exact versions of charms can be reproduced.
 ```
 
+If you prefer to use Poetry to manage your charm's dependencies, specify the {external+charmcraft:ref}`poetry plugin <craft_parts_poetry_plugin>` in `charmcraft.yaml`. In this case, your charm's lock file will be `requirements.txt`. The `poetry` plugin generates `requirements.txt` from `pyproject.toml`.
+
+If your charm uses Charmcraft's `charm` plugin, you'll need to generate `requirements.txt` from `pyproject.toml`. You can do this using a tool such as [pip-compile](https://pip-tools.readthedocs.io/en/latest/).
```
dimaqq: **question:** I wonder if we need this at all. The leadership want us to advertise "the one true way", and folks who are really into those tools are more than capable of figuring out the details. Somehow I feel that this content belongs to a charmhub discourse post, not the core docs. WDYT?

james-garner-canonical: I think it's a good idea to steer charmers towards using dependency constraints in `pyproject.toml` + a lockfile (`requirements.txt`) even if they're stuck with the `charm` plugin, so I'd be happy for this to stay, but up to you guys.

docs/howto/write-and-structure-charm-code.md
james-garner-canonical: I would too, though I can see how `>=3,<4` is more explicit. I suppose we should match whatever the templates have for now.

dwilding: I'll leave this as-is, because we specifically set up the templates this way too

docs/howto/write-and-structure-charm-code.md
dwilding: I'll raise this for a quick discussion at daily (and record the result here). There's a best practice about the `charm` plugin that we probably don't want to lose, as it's automatically pulled into the listing review template.

dwilding: We decided to steer charmers away from the `charm` plugin. We also decided to acknowledge the poetry plugin without replicating details from the Charmcraft docs.

I've replaced the discussion of `poetry` and `charm` plugins by this:

> **Best practice**
> Avoid using Charmcraft's `charm` plugin if possible. Instead, [migrate to the uv plugin](https://documentation.ubuntu.com/charmcraft/latest/howto/migrate-plugins/charm-to-uv/#howto-migrate-to-uv) or the [poetry plugin](https://documentation.ubuntu.com/charmcraft/latest/howto/migrate-plugins/charm-to-poetry/#howto-migrate-to-poetry).

FYI @tonyandrewmeyer

PR #2222, opened 2025-12-09
docs: use base 24.04 for httpbin-demo charm

dwilding:
This PR bumps the base of the [httpbin-demo](https://github.com/canonical/operator/tree/main/examples/httpbin-demo) charm from 22.04 to 24.04. (I'm classifying this as docs, as I consider our example charms loosely part of the docs)

I've opened https://github.com/canonical/charmcraft/pull/2522 to make the same change in the `kubernetes` and `machine` profiles. So I'll leave bumping our other example charms to 24.04 until that has been released.

### DIFF
--- a/examples/httpbin-demo/charmcraft.yaml
+++ b/examples/httpbin-demo/charmcraft.yaml
@@ -1,5 +1,8 @@
 type: charm
 base: ubuntu@24.04
+platforms:
+  amd64:
+  arm64:
 
 name: httpbin-demo
 title: |


examples/httpbin-demo/charmcraft.yaml
```diff
@@ -2,10 +2,10 @@ type: charm
 bases:
   - build-on:
```
dwilding: Is it worth separating `build-on` and `run-on`? The Charmcraft profile doesn't do this ([see here](https://github.com/canonical/charmcraft-profile-tools/blob/main/kubernetes/charmcraft.yaml#L18))

benhoyt: No, probably not.

benhoyt (approved):

dwilding:
Interesting - I see this error from our CI that tries to pack the charm:

> Not valid for use with the 'bases' key: 'ubuntu@24.04'
> Recommended resolution: Use the 'platforms' keyword in order to use newer bases.

I'll investigate then re-request review

dwilding:
TIL `bases` is deprecated. I've switched to `base` and `platforms` as in the Charmcraft profiles. Ready for review now @tonyandrewmeyer

tonyandrewmeyer (approved):
Thanks!

PR #2223, opened 2025-12-09
docs: clarify parts of the machine charm tutorial

dwilding:
Based on feedback we received about the new [machine charm tutorial](https://documentation.ubuntu.com/ops/latest/tutorial/write-your-first-machine-charm/), I'm making a few clarifications:

- Add a link for context about what a charm is
- Emphasise that the code is on GitHub, and mention this in a couple more places
- Emphasise that the charm class is newly instantiated on every event, and what that means for storing data in the charm class. I'm less sure this addition is needed - happy to revise

**[Preview doc](https://canonical-ubuntu-documentation-library--2223.com.readthedocs.build/ops/2223/tutorial/write-your-first-machine-charm/)**

### DIFF
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -342,10 +342,10 @@ In `src/charm.py`, replace the `_on_install` method of the charm class with:
 
 When your charm receives the "install" event from Juju, Ops runs this method and tells Juju the version of tinyproxy that's installed on the machine. Juju shows the version in its status output.
 
-As you write your charm, keep in mind that the charm code only runs when there's an event to handle. Behind the scenes, Juju runs `charm.py` with event data in the environment. Ops reads the event data, instantiates the charm class, then runs the appropriate method.
+As you write your charm, keep in mind that the charm code only runs when there's an event to handle.
 
 ```{important}
-The charm class will be newly instantiated on every event. In general, when you write a charm, you can't persist data between events by storing the data in the charm class.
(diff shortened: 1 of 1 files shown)


docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -340,6 +344,10 @@ When your charm receives the "install" event from Juju, Ops runs this method and
 
 As you write your charm, keep in mind that the charm code only runs when there's an event to handle. Behind the scenes, Juju runs `charm.py` with event data in the environment. Ops reads the event data, instantiates the charm class, then runs the appropriate method.
 
+```{important}
+The charm class will be newly instantiated on every event. In general, when you write a charm, you can't persist data between events by storing the data in the charm class.
```
benhoyt:
```suggestion
The charm class will be newly instantiated on every event. In general, when you write a charm, you can't persist data between events by storing the data in the charm instance.
```

james-garner-canonical: Maybe we should go even further here, as this makes it sound a bit like you could persist information in a global or similar.

```suggestion
Juju will execute your `charm.py` file on every event, and your call to `ops.main` creates a fresh instance of your charm class. You can't persist data between Juju events by storing it in memory.
```

dwilding: Good suggestions, thanks. I've combined James's suggestion with some of the preceding paragraph, to make the info more concise overall. It now looks like:

> As you write your charm, keep in mind that the charm code only runs when there's an event to handle.
>
> **Important**
> Juju executes `charm.py` on every event, with event data in the environment. Your call to `ops.main` creates a fresh instance of the charm class, then Ops runs the appropriate method on the charm class. You can't persist data between Juju events by storing it in memory.

PR #2225, opened 2025-12-10
ci: adjustments to the AI agent instructions

tonyandrewmeyer:
A small tweak to AGENTS.md to include suggestions on commits.

A copilot-instructions.md file that is solely focused on code review (Copilot's code review will not pick up AGENTS.md) and directs the agent to AGENTS.md in other cases. I think this should work, but it's difficult to tell without it being actually in place and used (I can put it on a fork or personal repo, but unless there's reviewing and developing going on, it's difficult to know how well it will work).

[Subagents](https://code.claude.com/docs/en/sub-agents), or [custom agents](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/coding-agent/create-custom-agents):
* Replaced the test agent with one that seems to work better and is more aligned with what is recommended.
* Added a (very basic) lint and doc agent, which are the other two that are most commonly recommended.
* Added symlinks so that Claude and Copilot can share the agent definitions.

[Claude skills](https://claude.com/skills):

* A `tmux` skill that Armin Ronacher developed. This is generally useful for running multiple commands at once, and I've found reasonably useful when working on Charm Tech issues.
* A `juju` skill. I find that Claude and Copilot struggle to actually use Juju, but it's very useful to be able to really test out changes against a real Juju. This is my initial attempt at helping the agents with that.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,6 +1,6 @@
 # Copilot Instructions for Ops Library Code Review
 
-These instructions are specifically for GitHub Copilot when reviewing code in the `ops` library repository. For more general instructions ignore this file and refer to [AGENTS.md](AGENTS.md).
+These instructions are specifically for GitHub Copilot when reviewing code in the `ops` library repository. For more general instructions ignore this file and refer to [AGENTS.md](../AGENTS.md).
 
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
Discarding this for now, will likely push some changes to copilot-collections, with a more data oriented approach, and move this repo to use that if we go that way.

PR #2226, opened 2025-12-11
chore: update the list of published charms

tonyandrewmeyer:
Update after running the update script, adjusting the monorepo additions.

[Run in my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/20121233031).

There are an increasing number of failures due to Python 3.8 / Ops 2.x requirements, but it's better that we have the known failing list rather than wait to add the extra ones.

### DIFF
--- a/.github/update-published-charms-tests-workflow.py
+++ b/.github/update-published-charms-tests-workflow.py
@@ -67,6 +67,8 @@ SKIP = {
     'charm-sysconfig',
     # A bundle, not a charm.
     'cos-lite-bundle',
+    # Source is not public.
+    'charm-weebl',
 }
 CHARM_ROOTS = {
     'argo-operators': ['charms/argo-controller'],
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
> Is this the process for updating the list? Should we run the published charm tests with the new charms and see if any need to be excluded, or added to `CHARM_ROOTS` (e.g. `katib-operators` and `kfp-operators` are named as if they're monorepos)?

Ah, I'd forgotten about the monorepo support. I guess that does need manual checking, which is annoying - I'd much rather this was totally automatic.

I'll run the tests to see what the results are.

james-garner-canonical:
> > Is this the process for updating the list? Should we run the published charm tests with the new charms and see if any need to be excluded, or added to `CHARM_ROOTS` (e.g. `katib-operators` and `kfp-operators` are named as if they're monorepos)?
> 
> Ah, I'd forgotten about the monorepo support. I guess that does need manual checking, which is annoying - I'd much rather this was totally automatic.

I guess if we standardise some charm monorepo metadata this could be automated easily. Or I guess we could probe for common monorepo layouts. Either way would require downloading the repos, which adds some overhead to the script. I wonder if it's worth creating an issue to track this, or wait and see how charm monorepo processes develop first.

dimaqq (approved):

PR #2228, opened 2025-12-11
chore: solve and remove the testing/tests linting ignores

tonyandrewmeyer:
Removes all linting exceptions from `testing/tests`, fixing the underlying issues.

Changes:
  - Added copyright notices: Added missing copyright headers to all test files
  - Fixed import ordering: Reorganised imports (standard library → third party → first party)
  - Improved type annotations:
    - Removed unnecessary quotes from type annotations
    - Added ClassVar[] annotations for mutable class attributes
  - Enhanced code safety:
    - Replaced blind exception assertions with specific exception types
    - Fixed useless attribute access (properly using noqa: B018 where intentional)
    - Used context managers for file operations to ensure timely cleanup
  - Code style improvements:
    - Fixed line length violations
    - Used next(iter()) instead of indexing with [0] for iterables
    - Changed hardcoded /tmp paths to more generic alternatives
    - Maintained CamelCase for class names

Each commit addresses (and removes) a single exception, so reviewing commit-by-commit may be the simplest approach.

### DIFF
--- a/testing/tests/test_e2e/test_deferred.py
+++ b/testing/tests/test_e2e/test_deferred.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import typing
+from collections.abc import Mapping
 
 import pytest
 from scenario import Context
@@ -20,7 +21,7 @@ CHARM_CALLED = 0
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(ops.CharmBase):
-        META: typing.ClassVar[dict[str, typing.Any]] = {
+        META: Mapping[str, typing.Any] = {
             'name': 'mycharm',
             'requires': {'foo': {'interface': 'bar'}},
             'containers': {'foo': {'type': 'oci-image'}},
--- a/testing/tests/test_e2e/test_event.py
+++ b/testing/tests/test_e2e/test_event.py
@@ -3,8 +3,6 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
-
 import pytest
 from scenario import Context
 from scenario.state import State, _CharmSpec, _Event, _EventType
@@ -65,10 +63,7 @@ def test_event_type(evt, expected_type):
 
 
 def test_emitted_framework():
-    class MyCharm(ops.CharmBase):
-        META: ClassVar[dict[str, Any]] = {'name': 'joop'}
-
-    ctx = Context(MyCharm, meta=MyCharm.META, capture_framework_events=True)
+    ctx = Context(ops.CharmBase, meta={'name': 'joop'}, capture_framework_events=True)
     ctx.run(ctx.on.update_status(), State())
     assert len(ctx.emitted_events) == 4
     assert list(map(type, ctx.emitted_events)) == [
@@ -81,8 +76,6 @@ def test_emitted_framework():
 
 def test_emitted_deferred():
     class MyCharm(ops.CharmBase):
-        META: ClassVar[dict[str, Any]] = {'name': 'joop'}
-
         def __init__(self, framework: ops.Framework):
             super().__init__(framework)
             framework.observe(self.on.update_status, self._on_update_status)
@@ -92,7 +85,7 @@ def test_emitted_deferred():
 
     ctx = Context(
         MyCharm,
-        meta=MyCharm.META,
+        meta={'name': 'joop'},
         capture_deferred_events=True,
         capture_framework_events=True,
     )
--- a/testing/tests/test_e2e/test_juju_log.py
+++ b/testing/tests/test_e2e/test_juju_log.py
@@ -4,7 +4,8 @@
 from __future__ import annotations
 
 import logging
-from typing import Any, ClassVar
+from collections.abc import Mapping
+from typing import Any
 
 import pytest
 from scenario import Context
@@ -18,7 +19,7 @@ logger = logging.getLogger('testing logger')
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(CharmBase):
-        META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
+        META: Mapping[str, Any] = {'name': 'mycharm'}
 
         def __init__(self, framework):
             super().__init__(framework)
--- a/testing/tests/test_e2e/test_manager.py
+++ b/testing/tests/test_e2e/test_manager.py
@@ -3,7 +3,8 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
+from collections.abc import Mapping
+from typing import Any
 
 import pytest
 from scenario import Context, State
@@ -16,8 +17,8 @@ from ops.charm import CharmBase, CollectStatusEvent
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(CharmBase):
-        META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
-        ACTIONS: ClassVar[dict[str, Any]] = {'do-x': {}}
+        META: Mapping[str, Any] = {'name': 'mycharm'}
+        ACTIONS: Mapping[str, Any] = {'do-x': {}}
 
         def __init__(self, framework):
             super().__init__(framework)
--- a/testing/tests/test_e2e/test_ports.py
+++ b/testing/tests/test_e2e/test_ports.py
@@ -3,7 +3,8 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
+from collections.abc import Mapping
+from typing import Any
 
 import pytest
 from scenario import Context, State
@@ -13,7 +14,7 @@ from ops import CharmBase, Framework, StartEvent, StopEvent
 
 
 class MyCharm(CharmBase):
-    META: ClassVar[dict[str, Any]] = {'name': 'edgar'}
+    META: Mapping[str, Any] = {'name': 'edgar'}
 
     def __init__(self, framework: Framework):
         super().__init__(framework)
--- a/testing/tests/test_e2e/test_storage.py
+++ b/testing/tests/test_e2e/test_storage.py
@@ -3,7 +3,8 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
+from collections.abc import Mapping
+from typing import Any
 
 import pytest
 from scenario import Context, State, Storage
@@ -12,14 +13,14 @@ from ops import CharmBase, ModelError
 
 
 class MyCharmWithStorage(CharmBase):
-    META: ClassVar[dict[str, Any]] = {
+    META: Mapping[str, Any] = {
         'name': 'charlene',
         'storage': {'foo': {'type': 'filesystem'}},
     }
 
 
 class MyCharmWithoutStorage(CharmBase):
-    META: ClassVar[dict[str, Any]] = {'name': 'patrick'}
+    META: Mapping[str, Any] = {'name': 'patrick'}
 
 
 @pytest.fixture
--- a/testing/tests/test_e2e/test_stored_state.py
+++ b/testing/tests/test_e2e/test_stored_state.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from typing import Any, ClassVar
 
 import pytest
@@ -16,7 +17,7 @@ from tests.helpers import trigger
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(ops.CharmBase):
-        META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
+        META: Mapping[str, Any] = {'name': 'mycharm'}
 
         _read: ClassVar[dict[str, Any]] = {}
         _stored = OpsStoredstate()
--- a/testing/tests/test_e2e/test_vroot.py
+++ b/testing/tests/test_e2e/test_vroot.py
@@ -4,8 +4,9 @@
 from __future__ import annotations
 
 import tempfile
+from collections.abc import Mapping
 from pathlib import Path
-from typing import Any, ClassVar
+from typing import Any
 
 import pytest
 import yaml
@@ -19,7 +20,7 @@ from ..helpers import trigger
 
 
 class MyCharm(CharmBase):
-    META: ClassVar[dict[str, Any]] = {'name': 'my-charm'}
+    META: Mapping[str, Any] = {'name': 'my-charm'}
 
     def __init__(self, framework: Framework):
         super().__init__(framework)
--- a/testing/tests/test_emitted_events_util.py
+++ b/testing/tests/test_emitted_events_util.py
@@ -3,7 +3,8 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
+from collections.abc import Mapping
+from typing import Any
 
 from scenario import Context, State
 from scenario.state import _Event
@@ -21,7 +22,7 @@ class MyCharmEvents(CharmEvents):
 
 
 class MyCharm(CharmBase):
-    META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
+    META: Mapping[str, Any] = {'name': 'mycharm'}
     on = MyCharmEvents()
 
     def __init__(self, *args, **kwargs):


testing/tests/helpers.py
```diff
@@ -42,10 +44,10 @@ def trigger(
     if isinstance(event, str):
         if event.startswith('relation_'):
             assert len(tuple(state.relations)) == 1, 'shortcut only works with one relation'
-            event = getattr(ctx.on, event)(tuple(state.relations)[0])
+            event = getattr(ctx.on, event)(next(iter(state.relations)))
         elif event.startswith('pebble_'):
             assert len(tuple(state.containers)) == 1, 'shortcut only works with one container'
-            event = getattr(ctx.on, event)(tuple(state.containers)[0])
+            event = getattr(ctx.on, event)(next(iter(state.containers)))
```
james-garner-canonical, line 50: These are improvements on their own, but feel weird given that we still call `tuple(...)` on the preceding line. Also unfortunate that we need to do `len(tuple(...` instead of just `len(` given that these are actually going to be `frozenset`.

I wonder if it would be worth adding custom `__init__` methods to the relevant classes (just `State`?) to correct the mismatch between the `relations` arg only needing to be `Iterable`, but the `relations` attribute actually always being a `frozenset` -- should I create an issue for this? Or discuss at daily?

In the meantime, if we're open to making some manual improvements in this PR, how about this:
```py
        if event.startswith('relation_'):
            (relation,) = state.relations  # shortcut only works with one relation
            event = getattr(ctx.on, event)(relation)
        elif event.startswith('pebble_'):
            (container,) = state.containers  # shortcut only works with one container
            event = getattr(ctx.on, event)(container)
```
But feel free to reject to minimise the diff.

testing/tests/test_e2e/test_deferred.py
```diff
@@ -16,13 +20,13 @@
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(ops.CharmBase):
-        META: dict[str, typing.Any] = {
+        META: typing.ClassVar[dict[str, typing.Any]] = {
```
james-garner-canonical: This is the 'correct' type annotation, but `Mapping` would feel cleaner to me if we don't mutate `META` in tests.
```suggestion
        META: typing.Mapping[str, typing.Any] = {
```

testing/tests/test_e2e/test_event.py
```diff
@@ -62,7 +67,7 @@ class MyCharm(CharmBase):
 
 def test_emitted_framework():
     class MyCharm(CharmBase):
-        META = {'name': 'joop'}
+        META: ClassVar[dict[str, Any]] = {'name': 'joop'}
 
     ctx = Context(MyCharm, meta=MyCharm.META, capture_framework_events=True)
```
james-garner-canonical: In the cases where we define the charm class in the test body, rather than in a fixture, it seems a lot cleaner to lift `META` out of the class entirely.
```suggestion

    ctx = Context(MyCharm, meta={'name': 'joop'}, capture_framework_events=True)
```

testing/tests/test_e2e/test_event.py
```diff
@@ -77,7 +82,7 @@ class MyCharm(CharmBase):
 
 def test_emitted_deferred():
     class MyCharm(CharmBase):
-        META = {'name': 'joop'}
+        META: ClassVar[dict[str, Any]] = {'name': 'joop'}
```
james-garner-canonical: This can also be lifted out of the class, but I can't make a `suggestion` because the in-lining site is too many lines down.

testing/tests/test_e2e/test_juju_log.py
```diff
@@ -1,20 +1,24 @@
...
+from ops.charm import CharmBase, CollectStatusEvent
+
 logger = logging.getLogger('testing logger')
 
 
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(CharmBase):
-        META = {'name': 'mycharm'}
+        META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
```
james-garner-canonical:
```suggestion
        META: Mapping[str, Any] = {'name': 'mycharm'}
```

testing/tests/test_e2e/test_manager.py
```diff
@@ -1,18 +1,23 @@
...
+from ops.charm import CharmBase, CollectStatusEvent
+
 
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(CharmBase):
-        META = {'name': 'mycharm'}
-        ACTIONS = {'do-x': {}}
+        META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
+        ACTIONS: ClassVar[dict[str, Any]] = {'do-x': {}}
```
james-garner-canonical:
```suggestion
        META: Mapping[str, Any] = {'name': 'mycharm'}
        ACTIONS: Mapping[str, Any] = {'do-x': {}}
```

testing/tests/test_e2e/test_pebble.py
```diff
@@ -75,30 +79,31 @@ def callback(self: CharmBase):
 
 def test_fs_push(charm_cls):
     text = 'lorem ipsum/n alles amat gloriae foo'
-    file = tempfile.NamedTemporaryFile()
-    pth = Path(file.name)
-    pth.write_text(text)
```
james-garner-canonical, line 80: WDYT about using the [tmp_path](https://docs.pytest.org/en/stable/how-to/tmp_path.html) fixture here?

```py
pth = tmp_path / 'textfile'
pth.write_text(text)
```

testing/tests/test_e2e/test_pebble.py
```diff
@@ -120,45 +125,43 @@ def callback(self: CharmBase):
             with pytest.raises((FileNotFoundError, pebble.PathError)):
                 container.pull('/foo/bar/baz.txt')
 
-    td = tempfile.TemporaryDirectory()
```
james-garner-canonical, line 123: Same here -- `tmp_path` would be equivalent to `td`, which I think would decrease the diff here.

testing/tests/test_e2e/test_ports.py
```diff
@@ -1,14 +1,19 @@
...
+import pytest
 from scenario import Context, State
 from scenario.state import Port, StateValidationError, TCPPort, UDPPort
 
+from ops import CharmBase, Framework, StartEvent, StopEvent
+
 
 class MyCharm(CharmBase):
-    META = {'name': 'edgar'}
+    META: ClassVar[dict[str, Any]] = {'name': 'edgar'}
```
james-garner-canonical:
```suggestion
    META: Mapping[str, Any] = {'name': 'edgar'}
```

testing/tests/test_e2e/test_resource.py
```diff
@@ -24,7 +24,7 @@ def test_get_resource():
             'resources': {'foo': {'type': 'file'}, 'bar': {'type': 'file'}},
         },
     )
-    resource1 = Resource(name='foo', path=pathlib.Path('/tmp/foo'))
+    resource1 = Resource(name='foo', path=pathlib.Path('/var/charm/foo'))
```
james-garner-canonical, line 27: Why this change?

testing/tests/test_e2e/test_storage.py
```diff
@@ -1,17 +1,25 @@
...
 
+import pytest
 from scenario import Context, State, Storage
 
+from ops import CharmBase, ModelError
+
 
 class MyCharmWithStorage(CharmBase):
-    META = {'name': 'charlene', 'storage': {'foo': {'type': 'filesystem'}}}
+    META: ClassVar[dict[str, Any]] = {
```
james-garner-canonical:
```suggestion
    META: Mapping[str, Any] = {
```

testing/tests/test_e2e/test_storage.py
```diff
@@ -1,17 +1,25 @@
...
+        'name': 'charlene',
+        'storage': {'foo': {'type': 'filesystem'}},
+    }
 
 
 class MyCharmWithoutStorage(CharmBase):
-    META = {'name': 'patrick'}
+    META: ClassVar[dict[str, Any]] = {'name': 'patrick'}
```
james-garner-canonical:
```suggestion
    META: Mapping[str, Any] = {'name': 'patrick'}
```

testing/tests/test_e2e/test_stored_state.py
```diff
@@ -1,22 +1,26 @@
...
-from scenario.state import State, StoredState
+from ops.framework import StoredState as OpsStoredstate
 from tests.helpers import trigger
 
 
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(ops.CharmBase):
-        META = {'name': 'mycharm'}
+        META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
```
james-garner-canonical:
```suggestion
        META: Mapping[str, Any] = {'name': 'mycharm'}
```

testing/tests/test_e2e/test_vroot.py
```diff
@@ -1,20 +1,25 @@
...
 from ops.framework import Framework
 from ops.model import ActiveStatus
 
-from scenario import Context, State
 from ..helpers import trigger
 
 
 class MyCharm(CharmBase):
-    META = {'name': 'my-charm'}
+    META: ClassVar[dict[str, Any]] = {'name': 'my-charm'}
```
james-garner-canonical:
```suggestion
    META: Mapping[str, Any] = {'name': 'my-charm'}
```

testing/tests/test_emitted_events_util.py
```diff
@@ -16,7 +21,7 @@ class MyCharmEvents(CharmEvents):
 
 
 class MyCharm(CharmBase):
-    META = {'name': 'mycharm'}
+    META: ClassVar[dict[str, Any]] = {'name': 'mycharm'}
```
james-garner-canonical:
```suggestion
    META: Mapping[str, Any] = {'name': 'mycharm'}
```

testing/tests/test_emitted_events_util.py
```diff
@@ -1,11 +1,16 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
-from ops.charm import CharmBase, CharmEvents, CollectStatusEvent, StartEvent
-from ops.framework import CommitEvent, EventBase, EventSource, PreCommitEvent
+from typing import Any, ClassVar
```
james-garner-canonical:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_vroot.py
```diff
@@ -1,20 +1,25 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
 import tempfile
 from pathlib import Path
+from typing import Any, ClassVar
```
james-garner-canonical:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_stored_state.py
```diff
@@ -1,22 +1,26 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
+from typing import Any, ClassVar
```
james-garner-canonical, line 7:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_storage.py
```diff
@@ -1,17 +1,25 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
-import pytest
-from ops import CharmBase, ModelError
+from typing import Any, ClassVar
```
james-garner-canonical:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_ports.py
```diff
@@ -1,14 +1,19 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
-import pytest
-from ops import CharmBase, Framework, StartEvent, StopEvent
+from typing import Any, ClassVar
```
james-garner-canonical:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_manager.py
```diff
@@ -1,18 +1,23 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
-import pytest
-from ops import ActiveStatus
-from ops.charm import CharmBase, CollectStatusEvent
+from typing import Any, ClassVar
```
james-garner-canonical:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_juju_log.py
```diff
@@ -1,20 +1,24 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 from __future__ import annotations
 
 import logging
+from typing import Any, ClassVar
```
james-garner-canonical:
```suggestion
from typing import Any, ClassVar, Mapping
```

testing/tests/test_e2e/test_resource.py
tonyandrewmeyer: ruff doesn't like hard-coded paths to `/tmp`. Normally, the change would be to use `tempfile.TemporaryFile` (or similar) but in this case the path isn't actually used, it's just checking that the path gets passed through to Ops.

It felt like a cleaner change than putting in a `noqa`, and we don't want to actually load `tempfile.get_tempdir()` I think.

james-garner-canonical: Makes sense, thanks for explaining.

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer: Good call, thanks!

testing/tests/test_emitted_events_util.py
tonyandrewmeyer: Needs to be from collections.abc, but yes, thanks.

testing/tests/test_emitted_events_util.py
tonyandrewmeyer: Fair point with these. I was thinking that some test or subclass might alter the meta, but it shouldn't really be done like that and certainly isn't now, so Mapping is cleaner indeed.

testing/tests/test_emitted_events_util.py
james-garner-canonical: Oops, guess I shouldn't have overzealously tried to streamline things with all those `suggestions` :smile:

tonyandrewmeyer:
It was a kind thought :)

testing/tests/helpers.py
tonyandrewmeyer: > I wonder if it would be worth adding custom `__init__` methods to the relevant classes (just `State`?) to correct the mismatch between the `relations` arg only needing to be `Iterable`, but the `relations` attribute actually always being a `frozenset` -- should I create an issue for this? Or discuss at daily?

Yeah, this has come up a couple of times before.

We did actually have a version where they all had an `__init__`. The primary reason was to support keyword-only arguments in Python 3.8, and we eventually decided that having in one place (the thing with the custom `__new__`) was superior and ripped them all out. But they did also nicely handle the "you can give me an X but the attribute will always be a Y" case, so we lost that at the same time. Obviously, we have since ripped out the custom solution.

I'm definitely not opposed to going back to `__init__`. I think if we fix it for one it would be best to fix it for them all. Looking through, that's:

* CloudCredential (Mapping->dict, Sequence->list, really not sure why that is not tuple!)
* CloudSpec(Sequence->list, really not sure why that is not tuple!)
* Secret(Mapping->RawSecretRevisionContents, could probably just be dict; Mapping->dict)
* BindAddress(Sequence->tuple, at the moment stays a list if passed a list)

And probably lots more (I wouldn't be surprised if it's like 80% of the classes). However, it could also potentially get rid of or incorporate the code we have for doing the copy (for somewhat immutability).

Feel free to open a PR or add to daily.

dimaqq (approved):

Issue #1785, opened 2025-06-04
Type-check the Scenario tests


tonyandrewmeyer:
We run pyright over the ops tests, but not the Scenario ones, which means we miss some issues, because the tests are our biggest set of code exercising the functionality.

We should add `testing/tests/*.py` and `testing/tests/test_e2e/*.py` to the pyright list, and also fix all the issues it finds:

```
1696 errors, 0 warnings, 0 informations 
static: exit 1 (31.27 seconds) /home/tameyer/code/operator> pyright pid=3163306
```

(I'm sure that a lot of those 1696 are duplicates).

tonyandrewmeyer:
Unassigning myself - I used this as a test-case in the Copilot workshop, but it turned out (see linked closed PR) that it was a poor choice, so that PR was rejected. I have no intention to work on this at the moment, so anyone else should feel free to pick it up.

Note that the conclusion in the rejected PR is useful: this change should be implemented as a series of small PRs rather than one large one, and each PR should finish enough work that it can be continually validated going forward (either fixing a subset of rules, or fixing a subset of files).

tonyandrewmeyer:
Done in the linked PRs.

PR #2230, opened 2025-12-12

tonyandrewmeyer:
`testing/tests` is added to the files that pyright should check, and the files that are left to do are added to the exclusion list.

Add type annotations to the test files (all of the ones outside of test_e2e, which are left for another PR).

To service this, add a type annotation to `Runtime.exec`.

### DIFF
--- a/testing/tests/test_context_on.py
+++ b/testing/tests/test_context_on.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 import copy
 import datetime
 import typing
-from collections.abc import Mapping
 from typing import Any, Literal
 
 import pytest
@@ -14,7 +13,7 @@ import scenario
 
 import ops
 
-META: Mapping[str, Any] = {
+META: dict[str, Any] = {
     'name': 'context-charm',
     'containers': {
         'bar': {},


tonyandrewmeyer:
This moves #1785 closer to completion but there are still many files to go (probably at least 2 more PRs, to keep the size manageable).

testing/tests/helpers.py
```diff
@@ -73,9 +71,6 @@ def jsonpatch_delta(self, other: 'State'):
...
         dict_other[attr] = [dataclasses.asdict(o) for o in dict_other[attr]]
         dict_self[attr] = [dataclasses.asdict(o) for o in dict_self[attr]]
-    patch = jsonpatch.make_patch(dict_other, dict_self).patch
-    return sort_patch(patch)
-
-
-def sort_patch(patch: list[dict], key=lambda obj: obj['path'] + obj['op']):
-    return sorted(patch, key=key)
+    patch = jsonpatch.make_patch(dict_other, dict_self).patch  # type: ignore
+    patch = cast('list[dict[str, Any]]', patch)
```
james-garner-canonical: Does this work?
```suggestion
    patch: list[dict[str, Any]] = jsonpatch.make_patch(dict_other, dict_self).patch  # type: ignore
```

testing/tests/test_charm_spec_autoload.py
```diff
@@ -70,43 +73,43 @@ def create_tempcharm(
...
             (root / 'charmcraft.yaml').write_text(yaml.safe_dump(unified_meta))
 
-    with import_name(name, charmpy) as charm:
-        yield charm
+    with import_name(name, charmpy) as charm_class:
+        yield charm_class
 
 
-def test_autoload_no_meta_fails(tmp_path):
+def test_autoload_no_meta_fails(tmp_path: Path) -> None:
```
james-garner-canonical: Didn't we settle on not doing `-> None` for all our tests?

testing/tests/test_context.py
```diff
@@ -100,27 +101,29 @@ def test_context_manager():
...
 
 @pytest.mark.parametrize('bare_charm_errors', ('1', '0'))
-def test_context_manager_uncaught_error(bare_charm_errors: str, monkeypatch: pytest.Monkeypatch):
-    class CrashyCharm(CharmBase):
-        def __init__(self, framework):
+def test_context_manager_uncaught_error(
+    bare_charm_errors: str, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    class CrashyCharm(ops.CharmBase):
+        def __init__(self, framework: ops.Framework) -> None:
```
james-garner-canonical: And didn't we settle on no `-> None` for `__init__`?

testing/tests/test_context_on.py
```diff
@@ -250,14 +261,15 @@ def test_relation_app_events(as_kwarg, event_name, event_kind):
         mgr.run()
         relation_event, collect_status = mgr.charm.observed
         assert isinstance(relation_event, event_kind)
+        relation_event = typing.cast('ops.RelationEvent', relation_event)
```
james-garner-canonical: How about annotating `event_kind` as `type[ops.RelationEvent]` instead, would that eliminate the need for this `cast`?

testing/tests/test_context_on.py
```diff
@@ -498,19 +517,19 @@ def __init__(self, framework: ops.Framework):
         framework.observe(self.consumer.on.state_event, self._on_event)
 
 
-def test_custom_event_no_args():
+def test_custom_event_no_args() -> None:
     ctx = scenario.Context(CustomCharm, meta=META, actions=ACTIONS)
-    with ctx(ctx.on.custom(MyConsumer.on.foo_started), scenario.State()) as mgr:
+    with ctx(ctx.on.custom(MyConsumer.on.foo_started), scenario.State()) as mgr:  # type: ignore
```
james-garner-canonical, line 516: These `type: ignore` comments seem unfortunate given how idiomatic this code looks -- why are they needed? For `on.foo_started` attribute access? I'm wondering here if there's some user facing typing stuff we should be noting down for future improvement, the `type: ignore` seems like a fine solution in this PR.

testing/tests/test_e2e/test_state.py
```diff
@@ -137,12 +137,16 @@ def call(charm: CharmBase, e):
...
-        {'op': 'replace', 'path': '/unit_status/name', 'value': 'active'},
-    ])
+    expected = sorted(
+        [
+            {'op': 'replace', 'path': '/app_status/message', 'value': 'foo barz'},
+            {'op': 'replace', 'path': '/app_status/name', 'value': 'waiting'},
+            {'op': 'replace', 'path': '/unit_status/message', 'value': 'foo test'},
+            {'op': 'replace', 'path': '/unit_status/name', 'value': 'active'},
+        ],
+        key=lambda obj: obj['path'] + obj['op'],
```
james-garner-canonical: Eliminating `sort_patch` earlier seemed like a nice cleanup, but do we want to be repeating this lambda? I guess it's OK if this is the only case, but otherwise maybe we want to factor it back into a `patch_key` function or something.

testing/tests/test_emitted_events_util.py
```diff
@@ -17,21 +19,21 @@ class MyCharmEvents(CharmEvents):
 
 class MyCharm(CharmBase):
     META = {'name': 'mycharm'}
-    on = MyCharmEvents()
+    on = MyCharmEvents()  # type: ignore[assignment]
```
james-garner-canonical: Does `pyright` actually respect this kind of ignore scope? I think it might ignore them entirely and treat all `type: ignore` comments as total.

testing/tests/test_emitted_events_util.py
```diff
@@ -17,21 +19,21 @@ class MyCharmEvents(CharmEvents):
...
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
james-garner-canonical: More idiomatic?
```suggestion
    def __init__(self, framework: ops.Framework) -> None:
```
Need to import `ops` or `Framework` though.
Feel free to reject as out of scope.

testing/tests/test_runtime.py
```diff
@@ -46,74 +48,80 @@ def test_event_emission():
...
         runtime = Runtime(
             'foo',
-            _CharmSpec(
-                my_charm_type,
-                meta=meta,
-            ),
+            charm_spec,
         )
 
+        ctx: Context[ops.CharmBase] = Context(my_charm_type, meta=meta)
```
james-garner-canonical: Is the type annotation needed here, with `my_charm_type` typed as `type[ops.CharmBase]` via `charm_type()`?

testing/tests/test_runtime.py
```diff
@@ -46,74 +48,80 @@ def test_event_emission():
...
             event=_Event('bar'),
-            context=Context(my_charm_type, meta=meta),
+            context=ctx,
         ) as manager:
             manager.run()
 
-        assert my_charm_type._event
-        assert isinstance(my_charm_type._event, MyEvt)
+        assert my_charm_type._event  # type: ignore[attr-defined]
+        assert isinstance(my_charm_type._event, MyEvt)  # type: ignore[attr-defined]
```
james-garner-canonical: Additional scoped ignores I'm curious about.

testing/tests/test_charm_spec_autoload.py
tonyandrewmeyer: Ah, I thought we settled on it's not *required* for tests. Maybe I'm remembering that incorrectly and it was "don't do it for tests". The STYLE.md section (which I wrote) is ambiguous on this :(

testing/tests/test_context.py
tonyandrewmeyer: Ah, this one was definitely "skip it", yes. Overzealous tab-acceptance on my part, will fix.

testing/tests/test_context_on.py
tonyandrewmeyer: This is tricky outside Scenario as well (see [this code in ops](https://github.com/canonical/operator/blob/6032adbd34cce4d8e620b60a276b79de5b9f752f/ops/charm.py#L1478).

We `type: ignore` the `.on` definition, because it doesn't match the parent `ops.Object`. Maybe that causes issues? The custom events are defined on the `CustomEvents` class.

It didn't seem like it was worth trying to get this to work, since if the test had the wrong event name or something like that it would still fail. But if you know how to make this work without jumping through lots of hoops, or in a way that would help actual Scenario use, I'm happy to do that.

testing/tests/test_charm_spec_autoload.py
james-garner-canonical: My understanding was that we don't want them as they just add noise.

benhoyt: Yeah, I think that's what we settled on for tests: they don't add any value for tests, just noise. I think it's pretty clear for test functions that we shouldn't do it.

testing/tests/helpers.py
tonyandrewmeyer: No:

```
/home/tameyer/w/type-check-scenario-tests2/testing/tests/helpers.py
  /home/tameyer/w/type-check-scenario-tests2/testing/tests/helpers.py:75:19 - error: Argument type is partially unknown
    Argument corresponds to parameter "iterable" in function "sorted"
    Argument type is "Unknown | list[dict[str, Any]]" (reportUnknownArgumentType)
```

testing/tests/test_context_on.py
james-garner-canonical: I don't have a quick fix for this. And I'm not sure how pervasive this problem is -- I have a feeling I've seen `type: ignore` comments like the one in `ops` on charm and/or charm library `on` attributes. 

The current ignore seems fine to me. I've created #2237 for the larger issue.

testing/tests/helpers.py
james-garner-canonical: Thanks for checking. Annoying how `Unknown` is handled after we've explicitly ignored the import for being untyped. The `cast` seems like the right approach then, sadly.

testing/tests/test_emitted_events_util.py
tonyandrewmeyer: Huh, I thought `pyright` did, but I did a quick check locally and it seems like `pyright` does indeed ignore them (I guess you have to use the `pyright`-specific syntax). And [the docs](https://microsoft.github.io/pyright/#/comments?id=line-level-diagnostic-suppression) back this up. `mypy` does let you use the scopes in `type: ignore`.

We have these in many other places across ops too. Maybe that was from a time when `mypy` was used, or maybe it's there for the clearer documentation? I'm tempted to leave them for those reasons - would you prefer they were gone?

james-garner-canonical: I feel like they're misleading with our current type-checking, as we might really be type ignoring a non-assignment issue here, for example.

I'd be on board with either using `pyright` specific ignores so long as we're using `pyright`, or using bare `type: ignore` comments with an explanation comment if we want more detail.

In this case the line is simple enough that I wouldn't complain about a bare `type: ignore` with no further comments, though [we do already use pyright: ignore in some places](https://github.com/search?q=repo%3Acanonical%2Foperator%20pyright%3A%20ignore&type=code).

tonyandrewmeyer: Discussed in daily: going forward the decision is not to add a scope to `type: ignore`, and to use `pyright: ignore[scope]` if needed.

testing/tests/test_e2e/test_state.py
tonyandrewmeyer: `sort_patch` is only used here and internally in `helpers.py`. I'll get rid of the lambda here, we don't really need to be calling `sorted()` on a hard-coded value.

testing/tests/test_context_on.py
```diff
@@ -3,14 +3,15 @@
...
+from typing import Any, Literal
 
 import ops
 import pytest
 
 import scenario
 
 
-META = {
+META: dict[str, Any] = {
```
james-garner-canonical, line 16:
```suggestion
META: Mapping[str, Any] = {
```
So the type checker helps us not to mutate this.

testing/tests/test_context_on.py
```diff
@@ -101,7 +102,12 @@ def test_simple_deprecated_events(event_name, event_kind):
         ('secret_rotate', ops.SecretRotateEvent, 'app'),
     ],
 )
-def test_simple_secret_events(as_kwarg, event_name, event_kind, owner):
+def test_simple_secret_events(
+    as_kwarg: bool,
+    event_name: str,
+    event_kind: type[ops.EventBase],
```
james-garner-canonical:
```suggestion
    event_kind: type[ops.SecretEvent],
```
And drop the `typing.cast` below.

testing/tests/test_context_on.py
```diff
@@ -146,13 +153,14 @@ def test_revision_secret_events(event_name, event_kind):
         mgr.run()
         secret_event, collect_status = mgr.charm.observed
         assert isinstance(secret_event, event_kind)
+        secret_event = typing.cast('ops.SecretExpiredEvent | ops.SecretRemoveEvent', secret_event)
```
james-garner-canonical: Again, we can probably drop the `cast` by making the annotation match this.

testing/tests/test_context_on.py
```diff
@@ -181,6 +189,7 @@ def test_storage_events(event_name, event_kind):
         mgr.run()
         storage_event, collect_status = mgr.charm.observed
         assert isinstance(storage_event, event_kind)
+        storage_event = typing.cast('ops.StorageEvent', storage_event)
```
james-garner-canonical: Likewise here.

testing/tests/test_context_on.py
```diff
@@ -257,7 +266,7 @@ def test_relation_app_events(as_kwarg, event_name, event_kind):
 
 
 def test_relation_complex_name():
-    meta = copy.deepcopy(META)
+    meta: dict[str, Any] = copy.deepcopy(META)
```
james-garner-canonical: I don't think this type annotation is needed.

testing/tests/test_context_on.py
```diff
@@ -299,8 +308,10 @@ def test_relation_unit_events_default_unit(event_name, event_kind):
         mgr.run()
         relation_event, collect_status = mgr.charm.observed
         assert isinstance(relation_event, event_kind)
+        relation_event = typing.cast('ops.RelationEvent', relation_event)
```
james-garner-canonical: `event_kind: type[ops.RelationEvent]`

testing/tests/test_context_on.py
```diff
@@ -322,8 +333,10 @@ def test_relation_unit_events(event_name, event_kind):
...
```
james-garner-canonical: `event_kind: type[ops.RelationEvent]`

testing/tests/test_context_on.py
```diff
@@ -524,11 +538,11 @@ def test_custom_event_with_args():
 def test_custom_event_is_hookevent():
     ctx = scenario.Context(CustomCharm, meta=META, actions=ACTIONS)
     with pytest.raises(ValueError):
-        ctx.on.custom(MyConsumer.on.foo_relation_changed)
+        ctx.on.custom(MyConsumer.on.foo_relation_changed)  # type: ignore
 
 
 def test_custom_event_with_scenario_args():
-    meta = META.copy()
+    meta: dict[str, Any] = META.copy()
```
james-garner-canonical: Unnecessary annotation.

testing/tests/test_e2e/test_state.py
```diff
@@ -137,12 +137,14 @@ def call(charm: CharmBase, e):
...
-    assert jsonpatch_delta(out_purged, state) == sort_patch([
+    # This should be sorted by path then op.
+    expected = [
         {'op': 'replace', 'path': '/app_status/message', 'value': 'foo barz'},
         {'op': 'replace', 'path': '/app_status/name', 'value': 'waiting'},
         {'op': 'replace', 'path': '/unit_status/message', 'value': 'foo test'},
         {'op': 'replace', 'path': '/unit_status/name', 'value': 'active'},
-    ])
+    ]
+    assert jsonpatch_delta(out_purged, state) == expected
```
james-garner-canonical, line 149: Does this fail because the sorting is different now? To avoid the `sort_patch` call, how about manually sorting these kinds by path and then op?

testing/tests/test_context_on.py
tonyandrewmeyer: https://github.com/canonical/operator/pull/2228 has changes related to this, it'd be nice to get that merged rather than making more conflicting changes.

james-garner-canonical: Agreed.

testing/tests/test_e2e/test_state.py
tonyandrewmeyer: It shouldn't fail, I ran `tox` before pushing. Also, I put the value in there by copying out the `sorted()` bit, running it in a REPL, then pasting it back, and doing `tox -e format`. Also, I don't see it failing in the CI.

james-garner-canonical: Oh I misread the paths, they're already sorted correctly, my mistake.

testing/src/scenario/_runtime.py
```diff
@@ -278,8 +281,8 @@ def exec(
         self,
         state: State,
         event: _Event,
-        context: Context,
-    ):
+        context: Context[CharmType],
+    ) -> Generator[Ops[CharmType]]:
```
dimaqq: Isn't that Py 3.13 thing? And older Python requiring `..., None, None]`?
Which is why we often use `Iterator[x]` instead.

testing/tests/helpers.py
```diff
@@ -75,9 +73,5 @@ def jsonpatch_delta(self, other: State):
...
         dict_other[attr] = [dataclasses.asdict(o) for o in dict_other[attr]]
         dict_self[attr] = [dataclasses.asdict(o) for o in dict_self[attr]]
-    patch = jsonpatch.make_patch(dict_other, dict_self).patch
-    return sort_patch(patch)
-
-
-def sort_patch(patch: list[dict], key=lambda obj: obj['path'] + obj['op']):
-    return sorted(patch, key=key)
+    patch = jsonpatch.make_patch(dict_other, dict_self).patch  # type: ignore
+    return sorted(patch, key=lambda obj: obj['path'] + obj['op'])  # type: ignore
```
dimaqq, line 79: Here, it's rather unclear what type ignores apply to.

testing/tests/test_charm_spec_autoload.py
```diff
@@ -36,12 +40,12 @@ def import_name(name: str, source: Path) -> type[CharmType]:
...
-    meta=None,
-    actions=None,
-    config=None,
+    meta: dict[str, Any] | None = None,
+    actions: dict[str, Any] | None = None,
+    config: dict[str, Any] | None = None,
     name: str = 'MyCharm',
     legacy: bool = False,
-):
+) -> Generator[type[CharmBase]]:
```
dimaqq: same wrt Generator[x]

testing/tests/test_plugin.py
```diff
@@ -5,13 +5,15 @@
...
 import sys
 
+import pytest
+
 pytest_plugins = 'pytester'
 sys.path.append('.')
 
 
-def test_plugin_ctx_run(pytester):
+def test_plugin_ctx_run(pytester: pytest.Pytester):
```
dimaqq: **Change request:** what is being tested here? Is this perhaps dead code?

Looking at Scenario history:
https://github.com/tonyandrewmeyer/operator/commit/4da1031d6d9686ecaf78ce790cc28a141673f9b8#diff-b62266dbe51dfb9358f9f7174a4dd8f165eb650d71cf59c42f3a933078e4a95b

I think this file was needed because Scenario used to bring in first-party fixtures, specifically `emitted_events`.

I don't think that's the case with modern ops[testing], in which case this file can be removed.

dimaqq (changes requested):

testing/src/scenario/_runtime.py
james-garner-canonical: Good catch, I'd be fine with any of
1. `Iterator[Ops[CharmType]]`
2. `Generator[Ops[CharmType], None, None]`
3. `from typing_extensions import Generator` (requires [typing_extensions >=4.12.0](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.Generator))

testing/tests/helpers.py
james-garner-canonical: They're needed because the `jsonpatch` library doesn't have type annotations, apparently due to supporting Python 2.7 (!)
- https://github.com/stefankoegl/python-json-patch/issues/158

But I swear one of these PRs fixed this with a `cast` instead, though I can't find that anymore. Note that there *is* a competing implementation of typing for this [in the other PR here](https://github.com/canonical/operator/pull/2234/files#diff-8cfebfb9f4f01a8925d6cd21fa816e7697c0ead8cab60e90d993b30380cfae8c). Would be good to settle on one approach or perhaps even split the `helpers.py` changes into a separate PR to merge first?

testing/tests/test_consistency_checker.py
```diff
@@ -51,17 +52,17 @@ def assert_inconsistent(
...
+    unit_id: int = 0,
 ):
     check_consistency(state, event, charm_spec, juju_version, unit_id)
 
 
 def test_base():
     state = State()
     event = _Event('update_status')
-    spec = _CharmSpec(MyCharm, {})
+    spec: _CharmSpec[ops.CharmBase] = _CharmSpec(MyCharm, {})
```
james-garner-canonical, line 65: I suspect this type annotation is needed because the `CharmType` type variable used in the `_CharmSpec` generic is invariant ([the default](https://docs.python.org/3/library/typing.html#typing.TypeVar)). It's probably intended to be covariant -- anywhere a `_CharmSpec[ops.CharmBase]` is required, a `_CharmSpec[MyCharmBaseSubClass]` should be fine.

I believe that moving from invariant to covariant would be backwards compatible, so we could fix that.

testing/tests/test_context_on.py
james-garner-canonical: Now that #2228 is merged, I actually think there's no overlap here, and we should just go with `Mapping` so the type checker helps us avoid mutating the `META` global.

testing/tests/test_plugin.py
james-garner-canonical: Well spotted, should probably be a separate PR either before or after this one though IMO.

testing/tests/test_runtime.py
```diff
@@ -48,74 +50,85 @@ def test_event_emission():
...
+            meta=meta,
+        )
         runtime = Runtime(
             'foo',
-            _CharmSpec(
-                my_charm_type,
-                meta=meta,
-            ),
+            charm_spec,
         )
```
james-garner-canonical, line 64: Minor nit, how about:
```py
        runtime = Runtime('foo', charm_spec)
```

testing/tests/test_runtime.py
```diff
@@ -48,74 +50,85 @@ def test_event_emission():
...
 
 
 @pytest.mark.parametrize('app_name', ('foo', 'bar-baz', 'QuX2'))
 @pytest.mark.parametrize('unit_id', (1, 2, 42))
-def test_unit_name(app_name, unit_id):
-    meta = {
+def test_unit_name(app_name: str, unit_id: int):
+    meta: dict[str, Any] = {
         'name': app_name,
     }
```
james-garner-canonical, line 86: Rather than adding a type annotation, how about inlining this in the `charm_spec` creation below?
```py
        meta = {'name': app_name},
```

testing/tests/test_runtime.py
```diff
@@ -48,74 +50,85 @@ def test_event_emission():
...
-        context=Context(my_charm_type, meta=meta),
+        context=ctx,
     ) as manager:
         assert manager.charm.unit.name == f'{app_name}/{unit_id}'
 
 
 def test_env_clean_on_charm_error(monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setenv('SCENARIO_BARE_CHARM_ERRORS', 'false')
-    meta = {'name': 'frank', 'requires': {'box': {'interface': 'triangle'}}}
+    meta: dict[str, Any] = {'name': 'frank', 'requires': {'box': {'interface': 'triangle'}}}
```
james-garner-canonical, line 111: I'd inline this one too.
```py
        meta={'name': 'frank', 'requires': {'box': {'interface': 'triangle'}}},
```

testing/tests/helpers.py
james-garner-canonical: Having scanned through this PR again, I don't think a separate PR is needed, as it seems like the only use of `sort_patch` is eliminated in this PR. So I think we just need something to make the `type: ignore` comments clear, and then make sure that this version of `helpers.py` is the one we land on in `main`.
```suggestion
    # The jsonpatch library is untyped.
    # See: https://github.com/stefankoegl/python-json-patch/issues/158
    patch = jsonpatch.make_patch(dict_other, dict_self).patch  # type: ignore
    return sorted(patch, key=lambda obj: obj['path'] + obj['op'])  # type: ignore
```

testing/tests/test_consistency_checker.py
james-garner-canonical: I'm fine with leaving this line as you have it now and tracking the covariance in a separate issue (#2242).

james-garner-canonical (commented):
I've finished my review of this PR, see individual comments and replies.

testing/tests/test_plugin.py
tonyandrewmeyer: https://github.com/canonical/operator/pull/2247

testing/tests/test_plugin.py
dimaqq, line 1: this file has been removed in the other PR, please merge main into your branch and resolve the conflicts

dimaqq (approved):

testing/tests/test_runtime.py
tonyandrewmeyer: Isn't it better to have the `Any`, which matches the correct type, rather than have it infer `str`, which is more restrictive than it really is?

tonyandrewmeyer:
@james-garner-canonical these PRs feel like dead weight sitting on me now 😞. Would you mind giving this another review? What I'd like to propose, if you're ok with it, is to focus on the changes that are in the PR (and getting the type checking running on these files without errors) and then do any additional improvements in follow-up PRs (not the other 2 already open, but follow-ups after those).

testing/tests/test_runtime.py
james-garner-canonical: This is an interesting case, as the underlying problem is that `_CharmSpec` should really be typed as accepting `Mapping` rather than `dict`, since it doesn't intend to mutate it. If we fixed that in future, or (in some even further future) had some even more precise typing for charm metadata, it seems like it would be a shame to have introduced an extra `Any` here.

That said, it's not a major deal. If you prefer `dict[str, Any]` to in-lining it, it's not a blocker for me.

testing/tests/test_e2e/test_relations.py
```diff
@@ -430,27 +425,30 @@ def callback(charm: CharmBase, event):
...
-        assert event.app  # that's always present
+        assert event.app  # type: ignore  # that's always present
         # .unit is always None for created and broken.
-        if isinstance(event, (RelationCreatedEvent, RelationBrokenEvent)):
-            assert event.unit is None
+        if isinstance(event, ops.RelationCreatedEvent | ops.RelationBrokenEvent):
+            assert event.unit is None  # type: ignore
         else:
-            assert event.unit
+            assert event.unit  # type: ignore
```
dimaqq, line 447: that's a lot of type: ignore's... are they all needed?
would a more precise ignorance be more readable perhaps?

Issue #2229, opened 2025-12-12
Errors are being reported by onboarders when following the K8s tutorial


james-garner-canonical:
Errors are being reported in the onboarding channel by onboarders following the K8s tutorial.

From [this message](https://chat.canonical.com/canonical/pl/fjwpa3sebb8aiekf1qr31ip6jo) the following steps:
```bash
multipass launch --cpus 4 --memory 8G --disk 50G --name charm-dev charm-dev
multipass shell charm-dev
git clone https://github.com/canonical/operator.git
cd examples/k8s-5-observe/
charmcraft fetch-libs
charmcraft pack
```

<details>
<summary>Results in the following error (click for more)</summary>

```
$ charmcraft pack
Failed to run the build script for part 'charm'.
Detailed information: 
:: + env -i LANG=C.UTF-8 LC_ALL=C.UTF-8 CRYPTOGRAPHY_OPENSSL_NO_LEGACY=true PIP_CONSTRAINT=/snap/charmcraft/7517/lib/python3.12/site-packages/charmcraft/parts/plugins/charm-constraints.txt PATH=/snap/charmcraft/7517/libexec/charmcraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin SNAP=/snap/charmcraft/7517 SNAP_ARCH=amd64 SNAP_NAME=charmcraft SNAP_VERSION=4.0.1 /snap/charmcraft/7517/bin/python -u -I /snap/charmcraft/7517/lib/python3.12/site-packages/charmcraft/charm_builder.py --builddir /root/parts/charm/build --installdir /root/parts/charm/install --entrypoint /root/parts/charm/build/src/charm.py -p pip -p setuptools -p wheel -r requirements.txt
:: Traceback (most recent call last):
::   File "/snap/charmcraft/7517/lib/python3.12/site-packages/charmcraft/charm_builder.py", line 499, in <module>
::     main()
::   File "/snap/charmcraft/7517/lib/python3.12/site-packages/charmcraft/charm_builder.py", line 494, in main
::     builder.build_charm()
::   File "/snap/charmcraft/7517/lib/python3.12/site-packages/charmcraft/charm_builder.py", line 95, in build_charm
::     self.handle_dependencies()
::   File "/snap/charmcraft/7517/lib/python3.12/site-packages/charmcraft/charm_builder.py", line 368, in handle_dependencies
::     self._install_dependencies(staging_venv_dir)
[...]

PR #2233, opened 2025-12-12
ci: add scheduled workflow for packing and integration testing example charms

dwilding:
As James suggested in #2229, this PR adds a scheduled workflow that packs and runs integration tests for our example charms. I've scheduled it to run each ~Tuesday~ Wednesday before NZ hours.

(We already have a workflow that lints and runs unit tests for the example charms. That workflow is fast and runs on PRs.)

I've organised the example charms into three jobs:
- The machine tutorial charm, based on a Charmcraft 4 profile.
- The httpbin-demo K8s charm, based on a Charmcraft 4 profile.
- The K8s tutorial charms, not based on a Charmcraft 4 profile. We're planning to refactor these later in the 26.04 cycle.

I tested the jobs by temporarily enabling them for this PR. All are passing after #2232 merged.

### DIFF
--- a/.github/workflows/example-charm-integration-tests.yaml
+++ b/.github/workflows/example-charm-integration-tests.yaml
@@ -3,7 +3,7 @@ name: Example Charm Integration Tests
 on:
   workflow_dispatch:
   schedule:
-    - cron: '50 15 * * *'  # 15:50 UTC daily
+    - cron: '50 15 * * 2'  # 15:50 UTC Tuesdays
 
 permissions: {}
 


.github/workflows/example-charm-integration-tests.yaml
```diff
@@ -0,0 +1,105 @@
+name: Example Charm Integration Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '50 15 * * 2'  # 15:50 UTC Tuesdays
```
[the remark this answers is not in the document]
dwilding: I think it's clear enough

james-garner-canonical (approved):
Looks great to me, thanks!

Thanks for running the new workflow in CI to test it -- I don't see the results linked, but I trust that everything is good like you said, and we can easily update this workflow as needed.

.github/workflows/example-charm-integration-tests.yaml
james-garner-canonical: Just had a thought: would daily (Mon-Fri) be a better cadence?

dwilding: I'm totally good with daily - the tests don't take _that_ long to run. I'll make the change now

dwilding:
For completeness, here are my test runs of the new CI:

- https://github.com/canonical/operator/actions/runs/20162959531 for all charms before k8s-5-observe was fixed by #2232
- https://github.com/canonical/operator/actions/runs/20163731133 after k8s-5-observe was fixed

.github/workflows/example-charm-integration-tests.yaml
```diff
@@ -0,0 +1,105 @@
+name: Example Charm Integration Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '50 15 * * *'  # 15:50 UTC daily
```
dimaqq: My vote would be to run these monthly instead.

james-garner-canonical:
> Looks good to me. Maybe cc @benhoyt if we should be running these tests at all.

I believe we discussed this PR in daily a while back.

.github/workflows/example-charm-integration-tests.yaml
james-garner-canonical: The benefit of running them daily would be to know as early as actionable if something has broken in our tutorials, like the recent breakage that we didn't know about until onboarders began discussing it and finding workarounds. If anything does break, I think we'd want to fix it asap, so these should stay passing.

What would be the benefit of running monthly instead?

benhoyt: Let's just run weekly for now and see how we go.

dwilding:
As discussed at daily today, I've switched back to running the new workflow on a weekly schedule.

dimaqq (approved):

dwilding:
The workflow ran successfully on schedule today: https://github.com/canonical/operator/actions/runs/20465329325

Well, almost successfully... I had expected it to run yesterday, but was apparently off by one day when doing my UTC conversion. So Wednesdays it is :man_dancing:

PR #2234, opened 2025-12-13

tonyandrewmeyer:
testing/tests is added to the files that pyright should check, and the files that are left to do are added to the exclusion list.

Add type annotations to roughly half of the test/test_e2e files (the rest are left for another PR).

See also #2230.

### DIFF
--- a/testing/tests/helpers.py
+++ b/testing/tests/helpers.py
@@ -7,7 +7,7 @@ import dataclasses
 import logging
 from collections.abc import Callable
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any
 
 import jsonpatch  # type: ignore
 from scenario.context import _DEFAULT_JUJU_VERSION, Context
--- a/testing/tests/test_e2e/test_actions.py
+++ b/testing/tests/test_e2e/test_actions.py
@@ -12,7 +12,6 @@ from scenario.state import State, _Action, _next_action_id
 from ops._private.harness import ActionFailed
 from ops.charm import ActionEvent, CharmBase
 from ops.framework import Framework
-from ops.version import version as ops_version
 
 
 @pytest.fixture(scope='function')
--- a/testing/tests/test_e2e/test_event.py
+++ b/testing/tests/test_e2e/test_event.py
@@ -3,8 +3,6 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
-
 import pytest
 from scenario import Context
 from scenario.state import State, _CharmSpec, _Event, _EventType


testing/tests/test_e2e/test_deferred.py
```diff
@@ -228,22 +227,24 @@ def __init__(self, charm: ops.CharmBase):
...
+            framework.observe(self.consumer.on.foo_changed, self._on_event)  # type: ignore[attr-defined]
 
-    ctx = Context(MyCharm, meta=mycharm.META, capture_deferred_events=True)
+    ctx = Context(MyCharm, meta=mycharm.META, capture_deferred_events=True)  # type: ignore[attr-defined]
 
-    mycharm.defer_next = 1
-    state_1 = ctx.run(ctx.on.custom(MyConsumer.on.foo_changed, 'foo', 28), State())
+    mycharm.defer_next = 1  # type: ignore[attr-defined]
+    state_1 = ctx.run(
+        ctx.on.custom(typing.cast('typing.Any', MyConsumer.on).foo_changed, 'foo', 28), State()
```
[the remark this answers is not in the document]
tonyandrewmeyer: No, we need the quotes, and linting enforces that.

testing/tests/test_e2e/test_state.py
```diff
@@ -780,13 +787,15 @@ def test_layer_from_rockcraft(rockcraft: dict[str, Any]):
         assert layer_check_level == check.get('level', ops.pebble.CheckLevel.UNSET.value)
         if 'exec' in check:
             assert layer_check.exec is not None and 'command' in check['exec']
+            assert 'command' in layer_check.exec
```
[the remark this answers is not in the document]
tonyandrewmeyer: No, check and laywr_check are not the same object.

testing/tests/test_e2e/test_state.py
```diff
@@ -780,13 +787,15 @@ def test_layer_from_rockcraft(rockcraft: dict[str, Any]):
...
             assert layer_check.exec['command'] == check['exec']['command']
         if 'tcp' in check:
             assert layer_check.tcp is not None and 'port' in check['tcp']
+            assert 'port' in layer_check.tcp
```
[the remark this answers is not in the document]
tonyandrewmeyer: No layer_check and check are not he same object.

testing/tests/test_e2e/test_config.py
```diff
@@ -71,18 +72,18 @@ def check_cfg(charm: CharmBase):
...
                 self.framework.observe(evt, self._on_event)
 
-        def _on_event(self, event):
+        def _on_event(self, event: ops.EventBase) -> None:
             # access the config to trigger a config-get
-            foo_cfg = self.config['foo']  # noqa: F841
-            baz_cfg = self.config['baz']  # noqa: F841
-            qux_cfg = self.config['qux']  # noqa: F841
+            _foo_cfg = self.config['foo']
+            _baz_cfg = self.config['baz']
```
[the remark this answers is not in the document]
tonyandrewmeyer: The underscore indicates that it's not used. Same below.

testing/tests/helpers.py
```diff
@@ -77,5 +77,8 @@ def jsonpatch_delta(self, other: 'State'):
     return sort_patch(patch)
 
 
-def sort_patch(patch: list[dict], key=lambda obj: obj['path'] + obj['op']):
+def sort_patch(
+    patch: list[dict[str, Any]],
+    key: Callable[[dict[str, Any]], str] = lambda obj: obj['path'] + obj['op'],
+) -> list[dict[str, Any]]:
```
james-garner-canonical: This gets factored out in one of these PRs doesn't it?

testing/tests/helpers.py
```diff
@@ -58,7 +58,7 @@ def trigger(
     return state_out
 
 
-def jsonpatch_delta(self, other: 'State'):
+def jsonpatch_delta(self: 'State', other: 'State') -> list[dict[str, Any]]:
```
james-garner-canonical: These probably shouldn't be quoted, should they?

testing/tests/helpers.py
tonyandrewmeyer: Yes. I mentioned it in the commit message but probably should have put it in the description as well. My intention is to merge each of these in turn, dealing with the merge conflicts at the time, rather than try to set them up as a sequence to go into a single merge into main or anything consistent like that.

testing/tests/helpers.py
tonyandrewmeyer: No, but that's a linting concern (fixed in https://github.com/canonical/operator/pull/2228) rather than one that pyright picks up.

pyproject.toml
```diff
@@ -269,8 +269,28 @@ convention = "google"
...
+    "testing/tests/test_e2e/test_relations.py",
+    "testing/tests/test_e2e/test_trace_data.py",
+    "testing/tests/test_e2e/test_juju_log.py",
+    "testing/tests/test_e2e/test_status.py",
+    "testing/tests/test_e2e/test_storage.py",
+    "testing/tests/test_e2e/test_manager.py",
+    "testing/tests/test_e2e/test_ports.py",
+    "testing/tests/test_e2e/test_rubbish_events.py",
+    "testing/tests/test_e2e/test_pebble.py",
+]
```
dimaqq, line 295: Is the idea to make this more incremental?

Also, how does it work when a path is excluded but also matches an included glob?

testing/tests/test_e2e/test_state.py
```diff
@@ -878,7 +887,9 @@ class Charm(ops.CharmBase):
     assert state.get_relations('relreq')[0].interface == 'across'
     assert state.get_relations('relpro')[0].interface == 'around'
     assert state.get_relations('sub')[0].interface == 'below'
-    assert state.get_relation(relation.id).remote_app_data == {'a': 'b'}
+    retrieved_relation = state.get_relation(relation.id)
+    assert isinstance(retrieved_relation, Relation)
```
dimaqq, line 894: **Aside:** I wish this was more ergonomic for our users.

Like if they could conceptually write:

```py
rel = state.get_relation<Relation>(rel.id)
# or
rel = state.get_relation(rel.id, kind=Relation)  # runtime error on mismatch
```

Perhaps more Pythonically:

```py
rel = state.get_regular_relation(rid)
per = state.get_peer_relation(pid)
sub = state.get_subordinate_relation(sid)
```

testing/tests/test_e2e/test_state.py
```diff
@@ -508,11 +510,14 @@ def test_state_immutable(obj_in, attribute: str, get_method: str, key_attr: str,
     elif attribute == 'secrets':
         # State.get_secret only takes keyword arguments, while the others take
         # only positional arguments.
+        assert isinstance(obj_in, Secret)
         obj_out = state_out.get_secret(id=obj_in.id)
     elif attribute == 'resources':
         # Charms can't change resources, so there's no State.get_resource.
         obj_out = [r for r in state_out.resources if r == obj_in][0]
     else:
+        # This handles: Relation, PeerRelation, SubordinateRelation, Network, Container, Storage, StoredState
```
dimaqq: **Minor:** would perhaps be clearer like so:

```suggestion
        # BaseRelation: {Relation, PeerRelation, SubordinateRelation}
```

Alternatively list all leaf classes in the `instance` tuple explicitly?

testing/tests/test_e2e/test_deferred.py
```diff
@@ -14,7 +13,7 @@
 
 
 @pytest.fixture(scope='function')
-def mycharm():
+def mycharm() -> type[ops.CharmBase]:
```
dimaqq, line 21: AI-generated alternative that would make things simpler, I think:

```py

class _MyCharmTemplate(ops.CharmBase):
    META: dict[str, typing.Any] = {
        "name": "mycharm",
        "requires": {"foo": {"interface": "bar"}},
        "containers": {"foo": {"type": "oci-image"}},
    }
    defer_next = 0
    captured: list[ops.EventBase] = []

        super().__init__(framework)
        for evt in self.on.events().values():
            if issubclass(evt.event_type, ops.LifecycleEvent):
                continue
            self.framework.observe(evt, self._on_event)

    def _on_event(self, event: ops.EventBase):
        self.captured.append(event)
        if self.defer_next > 0:
            self.defer_next -= 1
            event.defer()


def make_mycharm_class() -> type[ops.CharmBase]:
    # Fresh class object each call; class name can be constant.
    return type(
        "MyCharm",
        (_MyCharmTemplate,),
        {
            # fresh per-test class attributes
            "META": copy.deepcopy(_MyCharmTemplate.META),
            "defer_next": 0,
            "captured": [],
            "__module__": __name__,  # optional: nicer repr / pickling expectations
        },
    )


@pytest.fixture(scope="function")
def mycharm() -> type[ops.CharmBase]:
    return make_mycharm_class()
```

This way we can pass a real charm class type around, and not `ops.CharmBase`.

testing/tests/test_e2e/test_actions.py
```diff
@@ -158,7 +160,7 @@ def _on_foo_action(self, event):
     assert ctx.action_results == {'initial': 'result', 'final': 'result'}
 
 
-def _ops_less_than(wanted_major, wanted_minor):
+def _ops_less_than(wanted_major: int, wanted_minor: int) -> bool:
```
dimaqq: **Change request:** Can we rip this and all tests that use it out?

1. Python 3.8 gone, no point supporting ops 2.x
2. we release in lock-step, this test will never be run against some other ops

If a separate PR is better, then open an issue and add a comment here.

dimaqq (changes requested):

pyproject.toml
tonyandrewmeyer: > Is the idea to make this more incremental?

Not sure what you mean by this. Do you mean will there be follow-ups to do the rest? If so, yes, #2230, #2235. Do you mean incremental in this PR? It started as one file per commit, but then after review I have not done any squashing so there is a mixed bag after that.

> Also, how does it work when a path is excluded but also matches an included glob?

Include first, exclude second.

testing/tests/test_e2e/test_state.py
tonyandrewmeyer: Issues and PRs welcome 😉

testing/tests/test_e2e/test_actions.py
tonyandrewmeyer: https://github.com/canonical/operator/pull/2240

pyproject.toml
dimaqq: I guess what I'd like to see is that eventually the long hand-crafted list of excludes goes away.

tonyandrewmeyer: It goes away when all three PRs are merged; they each do about a third of the files.

testing/tests/test_e2e/test_deferred.py
dimaqq: I guess more pragmatically:

```py
class MyCharm(ops.CharmBase):
    META = ...
    ...

@pytest.fixture
def mycharm() -> type[MyCharm]:
    """Create a brand new MyCharm for each test."""
    return type("MyCharm", (MyCharm,), {"META": ..., ...})
```

testing/tests/test_e2e/test_state.py
james-garner-canonical: I created #2243 to track this.

testing/tests/test_e2e/test_deferred.py
james-garner-canonical: I reckon it's probably OK to land this PR without these extra changes, and create a follow-up issue to track tidying this up. When we tidy it up, `pyright` will help us remove any now-unnecessary `type: ignore` comments.

I think the PR is probably good to go, just need to see how it looks with `main` merged.

pyproject.toml
```diff
@@ -257,6 +257,13 @@ builtins-ignorelist = ["id", "min", "map", "range", "type", "TimeoutError", "Con
 include = ["ops/*.py", "ops/_private/*.py", "test/*.py", "test/charms/*/src/*.py", "testing/src/*.py", "testing/tests/*.py", "testing/tests/test_e2e/*.py"]
 exclude = [
     "tracing/*",
+    "testing/tests/helpers.py",
+    "testing/tests/test_charm_spec_autoload.py",
+    "testing/tests/test_consistency_checker.py",
+    "testing/tests/test_context_on.py",
+    "testing/tests/test_context.py",
+    "testing/tests/test_emitted_events_util.py",
+    "testing/tests/test_runtime.py",
```
james-garner-canonical: Should this PR be introducing these lines? Are they left over from when we weren't checking anything on `main`?

testing/tests/test_e2e/test_actions.py
```diff
@@ -160,44 +162,44 @@ def _on_foo_action(self, event):
...
+def test_action_event_has_override_id(mycharm: type[CharmBase]):
     uuid = '0ddba11-cafe-ba1d-5a1e-dec0debad'
 
     def handle_evt(charm: CharmBase, evt: ActionEvent):
         if not isinstance(evt, ActionEvent):
             return
         assert evt.id == uuid
 
-    mycharm._evt_handler = handle_evt
+    mycharm._evt_handler = handle_evt  # type: ignore
```
dimaqq, line 185: If I drop something like this at the top of the file:

```py
  class MyCharm(CharmBase):
      _evt_handler = Callable[[CharmBase, ActionEvent], None]
```

And update the test argument to `mycharm: type[MyCharm]`, then I don't need `# type: ignore` here and all the other places.

WDYT about that, or perhaps a Protocol?

testing/tests/test_e2e/test_deferred.py
```diff
@@ -160,18 +159,18 @@ def test_deferred_workload_event(mycharm):
...
-    ctx = Context(mycharm, meta=mycharm.META, capture_deferred_events=True)
+def test_defer_reemit_lifecycle_event(mycharm: type[ops.CharmBase]):
+    ctx = Context(mycharm, meta=mycharm.META, capture_deferred_events=True)  # type: ignore
 
-    mycharm.defer_next = 1
+    mycharm.defer_next = 1  # type: ignore
     state_1 = ctx.run(ctx.on.update_status(), State())
 
-    mycharm.defer_next = 0
+    mycharm.defer_next = 0  # type: ignore
```
dimaqq, line 173: I wonder if it's possible to type `defer_next` instead.

testing/tests/test_e2e/test_state.py
```diff
@@ -529,14 +533,16 @@ def test_state_immutable(obj_in, attribute: str, get_method: str, key_attr: str,
...
+    relation_type: type[RelationBase], mycharm: type[CharmBase]
+):
+    def event_handler(charm: CharmBase, _: EventBase):
         rel = charm.model.get_relation(relation_type.__name__)
         assert rel is not None
         rel.data[charm.app]['a'] = 'b'
         rel.data[charm.unit]['c'] = 'd'
 
-    mycharm._call = event_handler
+    mycharm._call = event_handler  # type: ignore
```
dimaqq, line 545: Same here wrt `type: ignore`.

I'd rather we didn't have to sprinkle static type checker suppression so much.

testing/tests/test_e2e/test_actions.py
james-garner-canonical: I guess you disagree with [this comment](https://github.com/canonical/operator/pull/2234#discussion_r2892920290)? I think we'd get value out of being able to turn on static type checking for these files now, and deferring further iteration til later. `pyright` will help us get rid of `type: ignore` comments when we eventually improve type visibility (by complaining about them once they're unnecessary).

testing/tests/test_e2e/test_deferred.py
tonyandrewmeyer: The issue is that the `mycharm` fixture says that it returns a `ops.CharmBase`, which doesn't have a `defer_next`. It would need to have a protocol or something like that I think.

pyproject.toml
tonyandrewmeyer: Bad merge.

testing/tests/test_e2e/test_deferred.py
james-garner-canonical: Yes, I think ideally in future we'd lift the class declarations to the top-level, but IMO that's out of scope for this PR.

tonyandrewmeyer:
This is ready for review again (it has the latest changes from main, and other cleanup).

My preference, like with the first one in this series, is to focus on getting the tests type-checked without `type: ignore` or `Any` *everywhere*, and then to make other improvements in the future. I'm afraid that this will drag on forever (it's already quite old) without that, and it's already three PRs that aren't super short. I think that some of the fixes really would be best done by some significant refactoring of the tests (particularly: less defining things on the fly, fewer generic helper systems), and I don't think that belongs in this PR.

Obviously, if you disagree with that, then push back, but I would appreciate (in that case) suggestions on how to manage the scope of the PR.

testing/tests/test_e2e/test_actions.py
```diff
@@ -160,44 +163,55 @@ def _on_foo_action(self, event):
...
+def _ops_less_than(wanted_major: int, wanted_minor: int) -> bool:
+    major, minor = (int(v) for v in ops_version.split('.')[:2])
+    if major < wanted_major:
+        return True
+    if major == wanted_major and minor < wanted_minor:
+        return True
+    return False
+
+
+@pytest.mark.skipif(_ops_less_than(2, 11), reason="ops 2.10 and earlier don't have ActionEvent.id")
```
james-garner-canonical: So shouldn't `_ops_less_than` be gone now thanks to #2240?

testing/tests/test_e2e/test_actions.py
```diff
@@ -160,44 +163,55 @@ def _on_foo_action(self, event):
...
 
-    mycharm._evt_handler = handle_evt
+    mycharm._evt_handler = handle_evt  # type: ignore
 
     ctx = Context(mycharm, meta={'name': 'foo'}, actions={'foo': {}})
     ctx.run(ctx.on.action('foo'), State())
 
 
-def test_action_event_has_override_id(mycharm):
+@pytest.mark.skipif(_ops_less_than(2, 11), reason="ops 2.10 and earlier don't have ActionEvent.id")
```
james-garner-canonical: Also here.

testing/tests/test_e2e/test_event.py
```diff
@@ -76,6 +78,8 @@ def test_emitted_framework():
 
 def test_emitted_deferred():
     class MyCharm(ops.CharmBase):
+        META: ClassVar[dict[str, Any]] = {'name': 'joop'}
+
```
james-garner-canonical: Doesn't look like this is used.

```suggestion
```

testing/tests/helpers.py
james-garner-canonical: Please make sure this doesn't get reintroduced by this PR, I'm not sure what order these PRs are ending up landing in now.

testing/tests/test_e2e/test_actions.py
```diff
@@ -111,27 +114,27 @@ def handle_evt(_: CharmBase, evt: ActionEvent):
...
 
-def test_action_event_fail(mycharm):
+def test_action_event_fail(mycharm: type[CharmBase]):
     def handle_evt(_: CharmBase, evt: ActionEvent):
         if not isinstance(evt, ActionEvent):
             return
         evt.fail('action failed!')
 
-    mycharm._evt_handler = handle_evt
+    mycharm._evt_handler = handle_evt  # type: ignore
```
dimaqq, line 122: I'm still not to happy about these `type: ignore` all over but I guess the change is an improvement overall and silencing can be removed later.

testing/tests/test_e2e/test_actions.py
dimaqq: #2240 got merged, let's clean up.

testing/tests/test_e2e/test_deferred.py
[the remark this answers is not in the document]
dimaqq: Is `typing.cast('typing.Any', ...)` substantially different from `# type: ignore`?

testing/tests/helpers.py
tonyandrewmeyer: The order has always been the same, tests2, then tests3, then tests4 (tests-no-number was something else, already merged, first of all).

But yes, will review again for mistaken merges.

james-garner-canonical: TBH I never noticed the branches were numbered :laughing:

PR #2235, opened 2025-12-14

tonyandrewmeyer:
testing/tests is added to the files that pyright should check, and the files that are left to do are added to the exclusion list.

Add type annotations to half of the test_e2e test files (the other half are in another PR, #2234).

### DIFF
--- a/testing/tests/test_e2e/test_actions.py
+++ b/testing/tests/test_e2e/test_actions.py
@@ -246,7 +246,7 @@ def test_action_get_returns_copy(mycharm: type[CharmBase]):
         second = backend.action_get()
         results.append((first, second))
 
-    mycharm._evt_handler = handle_evt  # type: ignore[attr-defined]
+    mycharm._evt_handler = handle_evt  # type: ignore
 
     ctx: Context[CharmBase] = Context(
         mycharm,


benhoyt:
Thanks for doing these. Out of interest, did you use a tool to help automate these?

tonyandrewmeyer:
> Thanks for doing these. Out of interest, did you use a tool to help automate these?

Not an agent, but I did use the tab-completion functionality in VS Code a reasonable amount. I find it's particularly useful when what you're doing is very repetitive like this was.

Not sure if you remember or not, but I tried getting Copilot to do this work in the Gothenburg workshop (and then over the following days) and it did a terrible job. I still feel that an agent ought to be able to do this and that I probably just needed a different prompt or configuration, but I wasn't willing to spend more time on that with this ticket.

benhoyt:
> Not an agent, but I did use the tab-completion functionality in VS Code a reasonable amount. I find it's particularly useful when what you're doing is very repetitive like this was.
>
> Not sure if you remember or not, but I tried getting Copilot to do this work...

Yeah, I did remember. I was more curious about whether you'd used a mechanical tool like https://github.com/orsinium-labs/infer-types (have no idea whether it works or is any good).

tonyandrewmeyer:
> I was more curious about whether you'd used a mechanical tool like https://github.com/orsinium-labs/infer-types (have no idea whether it works or is any good).

Ah, interesting. I wasn't aware of such tools, otherwise I probably would have given one a go :)

testing/tests/test_e2e/test_juju_log.py
```diff
@@ -14,14 +16,14 @@
 @pytest.fixture(scope='function')
 def mycharm():
     class MyCharm(CharmBase):
-        META = {'name': 'mycharm'}
+        META: dict[str, Any] = {'name': 'mycharm'}
```
james-garner-canonical: I don't think the type annotations on the `META` and `ACTIONS` attributes get us anything right now tbh. Though why aren't we getting complaints about mutable class attributes needing `ClassVar`? If we do annotate these it should probably be `Mapping[str, Any]`. And we should investigate why we're not getting the `ClassVar` complaint.

testing/tests/test_e2e/test_manager.py
```diff
@@ -28,7 +32,7 @@ def _on_event(self, e):
     return MyCharm
 
 
-def test_manager(mycharm):
+def test_manager(mycharm: Any) -> None:
```
james-garner-canonical: WDYT about lifting `MyCharm` and its `META` and `ACTIONS` attributes to the top-level like we did for https://github.com/canonical/grafana-agent-operator/pull/355?

I guess this PR will need the `-> None` annotations stripped as well.

testing/tests/test_e2e/test_juju_log.py
james-garner-canonical: Kind of superseded by lifting these to the top-level if we can do that, see other comment.

tonyandrewmeyer: I think the `ClassVar` bit comes from ruff, not from pyright. There's overlap here with https://github.com/canonical/operator/pull/2228

james-garner-canonical: Ah maybe that's where all the deja vu is coming from. Could we perhaps merge #2228 before working further on these then?

testing/tests/test_e2e/test_pebble.py
```diff
@@ -627,43 +635,43 @@ def _on_config_changed(self, _):
...
     # Verify that start_checks works.
     state_out = ctx.run(ctx.on.config_changed(), state=state_out)
     chk1_info = state_out.get_container('foo').get_check_info('chk1')
-    assert chk1_info.status == pebble.CheckStatus.UP
+    assert chk1_info.status == ops.pebble.CheckStatus.UP
 
 
 @pytest.fixture
-def reset_security_logging():
+def reset_security_logging() -> Iterator[None]:
```
dimaqq: **nit:** I think that this would be just as well without the return type annotation.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -143,16 +145,18 @@ def callback(self: CharmBase):
         file = tmp_path / 'bar' / 'baz.txt'
 
         # another is:
-        assert file == Path(out.get_container('foo').mounts['foo'].source) / 'bar' / 'baz.txt'
+        assert (
+            file == pathlib.Path(out.get_container('foo').mounts['foo'].source) / 'bar' / 'baz.txt'
+        )
```
dimaqq: **minor:** this assert and the one below would look better if split into two lines.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -143,16 +145,18 @@ def callback(self: CharmBase):
...
         # but that is actually a symlink to the context's root tmp folder:
         assert (
-            Path(ctx._tmp.name) / 'containers' / 'foo' / 'foo' / 'bar' / 'baz.txt'
+            pathlib.Path(ctx._tmp.name) / 'containers' / 'foo' / 'foo' / 'bar' / 'baz.txt'
         ).read_text() == text
         assert file.read_text() == text
 
         # shortcut for API niceness purposes:
-        file = container.get_filesystem(ctx) / 'foo' / 'bar' / 'baz.txt'
+        file = container.get_filesystem(ctx) / 'foo' / 'bar' / 'baz.txt'  # type: ignore
```
dimaqq: Why is `type: ignore` needed here?

testing/tests/test_e2e/test_pebble.py
```diff
@@ -221,15 +226,16 @@ def callback(self: CharmBase):
...
+    class MyCharm(ops.CharmBase):
+        def __init__(self, framework: ops.Framework):
             super().__init__(framework)
             self.framework.observe(self.on.foo_pebble_ready, self._on_ready)
 
-        def _on_ready(self, _):
+        def _on_ready(self, _: ops.EventBase):
             proc = self.unit.get_container('foo').exec(['ls'], stdin=stdin)
             if write:
+                assert proc.stdin is not None
```
dimaqq: I wonder if a slight refactor could help here.
I took a long dive into https://documentation.ubuntu.com/ops/latest/reference/pebble/#ops.pebble.ExecProcess.stdin to figure out why `stdin` which is only sometimes set (parametrised) seems to be always asserted on.

It turns out that stdin and write are careful kept exclusive in the test parameters.

How about:
```py
if stdin:
    proc = x.exec(stdin=stdin)
else:
    proc = x.exec()
    proc.stdin.write(write)
```

testing/tests/test_e2e/test_pebble.py
```diff
@@ -809,8 +815,8 @@ def test_layers_merge_in_plan(layer1_name, layer2_name):
                 'level': 'ready',
                 'startup': 'enabled',
                 'threshold': 10,
-                'period': 1,
-                'timeout': 28,
+                'period': '1s',
+                'timeout': '28s',
```
dimaqq, line 894: that's a good logical change... but it's a bit weird to see it in the type checking PR. were the values outright wrong before?

testing/tests/test_e2e/test_rubbish_events.py
```diff
@@ -43,19 +43,21 @@ def __init__(self, framework: Framework):
...
-        def _on_event(self, e):
+        def _on_event(self, e: EventBase):
             MyCharm.evts.append(e)
 
     return MyCharm
 
 
 @pytest.mark.parametrize('evt_name', ('rubbish', 'foo', 'bar'))
-def test_rubbish_event_raises(mycharm: CharmBase, evt_name: str):
+def test_rubbish_event_raises(mycharm: type[CharmBase], evt_name: str):
```
dimaqq, line 53: **question:** why are we not changing `CharmBase` to `ops.CharmBase` in this file? Is it because it's covered by another PR?

P.S. I wonder if a type alias for `type[ops.CharmBase]` would be handy here.

testing/tests/test_e2e/test_storage.py
```diff
@@ -24,57 +24,57 @@ class MyCharmWithoutStorage(CharmBase):
...
 
     assert storage_ctx.requested_storages['foo'] == n
 
 
-def test_storage_usage(storage_ctx):
+def test_storage_usage(storage_ctx: Context[MyCharmWithStorage]):
     storage = Storage('foo')
     # setup storage with some content
-    (storage.get_filesystem(storage_ctx) / 'myfile.txt').write_text('helloworld')
+    (storage.get_filesystem(storage_ctx) / 'myfile.txt').write_text('helloworld')  # type: ignore
```
dimaqq: Why is type ignore needed here?

testing/tests/test_e2e/test_storage.py
```diff
@@ -87,14 +87,16 @@ def test_storage_usage(storage_ctx):
         myfile.write_text('helloworlds')
 
     # post-mortem: inspect fs contents.
-    assert (storage.get_filesystem(storage_ctx) / 'path.py').read_text() == 'helloworlds'
+    assert (
+        storage.get_filesystem(storage_ctx) / 'path.py'  # type: ignore
```
dimaqq: and here

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer: They were the wrong type, yes.

testing/tests/test_e2e/test_rubbish_events.py
tonyandrewmeyer: I was trying to avoid doing the change from `from ops import CharmBase` to `import ops` as much as possible to keep the size/scope down. I was originally just doing it when there was a new import needed (most often `Framework`). Then when I merged in the `ruff` PR there were a bunch of merge conflicts in most of the imports so I also simplified those.

I guess I could uniformly make this change, but I think it would probably be better as a separate PR, and it would be nice if it was after this one so I don't have to redo those changes.

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer: I think the signature for `get_filesystem` needs to take a `Context[ops.CharmBase]` rather than plain `Context` or something like that.

testing/tests/test_e2e/test_rubbish_events.py
james-garner-canonical: +1 on leaving it for a separate PR

james-garner-canonical: +1 on leaving it for a separate PR

testing/tests/test_e2e/test_pebble.py
tonyandrewmeyer: We looked at this a bit yesterday together and couldn't find something that was significantly tidier.

testing/tests/test_e2e/test_storage.py
tonyandrewmeyer: Opened https://github.com/canonical/operator/pull/2248 to resolve.

testing/tests/test_e2e/test_storage.py
tonyandrewmeyer: Opened https://github.com/canonical/operator/pull/2248 to resolve.

testing/src/scenario/state.py
```diff
@@ -1139,7 +1139,7 @@ def services(self) -> dict[str, pebble.ServiceInfo]:
             infos[name] = info
         return infos
 
-    def get_filesystem(self, ctx: Context) -> pathlib.Path:
+    def get_filesystem(self, ctx: Context[CharmBase]) -> pathlib.Path:
```
dimaqq: Has this become `ctx: Context[Any]` in the other PR?

testing/tests/test_e2e/test_juju_log.py
tonyandrewmeyer: All the other PRs are merged now.

testing/tests/test_e2e/test_manager.py
tonyandrewmeyer: Out of date now that the other PRs are done.

testing/tests/test_e2e/test_actions.py
```diff
@@ -246,9 +246,9 @@ def handle_evt(_: CharmBase, evt):
         second = backend.action_get()
         results.append((first, second))
 
-    mycharm._evt_handler = handle_evt
+    mycharm._evt_handler = handle_evt  # type: ignore[attr-defined]
```
tonyandrewmeyer:
```suggestion
    mycharm._evt_handler = handle_evt  # type: ignore
```

tonyandrewmeyer:
@dimaqq @james-garner-canonical this last one is ready for review now. It's come out a bit simpler than the last two, I think, with only the Pebble tests being quite large changes.

PR #2244, opened 2025-12-17
test: update new scenario test to fix linting error

james-garner-canonical:
This PR adds a `noqa` directive to the test newly added in #2219, which is needed due to the linting updates in #2228.

### DIFF
--- a/testing/tests/test_context.py
+++ b/testing/tests/test_context.py
@@ -187,4 +187,4 @@ def test_run_env_cleared():
 def test_charm_spec_is_deprecated():
     ctx = Context(CharmBase, meta={'name': 'some-name'})
     with pytest.warns(DeprecationWarning):
-        ctx.charm_spec  # type: ignore  # noqa: B018
+        _ = ctx.charm_spec  # type: ignore


tonyandrewmeyer (approved):

dwilding (approved):
https://docs.astral.sh/ruff/rules/useless-expression/#known-problems suggests doing:

```py
with pytest.warns(DeprecationWarning):
    _ = ctx.charm_spec
```

if I understand correctly. What do you think?

testing/tests/test_context.py
```diff
@@ -187,4 +187,4 @@ def _on_start(self, event):
 def test_charm_spec_is_deprecated():
     ctx = Context(CharmBase, meta={'name': 'some-name'})
     with pytest.warns(DeprecationWarning):
-        ctx.charm_spec  # type: ignore
+        ctx.charm_spec  # type: ignore  # noqa: B018
```
james-garner-canonical: Good call @dwilding 
```suggestion
        _ = ctx.charm_spec  # type: ignore
```

Issue #2242, opened 2025-12-16
CharmType TypeVar used by _CharmSpec should probably be covariant


james-garner-canonical:
The `CharmType` type variable used in the `_CharmSpec` generic is invariant ([the default](https://docs.python.org/3/library/typing.html#typing.TypeVar)). The variable is defined [here](https://github.com/canonical/operator/blob/ffa669e9a6bf2a76b543e8a09963ec8c3a8cbbd1/testing/src/scenario/state.py#L71). It's probably intended to be covariant, at least as far as `_CharmSpec` is concerned. That is, anywhere a `_CharmSpec[ops.CharmBase]` is required, the more specific `_CharmSpec[MyCharmBaseSubClass]` should be fine.

Invariance is problematic in cases like [this one](https://github.com/canonical/operator/pull/2230/files/6aa58fcac7fd40a974d83e1d0c1c2aa2fb0ce760#diff-f3190d94568be90381db0085d0cfce3208215375bfa511046e42fa04981f392b) in #2230, which looks like this:
```py
def f(c: _CharmSpec[ops.CharmBase]):
    ...

charm_spec = _CharmSpec(MyCharmClass, ...)   # typed as _CharmSpec[MyCharmClass]
f(charm_spec)  # error because f needs exactly a _CharmSpec[ops.CharmBase]
```

It seems pretty clear that `_CharmSpec[Specific]` should be fine where `_CharmSpec[Base]` is declared since `_CharmSpec` is essentially an immutable container here (it's a frozen dataclass). I believe that moving from invariant to covariant here would be backwards compatible. However, we need to check all the uses of the existing `TypeVar`, or make a new one for this, in case there are any uses that *should* be invariant (or contravariant).

james-garner-canonical:
At the same time, we should consider adding a default to the `TypeVar` for better ergonomics in cases like #2248. This [requires Python 3.13](https://docs.python.org/3/library/typing.html#typing.TypeVar.__default__), so we'd need to use [TypeVar from typing_extensions](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeVar). We added a `testing` dependency on `typing_extensions` in #2219, so this should be OK.

PR #2248, opened 2025-12-18
chore: change the get_filesystem context argument from Context to Context[Any]

tonyandrewmeyer:
Neither of the `get_filesystem` methods make any use of the generic type of Context, but we need to provide a type to avoid having calls need a `type: ignore`, so `Any` seems the appropriate type.

Maybe #2242 should be done instead? But I think that can be done later, replacing this if needed.


james-garner-canonical (approved):
Seems like a good fix to me. In addition to the covariance issue, another thing that might improve the typing UX of `Context` would be adding a default to the generic as well. I'll add a note to the issue.

PR #2249, opened 2025-12-18
docs: update the action how-to to explain the additionalProperties default flip

tonyandrewmeyer:
Juju 4 flips the default value of `additionalProperties` from `true` to `false` (see https://github.com/juju/juju/issues/21294). Add a note to the actions how-to guide that lets people know this and that they need to explicitly include the field if they want to ensure compatibility across Juju 3 and Juju 4.

### DIFF
--- a/examples/k8s-4-action/charmcraft.yaml
+++ b/examples/k8s-4-action/charmcraft.yaml
@@ -43,8 +43,8 @@ actions:
       show-password:
         description: Show username and password in output information
         type: boolean
-        default: False
-    additionalProperties: False
+        default: false
+    additionalProperties: false
 
 containers:
   demo-server:


tonyandrewmeyer:
@james-garner-canonical I realised after opening this that it would be good to change the tutorial charm as well, so I've added that. Feel free to re-review or not as you please.

docs/howto/manage-actions.md
```diff
@@ -39,6 +39,8 @@ actions:
     additionalProperties: false
 ```
 
+Note that you should **always** include the `additionalProperties` field, which controls whether users can provide properties that are not in the definition, as the default changed from `true` in Juju 3 to `false` in Juju 4. If you don't explicitly include it in the YAML, then the charm behaviour will differ depending on which Juju version the charm is deployed to.
```
dwilding: A few minor suggestions for readability/impact

```suggestion
You should **always** include the `additionalProperties` field, which controls whether users can provide properties that are not in the definition. The default value of the field changed from `true` in Juju 3 to `false` in Juju 4. If you don't explicitly include the field, the charm behaviour will differ depending on which Juju version the charm is deployed to.
```

dwilding (approved):
Good catch, thanks!

PR #2250, opened 2025-12-18
fix: correct the value of `additional_properties` in the action meta in Juju 4

### LINKED ISSUE #2188: Update the Scenario behaviour when using action params not defined in the schema
We should make sure that Scenario properly reflects the `additionalProperties` field in `actions.yaml` or `charmcraft.yaml`. The consistency checker should error if additional properties are not allowed and the event provides one that isn't defined in the schema.

Note that [Juju 4 reverses the default value](https://github.com/juju/juju/issues/21294) so we need to handle that based on the mocked Juju version.

This should help people identify that they need to explicitly set `additionalProperties=true` for Juju 4, if their integration tests aren't covering this behaviour.

tonyandrewmeyer:
Juju 4 flips the default value of additionalProperties from true to false (see https://github.com/juju/juju/issues/21294).

We expose this value in `ActionMeta.additional_properties`. The PR adjusts this so that the default value (if `additionalProperties` is not in the YAML) is correct in both Juju 4 and earlier Juju.

This loads the version from the environment, which we try to avoid (using the existing `juju_context` on the model instead). However, `ActionMeta` doesn't have access to the model object, and nor does `CharmMeta`, so piping in access would be somewhat messy. I've gone with the simpler solution, but please consider this in the review.

Fixes #2188

### DIFF
--- a/test/test_charm.py
+++ b/test/test_charm.py
@@ -627,11 +627,17 @@ def test_config_from_raw():
     assert meta.config['sssh'].description == 'a user secret'
 
 
-def test_actions_from_charm_root():
+@pytest.mark.parametrize('additional_properties', [False, True])
+def test_actions_from_charm_root(additional_properties: bool):
     with tempfile.TemporaryDirectory() as d:
         td = pathlib.Path(d)
         (td / 'actions.yaml').write_text(
-            yaml.safe_dump({'foo': {'description': 'foos the bar', 'additionalProperties': False}})
+            yaml.safe_dump({
+                'foo': {
+                    'description': 'foos the bar',
+                    'additionalProperties': additional_properties,
+                }
+            })
         )
         (td / 'metadata.yaml').write_text(
             yaml.safe_dump({'name': 'bob', 'requires': {'foo': {'interface': 'bar'}}})
@@ -640,18 +646,21 @@ def test_actions_from_charm_root():
         meta = ops.CharmMeta.from_charm_root(td)
         assert meta.name == 'bob'
         assert meta.requires['foo'].interface_name == 'bar'
-        assert not meta.actions['foo'].additional_properties
+        assert meta.actions['foo'].additional_properties == additional_properties
         assert meta.actions['foo'].description == 'foos the bar'
 
 
 @pytest.mark.parametrize(
     'juju_version,additional_properties',
-    [('2.9', True), ('3.6.12', True), ('4.0.0', False), ('4.1', False)],
+    [('2.9', True), ('3.6.12', True), ('4.0.0', False), ('4.1', False), (None, True)],
 )
 def test_actions_additional_properties(
-    monkeypatch: pytest.MonkeyPatch, juju_version: str, additional_properties: bool
+    monkeypatch: pytest.MonkeyPatch, juju_version: str | None, additional_properties: bool
 ):
-    monkeypatch.setenv('JUJU_VERSION', juju_version)
+    if juju_version is None:
+        monkeypatch.delenv('JUJU_VERSION', raising=False)
+    else:
+        monkeypatch.setenv('JUJU_VERSION', juju_version)
     action = ops.ActionMeta('foo', {})
     assert action.additional_properties == additional_properties
 


ops/charm.py
```diff
@@ -2209,7 +2211,9 @@ def __init__(self, name: str, raw: dict[str, Any] | None = None):
         self.description = raw.get('description', '')
         self.parameters = raw.get('params', {})  # {<parameter name>: <JSON Schema definition>}
         self.required = raw.get('required', [])  # [<parameter name>, ...]
-        self.additional_properties = raw.get('additionalProperties', True)
+        # The default in Juju 4 is False. The default in earlier Juju versions is True.
+        juju_version = JujuVersion(os.environ.get('JUJU_VERSION', '0.0.0'))
```
james-garner-canonical, line 2215: Significantly, it seems like this is the first thing we have under `CharmMeta` that actually depends on the Juju context, rather than being purely derived from the charm's YAML files.

I wonder if it would be cleaner to update `ActionMeta` and `CharmMeta` to take a `JujuContext`. This would have to be optional, since the constructors are public, and I guess the default behaviour if the context isn't provided would still have to be loading it from the environment. And if we want this to continue to work outside a real Juju context, we'd still need to provide a default value for `JUJU_VERSION` (among others), since `JujuContext.from_environ` will error if it isn't set.

This seems like more trouble than it would be worth right now, so I guess just checking the version is the right call, even if it feels a little bit messy.

---

It does seem like a shame to essentially reimplement the deprecated `JujuVersion.from_environ` here. WDYT about moving the logic to a private `JujuVersion._from_environ` method and calling that here? Maybe it's simple enough that it's not worth it, but it could be tidier if we end up needing to do more of this in future.

james-garner-canonical (approved):
I've shared some thoughts in a separate comment, but ultimately I suspect this current implementation might be the right tradeoff, so I'm approving this PR.

ops/charm.py
tonyandrewmeyer: > This seems like more trouble than it would be worth right now, so I guess just checking the version is the right call, even if it feels a little bit messy.

Yes, that's basically what I said in the PR description 😉.

> It does seem like a shame to essentially reimplement the deprecated `JujuVersion.from_environ` here. WDYT about moving the logic to a private `JujuVersion._from_environ` method and calling that here? Maybe it's simple enough that it's not worth it, but it could be tidier if we end up needing to do more of this in future.

I'm not sure it's worth it for a single line. I guess there is the `JUJU_VERSION` constant and the weird "use 0.0.0 if not present" default, but I'm not sure those merit avoiding duplicating it.

james-garner-canonical: > > This seems like more trouble than it would be worth right now, so I guess just checking the version is the right call, even if it feels a little bit messy.
> 
> Yes, that's basically what I said in the PR description 😉.

Just reproducing your working.

> > It does seem like a shame to essentially reimplement the deprecated `JujuVersion.from_environ` here. WDYT about moving the logic to a private `JujuVersion._from_environ` method and calling that here? Maybe it's simple enough that it's not worth it, but it could be tidier if we end up needing to do more of this in future.
> 
> I'm not sure it's worth it for a single line. I guess there is the `JUJU_VERSION` constant and the weird "use 0.0.0 if not present" default, but I'm not sure those merit avoiding duplicating it.

Yeah, I'm on the fence -- I think it's more worth it for the `0.0.0` default than `JUJU_VERSION`, but probably fine either way.

dwilding: Based on a quick scan around the code, it seems that using `0.0.0` by default is an established pattern. So keeping the logic here seems fine to me (but I'd also be fine either way).

test/test_charm.py
```diff
@@ -644,6 +644,18 @@ def test_actions_from_charm_root():
...
+@pytest.mark.parametrize(
+    'juju_version,additional_properties',
+    [('2.9', True), ('3.6.12', True), ('4.0.0', False), ('4.1', False)],
+)
+def test_actions_additional_properties(
+    monkeypatch: pytest.MonkeyPatch, juju_version: str, additional_properties: bool
+):
+    monkeypatch.setenv('JUJU_VERSION', juju_version)
+    action = ops.ActionMeta('foo', {})
+    assert action.additional_properties == additional_properties
```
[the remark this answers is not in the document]
dwilding: @tonyandrewmeyer do you plan to add either of these? If so I'll wait before finishing my review

ops/charm.py
```diff
@@ -2209,7 +2211,9 @@ def __init__(self, name: str, raw: dict[str, Any] | None = None):
...
+        self.additional_properties = raw.get('additionalProperties', juju_version < '4.0.0')
```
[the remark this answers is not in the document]
tonyandrewmeyer: This is more-or-less the point I made in the description and James made above. I would rather plumb through the model's `juju_context` than put in a cached `JujuVersion` object (which will have all sorts of issues if you're running Scenario tests that change the version).

test/test_charm.py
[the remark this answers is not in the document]
tonyandrewmeyer: I've added them now, thanks.

dwilding (approved):

Issue #2059, opened 2025-09-26
httpbin-demo: Add unit tests to fully exercise the status reporting


dwilding:
Our [httpbin charm](https://github.com/canonical/operator/tree/main/examples/httpbin-demo) has a collect-status handler that accounts for various situations, but the unit tests don't cover all cases. The unit tests currently cover:

- The charm going into active status under normal operation
- The charm going into maintenance status if the container can't be reached - as a byproduct of trying to change the config in this situation

Other situations aren't covered. For example, the container is available but we haven't asked Pebble to create a service yet.

Some notes from discussion within Charm Tech:

> Generally speaking, we should cover each case of the status reporting. We can consider the status reporting part of the charm's "user interface". It's fine to specifically test the status reporting. In some cases, this will be cleaner than always testing it as a byproduct of some other event. Best practice: Run `update_status`, don't run `collect_unit_status`.

I'm tracking this as an issue because it would be nice to link to the httpbin unit tests from the Ops docs.

dwilding:
I had an initial go at this, but need to work on other things right now. I'll plan to revisit in 26.04

https://github.com/dwilding/operator/blob/httpbin-collect-status/examples/httpbin-demo/tests/unit/test_charm.py

PR #2254, opened 2025-12-23
docs: improve unit tests of httpbin demo charm

dwilding:
This PR makes several improvements to the unit tests of our httpbin demo charm:

- Include tests for the status reporting logic, per #2059
- Make the config-changed tests focused on config, not also testing the Pebble connection
- Refactor the config-changed tests. Parameterize valid and invalid values of `log-level`.
- Remove the Arrange, Act, Assert comments from all but the first (pebble-ready) test, as I feel the other tests don't really benefit from the comments

I also made a couple of small changes elsewhere:

- Used `ops.pebble` instead of importing `pebble` from `ops`
- Changed the charm code to set a different status message if `log-level` is empty
- Slightly refactored the charm code, after a suggestion from James
- Linked to the httpbin unit tests from the "See more" section of the Harness migration guide ([preview](https://canonical-ubuntu-documentation-library--2254.com.readthedocs.build/ops/2254/howto/legacy/migrate-unit-tests-from-harness/#see-more))

### DIFF
--- a/examples/httpbin-demo/src/charm.py
+++ b/examples/httpbin-demo/src/charm.py
@@ -42,8 +42,6 @@ class HttpbinConfig:
     def __post_init__(self):
         log_level = self.log_level.lower()
         valid_log_levels = {"info", "debug", "warning", "error", "critical"}
-        if not log_level:
-            raise ValueError(f"Empty log level. Valid values are: {', '.join(valid_log_levels)}.")
         if log_level not in valid_log_levels:
             raise ValueError(
                 f"Invalid log level: '{self.log_level}'. "
--- a/examples/httpbin-demo/tests/unit/test_charm.py
+++ b/examples/httpbin-demo/tests/unit/test_charm.py
@@ -131,21 +131,18 @@ def test_config_changed(user_log_level: str, gunicorn_log_level: str):
     assert state_out.unit_status == testing.ActiveStatus()
 
 
-def test_config_changed_empty():
+@pytest.mark.parametrize(
+    "user_log_level",
+    [
+        "",
+        "foobar",
+    ],
+)
+def test_config_changed_invalid(user_log_level: str):
     """Test a config-changed event when the config is invalid."""
     ctx = testing.Context(HttpbinDemoCharm)
     container = testing.Container(CONTAINER_NAME, can_connect=True)
-    state_in = testing.State(containers={container}, config={"log-level": ""})
+    state_in = testing.State(containers={container}, config={"log-level": user_log_level})
     state_out = ctx.run(ctx.on.config_changed(), state_in)
     assert isinstance(state_out.unit_status, testing.BlockedStatus)
-    assert state_out.unit_status.message.startswith("Empty log level.")
-
-
-def test_config_changed_invalid():
-    """Test a config-changed event when the config is invalid."""
-    ctx = testing.Context(HttpbinDemoCharm)
-    container = testing.Container(CONTAINER_NAME, can_connect=True)
-    state_in = testing.State(containers={container}, config={"log-level": "foobar"})
-    state_out = ctx.run(ctx.on.config_changed(), state_in)
-    assert isinstance(state_out.unit_status, testing.BlockedStatus)
-    assert state_out.unit_status.message.startswith("Invalid log level: 'foobar'.")
+    assert state_out.unit_status.message.startswith(f"Invalid log level: '{user_log_level}'.")


examples/httpbin-demo/src/charm.py
```diff
@@ -82,6 +80,10 @@ def _on_collect_status(self, event: ops.CollectStatusEvent):
             # It's technically possible (but unlikely) for Pebble to have an internal error.
             logger.error("Unable to fetch service info from Pebble")
             raise
+        else:
+            if not service.is_running():
+                # We can connect to Pebble in the container, but the service hasn't started yet.
+                event.add_status(ops.MaintenanceStatus("waiting for workload"))
```
dwilding, line 86: I refactored this part of the charm code after James's comment on similar code elsewhere: https://github.com/canonical/operator/pull/2062#discussion_r2389453295

examples/httpbin-demo/tests/unit/test_charm.py
```diff
@@ -48,72 +62,87 @@ def test_httpbin_pebble_ready():
...
-    state_in = testing.State(containers={container}, config={"log-level": invalid_level})
-
-    # Act:
+    state_in = testing.State(containers={container}, config={"log-level": user_log_level})
     state_out = ctx.run(ctx.on.config_changed(), state_in)
-
-    # Assert:
     assert isinstance(state_out.unit_status, testing.BlockedStatus)
-    assert invalid_level in state_out.unit_status.message
+    assert f"'{user_log_level}'" in state_out.unit_status.message
```
dimaqq: Minor: this assertion doesn't do all that much when the test is parametrised with an empty string. A pair of quotes could come from anywhere.

P.S. this could also be written as:

```suggestion
    assert f"{user_log_level!r}" in state_out.unit_status.message
```

though I'm not sure which is easier to understand.

dwilding: Good point, thanks. How about this - I've changed the charm code to give a different error message when the config value is empty. Then we can split into two (unparameterized) tests:

```python
def test_config_changed_empty():
    """Test a config-changed event when the config is invalid."""
    ctx = testing.Context(HttpbinDemoCharm)
    container = testing.Container(CONTAINER_NAME, can_connect=True)
    state_in = testing.State(containers={container}, config={"log-level": ""})
    state_out = ctx.run(ctx.on.config_changed(), state_in)
    assert isinstance(state_out.unit_status, testing.BlockedStatus)
    assert state_out.unit_status.message.startswith("Empty log level.")


def test_config_changed_invalid():
    """Test a config-changed event when the config is invalid."""
    ctx = testing.Context(HttpbinDemoCharm)
    container = testing.Container(CONTAINER_NAME, can_connect=True)
    state_in = testing.State(containers={container}, config={"log-level": "foobar"})
    state_out = ctx.run(ctx.on.config_changed(), state_in)
    assert isinstance(state_out.unit_status, testing.BlockedStatus)
    assert state_out.unit_status.message.startswith("Invalid log level: 'foobar'.")
```

examples/httpbin-demo/src/charm.py
```diff
@@ -42,6 +42,8 @@ class HttpbinConfig:
     def __post_init__(self):
         log_level = self.log_level.lower()
         valid_log_levels = {"info", "debug", "warning", "error", "critical"}
+        if not log_level:
+            raise ValueError(f"Empty log level. Valid values are: {', '.join(valid_log_levels)}.")
```
james-garner-canonical: This is nice but seems unnecessary since we'd get a `ValueError` with a reasonable helpful message anyway: `Invalid log level: ''.`

Better to keep the code simpler with one less special case? But see also my comment on your thread with Dima.

examples/httpbin-demo/src/charm.py
```diff
@@ -82,6 +82,10 @@ def _on_collect_status(self, event: ops.CollectStatusEvent):
             # It's technically possible (but unlikely) for Pebble to have an internal error.
             logger.error("Unable to fetch service info from Pebble")
             raise
+        else:
+            if not service.is_running():
+                # We can connect to Pebble in the container, but the service hasn't started yet.
+                event.add_status(ops.MaintenanceStatus("waiting for workload"))
```
james-garner-canonical, line 86: I'd avoid using the `else` clause of `try ... except` in this case for readability, since it's not strictly necessary.
```suggestion
        if not service.is_running():
            # We can connect to Pebble in the container, but the service hasn't started yet.
            event.add_status(ops.MaintenanceStatus("waiting for workload"))
```

examples/httpbin-demo/tests/unit/test_charm.py
```diff
@@ -13,20 +13,34 @@
...
 import ops
-from ops import testing
+import pytest
+from ops import pebble, testing
 
 from charm import CONTAINER_NAME, SERVICE_NAME, HttpbinDemoCharm
 
+# A mock Pebble layer - useful for testing the charm's status reporting code. The status reporting
+# code shouldn't care how the service is started, so the layer doesn't need the real command.
+MOCK_LAYER = pebble.Layer(
```
james-garner-canonical: While we recommend `from ops import testing`, I think that's the only exception to our usual practice of keeping everything under the top-level namespace -- that is, I think we use `ops.pebble` generally, don't we?
```suggestion
from ops import testing

from charm import CONTAINER_NAME, SERVICE_NAME, HttpbinDemoCharm

# A mock Pebble layer - useful for testing the charm's status reporting code. The status reporting
# code shouldn't care how the service is started, so the layer doesn't need the real command.
MOCK_LAYER = ops.pebble.Layer(
```
There are a couple of other direct uses of `pebble` in this file that would need to be updated too if you accepted this suggestion.

A bit nitpicky, but since this is example code for charmers it seems worthwhile to follow our preferred approach.

examples/httpbin-demo/tests/unit/test_charm.py
```diff
@@ -48,72 +62,90 @@ def test_httpbin_pebble_ready():
...
-    state_in = testing.State(containers={container}, config={"log-level": "debug"})
+    container = testing.Container(
+        CONTAINER_NAME,
+        layers={"base": MOCK_LAYER},
+        service_statuses={SERVICE_NAME: pebble.ServiceStatus.INACTIVE},
+        can_connect=True,
+    )
+    state_in = testing.State(containers={container})
+    state_out = ctx.run(ctx.on.update_status(), state_in)
+    assert state_out.unit_status == testing.MaintenanceStatus("waiting for workload")
```
james-garner-canonical, line 94: As a general rule, I don't _love_ asserting on the exact contents of user-facing messages (e.g. errors, statuses, etc) in unit tests, as they're quite brittle. However, I can see the value in also checking that the message is correct for the different cases, and your approach is nice and readable, so I think I'm on board with it here on balance.

james-garner-canonical (approved):
Approving because these are nice improvements that would be fine to merge as-is.

I've commented on some small things I'd change, but feel free to disagree.

examples/httpbin-demo/tests/unit/test_charm.py
dwilding: Good call, I agree - done

examples/httpbin-demo/src/charm.py
dwilding: The `else` is needed, isn't it, because two of the `except` clauses don't re-raise. So it's possible that `service` isn't defined after the `try` block - if I understand things correctly

james-garner-canonical: Ah I didn't notice that, I assumed they all reraised

examples/httpbin-demo/src/charm.py
dwilding: Which comment on the thread with Dima? (Sorry if I'm totally missing it!)

james-garner-canonical: Ugh, I think I didn't submit it correctly and it's lost now.

My thoughts were that the new `message.startswith("Invalid log level: '{user_log_level}'.")` form should take care of the concern about an empty string coming from anywhere, so a single test could be fine.

I don't really mind the separate tests, but I'm not a big fan of the (imo unnecessary) separate error case.

dwilding: I get you, thanks. I agree that `message.startswith` alone is a better change than splitting into two code paths. I've reverted to a single `test_config_changed_invalid` test with a parameterised value.

Issue #2095, opened 2025-10-08
K8s tutorial: Clarify how status is set if no relation data


dwilding:
In [Integrate your charm with PostgreSQL](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/), the definition of `fetch_postgres_relation_data` has a comment that says:

> [...] If no data is retrieved, the unit is set to waiting status and the program exits with a zero status code.

This is confusing for several reasons:

- The method itself doesn't set the unit status. Instead, later in the doc we're adding a collect-status handler that sets waiting status.
- Even in the collect-status handler, `fetch_postgres_relation_data` isn't called. Instead, both places use `self.database.fetch_relation_data`.
- "The program exists with a zero status" is technically true but not very helpful. We don't usually think of charms as programs that exit with particular status codes.

---

Discussion in Matrix: https://matrix.to/#/!LhFxJIPEcCacgdMghH:ubuntu.com/$yreC1AD_wDNvYYL18CyzXL6UPOp1yeCbVV28BpiIswc

PR #2256, opened 2025-12-26
docs: clarify handling of postgres relation data in K8s tutorial

dwilding:
In [Integrate your charm with PostgreSQL](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#add-relation-event-observers), the part from "Add relation event observers" to the end of "Share the authentication information with your application" needs improving:

- I find the sequence of file edits confusing. I don't know what the goal is until I get to the end and work backwards.
- A couple of the method explanations and docstrings are much wordier than they need to be.
- A couple of places still talk about setting status without collect-status: The diagram at the end and the part mentioned in [#2095](https://github.com/canonical/operator/issues/2095).

I've reorganised the flow and improved the explanations of what's happening. That includes replacing the diagram by more focused text.

**Previews of updated sections:**
- [Add relation event observers](https://canonical-ubuntu-documentation-library--2256.com.readthedocs.build/ops/2256/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#add-relation-event-observers)
- [Fetch the database authentication data](https://canonical-ubuntu-documentation-library--2256.com.readthedocs.build/ops/2256/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#fetch-the-database-authentication-data)
- [Share the authentication data with your application](https://canonical-ubuntu-documentation-library--2256.com.readthedocs.build/ops/2256/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#share-the-authentication-data-with-your-application)

Other updates to the tutorial:
- Renamed `self._update_layer_and_restart()` to `self._replan_workload()` - [see comment](https://github.com/canonical/operator/pull/2256#discussion_r2663887990)
- Renamed `self._on_database_created()` to `self._on_database_endpoint()` - [see comment](https://github.com/canonical/operator/pull/2256#discussion_r2663884987)
- Renamed `self.fetch_postgres_relation_data()` to `self.fetch_database_relation_data()` and removed other places where PostgreSQL is mentioned, to be more generic - [see comment](https://github.com/canonical/operator/pull/2256#discussion_r2680141727)
- Simplified `self.get_app_environment()` - [see comment](https://github.com/canonical/operator/pull/2256#discussion_r2680167809)

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -165,17 +165,12 @@ def get_app_environment(self) -> dict[str, str]:
     db_data = self.fetch_database_relation_data()
     if not db_data:
         return {}
-    env = {
-        key: value
-        for key, value in {
-            'DEMO_SERVER_DB_HOST': db_data.get('db_host', None),
-            'DEMO_SERVER_DB_PORT': db_data.get('db_port', None),
-            'DEMO_SERVER_DB_USER': db_data.get('db_username', None),
-            'DEMO_SERVER_DB_PASSWORD': db_data.get('db_password', None),
-        }.items()
-        if value is not None
+    return {
+        'DEMO_SERVER_DB_HOST': db_data['db_host'],
+        'DEMO_SERVER_DB_PORT': db_data['db_port'],
+        'DEMO_SERVER_DB_USER': db_data['db_username'],
+        'DEMO_SERVER_DB_PASSWORD': db_data['db_password'],
     }
-    return env
 ```
 
 This method depends on the following method, which extracts the database authentication data:
--- a/examples/k8s-3-postgresql/src/charm.py
+++ b/examples/k8s-3-postgresql/src/charm.py
@@ -167,17 +167,12 @@ class FastAPIDemoCharm(ops.CharmBase):
         db_data = self.fetch_database_relation_data()
         if not db_data:
             return {}
-        env = {
-            key: value
-            for key, value in {
-                'DEMO_SERVER_DB_HOST': db_data.get('db_host', None),
-                'DEMO_SERVER_DB_PORT': db_data.get('db_port', None),
-                'DEMO_SERVER_DB_USER': db_data.get('db_username', None),
-                'DEMO_SERVER_DB_PASSWORD': db_data.get('db_password', None),
-            }.items()
-            if value is not None
+        return {
+            'DEMO_SERVER_DB_HOST': db_data['db_host'],
+            'DEMO_SERVER_DB_PORT': db_data['db_port'],
+            'DEMO_SERVER_DB_USER': db_data['db_username'],
+            'DEMO_SERVER_DB_PASSWORD': db_data['db_password'],
         }
-        return env
 
     def fetch_database_relation_data(self) -> dict[str, str]:
         """Retrieve relation data from a database."""
--- a/examples/k8s-4-action/src/charm.py
+++ b/examples/k8s-4-action/src/charm.py
@@ -208,17 +208,12 @@ class FastAPIDemoCharm(ops.CharmBase):
         db_data = self.fetch_database_relation_data()
         if not db_data:
             return {}
-        env = {
-            key: value
-            for key, value in {
-                'DEMO_SERVER_DB_HOST': db_data.get('db_host', None),
-                'DEMO_SERVER_DB_PORT': db_data.get('db_port', None),
-                'DEMO_SERVER_DB_USER': db_data.get('db_username', None),
-                'DEMO_SERVER_DB_PASSWORD': db_data.get('db_password', None),
-            }.items()
-            if value is not None
+        return {
+            'DEMO_SERVER_DB_HOST': db_data['db_host'],
+            'DEMO_SERVER_DB_PORT': db_data['db_port'],
+            'DEMO_SERVER_DB_USER': db_data['db_username'],
+            'DEMO_SERVER_DB_PASSWORD': db_data['db_password'],
         }
-        return env
 
     def fetch_database_relation_data(self) -> dict[str, str]:
         """Retrieve relation data from a database."""
--- a/examples/k8s-5-observe/src/charm.py
+++ b/examples/k8s-5-observe/src/charm.py
@@ -229,17 +229,12 @@ class FastAPIDemoCharm(ops.CharmBase):
         db_data = self.fetch_database_relation_data()
         if not db_data:
             return {}
-        env = {
-            key: value
-            for key, value in {
-                'DEMO_SERVER_DB_HOST': db_data.get('db_host', None),
-                'DEMO_SERVER_DB_PORT': db_data.get('db_port', None),
-                'DEMO_SERVER_DB_USER': db_data.get('db_username', None),
-                'DEMO_SERVER_DB_PASSWORD': db_data.get('db_password', None),
-            }.items()
-            if value is not None
+        return {
+            'DEMO_SERVER_DB_HOST': db_data['db_host'],
+            'DEMO_SERVER_DB_PORT': db_data['db_port'],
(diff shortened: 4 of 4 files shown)


docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -135,29 +135,54 @@ Next, in the `__init__` method, define a new instance of the 'DatabaseRequires'
...
 ```python
 # See https://charmhub.io/data-platform-libs/libraries/data_interfaces
 framework.observe(self.database.on.database_created, self._on_database_created)
 framework.observe(self.database.on.endpoints_changed, self._on_database_created)
 ```
 
+Finally, define the method that is called on the database events:
+
+```python
+def _on_database_created(
```
dimaqq: Minor: I wonder if the method can be named something else, given that it's connected to two distinct synthetic events.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -135,29 +135,54 @@ Next, in the `__init__` method, define a new instance of the 'DatabaseRequires'
...
+    self, _: DatabaseCreatedEvent | DatabaseEndpointsChangedEvent
+) -> None:
+    """Event is fired when postgres database is created or endpoint is changed."""
+    self._update_layer_and_restart()
```
dimaqq: Minor: restart is implicit in Pebble when the plan is updated.

Shall we call it something like `self._replan_workload()` or simply `self._replan()` ?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -216,9 +243,9 @@ def _update_layer_and_restart(self) -> None:
         logger.info('Unable to connect to Pebble: %s', e)
 ```
 
-We've also removed three `self.unit.status = ` lines. We'll handle replacing those shortly.
+We removed three `self.unit.status = ` lines from this version of the method. We'll handle replacing those shortly.
 
-Now, update your `_get_pebble_layer()` method to use the passed environment:
+Next, update `_get_pebble_layer()` to put the environment variables in the Pebble layer:
```
dimaqq, line 243: I think we should only talk about "layers" in the context of e.g. merging some preexisting configuration with newly generated config.

Since we `override: replace`, perhaps it's cleaner to reference the concept as "the plan".

(Or if we wanted to be pedantic, "this service's slice of the plan").

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: I've renamed it to `_on_database_endpoint`, which I think more clearly matches the point of observing these events

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: I've renamed it to `_replan_workload`. I think it's important to include `workload` in the context of the tutorial.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: That reasoning makes sense to me. I won't make changes here, as I think the implementation will need more thought. I've tracked #2265

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -135,29 +135,54 @@ Next, in the `__init__` method, define a new instance of the 'DatabaseRequires'
...
+framework.observe(self.database.on.endpoints_changed, self._on_database_endpoint)
 ```
 
+Finally, define the method that is called on the database events:
+
+```python
+def _on_database_endpoint(
+    self, _: DatabaseCreatedEvent | DatabaseEndpointsChangedEvent
+) -> None:
+    """Event is fired when postgres database is created or endpoint is changed."""
```
tonyandrewmeyer:
```suggestion
    """Event is fired when PostgreSQL database is created or endpoint is changed."""
```

or maybe remove "Postgres" entirely? If this was switched out to use another database (of the ones the library supports) this code should be able to stay the same (which is meant to be one of the selling points of Juju) so perhaps we should be generic too.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -135,29 +135,54 @@ Next, in the `__init__` method, define a new instance of the 'DatabaseRequires'
...
+```python
+def get_app_environment(self) -> dict[str, str]:
+    """Return a dictionary of environment variables for the application."""
+    db_data = self.fetch_postgres_relation_data()
+    if not db_data:
+        return {}
+    env = {
+        key: value
+        for key, value in {
+            'DEMO_SERVER_DB_HOST': db_data.get('db_host', None),
```
tonyandrewmeyer: I know this is copied from the existing version, but perhaps we could improve it at the same time.

I find it odd to create a dict inside of a dict expression, which has None values that will get ignored by the expression, and tricky to read for the same reason - it's important to match up the "if value is not None" with the internal dict's None default. In practice, that doesn't happen: you get all the keys or none of them, and you don't get None values from db_data - but this is written as if you do.

Something like this maybe, although there are probably tidier ways to do this:

```python
key_to_env = {'db_host': 'DEMO_SERVER_DB_HOST', ...}
env = {key_to_env[key] for key, value in db_data}
```

I think it makes it clearer that we are just transforming the data about the DB into the same data for the environment.

It may not be worth it for a tutorial, but `fetch_postgres_relation_data` could return a typed dict that made it clearer that you always get those keys.

tonyandrewmeyer (approved):
Thanks for doing this! I agree the flow makes more sense now, and the renames are good.

I agree the simple text is clearer than the diagram being removed. I wonder if (later, in a different PR) some sort of diagrammatic summary at the end of the tutorial might be useful for people - often the event flow is challenging to understand and I think that's where a diagram can best help.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: Good suggestion - I've gone with removing postgres entirely, here and in other places.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: If we're going for clarity, how about this?

```python
db_data = self.fetch_database_relation_data()
if not db_data:
    return {}
return {
    'DEMO_SERVER_DB_HOST': db_data['db_host'],
    'DEMO_SERVER_DB_PORT': db_data['db_port'],
    'DEMO_SERVER_DB_USER': db_data['db_username'],
    'DEMO_SERVER_DB_PASSWORD': db_data['db_password'],
}
```

PR #2266, opened 2026-01-07
fix: clean up temp dir to avoid ResourceWarning from testing.Context

### LINKED ISSUE #2227: Scenario Context objects do not explicitly clean up the temporary directory used for containers and storages
The Scenario Context leaves behind a temporary directory:

```python
>>> import os
>>> import ops
>>> from ops import testing
>>> t = os.listdir("/tmp")
>>> ctx = testing.Context(ops.CharmBase, meta={"name": "test"})
>>> _ = ctx.run(ctx.on.start(), testing.State())
>>> set(os.listdir("/tmp")) - set(t)
{'tmpcd0o54a7'}
>>> os.listdir('/tmp/tmpcd0o54a7')
[]
```

This is the [`Context._tmp`](https://github.com/canonical/operator/blob/e7238c9b8c419bdd0e8de561d0c33e800267e8d7/testing/src/scenario/context.py#L712) directory. It's used for the container root and storage root (which also means that it's shared across `run` calls with the same Context, which is probably not desirable but perhaps is now baked in and not changeable?), so it's available outside of the `run` or context manager use.

Note that the directory doesn't exist when the process exits because Python cleans it up (emitting a `ResourceWarning`):

```python
$ python3 -W error -c 'import tempfile;t=tempfile.TemporaryDirectory()'
  File "/usr/lib/python3.12/weakref.py", line 666, in _exitfunc
    f()
  File "/usr/lib/python3.12/weakref.py", line 590, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/tempfile.py", line 1075, in _cleanup
    _warnings.warn(warn_message, ResourceWarning)
ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/tmp83ywjo1j'>
```

xdqi:
This pull request explicitly clean up the `TemporaryDirectory` instead of relying on `weakref.finalize` which is the default behaviour of `TemporaryDirectory`. `weakref.finalize` does the cleanup either on GC collection or when `atexit`, on both occasion, there will be a warning occurred from the interpreter.

Fixes #2227

### DIFF
--- a/testing/src/scenario/context.py
+++ b/testing/src/scenario/context.py
@@ -759,9 +759,7 @@ class Context(Generic[CharmType]):
 
     def __del__(self):
         """Clean up the temporary directory."""
-        tmp = getattr(self, '_tmp', None)
-        if tmp:
-            shutil.rmtree(tmp, ignore_errors=True)
+        shutil.rmtree(self._tmp, ignore_errors=True)
 
     def _get_container_root(self, container_name: str):
         """Get the path to a tempdir where this container's simulated root will live."""


xdqi:
Changed `TemporaryDirectory` to old-style `mkdtemp` for Python compatibility.

testing/src/scenario/context.py
```diff
@@ -711,7 +712,7 @@ def __init__(
         self._availability_zone = availability_zone
         self._principal_unit = principal_unit
         self.app_trusted = app_trusted
-        self._tmp = tempfile.TemporaryDirectory()
+        self._tmp = tempfile.mkdtemp()
```
benhoyt, line 716: It'd be good to add a comment here explaining why we're using mkdtemp instead of TemporaryDirectory. Perhaps:
```suggestion
        # use mkdtemp as TemporaryDirectory only added "delete=False" feature in Python 3.12
        self._tmp = tempfile.mkdtemp()
```

testing/tests/test_context.py
```diff
@@ -188,3 +188,40 @@ def test_charm_spec_is_deprecated():
...
+    """Test that the Context.__del__ properly cleans up the temporary directory.
+
+    This test verifies that:
+    1. The temp directory is created and exists after Context creation
+    2. The temp directory persists after run() calls
+    3. No ResourceWarning is triggered when the context is garbage collected
+    4. The temp directory is cleaned up after __del__ is called
+    """
+    import pathlib
+    import warnings
```
benhoyt: Imports should be at the top of the file with the other imports.

testing/tests/test_context.py
```diff
@@ -188,3 +188,40 @@ def test_charm_spec_is_deprecated():
...
+
+    ctx = Context(MyCharm, meta={'name': 'foo'})
+
+    # Get the temporary directory path
```
benhoyt: Most of these comments are not necessary, they just say exactly what the line of code says. In addition, because `ctx._tmp` is just a path string now, it's probably better to avoid pathlib and just use `assert os.path.exists(ctx._tmp)`.

testing/tests/test_context.py
```diff
@@ -188,3 +188,40 @@ def test_charm_spec_is_deprecated():
...
+
+    # Run a scenario to ensure the temp dir is used and persists after run()
+    ctx.run(ctx.on.start(), State())
+
+    # Temp directory should still exist after run()
+    assert tmp_path.exists()
+
+    # Call __del__ directly because cyclic references prevent garbage collection
+    # from calling it reliably.
+    with warnings.catch_warnings():
```
benhoyt: This doesn't actually test what we want (or what the function docstring says). What we want is to capture the warnings and ensure there's no ResourceWarning. I think you'd do this using `with warnings.catch_warnings(record=True) as warns` and then assert that that was empty.

However, there's another problem. Even with the old TemporaryDirectory code, the ResourceWarning isn't raised here -- so it's not testing what we think.

I've played around with it a bit and it's actually hard to get the ResourceWarning to be triggered. So maybe it's not worth adding an automated test for this after all. I've confirmed manually with this script that a ResourceWarning is raised before the fix, but not after:

```python
# Run with:
# uv run --with-editable ./testing python -W error t.py

import os
import ops
from ops import testing
t = os.listdir("/tmp")
ctx = testing.Context(ops.CharmBase, meta={"name": "test"})
ctx.run(ctx.on.start(), testing.State())
diff = set(os.listdir("/tmp")) - set(t)
tmpname = list(diff)[0]
print(tmpname)
print(os.listdir('/tmp/' + tmpname))
```

benhoyt (changes requested):
Thanks for this. Added a couple of comments.

The main thing is that the test isn't doing what you think it is. And it might actually be hard to do that. So I'm now thinking we should just remove the test for this (I've confirmed it works manually).

testing/src/scenario/context.py
```diff
@@ -755,13 +756,17 @@ def _set_output_state(self, output_state: State):
         """Hook for Runtime to set the output state."""
         self._output_state = output_state
 
+    def __del__(self):
+        """Clean up the temporary directory."""
+        shutil.rmtree(self._tmp)
```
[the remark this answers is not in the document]
benhoyt: @xdqi I don't think we should do the getattr and `if tmp` logic, but I do think adding `ignore_errors=True` is reasonable here for a `__del__` cleanup operation.

testing/tests/test_context.py
```diff
@@ -188,3 +188,40 @@ def test_charm_spec_is_deprecated():
...
+        warnings.simplefilter('error', ResourceWarning)
+
+        ctx.__del__()
```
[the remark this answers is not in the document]
benhoyt: We're probably going to remove this test (see above), so resolving this.

testing/src/scenario/context.py
```diff
@@ -755,13 +757,19 @@ def _set_output_state(self, output_state: State):
         """Hook for Runtime to set the output state."""
         self._output_state = output_state
 
+    def __del__(self):
+        """Clean up the temporary directory."""
+        tmp = getattr(self, '_tmp', None)
+        if tmp:
+            shutil.rmtree(tmp, ignore_errors=True)
+
```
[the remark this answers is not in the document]
xdqi: I don't think it's a good idea to either add a cleanup function, or setting `_tmp` to `None`.
Both will make the whole Context unusable again, easily cause problems like use-after-free.

testing/src/scenario/context.py
```diff
@@ -755,13 +757,19 @@ def _set_output_state(self, output_state: State):
         """Hook for Runtime to set the output state."""
         self._output_state = output_state
 
+    def __del__(self):
+        """Clean up the temporary directory."""
+        tmp = getattr(self, '_tmp', None)
+        if tmp:
+            shutil.rmtree(tmp, ignore_errors=True)
```
benhoyt: Note my previous comment:

> I don't think we should do the getattr and `if tmp` logic, but I do think adding ignore_errors=True is reasonable here for a __del__ cleanup operation.

So let's not do the getattr / `if tmp` part, as `self._tmp` is always set.
```suggestion
        shutil.rmtree(self._tmp, ignore_errors=True)
```

testing/src/scenario/context.py
```diff
@@ -711,7 +712,8 @@ def __init__(
         self._availability_zone = availability_zone
         self._principal_unit = principal_unit
         self.app_trusted = app_trusted
-        self._tmp = tempfile.TemporaryDirectory()
+        # use mkdtemp as TemporaryDirectory only added "delete=False" feature in Python 3.12
+        self._tmp = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
xdqi: You're right but we don't like warnings.

tonyandrewmeyer (changes requested):
Hi @xdqi - thanks for working on this, and my apologies for being slow at getting to the review.

This errors out as it currently is:

```
>>> import ops
>>> from ops import testing
>>> ctx = testing.Context(ops.CharmBase, meta={"name": "foo"})
>>> 
Exception ignored while calling deallocator <function Context.__del__ at 0x73e54cc55f30>:
  File "/tmp/xdqi-operator/testing/src/scenario/context.py", line 762, in __del__
TypeError: 'NoneType' object is not callable
```

(At the empty line I used control-D to exit).

I believe this is because during shutdown `shutil` can be set to `None` before the `Context` object's `__del__` is called. You can avoid the error by adding a simple `if shutil is not None`, but then (of course) you're not actually cleaning up any more.

I think solving this issue is actually quite hard to do without making API changes (like introducing a "cleanup" method on the Context), maybe not actually feasible at all.

If you want to explore this issue further and you'd like to chat about it (in a call or ticket or in the Charm Tech channel in Matrix) let me know.

xdqi:
I cannot replicate issue by @tonyandrewmeyer at my side:
```
~/canonical/operator$ source ~/canonical/operator/.venv/bin/activate
(ops) ~/canonical/operator$ python3
Python 3.12.3 (main, Jan  8 2026, 11:30:50) [GCC 13.3.0] on linux
Ctrl click to launch VS Code Native REPL
>>> import ops
>>> from ops import testing
>>> testing
<module 'ops.testing' from '~/canonical/operator/ops/testing.py'>
>>> ctx = testing.Context(ops.CharmBase, meta={"name": "foo"})
>>> ^D
(ops) ~/canonical/operator$ 

```

tonyandrewmeyer:
Looks like it's reliably reproducible with 3.14:

```
$ uvx --python=3.10 --with=.[testing] python -c 'import ops;from ops import testing;ctx=testing.Context(ops.CharmBase, meta={"name": "foo"})'
$ uvx --python=3.12 --with=.[testing] python -c 'import ops;from ops import testing;ctx=testing.Context(ops.CharmBase, meta={"name": "foo"})'
$ uvx --python=3.14 --with=.[testing] python -c 'import ops;from ops import testing;ctx=testing.Context(ops.CharmBase, meta={"name": "foo"})'
Exception ignored while calling deallocator <function Context.__del__ at 0x7bc2ea34a350>:
  File "/tmp/xdqi-operator/testing/src/scenario/context.py", line 762, in __del__
TypeError: 'NoneType' object is not callable
```

But, essentially, you can't really rely on things being available in `__del__`; there aren't guarantees about order as far as I can recall.

benhoyt:
The Charm Tech team spoke about this, and we decided it wasn't worth fixing this. There's no easy way to fix the ResourceWarning without an explicit cleanup API or with statement, which we're not going to add now. So closing won't-do. Sorry @xdqi, but at least you helped us close something! :-)

PR #2268, opened 2026-01-08
fix: prevent `KeyError` on `auth-type` when creating `CloudCredential` object

### LINKED ISSUE #2267: `model.get_cloud_spec()` fails with `KeyError: 'auth-type'`
Hey Team, I'm trying to use `model.get_cloud_spec()` to differentiate between different cloud types within a charm. Even though deployed with `--trust`, I still run into an error:

```
  File "/var/lib/juju/agents/unit-valkey-k8s-1/charm/src/charm.py", line 49, in _on_pebble_ready
    cloud_spec = self.model.get_cloud_spec()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-valkey-k8s-1/charm/venv/lib/python3.12/site-packages/ops/model.py", line 331, in get_cloud_spec
    return self._backend.credential_get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-valkey-k8s-1/charm/venv/lib/python3.12/site-packages/ops/model.py", line 4077, in credential_get
    return CloudSpec.from_dict(dataclasses.asdict(raw_spec))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-valkey-k8s-1/charm/venv/lib/python3.12/site-packages/ops/model.py", line 4283, in from_dict
    credential=CloudCredential.from_dict(d['credential']) if d.get('credential') else None,
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-valkey-k8s-1/charm/venv/lib/python3.12/site-packages/ops/model.py", line 4233, in from_dict
    auth_type=d['auth-type'],
              ~^^^^^^^^^^^^^
KeyError: 'auth-type'
```

I've reproduced this on microk8s and LXD controllers, both with Juju agent versions 3.6.12 and 3.6.9 (LXD only in this case).

The cloud bootstrap is creating a microk8s on [localhost](http://localhost/). If I run juju show-cloud microk8s I get a valid result:
```
$ juju show-cloud microk8s
Cloud "microk8s" from controller "k8s-controller":

defined: public
type: k8s
auth-types: [certificate, clientcertificate, oauth2, oauth2withcert, userpass]
endpoint: https://127.0.0.1:16443
credential-count: 1
regions:
  localhost: {}
```

reneradoi:
Fixes #2267

When the hook command `credential-get` is executed as part of ops' `model.get_cloud_spec()`, it currently fails with `KeyError` (see issue linked above).

This PR drops the use of `dataclasses.asdict` to get a Juju-style dictionary (to then convert into ops.model classes), since that doesn't undo the `-` to `_` that happened when making the data classes. Instead, new private methods to convert from a hookcmds class are added and used.

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -3664,7 +3664,7 @@ class _ModelBackend:
     def relation_model_get(self, relation_id: int) -> dict[str, Any]:
         with self._wrap_hookcmd('relation-model-get', relation_id=relation_id):
             raw = hookcmds.relation_model_get(relation_id)
-        return {"uuid": raw.uuid}
+        return {'uuid': raw.uuid}
 
     def config_get(self) -> dict[str, bool | int | float | str]:
         with self._wrap_hookcmd('config-get'):


test/test_model.py
```diff
@@ -4543,6 +4543,15 @@ def test_from_dict_full(self):
         assert cloud_cred.attributes == d['attrs']
         assert cloud_cred.redacted == d['redacted']
 
+    def test_from_dict_empty(self):
```
benhoyt: The test name indicates it's an empty dict, but the dict input has "redacted". Can we just pass in `{}` here to `from_dict`?

benhoyt:
Also adding Tony as a reviewer (for when he's back next week).

benhoyt:
I'd also like to understand why `auth-type` is not coming through. See [my comment an the issue](https://github.com/canonical/operator/issues/2267#issuecomment-3725699199).

test/test_model.py
reneradoi: Good point, I've changed the test name.

ops/hookcmds/_types.py
```diff
@@ -116,7 +116,7 @@ class CloudCredential:
     def _from_dict(cls, d: dict[str, Any]) -> CloudCredential:
         """Create a new CloudCredential object from a dictionary."""
         return cls(
-            auth_type=d['auth-type'],
+            auth_type=d['auth_type'],
```
benhoyt: With your new finding of auth_type vs auth-type and the asdict problem, this isn't a correct fix. See my comment at https://github.com/canonical/operator/issues/2267#issuecomment-3739608343 (though we can take it from here).

tonyandrewmeyer:
> Also adding Tony as a reviewer (for when he's back next week).

I've instead made adjustments as outlined in #2267 and our discussion earlier today (and I've removed myself as a reviewer since I'm now more of a co-author).

benhoyt (approved):
Looks good, thanks!

reneradoi (commented):
Thank you @tonyandrewmeyer these changes of course make much more sense, looks good from my point of view!

james-garner-canonical (commented):
The hyphens-to-underscores issue does prevent us from using `dataclasses.asdict` out of the box in this error case, and I definitely agree with switching away from it preemptively even in the case where none of the keys are hyphenated.

I'm wonder if going all the way to hardcoding the keys and values on the model side is the best approach.

Might it be cleaner to add a small function that fixes our mistaken assumption about the symmetry of `asdict` and `_from_dict`? This could be used as a drop-in replacement for `dataclasses.asdict` in `model.py`.
```py
def as_juju_dict(cls: Any) -> dict[str, Any]:
    return {k.replace('_', '-'): v for k, v in dataclasses.asdict(cls)}
```
This does add some public API surface with not-particularly-helpful typing. But it could help future users from shooting themselves in the foot the same way we did, and means we wouldn't need to add/maintain a third constructor for these model classes.

tonyandrewmeyer:
> I'm wonder if going all the way to hardcoding the keys and values on the model side is the best approach.
> 
> Might it be cleaner to add a small function that fixes our mistaken assumption about the symmetry of `asdict` and `_from_dict`? This could be used as a drop-in replacement for `dataclasses.asdict` in `model.py`.
> 
> ```python
> def as_juju_dict(cls: Any) -> dict[str, Any]:
>     return {k.replace('_', '-'): v for k, v in dataclasses.asdict(cls)}
> ```
> 
> This does add some public API surface with not-particularly-helpful typing. But it could help future users from shooting themselves in the foot the same way we did, and means we wouldn't need to add/maintain a third constructor for these model classes.

It's a bit messy because we retroactively inserted hookcmds into things. Really, there shouldn't be methods to create the model objects from Juju dicts, only from hookcmds objects, but we can't remove them since they're public.

We already have other examples where we create the object with explicit arguments (Port, SecretInfo) so the PR as-is is more consistent, plus one (Status) that has a `from_name` that's more like this than `from_dict`. The other backend methods that return something return stdlib types, mostly staying that way (with relation being a complicated exception).

I think secret-info-get is a good example of why we don't want a basic underscore-to-dash, since we change more than just that when converting from the low-level hookcmds class (and Juju structure) to the ops class.

I assume the users you're referring to are ones building things on top of hookcmds. I hope that in those cases people are not doing what ops does and making copies of the classes, and are instead building more interesting, different, exposure on top, or exposing the raw hookcmds objects.

So I'd definitely prefer to have the two explicit, slightly verbose, constructions instead of a overly-simple helper.

tonyandrewmeyer:
Sorry @dimaqq I misclicked on the re-review request next to your name when I was aiming for James.

james-garner-canonical (approved):
Makes sense, thanks for explaining, @tonyandrewmeyer

Issue #2269, opened 2026-01-14
_checks_action in pebble shouldn't return None


gabrielcocenza:
Currently the function `_checks_action` in pebble.yaml can return `None` instead of empty list and this makes the function not `idempotent` meaning that after stopping a check, the code fails if the function runs again after the check is stopped .

See the trackback:

```
2026-01-14T17:16:52.566Z [container-agent] 2026-01-14 17:16:52 ERROR juju-log Uncaught exception while in charm code:
2026-01-14T17:16:52.566Z [container-agent] Traceback (most recent call last):
2026-01-14T17:16:52.566Z [container-agent]   File "/var/lib/juju/agents/unit-keystone-0/charm/./src/charm.py", line 2847, in <module>
2026-01-14T17:16:52.566Z [container-agent]     ops.main(KeystoneOperatorCharm)
2026-01-14T17:16:52.566Z [container-agent]   File "/var/lib/juju/agents/unit-keystone-0/charm/venv/ops/__init__.py", line 360, in __call__
2026-01-14T17:16:52.566Z [container-agent]     return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
2026-01-14T17:16:52.566Z [container-agent]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-01-14T17:16:52.566Z [container-agent]   File "/var/lib/juju/agents/unit-keystone-0/charm/venv/ops/_main.py", line 504, in main
2026-01-14T17:16:52.566Z [container-agent]     manager.run()
2026-01-14T17:16:52.566Z [container-agent]   File "/var/lib/juju/agents/unit-keystone-0/charm/venv/ops/_main.py", line 488, in run
2026-01-14T17:16:52.566Z [container-agent]     self._emit()
2026-01-14T17:16:52.566Z [container-agent]   File "/var/lib/juju/agents/unit-keystone-0/charm/venv/ops/_main.py", line 423, in _emit
2026-01-14T17:16:52.566Z [container-agent]     self._emit_charm_event(self.dispatcher.event_name)
2026-01-14T17:16:52.566Z [container-agent]   File "/var/lib/juju/agents/unit-keystone-0/charm/venv/ops/_main.py", line 467, in _emit_charm_event
[...]

PR #2270, opened 2026-01-14
fix: _checks_action should return empty list when there are no changes

gabrielcocenza:
_checks_action cannot return None
Fix: #2269

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2531,6 +2531,9 @@ class Client:
             span.set_attribute('checks', checks)
             body = {'action': action, 'checks': checks}
             resp = self._request('POST', '/v1/checks', body=body)
+            # Pebble may return `changed: null` if nothing has been stopped or started.
+            # Remove this crutch when Jujus that include affected Pebble have reached EOL.
+            # https://github.com/canonical/pebble/issues/788
             return resp['result']['changed'] or []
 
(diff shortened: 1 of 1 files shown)


gabrielcocenza:
@tonyandrewmeyer I've added a unit test. Let me know if that is enough

ops/pebble.py
```diff
@@ -2531,7 +2531,7 @@ def _checks_action(self, action: str, checks: Iterable[str]) -> list[str]:
             span.set_attribute('checks', checks)
             body = {'action': action, 'checks': checks}
             resp = self._request('POST', '/v1/checks', body=body)
-            return resp['result']['changed']
+            return resp['result']['changed'] or []
```
dimaqq, line 2537:
```suggestion
            # Pebble may return `changed: null` if nothing has been stopped or started.
            # Remove this crutch when Jujus that include affected Pebble have reached EOL.
            # https://github.com/canonical/pebble/issues/788
            return resp['result']['changed'] or []
```

dimaqq (approved):

tonyandrewmeyer (approved):
I tweaked the test a little so that it's testing the public method rather than the private one.

Looks good to me - I agree an explanatory comment along the lines of Dima's suggestion would be good.

Thanks again!

gabrielcocenza:
Thanks @tonyandrewmeyer @dimaqq !

I've committed the suggested message.

PR #2271, opened 2026-01-14
docs: secrets over CMR comment added

Abuelodelanada:
This PR adds a comment in the Secrets How-To to reflect [this comment](https://github.com/juju/juju/issues/21248#issuecomment-3564906360).

### DIFF
--- a/docs/howto/manage-secrets.md
+++ b/docs/howto/manage-secrets.md
@@ -53,7 +53,8 @@ Note that:
 - We call `add_secret` on `self.app` (the application). That is because we want the secret to be owned by this application, not by this unit. If we wanted to create a secret owned by the unit, we'd call `self.unit.add_secret` instead.
 - The only data shared in plain text is the secret ID (a locator URI). The secret ID can be publicly shared. Juju will ensure that only remote apps/units to which the secret has explicitly been granted by the owner will be able to fetch the actual secret payload from that ID.
 - The secret needs to be granted to a remote entity (app or unit), and that always goes via a relation instance. By passing a relation to `grant` (in this case the event's relation), we are explicitly declaring the scope of the secret -- its lifetime will be bound to that of this relation instance.
-- In `Cross Model Relation` (`CMR`) scenario, the offering side of the relation is the one that can share secrets. Juju does not support `apps` on the consuming side sharing secrets.
+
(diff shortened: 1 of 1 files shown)


docs/howto/manage-secrets.md
```diff
@@ -53,6 +53,7 @@ Note that:
 - We call `add_secret` on `self.app` (the application). That is because we want the secret to be owned by this application, not by this unit. If we wanted to create a secret owned by the unit, we'd call `self.unit.add_secret` instead.
 - The only data shared in plain text is the secret ID (a locator URI). The secret ID can be publicly shared. Juju will ensure that only remote apps/units to which the secret has explicitly been granted by the owner will be able to fetch the actual secret payload from that ID.
 - The secret needs to be granted to a remote entity (app or unit), and that always goes via a relation instance. By passing a relation to `grant` (in this case the event's relation), we are explicitly declaring the scope of the secret -- its lifetime will be bound to that of this relation instance.
+- In `Cross Model Relation` (`CMR`) scenario, the offering side of the relation is the one that can share secrets. Juju does not support `apps` on the consuming side sharing secrets.
```
dwilding: Since this note doesn't directly relate to the code above, I recommend moving it to a separate paragraph. I'd also suggest adjusting the wording to match other places where we talk about cross-model relations:

```suggestion

If the relation is a cross-model relation, Juju only allows the provider application to grant access to secrets.
```

dimaqq: This reads better.

At a high level, does the unit even know whether its side of the relation is a provider or a consumer in the cross-model relation case? Perhaps this needs to be highlighted more prominently, because charm authors would have to leave instructions for Juju users along the lines of "if you're going to integrate on this endpoint to an app in a different Juju model, make sure that this (or the other) app is what offers the endpoint".

dimaqq: On the second thought, I think we're conflating provides/requires (charmcraft.yaml declaration) with offer/consume (Juju cross-model relation setup commands).

Let's change this to "the offering application".

dwilding: Sounds good to me, thanks Dima. So the change would be:

```suggestion

If the relation is a cross-model relation, Juju only allows the offering application to grant access to secrets.
```

Abuelodelanada:
@dwilding @dimaqq  Thank you very much for your suggestions!

Issue #1790, opened 2025-06-05
Passing check level from layer to CheckInfo causes type error


dwilding:
The testing code:

```python
ctx = testing.Context(MyCharm, meta={"name": "foo", "containers": {"my_container": {}}})
layer = pebble.Layer({
    "checks": {"http-check": {"override": "replace", "startup": "enabled", "failures": 7}},
})
check_info = testing.CheckInfo(
    "http-check",
    status=pebble.CheckStatus.UP,
    level=layer.checks["http-check"].level,
    startup=layer.checks["http-check"].startup,
    threshold=layer.checks["http-check"].threshold,
)
container = testing.Container("my_container", check_infos={check_info}, layers={"layer1": layer})
state = testing.State(containers={container})
ctx.run(ctx.on.pebble_check_failed(info=check_info, container=container), state=state)
```

produces an error from the type checker:

```
error: Argument of type "CheckLevel | str" cannot be assigned to parameter "level" of type "CheckLevel | None" in function "__init__"
```

It's possible to work around the issue by changing `level=layer.checks["http-check"].level` to `level=pebble.CheckLevel(layer.checks["http-check"].level)` (or directly hard-coding `level=pebble.CheckLevel.READY`).

The proposal to fix this in `ops.testing` is:

> change `testing.CheckInfo` so that `level` (maybe some of the others?) accepts `str` as well but probably converts to a `CheckLevel` in `__post_init__`

See https://github.com/canonical/charmcraft/pull/2309#issuecomment-2942189138

tonyandrewmeyer:
Other suggestions from the discussion in daily:
* Provide defaults in `testing.CheckInfo` matching the Pebble ones, and then be explicit duplicating the value if using something else than the default.
* Change the docs so that we don't recommend getting the value from the layer like Pebble does but instead just always hard-code/duplicate the values.
* Add some sort of `testing.CheckInfo.from_layer(failures)` method that would give back an appropriate object with all the values copied from the layer or computed (status) or passed in (failures).
* Change the docs to use the workaround from the description.

PR #2274, opened 2026-01-16
fix: make testing.CheckInfo level arg type match pebble.CheckInfo.level

james-garner-canonical:
This PR adds a custom `__init__` method to `testing.CheckInfo` to allow the `level` argument to accept a `str` (which is then converted to a `pebble.CheckInfo` before setting the `testing.CheckInfo.level` attribute). This allows the `level` argument to be a `pebble.CheckInfo.level` (e.g. pulled from a `pebble.Layer`) without the type checker complaining.

---

Original PR description:

The signature of `testing.CheckInfo` doesn't quite match that of `pebble.CheckInfo`. While `pebble.CheckInfo` will accept a `str` for `level`, the `testing` version only accepts a `pebble.CheckLevel | None`. This is a nicer experience for users of the `testing.CheckInfo.level` *attribute*, but leads to a type error when passing a `pebble.Check.level` to the `testing.CheckInfo` *constructor*. This was reported in #1790.

I believe the correct solution is to broaden the types accepted by the `testing.CheckInfo` constructor without broadening the corresponding `level` attribute. This PR does so by setting `init=False` and adding a custom `__init__` method to `testing.CheckInfo`. This makes the argument types match and resolves this type checking error, while retaining the nicer attribute typing.

While investigating this, I noticed that there is a similar issue with `pebble.Check.threshold` being `None`-able, while the `threshold` argument to `testing.CheckInfo` is not. This differs from the `level` case in that `pebble.CheckInfo` does not accept `None` for the `threshold` argument. ~Currently this PR 'fixes' this for `testing.CheckInfo` by again broadening the types accepted by the constructor, while keeping the attribute the same, however this introduces a new mismatch between the `pebble` and `testing` versions of this class. Perhaps this enhancement should be dropped from this PR, or perhaps the PR should also include applying it to `pebble.CheckInfo`.~ This was not addressed in this PR.

Fixes: #1790

---

The file inlined below (adapted from the example in the issue) can be used to validate these changes -- perhaps it's worth adding similar constructs to one of our tests?

```py
# Copyright 2026 Canonical Ltd.
# Add this file as e.g. _tmp.py and validate that type checking continues to pass

import ops
from ops import pebble, testing


def _(my_charm_type: type[ops.CharmBase]):
    ctx = testing.Context(my_charm_type, meta={'name': 'foo', 'containers': {'my_container': {}}})
    layer = pebble.Layer({
        'checks': {'http-check': {'override': 'replace', 'startup': 'enabled', 'threshold': 7}},
    })
    check_info = testing.CheckInfo(
        'http-check',
        status=pebble.CheckStatus.UP,
        level=layer.checks['http-check'].level,
        startup=layer.checks['http-check'].startup,
        threshold=layer.checks['http-check'].threshold,
    )
        'my_container', check_infos={check_info}, layers={'layer1': layer}
    )
    state = testing.State(containers={container})
    ctx.run(ctx.on.pebble_check_failed(info=check_info, container=container), state=state)
```

On `main` we get these errors:
```
_tmp.py:15:15 - error: Argument of type "CheckLevel | str" cannot be assigned to parameter "level" of type "CheckLevel | None" in function "__init__"
  Type "CheckLevel | str" is not assignable to type "CheckLevel | None"
    Type "str" is not assignable to type "CheckLevel | None"
      "str" is not assignable to "CheckLevel"
      "str" is not assignable to "None" (reportArgumentType)
tmp.py:17:19 - error: Argument of type "int | None" cannot be assigned to parameter "threshold" of type "int" in function "__init__"
  Type "int | None" is not assignable to type "int"
    "None" is not assignable to "int" (reportArgumentType)
```

### DIFF
--- a/testing/tests/test_check_info.py
+++ b/testing/tests/test_check_info.py
@@ -1,0 +1,54 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import pytest
+from scenario.state import CheckInfo
+
+import ops
+
+
+def test_check_info_defaults():
+    """The custom __init__ sets the expected default values."""
+    check_info = CheckInfo(name='Mairon')
+    assert check_info.name == 'Mairon'
+    assert check_info.level is None
+    assert check_info.startup is ops.pebble.CheckStartup.ENABLED
+    assert check_info.status is ops.pebble.CheckStatus.UP
+    assert check_info.successes == 0
+    assert check_info.failures == 0
+    assert check_info.threshold == 3
+    assert check_info.change_id is not None
+    assert isinstance(check_info.change_id, ops.pebble.ChangeID)
+
+
+@pytest.mark.parametrize(('level'), ['', 'alive', 'ready'])
+def test_check_info_good_level_converted(level: str):
+    """A valid level string is converted to the CheckLevel enum."""
+    check_info = CheckInfo(name='Fëanor', level=level)
+    assert check_info.level is ops.pebble.CheckLevel(level)
+
+
+def test_check_info_bad_level():
+    """An invalid level value raises a ValueError, and the type checker complains."""
+    with pytest.raises(ValueError):
+        CheckInfo(name='Melkor', level=1)  # type: ignore
+
+
+def test_check_info_bad_startup_and_status():
+    """Bad startup and status values are allowed at runtime, but the type checker complains."""
+    check_info = CheckInfo(
+        name='Maedhros',
+        startup='bad',  # type: ignore
+        status='worse',  # type: ignore
+    )
+    assert check_info.startup == 'bad'
+    assert check_info.status == 'worse'
+
+
+def test_check_info_threshold_none():
+    """A bad threshold value of None is allowed at runtime, but the type checker complains."""
+    check_info = CheckInfo(
+        name='Fingon',
+        threshold=None,  # type: ignore
+    )
+    assert check_info.threshold is None


testing/src/scenario/state.py
```diff
@@ -897,22 +897,20 @@ def _to_ops(self) -> pebble.Notice:
...
 class CheckInfo:
     """A health check for a Pebble workload container."""
 
     name: str
     """Name of the check."""
 
-    _: dataclasses.KW_ONLY
-
-    level: pebble.CheckLevel | None = None
+    level: pebble.CheckLevel | None
```
james-garner-canonical, line 907: With `init=False`, these are now only attribute annotations, so the default values can go, as they're redundant with those defined in the `__init__` method.

testing/src/scenario/state.py
```diff
@@ -922,36 +920,61 @@ class CheckInfo:
...
     :attr:`ops.pebble.CheckStatus.UP` or :attr:`ops.pebble.CheckStatus.DOWN`.
     """
 
-    def __post_init__(self):
-        if self.change_id is None:
+    def __init__(
+        self,
+        name: str,
+        *,
+        level: pebble.CheckLevel | str | None = None,
```
james-garner-canonical, line 950: This is the key change from previously -- we now accept `str` as well.

testing/src/scenario/state.py
```diff
@@ -922,36 +920,61 @@ class CheckInfo:
...
+        threshold: int | None = 3,
+        change_id: pebble.ChangeID | None = None,
+    ):
+        object.__setattr__(self, 'name', name)
+        if level is not None:
+            try:
+                level = pebble.CheckLevel(level)
+            except ValueError:
+                # preserve the existing behaviour of not crashing on arbitrary strings
+                logger.error('%s is not a valid Pebble check level.', level)
```
james-garner-canonical: To avoid introducing a new runtime error, we allow `level` strings that don't actually match `CheckLevel` -- but maybe we could be more aggressive here and require it to match. This would be a breaking change.

testing/src/scenario/state.py
```diff
@@ -922,36 +920,61 @@ class CheckInfo:
...
+        object.__setattr__(self, 'level', level)
+        object.__setattr__(self, 'startup', startup)
+        object.__setattr__(self, 'status', status)
+        # FIXME: is it safe to narrow the `successes` attribute type to `int` and convert it here?
+        object.__setattr__(self, 'successes', successes)
```
james-garner-canonical: It's odd that `successes` is `None`- able and `failures` isn't. We could retain the current constructor signature but narrow the attribute type, which should be pretty safe.

Or maybe `failures` should accept `None` as well?

testing/src/scenario/state.py
```diff
@@ -922,36 +920,61 @@ class CheckInfo:
...
+    def __init__(
+        self,
+        name: str,
+        *,
+        level: pebble.CheckLevel | str | None = None,
+        startup: pebble.CheckStartup = pebble.CheckStartup.ENABLED,
+        status: pebble.CheckStatus = pebble.CheckStatus.UP,
+        successes: int | None = 0,
+        failures: int = 0,
+        threshold: int | None = 3,
```
james-garner-canonical: This is also a change from the previous behaviour, as noted in the PR description -- accepting `None` (and converting it to 0 below). Perhaps it should be dropped from this PR, or perhaps `pebble.CheckInfo` should be updated to match, or perhaps it's fine as-is, I'm not sure.

benhoyt: Do we know that converting None (from a `pebble.CheckInfo`) to 0 is the correct move, from a Pebble point of view?

testing/src/scenario/state.py
```diff
@@ -922,36 +920,61 @@ class CheckInfo:
...
+        change_id: pebble.ChangeID | None = None,
+    ):
+        object.__setattr__(self, 'name', name)
```
benhoyt, line 958: I'm having a thinko -- why do we need `object.__setattr__` here, instead of just `self.foo = foo`?

james-garner-canonical: Because this `dataclass` has `frozen=True` -- the `__init__` method that `dataclasses` would generate with `init=True` would also use `object.__setattr__`, and we do the same in `__post_init__`.

testing/src/scenario/state.py
james-garner-canonical: Looking at the [pebble source code](https://github.com/canonical/pebble/blob/f7589e55ec596a3c69d342a9e5dfca751dc5db4b/client/checks.go#L98), `successes` is treated differently from `failures` there too (may be `nil`), so we shouldn't change that here.

testing/src/scenario/state.py
james-garner-canonical: `threshold` apparently being able to be `None` originates from [pebble.Check](https://github.com/canonical/operator/blob/a60c93a32742e16ed4ce3fea0678f25b5c0b4c2b/ops/pebble.py#L1111) (and [pebble.CheckDict](https://github.com/canonical/operator/blob/a60c93a32742e16ed4ce3fea0678f25b5c0b4c2b/ops/pebble.py#L137)). I assumed that `None` and `0` should be treated equivalently because [when we turn a Check into a dict](https://github.com/canonical/operator/blob/a60c93a32742e16ed4ce3fea0678f25b5c0b4c2b/ops/pebble.py#L1128) we skip all fields with `False`-y values.

On the `pebble` side, [threshold can't be nil](https://github.com/canonical/pebble/blob/f7589e55ec596a3c69d342a9e5dfca751dc5db4b/client/checks.go#L106). IIRC Pebble always uses 'omit empty' when sending JSON, so it would only be missing (and thus `None` on the Python side) if it was set to 0. However, the default threshold is 3, and it doesn't make any sense for the threshold to be 0 (health check fails if there are 0 failures in a row), so I'm guessing the zero-so-omit case never comes up in the real world?

This makes it seem like the right thing is to make `pebble.Check.threshold` not be `None`-able, which is a bit of a bigger question, so maybe it's best to leave changes to `testing.CheckInfo.threshold` out of this PR.

james-garner-canonical:
> I definitely think it'd be a good idea to add a test of this new behaviour / typing.

Will do.


> A thought on a possible way to simplify: could you define a `_CheckInfo` dataclass with most of the fields, and then inherit `CheckInfo` from that to add the special fields, calling the super()'s `__init__` so you only need to special case the special fields?

I think that would work, but I'm not convinced it's truly simpler.

TBH I think we should be moving in the direction of having a custom `__init__` for more of the Scenario `dataclasses`, for better type checking and auto-completion UX in the  cases where the acceptable argument types and what the field will actually end up as at runtime diverge -- for example, there are fields that are typed as `Iterable[...]` because that's all the argument requires, but the attribute is always a `frozenset` after being initialized.

benhoyt:
> > A thought on a possible way to simplify: could you define a `_CheckInfo` dataclass with most of the fields, and then inherit `CheckInfo` from that to add the special fields, calling the super()'s `__init__` so you only need to special case the special fields?
> 
> I think that would work, but I'm not convinced it's truly simpler.

Fair enough. It's definitely more direct and flatter to do what you're suggesting. I don't love the `object.__setattr__`'s, but oh well.

tonyandrewmeyer:
> TBH I think we should be moving in the direction of having a custom `__init__` for more of the Scenario `dataclasses`, for better type checking and auto-completion UX in the cases where the acceptable argument types and what the field will actually end up as at runtime diverge -- for example, there are fields that are typed as `Iterable[...]` because that's all the argument requires, but the attribute is always a `frozenset` after being initialized.

I agree with this (we've talked about this before, I think). Perhaps worth creating a ticket?

To record some history: we had a version with custom `__init__` for most of the the `State` classes, primarily to support keyword-only fields in Python 3.8, but removed it in favour of using `__new__`. I believe the main reason was that we felt that it was cleaner to maintain in a single place than have to maintain a lot of `__init__` methods.

However, we don't need that any more (since we're using 3.10+), so some of the classes are perfectly fine with the dataclasses default `__init__`. I feel like a custom `__init__` to say "you can pass an iterable but you will end up with a set" is valuable enough that maintaining a few `__init__`'s is worthwhile.

testing/src/scenario/state.py
tonyandrewmeyer: There is also a documentation change, from:

Screenshot 2026-01-20 at 3 15 59 PM

To:

Screenshot 2026-01-20 at 3 15 27 PM

I do find it convenient to look at the field and see that the default is `None`, rather than having to scroll up and look at the `__init__`. However, I don't think we want to have the default in both places since then there's the risk that they get out of sync.

testing/src/scenario/state.py
tonyandrewmeyer: `threshold` can be `None` in **`Check`** (the thing you use to build a layer), but it cannot be `None` in a **`CheckInfo`** (the thing you get from Pebble when you ask about the current status of a check).

It makes no sense for `threshold` to be `None` here, because it'll always be in the Pebble response, so we should definitely *not* make this change. `successes` can only be `None` because it's supporting old Pebble versions where we didn't have that field.

For a `Check`, `None` means "use the default value". That does make sense, and should also *not* be changed.

testing/src/scenario/state.py
tonyandrewmeyer: This would only be a breaking change if your static type checking was currently failing, right? For example:

```python
from ops import testing

chk = testing.CheckInfo('chk', level='unknown')
```

```
% uvx '--with=ops[testing]' pyright /tmp/2274.py
/tmp/2274.py
  /tmp/2274.py:3:38 - error: Argument of type "Literal['unknown']" cannot be assigned to parameter "level" of type "CheckLevel | None" in function "__init__"
    Type "Literal['unknown']" is not assignable to type "CheckLevel | None"
      "Literal['unknown']" is not assignable to "CheckLevel"
      "Literal['unknown']" is not assignable to "None" (reportArgumentType)
```

I think we have already signalled that this is an error, so would potentially be ok.

However, I believe the reason we don't make it a hard error with `ops.pebble.CheckInfo` is so that when we do add a new level, your existing charm code will not crash when run against a newer Juju (and Pebble).

I feel like it's ok if your tests break, though. If you're updating your charm and charm tests to understand the new `level` value, then you should also update your version of `ops` that has built-in support for it. You don't have the situation where your tests get mock Pebble data that has the new value without you changing anything.

So my preference here would be to remove the try/except and let the tests crash if your tests are passing in the wrong mock Pebble data (and/or your charm is trying to use a new `level` value when adding a layer, without updating the ops version).

testing/src/scenario/state.py
tonyandrewmeyer: `successes` being `None` means "the simulated version of Juju is one with an old version of Pebble, which didn't include this field". It should stay as `None` if that's what is passed, because that's the behaviour you get with ops with those old Pebble versions, and the point of Scenario is to behave like ops.

Tests that have `None` for `successes` are checking that the charm works with those old versions of Pebble/Juju, so it's important it stays `None`.

If we wanted to tighten this up, we could have a consistency check that verified that this is/is not `None` based on the Juju version that has been selected in the context.

testing/src/scenario/state.py
```diff
@@ -922,36 +920,61 @@ class CheckInfo:
...
+            except ValueError:
+                # preserve the existing behaviour of not crashing on arbitrary strings
+                logger.error('%s is not a valid Pebble check level.', level)
+        object.__setattr__(self, 'level', level)
+        object.__setattr__(self, 'startup', startup)
+        object.__setattr__(self, 'status', status)
+        # FIXME: is it safe to narrow the `successes` attribute type to `int` and convert it here?
+        object.__setattr__(self, 'successes', successes)
+        object.__setattr__(self, 'failures', failures)
+        object.__setattr__(self, 'threshold', threshold if threshold is not None else 0)
```
tonyandrewmeyer: Is this a typo? We definitely should not change the default to `0`.

```suggestion
        object.__setattr__(self, 'threshold', threshold if threshold is not None else 3)
```

It worries me a bit that we don't have a test that would catch an invalid threshold.

tonyandrewmeyer (changes requested):
I haven't tested this out directly, but will do that after the fixes are done.

testing/src/scenario/state.py
james-garner-canonical: Agreed.

testing/src/scenario/state.py
james-garner-canonical: Not a typo, but a misunderstanding.

testing/src/scenario/state.py
james-garner-canonical: > `threshold` can be `None` in **`Check`** (the thing you use to build a layer), but it cannot be `None` in a **`CheckInfo`** (the thing you get from Pebble when you ask about the current status of a check).
> 
> ...
>
> For a `Check`, `None` means "use the default value". That does make sense, and should also _not_ be changed.

I'm fine with dropping all the changes to `threshold` from this PR, and good to know that it's intended functionality for `threshold` to be `None`-able in `Check` objects (I guess when constructing them manually? Or does Pebble omit `threshold` from `Check` too?).

But I'm wondering, *if* `testing.CheckInfo(threshold=my_check.threshold, ...)` is a useful pattern, then should passing `None` to the `CheckInfo` constructor also mean "use the default value" (3)?

testing/src/scenario/state.py
james-garner-canonical: Good point, I'd be keen to drop this as well since this is in `testing`.

james-garner-canonical:
```suggestion
        if level is not None:
            level = pebble.CheckLevel(level)
```

testing/src/scenario/state.py
james-garner-canonical:
```suggestion
        object.__setattr__(self, 'successes', successes)
```
Makes sense, thanks for explaining.

testing/src/scenario/state.py
```diff
@@ -922,36 +920,56 @@ class CheckInfo:
...
+    ):
+        object.__setattr__(self, 'name', name)
+        if level is not None:
+            level = pebble.CheckLevel(level)
+        object.__setattr__(self, 'level', level)
+        object.__setattr__(self, 'startup', startup)
+        object.__setattr__(self, 'status', status)
+        object.__setattr__(self, 'successes', successes)
+        object.__setattr__(self, 'failures', failures)
+        object.__setattr__(self, 'threshold', threshold if threshold is not None else 3)
```
```suggestion
        if threshold is None:
            threshold = 3
        object.__setattr__(self, 'threshold', threshold)
```

james-garner-canonical:
Thanks for the feedback. Still need to figure out exactly what to do for `threshold` -- drop entirely or allow `testing.CheckInfo` to accept `threshold=None` as a signal to use the default value (3).

Tests are definitely needed here, I'll add them and re-request review at that point.

dimaqq:
This is a fun read :)

james-garner-canonical:
I've dropped the changes to `threshold` from this PR, so this PR is now limited to adding a custom `__init__` method to `CheckInfo` that widens the accepted type of the `level` argument to also accept `str`, converting it to a `CheckLevel` before setting the `level` attribute so that the attribute typing is unchanged.

I've added some unit tests to cover the defaults with the custom init, the new behaviour, as well as the (existing) behaviour of some edge cases.

tonyandrewmeyer (approved):
Sorry about the slow review here - everything looks good to me now. Thanks for pushing forward with this.

I'm not totally sure this is where the new test module should live or be named, but I think we want a bigger re-org of that at some point, so I'm happy with it as is for now.

james-garner-canonical:
> I'm not totally sure this is where the new test module should live or be named, but I think we want a bigger re-org of that at some point, so I'm happy with it as is for now.

Yeah I wasn't sure either. I couldn't find a natural inclusion point in the existing tests, so I figured I'd just keep them self-contained and reasonably tightly named for now.

PR #2277, opened 2026-01-19
ci: explicitly provide the charmcraft repo location in CI

tonyandrewmeyer:
Currently, the automated PR (e.g. #2178) loses the charmcraft docs, because the script doesn't find the charmcraft repository clone.

Two changes to solve that:
* Remove the default (sibling folders for operator and charmcraft, except that there was a buggy extra `.parent`), in favour of always requiring the location to be passed.
* Explicitly pass the correct location in CI, based on the checkout action usage in the workflow

### DIFF
--- a/.github/update-best-practice-table.py
+++ b/.github/update-best-practice-table.py
@@ -154,7 +154,7 @@ def main():
     parser.add_argument(
         '--path-to-charmcraft',
         type=pathlib.Path,
-        default=pathlib.Path(__file__).parent.parent.parent / 'charmcraft',
+        required=True,
         help='Path to a clone of canonical/charmcraft',
     )
     args = parser.parse_args()


.github/update-best-practice-table.py
```diff
@@ -154,7 +154,7 @@ def main():
     parser.add_argument(
         '--path-to-charmcraft',
         type=pathlib.Path,
-        default=pathlib.Path(__file__).parent.parent.parent.parent / 'charmcraft',
+        default=pathlib.Path(__file__).parent.parent.parent / 'charmcraft',
```
james-garner-canonical: I think I commented on this initially, but I find this quite confusing, and I think there's still not really much guarantee that it would match anyone's local system. I'd be on board with dropping the default entirely, but I'm also fine with having it set to whatever works for whoever runs this script the most.

.github/workflows/update-best-practice-doc.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
-      - run: python .github/update-best-practice-table.py > docs/reuse/best-practices.txt
+      - run: python .github/update-best-practice-table.py --path-to-charmcraft=../charmcraft > docs/reuse/best-practices.txt
```
james-garner-canonical: WDYT about something like this?
```suggestion
      - run: python .github/update-best-practice-table.py --path-to-charmcraft='${{ github.workspace }}/charmcraft' > docs/reuse/best-practices.txt
```
I guess it's probably fine as-is too. Maybe `../charmcraft` would be more intuitive if `working-directory: operator` was on the line before `run`.

.github/workflows/update-best-practice-doc.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
...
         shell: bash
```
james-garner-canonical: Why do we set `shell: bash` here btw? I know that it's [subtly different](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell) to letting the shell *default* to bash, but I'm curious which of those subtle differences we care about here.

james-garner-canonical (approved):
Change seems reasonable, and fine to merge, though it would be good to get a successful run of the workflow from this branch immediately before merging.

I've left a few comments on things that took me a little bit to get my head around.

.github/update-best-practice-table.py
james-garner-canonical: To be clear, I'm fine with merging this as-is, just noting that it still hurts my brain.

tonyandrewmeyer: This matches CI (even though CI is also now changing to not match it). I think that's the most reasonable default, but also that siblings is probably the most common layout. I'm not opposed to making it required.

.github/workflows/update-best-practice-doc.yaml
tonyandrewmeyer: I don't recall, but I suspect it was from something that needed bash rather than sh in an earlier version of this.

.github/update-best-practice-table.py
james-garner-canonical: > This matches CI (even though CI is also now changing to not match it).

😵‍💫

> I think that's the most reasonable default, but also that siblings is probably the most common layout.

Agreed.

> I'm not opposed to making it required.

I suspect that would be the path of least confusion, but if you do prefer the default then I'm OK with leaving it as-is (though I reserve the right to bring this up next time we change the default).

tonyandrewmeyer:
> Change seems reasonable, and fine to merge, though it would be good to get a successful run of the workflow from this branch immediately before merging.

I can (and intend to) do it immediately after merging, but I don't believe I can do it before merging, because I can't run it from my fork in canonical/operator, and I can't run it in my fork because it's missing the permissions to clone the repos.

I think I would have to redo the branch as one on canonical/operator to be able to do that.

james-garner-canonical:
> > Change seems reasonable, and fine to merge, though it would be good to get a successful run of the workflow from this branch immediately before merging.
> 
> I can (and intend to) do it immediately after merging, but I don't believe I can do it before merging, because I can't run it from my fork in canonical/operator, and I can't run it in my fork because it's missing the permissions to clone the repos.
> 
> I think I would have to redo the branch as one on canonical/operator to be able to do that.

Makes sense, afterwards should be OK then.

dwilding (approved):

PR #2278, opened 2026-01-19
chore: bump urllib3 to resolve security alert

tonyandrewmeyer:
[Various security fixes and other minor changes](https://github.com/urllib3/urllib3/releases), none of which should impact us.

urllib3 gets pulled in as a dependency for the integration tests (via `minio`) and the release script (via `pygithub`).

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1313,11 +1313,11 @@ wheels = [
 
 [[package]]
 name = "urllib3"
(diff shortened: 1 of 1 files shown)


dimaqq (approved):
It took a while to actually read the urllib3 release notes, sorry for the delay.
Good to go, there's nothing that should affect us directly.

PR #2283, opened 2026-01-22
chore: relock the deps to get newer pynacl

dimaqq:
https://github.com/canonical/operator/security/dependabot/21 complains about some minor pynacl bug that's brought in via pygithub which is only use for the release script.

this is a dumb attempt to bump the dep and relock.


benhoyt:
Done (slightly more constrained) in https://github.com/canonical/operator/pull/2288

PR #2289, opened 2026-01-27
docs: add missing "How to" in page titles

dwilding:
A tiny PR to add "How to" in the titles of a few how-to guides.

### DIFF
--- a/docs/howto/make-your-charm-discoverable.md
+++ b/docs/howto/make-your-charm-discoverable.md
@@ -1,5 +1,5 @@
 (make-your-charm-discoverable)=
-# Make your charm discoverable
+# How to make your charm discoverable
 
 > See first: [Charm maturity](#charm-maturity)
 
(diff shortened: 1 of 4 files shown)


dimaqq:
I'm not sure about the proposed change.

Why change these four but the other two dozen?
Why change at all, when these are entries under the How-to section in the TOC?

Screenshot 2026-01-29 at 19 28 54

dwilding:
Most of our how-to pages follow this convention:
- Page title (displayed on the page itself) is "How to do the thing"
- What's displayed in the site navigation is "Do the thing", for a more efficient use of space

Arguably we could drop "How to" from all pages, yes. But the majority already follow the established pattern - in this PR I'm fixing up the four that don't.

dwilding:
I should add: The title in each `foo.md` file isn't what controls what appears in the site navigation. The site navigation entries are defined separately in the `index.md` files under each category.

dimaqq (approved):
Sorry I didn't realised the discrepancy between the titled and TOC entries!

PR #2296, opened 2026-01-30
feat: support charmcraft.yaml format as meta for testing.Context

### LINKED ISSUE #1424: Add support for passing charmcraft.yaml as metadata to testing.Context

james-garner-canonical:
This PR adds support for providing `charmcraft.yaml` formatted data to the `meta` argument of `testing.Context`.

It is an error to specify `actions` or `config` both as separate arguments and as keys in `meta`.

Resolves #1424


testing/tests/test_context.py
```diff
@@ -188,3 +190,112 @@ def test_charm_spec_is_deprecated():
     ctx = Context(CharmBase, meta={'name': 'some-name'})
     with pytest.warns(DeprecationWarning):
         _ = ctx.charm_spec  # type: ignore
+
+
+CONFIG: Mapping[str, Any] = {
+    'options': {'perambulator-abbreviation': {'type': 'string', 'default': 'pram'}}
```
dimaqq, line 197: 😆

testing/tests/test_context.py
```diff
@@ -188,3 +190,112 @@ def test_charm_spec_is_deprecated():
...
+    spec = ctx._charm_spec
+    assert spec.meta == {'name': 'MyCharm'}
+    assert spec.config is None
+    assert spec.actions == ACTIONS
+    ctx.run(ctx.on.update_status(), State())
+
+
+def test_init_with_no_meta():
+    with pytest.raises(ContextSetupError):
+        _ = Context(MyCharm)
```
dimaqq: **question:** does the tooling require the assignment, or do you reckon it's more legible this way, vs. the below?

```suggestion
        Context(MyCharm)
```

dimaqq (approved):
Looks good!

testing/src/scenario/context.py
```diff
@@ -654,11 +655,18 @@ def __init__(
 
         :arg charm_type: the :class:`ops.CharmBase` subclass to handle the event.
         :arg meta: charm metadata to use. Needs to be a valid metadata.yaml format (as a dict).
-            If none is provided, we will search for a ``metadata.yaml`` file in the charm root.
+            Alternatively, can be a full ``charmcraft.yaml`` dict, in which case ``config`` and
```
dimaqq, line 658: The docstring update makes sense.

I wonder if there are some other doc places to update though.
Would you mind going through sections like "how to write unit tests" to see if something can be simplified with the new way?

testing/tests/test_context.py
james-garner-canonical: I thought tooling required it, but it seems that it doesn't.

testing/src/scenario/context.py
james-garner-canonical: Our how to guides assume these arguments are left empty and the data is searched for in the appropriate files in the charm root.

dimaqq: I was thinking about examples like:
`Context(charm_type=MyCharmType, meta={'name': 'my-charm-name'})`

That alone seems fine, but if there was `meta=..., config={...}`, then we'd probably want to update that?

testing/src/scenario/context.py
```diff
@@ -677,6 +685,19 @@ def __init__(
             ``juju trust``).
         :arg charm_root: virtual charm filesystem root the charm will be executed with.
         """
+        meta = copy.deepcopy(meta)
+        config = copy.deepcopy(config)
+        actions = copy.deepcopy(actions)
```
dimaqq: Given that we only pop two top-level keys from meta, why **deep**copy?
Also, why copy config and actions, these don't seem to be modified.

james-garner-canonical: My thinking is that since we definitely need a copy if we're going to pop, why not go all the way and only hold onto copies, preventing any future errors or unintended interactions stemming from the same underlying issue.

dimaqq: So it's a drive-by change.
I think a stronger justification is needed.

james-garner-canonical: Holding onto an internal reference to the actual dictionary object passed by the caller makes both user code and our internal code error prone. This is highlighted by the fact that it's unsafe for us to `pop` from it as-is. It's true that we could try to preserve the problem in this PR, but why?

testing/src/scenario/context.py
james-garner-canonical: None of our examples appear to use both `meta` and `config`, and it seems that none of them use `actions`. For the ones that only use `config`, I think leaving them as `config={...}` is nicer than changing to. `meta={'config': {...}}`.

dimaqq (changes requested):
Let's discuss 3x `deepcopy`.

testing/src/scenario/context.py
```diff
@@ -677,6 +685,19 @@ def __init__(
...
+        # Extract config and actions from meta if it contains them (e.g., charmcraft.yaml)
+        # and they were not explicitly provided
+        if meta is not None:
+            meta_config = meta.pop('config', None)
+            if config is None:
+                config = meta_config
+            meta_actions = meta.pop('actions', None)
+            if actions is None:
+                actions = meta_actions
+
```
dimaqq: Option 1:

```suggestion
        # Extract config and actions from meta if it contains them (e.g., charmcraft.yaml)
        # and they were not explicitly provided
        if meta is not None:
            meta = {**meta}
            meta_config = meta.pop('config', None)
            if config is None:
                config = meta_config
            meta_actions = meta.pop('actions', None)
            if actions is None:
                actions = meta_actions

```

Option 2:

```suggestion
        # Extract config and actions from meta if it contains them (e.g., charmcraft.yaml)
        # and they were not explicitly provided
        if meta is not None:
            meta_config = meta.get('config')
            meta_actions = meta.get('actions')
            # Optional, I think CharmSpec should be doing this instead:
            meta = {k: v for k, v in meta.items if k not in ("config", "actions")}
            if config is None:
                config = meta_config
            if actions is None:
                actions = meta_actions

```

Option 3, if CharmSpec does the "right thing":

```suggestion
        # Extract config and actions from meta if it contains them (e.g., charmcraft.yaml)
        # and they were not explicitly provided
        if meta is not None:
            if config is None:
                config = meta.get('config')
            if actions is None:
                actions = meta.get('actions')

```

Doesn't that look so much simpler?

james-garner-canonical: [_CharmSpec](https://github.com/canonical/operator/blob/main/testing/src/scenario/state.py#L1864) just holds onto references to the three dictionaries.

The dicts are then used via `_CharmSpec` by the Scenario machinery to check consistency, write the metadata files to the virtual charm root, etc., or by user-facing methods like generating a `State` from a `Context`.

This is why I'm proposing that we should work with copies instead of the objects the user passes in.

---

Re: Options 1 and 2, either of those shallow copy based approaches would do the job for this PR's feature. I just think it would be worth making deep copies instead to remove the underlying problem that requires us to have to think about making copies here once and for all.

testing/tests/test_context.py
```diff
@@ -188,3 +190,112 @@ def test_charm_spec_is_deprecated():
...
+
+
+def test_init_with_no_meta():
+    with pytest.raises(ContextSetupError):
+        Context(MyCharm)
+
+
+def test_init_with_bad_meta():
+    ctx = Context(MyCharm, meta={'a truth universally acknowledged': 'it'})
+    with pytest.raises((UncaughtCharmError, KeyError)):
```
tonyandrewmeyer: With https://github.com/canonical/operator/pull/2314 coming, it's probably easier to preemptively use that approach here.

testing/src/scenario/context.py
```diff
@@ -654,11 +655,18 @@ def __init__(
...
         :arg meta: charm metadata to use. Needs to be a valid metadata.yaml format (as a dict).
-            If none is provided, we will search for a ``metadata.yaml`` file in the charm root.
+            Alternatively, can be a full ``charmcraft.yaml`` dict, in which case ``config`` and
+            ``actions`` will be automatically extracted from it (if not explicitly provided).
+            Otherwise, if no meta, actions, or config were provided, will be loaded from the
+            ``charmcraft.yaml`` file in the charm root (falling back to ``metadata.yaml``).
         :arg actions: charm actions to use. Needs to be a valid actions.yaml format (as a dict).
-            If none is provided, we will search for a ``actions.yaml`` file in the charm root.
+            If none is provided, will be extracted from ``meta`` if present.
+            Otherwise, if no meta, actions, or config were provided, will be loaded from the
```
tonyandrewmeyer, line 667: Nice update fixing these to mention charmcraft.yaml in the autospec mode.

testing/src/scenario/context.py
tonyandrewmeyer: I agree that it would be cleaner if we broke the link, whether that's by a deep copy or other means (probably it has to be a deep copy at some point, since we use it for things like network-get mocking). However, I also agree that it doesn't really fit with this PR.

My suggestion is that we have a separate PR that does do a deep copy at this point in the code (I think a deep copy is probably simpler and more correct than any of the other options), merge that, and then do this PR.

tonyandrewmeyer (approved):
I feel like it may be better to raise an error if there's an actions/config dict provided *both* in the `meta` argument and separately, rather than let one of them 'win'.

I wonder if we should add an example somewhere in the documentation (maybe in one of the library docs that isn't going to be moved to charmlibs?). In most cases, charms are ideally never using this and always auto-spec'ing. However, if they do use it, I would rather we encourage:

```python
ctx = Context(MyCharm, charmcraft_dict)
```

rather than

```
ctx = Context(MyCharm, meta=charmcraft_dict)
```

And reserve the `meta=` keyword passing for when it's specifically the metadata dict.

I haven't tried running this since it's unclear what the exact implementation of the core piece will be, but I'm happy to approve trusting that it'll work at that point.

testing/src/scenario/context.py
james-garner-canonical: I've made a separate PR for discussion as suggested: #2349

dimaqq: If there's a separate PR, can this PR ship without copies?

james-garner-canonical: If we merge #2349, this PR should be updated accordingly. If we don't merge #2349 (rejecting making copies) this PR would have to adopt a different approach.

dimaqq:
> I feel like it may be better to raise an error if there's an actions/config dict provided _both_ in the `meta` argument and separately, rather than let one of them 'win'.


Let's debate and make a decision on this.

testing/tests/test_e2e/conftest.py
```diff
@@ -15,7 +15,7 @@
 
 @pytest.fixture
 def secrets_context(secrets_charm_meta: dict[str, Any]):
-    return Context(SecretsCharm, meta=secrets_charm_meta, actions=secrets_charm_meta['actions'])
+    return Context(SecretsCharm, meta=secrets_charm_meta)
```
james-garner-canonical, line 18: Making it an error to pass `actions` both as a separate argument and as an entry in `meta` actually broke this fixture. It loads `secrets_charm_meta` from a `charmcraft.yaml` file, and passes `['actions']` explicitly (since it was written before this PR :)).

I think this is a sign this change might be too breaking.

testing/src/scenario/context.py
james-garner-canonical: Resolving this thread since we merged #2349.

testing/src/scenario/context.py
james-garner-canonical: Resolving this thread since we merged the separate `deepcopy` PR.

testing/tests/test_context.py
james-garner-canonical: TBH I think this is lower weight than adding `monkeypatch` and using `monkeypatch.setenv`.

james-garner-canonical:
Requesting re-review as things have changed significantly:
- `deepcopy` change landed separately, so this PR builds on that
- We explicitly error if `actions` or `config` are passed in `meta` and separately at the same time -- but this actually broke some of our tests, so I wonder if this is too breaking a change.
  - We could let the separate argument take precedence like it does in `main`, and perhaps warn in this case.
  - Or try a gentler breaking change, like making it an error if they're passed separately and aren't identical.

testing/tests/test_e2e/conftest.py
tonyandrewmeyer: Could we try it across our inventory of charms?

testing/src/scenario/context.py
```diff
@@ -689,13 +696,14 @@ def __init__(
...
-                config=copy.deepcopy(config),
-            )
+            meta = copy.deepcopy(meta) if meta is not None else {'name': str(charm_type.__name__)}
+            if actions is not None and 'actions' in meta:
+                raise ValueError('Cannot specify actions in both charmcraft.yaml and separately')
+            actions = copy.deepcopy(actions) if actions is not None else meta.pop('actions', None)
+            if config is not None and 'config' in meta:
+                raise ValueError('Cannot specify config in both charmcraft.yaml and separately')
+            config = copy.deepcopy(config) if config is not None else meta.pop('config', None)
+            spec = _CharmSpec(charm_type=charm_type, meta=meta, actions=actions, config=config)
```
tonyandrewmeyer: Freel free to push back on this; it's pretty subjective. I don't find this very readable.

Some of it maybe ruff pushes us into, like always using the ternary operator? If so then I would say that the consistency gained through tooling-enforced rules wins out.

Maybe a comment would help? Maybe if it was in a loop? (but it's only two items so ...)

I find my argument weakening by not being able to immediately see how I would write a nicer version of this.

james-garner-canonical: Maybe comments to divide into logical sections and explain what's up? Maybe flipping the ternary to use `is` rather than `is not`?
```py
            # Metadata: may be in charmcraft.yaml format
            meta = {'name': str(charm_type.__name__)} if meta is None else copy.deepcopy(meta)
            # Actions: it is an error if actions is specified both separately and in charmcraft.yaml
            if actions is not None and 'actions' in meta:
                raise ValueError('Cannot specify actions in both charmcraft.yaml and separately')
            actions = meta.pop('actions', None) if actions is None else copy.deepcopy(actions)
            # Config: it is an error if config is specified both separately and in charmcraft.yaml
            if config is not None and 'config' in meta:
                raise ValueError('Cannot specify config in both charmcraft.yaml and separately')
            config = meta.pop('config', None) if config is None else copy.deepcopy(config)
            # Charm spec from user specified charm type and metadata
            spec = _CharmSpec(charm_type=charm_type, meta=meta, actions=actions, config=config)
```

testing/tests/test_e2e/conftest.py
james-garner-canonical: Good idea, I'm running it now, but I'm not confident in my ability to actually detect new failures from the output. Would you mind running it too?

tonyandrewmeyer: I get the same number of successes with the branch, so it seems safe to me.

However, I'm open to making it a warning (maybe a deprecation warning?) instead if you think it's better to be safe.

dimaqq:
Sorry I lost track about what we decided for copying stuff exactly (same pr? separate? specific copies? whack-all deepcopy?).
Will put this on discussion tomorrow.

testing/src/scenario/context.py
james-garner-canonical: Maybe:
```py
            meta = copy.deepcopy(meta) if meta else {'name': str(charm_type.__name__)}
            # actions
            if actions is None:
                actions = meta.pop('actions', None)
            elif 'actions' in meta:
                raise ValueError('Cannot specify actions in both charmcraft.yaml and separately')
            else:
                actions = copy.deepcopy(actions)
            # config
            if config is None:
                config = meta.pop('config', None)
            elif 'config' in meta:
                raise ValueError('Cannot specify config in both charmcraft.yaml and separately')
            else:
                config = copy.deepcopy(config)
            spec = _CharmSpec(charm_type=charm_type, meta=meta, actions=actions, config=config)

File: testing/tests/test_e2e/conftest.py (hunk 8 above)
Comment (james-garner-canonical) [reply testing/tests/test_e2e/conftest.py:18]: Since this doesn't break any charm tests we can detect, let's keep it as an error.

**Comment by james-garner-canonical** [comment]
> Sorry I lost track about what we decided for copying stuff exactly (same pr? separate? specific copies? whack-all deepcopy?). Will put this on discussion tomorrow.

We decided on doing the deep copies in:
- #2349

This PR doesn't introduce any new copying, just pulls `'config'` and `'actions'` out of `meta`.

File: testing/src/scenario/context.py [hunk 10]
```diff
@@ -692,13 +699,14 @@ def __init__(
...
 
         else:
-            meta = copy.deepcopy(meta) if meta else {'name': str(charm_type.__name__)}
-            spec = _CharmSpec(
-                charm_type=charm_type,
-                meta=meta,
-                actions=copy.deepcopy(actions),
-                config=copy.deepcopy(config),
-            )
+            meta = copy.deepcopy(meta) if meta is not None else {'name': str(charm_type.__name__)}
```
dimaqq: This is a subtle change for the case of explicit `meta={}` and supplying a non-empty `actions` or `config`.

I'm not sure why someone would do that, so I think it's OK.

I wonder if the code on line 691 should be changed to:

`if not any(x is not None for x in (meta, actions, config)):`

Though against, I've no clue why someone may override action or config with an empty dict (no actions or no config), and yet supply a not-None, yet-empty meta.

So maybe current code is fine?
Please double-check my logic.

james-garner-canonical: You're right about the subtle behaviour change, and I'm not 100% sure why I made the decision to test meta for `None` instead of falsiness other than perhaps for symmetry with config and actions. I'll revert to the original behaviour for meta.

PR #2306, opened 2026-02-04
fix: support Pydantic MISSING sentinel in ops.Relation.save

### LINKED ISSUE #2299: Support pydantic.experimental.missing_sentinel.MISSING in ops.Relation.save
I've extended the existing test like so:

```diff
     class MyPydanticDatabag(pydantic.BaseModel):
-        foo: str
+        foo: str | MISSING = MISSING
         bar: int = pydantic.Field(default=0, ge=0)
         baz: list[str] = pydantic.Field(default_factory=list)
         quux: Nested = pydantic.Field(default_factory=Nested)
+        hah: int | MISSING = MISSING
```

And that fails with `KeyError('hah')` at:
```py
/code/operator/test/test_model_relation_data_class.py(407)_on_relation_changed()
-> event.relation.save(data, self.app, encoder=self.encoder)
> /code/operator/ops/model.py(1918)save()
-> data = {field: encoder(values[attr]) for attr, field in sorted(fields.items())}
```

dimaqq:
Support Pydantic's experimental `MISSING` sentinel field types and values.

`ops.Relation.load()` does the right thing already.
`ops.Relation.save()` needs a small fix to erase missing fields

Check list:
- [x] core code change
- [x] unit tests pass
- [x] ruff format
- [x] docs update
- [x] test on real Juju --> https://github.com/dimaqq/hexanator/pull/7
- [x] separate new test case
- [x] make the new test conditional on being able to import MISSING
- [x] (whatever reviews bring)

Fixes #2299

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1873,9 +1873,9 @@ class Relation:
                 # data.destination will be stored under the Juju relation key 'to'
                 relation.save(data, self.unit)
 
-        If a class declares a field, but the object does not have a value for it,
-        the field will be erased from the relation data.
-        This is possible when using Pydantic's ``MISSING`` sentinel.
+        If a Pydantic model's ``model_dump`` method omits any field (e.g. if its
+        value is Pydantic's ``MISSING`` sentinel) the field will be erased from
+        the relation data.
 
         Args:
             obj: an object with attributes to save to the relation data, typically
@@ -1919,7 +1919,7 @@ class Relation:
             values = {field: getattr(obj, field) for field in fields}
 
         # Encode each value, and then pass it over to Juju.
-        # Missing values are erased from the databag via an empty string.
+        # Missing values are erased from the databag using empty string values.
         data = {
             field: encoder(values[attr]) if attr in values else ''
             for attr, field in sorted(fields.items())


ops/model.py
```diff
@@ -1915,7 +1915,7 @@ def _add_transfer(self):
             values = {field: getattr(obj, field) for field in fields}
 
         # Encode each value, and then pass it over to Juju.
-        data = {field: encoder(values[attr]) for attr, field in sorted(fields.items())}
+        data = {field: encoder(values[attr]) if attr in values else '' for attr, field in sorted(fields.items())}
```
tonyandrewmeyer: I feel that this is the wrong behaviour, that if it's missing that should be missing in the dump as well, leaving it unchanged.

Pydantic says:

> The MISSING sentinel is a singleton indicating a field value was not provided during validation.
>
> This singleton can be used as a default value, as an alternative to None when it has an explicit meaning. During serialization, any field with MISSING as a value is excluded from the output.

I don't get "delete this" from that definition. To me "excluded from the output" means that it is like you didn't have it in the class, and our behaviour in Ops (reflecting that of Juju) is that if it's not there, it means don't touch it. If you want to delete something, you need to do that explicitly.

As discussed in daily today, I would be fine if there's a use-case for entirely replacing to add that as an argument for `save`. Maybe even a helper that does `del self.data[app_or_unit][key]` for all keys that have a specific value or something like that (again, assuming there's a clear use-case).

I think the correct fix here is something like this:

```python
            values = obj.model_dump(mode='json', by_alias=True, exclude_defaults=False)  # type: ignore
            for name in values:
                # Pydantic takes care of the alias.
                field = obj.__class__.model_fields[name]
                fields[field.alias or name] = field.alias or name  # type: ignore
```

Or something similar: basically, only include in the dump the keys that `model_dump` provides, rather than for all fields.

james-garner-canonical: I somehow missed that nuance when looking at this, good point, Tony.

I guess we could also avoid deletion like this:
```suggestion
        data = {field: encoder(values[attr]) for attr, field in sorted(fields.items()) if attr in values}
```

tonyandrewmeyer: Yes, that's the suggestion I made wherever this was first brought up. I wondered if it was more efficient to do the loop the other way around, but probably in reality it doesn't make any difference either way. The @james-garner-canonical version is probably clearer to read.

dimaqq: Yep, that's the question: should we whack fields that are both **declared** and **unset**?

This PR is the most narrow approach: since this path was broken (run-time exception, #2299), implementing the missing value functionality this way is not a breaking change.

The issue I have with Tony's alternative, is that it leaves the `MISSING` a no-op on the write path. 

Which means the charmers would have to remove the values for declared fields out of band: there's no way to do that using `relation.save(...)`, they'd have to do `for field in some_fields: del relation[Unit/App][field]` I think?

tonyandrewmeyer: > This PR is the most narrow approach: since this path was broken (run-time exception, #2299), implementing the missing value functionality this way is not a breaking change.

I agree it's not a breaking change. But my argument is that it is narrower, less surprising, more consistent, and more correct to fix it the other way.

> Which means the charmers would have to remove the values for declared fields out of band: there's no way to do that using `relation.save(...)`, they'd have to do `for field in some_fields: del relation[Unit/App][field]` I think?

Yes, if someone wants to remove a field, then they need to explicitly do so. I did say that it seems reasonable to provide a helper for this or an argument on the save method, if there is a compelling use case.

I'm still not convinced about the removal. If you downgrade, you've lost data, so behaviour changes (quite likely breaking the relation) and that doesn't seem like the happy path. If you stay at the same schema then there's some wasted space, but that doesn't seem so bad, and a "explicitly delete once all units have updated" path could assist with that. If you're upgrading, you've lost data mustn't reuse or change the field, so it's not in the way at all. We deliberately designed `load` so that if there are fields that aren't in the schema that's ignored.

dimaqq: Based on charming feedback https://discourse.charmhub.io/t/supporting-pydantics-missing-sentinel-in-ops/19706 and stand-up discussion, this PR proceeds with "remove declared fields with MISSING values from the databag on .save()".

ops/model.py
dimaqq, line 1912: This is where we use Pydantic's slightly magic method to get the values from the Pydantic object.

The `model_dump()` drops the fields with the `MISSING` value.

james-garner-canonical (commented):
The implementation looks good based on our team discussion.

The addition of a class using this sentinel to the tests is great, ensuring that we don't crash on `MISSING` anymore. I think it would be a good idea to add a test that explicitly asserts that `MISSING` doesn't get serialized as well.

Are you still planning to add some real Juju tests, per the checklist in the PR description? I'm not sure if they're necessary, but I'm definitely not opposed to us adding more integration tests.

ops/model.py
```diff
@@ -1915,7 +1915,10 @@ def _add_transfer(self):
             values = {field: getattr(obj, field) for field in fields}
 
         # Encode each value, and then pass it over to Juju.
-        data = {field: encoder(values[attr]) for attr, field in sorted(fields.items())}
+        data = {
```
tonyandrewmeyer, line 1923:
```suggestion
        # If a field does not have a provided value, ask Juju to delete that key.
        data = {
```

dimaqq: Good point that the docs could use small update.

dimaqq: Done, PTAL.

test/test_model_relation_data_class.py
dimaqq, line 450: @james-garner-canonical this assertion ensures that the field with a `MISSING` value does not end up in the databag.

ops/model.py
```diff
@@ -1853,8 +1853,9 @@ def save(
         fields will be saved through to the relation data. Pydantic fields that
         have an ``alias``, or dataclasses that have a ``metadata{'alias'=}``,
         will have the object's value saved to the Juju relation data with the
-        alias as the key. For other classes, all of the object's attributes that
-        have a class type annotation and value set on the object will be saved
+        alias as the key. Fields without a value (e.g. Pydantic's ``MISSING``
+        sentinel) will be erased. For other classes, all of the object's attributes
+        that have a class type annotation and value set on the object will be saved
```
james-garner-canonical: I wonder if describing the new erase behaviour should come at the very end, after we finish describing the supported save behaviour, like this:

```suggestion
        alias as the key. For other classes, all of the object's attributes that
        have a class type annotation and value set on the object will be saved
        through to the relation data.
        
        If a class declares a field, but the object does not have a value for it, 
        the field will be erased from the relation data.
        This is possible when using Pydantic's ``MISSING`` sentinel.
```

ops/model.py
```diff
@@ -1853,8 +1853,9 @@ def save(
         fields will be saved through to the relation data. Pydantic fields that
         have an ``alias``, or dataclasses that have a ``metadata{'alias'=}``,
         will have the object's value saved to the Juju relation data with the
-        alias as the key. For other classes, all of the object's attributes that
-        have a class type annotation and value set on the object will be saved
+        alias as the key. Fields without a value (e.g. Pydantic's ``MISSING``
+        sentinel) will be erased. For other classes, all of the object's attributes
```
james-garner-canonical: Saying that fields without a value will be erased from the relation data is clear, but not quite true:
- If a field from `dataclasses.fields` isn't present on the object, `dataclasses.asdict` will raise an `AttributeError`.
- If a field from `get_type_hints` is not present on a vanilla class, `getattr(obj, field)` will raise an `AttributeError`.

We could 'fix' this in the vanilla class case (`getattr(obj, field) for field in fields if hasattr(obj, field)`), but the `dataclasses.asdict` case would be trickier to change, and I wouldn't suggest it.

Whether we change the behaviour for vanilla classes or not, I think we should document the expected behaviour in all three cases: Pydantic=erase, dataclasses=error, vanilla=?

dimaqq: I think that the only other "thing" where a field can be legally missing is a `TypedDict` with a `NotRequired` annotation.

Today, these are not supported:
- `.load(SomeTypeDict, rel)` will return a plain dict with everything in the databag, ignoring the typed keys
- `.save(SomeTypeDict(foo="bar"), rel)` will write nothing, whether values match declared keys or not

neither raises any errors...

So, I'd punt any discussion on other data-ish classes.
Let's solve Pydantic specifically in this PR.

ops/model.py
dimaqq: That's fair. I've added the suggested paragraph, though after the example, as that illustrates aliases.

ops/model.py
james-garner-canonical: Yes, let's leave `TypedDict` out of scope for this PR. I'm on board with not making any further behaviour changes.

However, I think we should either:
- Be clear that for `dataclasses` and vanilla classes, a 'field' not having a corresponding 'value' will lead to an error instead.
- Or be clear that omitting a value is only really possible with Pydantic classes via `model_dump`.

I think the new docstring gets pretty close on the latter. But maybe it could be:
```
        If a Pydantic model's ``model_dump`` method omits any field
        (e.g. if its value is Pydantic's ``MISSING`` sentinel)
        the field will be erased from the relation data.
```

test/test_model_relation_data_class.py
james-garner-canonical: Ah, good point!

ops/model.py
```diff
@@ -1915,7 +1919,11 @@ def _add_transfer(self):
             values = {field: getattr(obj, field) for field in fields}
 
         # Encode each value, and then pass it over to Juju.
-        data = {field: encoder(values[attr]) for attr, field in sorted(fields.items())}
+        # Missing values are erased from the databag via an empty string.
```
benhoyt: Nit:
```suggestion
        # Missing values are erased from the databag using an empty string.
```

benhoyt (approved):
Seems reasonable to me, thanks.

ops/model.py
dimaqq: Sure, thanks!

PR #2313, opened 2026-02-09
chore: clean up ops-testing project file

### LINKED ISSUE #2311: Clean up ops-testing project file
From #2310 

1. We could totally switch to `hatchling` as we do in `ops-tracing`.
   * correction: switch to `uv-build`
3. We should drop `[bdist_wheel] universal = 1`, as we only support Python 3.

dimaqq:
Fixes #2311 -- switches the build backend to Hatchling, better links, etc.

Open questions:
- [ ] I've renamed `testing/src/scenario` to `testing/src/ops_scenario`. The alternative would be to tweak some project setting to our setup (package name doesn't match directory)
- [ ] If we go renaming things, how about dropping the src layout?


tonyandrewmeyer:
> * [ ]  I've renamed `testing/src/scenario` to `testing/src/ops_scenario`. The alternative would be to tweak some project setting to our setup (package name doesn't match directory)

I'm pretty sure this is wrong, whether it's `src` layout or not. The PyPI package name is `ops-scenario`, but the package is `scenario` (it's "import scenario" not "import ops_scenario"). The filesystem should match the namespace not the PyPI package.

> * [ ]  If we go renaming things, how about dropping the src layout?

What about adopting it instead? It still seems to be that this is the current common practice, and it makes it simpler for the monorepo with tooling, based on the experimentation last year.

dimaqq:
> I'm pretty sure this is wrong, whether it's `src` layout or not. The PyPI package name is `ops-scenario`, but the package is `scenario` (it's "import scenario" not "import ops_scenario"). The filesystem should match the namespace not the PyPI package.

I tried that, but `uv-build` disagrees:

```
🦐/c/operator (chore-clean-up-ops-testing-project-file)> tox -e unit -- -vx
unit: uv-sync> uv sync --locked --python-preference system --no-default-groups --group unit --group xdist -p /Users/dima/.local/share/uv/tools/tox/bin/python
Resolved 59 packages in 3ms
  × Failed to build `ops-scenario @ file:///code/operator/testing`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `uv_build.build_editable` failed (exit status: 1)

      [stderr]
      Error: Expected a Python module at: src/ops_scenario/__init__.py
```

P.S. and weirdly your advice works if I move everything under `packages/`.
So confusing!

P.P.S. I think there's a subtle difference between editable install and sdist build.
One wants it one way, the other, the opposite way.

dimaqq:
I think this particular attempt is a dead end. I'll have to try again from scratch.

PR #2315, opened 2026-02-09
chore: include ops.hookcmds in Ops setuptools package discovery

### LINKED ISSUE #2312: Ops build warning about ops.hookcmds
```
/Users/dima/.cache/uv/builds-v0/.tmpUnI41W/lib/python3.14/site-packages/setuptools/command/build_py.py:215: _Warning: Package 'ops.hookcmds' is absent from the `packages` configuration.
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'ops.hookcmds' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'ops.hookcmds' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'ops.hookcmds' to be distributed and are
        already explicitly excluding 'ops.hookcmds' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************

!!
  check.warn(importable)
```

dimaqq:
Fixes #2312 

When `ops.hookcmds` has landed, we forgot to add the submodule to the setuptools config in pyproject.toml and now we get warnings. The PR addresses that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,7 +94,7 @@ ops-scenario = { workspace = true }
 ops-tracing = { workspace = true }
 
 [tool.setuptools.packages.find]
-include = ["ops", "ops._private", "ops.lib"]
+include = ["ops", "ops._private", "ops.hookcmds", "ops.lib"]
 
 [tool.setuptools.dynamic]
 version = {attr = "ops.version.version"}


james-garner-canonical (approved):
This looks like what we're supposed to do based on the docs and what we were already doing. But what effect does this actually have? We were shipping `ops.hookcmds` anyway, right?

dimaqq:
Yes we are shipping ops.hookcmds.
Right now, the only difference is having or not having a warning when building the wheels.

tonyandrewmeyer (approved):
Thanks!

PR #2316, opened 2026-02-10
feat: bump default Juju version in ops.testing.Context

dimaqq:
As decided at the standup, bumping the default Juju version in Scenario Context to the version of the current Juju LTS, 3.6.14.


dimaqq:
Oddly enough, I didn't have to adjust a single test 🎉

PR #2317, opened 2026-02-10
ci: adjust minimum Python version in broad charm compatibility tests

tonyandrewmeyer:
A large number of our [broad charm compatibility tests](https://github.com/canonical/operator/actions/workflows/published-charms-tests.yaml) fail because the charm declares that it works with Python 3.8. That conflicts with us pushing the latest ops 3.x into the dependencies.

Generally, if we've made a change to ops[...] that breaks charms, that will be across Python versions, and we're more concerned with 3.10+ anyway (and in main, we're only supporting 3.10+). We could have some compatibility tests that check the latest 2.x branch for charms that expect 3.8, but that doesn't seem as useful.

This PR adjusts the dependency patching to also handle bumping the Python version to 3.10. It [reduces the number of failures to around 6](https://github.com/tonyandrewmeyer/operator/actions/runs/23189775456), some of which are only for static checks (which I haven't worked on in this PR at all).

Since the changes are more complex now, they are moved out of the action file (YAML+bash) and into a helper (Python) script. Primarily, we do the work through tox, which I found to be the most consistent way to handle things across packaging tools (particularly given that we are using wheels now rather than git references).

The script also adds support for `tox.toml`, as some charms use that rather than `tox.ini`.

### DIFF
--- a/.github/update-published-charms-tests-workflow.py
+++ b/.github/update-published-charms-tests-workflow.py
@@ -80,7 +80,9 @@ CHARM_ROOTS = {
     'argo-operators': ['charms/argo-controller'],
     'catalogue-k8s-operator': ['charm'],
     'jimm': ['charms/jimm'],
-    'k8s-operator': ['charms/worker/k8s'],
+    # k8s-operator has an odd structure where there is a charm in charms/worker, and
+    # also another separate charm not as a sibling, but inside a subfolder.
+    'k8s-operator': ['charms/worker', 'charms/worker/k8s'],
     'katib-operators': ['charms/katib-controller', 'charms/katib-db-manager', 'charms/katib-ui'],
     'kfp-operators': [
         'charms/kfp-api',
--- a/.github/workflows/example-charm-tests.yaml
+++ b/.github/workflows/example-charm-tests.yaml
@@ -54,3 +54,27 @@ jobs:
         run: |
           cd ${{ matrix.dir }}
           tox -e unit
+      - name: Write mypy-requirements.txt
+        shell: uv run --script --no-project {0}
+        run: |
+          # /// script
+          # dependencies = ["tomli"]  # drop for tomllib when we're Python 3.11+
+          # ///
+          import tomli, pathlib
+          pyproject_toml = tomli.loads(pathlib.Path('pyproject.toml').read_text())
+          deps = '\n'.join(pyproject_toml['dependency-groups']['mypy'])
+          print(deps)  # logging
+          pathlib.Path('mypy-requirements.txt').write_text(deps)
+      - name: Run mypy
+        working-directory: ${{ matrix.dir }}
+        env:
+          MYPYPATH: 'src:lib:../..'
+        run: |
+          uv run \
+            --with-requirements=../../mypy-requirements.txt \
+            --with=../.. \
+            --with=../../testing \
+            --group=lint \
+            --group=unit \
+            --group=integration \
+            mypy --explicit-package-bases --follow-imports=silent src tests
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -12,6 +12,7 @@ Dependabot
 dict's
 failover
 IDE's
+jhack
 Jubilant's
 lightkube
 liveness
--- a/docs/explanation/state-transition-testing.md
+++ b/docs/explanation/state-transition-testing.md
@@ -4,6 +4,9 @@ The framework for state-transition testing in Ops was originally developed under
 
 This page compares Scenario to Harness and explains how state-transition tests work. For a summary of how to write unit tests, see [](/howto/write-unit-tests-for-a-charm).
 
+If you are migrating an existing charm and need realistic starting data or writing a regression test, you can capture a live model snapshot and convert it into a state-transition test.
+For a workflow focused on relation debugging, see [](#generate-tests-from-a-deployed-model).
+
 ```{note}
 In your testing dependencies, add `ops[testing]` rather than `ops-scenario`, and
 this will ensure that an appropriate version of the framework is installed.
--- a/docs/howto/manage-relations.md
+++ b/docs/howto/manage-relations.md
@@ -361,3 +361,24 @@ def test_active_with_another_app(juju: jubilant.Juju):
 
     juju.wait(jubilant.all_active)
 ```
+
+(generate-tests-from-a-deployed-model)=
+### Generate tests from a deployed model
+
+When a relation bug is only reproducible in a live deployment, you can snapshot
+the live relation state and turn it into a state-transition test.
+
+A practical workflow is:
+
+1. Reproduce the issue in a deployed model.
+2. Use `jhack scenario snapshot` to export representative relation state.
+3. Adapt the generated test scaffolding to your `ops.testing` setup (`Context`,
+    `State`, and `ctx.on.relation_*` events).
+4. Keep one failing test that captures the bug, then fix charm code until the
+    test passes.
+
+This is especially useful when relation databags, relation IDs, or unit/app
+combinations are difficult to reconstruct by hand.
+
+For command syntax and tool updates, see the
+[jhack repository](https://github.com/canonical/jhack).
--- a/docs/howto/migrate/migrate-integration-tests-from-pytest-operator.md
+++ b/docs/howto/migrate/migrate-integration-tests-from-pytest-operator.md
@@ -3,6 +3,10 @@
 
 Many charm integration tests use [pytest-operator](https://github.com/charmed-kubernetes/pytest-operator) and [python-libjuju](https://github.com/juju/python-libjuju). This guide explains how to migrate your integration tests from those libraries to Jubilant.
 
+```{tip}
+Try bootstrapping your migration with an AI Agent (such as GitHub Copilot or Claude Code). Instruct the agent to clone the canonical/jubilant and canonical/pytest-jubilant repositories, study them, and then migrate the charm integration tests to Jubilant. You should end up with a great starting point to then continue as outlined in the rest of this guide.
+```
+
 To get help while you're migrating tests, please keep the {external+jubilant:doc}`Jubilant API Reference <reference/jubilant>` handy, and make use of your IDE's autocompletion -- Jubilant tries to provide good type annotations and docstrings.
 
 Migrating your tests can be broken into three steps:
--- a/docs/howto/write-unit-tests-for-a-charm.md
+++ b/docs/howto/write-unit-tests-for-a-charm.md
@@ -134,6 +134,13 @@ will raise a `ValueError`, matching the behaviour of `charmcraft pack`.
 Rename or remove the overlapping keys to fix this.
 ```
 
+## Generate tests from a deployed model
+
+If your test needs realistic relation inputs from a deployed model, you can use `jhack scenario snapshot` to capture state, then adapt the generated data to `ops.testing` code.
+
+For a workflow focused on relation debugging, see
+[](#generate-tests-from-a-deployed-model).
+
 ## Mock beyond the State
 
 If you wish to use the framework to test an existing charm type, you will probably need to mock out certain calls that are not covered by the `State` data structure. In that case, you will have to manually mock, patch or otherwise simulate those calls.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,10 @@ lint = [
     "pyright~=1.1",
     "typing_extensions~=4.2",
 ]
+mypy = [  # used to type check example charms in CI
+    "mypy>=1.19",
+    "types-PyYAML",
+]
 unit = [
     "ops[testing,tracing]",
     "pytest~=8.4",
--- a/testing/src/scenario/context.py
+++ b/testing/src/scenario/context.py
@@ -655,14 +655,21 @@ class Context(Generic[CharmType]):
 
         :arg charm_type: the :class:`ops.CharmBase` subclass to handle the event.
         :arg meta: charm metadata to use. Needs to be a valid metadata.yaml format (as a dict).
-            If none is provided, we will search for a ``metadata.yaml`` file in the charm root.
+            Alternatively, can be a full ``charmcraft.yaml`` dict, in which case ``config`` and
+            ``actions`` will be automatically extracted from it (if not explicitly provided).
+            Otherwise, if no meta, actions, or config were provided, will be loaded from the
+            ``charmcraft.yaml`` file in the charm root (falling back to ``metadata.yaml``).
             If the ``charmcraft.yaml`` contains an ``extensions`` key (e.g.
             ``extensions: [flask-framework]``), the extension's metadata,
             config, and actions will be automatically merged in.
         :arg actions: charm actions to use. Needs to be a valid actions.yaml format (as a dict).
-            If none is provided, we will search for a ``actions.yaml`` file in the charm root.
+            If none is provided, will be extracted from ``meta`` if present.
+            Otherwise, if no meta, actions, or config were provided, will be loaded from the
+            ``charmcraft.yaml`` file in the charm root (falling back to ``actions.yaml``).
         :arg config: charm config to use. Needs to be a valid config.yaml format (as a dict).
-            If none is provided, we will search for a ``config.yaml`` file in the charm root.
+            If none is provided, will be extracted from ``meta`` if present
+            Otherwise, if no meta, actions, or config were provided, will be loaded from the
+            ``charmcraft.yaml`` file in the charm root (falling back to ``config.yaml``).
         :arg juju_version: Juju agent version to simulate.
         :arg app_name: App name that this charm is deployed as. Defaults to the charm name as
             defined in the metadata.
@@ -693,12 +700,19 @@ class Context(Generic[CharmType]):
 
         else:
             meta = copy.deepcopy(meta) if meta else {'name': str(charm_type.__name__)}
-            spec = _CharmSpec(
-                charm_type=charm_type,
-                meta=meta,
-                actions=copy.deepcopy(actions),
-                config=copy.deepcopy(config),
-            )
+            if actions is None:
+                actions = meta.pop('actions', None)
+            elif 'actions' in meta:
+                raise ValueError('Cannot specify actions in both charmcraft.yaml and separately')
+            else:
+                actions = copy.deepcopy(actions)
+            if config is None:
+                config = meta.pop('config', None)
+            elif 'config' in meta:
+                raise ValueError('Cannot specify config in both charmcraft.yaml and separately')
+            else:
+                config = copy.deepcopy(config)
+            spec = _CharmSpec(charm_type=charm_type, meta=meta, actions=actions, config=config)
 
         self._charm_spec = spec
         self.charm_root = charm_root
--- a/testing/tests/test_context.py
+++ b/testing/tests/test_context.py
@@ -4,11 +4,12 @@
 from __future__ import annotations
 
 import os
+from typing import Any
 from unittest.mock import patch
 
 import pytest
 from scenario import Context, State
-from scenario.errors import UncaughtCharmError
+from scenario.errors import ContextSetupError, UncaughtCharmError
 from scenario.state import _Event, _next_action_id
 
 import ops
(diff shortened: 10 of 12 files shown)


dimaqq:
+1 on skipping wordpress

.github/workflows/published-charms-tests.yaml
```diff
@@ -26,8 +26,6 @@ jobs:
             charm-root: .
           - charm-repo: canonical/airbyte-k8s-operator
             charm-root: .
-          - charm-repo: canonical/airbyte-ui-k8s-operator
```
dimaqq: **question**: did that get rolled into airbyte-k8s or what happened?

.github/workflows/published-charms-tests.yaml
```diff
@@ -162,9 +166,7 @@ jobs:
             charm-root: .
           - charm-repo: canonical/superset-k8s-operator
             charm-root: .
-          - charm-repo: canonical/tempo-coordinator-k8s-operator
-            charm-root: .
-          - charm-repo: canonical/tempo-worker-k8s-operator
+          - charm-repo: canonical/tempo-operators
```
dimaqq, line 167: Are subdirs automatically detected?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+                basepython = config.get(section, 'basepython')
+                if basepython in ('python3.8', 'python3.9'):
+                    config.set(section, 'basepython', 'python3.10')
+                    modified = True
+
+            # Update envlist to replace py38/py39 with py310
+            if config.has_option(section, 'envlist'):
+                envlist = config.get(section, 'envlist')
+                new_envlist = envlist.replace('py38', 'py310').replace('py39', 'py310')
+                new_envlist = new_envlist.replace('{py38}', '{py310}').replace('{py39}', '{py310}')
```
dimaqq: Should this only be done if there's no py310 env yet?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+                if new_envlist != envlist:
+                    config.set(section, 'envlist', new_envlist)
+                    modified = True
+
+        if modified:
+            with open(tox_config, 'w') as f:
+                config.write(f)
+            print(f'  ✓ Updated {tox_config.relative_to(charm_root)}')
+
+    elif tox_config.suffix == '.toml':
```
dimaqq, line 152: **TIL** 🚀

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+    """Update .python-version file.
+
+    Args:
+        python_version_file: Path to .python-version
+        charm_root: Root directory of charm (for relative path display)
+        max_version: Optional maximum Python version (e.g., "3.12")
+    """
+    content = python_version_file.read_text().strip()
+    # Replace any 3.8 or 3.9 version with 3.10 (or 3.11 if max_version is 3.12)
+    if re.match(r'^3\.[89]', content):
```
dimaqq: `re.match` implies start of input marker.
P.S. this will fail wonderfully when Python reaches versions like 3.80 :P

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+    if not has_allowlist:
+        print('    Creating new allowlist_externals with pip')
+        # Add after section header - find the line after the section header
+        content = re.sub(
+            rf'(\[{re.escape(toml_section)}\]\s*\n)',
+            r'\1allowlist_externals = ["pip"]\n',
+            content,
+        )
+    else:
+        print('    Found existing allowlist_externals, appending pip')
```
dimaqq: This is pretty complex.
That being said I can't necessarily think of anything else, apart from maybe `python -m pip ...`

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+                    use_uv_pip = True
+                    break
+            if config.has_option(section, 'package'):
+                package = config.get(section, 'package')
+                if 'uv' in package:
+                    use_uv_pip = True
+                    break
+
+        if use_uv_pip:
+            print("  Detected tox-uv, will use 'uv pip install'")
```
dimaqq: Would it be better to match pyproject.toml and relock instead?

.github/patch-charm-deps.py
dimaqq, line 1: This script will be hard to debug when one day something breaks.

I don't have a good idea how to make it better, but I feel I have to call this out.

dimaqq (commented):
I'm generally on board with this.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: Yes (see https://github.com/canonical/airbyte-ui-k8s-operator). But it's an automated change picked up by re-running the update script, not something I've manually done.

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: No. This is another change that's been picked up automatically.

.github/patch-charm-deps.py
tonyandrewmeyer: I started out with separate solutions that focused on pyproject.toml. The problem is that there are so many ways that the dependencies can be present, sometimes multiple within a charm:

* pyproject.toml, uv
* pyproject.toml, poetry
* requirements.txt, pip
* dependencies listed in tox

The only thing we definitely have is `tox`, because we don't support running tests with anything else (`make`, `just`) at the moment. So it ended up being cleanest to just lean into using tox, even though it's not what you'd typically do for really managing these dependencies.

.github/patch-charm-deps.py
tonyandrewmeyer:
```suggestion
    if re.fullmatch(r"3\.(8|9)(\.\d+)?", version):
```

.github/workflows/published-charms-tests.yaml
tonyandrewmeyer: I added in the roots and re-ran the update script.

.github/patch-charm-deps.py
tonyandrewmeyer: I don't disagree, although I would say that the previous YAML file approach was also not particularly simple to debug either. I think it is made easier by being a standalone script, and also by being Python rather than a bunch of shell commands. However, that's countered by it doing much more than before :(

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+            if config.has_option(section, 'runner'):
+                runner = config.get(section, 'runner')
+                if 'uv-venv' in runner:
+                    use_uv_pip = True
+                    break
+            if config.has_option(section, 'package'):
+                package = config.get(section, 'package')
+                if 'uv' in package:
+                    use_uv_pip = True
+                    break
```
james-garner-canonical: How about factor this into a helper with early return True and fallthrough return False?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+        True if any sections were patched, False otherwise
+    """
+    tox_config = get_tox_config_path(charm_root)
+    if not tox_config:
+        return False
+
+    is_toml = tox_config.suffix == '.toml'
+
+    if is_toml:
+        # Parse TOML to find sections and detect tox-uv
```
james-garner-canonical: I think it would read easier if we factored these two branches into helpers. It took me a while to realise the core logic, that in both branches we either early return False or fall through return True.
```py
    if not tox_config:
        return False
    if tox_config.suffix == '.toml':
        return patch_tox_testenv_sections_toml(...)
    assert tox_config.suffix == '.ini'
    return patch_tox_testenv_sections_ini(...)
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+    print('✓ Found requirements.txt-based charm')
+    updated = False
+
+    for req_file_name in [
+        'test-requirements.txt',
+        'requirements-charmcraft.txt',
+        'requirements.txt',
+    ]:
+        req_file = charm_root / req_file_name
+        if req_file.exists():
```
james-garner-canonical: What about `for req_file in charm_root.glob('*requirements*.txt'):`?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+                    or re.match(r'^ops\[testing\][ ><=]', line)
+                )
+            ]
+
+            # Add wheel paths
+            lines.append('')
+            lines.append(ops_wheel)
+            lines.append('')
+            lines.append(ops_scenario_wheel)
+
```
james-garner-canonical:
```suggestion
            lines.extend(['', ops_wheel, '', ops_scenario_wheel])
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+    tox_config = get_tox_config_path(charm_root)
+    if tox_config and tox_config.suffix == '.ini':
+        config = configparser.ConfigParser()
+        config.read(tox_config)
+
+        modified = False
+        for section in config.sections():
+            if config.has_option(section, 'deps'):
+                deps = config.get(section, 'deps')
+                # Remove ops and ops-scenario deps
```
james-garner-canonical: Let's define `deps_lines = [l for line in deps.splitlines() if (l := line.strip())]` (or repeat `line.strip()` to avoid the walrus).

Then we can have `new_deps_lines = [line for line in deps_lines if not(re.match ...)]` and `if new_deps_lines != deps_lines:` (or keep `len` for efficiency).

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+
+    # Also patch inline deps in tox config if present
+    tox_config = get_tox_config_path(charm_root)
+    if tox_config and tox_config.suffix == '.ini':
+        config = configparser.ConfigParser()
+        config.read(tox_config)
+
+        modified = False
+        for section in config.sections():
+            if config.has_option(section, 'deps'):
```
james-garner-canonical: How about `if not ...: continue` to keep things flatter?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+                    ):
+                        new_deps_lines.append(line)
+
+                if len(new_deps_lines) != len([ln for ln in deps.split('\n') if ln.strip()]):
+                    print('  Found inline ops deps in tox.ini, patching...')
+                    config.set(
+                        section,
+                        'deps',
+                        '\n    ' + '\n    '.join(new_deps_lines) if new_deps_lines else '',
+                    )
```
james-garner-canonical: I thought the operator precedence was such that this would come out wrong at first, but of course `+` binds tighter than `if`. Some extra brackets might not hurt though -- and here's my preferred spelling while we're at it.
```suggestion
                    content = ('\n    ' + '\n    '.join(new_deps_lines)) if new_deps_lines else ''
                    config.set(section, 'deps', content)
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,649 @@
...
+            lines.append('')
+            lines.append(ops_scenario_wheel)
+
+            req_file.write_text('\n'.join(lines))
+            print(f'    ✓ Updated {req_file_name} with ops 3.x')
+            updated = True
+
+    # Also patch inline deps in tox config if present
+    tox_config = get_tox_config_path(charm_root)
+    if tox_config and tox_config.suffix == '.ini':
```
james-garner-canonical: Flip condition and early return here too?

```suggestion
    if not (tox_config and tox_config.suffix == '.ini'):
        return True
```

james-garner-canonical (commented):
I like the overall approach of switching to a Python script for this more complicated logic. Going via `tox` to avoid too much special casing for all the different ecosystem tools seems like a great idea.

I do find the script hard to digest, I think primarily because there's a lot of long nesting conditionals. I think it would be a great idea to refactor a little bit to use more helpers and early returns / continues to make the high-level logic easier to follow. I've left some suggestions along these lines as I've scanned through.

I'm also wondering if unit tests for the factored out units might be nice to have, but they're not a blocker for me. I do wonder how we'll keep track of what's working as intended, but tracking what's passing and failing across multiple runs could be a separate project altogether.

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+    if use_uv_pip:
+        print("  Detected tox-uv, will use 'uv pip install'")
+
+    # Patch env_run_base (equivalent to [testenv])
+    if 'env_run_base' in data:
+        add_tox_pip_commands_toml(tox_config, 'testenv', ops_wheel, ops_scenario_wheel, use_uv_pip)
+
+    # Patch specific envs
+    if 'env' in data:
+        for env_name in sorted(data['env'].keys()):
```
dimaqq:
```suggestion
        for env_name in data['env']:
```

I would imagine that TOML loader preserves order in dicts, and the output is therefore in stable order.

Or am I missing the reason for sorting the environments?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+    use_uv_pip = _detect_tox_uv_ini(config)
+    if use_uv_pip:
+        print("  Detected tox-uv, will use 'uv pip install'")
+
+    # Find all testenv sections
+    testenv_sections = [s for s in config.sections() if s.startswith('testenv')]
+    if not testenv_sections:
+        return False
+
+    for section in sorted(testenv_sections):
```
dimaqq:
```suggestion
    for section in testenv_sections:
```

Same logic here, I would think.

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+        deps = config.get(section, 'deps')
+        # Remove ops and ops-scenario deps using the helper function
+        new_deps_lines = [
+            line.strip()
+            for line in deps.split('\n')
+            if line.strip() and not _is_ops_dependency_line(line.strip())
+        ]
+
+        original_count = len([ln for ln in deps.split('\n') if ln.strip()])
+        if len(new_deps_lines) == original_count:
```
james-garner-canonical:
```suggestion
        deps_lines = [l for line in deps.splitlines() if (l := line.strip())]
        new_deps_lines = [l for l in deps_lines if not _is_ops_dependency_line(l)]
        if deps_lines == new_deps_lines:
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+        config.read(tox_config)
+        modified = False
+
+        for section in config.sections():
+            # Update basepython to use python3.10 if it's set to 3.8 or 3.9
+            if config.has_option(section, 'basepython'):
+                basepython = config.get(section, 'basepython')
+                if basepython in ('python3.8', 'python3.9'):
+                    config.set(section, 'basepython', 'python3.10')
+                    modified = True
```
james-garner-canonical:
```suggestion
            if config.get(section, 'basepython', fallback=None) in ('python3.8', 'python3.9'):
                config.set(section, 'basepython', 'python3.10')
                modified = True
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+                    config.set(section, 'envlist', new_envlist)
+                    modified = True
+
+        if modified:
+            with open(tox_config, 'w') as f:
+                config.write(f)
+            print(f'  ✓ Updated {tox_config.relative_to(charm_root)}')
+
+    elif tox_config.suffix == '.toml':
+        # Handle tox.toml using regex (no standard library for writing TOML)
```
james-garner-canonical: WDYT about using a library for this?

For example, we could do something like this (just need to install `uv` in the workflow):
```
#!/usr/bin/env -S uv run --script --no-project

# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "tomli-w~=1.2.0",
# ]
# ///
```

and then

```py
original = tomllib.loads(tox_config.read_text())
content = copy.deepcopy(original)
content['basepython'] = 'python3.10'
# ...
if content != original:
    tox_config.write_text(tomli_w.dumps(content))
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+
+            # Update envlist to replace py38/py39 with py310
+            # Only replace if py310 doesn't already exist to avoid duplicates
+            if config.has_option(section, 'envlist'):
+                envlist = config.get(section, 'envlist')
+                new_envlist = envlist
+                if 'py310' not in envlist:
+                    new_envlist = new_envlist.replace('py38', 'py310').replace('py39', 'py310')
+                    new_envlist = new_envlist.replace('{py38}', '{py310}')
+                    new_envlist = new_envlist.replace('{py39}', '{py310}')
```
james-garner-canonical: Are these redundant with the previous line?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+                else:
+                    # py310 already exists, just remove py38/py39 references
+                    new_envlist = re.sub(r'\bpy38\b,?\s*', '', new_envlist)
+                    new_envlist = re.sub(r'\bpy39\b,?\s*', '', new_envlist)
+                    new_envlist = re.sub(r'\{py38\},?\s*', '', new_envlist)
+                    new_envlist = re.sub(r'\{py39\},?\s*', '', new_envlist)
+                    # Clean up any trailing commas or whitespace
+                    new_envlist = re.sub(r',\s*$', '', new_envlist)
+                    new_envlist = re.sub(r',\s*\n', '\n', new_envlist)
```
james-garner-canonical: Would this do the job?
```suggestion
                    envs = [e for e in envlist.split(',') if not re.search(r'py3[89]', e)]
                    new_envlist = ','.join(envs)
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+
+
+def update_tox_python_version(tox_config: Path, charm_root: Path) -> None:
+    """Update Python version in tox configuration (tox.ini or tox.toml).
+
+    Args:
+        tox_config: Path to tox configuration file
+        charm_root: Root directory of charm (for relative path display)
+    """
+    if tox_config.suffix == '.ini':
```
james-garner-canonical, line 150: I think it would improve readability to factor these two branches into helpers.

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+def main() -> int:
+    """Main entry point."""
+    parser = argparse.ArgumentParser(
+        description='Patch charm dependencies to use ops 3.x for compatibility testing'
+    )
+    parser.add_argument('ops_wheel', help='Path to ops wheel file')
+    parser.add_argument('ops_scenario_wheel', help='Path to ops-scenario wheel file')
+    parser.add_argument(
+        '--max-python-version',
+        help="Maximum Python version to support (e.g., '3.12'). "
```
james-garner-canonical:
```suggestion
        help="Maximum Python version to support (e.g., '3.12'). "
        "Must be strictly greater than '3.10'. "
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+            content = re.sub(
+                r'requires-python = ["\']>=3\.[89](\.[0-9]+)?["\']',
+                'requires-python = ">=3.10"',
+                content,
+            )
+        if content != original:
+            pyproject.write_text(content)
+            print(f'  ✓ Updated {pyproject.relative_to(charm_root)}')
+        return
+
```
james-garner-canonical: Can we drop the try/except? I'm OK with charm tests failing if their `pyproject.toml` isn't valid TOML.

```suggestion
    data = tomllib.loads(content)
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+        if new_requires:
+            # Use regex to preserve the exact format (quotes, whitespace)
+            content = re.sub(
+                r'(requires-python\s*=\s*)["\'][^"\']+["\']', f'\\1"{new_requires}"', content
+            )
+            modified = True
+
+    if modified:
+        pyproject.write_text(content)
+        print(f'  ✓ Updated {pyproject.relative_to(charm_root)}')
```
james-garner-canonical: We can simplify the control flow with early returns. I'm also suggesting writing back out with `tomli_w`, but the general suggestion applies even if we stick with the regex based approach.
```suggestion
    requires_python = data.get('project', {}).get('requires-python')
    if requires_python is None:
        return
    if max_version:
        new_requires = f'>=3.10,<{max_version}'
    elif re.match(r'>=3\.[89]', requires_python):
        new_requires = '>=3.10'
    else:
        return
    data['project']['requires-python'] = new_requires
    pyproject.write_text(tomli_w.dumps(data))
    print(f'  ✓ Updated {pyproject.relative_to(charm_root)}')
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+    Args:
+        python_version_file: Path to .python-version
+        charm_root: Root directory of charm (for relative path display)
+        max_version: Optional maximum Python version (e.g., "3.12")
+    """
+    content = python_version_file.read_text().strip()
+    # Replace any 3.8 or 3.9 version with 3.10 (or 3.11 if max_version is 3.12)
+    # Use fullmatch with explicit minor versions to avoid matching 3.80, 3.81, etc.
+    if re.fullmatch(r'3\.(8|9)(\.\d+)?', content):
+        new_version = '3.11' if max_version == '3.12' else '3.10'
```
james-garner-canonical: This logic is surprising. Does our one case that needs Python capped at 3.12 also need a minimum version of 3.11? One approach might be for the script to take a min_version parameter as well, which could be set to 3.11 in that case. A simpler fix might be this:

```suggestion
        new_version = max_version or '3.10'
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+
+    Returns:
+        True if any sections were patched, False otherwise
+    """
+    try:
+        with open(tox_config, 'rb') as f:
+            data = tomllib.load(f)
+    except tomllib.TOMLDecodeError:
+        print(f'  ⚠ Warning: Could not parse {tox_config.name}')
+        return False
```
james-garner-canonical: Shouldn't we just crash if the charm has a `tox.toml` that isn't valid TOML?
```suggestion
    data = tomllib.loads(tox_config.read_text())
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+
+    # Track whether we successfully updated any dependencies
+    updated = False
+
+    # Detect dependency management system and patch accordingly
+    print('\nDetecting dependency management system...')
+
+    # 1. Handle requirements.txt-based charms
+    req_files = list(args.charm_root.glob('*requirements*.txt'))
+    if req_files:
```
james-garner-canonical:
```suggestion
    if list(args.charm_root.glob('*requirements*.txt')):
```

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+    if args.max_python_version:
+        print(f'MAX PYTHON VERSION: {args.max_python_version}')
+    print('=========================================')
+
+    # Update Python version requirements
+    update_python_version_requirements(args.charm_root, args.max_python_version)
+
+    # Track whether we successfully updated any dependencies
+    updated = False
+
```
james-garner-canonical: How about moving `updated = False` to the `else` branch?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+    # Add pip to allowlist_externals
+    if config.has_option(section, 'allowlist_externals'):
+        allowlist = config.get(section, 'allowlist_externals')
+        if 'pip' not in allowlist.split():
+            print('    Found existing allowlist_externals, appending pip')
+            # Preserve multi-line format if it exists
+            if '\n' in allowlist:
+                config.set(section, 'allowlist_externals', allowlist + '\n    pip')
+            else:
+                config.set(section, 'allowlist_externals', allowlist + '\n    pip')
```
james-garner-canonical: What we do in each branch looks identical to me.

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,714 @@
...
+        if 'package' in env_data and 'uv' in str(env_data['package']):
+            return True
+
+    return False
+
+
+def add_tox_pip_commands_ini(
+    tox_ini_path: Path, section: str, ops_wheel: str, ops_scenario_wheel: str, use_uv_pip: bool
+) -> None:
+    """Add pip to allowlist_externals and commands_post to force-reinstall ops wheels (INI format).
```
james-garner-canonical: I don't understand this approach. Won't `commands_post` install our wheels *after* the `commands` (tests) have run? Should we be updating `commands_pre` instead?

james-garner-canonical (changes requested):
I'm getting my head around this new approach. My biggest question is: is patching `commands_post` actually doing what we want, or should we be patching `commands_pre`?

I've made a number of suggestions that I think will improve the script's digestibility as I go through the logic. The logic itself seems solid, though it's difficult to verify.

tonyandrewmeyer:
I've made a bunch of adjustments based on the review feedback, and working through the latest version in more detail (after lots of patching here and there, looking more holistically). The failure count is also down to 6 or 7, which is great, and with the extra logging in the output, I'm pretty confident that the patching is actually happening.

.github/workflows/published-charms-tests.yaml
```diff
@@ -202,67 +217,66 @@ jobs:
...
-            uv lock
+
+          # Use --max-python-version for wordpress-k8s to cap at 3.11
+          if [ "${{ matrix.charm-repo }}" = "canonical/wordpress-k8s-operator" ]; then
+            python /tmp/operator-scripts/patch-charm-deps.py "${ops_wheel}" "${ops_scenario_wheel}" --max-python-version=3.12
           else
-            echo "Error: No requirements.txt or poetry.lock or uv.lock file found"
-            exit 1
+            python /tmp/operator-scripts/patch-charm-deps.py "${ops_wheel}" "${ops_scenario_wheel}"
           fi
```
james-garner-canonical: The script has a shebang, so we should presumably make sure it's marked as executable when checked in, and run it directly here instead of with `python`.

But running with `python` like we're doing here instead of executing the script directly should bypass the shebang, shouldn't it?

Which means we're not running with `uv` here, and presumably not getting our deps (`tomli-w`) installed) ... so what's going on here?

.github/workflows/published-charms-tests.yaml
```diff
@@ -202,67 +217,66 @@ jobs:
...
-            poetry add "${ops_wheel}" --lock
-            poetry add "${ops_scenario_wheel}" --lock
-          elif [ -e "uv.lock" ]; then
-            uv add --frozen --raw-sources "${ops_wheel}"
-            uv add --frozen --raw-sources "${ops_scenario_wheel}"
-            uv lock
+
+          # Use --max-python-version for wordpress-k8s to cap at 3.11
+          if [ "${{ matrix.charm-repo }}" = "canonical/wordpress-k8s-operator" ]; then
+            python /tmp/operator-scripts/patch-charm-deps.py "${ops_wheel}" "${ops_scenario_wheel}" --max-python-version=3.12
```
james-garner-canonical: It seems that `--max-python-version` is a bit of a misnomer since we actually cap to `<` it. A bit of a pain, but could be a source of confusion in future, so perhaps worth either calling here with `--max-python-version=3.11` and writing `<=`; or renaming the option to something like `--python-less-than`.

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,768 @@
...
+        result = subprocess.run(
+            ['uv', 'lock', '--python-preference', 'system'],
+            cwd=charm_root,
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode == 0:
+            print('  ✓ Updated uv.lock')
+        else:
+            print(f'  ✗ Failed to update uv.lock: {result.stderr.strip()}')
```
james-garner-canonical, line 192: Should the script fail if we can't re-lock?

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,768 @@
...
+    modified = False
+
+    # Update allowlist_externals
+    if 'allowlist_externals' not in current:
+        print('    Creating new allowlist_externals with pip')
+        current['allowlist_externals'] = ['pip']
+        modified = True
+    elif (
+        isinstance(current['allowlist_externals'], list)
+        and 'pip' not in current['allowlist_externals']
```
james-garner-canonical: Why the `isinstance` check here? Is this more about convincing the type checker? If it's not a list, I reckon we should probably bail out rather than silently skipping.

.github/patch-charm-deps.py
```diff
@@ -0,0 +1,768 @@
...
+    pyproject = charm_root / 'pyproject.toml'
+    scenario_group = None
+    if pyproject.exists():
+        data = tomllib.loads(pyproject.read_text())
+        # Check dependency-groups for ops-scenario
+        for group_name, group_deps in data.get('dependency-groups', {}).items():
+            for dep in group_deps:
+                if isinstance(dep, str) and re.match(r'^ops-scenario\b', dep):
+                    scenario_group = group_name
+                    break
```
james-garner-canonical: This `break` will only break out of the inner loop. So the semantics here are that `scenario_group` is the last group that has `ops-scenario` in it.

dimaqq:
@tonyandrewmeyer pls add a note on the direction based on today's standup discussion.

.github/patch-charm-deps.py
tonyandrewmeyer: Yeah, really we should be handling multiple groups if that is the case. So not breaking out at all.

tonyandrewmeyer:
It seems to me that this "patch ops" code would be useful elsewhere (particularly in [tonyandrewmeyer/charm-analysis](https://github.com/tonyandrewmeyer/charm-analysis). So moving it to a small package and having ops use that may be the better choice. It would also make it simpler to break the big module into several smaller ones, add tests, and so on.

However, we're fine with it going into canonical/operator like it is for now, and can factor it out later.

PR #2322, opened 2026-02-11
chore: bump versions and update changelog for 2.23.2 release

james-garner-canonical:
This PR bumps the package versions and updates the changelog for the 2.23.2 release.

Because `release.py` doesn't exist on this branch, this PR was partially created manually. I ran `release.py` to successfully update the changelog (and manually added the fix we're releasing for as we did when releasing from `main`) and bump the versions, but it then crashed due to `versions.md` not existing. I ran `uv lock` manually, and created a branch and PR manually too.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # 2.23.2 - 11 February 2026
 
+## Fixes
+
+* Drop unused `setuptools_scm` build dependency (#2318)
+
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
Annoyingly, `uv lock` has produced many lines of diff, but I guess we should let it do its thing.

tonyandrewmeyer (approved):

PR #2326, opened 2026-02-13
feat: typed testing relation getters

### LINKED ISSUE #2243: Improve typing UX of ops.testing.State.get_relation
[ops.testing.State.get_relation(relation: int)](https://documentation.ubuntu.com/ops/latest/reference/ops-testing/#ops.testing.State.get_relation) is typed as returning an `ops.testing.RelationBase`, because it could return any of `Relation`, `PeerRelation`, and `SubordinateRelation`.

When type checking tests, code like this is often needed:
```
retrieved_relation = state.get_relation(relation.id)
assert isinstance(retrieved_relation, Relation)
```
Which isn't terrible, but as @dimaqq [noted](https://github.com/canonical/operator/pull/2234/files#r2621201824), could be better:
> **Aside:** I wish this was more ergonomic for our users.
> 
> Like if they could conceptually write:
> 
> ```python
> rel = state.get_relation<Relation>(rel.id)
> # or
> rel = state.get_relation(rel.id, kind=Relation)  # runtime error on mismatch
> ```
> 
> Perhaps more Pythonically:
> 
> ```python
> rel = state.get_regular_relation(rid)
> per = state.get_peer_relation(pid)
> sub = state.get_subordinate_relation(sid)
> ```

james-garner-canonical:
This PR adds more narrowly typed methods alongside `get_relation` and `get_relations`, like `get_subordinate_relation` and `get_peer_relations`. These are typed as returning the correct relation type, rather than just `RelationBase`, removing the need to narrow the relation type afterwards in tests. If the relation type is incorrect, a `TypeError` is raised.

Resolves #2243.

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1818,7 +1818,9 @@ class State:
         """
         return self._get_typed_relations(endpoint, kind=SubordinateRelation)
 
-    def _get_typed_relations(self, endpoint: str, kind: type[_RelationType]) -> tuple[_RelationType, ...]:
+    def _get_typed_relations(
+        self, endpoint: str, kind: type[_RelationType]
+    ) -> tuple[_RelationType, ...]:
         rels = self.get_relations(endpoint)
         for rel in rels:
             if not isinstance(rel, kind):


testing/src/scenario/state.py
```diff
@@ -1746,6 +1747,35 @@ def get_relation(self, relation: int, /) -> RelationBase:
                 return state_relation
         raise KeyError(f'relation: id={relation} not found in the State')
 
+    def get_regular_relation(self, relation: int, /) -> Relation:
```
tonyandrewmeyer, line 1750: I don't really love the word "regular here", but I'm also not sure what to suggest instead.

testing/src/scenario/state.py
```diff
@@ -1746,6 +1747,35 @@ def get_relation(self, relation: int, /) -> RelationBase:
...
+        """Get a regular relation from this State by relation id.
+
+        Raises a ``TypeError`` if this relation is not a regular ``Relation``.
```
tonyandrewmeyer: This (and the ones below) should use the standard `Raises:` format, I would think.

testing/src/scenario/state.py
```diff
@@ -1746,6 +1747,35 @@ def get_relation(self, relation: int, /) -> RelationBase:
...
+
+        Raises a ``TypeError`` if this relation is not a ``SubordinateRelation``.
+        """
+        return self._get_typed_relation(relation, kind=SubordinateRelation)
+
+    def _get_typed_relation(self, relation: int, kind: type[RelationType]) -> RelationType:
+        rel = self.get_relation(relation)
+        if not isinstance(rel, kind):
+            raise TypeError(
+                f"Relation {relation} is not a {kind.__name__}, it's {rel.__class__.__name__}"
```
```suggestion
                f"Relation {relation} is not a {kind.__name__}, it's a {rel.__class__.__name__}"
```

testing/src/scenario/state.py
```diff
@@ -1758,6 +1788,40 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+        """
+        return self._get_typed_relations(endpoint, kind=SubordinateRelation)
+
+    def _get_typed_relations(self, endpoint: str, kind: type[RelationType]) -> RelationType:
+        rels = self.get_relations(endpoint)
+        for rel in rels:
+            if not isinstance(rel, kind):
+                raise TypeError(
+                    f'Relation on endpoint {endpoint} is not a {kind.__name__}'
+                    f", it's {rel.__class__.__name__}"
```
```suggestion
                    f", it's a {rel.__class__.__name__}"
```

testing/src/scenario/state.py
```diff
@@ -1758,6 +1788,40 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
             r for r in self.relations if _normalise_name(r.endpoint) == normalized_endpoint
         )
 
+    def get_regular_relations(self, endpoint: str) -> tuple[Relation, ...]:
```
tonyandrewmeyer, line 1794: I'm not convinced about adding the plural/endpoint ones. I'm pretty sure when we've looked at use of this in the past, it's almost always code like:

```python
state_in = State(..., relations={relation})
state_out = ctx.run(..., state=state_in)
assert state_out.get_relations('endpoint')[0]...
```

And that is cleaner as:

```python
state_in = State(..., relations={relation})
state_out = ctx.run(..., state=state_in)
assert state_out.get_relation(relation.id)...
```

I had a quick look over a few random selections in charms, and they were all of that type. I'm sure there are exceptions, but this is already adding 3 new methods for getting by ID - I'm not convinced there is enough value to add another 3 as well.

testing/src/scenario/state.py
```diff
@@ -69,6 +69,7 @@ class _StateKwargs(TypedDict, total=False):
 UnitID = int
 
 CharmType = TypeVar('CharmType', bound=CharmBase)
+RelationType = TypeVar('RelationType', bound='RelationBase')
```
tonyandrewmeyer: Since this is only used by private methods, can it also be private?

testing/src/scenario/state.py
james-garner-canonical: Yeah I don't love it either. I went with it because the Juju [docs on relations](https://documentation.ubuntu.com/juju/3.6/reference/relation/) basically go Peer, Non-Peer, Subordinate, Non-Subordinate, and then:
> A non-subordinate relation (aka ‘regular’) is a [non-peer](https://documentation.ubuntu.com/juju/3.6/reference/relation/#non-peer-relation) relation where the applications are both principal.

I'm guessing that's why Dima suggested it too. Open to alternatives though.

testing/src/scenario/state.py
james-garner-canonical: Thanks, fixed.

testing/src/scenario/state.py
james-garner-canonical: Good idea, forgot this whole module is public.

testing/src/scenario/state.py
james-garner-canonical: Makes sense. I'll let @dimaqq weigh in too, but I'm happy to remove the plural ones.

james-garner-canonical: I suppose my one hesitation would be that getting relations by endpoint name rather than by ID feels more like the right level of expression for state-transition tests. But the current usage argument is pretty compelling.

testing/src/scenario/state.py
```diff
@@ -1746,6 +1747,38 @@ def get_relation(self, relation: int, /) -> RelationBase:
...
+
+        Raises:
+            TypeError: If this relation is not a ``SubordinateRelation``.
+        """
+        return self._get_typed_relation(relation, kind=SubordinateRelation)
+
+    def _get_typed_relation(self, relation: int, kind: type[_RelationType]) -> _RelationType:
+        rel = self.get_relation(relation)
+        if not isinstance(rel, kind):
+            raise TypeError(
```
dimaqq, line 1777: I wonder if this should be:
- a `TypeError` (wrong type of relation about to be returned), or
- a `ValueError` (wrong `relation: int` argument passed in)

testing/src/scenario/state.py
```diff
@@ -1758,6 +1791,45 @@ def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
...
+                a ``SubordinateRelation``.
+        """
+        return self._get_typed_relations(endpoint, kind=SubordinateRelation)
+
+    def _get_typed_relations(
+        self, endpoint: str, kind: type[_RelationType]
+    ) -> tuple[_RelationType, ...]:
+        rels = self.get_relations(endpoint)
+        for rel in rels:
+            if not isinstance(rel, kind):
```
dimaqq, line 1826: A zero-length return is valid, isn't it?
Then there's a possibility of a false positive, that is not raising an exception when we would have if there was a relation.

dimaqq:
Overall I'm not opposed to it.

james-garner-canonical:
The API surface area (+3 or +6 methods) is actually swaying me towards the idea of adding a `kind=...` argument to the existing 2 methods. If we were designing the API from scratch, I think the separate getters would be the way to go (and we would have `get_relation` instead of `get_regular_relation`). But given that `get_relation` and `get_relations` already exist, it seems like more functionality for less extra API commitment to add the `kind` argument.

I think I'll create an alternative PR with that implementation.

testing/src/scenario/state.py
james-garner-canonical: > A zero-length return is valid, isn't it? Then there's a possibility of a false positive, that is not raising an exception when we would have if there was a relation.

That's correct, though I don't think we can do any better here. The `Context` knows how the relation is defined in the metadata, but the `State` does not, it only knows what's in the current state.

james-garner-canonical:
Closing in favour of:
- #2359

PR #2329, opened 2026-02-16
fix: correct the `Model.get_binding()` return type

tonyandrewmeyer:
Although the `._bindings` attribute is a mapping, and we are using `get()`, the implementation never returns `None`, it always creates a new `Binding` if there isn't one. The signature for `get()` is correct and does not include `None` as a possible return type. This PR extends that to the `get_binding()` method.

Issue reported by @Gu1nness on Matrix.


james-garner-canonical (approved):
Nice find, @Gu1nness

PR #2330, opened 2026-02-17
docs: use "true" and "false" consistently in the reference documentation

### LINKED ISSUE #2291: Use consistent "true" strings in reference documentation
We're a bit inconsistent across our reference docs in terms of "true", "True", and "`True`" (and the same for false).

We should standardise on:

* "true" essentially means "truth-y", like "if true, then foo will bar"
* "`True`" means the Python `True` object, like "pass `True` to also foo the value"
* Not using "True" (or any other variation).

We should make a small note about this in our style guide and update the existing docs. In some cases, we might be able to get consistency through adjusting some phrases, like unifying on "if true, then x" rather than having both that and "set to `True` to x".

tonyandrewmeyer:
Adjust the docstrings so that true and false are used consistently, and also adjust the style guide to explain our practice.

"true" and "false" mean "truth-y" and "false-y". `True` and `False` mean the Python objects. We don't use "True" or "False" (unless the capitalisation is from the start of a sentence).

Fixes #2291

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -184,7 +184,7 @@ class ActionFailed(Exception):  # noqa: N818 (name doesn't end with "Error")
     state: State | None
     """The Juju state after the action has been run.
 
-    When using Harness.run_action, this will be None.
+    When using Harness.run_action, this will be ``None``.
     """
 
     def __init__(
@@ -1859,7 +1859,7 @@ class Harness(Generic[CharmType]):
 
         Args:
             secret_id: The ID of the secret associated with the event.
-            label: Label value to send to the event. If None, the secret's
+            label: Label value to send to the event. If ``None``, the secret's
                 label is used.
         """
         secret = self._ensure_secret(secret_id)
@@ -1880,7 +1880,7 @@ class Harness(Generic[CharmType]):
             secret_id: The ID of the secret associated with the event.
             revision: Revision number to provide to the event. This should be
                 an item from the list returned by :meth:`get_secret_revisions`.
-            label: Label value to send to the event. If None, the secret's
+            label: Label value to send to the event. If ``None``, the secret's
                 label is used.
         """
         secret = self._ensure_secret(secret_id)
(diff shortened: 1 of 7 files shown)


ops/_private/harness.py
```diff
@@ -450,7 +450,7 @@ def begin_with_initial_hooks(self) -> None:
         should be active when the charm starts, and then call this method. This method will
         automatically create and add peer relations that are specified in metadata.yaml.
 
-        If the charm metadata specifies containers, this sets can_connect to True for all
+        If the charm metadata specifies containers, this sets can_connect to `True` for all
```
dimaqq: Docstrings are RST and not MD, aren't they?

```suggestion
        If the charm metadata specifies containers, this sets can_connect to ``True`` for all
```

Here and in many other places.

tonyandrewmeyer: Oops, thanks!

ops/_private/harness.py
```diff
@@ -2599,7 +2599,10 @@ def _storage_detach(self, storage_id: str):
             self._storage_attached[name].remove(index)
 
     def _storage_attach(self, storage_id: str):
-        """Mark the named storage_id as attached and return True if it was previously detached."""
+        """Mark the named storage_id as attached.
+
+        Return ``True`` if it was previously detached.
```
james-garner-canonical:
```suggestion
        Returns:
            ``True`` if it was previously detached.
```

james-garner-canonical (approved):
Changes look good and more consistent than we had previously. The updated contribution guidance is nice.

I think sometimes it's not clear whether 'true' or '`True`' would fit better, but I think it's a nice improvement just eliminating 'True'.

ops/pebble.py
```diff
@@ -540,9 +540,9 @@ class ExecError(Error, Generic[AnyStr]):
     """Standard error from the process.
 
     If :meth:`ExecProcess.wait_output` was being called and ``combine_stderr``
-    was False, this is the captured stderr as a str (or bytes if encoding was
+    was ``False``, this is the captured stderr as a str (or bytes if encoding was
     None). If :meth:`ExecProcess.wait` was being called or ``combine_stderr``
-    was True, this is None.
+    was ``True``, this is None.
```
dimaqq, line 545: **possibly out of scope**: should be double-back-tick all the None's?

dimaqq (approved):

dimaqq (approved):

PR #2331, opened 2026-02-17
refactor: distinguish between init and attribute types in testing state classes

### LINKED ISSUE #2152: ops[testing] State.config type, should it be dict or Mapping?
Currently it's declared like this:

```py
  class State(_max_posargs(0)):
      """Represents the Juju-owned portion of a unit's state.

      Roughly speaking, it wraps all hook-tool- and pebble-mediated data a charm can access in its
      lifecycle. For example, status-get will return data from `State.unit_status`, is-leader will
      return data from `State.leader`, and so on.
      """

      config: dict[str, str | int | float | bool] = dataclasses.field(
          default_factory=dict,
      )
```

I'm trying to use it like so and pyright is unhappy:

```py
def caller():
    config = {"init": init}
    # reportArgumentType: Argument of type "dict[str, str]" cannot be assigned to pa...
    # Consider switching from "dict" to "Mapping" which is covariant in the value type
    do_scenario_thing(config)

def do_scenario_thing(config: dict[str, str | int | float | bool]):
    state = State(..., config=config)
```

But if I switch to Mapping, pyright is still unhappy:

```py
def caller():
    config = {"init": init}
    do_scenario_thing(config)

def do_scenario_thing(config: Mapping[str, str | int | float | bool]):
    # reportArgumentType: Argument of type "Mapping[str, str | int | float | bool]" is not assignable to "dict[str, str | int | float | bool]"
    state = State(..., config=config)
```

I wonder if this is a fundamental issue, because a dataclass field's type is used both when:
- dataclass is created with a value
- dataclass field is read

tonyandrewmeyer:
When designing the Scenario 7 API we introduced kw-only args, and originally had custom `__init__` for each state class to support that. We decided to change that because it felt busy and a lot of maintenance.

However, we currently have an unfortunate mismatch between some of the types accepted to create an instance of a state class and the type the corresponding attribute will be. For example, the init might accept any `Mapping` but we know the attribute will always be a `dict`. It would be nice to provide that information to users.

Now that we are using Python 3.10+, we do have some classes without this issue that can continue to use the dataclasses generated `__init__`. However, there are many that would be better as more explicit, and I am not convinced it's too much work to maintain.

We opened the door to this in #2274 adjusting `CheckInfo`. This PR applies the same improvement to the rest of the state classes.

A couple of small behaviour changes fall out of this:

- The `*_addresses`, `*_subnets`, `redacted`, `ca_certificates`, and `Secret.remote_grants` arguments now reject a bare `str` (or `bytes`) up front. Previously this would have silently iterated the string character-by-character, leaving you with `['s', 'o', 'm', ...]` in the attribute.
- Passing `app_status=None` or `unit_status=None` to `State` now coerces to `UnknownStatus()`, where previously it would have raised `TypeError`. Explicit `None` is now treated the same as omitting the argument.

Fixes #2152


testing/src/scenario/state.py
```diff
@@ -179,43 +188,66 @@ def _to_ops(self) -> CloudCredential_Ops:
...
+        self,
+        type: str,
+        *,
+        name: str = 'localhost',
+        region: str | None = None,
+        endpoint: str | None = None,
+        identity_endpoint: str | None = None,
+        storage_endpoint: str | None = None,
+        credential: CloudCredential | None = None,
+        ca_certificates: Sequence[str] | None = None,
```
james-garner-canonical: How about `Iterable[str] | None = None` and make the logic `list(ca_certificates) if ca_certificates is not None else []`?

testing/src/scenario/state.py
```diff
@@ -253,59 +285,83 @@ class Secret:
...
-    _latest_revision: int = 1
+    _latest_revision: int
+
+    def __init__(
+        self,
+        tracked_content: RawSecretRevisionContents,
+        *,
+        latest_content: RawSecretRevisionContents | None = None,
+        id: str | None = None,
+        owner: Literal['unit', 'app', None] = None,
```
james-garner-canonical: This feels more idiomatic to me:
```suggestion
        owner: Literal['unit', 'app'] | None = None,
```

testing/src/scenario/state.py
```diff
@@ -253,59 +285,83 @@ class Secret:
     This is the content the charm will receive with a
     :meth:`ops.Secret.get_content` call."""
 
-    _: dataclasses.KW_ONLY
-
-    latest_content: RawSecretRevisionContents | None = None
+    latest_content: RawSecretRevisionContents | None
```
james-garner-canonical: Looks like this shouldn't be `None`-able, since we set it to `tracked_content` if the argument is `None`.
```suggestion
    latest_content: RawSecretRevisionContents
```

testing/src/scenario/state.py
```diff
@@ -253,59 +285,83 @@ class Secret:
...
+        latest_content: RawSecretRevisionContents | None = None,
+        id: str | None = None,
+        owner: Literal['unit', 'app', None] = None,
+        remote_grants: Mapping[int, set[str]] | None = None,
+        label: str | None = None,
+        description: str | None = None,
+        expire: datetime.datetime | None = None,
+        rotate: SecretRotate | None = None,
+        _tracked_revision: int = 1,
+        _latest_revision: int = 1,
```
james-garner-canonical: It's odd that these arguments are underscored, but it has to stay this way for backwards compatibility. WDYT about adding a comment as to why they're underscored, if you know? Are they intended to be private arguments, not set by users directly?

testing/src/scenario/state.py
```diff
@@ -253,59 +285,83 @@ class Secret:
...
+        object.__setattr__(self, 'id', id if id is not None else _generate_secret_id())
+        object.__setattr__(self, 'owner', owner)
+        object.__setattr__(self, 'remote_grants', dict(remote_grants) if remote_grants else {})
+        object.__setattr__(self, 'label', label)
+        object.__setattr__(self, 'description', description)
+        object.__setattr__(self, 'expire', expire)
+        object.__setattr__(self, 'rotate', rotate)
+        object.__setattr__(self, '_tracked_revision', _tracked_revision)
+        object.__setattr__(self, '_latest_revision', _latest_revision)
+        _deepcopy_mutable_fields(self)
```
james-garner-canonical: WDYT about inlining the `deepcopy` calls above? Looks like it would just be `tracked_content`, `latest_content`, and `remote_grants`.

I'm OK with leaving it as-is to keep the PR simpler if that's your preference.

testing/src/scenario/state.py
```diff
@@ -150,25 +150,34 @@ class JujuLogLine:
...
+        self,
+        *,
+        auth_type: str,
+        attributes: Mapping[str, str] | None = None,
+        redacted: Sequence[str] | None = None,
+    ):
+        object.__setattr__(self, 'auth_type', auth_type)
+        object.__setattr__(self, 'attributes', dict(attributes) if attributes else {})
+        object.__setattr__(self, 'redacted', list(redacted) if redacted else [])
         _deepcopy_mutable_fields(self)
```
james-garner-canonical: None of these fields should need a `deepcopy`, since the collections are all flat, and we already call `dict` and `list` on them, so we could drop this line.

testing/src/scenario/state.py
```diff
@@ -179,43 +188,66 @@ def _to_ops(self) -> CloudCredential_Ops:
...
+        object.__setattr__(self, 'endpoint', endpoint)
+        object.__setattr__(self, 'identity_endpoint', identity_endpoint)
+        object.__setattr__(self, 'storage_endpoint', storage_endpoint)
+        object.__setattr__(self, 'credential', credential)
+        object.__setattr__(
+            self, 'ca_certificates', list(ca_certificates) if ca_certificates else []
+        )
+        object.__setattr__(self, 'skip_tls_verify', skip_tls_verify)
+        object.__setattr__(self, 'is_controller_cloud', is_controller_cloud)
         _deepcopy_mutable_fields(self)
```
james-garner-canonical: Do any of these actually need a `deepcopy`?

testing/src/scenario/state.py
```diff
@@ -423,28 +479,45 @@ class Network:
...
-    egress_subnets: Sequence[str] = dataclasses.field(
-        default_factory=lambda: ['192.0.2.0/24'],
-    )
+    egress_subnets: list[str]
     """Subnets that other units will see the charm connecting from."""
 
+    def __init__(
+        self,
+        binding_name: str,
+        bind_addresses: Sequence[BindAddress] | None = None,
```
james-garner-canonical: `BindAddress` and `Address` are both immutable (frozen dataclasses), so how about this:
```suggestion
        bind_addresses: Sequence[BindAddress] = (BindAddress((Address('192.0.2.0'),)),),
```
Would let the init logic be:
```py
        object.__setattr__(self, 'bind_addresses', list(bind_addresses))
```
Though the mess of commas in the single item tuples is basically enough to put me off this idea ...

testing/src/scenario/state.py
```diff
@@ -423,28 +479,45 @@ class Network:
...
+        *,
+        ingress_addresses: Sequence[str] | None = None,
+        egress_subnets: Sequence[str] | None = None,
+    ):
+        object.__setattr__(self, 'binding_name', binding_name)
+        object.__setattr__(
+            self,
+            'bind_addresses',
+            list(bind_addresses)
```
james-garner-canonical: What if we used `tuple` instead of `list` throughout this PR? Since the typing was `Sequence` previously, this should still be backwards compatible. And it reflects the intention of the class to be immutable/frozen.

testing/src/scenario/state.py
```diff
@@ -842,43 +923,72 @@ class Notice:
...
-    id: str = dataclasses.field(default_factory=_next_notice_id)
+    id: str
     """Unique ID for this notice."""
 
-    user_id: int | None = None
+    user_id: int | None
     """UID of the user who may view this notice (None means notice is public)."""
 
-    type: pebble.NoticeType | str = pebble.NoticeType.CUSTOM
+    type: pebble.NoticeType | str
```
james-garner-canonical, line 968: Could we narrow this further to `pebble.NoticeType` and convert in the constructor? Perhaps that would be best left for a separate PR.

testing/src/scenario/state.py
```diff
@@ -1065,21 +1171,43 @@ class Container:
...
+        self,
+        name: str,
+        *,
+        can_connect: bool = False,
+        _base_plan: Mapping[str, Any] | None = None,
+        layers: Mapping[str, pebble.Layer] | None = None,
+        service_statuses: Mapping[str, pebble.ServiceStatus] | None = None,
+        mounts: Mapping[str, Mount] | None = None,
+        execs: Iterable[Exec] | None = None,
+        notices: Sequence[Notice] | None = None,
```
james-garner-canonical: Could this be `Iterable` instead?

testing/src/scenario/state.py
```diff
@@ -1065,21 +1171,43 @@ class Container:
...
+        check_infos: Iterable[CheckInfo] | None = None,
```
james-garner-canonical: The default could be `()` and we could drop `| None`, right? Since these arguments didn't accept `None` previously.

testing/tests/test_e2e/test_state.py
```diff
@@ -385,7 +385,6 @@ def test_replace_state():
         (Container, 'layers', {'name': 'foo'}),
         (Container, 'service_statuses', {'name': 'foo'}),
         (Container, 'mounts', {'name': 'foo'}),
-        (Container, 'notices', {'name': 'foo'}),
```
james-garner-canonical, line 389: Is moving this from one test to the other just catching an existing error in categorisation, and essentially otherwise unrelated to this PR? Since the tests use `__setattr__`, I guess it works either way -- is that right?

james-garner-canonical (commented):
I'm a big fan of making this change, thanks for taking care of this. I have a number of suggestions around typing and defaults, which I've made on individual lines, though they typically apply to more lines across the PR -- but I figured I'd keep the comments fewer than they'd otherwise be ...

Do you think this change warrants some additional unit tests, or are you happy that the existing tests would catch any errors in this PR? If the latter, please mention the relevant test suites.

tonyandrewmeyer:
> Do you think this change warrants some additional unit tests, or are you happy that the existing tests would catch any errors in this PR? If the latter, please mention the relevant test suites.

In terms of not breaking the `__init__` signature, I feel like the [tests from here downwards](https://github.com/canonical/operator/blob/7d5b9bab60672c5a076b4f84008b6c013d693b7c/testing/tests/test_e2e/test_state.py#L302) should cover that (right number of positional/keyword-only arguments in particular), and they should also cover the behaviour of `__init__` in terms of forcing immutability (same file, the tests following on from the previous ones). So I feel comfortable that any regressions in this PR would be caught.

In terms of tests for the changes, I'm not super keen on having tests like:

```python
c = CloudCredential(auth_type="foo", redacted=['a', 'b', 'c'])
assert isinstance(c.redacted, list)
```

I know we have some tests where we expect pyright to find issues, but I'm not sure it's the right move to add something like that for this either.

Do you have any suggestions in terms of tests?

testing/src/scenario/state.py
tonyandrewmeyer: Actually, I guess because they're private, I can break backwards compatibility. It would be better if they were not in the `__init__` arguments at all (but would still have the default value of `1`). The value gets changed only by internal code - the user is not meant to touch these at all (hence the private-ness).

james-garner-canonical:
> > Do you think this change warrants some additional unit tests, or are you happy that the existing tests would catch any errors in this PR? If the latter, please mention the relevant test suites.
> 
> In terms of not breaking the `__init__` signature, I feel like the [tests from here downwards](https://github.com/canonical/operator/blob/7d5b9bab60672c5a076b4f84008b6c013d693b7c/testing/tests/test_e2e/test_state.py#L302) should cover that (right number of positional/keyword-only arguments in particular), and they should also cover the behaviour of `__init__` in terms of forcing immutability (same file, the tests following on from the previous ones). So I feel comfortable that any regressions in this PR would be caught.

Nice!
 
> In terms of tests for the changes, I'm not super keen on having tests like:
> 
> ```python
> c = CloudCredential(auth_type="foo", redacted=['a', 'b', 'c'])
> assert isinstance(c.redacted, list)
> ```
> 
> I know we have some tests where we expect pyright to find issues, but I'm not sure it's the right move to add something like that for this either.
> 
> Do you have any suggestions in terms of tests?

I wouldn't mind seeing tests a bit like the one you're not keen on, explicitly encoding (from a user perspective) the type conversion and copying behaviour that we're implementing.
```py
redacted = ['a', 'b', 'c']
...
c = CloudCredential(auth_type="foo", redacted=redacted, ...)
assert isinstance(c.redacted, list)  # or tuple if we go that way
assert c.redacted == redacted
assert c.redacted is not redacted
...
```

The existing tests probably do cover a lot of this, but a lot of them are hard to follow at a glance due to the parametrization and abstraction.

testing/src/scenario/state.py
tonyandrewmeyer: We somewhat permit `str` so that you can write charm code and corresponding tests to use new features that aren't available in ops yet. But I think that's probably overkill for test code, rather than code that has to deal with values from Juju (Pebble), since we always promptly update Ops (in fact, you're more likely to get it in Ops before you get it in a Juju'd Pebble). And I think we have made some other changes like that recently.

So I'm ok with dropping `str`, although I do think it would be better as its own PR. I assume we couldn't really break anything for anyone by making it *narrower*.

testing/src/scenario/state.py
james-garner-canonical: Good call. I guess the simplest thing is to set them in the class body without type annotations (will be inferred as `int` by type checker, but ignored by `dataclasses`).

testing/src/scenario/state.py
james-garner-canonical: Yeah I think that reasoning applies for when you have a charm running in production with an older `ops` and a newer `pebble`. But we probably don't need the same degree of safety/flexibility in unit tests as you say.

In #2274 we [went with](https://github.com/canonical/operator/pull/2274/changes#diff-b9a3d0885d1495621a0fe09c19d058f7d0233b14bc267cde4fc9cbb8350faf6cR959-R960) converting `str` `level` arguments to `pebble.CheckLevel`, but the backwards compatibility story is different. In #2274 we expanded the argument type to include `str` while leaving the attribute type as `pebble.CheckLevel`. In this case we'd be doing the opposite -- keeping the argument type broad while narrowing the attribute type. Since the attributes are read only, you're right that this is backwards compatible in principle.

Though if we do get an unexpected `str` for the argument, should we crash or let it through with a warning? I'm on board with raising, which *could* break some existing tests, but if so they're probably wrong ...

(I guess narrowing the attribute type could be a separate PR too.)

testing/src/scenario/state.py
tonyandrewmeyer: This was discussed when we added the `deepcopy` method - that method has this comment:

```python
    # We don't have a frozendict to freeze any dictionaries, and while we
    # could freeze a list into a tuple, we would break tests that currently
    # assert that they get a list, so we can't do that until 8.0. That means
    # we can't actually freeze the content here, but we can at least
    # disassociate it from the original object.
```

If the type had always been `Sequence` then it would probably be ok to use `tuple` - a type checker should have been warning about any time it wasn't, although people often aren't type checking their tests, I think.

However, before https://github.com/canonical/operator/commit/34bcab02f70c21aed0ca40d4dac5a97f3e578b3a the type was `list` (etc) - we made it `Sequence` in that change when we added in the deep copy. So there could still be code that's relying on it being a list, I think. (Maybe that wasn't the right choice at the time, but we're stuck with it now).

So I think we are forced to continue with `list`, even though `tuple` would indeed be better as you say.

james-garner-canonical: Ah, fair, guess we're stuck with `list` for backwards compatibility for now. The reasoning is documented in that copy method, but maybe it's worth inlining comments about this where we use `list` now?

testing/src/scenario/state.py
dimaqq: Indeed, it should be OK to pass a value that happens to be None, doesn't need to be a literal.

testing/src/scenario/state.py
```diff
@@ -1586,69 +1721,99 @@ class State:
...
-        for name in [
-            'relations',
-            'containers',
-            'storages',
-            'networks',
-            'opened_ports',
-            'secrets',
-            'resources',
+        object.__setattr__(
+            self, 'relations', frozenset(relations) if relations is not None else frozenset()
```
dimaqq: **minor**:
a part of this PR follows `container(x) is x else container()`.
another part follows `container(x) if x is not None else container()`.

is the distinction deliberate? e.g. because we'd want to drain custom falsey generator objects?

testing/src/scenario/state.py
```diff
@@ -253,59 +285,83 @@ class Secret:
...
 
-    id: str = dataclasses.field(default_factory=_generate_secret_id)
+    id: str
     """The Juju ID of the secret.
 
     This is automatically assigned and should not usually need to be explicitly set.
     """
 
-    owner: Literal['unit', 'app', None] = None
+    owner: Literal['unit', 'app', None]
```
dimaqq: Same wrt. literal
```suggestion
    owner: Literal['unit', 'app'] | None
```

dimaqq (approved):
I kinda like this.
If it works, it's fair to merge :)

Happy to leave the details to James.

testing/src/scenario/state.py
tonyandrewmeyer: Nice catch, thanks!

testing/src/scenario/state.py
tonyandrewmeyer: I'll make it consistent.

testing/tests/test_e2e/test_state.py
tonyandrewmeyer: Yeah, an existing error drive-by fix.

testing/src/scenario/state.py
tonyandrewmeyer: I'm not super keen on adding lots of separate comments. I think having the one prominent one in the module should be enough. Please re-open if you disagree.

testing/src/scenario/state.py
tonyandrewmeyer: Agreed with the putting off :)

testing/src/scenario/state.py
tonyandrewmeyer: I'd prefer to leave, to keep it simpler and so there's that large comment explaining things.

tonyandrewmeyer:
@james-garner-canonical brought up the excellent point that these are frozen dataclasses that we want people to treat as immutable. So giving their type checker information that they have a list (rather than an immutable Sequence) or a dict (rather than an immutable Mapping) leads them to where we don't want to go.

So rejecting this instead.

tonyandrewmeyer:
@james-garner-canonical I've adjusted per the discussion we had earlier in the week, and this should be good for reviewing again now, thanks!

testing/src/scenario/state.py
```diff
@@ -257,62 +285,84 @@ class Secret:
...
+    owner: Literal['unit', 'app'] | None
     """Indicates if the secret is owned by *this* unit, *this* application, or
     another application/unit.
 
     If ``None``, the implication is that read access to the secret has been granted
     to this unit.
     """
 
-    remote_grants: Mapping[int, set[str]] = dataclasses.field(default_factory=dict)
+    remote_grants: Mapping[int, set[str]]
```
james-garner-canonical: I think this would ideally be `frozenset[str]` for immutability, but that would be a backwards incompatible change.
```suggestion
    remote_grants: Mapping[int, set[str]]  # ideally frozenset[str] but set[str] for backwards compatibility
```
And/or add this to the list of things to fix in the next breaking release?

testing/src/scenario/state.py
```diff
@@ -257,62 +285,84 @@ class Secret:
...
+    _latest_revision = 1
+
+    def __init__(
+        self,
+        tracked_content: RawSecretRevisionContents,
+        *,
+        latest_content: RawSecretRevisionContents | None = None,
+        id: str | None = None,
+        owner: Literal['unit', 'app'] | None = None,
+        remote_grants: Mapping[int, set[str]] = {},
```
james-garner-canonical: We shouldn't enforce `set[str]` for the init argument IMO, since we'd really like it to be `frozenset` on the attribute in future.
```suggestion
        remote_grants: Mapping[int, Iterable[str]] = {},
```
This would require us to change how we set the `remote_grants` attribute below, like this:
```py
        object.__setattr__(self, 'remote_grants', {k: set(v) for k, v in remote_grants.items()})
```

testing/src/scenario/state.py
```diff
@@ -257,62 +285,84 @@ class Secret:
...
+            self._validate_content(latest_content, 'latest_content')
+        object.__setattr__(self, 'tracked_content', tracked_content)
+        object.__setattr__(
+            self,
+            'latest_content',
+            latest_content if latest_content is not None else tracked_content,
+        )
+        object.__setattr__(self, 'id', id if id is not None else _generate_secret_id())
+        object.__setattr__(self, 'owner', owner)
+        object.__setattr__(self, 'remote_grants', dict(remote_grants))
```
```suggestion
        # Ideally we'd use frozenset(v) but changing from set would be backwards incompatible.
        object.__setattr__(self, 'remote_grants', {k: set(v) for k, v in remote_grants.items()})
```

testing/src/scenario/state.py
james-garner-canonical: I'm increasingly skeptical of this. It would now only matter for `tracked_content` and `latest_content` if we make the `remote_grants` arg's values `Iterable[str]`. The idea of a one liner that makes everything safe is nice, but I think the abstraction makes this a lot less clear, and it would be much better to have things inlined.

For example, I was surprised to notice just now that `_deepcopy_mutable_fields` only copies `dict` and `list`, not the other built in mutable collection `set` (which I think would ideally be made a `frozenset` in that method, but probably can't be right now in the general case for backwards compatibility, with the same reasoning that it doesn't convert `list` to `tuple`).

My point is, you have to read it to check that we're making things immutable correctly anyway, so it doesn't actually make this simpler for readers.

testing/src/scenario/state.py
james-garner-canonical: I do still think this deserves local comments, either here where we setattr or on the attribute with the type annotation (or both?). What we're saying here is "the type annotations guarantee `Sequence`, but we're actually committed to guaranteeing `list`".

I think the case we're trying to avoid breaking is something like this example, since users shouldn't be making use of any mutable operations.
```py
assert network.bind_addresses == [...]
```
But that's not immediately clear to readers of the code (me specifically), even with the module level comment.

I also think local comments rather than global are very important, as they make it clear where we're committed to this and where we aren't. For example, if we added a totally new field, we could type it as `tuple` without any backwards compatibility concerns.

testing/src/scenario/state.py
```diff
@@ -60,15 +60,15 @@ class _StateKwargs(TypedDict, total=False):
...
+        deferred: Iterable[DeferredEvent]
         stored_states: Iterable[StoredState]
         app_status: _EntityStatus
         unit_status: _EntityStatus
         workload_version: str
 
 
 AnyJson = str | bool | dict[str, 'AnyJson'] | int | float | list['AnyJson']
-RawSecretRevisionContents = RawDataBagContents = dict[str, str]
+RawSecretRevisionContents = RawDataBagContents = Mapping[str, str]
```
james-garner-canonical, line 71: Let's in-line at least `RawSecretRevisionContents`, I'm not convinced that it adds any clarity over `Mapping[str, str]` -- in fact I find the opposite everywhere it's used.

I suppose we have to keep the alias itself for backwards compatibility (unless we make sure it's unused), so this might look like this:
```suggestion
 # unused, but exported for backwards compatibility, will be removed in the next major release
RawSecretRevisionContents = RawDataBagContents = Mapping[str, str]
```
Or however we want to track things like that.

testing/src/scenario/state.py
james-garner-canonical: And we don't actually need deep copies for `tracked_content` and `latest_content` since they're flat `Mapping[str, str]`, so we can just call `dict` in-line. This would be consistent with what we're now doing in `Network.__init__`.

testing/src/scenario/state.py
```diff
@@ -448,28 +498,44 @@ class Network:
...
+    egress_subnets: Sequence[str]
     """Subnets that other units will see the charm connecting from."""
 
+    def __init__(
+        self,
+        binding_name: str,
+        bind_addresses: Iterable[BindAddress] | None = None,
+        *,
+        ingress_addresses: Iterable[str] | None = None,
+        egress_subnets: Iterable[str] | None = None,
```
james-garner-canonical: How about we avoid introducing `None` here, and use immutable values for the defaults?
```suggestion
        bind_addresses: Iterable[BindAddress]  = (BindAddress([Address('192.0.2.0')]),),
        *,
        ingress_addresses: Iterable[str] = ('192.0.2.0',),
        egress_subnets: Iterable[str] = ('192.0.2.0/24',),
```
Then the `object.__setatrr__` calls in `__init__` can be much simpler since they don't need to test for `None` and can just be:
```py
        object.__setattr__(self, 'bind_addresses', list(bind_addresses))
        object.__setattr__(self, 'ingress_addresses', list(ingress_addresses))
        object.__setattr__(self, 'egress_subnets', list(egress_subnets),
```
This also keeps the defaults more clearly exposed for users.

testing/src/scenario/state.py
```diff
@@ -603,7 +669,7 @@ class Relation(RelationBase):
 
     remote_app_data: RawDataBagContents = dataclasses.field(default_factory=dict)
     """The current content of the application databag."""
-    remote_units_data: dict[UnitID, RawDataBagContents] = dataclasses.field(
+    remote_units_data: Mapping[UnitID, RawDataBagContents] = dataclasses.field(
```
james-garner-canonical, line 686: This is a backwards incompatible change since it applies to the attribute as well as the `__init__` argument. We can safely change the `__init__` argument to `Mapping`, since that's broader, but broadening the attribute type is not backwards compatible.

So here we should either:
1. Make no change for now, perhaps leaving a note that this should be `Mapping` in future.
2. Add a custom `__init__` like we're doing for the other classes.
3. Investigate if it's safe to make this technically breaking change, and make it.

testing/src/scenario/state.py
```diff
@@ -688,7 +754,7 @@ def remote_unit_name(self) -> str:
 class PeerRelation(RelationBase):
     """A relation to share data between units of the charm."""
 
-    peers_data: dict[UnitID, RawDataBagContents] = dataclasses.field(default_factory=dict)
+    peers_data: Mapping[UnitID, RawDataBagContents] = dataclasses.field(default_factory=dict)
```
james-garner-canonical, line 771: Same comment as `Relation.remote_units_data`.

testing/src/scenario/state.py
```diff
@@ -786,41 +852,49 @@ def _generate_new_change_id():
     return _CHANGE_IDS
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, init=False)
 class Exec:
     """Mock data for simulated :meth:`ops.Container.exec` calls."""
 
-    command_prefix: Sequence[str]
+    command_prefix: Sequence[str, ...]
```
james-garner-canonical: This isn't valid. `...` has special meaning in some contexts, like `tuple`, where `tuple[str, ...]` means a variable length `tuple` of `str` items, but `pyright` should complain about its use here. Why is our linting passing?

An easy fix, but the more pressing concern is why our linting isn't complaining.

testing/src/scenario/state.py
```diff
@@ -786,41 +852,49 @@ def _generate_new_change_id():
...
     # change ID: used internally to keep track of mocked processes
-    _change_id: int = dataclasses.field(default_factory=_generate_new_change_id)
+    _change_id: int
 
-    def __post_init__(self):
-        # The command prefix can be any sequence type, and a list is tidier to
-        # write when there's only one string. However, this object needs to be
-        # hashable, so can't contain a list. We 'freeze' the sequence to a tuple
-        # to support that.
-        object.__setattr__(self, 'command_prefix', tuple(self.command_prefix))
```
james-garner-canonical, line 823: Here's an individual example that contradicts the global statement that all our `Sequence` types are really `list`. This is the right way to document that, of course, but it makes me think that we should be documenting the specific cases since there are exceptions, and when you make a change or add a new field you'll want to know the details.

testing/src/scenario/state.py
```diff
@@ -786,41 +852,49 @@ def _generate_new_change_id():
     return _CHANGE_IDS
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, init=False)
 class Exec:
```
james-garner-canonical, line 870: I don't think this class needs a custom `__init__` anymore -- I guess it's an artifact of considering making the attributes `tuple`.

Making the attribute `tuple[str, ..]` would be backwards compatible, since `Exec` doesn't support assignment, but I think leaving the attribute as `Sequence[str]` is fine -- there aren't any `tuple` specific methods that we want to expose to users here, right? We use `tuple` for hashability of `Exec`, but we don't care if the user knows that `Exec.command_prefix` specifically is hashable.

Assuming we don't want to change the type to `tuple`, the question is: are we converting all the classes to have a custom `__init__` for consistency and to make it easier to add a new argument that should have (e.g.) attribute of `frozenset` and argument of `Iterable`? Or are we converting only the ones we need to make the typing changes introduced by this PR?

testing/src/scenario/state.py
james-garner-canonical: Let's leave narrowing the attribute type for a separate PR if we do it at all.

testing/src/scenario/state.py
```diff
@@ -856,7 +930,7 @@ def _next_notice_id(*, update: bool = True):
     return str(cur)
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, init=False)
 class Notice:
```
james-garner-canonical, line 952: This class doesn't actually need a custom `__init__` right now. Same question as above:

> are we converting all the classes to have a custom `__init__` for consistency and to make it easier to add a new argument that should have (e.g.) attribute of `frozenset` and argument of `Iterable`? Or are we converting only the ones we need to make the typing changes introduced by this PR?

Adding a custom init is a bit error prone, so it might pay to have it done once and for all. But on the other hand, it's only really error prone for the cases which needed types changed, it's pretty straightforward for cases like this one (though the `_now_utc` default adds a little mistake that would be easy to make but is handled correctly here).

testing/src/scenario/state.py
```diff
@@ -1090,21 +1188,39 @@ class Container:
...
+        object.__setattr__(self, 'name', name)
+        object.__setattr__(self, 'can_connect', can_connect)
+        object.__setattr__(self, '_base_plan', dict(_base_plan))
+        object.__setattr__(self, 'layers', dict(layers))
+        object.__setattr__(self, 'service_statuses', dict(service_statuses))
+        object.__setattr__(self, 'mounts', dict(mounts))
+        object.__setattr__(self, 'execs', frozenset(execs))
+        object.__setattr__(self, 'notices', list(notices))
+        object.__setattr__(self, 'check_infos', frozenset(check_infos))
+        _deepcopy_mutable_fields(self)
```
james-garner-canonical: Here I again think it would be clearer if we had:
```py
        object.__setattr__(self, '_base_plan', copy.deepcopy(_base_plan))
        object.__setattr__(self, 'layers', copy.deepcopy(layers))
```
It just seems odd to have `_deepcopy_mutable_fields` here making additional unnecessary copies of `service_statuses` , `mounts`, and `notices`.

testing/src/scenario/state.py
```diff
@@ -1069,7 +1167,7 @@ class Container:
     be safely modified.
     """
 
-    execs: Iterable[Exec] = frozenset()
+    execs: frozenset[Exec]
```
james-garner-canonical, line 1188: Do we want to guarantee `frozenset` here? We do want people to be able to do `Collection` operations conveniently, `Iterable` is the wrong typing for an attribute that's a collection. But if there aren't any `frozenset` specific operations we expect users to make use of, then maybe we should use `Collection` instead?

I guess there's set operations like union. But are they useful here?

testing/src/scenario/state.py
```diff
@@ -1090,21 +1188,39 @@ class Container:
         )
     """
 
-    notices: Sequence[Notice] = dataclasses.field(default_factory=list)
+    notices: Sequence[Notice]
     """Any Pebble notices that already exist in the container."""
 
-    check_infos: Iterable[CheckInfo] = frozenset()
+    check_infos: frozenset[CheckInfo]
```
james-garner-canonical, line 1212: Ditto, could be `Collection` perhaps.

testing/src/scenario/state.py
```diff
@@ -1090,21 +1188,39 @@ class Container:
...
+        execs: Iterable[Exec] = (),
+        notices: Iterable[Notice] = (),
+        check_infos: Iterable[CheckInfo] = (),
+    ):
+        object.__setattr__(self, 'name', name)
+        object.__setattr__(self, 'can_connect', can_connect)
+        object.__setattr__(self, '_base_plan', dict(_base_plan))
+        object.__setattr__(self, 'layers', dict(layers))
+        object.__setattr__(self, 'service_statuses', dict(service_statuses))
+        object.__setattr__(self, 'mounts', dict(mounts))
```
james-garner-canonical: We need to convert the `Iterable` arguments to something concrete, whether that's `frozenset`, `tuple`, or `list`. But do we need to convert the `Mapping` arguments, since the attribute is also `Mapping`.

Maybe we should just call `copy.deepcopy` instead. Or `copy.copy` for the flat mappings with immutable values for efficiency.

(Now that I think of it, converting to `dict` is *technically* breaking, since the user could previously assume that their exact object type was passed through, though I expect that this affects absolutely nobody.)

testing/src/scenario/state.py
```diff
@@ -1090,21 +1188,39 @@ class Container:
...
+        object.__setattr__(self, 'execs', frozenset(execs))
+        object.__setattr__(self, 'notices', list(notices))
```
james-garner-canonical, line 1244: We're taking our typing queue from the original default factory here. What's not immediately clear to me is whether this is a case like the other `list` ones where we typed as `Sequence` but really want to guarantee `list` (at least for equality comparisons?), so a local comment would be nice IMO.

testing/src/scenario/state.py
```diff
@@ -1090,21 +1188,39 @@ class Container:
...
+        object.__setattr__(self, 'check_infos', frozenset(check_infos))
+        _deepcopy_mutable_fields(self)
+
     def __hash__(self) -> int:
         return hash(self.name)
 
-    def __post_init__(self):
-        if not isinstance(self.execs, frozenset):
-            # Allow passing a regular set (or other iterable) of Execs.
-            object.__setattr__(self, 'execs', frozenset(self.execs))
```
james-garner-canonical, line 1105: Interestingly it looks like we previously didn't convert `check_infos` huh?

testing/src/scenario/state.py
```diff
@@ -1387,18 +1501,29 @@ class MyCharm(ops.CharmBase):
...
+        *,
+        owner_path: str | None = None,
+        content: Mapping[str, Any] = {},
+        _data_type_name: str = 'StoredStateData',
+    ):
+        object.__setattr__(self, 'name', name)
+        object.__setattr__(self, 'owner_path', owner_path)
+        object.__setattr__(self, 'content', dict(content))
+        object.__setattr__(self, '_data_type_name', _data_type_name)
+        _deepcopy_mutable_fields(self)
```
```suggestion
        object.__setattr__(self, 'content', copy.deepcopy(content))
        object.__setattr__(self, '_data_type_name', _data_type_name)
```

testing/src/scenario/state.py
```diff
@@ -1568,13 +1693,11 @@ class State:
     return data from `State.leader`, and so on.
     """
 
-    config: dict[str, str | int | float | bool] = dataclasses.field(
-        default_factory=dict,
-    )
+    config: Mapping[str, str | int | float | bool]
     """The present configuration of this charm."""
-    relations: Iterable[RelationBase] = dataclasses.field(default_factory=frozenset)
+    relations: frozenset[RelationBase]
```
james-garner-canonical, line 1727: Should we use `Collection` instead for all these attributes?

testing/tests/test_e2e/test_state.py
```diff
@@ -464,6 +476,53 @@ def test_immutable_content_dict_of_dicts(
...
+def test_init_converts_to_concrete_type(
+    component: type[object],
+    attribute: str,
+    expected_type: type,
+    input_value: Any,
+    required_args: dict[str, Any],
+):
+    """Verify that __init__ converts broader input types to concrete attribute types."""
+    obj = component(**required_args, **{attribute: input_value})
+    assert isinstance(getattr(obj, attribute), expected_type)
```
james-garner-canonical, line 542: I understand the intent here, but reading it breaks my brain a little.

Minor suggestion: reorder params to `component, required_args, attribute, input_value, expected_type` to follow the logic of the test.

More radical suggestion: have AI unroll this into a series of unparametrised tests like:
```py
def test_cloud_credential_init_converts_args():
    obj = CloudCredential(
        auth_type='foo',  # required
        attributes={'a', 'b'},
        redacted=('a', 'b'),
    )
    assert obj.attributes == ['a']
    assert obj.redacted == ['a', 'b']
```

testing/src/scenario/state.py
james-garner-canonical: Though it would be neat if we could get away with making it `Collection[str]` in typing and `frozenset` at runtime without breaking anyone ...

testing/src/scenario/state.py
james-garner-canonical: Also we may want to retain a note about why this is `tuple` (for hashability).

james-garner-canonical (changes requested):
I really like the direction here, and I definitely think it's worth making these changes to decouple the `__init__` argument typing from the attribute typing.

I've flagged a number of items that I think require some further thought before merging.

testing/src/scenario/state.py
tonyandrewmeyer:
```suggestion
    remote_grants: Mapping[int, frozenset[str]]
```

testing/src/scenario/state.py
tonyandrewmeyer:
```suggestion
        object.__setattr__(self, 'remote_grants', {k: frozenset(v) for k, v in remote_grants.items()})
```

testing/src/scenario/state.py
tonyandrewmeyer:
```suggestion
    command_prefix: Sequence[str]
```

```suggestion
    command_prefix: Sequence[str]
```

dimaqq:
I'll wait for Tony and James to hash it out.

testing/src/scenario/state.py
```diff
@@ -49,7 +49,7 @@
     from . import Context
 
     class _StateKwargs(TypedDict, total=False):
-        config: dict[str, str | int | float | bool]
+        config: Mapping[str, str | int | float | bool]
```
dimaqq, line 52: This is pretty clear and a good change, IMO.

testing/src/scenario/state.py
dimaqq, line 71: This one I'm not convinced about, because it's an "in-out" field:

```py
# user passes data in, Mapping is great here
rel = Relation(..., remote_app_data=this_is_now_a_mapping)
state = State(relations={rel})
```

```py
# user inspects of modified data, dict is better here
state = ctx.run(..., state)
rel = state.get_relation(rel)

# what we expect charmers to do, Mapping is helpful
assert rel.remote_units_data[unit]["foo"] == '"bar"'

# what charmers may also do, I think thaththis PR is a breaking change wrt. typing
rel.remote_units_data[unit].setdefault("foo", '"bar"')
```

The intention of having this field typed as a Mapping is good, IMO.
The practical implications though, I'm not sure about.

dimaqq (commented):
WDYT about settings up a working group (maybe James and Tony only) to figure out the minutae?

testing/src/scenario/state.py
tonyandrewmeyer: I've removed `_deepcopy_mustable_fields`.

testing/src/scenario/state.py
tonyandrewmeyer: I agree that it's odd to have this one specific type. We do need to keep it for backwards compatibility, though, so I don't see much value in changing it now. Let's chat about this outside the issue if you disagree.

testing/src/scenario/state.py
tonyandrewmeyer: Done.

testing/src/scenario/state.py
tonyandrewmeyer: Fixed. I don't particularly want to spend time investigating why the linter didn't find it at this time. Please open an issue for someone to do that if you're really keen on it.

testing/src/scenario/state.py
tonyandrewmeyer: Added a note.

testing/tests/test_e2e/test_state.py
tonyandrewmeyer: I've re-ordered. I don't really want to unroll, and I think it would definitely be better in a separate PR (it will be a lot of extra lines and a lot of extra checking to make sure nothing was missed) if it is done.

testing/src/scenario/state.py
tonyandrewmeyer: I feel that this falls into the backwards incompatible changes we have made previously and been ok with, in that we believe everyone should always have been treating these as immutable, so it's a bug if they were not (and this won't break people's tests, only the type checking of their tests).

I'll do a super-tox run to check all the changes in this PR when everything's otherwise settled, although I'm not sure that enough charms are type-checking their tests in `lint` for that to tell us a lot.

Let's chat about this outside the issue if you disagree.

testing/src/scenario/state.py
tonyandrewmeyer: I agree that leaving the attribute typed as `Sequence` is fine, so the init is really just doing the `tuple()`, which could indeed be a `post_init` like before. I don't think we need to convert everything to use a custom `__init__` but since there's something to do be done here, I'm fine leaving it that way.

Let me know if you disagree; I don't feel strongly about it.

testing/src/scenario/state.py
tonyandrewmeyer: Same with this one, except I feel it's more useful here (slightly).

testing/src/scenario/state.py
tonyandrewmeyer: I lean towards leaving it as `frozenset` in case there are uses that I'm not seeing, although I don't have any concrete ones. If you think `Collection` would be much better, I'm fine moving to that (same below).

testing/src/scenario/state.py
tonyandrewmeyer: Added a comment.

dimaqq (approved):

testing/src/scenario/state.py
```diff
@@ -183,44 +178,64 @@ def _to_ops(self) -> CloudCredential_Ops:
...
+        self,
+        type: str,
+        *,
+        name: str = 'localhost',
+        region: str | None = None,
+        endpoint: str | None = None,
+        identity_endpoint: str | None = None,
+        storage_endpoint: str | None = None,
+        credential: CloudCredential | None = None,
+        ca_certificates: Iterable[str] = (),
```
james-garner-canonical, line 240: This applied before the PR when the argument was typed as `Sequence[str]` -- `ca_credentials='some string'` is legal at type checking time.

Previously, `_deepcopy_mutable_variables` would have left us with `ca_certificates = 'some string'`. Now we'd get `ca_certificates = ['s', 'o', 'm', ...]`.

I assume that both of these are bad and may result in an error at some later point, but maybe they can silently pass through a test in some cases? Presumably this would ideally be an error straight away. Alternatively, a legal shorthand for `ca_certificates = ['some string']`.

james-garner-canonical (approved):
Looks good, apologies for the delays with this one. Thanks for all the work and super-tox-ing.

---

I've commented in one place about `Iterable/Sequence[str]` typing permitting bare `str` arguments. This is largely a pre-existing issue, except that inlining `_deepcopy_mutable_fields` leads to a behaviour change where a bare `str` now gets `list` called on it. Either way it would have been a technical error that *should* always have resulted in a runtime error at some point (though would it always have?). But if it's not a runtime error then we are introducing a behaviour change that could show up in state.

I'd be on board with adding an explicit check and error for bare `str` if that doesn't break existing tests.

testing/src/scenario/state.py
tonyandrewmeyer: Good catch. Added a `_list_of_str` helper that raises `StateValidationError` when a bare `str` is passed where an iterable of strings is expected, and applied it to `redacted`, `ca_certificates`, `ingress_addresses`, `egress_subnets`, and the values in `Secret.remote_grants`. Existing tests still pass and there's a new `test_bare_str_rejected` parametrised test covering each site.

tonyandrewmeyer:
This has two ticks, but one is Dima's so it doesn't pass any more. @tromai would you mind having a look?

james-garner-canonical:
New concern that occurred to me today: adding a custom `__init__` (which doesn't call `__post_init__`) breaks anyone who subclassed one of these classes and defined their own `__post_init__` (it would no longer be called).

I guess we should (re-)check with Hyrum? I suppose checking for new failures in unit tests would be sufficient, but it also feels like something it should be possible to analyse statically ... perhaps a feature for another time.

benhoyt:
We can definitely re-check with Hyrum. However, I guess we have to step back and decide what we're promising when exposing dataclasses as part of the public API (and document that). It seems to me people shouldn't need to subclass testing state classes, and maybe we should rule that out.

tromai (approved):
Thank you for waiting. The current changes look good to me.

It took me sometime to understand that dataclasses attribute type definition can be different from the types in `__init__` signature. These classes need additional logic in `__init__` to "pre-process" the parameters.

The changes make sense to me after I wrap my head around it.

tonyandrewmeyer:
Baseline pass rate: **73%** (78/107). Current pass rate: **71%** (76/107). 2 new failure(s), 0 resolved, 0 new error(s).

| Charm | Baseline | Current |
| --- | --- | --- |
| charm-ubuntu | failed: 3 errors; PendingDeprecationWarning: Harness is deprecated. For the recommended approach, see: https://documentation.ubuntu.com/ops/latest/howto/write-unit-tes… | same |
| charmed-etcd-operator | failed: 13 failed, 74 passed; InconsistentScenarioError x8 | same |
| exim-operator | failed: Exception: Can not find valid pkg-config name. | same |
| indico-operator | failed: uv resolve: no solution | same |
| kafka-k8s-operator | failed: TypeError: Link.__init__() got an unexpected keyword argument 'size' | same |
| kafka-operator | failed: collection error in tests/unit/test_tls_manager.py: subprocess.CalledProcessError: Command 'keytool' returned non-zero exit status 127. | same |
| ks-charmed | failed: FutureWarning: The 'wheel' package is no longer the canonical location | same |
| kubeflow-roles-operator | failed: 1 error; ModuleNotFoundError: No module named 'ops' | same |
| kyuubi-k8s-operator | failed: psycopg2 build: pg_config not found | same |
| manual-tls-certificates-operator | passed | failed: 28 failed, 4 passed; UncaughtCharmError x20 (+1 other) |
| mattermost-k8s-operator | failed: uv resolve: no solution | same |
| netbox-k8s-operator | failed: uv resolve: URL dep must be a direct requirement | same |
| opensearch-operator | failed: pytest: no tests ran | same |
| parca-agent-operator | failed: 5 failed, 17 passed; UncaughtCharmError x5 | same |
| parca-k8s-operator | passed | failed: 64 failed, 68 passed; TypeError x64 |
| pgbouncer-k8s-operator | failed: psycopg2 build: pg_config not found | same |
| pgbouncer-operator | failed: psycopg2 build: pg_config not found | same |
| postgresql-k8s-operator | failed: psycopg2 build: pg_config not found | same |
| postgresql-operator | failed: psycopg2 build: pg_config not found | same |
| powerflex-cinder-charm | failed: C extension build: Python.h not found | same |
| powerflex-nova-charm | failed: C extension build: Python.h not found | same |
| redis-operator | failed: 1 failed, 3 passed; ConnectionError x1 | same |
| self-signed-certificates-operator | failed: 5 errors; ModuleNotFoundError: No module named 'charms' | same |
| snappass-test | failed: 6 failed; TypeError x4 (+2 other) | same |
| synapse-operator | failed: uv resolve: URL dep must be a direct requirement | same |
| temporal-k8s-operator | failed: 13 failed, 40 passed; InconsistentScenarioError x13 | same |
| temporal-ui-k8s-operator | failed: 9 failed, 4 passed; InconsistentScenarioError x9 | same |
| timescaledb-charm | failed: 3 failed, 3 passed; InvalidSpecError x3 | same |
| traefik-k8s-operator | failed: uv resolve: no solution | same |
| zookeeper-k8s-operator | failed: 8 failed, 151 passed; InconsistentScenarioError x6 (+2 other) | same |
| zookeeper-operator | failed: 67 failed, 93 passed; UncaughtCharmError x58 (+9 other) | failed: 59 failed, 101 passed; UncaughtCharmError x48 (+11 other) |

Bit suspicious it's zookeeper again. I'll look into that.

PR #2334, opened 2026-02-19
docs: add cli args as another place to not put sensitive data

tonyandrewmeyer:
Command line arguments can easily end up in traces, logs, and/or exceptions, so call them out explicitly in the security doc guidance as another place where sensitive data should not be found.

(Based on a suggestion from @benhoyt in MM).


benhoyt (approved):
Thanks!

PR #2335, opened 2026-02-19
fix: use immutable types for state component attributes to reflect that they are immutable

tonyandrewmeyer:
We expect all of the state attributes to be treated as immutable, but some are typed as `dict`. Clean that up so that although the attribute may be a `dict`, it is typed as `Mapping`, so that a type checker will flag people wrongly mutating the state.

Also adds a custom `__init__` for `Secret`, which cleans up the private attributes (tracking revisions) and makes sure that the `remote_grants` mapping has frozenset items rather than set.

### DIFF
--- a/testing/tests/test_e2e/test_state.py
+++ b/testing/tests/test_e2e/test_state.py
@@ -371,7 +371,6 @@ def test_replace_state():
         (CloudCredential, 'attributes', {'auth_type': 'foo'}),
         (Secret, 'tracked_content', {}),
         (Secret, 'latest_content', {'tracked_content': {'password': 'password'}}),
-        (Secret, 'remote_grants', {'tracked_content': {'password': 'password'}}),
         (Relation, 'local_app_data', {'endpoint': 'foo'}),
         (Relation, 'local_unit_data', {'endpoint': 'foo'}),
         (Relation, 'remote_app_data', {'endpoint': 'foo'}),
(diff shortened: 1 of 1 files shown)


testing/src/scenario/state.py
```diff
@@ -253,59 +253,82 @@ class Secret:
...
+        object.__setattr__(self, 'id', id if id is not None else _generate_secret_id())
+        object.__setattr__(self, 'owner', owner)
+        remote_grants = {k: frozenset(v) for k, v in remote_grants.items()}
+        object.__setattr__(self, 'remote_grants', remote_grants)
+        object.__setattr__(self, 'label', label)
+        object.__setattr__(self, 'description', description)
+        object.__setattr__(self, 'expire', expire)
+        object.__setattr__(self, 'rotate', rotate)
+        object.__setattr__(self, '_tracked_revision', 1)
+        object.__setattr__(self, '_latest_revision', 1)
```
james-garner-canonical: These lines aren't necessary since we set the attributes directly in the class direclaration.

testing/src/scenario/state.py
```diff
@@ -253,59 +253,82 @@ class Secret:
...
+        _deepcopy_mutable_fields(self)
```
james-garner-canonical, line 326: This should be unnecessary (and is inefficient) since we copy `remote_grants` manually above.

testing/src/scenario/state.py
```diff
@@ -253,59 +253,82 @@ class Secret:
...
+        object.__setattr__(self, 'tracked_content', tracked_content)
+        object.__setattr__(
+            self,
+            'latest_content',
+            latest_content if latest_content is not None else tracked_content,
+        )
+        object.__setattr__(self, 'id', id if id is not None else _generate_secret_id())
+        object.__setattr__(self, 'owner', owner)
+        remote_grants = {k: frozenset(v) for k, v in remote_grants.items()}
+        object.__setattr__(self, 'remote_grants', remote_grants)
```
james-garner-canonical: I normally prefer the style that you have here, but I think we should stick to this pattern for consistency and clearer typing of the `remote_grants` variable.
```suggestion
        object.__setattr__(
            self, 'remote_grants', {k: frozenset(v) for k, v in remote_grants.items()}
        )
```

james-garner-canonical (commented):
Changing from `dict` -> `Mapping` in the type annotations, and from `set` -> `frozenset` both in annotation and in actual code are technically not backwards compatible, though it _should_ only be a problem if anyone is mutating their 'frozen' state. I'm in favour of making this change, but is there any due diligence we should do first?

benhoyt:
> I'm in favour of making this change, but is there any due diligence we should do first?

Can we run type checking (and unit tests) on the supertox charms?

tonyandrewmeyer:
> Can we run type checking (and unit tests) on the supertox charms?

Yeah, I've done that before for other things. We also have the published charms list that get checked, although it still needs to love to handle the dropping of Python 3.8.

The trickiest part is having to do many variants, because it might be "lint" or "static" or variants of those.

tonyandrewmeyer:
Discussing with James whether to continue with this branch/PR or to revive the previous one. Maybe we want to do more than just the changes in here.

testing/src/scenario/state.py
tonyandrewmeyer: But not the secret contents.

tonyandrewmeyer:
Closing in favour of #2331.

PR #2337, opened 2026-02-20
docs: clarify that secret labels are not names

### LINKED ISSUE #2284: docs: How to manage secrets > Label creates confusion between name and label
https://documentation.ubuntu.com/ops/latest/howto/manage-secrets/#when-to-use-labels says

> A label is basically the secret’s name (local to the charm),

This is incorrect/misleading: Names are created by users for themselves (which is why only user secrets have a name); in contrast, labels are created by charms for their own use, and will also be different for an owner/manager vs. a consumer. To give some examples that might help: 
- name vs. label: I as a user might create a secret and call it "my-api-key". I might configure a charm to use it. The charm might give it a label "vault-api-token". 
- label vs. label: Suppose you are a leader unit creating an application secret for peers units to consume -- you'd assign a label as an owner, and then when consuming the secret you'd assign another label as a consumer (because the first time you consume a secret you give it a label).

I know the "basically" and "(local to the charm)" are probably meant to help, but  [users are still confused](https://github.com/juju/juju/issues/21591#issuecomment-3783140135).

I suggest rephrasing this to something clearer, e.g.,

> A label is a secret tag created locally by charms (whether owners or consumers).

(Note: The "owners" bit seems to be missing from the how-to.)

I will create an issue to address this in Juju docs too, so this paragraph in Ops can just link there for further context, if necessary. >> https://github.com/juju/juju/issues/21633

dimaqq:
Don't confuse secret labels with names.
Fixes #2284 

I went for smallest possible change, not going into "charm" vs "deployed app" distinction or whether charmers "should" or are "recommended to" label secrets or what to do with unbounded number of secrets, etc.

### DIFF
--- a/docs/howto/manage-secrets.md
+++ b/docs/howto/manage-secrets.md
@@ -253,7 +253,7 @@ Note that:
 ### Label the secrets you're observing
 
 Sometimes a charm will observe multiple secrets. In the `secret-changed` event handler above, you might ask yourself: How do I know which secret has changed?
-The answer lies with **secret labels**: a label is a charm-local name that you can assign to a secret. Let's go through the following code:
+The answer lies with **secret labels**: you can assign a charm-local label to a secret. Let's go through the following code:
 
 ```python
 class MyWebserverCharm(ops.CharmBase):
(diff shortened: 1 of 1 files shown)


dimaqq:
cc @tmihoc

tonyandrewmeyer (approved):
Works for me, thanks!

PR #2338, opened 2026-02-20
chore: update tox-uv advice to include explicit Python version

dimaqq:
Our current advice is to install tox with tox-uv as follows:
```
uv tool install tox --with tox-uv
```

I've tried this today, and the result was `tox` that could not run unit tests.

The cause was that the tool was installed with Python 3.8.
(which I think pulls in ancient version of tox-uv).

Setting Python version explicitly works:
```
uv tool install tox --python 3.12 --with tox-uv
```

I'm a little unsure how far we should go with this advice... is this a problem only for me? my machine? any charmer? only charmers that have Py 3.8 available?

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -14,7 +14,7 @@ sudo snap install astral-uv --classic
 Then install `tox` with extensions, as well as a range of Python versions:
 
 ```sh
-uv tool install tox --with tox-uv
+uv tool install tox --python 3.12 --with tox-uv
 uv tool update-shell
 ```
 
--- a/docs/explanation/testing.md
+++ b/docs/explanation/testing.md
@@ -148,7 +148,7 @@ on:
       - name: Set up uv
         uses: astral-sh/setup-uv@7
       - name: Set up tox and tox-uv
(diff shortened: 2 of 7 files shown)


dimaqq:
If either you has an idea why this happens, do tell!

dimaqq:
I was able to recover my system by forcibly removing py 3.8, so now it's not available, and thus doesn't get picked up.

I was able to narrow down the offending config to:
- I have many Pythons installed system-wide (from python.org)
- I don't have modern pythons installed via `uv python install`
- I have 3.8 installed using `uv python install 3.8`

In this configuration, `uv python list` shows 10 pythons to choose from (across 46 lines of output), yet chooses crappiest for the tool. I'll open a ticket against `uv` for help. --> https://github.com/astral-sh/uv/issues/18111

dimaqq:
It's unfortunate, but I don't think we can "protect" charmers from particular usage of popular tools like `uv`.

dimaqq:
I'm still talking to `uv` folks, but I gather we can:
* either leave it alone
* or add `uv python install 3.12` to our instructions

PR #2339, opened 2026-02-23
ci: check httpbin-demo example charm with mypy

james-garner-canonical:
This PR updates the `httpbin-demo` example charm to add `mypy` to its existing linting. I've verified that this linting fails with the reported error without the update to `ops/testing.py` made in #2343.

To keep this check current, I've updated `tool.uv.sources` for the charm to pull `ops` from the local files -- but this introduces a gotcha, since we'd need to relock every time we change the `ops` version. I've updated the release script to do this, but I wonder if we'd be better off adding a separate CI step instead.

---

We could add `mypy` to the linting for our other example charms, the K8s and machine tutorial charms. However, this wouldn't reflect what we actually want charmers to do when following the tutorial, so we shouldn't do so on disc. Instead, it would probably be cleaner to add a custom CI job like below. If we did so, we could use the same approach for the `httpbin-demo` charm. #2360 implements this approach.

### DIFF
--- a/release.py
+++ b/release.py
@@ -54,6 +54,9 @@ VERSION_FILES = {
     'uvlock': pathlib.Path('uv.lock'),
     'versions_doc': pathlib.Path('docs/explanation/versions.md'),
 }
+COUPLED_EXAMPLES = (  # our example charms that use tool.uv.sources to depend on local ops
+    pathlib.Path('examples/httpbin-demo'),
+)
 CHANGE_LINE_REGEX = (
     r'^\* (?P<category>\w+)(?P<breaking>!?): (?P<summary>.*) by [^ ]+ in (?P<pr>.*)'
 )
@@ -432,9 +435,11 @@ def update_versions_doc(version: str):
     )
 
 
-def update_uv_lock():
+def update_uv_locks():
     """Update the uv.lock file with the new versions."""
     subprocess.run(['uv', 'lock'], check=True)  # noqa: S607
+    for example in COUPLED_EXAMPLES:
+        subprocess.run(['uv', 'lock'], cwd=example, check=True)  # noqa: S607
 
 
 def get_scenario_version() -> packaging.version.Version:
@@ -467,7 +472,7 @@ def update_versions_for_release(tag: str):
     update_testing_version(tag, scenario_version)
     update_tracing_version(tag)
     update_versions_doc(tag)
-    update_uv_lock()
+    update_uv_locks()
 
 
 def get_new_version_post_release(branch_name: str) -> str:
@@ -506,7 +511,7 @@ def update_versions_for_post_release(branch_name: str):
     update_ops_version(ops_version, scenario_version)
     update_testing_version(ops_version, scenario_version)
     update_tracing_version(ops_version)
-    update_uv_lock()
+    update_uv_locks()
 
 
 def check_update_charm_pins_prs(repo: github.Repository.Repository):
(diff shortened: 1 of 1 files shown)


benhoyt:
Excellent, thanks for this. Let's discuss in daily when Tony and Dima are both back (I've added it to the top of the doc).

dimaqq:
my 2c: let's bite the bullet: merge, call it out in release notes, but don't bump the major version.

ops/testing.py
```diff
@@ -169,7 +169,6 @@
 _compatibility_names = [
     'CharmBase',
     'CharmMeta',
-    'Container',  # If Scenario has been installed, then this will be scenario.Container.
```
dimaqq, line 172: Do we have containers in Harness, at all?

tonyandrewmeyer: Harness exported the name `Container` (the object `ops.model.Container`) for a long time, and that was kept when we added Scenario to the `ops.testing` namespace to avoid breaking test code that (lazily / incorrectly) would use `Container` imported from there, like `from ops.testing import Container, Harness, ...`.

tonyandrewmeyer (changes requested):
I don't like that this makes CI changes in the same PR (and that's not mentioned at all in the PR summary even though it's in the description). If we want to do that (and I think it's an entirely separate question) then I think it belongs in a dedicated PR.

james-garner-canonical:
> I don't like that this makes CI changes in the same PR (and that's not mentioned at all in the PR summary even though it's in the description). If we want to do that (and I think it's an entirely separate question) then I think it belongs in a dedicated PR.

The CI changes are essentially tests that exercise fix in the PR and would fail on `main` without the fix.

EDIT: Updated PR summary to mention this.

tonyandrewmeyer:
> > I don't like that this makes CI changes in the same PR (and that's not mentioned at all in the PR summary even though it's in the description). If we want to do that (and I think it's an entirely separate question) then I think it belongs in a dedicated PR.
> 
> The CI changes are essentially tests that exercise fix in the PR and would fail on `main` without the fix.

My concern here is that I think the CI change (basically that we are committing to some increased level of support for mypy) is the biggest change here, and adjusting `ops.testing` is the small side change being done to support that - but that's not how the PR summary or description is written.

If we are going to explicitly support `mypy` in this way, then I think that merits a dedicated discussion, and a dedicated PR. It can fail at first, or we can make this change first (without needing automated tests), or (my least favourite) with failing tests that we expect to resolve in a follow-up quickly afterwards.

I don't think it's the right approach to have the more significant change as a drive-by.

james-garner-canonical:
> > > I don't like that this makes CI changes in the same PR (and that's not mentioned at all in the PR summary even though it's in the description). If we want to do that (and I think it's an entirely separate question) then I think it belongs in a dedicated PR.
> > 
> > 
> > The CI changes are essentially tests that exercise fix in the PR and would fail on `main` without the fix.
> 
> My concern here is that I think the CI change (basically that we are committing to some increased level of support for mypy) is the biggest change here, and adjusting `ops.testing` is the small side change being done to support that - but that's not how the PR summary or description is written.
> 
> If we are going to explicitly support `mypy` in this way, then I think that merits a dedicated discussion, and a dedicated PR. It can fail at first, or we can make this change first (without needing automated tests), or (my least favourite) with failing tests that we expect to resolve in a follow-up quickly afterwards.
> 
> I don't think it's the right approach to have the more significant change as a drive-by.

Happy to reframe the PR that way. I guess if we opt to merge #2343 (with manual verification) this PR can become CI changes only.

benhoyt:
Per discussion, change this PR to just "add mypy to the example charms".

james-garner-canonical:
Closing in favour of:
- #2360

PR #2343, opened 2026-02-25
fix: Move the testing.Container compatibility import so that mypy style checkers understand it

tonyandrewmeyer:
Test code, based on the code in the Matrix thread, but fixed:

```python
from typing import Iterable

import ops
import ops.testing
import pytest

execs: Iterable[ops.testing.Exec] = set()


@pytest.fixture()
def mediawiki_container():
    return ops.testing.Container(
        name="mediawiki",
        can_connect=True,
        execs=execs,
    )


class TestPebbleReadyEvent:
    def test_can_connect(self, ctx: ops.testing.Context[ops.CharmBase], mediawiki_container: ops.testing.Container) -> None:
        state_in = ops.testing.State(containers=[mediawiki_container], leader=True)
        ctx.run(ctx.on.update_status(), state_in)
```

With main:

```
tameyer@tam-canoncial-1:~/w/operator$ uvx --with=pytest --with=ops[testing] mypy --follow-imports=silent /tmp/mypy_example.py                                  
/tmp/mypy_example.py:21: error: List item 0 has incompatible type "ops.model.Container"; expected "scenario.state.Container"  [list-item]
Found 1 error in 1 file (checked 1 source file)
tameyer@tam-canoncial-1:~/w/operator$ uvx --with=pytest --with=ops[testing] pyright /tmp/mypy_example.py 
```

With this branch:

```
tameyer@tam-canoncial-1:~/w/fix-mypy-testing-container$ uvx --with=pytest --with=ops[testing] mypy --follow-imports=silent /tmp/mypy_example.py 
Success: no issues found in 1 source file
tameyer@tam-canoncial-1:~/w/fix-mypy-testing-container$ uvx --with=pytest --with=ops[testing] pyright /tmp/mypy_example.py 
```

By moving the import from `ops.model` to the `except` block, type checkers like mypy understand that it's one or the other. We also need a `type: ignore` there because mypy otherwise complains about incompatible imports.

This is an alternative to #2339.

### DIFF
--- a/ops/testing.py
+++ b/ops/testing.py
@@ -118,7 +118,6 @@ try:
         layer_from_rockcraft,
     )
 except ImportError:
-    # mypy requires the type: ignore, pyright complains about the type: ignore
     from .model import Container
 else:
     # The Scenario unit testing framework.


james-garner-canonical:
I think an approach like this is a fine alternative if we really do want to keep preserving these compatibility names. At the very least it lets us drop them in a more principled way later.

We should still merge the CI changes from #2339. Happy for them to be rolled into this PR or to update #2339 if we prefer an approach like this. (They can't be merged separately first since they'd fail on `main`. They shouldn't be merged separately afterwards as they're good verification for whatever approach we go with.)

ops/testing.py
```diff
@@ -118,6 +119,7 @@
         layer_from_rockcraft,
     )
 except ImportError:
+    # mypy requires the type: ignore, pyright complains about the type: ignore
```
james-garner-canonical: The `ops` codebase doesn't need to comply with `mypy`, so can we just drop the `type: ignore` for `pyright`?

I know `mypy` will follow imports and complain about them by default, but in your example you use `--follow-imports=silent`, so this shouldn't be needed.

ops/testing.py
```diff
@@ -118,7 +119,8 @@
         layer_from_rockcraft,
     )
 except ImportError:
-    pass
+    # mypy requires the type: ignore, pyright complains about the type: ignore
+    from .model import Container  # type: ignore[assignment]
```
dimaqq: The way it is now, it's unclear why Container is being re-exported here in the except clause.
Let's add an addition comment smth like "for backwards compat, because we used to re-export since ops v1.2.3, blah blah" whatever the concern is.

In other words, consider someone trying to clean this up years from now.

tonyandrewmeyer: There's [already a comment explaining it](https://github.com/canonical/operator/blob/28fc366edccac45eb5b01aa7a2c49fcb6a7ed323/ops/testing.py#L168), it's just that it already exists so does not show up in this PR.

ops/testing.py
tonyandrewmeyer: I wasn't originally using `follow-imports=silent` (I added it because there was so much noise from other issues with `harness.py` and elsewhere). I can check whether this is still needed or not, if we decide to go with this approach.

tonyandrewmeyer:
> I think an approach like this is a fine alternative if we really do want to keep preserving these compatibility names.

I don't *want* to keep them, but so far we have done so across multiple changes, not just the inclusion of Scenario, and if we don't have a need to change that policy, I don't see reasoning for why we would change now.

> At the very least it lets us drop them in a more principled way later.

That would be my preference. At the next sprint, I would like us to discuss using the next major version for more cleanup than just dropping a Python version, including things like this.

ops/testing.py
dimaqq: Right... I didn't see that. I guess that's enough.

dimaqq (approved):
Code-wise LGTM, let's decide which way we go at the next stand-up.

benhoyt:
Let's do this, thanks. It would be interesting to see what "revealtype" says about `ops.testing.Container` for mypy and pyright.

ops/testing.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# pyright: reportUnnecessaryTypeIgnoreComment=false
```
tonyandrewmeyer:
```suggestion
```

ops/testing.py
```diff
@@ -118,7 +118,8 @@
         layer_from_rockcraft,
     )
 except ImportError:
-    pass
+    # mypy requires the type: ignore, pyright complains about the type: ignore
+    from .model import Container  # type: ignore[assignment]
```
tonyandrewmeyer:
```suggestion
    from .model import Container
```

tonyandrewmeyer:
> It would be interesting to see what "revealtype" says about `ops.testing.Container` for mypy and pyright.

In main, with `mypy`, `reveal_type(ops.testing.Container)` is `ops.model.Container` both for plain `ops` and also `ops[testing]`. In this branch, it changes to `scenario.state.Container` for `ops[testing]`.

For `pyright`, it's annoyingly `"Container"` in all cases, which is not particularly informative. I'm not sure if there's a way to have it output more information. We do know that it works for `pyright`, so I don't think it's worth spending a lot of time on.

ops/testing.py
james-garner-canonical: Can we drop or update this comment? I'm happy with dropping it. If you want it to stay, how about:

```suggestion
    # mypy would require a type ignore since Container is already defined, but pyright doesn't
```

james-garner-canonical (approved):
> For `pyright`, it's annoyingly `"Container"` in all cases, which is not particularly informative. I'm not sure if there's a way to have it output more information. We do know that it works for `pyright`, so I don't think it's worth spending a lot of time on.

I did a quick manual test that `pyright` via Pylance in VSCode understands the updated form, and it behaves as expected -- `scenario.state.Container` if running in a venv with `ops[testing]`, and `ops.model.Container` otherwise.

ops/testing.py
tonyandrewmeyer: We can drop it - I missed it when I was removing the rest of the changes around this, thanks!

tonyandrewmeyer:
```suggestion
```

PR #2348, opened 2026-02-26
fix: correct type annotation for StorageMeta.properties

yangdanny97:
I'm pretty sure this is supposed to be a type annotation, not initializing the attribute to `list[str]`

It's assigned with a list in the constructor on L2103

### DIFF
--- a/ops/charm.py
+++ b/ops/charm.py
@@ -2079,7 +2079,7 @@ class StorageMeta:
     multiple_range: tuple[int, int | None] | None
     """Range of numeric qualifiers when multiple storage units are used."""
 
-    properties = list[str]
+    properties: list[str]
     """List of additional characteristics of the storage."""
 
     def __init__(self, name: str, raw: _StorageMetaDict):


benhoyt:
This will fix the [StorageMeta.properties reference docs](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.StorageMeta.properties) too.

yangdanny97:
@benhoyt I'm a maintainer of [Pyrefly](https://pyrefly.org/), a Python type checker.

As part of our CI we run [mypy primer](https://github.com/hauntsaninja/mypy_primer), which runs Pyrefly on a corpus of open-source Python projects and reports the delta in type errors.

I stumbled onto this when I was spot-checking the delta for https://github.com/facebook/pyrefly/pull/2356

asukaminato0721:
> switching to

That pr is a test to see real world result, and maybe provide a configure to change in different mode?

Glad it can find some hidden bugs :)

james-garner-canonical:
> > switching to
> 
> That pr is a test to see real world result, and maybe provide a configure to change in different mode?
> 
> Glad it can find some hidden bugs :)

Thanks @asukaminato0721. There are definitely interesting tradeoffs here between developer ergonomics and subtle/hidden bugs. I look forward to seeing where `pyrefly` goes with this.

PR #2352, opened 2026-02-27
chore: update pre-commit to use the tools locked in the project

### LINKED ISSUE #1892: Update pre-commit to monorepo
We've switched to monorepo, but pre-commit got left behind.

Today it looks like:

```yaml
    - repo: https://github.com/astral-sh/ruff-pre-commit
      rev: v0.11.2
      hooks:
        - id: ruff
E         args: [ --preview ] # E: too many spaces inside brackets
        - id: ruff-format
E         args: [ --preview ] # E: too many spaces inside brackets
    # Spellcheck the code.
    - repo: https://github.com/codespell-project/codespell
      rev: v2.4.1
      hooks:
E     - id: codespell # E: wrong indentation: expected 6 but found 4
        additional_dependencies:
          - tomli
    # Check workflow security.
```

And it could be something like:

```yaml
  - repo: local
    hooks:
      - id: ruff
        name: ruff (uv)
        entry: uv run -- ruff check --preview
        language: system
        types: [python]

      - id: pyright
        name: pyright (uv)
        entry: uv run -- pyright
        language: system
        types: [python]
```

dimaqq:
Fixes #1892 

This changes pre-commit config to use the tools that are specified in pyproject.toml with versions locked in `uv.lock`.

Drive-by `pre-commit run -a` which fixes trailing whitespace in docs.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -18,14 +18,8 @@ uv tool install tox --with tox-uv
 uv tool update-shell
 ```
 
-To run checks automatically before each commit, install pre-commit and the lint
-tooling:
-
-```sh
-uv tool install pre-commit
-uv sync --group lint
-pre-commit install
-```
+Optionally, to run checks automatically before each commit, install
+[pre-commit](https://pre-commit.com/#install) and run `pre-commit install`.
 
 You can validate that you have a working installation by running:
 


james-garner-canonical (commented):
I don't use `pre-commit`, so I'm not the expert here. But IIRC there was some desire previously to use the tool-defined `pre-commit` actions? IMO the goal of this PR would be best addressed by having `pre-commit` run `tox -e format` and `tox -e lint` rather than copying the commands manually. But if there's some reason that wouldn't work, I do think that having them at least match our `tox` commands right now would be nice.

tonyandrewmeyer:
> But IIRC there was some desire previously to use the tool-defined `pre-commit` actions?

We definitely want to have `pre-commit` use the same *config* as running the command with `tox`, otherwise we get inconsistencies between the two.

Ideally, `pre-commit` and `tox` run the same *version* of the tools - it's less likely that you get inconsistencies, but it does make it possible.

One way of achieving this is to have `pre-commit` just run `tox -e blah`. You lose a bit of what `pre-commit` provides (like passing in only the relevant files) and you're using a tool runner to run a tool runner so it's not the most efficient, but it does give the most consistency. But just keeping the version in sync and ensuring we use tools that get the config from `pyproject.toml` also works (we could have an infra test if we wanted to be sure they match). I'm ok with either approach (and use `pre-commit` everywhere I can, and find that it's extra value when using agents to code).

HACKING.md
```diff
@@ -18,6 +18,9 @@ uv tool install tox --with tox-uv
 uv tool update-shell
 ```
 
+Optionally, to run checks automatically before each commit, install
+[pre-commit](https://pre-commit.com/#install) and run `pre-commit install`.
```
tonyandrewmeyer, line 22: ❤️

.pre-commit-config.yaml
```diff
@@ -14,27 +14,31 @@ repos:
       - id: check-toml
       - id: mixed-line-ending
       - id: end-of-file-fixer
+        exclude: '^docs/reuse/best-practices\.txt$'
       - id: trailing-whitespace
+        exclude: '^docs/reuse/best-practices\.txt$'
```
tonyandrewmeyer, line 19: Should we fix the script that generates this file instead?

.pre-commit-config.yaml
```diff
@@ -14,27 +14,31 @@ repos:
...
+        name: ruff-format (uv)
+        entry: uv run --group lint -- ruff format --preview --force-exclude
+        language: system
+        types_or: [python, pyi]
+        exclude: '^(examples/|docs/conf\.py|docs/\.sphinx/|tracing/ops_tracing/vendor/)'
+      - id: codespell
+        name: codespell (uv)
+        entry: uv run --group lint -- codespell
+        language: system
+        types: [text]
```
tonyandrewmeyer, line 39: Feel free to push back for this being outside the scope of the PR - I wonder if we should have `pre-commit` and the tox commands be more similar, so - since we don't have codespell running with tox - maybe we should drop it from `pre-commit`. I think maybe we had it in the past and found it too much noise for too little signal?

james-garner-canonical: It looks like we do still use `codespell` in our `tox`. You might be thinking of when we dropped it from `charmlibs`, it was much more painful in a multiple codeowners context. Of course if we just ran `tox` we wouldn't need to think about this :smile:

tonyandrewmeyer: > It looks like we do still use `codespell` in our `tox`.

Huh, I swear I searched for "spell" before writing that. Maybe I misspelt "spell" 😆.

> Of course if we just ran `tox` we wouldn't need to think about this 😄

Touché!

.pre-commit-config.yaml
dimaqq: I'd do that in a separate PR

PR #2358, opened 2026-03-01
feat: add kind argument to testing.State.get_relation + get_relations

james-garner-canonical:
Alternative to #2326 for comparison.

This PR adds an optional `kind` keyword-only argument to `testing.State.get_relation` and `testing.State.get_relations` that can be used to narrow the type of the returned relation object(s).

In comparison to #2326 which adds 3 (or 6) new methods and essentially soft deprecates 1 (or 2) methods, this PR would result in less API churn, adding only the `kind` keyword arguments.

Open question: as @dimaqq [commented](https://github.com/canonical/operator/pull/2326#discussion_r2810990451), while I'm proposing raising a `TypeError` on a mismatch, a `ValueError` would make sense too.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -510,6 +510,7 @@ nitpick_ignore = [
     ('py:class', 'scenario.state.CharmType'),
     ('py:class', 'scenario.state._EntityStatus'),
     ('py:class', 'scenario.state._Event'),
+    ('py:class', 'scenario.state._RelationType'),
 ]
 
 # Pull in fix from https://github.com/sphinx-doc/sphinx/pull/11222/files to fix


testing/src/scenario/state.py
```diff
@@ -1739,24 +1741,58 @@ def get_storage(
...
+    def get_relation(
+        self, relation: int, /, *, kind: type[_RelationType] = RelationBase
+    ) -> _RelationType:
+        """Get relation from this State, based on the relation's id.
+
+        If the relation is not an instance of ``kind``, a ``TypeError`` is raised.
+        This allows type checkers to know the return type more precisely, for example::
+
+            rel = state.get_relation(my_id, kind=testing.PeerRelation)
+            reveal_type(rel)  # testing.PeerRelation
```
dimaqq, line 1753: **minor**: while reveal_type is amazing, I dunno if we want to include that in a docstring.

could there be a duck-typed example of `rel.something` that's only available for one relation kind?

testing/src/scenario/state.py
```diff
@@ -1739,24 +1741,58 @@ def get_storage(
             f'storage: name={storage}, index={index} not found in the State',
         )
 
-    def get_relation(self, relation: int, /) -> RelationBase:
-        """Get relation from this State, based on the relation's id."""
+    def get_relation(
+        self, relation: int, /, *, kind: type[_RelationType] = RelationBase
+    ) -> _RelationType:
```
dimaqq, line 1746: This looks alright.

However, I wonder if there's a chance we'd come to regret the somewhat quirky API... I mean it's hard to imagine why exactly, maybe a custom RelationBase subclass? or if one day pyright disagrees with mypy and we'd have to switch the default to `kinda=None` with overloads? Charmers ending up with code like `PEER=testing.PeerRelation; rel = state.get_relation(..., PEER)` to emulate an enum? Or someone writing generic code and introducing `NonPeerRelation = Relation|SubordinateRelation`?

P.S. my LLM tool suggests `type=...` or `cls=...` over `kind=...` as more intuitive. I'm not sure, but PTAL.

testing/src/scenario/state.py
```diff
@@ -1739,24 +1741,58 @@ def get_storage(
...
+        """Get relation from this State, based on the relation's id.
+
+        If the relation is not an instance of ``kind``, a ``TypeError`` is raised.
+        This allows type checkers to know the return type more precisely, for example::
+
+            rel = state.get_relation(my_id, kind=testing.PeerRelation)
+            reveal_type(rel)  # testing.PeerRelation
+        """
```
tonyandrewmeyer, line 1754: This docstring should really already have a "Raises" section that covers the `KeyError` possibility. The `TypeError` (I think `TypeError` is a better choice than `ValueError`) possibility should be there too. Could you add that in?

testing/src/scenario/state.py
```diff
@@ -1739,24 +1741,58 @@ def get_storage(
...
+                f'relation id={relation} is not a {kind.__name__}'
+                f", it's a {type(rel).__name__}: {rel}"
+            )
+        return rel
 
-    def get_relations(self, endpoint: str) -> tuple[RelationBase, ...]:
-        """Get all relations on this endpoint from the current state."""
+    def get_relations(
+        self, endpoint: str, *, kind: type[_RelationType] = RelationBase
+    ) -> tuple[_RelationType, ...]:
```
tonyandrewmeyer, line 1770: The choice in this PR to have a `kind=` rather than increase the number of methods does somewhat reduce my objection to leaving `get_relations` alone, but I would still prefer that we leave this untouched, because it's an antipattern in the examples we have, so we want to push people towards `get_relation` rather than using this, and giving `get_relation` more advantages over it helps do that.

If, in the future, there were valid use-cases for `get_relations` that would really benefit from also having this change, it would be easy to add it at that time.

testing/src/scenario/state.py
```diff
@@ -1739,24 +1741,58 @@ def get_storage(
...
+        """Get all relations on this endpoint from the current state.
+
+        If any relation on this endpoint is not an instance of ``kind``, a ``TypeError`` is raised.
+        This allows type checkers to know the return type more precisely, for example::
+
+            rels = state.get_relations(my_endpoint, kind=testing.SubordinateRelation)
+            for rel in rels:
+                reveal_type(rels)  # testing.PeerRelation
+        """
```
tonyandrewmeyer, line 1779: Probably obvious, but just to avoid it being missed: if we do include this change, then let's avoid `reveal_type` here and also have a "Raises" block, like with the above.

testing/src/scenario/state.py
```diff
@@ -1739,24 +1741,58 @@ def get_storage(
...
         # we rather normalize the endpoint than worry about cursed metadata situations such as:
         # requires:
         #   foo-bar: ...
         #   foo_bar: ...
 
         normalized_endpoint = _normalise_name(endpoint)
-        return tuple(
+        rels = tuple(
```
tonyandrewmeyer, line 1786: I think in a situation like this, it's better to just have the loop building the response rather than to have a tuple from a genexp and then run through the tuple.

I haven't actually checked the performance, but I imagine it's pretty similar (one list->tuple conversion at the end, countered by having two loops). In terms of readability I think that the one loop is much clearer.

testing/tests/test_e2e/test_relations.py
```diff
@@ -672,6 +672,105 @@ def _on_config_changed(self, _):
     assert isinstance(exc.value, KeyError) or isinstance(exc.value.__cause__, KeyError)
 
 
+def test_get_regular_relation():
```
tonyandrewmeyer, line 675: It seems like this would be simpler parametrized. It could be three values - each of the subclasses - and then a success case and the two failure cases. Or a pair of functions, one for the successes and one for the failures.

I think I've seen comments from you that you don't like parametrizing for cases like this, so if that's the case, then feel free to ignore this - I don't mind the explicitness, but I do slightly feel like the simpler parameterized version would be cleaner.

testing/src/scenario/state.py
james-garner-canonical: Thanks for the review. I've been wondering if there's a third way that might be less warty than either of these, or if we might be better off not making a change here at all.

I think doing something at all is much more useful for `get_relations` than for `get_relation`, since `rel = ...get_relation(...)` can always be followed by `assert isinstance(rel, ...)`.

Why is `get_relations` an anti-pattern? It seems to me like it's the more ideal level of abstraction for charm tests to be thinking about.

tonyandrewmeyer: > Why is `get_relations` an anti-pattern? It seems to me like it's the more ideal level of abstraction for charm tests to be thinking about.

This is mostly copied from the other PR:

I'm pretty sure when we've looked at use of this in the past, it's almost always code like:

```python
state_in = State(..., relations={relation})
state_out = ctx.run(..., state=state_in)
assert state_out.get_relations('endpoint')[0]...
```

And that is cleaner as:

```python
state_in = State(..., relations={relation})
state_out = ctx.run(..., state=state_in)
assert state_out.get_relation(relation.id)...
```

I had a quick look over a few random selections in charms, and they were all of that type.

Typically, in a test you want to be asserting that the relation data is what you expect given the event that just happened. So if you ran a `relation_changed(relation_x)` then you're wanting to assert things on the data in relation x, and that relation is identified by `relation_x.id`, so you want to pull it out of the output state by ID, to make sure that you're working with that same relation.

The `[0]` is the particular anti-pattern that I'm thinking of, or (I've seen this somewhere too, but I don't remember exactly where and haven't gone searching for it) something like:

```python
for rel in state_out.get_relations(endpoint):
    if rel.id == some_specific_id:
        assert rel...
```

If you're wanting to assert something about all relations for a specific endpoint, then `get_relations` is reasonable, but I don't think it's very common that a hook for a specific relation would result in changes to other relations on the same endpoint.

Very happy to be proven wrong with some examples, but I couldn't find any with quick searching (or from memory).

james-garner-canonical: I agree that `get_relations('endpoint')[0]` is less than desirable, but I don't think that means using endpoint names to work with relations in tests is an anti-pattern -- it still seems better for charm tests to be thinking about endpoints rather than relation IDs.

Getting the same relation out after the event is different, since you don't care about the actual ID, just that it's the same relation. This actually suggests another interesting alternative API that solves our type checking woes in this specific case: extend `get_relation` to accept a relation instance as an alternative to the `int` ID.
```py
@typing.overload
def get_relation(relation: int, /) -> RelationBase: ...
@typing.overload
def get_relation(relation: _RelationType, /) -> _RelationType: ...
def get_relation(relation: int | _RelationType, /) -> RelationBase:
    if not isinstance(relation, int):
        relation = relation.id
    ...
```

tonyandrewmeyer: > I agree that `get_relations('endpoint')[0]` is less than desirable, but I don't think that means using endpoint names to work with relations in tests is an anti-pattern -- it still seems better for charm tests to be thinking about endpoints rather than relation IDs.

I agree that you shouldn't be thinking about relation IDs in a way that means you need to know what the ID is. But passing in the ID is a convenient way to say "I want that same object as before".

> Getting the same relation out after the event is different, since you don't care about the actual ID, just that it's the same relation. This actually suggests another interesting alternative API that solves our type checking woes in this specific case: extend `get_relation` to accept a relation instance as an alternative to the `int` ID.
> 
> ```python
> @typing.overload
> def get_relation(relation: int, /) -> RelationBase: ...
> @typing.overload
> def get_relation(relation: _RelationType, /) -> _RelationType: ...
> def get_relation(relation: int | _RelationType, /) -> RelationBase:
>     if not isinstance(relation, int):
>         relation = relation.id
>     ...
> ```

In the Scenario 7 design, we did at some point have a system where you'd pass in the actual input object itself, like `rel_out = state_out.get_relation(rel_in)` (and even a generic one where you could pass any kind of state object, which would be pretty annoying from a typing point of view, I guess). So that's pretty similar to this. I'll try to find the references to that - I can't remember why we dropped it (I think maybe with some objects it didn't work very well and we wanted it to be consistent).

tonyandrewmeyer: @james-garner-canonical take a look at https://github.com/canonical/ops-scenario/pull/134 and let me know what you think of that (here or in daily - maybe I was premature adding it to the agenda there...).

james-garner-canonical: Thanks for linking to that, good to see how this was decided originally. I agree with [your decision to start simple with ID only](https://github.com/canonical/ops-scenario/pull/134#issuecomment-2149144885). As you said:

> We can always expand on them later if people find that passing in the object really would be a significant improvement.

james-garner-canonical:
Closing in favour of:
- #2359

PR #2359, opened 2026-03-03
feat: allow testing.State.get_relation to accept relation objects

james-garner-canonical:
This PR updates `testing.State.get_relation` to accept a relation object. The means that users can write `state_out.get_relation(rel_in)` instead of `state_out.get_relation(rel_in.id)`, and if they do so, the returned type is guaranteed to match the relation type passed in (peer, subordinate, or regular) -- both statically and at runtime.

Resolves: #2243

Alternative to:
- #2326 
- #2358

---

For context on why `testing.State.get_relation` only accepts `int` currently, see discussion on the 2024 PR where this was designed: https://github.com/canonical/ops-scenario/pull/134

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1768,7 +1768,7 @@ class State:
                     raise ValueError(
                         f'State.get_relation() result does not match\n'
                         f'Called with:\n{relation!r}\n'
-                        Found:\n{state_relation!r}\n'
+                        f'Found:\n{state_relation!r}\n'
                     )
                 return state_relation
         raise KeyError(f'relation: id={rel_id} not found in the State')


testing/src/scenario/state.py
```diff
@@ -1739,12 +1741,44 @@ def get_storage(
...
+            relation: A relation ID, or a relation object from a previous state.
+                If a relation object is passed, its ID is used to find the relation, and the
+                type, endpoint and interface of the found relation are validated against it.
+
+        Raises:
+            KeyError: If no relation matching the id is found.
+            ValueError: If a relation object is passed but the found relation
+                has a mismatched type, endpoint, or interface.
+        """
+        rel_id = relation if isinstance(relation, int) else relation.id
```
dimaqq: I'd personally invert this, and test `isinstance(rel, RelationBase)`.

Somehow, it feels more Pythonic to test against a narrower (user-defined in this case) class.

Like.. I don't expects charmers to subclass `int` so much, so I guess it's more of a sanity thing / less confusion when reading the code.

WDYT about:

```suggestion
        if isinstance(relation, RelationBase):
            rel_in = relation
            rel_id = relation.id
        else:
            rel_in = None
            rel_id = relation
```

Then down below, instead of `if not isinstance(relation, int): do_checks`, the comparison would be `if rel_in is not None: do_checks`.

james-garner-canonical: That would work, though I kind of prefer the one line conditional assignment to the six line version given that we only check the extra variable in one case. How about we invert it but keep it terse?
```py
        rel_id = relation.id if isinstance(relation, RelationBase) else relation
        # ...
                if isinstance(relation, RelationBase):
                    # do checks
```

dimaqq: Either is fine.
I figure it's an implementation detail, so should not be blocking -- could be changed back and forth in any micro release.

testing/src/scenario/state.py
```diff
@@ -1739,12 +1741,44 @@ def get_storage(
...
         for state_relation in self.relations:
-            if state_relation.id == relation:
+            if state_relation.id == rel_id:
+                if not isinstance(relation, int):
+                    if type(state_relation) is not type(relation):
+                        raise ValueError(
+                            f'relation with id={rel_id} has type {type(state_relation).__name__!r}'
+                            f', but get_relation was called with {relation!r}'
```
dimaqq: On the fourth thought, I figure the error is not particularly ergonomic here.

We'll only hit this when the user supplies the relation object, which hey would presumably not hardcode the relation id.

Thus telling them the problematic relation's id gets them nowhere.

Would it be better to report a relation type mismatch, requested `relation!r`, found in state `state_relation!r`?

(Or even more concisely, `{relation=} {state_relation=}`).

testing/src/scenario/state.py
james-garner-canonical: Done.

testing/src/scenario/state.py
james-garner-canonical: It should be essentially impossible to get this error state without manually messing with IDs somehow, since Scenario uses auto-incrementing relation IDs behind the scenes, so I think we should report the offending ID.

However, I like your idea of collapsing the error, let me know what you think of the new approach.

PS: I'm dropping validating the interface from this PR, because I'm worried that it may lead to unexpected issues in future. The documentation threatens that it will be auto-populated from the charm's metadata if not specified, but I think this only happens with `State.from_context` (where you don't specify anything manually). But if we ever did start auto-populating this during `Context.run` or similar it would be easy to get an unexpected mismatch here (manually create `Relation` without `interface` name, then query post-run `State` and get mismatch).

testing/src/scenario/state.py
```diff
@@ -1758,25 +1758,18 @@ def get_relation(self, relation: int | RelationBase, /) -> RelationBase:
...
-                            f', but get_relation was called with {relation!r}'
-                        )
+                if isinstance(relation, RelationBase) and (
+                    (type(state_relation), state_relation.endpoint)
+                    != (type(relation), relation.endpoint)
+                ):
+                    raise ValueError(
+                        f'State.get_relation() result does not match\n'
+                        f'Called with:\n{relation!r}\n'
+                        f'Result:\n{state_relation!r}\n'
```
dimaqq: s/Result/Found/ probably? because it aint the result yet 😃

testing/src/scenario/state.py
```diff
@@ -1758,25 +1758,18 @@ def get_relation(self, relation: int | RelationBase, /) -> RelationBase:
...
-                            f', but get_relation was called with {relation!r}'
-                        )
-                    if state_relation.interface != relation.interface:
-                        raise ValueError(
-                            f'relation with id={rel_id} has interface {state_relation.interface!r}'
-                            f', but get_relation was called with {relation!r}'
-                        )
+                if isinstance(relation, RelationBase) and (
+                    (type(state_relation), state_relation.endpoint)
+                    != (type(relation), relation.endpoint)
```
dimaqq, line 1766: Is it the case that we don't care about the interface, because realistically we infer it from charm metadata, so it can't really change?

james-garner-canonical: I think validating interfaces here might come back to haunt us in the future, as there are behind the scenes shenanigans involved. See [other comment](https://github.com/canonical/operator/pull/2359#discussion_r2915289081) (relevant bit quoted below).

> PS: I'm dropping validating the interface from this PR, because I'm worried that it may lead to unexpected issues in future. The documentation threatens that it will be auto-populated from the charm's metadata if not specified, but I think this only happens with `State.from_context` (where you don't specify anything manually). But if we ever did start auto-populating this during `Context.run` or similar it would be easy to get an unexpected mismatch here (manually create `Relation` without `interface` name, then query post-run `State` and get mismatch).

testing/src/scenario/state.py
james-garner-canonical:
```suggestion
                        Found:\n{state_relation!r}\n'
```

testing/src/scenario/state.py
```diff
@@ -1739,12 +1741,37 @@ def get_storage(
...
-            if state_relation.id == relation:
+            if state_relation.id == rel_id:
+                if isinstance(relation, RelationBase) and (
+                    (type(state_relation), state_relation.endpoint)
+                    != (type(relation), relation.endpoint)
+                ):
+                    raise ValueError(
+                        f'State.get_relation() result does not match\n'
+                        f'Called with:\n{relation!r}\n'
+                        Found:\n{state_relation!r}\n'
```
```suggestion
                        f'Found:\n{state_relation!r}\n'
```

PR #2360, opened 2026-03-03
ci: check example charms with mypy in CI

james-garner-canonical:
This PR updates CI for our example charms to run the `mypy` type checker. We recommend using `pyright` in our examples, so we don't want to add `mypy` to the examples themselves, but we'd like to make sure that `ops` is usable by `mypy` users. This CI change will help us preemptively spot issues like the one we recently fixed in #2343, whether they come up due to changes in `ops` or changes in `mypy`.

It would be nice to be able to run this check locally, but I don't think we want to put `mypy` in the examples. It would be cool if our top-level `tox -e lint` and `tox -e unit` would run all the example charm checks, including this one though, right? Something for future.

This PR is an alternative to #2339, which proposes adding `mypy` to `tox -e lint` for the `httpbin-demo` charm only. I prefer this PR.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -559,6 +559,91 @@ wheels = [
 ]
 
 [[package]]
+name = "librt"
+version = "0.8.1"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/56/9c/b4b0c54d84da4a94b37bd44151e46d5e583c9534c7e02250b961b1b6d8a8/librt-0.8.1.tar.gz", hash = "sha256:be46a14693955b3bd96014ccbdb8339ee8c9346fbe11c1b78901b55125f14c73", size = 177471, upload-time = "2026-02-17T16:13:06.101Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/7c/5f/63f5fa395c7a8a93558c0904ba8f1c8d1b997ca6a3de61bc7659970d66bf/librt-0.8.1-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:81fd938344fecb9373ba1b155968c8a329491d2ce38e7ddb76f30ffb938f12dc", size = 65697, upload-time = "2026-02-17T16:11:06.903Z" },
+    { url = "https://files.pythonhosted.org/packages/ff/e0/0472cf37267b5920eff2f292ccfaede1886288ce35b7f3203d8de00abfe6/librt-0.8.1-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:5db05697c82b3a2ec53f6e72b2ed373132b0c2e05135f0696784e97d7f5d48e7", size = 68376, upload-time = "2026-02-17T16:11:08.395Z" },
+    { url = "https://files.pythonhosted.org/packages/c8/be/8bd1359fdcd27ab897cd5963294fa4a7c83b20a8564678e4fd12157e56a5/librt-0.8.1-cp310-cp310-manylinux1_i686.manylinux_2_28_i686.manylinux_2_5_i686.whl", hash = "sha256:d56bc4011975f7460bea7b33e1ff425d2f1adf419935ff6707273c77f8a4ada6", size = 197084, upload-time = "2026-02-17T16:11:09.774Z" },
+    { url = "https://files.pythonhosted.org/packages/e2/fe/163e33fdd091d0c2b102f8a60cc0a61fd730ad44e32617cd161e7cd67a01/librt-0.8.1-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.manylinux_2_28_aarch64.whl", hash = "sha256:5cdc0f588ff4b663ea96c26d2a230c525c6fc62b28314edaaaca8ed5af931ad0", size = 207337, upload-time = "2026-02-17T16:11:11.311Z" },
+    { url = "https://files.pythonhosted.org/packages/01/99/f85130582f05dcf0c8902f3d629270231d2f4afdfc567f8305a952ac7f14/librt-0.8.1-cp310-cp310-manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_28_x86_64.whl", hash = "sha256:97c2b54ff6717a7a563b72627990bec60d8029df17df423f0ed37d56a17a176b", size = 219980, upload-time = "2026-02-17T16:11:12.499Z" },
+    { url = "https://files.pythonhosted.org/packages/6f/54/cb5e4d03659e043a26c74e08206412ac9a3742f0477d96f9761a55313b5f/librt-0.8.1-cp310-cp310-manylinux_2_31_riscv64.manylinux_2_39_riscv64.whl", hash = "sha256:8f1125e6bbf2f1657d9a2f3ccc4a2c9b0c8b176965bb565dd4d86be67eddb4b6", size = 212921, upload-time = "2026-02-17T16:11:14.484Z" },
+    { url = "https://files.pythonhosted.org/packages/b1/81/a3a01e4240579c30f3487f6fed01eb4bc8ef0616da5b4ebac27ca19775f3/librt-0.8.1-cp310-cp310-musllinux_1_2_aarch64.whl", hash = "sha256:8f4bb453f408137d7581be309b2fbc6868a80e7ef60c88e689078ee3a296ae71", size = 221381, upload-time = "2026-02-17T16:11:17.459Z" },
+    { url = "https://files.pythonhosted.org/packages/08/b0/fc2d54b4b1c6fb81e77288ff31ff25a2c1e62eaef4424a984f228839717b/librt-0.8.1-cp310-cp310-musllinux_1_2_i686.whl", hash = "sha256:c336d61d2fe74a3195edc1646d53ff1cddd3a9600b09fa6ab75e5514ba4862a7", size = 216714, upload-time = "2026-02-17T16:11:19.197Z" },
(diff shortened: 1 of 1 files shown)


.github/workflows/example-charm-tests.yaml
dimaqq, line 1: WDYT about adding this command into tox.ini, perhaps as a step for `tox -e lint`.

that would pin the mypy version like other tools
that still allows the fancy command line arguments

james-garner-canonical: It would have to go in the tox.ini files for our examples. But we don't actually want to recommend this check to users who refer to our examples.

An alternative would be to include both the iteration over the examples and the mypy check in an environment defined in the top-level tox.ini. That would be nice for running the check locally, but I'm not sure how best to define the iteration there.

dimaqq: Ah I see the issue...

One option would be to name the tox environment something like `ops-private-additional-lint` or something.

.github/workflows/example-charm-tests.yaml
```diff
@@ -54,3 +54,17 @@ jobs:
         run: |
           cd ${{ matrix.dir }}
           tox -e unit
+      - name: Run mypy
```
dimaqq, line 68: **super-minor**: if the new checks are only expressed in actions, one alternative would be to run `mypy` right after `tox -e lint` rather than here.

Pros/cons are only in terms of run time and what is surfaced when one (or both) of the linters produce an error.

.github/workflows/example-charm-tests.yaml
```diff
@@ -54,3 +54,17 @@ jobs:
...
+        working-directory: ${{ matrix.dir }}
+        env:
+          MYPYPATH: 'src:lib:../..'
+        run: |
+          uv run \
```
dimaqq, line 73: Should it be uvx?

```suggestion
          uvx run \
```

It's not like we're relying on the `pyproject.toml` to source the locally defined tool.

.github/workflows/example-charm-tests.yaml
```diff
@@ -54,3 +54,17 @@ jobs:
...
+            --with 'mypy~=1.19.0' \
```
dimaqq: I would have expected looser version req:

```suggestion
            --with 'mypy~=1.19' \
```

That is when mypy 1.20 is released, I'd want to have that run.

That's more of an issue specifically because it's only defined in this obscure action file, and not subject to regular dependency bumps.

dimaqq:
It's a good idea overall, let's figure out the maintainability aspect.

.github/workflows/example-charm-tests.yaml
tonyandrewmeyer: Also, `~1.19.0` isn't valid anyway. If you want to pin to anything compatible with 1.19 (including 1.20), that'd be `~1.19`.

If I recall correctly, dependabot isn't going to find this version, so it'll be stuck at whatever we have here until someone manually notices that it's years out of date. I think it'd be nice if we had it coming from `pyproject.toml` even though it's not used for ops itself (maybe it's in a dedicated group or something).

james-garner-canonical: Because this is `uv run`-ing from the example charm's `pyproject.toml`, and installing Ops from `../..` as a dependency, it's tricky (impossible cleanly?) to use dependency groups from the Ops `pyproject.toml`.

Charmlibs had a similar problem, which I initially worked around by hardcoding the versions in the command runner (like this), and eventually moved to a top level `testing-requirements.txt` file. We could perhaps add a `mypy-requirements.txt` file somewhere.

---

Regarding the version, I suppose we should really just use `>=` since if there's a 2.0 we'd probably want to notice and start testing with it too.

tonyandrewmeyer: You could cheat by having an otherwise unused group in `pyproject.toml` and then a `test_infra` test that verified that they matched, I guess?

> Regarding the version, I suppose we should really just use >= since if there's a 2.0 we'd probably want to notice and start testing with it too.

👍

.github/workflows/example-charm-tests.yaml
james-garner-canonical: We're relying on the example charm's `pyproject.toml` to install the charm's deps though (including the lint, unit, and integration dependency groups).

.github/workflows/example-charm-tests.yaml
james-garner-canonical: > You could cheat by having an otherwise unused group in `pyproject.toml` and then a `test_infra` test that verified that they matched, I guess?

If we want it in `pyproject.toml`, and would be willing to write code that pulls it out of there to check that it matches, what do you think about having similar code to make it available in this workflow instead?

For example:
```
      - name: 
        shell: python
        run: |
          import tomllib, pathlib
          pyproject_toml = tomllib.loads(pathlib.Path('pyproject.toml').read_text())
          deps = '\n'.join(pyproject_toml['dependency-groups']['mypy'])
          print(deps)  # logging
          pathlib.Path('mypy-requirements.txt').write_text(deps)

      - name: Run mypy
        working-directory: ${{ matrix.dir }}
        env:
          MYPYPATH: 'src:lib:../..'
        run: |
          uv run \
            --with-requirements=../../mypy-requirements.txt \
            --with=../.. \
            --with=../../testing \
            --group=lint \
            --group=unit \
            --group=integration \
            mypy --explicit-package-bases --follow-imports=silent src tests
```

tonyandrewmeyer: Yeah, that probably is better than a `test_infra` test.

james-garner-canonical: @dimaqq WDYT about putting the `mypy` requirements either in top-level `pyproject.toml` (with a CI step to pull them out ephemerally for use here)?

dimaqq: `mypy` in top-level project file sounds good to me.
I'd like to see what it takes to make that work.

james-garner-canonical: > `mypy` in top-level project file sounds good to me. I'd like to see what it takes to make that work.

A workflow step like this:

> If we want it in `pyproject.toml`, and would be willing to write code that pulls it out of there to check that it matches, what do you think about having similar code to make it available in this workflow instead?
> 
> For example:
> 
> ```
>       - name: 
>         shell: python
>         run: |
>           import tomllib, pathlib
>           pyproject_toml = tomllib.loads(pathlib.Path('pyproject.toml').read_text())
>           deps = '\n'.join(pyproject_toml['dependency-groups']['mypy'])
>           print(deps)  # logging
>           pathlib.Path('mypy-requirements.txt').write_text(deps)
> 
>       - name: Run mypy
>         working-directory: ${{ matrix.dir }}
>         env:
>           MYPYPATH: 'src:lib:../..'
>         run: |
>           uv run \
>             --with-requirements=../../mypy-requirements.txt \
>             --with=../.. \
>             --with=../../testing \
>             --group=lint \
>             --group=unit \
>             --group=integration \
>             mypy --explicit-package-bases --follow-imports=silent src tests
> ```

.github/workflows/example-charm-tests.yaml
james-garner-canonical: I see the benefit of running directly after `tox -e lint` for logical organisation. But I prefer running at the end because I think the locally runnable `tox -e lint` and `tox -e unit` are both more important to know if they fail, and less likely to fail.

.github/workflows/example-charm-tests.yaml
james-garner-canonical: That makes sense. On balance I'd prefer to keep this in CI only rather than adding a `tox` environment.

.github/workflows/example-charm-tests.yaml
james-garner-canonical: I've added the discussed step to extract the `mypy` deps from `pyproject.toml`. I'll resolve this thread now and request re-review.

james-garner-canonical:
Requesting re-review as I've changed the approach to use our `pyproject.toml` while staying CI only (no new `tox` environment).

.github/workflows/example-charm-tests.yaml
```diff
@@ -54,3 +54,27 @@ jobs:
         run: |
           cd ${{ matrix.dir }}
           tox -e unit
+      - name: Write mypy-requirements.txt
+        shell: uv run --script --no-project {0}
+        run: |
+          # /// script
+          # dependencies = ["tomli"]  # drop for tomllib when we're Python 3.11+
```
tonyandrewmeyer, line 61: If the above was `uv --python=3.11 run ...` would that avoid this? Or would that not work for some reason?

james-garner-canonical: Yeah, I also considered that approach (though with `requires-python = ">=3.11"` in the script header rather than on the CLI), but I figured (after thinking very briefly) that installing an extra package was likely to be lighter weight than installing a whole Python version.

tonyandrewmeyer: Fair enough. I imagine a Python version is more likely to be in a very local cache, but either should be fine.

tonyandrewmeyer (approved):
LGTM

.github/workflows/example-charm-tests.yaml
james-garner-canonical: A uv cache? Github's caching for the runner? I kind of like the `requires-python` version if we can justify it, since it doesn't need any tidying up once we drop 3.10.

tonyandrewmeyer: I would have thought that GitHub would have put some sort of caching layer in so that it's not pulling from Astral all the time, because it's pretty common and they'd want to save bandwidth. But maybe not, and I don't think there's anything public that says this.

If it's uv I think the cache is off by default now? So we would need to turn it on, which would cache it I believe for runs of that workflow in this repo. You can poison that cache I think even as an external contributor so that's probably not worth doing even though there isn't much at risk in this particular workflow.

So I think let's just stick with what you have now and your reasoning that tomli is small.

PR #2363, opened 2026-03-04
feat: enable interactive debugging via `breakpoint` in testing

### LINKED ISSUE #2166: `breakpoint()` doesn't work in charm code during unit tests
### Problem

When using Scenario for unit tests, calling `breakpoint()` in charm code does nothing. It only works in test code or when using import `pdb; pdb.set_trace()` directly.

### Root Cause

The [ops Framework](https://github.com/canonical/operator/blob/main/ops/framework.py#L651) overrides `sys.breakpointhook` with a custom implementation that only activates when `JUJU_DEBUG_AT` environment variable is set. During unit tests, this variable isn't set, so `breakpoint()` silently returns without doing anything.

### Reproduction

1. Add `breakpoint()` to a charm event handler
1. Write a scenario test that invokes that handler
1. Run `pytest`
1. Breakpoint is never triggered

james-garner-canonical:
This PR enables interactive debugging via `breakpoint` calls in charm code in state-transition tests, by preventing Ops from redefining `sys.breakpointhook` when charm code is executed with `testing.Context.run`.

Resolves #2166.

## Issue

In `main`, Ops unconditionally overrides `sys.breakpointhook` with the `ops.Framework.breakpoint` method. This happens when `ops._main._Manager._make_framework` calls `ops.Framework.set_breakpointhook`.

`Framework.breakpoint` does nothing unless `JUJU_DEBUG_AT` is set. This means you can deploy a charm that calls `breakpoint` (or `self.framework.breakpoint`), and those breakpoints will be a no-op until you use `juju debug-code`.

A consequence of this is that `breakpoint` calls in charm code that are being driven by `ops.testing.Context.run` are no-ops, as reported in #2166.

## Fix

This PR updates the `CapturingFramework` of `ops.testing` to override the `set_breakpointhook` method with an implementation that does nothing. This means that in tests using `Context.run`, breakpoints in charm code will work the same as breakpoints in the test code itself -- dropping the user into an interactive debugger session (or not, if `PYTHONBREAKPOINT=0`; or raising an error if running tests across multiple processes with `pytest-xdist`).

## Implications

Searching for `breakpoint` in our known charms shows that it is common to set `PYTHONBREAKPOINT` to `pdb.set_trace` or `ipdb.set_trace`, but that `breakpoint` calls in Python code are not checked in.

To set breakpoints in charm code that are not triggered during state-transition tests, but will be active when running `juju debug-code`, users can either set `PYTHONBREAKPOINT=0` when running state-transition tests, or use `self.framework.breakpoint` instead.

## Documentation

I will be improving the docs for debugging a charm in a separate PR. Unless there's a great place to fit this in right now, I think we can defer documenting this change to that PR.

### DIFF
--- a/testing/tests/test_e2e/test_breakpoint.py
+++ b/testing/tests/test_e2e/test_breakpoint.py
@@ -1,0 +1,46 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+from __future__ import annotations
+
+import logging
+import sys
+import unittest.mock
+
+import pytest
+from scenario import Context
+from scenario.state import State
+
+import ops
+
+logger = logging.getLogger('testing logger')
+
+
+META = {'name': 'mycharm'}
+
+
+class MyCharm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        breakpoint()
+
+
+def test_breakpoint_calls_sys_breakpointhook(monkeypatch: pytest.MonkeyPatch):
+    """breakpoint() in charm code calls through to sys.breakpointhook during tests."""
+    mock_hook = unittest.mock.MagicMock()
+    monkeypatch.setattr(sys, 'breakpointhook', mock_hook)
+    ctx = Context(MyCharm, meta=META)
+    ctx.run(ctx.on.install(), State())
+    mock_hook.assert_called_once()
+
+
+def test_breakpoint_suppressed_by_pythonbreakpoint_env_var(monkeypatch: pytest.MonkeyPatch):
+    """When ``PYTHONBREAKPOINT=0``, ``breakpoint()`` should be a no-op.
+
+    Note that since we use pytest-xdist, we would get a BdbQuit exception if sys.breakpointhook
+    was called. If we run with ``-n0 --dist=no``, we would land in an interactive session.
+    But since we set ``PYTHONBREAKPOINT=0``, neither happens and the test completes successfully.
+    """
+    monkeypatch.setenv('PYTHONBREAKPOINT', '0')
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):
This solution works for me.

I wonder if it would be cleaner if framework.set_breakpointhook had the logic to check the environment variable, but that is more of an invasive change, and maybe the `hook` breakpoint wouldn't work with that? I haven't looked at the Ops code in detail.

james-garner-canonical:
> I wonder if it would be cleaner if framework.set_breakpointhook had the logic to check the environment variable, but that is more of an invasive change, and maybe the `hook` breakpoint wouldn't work with that? I haven't looked at the Ops code in detail.

I'm not sure I understand how that approach would work.

Do you mean if `ops.Framework.set_breakpointhook` only changes `sys.breakpointhook` when `JUJU_DEBUG_AT` is set? I think that would be a concerning change because we'd go from `breakpoint` being a no-op in charm code unless `JUJU_DEBUG_AT` was set, to triggering the default breakpoint handler if it's not set. I could see that working if Juju set `PYTHONBREAKPOINT` according to whether it was running with `juju debug-code` or not, but the [only hits in the juju repo](https://github.com/search?q=repo%3Ajuju%2Fjuju%20PYTHONBREAKPOINT&type=code) are in test charms.

tonyandrewmeyer:
> Do you mean if `ops.Framework.set_breakpointhook` only changes `sys.breakpointhook` when `JUJU_DEBUG_AT` is set?

Yes. But like I said I hadn't looked into it to see if it would work or not, it just seemed like maybe the framework approach wasn't the right one given how this is used. But your explanation makes sense, and I'm still good with doing this specifically in Scenario.

PR #2364, opened 2026-03-05
chore: tidy up SSL context creation now that we've dropped Python 3.8

### LINKED ISSUE #2055: Clean up SSLContext handling in ops-tracing
The comment states:

```py
        # Note that ssl.create_default_context() doesn't allow setting the context.protocol in a
        # way that's the same across Python 3.8 and 3.10 onwards. Whip the context up by hand.
```

Now that Python 3.8 support has been dropped, the code can be simplified.

https://github.com/canonical/operator/blob/e5ee94a7cf052659edcae0ce7a28f73c96d15b40/tracing/ops_tracing/_export.py#L108-L128

james-garner-canonical:
This PR tidies up SSL context in `ops-tracing`, using `create_default_context` now that we've dropped Python 3.8.

Due to changes to the default behaviour in Python 3.13, we still need two lines of manual changes afterwards. The [docs](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) say:

> The protocol, options, cipher and other settings may change to more restrictive values anytime without prior deprecation. The values represent a fair balance between compatibility and security.
If your application needs specific settings, you should create a [SSLContext](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) and apply the settings yourself.

~This makes me wonder if we should stick to manually creating the `SSLContext` ourselves, though the docs do indicate that applying the settings is OK too -- I'm just wondering if anything else relevant might change underneath us.~

Resolves #2055


tracing/ops_tracing/_export.py
```diff
@@ -106,25 +106,18 @@ def ssl_context(self, ca: str | None) -> ssl.SSLContext:
         return context
 
     def _ssl_context(self, ca: str | None) -> ssl.SSLContext:
-        # Note that ssl.create_default_context() doesn't allow setting the context.protocol in a
-        # way that's the same across Python 3.8 and 3.10 onwards. Whip the context up by hand.
-        context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        context = ssl.create_default_context(cadata=ca)
```
james-garner-canonical, line 109: `ssl.create_default_context` [defaults to](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) `purpose=Purpose.SERVER_AUTH`. This [uses](https://github.com/python/cpython/blob/main/Lib/ssl.py#L682-L705) `SSLContext(ssl.PROTOCOL_TLS_CLIENT)`.

tracing/ops_tracing/_export.py
```diff
@@ -106,25 +106,18 @@ def ssl_context(self, ca: str | None) -> ssl.SSLContext:
...
         context.minimum_version = ssl.TLSVersion.TLSv1_2  # See comment at the top of module
         context.set_alpn_protocols(['http/1.1'])
+        # The partial chain flag allows trusting intermediate CAs in the CA list
+        # without the matching root CA.
+        # Enabled by default in Python 3.13+, needed explicitly before that.
+        context.verify_flags |= ssl.VERIFY_X509_PARTIAL_CHAIN
```
james-garner-canonical, line 115: This diff is annoyingly interleaved here. This chunk of additions corresponds to these lines below:
```diff
-        if partial_chain := getattr(ssl, 'VERIFY_X509_PARTIAL_CHAIN', None):
-            # Available starting from Python 3.10. The partial chain flag allows trusting an
-            # intermediate CAs in the CA list without the matching root CA.
-            context.verify_flags |= partial_chain
```

tracing/ops_tracing/_export.py
```diff
@@ -106,25 +106,18 @@ def ssl_context(self, ca: str | None) -> ssl.SSLContext:
...
-            # intermediate CAs in the CA list without the matching root CA.
-            context.verify_flags |= partial_chain
-
-        if ca is not None:
-            context.load_verify_locations(cadata=ca)
-        else:
-            context.load_default_certs()
-
+        # Enabled by default in Python 3.13+, so must be explicitly cleared.
+        context.verify_flags &= ~ssl.VERIFY_X509_STRICT
```
james-garner-canonical, line 120: This diff is annoyingly interleaved here. This addition corresponds to this line above:
```diff
-        # context.verify_flags |= ssl.VERIFY_X509_STRICT
```
And matches this retained comment:
```
        # Can't use strict certificate chain validation until our self-signed ca is fixed
        # https://github.com/canonical/self-signed-certificates-operator/issues/330
        # https://github.com/canonical/tls-certificates-interface/pull/333
```

tracing/ops_tracing/_export.py
```diff
@@ -106,25 +106,18 @@ def ssl_context(self, ca: str | None) -> ssl.SSLContext:
...
-        # context.verify_flags |= ssl.VERIFY_X509_STRICT
-        if partial_chain := getattr(ssl, 'VERIFY_X509_PARTIAL_CHAIN', None):
-            # Available starting from Python 3.10. The partial chain flag allows trusting an
-            # intermediate CAs in the CA list without the matching root CA.
-            context.verify_flags |= partial_chain
-
-        if ca is not None:
-            context.load_verify_locations(cadata=ca)
-        else:
-            context.load_default_certs()
```
james-garner-canonical, line 126: This is now handled by `create_default_context`. Per the [docs](https://docs.python.org/3/library/ssl.html#ssl.create_default_context):
> Passing [SERVER_AUTH](https://docs.python.org/3/library/ssl.html#ssl.Purpose.SERVER_AUTH) as purpose sets [verify_mode](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_mode) to [CERT_REQUIRED](https://docs.python.org/3/library/ssl.html#ssl.CERT_REQUIRED) and either loads CA certificates (when at least one of cafile, capath or cadata is given) or uses [SSLContext.load_default_certs()](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs) to load default CA certificates.

tracing/ops_tracing/_export.py
```diff
@@ -106,25 +106,18 @@ def ssl_context(self, ca: str | None) -> ssl.SSLContext:
...
-        context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        context = ssl.create_default_context(cadata=ca)
         context.minimum_version = ssl.TLSVersion.TLSv1_2  # See comment at the top of module
         context.set_alpn_protocols(['http/1.1'])
+        # The partial chain flag allows trusting intermediate CAs in the CA list
+        # without the matching root CA.
+        # Enabled by default in Python 3.13+, needed explicitly before that.
+        context.verify_flags |= ssl.VERIFY_X509_PARTIAL_CHAIN
         # Can't use strict certificate chain validation until our self-signed ca is fixed
         # https://github.com/canonical/self-signed-certificates-operator/issues/330
```
dimaqq, line 117: I though we'd be able to drop that hack by now, but alas!
https://matrix.to/#/!yAkGlrYcBFYzYRvOlQ:ubuntu.com/$ibCVWod_F_dRl7afdHzfgDI5lGTbFhZI3jKtaQVbX50?via=ubuntu.com&via=matrix.org&via=laquadrature.net

dimaqq:
How I would validate this change:

1. Hack up Ops (ipython or unit test) to try to connect to (somewhere).
2. Run the {old,new} code x Python {3.8, 3.10, 3.12}.
3. Capture the traffic (e.g. wireshark) and inspect the initial handshake

How to inspect: one of
- A public HTTPS endpoint (like google.com) + Wireshark, or
- `openssl s_server` with some cert and cmd line flags that show decoded packets

(We've got integration tests, one uses self-signed-certs, but we don't have an integration test against a real cert provider).

james-garner-canonical:
> How I would validate this change:
> 
> 1. Hack up Ops (ipython or unit test) to try to connect to (somewhere).
> 2. Run the {old,new} code x Python {3.8, 3.10, 3.12}.
> 3. Capture the traffic (e.g. wireshark) and inspect the initial handshake
> 
> How to inspect: one of
> 
> * A public HTTPS endpoint (like google.com) + Wireshark, or
> * `openssl s_server` with some cert and cmd line flags that show decoded packets

I'm not sure that this is necessary, as the `SSLContext` objects created with code on `main` and the code from this PR should be identical on all the Python versions we test on (Python 3.10, 3.12, 3.14). I've added a test to demonstrate this, though I'm not sure it's worth keeping it around.

> (We've got integration tests, one uses self-signed-certs, but we don't have an integration test against a real cert provider).

I forgot that we don't run integration tests on PRs, I've kicked off a run [here](https://github.com/james-garner-canonical/operator/actions/runs/22706912502).

tracing/test/test_backend.py
```diff
@@ -85,3 +85,52 @@ def test_exporter_ssl_context(tmp_path: pathlib.Path):
...
+    ca = _CA if use_ca else None
+    new = BufferingSpanExporter(tmp_path / 'buffer').ssl_context(ca)
+    original = _ssl_context_original(ca)
+    assert type(new) is type(original)
+    assert dir(new) == dir(original)
+    for attr in dir(new):
+        new_attr = getattr(new, attr)
+        if callable(new_attr):
+            continue
+        assert new_attr == getattr(original, attr)
```
dimaqq: If this only were so simple.

IIRC we set ALPN or NPN protocols, but those can't be inspected.
I can't be sure if that's the only setting like that.

The point is that the change needs to be validated outside of Python process.

P.S. this equivalence test should be transient, I don't think that it needs to be checked in.

tracing/ops_tracing/_export.py
dimaqq, line 120: Having eyeballed this, it looks good.

tracing/test/test_backend.py
james-garner-canonical: True, one of the unchanged lines in in updated method is `context.set_alpn_protocols(['http/1.1'])`, and it doesn't look like we can get them back from the context. But given that `context` is still an `ssl.SSLContext`, why would we expect this line to do anything different now? [create_default_context](https://github.com/python/cpython/blob/main/Lib/ssl.py#L681) doesn't touch ALPN protocols.

I guess the real concern is that there are some other under the hood settings that we care about that are broken for us with `create_default_context`. If that's the case, then given that the docs indicate that `create_default_context` may change settings to be 'more restrictive' at any time, I wonder if `create_default_context` is the right fit for us, as we'd need to repeat manual validation whenever `ssl` changes.

> P.S. this equivalence test should be transient, I don't think that it needs to be checked in.

Agreed.

tonyandrewmeyer (commented):
I haven't reviewed this in detail - one question first: it feels like the goal of the issue was to get rid of creating a custom context that was required to support 3.8. However, because of other issues, we're still creating a custom context. What are we gaining here? Should we be waiting for some future Python version instead?

james-garner-canonical:
> I haven't reviewed this in detail - one question first: it feels like the goal of the issue was to get rid of creating a custom context that was required to support 3.8. However, because of other issues, we're still creating a custom context. What are we gaining here? Should we be waiting for some future Python version instead?

Yeah I feel the same way, and I suspect that `create_default_context` is always going to have this issue, or at least require careful attention with every Python release.

tonyandrewmeyer:
> Yeah I feel the same way, and I suspect that `create_default_context` is always going to have this issue, or at least require careful attention with every Python release.

Does that mean that it's not worth doing this issue, or that we need to have an issue like this with each Python release to handle whatever quirks are there? Or something else?

dimaqq:
Given that all we promise is this: https://documentation.ubuntu.com/ops/latest/reference/ops-tracing/#ops-tracing-security

I think that our CI should cover what we need.

Let's remove the transient equivalence test and merge.

dimaqq:
P.S. we could push min TLS to 1.3, because:
* https://charmhub.io/tempo-coordinator-k8s (min published revision is 79)
* https://github.com/canonical/tempo-coordinator-k8s-operator/commit/978259a5ac7351409a3bcfdd03e00232b771aed7 (merged in rev 75)

However, that could be a subtle breaking change for someone with an older, deployed COS.
So I'll punt that change to Ops 4.0 list.

james-garner-canonical:
Sorry for the delay in merging this. My hesitation is because I'm not sure that switching to `create_default_context` is really right for us, and I wanted to discuss that first.

My concern is that the defaults used by `create_default_context` may change with each Python release. From [the docs](https://docs.python.org/3/library/ssl.html#ssl.create_default_context):

> **Note:** The protocol, options, cipher and other settings may change to more restrictive values anytime without prior deprecation. The values represent a fair balance between compatibility and security.
>
> If your application needs specific settings, you should create a [SSLContext](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) and apply the settings yourself.

I don't know whether this is relevant to us or not -- how specific are our needs really? Would we expect breakages if the settings changed under us? If so, it seems like we'd be better off sticking with creating the context manually (with a little code cleanup since we've dropped Python 3.8). If not, then using `create_default_context` is probably correct, since we may get more secure settings in future 'for free'.

(Still TODO before merge: drop the equivalence test. Retaining for now in case we want further changes.)

dimaqq:
I think it's fine to trust the backwards compatibility in Python.

One could argue that we get OpenSSL from the Ubuntu base, which is also different version with different set of support algorithms and probably defaults base-to-base.

benhoyt:
Per discussion, let's use `create_default_context` and having it improve over time is okay (good).

PR #2365, opened 2026-03-05
fix: warn before clearing non-empty container in testing

### LINKED ISSUE #2041: Scenario silently wipes the container's filesystem if populated prior to running the context
related: https://github.com/canonical/operator/issues/1430

I just spent a good 40m debugging this with a colleague.
The situation is just as described in the previous ticket: instead of using a mount to pass a tempdir's contents to the charm runtime, we were attempting to directly write the container filesystem prior to running scenario.
However, as soon as the Container is initialized, the container's tempdir is wiped at the root by the `_MockPebbleClient` on `__init__`.

How about we add a warning, an error message, or even an exception if the filesystem has contents?
What legitimate use case or situation are we catering for by not doing it?
OTOH one might be doing:

```
state2 = ctx.run(state1)
ctx.run(state2)
```

which means all containers will get wiped by the second call, to avoid polluting the env of the second execution with leftovers from the first one. That's why the functionality was designed this way and pretty much the only semi-legitimate reason why we allow the filesystem to be 'nonempty' when we initialize a new run. 
However, this is also not 'the way' scenario was meant to be used, so a sternly worded warning is IMO warranted: "WARNING: container initialized with a non-empty filesystem (which will now be totally wiped). If you're trying to mock filesystem contents, you should be using Mounts. Otherwise this may mean you've been recycling context instances to run multiple events on a charm and you should be fine."

james-garner-canonical:
In testing, `Context.run` will clear the directories created for fake container filesystems before executing the charm code. As #2041 reported, this may be unexpected if the user is trying to populate the container filesystem before calling `Context.run`. This PR therefore adds a warning if the directory to be cleared has any contents.

Resolves #2041

#2041 suggests that there are legitimate reasons to call `Context.run` twice with the same `Context` object:

> OTOH one might be doing:
>
> ```
> state2 = ctx.run(state1)
> ctx.run(state2)
> ```
>
> which means all containers will get wiped by the second call, to avoid polluting the env of the second execution with leftovers from the first one. That's why the functionality was designed this way and pretty much the only semi-legitimate reason why we allow the filesystem to be 'nonempty' when we initialize a new run. 

However, the `Context` docstring says:

```
    Note that you can't call ``run()`` multiple times. The context 'pauses' ops
    right before emitting the event, but otherwise this is a regular test; you
    can't emit multiple events in a single charm execution.
```

If any charm tests were calling `Context.run` multiple times with the same `Context` object, and expecting the container directory to be wiped, they will have a new warning in their test output. I think this is OK, because we recommend against using a context this way.

What I'm wondering is if we should go further: this could be a logging `ERROR` now, and we could make it a runtime exception in our next major release.

### DIFF
--- a/testing/tests/test_e2e/test_pebble.py
+++ b/testing/tests/test_e2e/test_pebble.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 import dataclasses
 import datetime
 import io
-import logging
 from pathlib import Path
 
 import pytest
@@ -1003,8 +1002,7 @@ def test_warning_on_non_empty_container():
     ctx.run(ctx.on.start(), state)
 
     assert any(
-        'mycontainer' in line.message and 'non-empty' in line.message
-        for line in ctx.juju_log
+        'mycontainer' in line.message and 'non-empty' in line.message for line in ctx.juju_log
     )
 
 
@@ -1023,6 +1021,5 @@ def test_no_warning_on_empty_container():
     ctx.run(ctx.on.start(), state)
 
     assert not any(
-        'mycontainer' in line.message and 'non-empty' in line.message
-        for line in ctx.juju_log
+        'mycontainer' in line.message and 'non-empty' in line.message for line in ctx.juju_log
     )


testing/tests/test_e2e/test_pebble.py
```diff
@@ -880,3 +881,53 @@ def test_layers_merge_in_plan(layer1_name, layer2_name):
...
+    )
+    container = Container(name='mycontainer', can_connect=True)
+    state = State(containers={container})
+
+    # First run creates the container root.
+    ctx.run(ctx.on.start(), state)
+
+    # Second run should not warn since the container root is empty.
+    with caplog.at_level(logging.WARNING, logger='ops-scenario.mocking'):
+        ctx.run(ctx.on.start(), state)
```
dimaqq: I reuse the Context like this at times. I'm not sure I want to see warnings.
(to be fair, not start+start, rather start+config-changed+pebble-ready)

Would there always be a warning (because scenario dropped some files in), or only if my python-code-under-test explicitly modified container contents?

james-garner-canonical: Only if your code under test explicitly modifies container contents, as this test asserts.

It would be good to see some examples of `Context` reuse if you have any handy, given the docs say:
```
    Note that you can't call ``run()`` multiple times. The context 'pauses' ops
    right before emitting the event, but otherwise this is a regular test; you
    can't emit multiple events in a single charm execution.
```

tonyandrewmeyer: > It would be good to see some examples of `Context` reuse if you have any handy, given the docs say:
> 
> ```
>     Note that you can't call ``run()`` multiple times. The context 'pauses' ops
>     right before emitting the event, but otherwise this is a regular test; you
>     can't emit multiple events in a single charm execution.
> ```

I think what that means (maybe the wording should be clearer? I think it's pretty old and probably aimed more at how Scenario used to be) is that you cannot do something like:

```python
ctx = Context(MyCharm)
with ctx(ctx.on.config_changed(), State()):
    ctx.run(ctx.on.update_status(), State()
```

A "single charm execution" is, effectively, `ctx.run`. You can't run multiple events inside a "run".

But, like Dima, I feel it's legitimate to do:

```python
ctx = Context(MyCharm)
s2 = ctx.run(ctx.on.relation_joined(rel), s1)
s3 = ctx.run(ctx.on.relation_changed(rel), s2)
```

However, I think Pietro disagrees, based on the linked issue.

However, if we're only warning when there are files that will be destroyed, I think that's a reasonable compromise.

dimaqq: > It would be good to see some examples of Context reuse if you have any handy

`test_workload_version_is_setted` in zookeeper-k8s-operator/tests/unit/test_charm.py

---

See, Pietro is a reconciler, he'd have a fresh Context every time.
Delta charmers are not so lucky, they have build charm inner state bit by bit.

james-garner-canonical: Maybe container directory contents should be captured in the output `State` so that passing a `State` along between multiple `run` calls with the same `Context` matches real behaviour.

PR #2366, opened 2026-03-05
feat: support port ranges and endpoints

### LINKED ISSUE #2035: Add port range and endpoint support to open/close/set ports
Juju [supports](https://documentation.ubuntu.com/juju/3.6/reference/hook-command/list-of-hook-commands/open-port/) opening/closing ports as a range (e.g. 8000-8999) as well as single ports, and also for specific endpoints. In Ops, we only support the simpler single port approach.

It seems reasonable to support port ranges and endpoints in `open_port`, `close_port`, and `set_ports`. We should definitely support these in `opened_ports`, because charmers might have used a range or endpoint outside of Ops, and we should parse the returned data correctly.

When we do this, we can unify the dataclass for `Port` used in hookcmds and model.

james-garner-canonical:
Resolves #2035

TODO: Update Harness backend port methods with new arguments (that's why DB charm tests are failing).
TODO: Update Scenario port methods and classes.


ops/hookcmds/_port.py
```diff
@@ -150,7 +150,9 @@ def opened_ports(*, endpoints: bool = False) -> list[Port]:
     for port in result:
         if endpoints:
             port, port_endpoints = port.rsplit(' ', 1)
-            port_endpoints = [e.strip() for e in port_endpoints.strip('()').split(',')]
+            port_endpoints = [
+                endpoint for e in port_endpoints.strip('()').split(',') if (endpoint := e.strip())
```
james-garner-canonical: This is a bugfix: if we received an empty endpoints result `()`, the result should be `[]`, not `['']`.

ops/hookcmds/_types.py
```diff
@@ -249,7 +249,7 @@ def _from_dict(cls, d: dict[str, Any]) -> Network:
 class Port:
     """A port that Juju has opened for the charm."""
 
-    protocol: Literal['tcp', 'udp', 'icmp'] | None = 'tcp'
+    protocol: Literal['tcp', 'udp', 'icmp'] = 'tcp'
```
james-garner-canonical: This is a breaking change. But it's necessary if we want `hookcmds.Port` to be able to be used where `ops.Port` was.  It breaks anyone who called `Port(None)` or `Port(protocol=None)` -- but I can find no instances of this in our known charms.

ops/hookcmds/_types.py
```diff
@@ -258,9 +258,21 @@ class Port:
     to_port: int | None = None
     """The final port number if this is a range of ports."""
 
-    endpoints: list[str] | None = None
+    endpoints: tuple[str, ...] | None = None
```
james-garner-canonical: This is a breaking change too. It's necessary if we want ops to use `hookcmds.Port` instead of `ops.Port`, because all its attributes need to be hashable.

We add a custom `__init__`, so a `list` (or any iterable now) can still be passed in. However, this change still breaks anyone who mutated `Port.endpoints` after it was created, e.g. calling `endpoints.sort()`, or `endpoints.append(...)` or `endpoints[0] = ...`, etc. I haven't looked for uses like this.

ops/model.py
```diff
@@ -801,15 +801,23 @@ def set_ports(self, *ports: int | Port) -> None:
...
+            ('tcp', p, None, ())
+            if isinstance(p, int)
+            # If a user specifies a Port with endpoints=None (default) we normalise to endpoints=()
+            # but maybe this is too simple -- if the user didn't specify endpoints then do we
+            # actually want to take action if the port is already open but with some endpoints?
+            # In terms of backwards compatibility, probably not, since previously the endpoints
+            # would have been ignored ...
+            # FIXME: we might need to drop this nice set logic and do an explicit loop instead
+            else (p.protocol, p.port, p.to_port, p.endpoints or ())
+            for p in ports
```
james-garner-canonical: In terms of backwards compatibility, this problem applies equally to `to_port`, since previously we'd treat a request for `Port(port=80)` to be satisfied if the open port was `Port(port=80, to_port=90, endpoints=('foo', 'bar'))`.

ops/model.py
```diff
@@ -4037,25 +4035,40 @@ def secret_remove(self, id: str, *, revision: int | None = None):
...
-    def open_port(self, protocol: str, port: int | None = None):
-        with self._wrap_hookcmd('open-port', protocol=protocol, port=port):
-            hookcmds.open_port(protocol, port)
+    def open_port(
+        self,
+        protocol: str,
+        port: int | None = None,
+        *,
+        to_port: int | None = None,
+        endpoints: str | Iterable[str] | None = None,
```
james-garner-canonical: It feels awkward exposing this `str | Iterable[str]` API at this level, but maybe it's the right call since `str` is an `Iterable[str]` ... but it feels odd since Ops is normally higher level.

ops/hookcmds/_port.py
james-garner-canonical: The question is though, would we ever receive such a result? The comment above indicates yes, but I'm not sure I understand when this would happen.
```
    # If ``--endpoints`` is used, then each port will be followed by a
    # (possibly empty) tuple of endpoints.
```

tonyandrewmeyer: I think that is also a bug. I remember finding it a bit tricky to determine all the possible outputs here, and not being sure how to handle `*` as an endpoint, so I probably left in something that actually isn't possible.

I've tried various things locally and I can't get an empty list of endpoints; it's `*` or at least one.

Looking at the Juju code:

### Default mode (no `--endpoints` flag)                                                                                      
                  
One entry per line, each being a port range string in one of three forms:

| Format | Example | When |
|--------|---------|------|
| `<port>/<protocol>` | `80/tcp` | Single port |
| `<from>-<to>/<protocol>` | `8080-8088/udp` | Port range |
| `icmp` | `icmp` | ICMP protocol (no port numbers) |

Protocol is always lowercase. Entries are sorted by port range. Duplicate port ranges across different endpoints are deduplicated.

Example default output:

```
10-20/tcp
53-55/udp
63/udp
80/tcp
```

### With `--endpoints` flag

Same port range format, but each entry is suffixed with endpoint info in parentheses:

| Format | Example | When |
|--------|---------|------|
| `<portrange> (*)` | `80/tcp (*)` | Port applies to all endpoints (opened with empty endpoint `""`) |
| `<portrange> (<endpoint>)` | `63/udp (bar)` | Port applies to a single endpoint |
| `<portrange> (<ep1>, <ep2>)` | `80/tcp (endpoint1, endpoint2)` | Port applies to multiple specific endpoints (comma+space separated, sorted alphabetically) |

Example `--endpoints` output:

```
10-20/tcp (foo)
53-55/udp (*)
63/udp (bar)
80/tcp (*)
```

(The summary is from my friend Claude, to be nicely formatted, but I manually looked over the code).

ops/hookcmds/_types.py
tonyandrewmeyer: I don't think that is still a goal (the comment is out of date, I believe). We were going to do that, and wanted to have `model` use the raw `hookcmds` classes, but there were pros & cons to that approach and our team decision was to not do that. That's why we have `CloudCredential` and `CloudSpec` classes defined in both `hookcmds` and `model`, for example.

*However*, as per our Matrix discussion, I think the `None` was a mistake. `None` effectively would mean "use the Juju default protocol" but I think it's cleaner to just hard-code that in as always being `tcp` rather than having the type be the string or `None`. I don't think Juju is likely to change the default and if it does we'd want to discuss what to do here anyway.

ops/hookcmds/_types.py
tonyandrewmeyer: See the comment about about having only `hookcmds.Port` and not also `model.Port`.

In this case, there seems to be less reason to make the change. I do think it would be a bit odd to change the list (somewhat like Scenario, these are frozen and meant to be immutable, but are not deep frozen).

I think I would prefer to keep the simplicity of using `list`, though. I don't remember exactly, but I think the intention here was to stick to the types returned from the JSON where that was reasonably sensible.

Also, it seems unlikely anyone other than hookcmds is creating a `Port` object, so we know what the types of the arguments will be. Maybe you'd create one in some sort of tests, but I don't think we need a custom init for that.

ops/model.py
```diff
@@ -836,17 +844,7 @@ def reboot(self, now: bool = False) -> None:
...
-    """Represents a port opened by :meth:`Unit.open_port` or :meth:`Unit.set_ports`."""
-
-    protocol: typing.Literal['tcp', 'udp', 'icmp']
-    """The IP protocol."""
-
-    port: int | None
-    """The port number. Will be ``None`` if protocol is ``'icmp'``."""
-
-
+Port = hookcmds.Port
```
tonyandrewmeyer: As per above comments, I don't think we want to do this any more.

ops/model.py
```diff
@@ -801,15 +801,23 @@ def set_ports(self, *ports: int | Port) -> None:
...
+            for p in ports
         }
-        for protocol, port in existing - desired:
-            self._backend.close_port(protocol, port)
-        for protocol, port in desired - existing:
-            self._backend.open_port(protocol, port)
+        for protocol, p, to_port, endpoints in existing - desired:
+            self._backend.close_port(protocol, p, to_port=to_port, endpoints=endpoints)
+        for protocol, p, to_port, endpoints in desired - existing:
+            self._backend.open_port(protocol, p, to_port=to_port, endpoints=endpoints)
```
tonyandrewmeyer: Can we keep the name `port` here? It's clearer and matches `to_port`, which isn't `to_p`.

ops/model.py
tonyandrewmeyer: I think at the ModelBackend level, it's fine to keep this like it is, it's only for use by internal ops code. For charmers, I think we would want a higher level.

tonyandrewmeyer (commented):
Some initial notes on the code - I haven't looked at the tests at all.

One thing I think that is missing here is that as far as I can see, this only changes ModelBackend so it doesn't actually expose the functionality to the user for *setting*, only *getting* (which is the most important part, to be fair).

I can see an argument that this PR should *only* do that and there should be a follow-up that provides the ability to set as well, which has some additional design (for example, do you set ports for an endpoint via `Model` or `Relation`?).

I haven't checked what the behaviour is if you try to close a port that is inside a range or don't provide the endpoint but it was opened with an endpoint and things like that. If we're adding *setting* as well, then we should make sure we know what happens and probably have it in the docstring (maybe it's just a regular ModelError?). Also how `set_ports` works in those sorts of cases (possibly it's fine, since it's just closing exactly everything that isn't in the desired list, and opening the others).

ops/model.py
james-garner-canonical: Good catch, this was an accidental rename. I've refactored, so we're still using `p` here, but it's a `Port` object, so `p.protocol`, etc.

ops/model.py
james-garner-canonical: I think this partially goes away with a separate `ops.Port` type that has clearer defaults for endpoint, though backwards compatibly behaviour when requesting `set_ports(1)` if `1-1000` is open would still be a problem and might be desirable ... on the other hand, it's kind of a bugfix, but a very longstanding bug. On the other other hand, unlikely to be a problem since who would have opened the port range?

ops/hookcmds/_types.py
james-garner-canonical: I'll leave the `hookcmds.Port` type unchanged and bring back `ops.Port`, as only the latter needs to be hashable.

ops/hookcmds/_types.py
james-garner-canonical: Since we're not using `ops.hookcmds.Port` in place of `ops.Port` now, this doesn't need to be changed for this PR. I think it kind of makes sense to let it stay `None`-able in the low-level representation so you can encode 'let Juju set the default' -- although the actual hook commands take separate arguments rather than an `ops.hookcmds.Port` object anyway.

ops/hookcmds/_port.py
james-garner-canonical: Hmm, so because we'll never get `()` from Juju, the code is fine as-is -- it doesn't parse `()` correctly, but it's not a generalised `str(tuple(str, ...))` parser, it requires them to have at least length 1. I'll revert this change and instead clean up the comment indicating it could be `()`.

ops/model.py
james-garner-canonical: Agreed, I had forgotten that this was in `ModelBackend`. Resolving this and we can discuss the higher-level API elsewhere.

ops/model.py
```diff
@@ -765,21 +781,29 @@ def close_port(
...
+        if not endpoints:
+            raise TypeError('endpoints must be a non-empty string or sequence of strings')
+        self._backend.close_port(protocol.lower(), port, to_port=to_port, endpoints=endpoints)
 
     def opened_ports(self) -> set[Port]:
         """Return a list of opened ports for this unit."""
         return self._backend.opened_ports()
 
-    def set_ports(self, *ports: int | Port) -> None:
+    def set_ports(self, *ports: int | tuple[int, int | None] | Port) -> None:
```
benhoyt, line 806: Per discussion, I'd be inclined to drop this change, and they can pass in `Port` with `to_port` if they need it.

ops/hookcmds/_port.py
```diff
@@ -29,22 +29,22 @@ def close_port(
     *,
     to_port: int | None = None,
     endpoints: str | Iterable[str] | None = None,
-) -> None: ...
+) -> str | None: ...
```
benhoyt, line 32: I think this is a bug in Juju, or at least a wart. I don't think we should return stdout from hookcmds, but detect the error (from stdout) and raise an exception here. We should also open a bug on Juju to return a non-zero exit code for this case.

ops/model.py
```diff
@@ -719,7 +719,11 @@ def add_secret(
         )
 
     def open_port(
-        self, protocol: typing.Literal['tcp', 'udp', 'icmp'], port: int | None = None
+        self,
+        protocol: typing.Literal['tcp', 'udp', 'icmp'],
+        port: int | tuple[int, int | None] | None = None,
+        *,
+        endpoints: Sequence[str] = '*',
```
benhoyt, line 726: I think it'd be better / more Pythonic for this input parameter, so `endpoints: Sequence[str] | None = None`.

benhoyt:
Closing: https://github.com/canonical/operator/issues/2035#issuecomment-4449792576

PR #2367, opened 2026-03-06
feat: ops.testing autoload support for charmcraft extensions

### LINKED ISSUE #1699: Support charmcraft extensions in Scenario auto-load
When Scenario autoloads the metadata, it does so via charmcraft.yaml (or the older metadata.yaml, actions.yaml, config.yaml). However, if an extension is being used (such as for the 12-factor apps) then there is additional metadata provided by the extension (for example, see the [code in charmcraft](https://github.com/canonical/charmcraft/blob/1a19ec0b50db488c83ac62b6f5d1f4160fd7c894/charmcraft/extensions/app.py#L231)).

For the moment, if you're writing tests for these, then you need to explicitly provide the metadata rather than use autoload (for example, you could copy it from the packed charm).

12-factor apps are probably less likely to have additional unit tests, but it would be nice if we did make this easier to do.

It seems likely that we'd need to either run charmcraft to get the generated metadata, or import charmcraft to figure it out. I don't think we would want to try to manually maintain a copy of what metadata each extension adds.

tonyandrewmeyer:
When using a `charmcraft` extension (such as the 12-factor ones), the charm's charmcraft.yaml` has metadata, config, and actions added before being written out to the individual YAML files in the packed charm. Scenario doesn't understand this, so tests fail to make the right events available, and so on.

The PR adds a script that extracts the changes that each extension makes by running the `charmcraft` CLI tool, and saves that as a series of dictionaries in a private Scenario module. When Scenario autoloads, if there is an extension specified then it starts from the extension data and merges in the local data, simulating what charmcraft does when packing.

Changes to the extensions are infrequent, so we can start by manually regenerating the module when needed. If necessary, we could add a workflow that would automate this in the future.

Fixes #1699

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1999,7 +1999,7 @@ class _CharmSpec(Generic[CharmType]):
         return meta, config, actions
 
     @staticmethod
-    def autoload(charm_type: type[CharmBase]) -> _CharmSpec[CharmType]:
+    def autoload(charm_type: type[CharmBase]) -> _CharmSpec[CharmBase]:
         """Construct a ``_CharmSpec`` object by looking up the metadata from the charm's repo root.
 
         Will attempt to load the metadata off the ``charmcraft.yaml`` file
--- a/testing/tests/test_charm_spec_autoload.py
+++ b/testing/tests/test_charm_spec_autoload.py
@@ -30,7 +30,7 @@ def import_name(name: str, source: Path) -> Iterator[type[CharmBase]]:
     pkg_path = str(source.parent)
     sys.path.append(pkg_path)
     charm = importlib.import_module('mycharm')
-    obj = getattr(charm, name)
+    obj: type[CharmBase] = getattr(charm, name)
     sys.path.remove(pkg_path)
     yield obj
     del sys.modules['mycharm']
@@ -174,7 +174,7 @@ def test_config_defaults(tmp_path: Path, legacy: bool):
 class TestExtensions:
     """Tests for charmcraft extension autoloading."""
 
-    def test_extension_adds_metadata(self, tmp_path):
+    def test_extension_adds_metadata(self, tmp_path: Path):
         """An extension injects its metadata (containers, requires, etc.)."""
         with create_tempcharm(
             tmp_path,
@@ -194,7 +194,7 @@ class TestExtensions:
             assert 'flask-app-image' in spec.meta.get('resources', {})
             assert 'k8s-api' in spec.meta.get('assumes', [])
 
-    def test_extension_adds_config(self, tmp_path):
+    def test_extension_adds_config(self, tmp_path: Path):
         """An extension injects its config options."""
         with create_tempcharm(
             tmp_path,
@@ -212,7 +212,7 @@ class TestExtensions:
             assert 'flask-debug' in options
             assert 'webserver-workers' in options
 
-    def test_extension_adds_actions(self, tmp_path):
+    def test_extension_adds_actions(self, tmp_path: Path):
         """An extension injects its actions."""
         with create_tempcharm(
             tmp_path,
@@ -228,7 +228,7 @@ class TestExtensions:
             assert spec.actions is not None
             assert 'rotate-secret-key' in spec.actions
 
-    def test_local_meta_overlaps_extension_errors(self, tmp_path):
+    def test_local_meta_overlaps_extension_errors(self, tmp_path: Path):
         """Overlapping metadata keys with extension cause an error."""
         with create_tempcharm(
             tmp_path,
@@ -246,7 +246,7 @@ class TestExtensions:
             with pytest.raises(ValueError, match=r'overlapping keys.*requires.*flask-framework'):
                 _CharmSpec.autoload(charm)
 
-    def test_local_meta_no_overlap_with_extension(self, tmp_path):
+    def test_local_meta_no_overlap_with_extension(self, tmp_path: Path):
         """Non-overlapping local metadata keys merge with extension."""
         with create_tempcharm(
             tmp_path,
@@ -268,7 +268,7 @@ class TestExtensions:
             assert 'ingress' in spec.meta['requires']
             assert 'logging' in spec.meta['requires']
 
-    def test_local_config_overlaps_extension_errors(self, tmp_path):
+    def test_local_config_overlaps_extension_errors(self, tmp_path: Path):
         """Overlapping config options with extension cause an error."""
         with create_tempcharm(
             tmp_path,
@@ -291,7 +291,7 @@ class TestExtensions:
             ):
                 _CharmSpec.autoload(charm)
 
-    def test_local_config_no_overlap_with_extension(self, tmp_path):
+    def test_local_config_no_overlap_with_extension(self, tmp_path: Path):
         """Non-overlapping local config options merge with extension."""
         with create_tempcharm(
             tmp_path,
@@ -309,6 +309,7 @@ class TestExtensions:
             },
         ) as charm:
             spec = _CharmSpec.autoload(charm)
+            assert spec.config is not None
             options = spec.config['options']
             # Local-only option is present.
             assert 'my-custom-option' in options
@@ -316,7 +317,7 @@ class TestExtensions:
             assert 'flask-debug' in options
             assert 'webserver-workers' in options
 
-    def test_local_actions_overlap_extension_errors(self, tmp_path):
+    def test_local_actions_overlap_extension_errors(self, tmp_path: Path):
         """Overlapping actions with extension cause an error."""
         with create_tempcharm(
             tmp_path,
@@ -335,7 +336,7 @@ class TestExtensions:
             with pytest.raises(ValueError, match=r'overlapping keys.*actions.*flask-framework'):
                 _CharmSpec.autoload(charm)
 
-    def test_local_actions_no_overlap_with_extension(self, tmp_path):
+    def test_local_actions_no_overlap_with_extension(self, tmp_path: Path):
         """Non-overlapping local actions merge with extension."""
         with create_tempcharm(
             tmp_path,
@@ -351,12 +352,13 @@ class TestExtensions:
             },
         ) as charm:
             spec = _CharmSpec.autoload(charm)
+            assert spec.actions is not None
             # Local-only action is present.
             assert 'my-action' in spec.actions
             # Extension action is still present.
             assert 'rotate-secret-key' in spec.actions
 
-    def test_local_assumes_merged_with_extension(self, tmp_path):
+    def test_local_assumes_merged_with_extension(self, tmp_path: Path):
         """Local assumes list is merged with extension assumes."""
         with create_tempcharm(
             tmp_path,
@@ -374,7 +376,7 @@ class TestExtensions:
             assert 'k8s-api' in assumes
             assert 'juju >= 3.1' in assumes
 
-    def test_django_extension_has_create_superuser(self, tmp_path):
+    def test_django_extension_has_create_superuser(self, tmp_path: Path):
         """Django extension adds the create-superuser action."""
         with create_tempcharm(
             tmp_path,
@@ -387,10 +389,11 @@ class TestExtensions:
             },
         ) as charm:
             spec = _CharmSpec.autoload(charm)
+            assert spec.actions is not None
             assert 'create-superuser' in spec.actions
             assert 'rotate-secret-key' in spec.actions
 
-    def test_unknown_extension_warns(self, tmp_path):
+    def test_unknown_extension_warns(self, tmp_path: Path):
         """An unknown extension name emits a warning and is skipped."""
         with create_tempcharm(
             tmp_path,
@@ -407,7 +410,7 @@ class TestExtensions:
             # No extension data merged, but the charm still loads.
             assert spec.meta['name'] == 'my-app'
 
-    def test_extension_stripped_from_meta(self, tmp_path):
+    def test_extension_stripped_from_meta(self, tmp_path: Path):
         """The 'extensions' key should not remain in the loaded meta."""
         with create_tempcharm(
             tmp_path,
@@ -422,7 +425,7 @@ class TestExtensions:
             spec = _CharmSpec.autoload(charm)
             assert 'extensions' not in spec.meta
 
-    def test_extension_with_relations_in_context(self, tmp_path):
+    def test_extension_with_relations_in_context(self, tmp_path: Path):
         """Relations from an extension can be used in a Context run."""
         with create_tempcharm(
             tmp_path,


tonyandrewmeyer:
Examples of using this:
* [cs.canonical.com](https://github.com/tonyandrewmeyer/cs.canonical.com/pull/1)
* [forms.canonical.com](https://github.com/tonyandrewmeyer/forms.canonical.com/pull/1)
* [university.canonical.com](https://github.com/tonyandrewmeyer/university.canonical.com/pull/1)
* [support-portal](https://github.com/tonyandrewmeyer/support-portal/pull/1)
* [dashboard](https://github.com/tonyandrewmeyer/dashboard/pull/1)
* [github-runner-webhook-router](https://github.com/tonyandrewmeyer/github-runner-webhook-router/pull/1) 
* [httprequest-lego-provider](https://github.com/tonyandrewmeyer/httprequest-lego-provider/pull/1) (includes an example of using `State.from_context` with this new functionality)
* [documentationacademy.org](https://github.com/tonyandrewmeyer/documentationacademy.org/pull/1) (this charm uses `xiilib` rather than paas-charm (the domain also errors out, so maybe it's just entirely dead and someone should archive the repo), so I've also migrated it to pass-charm to make it simpler -- *but* then it's just another "paas-charm has all the functionality" example, so it's probably not worth looking at this or taking this PR any further)

Those are roughly in order of complexity, so if you don't want to read them all, choose as you please.

tonyandrewmeyer:
@james-garner-canonical , @dimaqq please see #1699 for some thoughts on how the issue could be solved. If you think that the one in this PR isn't the right one, let's have a team chat at daily about what we want to do. I feel like this is the right approach for the moment, and it's not locking us in too much (we could definitely import from charmcraft or some vendored charmcraft code without changing the API, we could definitely automate getting the data without changing the API, if we added a "run charmcraft expand-extensions" system we'd probably want to keep what's here as a fallback, etc).

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.10"
```
dimaqq: I suppose shebang is redundant: plain Python doesn't grok pep-723.

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+    'charm-libs',
+}
+
+OUTPUT_FILE = (
+    pathlib.Path(__file__).resolve().parent.parent
+    / 'testing'
+    / 'src'
+    / 'scenario'
+    / '_charmcraft_extensions.py'
+)
```
dimaqq, line 62: That's...  creative... is that the formatting style preferred by Ruff?

testing/src/scenario/_charmcraft_extensions.py
```diff
@@ -0,0 +1,417 @@
...
+        },
+        'resources': {
+            'app-image': {'description': 'spring-boot application image.', 'type': 'oci-image'}
+        },
+    },
+}
+
+# Config options added by each charmcraft extension.
+CONFIG: dict[str, dict[str, Any]] = {
+    'django-framework': {
```
dimaqq, line 123: That's not how I would do it, but maybe it's just fine.

I'd make something like:
```py
@dataclass
class Extension:
    meta: dict[str, Any]  # or a TypeDict?
    config: ...
    actions: ... | None

Django = Extension(..., ..., ...)
Flask = Extension(..., ..., ...)

bleh = {"django": Django, "flask": Flask, ...}
```

totally not a blocking comment!

.github/generate_charmcraft_extensions.py
james-garner-canonical: Since we use `uv` in this project, how about this?
```suggestion
#!/usr/bin/env -S uv run --script --no-project
# /// script
# requires-python = ">=3.10"
```
And I guess the file should be executable if we're including a shebang.

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+        ['tox', '-e', 'format', '--', str(OUTPUT_FILE)],
+        cwd=OUTPUT_FILE.parent,
+        check=True,
+    )
+
+    return 0
+
+
+if __name__ == '__main__':
+    raise SystemExit(main())
```
james-garner-canonical: TIL. This seems more idiomatic to me:

```suggestion
    sys.exit(main())
```

Not a blocker.

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def run_charmcraft(profile: str, workdir: pathlib.Path) -> dict[str, Any]:
+    """Run charmcraft init + expand-extensions and return the expanded YAML as a dict."""
+    subprocess.run(
+        ['charmcraft', 'init', '--profile', profile, '--name', 'test-charm'],
+        cwd=workdir,
+        check=True,
+        capture_output=True,
+        text=True,
```
james-garner-canonical: How about using `check_call`?
```suggestion
    subprocess.check_call(
        ['charmcraft', 'init', '--profile', profile, '--name', 'test-charm'], cwd=workdir
```

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+    result = subprocess.run(
+        ['charmcraft', 'expand-extensions'],
+        cwd=workdir,
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    return yaml.safe_load(result.stdout)
```
james-garner-canonical: I think `check_output` would be good here, it takes care of `check=True` and `capture_output=True`.
```suggestion
    result = subprocess.check_output(['charmcraft', 'expand-extensions'], cwd=workdir, text=True)
    return yaml.safe_load(result)
```

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def extract_extension_data(
+    expanded: dict[str, Any],
+) -> tuple[dict[str, Any], dict[str, Any], dict[str, Any]]:
+    """Extract metadata, config options, and actions from expanded charmcraft YAML."""
+    metadata: dict[str, Any] = {}
+    for key in METADATA_KEYS:
+        if key in expanded:
+            metadata[key] = expanded[key]
```
james-garner-canonical: IMO a dict comprehension has less mental overhead for readers (for me at least!), and as an added bonus it does away with the need for a type annotation.
```suggestion
    metadata = {key: expanded[key] for key in METADATA_KEYS if key in expanded}
```

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+
+    config: dict[str, Any] = {}
+    if 'config' in expanded and 'options' in expanded['config']:
+        config = expanded['config']['options']
+
```
james-garner-canonical: Similarly:
```suggestion
    config = expanded.get('config', {}).get('options', {})
```

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+def format_dict(d: dict[str, Any], indent: int = 4) -> str:
+    """Format a dict as a Python literal string."""
+    formatted = pprint.pformat(d, width=100, sort_dicts=True)
+    if '\n' in formatted:
+        lines = formatted.splitlines()
+        result = lines[0]
+        for line in lines[1:]:
+            result += '\n' + ' ' * indent + line
+        return result
+    return formatted
```
james-garner-canonical: Since we run `tox -e format` at the end, can we drop this function? See also further comments below on the file generation loop.

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+            all_data[profile] = extract_extension_data(expanded)
+
+    module_source = generate_module(all_data)
+
+    if OUTPUT_FILE.exists():
+        response = input(f'{OUTPUT_FILE} already exists. Overwrite? [y/N] ').strip().lower()
+        if response != 'y':
+            print('Aborted.')
+            return 1
+
```
james-garner-canonical: Since `OUTPUT_FILE` will be under version control and we may want to run this step automatically, how about we drop this?

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+    ]
+
+    # Build the three top-level dicts.
+    metadata_entries: dict[str, dict[str, Any]] = {}
+    config_entries: dict[str, dict[str, Any]] = {}
+    action_entries: dict[str, dict[str, Any]] = {}
+    for profile, (metadata, config, actions) in sorted(all_data.items()):
+        metadata_entries[profile] = metadata
+        config_entries[profile] = config
+        action_entries[profile] = actions
```
james-garner-canonical: If we pre-sort here, then if we drop formatting logic from the loop that build the lines, the loop can be very simple -- see comment below.
```suggestion
        metadata_entries[profile] = {k: metadata[k] for k in sorted(metadata)}
        config_entries[profile] = {k: config[k] for k in sorted(config)}
        action_entries[profile] = {k: actions[k] for k in sorted(actions)}
```
Or perhaps `dict(sorted(metadata.items()))` etc

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+        lines.append(f'# {docstring}')
+        lines.append(f'{var_name}: dict[str, dict[str, Any]] = {{')
+        for profile in sorted(data):
+            lines.append(f'    {profile!r}: {{')
+            for key in sorted(data[profile]):
+                val = data[profile][key]
+                formatted = format_dict(val, indent=8)
+                lines.append(f'        {key!r}: {formatted},')
+            lines.append('    },')
+        lines.append('}')
```
james-garner-canonical: Here I wonder if we can dump the string representation of the dictionary without worrying about formatting at all, since we run `tox -e format` at the end.

We sort by profile when we build the three top-level dicts, so we don't need to call sorted(data), and I suggest (above) that we could sort the dicts there too.

Then we could:
```suggestion
        lines.append(f'{var_name}: dict[str, dict[str, Any]] = {data}')
```

docs/explanation/state-transition-testing.md
```diff
@@ -231,6 +231,31 @@ ctx = testing.Context(
...
+#     - flask-framework
+
+ctx = testing.Context(MyFlaskCharm)
+# The 'ingress' relation is provided by the flask-framework extension.
+state = ctx.run(ctx.on.start(), testing.State(relations={testing.Relation('ingress')}))
+```
+
+Local values in `charmcraft.yaml` take precedence over extension defaults. For
+example, if you define a custom `ingress` relation in your `charmcraft.yaml`, it
+will override the one provided by the extension.
```
james-garner-canonical: I guess this is true on Charmcraft's side too? If so, worth making that explicit here?

testing/src/scenario/_charmcraft_extensions.py
james-garner-canonical: I guess there are two comments here:
- Should the top-level organisation be
    - `extension-name` -> `meta|config|actions` -> `stuff` (proposed)
    - `meta|config|actions` -> `extension-name` -> `stuff` (current)
- Should the top-level objects be classes rather than dictionaries

I think top-level classes are easier on the eyes, but (1) this would make the code generation logic slightly more complicated; (2) if the top-level classes are `Django` etc, we'd have to call `getattr(this_module, extension_name)` which seems ugly; while if the top-level classes are `Meta` etc. we call `getattr(this_module.Meta, extension_name)`, which doesn't seem great either. Oh and we'd need to map from the actual extension name (e.g.) `django-framework` to permitted Python identifiers, either with a rule like `name.split('-')[0].capitalize()`, or perhaps with a mapping somewhere.

Sticking with dictionaries, but having the top-level names be the extension names rather than the three known names `METADATA`, `CONFIG`, and `ACTIONS` leads to the same pattern of `getattr` on the module with a mapping from valid extension names to valid Python identifiers.

Having thought that through by typing it up, I think the current structure is the right choice both for ease of generation and ease of use by Scenario.

testing/src/scenario/state.py
```diff
@@ -1860,6 +1862,63 @@ def _is_valid_charmcraft_25_metadata(meta: dict[str, Any]):
     return True
 
 
+def _apply_extensions(
+    meta: dict[str, Any],
+    extensions: list[str],
+) -> dict[str, Any]:
```
```suggestion
def _apply_extensions(meta: dict[str, Any], extensions: list[str]) -> dict[str, Any]:
```

testing/src/scenario/state.py
```diff
@@ -1860,6 +1862,63 @@ def _is_valid_charmcraft_25_metadata(meta: dict[str, Any]):
...
+    """Merge charmcraft extension defaults into the charm metadata.
```
james-garner-canonical: Let's either mutate `meta` in place, *or* return a new dict with the extensions applied, but not both.

I don't really mind which in principle, since the dictionary we're working with is a private copy that we loaded from disc, so it doesn't affect anyone else. I assume mutation in place is slightly more performant, but I think we should prefer whatever is more readable and maintainable. My personal bias is towards not mutating in place.

testing/src/scenario/state.py
```diff
@@ -1860,6 +1862,63 @@ def _is_valid_charmcraft_25_metadata(meta: dict[str, Any]):
...
+    """
+    for ext_name in extensions:
+        ext_meta = _charmcraft_extensions.METADATA.get(ext_name, {})
+        ext_config = _charmcraft_extensions.CONFIG.get(ext_name, {})
+        ext_actions = _charmcraft_extensions.ACTIONS.get(ext_name, {})
+
+        if not ext_meta and not ext_config and not ext_actions:
+            warnings.warn(
+                f'Unknown charmcraft extension {ext_name!r}; '
+                f'ignoring. You may need to update to a newer ops.',
```
```suggestion
                f'Unknown charmcraft extension {ext_name!r}; ignoring.'
                "Ensure you're running the latest ops, and open an issue if this persists.",
```
Maybe we can give a more opinionated and direct call to action here?

testing/src/scenario/state.py
```diff
@@ -1860,6 +1862,63 @@ def _is_valid_charmcraft_25_metadata(meta: dict[str, Any]):
...
+                stacklevel=2,
+            )
+            continue
+
+        # Merge metadata: for dicts, extension provides defaults that
+        # the local yaml overrides. For lists, combine them.
+        for key, ext_value in ext_meta.items():
+            if key not in meta:
+                meta[key] = copy.deepcopy(ext_value)
+            elif isinstance(ext_value, dict) and isinstance(meta[key], dict):
```
james-garner-canonical, line 1895: Should a type mismatch be an error, or at least a warning? What does Charmcraft do?

testing/src/scenario/state.py
```diff
@@ -1860,6 +1862,63 @@ def _is_valid_charmcraft_25_metadata(meta: dict[str, Any]):
...
+                merged = copy.deepcopy(ext_value)
+                merged.update(meta[key])
+                meta[key] = merged
+            elif isinstance(ext_value, list) and isinstance(meta[key], list):
+                merged = copy.deepcopy(ext_value)
+                for item in meta[key]:
+                    if item not in merged:
+                        merged.append(item)
```
james-garner-canonical: Personal taste, but I like how it makes this branch more obviously follow the same pattern as the `dict` branch.
```suggestion
                merged.extend(i for i in meta[key] if i not in merged)
```

testing/src/scenario/state.py
```diff
@@ -1892,13 +1951,25 @@ def _load_metadata_legacy(charm_root: pathlib.Path):
...
             yaml.safe_load(metadata_path.open()) if metadata_path.exists() else {}
         )
         if not _is_valid_charmcraft_25_metadata(meta):
             meta = {}
+
+        # Apply charmcraft extensions before extracting config/actions.
+        extensions = meta.pop('extensions', None)
+        if extensions:
+            meta = _apply_extensions(meta, extensions)
+
```
james-garner-canonical, line 1996: I was wondering if we should support separate `metadata`, `config`, and `actions` files when applying extensions. I know we call that path `_load_metadata_legacy`, but I believe separate  files are still valid. I assume they're not common with 12-factor extensions though.

I think Scenario is actually wrong (doesn't match Charmcraft) currently in its handling of `charmcraft.yaml` vs separate files. It's valid in Charmcraft to have both `charmcraft.yaml` and `metadata.yaml`, `config.yaml`, etc. We ignore the separate files if there's a `charmcraft.yaml` file with metadata in it.

Given our current approach, I think we're applying the extensions in the right place. But I wonder if we should consider supporting the same combinations that Charmcraft does, so that state-transition tests better match the real behaviour. Do you think it's worth opening an issue to track this?

testing/src/scenario/state.py
```diff
@@ -1892,13 +1951,25 @@ def _load_metadata_legacy(charm_root: pathlib.Path):
...
+        """
         metadata_path = charm_root / 'charmcraft.yaml'
         meta: dict[str, Any] = (
             yaml.safe_load(metadata_path.open()) if metadata_path.exists() else {}
         )
         if not _is_valid_charmcraft_25_metadata(meta):
             meta = {}
+
+        # Apply charmcraft extensions before extracting config/actions.
+        extensions = meta.pop('extensions', None)
```
james-garner-canonical, line 1993: Removing the `'extensions'` field from metadata is a behaviour change, but since autoloaded metadata is only visible to users via the deprecated `charm_spec` attribute, it should be fine.

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,218 @@
...
+    for var_name, data, docstring in [
+        (
+            'METADATA',
+            metadata_entries,
+            'Metadata added by each charmcraft extension.',
+        ),
+        ('CONFIG', config_entries, 'Config options added by each charmcraft extension.'),
+        ('ACTIONS', action_entries, 'Actions added by each charmcraft extension.'),
+    ]:
+        lines.append(f'# {docstring}')
```
james-garner-canonical: Very minor but I think it improves readability:
```suggestion
    for var_name, data, doc_name in [
        ('METADATA', metadata_entries, 'Metadata'),
        ('CONFIG', config_entries, 'Config options'),
        ('ACTIONS', action_entries, 'Actions'),
    ]:
        lines.append(f'# {doc_name} added by each charmcraft extension.')
```

james-garner-canonical (changes requested):
I think this is a nice approach to supporting Charmcraft extensions, and I agree with your decision to defer updating automatically until we need to.

I'm marking my review as 'request changes' since I'd like these specific concerns to be addressed before merging:
1. On the code generation side, I think we should drop all the manual formatting code since we run `tox -e format` at the end.
2. On the Scenario side, `_apply_extensions` should either mutate in place, or return a new dict, but not both.
3. Should it be an error (or at least a logging error or warning) if the extension defines a field and the charm overrides it with a different type (dict vs list)? How does Charmcraft handle this?

I've made a number of other comments and suggestions as I went, just putting into writing my thoughts as I reviewed. I think they'd improve readability going forward, but if you disagree then don't consider them blocking.

tonyandrewmeyer:
> 3. Should it be an error (or at least a logging error or warning) if the extension defines a field and the charm overrides it with a different type (dict vs list)? How does Charmcraft handle this?

Charmcraft does have some [validation](https://github.com/canonical/charmcraft/blob/e66a25b5e80b6a83603a823615e00d0bcc98a0b6/charmcraft/extensions/app.py#L124) that prevents some clashes. I'm not particularly keen on copying that into Scenario, and I think people will run into it at packing time, so it shouldn't really be an issue.

If you do something simple like redefine an config option then you get an error (not just in pack, but also e.g. in clean) that says that there are overlapping keys and you need to remove or remove that key. I think it'd be reasonable to make this an error in Scenario as well, even though you're going to quickly hit it elsewhere anyway.

.github/generate_charmcraft_extensions.py
tonyandrewmeyer: Generally, yes, it wants the operator (the stupid pathlib `/`) to be the first thing on the continuation line. I could collapse it more if you like, but it seemed reasonably readable to me.

james-garner-canonical: I like it as-is, but if you want to avoid the pathlib '/' then perhaps consider using `Path.joinpath` or smooshing the parts into one slash-delimited string (I assume ruff doesn't object to this?) (or both).

```suggestion
OUTPUT_FILE = pathlib.Path(__file__).resolve().parent.parent.joinpath(
    'testing', 'src', 'scenario', '_charmcraft_extensions.py'
)
```

```suggestion
OUTPUT_FILE = (
    pathlib.Path(__file__).resolve().parent.parent
    / 'testing/src/scenario/_charmcraft_extensions.py'
)
```

tonyandrewmeyer: Thanks, but I prefer it as-is as well.

james-garner-canonical:
> > 3. Should it be an error (or at least a logging error or warning) if the extension defines a field and the charm overrides it with a different type (dict vs list)? How does Charmcraft handle this?
> 
> Charmcraft does have some [validation](https://github.com/canonical/charmcraft/blob/e66a25b5e80b6a83603a823615e00d0bcc98a0b6/charmcraft/extensions/app.py#L124) that prevents some clashes. I'm not particularly keen on copying that into Scenario, and I think people will run into it at packing time, so it shouldn't really be an issue.
> 
> If you do something simple like redefine an config option then you get an error (not just in pack, but also e.g. in clean) that says that there are overlapping keys and you need to remove or remove that key. I think it'd be reasonable to make this an error in Scenario as well, even though you're going to quickly hit it elsewhere anyway.

Agreed on not copying that logic. I'm happy with documenting that Scenario won't catch metadata that Charmcraft will consider invalid and/or adding a simple error case -- e.g. should the final `else` branch when we combine the metadata be an error?

docs/explanation/state-transition-testing.md
tonyandrewmeyer: It's not really true on the charmcraft side. We should probably make it closer to charmcraft (as per the other comment) and then this doc should be updated too.

dimaqq:
```
.github/generate_charmcraft_extensions.py:205:5: F821 Undefined name `sys`
    |
204 | if __name__ == '__main__':
205 |     sys.exit(main())
    |     ^^^ F821
    |

Found 1 error.
```

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,205 @@
...
+        ['tox', '-e', 'format', '--', str(OUTPUT_FILE)],
+        cwd=OUTPUT_FILE.parent,
+        check=True,
+    )
+
+    return 0
+
+
+if __name__ == '__main__':
+    sys.exit(main())
```
dimaqq, line 180: I thought there was something fishy here, indeed, `sys` is not imported.

.github/generate_charmcraft_extensions.py
```diff
@@ -0,0 +1,205 @@
...
+def get_extensions() -> list[str]:
+    """Get the list of available charmcraft extensions via ``charmcraft list-extensions``."""
+    result = subprocess.run(
+        ['charmcraft', 'list-extensions'],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    extensions = []
+    for line in result.stdout.splitlines():
```
james-garner-canonical: I missed a `subprocess.run` on my first scan through, sorry.
```suggestion
    result = subprocess.check_output(['charmcraft', 'list-extensions'], text=True)
    extensions = []
    for line in result.splitlines():
```

james-garner-canonical: But also if we're wanting to suppress/capture both stdout and stderr, maybe what you had originally is best.

testing/src/scenario/state.py
tonyandrewmeyer: I'd also argue it's a bug if the charm is doing this, since it doesn't end up in the packed files. I suppose the tests might be using it (to do something exactly like this PR, perhaps) but I agree that it should be fine, and it doesn't show up in any examples of 12-factor charms I can find.

testing/src/scenario/state.py
tonyandrewmeyer: I suppose people might manually add an extension to charmcraft.yaml, but realistically I think people are using the init command with a profile, and that produces the single file. I don't think we want to encourage going backwards by supporting the separate files here.

For the more general question, we could perhaps have an issue, although my preferred solution would be for Scenario to error and so we more strongly push charmers to unify into the single file.

testing/src/scenario/state.py
tonyandrewmeyer: Is this relevant now that we're erroring on name conflicts?

 1. If key not in meta → use extension value (no conflict)                                                                                                                    
 2. If both are dicts → error on overlapping sub-keys, then merge
 3. If both are lists → combine them                                                                                                                                          
 4. (implicit) Otherwise → do nothing 

For case 4 to happen, the extension and the charm would need the same top-level metadata key (like assumes, containers, etc.) but with different types - e.g. the extension provides assumes as a list but the charm has assumes as a dict. I believe in practice this can't happen with valid charmcraft.yaml.

testing/src/scenario/_charmcraft_extensions.py
tonyandrewmeyer: I don't think we benefit from generating dataclasses here. The use case for this data is to produce a dictionary, and they are autogenerated, so it seems like it's extra work that never gets used to put a dataclass in the middle.

I also prefer a dict with a constant name and the extensions as keys for the reasons James outlines.

testing/src/scenario/state.py
james-garner-canonical: I wouldn't mind seeing a ValueError for the final else branch saying basically that. But if you'd prefer it to silently skip in that case, I assume it won't make much difference in practice.

PR #2368, opened 2026-03-09
docs: replace links to juju.is by canonical.com/juju

dwilding:
[juju.is](https://juju.is) now redirects to [canonical.com/juju](https://canonical.com/juju). No links are broken - this PR updates the links to keep things tidy.

I'm adjusting the documentation header to point to [canonical.com/juju/docs](https://canonical.com/juju/docs) instead of canonical.com/juju, matching the Juju RTD docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ![CI Status](https://github.com/canonical/operator/actions/workflows/framework-tests.yaml/badge.svg)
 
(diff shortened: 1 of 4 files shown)


dwilding:
> Do we have some automation to warn us about linking to 301/308s?

We don't. The docs link checker (`make -C docs linkcheck`) reports redirections, so in principle we could have automation that checks the output for redirections. But we haven't enabled the link checker in Ops CI because it's quite slow and flaky due to rate-limiting by external sites.

PR #2370, opened 2026-03-10
docs: refactor homepage to better put Ops in context

dwilding:
The goal of this PR is to make the Ops docs homepage more useful as a general place for working charm developers to locate resources in Ops and other ecosystem docs.

My core assumption is that the reader already has a general understanding of where a charm fits in the Juju architecture and what they want to achieve with their charm. Supporting readers who lack this understanding is better handled by the [ecosystem docs](https://canonical.com/juju/docs), in my opinion.

**[Preview of new homepage](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/)**

Main improvements:

- Added a "Get started" section that presents the Ops tutorials and the 12-factor (Charmcraft) tutorials as equally important paths. (I personally think that every working charmer should have completed at least one Ops tutorial and one 12-factor tutorial)

- Reworked the "In this documentation" section so that it's organised according to the activities that a charmer is likely to care about. I intend for us to expand this slightly in the future, as we add more general content. Eg as we add more on CI, libraries, and publishing.

- Replaced the "Related links" list by a "Related documentation" section that briefly explains the purpose of each ecosystem component. In a few cases, I've provided links to frequently-used external pages. My intention here isn't to be comprehensive or link to a representative page - the pages I'm linking to are known to be traffic outliers, so these links are a special convenience for charmers.

### DIFF
--- a/docs/_static/project_specific.css
+++ b/docs/_static/project_specific.css
@@ -17,8 +17,25 @@ table.docutils.top-aligned tbody > tr:first-child > td {
     border-top: 1px solid var(--color-table-border);
 }
 
-/* Define icon positioning, for instance for an external link icon */
-svg.icon-inline {
-    display: inline-block;
-    vertical-align: middle;
+/* Show a box-with-arrow icon after links */
+a.show-external::after {
+    content: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M9.157 3l-1.5 1.5H6a.5.5 0 00-.492.41L5.5 5v5a.5.5 0 00.41.492L6 10.5h5a.5.5 0 00.492-.41L11.5 10V8.538l1.5-1.5V10a2 2 0 01-2 2H6a2 2 0 01-2-2V5a2 2 0 012-2h3.157zm5.593-1.75V6h-1.5V3.81L8.5 8.56 7.44 7.5l4.748-4.75H10v-1.5h4.75z' fill='%23333' fill-rule='evenodd'/%3E%3C/svg%3E");
+    margin-left: 0.25rem;
+    vertical-align: text-top;
+}
+
+/* Make sure that our customisations support the dark theme */
+@media not print {
+    body[data-theme="dark"] {
+        a.show-external::after {
+            content: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M9.157 3l-1.5 1.5H6a.5.5 0 00-.492.41L5.5 5v5a.5.5 0 00.41.492L6 10.5h5a.5.5 0 00.492-.41L11.5 10V8.538l1.5-1.5V10a2 2 0 01-2 2H6a2 2 0 01-2-2V5a2 2 0 012-2h3.157zm5.593-1.75V6h-1.5V3.81L8.5 8.56 7.44 7.5l4.748-4.75H10v-1.5h4.75z' fill='%23CDCDCD' fill-rule='evenodd'/%3E%3C/svg%3E");
+        }
+    }
+    @media (prefers-color-scheme: dark) {
+        body[data-theme="auto"] {
+            a.show-external::after {
+                content: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M9.157 3l-1.5 1.5H6a.5.5 0 00-.492.41L5.5 5v5a.5.5 0 00.41.492L6 10.5h5a.5.5 0 00.492-.41L11.5 10V8.538l1.5-1.5V10a2 2 0 01-2 2H6a2 2 0 01-2-2V5a2 2 0 012-2h3.157zm5.593-1.75V6h-1.5V3.81L8.5 8.56 7.44 7.5l4.748-4.75H10v-1.5h4.75z' fill='%23CDCDCD' fill-rule='evenodd'/%3E%3C/svg%3E");
+            }
+        }
+    }
 }
--- a/docs/index.md
+++ b/docs/index.md
@@ -22,7 +22,7 @@ Ops promotes consistent and maintainable charm code. Its APIs help you separate 
 ```{grid-item-card} Generate charm code for a web app
 Use our CLI tools to turn your 12-factor app into a charm that's ready to deploy. We support Django, FastAPI, Go, and more!
 
-<a class="reference external" href="https://documentation.ubuntu.com/charmcraft/latest/tutorial/" title="(in Charmcraft latest)" target="_blank">Write your first 12-factor app charm <svg class="icon-inline" width="1em" height="1em" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="M9.157 3l-1.5 1.5H6a.5.5 0 00-.492.41L5.5 5v5a.5.5 0 00.41.492L6 10.5h5a.5.5 0 00.492-.41L11.5 10V8.538l1.5-1.42V10a2 2 0 01-2 2H6a2 2 0 01-2-2V5a2 2 0 012-2h3.157zm5.593-1.75V6h-1.5V3.81L8.5 8.56 7.44 7.5l4.748-4.75H10v-1.5h4.75z" fill="currentColor" fill-rule="evenodd"/></svg></a>
+<a class="show-external" href="https://documentation.ubuntu.com/charmcraft/latest/tutorial/" title="(in Charmcraft latest)" target="_blank">Write your first 12-factor app charm</a>
 ```
 ```{grid-item-card} Write your first charm from scratch
 For a hands-on introduction to charm development with Ops, try our tutorials:


docs/index.md
```diff
@@ -16,49 +12,70 @@ explanation/index
...
-```{grid-item-card} [Explanation](explanation/index)
-**Discussion and clarification** of key topics
-- [Testing](explanation/testing)
-- [Tracing](explanation/tracing)
-- [Security](explanation/security)
-```
+**Running workloads:** [Manage packages on machines](howto/run-workloads-with-a-charm-machines) • [Manage Kubernetes workloads](howto/manage-containers/index)
 
-````
+**Adding functionality:** [Manage relations](howto/manage-relations) • [Manage configuration](howto/manage-configuration) • [More Juju features](#how-to-guides-managing-features)
```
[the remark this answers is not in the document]
dwilding: This won't be a problem in the Sphinx build.

james-garner-canonical (approved):
This looks good to me. A very nice refresh for the landing page, and a more approachable overview for new arrivals.

One minor thing that stood out to me is this line:
> [Reference](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/reference/), including [ops](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/reference/ops/), [ops.testing](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/reference/ops-testing/), [ops.tracing](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/reference/ops-tracing/), and [ops.pebble](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/reference/pebble/)

I wonder if `pebble` should really be here -- I can see how it was grandfathered in from the original layout, where I guess we had `ops` with `testing` and `pebble` as the more self-contained modules originally, and then added `tracing` alongside `testing` since their distribution mechanisms are the most similar. But I don't think `pebble` is really a library entrypoint for charming, you'll end up there naturally via `ops.Container` methods -- or not, and stay using the higher level API.

docs/index.md
```diff
@@ -1,7 +1,3 @@
----
-relatedlinks: "[Charmcraft](https://documentation.ubuntu.com/charmcraft/stable/), [Charmlibs](https://documentation.ubuntu.com/charmlibs/), [Concierge](https://github.com/canonical/concierge), [Jubilant](https://documentation.ubuntu.com/jubilant/), [Juju](https://documentation.ubuntu.com/juju/3.6/), [Pebble](https://documentation.ubuntu.com/pebble/)"
```
james-garner-canonical, line 2: I like the related documentation section, but I wonder if the related links item in the sidebar would still be nice to have. It's a bit of a pain that it requires repeating the URLs, but it does feel like a tasteful and useful use of space to me. Not a deal breaker either way though, and I guess maybe we're moving away from the related links item generally?

docs/index.md
```diff
@@ -16,49 +12,70 @@ explanation/index
...
 
-```{grid-item-card} [How-to guides](howto/index)
-**Step-by-step guides** covering key operations and common tasks
-- [Manage charms](howto/manage-charms)
-- [Manage relations](howto/manage-relations)
-- [Manage containers](howto/manage-containers/index)
+```{grid-item-card} Charm a 12-factor app
+Alternatively, use Charmcraft to generate charm code for your web app. Charmcraft supports Django, FastAPI, Go, and more!
+
+[Try a Charmcraft tutorial](https://documentation.ubuntu.com/charmcraft/latest/tutorial/)
```
tonyandrewmeyer: It's technically correct but "a Charmcraft tutorial" feels wrong to me here. What about "a 12-factor tutorial" or something similar instead? Or "Write your first web app charm"? We don't say "Try an ops tutorial", for what it's worth.

docs/index.md
```diff
@@ -16,49 +12,70 @@ explanation/index
...
+Ops promotes consistent and maintainable charm code. Its APIs help you separate different aspects of your charm, such as managing the application's state and integrating with other charms.
 
-## In this documentation
+## Get started
 
 ````{grid} 1 1 2 2
 
-```{grid-item-card} [Tutorials](tutorial/index)
-**Start here:** hands-on introductions to Ops, guiding you through writing charms
+```{grid-item-card} Write your first charm from scratch
```
tonyandrewmeyer, line 27: Perhaps this is blasphemous considering these are the ops docs but maybe the 12-factor one should be on the left where (English) readers will see it first, since it's the simpler place to start?

docs/index.md
```diff
@@ -16,49 +12,70 @@ explanation/index
...
-- [Testing](explanation/testing)
-- [Tracing](explanation/tracing)
-- [Security](explanation/security)
-```
+**Running workloads:** [Manage packages on machines](howto/run-workloads-with-a-charm-machines) • [Manage Kubernetes workloads](howto/manage-containers/index)
 
-````
+**Adding functionality:** [Manage relations](howto/manage-relations) • [Manage configuration](howto/manage-configuration) • [More Juju features](#how-to-guides-managing-features)
+
+**Testing & CI:** [Write unit tests](howto/write-unit-tests-for-a-charm) • [Write integration tests with Jubilant](howto/write-integration-tests-for-a-charm)
```
tonyandrewmeyer: Is the "with Jubilant" here because Jubilant is outside of what you get with ops? It feels like it could be "write unit tests" and "write integration tests", but if it's because it's involving an external tool that's reasonable too.

docs/index.md
```diff
@@ -16,49 +12,70 @@ explanation/index
...
+
+**Design & best practices:** [Holistic vs delta charms](explanation/holistic-vs-delta-charms) • [Follow best practices](#follow-best-practices) • [Trace your charm](howto/trace-your-charm)
```
tonyandrewmeyer: I think linking to holistic/delta makes sense, and you do want to learn and think about that early. I wonder if the home page is a good place to introduce these terms, which (unlike most of the rest of the page) aren't something that you'd already know even with some Juju familiarity. That's the name of the page, of course, but we could also change that for consistency. Maybe there's some more generic (and short, title-sized) way of saying that you need to consider whether you're going to respond in the same way to most events or have bespoke handlers for each? We seem to be pushing reconciling heavily at the moment, maybe we should be titling the page and this link to reflect that?

docs/index.md
```diff
@@ -16,49 +12,70 @@ explanation/index
...
+## Related documentation
 
-## Releases
+- **{external+charmcraft:doc}`Charmcraft <index>`** - The CLI tool for initialising charms, packing charms, and interacting with [Charmhub](https://charmhub.io/). You'll find the {external+charmcraft:ref}`charmcraft.yaml specification <charmcraft-yaml-file>` especially helpful.
+
+- **{external+charmlibs:doc}`Charmlibs <index>`** - A listing of charm libraries and guidance on how to distribute your own libraries.
+
+- **[Concierge](https://github.com/canonical/concierge)** - A CLI tool for setting up charm development environments.
+
+- **{external+jubilant:doc}`Jubilant <index>`** - A Python library that wraps the Juju CLI. Recommended for use in integration tests.
```
tonyandrewmeyer: Should the latter part be more forceful? "Use for your integration tests" or something?

dwilding:
> There's an "expand all options" button on the page, which I see is there on the current one too. Does that do anything? It doesn't seem to. Can we remove it if not?

I've never seen "expand all options" myself, but you're not the first person to report seeing it in Canonical docs. It's coming from a bug in the `sphinx-config-options` package, which is enabled by default in the doc starter pack. I've opened https://github.com/canonical/sphinx-config-options/pull/14 to fix the bug. We could remove `sphinx-config-options` as we don't need it in Ops, but I'd rather keep our setup closer to the starter pack and wait for the fix.

tonyandrewmeyer:
> I've never seen "expand all options" myself, but you're not the first person to report seeing it in Canonical docs.

It looks like this, in case you haven't seen it from someone else:

IMG_5351

 > I'd rather keep our setup closer to the starter pack and wait for the fix.

👍

docs/index.md
dwilding: Good feedback, thanks. I've taken another pass - including your blasphemous suggestion (which I think I support too :face_with_open_eyes_and_hand_over_mouth:)

How does this look to you? https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/#get-started

docs/index.md
dwilding: Makes sense. I've changed it to "Use Jubilant for your integration tests"

docs/index.md
dwilding: I think I added "with Jubilant" in my initial draft before I'd settled on how much detail to include under "Related documentation". I agree that it's OK not to emphasise Jubilant here.

docs/index.md
dwilding: I'm quite partial to the title "Holistic vs delta charms" - I like its personality. But I see your point about how it looks on the homepage. How about on the homepage we go with "Approaches to handling events"? We wouldn't necessarily have to change the doc title to match. What do you think @dimaqq?

dwilding:
> I wonder if `pebble` should really be here

@james-garner-canonical I'm OK with removing `ops.pebble`. We do say "including" ;)

I've removed it - let's see if anyone has a different preference

docs/index.md
james-garner-canonical: FWIW I wouldn't mind something more opinionated here, like "The one correct way to handle events" (number two may surprise you ;)

Just my 2c, don't let it block anything.

dwilding:
> It's a bit odd not having the Diátaxis box

@tonyandrewmeyer Yes, this new layout is the standard that all Canonical doc sites should be adopting.

The full standard for "In this documentation" looks more like this:

> - The <ins>Tutorial</ins> takes you step-by-step through the [product specifics]
> - <ins>How-to guides</ins> assume you have basic familiarity with \<product>. [More product specifics]
> - <ins>Reference</ins> provides a guide to APIs, key classes and functions.
> - <ins>Explanation</ins> includes topic overviews, background and context and detailed discussion.

I strongly disagree with the need for the extra text. There's already enough for people to read, and between the "Get started" and "In this documentation" sections, I think we've already covered the specifics well.

(You might argue, why have the four categories linked at all, when we already have the site nav tree. It's nice listing the categories in the body text because the nav tree is hidden if the page is narrow)

---

FYI - I made one adjustment after talking with @secondskoll today. I switched "In this documentation" and "Related documentation" to tables for improved readability. Thanks for the suggestion!

docs/index.md
dwilding: On balance I'd rather not repeat the same links in the sidebar, even though they are nice shortcut. So I'll continue with having "Related documentation" only. I think this is something we can revisit when we group all doc sites under the Juju ecosystem docs.

dwilding:
I made one additional change today: It wasn't clear enough that the "Write your first 12-factor app charm" link takes you outside the Ops docs - thanks @tmihoc for flagging this! I've set the link to open in a new tab and added a box-with-arrow icon as a visual clue. [Latest preview](https://canonical-ubuntu-documentation-library--2370.com.readthedocs.build/ops/2370/)

docs/_static/project_specific.css
```diff
@@ -8,3 +8,17 @@ body {
...
+}
+table.docutils.top-aligned tbody > tr:first-child > td {
+    border-top: 1px solid var(--color-table-border);
+}
+
+/* Show a box-with-arrow icon after links */
+a.show-external::after {
+    content: url("data:image/svg+xml;charset=utf-8,%3Csvg width='12' height='12' xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' stroke-width='1.5' stroke='%23333' fill='none' stroke-linecap='round' stroke-linejoin='round'%3E%3Cpath d='M0 0h24v24H0z' stroke='none'/%3E%3Cpath d='M11 7H6a2 2 0 0 0-2 2v9a2 2 0 0 0 2 2h9a2 2 0 0 0 2-2v-5M10 14 20 4M15 4h5v5'/%3E%3C/svg%3E");
+    margin-left: 0.25rem;
+}
```
dimaqq: At first, I thought this wasn't working, but as it turned out, it's gray or black on transparent, which means invisible in dark mode.

dimaqq: P.S. WDYT about one of these Unicode symbols instead?
Or is the svg dropped in here part of our corp style guide?
↗ U+2197 north east arrow
⤴ U+2934 arrow pointing right then curving up
🔗 U+1F517 link symbol

docs/index.md
```diff
@@ -16,49 +12,78 @@ explanation/index
...
+Ops promotes consistent and maintainable charm code. Its APIs help you separate different aspects of your charm, such as managing the application's state and integrating with other charms.
 
-## In this documentation
+## Get started
 
 ````{grid} 1 1 2 2
+```{grid-item-card} Generate charm code for a web app
+Use our CLI tools to turn your 12-factor app into a charm that's ready to deploy. We support Django, FastAPI, Go, and more!
+
+<a class="reference external" href="https://documentation.ubuntu.com/charmcraft/latest/tutorial/" title="(in Charmcraft latest)" target="_blank">Write your first 12-factor app charm <svg class="icon-inline" width="1em" height="1em" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="M9.157 3l-1.5 1.5H6a.5.5 0 00-.492.41L5.5 5v5a.5.5 0 00.41.492L6 10.5h5a.5.5 0 00.492-.41L11.5 10V8.538l1.5-1.42V10a2 2 0 01-2 2H6a2 2 0 01-2-2V5a2 2 0 012-2h3.157zm5.593-1.75V6h-1.5V3.81L8.5 8.56 7.44 7.5l4.748-4.75H10v-1.5h4.75z" fill="currentColor" fill-rule="evenodd"/></svg></a>
```
dimaqq: I'm not a fan of inline SVG in markdown.
That's hardly human-readable.

If we decided do go this way, I would imagine that `a class="reference external"` may be trimmed the `external` class is no longer defined in CSS, is it?

dimaqq: Additionally... inline SVG behaves badly in HTML.
(typically it's to do with sizing, viewbox/viewport, styles applying weirdly, unexpected spacing).

If you're determined to make this inline, I would recommend to make it an `` tag with a data url (base-64 encoded svg) instead.

dwilding: Not determined to make it inline - I'm copying what the [sphinx-new-tab-link](https://github.com/ftnext/sphinx-new-tab-link) extension generates, but without applying the icon to every link. Agree on the ugliness in the .md source (although I want at least an `<a>` element to be able to open the link in a new tab).

Maybe the best thing is to switch back to applying the icon with a CSS class? Later I'm going to open a starter pack issue to ask for an official way to achieve this. In the meantime I want to wrap this up somehow :slightly_smiling_face:

benhoyt: I think let's go ahead on this PR with whatever works (inline SVG, CSS class with it in another file, whatever) and then open a starter pack issue to address.

dwilding: I've switched back to a CSS class, which I think is an overall neater solution for now. It looks like this:

Screenshot from 2026-03-17 08-16-25

(It does the right thing in dark mode)

This addresses Dima's comments, so I'm going to merge. @dimaqq we can review together next time in case you have additional comments.

PR #2371, opened 2026-03-11
chore: tidy up old Claude Code files

tonyandrewmeyer:
Remove the after-edit hook. It turns out that these can actually be detrimental because an agent (like a human) will save files after doing an edit in one place intending to edit another, and that mid-save is not really the right time for running hooks for formatting or linting. The much better practice here is to have pre-commit installed and the agent commit at appropriate times and that "keeps it honest".

Remove outdated Claude settings that don't do anything any more. Also drop the default model choice, because the tool does a much better job now of automatically determining which model to use, and switching between models as appropriate. It's also much less clear whether opus or sonnet should be used by default. Also update the allowed tools format to the latest expectation.

Removes the subagents. This is now (so many months later 😂) a very outdated way to do this, and also these kinds of specialities are that are common in software engineering and not at all specific to charming or to this repo are essentially baked into most agents (including GitHub Copilot and Claude Code) so there's no point having them here.

The README was always of questionable worth, but now that there aren't other files, it definitely is not required.

Note that I have purposely not added anything new here or modified AGENTS.md. I would prefer to keep this PR focused on removing things that we don't need and updating anything that's objectively out of date, rather than widening the scope to "let's make our agents run better". We can do more of that in other PRs 😄.


dwilding:
> The much better practice here is to have pre-commit installed

I know it's out of scope for this PR, but I'm curious - how do we force agents to install pre-commit? Is it as simple as requesting it in AGENTS.md?

tonyandrewmeyer:
> I know it's out of scope for this PR, but I'm curious - how do we force agents to install pre-commit? Is it as simple as requesting it in AGENTS.md?

If you want to truly force it, then Claude Code has hooks (like the one I'm dropping here) so you could set it up to do `precommit install` on session start. That's idempotent and fast if it's already done so should work fine.

If you want to have it happen essentially all the time, then yeah the simpler thing would be to put it in AGENTS.md in a 'forceful' sort of wording ("you must ...").

If your method of working is to create a branch and do things locally (rather than the Copilot on the web and assigning issues style) then you probably have it installed anyway (including with your gimmegit branch or worktree).

PR #2376, opened 2026-03-15
fix: return copy of relation data from Harness backend

tonyandrewmeyer:
## Summary

- `relation_get()` in the Harness backend (`ops/_private/harness.py`) returned a direct reference to the internal `_relation_data_raw` dict rather than a copy. This meant any mutation of the returned dict by charm code would silently corrupt the Harness's internal state, leading to subtle and hard-to-diagnose test failures.
- The Scenario mock backend already fixed this same issue (see `testing/src/scenario/mocking.py:257` which returns `data.copy()`), creating an inconsistency where tests behaved differently depending on which backend was used.
- This change aligns the Harness backend with the Scenario backend by returning `.copy()`, and adds a regression test to verify the copy semantics.

### DIFF
--- a/ops/_private/harness.py
+++ b/ops/_private/harness.py
@@ -2455,7 +2455,7 @@ class _TestingModelBackend:
             member_name = member_name.split('/')[0]
         if relation_id not in self._relation_data_raw:
             raise model.RelationNotFoundError()
-        return self._relation_data_raw[relation_id][member_name]
+        return self._relation_data_raw[relation_id][member_name].copy()
 
     def update_relation_data(
         self,
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -3782,6 +3782,30 @@ class TestTestingModelBackend:
         with pytest.raises(ops.RelationNotFoundError):
             backend.relation_get(1234, 'unit/0', False)
 
+    def test_relation_get_returns_copy(self, request: pytest.FixtureRequest):
+        harness = ops.testing.Harness(
+            ops.CharmBase,
+            meta="""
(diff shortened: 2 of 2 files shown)


tonyandrewmeyer:
We're generally not doing Harness fixes, but it's a pretty trivial change, so maybe worth it?

james-garner-canonical:
I'm more hesitant about this fix than fixes to Scenario. Questions:

1. Should this be a deep copy, or is our storage flat structure like `dict[str, str]`?
2. Could this break anyone's Harness tests? I think we should be reluctant to do so -- my model of legacy Harness tests is that they may break as new functionality is added to Ops, which Harness isn't properly aware of, but should otherwise be stable. I'd be satisfied with a `super-tox` run showing no breakages.

ops/_private/harness.py
```diff
@@ -2455,7 +2455,7 @@ def relation_get(self, relation_id: int, member_name: str, is_app: bool):
             member_name = member_name.split('/')[0]
         if relation_id not in self._relation_data_raw:
             raise model.RelationNotFoundError()
-        return self._relation_data_raw[relation_id][member_name]
+        return self._relation_data_raw[relation_id][member_name].copy()
```
dimaqq, line 2458: Function is not annotated, so I have to guess what the return type is.
`.copy` exists for list and dict but not e.g. None.
Should be fine following the member declaration and initialisation.

tonyandrewmeyer:
Ok, I was 50-50 on this one myself so let's leave it unchanged. I'll provide the backstory later in the week.

PR #2377, opened 2026-03-15
fix: restrict permissions on secret temporary files

tonyandrewmeyer:
## Summary

- `secret_add()` and `secret_set()` in `ops/hookcmds/_secret.py` write secret content to temporary files using `open()` with default permissions (typically 0o644 depending on umask), leaving secret values world-readable on disk until cleanup.
- Replaced `open()` with `os.open(path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)` followed by `os.fdopen()` to ensure temporary files are created with owner-only read/write permissions.
- This follows the precedent set by commit `e4b0f9d4` ("fix: reduce mode of SQLite storage file to 0o600"), which established that files containing sensitive data must be created with restrictive permissions.

### DIFF
--- a/ops/hookcmds/_secret.py
+++ b/ops/hookcmds/_secret.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import datetime
 import json
+import os
 import tempfile
 from typing import Any, Literal, overload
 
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
This is invalid, the `TemporaryDirectory` protects things.

PR #2378, opened 2026-03-15
fix: use timezone-aware datetimes in expiry calculation

tonyandrewmeyer:
## Summary

- **`_calculate_expiry()` (line 539)**: Changed `datetime.datetime.now()` to `datetime.datetime.now(tz=datetime.timezone.utc)` so that when a charm passes a `timedelta` for secret expiry, the resulting datetime is timezone-aware. Previously, the naive datetime would be serialised via `datetime_to_rfc3339` (which calls `.isoformat()`), producing a string without a timezone suffix. Juju expects RFC 3339 timestamps with timezone information.

- **`Container._build_fileinfo()` (line 3047)**: Changed `datetime.datetime.fromtimestamp(info.st_mtime)` to `datetime.datetime.fromtimestamp(info.st_mtime, tz=datetime.timezone.utc)` so that `FileInfo.last_modified` carries UTC timezone information rather than being a naive datetime in the local timezone.

Both fixes ensure that datetimes produced by these functions include explicit UTC timezone data, which is necessary for correct RFC 3339 serialisation and consistent behaviour across different host timezone configurations.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -3507,9 +3507,7 @@ class TestSecrets:
 
         calls = fake_script.calls(clear=True)
         expire_arg = calls[0][calls[0].index('--expire') + 1]
-        assert expire_arg.endswith('Z'), (
-            f'Expected UTC timezone suffix Z, got {expire_arg!r}'
-        )
+        assert expire_arg.endswith('Z'), f'Expected UTC timezone suffix Z, got {expire_arg!r}'
 
     def test_unit_add_secret_simple(self, fake_script: FakeScript, model: ops.Model):
         fake_script.write('secret-add', 'echo secret:345')
@@ -4788,8 +4786,6 @@ def test_departing_unit_data_available(fake_script: FakeScript):
     ]
     assert ['relation-get', '--format=json', '-r', '1', '-', 'db/0'] in calls
     assert ['relation-get', '--format=json', '-r', '1', '-', 'db/1'] in calls
-
-
 
 
 if __name__ == '__main__':


ops/_private/harness.py
```diff
@@ -2455,7 +2455,7 @@ def relation_get(self, relation_id: int, member_name: str, is_app: bool):
             member_name = member_name.split('/')[0]
         if relation_id not in self._relation_data_raw:
             raise model.RelationNotFoundError()
-        return self._relation_data_raw[relation_id][member_name]
+        return self._relation_data_raw[relation_id][member_name].copy()
```
tonyandrewmeyer:
```suggestion
```

This is for a different PR.

ops/model.py
```diff
@@ -3044,7 +3044,7 @@ def _build_fileinfo(path: str | Path) -> pebble.FileInfo:
             type=ftype,
             size=info.st_size,
             permissions=stat.S_IMODE(info.st_mode),
-            last_modified=datetime.datetime.fromtimestamp(info.st_mtime),
+            last_modified=datetime.datetime.fromtimestamp(info.st_mtime, tz=datetime.timezone.utc),
```
james-garner-canonical: I was initially concerned that this might be a breaking change, since `file_info.last_modified` comparisons with timezone-naive `datetime` objects would now raise an exception, but it seems that this is only used internally in the `Container.push_path` helper (and in harness). The modified time isn't used in `Container.push_path`, so this change is essentially a no-op.

ops/model.py
```diff
@@ -536,7 +536,7 @@ def _calculate_expiry(
     if isinstance(expire, datetime.datetime):
         return expire
     elif isinstance(expire, datetime.timedelta):
-        return datetime.datetime.now() + expire
+        return datetime.datetime.now(tz=datetime.timezone.utc) + expire
```
james-garner-canonical, line 539: I was also concerned that this might actually be a breaking change, but it looks like the resulting expiry times are never exposed to the user since we don't cache them, and always return the result of `secret-info-get` when the user calls `Secret.get_info`.

`_calculate_expiry` is only used in `Application.add_secret`, `Unit.add_secret`, and `Secret.set_info`, to calculate the `expire` argument for `_ModelBackend.secret_set`.

test/test_model.py
```diff
@@ -4777,5 +4777,46 @@ def test_departing_unit_data_available(fake_script: FakeScript):
...
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = pathlib.Path(tmpdir) / 'test.txt'
+            path.write_text('hello')
+            fileinfo = ops.Container._build_fileinfo(path)
+
+        assert fileinfo.last_modified.tzinfo is not None, (
+            '_build_fileinfo should return a timezone-aware last_modified'
+        )
+        assert fileinfo.last_modified.tzinfo == datetime.timezone.utc
```
james-garner-canonical: We probably don't really need a test for this, since it's internal only and doesn't affect anything as far as I can tell.

test/test_model.py
```diff
@@ -4777,5 +4777,46 @@ def test_departing_unit_data_available(fake_script: FakeScript):
...
+        assert result.tzinfo is not None, (
+            '_calculate_expiry should return a timezone-aware datetime'
+        )
+        assert result.tzinfo == datetime.timezone.utc
+
+    def test_calculate_expiry_with_datetime_returns_as_is(self):
+        """_calculate_expiry with a datetime should return it unchanged."""
+        dt = datetime.datetime(2025, 6, 1, 12, 0, 0, tzinfo=datetime.timezone.utc)
+        result = _calculate_expiry(dt)
+        assert result is dt
```
james-garner-canonical: Perhaps more useful to highlight in this test that a timezone-naive datetime is returned unchanged.

ops/model.py
tonyandrewmeyer: Fair point. And if Pebble grew the ability to set the last modified time (I'm not sure whether it would even be correct to do that or not - should it be when it was modified in the container, or in the source?) then local times would be fine.


```suggestion
            # This is unused, but a required FileInfo field.
            last_modified=datetime.datetime.fromtimestamp(info.st_mtime),
```

tonyandrewmeyer:
> Also, please validate that some old Juju, like 2.9.x honours the new timestamps. I'd assume that Juju did from the get-go, being coded in Go, but better safe than sorry.

The change is in secrets, which aren't in 2.9, so that's not something that can be validated. (The other case was internal Ops code, but has been removed as suggested elsewhere in the PR).

test/test_model.py
tonyandrewmeyer: I replaced this test with one that's more similar to our others, where it's testing at the event level.

PR #2379, opened 2026-03-15
fix: return copies from Scenario secret_get and action_get

tonyandrewmeyer:
## Summary

- **`secret_get()`** in `_MockModelBackend` returned `secret.latest_content` and `secret.tracked_content` directly, without copying. In production, `Secret.get_content()` returns `self._content.copy()`. Mutating the returned dict would silently corrupt internal Scenario state.
- **`action_get()`** in `_MockModelBackend` returned `action.params` directly. In production, `action_get()` returns fresh data from Juju, and in Harness it builds a new dict each time. Same mutation-leaks-into-state issue.

Both are the same class of bug fixed in commit `be090122` for `relation_get`, which returned `data.copy()` to prevent the mock's internal state from being shared with the caller.

## Changes

- `testing/src/scenario/mocking.py`: `secret_get()` now returns `.copy()` on both the `latest_content` and `tracked_content` paths; `action_get()` now returns `dict(action.params)`.
- Regression tests added in `test_secrets.py` and `test_actions.py` that verify mutating the returned dict does not affect subsequent calls.

### DIFF
--- a/testing/tests/test_consistency_checker.py
+++ b/testing/tests/test_consistency_checker.py
@@ -793,7 +793,11 @@ def test_storedstate_consistency():
         ),
     )
     assert_inconsistent(
-        State(stored_states={StoredState(owner_path=None, content={'secret': Secret({})})}),
+        State(
+            stored_states={
+                StoredState(owner_path=None, content={'secret': Secret({'key': 'value'})})
+            }
+        ),
         _Event('start'),
         _CharmSpec(
             MyCharm,


testing/src/scenario/mocking.py
```diff
@@ -481,9 +481,9 @@ def secret_get(
             if refresh:
                 secret._track_latest_revision()
             assert secret.latest_content is not None
-            return secret.latest_content
+            return secret.latest_content.copy()
 
-        return secret.tracked_content
+        return secret.tracked_content.copy()
```
dimaqq, line 487: I'm worried about footgun potential: I don't think we validate the shape of `.tracked_content` and the charmers supply it.

Juju ofc requires at least one key when the secret is created.

Would it be a good idea to enforce that when a testing.Secret is just created, as opposed to unlucky charmer hitting a confusing exception at the very end of `.run`?

testing/src/scenario/mocking.py
```diff
@@ -650,7 +650,7 @@ def action_get(self):
             raise ActionMissingFromContextError(
                 'not in the context of an action event: cannot action-get',
             )
-        return action.params
+        return dict(action.params)
```
dimaqq: We type `_Action.params` to be a mapping, but we don't validate that when the action is instantiated from `name, **kwargs`... Would it be a good idea to add validation at that point / or in `_Action.post_init()`?

testing/src/scenario/mocking.py
tonyandrewmeyer: This feels a little orthogonal to the fix in the PR, but it seems a reasonable addition to me, and does make the `.copy` more reliable (particularly that we know for sure that a deep copy is not required).

testing/src/scenario/mocking.py
tonyandrewmeyer: I'd rather not include this one in the PR, though. A type checker should find it and although lots of charms are not type checking their tests, they should be, so I'd rather encourage that than adding in `TypeError` raising where we do it ourselves.

testing/src/scenario/state.py
```diff
@@ -304,7 +304,30 @@ class Secret:
...
+        bad = {
+            k: v for k, v in content.items()
+            if not isinstance(k, str) or not isinstance(v, str)
+        }
+        if bad:
+            raise StateValidationError(
+                f'Secret.{name} should be dict[str, str]; '
+                f'found non-string key(s)/value(s): {bad}',
+            )
+
```
james-garner-canonical, line 333: Minor: I'd prefer this method definition to go after `__post_init__`

testing/tests/test_e2e/test_secrets.py
```diff
@@ -722,3 +722,33 @@ def common_assertions(scenario_secret: Secret | None, result: Result):
...
+
+        # A subsequent call must return the original data, unaffected.
+        assert secret_obj.get_content() == {'a': 'b'}
+
+        # Same check for refresh (latest_content path).
+        refreshed = secret_obj.get_content(refresh=True)
+        assert refreshed == {'a': 'c'}
+        refreshed['a'] = 'MUTATED'
+
+        assert secret_obj.get_content() == {'a': 'c'}
```
james-garner-canonical: Minor: can we dedent anything that doesn't need to be inside the context manager scope?

testing/tests/test_e2e/test_actions.py
```diff
@@ -223,3 +223,41 @@ def test_default_arguments():
...
+
+    ctx = Context(
+        mycharm,
+        meta={'name': 'foo'},
+        actions={'foo': {'params': {'bar': {'type': 'number'}}}},
+    )
+    ctx.run(ctx.on.action('foo', params={'bar': 10}), State())
+
+    assert len(results) == 1
+    mutated, fresh = results[0]
```
james-garner-canonical, line 257: I know we tend to prefer styling this as written already, but doesn't this look cool?
```py
    [(mutated, fresh)] = results
```

testing/src/scenario/mocking.py
james-garner-canonical: Is a shallow copy OK here?

tonyandrewmeyer: I guess not, given how complex action params could be.

```suggestion
        return copy.deepcopy(action.params)
```

testing/tests/test_e2e/test_actions.py
tonyandrewmeyer: Um, no? 😂

james-garner-canonical: 🥲

dimaqq: James, why not make it more magic: `(mut, fsh), = results` 🙃

testing/src/scenario/state.py
```diff
@@ -305,11 +305,34 @@ def __hash__(self) -> int:
...
+            raise StateValidationError(
+                f'Secret.{name} should be a dict, not {type(content)}',
+            )
+        if not content:
+            raise StateValidationError(
+                f'Secret.{name} must not be empty; Juju requires at least one key',
+            )
+        bad = {
+            k: v for k, v in content.items()
+            if not isinstance(k, str) or not isinstance(v, str)
```
dimaqq: ruff wants these on one line...

testing/src/scenario/state.py
```diff
@@ -305,11 +305,34 @@ def __hash__(self) -> int:
...
+        }
+        if bad:
+            raise StateValidationError(
+                f'Secret.{name} should be dict[str, str]; '
+                f'found non-string key(s)/value(s): {bad}',
```
dimaqq: And here too, ruff prefers a single line.

I think the tooling that you use didn't catch our custom line lenght limit.

dimaqq:
`test_storedstate_consistency` fails, the secret included into stored state (whatever what means!) needs a dummy field.

testing/tests/test_e2e/test_actions.py
james-garner-canonical: Doesn't look as cool :nerd_face:

PR #2380, opened 2026-03-15
fix: deep-copy layer objects during Scenario plan rendering

tonyandrewmeyer:
## Summary

- **Bug**: `Container._render_services()`, `_render_checks()`, and `_render_log_targets()` stored direct references to original Layer objects in their result dictionaries. When a subsequent layer merged into those references, the `_merge()` call mutated the original Layer's objects in place. This meant that accessing `.plan` (or `.services`) multiple times caused list fields (`after`, `before`, `requires`, `services`) to accumulate duplicates with each access.
- **Fix**: Use `copy.deepcopy()` when storing objects in the `else` branch (i.e. when a service/check/log_target is first encountered or uses `replace` override), so that merge operations never mutate the original Layer data.
- **Precedent**: This is the same class of bug that was previously fixed in Harness in commit `3dda5b5f` ("Pebble layer merging in Harness mutated the original layer objects").

### DIFF
--- a/testing/tests/test_e2e/test_pebble.py
+++ b/testing/tests/test_e2e/test_pebble.py
@@ -947,9 +947,6 @@ def test_plan_accessed_twice_does_not_accumulate_list_fields():
         layers={'base': layer1, 'override': layer2},
     )
 
-    # Access plan twice and snapshot list fields immediately.  Without
-    # deepcopy the plans share the same underlying objects, so comparing
-    # plan1 vs plan2 attributes directly would always pass (same object).
     plan1 = container.plan
     svc1_after = list(plan1.services['svc-a'].after)
     svc1_before = list(plan1.services['svc-a'].before)


testing/tests/test_e2e/test_pebble.py
```diff
@@ -880,3 +880,86 @@ def test_layers_merge_in_plan(layer1_name, layer2_name):
...
+    # Service list fields must not accumulate duplicates.
+    assert plan1.services['svc-a'].after == plan2.services['svc-a'].after
+    assert plan1.services['svc-a'].before == plan2.services['svc-a'].before
+    assert plan1.services['svc-a'].requires == plan2.services['svc-a'].requires
+
+    # Check fields must be stable across accesses.
+    assert plan1.checks['chk-a'].level == plan2.checks['chk-a'].level
+
+    # Log target list fields must not accumulate duplicates.
+    assert plan1.log_targets['lt-a'].services == plan2.log_targets['lt-a'].services
```
dwilding: Would there be a way to do a "deep equality" check between `plan1` and `plan2`? The standard library doesn't provide anything for that does it? In which case this seems fine

dwilding (commented):
I ran the new test with `testing/src/scenario/state.py` unmodified, and the test passed, but I was expecting it to fail. So it looks like something isn't quite right (perhaps my understanding is off!)

testing/tests/test_e2e/test_pebble.py
james-garner-canonical: We could do a deep equality check with `plan1.to_dict() == plan2.to_dict()`. But I don't think this test, as written, is actually exercising anything, since that just means `container.plan.to_dict() == container.plan.to_dict()`.

In order to actually test what this PR is trying to fix, the test should instead be mutating `layer1` and ensuring the changes aren't reflected here in the plan, or taking a deep copy of `layer1` before passing it into `Container` and then asserting that `layer1` wasn't mutated.

For example:
```py
    layer1_snapshot = copy.deepcopy(layer1.to_dict())  # idk if the copy is actually needed
    container = Container(
        'my-container',
        can_connect=True,
        layers={'base': layer1, 'override': layer2},
    )
    assert layer1.to_dict() == layer1_snapshot
```

testing/src/scenario/state.py
```diff
@@ -1091,7 +1091,7 @@ def _render_services(self):
                 if name in services and service.override == 'merge':
                     services[name]._merge(service)
                 else:
-                    services[name] = service
+                    services[name] = copy.deepcopy(service)
```
dimaqq, line 1095: **debatable:**

Arguably, the `_merge()` result two lines above could use a comment that it's relying on `ops.pebble.Service._merge()` to splatter the RHS and thus doesn't need an explicit copy.

And same for the two similar occurrences below.

testing/tests/test_e2e/test_pebble.py
```diff
@@ -880,3 +880,86 @@ def test_layers_merge_in_plan(layer1_name, layer2_name):
...
+    )
+
+    # Access plan twice: the second access should return identical results.
+    plan1 = container.plan
+    plan2 = container.plan
+
+    # Service list fields must not accumulate duplicates.
+    assert plan1.services['svc-a'].after == plan2.services['svc-a'].after
+    assert plan1.services['svc-a'].before == plan2.services['svc-a'].before
+    assert plan1.services['svc-a'].requires == plan2.services['svc-a'].requires
```
dimaqq: This tests implementation but not the API.

Imagine one day we'd update runtime pebble code to extend these fields omitting duplicates (instead of plain extend as it is today).

Suddenly this test doesn't actually test anything any more.

A better test could be something like:

```py
initial_state = ...
state_a = ctx.run(on_smth_that_merges_plan, initial_state)
assert "a" in state_a.get_container(...).foo

state_b = ctx.run(on_smthn_else_that_merges_plan, initial_state)
assert "a" not in state_b.get_container(...).foo
```

testing/tests/test_e2e/test_pebble.py
james-garner-canonical: I hadn't understood that `plan` is a property that computes the plan on the fly.

Maybe `assert container.plan.to_dict() == container.plan.to_dict()` would do what we want here?

testing/tests/test_e2e/test_pebble.py
```diff
@@ -880,3 +880,103 @@ def test_layers_merge_in_plan(layer1_name, layer2_name):
...
+
+    container = Container(
+        'my-container',
+        can_connect=True,
+        layers={'base': layer1, 'override': layer2},
+    )
+
+    # Access plan twice and snapshot list fields immediately.  Without
+    # deepcopy the plans share the same underlying objects, so comparing
+    # plan1 vs plan2 attributes directly would always pass (same object).
```
tonyandrewmeyer:
```suggestion
```

tonyandrewmeyer:
> I ran the new test with `testing/src/scenario/state.py` unmodified, and the test passed, but I was expecting it to fail. So it looks like something isn't quite right (perhaps my understanding is off!)

Your understanding is fine, the test was off. It should be fixed now (and fails for me if I remove the fixes). Sorry about that - I should have done that check myself.

james-garner-canonical (approved):

PR #2382, opened 2026-03-18
docs: add a tip about AI help in the Jubilant migration guide

tonyandrewmeyer:
As per the recent findings, a simple prompt that tells an agent to look over the repo (which, of course, includes the docs, handily in markdown/ReST rather than HTML) and then do the migration will go a long way. This PR adds a tip about this, without going into more detail on how you might use AI for this.

[Preview](https://canonical-ubuntu-documentation-library--2382.com.readthedocs.build/ops/2382/howto/migrate/migrate-integration-tests-from-pytest-operator/)

### DIFF
--- a/docs/howto/migrate/migrate-integration-tests-from-pytest-operator.md
+++ b/docs/howto/migrate/migrate-integration-tests-from-pytest-operator.md
@@ -3,6 +3,10 @@
 
 Many charm integration tests use [pytest-operator](https://github.com/charmed-kubernetes/pytest-operator) and [python-libjuju](https://github.com/juju/python-libjuju). This guide explains how to migrate your integration tests from those libraries to Jubilant.
 
+```{tip}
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
> I at first parsed this as saying "This PR adds a tip about this, ..." but I half wonder if it's an AI instruction given the imperative voice?

it's a typo missing the "s" (and skipping "this PR" as I often do). Will fix.

> I'm curious what details about using AI for this are being elided as the instructions seem pretty clear.

I think you'd get even better results if you set up a PR, either after the initial run or telling the agent to do it, and instructing it to monitor the results in CI and fix issues as they come up (or run the tests locally, but being able to do that seems uncommon at the moment). And also a reminder, assuming it's not in some AGENTS.md or so, to make sure that linting and formatting continue to pass. I saw issues with those when I was trying this out. But that seems to make it a bit wordy for a tip.

james-garner-canonical:
> > I'm curious what details about using AI for this are being elided as the instructions seem pretty clear.
> 
> I think you'd get even better results if you set up a PR, either after the initial run or telling the agent to do it, and instructing it to monitor the results in CI and fix issues as they come up (or run the tests locally, but being able to do that seems uncommon at the moment). And also a reminder, assuming it's not in some AGENTS.md or so, to make sure that linting and formatting continue to pass. I saw issues with those when I was trying this out. But that seems to make it a bit wordy for a tip.

Makes sense, thanks for elaborating!

PR #2386, opened 2026-03-22
docs: add a how-to guide focused on debugging charms

tonyandrewmeyer:
This PR adds a new how-to guide that explains the tooling that's available for debugging charms. It's limited to what's available in Juju and jhack, plus an extra section that explains how to use VS Code rather than pdb. I haven't tried to be exhaustive with jhack, just cover the commands that seem most useful in general for debugging.

It does not include information about debugging K8s; I feel like that would quickly rabbit-hole into a lot of extra content, and, hopefully, Juju is sufficient in most cases, and, where not, there are other guides that can help with anything K8s-specific.

I haven't included anything along the lines of "ssh in and look around", as that seems a bit obvious and not a great practice, but I think it probably is actually fairly common.

In the index page, I've put it with logging, since they seem quite related; however, this does mean that it's quite high up in the sidebar listing, perhaps more so than it should be.

[Preview](https://canonical-ubuntu-documentation-library--2386.com.readthedocs.build/ops/2386/howto/debug-your-charm/)

### DIFF
--- a/docs/howto/debug-your-charm.md
+++ b/docs/howto/debug-your-charm.md
@@ -83,6 +83,8 @@ A service in `backoff` or `error` state tells you the workload has been crashing
 
 ```{note}
 Pebble keeps the most recent output from each service in a 100 KB ring buffer. Older output is discarded, so if you need persistent logs consider configuring a Pebble {external+pebble:doc}`log forwarding target <how-to/forward-logs-to-loki>`.
+
+For a full observability setup -- including log forwarding, metrics, and dashboards -- see {ref}`Observe your charm with COS Lite <observe-your-charm-with-cos-lite>` and the [Canonical Observability Stack documentation](https://documentation.ubuntu.com/observability/).
 ```
 
 ### Run commands in the container
@@ -122,7 +124,7 @@ If the charm configures Pebble health checks, `pebble checks` shows their curren
 (read-charm-logs)=
 ## Read charm logs with `juju debug-log`
 
-The `juju debug-log` command streams log messages from every agent in a model. It is the first tool to reach for when something goes wrong. By default it shows recent log lines and then tails new output. Common flags:
+The {external+juju:ref}`juju debug-log <command-juju-debug-log>` command streams log messages from every agent in a model. It is the first tool to reach for when something goes wrong. By default it shows recent log lines and then tails new output. Common flags:
 
 ```shell
 juju debug-log --replay                          # show full history, then tail
@@ -160,7 +162,7 @@ This tells Juju to store DEBUG-level messages from charm units while keeping eve
 If you raise the stored log level for debugging (e.g. to DEBUG or TRACE), remember to restore it to the default once you are done. Verbose logs consume storage in the Juju database and can affect controller performance.
 ```
 
-> See more: {external+juju:ref}`Juju | juju debug-log <command-juju-debug-log>`, {external+juju:ref}`Juju | logging-config <model-config-logging-config>`
+> See more: {external+juju:ref}`Juju | logging-config <model-config-logging-config>`
 
 (use-jhack)=
 ## Use jhack for a faster debugging workflow
@@ -230,7 +232,7 @@ jhack show-relation myapp:database postgresql:database
 (debug-hooks)=
 ## Interactively debug hooks with `juju debug-hooks`
 
-The `juju debug-hooks` command opens a [`tmux`](https://github.com/tmux/tmux/wiki) session on a unit. When a matching hook fires, the session navigates to the charm directory with the full hook environment configured -- but the hook is *not* executed automatically. This gives you a chance to inspect the environment, modify files, and run the hook yourself.
+The {external+juju:ref}`juju debug-hooks <command-juju-debug-hooks>` command opens a [`tmux`](https://github.com/tmux/tmux/wiki) session on a unit. When a matching hook fires, the session navigates to the charm directory with the full hook environment configured -- but the hook is *not* executed automatically. This gives you a chance to inspect the environment, modify files, and run the hook yourself.
 
 ```shell
 juju debug-hooks myapp/0                 # intercept all hooks and actions
@@ -247,12 +249,10 @@ juju debug-hooks myapp/0 config-changed  # intercept only config-changed
 
 While a hook is being debugged, the unit is paused. Other hooks queue up and execute in order once you exit. Keep your debugging sessions short to avoid blocking the unit for too long.
 
-> See more: {external+juju:ref}`Juju | juju debug-hooks <command-juju-debug-hooks>`
-
 (debug-code)=
 ## Step through charm code with `juju debug-code`
 
-The `juju debug-code` command is similar to `debug-hooks`, but the hook *is* executed automatically. Juju sets the `JUJU_DEBUG_AT` environment variable, which Ops uses to activate breakpoints. When execution reaches a breakpoint, you are dropped into a {external+python:mod}`pdb` session where you can inspect variables and step through the code.
+The {external+juju:ref}`juju debug-code <command-juju-debug-code>` command is similar to `debug-hooks`, but the hook *is* executed automatically. Juju sets the `JUJU_DEBUG_AT` environment variable, which Ops uses to activate breakpoints. When execution reaches a breakpoint, you are dropped into a {external+python:mod}`pdb` session where you can inspect variables and step through the code.
 
 ```shell
 juju debug-code myapp/0                 # debug all hooks
@@ -284,8 +284,6 @@ Python's built-in {external+python:func}`breakpoint` also works when `JUJU_DEBUG
 
 ```{note}
 `ops.Framework.breakpoint()` won't trigger during state-transition tests, but the built-in `breakpoint()` will trigger as if `JUJU_DEBUG_AT` was set to `all`.
-
-> See more: {external+juju:ref}`Juju | juju debug-code <command-juju-debug-code>`
 
 (remote-debugging-with-vs-code)=
 ## Remote debugging with VS Code
@@ -349,8 +347,7 @@ juju show-unit myapp/0 | yq '.*.address'
 2. Trigger the hook you want to debug (or wait for it to fire naturally). The charm will start `debugpy` and block until a client connects.
 3. In VS Code, set your breakpoints and press **F5** (or click **Run > Start Debugging**). To get access to an Ops event object, you'll need to set a breakpoint inside the corresponding event handler.
 
-````{tip}
-**Debugging when Juju runs inside a Multipass VM**
+### Debugging from outside a Multipass VM
 
 If your Juju model is inside a Multipass VM (a common setup for local charm development), VS Code on your host machine cannot reach the charm unit's IP directly. Use an SSH port forward through the VM to bridge the gap:
 
@@ -369,4 +366,3 @@ ssh -N -L 5678:${UNIT_IP}:5678 ubuntu@${VM_IP}
 ```
 
 Then, in your `launch.json`, set `"host"` to `"localhost"` instead of the unit IP. VS Code will connect to the forwarded port on your host, and the SSH tunnel will relay traffic to `debugpy` on the charm unit inside the VM.
-````
--- a/docs/howto/manage-charms.md
+++ b/docs/howto/manage-charms.md
@@ -71,6 +71,7 @@ and the charm workflow.
 > - {ref}`write-unit-tests-for-a-charm`
 > - {ref}`write-integration-tests-for-a-charm`
 > - {ref}`manage-logs`
+> - {ref}`debug-your-charm`
 
 The next thing to do is add functionality to your charm.
 As you do that, you'll frequently pack, test, and debug your charm.


docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+```
+````
+
+The `--level` and `--debug` flags on `juju debug-log` only filter what is *displayed* -- they do not change what Juju actually records. To control which log levels are *stored*, use the `logging-config` model setting:
+
+```shell
+juju model-config logging-config="<root>=WARNING;unit=DEBUG"
+```
+
+This tells Juju to store DEBUG-level messages from charm units while keeping everything else at WARNING.
```
james-garner-canonical: Perhaps obvious given what you've already said, but how about making it even more explicit:
```suggestion
This tells Juju to store DEBUG-level messages from charm units while keeping everything else at WARNING. For this to show up in your logs, you'll need to set `logging-config` *before* the event you're interested in runs.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+
+```{tip}
+If you raise the stored log level for debugging (e.g. to DEBUG or TRACE), remember to restore it to the default once you are done. Verbose logs consume storage in the Juju database and can affect controller performance.
+```
+
+> See more: {external+juju:ref}`Juju | juju debug-log <command-juju-debug-log>`
```
james-garner-canonical: Perhaps we should also link to the `logging-config` docs here or earlier, since it supports some extra stuff like setting verbosity for separate units etc.

https://documentation.ubuntu.com/juju/3.6/reference/configuration/list-of-model-configuration-keys/#logging-config

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+
+In your charm code, call [](ops.Framework.breakpoint) to define breakpoints that you can selectively activate:
+
+```python
+class MyCharm(ops.CharmBase):
+    def _on_config_changed(self, event: ops.ConfigChangedEvent):
+        self.framework.breakpoint('config-start')
+        new_val = self.config['setting']
+        # ... process the new value ...
+        self.framework.breakpoint('config-end')
```
james-garner-canonical: Don't the breakpoint names need to match the name of the Juju event?

Maybe we could have an example of a bare `self.framework.breakpoint()` that will only trigger with `JUJU_DEBUG_AT` set to `all` or `hook`, and one with a name that will also trigger when `JUJU_DEBUG_AT` explicitly includes that hook name.

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+By default, `juju debug-code` sets `JUJU_DEBUG_AT=all`, activating every breakpoint. To activate only specific breakpoints, use `--at`:
+
+```shell
+juju debug-code --at=config-start myapp/0 config-changed
+```
+
+Python's built-in {external+python:func}`breakpoint` also works when `JUJU_DEBUG_AT` is set, so you can use either form.
```
james-garner-canonical: Worth adding something like this?
```suggestion
Python's built-in {external+python:func}`breakpoint` also works when `JUJU_DEBUG_AT` is set, so you can use either form. The main difference is that an `ops.Framework.breakpoint` can be set to trigger only on specific events. Note also that `ops.Framework.breakpoint` won't trigger during state-transition tests, but the built-in `breakpoint()` will trigger as if `JUJU_DEBUG_AT=all`.
```
I think so for the first new sentence, maybe for the second.

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+```python
+import os
+
+if os.getenv('JUJU_DEBUG_AT'):
+    import debugpy
+    debugpy.listen(('0.0.0.0', 5678))
+    debugpy.wait_for_client()
+```
+
+Place this at the top of the event handler you want to debug, or in your charm's `__init__` to intercept every hook.
```
james-garner-canonical:
```suggestion
Place this at the top of the event handler you want to debug, or in your charm's `__init__` to intercept every hook. Note that in the `__init__` context, your code won't have access to the current Juju event via Ops.
```
Or relation data? Could maybe mention the possibility of using `hookcmds` to poke at stuff? Or perhaps better to just steer users to debugging in event handlers instead. Perhaps a note to this effect would be better placed earlier in the doc since it applies equally to all methods of triggering a debugger.

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+Then, in your charm code, add a `debugpy` listener that activates when `JUJU_DEBUG_AT` is set (i.e. when you run `juju debug-code`):
+
+```python
+import os
+
+if os.getenv('JUJU_DEBUG_AT'):
+    import debugpy
+    debugpy.listen(('0.0.0.0', 5678))
+    debugpy.wait_for_client()
+```
```
james-garner-canonical: I wonder if it would be cleaner to provide a recipe that uses the built-in `breakpoint()` to trigger `debugpy`. Ops overrides `sys.breakpointhook` to point to `Framework.breakpoint`, so we'd get the correct `JUJU_DEBUG_AT` semantics for free on breakpoints.

The debugpy readme gives [an example](https://github.com/microsoft/debugpy?tab=readme-ov-file#breakpoint-function) which looks promising. Perhaps we could adapt it like this, with `listen` at the top of `charm.py` and then using `breakpoint()` as usual. I haven't tested this though.

````suggestion
Then, in your `charm.py`, add a `debugpy` listener that activates when `JUJU_DEBUG_AT` is set (i.e. when you run `juju debug-code`), and use `breakpoint()` like normal:

```python
import os

if os.getenv('JUJU_DEBUG_AT'):
    import debugpy
    debugpy.listen(('0.0.0.0', 5678))


class MyCharm(ops.CharmBase):
    # ...
    def _on_some_event(event: ops.EventBase):
        breakpoint() 
```
````

File: docs/howto/debug-your-charm.md [hunk 7]
```diff
@@ -0,0 +1,264 @@
...
+
+```shell
+juju show-unit myapp/0 | yq '.*.address'
+```
+
+### Start a debug session
+
+1. Run `juju debug-code myapp/0` to tell Juju to set `JUJU_DEBUG_AT` on the next hook execution.
+2. Trigger the hook you want to debug (or wait for it to fire naturally). The charm will start `debugpy` and block until a client connects.
+3. In VS Code, set your breakpoints and press **F5** (or click **Run > Start Debugging**).
```
james-garner-canonical: I guess this also obviates calling `wait_for_client` manually. But wait, does setting breakpoints in the editor after packing actually work? It definitely shouldn't, right? So you'd need to add `breakpoint()` to the charm code before packing.

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+
+# 4. Forward the debugpy port through the VM to your host:
+ssh -N -L 5678:${UNIT_IP}:5678 ubuntu@${VM_IP}
+```
+
+Then, in your `launch.json`, set `"host"` to `"localhost"` instead of the unit IP. VS Code will connect to the forwarded port on your host, and the SSH tunnel will relay traffic to `debugpy` on the charm unit inside the VM.
+````
+
+(use-jhack)=
+## Use jhack for a faster debugging workflow
```
james-garner-canonical: I wonder if `jhack` should come before the VSCode section, particularly given how specific and in the weeds it gets at the end with the VM port forwarding. I think we'd prefer users to be guided more to `jhack` than to the VSCode setup, and it applies more broadly too (not IDE specific). WDYT?

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,264 @@
...
+
+[jhack](https://github.com/canonical/jhack) is a toolkit that provides higher-level utilities on top of Juju. Several of its commands are particularly useful during charm development and debugging.
```
james-garner-canonical:
```suggestion
[jhack](https://github.com/canonical/jhack) is a toolkit that provides higher-level utilities on top of Juju. Several of its commands are particularly useful during charm development and debugging. You don't need to add anything special to your charm code to use `jhack`, just install it and run the commands.
```
Worth clarifying this since it contrasts with everything else in this guide IMO.

Also maybe we should add a warning box that `jhack` is pretty spicy (I know its own README does, but I think we should too).

docs/howto/debug-your-charm.md
tonyandrewmeyer: Yes, if you're using the VS Code approach, then you use the breakpoints in VS Code, you don't put `breakpoint()` or `framework.breakpoint()` calls in the code. I don't know what's going on under the hood, but that's the exact case that I validated works.

And, I believe, you *shouldn't* put breakpoints into the code, because those will try to go out to a tmux session that you're not using.

docs/howto/debug-your-charm.md
tonyandrewmeyer: That's the order I originally had, but I felt that the VS Code section logically belongs with `debug-hooks` and (particularly) `debug-code`, because they're all variations on the same thing, one giving you a plain SSH session, one giving you pdb, one giving you VS Code debugging.

I could move `jhack` above all of those instead, perhaps? It does seem likely that someone would reach for jhack before they reach for `debug-code` or `debug-hooks`.

docs/howto/debug-your-charm.md
tonyandrewmeyer: > Also maybe we should add a warning box that `jhack` is pretty spicy (I know its own README does, but I think we should too).

What do you mean by spicy here? That it's hacking into things? That the language is more fun that you might find elsewhere? That you'll get big warnings about destructive things? Something else?

docs/howto/debug-your-charm.md
james-garner-canonical: That sounds reasonable to me, it makes sense to keep VS Code with the other approaches.

docs/howto/debug-your-charm.md
james-garner-canonical: Basically that it depends on implementation details of Ops (and Juju?), does some potentially destructive stuff (e.g. you call out the risk of firing an event specifically) and is very much not an advisable way to interact with a production Juju deployment.

docs/howto/debug-your-charm.md
james-garner-canonical: My skim of the `debugpy` readme seemed to indicate that after calling `listen`, `breakpoint` would go to `debugpy` instead -- presumably they override `sys.breakpointhook`.

Though `ops.Framework` calls `pdb` directly in the `breakpoint` method it overrides `sys.breakpointhook` with, so I guess that breaks calling `listen` outside the charm class, since `Framework` will do its override afterwards.

It's out of scope for this PR, but maybe this suggests a feature we could add, making providing your own debug hook easier.

So I guess no change required here (if I have the control flow right now). And probably my [suggestion](https://github.com/canonical/operator/pull/2386#discussion_r2991138368) doesn't work. If that's the case, then what you have seems like a reasonable approach for now, thanks.

tonyandrewmeyer: When I was testing this, if I put `breakpoint()` in the code (which I had originally because I had first tested the bit about `debug-code`) it looked like things hung, because it was going to the wrong place.

I did also wonder about how we could make this easier, either in Ops or in Juju. I do think that VS Code is common enough that it would be ok to have something specific for it. My understanding from the [Discourse post](https://discourse.charmhub.io/t/remote-debugging-a-charm-with-vs-code/12984?u=tony-meyer) is that [Pycharm could support this](https://youtrack.jetbrains.com/issue/PY-63403/Support-debugpy) (but don't, at least not yet), and that the [protocol is fairly open](https://microsoft.github.io/debug-adapter-protocol/) with quite a lot of usage.

However, it's definitely out of scope for a docs PR as you said :). My thinking was to see whether this content is useful and whether people are doing this, and if so we could then work on making it simpler. But feel free to open an issue if you think we should be more proactive on this.

docs/howto/debug-your-charm.md
tonyandrewmeyer: As per other thread, calling `breakpoint()` in the code is bad: it drops you into `pdb` in the place you ran `debug-code`, and there's a weird message about that in VS Code rather than actually connecting to the debugging functionality in VS Code.

james-garner-canonical: Thinking further, I wonder if you put your `listen` call in your charm's `__init__`, if subsequent `breakpoint()` calls would work. Depending on when `debugpy` patches `sys.breakpointhook`, it should just work if we can get it to run its patching after Ops's.

docs/howto/debug-your-charm.md
tonyandrewmeyer: > Don't the breakpoint names need to match the name of the Juju event?

No, they are arbitrary strings. You run `juju debug-code [--at=<name>] <unit/x> [<hook or action name>]`. The last bit does need to be a hook or action name, but the `at` part is either "all" (the default) or "hook" (which is pre-configured to break at the start of each handler), or anything you choose (I just verified this with "foo", for example).

docs/howto/debug-your-charm.md
james-garner-canonical: > I did also wonder about how we could make this easier, either in Ops or in Juju. I do think that VS Code is common enough that it would be ok to have something specific for it. My understanding from the [Discourse post](https://discourse.charmhub.io/t/remote-debugging-a-charm-with-vs-code/12984?u=tony-meyer) is that [Pycharm could support this](https://youtrack.jetbrains.com/issue/PY-63403/Support-debugpy) (but don't, at least not yet), and that the [protocol is fairly open](https://microsoft.github.io/debug-adapter-protocol/) with quite a lot of usage.
> 
> However, it's definitely out of scope for a docs PR as you said :). My thinking was to see whether this content is useful and whether people are doing this, and if so we could then work on making it simpler. But feel free to open an issue if you think we should be more proactive on this.

I had something more Ops-specific in mind, like letting you all `framework.set_breakpointhook` and pass your own callable that it calls instead of `pdb`'s method. It would be a bit of an odd mixture of debugging concerns into charm code though (at least more than having `breakpoint()` calls in the first place). But it would be cool if it just worked!

I'll resolve this thread for now.

docs/howto/debug-your-charm.md
james-garner-canonical: Ah right, thanks. I see you explain `--at` right below this example, but I still somehow misunderstood. Worth making clearer somehow that they're arbitrary strings?

docs/howto/debug-your-charm.md
james-garner-canonical: So I guess this is not "can be set to trigger only on specific events" but "can be named so that `--at` only triggers specific named breakspoints".

james-garner-canonical: The note about unit test behaviour is a bit out of scope for this doc, but I don't think we document it anywhere. I'd mentioned in #2363 that we would document it as part of this debugging guide, but maybe we shouldn't.

tonyandrewmeyer:
> I do wonder if we should mention "ssh in and look around". It's pretty basic, as you said, but it might be good to make it explicit as a first step.

Sure. David also reminded me that there's some Pebble debugging tools that should be mentioned as well, so I'll add that also.

> Perhaps a sentence or two near the top, before or after the logging content, and an item in the numbered list at the bottom.

I'm thinking of dropping the bit at the bottom. It isn't necessarily the order you'd do things in, and it feels a bit filler, particularly given how long this already is (and I'm about to add two new bits). David suggested maybe having something like it at the top, but I feel like maybe that would just duplicate the ToC on the side. So I'm tempted to just remove it.

tonyandrewmeyer:
@james-garner-canonical there's been a non-trivial amount of editing here, partly as a result of your suggestions, but also adding in the Pebble bit and dropping the bit at the end. Feel free to re-review if you like (but also fine to just leave it as-is).

@dwilding it should be good for review now - but no hurry! Thanks!

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+(ssh-into-a-unit)=
+## Get a shell on a running unit with `juju ssh`
+
+The quickest way to poke around a live unit is to SSH into it with `juju ssh`. What you connect to depends on the substrate:
+
+**Machine charms.** `juju ssh` connects you to the machine itself. You land in a shell as the `ubuntu` user, which has password-less `sudo`:
+
+```shell
+juju ssh myapp/0
+```
```
dwilding: Very minor - I'd put the "You land in" part after the command. Otherwise it looks like we're introducing a command related to the `ubuntu` user.

````suggestion
**Machine charms.** `juju ssh` connects you to the machine itself:

```shell
juju ssh myapp/0
```

You land in a shell as the `ubuntu` user, which has password-less `sudo`.
````

File: docs/howto/debug-your-charm.md [hunk 11]
```diff
@@ -0,0 +1,367 @@
...
+
+**Kubernetes charms.** A K8s charm pod typically has multiple containers (the charm container and one or more workload containers). Use `--container` to choose which container to connect to:
+
+```shell
+juju ssh --container myworkload myapp/0
+```
+
+If you omit `--container`, `juju ssh` targets the charm container by default.
```
dwilding: Minor - Suggest more precise wording that matches the previous wording.

```suggestion
If you omit `--container`, `juju ssh` connects to the charm container by default.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+````{important}
+**Juju 4: SSH keys are no longer added automatically.**
```
dwilding, line 32: If I don't have an SSH key, can I ignore this info? E.g., what happens if I'm working inside a VM? I don't think we necessarily need to update the doc - just curious.

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+juju import-ssh-key gh:<your-github-username>
+```
+````
+
+> See more: {external+juju:ref}`Juju | juju ssh <command-juju-ssh>`
+
+(debug-with-pebble)=
+## Inspect the workload with Pebble (Kubernetes charms)
+
+In Kubernetes charms, each workload container runs {external+pebble:doc}`Pebble <index>` as its init system. You can use Pebble commands to inspect and interact with the workload directly. First, SSH into the workload container:
```
dwilding: I haven't seen us use "init system" before. How about "service manager" instead?


```suggestion
In Kubernetes charms, each workload container runs {external+pebble:doc}`Pebble <index>` as its service manager. You can use Pebble commands to inspect and interact with the workload directly. First, SSH into the workload container:
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
+(debug-your-charm)=
+# How to debug your charm
+
+> See first: {ref}`log-from-your-charm`, {external+juju:ref}`Juju | How to manage logs <manage-logs>`
+
+When your charm isn't behaving as expected, Juju and the broader charming ecosystem provide several tools to help you investigate. This guide covers how to get a shell on a running unit, read charm logs, interactively debug hook execution, and use third-party tools to accelerate your debugging workflow.
+
```
dwilding, line 7: How about adding a link to the 12-factor advice, which is pretty comprehensive:

```suggestion

For advice about debugging a 12-factor app charm, also see {external+charmcraft:ref}`Charmcraft | Use a 12-factor app charm <use-12-factor-charms-troubleshoot>`.

```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+A service in `backoff` or `error` state tells you the workload has been crashing.
+
+### View service logs
+
+`pebble logs` shows recent stdout and stderr from services. Use `-f` to follow in real time:
+
+```shell
+/charm/bin/pebble logs              # last 30 lines from all services
+/charm/bin/pebble logs -f           # tail and follow
+/charm/bin/pebble logs -n all       # show all buffered output
```
dwilding:
```suggestion
/charm/bin/pebble logs         # last 30 lines from all services
/charm/bin/pebble logs -f      # tail and follow
/charm/bin/pebble logs -n all  # show all buffered output
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+Pebble keeps the most recent output from each service in a 100 KB ring buffer. Older output is discarded, so if you need persistent logs consider configuring a Pebble {external+pebble:doc}`log forwarding target <how-to/forward-logs-to-loki>`.
+```
+
+### Run commands in the container
+
+`pebble exec` runs a one-off command inside the container. This is useful for checking files, environment variables, or connectivity:
+
+```shell
+/charm/bin/pebble exec -- ls /etc/myapp/
+/charm/bin/pebble exec --context myworkload -- env   # inherit the service's environment
```
dwilding:
```suggestion
/charm/bin/pebble exec --context myworkload -- env  # inherit the service's environment
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+`pebble logs` shows recent stdout and stderr from services. Use `-f` to follow in real time:
+
+```shell
+/charm/bin/pebble logs              # last 30 lines from all services
+/charm/bin/pebble logs -f           # tail and follow
+/charm/bin/pebble logs -n all       # show all buffered output
+```
+
+```{note}
+Pebble keeps the most recent output from each service in a 100 KB ring buffer. Older output is discarded, so if you need persistent logs consider configuring a Pebble {external+pebble:doc}`log forwarding target <how-to/forward-logs-to-loki>`.
```
dwilding, line 85: What do you think about plugging COS here?

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+```text
+$ /charm/bin/pebble plan
+services:
+    myworkload:
+        summary: my workload service
+        startup: enabled
+        override: replace
+        command: my-workload
+```
```
dwilding: Very minor suggestions for readability:

````suggestion
`pebble plan` prints the merged configuration that Pebble is currently using:

```text
$ /charm/bin/pebble plan
services:
    myworkload:
        summary: my workload service
        startup: enabled
        override: replace
        command: my-workload
```

This is helpful to verify that your charm wrote the correct layers.
````

File: docs/howto/debug-your-charm.md [hunk 19]
```diff
@@ -0,0 +1,367 @@
...
+
+### Check health checks
+
+If the charm configures Pebble health checks, `pebble checks` shows their current status:
+
+```shell
+/charm/bin/pebble checks
+```
+
+> See more: {external+pebble:doc}`Pebble | CLI commands <reference/cli-commands>`
```
dwilding: This is logically at the end of the "Inspect the workload with Pebble (Kubernetes charms)" section, but on the rendered page I don't think it's ideal because it looks only attached to "Check health checks". I suggest removing it from here. I have a separate suggestion for adding it further up the page.


```suggestion
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+(debug-with-pebble)=
+## Inspect the workload with Pebble (Kubernetes charms)
+
+In Kubernetes charms, each workload container runs {external+pebble:doc}`Pebble <index>` as its init system. You can use Pebble commands to inspect and interact with the workload directly. First, SSH into the workload container:
+
+```shell
+juju ssh --container <container-name> <unit>
+```
+
+Then use the Pebble CLI (available at `/charm/bin/pebble`) to inspect the workload.
```
dwilding: I think there's a (low) risk here that someone might think the discussion of the Pebble CLI is done. I think we should give another signal for what's coming and link to the CLI reference up-front. Something like this:


```suggestion
Then use the Pebble CLI (available at `/charm/bin/pebble`) to inspect the workload. We've included several examples below.

> See also: {external+pebble:doc}`Pebble | CLI commands <reference/cli-commands>`
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Inspect relation data with `jhack show-relation`
+
+`jhack show-relation` displays the relation databags for all units involved in a relation:
+
+```shell
+jhack show-relation myapp:database postgresql:database
+```
+
+> See more: [jhack](https://github.com/canonical/jhack)
```
dwilding: Similar comment as the Pebble CLI reference. I think it looks odd to have this at the end of the last subsection. We already have a link to jhack at the beginning of the section, so I'd just remove this one.

```suggestion
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+Firing events manually can desynchronise charm state from Juju state if your event handlers are not idempotent. Use this only in development and test environments.
+```
+
+### Push local changes with `jhack sync`
+
+`jhack sync` watches local directories and automatically pushes file changes to remote charm units. Combined with `jhack fire`, this enables a rapid edit-trigger-observe loop:
+
+```shell
+jhack sync myapp/0 --source ./src --source ./lib
+```
```
dwilding: Similar suggestion as in a couple of other places

````suggestion
`jhack sync` watches local directories and automatically pushes file changes to remote charm units:

```shell
jhack sync myapp/0 --source ./src --source ./lib
```

Combined with `jhack fire`, this enables a rapid edit-trigger-observe loop.
````

File: docs/howto/debug-your-charm.md [hunk 23]
```diff
@@ -0,0 +1,367 @@
...
+```shell
+jhack show-relation myapp:database postgresql:database
+```
+
+> See more: [jhack](https://github.com/canonical/jhack)
+
+(debug-hooks)=
+## Interactively debug hooks with `juju debug-hooks`
+
+The `juju debug-hooks` command opens a [`tmux`](https://github.com/tmux/tmux/wiki) session on a unit. When a matching hook fires, the session navigates to the charm directory with the full hook environment configured -- but the hook is **not** executed automatically. This gives you a chance to inspect the environment, modify files, and run the hook yourself.
```
dwilding: I think italics would be fine for emphasis, matching elsewhere on the page

```suggestion
The `juju debug-hooks` command opens a [`tmux`](https://github.com/tmux/tmux/wiki) session on a unit. When a matching hook fires, the session navigates to the charm directory with the full hook environment configured -- but the hook is *not* executed automatically. This gives you a chance to inspect the environment, modify files, and run the hook yourself.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+```{note}
+While a hook is being debugged, the unit is paused. Other hooks queue up and execute in order once you exit. Keep your debugging sessions short to avoid blocking the unit for too long.
+```
+
+> See more: {external+juju:ref}`Juju | juju debug-hooks <command-juju-debug-hooks>`
+
+(debug-code)=
+## Step through charm code with `juju debug-code`
+
+The `juju debug-code` command is similar to `debug-hooks`, but the hook **is** executed automatically. Juju sets the `JUJU_DEBUG_AT` environment variable, which Ops uses to activate breakpoints. When execution reaches a breakpoint, you are dropped into a {external+python:mod}`pdb` session where you can inspect variables and step through the code.
```
dwilding:
```suggestion
The `juju debug-code` command is similar to `debug-hooks`, but the hook *is* executed automatically. Juju sets the `JUJU_DEBUG_AT` environment variable, which Ops uses to activate breakpoints. When execution reaches a breakpoint, you are dropped into a {external+python:mod}`pdb` session where you can inspect variables and step through the code.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+```shell
+juju debug-code myapp/0                        # debug all hooks
+juju debug-code myapp/0 config-changed         # debug a specific hook
```
dwilding:
```suggestion
juju debug-code myapp/0                 # debug all hooks
juju debug-code myapp/0 config-changed  # debug a specific hook
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+> See more: [jhack](https://github.com/canonical/jhack)
+
+(debug-hooks)=
+## Interactively debug hooks with `juju debug-hooks`
+
+The `juju debug-hooks` command opens a [`tmux`](https://github.com/tmux/tmux/wiki) session on a unit. When a matching hook fires, the session navigates to the charm directory with the full hook environment configured -- but the hook is **not** executed automatically. This gives you a chance to inspect the environment, modify files, and run the hook yourself.
+
+```shell
+juju debug-hooks myapp/0                       # intercept all hooks and actions
+juju debug-hooks myapp/0 config-changed        # intercept only config-changed
```
dwilding:
```suggestion
juju debug-hooks myapp/0                 # intercept all hooks and actions
juju debug-hooks myapp/0 config-changed  # intercept only config-changed
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+```
+
+*Once a hook fires*, the `tmux` session lands in the charm directory. From there you can:
+
+- Inspect the environment variables that Juju provides (e.g. `JUJU_DISPATCH_PATH`).
```
dwilding:
```suggestion
- Inspect the environment variables that Juju provides, such as `JUJU_DISPATCH_PATH`.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+- Examine or modify files under `src/`.
+- Run `./dispatch` to execute the hook manually.
+- Run `./dispatch` again after making changes, to iterate.
+- Exit the `tmux` session to let the unit resume normal operation.
+
+```{note}
+While a hook is being debugged, the unit is paused. Other hooks queue up and execute in order once you exit. Keep your debugging sessions short to avoid blocking the unit for too long.
+```
```
dwilding: This seems fine as body text

```suggestion
While a hook is being debugged, the unit is paused. Other hooks queue up and execute in order once you exit. Keep your debugging sessions short to avoid blocking the unit for too long.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+juju debug-code --at=config-start myapp/0 config-changed
+```
+
+The `--at` value can be a string you have used with the `ops.Framework.breakpoint()` method, or the special value `hook`, which will break into the code at the start of each event handler.
+
+Python's built-in {external+python:func}`breakpoint` also works when `JUJU_DEBUG_AT` is set, so you can use either form. The advantage of `ops.Framework.breakpoint` is that you can combine it with labels that you use with `--at` to control where you breakpoint without changing the code.
+
+```{tip}
+`ops.Framework.breakpoint()` won't trigger during state-transition tests, but the built-in `breakpoint()` will trigger as if `JUJU_DEBUG_AT` was set to `all`.
+```
```
dwilding: This feels more like a note (or maybe it should be phrased more like a tip?)


```suggestion
```{note}
`ops.Framework.breakpoint()` won't trigger during state-transition tests, but the built-in `breakpoint()` will trigger as if `JUJU_DEBUG_AT` was set to `all`.
```
```

File: docs/howto/debug-your-charm.md [hunk 30]
```diff
@@ -0,0 +1,367 @@
...
+juju model-config logging-config="<root>=WARNING;unit=DEBUG"
+```
+
+This tells Juju to store DEBUG-level messages from charm units while keeping everything else at WARNING. For this to show up in your logs, you'll need to set `logging-config` *before* the event you're interested in runs.
+
+```{tip}
+If you raise the stored log level for debugging (e.g. to DEBUG or TRACE), remember to restore it to the default once you are done. Verbose logs consume storage in the Juju database and can affect controller performance.
+```
+
+> See more: {external+juju:ref}`Juju | juju debug-log <command-juju-debug-log>`, {external+juju:ref}`Juju | logging-config <model-config-logging-config>`
```
dwilding: I'd keep the logging-config link as a "see more", as it follows on from the part just above. But I'd remove the debug-log link and instead use an inline link in the very first sentence ("The `juju debug-log` command streams...")


```suggestion
> See more: {external+juju:ref}`Juju | logging-config <model-config-logging-config>`
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+- Examine or modify files under `src/`.
+- Run `./dispatch` to execute the hook manually.
+- Run `./dispatch` again after making changes, to iterate.
+- Exit the `tmux` session to let the unit resume normal operation.
+
+```{note}
+While a hook is being debugged, the unit is paused. Other hooks queue up and execute in order once you exit. Keep your debugging sessions short to avoid blocking the unit for too long.
+```
+
+> See more: {external+juju:ref}`Juju | juju debug-hooks <command-juju-debug-hooks>`
```
dwilding: I'd remove this and add an inline link to the first sentence ("The `juju debug-hooks` command opens...")


```suggestion
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+The `--at` value can be a string you have used with the `ops.Framework.breakpoint()` method, or the special value `hook`, which will break into the code at the start of each event handler.
+
+Python's built-in {external+python:func}`breakpoint` also works when `JUJU_DEBUG_AT` is set, so you can use either form. The advantage of `ops.Framework.breakpoint` is that you can combine it with labels that you use with `--at` to control where you breakpoint without changing the code.
+
+```{tip}
+`ops.Framework.breakpoint()` won't trigger during state-transition tests, but the built-in `breakpoint()` will trigger as if `JUJU_DEBUG_AT` was set to `all`.
+```
+
+> See more: {external+juju:ref}`Juju | juju debug-code <command-juju-debug-code>`
```
dwilding: I'd remove this and add an inline link to the first sentence ("The `juju debug-code` command is similar...")

```suggestion
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+```shell
+juju show-unit myapp/0 | yq '.*.address'
+```
+
+### Start a debug session
+
+1. Run `juju debug-code myapp/0` to tell Juju to set `JUJU_DEBUG_AT` on the next hook execution.
+2. Trigger the hook you want to debug (or wait for it to fire naturally). The charm will start `debugpy` and block until a client connects.
+3. In VS Code, set your breakpoints (note that you'll need to have a breakpoint in an event handler to get access to the Ops event object) and press **F5** (or click **Run > Start Debugging**).
```
dwilding: Is my suggestion still accurate?

```suggestion
3. In VS Code, set your breakpoints and press **F5** (or click **Run > Start Debugging**). To get access to an Ops event object, you'll need to set a breakpoint inside the corresponding event handler.
```

docs/howto/debug-your-charm.md
```diff
@@ -0,0 +1,367 @@
...
+
+````{tip}
```
dwilding: How about promoting this to a subsection? It feels important enough. Something like:

```
### Debugging from outside a Multipass VM
...
```

*Edit:* Or even, "Debugging outside a Multipass VM"

docs/howto/debug-your-charm.md
tonyandrewmeyer: Nice! I was not aware of that doc, I don't think.

docs/howto/debug-your-charm.md
tonyandrewmeyer: With Juju 3, I don't know what happens, actually. Maybe Juju makes one for you? With Juju 4, you have to have an SSH key and add it manually, so if you don't have one then you need to make one (or import one from GitHub or Launchpad).

Inside a VM that would still apply.

docs/howto/debug-your-charm.md
tonyandrewmeyer: Sure. Do you have a suggestion where we should link? https://documentation.ubuntu.com/observability/track-2/ maybe?

dwilding: Maybe we should link to the COS Lite chapter of our K8s tutorial, as that shows what might need adding to charm code to forward logs. I couldn't find anything directly comparable in the Observability docs, so maybe include that as a secondary link?

dwilding:
One other thing I just thought of. We probably want to list the new page somewhere on https://documentation.ubuntu.com/ops/latest/howto/manage-charms/. I suppose after the logging doc under "Develop your charm"?

PR #2387, opened 2026-03-23
docs: Juju secrets identifier is now an opaque string

dimaqq:
Starting from Juju 3.6.19, the secrets identifier is no longer an Xid, as a mitigation for https://github.com/juju/juju/security/advisories/GHSA-5cj2-rqqf-hx9p

### DIFF
--- a/ops/model.py
+++ b/ops/model.py
@@ -1391,7 +1391,7 @@ class Secret:
     def unique_identifier(self) -> str | None:
         """Unique identifier of this secret.
 
-        This is the secret's globally-unique identifier (20 characters).
+        This is the secret's globally-unique identifier (alphanumeric).
 
         Charms should use :attr:`id` (the secret's locator ID) to send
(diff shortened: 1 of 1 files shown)


ops/model.py
```diff
@@ -1391,8 +1391,7 @@ def id(self) -> str | None:
     def unique_identifier(self) -> str | None:
         """Unique identifier of this secret.
 
-        This is the secret's globally-unique identifier (currently a
-        20-character Xid, for example "9m4e2mr0ui3e8a215n4g").
+        This is the secret's globally-unique identifier (20 characters).
```
dwilding: I was surprised to see that the length is still 20 characters. Just want to double check that's right

tonyandrewmeyer: Even though it is, I don't think that's important or something that charms should rely on, or necessarily something that Juju guarantees. I'd suggest it's removed.

dimaqq: fair point; I'll just mention that it's "alphanumeric"

PR #2388, opened 2026-03-23
ci: drop external action for checking PR title format

james-garner-canonical:
This PR drops the use of an external, third-party action for checking PR title format, replacing it with a small, locally-defined Python script. This has the added security benefit of letting us run this check in the safer `pull_request` context, rather than on `pull_request_target`, which the action apparently required for creating a [commit status](https://docs.github.com/en/rest/commits/statuses?apiVersion=2026-03-10).

I'm proposing the same approach for `charmlibs` in https://github.com/canonical/charmlibs/pull/380

The script in this repo disallows the use of conventional commit scopes, and uses the same set of commits that Ops uses.

---

In addition to running sanity tests locally with the environment variable set to different values, I've tested it in our CI by editing this PRs title several times.

Due to the workflow's concurrency definition, you'll notice that the `pull_request` version and `pull_request_target` version race and one gets cancelled. This will be resolved when this PR is merged and only one check is defined.

### DIFF
--- a/.github/check-conventional-pr-title.py
+++ b/.github/check-conventional-pr-title.py
@@ -31,7 +31,7 @@ _TYPES = frozenset({
 
 # <type>[optional scope][optional !]: <description>
 _PATTERN = re.compile(
-    r'^(?p<type>[a-za-z]+)'  # lower-case only, but let this be validated by _TYPES
+    r'^(?P<type>[A-Za-z]+)'  # lower-case only, but let this be validated by _TYPES
     r'(?:\((?P<scope>[^()]+)\))?'
     r'(?P<breaking>!)?'
     r': '


.github/workflows/validate-pr-title.yaml
```diff
@@ -18,18 +17,9 @@ jobs:
...
-            docs
-            feat
-            fix
-            perf
-            refactor
-            revert
-            test
-          disallowScopes: ".*"
+          persist-credentials: false
+      - run: python3 .github/check-conventional-pr-title.py
```
dimaqq, line 19: WDYT about attempting a super-compact inline Python block instead?

.github/workflows/validate-pr-title.yaml
```diff
@@ -3,11 +3,10 @@ name: "Validate PR Title"
 # Ensure that the PR title conforms to the Conventional Commits and our choice of types and scopes, so that library version bumps can be detected automatically
 
 on:
-  pull_request_target:  # zizmor: ignore[dangerous-triggers] Doesn't touch code - pull_request results in cancellation errors.
+  pull_request:
```
dimaqq, line 6: I think we'd be back to cancellation errors, unless the concurrency group is removed or changed.

Maybe it's cleaner to the remove concurrency seeting, given how simple the job is.

james-garner-canonical: I believe the cancelation errors are only due to there being two workflows running right now, the pull_request trigger one from this branch, and the pull_request_target one from main.

.github/workflows/validate-pr-title.yaml
james-garner-canonical: I don't love the idea, I think that will be harder to maintain in future

.github/workflows/validate-pr-title.yaml
dimaqq: I doubt that. We used to have `on: pull_request` originally, and had those cancellation errors. I think it's because pushing the branch and creating a PR triggers both "pr created" and "code pushed" events. TBH I never found the root cause.

james-garner-canonical: I'm certain that the cancellations we see in this PR are because both the `pull_request` trigger from this branch, and the `pull_request_target` trigger from `main` are running here, and they define the same concurrency group.

I know those aren't the errors you're talking about, just to be clear.

I suggest we merge this PR and keep an eye out for cancellation errors thereafter. If we see the issue you remember, then let's remove the concurrency. I believe the idea with concurrency for this job is to avoid multiple quick PR edits resulting in spurious failures if an intermediate edit made a mistake.

.github/workflows/validate-pr-title.yaml
dimaqq: How about this?

```yaml
- uses: actions/github-script@v7
  with:
    script: |
      /^(chore|ci|docs|feat|fix|perf|refactor|revert|test)!?: .+$/.test(context.payload.pull_request.title) || core.setFailed('PR title must be "<type>[!]: <description>", with no scope')
```

dimaqq: Or inline Python:

```py
prefixes = ["chore: ", "chore!: ", ...]
if not title.startswith(prefixes): raise ...
```

.github/check-conventional-pr-title.py
```diff
@@ -0,0 +1,84 @@
...
+            f'Got: {title!r}\n'
+            'Read more: https://github.com/canonical/operator/blob/main/CONTRIBUTING.md#pull-requests',
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    scope = match.group('scope')
+    if scope is not None:
+        print(
+            f'Scopes are not used in PR titles.\n'
```
benhoyt:
```suggestion
            f'Scopes must not be used in PR titles.\n'
```

.github/workflows/validate-pr-title.yaml
benhoyt: Dima's github-script one-liner is interesting, but overall I think the approach here is cleaner, and it has the precedent of already being in charmlibs.

james-garner-canonical:
> This works for me, especially seeing you've already merged something similar over in charmlibs. If there were any changes there, let's update this PR to match.

Thanks. This PR is up-to-date with the changes made in `charmlibs` thanks to Tony's review there.

.github/workflows/validate-pr-title.yaml
james-garner-canonical: I'm wondering if we should drop the concurrency group before merging, as you suggested initially @dimaqq. We've been running this in `charmlibs` for a bit now without `concurrency`, and the UX seems to be fine.

james-garner-canonical: I've dropped concurrency.

.github/check-conventional-pr-title.py
```diff
@@ -0,0 +1,84 @@
...
+    'fix',
+    'perf',
+    'refactor',
+    'revert',
+    'test',
+})
+
+# <type>[optional scope][optional !]: <description>
+_PATTERN = re.compile(
+    r'^(?p<type>[a-za-z]+)'  # lower-case only, but let this be validated by _TYPES
```
james-garner-canonical:
```suggestion
    r'^(?P<type>[A-Za-z]+)'  # lower-case only, but let this be validated by _TYPES
```

.github/workflows/validate-pr-title.yaml
dimaqq: ok

PR #2393, opened 2026-03-29
docs: add a note to our tutorials about the confusing concierge output

james-garner-canonical:
We plan to improve `concierge` output in future, but right now it confuses users who follow our tutorial. This PR adds a note about the "error" messages to setup step. This can be reverted when we address this in `concierge`.

- https://github.com/canonical/concierge/issues/102
- https://github.com/canonical/concierge/issues/171

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/set-up-your-development-environment.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/set-up-your-development-environment.md
@@ -60,7 +60,7 @@ sudo concierge prepare -p k8s --extra-snaps astral-uv
 
 This first installs Concierge, then uses Concierge to install and configure the other tools (except tox). The option `-p k8s` tells Concierge that we want tools for developing Kubernetes charms, with a local cloud managed by Canonical Kubernetes.
 
-This step should take less than 15 minutes, but the time depends on your computer and network. When the tools have been installed, you'll see a message that ends with:
+This step should take less than 15 minutes, but the time depends on your computer and network. As `concierge` runs, you may see a number of messages that appear to be errors, like `ERROR controller ... not found`. Don't worry about these, they're because of how `concierge` currently exposes the output of the commands it runs internally. You only need to worry if your `concierge prepare` call fails -- then you should investigate the output, and reach out to us if you can't solve the problem yourself. When the tools have been installed, you'll see a message that ends with:
 
 ```text
 msg="Bootstrapped Juju" provider=k8s
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -95,7 +95,7 @@ sudo concierge prepare -p machine --extra-snaps astral-uv
 
 This first installs Concierge, then uses Concierge to install and configure the other tools (except tox). The option `-p machine` tells Concierge that we want tools for developing machine charms.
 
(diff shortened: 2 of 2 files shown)


james-garner-canonical:
> I would much rather we just fixed the issue in Concierge.

Me too, but the issue has been open in Concierge for 6 months, so a note to help people in the meantime seems reasonable.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/set-up-your-development-environment.md
```diff
@@ -60,7 +60,7 @@ sudo concierge prepare -p k8s --extra-snaps astral-uv
 
 This first installs Concierge, then uses Concierge to install and configure the other tools (except tox). The option `-p k8s` tells Concierge that we want tools for developing Kubernetes charms, with a local cloud managed by Canonical Kubernetes.
 
-This step should take less than 15 minutes, but the time depends on your computer and network. When the tools have been installed, you'll see a message that ends with:
+This step should take less than 15 minutes, but the time depends on your computer and network. As `concierge` runs, you may see a number of messages that appear to be errors, like `ERROR controller ... not found`. Don't worry about these, they're because of how `concierge` currently exposes the output of the commands it runs internally. You only need to worry if your `concierge prepare` call fails -- then you should investigate the output, and reach out to us if you can't solve the problem yourself. When the tools have been installed, you'll see a message that ends with:
```
dimaqq, line 63: Nice.
I would personally use a different form, if there's more than one kind of message:

The following errors are transient:
- `ERROR controller ... not found`
- `yada yada bleh lbeh`

james-garner-canonical: Good idea, but I think it's best not to try to be exhaustive here, just give an idea of the general type of problem, especially as this is intended as a temporary solution.

benhoyt:
Let's be careful here -- as discussed in daily, I'd really like us to spend our time pushing and reviewing a Concierge fix rather than reviewing temporary "doc workarounds". It's possible we could have had the fix already if we'd focussed our effort there. If that fix turns out to be much larger than we expect, we can revisit a doc workaround.

dwilding:
> Let's be careful here -- as discussed in daily, I'd really like us to spend our time pushing and reviewing a Concierge fix rather than reviewing temporary "doc workarounds".

I get what you're saying and don't disagree, in general. In this instance, we already have the doc update from James, and the tutorials are high visibility. Two instances of confusion related to this have been raised within the past week. I'd still push for publishing the workaround in this instance, if Tony is onboard.

tonyandrewmeyer:
> Concierge is causing confusion and we're still at least a release cycle away from an improvement.

"Release cycle" is < 1 month, though. That's not really that long. And if this really is critical, then we could do a bug fix release mid-month. (I'm not convinced it is).

> I'd still push for publishing the workaround in this instance, if Tony is onboard.

Unfortunately, Tony is not. I agree with Ben: let's just fix the problem, not document that there is a problem. I think it's confusing, but not exceedingly so. I already said last week that I'd prioritise getting this done before Nhan starts, so we can be confident that it'll be in the next release of Concierge, and (as above) could always push out an extra release if needed.

I think documenting problems where we can fix them, instead of just putting the effort into fixing them, is a bad precedent for how our team spends time, and also gives a bad impression of the tools we produce.

tonyandrewmeyer:
Can we close this? If we feel this is urgent I would rather someone advocated for a new Concierge release with the improvement rather than adding this to our docs.

PR #2404, opened 2026-03-30
chore: bump pygments, requests, and cryptography versions

tonyandrewmeyer:
pygments is used by pytest. This release fixes a [security issue](https://github.com/canonical/operator/security/dependabot/40), but does have other changes (dropping Python 3.8 support, and a bunch more). See https://github.com/canonical/operator/pull/2403 for more details. The security issue is not one that we are likely to be impacted by, but we might as well move to the new version, particularly since this is a CI/dev dependency.

Also bumps requests, which has another [security issue](https://github.com/psf/requests/releases/tag/v2.33.0) that would also not impact us.

Also bumps cryptography, which has another [security issue](https://cryptography.io/en/latest/changelog/#v46-0-6) that as far as I can tell, likely also does not impact us.

(We could wait a week before merging this if we decided that was safer.)


dwilding (approved):
I guess we ought to bump pygments in the Charmcraft profiles too? I can take care of that.

dwilding:
FYI here's the corresponding Charmcraft PR: https://github.com/canonical/charmcraft/pull/2622

PR #2405, opened 2026-03-31
docs: add how-to guide about CI

dwilding:
This PR consolidates our various CI examples and puts them into a how-to guide so they are much easier to find.

**[Preview of the new guide](https://canonical-ubuntu-documentation-library--2405.com.readthedocs.build/ops/2405/howto/set-up-continuous-integration-for-a-charm/)**

I tried to keep the guidance minimal and close to what we currently have. We can build on the guidance in follow-up PRs. But if the examples themselves need tweaking/improving, let's do that in this PR.

---

Merge this after https://github.com/canonical/operator/pull/2406


docs/howto/manage-charms.md
```diff
@@ -70,7 +71,8 @@ and the charm workflow.
 > - {ref}`write-and-structure-charm-code`
 > - {ref}`write-unit-tests-for-a-charm`
 > - {ref}`write-integration-tests-for-a-charm`
-> - {ref}`manage-logs`
+> - {ref}`set-up-ci`
+> - {ref}`log-from-your-charm`
```
dwilding, line 75: `manage-logs` pointed to the Juju docs. I'm fixing the link so that it points to our Ops doc instead.

docs/howto/manage-charms.md
```diff
@@ -17,8 +17,9 @@ To deploy your charm locally and to run integration tests, you'll also need a Ju
 
 > See more:
 >
+> - [Prepare your environment to develop Kubernetes charms](#set-up-your-development-environment)
```
dwilding, line 20: I meant to add this after switching the K8s charm tutorial to use Concierge. Adding it now.

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,102 @@
...
+
+```{admonition} Best practice
+:class: hint
+
+The quality assurance pipeline of a charm should be automated using a continuous integration (CI) system.
+```
+
+This guide demonstrates how to automatically run your charm's tests against any PR into the main branch of your GitHub repository.
+
+You might also want to automatically publish your charm on Charmhub or publish charm libraries on PyPI. [charming-actions](https://github.com/canonical/charming-actions) has some useful GitHub actions for publishing on Charmhub. For guidance about publishing on PyPI, see {external+charmlibs:ref}`How to distribute charm libraries <python-package-distribution-pypi>`.
```
james-garner-canonical: I wonder if we should keep endorsing `charming-actions`.

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,102 @@
...
+      - name: Run unit tests
+        run: tox -e unit
+```
+
+(set-up-ci-integration)=
+## Create a workflow for integration tests
+
+Integration tests require a Juju controller and a cloud in which to deploy your charm. We recommend that you use [Concierge](https://github.com/canonical/concierge) to prepare the CI environment.
+
+If your charm is a Kubernetes charm, create a file called `.github/workflows/integration-tests.yaml`:
```
james-garner-canonical: How about we recommend putting everything in a `ci.yaml` file for now? The integration test job is simple enough to live in the same file as the rest.

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,102 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  integration:
```
james-garner-canonical, line 77: If we put this in the same file, how about adding `needs: [lint, unit]`?

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,102 @@
...
+        uses: actions/checkout@v6
+      - name: Set up uv
+        uses: astral-sh/setup-uv@7
+      - name: Set up tox and tox-uv
+        run: uv tool install tox --with tox-uv
+      - name: Set up Concierge
+        run: sudo snap install --classic concierge
+      - name: Set up Juju and charm development tools
+        run: sudo concierge prepare -p k8s
+      - name: Run integration tests
```
james-garner-canonical, line 99: We should add a `charmcraft pack` here, and either the `CHARM_PATH` environment variable we recommend using, or add an explicit comment that `tox -e integration` will find the charms by path.

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,102 @@
...
+        # Set a predictable model name so it can be consumed by charm-logdump-action.
+        run: tox -e integration -- --model testing
+      - name: Dump logs
+        uses: canonical/charm-logdump-action@main
+        if: failure()
+        with:
+          app: your-app
+          model: testing
```
james-garner-canonical: Maybe we could do [this](https://github.com/canonical/operator/pull/2406/changes#r3013101198) instead?

dwilding: Would we need to make sure the charm has `pytest-jubilant` in its integration testing deps?

docs/howto/set-up-continuous-integration-for-a-charm.md
dwilding: Oops, yes, thanks. Added a pack step

docs/howto/set-up-continuous-integration-for-a-charm.md
james-garner-canonical: Yes. I think that's OK, as [charm-logdump-action](https://github.com/canonical/charm-logdump-action) makes some pretty strong assumptions too (microk8s only, models aren't torn down by integration tests). We should document that we're assuming `pytest-jubilant` integration tests, of course.

docs/howto/manage-charms.md
tonyandrewmeyer: Huh. I think we used to have a `manage-logs` reference in Ops, and it was removed when we refactored things. I would have expected the link to break in that case, and only try to match with inter-sphinx links if there was "external" here. I'll need to keep this fallback behaviour in mind.

docs/howto/set-up-continuous-integration-for-a-charm.md
tonyandrewmeyer: I think we should *not*. But I'm not sure whether we should change that in this PR or later on. I feel like it's a separate PR where we figure out for each feature that `charming-actions` offers what the Charm Tech approved alternative is, and then update our docs (and potentially your spec) accordingly.

There's also the charmlibs bit. There's still value in the various actions for managing Charmhub charmlibs, but I'm not sure we want to promote those in our docs, unless it's in a migration guide or legacy page.

Maybe this should be a roadmap item for 26.10?

It seems to me like it would be simpler to keep the scope of this PR smaller, and focus on consolidating the content more than updating it, except in very straightforward cases.

But also:

```suggestion
You might also want to automatically publish your charm on Charmhub or publish charm libraries on PyPI. [charming-actions](https://github.com/canonical/charming-actions) has some useful GitHub actions for publishing on Charmhub. For guidance about publishing libraries on PyPI, see {external+charmlibs:ref}`How to distribute charm libraries <python-package-distribution-pypi>`.
```

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,106 @@
...
+
+```yaml
+name: Linting and unit tests
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_call:
+  workflow_dispatch:
```
tonyandrewmeyer, line 33: Is this useful for linting and unit tests? I feel like if you want to run those manually, you ought to be able to do that locally.

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+This guide demonstrates how to automatically run your charm's tests against any PR into the main branch of your GitHub repository.
+
+You might also want to automatically publish your charm on Charmhub or publish charm libraries on PyPI. [charming-actions](https://github.com/canonical/charming-actions) has some useful GitHub actions for publishing on Charmhub. For guidance about publishing on PyPI, see {external+charmlibs:ref}`How to distribute charm libraries <python-package-distribution-pypi>`.
+
+(set-up-ci-linting-unit)=
+## Create a workflow for linting and unit tests
+
+Create a file called `.github/workflows/tests.yaml`:
```
tonyandrewmeyer: Let's Zizmor this file and fix the issues it'll report (like not storing git credentials, default permissions, etc). People will copy it exactly, so we should make it as safe as possible.

docs/howto/set-up-continuous-integration-for-a-charm.md
tonyandrewmeyer: If we do that, then we should leave the `workflow_dispatch` in place.

docs/howto/set-up-continuous-integration-for-a-charm.md
tonyandrewmeyer: I don't think integration tests need linting and unit tests to pass in order to run. In CI, I would rather see everything that's going wrong rather than seeing some formatting failure and having to fix that before I can see the integration test failures, rather than seeing all of them at once.

docs/howto/set-up-continuous-integration-for-a-charm.md
```diff
@@ -0,0 +1,106 @@
...
+        run: sudo concierge prepare -p k8s
+      - name: Pack the charm
+        # The integration tests don't pack the charm. Instead, they look for a .charm
+        # file in the project dir (or use CHARM_PATH, if set).
+        run: charmcraft pack
+      - name: Run integration tests
+        # Set a predictable model name so it can be consumed by charm-logdump-action.
+        run: tox -e integration -- --model testing
+      - name: Dump logs
+        uses: canonical/charm-logdump-action@main
```
tonyandrewmeyer: I wonder about recommending this, too. It's a lot more than the Juju logs, but is it something that should be used by all (K8s) charms? If so, should Charm Tech be providing something for this ourselves?

tonyandrewmeyer (approved):
Thanks! This looks good to me. I think it's best to keep it more-or-less like it is now, and then make adjustments to what we recommend, particularly with regards to external actions/workflows, in subsequent PRs. Partly so that this one stays focused, but partly because I think we haven't answered for ourselves exactly what we're recommending yet.

docs/howto/set-up-continuous-integration-for-a-charm.md
james-garner-canonical: Well maybe just `needs: [unit]` then if you want to reduce the chance that we're blocking an integration test run that would otherwise succeed or fail in an interesting way?

I personally find it's well worth having to quickly fix bugs that would be caught by lint/unit before the integration tests will run, if only to avoid getting emails 15 minutes later because bad code from an old commit was being packed and deployed in the background and now finally failed due to a missing import or similar that was fixed when unit tests run.

I guess it is a matter of taste though, and arguably a skill issue if you push code that would fail lint/unit, so if we want to stick with Charm Tech's recommendation being to start your integration tests ASAP, it's not a blocker for me.

docs/howto/set-up-continuous-integration-for-a-charm.md
dwilding: The comment link doesn't seem to be working. Trying this link instead https://github.com/canonical/operator/pull/2406#discussion_r3013101198

docs/howto/set-up-continuous-integration-for-a-charm.md
dwilding: Let's leave `charming-actions` in there for now. I've created a placeholder issue for improvements (https://github.com/canonical/operator/issues/2410), which we can prioritise later.

docs/howto/set-up-continuous-integration-for-a-charm.md
dwilding: I think the argument for `unit` is convincing and also the argument against linting/formatting is also convincing. So I'll go with `needs: [unit]` and we can always change our mind later.

docs/howto/set-up-continuous-integration-for-a-charm.md
dwilding: OK, I'll switch to a single file

docs/howto/set-up-continuous-integration-for-a-charm.md
dwilding: I switched to the `--juju-dump-logs` approach and have added a sentence with more detail:

```
The "Upload logs" step assumes that your integration tests use Jubilant together
with `pytest-jubilant`. See [How to write integration tests for a charm](#write-
integration-tests-for-a-charm-view-juju-logs).
```

That section of the how-to guide is being added in #2406.

I'll leave this comment thread open because the approach didn't work as expected when I tried it in GitHub. See https://github.com/canonical/operator/pull/2406#discussion_r3031101959.

dwilding: I've updated the example to set the log directory to a non-hidden directory, following https://github.com/canonical/pytest-jubilant/pull/55:

```
      - name: Run integration tests
        run: tox -e integration -- --juju-dump-logs logs
      - name: Upload logs
        if: ${{ !cancelled() }}
        uses: actions/upload-artifact@v7
        with:
          name: juju-dump-logs
          path: logs
```

PR #2426, opened 2026-04-14
ci: configure dependabot for the example charms

tonyandrewmeyer:
This PR extends the dependabot configuration to also include the example charms, with the same settings (`uv`, monthly, 7-day cooldown) as the main dependencies.

For the two tutorial charms, which start off with a `charmcraft init`, the dependencies that `init` brings in are ignored, as those will be picked up in the `charmcraft` or `charmcraft-profile-tools` repositories.

For the K8s tutorial, only configure for the final chapter, to avoid noise. When reviewing those dependabot PRs, people should consider whether earlier chapters also need to be updated. For both tutorials, people should consider at the time whether any docs also need to be updated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,6 +37,7 @@ updates:
       - dependency-name: "pyright"
       - dependency-name: "coverage"
       - dependency-name: "pytest"
+      - dependency-name: "pytest-jubilant"
       - dependency-name: "jubilant"
       - dependency-name: "PyYAML"
     labels:
@@ -54,7 +55,7 @@ updates:
       - dependency-name: "pyright"
       - dependency-name: "coverage"
       - dependency-name: "pytest"
+      - dependency-name: "pytest-jubilant"
       - dependency-name: "jubilant"
-      - dependency-name: "PyYAML"
     labels:
       - "dependencies"
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -13,3 +13,45 @@ updates:
     open-pull-requests-limit: 0  # Security updates only
     labels:
       - "dependencies"
+  - package-ecosystem: "uv"
+    directory: "/examples/httpbin-demo"
+    schedule:
+      interval: "monthly"
+    cooldown:
+      default-days: 7
+    labels:
+      - "dependencies"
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -16,3 +16,45 @@ updates:
...
+      default-days: 7
+    ignore:
+      - dependency-name: "ops"
+      - dependency-name: "ruff"
+      - dependency-name: "codespell"
+      - dependency-name: "pyright"
+      - dependency-name: "coverage"
+      - dependency-name: "pytest"
+      - dependency-name: "jubilant"
+      - dependency-name: "PyYAML"
```
dwilding: PyYAML isn't needed for the machine profile, so shall we remove it?

.github/dependabot.yml
```diff
@@ -16,3 +16,45 @@ updates:
...
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "ops"
+      - dependency-name: "ruff"
+      - dependency-name: "codespell"
+      - dependency-name: "pyright"
+      - dependency-name: "coverage"
+      - dependency-name: "pytest"
+      - dependency-name: "jubilant"
```
dwilding: We're adding pytest-jubilant to the profile in https://github.com/canonical/charmcraft/pull/2623, so might as well preemptively add it here too

```suggestion
      - dependency-name: "pytest"
      - dependency-name: "pytest-jubilant"
      - dependency-name: "jubilant"
```

.github/dependabot.yml
```diff
@@ -16,3 +16,45 @@ updates:
...
```
dwilding, line 41:
```suggestion
      - dependency-name: "pytest"
      - dependency-name: "pytest-jubilant"
      - dependency-name: "jubilant"
```

dwilding (approved):
Thanks a lot!

.github/dependabot.yml
tonyandrewmeyer: Hmm, good question. It's pulled in by ops, but so is websocket and we're not ignoring that. I guess remove it and see how we go.

```suggestion
```

james-garner-canonical (approved):

PR #2439, opened 2026-04-21
Security: Potential GitHub token exfiltration via unvalidated repository path in API requests

tomaioo:
## Summary

Security: Potential GitHub token exfiltration via unvalidated repository path in API requests

## Problem

**Severity**: `High` | **File**: `.github/actions/update-charm-pins/main.py:L31`

The workflow updater reads `charm-repo` values from YAML and interpolates them directly into `github.get(f'{charm_repo}/commits')` and `github.get(f'{charm_repo}/tags')`. If `charm-repo` is attacker-controlled (for example via a modified workflow in a PR), an absolute URL may bypass the intended `base_url` and cause requests to be sent to an attacker-controlled host with the configured `Authorization` header, leaking `GITHUB_TOKEN`.

## Solution

Validate `charm-repo` against a strict pattern such as `^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$` before use, reject absolute URLs, and construct URLs via safe path joining. Optionally instantiate a separate unauthenticated client for untrusted input or only add `Authorization` per-request after validation.

## Changes

- `.github/actions/update-charm-pins/main.py` (modified)

### DIFF
--- a/.github/actions/update-charm-pins/main.py
+++ b/.github/actions/update-charm-pins/main.py
@@ -4,6 +4,7 @@
 
 import logging
 import os
+import re
 import sys
 
 from httpx import Client
@@ -20,6 +21,8 @@ github = Client(
     },
 )
 
+charm_repo_pattern = re.compile(r'^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$')
+
 
 def update_charm_pins(workflow):
     """Update pinned versions of charms in the given GitHub Actions workflow."""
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
This looks like AI to me. Please see [our contribution guidelines](https://github.com/canonical/operator/blob/main/CONTRIBUTING.md#ai). If I'm incorrect, please explain.

Also: this is called with hard-coded paths, so would never be an actual security issue. We could consider adding a safety check, as proposed here, but there is definitely no current risk.

tonyandrewmeyer:
Also: please see SECURITY.md for instructions for responsibly reporting security issues (real ones, not like this).

PR #2443, opened 2026-04-22
docs: refresh the machine workloads how-to

tonyandrewmeyer:
Adjust the "how to run workloads on a machine charm" how-to guide:

* Add a HTML meta description.
* Add a small section on splitting out the workload logic from the charm logic.
* Add a section for installing the workload, with the charmlibs apt or snap packages.
* Mention the uninstall dilemma.
* Add a section with examples of unit tests for the charm (with Scenario) and the workload (with plain mocking).
* Add a section with a small Jubilant example integration test suite.
* Link out to some examples, particularly our tutorial charm, and our two machine reference charms.

[Preview](https://canonical-ubuntu-documentation-library--2443.com.readthedocs.build/ops/2443/howto/run-workloads-with-a-charm-machines/)


docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,374 @@
...
+
+Be more cautious about the package itself. If you can't be sure that you're the only consumer, leave the snap or deb installed and just stop the service you started. The cost of leaving an unused package on a machine is much smaller than the cost of breaking another charm.
+
+## Manage the service lifecycle
+
+How you start, stop, and signal the workload depends on how the package runs it:
+
+- **systemd units** (most APT packages) — use {external+charmlibs:ref}`charmlibs-systemd <charmlibs-systemd>`, or call `systemctl` as a subprocess.
+- **snap services** — use `snap.Snap.start` / `.stop` / `.restart` from the snap library.
+- **A process you launch directly** — use `subprocess.run` to start it, and send signals with `os.kill` (e.g. `SIGTERM` to stop, `SIGUSR1` to reload config). Read the workload's man page for the signals it supports.
```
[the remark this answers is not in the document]
tonyandrewmeyer:
```suggestion
- **A process you launch directly** — use `subprocess.run` to start the daemon (note that the charm process is short-lived, so the command you run should return immediately and have a daemonized process), and send signals with `os.kill` (e.g. `SIGTERM` to stop, `SIGUSR1` to reload config). Read the workload's man page for the signals it supports.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
 # How to run workloads with a machine charm
 
-There are several ways your charm might start a workload, depending on the type of charm you're authoring.
+A machine charm typically installs one or more system packages (from APT or as snaps), then starts and manages the workload as a long-running service. This how-to shows a pattern for structuring that code, installing packages, managing the service lifecycle, and testing the charm.
+
+For a complete worked example, see the [machine-tinyproxy](https://github.com/canonical/operator/tree/main/examples/machine-tinyproxy) example charm and the {ref}`machine-charm-tutorial` tutorial. For a production charm, see [ubuntu-manpages-operator](https://github.com/canonical/ubuntu-manpages-operator).
+
+## Put workload logic in its own module
+
+Keep charming concerns (event handlers, status, config parsing) in `src/charm.py`, and put workload-specific logic (installing, starting, configuring, stopping the workload) in a separate module such as `src/myworkload.py`. The charm calls the module; the module doesn't know about Ops.
```
tonyandrewmeyer, line 16:
```suggestion
Keep charming concerns (event handlers, status, config parsing) in `src/charm.py`, and put workload-specific logic (installing, starting, configuring, stopping the workload) in a separate module such as `src/myworkload.py`. The charm calls the module; the module doesn't know about Ops. If you use `charmcraft init --profile machine`, Charmcraft creates a `charm.py` and `<workload>.py` placeholder.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+    apt.remove_package("tinyproxy-bin")
 ```
 
-```{tip}
-Execute processes directly rather than via the shell.
+:class: hint
+
+Pin workload versions rather than installing the latest available package. A charm that silently upgrades between reconciliations is hard to debug, and can break if upstream introduces a breaking change.
 ```
```
tonyandrewmeyer, line 107: I'm not sure whether we consider this best practice or not. There was a discussion about this recently, I think, but I can't remember/find where 😞. Maybe we shouldn't include this (yet)?

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,374 @@
...
+
+
+def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
+    juju.deploy(charm.resolve(), app="myworkload")
+    juju.wait(jubilant.all_active, timeout=600)
+
+
+def test_workload_version(juju: jubilant.Juju):
+    version = juju.status().apps["myworkload"].version
+    assert version == "1.11.1"  # The version we pinned in install().
```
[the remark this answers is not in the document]
tonyandrewmeyer: Copilot's suggestion here isn't quite right, because it's really the version we set in the workload.py module, and implying that *Juju* is deciding what to return is misleading. However, the point about it being different than the pinned deb version is valid, though. Any suggestions for alternatives here?

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+myst:
+  html_meta:
+    description: Manage system packages (APT and snaps) and workloads in a machine charm, with unit tests using ops.testing and integration tests using Jubilant.
+---
+
 (run-workloads-with-a-charm-machines)=
 # How to run workloads with a machine charm
 
-There are several ways your charm might start a workload, depending on the type of charm you're authoring.
+A machine charm typically installs one or more system packages (from APT or as snaps), then starts and manages the workload as a long-running service. This how-to shows a pattern for structuring that code, installing packages, managing the service lifecycle, and testing the charm.
```
dwilding: "how-to" -> "guide" for consistency with other recent how-to guides

```suggestion
A machine charm typically installs one or more system packages (from APT or as snaps), then starts and manages the workload as a long-running service. This guide shows patterns for structuring that code, installing packages, managing the service lifecycle, and testing the charm.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+## Put workload logic in its own module
+
+Keep charming concerns (event handlers, status, config parsing) in `src/charm.py`, and put workload-specific logic (installing, starting, configuring, stopping the workload) in a separate module such as `src/myworkload.py`. The charm calls the module; the module doesn't know about Ops. If you use `charmcraft init --profile machine`, Charmcraft creates a `charm.py` and `<workload>.py` placeholder.
 
-For a machine charm, it is likely that packages will need to be fetched, installed and started to provide the desired charm functionality. This can be achieved by interacting with the system's package manager, ensuring that package and service status is maintained by reacting to events accordingly.
+This separation:
 
-It is important to consider which events to respond to in the context of your charm. A simple example might be:
+- Makes the workload code reusable and easy to read.
+- Lets you unit test the charm by mocking the module (no `subprocess` patching at the charm-test layer).
```
dwilding: How about this?

```suggestion
- Lets you unit test the charm by mocking the module (no `subprocess` patching in state-transition tests).
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
 # How to run workloads with a machine charm
 
-There are several ways your charm might start a workload, depending on the type of charm you're authoring.
+A machine charm typically installs one or more system packages (from APT or as snaps), then starts and manages the workload as a long-running service. This how-to shows a pattern for structuring that code, installing packages, managing the service lifecycle, and testing the charm.
+
+For a complete worked example, see the [machine-tinyproxy](https://github.com/canonical/operator/tree/main/examples/machine-tinyproxy) example charm and the {ref}`machine-charm-tutorial` tutorial. For a production charm, see [ubuntu-manpages-operator](https://github.com/canonical/ubuntu-manpages-operator).
+
+## Put workload logic in its own module
+
+Keep charming concerns (event handlers, status, config parsing) in `src/charm.py`, and put workload-specific logic (installing, starting, configuring, stopping the workload) in a separate module such as `src/myworkload.py`. The charm calls the module; the module doesn't know about Ops. If you use `charmcraft init --profile machine`, Charmcraft creates a `charm.py` and `<workload>.py` placeholder.
```
dwilding: I'd remove the last sentence and put it after the bullets (see a separate suggestion)


```suggestion
Keep charming concerns (event handlers, status, config parsing) in `src/charm.py`, and put workload-specific logic (installing, starting, configuring, stopping the workload) in a separate module such as `src/myworkload.py`. The charm calls the module; the module doesn't know about Ops.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
 
-For a machine charm, it is likely that packages will need to be fetched, installed and started to provide the desired charm functionality. This can be achieved by interacting with the system's package manager, ensuring that package and service status is maintained by reacting to events accordingly.
+This separation:
 
-It is important to consider which events to respond to in the context of your charm. A simple example might be:
+- Makes the workload code reusable and easy to read.
+- Lets you unit test the charm by mocking the module (no `subprocess` patching at the charm-test layer).
+- Lets you unit test the module on its own, patching only its direct system calls.
 
 ```python
```
dwilding, line 28: I think it would be helpful to introduce the example. How about:


````suggestion
If you use `charmcraft init --profile machine`, Charmcraft creates `charm.py` and `<workload>.py` placeholders in the `src` directory.

Below is an example of how `charm.py` calls the workload module. For more guidance, see {ref}`design-your-python-modules`.

```python
````

File: docs/howto/run-workloads-with-a-charm-machines.md [hunk 9]
```diff
@@ -1,112 +1,377 @@
...
-import subprocess
+# src/charm.py
```
dwilding:
```suggestion
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+            self.unit.set_workload_version(myworkload.get_version())
+
+    def _on_start(self, event: ops.StartEvent) -> None:
+        myworkload.start()
+
+    def _on_stop(self, event: ops.StopEvent) -> None:
+        myworkload.stop()
+
+    def _on_remove(self, event: ops.RemoveEvent) -> None:
+        # On shared machines, avoid uninstalling system packages by default.
```
dwilding: "by default" sounds a bit odd to me. How about:

```suggestion
        # On shared machines, avoid automatically uninstalling system packages.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+            event.add_status(ops.MaintenanceStatus("Starting workload"))
+        event.add_status(ops.ActiveStatus())
 ```
 
-```{tip}
-When running subprocesses, log the return (exit) code as well as `stderr` when
-errors occur.
+## Install system packages
+
+Prefer purpose-built Python libraries over shelling out to `apt-get` or `snap`. Libraries give you typed errors, idempotent operations, and avoid the pitfalls of parsing CLI output.
```
dwilding: I like the personality of "shelling out" 😁 but I think this might be more accessible:

```suggestion
Prefer purpose-built Python libraries over running `apt-get` or `snap`. Libraries give you typed errors, idempotent operations, and avoid the pitfalls of parsing CLI output.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+### APT packages
+
+Use {external+charmlibs:ref}`charmlibs-apt <charmlibs-apt>`. Add it to `pyproject.toml`:
```
dwilding, line 75: Do you think we should mention `uv add charmlibs-apt`? I suppose we might be better not to assume uv here.

docs/howto/run-workloads-with-a-charm-machines.md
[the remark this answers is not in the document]
dwilding: I think this would be sufficient:

```suggestion
    assert version == "1.11.1"  # The version we pinned in install(), as reported by the workload.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
 ```
 
-If the machine is likely to be long-running and endure multiple upgrades throughout its life, it may be prudent to ensure the package is installed more regularly, and handle the case where it needs upgrading or reinstalling. Consider this excerpt from the [ubuntu-advantage charm code](https://git.launchpad.net/charm-ubuntu-advantage/tree/src/charm.py) (with some additional comments):
+### Snap packages
+
+Use {external+charmlibs:ref}`charmlibs-snap <charmlibs-snap>`:
+
+```toml
+dependencies = [
+    "charmlibs-snap~=1.0",
```
dwilding: Matching APT

```suggestion
    "charmlibs-snap>=1,<2",
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+
+def stop() -> None:
+    snap.SnapCache()["my-workload"].stop(disable=True)
 ```
 
-In the example above, the package install status is ensured each time the charm's `config-changed` event fires, which should ensure correct state throughout the charm's deployed lifecycle.
+### When there's no library
+
+If no library is available for the workload you need, call out to the tool with `subprocess`. Keep these calls isolated in the workload module.
```
dwilding: This seems to be talking about running the workload, but we also care about installing the workload. How about this:

```suggestion
If no library is available for installing the workload, use `subprocess` to run commands that install and start the workload. Keep these calls isolated in the workload module.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+```{admonition} Best practice
+:class: hint
+
+When running subprocesses:
+- Use absolute paths (e.g. `/usr/bin/apt` rather than `apt`) to avoid PATH-based attacks.
```
dwilding: Suggestion to avoid "e.g."

```suggestion
- Use absolute paths to avoid PATH-based attacks. For example, `/usr/bin/apt` rather than `apt`.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+- Pass arguments as a list, not a shell string, so the shell doesn't interpret them.
+- Use `check=True` and `capture_output=True` so failures raise, and log the return code and `stderr` when they do.
```
dwilding: Suggestions for readability

```suggestion
- Use `check=True` and `capture_output=True` so that failures raise. In case of failure, log the return code and `stderr`.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+```
+
+### Uninstall with care on shared machines
+
+A machine charm doesn't necessarily own its machine. Another charm may have installed the same package before you did, may install it after you with `--skip-install`-style logic, or may rely on it for an unrelated purpose. Removing the package on `remove` can break those other consumers.
```
dwilding: I don't think this detail is needed, but happy if you prefer to keep it

```suggestion
A machine charm doesn't necessarily own its machine. Another charm may have installed the same package before you did, may install it after you, or may rely on it for an unrelated purpose. Removing the package on `remove` can break those other consumers.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+In your `remove` handler, always clean up the things that are unambiguously yours:
+
+- Config files and data directories your charm wrote.
+- systemd drop-ins or unit files your charm created.
+- Workload state that no other consumer would expect to find (e.g. a PID file you maintain).
```
dwilding:
```suggestion
- Workload state that no other consumer would expect to find, such as a PID file you maintain.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+Be more cautious about the package itself. If you can't be sure that you're the only consumer, leave the snap or deb installed and just stop the service you started. The cost of leaving an unused package on a machine is much smaller than the cost of breaking another charm.
```
dwilding:
```suggestion
Be more cautious about the package itself. If you can't be sure that you're the only consumer, leave the package installed and just stop the service you started. The cost of leaving an unused package on a machine is much smaller than the cost of breaking another charm.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+## Manage the service lifecycle
+
+How you start, stop, and signal the workload depends on how the package runs it:
+
+- **systemd units** (most APT packages) — use {external+charmlibs:ref}`charmlibs-systemd <charmlibs-systemd>`, or call `systemctl` as a subprocess.
+- **snap services** — use `snap.Snap.start` / `.stop` / `.restart` from the snap library.
+- **A process you launch directly** — use `subprocess.run` to start the daemon (note that the charm process is short-lived, so the command you run should return immediately and have a daemonized process), and send signals with `os.kill` (e.g. `SIGTERM` to stop, `SIGUSR1` to reload config). Read the workload's man page for the signals it supports.
```
dwilding: Suggestions for readability

```suggestion
- **A process you launch directly** — use `subprocess.run` to start the daemon. The charm process is short-lived, so the command you run should return immediately and have a daemonized process. Send signals with `os.kill` (such as `SIGTERM` to stop and `SIGUSR1` to reload config). Read the workload's man page for the signals it supports.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+For example, signalling a directly-launched process to reload its config:
+
+```python
```
dwilding, line 174:
```suggestion
```python
# src/myworkload.py
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+- `start` — start the service.
+- `config_changed` — write a new config file and signal the workload (or restart it).
+- `stop` / `remove` — stop the service and uninstall packages.
+- `collect_unit_status` — report status to Juju at the end of every hook, based on the current state of the workload.
+
+For a longer-lived charm that may be upgraded in place, also observe `upgrade_charm` and re-run the install and config steps so that packages and config stay in sync with the charm revision.
+
+## Write unit tests
+
+Unit tests for a machine charm come in two layers, matching the module split above. Together they cover the whole charm without ever installing the real package.
```
dwilding:
```suggestion
Unit tests for a machine charm come in two layers, matching the charm's modules. Together they cover the whole charm without ever installing the real package.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+    workload.installed = True
+    workload.running = True
+    ctx = testing.Context(MyCharm)
+    ctx.run(ctx.on.stop(), testing.State())
+    assert not workload.is_running()
+```
+
+### Tests for the workload module
+
+Test the module directly by patching the things it actually calls — `apt`, `subprocess.run`, `os.kill`, the snap cache. Keep these tests small: they exist to check that the module invokes its dependencies correctly, not to test those dependencies.
```
dwilding:
```suggestion
Test the module directly by patching the things it actually calls — `apt`, `subprocess.run`, `os.kill`, the snap cache, and so on. Keep these tests small: they exist to check that the module invokes its dependencies correctly, not to test those dependencies.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+    commands: list[list[str]] = []
+    monkeypatch.setattr(
+        "subprocess.run",
+        lambda cmd, **kwargs: commands.append(cmd) or None,
+    )
+    myworkload.start()
+    assert commands == [["myworkload"]]
+```
+
+Run all your unit tests with:
```
dwilding, line 333: For easier discoverability

```suggestion
### Run the tests

Run all your unit tests with:
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+    assert version == "1.11.1"  # The version we pinned in install().
+
+
+def test_blocks_on_invalid_config(juju: jubilant.Juju):
+    juju.config("myworkload", {"slug": "not/valid"})
+    juju.wait(jubilant.all_blocked)
+    juju.config("myworkload", reset="slug")
+```
+
+The `juju` fixture from `pytest-jubilant` creates a temporary model per test file and tears it down afterwards. You supply a `charm` fixture that locates the packed `.charm` file — see the [machine-tinyproxy integration `conftest`](https://github.com/canonical/operator/blob/main/examples/machine-tinyproxy/tests/integration/conftest.py) for a reusable template.
```
dwilding: I think we should mention the Charmcraft profile here. How about:

```suggestion
The `juju` fixture from `pytest-jubilant` creates a temporary model per test file and tears it down afterwards. You supply a `charm` fixture that locates the packed `.charm` file. For an example, see [`conftest.py` in machine-tinyproxy's integration tests](https://github.com/canonical/operator/blob/main/examples/machine-tinyproxy/tests/integration/conftest.py).

If you use `charmcraft init --profile machine`, Charmcraft creates a `charm` fixture and placeholder files for your tests.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+
+For how to pack the charm, set up a virtual machine to run the tests, and configure CI, see {ref}`write-integration-tests-for-a-charm` and {ref}`set-up-ci`.
```
dwilding: I don't think it's necessary to have all the intro text. How about:

```suggestion
For guidance on running the tests, see:

- {ref}`write-integration-tests-for-a-charm`
- {ref}`set-up-ci`
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,377 @@
...
+- [ubuntu-manpages-operator](https://github.com/canonical/ubuntu-manpages-operator) — a production machine charm. See its [`tests/unit`](https://github.com/canonical/ubuntu-manpages-operator/tree/main/tests/unit) for a real-world example of the test patterns above.
+- [openstack-exporter-operator](https://github.com/canonical/openstack-exporter-operator) — a production machine charm that installs its workload as a snap. The workload module is in [`src/service.py`](https://github.com/canonical/openstack-exporter-operator/blob/main/src/service.py).
+
+## See also
+
+- {ref}`machine-charm-tutorial` — build a machine charm from scratch.
+- {ref}`write-unit-tests-for-a-charm` — more on `ops.testing`.
+- {ref}`write-integration-tests-for-a-charm` — more on Jubilant.
+- {external+charmlibs:ref}`charmlibs-apt <charmlibs-apt>`, {external+charmlibs:ref}`charmlibs-snap <charmlibs-snap>`, {external+charmlibs:ref}`charmlibs-systemd <charmlibs-systemd>`, and {external+charmlibs:ref}`charmlibs-pathops <charmlibs-pathops>`.
```
dwilding: I don't think it's necessary to have the first three links, as we already have them in discoverable places. But I think it's especially helpful to provide the lib links in one place. So how about this:

```suggestion
- {external+charmlibs:ref}`charmlibs-apt <charmlibs-apt>`
- {external+charmlibs:ref}`charmlibs-snap <charmlibs-snap>`
- {external+charmlibs:ref}`charmlibs-systemd <charmlibs-systemd>`
- {external+charmlibs:ref}`charmlibs-pathops <charmlibs-pathops>`
```

dwilding (approved):
Thanks for spending time on this. It's a massive improvement over what we currently have!

docs/howto/run-workloads-with-a-charm-machines.md
tonyandrewmeyer: I don't like "running" because I don't think it's clear enough that it's you doing a subprocess call - the library is "running" them too. But shelling out is also bad because we don't actually do it through the shell.

What about this?


```suggestion
Prefer purpose-built Python libraries over subprocess calls to `apt-get` or `snap`. Libraries give you typed errors, idempotent operations, and avoid the pitfalls of parsing CLI output.
```

docs/howto/run-workloads-with-a-charm-machines.md
tonyandrewmeyer: Yeah, I'm not sure how much we want to push uv vs. let people pick other things. I think I would prefer to leave it as-is in this how-to.

docs/howto/run-workloads-with-a-charm-machines.md
tonyandrewmeyer: What about this?

```
- Use `check=True` so that failures raise -- generally you will want to catch that exception and log the return code and possibly `stderr`. Use `capture_output=True` so that output from the command doesn't leak into the Juju log.
```

docs/howto/run-workloads-with-a-charm-machines.md
dwilding: "subprocess calls" sounds great

docs/howto/run-workloads-with-a-charm-machines.md
dwilding: How about these tiny adjustments?

> - Use `check=True` so that failures raise. Generally you will want to catch that exception, then log the return code and possibly `stderr`. Use `capture_output=True` so that output from the command doesn't leak into the Juju log.

tonyandrewmeyer:
```suggestion
- Use `check=True` so that failures raise. Generally you will want to catch that exception, then log the return code and possibly `stderr`. Use `capture_output=True` so that output from the command doesn't leak into the Juju log.
```

docs/howto/run-workloads-with-a-charm-machines.md
james-garner-canonical: This seems reasonable as-is to me.

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,389 @@
...
+- Workload state that no other consumer would expect to find, such as a PID file you maintain.
+
+Be more cautious about the package itself. If you can't be sure that you're the only consumer, leave the package installed and just stop the service you started. The cost of leaving an unused package on a machine is much smaller than the cost of breaking another charm.
+
+## Manage the service lifecycle
+
+How you start, stop, and signal the workload depends on how the package runs it:
+
+- **systemd units** (most APT packages) — use {external+charmlibs:ref}`charmlibs-systemd <charmlibs-systemd>`, or call `systemctl` as a subprocess.
+- **snap services** — use `snap.Snap.start` / `.stop` / `.restart` from the snap library.
```
james-garner-canonical: Maybe we can write this in a way that persists across Snap library versions.
```suggestion
- **snap services** — use `start`, `stop`, and `restart` methods of the {external+charmlibs:ref}`charmlibs-snap <charmlibs-snap>` library.
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,389 @@
...
+- **A process you launch directly** — use `subprocess.run` to start the daemon. The charm process is short-lived, so the command you run should return immediately and have a daemonized process. Send signals with `os.kill` (such as `SIGTERM` to stop and `SIGUSR1` to reload config). Read the workload's man page for the signals it supports.
```
james-garner-canonical: Should we be recommending `subprocess.Popen` with args like `stdout/err/in=subprocess.DEVNULL` and `start_new_session=True` for starting long-running processes that don't take care of launching a separate daemon themselves?

Should we give explicit advice for how to get/store the daemon's PID? e.g. ideally the workload writes it to a known path, otherwise you should do so yourself.

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,389 @@
...
+
+## Observe the right events
+
+For a long-running workload, the core lifecycle is:
+
+- `install` — install packages and set the workload version.
+- `start` — start the service.
+- `config_changed` — write a new config file and signal the workload (or restart it).
+- `stop` / `remove` — stop the service and uninstall packages.
+- `collect_unit_status` — report status to Juju at the end of every hook, based on the current state of the workload.
```
james-garner-canonical: I'm OK with having this in the list, and we definitely should recommend using it, but it feels odd listing it alongside the actual Juju lifecycle events, doesn't it? Also makes the term 'lifecycle' ambiguous to me -- are we talking about `__init__` -> `collect_unit_status` or `install` -> `remove`?

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,389 @@
...
+
+Run all your unit tests with:
+
+```text
+tox -e unit
+```
+
+For more on state-transition testing — including `State.from_context`, reusing state across events, and accessing the charm instance — see {ref}`write-unit-tests-for-a-charm`.
+
+## Write integration tests
```
james-garner-canonical, line 376: Should we mention 'functional' style tests as well (in the sense used in `charmlibs`) -- non-Juju integration tests? These are typically a lot faster than Juju tests while giving you coverage of interactions with the real workload.

james-garner-canonical (approved):

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -1,112 +1,389 @@
...
+
+
+def reload_config() -> None:
+    pid = int(PID_FILE.read_text())
+    os.kill(pid, signal.SIGUSR1)
+```
+
+## Observe the right events
+
+For a long-running workload, the core lifecycle is:
```
tonyandrewmeyer:
```suggestion
For a long-running workload, the core events are:
```

docs/howto/run-workloads-with-a-charm-machines.md
tonyandrewmeyer: Lifecycle is too ambiguous a word, yes, so I've removed that. I think maybe this should be removed, because as you say it's not really like the others and it's more a general concern than a machine charm one.


```suggestion
```

docs/howto/run-workloads-with-a-charm-machines.md
tonyandrewmeyer: Good suggestion, yes. I'll add a section for that.

docs/howto/run-workloads-with-a-charm-machines.md
tonyandrewmeyer: I'm not sure. I somewhat feel like our guidance should be "make a snap" if there isn't one, and then just use the snap instructions. But I felt that was maybe too aggressive and so included this fairly basic bit about not using snaps or debs. I'm not sure how much detail we want to go into - do we consider this a reasonable way to manage workloads?

james-garner-canonical: I like the idea of recommending making a snap for production use if there isn't one. Then we can frame launching a process directly as being for prototyping, and then I think it's fine as-is.

docs/howto/run-workloads-with-a-charm-machines.md
james-garner-canonical: Agree on removing the word lifecycle. I do like mentioning `collect_unit_status` when we talk about events, since repetition helps. But fine as-is now.

PR #2449, opened 2026-04-24
docs: add how-to guide for naming and initialising a charm

dwilding:
This PR adds a new how-to guide called "How to initialise your project". I'm doing this for two reasons:

1. Our [current guidance](https://documentation.ubuntu.com/ops/latest/howto/write-and-structure-charm-code/#create-a-repository-and-initialise-it) is not so easy to find. This topic deserves to be discoverable from the Ops homepage and the list of how-to guides.

2. In the charming docs generally, it could be clearer how developers are supposed to name their repos and charms. See [charmcraft#2582](https://github.com/canonical/charmcraft/issues/2582). I'm taking this opportunity to improve our naming guidelines.

**[Preview of new how-to guide](https://canonical-ubuntu-documentation-library--2449.com.readthedocs.build/ops/2449/howto/initialise-your-project/)**

Also updated:

- [How to manage charms > Initialise your charm project](https://canonical-ubuntu-documentation-library--2449.com.readthedocs.build/ops/2449/howto/manage-charms/#initialise-your-charm-project)
- [How to write and structure charm code](https://canonical-ubuntu-documentation-library--2449.com.readthedocs.build/ops/2449/howto/write-and-structure-charm-code/) - Removed the current guidance and rewrote the intro

### DIFF
--- a/docs/howto/initialise-your-project.md
+++ b/docs/howto/initialise-your-project.md
@@ -62,14 +62,7 @@ Important files to be aware of:
 - `pyproject.toml` - Python project configuration, including the dependencies of your charm.
 - `src/charm.py` - The Python file that will contain the main logic of your charm.
 
-(alternative-project-structures)=
-## Alternative project structures
-
-If your repository will hold multiple charms, or a charm and source for another artifact, such as a [rock](https://ubuntu.com/containers/docs), create a `charms` directory at the repository root. Then create a subdirectory for each charm and run `charmcraft init` in each subdirectory.
-
 If your charm's workload was built with a web framework such as Django or FastAPI, consider using one of Charmcraft's 12-factor app profiles instead of `kubernetes` or `machine`. These profiles accelerate development by generating charms that are ready to deploy. {external+charmcraft:ref}`Write your first 12-factor app charm <tutorial>`.
-
-<!-- Can we link to a real example of a charm monorepo? -->
 
 ## Next steps
 
--- a/docs/howto/initialise-your-project.md
+++ b/docs/howto/initialise-your-project.md
@@ -69,7 +69,7 @@ If your repository will hold multiple charms, or a charm and source for another 
 
 If your charm's workload was built with a web framework such as Django or FastAPI, consider using one of Charmcraft's 12-factor app profiles instead of `kubernetes` or `machine`. These profiles accelerate development by generating charms that are ready to deploy. {external+charmcraft:ref}`Write your first 12-factor app charm <tutorial>`.
 
-<!-- TODO: Can we link to a real example of a charm monorepo? -->
+<!-- Can we link to a real example of a charm monorepo? -->
 
 ## Next steps
 
--- a/docs/howto/initialise-your-project.md
+++ b/docs/howto/initialise-your-project.md
@@ -1,0 +1,77 @@
+---
+myst:
+  html_meta:
+    description: Decide your charm's name and use Charmcraft to generate the recommended project structure.
+---
+
+(init-charm)=
+# How to initialise your project
+
+Before you initialise your project, install charm development tools. See {ref}`prepare-your-environment`.
+
+This guide demonstrates how to name your charm and generate the recommended project structure. You should decide your charm's name before you create a repository for your charm.
+
+(decide-your-charms-name)=
+## Decide your charm's name
+
+Your charm's name should be based on the name of the workload and should be short and memorable. Examples:
+
+- `mega-calendar` for a machine charm that operates a workload called Mega Calendar
+- `mega-calendar-k8s` for a Kubernetes charm that operates the same workload
+
+To check that your preferred name isn't already in use, visit the (future) Charmhub URL, which should be an error page:
+
+```text
+https://charmhub.io/mega-calendar-k8s
+```
+(create-a-repository-and-initialise-it)=
+## Create a repository and initialise it
+
+Create a repository with your source control of choice.
+
+```{admonition} Best practice
+:class: hint
+
+Name the repository using the pattern ``<charm name>-operator`` for a single
+charm, or ``<base charm name>-operators`` when the repository will hold
+multiple related charms. For the charm name, see [](#decide-your-charms-name).
+```
+
+For example, name the repository `mega-calendar-k8s-operator` if your charm will be called `mega-calendar-k8s`.
+
+Next, use {external+charmcraft:doc}`Charmcraft <index>` to generate the recommended project structure in the repository:
+
+```text
+charmcraft init --name mega-calendar-k8s --profile kubernetes
+```
+
+Or for a machine charm:
+
+```text
+charmcraft init --name mega-calendar --profile machine
+```
+
+<!-- Mention that --name is optional? (Omitting it can cause confusion) -->
+
+Charmcraft's `kubernetes` and `machine` profiles are minimal charms that contain placeholder code and configuration.
+These profiles also give you starting points for unit tests and integration tests.
+
+Important files to be aware of:
+
+- `charmcraft.yaml` - Metadata about your charm.
+- `pyproject.toml` - Python project configuration, including the dependencies of your charm.
+- `src/charm.py` - The Python file that will contain the main logic of your charm.
+
+(alternative-project-structures)=
+## Alternative project structures
+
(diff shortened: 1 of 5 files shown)


docs/howto/initialise-your-project.md
```diff
@@ -0,0 +1,77 @@
...
+charmcraft init --name mega-calendar-k8s --profile kubernetes
+```
+
+Or for a machine charm:
+
+```text
+charmcraft init --name mega-calendar --profile machine
+```
+
+<!-- Mention that --name is optional? (Omitting it can cause confusion) -->
```
dwilding: I'd like opinions on this approach.

In our [current guidance](https://documentation.ubuntu.com/ops/latest/howto/write-and-structure-charm-code/#create-a-repository-and-initialise-it) we don't mention `--name`. But here's the problem: If you create your repo following the naming conventions, then init your charm, your charm will have the wrong name. I think it's better for us to tell charmers to explicitly set the right charm name.

Should we mention any of this in the doc?

docs/howto/initialise-your-project.md
```diff
@@ -0,0 +1,77 @@
...
+- `src/charm.py` - The Python file that will contain the main logic of your charm.
+
+(alternative-project-structures)=
+## Alternative project structures
+
+If your repository will hold multiple charms, or a charm and source for another artifact, such as a [rock](https://ubuntu.com/containers/docs), create a `charms` directory at the repository root. Then create a subdirectory for each charm and run `charmcraft init` in each subdirectory.
+
+If your charm's workload was built with a web framework such as Django or FastAPI, consider using one of Charmcraft's 12-factor app profiles instead of `kubernetes` or `machine`. These profiles accelerate development by generating charms that are ready to deploy. {external+charmcraft:ref}`Write your first 12-factor app charm <tutorial>`.
+
+<!-- Can we link to a real example of a charm monorepo? -->
```
dwilding: Our [current guidance](https://documentation.ubuntu.com/ops/latest/howto/write-and-structure-charm-code/#create-a-repository-and-initialise-it) includes a directory tree (which is not so easy to interpret). I don't think we need to show a tree, but it would be nice to link to a real charm monorepo. Any suggestions?

docs/howto/initialise-your-project.md
james-garner-canonical: Why will your charm have the wrong name? Because of the `-operator` suffix?

I think recommending the use of the `--name` option is a nice idea, since adding fancy name transformation logic to the charmcraft profiles isn't on our to do list any time soon IIRC.

dwilding: > Why will your charm have the wrong name? Because of the -operator suffix?

Exactly

docs/howto/initialise-your-project.md
```diff
@@ -0,0 +1,77 @@
+---
+myst:
+  html_meta:
+    description: Decide your charm's name and use Charmcraft to generate the recommended project structure.
+---
+
+(init-charm)=
+# How to initialise your project
```
dwilding, line 8: Other ideas:
- How to initialise your charm project
- How to initialise your charm

docs/howto/initialise-your-project.md
james-garner-canonical: This is tricky because the core problems with charm monorepos are still unsolved. I'm not sure if there are any good examples we can link to that comply with current recommendations. That may change with https://github.com/canonical/charmcraft/issues/2573

I do think we should drop the current tree example though, and probably our current sentence of advice.

docs/howto/initialise-your-project.md
```diff
@@ -0,0 +1,77 @@
...
+:class: hint
+
+Name the repository using the pattern ``<charm name>-operator`` for a single
+charm, or ``<base charm name>-operators`` when the repository will hold
+multiple related charms. For the charm name, see [](#decide-your-charms-name).
+```
+
+For example, name the repository `mega-calendar-k8s-operator` if your charm will be called `mega-calendar-k8s`.
+
+Next, use {external+charmcraft:doc}`Charmcraft <index>` to generate the recommended project structure in the repository:
```
dwilding, line 42: I'm not telling people to use Charmcraft from a commit hash. But maybe we should? What do you think @james-garner-canonical?

docs/howto/initialise-your-project.md
james-garner-canonical: I'm definitely not opposed to adding a note explaining that `--name` is optional in general, but required here to avoid the `-operator` suffix being part of the name.

docs/howto/initialise-your-project.md
```diff
@@ -0,0 +1,77 @@
...
+Important files to be aware of:
+
+- `charmcraft.yaml` - Metadata about your charm.
+- `pyproject.toml` - Python project configuration, including the dependencies of your charm.
+- `src/charm.py` - The Python file that will contain the main logic of your charm.
+
+(alternative-project-structures)=
+## Alternative project structures
+
+If your repository will hold multiple charms, or a charm and source for another artifact, such as a [rock](https://ubuntu.com/containers/docs), create a `charms` directory at the repository root. Then create a subdirectory for each charm and run `charmcraft init` in each subdirectory.
```
james-garner-canonical: This is our current recommendation, but it doesn't really make sense to me, and I don't think it follows current practice. I think we should drop it entirely for now. The 12-factor part could then just sit at the end of the previous section rather than needing a section of its own.

docs/howto/initialise-your-project.md
james-garner-canonical: It seems reasonable to duplicate the current tip in a third place here since we're explicitly talking about `charmcraft init`, and the other docs may be confusing if you're not using up-to-date profiles, and we want people using our latest recommendations anyway.

james-garner-canonical: Maybe we should update your charmcraft profiles tooling repo to explain how to find the commits in our docs and when they should be updated (after charmcraft PR merge).

dwilding: What do you think about a scheduled workflow in Ops that checks:

1. That we use the same commit hash in all places (although, we could single-source this in the docs)
2. That initing charms based on that commit hash gives the same result as initing charms based on main

james-garner-canonical: Single sourcing the hash in the docs would be nice, how would that work?

I don't know about the scheduled workflow, maybe for now we should just stick with it as a follow-up task when updating the profiles.

dwilding: > Single sourcing the hash in the docs would be nice, how would that work?

I'll try adding it to this PR and we can see how we like it.

> I don't know about the scheduled workflow, maybe for now we should just stick with it as a follow-up task when updating the profiles.

Yeah, that's fair. I think my time would be better spent properly documenting the full process for updating Charmcraft profiles including downstream actions.

docs/howto/initialise-your-project.md
tromai: `How to initialise your project` sounds okay to me.

I personally prefer mentioning of `project` in the title because `initialise a charm` can be mistaken as initialising a Charm during run-time.

docs/howto/initialise-your-project.md
tromai: I think it is useful to mention why we use `--name` in this section.

It certainly confused me a little bit the first time I used `charmcraft init`, given the default behavior of `--name`.

docs/howto/initialise-your-project.md
dwilding: James and I had some internal discussion. We're going to change the approach of the tutorials to include only one `charmcraft init` instruction (the `uvx ...` one). I've directly included the tip here.

docs/howto/initialise-your-project.md
dwilding: Sounds good - I've dropped the part about repo structure. As I was restructuring, I figured it's best to put the 12-factor part at the very end of the doc, as a "next step" of learning about 12-factor. For the main doc, I think we can reasonably assume that someone doesn't want a 12-factor charm.

docs/howto/initialise-your-project.md
dwilding: Thanks for your feedback! I've added a couple of sentences about this:

> If you don't specify `--name` when running `charmcraft init`, Charmcraft uses the parent directory name for your charm. For example, `mega-calendar-k8s-operator`, from the name of the repository. So we recommend specifying `--name` to ensure that your charm's name doesn't end with `-operator`.

james-garner-canonical (approved):

tromai (approved):
Thank you. It looks good to me.

PR #2450, opened 2026-04-24
ci: add 26.04 to smoke tests and remove outdated metadata.yaml

tonyandrewmeyer:
A recent Juju 2.9 [broke our smoke tests for 2.9](https://github.com/juju/juju/issues/21562). That issue has since been fixed with a later 2.9 release, but one of the reasons it broke was the way that we're using the metadata.yaml file.

This PR removes the `metadata.yaml` file in favour of just using `charmcraft.yaml` as expected in modern Charmcraft.

Now that Ubuntu 26.04 has been released, that is also added to the set of tests, except that it's currently not supported in Charmcraft, so an xfail marker is added (which means we'll notice when it does get support and can adjust appropriately).


tromai (approved):
Thanks for the fix. It looks great to me.

james-garner-canonical (approved):
Dropping `metadata.yaml` seems reasonable.

I like the use of `xfail` here.

PR #2420, opened 2026-04-11
docs: add a how-to guide about publishing a charm

dwilding:
This PR adds a how-to guide to fill a gap in the Ops docs: We talk about how [make a charm discoverable](https://documentation.ubuntu.com/ops/latest/howto/make-your-charm-discoverable/) on Charmhub, but not how to publish to Charmhub in the first place. I want to make sure that our homepage and how-to guides have a clear entrypoint for this topic, even if we are linking out to Charmcraft for the specifics about how to publish.

**[Preview of new doc](https://canonical-ubuntu-documentation-library--2420.com.readthedocs.build/ops/2420/howto/publish-your-charm-on-charmhub/)**

PR #2451, opened 2026-04-25
docs: explain how to add an icon to a charm

### LINKED ISSUE #2385: Maybe we should document the charm icon
Our docs contain a mix of charmcraft plugins: `charm`, `uv`, and `dump`

These behave differently wrt. the magic `icon.svg` file that's shown on charm hub: https://github.com/canonical/charmcraft/issues/2593

gcomneno:
## Summary

Add a note to the charm management docs clarifying that `icon.svg` handling depends on the Charmcraft plugin or profile in use.

The note calls out that modern charms in this repository typically use the `uv` plugin, while older charms may use `dump`, so the behavior is not identical in every case.

## Testing

- `make -C docs html`
- `make -C docs vale TARGET='docs/howto/manage-charms.md'`
- `make -C docs spelling TARGET='docs/howto/manage-charms.md'`


dwilding:
Thanks for the addition @gcomneno! Just a heads up - I won't review or merge this until I've finished https://github.com/canonical/operator/pull/2420, which affects where we might put advice about icons.

gcomneno:
Thanks for the heads up! That makes sense. I’ll leave this as-is for now and wait for #2420 to settle before making any further changes.

tonyandrewmeyer (commented):
This is also not quite correct. If you're using `uv` for the charm plugin, then you *need* to use the `dump` plugin to include the icon file. I think that is the case for `poetry` and `python` too, but I'm not 100% sure on that. I think the old `charm` plugin is the only one that pulls in the icon by default.

I think we would also want to have an example of what the YAML would look like to add the icon.

gcomneno:
Thanks Tony, that's helpful. I agree the current wording is too broad.

I'll hold off on revising this until #2420 settles where this advice should live, but the next version should distinguish the old charm plugin from uv/poetry/python-based projects and include a concrete YAML example using the dump plugin to include the icon file.

dwilding:
This PR is also relevant for https://github.com/canonical/charmcraft/issues/2593

dwilding:
@gcomneno I revisited this PR to think about the best place to mention icon handling.

I hope you don't mind, I directly pushed to your branch with my preference for where to mention this. Please pull your branch to get the changes.

I've switched to a dedicated section called "Review metadata and add an icon" in [How to publish your charm on Charmhub](https://documentation.ubuntu.com/ops/latest/howto/publish-your-charm-on-charmhub/). I've also changed the positioning:

- Assume that people are using a modern plugin that requires Dump (with an example).
- Mention the Charm plugin after the example, but steer people away from the Charm plugin.

tonyandrewmeyer (approved):
Thanks!

I think the Charmhub page should also explain that the icon doesn't automatically get added if not using the old plug-in. @dwilding do we already have plans to update there too?

dwilding:
> I think the Charmhub page should also explain that the icon doesn't automatically get added if not using the old plug-in. @dwilding do we already have plans to update there too?

Good idea. I'll coordinate with the other authors. For now I've tracked this internally.

dwilding (approved):
@gcomneno I'll merge if this version looks good to you. You can see a preview at https://canonical-ubuntu-documentation-library--2451.com.readthedocs.build/ops/2451/howto/publish-your-charm-on-charmhub/#review-metadata-and-add-an-icon

gcomneno:
Thanks Dave, this version looks good to me. The placement on the Charmhub publishing page makes sense, and the wording is clearer than my original version.

Happy for you to merge it.

PR #2453, opened 2026-04-25
docs: add git workflow guidance for contributors

### LINKED ISSUE #2395: Add more git workflow details to CONTRIBUTING.md
From the 26.04 survey:

If you could document the expected git workflow for "flesh and bone" external contributors in the CONTRIBUTING.md file that would be great. I've found through several contributors that I often spend more time fighting with git than I do actually crafting the code for the contribution since I use my preferred workflow rather than the charm tech team's preferred workflow. Not a huge issue, but I spent like 2 hours trying to figure out how to use `git reflog` when I contributed the systemd library to charm libs. I would've preferred to have had contributor documentation that was like "no force pushes please we squash merge PRs at the end."

Here's how we communicate our expected git workflow in the HPC team: https://github.com/charmed-hpc/.github/blob/main/CONTRIBUTING.md#pull-requests

---

We should do this across all of the Charm Tech CONTRIBUTING.md documents (except perhaps Pebble?), or in pull request templates, or some other discoverable way.

gcomneno:
## Summary

Add a short git workflow section to `CONTRIBUTING.md` for external contributors.

The new guidance explains that contributors should work on branches in their forks, rebase before asking for review, avoid force-pushing after review starts when possible, and prefer keeping the pull request easy to review over maintaining a perfect branch history. It also clarifies that pull requests are squash-merged.

## Testing

- `make -C docs vale TARGET='CONTRIBUTING.md'`
- `make -C docs spelling TARGET='CONTRIBUTING.md'`

Fixes #2395


dwilding:
Thanks for contributing this @gcomneno!

I think the info you added is helpful, but I'd prefer if the section requires less reading for people to get the idea. I've pushed to your branch to put the most important points (in my opinion) up front and generally reduce the amount of reading needed. I hope that's OK. You'll need to pull changes.

If this new structure doesn't work for you as a contributor, let's revise it further! Thanks again :slightly_smiling_face:

gcomneno:
Thanks Dave, this structure works for me as a contributor. Putting the branch/update guidance before the detailed PR title section makes the key points easier to find.

CONTRIBUTING.md
```diff
@@ -14,6 +14,19 @@ Please do not use tools (such as GitHub Copilot) to provide PR reviews. The Char
 
 Changes are proposed as [pull requests on GitHub](https://github.com/canonical/operator/pulls).
 
+- Work on a branch in your own fork.
+- Sequence your commits logically if possible. But don't worry too much -- we'll squash to `main` after review.
+- Don't force-push after review has started.
+- Follow conventional commit style for the PR title (not required for individual commits).
```
tonyandrewmeyer: I'm not sure we should have this here and also ten lines down. It looks like the intention here is to move from the existing paragraphs to bullets. If that's the case, then let's pull in the note about scopes, drop the list of types (or smush it down into a comma-separated list), and remove the title section.

tonyandrewmeyer (approved):
The [CPC one](https://github.com/charmed-hpc/.github/blob/main/CONTRIBUTING.md#pull-requests) has a lot more details (example commands and so on) but I think we're ok with this more minimal approach, since people can look up the commands, or, more likely, are using either an agent or some IDE functionality to do the actual work.

CONTRIBUTING.md
dwilding: I think it's OK to drop the full list of types, but I like the examples. How does it look now, after https://github.com/canonical/operator/pull/2453/commits/4a29db64bdae1514085770a10f6b6488667e9ff3?

gcomneno:
Looks good to me. The shorter PR title guidance keeps the useful examples while avoiding duplication, and the branch update advice is still clear from a contributor point of view.

dwilding (approved):
Thanks again @gcomneno for your contribution & feedback!

PR #2454, opened 2026-04-25
fix: treat remote unit zero as explicit

### LINKED ISSUE #2341: state transition testing relation events don't like remote unit zero
This test:

```py
    ctx = testing.Context(DummyCharm, meta=META)
    secret = testing.Secret(tracked_content=BAD_SECRET_DATA)
    rel = testing.Relation(
        "rel", "intr",
        remote_app_data={"secret-uri": f'"secret://{secret.id}"'},
        remote_units_data={0: {}},
    )

    state_in = testing.State(relations={rel}, secrets={secret})
    state_out = ctx.run(ctx.on.relation_changed(relation=rel, remote_unit=0), state_in)
```

Yields ⚠️ `WARNING  ops-scenario.runtime.consistency_checker:_consistency_checker.py:122 This scenario is probably inconsistent. Double check, and ignore this warning if you're sure. The following warnings were found: 'rel_relation_changed' is implicitly using 0 as the remote unit. Consider passing `remote_unit` explicitly.`

If I change the remote unit to `1` in both `remote_units_data=...` and `remote_unit=...` the warning disappears.

I think this means we're confusing the explicit remote unit `1` with the implicit/default.

Given that unit 0 is legal in Juju, we should fix this.

gcomneno:
Fixes #2341.

This updates the Scenario consistency checker to distinguish between a missing remote unit id and an explicit remote unit id of `0`.

`0` is a valid Juju unit id, so the checker should not treat it as falsy/absent.

Changes:
- check `relation_remote_unit_id is None` instead of relying on truthiness
- add a regression test for `relation_remote_unit_id=0`

Validation:
- `uv tool run --with tox-uv tox -e unit -- testing/tests/test_consistency_checker.py -k "relation_not_in_state or remote_unit_zero" -rs -vv --tb=short`
- `uv tool run --with tox-uv tox -e unit -- testing/tests/test_consistency_checker.py -rs`
- `ruff check testing/src/scenario/_consistency_checker.py testing/tests/test_consistency_checker.py`

### DIFF
--- a/testing/tests/test_consistency_checker.py
+++ b/testing/tests/test_consistency_checker.py
@@ -219,19 +219,17 @@ def test_evt_bad_container_name():
     ],
 )
 def test_checkinfo_matches_layer(check: CheckInfo, consistent: bool):
-    layer = ops.pebble.Layer(
-        {
-            'checks': {
-                'chk1': {
-                    'override': 'replace',
-                    'level': 'ready',
-                    'startup': 'disabled',
-                    'threshold': 42,
-                    'http': {'url': 'http://localhost:5000/version'},
-                }
+    layer = ops.pebble.Layer({
+        'checks': {
+            'chk1': {
+                'override': 'replace',
+                'level': 'ready',
+                'startup': 'disabled',
+                'threshold': 42,
+                'http': {'url': 'http://localhost:5000/version'},
             }
         }
-    )
+    })
     state = State(containers={Container('foo', check_infos={check}, layers={'base': layer})})
     asserter = assert_consistent if consistent else assert_inconsistent
     asserter(
--- a/testing/tests/test_consistency_checker.py
+++ b/testing/tests/test_consistency_checker.py
@@ -488,6 +488,15 @@ def test_relation_changed_remote_unit_zero_is_explicit():
     )
 
 
+def test_relation_changed_without_available_remote_unit_is_inconsistent():
+    relation = Relation('foo', remote_units_data={})
+    assert_inconsistent(
+        State(relations={relation}),
+        _Event('foo_relation_changed', relation=relation),
+        _CharmSpec(MyCharm, {'requires': {'foo': {'interface': 'bar'}}}),
+    )
+
+
 def test_action_not_in_meta_inconsistent():
     ctx = Context(MyCharm, meta={'name': 'foo'}, actions={'foo': {}})
     assert_inconsistent(
--- a/testing/tests/test_consistency_checker.py
+++ b/testing/tests/test_consistency_checker.py
@@ -219,17 +219,19 @@ def test_evt_bad_container_name():
     ],
 )
 def test_checkinfo_matches_layer(check: CheckInfo, consistent: bool):
-    layer = ops.pebble.Layer({
-        'checks': {
-            'chk1': {
-                'override': 'replace',
-                'level': 'ready',
-                'startup': 'disabled',
-                'threshold': 42,
-                'http': {'url': 'http://localhost:5000/version'},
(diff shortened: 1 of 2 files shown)


gcomneno:
Hi! Just gently following up on this one.

I rebased the branch onto current `upstream/main` and re-ran the focused checks locally:

- `PYTHONPATH="$PWD:$PWD/testing/src" uv run --with pytest --with websocket-client python -m pytest -q testing/tests/test_consistency_checker.py -rs`
- `uv run ruff check testing/src/scenario/_consistency_checker.py testing/tests/test_consistency_checker.py`
- `uv run ruff format --check testing/src/scenario/_consistency_checker.py testing/tests/test_consistency_checker.py`

The focused test run passes with `67 passed, 6 skipped`, and Ruff is clean.

Happy to adjust anything else if useful.

tonyandrewmeyer (approved):
Apologies for the delay here; the last 3-4 weeks have been very busy with other things.

The fix looks correct to me -- zero is indeed a valid ID and we were wrongly checking false-iness.

Thanks!

tonyandrewmeyer:
@copilot please fix the linting (run `tox -e format` and verify `tox -e lint` shows no issues).

testing/tests/test_consistency_checker.py
```diff
@@ -219,17 +219,19 @@ def test_evt_bad_container_name():
...
+                    'override': 'replace',
+                    'level': 'ready',
+                    'startup': 'disabled',
+                    'threshold': 42,
+                    'http': {'url': 'http://localhost:5000/version'},
+                }
             }
         }
-    })
+    )
```
james-garner-canonical: Spurious change?

testing/tests/test_consistency_checker.py
```diff
@@ -479,6 +481,15 @@ def test_relation_not_in_state():
...
 
+def test_relation_changed_remote_unit_zero_is_explicit():
+    relation = Relation('foo', remote_units_data={0: {}})
+    assert_consistent(
+        State(relations={relation}),
+        _Event('foo_relation_changed', relation=relation, relation_remote_unit_id=0),
+        _CharmSpec(MyCharm, {'requires': {'foo': {'interface': 'bar'}}}),
+    )
+
+
```
james-garner-canonical, line 490: Is there still an implicit remote unit zero case that should raise a consistency checker error? Can we add a test for it?

testing/tests/test_consistency_checker.py
tonyandrewmeyer: I think perhaps this is why the linter is unhappy, but for some reason Copilot has not followed up on my request to fix it, and I was too lazy to check out the branch and do it myself.

@gcomneno can you run `tox -e format` and then `tox -e lint`? You should get the same result as in CI (the `format` should fix whatever it is). Generally we shouldn't end up with style changes like the above.

gcomneno:
Thanks, good catch.

I ran `tox -e format` and `tox -e lint` via `uv tool run --with tox-uv`; both pass now.

For the extra negative coverage: `Relation('foo')` defaults to remote unit 0, so omitting `remote_unit` in that case only produces the existing implicit-remote-unit warning. I added the actual error-path test with `remote_units_data={}`, which verifies that a relation-changed event without an explicit remote unit is inconsistent when no remote unit is available.

I also verified the focused tests with:

`tox -e py3.12-unit -- testing/tests/test_consistency_checker.py -k "remote_unit_zero or available_remote_unit" -rN`

james-garner-canonical (approved):

james-garner-canonical:
Thanks @gcomneno!

PR #2455, opened 2026-04-27
docs: add doc about subordinate charms

dwilding:
This PR adds an explanation doc about subordinate charms. My intention is to give a summary of what a subordinate charm is, when it's appropriate to write a subordinate charm, and some of the considerations. This isn't meant to be a detailed guide on how to write a subordinate charm.

**[Preview doc](https://canonical-ubuntu-documentation-library--2455.com.readthedocs.build/ops/2455/explanation/subordinate-charms/)**

### DIFF
--- a/docs/explanation/subordinate-charms.md
+++ b/docs/explanation/subordinate-charms.md
@@ -9,8 +9,6 @@ myst:
 
 A "subordinate" charm is a machine charm where each unit runs alongside a unit of another charm, called the "principal" charm. In this setup, the principal unit and corresponding subordinate unit always run on the same machine.
 
-A subordinate charm typically communicates with the principal charm out of band. There isn't a dedicated Juju relation between a principal unit and the corresponding subordinate unit.
-
 Subordinate charms aren't always appropriate. If possible, instead of writing a subordinate charm, write a regular charm that communicates over relations.
 
 ## When to use a subordinate charm
@@ -29,18 +27,12 @@ A subordinate charm is declared in `charmcraft.yaml`:
 subordinate: true
 ```
 
-`charmcraft.yaml` doesn't specify a principal charm. A subordinate charm doesn't know which principal charm it will be deployed alongside. A Juju user is free to deploy a subordinate charm alongside any principal charm; your subordinate charm's code needs to decide whether it's running in the correct context.
+`charmcraft.yaml` doesn't specify a principal charm. A subordinate charm doesn't know which principal charm it will be deployed alongside. A Juju user is free to deploy a subordinate charm alongside any principal charm; your subordinate charm's code needs to determine whether it's running in the correct context.
 
 When you document a subordinate charm, clearly state the intended principal charm.
 
-```{important}
+## Runtime constraints
+
 A subordinate charm can't control the Ubuntu base that its units run on. If the principal charm supports multiple bases, consider publishing a separate revision of the subordinate charm for each base.
-```
 
-## After deployment
-
-TODO: What actually gets put on the machine? If I have two regular applications in my model, do I automatically get a unit of the subordinate for each unit of each regular application? So there isn't really a notion of "the" principal charm?
-
-TODO: Watch out - other subordinate charms might be configuring the same machine.
-
-TODO: What data can subordinate units see? Is there a peer relation with other units of the same subordinate charm? Review the relation how-to in Ops - "Note however that subordinate units cannot see each other's peer data".
+A subordinate unit shouldn't assume it's the only unit trying to configure the hardware or system. Subordinate units should minimise side effects when writing configuration and installing/uninstalling packages. For general advice, see {ref}`run-workloads-with-a-charm-machines`.
--- a/docs/explanation/subordinate-charms.md
+++ b/docs/explanation/subordinate-charms.md
@@ -1,0 +1,46 @@
+---
+myst:
+  html_meta:
+    description: A subordinate charm runs alongside another charm, typically for managing hardware or system interactions.
+---
+
+(subordinate-charms)=
+# Subordinate charms
+
+A "subordinate" charm is a machine charm where each unit runs alongside a unit of another charm, called the "principal" charm. In this setup, the principal unit and corresponding subordinate unit always run on the same machine.
+
+A subordinate charm typically communicates with the principal charm out of band. There isn't a dedicated Juju relation between a principal unit and the corresponding subordinate unit.
+
+Subordinate charms aren't always appropriate. If possible, instead of writing a subordinate charm, write a regular charm that communicates over relations.
+
+## When to use a subordinate charm
+
+A subordinate charm is appropriate if you need to monitor or configure the underlying hardware. For example, [`hardware-observer`](https://github.com/canonical/hardware-observer-operator).
+
+A subordinate charm is also appropriate if you need to manage how a particular workload interacts with the system. For example, how data is backed up or how connections are pooled. See [`pgbouncer`](https://github.com/canonical/pgbouncer-operator).
+
+In general, a subordinate charm should be lighter weight than its intended principal charm.
+
+## Declaring a subordinate charm
+
+A subordinate charm is declared in `charmcraft.yaml`:
+
+```yaml
+subordinate: true
+```
+
+`charmcraft.yaml` doesn't specify a principal charm. A subordinate charm doesn't know which principal charm it will be deployed alongside. A Juju user is free to deploy a subordinate charm alongside any principal charm; your subordinate charm's code needs to decide whether it's running in the correct context.
+
+When you document a subordinate charm, clearly state the intended principal charm.
+
+```{important}
+A subordinate charm can't control the Ubuntu base that its units run on. If the principal charm supports multiple bases, consider publishing a separate revision of the subordinate charm for each base.
+```
+
+## After deployment
+
+TODO: What actually gets put on the machine? If I have two regular applications in my model, do I automatically get a unit of the subordinate for each unit of each regular application? So there isn't really a notion of "the" principal charm?
+
+TODO: Watch out - other subordinate charms might be configuring the same machine.
+
+TODO: What data can subordinate units see? Is there a peer relation with other units of the same subordinate charm? Review the relation how-to in Ops - "Note however that subordinate units cannot see each other's peer data".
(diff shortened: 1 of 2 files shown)


docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,46 @@
...
+A subordinate charm can't control the Ubuntu base that its units run on. If the principal charm supports multiple bases, consider publishing a separate revision of the subordinate charm for each base.
+```
+
+## After deployment
+
+TODO: What actually gets put on the machine? If I have two regular applications in my model, do I automatically get a unit of the subordinate for each unit of each regular application? So there isn't really a notion of "the" principal charm?
+
+TODO: Watch out - other subordinate charms might be configuring the same machine.
+
+TODO: What data can subordinate units see? Is there a peer relation with other units of the same subordinate charm? Review the relation how-to in Ops - "Note however that subordinate units cannot see each other's peer data".
```
dwilding: @tonyandrewmeyer & @james-garner-canonical do you have any thoughts about these questions? I'm hoping to provide answers in the doc. But if we're unsure what to say, let's leave these out of scope for now.

docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,46 @@
...
+  html_meta:
+    description: A subordinate charm runs alongside another charm, typically for managing hardware or system interactions.
+---
+
+(subordinate-charms)=
+# Subordinate charms
+
+A "subordinate" charm is a machine charm where each unit runs alongside a unit of another charm, called the "principal" charm. In this setup, the principal unit and corresponding subordinate unit always run on the same machine.
+
+A subordinate charm typically communicates with the principal charm out of band. There isn't a dedicated Juju relation between a principal unit and the corresponding subordinate unit.
```
tonyandrewmeyer: FYI, we should probably mention `juju-info` somewhere here.

docs/explanation/subordinate-charms.md
tonyandrewmeyer: For the "watch out" bit, I think we don't really have much to offer, apart from warning people. Perhaps the "if you're not sure that it's only yours leave it alone" advice, but I don't think we have much else.

Maybe here is where the `juju-info` relation gets mentioned? I think it somewhat answers the last question. I think it's missing from the main Juju docs (or the search just doesn't find it), so maybe that should also get fixed so that we have an upstream to link to. I believe it's how the subordinate should get information about the primary.

Subordinates are very much a weak spot for me, so I would need to verify things. But I think the idea is that a subordinate works with the primary rather than with other instances of itself, and that's what the peer relation bit is about.

For the first question, I'm not sure, would have to try it out, sorry.

dwilding: We discussed this internally today and decided to remove the first and third TODOs. I've kept the "watch out" TODO and worked it into the doc. I'll open a separate issue for the removed questions, including whether we mention `juju-info`.

I also removed the part of the doc that talked about there being no dedicated Juju relation.

dwilding:
[Draft doc](https://canonical-ubuntu-documentation-library--2455.com.readthedocs.build/ops/2455/explanation/subordinate-charms/) updated per our discussion at daily. Ready for review - thanks

docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,38 @@
...
+
+Subordinate charms aren't always appropriate. If possible, instead of writing a subordinate charm, write a regular charm that communicates over relations.
+
+## When to use a subordinate charm
+
+A subordinate charm is appropriate if you need to monitor or configure the underlying hardware. For example, [`hardware-observer`](https://github.com/canonical/hardware-observer-operator).
+
+A subordinate charm is also appropriate if you need to manage how a particular workload interacts with the system. For example, how data is backed up or how connections are pooled. See [`pgbouncer`](https://github.com/canonical/pgbouncer-operator).
+
+In general, a subordinate charm should be lighter weight than its intended principal charm.
```
james-garner-canonical: Is this really true? And are we measuring the weight of the charm or the thing the charm manages?

docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,38 @@
...
+subordinate: true
+```
+
+`charmcraft.yaml` doesn't specify a principal charm. A subordinate charm doesn't know which principal charm it will be deployed alongside. A Juju user is free to deploy a subordinate charm alongside any principal charm; your subordinate charm's code needs to determine whether it's running in the correct context.
+
+When you document a subordinate charm, clearly state the intended principal charm.
+
+## Runtime constraints
+
+A subordinate charm can't control the Ubuntu base that its units run on. If the principal charm supports multiple bases, consider publishing a separate revision of the subordinate charm for each base.
```
james-garner-canonical: Talking about "the" principal charm like this seems wrong, since the subordinate doesn't know what principal it is deployed with.

I'm still not clear why a subordinate charm can't control the base that it runs on. I assume it still has to declare its bases in `charmcraft.yaml`. Does Juju just ignore it?

docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,38 @@
...
+In general, a subordinate charm should be lighter weight than its intended principal charm.
+
+## Declaring a subordinate charm
+
+A subordinate charm is declared in `charmcraft.yaml`:
+
+```yaml
+subordinate: true
+```
+
```
tonyandrewmeyer, line 27: The [juju docs](https://documentation.ubuntu.com/juju/latest/reference/juju-cli/list-of-juju-cli-commands/integrate/#subordinate-applications) say that:

> They must define a container scoped endpoint

Maybe we should mention this? Or maybe it's covered by the follow-up issue?

docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,38 @@
...
+`charmcraft.yaml` doesn't specify a principal charm. A subordinate charm doesn't know which principal charm it will be deployed alongside. A Juju user is free to deploy a subordinate charm alongside any principal charm; your subordinate charm's code needs to determine whether it's running in the correct context.
+
+When you document a subordinate charm, clearly state the intended principal charm.
```
tonyandrewmeyer: I feel like this too strongly indicates that a subordinate is designed for a specific principal, whereas it might be that it's a log collector for example, which will work with a lot of different principals. Can we adjust it so that it's more saying that you need to clearly state what types of charm it works with? Or more generally the purpose and if specific charms are required?

docs/explanation/subordinate-charms.md
tonyandrewmeyer: I believe it will error if the charm isn't available for the base the principal is delayed on.

Maybe this would be better as encouragement to support as many bases as possible? Something about the base having to match the principal, so it's good to handle as many LTS bases as feasible.

tonyandrewmeyer (approved):
I think we should link to the Juju docs, at least https://documentation.ubuntu.com/juju/latest/reference/charm/#subordinate-charm and maybe https://documentation.ubuntu.com/juju/latest/reference/relation/#subordinate-relation.

I think what we really need is a how-to for writing a subordinate that goes alongside this explanation. But happy to leave that for another cycle.

A few general comments noted as well. I'm approving now so I don't block things while travelling. I assume you and James can narrow down the bits left, including the comments I've left and whether they should be addressed or not.

docs/explanation/subordinate-charms.md
james-garner-canonical: > I believe it will error if the charm isn't available for the base the principal is delayed on.

Like if you try to deploy a regular charm on a machine with the wrong base?

I think what we're getting at here is that with a principal charm, you can pretty easily say you run on only one specific base, and only your direct users care. Whereas a subordinate charm may have interest from users of different principal charms across different bases.

But TBH I don't think this difference is all that substantial -- in either case you have to find out what your users need and what you can afford to support.

> Maybe this would be better as encouragement to support as many bases as possible? Something about the base having to match the principal, so it's good to handle as many LTS bases as feasible.

What I want to push back against (until I understand it further, at least) is the idea that subordinate charms need to support arbitrary bases.

Sure, someone might want to deploy you on Ubuntu 20.04 with their principal charm that runs there, and it would be a shame if they couldn't, but the same is true of principal charms. I just don't see the difference between "there's a user who wants to deploy our charm on an old base", and "there's a user who wants to deploy our charm on an old base alongside another charm".

It seems like developers of subordinate charms should be able to be just as clear about their supported bases and support periods as any other charm. But maybe there's something I'm missing here.

docs/explanation/subordinate-charms.md
dwilding: I don't think the sentence added much explanatory value. I've removed it.

docs/explanation/subordinate-charms.md
dwilding: I've covered this and linked to the extra info you've prepared in https://github.com/juju/juju/pull/22373. Link is a comment until that merges.

docs/explanation/subordinate-charms.md
dwilding: I've changed the wording:

> When you document a subordinate charm, clearly state which charms your subordinate charm is intended to be used with and how they should be integrated.

docs/explanation/subordinate-charms.md
dwilding: James and I discussed this today. Our thinking is that for now it's best to focus on the importance of documenting the subordinate charm's support for bases - at least until we work on https://github.com/canonical/operator/issues/2460.

Updated wording:

> A subordinate charm can’t control the Ubuntu base that its units run on. When you document which charms your subordinate charm is intended to be used with, mention any differences in supported bases.

docs/explanation/subordinate-charms.md
```diff
@@ -0,0 +1,43 @@
...
+```yaml
+subordinate: true
+```
+
+`charmcraft.yaml` doesn't specify a principal charm. A subordinate charm doesn't know which principal charm it will be deployed alongside.
+
+Instead of specifying a principal charm, you define an endpoint with `scope: container`. The endpoint can use an application-specific interface or the generic `juju-info` interface.
+
+<!-- TODO when ref target is available (from https://github.com/juju/juju/pull/22373) -->
+<!-- See {external+juju:ref}`Juju | The implicit juju-info relation <the-implicit-juju-info-relation-endpoint>`. -->
```
dwilding: Uncomment this before merging

james-garner-canonical (approved):
Looks great, thanks @dwilding.

PR #2457, opened 2026-04-28
docs: pin to 24.04 when launching VMs

dwilding:
By default, Multipass now creates 26.04 VMs. But when packing on 26.04 I get an error:

```
charmcraft internal error: ValueError("'26.04' is not a valid BuilddBaseAlias")
```

This PR updates our docs to (hopefully temporarily) pin to 24.04 VMs.


dwilding:
FYI @lengau. What's the latest on 26.04 support in Charmcraft?

tromai (approved):
Thank you. Pinning it sounds good to me.

tonyandrewmeyer (approved):
Works as a short-term solution. It's probably reasonable to do until we have tested 26.04 on the tutorial end to end, or at least until we have the integration tests doing so.

PR #2468, opened 2026-05-04
ci: add workflow_dispatch trigger to workflows where it is useful

james-garner-canonical:
We already have `workflow_dispatch` triggers on many workflows, which is very useful when something goes wrong (especially externally), and we want to validate how the workflow currently runs on `main`. This PR adds `workflow_dispatch` triggers to workflows that didn't have this trigger, where this would also be useful. This excludes workflows like publishing (which runs on pushing tags), and workflows that are only called by others.

---

This PR is motivated by the [db test failure](https://github.com/canonical/operator/actions/runs/25240130398/job/74014364316?pr=2466) in a seemingly unrelated PR. The last run on `main` was a success, but that was a few days ago, so I would have liked to run the tests on `main`  manually as a quick sanity check.

### DIFF
--- a/.github/workflows/charmcraft-pack.yaml
+++ b/.github/workflows/charmcraft-pack.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
 
 permissions: {}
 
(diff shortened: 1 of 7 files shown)


tonyandrewmeyer (approved):
Seems reasonable to me. These are all "check something" (even the oddly named "docs") rather than make/change/do something, so opening up how they can be triggered seems ok.

I don't think it should be necessary very often, vs re-running in a PR, but it also doesn't seem like it adds much noise or other issues.

dwilding (approved):

PR #2470, opened 2026-05-06
docs: improve how-to about managing files in containers

dwilding:
This PR is a complete rewrite of [How to manage files in the workload container](https://documentation.ubuntu.com/ops/latest/howto/manage-containers/manage-files-in-the-workload-container/). I recommend reviewing the preview doc instead of trying to review the diff.

**[Preview doc](https://canonical-ubuntu-documentation-library--2470.com.readthedocs.build/ops/2470/howto/manage-containers/manage-files-in-the-workload-container/)**

Main improvements:

- Switch to pathops where possible
- Focus the examples, making them consistent and not workload-specific
- Focus and modernise the unit tests


docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
+    description: Use the pathops library and ops.Container methods to manage files in a charm's workload container.
+---
+
 (files-in-containers)=
 # How to manage files in the workload container
 
-Pebble's files API allows charm authors to read and write files on the workload container. You can write files ("push"), read files ("pull"), list files in a directory, make directories, and delete files or directories.
+The [](ops.Container) class provides methods for managing files in a container.
+
+Instead of using `ops.Container` directly, we recommend using the {external+charmlibs:ref}`pathops <charmlibs-pathops>` library. `pathops` provides a `ContainerPath` class that uses a `pathlib`-like approach for managing files in a container.
```
tonyandrewmeyer: I don't think we'd want to do it across all the examples, but I think just for this first mention, it would be worth linking to the Python docs (an intersphinx link for "pathlib").

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
+
+To write a file to the workload container, use {external+charmlibs:meth}`ContainerPath.write_text <pathops.ContainerPath.write_text>` or {external+charmlibs:meth}`ContainerPath.write_bytes <pathops.ContainerPath.write_bytes>`:
+
+```python
+config = '...'
+(self.myapp_root / 'config.yaml').write_text(config)
+```
 
-There are many additional features, including the ability to send raw bytes (by providing a Python `bytes` object as the second argument) and write data from a file-like object. You can also specify permissions and the user and group for the file. See the [API documentation](ops.Container.push) for details.
+`pathops` also has a function {external+charmlibs:func}`ensure_contents <pathops.ensure_contents>` that ensures a file exists with the given contents:
```
tonyandrewmeyer: Shouldn't this be "with the given content"? (Which makes me wonder why we didn't call the method `ensure_content`.)

More generally, I feel like we should explain this without basically just repeating the name. Maybe saying that it checks the content of the file, and creates or updates it if it doesn't match?

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
+
+```python
+class MyCharm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        self.container = self.unit.get_container('myapp')
+        self.myapp_root = pathops.ContainerPath(
+            '/etc/myapp',
+            container=self.container,
+        )
```
james-garner-canonical:
```suggestion
        self.myapp_root = pathops.ContainerPath('/etc/myapp', container=self.container)
```

docs/howto/manage-containers/manage-files-in-the-workload-container.md
james-garner-canonical:
```suggestion
`pathops` also has a function {external+charmlibs:func}`ensure_contents <pathops.ensure_contents>` that ensures a file exists with the given contents, ownership, and permissions, making no changes if the conditions are already satisfied:
```

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
 A trailing "/*" on the source directory is the only supported globbing/matching.
 
 ## List files
 
-To list the contents of a directory or return stat-like information about one or more files, use [`Container.list_files`](ops.Container.list_files). It returns a list of [`pebble.FileInfo`](ops.pebble.FileInfo) objects for each entry (file or directory) in the given path, optionally filtered by a glob pattern. For example:
+To iterate over a directory, use {external+charmlibs:meth}`ContainerPath.iterdir <pathops.ContainerPath.iterdir>` or {external+charmlibs:meth}`ContainerPath.glob <pathops.ContainerPath.glob>`:
+
+```python
+paths = self.myapp_root.iterdir()
+yaml_paths = self.myapp_root.glob('*.yaml')
```
james-garner-canonical: Since these methods give an iterator:
```suggestion
paths = list(self.myapp_root.iterdir())
for yaml_path in self.myapp_root.glob('*.yaml'):
    # ...
```

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
+```
+
+Alternatively, to list the contents of a directory or return stat-like information about one or more files, use [`Container.list_files`](ops.Container.list_files). This method returns a list of [`pebble.FileInfo`](ops.pebble.FileInfo) objects for each entry (file or directory) in the given path, optionally filtered by a glob pattern. For example:
```
james-garner-canonical, line 120: It's annoying that `pathops` doesn't provide a unified way of getting info for local and container paths. The utilities are all implemented and used internally (e.g. in `ensure_path`), but not exposed publicly. This means users have to write their own substrate agnostic helpers for cases like this, using `path.stat` for local paths, and `container.list_files` for container paths. But probably not worth adding this to the `pathops` API without a user feature request.

james-garner-canonical (approved):
Looks great, thanks!

docs/howto/manage-containers/manage-files-in-the-workload-container.md
dwilding: Good suggestion, thanks. I've put "making no changes" after the example for ease of reading.

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
+self.container.push_path('/source/dir/*', '/destination')
 ```
 
 A trailing "/*" on the source directory is the only supported globbing/matching.
 
-## Pull recursive
+### From the container
 
-To copy several files to the workload, use [`Container.pull_path`](ops.Container.pull_path), which copies files recursively into a specified destination directory.  The API docs contain detailed examples of source and destination semantics and path handling.
+To copy several files from the workload container, use [`Container.pull_path`](ops.Container.pull_path), which copies files recursively into a specified destination directory. The API docs contain detailed examples of source and destination semantics and path handling.
```
tonyandrewmeyer, line 98: Eeek, that's fixing a rather ugly mistake!

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -1,209 +1,213 @@
...
+    assert config_file.exists()
+    assert my_custom_checks(config_file)
+```
 
-    def _on_pebble_ready(self, _):
-        foo = self.unit.get_container('foo')
-        foo.push('/local/share/config.yaml', 'TEST', make_dirs=True)
+`get_filesystem` returns a temporary directory that simulates the container's filesystem.
 
+Don't write to the filesystem that `get_filesystem` returns. If a unit test needs to ensure that particular data exists in the container, use [`Container.mounts`](ops.testing.Container.mounts):
```
tonyandrewmeyer, line 188: I'm a bit hesitant on how strong this is (I do remember we talked about it before the PR was opened). Mounts are definitely better if you're providing the content of an existing file. If it's something more dynamic, then writing to the (temporary) filesystem isn't that bad, and maybe cleaner than putting up a temporary file(like) object. But I think on balance I'm ok with this and seeing if we get any pushback.

tonyandrewmeyer (approved):
I'm so sorry for being so slow with this! I had pending comments from 3 weeks ago, and then failed to either submit a WIP review, or to complete it during the sprints, and then failed to get to it last week as well :(

This does feel like a bit of a step-change: rather than just saying "you can also do this via a library", we're explicitly saying to use the library and *not* use the Ops functionality. It's similar to saying that people should use the snap (etc) libraries instead of subprocess, but it feels further given that there's no Ops-native way to do it. On balance, I think this is fine -- it goes towards the "charming with ops" / "centre of charming excellence" pattern that we're working towards, beyond just what Ops offers. I think we will need to be careful about when we choose to do this, particularly if it's using a library that we don't own/maintain ourselves.

It does feel very smooth that you click through to reference links that happen to be in charmlibs and not here. Given the consistent Canonical styling, you barely notice that you're somewhere else. I think in practice this will be fine, because people aren't browsing here, they're opening the reference page to dig specifically into that. I think we should try to figure out if people do get confused and don't understand that they've moved sites and can't continue to browse around ops from following that link (if so, maybe we need signals that you're switching).

Overall, this is a much nicer doc with instructions that are far easier to follow. Thanks for all the work on it (and again sorry for being so slow).

docs/howto/manage-containers/manage-files-in-the-workload-container.md
dwilding: I improved this part after feedback from James. Latest preview: http://localhost:8080/_build/howto/manage-containers/manage-files-in-the-workload-container/#write-a-file

I dunno about contents vs content. A bit like "data is" vs "data are", saying contents sounds more natural to me. Also I suppose we're locked in with the method name.

PR #2476, opened 2026-05-14
fix: silence 331 ty issues

lengau:
This removes 331 diagnostic issues when running `ty check`

It came from running `ty check --add-ignore`. Good luck!


tonyandrewmeyer:
I'm not sure we really want to sprinkle all these ignores everywhere. I think we'd want to try to figure out if they could be resolved, or if ty is not yet mature enough for us to us.

james-garner-canonical:
I suggest we close this PR for now, but it's a useful point of reference, thanks @lengau!

I see a couple of systematic issues at a glance (unresolved imports for Sphinx, so we need to configure `ty` to know about or ignore those files; and it seems to have trouble with a lot of our overloads and inheritance). If we do want to migrate to `ty` in future, I'd suggest we take this as a baseline and resolve some of these systematic issues before adding it to our CI (maybe we could add a local entrypoint a manual workflow for easy reference?).

PR #2477, opened 2026-05-15
chore: use consistent name for ops.testing.Context when used as a context manager

### LINKED ISSUE #2458: Use consistent name for ops.testing.Context when used as a context manager
We should be consistent about whether to write `with context(...) as mgr` or `with context(...) as manager`, especially in our docs.

```
> ag 'as manager' | wc -l
368
> ag 'as mgr' | wc -l
2760
```

tromai:
Fixes #2458 .

I picked `mgr` as the target because `mgr` is used more widely in the repo.

`as manager` were mainly used in `./testing`.

```
x grep -r -F "as manager" ./
./testing/src/scenario/context.py:        with ctx(ctx.on.start(), State()) as manager:
./testing/src/scenario/context.py:            with ctx(ctx.on.start(), State()) as manager:
./testing/UPGRADING.md:with ctx(ctx.on.start(), State()) as manager:
./testing/tests/test_e2e/test_manager.py:    with Manager(ctx, ctx.on.start(), State()) as manager:
./testing/tests/test_e2e/test_manager.py:    with Manager(ctx, ctx.on.start(), State()) as manager:
./testing/tests/test_e2e/test_manager.py:    with Manager(ctx, ctx.on.start(), State()) as manager:
./testing/tests/test_e2e/test_manager.py:    with ctx(ctx.on.start(), State()) as manager:
./testing/tests/test_e2e/test_manager.py:    with ctx(ctx.on.action('do-x'), State()) as manager:
./testing/tests/test_runtime.py:        ) as manager:
./testing/tests/test_runtime.py:    ) as manager:
./testing/tests/test_runtime.py:        ) as manager:
```

I don't have a strong opinion on any of them, and am happy to change if need.


dwilding (approved):
Changes look good to me, thanks.

As I was reviewing, I noticed some small doc updates that would be good too. But feel free to leave those to a separate PR if you don't think they fit here.

1. In [State transition testing](https://documentation.ubuntu.com/ops/latest/explanation/state-transition-testing/) we say:

    > The `Context` object can be used as a context manager for this use case. Note that you can't call `manager.run()` multiple times ...

    But we don't show any code with a context manager. I think it would be helpful to insert a very short code snippet (with `mgr` instead of `manager`).

2. [How to manage libraries](https://documentation.ubuntu.com/ops/latest/howto/manage-libraries/) has an unusual pattern in one place. It defines a `context()` fixture. I haven't seen this anywhere else. I think we should switch to the usual `ctx = Context(...)` pattern. Should be a quick change.

tromai:
@dwilding Thanks for the review and suggestion. I am happy with putting those changes in this PR. Please let me know if it looks okay.
- In [State transition testing](https://documentation.ubuntu.com/ops/latest/explanation/state-transition-testing/): I used a small modified example, instead of a full one. I'm not sure if this would confuse the readers.
- [How to manage libraries](https://documentation.ubuntu.com/ops/latest/howto/manage-libraries/): the fixture was removed, and `ctx` is created within the test function.

docs/explanation/state-transition-testing.md
```diff
@@ -195,6 +195,19 @@ that ensures that `ops.main` 'pauses' right before emitting the event to hand
...
+
+```python
+@pytest.mark.parametrize('leader', (True, False))
+def test_status_leader(leader):
+    ctx = testing.Context(MyCharm, meta={'name': 'foo'})
+    with ctx(ctx.on.start(), testing.State(leader=leader)) as mgr:
+        assert mgr.charm.unit.status == testing.UnknownStatus()
+        mgr.run()
+        assert mgr.charm.unit.status == testing.ActiveStatus('I rule' if leader else 'I am ruled')
+```
```
dwilding: Nice idea to show how to adapt one of the existing examples. I think it would be good to split the explanation with the example, then we can refer to `mgr.run()` after the example has introduced it. Plus a couple of very minor suggestions:

- There a different examples above, so I'd probably say "a previous example"
- Adjusted the quotes (the double backticks would be needed in docstrings but not in the .md docs)

````suggestion
The `Context` object can be used as a context manager for this use case.
Here's a modified version of a previous example:

```python
@pytest.mark.parametrize('leader', (True, False))
def test_status_leader(leader):
    ctx = testing.Context(MyCharm, meta={'name': 'foo'})
    with ctx(ctx.on.start(), testing.State(leader=leader)) as mgr:
        assert mgr.charm.unit.status == testing.UnknownStatus()
        mgr.run()
        assert mgr.charm.unit.status == testing.ActiveStatus('I rule' if leader else 'I am ruled')
```

Before `mgr.run()`, no event is emitted, hence the 'unknown' status.

Note that you can't call `mgr.run()` multiple times: the object is a context
that ensures that `ops.main` 'pauses' right before emitting the event to hand
you some introspection hooks, but for the rest this is a regular test: you can't
emit multiple events in a single charm execution.
````

**Review by dwilding — APPROVED** [review]

File: docs/explanation/state-transition-testing.md (hunk 1 above)
Comment (dwilding) [reply docs/explanation/state-transition-testing.md]: Oh, one more suggestion while we're here: I'd drop "Note that" from "Note that you can't call `mgr.run()` multiple times". It's fine to start the sentence with "You can't `mgr.run()` multiple times" (I usually recommend dropping "Note that" because I don't think it adds much)

Comment (james-garner-canonical) [reply docs/explanation/state-transition-testing.md]: I do find the text around this example a bit confusing TBH. It's not directly in scope for this PR, but if we are rewriting it now, how about this for before the example:

```md
Using the `Context` object as a context manager (instead of calling `Context.run` directly) produces a [](ops.testing.Manager) object with access to the charm object instantiated for this event.

Let's rewrite `test_status_leader` to make assertions about the charm instance rather than the state:
```

And then after the example:
```md
Before calling [](ops.testing.Manager.run), the charm object was initialised based on the context's state. The state didn't specify a status for the unit, so at this point it's `unknown`. The charm's `_on_start` method runs and sets the status when `mgr.run()` is called.

Calling `mgr.run()` multiple times is an error, as the context manager represents a single event and charm instance.
```

I'd also be tempted to add a note that this is bad example of when to use the context manager form, since everything we're asserting on is visible in the state ... maybe we should update the example instead though. We could also make it clearer what you need to do in the context manager scope (call `run()`) vs what you can call outside. And also how to get the output state.

```py
import pytest


class MyCharm(ops.CharmBase):
    msg = ''

        super().__init__(framework)

    def _on_start(self, _):
        if self.unit.is_leader():
            self.msg = 'I rule'
        else:
            self.msg = 'I am ruled'
        self.unit.status = ops.ActiveStatus(self.msg)


@pytest.mark.parametrize('leader', (True, False))
def test_status_leader(leader):
    ctx = testing.Context(MyCharm, meta={'name': 'foo'})
    with ctx(ctx.on.start(), testing.State(leader=leader)) as mgr:
        charm = mgr.charm
        assert charm.msg == ''
        assert charm.unit.status == testing.UnknownStatus()
        state_out = mgr.run()
    msg = 'I rule' if leader else 'I am ruled'
    assert charm.msg = msg
    assert charm.unit.status == testing.ActiveStatus(msg)
    assert state_out.unit_status == charm.unit.status
```

james-garner-canonical (approved):
Thanks for tackling this one! Approving as this is a nice win regardless of other docs improvements -- but some suggestions building on @dwilding's for the example in the docs.

docs/explanation/state-transition-testing.md
james-garner-canonical: While we're at it, the opening paragraph for this section could be clearer too

> The testing framework is a black-box, state-transition testing framework. It
makes it trivial to assert that a status went from A to B, but not to assert
that, in the context of this charm execution, with this state, a certain
charm-internal method was called and returned a given piece of data, or would
return this and that _if_ it had been called.

How about:

```md
A typical state-transition test calls [`Context.run`](ops.testing.Context.run) to produce the [`State`](ops.testing.State) after some event has been handled, and makes assertions about that state. However, sometimes we'd also like to make assertions about the Python charm object that handled the event.
```

tromai: Thanks @james-garner-canonical and @dwilding . I really like the suggestions. I added them in https://github.com/canonical/operator/pull/2477/commits/e60d0b47fa9e3ee4daee7eb6e7cb51ad69b3f14d:
- Rewrite the example to emphasize on asserting charm's state.
- Refer to the example as `test_status_leader`, which also indicates it being a previous example.
- Rewrite the text following James's suggestion, which removed the "Note that" phrase.
- Use the example to split the text
- Use single backticks

dwilding: The new doc structure looks great to me, thanks!

docs/explanation/state-transition-testing.md
```diff
@@ -183,17 +183,54 @@ execution context.
...
+    ctx = testing.Context(MyCharm, meta={'name': 'foo'})
+    with ctx(ctx.on.start(), testing.State(leader=leader)) as mgr:
+        charm = mgr.charm
+        assert charm.msg == ''
+        assert charm.unit.status == testing.UnknownStatus()
+        state_out = mgr.run()
+    msg = 'I rule' if leader else 'I am ruled'
+    assert charm.msg == msg
+    assert charm.unit.status == testing.ActiveStatus(msg)
+    assert state_out.unit_status == charm.unit.status
```
dwilding, line 226: One question, more for my education: Does it matter whether these lines are inside/outside the context manager?

PR #2478, opened 2026-05-16
docs: add COS Lite integration tests to K8s tutorial

### LINKED ISSUE #1838: K8s tutorial: add integration tests for COS Lite relations
Per [Tony's comment](https://github.com/canonical/operator/pull/1734/files/98c7b20d862eebadfe41db03679ffeb4800cbe71#r2151137612), we should expand [Observe your charm with COS Lite](https://ops.readthedocs.io/en/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.html) to include integration tests for the relations added in the chapter.

This could be done before or after switching to Jubilant (which is planned as part of #1749).

dwilding:
This PR follows on from https://github.com/canonical/operator/pull/2411 and resolves https://github.com/canonical/operator/issues/1838.

I'm adding a new section [Write integration tests (preview)](https://canonical-ubuntu-documentation-library--2478.com.readthedocs.build/ops/2478/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite/#write-integration-tests) to the last chapter of the K8s tutorial. The tests in this section match what was added to the example charm in the previous PR.

Other changes:

- Adjusted the part where we talk about adding `cosl` to dependencies. It wasn't clear why 1.9.1 should be added based on the versions required by libs, so I'm explicitly saying to add the latest version of the package. [Preview](https://canonical-ubuntu-documentation-library--2478.com.readthedocs.build/ops/2478/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite/#add-dependencies-from-libraries)

- During earlier testing, I found that the postgres deployment can cause the integration tests to time out. We don't want to specify a timeout in that test, so I've added some info to the tutorial chapter ([Preview](https://canonical-ubuntu-documentation-library--2478.com.readthedocs.build/ops/2478/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#write-an-integration-test)):

    > If the test fails with a timeout error, override the default timeout in test_database_integration:
    > ...

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
@@ -539,18 +539,22 @@ def test_loki_data(cos: jubilant.Juju):
     results = json.loads(task.results["proxied-endpoints"])
     loki_url = results["loki/0"]["url"]
     loki_api_url = f"{loki_url}/loki/api/v1/label/juju_application/values"
+    juju_applications = _get_loki_logs(loki_api_url)
+    assert juju_applications is not None, "No logs available from Loki"
+    assert APP_NAME in juju_applications
 
-    # Wait for logs to be available from Loki, then check for our app.
-    for _ in range(2 * 60):
+
+def _get_loki_logs(loki_api_url: str) -> list[str] | None:
+    """Wait for logs to be available from Loki and return them."""
+    for attempt in range(60):
+        if attempt:  # If not the first attempt, wait before retrying.
+            time.sleep(1)
         response = requests.get(loki_api_url)
         if response.status_code == 200:
             response_decoded = response.json()
             if "data" in response_decoded:
-                juju_applications: list[str] = response_decoded["data"]
-                assert APP_NAME in juju_applications
-                return
-        time.sleep(1)
-    pytest.fail("No logs available from Loki")
+                return response_decoded["data"]
+    return None
 ```
 
 ### Run the tests
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
@@ -511,6 +511,8 @@ We're now ready to write the tests.
 
 ### Deploy COS Lite and integrate with Loki
 
+Add two test functions to `tests/integration/test_charm.py`:
+
 ```python
 def test_deploy_cos(cos: jubilant.Juju):
     """Deploy COS Lite in a separate model."""
@@ -527,6 +529,8 @@ def test_integrate_loki(juju: jubilant.Juju, cos: jubilant.Juju):
 ```
 
 ### Request logs from Loki
+
+Add a test function to `tests/integration/test_charm.py`:
 
 ```python
 def test_loki_data(cos: jubilant.Juju):
@@ -557,9 +561,20 @@ def _get_loki_logs(loki_api_url: str) -> list[str] | None:
     return None
 ```
 
+For more information, see:
+
+- [Traefik actions](https://charmhub.io/traefik-k8s/actions)
+- [Loki HTTP API](https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-label-values)
+
 ### Run the tests
 
-TODO
+Run the following command from anywhere in the `~/fastapi-demo` directory:
+
+```text
+tox -e integration
+```
+
+The tests take 10-15 minutes to run, depending on your computer and network.
 
 ## Review the final code
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
@@ -85,7 +85,7 @@ If you open `lib/charms/grafana_k8s/v0/grafana_dashboard.py` and the other libra
 - `loki_push_api.py` specifies `PYDEPS = ["cosl"]`
 - `prometheus_scrape.py` specifies `PYDEPS = ["cosl>=0.0.53"]`
 
-This means that you need to add `cosl>=1.9.1` to your charm's dependencies.
+You should add the latest version of the [cosl package](https://pypi.org/project/cosl/) to your charm's dependencies.
 
 To update your charm's dependencies in `pyproject.toml`, run:
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
@@ -468,6 +468,97 @@ The search result should look like:
 
 Which means that `/names` has received 150 requests so far.
 
+## Write integration tests
+
+Let's write some integration tests to check that our application works correctly with COS Lite:
+
+1. Deploy COS Lite in a separate Juju model.
+2. Integrate our application with Loki from the COS Lite model.
+3. Ask Loki for logs related to our application.
+
+### Create a model for COS Lite
+
+The existing integration tests use a model that is created by the `juju` fixture. We'll define a similar fixture that creates a separate model for COS Lite.
+
+First, update the imports at the beginning of `tests/integration/test_charm.py`:
+
+```python
+import json
+import logging
+import pathlib
+import time
+
+import jubilant
+import pytest
+import pytest_jubilant
+import requests
+import yaml
+```
+
+We need `pytest` and `pytest_jubilant` for the new fixture. We'll use the other additions (`json`, `time`, and `requests`) when we write the tests.
+
+Next, still in `tests/integration/test_charm.py`, define the new fixture:
+
+```python
+@pytest.fixture(scope="module")
+def cos(juju_factory: pytest_jubilant.JujuFactory):
+    yield juju_factory.get_juju(suffix="cos")
+```
+
+`get_juju` creates a model called `jubilant-<randomhex>-cos`.
+
+We're now ready to write the tests.
+
+### Deploy COS Lite and integrate with Loki
+
+```python
+@pytest.mark.juju_setup
+def test_deploy_cos(cos: jubilant.Juju):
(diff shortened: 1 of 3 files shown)


docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: Should we tell people to use `cosl>=1.9.1,<2`?

tromai: `cosl>=1.9.1,<2` would be stricter than `loki_push_api.py`'s dependency definition `PYDEPS = ["cosl"]`.
I see no issue with `cosl>=1.9.1,<2`, but may be it's better to keep it consistent with the charmlibs deps.

tromai (approved):
The docs changes look good to me. Thank you.

I agree with letting the users know about overriding the timeout period.

There might be an issue that makes the test timeout, and increasing the period doesn't help. In such case, the reader should debug their tests or look for help. This could be a small "note" or "warning" at the end. I'm not sure whether we should include it, as it's hard to know how long is _too long_.

dwilding:
> There might be an issue that makes the test timeout, and increasing the period doesn't help. In such case, the reader should debug their tests or look for help. This could be a small "note" or "warning" at the end. I'm not sure whether we should include it, as it's hard to know how long is _too long_.

Yeah, I think it's hard to give general advice. But I like your thinking. I've updated the advice slightly:

> Then run `tox -e integration` again. If the test still fails, try [our example charm for this chapter](https://github.com/canonical/operator/tree/main/examples/k8s-3-postgresql) instead, in case there's a mistake in your code.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: I suppose, really, the Loki library should have tighter constraints on `cosl` (!) I'm not sure whether charmers ought to constrain both the lower and upper bound of a transitive dep. What do you think @james-garner-canonical?

tonyandrewmeyer: For what it's worth, the PYDEPS value is ignored in this case; you have to manually add the restriction. So I don't think it really matters here. When (if) they move to a Python package on charmlibs putting a maximum major would be a good move.

tromai:
> Then run tox -e integration again. If the test still fails, try [our example charm for this chapter](https://github.com/canonical/operator/tree/main/examples/k8s-3-postgresql) instead, in case there's a mistake in your code.

This is perfect. Thanks!

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
james-garner-canonical: Agreed that we should use `cosl>=1.9.1,<2`.

For Charmhub-hosted libraries today, charmers should specify the bounds on the libraries' dependencies themselves, manually resolving any differences between libraries' dependencies (and of course they should also be using a lockfile to pin the exact version tested with).

Hopefully `loki_push_api` will move off Charmhub to a regular Python package eventually, and then we (and charmers) won't have to specify the transitive dependency ourselves.

```suggestion
uv add 'cosl>=1.9.1,<2'
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -468,6 +468,114 @@ The search result should look like:
 
 Which means that `/names` has received 150 requests so far.
 
+## Write integration tests
+
+Let's write some integration tests to check that our application works correctly with COS Lite:
```
james-garner-canonical:
```suggestion
Let's write some integration tests to check that our application works correctly with COS Lite. The tests will:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -468,6 +468,114 @@ The search result should look like:
...
+import time
+
+import jubilant
+import pytest
+import pytest_jubilant
+import requests
+import yaml
+```
+
+We need `pytest` and `pytest_jubilant` for the new fixture. We'll use the other additions (`json`, `time`, and `requests`) when we write the tests.
```
james-garner-canonical: I see the value in providing a snapshot for easy verification that you have the right contents in your file, but for easy reading it would be good to see what's new. Maybe it's possible to use diff formatting? Though I guess you then can't copy paste the whole thing easily ...

```diff
+ import json
  import logging
  import pathlib
+ import time

  import jubilant
+ import pytest
+ import pytest_jubilant
+ import requests
  import yaml
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -468,6 +468,114 @@ The search result should look like:
...
+
+Next, still in `tests/integration/test_charm.py`, define the new fixture:
+
+```python
+@pytest.fixture(scope="module")
+def cos(juju_factory: pytest_jubilant.JujuFactory):
+    yield juju_factory.get_juju(suffix="cos")
+```
+
+`get_juju` creates a model called `jubilant-<randomhex>-cos`.
```
james-garner-canonical, line 506: Though if the user specifies `--juju-model` on the CLI it will create `<user specified model prefix>-cos` -- do we want to get into that here?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -468,6 +468,114 @@ The search result should look like:
...
+
+We need `pytest` and `pytest_jubilant` for the new fixture. We'll use the other additions (`json`, `time`, and `requests`) when we write the tests.
+
+Next, still in `tests/integration/test_charm.py`, define the new fixture:
+
+```python
+@pytest.fixture(scope="module")
+def cos(juju_factory: pytest_jubilant.JujuFactory):
+    yield juju_factory.get_juju(suffix="cos")
+```
```
james-garner-canonical, line 504: I kind of wonder if the COS tests should be a separate module. Sorry, I know I reviewed the initial tests-only PR, it didn't occur to me at the time. But seeing the tutorial steps make me wonder if that would be clearer.

Though it would make the tests a bit heavier since we'd need to deploy our charm a second time as well (since the Juju models are module-scoped) -- and decide whether to deploy our charm in the same model as cos or a different one.

Not a blocker, can stay as-is for now.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -468,6 +468,114 @@ The search result should look like:
...
+`get_juju` creates a model called `jubilant-<randomhex>-cos`.
+
+We're now ready to write the tests.
+
+### Deploy COS Lite and integrate with Loki
+
+Add two test functions to `tests/integration/test_charm.py`:
+
+```python
+def test_deploy_cos(cos: jubilant.Juju):
```
james-garner-canonical, line 515: Presumably we'd want this.
```suggestion
@pytest.mark.juju_setup
def test_deploy_cos(cos: jubilant.Juju):
```

And I guess we want it on the integrate test as well.

Again, something I should have noticed in the previous PR.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
james-garner-canonical: This wouldn't be a question if it was a separate file :)

But if we stick with the all in one file, maybe we should say something like:

> We'll need to import `json` and `time` from the standard library, as well as `pytest`, `pytest_jubilant`, and `requests`. Your imports should now look like this:

And drop the follow up paragraph about why we need them, as that should be clear from the code we add.

james-garner-canonical (approved):
Looks good to me, thanks @dwilding. My main suggestion is that we add `@pytest.mark.juju_setup` to the deploy and integrate tests (should add them to the examples too in this PR if so).

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: I don't think we need to get into that, yeah

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: I'll leave it as is for now, as I think it's going to get quite fiddly in the tutorial

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: Pretty sure you did notice it in the previous PR. I deferred doing it there because I want to get `@pytest.mark.juju_setup` into the profiles first. I'm going to use the same excuse here :slightly_smiling_face: When https://github.com/canonical/charmcraft/pull/2695 merges, I'll open an Ops PR to add `@pytest.mark.juju_setup` to the tutorials and example charms.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: > But if we stick with the all in one file, maybe we should say something like

Nice suggestion, thanks. I've gone with that

PR #2482, opened 2026-05-18
chore: rename license file to avoid dependabot bug

tonyandrewmeyer:
Dependabot, for some time, believed every text file was a dependency file, ludicrously.


tonyandrewmeyer:
This was apparently fixed upstream a few days ago: https://github.com/dependabot/dependabot-core/pull/14986

Testing now to make sure it works.

tonyandrewmeyer:
It does work ([example](https://github.com/canonical/operator/pull/2483)), so the long nightmare is over, and we can keep our .txt extension.

I think it's better without, but also that the noise of renaming isn't needed any more.

PR #2486, opened 2026-05-18
docs: update profile guidance in HACKING.md

dwilding:
In Madrid we decided to archive [https://github.com/canonical/charmcraft-profile-tools](https://github.com/canonical/charmcraft-profile-tools), partly because we plan to stop shipping uv.lock with the Charmcraft profiles.

This PR updates our guidance in HACKING.md to recommend a more direct way of running tests when updating the profiles.

I've also included the follow-up step of updating the Charmcraft commit hash in the Ops docs (which won't be needed when we switch back to Charmcraft stable releases - another Madrid decision).

**[Preview of updated guidance](https://github.com/dwilding/operator/blob/charmcraft-profile-guidance/HACKING.md#updating-the-charmcraft-profiles)**

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -348,49 +348,62 @@ Then, check out the main branch of your forked operator repo and pull upstream t
 
 If the release automation script fails, delete the draft release and the newly created branches (`release-prep-*`, `post-release-*`) both locally and in the origin, fix issues, and retry.
 
-## Updating the Ops versions in the Charmcraft profiles
+## Updating the Charmcraft profiles
 
-The Charmcraft `kubernetes` and `machine` profiles specify a minimum Ops version in their `pyproject.toml` templates. If an Ops release includes a major new feature or resolves a dependency issue, open a PR to Charmcraft to increase the minimum Ops version in the profiles and refresh the `uv.lock` templates.
+The Charmcraft `kubernetes` and `machine` profiles specify a minimum Ops version in their `pyproject.toml` templates. If an Ops release includes a major new feature or resolves a dependency issue, open a Charmcraft PR to increase the minimum Ops version in the profiles.
 
-1. Fork the [Charmcraft repo](https://github.com/canonical/charmcraft) and create a branch for local development. In your branch, run `make setup` to create a virtual environment, then run `source .venv/bin/activate`.
+Here's the general maintenance process for the Charmcraft profiles.
 
-    See also: Charmcraft's [contributing guide](https://github.com/canonical/charmcraft/blob/main/CONTRIBUTING.md)
+### Editing the profiles
 
-2. In your Charmcraft development branch, modify the Ops version specifier in:
+In your Charmcraft clone, check out a new branch, then edit the .j2 template files in these directories:
 
-    - `charmcraft/templates/init-kubernetes/pyproject.toml.j2`
-    - `charmcraft/templates/init-machine/pyproject.toml.j2`
+- `charmcraft/templates/init-kubernetes`
+- `charmcraft/templates/init-machine`
 
-2. Clone [charmcraft-profile-tools](https://github.com/canonical/charmcraft-profile-tools) locally.
+Don't commit changes yet. Wait until you've tested the charms that `charmcraft init` generates.
 
-3. In your clone of charmcraft-profile-tools, run:
+### Testing the profiles
 
-    ```text
-    CHARMCRAFT_DIR=/path/to/charmcraft just init
-    ```
+At the Charmcraft root, create a directory `.charms` and a script `.charms/generate.py`:
 
-    This initialises a Kubernetes charm and a machine charm based on your Charmcraft development branch.
+```python
+import pathlib
+import shutil
+import subprocess
 
-    If you don't have [just](https://just.systems/man/en/) installed, use `uvx --from rust-just just` instead.
+INIT = ["uv", "run", "--python", "3.10", "--no-dev", "charmcraft", "init"]
+LOCK = ["uv", "lock", "--python", "3.10"]
+TOX = ["uvx", "--python", "3.10", "--with", "tox-uv", "tox"]
 
-4. Lock the dependencies of the charms and generate `uv.lock.j2` files:
 
-    ```text
-    just lock
-    ```
+def init_and_test(profile: str):
+    name = f"myapp-{profile}"
+    if pathlib.Path(name).exists():
+        shutil.rmtree(name)
+    subprocess.run([*INIT, "--profile", profile, "--project-dir", name], check=True)
+    subprocess.run(LOCK, cwd=name, check=True)
+    subprocess.run([*TOX, "-e", "lint,unit"], cwd=name, check=True)
 
-5. Copy the `uv.lock.j2` files to your Charmcraft development branch:
 
-    ```text
-    cp .templates/init-kubernetes/uv.lock.j2 /path/to/charmcraft/charmcraft/templates/init-kubernetes
-    cp .templates/init-machine/uv.lock.j2 /path/to/charmcraft/charmcraft/templates/init-machine
-    ```
+init_and_test("kubernetes")
+init_and_test("machine")
+```
 
-6. In your Charmcraft development branch, commit your changes.
+In the `.charms` directory, run `uv run --script generate.py`.
 
-    You should have changed these files:
+### Opening a Charmcraft PR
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -348,49 +348,62 @@ Then, check out the main branch of your forked operator repo and pull upstream t
...
-    This initialises a Kubernetes charm and a machine charm based on your Charmcraft development branch.
+```python
+import pathlib
+import shutil
+import subprocess
 
-    If you don't have [just](https://just.systems/man/en/) installed, use `uvx --from rust-just just` instead.
+INIT = ["uv", "run", "--python", "3.10", "--no-dev", "charmcraft", "init"]
+LOCK = ["uv", "lock", "--python", "3.10"]
+TOX = ["uvx", "--python", "3.10", "--with", "tox-uv", "tox"]
```
dwilding: Use Python 3.10 for all operations/tests on the generated charms - the minimum Python version that we need to guarantee the profiles support. Maybe this is too fussy?

dwilding:
Very low priority for review

HACKING.md
james-garner-canonical: Seems unnecessary for `charmcraft init` and `uv lock`, but sensible for `tox -e lint,unit`.

HACKING.md
```diff
@@ -348,49 +348,62 @@ Then, check out the main branch of your forked operator repo and pull upstream t
...
-2. Clone [charmcraft-profile-tools](https://github.com/canonical/charmcraft-profile-tools) locally.
+Don't commit changes yet. Wait until you've tested the charms that `charmcraft init` generates.
 
-3. In your clone of charmcraft-profile-tools, run:
+### Testing the profiles
 
-    ```text
-    CHARMCRAFT_DIR=/path/to/charmcraft just init
-    ```
+At the Charmcraft root, create a directory `.charms` and a script `.charms/generate.py`:
```
james-garner-canonical: This is a rather nice approach. There are two subtleties that I'm not sure we convey clearly enough here:
1. `uv run` is what lets us run `charmcraft init` from the local `charmcraft` code + profiles -- we doubly ensure it currently with `uv run --script generate.py`, and `uv run charmcraft init` in the script itself (which I found a little confusing).
2. The directory we init the charms to is controlled by the script's working directory -- the instruction to run it from `.charms` is to ensure that they end up at `.charms/my-app-kubernetes` etc.

Maybe this is clearer as a bash script?
```bash
#!/usr/bin/env bash
set -xueo pipefail

for profile in kubernetes machine; do
    project="myapp-${profile}"
    rm -rf "${project}"
    uv run --script --no-dev \
        charmcraft init --profile "${profile}" --project-dir "${project}"
    pushd "${project}"
    uv lock
    uvx --python 3.10 --with tox-uv tox -e lint,unit
    popd
done
```

Though the CWD bit is still a bit subtle. Maybe clearer if the script sits at the repository root, and manages `.charms`? Like
```bash
rm -rf .charms
mkdir .charms

for profile in kubernetes machine; do
    project=".charms/myapp-${profile}"
```

HACKING.md
```diff
@@ -348,49 +348,62 @@ Then, check out the main branch of your forked operator repo and pull upstream t
...
+
+In the `.charms` directory, run `uv run --script generate.py`.
+
+### Opening a Charmcraft PR
+
+Use a conventional commit type **for each commit**. For example, `chore(templates):`.
+
+Don't commit the `.charms` directory!
+
+### Updating the Ops docs and example charms
```
james-garner-canonical: I'm not sure about this bit. It does reflect what we have right now, but at the sprint we discussed coordinating with the starcraft folks to align our docs updates with charmcraft releases going forward. Should we discuss this further?

dwilding: My thinking is that after https://github.com/canonical/charmcraft/pull/2695 merges, we can update the commit hashes in Ops one last time. Then there'll be a Charmcraft release at some point, which will go to `latest/stable`, and then we can revert the Ops docs to using installed Charmcraft.

So thinking through the timeline, I'd be OK with removing this bit if we're only going to need it one more time in practice.

HACKING.md
dwilding: A bash script sounds good. I also prefer the idea of having a script at the repo root. The only reason I put it in the `.charms` directory was so that I could more easily find it again (there's a lot of stuff at the repo root). But I think that's a less important concern than the one you raised.

HACKING.md
dwilding: To clarify: I think one hash bump is acceptable because our docs are already a couple of PRs ahead of the latest Charmcraft release. So we can't switch the Ops docs back to installed Charmcraft until after the _subsequent_ Charmcraft release goes to `latest/stable`.

tonyandrewmeyer: I think it's ok to land this as it is now, as long as we do quickly follow up the next charmcraft release (candidate today and stable a week from now, if I understand correctly).

HACKING.md
```diff
@@ -348,49 +348,62 @@ Then, check out the main branch of your forked operator repo and pull upstream t
...
+init_and_test("machine")
+```
+
+In the `.charms` directory, run `uv run --script generate.py`.
+
+### Opening a Charmcraft PR
+
+Use a conventional commit type **for each commit**. For example, `chore(templates):`.
+
+Don't commit the `.charms` directory!
```
tonyandrewmeyer: I feel like this is awkward and someone almost certainly will accidentally do this. Can we put the directory outside of the charmcraft repo? Or add it to `.gitignore` or use an existing ignored location?

HACKING.md
tonyandrewmeyer: I'm generally not a fan of bash scripts, but here I do agree that it seems a bit messy to have a Python script that's almost entirely just calling subprocesses. @james-garner-canonical's bash version seems quite readable to me.

tonyandrewmeyer (approved):
I agree with the James comments, but in general seems like a nice simplification.

Approving in advance; I assume you'll resolve the various comments as appropriate.

dwilding:
Thanks for the comments. Addressing several at the same time...

How about we say to put the script somewhere external to the Charmcraft dir and expect the Charmcraft dir as an argument:

```sh
#!/usr/bin/env bash
set -xueo pipefail

charmcraft_dir="$1"

for profile in kubernetes machine; do
    project="myapp-${profile}"
    rm -rf "${project}"
    uv run --project "$charmcraft_dir" --no-dev \
        charmcraft init --profile "${profile}" --project-dir "${project}"
    pushd "${project}"
    uv lock
    uvx --python 3.10 --with tox-uv tox -e lint,unit
    popd
done
```

(Modified from James's version as I don't think we want `--script` on `uv run`, do we?)

HACKING.md
dwilding: https://github.com/canonical/charmcraft/pull/2695 has merged and I'm working on a PR to bump the commit hash for the final time. So I've reduced the Ops guidance to:

> After your PR has merged and Charmcraft has released to `latest/stable`, make sure that the Ops tutorials and example charms are consistent with your profile changes.

dwilding:
I updated HACKING.md with the bash script from my previous comment - [preview](https://github.com/dwilding/operator/blob/charmcraft-profile-guidance/HACKING.md#updating-the-charmcraft-profiles)

HACKING.md
```diff
@@ -348,49 +348,49 @@ Then, check out the main branch of your forked operator repo and pull upstream t
...
-    cp .templates/init-kubernetes/uv.lock.j2 /path/to/charmcraft/charmcraft/templates/init-kubernetes
-    cp .templates/init-machine/uv.lock.j2 /path/to/charmcraft/charmcraft/templates/init-machine
-    ```
+### Opening a Charmcraft PR
 
-6. In your Charmcraft development branch, commit your changes.
+Use a conventional commit type **for each commit**. For example, `chore(templates):`.
 
-    You should have changed these files:
+Don't commit the `.charms` directory!
```
james-garner-canonical:
```suggestion
Don't commit the generated charms!
```

james-garner-canonical (approved):
Looks good, thanks! I wonder if charmcraft would accept an addition to their makefile that runs the checks we want, so we can replace the bash script instructions with "run `make quick-check-profiles`" or similar.

HACKING.md
dwilding: Oops, yes, I forgot to remove this line before asking for review

PR #2492, opened 2026-05-19
chore: backport fixes from 3.x to 2.23

tonyandrewmeyer:
This PR contains cherry-picked fixes from 3.x (main) to the 2.23 branch, as selected in the recent sprint. This adjusts the 2.23 branch from critical fixes only to a more usual LTS release that (for some time) includes regular bug fixes.

There are a few conflicts that were resolved:
* #1918, #1925: resolved tests in test_model.py to match the final state after both PRs.
* #2127, #2379: the added tests reference fixtures/classes not in the 2.23 branch. The source fix was kept and the regression test ported.
* #2264: kept the timeconv changes but skipped the hookcmds changes (hookcmds is not in the 2.23 branch).
* #2274: switched `CheckInfo` to `init=False` with a custom `__init__`
* #2343: adapted the `Container` fallback import into the 2.23 style version-gated try/else.


testing/tests/test_e2e/test_secrets.py
dwilding, line 165: Copilot suggested porting an assertion from https://github.com/canonical/operator/pull/2127:

```suggestion
assert secret.owner == ('app' if app else 'unit')

```

dwilding (approved):
I spent some time with Copilot reviewing the consistency between the backported fixes and `main`. In the end it only had one minor suggestion (see comment).

I wonder whether we should change the positioning of the [2.23 docs](https://documentation.ubuntu.com/ops/2.x/). We say:

> Ops 2.x will continue to receive security and critical bug fixes.

Should we broaden this statement to cover less critical fixes?

testing/src/scenario/mocking.py
```diff
@@ -415,7 +425,7 @@ def secret_add(
         description: str | None = None,
         expire: datetime.datetime | None = None,
         rotate: SecretRotate | None = None,
-        owner: Literal['unit', 'app'] | None = None,
+        owner: Literal['unit', 'application'] | None = None,
```
james-garner-canonical, line 428: This change jumps out as kinda breaky  (type checking only, tests only), but I guess we decided it was OK originally.

james-garner-canonical (approved):
I had a scan through all the non-test changes, and they look good to me.

ops/testing.py
```diff
@@ -75,8 +75,10 @@
 try:
     _version = importlib.metadata.version('ops-scenario')
 except importlib.metadata.PackageNotFoundError:
-    pass
+    from .model import Container
 else:
+    if not (_version and int(_version.split('.', 1)[0]) >= 7):
+        from .model import Container
     if _version and int(_version.split('.', 1)[0]) >= 7:
```
james-garner-canonical, line 87: I'm not 100% sure why we need this specific adaptation in the 2.23 branch. Maybe a comment would be handy.

The `if not (foo)` followed by `if foo` feels a little odd too, could that be an if/else? I guess we don't want to modify the existing `if _version` line if possible for easier merges in future, and I guess `from .model import Container` being in a trailing `else` branch might not satisfy the type checker ... but it feels odd.

testing/src/scenario/mocking.py
tonyandrewmeyer: Yes, and I believe we figured that is was broken before so even though it's not backwards compatible it's better to be non-broken.

ops/testing.py
tonyandrewmeyer: Added a comment. Indeed, trying to keep future merges simple.

PR #2498, opened 2026-05-22
docs: add how-to guide for debugging Kubernetes charms

### LINKED ISSUE #2489: Extend the debugging how-to with Pebble-specific advice
This was suggested by Jon in the Madrid team-only, and also (perhaps not coincidentally) by Sebastian from Ubuntu.

We cover the debug-* functionality now, but we should add more Pebble/K8s specific debugging, such as getting logs from Pebble, running commands, getting files, and so on.

We should possibly also have a charm debugging skill that utilises both of these.

Also: check if the pebble binary is on the path from Juju and if not, convince the Juju team it should be. Also, can we make it easier to run pebble without the long command line and explicit socket environment variable?

tonyandrewmeyer:
This PR adds a follow-on guide to the recent how-to for debugging, specifically focused on K8s charms and using Pebble.

At the recent sprints we received a couple of comments that more information was needed for debugging in this specific case, so this is addressing those.

The main focus is on Pebble, but there's a little bit for K8s directly, without going all the way into being a guide for debugging K8s itself.

[Preview](https://canonical-ubuntu-documentation-library--2498.com.readthedocs.build/ops/2498/howto/debug-a-kubernetes-charm/)

Fixes #2489


docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,183 @@
...
+  html_meta:
+    description: Debug Kubernetes charms and their Pebble-managed workloads -- inspect the charm and workload containers, read Pebble services, logs, plans, changes, checks, and notices, and diagnose common sidecar failure modes.
+---
+
+(debug-a-kubernetes-charm)=
+# How to debug a Kubernetes charm
+
+> See first: {ref}`debug-your-charm`, {ref}`workload-containers`, {external+juju:ref}`Juju | How to manage logs <manage-logs>`
+
+A Kubernetes charm runs as a sidecar: each unit is a pod with a *charm container* (running your charm code) alongside zero or more *workload containers*. Juju injects {external+pebble:doc}`Pebble <index>` into every workload container as its service manager, and your charm talks to each workload's Pebble over an HTTP-on-Unix-socket API.
```
tonyandrewmeyer:
```suggestion
A Kubernetes charm runs as a sidecar: each unit is a pod with a *charm container* (running your charm code) alongside zero or more *workload containers*. Juju injects {external+pebble:doc}`Pebble` into every workload container as its service manager, and your charm talks to each workload's Pebble over an HTTP-on-Unix-socket API.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,183 @@
...
+| The pod, image pulls, scheduling | Kubernetes | `kubectl` |
+
+The charm and workload containers each have their own filesystem and process space. The charm reaches a workload's Pebble through a Unix socket that Juju mounts into both containers:
+
+- In the **workload container**, the socket is at `/var/lib/pebble/default/pebble.sock`.
+- In the **charm container**, the same socket is mounted at `/charm/<container>/pebble.sock`, and the Pebble CLI binary is at `/charm/bin/pebble`.
+
+```{tip}
+If [`Container.can_connect()`](ops.Container.can_connect) returns `False` or your charm raises [`ops.pebble.ConnectionError`](ops.pebble.ConnectionError), the charm container cannot reach the workload's Pebble over that socket. This usually means the workload container hasn't started yet (no [`PebbleReadyEvent`](ops.PebbleReadyEvent) has fired) -- look at the pod first (see [](#k8s-inspect-the-pod)), not at your charm code.
+```
```
dwilding: Suggest migrating this tip

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,183 @@
...
+```shell
+pebble notices                # notices not yet acknowledged
+pebble notice 4               # detail for one notice by ID
+pebble warnings --all         # Pebble's own warnings (deprecations, config issues)
+```
+
+(k8s-debug-from-charm-container)=
+## Debug from the charm container
+
+Many production workload images are stripped down to just the application -- with no shell or utilities -- so `juju ssh --container` lands you nowhere useful. You can still run Pebble commands against that workload from the charm container, because the workload's socket is mounted there:
```
dwilding: Suggest migrating most of this

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,183 @@
...
+/charm/bin/pebble logs
+/charm/bin/pebble exec -- cat /etc/myapp/config.yaml
+```
+
+This is also the most faithful way to reproduce what your charm sees, since your charm talks to exactly this socket. If a Pebble command works here but your charm raises an error, the problem is in the charm code rather than the Pebble configuration.
+
+(k8s-inspect-the-pod)=
+## Inspect the pod at the Kubernetes layer
+
+When a unit is stuck before Pebble is even reachable -- the container is `waiting`, the image won't pull, or the pod won't schedule -- the answer is below Juju, at the Kubernetes layer. Juju puts each model in its own namespace, and names each unit's pod `<app>-<unit-number>`.
```
dwilding: Suggest migrating the first sentence

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,183 @@
...
+## Common failure modes
+
+| Symptom | Where to look |
+| --- | --- |
+| Charm stuck in `maintenance`/`waiting`; `can_connect()` is `False` | The workload container hasn't started -- `kubectl describe pod` for image-pull or scheduling errors ([](#k8s-inspect-the-pod)). |
+| Service shows `backoff` or `error` | `pebble logs` for the crash output, then `pebble changes` / `pebble tasks` for the start failure ([](#k8s-pebble-cli)). |
+| Config change has no effect on the running process | The charm added a layer but didn't [`replan`](#run-workloads-with-a-charm-kubernetes-replan); confirm with `pebble plan` and `pebble services`. |
+| Charm raises `ConnectionError` mid-handler | The workload's Pebble became unreachable -- guard Pebble calls with `try`/`except` rather than `can_connect()` ([](ops.Container.can_connect)). |
+| `pebble_custom_notice` never fires | Confirm the notice was recorded with `pebble notices`; check the `key` your handler matches on ([](#k8s-pebble-cli)). |
+| Workload won't go ready despite running | A health check is failing -- `pebble checks` and `pebble check <name> --refresh` ([](#k8s-pebble-cli)). |
```
dwilding: Suggest migrating this

dwilding (commented):
Thanks for compiling this! I need to review in more detail, but have taken a first pass.

I think it would be easier for people to orient themselves if we move "Common failure modes" nearer the beginning of the doc - probably after "Know which container you’re looking at". I think that section is a great quick reference and should be (slightly) expanded by migrating other content from around the doc.

I've commented on the pieces I think we should move.

My thinking is that we should make a cleaner split between the _why_ and the _how_. If you already know _why_ you need to be reading a particular section, there should be minimal intro text. Get right into the _how_. But if you don't know which section you should be reading, "Common failure modes" points you in the right direction and helps you understand why.

Let me know if you'd like to discuss this suggestion together. I'm also very happy to experiment with different structures if that would help.

tonyandrewmeyer:
@dwilding could you take another look when you have some time? No rush :)

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+  html_meta:
+    description: Debug Kubernetes charms and their Pebble-managed workloads -- inspect the charm and workload containers, read Pebble services, logs, plans, changes, checks, and notices, and diagnose common sidecar failure modes.
+---
+
+(debug-a-kubernetes-charm)=
+# How to debug a Kubernetes charm
+
+> See first: {ref}`debug-your-charm`, {ref}`workload-containers`, {external+juju:ref}`Juju | How to manage logs <manage-logs>`
+
+A Kubernetes charm runs as a sidecar: each unit is a pod with a *charm container* (running your charm code) alongside zero or more *workload containers*. Juju injects {external+pebble:doc}`Pebble <index>` into every workload container as its service manager, and your charm talks to each workload's Pebble over an HTTP-on-Unix-socket API.
```
dwilding: Very minor readability/style things

```suggestion
A Kubernetes charm runs as a sidecar: each unit is a pod with a *charm container* (running your charm code) alongside zero or more *workload containers*. Juju injects {external+pebble:doc}`Pebble <index>` into each workload container as its service manager. Your charm talks to each workload's Pebble over an HTTP-on-Unix-socket API.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+This split is what makes debugging a Kubernetes charm different from debugging a machine charm: a problem can live in the charm code, in the Pebble configuration, in the workload process itself, or at the Kubernetes layer below all of them. This guide covers the Kubernetes- and Pebble-specific tools for narrowing that down. For the substrate-agnostic tools (`juju debug-log`, `juju debug-hooks`, `juju debug-code`, jhack, and remote debugging with VS Code), see {ref}`debug-your-charm`.
```
dwilding: A couple of suggestions for easier reading. I think we can drop "Pebble-specific" because we're using Kubernetes in the sense of Kubernetes the substrate.

```suggestion
This split is what makes debugging a Kubernetes charm different from debugging a machine charm: a problem can live in the charm code, in the Pebble configuration, in the workload process itself, or at the Kubernetes layer below all of them. This guide covers the Kubernetes-specific tools for narrowing that down.

For substrate-agnostic tools (`juju debug-log`, `juju debug-hooks`, `juju debug-code`, jhack, and remote debugging with VS Code), see {ref}`debug-your-charm`.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+(k8s-two-container-model)=
+## Know which container you're looking at
+
+Each piece of the system lives in a specific place, and reaching for the wrong one is the most common way to waste time:
```
dwilding: Very minor readability/style things

```suggestion
Each piece of the system lives in a specific place. Reaching for the wrong one is the most common way to waste time.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+| What | Where it runs | How to reach it |
+| --- | --- | --- |
+| Your charm code (`src/charm.py`, hooks, logs) | charm container (named `charm`) | `juju ssh <unit>`, `juju debug-log` |
```
dwilding: "hooks" meaning what you'd see with `debug-hooks`? I think I'd arrange the text differently to avoid implying that this is part of the charm code:


```suggestion
| Your charm code (`src/charm.py`), logs, hooks | charm container (named `charm`) | `juju ssh <unit>`, `juju debug-log`, `juju debug-hooks` |
```

dwilding: Could someone get confused and think that they need to run `juju ssh charm`? It might be clearer not to mention the name of the charm container, if you can always SSH in by unit name.


```suggestion
| Your charm code (`src/charm.py`, hooks, logs) | charm container | `juju ssh <unit>`, `juju debug-log` |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| Pebble and the workload process | workload container (named after the `containers` entry in `charmcraft.yaml`) | `juju ssh --container <name> <unit>`, the Pebble CLI |
+| The pod, image pulls, scheduling | Kubernetes | `kubectl` |
+
+The charm and workload containers each have their own filesystem and process space. The charm reaches a workload's Pebble through a Unix socket that Juju mounts into both containers:
+
+- In the **workload container**, the socket is at `/var/lib/pebble/default/pebble.sock`.
+- In the **charm container**, the same socket is mounted at `/charm/<container>/pebble.sock`, and the Pebble CLI binary is at `/charm/bin/pebble`.
```
dwilding: Very minor readability/style things

```suggestion
- **Workload container** -- The socket is at `/var/lib/pebble/default/pebble.sock`.
- **Charm container** -- The same socket is mounted at `/charm/<container>/pebble.sock`. The Pebble CLI binary is at `/charm/bin/pebble`.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+(k8s-common-failure-modes)=
+## Common failure modes
+
+If you're not sure where to start, find your symptom here and jump to the section that covers it:
+
+| Symptom | Where to look |
+| --- | --- |
+| Charm stuck in `maintenance`/`waiting`; `can_connect()` is `False` (or a Pebble call raises `ConnectionError`) at startup | The charm container can't reach the workload's Pebble -- usually the workload container hasn't started yet (no [`PebbleReadyEvent`](ops.PebbleReadyEvent) has fired). Look at the pod, not your charm code -- `kubectl describe pod` for image-pull or scheduling errors ([](#k8s-inspect-the-pod)). |
```
dwilding: Minor readability/style things. I recommend ending with "See foo" (no punctuation) to reduce visual noise. Similarly for the other rows of the table.

```suggestion
| Charm stuck in `maintenance`/`waiting`; `can_connect()` is `False` (or a Pebble call raises `ConnectionError`) at startup | The charm container can't reach the workload's Pebble. Usually the workload container hasn't started yet (no [`PebbleReadyEvent`](ops.PebbleReadyEvent) has fired). Look at the pod, not your charm code -- `kubectl describe pod` for image-pull or scheduling errors. See [](#k8s-inspect-the-pod) |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| Service shows `backoff` or `error` | `pebble logs` for the crash output, then `pebble changes` / `pebble tasks` for the start failure ([](#k8s-pebble-cli)). |
+| Config change has no effect on the running process | The charm added a layer but didn't [`replan`](#run-workloads-with-a-charm-kubernetes-replan); confirm with `pebble plan` and `pebble services`. |
```
dwilding:
```suggestion
| Config change has no effect on the running process | The charm added a layer but didn't [`replan`](#run-workloads-with-a-charm-kubernetes-replan). Confirm with `pebble plan` and `pebble services`. |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| Charm raises `ConnectionError` mid-handler | The workload's Pebble became unreachable -- guard Pebble calls with `try`/`except` rather than `can_connect()` ([](ops.Container.can_connect)). |
+| `pebble_custom_notice` never fires | Confirm the notice was recorded with `pebble notices`; check the `key` your handler matches on ([](#k8s-pebble-cli)). |
```
dwilding: The table contains a mix of `--` and `;` as connectives. In most cases a full stop would be fine, as it's clear from the context of the table that each sentence is continuing the same topic. I haven't added suggestions about the remaining ones.

```suggestion
| `pebble_custom_notice` never fires | Confirm the notice was recorded with `pebble notices`. Check the `key` your handler matches on ([](#k8s-pebble-cli)). |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+## Common failure modes
+
+If you're not sure where to start, find your symptom here and jump to the section that covers it:
+
+| Symptom | Where to look |
+| --- | --- |
+| Charm stuck in `maintenance`/`waiting`; `can_connect()` is `False` (or a Pebble call raises `ConnectionError`) at startup | The charm container can't reach the workload's Pebble -- usually the workload container hasn't started yet (no [`PebbleReadyEvent`](ops.PebbleReadyEvent) has fired). Look at the pod, not your charm code -- `kubectl describe pod` for image-pull or scheduling errors ([](#k8s-inspect-the-pod)). |
+| Service shows `backoff` or `error` | `pebble logs` for the crash output, then `pebble changes` / `pebble tasks` for the start failure ([](#k8s-pebble-cli)). |
+| Config change has no effect on the running process | The charm added a layer but didn't [`replan`](#run-workloads-with-a-charm-kubernetes-replan); confirm with `pebble plan` and `pebble services`. |
+| Charm raises `ConnectionError` mid-handler | The workload's Pebble became unreachable -- guard Pebble calls with `try`/`except` rather than `can_connect()` ([](ops.Container.can_connect)). |
```
dwilding:
```suggestion
| Charm raises `ConnectionError` mid-handler | The workload's Pebble became unreachable. Guard Pebble calls with `try`/`except` rather than `can_connect()` ([](ops.Container.can_connect)). |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| `pebble_custom_notice` never fires | Confirm the notice was recorded with `pebble notices`; check the `key` your handler matches on ([](#k8s-pebble-cli)). |
+| Workload won't go ready despite running | A health check is failing -- `pebble checks` and `pebble check <name> --refresh` ([](#k8s-pebble-cli)). |
```
dwilding: Would this be OK?

```suggestion
| Workload isn't ready despite running | A health check is failing -- `pebble checks` and `pebble check <name> --refresh` ([](#k8s-pebble-cli)). |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| `juju ssh --container` lands in an image with no shell or tools | The workload image is stripped down -- run Pebble against it from the charm container instead, where the socket is mounted ([](#k8s-debug-from-charm-container)). |
+
+(k8s-pebble-cli)=
+## Inspect the workload with the Pebble CLI
+
+SSH into the workload container and use the Pebble CLI to see what Pebble thinks is going on. The {ref}`debug-your-charm` guide covers `pebble services`, `pebble logs`, `pebble exec`, `pebble plan`, and `pebble checks`. The commands below go deeper and are the ones you'll reach for when a workload won't start or keeps crashing.
```
dwilding:
```suggestion
SSH into the workload container and use the Pebble CLI to see what Pebble thinks is going on. {ref}`debug-your-charm` covers `pebble services`, `pebble logs`, `pebble exec`, `pebble plan`, and `pebble checks`. The commands below go deeper and are the ones you'll reach for when a workload won't start or keeps crashing.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+```shell
+juju ssh --container myapp myapp/0
+```
+
+All of the examples below assume you're running them inside the workload container, where `pebble` is on the `PATH`. (To run them from the charm container instead, see [](#k8s-debug-from-charm-container).)
```
dwilding:
```suggestion
All of the examples below assume you're running them inside the workload container, where `pebble` is on the `PATH`. To run them from the charm container instead, see [](#k8s-debug-from-charm-container).
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| `juju ssh --container` lands in an image with no shell or tools | The workload image is stripped down -- run Pebble against it from the charm container instead, where the socket is mounted ([](#k8s-debug-from-charm-container)). |
+
+(k8s-pebble-cli)=
+## Inspect the workload with the Pebble CLI
+
+SSH into the workload container and use the Pebble CLI to see what Pebble thinks is going on. The {ref}`debug-your-charm` guide covers `pebble services`, `pebble logs`, `pebble exec`, `pebble plan`, and `pebble checks`. The commands below go deeper and are the ones you'll reach for when a workload won't start or keeps crashing.
+
+```shell
+juju ssh --container myapp myapp/0
+```
```
dwilding, line 58: I'd move this command to a dedicated subsection to make sure it doesn't get overlooked. Like this:

````
### SSH into the workload container

```shell
juju ssh --container myapp myapp/0
```

### Read service states

...
````

File: docs/howto/debug-a-kubernetes-charm.md [hunk 19]
```diff
@@ -0,0 +1,176 @@
...
+
+### Read service states
+
+`pebble services` reports a `Current` state for each service. The state tells you most of what you need to know:
+
+| State | Meaning |
+| --- | --- |
+| `active` | The service is running normally. |
+| `inactive` | The service is not running. It was never started (`startup: disabled`), was stopped, or its command could not be executed at all (a wrong path or a binary missing from the image). |
+| `backoff` | The service started but exited, and Pebble is restarting it on a backoff schedule -- the workload is crash-looping. |
```
dwilding: I'm pretty sure I get what you mean by "crash-looping". I think it would be clearer to spell it out:

```suggestion
| `backoff` | The service started but exited, and Pebble is restarting it on a backoff schedule. The workload is crashing each time it starts. |
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+| `error` | The service failed and Pebble has stopped trying to restart it. |
+
+`backoff` and `error` mean the process ran and then died -- look at the workload itself with `pebble logs` next. An unexpected `inactive` (a service you expected to be running) usually means the command never executed; the reason is in `pebble changes` / `pebble tasks` (see below) rather than in the service's logs.
```
dwilding: I find this dense to read. I would simplify the structure and split paragraphs so that you don't need dashes or semicolons. Something like:


```suggestion
`backoff` and `error` mean the process ran and then died. Look at the workload itself with `pebble logs`.

An unexpected `inactive` (a service you expected to be running) usually means the command never executed. The reason is in `pebble changes` / `pebble tasks`, rather than in the service's logs. We'll look at that next.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+Start service "myapp"
+
+2026-05-22T02:09:01Z INFO Most recent service output:
+    Traceback (most recent call last):
+      ...
+    KeyError: 'DATABASE_URL'
+2026-05-22T02:09:01Z ERROR service start attempt: exited quickly with code 1, will restart
+```
+
+The captured "Most recent service output" is the workload's own stdout/stderr, so a stack trace, a missing-config error, or a permission error shows up right here. A service whose command can't be executed at all (a wrong path, or a binary missing from the image) fails differently -- `cannot start service: fork/exec ...: no such file or directory`. Use `pebble tasks --last=start` to jump straight to the most recent service start without looking up its ID.
```
dwilding: This paragraph covers a few different aspects, so I'd probably break it up. I also suggest putting the usage advice immediately after the example. Plus some other suggestions:

- Say "Instead of `pebble tasks <change-id>`" rather than "without looking up its ID"
- "most recent service start" --> "most recent service start attempt"
- Say "For example" in the part about commands that can't be executed

```suggestion
Instead of `pebble tasks <change-id>`, use `pebble tasks --last=start` to jump straight to the most recent service start attempt.

The captured "Most recent service output" is the workload's own stdout/stderr, so a stack trace, a missing-config error, or a permission error shows up right here.

A service whose command can't be executed at all (a wrong path, or a binary missing from the image) fails differently. For example, `cannot start service: fork/exec ...: no such file or directory`.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+$ pebble plan
+services:
+    myapp:
+        summary: my application
+        startup: enabled
+        override: replace
+        command: /bin/myapp --port 8080
+```
+
+`pebble plan` is the right tool for "did my charm's configuration take effect?" -- it shows the live, in-memory result. A charm adds its configuration at runtime with [`Container.add_layer()`](ops.Container.add_layer), which sends the layer to Pebble over the API; those layers are held in memory and are *not* written to `/var/lib/pebble/default/layers/`. That directory contains only the layers baked into the container image (a rock built with Rockcraft may ship some), which Pebble reads once at startup -- so an empty or sparse layers directory is normal and doesn't mean your charm's `add_layer()` call failed. Trust `pebble plan`, not the directory listing.
```
dwilding: I suggest splitting this for easier reading. Plus some other suggestions:

- "those layers" --> "Added layers"
- Splitting the penultimate sentence

```suggestion
`pebble plan` is the right tool for "did my charm's configuration take effect?" -- it shows the live, in-memory result.

A charm adds its configuration at runtime with [`Container.add_layer()`](ops.Container.add_layer), which sends the layer to Pebble over the API. Added layers are held in memory and are *not* written to `/var/lib/pebble/default/layers/`. That directory contains only the layers baked into the container image (a rock built with Rockcraft may ship some), which Pebble reads once at startup. An empty or sparse layers directory is normal and doesn't mean your charm's `add_layer()` call failed. Trust `pebble plan`, not the directory listing.
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+If you change a service's configuration but the running process doesn't change, the usual cause is a missing [`replan`](#run-workloads-with-a-charm-kubernetes-replan): adding a layer updates the plan but does not restart services on its own.
```
dwilding: I think it would be good to explicitly say what `replan` does. Something like:

```suggestion
If you change a service's configuration but the running process doesn't change, the usual cause is a missing [`replan`](#run-workloads-with-a-charm-kubernetes-replan): Pebble updates the plan when you add a layer, but doesn't restart services until you call `replan`.
```

docs/howto/debug-your-charm.md
```diff
@@ -63,6 +63,10 @@ juju ssh --container <container-name> <unit>
 
 Then use the Pebble CLI (available at `/charm/bin/pebble`) to inspect the workload. We've included several examples below.
 
+```{tip}
+For Kubernetes- and Pebble-specific debugging in more depth -- tracing Pebble changes and tasks, working with minimal workload images, and diagnosing common sidecar failure modes -- see {ref}`debug-a-kubernetes-charm`.
+```
+
 > See also: {external+pebble:doc}`Pebble | CLI commands <reference/cli-commands>`
```
dwilding: I think I'd shift the tip to be part of the "see also":

```suggestion
See also:

- {ref}`debug-a-kubernetes-charm`, which covers Kubernetes charm failure modes and debugging tools in more depth
- {external+pebble:doc}`Pebble | CLI commands <reference/cli-commands>`
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Check health checks
+
+A failed {ref}`Pebble health check <pebble-health-checks>` behaves differently depending on how it's configured: going "down" can restart the service (via `on-check-failure`), and a `level: alive` or `level: ready` check is wired to the container's Kubernetes liveness or readiness probe -- so a failing `alive` check makes Kubernetes restart the container, while a failing `ready` check marks the pod not-ready and removes it from its Service's endpoints. Inspect checks with:
```
dwilding: This part is quite dense. How about:

```suggestion
A failed {ref}`Pebble health check <pebble-health-checks>` behaves differently depending on how it's configured:
- Going "down" can restart the service (set with `on-check-failure`).
- A `level: alive` or `level: ready` check is wired to the container's Kubernetes liveness or readiness probe. This means that a failing `alive` check makes Kubernetes restart the container, while a failing `ready` check marks the pod not-ready and removes it from its Service's endpoints.

Inspect checks with:
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+
+```shell
+pebble checks                 # status of all checks
+pebble check myapp-ready      # full detail for one check, in YAML
+pebble check myapp-ready --refresh   # run it now instead of waiting for the next interval
+pebble health                 # exit code 0 if all checks healthy, 1 otherwise
+```
+
+`pebble check <name>` shows the failure count, the threshold, and the error from the most recent run -- useful for distinguishing "the check is configured incorrectly" from "the workload is genuinely unhealthy".
```
dwilding:
```suggestion
`pebble check <name>` shows the failure count, the threshold, and the error from the most recent run. This is useful for distinguishing "the check is configured incorrectly" from "the workload is genuinely unhealthy".
```

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,176 @@
...
+kubectl -n <model> get pods
+kubectl -n <model> describe pod myapp-0        # events: image pulls, scheduling, OOM kills, restarts
+kubectl -n <model> logs myapp-0 -c charm       # charm container stdout
+kubectl -n <model> logs myapp-0 -c myapp       # workload container stdout (Pebble's own output)
+kubectl -n <model> logs myapp-0 -c myapp --previous   # output from the last crashed instance
+```
+
+The `Events` section of `describe pod` is where you'll find `ImagePullBackOff`, `CreateContainerConfigError`, failed liveness probes, and out-of-memory kills -- none of which appear in `juju debug-log` or the Pebble logs.
+
+```{note}
```
dwilding: Promote this to a tip?

dwilding (approved):
Great doc, thanks! Sorry for the delay... I had an almost-finished review pending but got sidetracked with other stuff.

docs/howto/debug-a-kubernetes-charm.md
tonyandrewmeyer: You do use the word `charm` to select it when picking a container to ssh to, though.

docs/howto/debug-a-kubernetes-charm.md
tonyandrewmeyer: Sure.

```suggestion
```{tip}
```

docs/howto/debug-a-kubernetes-charm.md
dwilding: I'm still a bit confused. Doesn't `juju ssh <unit>` get you into the charm container for the unit? No need to specify `charm` anywhere. If we mention the name `charm`, I think it's best if we also mention where that name is used: `--container charm`. But that's optional, if I understand right, so overall I think it's clearer not to mention `charm` here. But maybe I'm overthinking this (please tell me if so :slightly_smiling_face:)

tonyandrewmeyer: It's the default value yes. But you can be explicit:

```
ubuntu@cascade:~$ juju ssh snappass-test/0 ls
agents  template-agent.conf  tools
ubuntu@cascade:~$ juju ssh --container charm snappass-test/0 ls
agents  template-agent.conf  tools
ubuntu@cascade:~$ juju ssh --container snappass snappass-test/0 ls
AUTHORS.rst  MANIFEST.in  README.rst  build  dist  setup.py  snappass  snappass.egg-info
ubuntu@cascade:~$ juju ssh --container redis snappass-test/0 ls
ubuntu@cascade:~$ 
```

If you use an invalid name, Juju will tell you that "charm" is an option, not that it's the default and you can leave it out.

I'm ok taking it out.

docs/howto/debug-a-kubernetes-charm.md
```diff
@@ -0,0 +1,192 @@
...
+| Service shows `backoff` or `error` | `pebble logs` for the crash output, then `pebble changes` / `pebble tasks` for the start failure ([](#k8s-pebble-cli)). |
+| Config change has no effect on the running process | The charm added a layer but didn't [`replan`](#run-workloads-with-a-charm-kubernetes-replan). Confirm with `pebble plan` and `pebble services`. |
+| Charm raises `ConnectionError` mid-handler | The workload's Pebble became unreachable. Guard Pebble calls with `try`/`except` rather than `can_connect()` ([](ops.Container.can_connect)). |
+| `pebble_custom_notice` never fires | Confirm the notice was recorded with `pebble notices`. Check the `key` your handler matches on ([](#k8s-pebble-cli)). |
+| Workload isn't ready despite running | A health check is failing. Run `pebble checks` and `pebble check <name> --refresh` ([](#k8s-pebble-cli)). |
+| `juju ssh --container` lands in an image with no shell or tools | The workload image is stripped down. Run Pebble against it from the charm container instead, where the socket is mounted ([](#k8s-debug-from-charm-container)). |
```
hpidcock: Maybe have a table below that shows common shell commands translated into pebble commands:

e.g.
`ls` -> `pebble ls`
`./myprogram` -> `pebble exec`

Maybe also show injecting a rescue shell:
```
$ apt install busybox-static
$ pebble push /usr/bin/busybox /charm/bin/busybox
$ pebble exec /charm/bin/busybox
# hostname
mycharm-0
```

And lastly, if the user has access to the host (e.g. the container is running on their system) and they have sufficient capabilities, show them how to run a process as root inside that container where they know the real PID of a process in that container:
```
$ apt install busybox-static
$ cp /usr/bin/busybox /proc/<real-pebble-PID>/root/charm/bin/busybox
$ nsenter -a -t <real-pebble-PID> -S 0 -G 0 /charm/bin/busybox
# whoami
root
```

hpidcock (approved):
I like it, but I think we can add some extra help for users who need it in limited containers (e.g. rootless/no-shell)

PR #2499, opened 2026-05-24
fix: pass the endpoint name through to relation-get

### LINKED ISSUE #2327: ops returns databag for non-existing relation.
It is possible to create a relation Object for a relation that does not exist, as long as the relation ID exists,
Given a charm with a relation integrated, with id 1, I can run this:

```python
relation = self.model.get_relation("not-integrated", 1)
relation.data[self.app]
```

Interestingly:
 * The relation object creation is not blocked
 * accessing the relation databag returns the databag of the relation with id 1.

The reason for the first part is [here](https://github.com/canonical/operator/blob/deba324055802b34d4b320655ac00a4ffd85fc66/ops/model.py#L970)
We provide both the relation name and the relation id. We fall in the else case [here](https://github.com/canonical/operator/blob/deba324055802b34d4b320655ac00a4ffd85fc66/ops/model.py#L980) and return a relation object.
The thing is: we're not providing the information for a dead relation, AND we do have a relation existing with that id to a different relation name.
I think this case should raise an error. It's not that the relation is dead but we refuse to forget it.
Maybe the loop a few lines above should also check if there isn't a relation that has this relation ID.
This would work as relation ids are unique throughout a model lifetime.

The reason for the second part is [this line](https://github.com/canonical/operator/blob/deba324055802b34d4b320655ac00a4ffd85fc66/ops/model.py#L1998): while the relation object is created with a relation name and a relation id, loading the databag is ultimately done using only the relation id.

The reason: we pass a relation with an invalid name BUT a valid relation id. And when building the `RelationDataContent` object it uses only the relation ID. And that relation id is valid and does exist and hence we're accessing its databag.

The easy way to go would be to craft a valid identifier when running the `relation-get` hook command: it accepts `relation-get -r <relation_id>` but also `relation-get -r <relation_name>:<relation_id>`.
This would ensure that we're accessing exactly what is needed instead of returning data for a non existing relation.

tonyandrewmeyer:
This PR passes the endpoint name from the Relation down to the actual hook command, where Juju will use it to validate that it matches the provided relation ID. Without this, you can trick Ops into a mismatch between ID and name and get misleading data (although Juju enforces security, so there is no leakage of data that you couldn't get anyway).

This is mostly just plumbing the extra information through the backend and into the hookcmds call from the various places we call this command. There are also a bunch of test updates to include the relation name to match.

Scenario is updated to reflect the Juju behaviour more closely, and otherwise expect / accept the name in calls.

I've updated Harness to match, even though we are not typically doing Harness fixes, because the signature changes and it seems less likely we would break anyone's existing Harness tests this way, and it's a pretty straightforward set of changes.

Fixes #2327


ops/model.py
```diff
@@ -2002,7 +2002,9 @@ def _hook_is_running(self) -> bool:
     def _load(self) -> _RelationDataContent_Raw:
         """Load the data from the current entity / relation."""
         try:
-            return self._backend.relation_get(self.relation.id, self._entity.name, self._is_app)
+            return self._backend.relation_get(
+                self.relation.id, self._entity.name, self._is_app, self.relation.name
```
james-garner-canonical: How about keyword arguments here? Even the calls above with just two arguments would be improved with keyword arguments IMO.

ops/model.py
```diff
@@ -3614,11 +3619,13 @@ def relation_ids(self, relation_name: str) -> list[int]:
...
-    def relation_list(self, relation_id: int) -> list[str]:
-        with self._wrap_hookcmd('relation-list', relation_id=relation_id):
-            return hookcmds.relation_list(relation_id)
+    def relation_list(self, relation_id: int, relation_name: str | None = None) -> list[str]:
+        with self._wrap_hookcmd('relation-list', relation_id=relation_id, endpoint=relation_name):
+            return hookcmds.relation_list(relation_id, endpoint=relation_name)
 
-    def relation_remote_app_name(self, relation_id: int) -> str | None:
+    def relation_remote_app_name(
+        self, relation_id: int, relation_name: str | None = None
```
james-garner-canonical: Should we make the new argument keyword only, here and elsewhere?

```suggestion
        self, relation_id: int, *, relation_name: str | None = None
```

If so, how about using the Juju terminology `endpoint`, as `hookcmds` does?


```suggestion
        self, relation_id: int, *, endpoint: str | None = None
```

testing/src/scenario/mocking.py
```diff
@@ -252,13 +256,27 @@ def _check_app_data_access(self, is_app: bool):
                 f'setting application data is not supported on Juju version {version}',
             )
 
-    def relation_get(self, relation_id: int, member_name: str, is_app: bool):
+    def relation_get(
+        self,
+        relation_id: int,
+        member_name: str,
+        is_app: bool,
+        relation_name: str | None = None,
```
james-garner-canonical, line 265: Match the keyword only decision from Ops itself, here and below.

ops/_private/harness.py
```diff
@@ -2438,19 +2438,40 @@ def relation_ids(self, relation_name: str) -> list[int]:
             no_ids: list[int] = []
             return no_ids
 
-    def relation_list(self, relation_id: int):
```
james-garner-canonical, line 2441: I'm mildly against updating Harness to support feature type changes (adding a new optional argument in this case). But does this make our own testing story better? (`test_model.py`, `test_testing.py`)

test/test_model.py
```diff
@@ -236,13 +236,29 @@ def test_get_relation(self, harness: ops.testing.Harness[ops.CharmBase]):
...
         )
 
+    def test_get_relation_wrong_endpoint(self, harness: ops.testing.Harness[ops.CharmBase]):
+        # Regression test for https://github.com/canonical/operator/issues/2327
+        # Passing an endpoint name that doesn't match the relation_id must not
+        # leak the databag of the relation that actually owns that id.
+        rel_id = harness.add_relation('db1', 'remoteapp1')
+        harness.add_relation_unit(rel_id, 'remoteapp1/0')
+        with harness._event_context('foo_event'):
+            harness.update_relation_data(rel_id, 'remoteapp1', {'secret': 'cafedeadbeef'})
```
james-garner-canonical: Seems like an odd example given that relation data shouldn't contain secrets. The 'leak' description above feels a bit odd too -- isn't the mismatched endpoint name handling behaviour more about modeling Juju cleanly than about security? After all, a user in the charm context can just query Juju directly with a `subprocess.run` if it really wants the data, can't they?

test/test_hookcmds.py
```diff
@@ -540,6 +540,41 @@ def test_relation_set(run: Run, mock_file: NamedTemporaryFile, id: int | None, a
     assert run.calls[0].stdin == '{"foo": "bar"}'
 
 
+# Regression tests for https://github.com/canonical/operator/issues/2327:
+# when an ``endpoint`` is provided together with an ``id``, the relation
+# reference passed to Juju is ``endpoint:id`` so that Juju errors out if the
+# given id doesn't actually belong to that endpoint.
```
james-garner-canonical: This description confuses me a bit. The new tests are all of the happy path.

test/bin/relation-list
```diff
@@ -40,6 +40,8 @@ done
 
 get_relation_data() {
     local rel_id="$1"
+    # Strip the optional ``endpoint:`` prefix so the case-statement keeps working.
+    rel_id="${rel_id##*:}"
```
james-garner-canonical, line 45: Perhaps an example transformation? I guess it's just this?
```suggestion
    # foo:1 -> 1
    rel_id="${rel_id##*:}"
```

And I guess we're just not directly modeling the mismatched name behaviour in any of these bash tests.

ops/hookcmds/_relation.py
```diff
@@ -22,24 +22,39 @@
...
+def _format_relation_ref(id: int, endpoint: str | None) -> str:
+    """Build the ``-r`` argument for relation hook commands.
+
+    Including the endpoint name lets Juju validate that ``id`` actually belongs
+    to ``endpoint``, rather than silently operating on whichever relation has
+    that ID.
+    """
+    if endpoint is None:
+        return str(id)
+    return f'{endpoint}:{id}'
```
james-garner-canonical, line 34: Is it intentional that `(1, endpoint='')` results in `:1`? I guess it's fine either way though since Juju handles the resulting string.

james-garner-canonical (commented):
Design question: how about making the new ops-level arguments keyword only? How about using the Juju name `endpoint` instead of `relation_name`?

Also some non-blocking questions about the tests and harness update. I suspect I'm just a bit confused due to being in unfamiliar territory, but figured it's worth raising the questions anyway.

ops/_private/harness.py
tonyandrewmeyer: Yes, I think so, and I think there's more risk not doing so (as per PR description). The cost also seems very low.

ops/hookcmds/_relation.py
tonyandrewmeyer: In general, hookcmds leaves it to Juju to raise errors for incorrect use. Would you want the code to raise ValueError instead?

ops/model.py
tonyandrewmeyer: @copilot make the suggestion to change to pass these args by keyword.

ops/model.py
tonyandrewmeyer: In terms of "endpoint" vs. "relation name", I believe Juju uses "endpoint" but ops generally uses "relation name", which is why I've following that in the PR. I'm not against standardising on "endpoint", but I worry that it would be more confusing.

Given that this is a method in a private class, it probably doesn't matter too much. I didn't make it kwonly because I feel that for internal use we would want to provide it (even though there is a default value) to get the extra resilience that was missing and caused the original issue.

test/bin/relation-list
tonyandrewmeyer: Is your suggestion just the example? Isn't it clear what it's doing from the existing comment?

ops/_private/harness.py
james-garner-canonical: I reread the PR description and I'm not sure what the risk of not updating Harness is.

test/bin/relation-list
james-garner-canonical: > Is your suggestion just the example?

Yeah.

> Isn't it clear what it's doing from the existing comment?

I had to think about it a bit when reading it the first time, but I agree that in theory it's clear as-is.

ops/hookcmds/_relation.py
james-garner-canonical: I guess it's not obvious to me whether `:1` is legal, and if so, whether it's any different from `1`. Probably the current approach of leaving it entirely to Juju is the right thing, as it shields us from any changes in that behaviour.

ops/_private/harness.py
tonyandrewmeyer: If charms are using Harness and they start using this new keyword argument (indirectly, because Ops is), it seems like we could inadvertently break their tests.

test/test_model.py
tonyandrewmeyer: It comes from the original issue/repro. It's not leaking anything the charm as a whole couldn't get, Juju handles that. But one piece of code could get data it shouldn't have. I'll try to make it clearer.

test/test_hookcmds.py
tonyandrewmeyer: Reworded.

ops/hookcmds/_relation.py
tonyandrewmeyer: Yes, that's my preference.

james-garner-canonical (approved):

tromai (approved):
Thanks for the fix!

PR #2505, opened 2026-05-26
Security: YAML Safe Load Bypass via Custom Loader

tomaioo:
## Summary

Security: YAML Safe Load Bypass via Custom Loader

## Problem

**Severity**: `Low` | **File**: `ops/_private/yaml.py:L28`

In `ops/_private/yaml.py`, the `safe_load` function uses a custom `_safe_loader` which defaults to `yaml.CSafeLoader` if available. While `yaml.safe_load` is generally safe, the comment `# noqa: S506` explicitly suppresses a security warning about using `yaml.load` instead of `yaml.safe_load`. The function does use `yaml.load` with a SafeLoader-derived class, but this pattern is risky if the loader class is ever modified. Additionally, the `safe_dump` function has a type ignore comment that could mask issues.

## Solution

Remove the `# noqa: S506` suppression and ensure the code passes security linting. Consider using `yaml.safe_load` directly instead of `yaml.load` with a custom loader to reduce attack surface. Add explicit documentation about why the custom loader is necessary.

## Changes

- `ops/_private/yaml.py` (modified)

### DIFF
--- a/ops/_private/yaml.py
+++ b/ops/_private/yaml.py
@@ -20,16 +20,11 @@ from typing import Any, TextIO
 
 import yaml
 
-# Use C speedups if available
-_safe_loader = getattr(yaml, 'CSafeLoader', yaml.SafeLoader)
-_safe_dumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
-
-
 def safe_load(stream: str | TextIO) -> Any:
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
The current implementation is correct, allowing us to use `CSafeLoader` if it's available.

PR #2506, opened 2026-05-26
fix: ensure resources are cleaned up in testing.Context

tonyandrewmeyer:
A few fixes to ensure we do not leak resources (until gc time) in Scenario:

* In runtime, move the cleanup to a finally block.
* In the Context, instead of using TemporaryDirectory, use mktmpdir directly, and register a weakref cleanup -- but also let users explicitly clean up if they prefer  (and make the context manager case do that automatically).
* In state, replace a few opens with read_texts (we could do the open with a context manager is read but this seemed the more minimal change, and none of the YAML files should be large enough that there is any difference).
* In a couple of Scenario's own tests, avoid leaking when using Container.pull.

After these changes, the tests can be run with -Werror and not have resource warnings cause failures.


tromai (approved):

testing/src/scenario/context.py
```diff
@@ -790,6 +798,23 @@ def _record_status(self, state: State, is_app: bool):
         else:
             self.unit_status_history.append(state.unit_status)
 
+    def close(self) -> None:
+        """Release the per-context tempdir used for simulated container/storage roots.
+
+        Calling ``close()`` (or using the Context as a context manager) is optional;
+        the tempdir is cleaned up at garbage-collection time either way. Eager
+        cleanup is recommended when running tests with ``-W error``, where the
+        non-deterministic GC-time cleanup can interleave with pytest's finalisers.
```
james-garner-canonical: What does eager cleanup mean for users?

testing/src/scenario/state.py
```diff
@@ -2014,14 +2014,14 @@ def _load_metadata_legacy(charm_root: pathlib.Path):
         # files for charm metadata.
         metadata_path = charm_root / 'metadata.yaml'
         meta: dict[str, Any] = (
-            yaml.safe_load(metadata_path.open()) if metadata_path.exists() else {}
+            yaml.safe_load(metadata_path.read_text()) if metadata_path.exists() else {}
```
james-garner-canonical, line 2017: 🙀

james-garner-canonical (approved):
Very nice cleanup.

james-garner-canonical:
I see in the harness PR you dropped the weakref addition in favour of the existing pattern of users calling `cleanup`. I think the weakref is a nice feature here as it nicely supports the pattern we document of not using `Context` as a context manager.

tonyandrewmeyer:
> I see in the harness PR you dropped the weakref addition in favour of the existing pattern of users calling `cleanup`. I think the weakref is a nice feature here as it nicely supports the pattern we document of not using `Context` as a context manager.

Yeah, whereas we have always told people to call `cleanup` for Harness, so although we *could* auto clean up for them, it was the wrong call, I think. Good push back from David :)

testing/src/scenario/context.py
tonyandrewmeyer: Calling this new `close()` method. Or was that a suggestion to make the language here more understandable?

james-garner-canonical: Yeah, I didn't fully grok that it was an instruction to use `close()` in your tests if you use `-W error`. Should we be more pushy about that since we want people to run their tests that way?

PR #2507, opened 2026-05-26
fix: close SQLite storage in Harness.cleanup()

tonyandrewmeyer:
We don't normally do fixes in Harness, but this will allow charms that are not yet migrated to Scenario to run with `-Werror` and not get hit by a resource warning we cause.

`Harness` builds a `framework.Framework` backed by an in-memory `SQLiteStorage`, but `Harness.cleanup()` never closes it. The sqlite3 connection is only released when the `Harness` is garbage-collected, at which point its destructor emits `unclosed database in <sqlite3.Connection ...>`. pytest wraps that as `PytestUnraisableExceptionWarning` and `-W error` turns it into a test failure — even for callers who correctly call `harness.cleanup()`.

This is one of the bug classes flagged by a recent `canonical/hyrum` (super-tox) run across the canonical charm collection: under `-Werror`, charms whose tests use `ops.testing.Harness` regress even though their own code is clean.

Changes:

- `Harness.cleanup()`: call `self._framework.close()` after the backend cleanup so the `SQLiteStorage` connection is closed eagerly. sqlite3's `close()` is idempotent, so repeated `cleanup()` calls stay safe.
- Add a regression test that asserts the storage's underlying connection is closed after `harness.cleanup()`.


dwilding:
> under `-Werror`, charms whose tests use `ops.testing.Harness` regress even though their own code is clean

To help my understanding: Are those charm tests not using `harness.cleanup()`? Or would the warnings happen even if they're using `harness.cleanup()`?

tonyandrewmeyer:
> > under `-Werror`, charms whose tests use `ops.testing.Harness` regress even though their own code is clean
> 
> To help my understanding: Are those charm tests not using `harness.cleanup()`? Or would the warnings happen even if they're using `harness.cleanup()`?

There are two pieces in the PR:

* closing the framework's sqlite3 storage -- without the PR this is always leaked until gc
* the temp directory cleanup switch to weaker -- this is a safety net for not calling cleanup()

Thinking about this more, I think the second case should be dropped; people should explicitly call `cleanup()`. I'll remove that part of the PR.

tonyandrewmeyer:
Hmm, that data failure looks like it may be a real issue. Moving to draft while I figure that out.

tonyandrewmeyer:
> Hmm, that data failure looks like it may be a real issue. Moving to draft while I figure that out.

It's a downstream bug. We should wait for the [PR with a fix](https://github.com/canonical/postgresql-k8s-operator/pull/1551) to land before merging this, to avoid disabling our CI. There's no particular urgency.

tonyandrewmeyer:
Downstream fix merged :)

james-garner-canonical (approved):

dwilding (approved):

PR #2508, opened 2026-05-26
docs: Improve Container.pull docs to avoid leaky files

### LINKED ISSUE #2504: Improve Container.pull docs to avoid leaky files
Using `Container.pull` you get a file that has been `open`ed, but nothing automatically closes it. This means that doing something like `foo = my_container.pull(path).read()` is leaky, but we suggest that in our documentation.

This demonstrates the issue:

```python
"""Run with: python -W error::ResourceWarning repro_pull_leak.py

Demonstrates that pebble.Client.pull() returns a file-like object that, if not
closed, raises a ResourceWarning (fatal under -Werror). The fix is to use it
as a context manager: `with client.pull(path) as f: ...`.

Requires the `pebble` binary on PATH.
"""

import os
import subprocess
import tempfile
import time

from ops import pebble


def wait_for_socket(path: str, timeout: float = 5.0) -> None:
    deadline = time.monotonic() + timeout
    while time.monotonic() < deadline:
        if os.path.exists(path):
            return
        time.sleep(0.05)
    raise RuntimeError(f"pebble socket {path} did not appear within {timeout}s")


with tempfile.TemporaryDirectory() as pebble_dir:
    socket_path = os.path.join(pebble_dir, ".pebble.socket")
    env = {**os.environ, "PEBBLE": pebble_dir}
    proc = subprocess.Popen(
        ["pebble", "run", "--create-dirs", "--hold"],
        env=env,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
    )
    try:
        wait_for_socket(socket_path)
        client = pebble.Client(socket_path=socket_path)
        remote_path = os.path.join(pebble_dir, "x")
        client.push(remote_path, "hello\n")
        # Correct usage: file object closed by context manager.
        with client.pull(remote_path) as f:
            _ = f.read()
        # LEAK: file object discarded without close().
        _ = client.pull(remote_path).read()
    finally:
        proc.terminate()
        proc.wait()
```

We should update our documentation to use the:

```python
with container.pull(remote_path) as f:
    content = f.read()
```

pattern.

tonyandrewmeyer:
`Container.pull` (and the `Client.pull` it wraps) return an opened file that is never explicitly closed. This leaks a resource until garbage collection (which then emits a warning), and in the case of `encoding=None` can leak that resource until process end. This is not a significant problem, but we should document the clean, rather than leaky, way to use these methods.

The how-to guide has poor suggestions as well, but those are all being replaced in #2470 (to use pathops) so are not worth fixing here.

Fixes #2504


docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -21,13 +21,14 @@ There are many additional features, including the ability to send raw bytes (by
 
 ## Pull
 
-To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`.
+To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`. Use it as a context manager (`with` block) so the underlying file is closed promptly — leaving it for the garbage collector emits an `unclosed file` `ResourceWarning` (and, with `encoding=None`, can pin a connection-side file descriptor until the next collection).
```
tonyandrewmeyer:
```suggestion
To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`. Use it as a context manager (`with` block) so the underlying file is closed promptly.
```

ops/model.py
```diff
@@ -2762,6 +2762,15 @@ def pull(self, path: str | PurePath, *, encoding: str = 'utf-8') -> TextIO: ...
     def pull(self, path: str | PurePath, *, encoding: str | None = 'utf-8') -> BinaryIO | TextIO:
         """Read a file's content from the remote system.
 
+        The returned object is a context manager; use ``with`` so the
+        underlying file is closed promptly, rather than relying on garbage
+        collection (which can warn about unclosed files and, with
+        ``encoding=None``, leak a file descriptor on the local end of the
+        Pebble HTTP connection until the next collection cycle)::
```
```suggestion
        underlying file is closed promptly, rather than relying on garbage
        collection::
```

tonyandrewmeyer:
@dwilding and @james-garner-canonical: in the issue, James suggested using `pathops`, which handles `pull` correctly. I'm not sure how much we are using `pathops` within our docs - I imagine not in reference docs, but maybe in the how-to guides? If we would rather adapt the bits of doc to use that approach, I'm happy to do that.

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -21,13 +21,14 @@ There are many additional features, including the ability to send raw bytes (by
 
 ## Pull
 
-To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`.
+To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`. Use it as a context manager (`with` block) so the underlying file is closed promptly.
```
james-garner-canonical: The changes to this file clash with the changes in #2470

ops/model.py
```diff
@@ -2762,6 +2762,13 @@ def pull(self, path: str | PurePath, *, encoding: str = 'utf-8') -> TextIO: ...
     def pull(self, path: str | PurePath, *, encoding: str | None = 'utf-8') -> BinaryIO | TextIO:
         """Read a file's content from the remote system.
 
+        The returned object is a context manager; use ``with`` so the
+        underlying file is closed promptly, rather than relying on garbage
+        collection::
```
james-garner-canonical: Perhaps more directly:
```suggestion
        Use the returned object as a context manager, otherwise ``close()``
        must be called manually to avoid memory leaks::
```

ops/model.py
```diff
@@ -2770,7 +2777,8 @@ def pull(self, path: str | PurePath, *, encoding: str | None = 'utf-8') -> Binar
         Returns:
             A readable file-like object, whose ``read()`` method will return
             strings decoded according to the specified encoding, or bytes if
-            encoding is ``None``.
+            encoding is ``None``. Close it when done, or use it as a context
+            manager.
```
james-garner-canonical: Could order the preferred usage first?
```suggestion
            encoding is ``None``. Use as a context manager, or call ``close()`` when done.
```

ops/pebble.py
```diff
@@ -2602,6 +2602,10 @@ def pull(
     ) -> BinaryIO | TextIO:
         """Read a file's content from the remote system.
 
+        The returned object is a context manager; use ``with`` so the
+        underlying file is closed promptly rather than relying on garbage
+        collection.
```
james-garner-canonical: For convenience, the same change as above, though TBH I don't mind if the lower level one is phrased differently.
```suggestion
        Use the returned object as a context manager, otherwise ``close()``
        must be called manually to avoid memory leaks::
```

ops/pebble.py
```diff
@@ -2610,7 +2614,8 @@ def pull(
         Returns:
             A readable file-like object, whose read() method will return str
             objects decoded according to the specified encoding, or bytes if
-            encoding is ``None``.
+            encoding is ``None``. Close it when done, or use it as a context
+            manager.
```
james-garner-canonical: Ditto
```suggestion
            encoding is ``None``. Use as a context manager, or call ``close()`` when done.
```

james-garner-canonical (changes requested):
The changes to the docstrings and the custom notices example are a nice fix, thanks!

The changes to the managing files in the workload container guide clash with #2470. I'd suggest dropping the changes to the file made in this PR, and adding any that are relevant to #2470 (though I'm not sure if any of them are since we're switching to ``pathops`` for most examples -- but maybe worth mentioning the file closing issue doesn't affect ``pathops``?).

Other suggestions are all non-blocking.

tonyandrewmeyer:
> The changes to the managing files in the workload container guide clash with #2470.

Ah yes, the PR I keep forgetting that I have only half-reviewed. Fixed now! Thanks for calling that out.

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -137,7 +138,8 @@ In this case, if the charm were to:
 ```python
 def _on_start(self, _):
     foo = self.unit.get_container('foo')
-    content = foo.pull('/local/share/config.yaml').read()
+    with foo.pull('/local/share/config.yaml') as f:
+        content = f.read()
```
tonyandrewmeyer:
```suggestion
    content = foo.pull('/local/share/config.yaml').read()
```

docs/howto/manage-containers/manage-files-in-the-workload-container.md
```diff
@@ -21,13 +21,14 @@ There are many additional features, including the ability to send raw bytes (by
...
-To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`.
+To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`. Use it as a context manager (`with` block) so the underlying file is closed promptly.
 
 The files API doesn't currently support update, so to update a file you can use `pull` to perform a read-modify-write operation, for example:
 
 ```python
 # Update port to 8888 and restart service
-config = container.pull('/etc/pg/postgresql.conf').read()
+with container.pull('/etc/pg/postgresql.conf') as f:
+    config = f.read()
```
tonyandrewmeyer:
```suggestion
config = container.pull('/etc/pg/postgresql.conf').read()
```

docs/howto/manage-containers/manage-files-in-the-workload-container.md
tonyandrewmeyer:
```suggestion
To read a file from the workload, use [`Container.pull`](ops.Container.pull), which returns a file-like object that you can `read()`.
```

docs/howto/manage-containers/manage-pebble-custom-notices.md
```diff
@@ -31,8 +31,8 @@ class PostgresCharm(ops.CharmBase):
         if event.notice.key == "canonical.com/postgresql/backup-done":
             path = event.notice.last_data["path"]
             logger.info("Backup finished, copying %s to the cloud", path)
-            f = event.workload.pull(path, encoding=None)
-            s3_bucket.upload_fileobj(f, "db-backup.sql")
+            with event.workload.pull(path, encoding=None) as f:
+                s3_bucket.upload_fileobj(f, "db-backup.sql")
```
tonyandrewmeyer:
```suggestion
            f = event.workload.pull(path, encoding=None)
            s3_bucket.upload_fileobj(f, "db-backup.sql")
```

ops/model.py
tonyandrewmeyer: I was reluctant to tell people to `close()` because you really ought to put that in a `finally` or `contextlib.closing` unless you're *really* sure nothing can stop it happening, and if you're doing that then you might as well just be using the returned file as a context manager.

james-garner-canonical: Good point, what you have is better than my suggestion. An alternative suggestion then would be:
```suggestion
        The returned object is a context manager; use ``with`` so the
        underlying file is closed promptly and avoids memory leaks::
```
I guess I feel that mentioning garbage collection is both too low-level and might lead to a false sense of security (given that with `encoding=None` we leak till process end).

james-garner-canonical (approved):

ops/pebble.py
dwilding: I agree with this and James's other comments about mentioning the context manager first. For this instance, we don't need `::` at the end, right?

dwilding (approved):

ops/pebble.py
tonyandrewmeyer: I never remember when you need two colons with ReST. I'll double check :)

ops/pebble.py
```diff
@@ -2602,6 +2602,9 @@ def pull(
     ) -> BinaryIO | TextIO:
         """Read a file's content from the remote system.
 
+        Use the returned object as a context manager, otherwise ``close()``
+        must be called manually to avoid memory leaks::
```
tonyandrewmeyer:
```suggestion
        must be called manually to avoid memory leaks.
```

PR #2509, opened 2026-05-26
test: wait on ExecProcess after expected TypeError

tonyandrewmeyer:
`test_wait_output_no_stdout` provokes a `TypeError` from `wait_output()` without subsequently waiting on the process, so the `ExecProcess` is GC'd without `_waited` being set and `ExecProcess.__del__` emits:

```
ResourceWarning: ExecProcess instance garbage collected without call to wait() or wait_output()
```

Under `-W error` that fails the test.

The makes a small (tiny!) change to call `process.wait()` after the expected `TypeError` so the leak detector doesn't fire on a test that's intentionally exercising a misuse-error path.

Context: surfaced while making operator's test suite pass under `-W error`.


dwilding (approved):
I can't confidently review the `pebble.py` change, as I'm not so familiar with threading - but I get the point of the change. I also get the point of the update to the test. And I checked that we already have a test `test_no_wait_call` that covers the `ResourceWarning`.

tromai (approved):

PR #2510, opened 2026-05-27
docs: clarify Secret manage-permission error in testing.State

### LINKED ISSUE #2077: Fix docs/impl inconsistency in Secret ownership in ops[testing]
Code:

```py
      def _check_can_manage_secret(
          self,
          secret: Secret,
      ):
          if secret.owner is None:
              raise SecretNotFoundError(
                  'this secret is not owned by this unit/app or granted to it. '
                  'Did you forget passing it to State.secrets?',
              )
          ...
```

Docs:

> owner: [Literal](https://docs.python.org/3/library/typing.html#typing.Literal)['unit', 'app', None] = None
> Indicates if the secret is owned by this unit, this application, or another application/unit.
>
> If None, the implication is that read access to the secret has been granted to this unit.

tonyandrewmeyer:
The error raised by `_check_can_manage_secret` when `secret.owner` is `None` is currently 'this secret is not owned by this unit/app or granted to it. Did you forget passing it to State.secrets?', but by the time that branch runs, the secret *is* in `State.secrets` (`_get_secret` would have raised otherwise) and `owner=None` is a valid state (meaning read access has been granted to this unit). The message contradicted the docs on `Secret.owner` (which are correct) and `State.secrets` and pointed users at a fix that wasn't the real problem.

This PR replaces it with 'You must own secret {id!r} to perform this operation', matching the equivalent error raised by Harness's `secret_info_get`. It also adds a code comment explaining the precondition, since there was confusion around this in an earlier attempt to fix the issue. It also tightens the existing `test_secret_permission_model` parametrized test to match on the message form so the fix can't regress silently.

Fixes #2077

### DIFF
--- a/testing/src/scenario/mocking.py
+++ b/testing/src/scenario/mocking.py
@@ -478,9 +478,12 @@ class _MockModelBackend(_ModelBackend):  # type: ignore
         secret: Secret,
     ):
         if secret.owner is None:
+            # The secret is in State.secrets (otherwise _get_secret would have
+            # raised already), so the charm has read access; it just doesn't
+            # own the secret and so cannot manage it. Set `Secret.owner` to
+            # 'unit' or 'app' in the test setup if the charm is meant to own it.
             raise SecretNotFoundError(
-                'this secret is not owned by this unit/app or granted to it. '
-                'Did you forget passing it to State.secrets?',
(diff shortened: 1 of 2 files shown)


dwilding (approved):
Thanks for the clear comments - very helpful

tromai (approved):
Thank you!

PR #2515, opened 2026-05-28
fix: preserve hyphens in Juju hook names in ops.testing

### LINKED ISSUE #2511: Does Scenario produces incorrect JUJU_HOOK_NAME?
While I was working on [this PR](https://github.com/canonical/opentelemetry-collector-operator/pull/293), I discovered some behaviour in Scenario that appears to be incorrect. 

When simulating events for relations whose endpoint names contain hyphens (e.g. `receive-ca-cert`), Scenario sets `JUJU_HOOK_NAME` to a mixed format with underscores in the prefix and hyphens in the suffix (e.g. `receive_ca_cert-relation-changed`), instead of the all-hyphens format that real Juju provides (`receive-ca-cert-relation-changed`).

This causes unit tests to pass against incorrect hook name values, masking real bugs in charm code that depends on `JUJU_HOOK_NAME`.

### Step 1 — Normalisation

In the function [`_normalise_name`](https://github.com/canonical/operator/blob/bb9ae3dc2e7d45a734bf9bfe217e9aaafcc68778/testing/src/scenario/state.py#L369)...

```python
def _normalise_name(s: str):
    """Event names, in Scenario, uniformly use underscores instead of dashes."""
    return s.replace('-', '_')
```

all dashes become underscores: `"receive-ca-cert-relation-changed"` → `"receive_ca_cert_relation_changed"`

### Step 2 — `_EventPath.__new__`

Then in [`_EventPath.__new__`](https://github.com/canonical/operator/blob/bb9ae3dc2e7d45a734bf9bfe217e9aaafcc68778/testing/src/scenario/state.py#L2146)

```python
def __new__(cls, string: str):
    string = _normalise_name(string)
    instance = super().__new__(cls, string)
    instance.name = name = string.split('.')[-1]
    instance.owner_path = string.split('.')[:-1] or ['on']
    return instance
```

For `"receive_ca_cert_relation_changed"`:
- `suffix` = `"_relation_changed"` (matched from [`_RELATION_EVENTS_SUFFIX`](https://github.com/canonical/operator/blob/bb9ae3dc2e7d45a734bf9bfe217e9aaafcc68778/testing/src/scenario/state.py#L110))
- `prefix` = `"receive_ca_cert"`

### Step 3 — [`_juju_name` construction](https://github.com/canonical/operator/blob/bb9ae3dc2e7d45a734bf9bfe217e9aaafcc68778/testing/src/scenario/state.py#L2253)

```python
# This is the event name as Juju provides it, with dashes not underscores.
object.__setattr__(self, '_juju_name', f'{path.prefix}{path.suffix.replace("_", "-")}')
```

Only the **suffix** gets underscores converted back to hyphens:
- `path.prefix` = `"receive_ca_cert"` (underscores preserved)
- `path.suffix.replace("_", "-")` = `"-relation-changed"`
- Result: `"receive_ca_cert-relation-changed"` ← **WRONG**

### Step 4 — [Assignment to `JUJU_HOOK_NAME` ](https://github.com/canonical/operator/blob/bb9ae3dc2e7d45a734bf9bfe217e9aaafcc68778/testing/src/scenario/_runtime.py#L77)

```python
'JUJU_HOOK_NAME': '' if event._is_action_event else event._juju_name,
```

## Expected behaviour

`JUJU_HOOK_NAME` should be set to `"receive-ca-cert-relation-changed"` (all hyphens), matching real Juju behaviour.

[From Juju logs](https://github.com/canonical/opentelemetry-collector-operator/issues/288) I see:
```
ran "receive-ca-cert-relation-created" hook (via hook dispatching script: dispatch)
```

## how to reproduce it

```python
from ops.testing import Context, Relation, State
from charm import MyCharm

ctx = Context(MyCharm)
rel = Relation("receive-ca-cert", remote_app_data={"certificates": "[]"})
state = State(leader=True, relations=[rel])

# During ctx.run(), JUJU_HOOK_NAME will be "receive_ca_cert-relation-changed"
# instead of the correct "receive-ca-cert-relation-changed"
ctx.run(ctx.on.relation_changed(rel), state)
event = ctx.on.relation_changed(rel)
print(event._juju_name) # receive_ca_cert-relation-changed
```

gcomneno:
Fixes #2511.

## Summary

Preserve Juju hook-name spelling in `ops.testing` for events whose entity names contain dashes or underscores.

Juju keeps the entity name portion of a hook name exactly as declared in charm metadata, for example relation endpoints, storage names, or workload/container names. The event-type suffix uses dashes.

For example:

- endpoint `receive-ca-cert` + relation changed -> `receive-ca-cert-relation-changed`
- endpoint `foo_bar` + relation changed -> `foo_bar-relation-changed`

Previously, Scenario normalised the full event path to underscores and then converted only the event suffix back to dashes. For endpoints such as `receive-ca-cert`, this produced mixed hook names like:

`receive_ca_cert-relation-changed`

This change preserves the original entity prefix for relation, storage, and workload events, while still rendering the event suffix in Juju's dash-separated form.

## Testing

```bash
uv tool run --with tox-uv tox -e unit -- testing/tests/test_runtime.py -k 'juju_hook_name or env_clean' -rs -vv --tb=short

uv tool run --with tox-uv tox -e unit -- \
  testing/tests/test_runtime.py \
  testing/tests/test_context_on.py \
  testing/tests/test_consistency_checker.py \
  -k 'juju_hook_name or relation_complex_name or relation_unit_events or relation_changed' \
  -rs -vv --tb=short

ruff check testing/src/scenario/state.py testing/tests/test_runtime.py testing/tests/test_e2e/test_event.py
ruff format --check testing/src/scenario/state.py testing/tests/test_runtime.py testing/tests/test_e2e/test_event.py
git diff --check
```

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -2245,12 +2245,41 @@ class _Event:  # type: ignore
 
     _owner_path: list[str] = dataclasses.field(default_factory=list)
 
+    # The event name as Juju provides it. Set in __post_init__ via
+    # object.__setattr__ (frozen dataclass), so it's excluded from init.
+    _juju_name: str = dataclasses.field(init=False)
+
     def __post_init__(self):
+        # The original (un-normalised) name, so that we can recover the exact
+        # spelling of the entity the event refers to (see _juju_name below).
+        original_name = str(self.path).split('.')[-1]
         path = _EventPath(self.path)
         # bypass frozen dataclass
         object.__setattr__(self, 'path', path)
-        # This is the event name as Juju provides it, with dashes not underscores.
-        object.__setattr__(self, '_juju_name', path.name.replace('_', '-'))
+        # This is the event name as Juju provides it. Juju keeps the entity
+        # name (a relation endpoint, storage name, or container name) exactly
+        # as it is declared in the charm metadata -- which may contain dashes
+        # *or* underscores -- and only ever uses dashes in the event-type
+        # suffix. For events that are not tied to such an entity, the whole
+        # name uses dashes. See #2511.
+        object.__setattr__(self, '_juju_name', self._build_juju_name(original_name, path))
+
+    @staticmethod
+    def _build_juju_name(original_name: str, path: _EventPath) -> str:
+        suffix = path.suffix
+        if suffix and path.type in (
+            _EventType.RELATION,
+            _EventType.STORAGE,
+            _EventType.WORKLOAD,
+        ):
+            # Preserve the entity name (the prefix) verbatim; only the suffix
+            # is normalised to dashes. ``original_name`` and the normalised
+            # ``path.name`` have the same length (``_normalise_name`` only
+            # swaps dashes for underscores), so the suffix can be stripped by
+            # length to recover the original prefix spelling.
+            prefix = original_name[: len(original_name) - len(suffix)]
+            return f'{prefix}{suffix.replace("_", "-")}'
+        return path.name.replace('_', '-')
 
     @property
     def _path(self) -> _EventPath:
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -2250,7 +2250,7 @@ class _Event:  # type: ignore
         # bypass frozen dataclass
         object.__setattr__(self, 'path', path)
         # This is the event name as Juju provides it, with dashes not underscores.
-        object.__setattr__(self, '_juju_name', f'{path.prefix}{path.suffix.replace("_", "-")}')
+        object.__setattr__(self, '_juju_name', path.name.replace('_', '-'))
 
     @property
     def _path(self) -> _EventPath:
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
Thanks @gcomneno !

I expanded the code and tests somewhat so I shouldn't review this.

james-garner-canonical (commented):
I wonder if it would be cleaner to extend `_EventPath` to store the original prefix as well (`instance._original_prefix = original_string[: len(instance.prefix)]`).

Then `_Event._build_juju_name` doesn't need to extract the prefix with arithmetic.

james-garner-canonical:
The PR description also needs to be updated to reflect the actually desired behaviour (see comments in #2511).

gcomneno:
Thanks for the review. I pushed a small refactor that keeps the original prefix on `_EventPath`, so `_Event._build_juju_name()` no longer has to recover it with suffix-length arithmetic. I also updated the PR description to reflect the dash/underscore behaviour discussed in #2511.

PR #2516, opened 2026-05-28
docs: add k8s tutorial integration test examples

### LINKED ISSUE #2514: Add config and action integration tests to K8s tutorial charms
These chapters of the K8s tutorial don't show how to write integration tests:

- [Make your charm configurable](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable/)
- [Expose operational tasks via actions](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions/)

@tonyandrewmeyer and I discussed this today. We don't think it's valuable to extend the time it takes to complete the tutorial with another two rounds of running integration tests. So here's our plan:

1. Add integration tests to the relevant example charms. (Probably a good idea to comment that the tests go beyond the tutorial.)

2. In the tutorial chapters, mention that adding integration tests is an exercise for the reader. They can refer to [How to write integration tests for a charm](https://documentation.ubuntu.com/ops/latest/howto/write-integration-tests-for-a-charm/), which has clearly-named sections that should help. If the reader is stuck, they can inspect the example charms.

gcomneno:
Fixes #2514.

## Summary

Add integration-test examples for the Kubernetes tutorial chapters that introduce configuration and actions.

This change:

- adds an integration test to `examples/k8s-2-configurable` that verifies invalid `server-port` config blocks the charm, then resets the config and waits for the unit to become active again
- extends the `examples/k8s-4-action` integration test to run the `get-db-info` action after the PostgreSQL integration is established
- updates both tutorial chapters to point readers toward integration testing as a follow-up exercise
- links readers to the integration testing guide

## Testing

```bash
ruff check \
  examples/k8s-2-configurable/tests/integration/test_charm.py \
  examples/k8s-4-action/tests/integration/test_charm.py

ruff format --check \
  examples/k8s-2-configurable/tests/integration/test_charm.py \
  examples/k8s-4-action/tests/integration/test_charm.py

git diff --check
```

The integration tests were updated but not run locally because they require a Juju/Kubernetes environment and packed charms.

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
@@ -217,6 +217,12 @@ def test_get_db_info_action_show_password():
 
 Run `tox -e unit` to check that all tests pass.
 
+## Go further with integration tests
+
+Adding an integration test for this feature is left as an exercise for the reader. For guidance, see {ref}`How to write integration tests for a charm <write-integration-tests-for-a-charm>`.
+
+The example charm for this chapter includes an integration test that runs the `get-db-info` action against a deployed charm.
+
 ## Review the final code
 
 For the full code, see [our example charm for this chapter](https://github.com/canonical/operator/tree/main/examples/k8s-4-action).
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
@@ -279,6 +279,12 @@ def test_config_changed_invalid_port():
 
 Run `tox -e unit` to check that all tests pass.
 
+## Go further with integration tests
+
+Adding an integration test for this feature is left as an exercise for the reader. For guidance, see {ref}`How to write integration tests for a charm <write-integration-tests-for-a-charm>`.
+
+The example charm for this chapter includes an integration test that checks the charm's behaviour when `server-port` is set to an invalid value.
+
 ## Review the final code
 
 For the full code, see [our example charm for this chapter](https://github.com/canonical/operator/tree/main/examples/k8s-2-configurable).
--- a/examples/k8s-2-configurable/tests/integration/test_charm.py
+++ b/examples/k8s-2-configurable/tests/integration/test_charm.py
@@ -39,3 +39,16 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     }
     juju.deploy(charm, app=APP_NAME, resources=resources)
     juju.wait(jubilant.all_active)
+
+
+@pytest.mark.juju_setup
+def test_invalid_server_port_blocks_unit(juju: jubilant.Juju):
+    """Verify that invalid config blocks the charm.
+
+    This integration test goes beyond the tutorial instructions.
+    """
+    juju.config(APP_NAME, {"server-port": "22"})
+    juju.wait(jubilant.all_blocked)
+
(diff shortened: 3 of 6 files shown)


dwilding (changes requested):
On an initial review, I see a couple of issues with the implementation:

- When updating one charm in the sequence, we need to update all following charms in the sequence accordingly.

- I'm not keen on expanding `test_database_integration` to test more things. This test is marked with `juju_setup`, but the action test shouldn't be. The config test shouldn't have `juju_setup` either - it's best to use that marker for tests that deploy and integrate charms.

My main concern is that the additions in this PR haven't been run for real.

I'd really like to see that you've gone through the K8s tutorial for yourself, packed the charms, and successfully run the expanded integration tests.

Please could you rework this PR and go through the tutorial as you do that? (Also any first-hand feedback you have from the tutorial would be very much appreciated!)

gcomneno:
Thanks for the review, Dave.

I’ve reworked the PR to address the points you raised:

- updated the later K8s tutorial examples consistently after changing the earlier ones;
- kept the database integration test focused on deploying/integrating PostgreSQL;
- split the action check into a separate test without `juju_setup`;
- added the invalid config test separately, also without `juju_setup`;
- changed blocked waits that are meant to validate only the tutorial app to scope `jubilant.all_blocked` to `APP_NAME`;
- increased PostgreSQL-related active waits to 10 minutes, since the database charm can take longer to settle locally.

Validation I was able to complete locally:

- `examples/k8s-2-configurable`: full integration passed, `2 passed in 59.55s`;
- `examples/k8s-3-postgresql`: lint passed; full integration passed, `3 passed in 301.22s`;
- `examples/k8s-4-action`: lint passed; full integration passed, `4 passed in 249.03s`;
- `examples/k8s-5-observe`: lint passed; non-COS pytest collection was verified as selecting only the intended 4 tests and deselecting the 3 COS tests.

For `k8s-5-observe`, I also started a non-COS run and the PostgreSQL integration path reached `PASSED`, with `fastapi-demo` becoming active after integrating `postgresql-k8s`. However, I then hit local environment problems while cleaning up/rerunning: after removing generated charmcraft artifacts, `charmcraft pack` now fails inside the LXD build instance during `apt-get update` with a network error reaching `archive.ubuntu.com:80`. Host networking works, DNS inside the container works, LXD bridge NAT/forwarding appear configured, and recreating the charmcraft base instance did not resolve it.

So I don’t want to claim a clean full rerun for `k8s-5-observe` locally. The code changes are pushed, and CI or another clean environment should give a more reliable signal for that final case than my currently-broken local charmcraft/LXD networking.

dwilding:
I'm afraid this PR isn't going in a direction I like. The latest comment looks very much like an agent response to my previous comment, when I need to see that a human is working through the proposed changes in the context of the tutorial. I'm going to close the PR but would be very happy to consider a human-driven contribution in the future.

gcomneno:
Thanks for the feedback, Dave.

I understand closing the PR if you feel the proposed change does not fit the direction or flow of the tutorial.

However, I want to clarify one point. The project’s CONTRIBUTING guide says that PRs partly or entirely generated using AI tools are welcome, provided that the human contributor reviews the work and takes personal responsibility for its quality and suitability. That is how I treat the contributions I submit.

The relevant question should be whether I understand and stand behind the work, not whether every sentence was produced with bare hands and a candle. We all use tools while working, from compilers and linters to calculators. I would find it odd to question whether someone owns a calculation because they did not compute a cube root by hand.

If the issue is the technical or documentation fit of this specific change, that is fair feedback. But I do not think it is fair to frame the concern as whether a human is working on it based on how a comment reads.

Given this, I do not think this is the right repository for me to keep contributing to. Thanks for the reviews and for the time spent on my previous contributions.

PR #2518, opened 2026-05-29
ci: use the sys.monitoring coverage core

tonyandrewmeyer:
Set COVERAGE_CORE=sysmon for the coverage and coverage-tracing tox environments. The sys.monitoring core has much lower overhead than the default C trace core, especially with branch coverage. For me, with Python 3.14, this cut the coverage run from ~64s to ~40s on the unit suite, with identical results. It is a no-op on Python < 3.12, where coverage falls back to the C trace core.

See https://nedbatchelder.com/blog/202312/coveragepy_with_sysmonitoring for more background.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -84,6 +84,11 @@ commands =
 [testenv:coverage]
 description = Run unit tests with coverage
 allowlist_externals = mv
+setenv =
+    # Use the sys.monitoring core (PEP 669): much lower overhead than the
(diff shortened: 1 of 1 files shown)


tromai (approved):
Thanks.

I verified that it worked on my local setup:
- With `python3.12`: it issued a warning and fell back to `sys.settrace`.
- With `python3.14`: it didn't issue any warning.

james-garner-canonical (approved):
Nice!

PR #2520, opened 2026-05-31
test: add integration tests for ops.hookcmds

tonyandrewmeyer:
When working on the ops.hookcmds addition, I ran out of time to finish adding integration tests, so that was dropped from the immediate deliverables. However, these are particularly well suited for integration tests, since the rest of the ops model is built on top of them, and they ensure ops has a validated picture of what Juju (hook commands) provide.

This PR adds a set of integration tests that validate ops.hookcmds against Juju 3.6 and 4.0, on both k8s and machine controllers. They are not intended to be comprehensive, but cover the core hook command functionality. Like the existing integration tests, they run on a schedule or manual trigger, as they are too slow to run on PRs.

Like the existing integration tests, a new test charm is added to exercise the functionality. This mostly has a bunch of actions that trigger sets of hook commands.

One weakness is that almost everything happens in a single hook, so the behaviour of Juju committing and presenting information in the next hook isn't exercised. However, this keeps the tests much simpler, and that seems more like Juju functionality than ops.hookcmds functionality. The tests are more about running the commands and parsing the output successfully, than the actual Juju behaviour.

[Passing run on my fork](https://github.com/tonyandrewmeyer/operator/actions/runs/26705626487).

It seems like I found two Juju 4 bugs, so I opened issues for those, and they are currently marked as xfail.


ops/hookcmds/_state.py
```diff
@@ -54,7 +54,7 @@ def state_get(key: str | None) -> dict[str, str] | str:
     if key is not None:
         key_result: str = json.loads(stdout)
         return key_result
-    result: dict[str, str] = json.loads(stdout)
+    result: dict[str, str] = json.loads(stdout) or {}
```
tonyandrewmeyer, line 60: This is needed for Juju 4. I'm not sure if it's a deliberate change or not. I've asked in https://github.com/juju/juju/issues/22523 so the Juju team should clarify.

If it's a Juju bug, we could not include this and add more xfail, although that wouldn't help ops users on Juju 4 (although we generally avoid working around Juju bugs in ops, these days).

test/integration/conftest.py
```diff
@@ -187,6 +189,19 @@ def cleanup():
...
+        lock_path = charm_dir / 'uv.lock'
+        lock_text = lock_path.read_text()
+        lock_text = re.sub(
+            r'(name = "ops".*?sdist = \{ hash = "sha256:)[0-9a-f]+("\s*\})',
+            rf'\g<1>{ops_hash}\g<2>',
+            lock_text,
+            flags=re.DOTALL,
+        )
+        lock_path.write_text(lock_text)
     except subprocess.CalledProcessError as e:
```
dwilding, line 191: I don't understand why this is needed. Is it because we won't maintain the hash of in-development ops in this charm's uv.lock file? I would have thought that uv.lock would get updated before the charm is packed. So is this because of a format/calculation difference between different uv versions?

.github/workflows/integration.yaml
```diff
@@ -69,5 +69,31 @@ jobs:
...
+      - run: tox -e integration -- --log-cli-level=INFO -s -k test_secrets
+        timeout-minutes: 20  # Juju 4 gets stuck while destroying the model
+
+  hookcmds:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: ['k8s', 'machine']
```
tromai, line 71: I wonder if we can define both `preset` and `juju-channel` in the same style? I assume they are both list of strings.

test/charms/test_hookcmds/src/charm.py
```diff
@@ -0,0 +1,381 @@
...
+        retrieved = hookcmds.state_get(key)
+        all_state = hookcmds.state_get(None)
+
+        hookcmds.state_delete(key)
+
+        after_delete = hookcmds.state_get(None)
+
+        event.set_results({
+            'retrieved': retrieved,
+            'types-match': str(retrieved == value).lower(),
```
tromai: Does `types-match` here mean comparing the type of `retrieved` and `value`?

test/charms/test_hookcmds/src/charm.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # --- Remove first revision only ---
+        hookcmds.secret_remove(secret_id, revision=1)
+
+        # --- Remove all remaining revisions ---
+        hookcmds.secret_remove(secret_id)
+
+        event.set_results({
+            'secret-id': secret_id,
+            'in-ids-list': str(any(i in secret_id or secret_id in i for i in ids)).lower(),
```
tromai: Just curious, what is the difference between `secret_id` and the ids in `ids`? This logic indicates that they might be sub-string of each other.

test/charms/test_hookcmds/src/charm.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if data_storage:
+            info = hookcmds.storage_get(data_storage[0])
+            event.set_results({
+                'total-storage-count': str(len(all_storage)),
+                'data-storage-count': str(len(data_storage)),
+                'storage-id': data_storage[0],
+                'storage-kind': info.kind,
+                'storage-location': str(info.location),
+                'has-location': str(bool(info.location)).lower(),
```
tromai: Does `str(bool(info.location)).lower()` suppose to check if `location` (of type `Path`) exist? Or does it only check if `info.location` is not None.

test/charms/test_hookcmds/src/charm.py
```diff
@@ -0,0 +1,381 @@
...
+    # Error path
+
+    def _on_trigger_relation_error(self, event: ops.ActionEvent):
+        """relation_list on a bogus id should raise hookcmds.Error."""
+        relation_id = int(event.params['relation-id'])
+        try:
+            hookcmds.relation_list(relation_id)
+        except hookcmds.Error as exc:
+            event.set_results({
+                'raised': type(exc).__name__,
```
tromai, line 333: `exc` here has type `hookcmds.Error` because of `except ... as ...`. Is the invocation of `type()` 's purpose: to easily add another excepted error type here later on ?

tromai (commented):
Thanks for the PR. I have some comments.

test/integration/conftest.py
tonyandrewmeyer: As discussed, I got fooled into keeping this even after I also was unsure. Verified it is not required.

.github/workflows/integration.yaml
tonyandrewmeyer: We could, and they are. Would you prefer that?

.github/workflows/integration.yaml
```diff
@@ -59,5 +59,33 @@ jobs:
...
+        juju-channel:
+          - 3/stable
+          - 4.0/stable
+          - 4.0/beta
+          - 4.1/beta
```
tonyandrewmeyer:
```suggestion
          - 4.0/edge
          - 4.1/edge
```

.github/workflows/integration.yaml
tromai: Sorry, on second thought, I think changing it doesn't create a major difference. And I don't have a preferred style. 

There is no need to change. I will close this discussion.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: Yes, it compared type(retrieved) == type(value). That handler-side check is gone now (refactoring after David's review): the handler returns the raw value and the test asserts on it directly.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: Yes, just a truthiness check. It was redundant with storage-location and got dropped in the refactoring.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: Yes, partly. More importantly, `except hookcmds.Error as exc` also catches subclasses, so `type(exc).__name__` records the actual concrete class that was raised rather than just  "Error". The test asserts on that name, so if relation_list ever raised a more specific subclass we'd see it directly.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: Yes, secret_add() and secret_ids() returned IDs in different forms across Juju versions (short uuid vs prefixed secret:uuid), and the substring check was a lazy workaround. Removed in the refactoring: the handler now returns the raw list and the test matches properly.

.github/workflows/integration.yaml
```diff
@@ -59,5 +59,33 @@ jobs:
...
+          - 4.0/stable
+          - 4.0/edge
+          - 4.1/edge
+
+    steps:
+      - run: sudo snap install --classic concierge
+      - run: >
+          sudo concierge prepare
+          --juju-channel=${{ matrix.juju-channel }}
+          --charmcraft-channel=3.x/stable
```
tonyandrewmeyer:
```suggestion
```

tonyandrewmeyer:
Heads up: I've got a follow-up branch (`hookcmds-followups-3-4-5` on my fork) that I'll open as a separate PR once this one merges, so as not to disturb the review state here.

What it adds:

- **Follow-up 3** — `test_secret_grant_revoke`. Deploys `any-charm` from `latest/beta` and integrates it via the test-hookcmds `anycharm` endpoint, then exercises `secret_grant` / `secret_revoke` across a real cross-app relation. Covers the grant path that the existing `test_secret_full_lifecycle` couldn't reach (no related app).
- **Follow-up 4** — `test_ports_endpoint_scoped`. New `test-ports-endpoint-scoped` action that opens TCP 7766 scoped to the `peer` endpoint, verifies `opened_ports(endpoints=True)` returns `endpoints=['peer']`, then closes it. Complements the existing `test_ports_open_close_cycle` (no-endpoint variant).
- **Follow-up 5** — `relation_model_get` was *not* added as a test. In 3.6 and 4.0 it always returns the local model UUID for peer relations (same UUID as `JUJU_MODEL_UUID`), so there's nothing distinct to assert against. Worth revisiting if/when the suite grows a cross-model relation, but a peer-only test would be tautological. Documented in PROGRESS notes alongside the branch.

Plus two non-test changes that landed on the same branch while validating the matrix:

- Drop `test/charms/test_hookcmds/uv.lock`. The lockfile pinned `ops==3.8.0.dev0` from before 3.7.1 was tagged, and uv refused to build `ops.tar.gz` ("Package metadata version `3.7.1` does not match given version `3.8.0.dev0`") on a charmcraft pack. The three sibling test charms (`test_relation`, `test_secrets`, `test_tracing`) ship the same pyproject shape with no lockfile and resolve cleanly against whatever the local tarball provides — matching that pattern means a future ops version bump won't break the hookcmds matrix again.
- Three `xfail` guards for Juju 4.0/k8s failures:
  - `test_network_get_returns_addresses` (pre-existing test from this PR): on 4.0/k8s, `network-get --format=json` emits two `bind-addresses` entries with the empty "k8s service" placeholder first and the cloud-container with the pod IP second. So `bind_addresses[0].addresses` is `[]`, and the action's `first-address` field is absent. Juju 3.6 returns a single non-placeholder entry; Juju 4.1 emits the same two entries but with the cloud-container first. Filed as juju/juju#22616.
  - `test_secret_grant_revoke` (follow-up 3): same uniter commit-phase root cause as the existing `_JUJU4_COMMIT_BUG` xfails. Symptom on 4.0/k8s is a driver-side `TimeoutError` rather than `jubilant.TaskError`, so the except clause widens to cover both.
  - `test_ports_endpoint_scoped` (follow-up 4): same timeout symptom; xfail with the same constant pending a more specific upstream issue.

Matrix is green on the fork: https://github.com/tonyandrewmeyer/operator/actions/runs/27271484908 (in flight as I type this; will update if anything red survives).

tonyandrewmeyer:
Matrix green: https://github.com/tonyandrewmeyer/operator/actions/runs/27272486017

All four `hookcmds` cells pass (machine/k8s × 3.6/4.0). The four 4.0/k8s xfails fire as expected (`test_network_get_returns_addresses` against juju/juju#22616; `test_secret_full_lifecycle`, `test_secret_grant_revoke`, `test_ports_endpoint_scoped` against the existing `_JUJU4_COMMIT_BUG` constant for juju/juju#22523, juju/juju#22524). Two unrelated cells are red: `secrets (4.0/stable)` (an existing operator-suite regression I'm seeing on every recent run — `test_secrets.py::test_add_with_meta[]` times out on `juju run … add-secret`, looks like the same uniter commit-phase family) and `test (microk8s, test_relation_units)` (probably flaky — green on the previous run on the same branch).

I'll open the follow-up PR off `hookcmds-followups-3-4-5` once this one lands.

test/charms/test_hookcmds/src/charm.py
```diff
@@ -0,0 +1,408 @@
...
+        gs = hookcmds.goal_state()
+        unit_names = sorted(gs.units.keys())
+        unit_statuses = [gs.units[n].status for n in unit_names]
+        # Relations may be empty if no peer has joined yet.
+        event.set_results({
+            'unit-count': str(len(unit_names)),
+            'unit-names': ','.join(unit_names),
+            'unit-statuses': ','.join(unit_statuses),
+            'relation-endpoint-count': str(len(gs.relations)),
+        })
```
dwilding, line 249: Since we don't assert on `relation-endpoint-count`, how about:

```suggestion
        # Don't process gs.relations; it may be empty if no peer has joined yet.
        event.set_results({
            'unit-count': str(len(unit_names)),
            'unit-names': ','.join(unit_names),
            'unit-statuses': ','.join(unit_statuses),
        })
```

Or is it preferable to access `gs.relations` in code, effectively checking that it exists?

test/integration/test_hookcmds.py
```diff
@@ -0,0 +1,537 @@
...
+# Storage (storage_list / storage_get)
+
+
+def test_storage_list_and_get(juju: jubilant.Juju, any_unit: str):
+    """storage_list returns the attached data storage; storage_get has a path."""
+    task = juju.run(any_unit, 'test-storage')
+    assert task.success
+    r = task.results
+    data_count = int(r['data-storage-count'])
+    assert data_count >= 1, 'Expected at least one data storage instance'
```
dwilding, line 393: I don't fully understand this line. One of the paths in `_on_test_storage` sets `data-storage-count` to 0.

Also, should we assert on `total-storage-count`?

test/integration/test_hookcmds.py
```diff
@@ -0,0 +1,537 @@
...
+
+# Resources (resource_get)
+
+
+def test_resource_get(juju: jubilant.Juju, any_unit: str):
+    """resource_get returns a non-empty path to the cached resource file."""
+    task = juju.run(any_unit, 'test-resource-get')
+    assert task.success
+    path = task.results['path']
+    assert path  # non-empty path string
```
dwilding: Shall we assert on `path.name`? Since we know that (I think)

test/integration/test_hookcmds.py
```diff
@@ -0,0 +1,537 @@
...
+    """goal_state() returns GoalState with the expected unit names."""
+    task = juju.run(any_unit, 'test-goal-state')
+    assert task.success
+    r = task.results
+    unit_count = int(r['unit-count'])
+    assert unit_count >= 1
+    unit_names = r['unit-names'].split(',')
+    assert any(u.startswith('test-hookcmds/') for u in unit_names)
+    # Every unit should be in an 'alive' or 'waiting' goal state.
+    statuses = r['unit-statuses'].split(',')
```
dwilding, line 355: Should we assert that `len(statuses) == len(unit_names) == unit_count`? (I'm hypothesising that we'd expect that)

test/integration/test_hookcmds.py
```diff
@@ -0,0 +1,537 @@
...
+    """Full state lifecycle: set, get by key, get all, delete, verify gone."""
+    try:
+        task = juju.run(
+            any_unit,
+            'test-unit-state',
+            params={'key': 'testkey', 'value': 'testvalue'},
+        )
+    except jubilant.TaskError:
+        if _juju_major(juju) >= 4:
+            pytest.xfail(_JUJU4_COMMIT_BUG)
```
dwilding: Shall we add `strict=True` in case we forget to remove `xfail` when a fixed Juju is available? Similarly for the other `xfail`s.

dwilding (approved):
I think this logic split is clearer than previously, despite having to convert non-string values to strings for the action results. Thank you!

tromai (approved):
Thank you for replying to my questions. And I appreciate the heads up too.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: Good catch. Accessing gs.relations is deliberate, it makes sure that you can do that without crashing, so I'd rather assert on the count than drop the field. I'll add an assert and similar for the goal-state ones.

test/integration/test_hookcmds.py
tonyandrewmeyer: Nice catch. The else branch is really dead code. The storage is always attached in this deployment, so the only readers of data-storage-count: '0' would be a broken setup. I've switched it to event.fail(...) to be explicit, and added an assert as you suggested.

test/integration/test_hookcmds.py
tonyandrewmeyer: Ah, either TIL or today I re-learnt. Yes, thanks!

PR #2521, opened 2026-06-01
chore: drop the unused importlib-metadata dependency declaration

tonyandrewmeyer:
ops uses the stdlib importlib.metadata module (in ops/testing.py for the ops-scenario version check), not the PyPI importlib-metadata backport package. The two APIs we use, `.version()` and `PackageNotFoundError`, have been in stdlib since Python 3.8, which is the minimum supported version, so the backport is not required at any supported Python.

It remains a transitive dependency of opentelemetry-api on Python < 3.10.2, so the resolved install set on those Pythons is functionally unchanged; on newer Pythons it is no longer present in the resolved tree.

This aligns 2.23-maintenance with main, where #1980 dropped the direct declaration as part of a broader cleanup. More importantly, it makes it clearer what dependencies are actually required for building ops.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,6 @@ dependencies = [
     "websocket-client==1.*",
     # Wide version range to preempt conflicts when charms pin a version.
     "opentelemetry-api~=1.0",
-    "importlib-metadata",
 ]
 dynamic = ["version"]
 
--- a/uv.lock
+++ b/uv.lock
@@ -1939,8 +1939,6 @@ wheels = [
 name = "ops"
 source = { editable = "." }
 dependencies = [
-    { name = "importlib-metadata", version = "8.5.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
-    { name = "importlib-metadata", version = "8.7.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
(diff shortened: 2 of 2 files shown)


dwilding (approved):
> This aligns 2.23-maintenance with main, where https://github.com/canonical/operator/pull/1980 dropped the direct declaration as part of a broader cleanup. More importantly, it makes it clearer what dependencies are actually required for building ops.

Why don't we do the broader cleanup here too? Since Ops 2.23 pins to an exact version of `ops-scenario`, it seems like the same justification as #1980 would apply here. I suppose we need to limit the code changes in 2.23, whereas removing a dependency has a tangible upside.

tonyandrewmeyer:
> Why don't we do the broader cleanup here too?

I felt like it was a more substantive change and all I really need is getting rid of the unused dependency. However, it's a good point that the same reasoning applies, and we're being wider with what we backport to 2.23.

However, there is a difference in behaviour if instead of installing `ops[testing]`, which pins an exact version of `ops-scenario`, you install `ops` and `ops-scenario` explicitly, where nothing forces a specific Scenario version. The 2.23 ops handles that fairly gracefully, whereas in 3.x we basically said "use ops[testing] not ops-scenario directly" (plus ops-scenario pins *ops* specifically now, so you somewhat get it the other way around, but older versions of `ops-scenario` didn't do that, and maybe people are using those with older ops.

So on balance my preference is to play it safe and just do the minimal change.

tromai (approved):

PR #2523, opened 2026-06-02
refactor: avoid yaml.load() to silence security scanner false positives

tonyandrewmeyer:
This PR replaces `yaml.load()` calls with direct loader instantiation in `ops/_private/yaml.py` and `ops/storage.py`. This sidesteps security scanner findings without changing behaviour: both paths use `SafeLoader` and produce identical output.

See https://github.com/canonical/jubilant/pull/331 where the same change was made (with some short discussion in the threads).

The PR also renames the hardcoded-secret default in `add_oci_image_resource` (`harness.py`) via a local variable, which I think will avoid another scanner false positive and is a small enough change that it seems worth doing/

Finally, a comment is added on the `pickle.loads` call in `storage.py` explaining that the serialisation format cannot be changed without breaking backwards compatibility for charms with persisted `StoredState`. This will not help the scanning, but should be of use to anyone investigating in the future.


ops/_private/harness.py
```diff
@@ -701,10 +701,13 @@ def add_oci_resource(
...
         """
         if not contents:
+            # Use a variable to avoid pattern-based hardcoded-secret scanners
+            # flagging this test-fixture default value.
+            default_cred = 'password'
             contents = {
                 'registrypath': 'registrypath',
                 'username': 'username',
-                'password': 'password',
+                'password': default_cred,
```
james-garner-canonical, line 711: This is hilarious, yet helpful.

ops/storage.py
```diff
@@ -330,7 +335,7 @@ def _save_notice_list(self, notices: _Notices) -> None:
 
 # we load yaml.CSafeX if available, falling back to slower yaml.SafeX.
 _BaseDumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
-_BaseLoader = getattr(yaml, 'CSafeLoader', yaml.SafeLoader)
+_BaseLoader: type[_SafeLoader] = getattr(yaml, 'CSafeLoader', yaml.SafeLoader)
```
james-garner-canonical, line 339: Non-blocking idea, curious if it's worth considering for future:

Do you think it would be cleaner to inherit from `_private._yaml._safe_loader` instead of performing the `getattr` dance in two places?

ops/storage.py
```diff
@@ -407,7 +412,13 @@ def get(self, key: str) -> Any:
...
         if p.stdout == '' or p.stdout == '\n':
             raise KeyError(key)
-        return yaml.load(p.stdout, Loader=_SimpleLoader)  # noqa: S506
+        # Instantiate the loader directly rather than via yaml.load() to avoid
+        # false-positive "unsafe deserialization" warnings from pattern-based scanners.
+        loader = cast('_SafeLoader', _SimpleLoader(p.stdout))
+        try:
+            return loader.get_single_data()
+        finally:
+            loader.dispose()
```
james-garner-canonical: Non-blocking idea, curious if it's worth considering for future:

The repetition of this logic makes me wonder if it would be worth defining `_private._yaml.safe_load` so that we can write `_private._yaml.safe_load(p.stdout, safe_loader=_SimpleLoader)` here. For example:

```py
def safe_load(stream: str | TextIO, safe_loader: type[_SafeLoader] = _safe_loader) -> Any:
    """Same as yaml.safe_load, but use fast C loader if available."""
    # Instantiate the loader directly rather than via yaml.load() to avoid
    # false-positive "unsafe deserialization" warnings from pattern-based scanners.
    loader = safe_loader(stream)
    try:
        return loader.get_single_data()
    finally:
        loader.dispose()
```

ops/storage.py
```diff
@@ -407,7 +412,13 @@ def get(self, key: str) -> Any:
         p = _run(['state-get', key], stdout=subprocess.PIPE, check=True)
         if p.stdout == '' or p.stdout == '\n':
             raise KeyError(key)
-        return yaml.load(p.stdout, Loader=_SimpleLoader)  # noqa: S506
+        # Instantiate the loader directly rather than via yaml.load() to avoid
+        # false-positive "unsafe deserialization" warnings from pattern-based scanners.
+        loader = cast('_SafeLoader', _SimpleLoader(p.stdout))
```
james-garner-canonical: Is this `cast` necessary given that `_SimpleLoader` inherits from `_BaseLoader`, which has `type[_SafeLoader]`?

james-garner-canonical (approved):
Thanks for making the security scanner happy for us. This PR can be merged as-is IMO.

tromai (approved):
Thanks for the fix. 

I also agree with the PR description and discussions in https://github.com/canonical/jubilant/pull/331.

ops/storage.py
tonyandrewmeyer: I tired this, but the inheritance route trades the duplicated getattr for an additional cast in yaml.py, and I don't particularly like casts. You might be able to do a better job than me with this, feel free to open a PR if you're motivated.

james-garner-canonical: All good, as long as the low hanging fruit has been picked :)

PR #2524, opened 2026-06-02
docs: give each best-practice admonition a stable :name: anchor

tonyandrewmeyer:
This PR adds a `:name:` option to every best practice admonition in the docs, so each practice has its own anchor in the rendered HTML page (e.g. manage-libraries.html#best-practice-libraries-no-status-mutation) and a stable identifier independent of its surrounding text. A little JS makes the anchor visible so that people can copy it.

There are two benefits here:
* it's possible to link directly to the individual best practices, not just the surrounding block, or the collated list.
* other systems, like the automated checks on charmhub-listing-review, can use the IDs to uniquely track each item.

The `.github/update-best-practice-table.py` script is updated to extract `:name:` from each admonition and emit, in the generated docs/reuse/best-practices.txt, a trailing `<!-- id: <name> -->` HTML comment, so downstream automation (particularly canonical/charmhub-listing-review) can match practices by ID rather than fuzzy-matching the description text

A header comment at the top of best-practices.txt documents the convention.

A matching PR against charmcraft will add :name: to the Best practice admonitions there (assuming this lands). Admonitions without :name: continue to render exactly as before.

[Preview](https://canonical-ubuntu-documentation-library--2524.com.readthedocs.build/ops/2524/)


dwilding (commented):
Do you have a plan to make use of the ability to link to the bullets under "Follow best practices"? The anchors aren't discoverable when looking at the page in a browser, so unless you have anything planned for the review process, I think we should drop those anchors.

I think the ability to link directly to the best practice notes is more important. But this PR doesn't implement that for the "See foo" links at the end of each bullet. I'd be happy to review that in the same PR, if you want to implement it.

tonyandrewmeyer:
> Do you have a plan to make use of the ability to link to the bullets under "Follow best practices"? The anchors aren't discoverable when looking at the page in a browser, so unless you have anything planned for the review process, I think we should drop those anchors.

Ah, I thought that they would be there in the HTML, so usable even if not easily. But it seems not.

> I think the ability to link directly to the best practice notes is more important. But this PR doesn't implement that for the "See foo" links at the end of each bullet. I'd be happy to review that in the same PR, if you want to implement it.

I'll see if this is simple enough to do in this PR.

tonyandrewmeyer:
@dwilding i added a wee anchor via JS, what do you think?

dwilding:
Sorry for the delay in getting back to this one. I like the little link anchors on the best practice notes!

Unless I'm missing something, I still think it would be worth removing the targets from `best-practices.txt`:

```diff
- (best-practice-no-duplicate-model-config)=
  - Don't duplicate model-level configuration options that are controlled by {external+juju:ref}`juju model-config <command-juju-model-config>`. See [Define a configuration option](#define-a-configuration-option). <!-- id: best-practice-no-duplicate-model-config -->
```

I don't see a strong need to be able to deep link to the bullets.

dwilding (approved):
Will you do a separate PR to update the links at the end of each bullet so that they point to the `:name:` anchors?

tonyandrewmeyer:
> Will you do a separate PR to update the links at the end of each bullet so that they point to the `:name:` anchors?

I was going to but it turns out it's only a few lines of code so I've done it now. What do you think? It's nice they go straight there and it adds links to a couple that were missing. But you do jump to a place that is maybe *after* the context, since the best practice is often after the explanation. I'm not sure this is better than linking to the section containing the best practice.

dwilding:
There's also the accessibility angle of having four (!) identically-named links that go to different places:

> Screenshot from 2026-06-11 09-19-48

I think we revert this change. It was nice in principle, but I don't think it's better in practice. I'd say keep the visible anchors you added to the best practice notes though!

hpidcock (approved):

PR #1481, opened 2024-12-03
docs: go full Diátaxis, ingesting the relevant juju.is/docs/sdk documentation

tmihoc:
[**Preview on ReadTheDocs**](https://ops--1481.org.readthedocs.build/en/1481/)

This PR adds all the juju.is/docs content featuring Ops.

The content was originally formatted for Discourse; this PR reformats it to serve the RTD project the docs are destined for.

The PR also retitles all how-to guides as "Manage x", often with sections of the form "Implement the feature" and "Test the feature". This extends a pattern we already had in juju docs, which will continue in the juju RTD docs, and which we've also implemented in the new charmcraft RTD docs. Whereas in the past we tended to associate "manage" with a juju admin, here it's used as a catch-all for "how-to do all the things that it is possible to do about x in juju/charmcraft/ops". I believe the pattern is no clumsier than what we had before and having the same terminology across juju/charmcraft/ops , along with cross-referencing, will make it easier for people to piece together the story. (E.g., in Juju, "manage relations" will include `juju integrate`, while in charmcraft it'll mean declaring a particular endpoint and in ops it'll mean reacting to certain events and using or implementing an interface.)

Finally, the PR trims down the Reference and Explanation docs to just content directly relevant to ops. E.g., the charm taxonomy, events / lifecycle bits, charm maturity, charm best practices are all things that belong at a higher level of abstraction, directly in juju docs. E.g., in the charm best practices doc the first best practice we speak about is using ops. (A lot of that content should also be a lot leaner -- e.g., I'm working to incorporate all the charm SDK event content into the juju doc on hooks, and the charm maturity and charm best practices could really use a complete rethink as well, as they are way too verbose, always understandable, and often duplicate content that's better expressed at a more specific level.)

Some things that I think should be addressed in subsequent PRs:

[...]

PR #2539, opened 2026-06-03
chore: bump ruff to 0.15.14 and pyright to 1.1.410

tonyandrewmeyer:
This PR brings `ruff` and `pyright` up to newer versions and makes adjustments for those changes.

- Bump `ruff` from 0.11.2 to 0.15.14 and relock to pick up `pyright` 1.1.410.
- Add a ruff.toml in `docs/` that limits lines to 80 characters to avoid horizontal scrolling in the width-constrained doc view.
- Apply auto-fixes for the new ruff rules (RUF052/061/067/069/071/075/100, D420, F401) and configure ignores for rules that conflict with intentional patterns.
- Wrap `ops.framework.Framework._event_context`'s `yield` in `try/finally` so backend state is restored on exceptions (real cleanup bug surfaced by RUF075, although extremely unlikely to ever happen given where it is).
- Replace `os.path.commonprefix` in `ops.model.Container._build_destpath` with `os.path.commonpath` (ruff's RUF071 auto-fix). Same shape as the original, just component-based rather than character-by-character, so the partial-component false-match can't happen.
- Switch `@contextmanager` return annotations from `Iterator` to `Generator` in `testing/tests/test_charm_spec_autoload.py` per the new pyright deprecation.

This is a big one, but I think even though there is some noise now, it's good as a linter going forward:
- `--preview` formatter now reformats Python code blocks inside markdown — accepted across docs/, README, and the testing/ readmes.

There is a catch, though: our docs double-quote strings, and our code single-quotes. ruff doesn't seem to have any clean way of handling this (to be fair, it's a bit odd). One option would be to put a config file in `docs/` that inherits from the top level but changes this setting. However, that would miss the README and other similar files, and I assume we don't want those to change either (if we do, then I think it may be the tidier change). I went with running ruff a second time in tox, with a config option overriding the config.


tox.ini
```diff
@@ -45,13 +45,17 @@ dependency_groups = lint
 commands =
     ruff format --preview
     ruff check --preview --fix
+    # Second pass over markdown: format Python code blocks but preserve their
+    # existing quote style so doc examples keep their double quotes.
+    ruff format --preview --config "format.quote-style = \"preserve\"" --config "format.exclude = [\"**/*.py\"]" .
```
tonyandrewmeyer: Instead of "preserve", we could go with "double" but I think we want to keep the hand-crafted choice maybe?

james-garner-canonical: We should preserve initially to keep this PR sane to review, but a separate PR standardising either way would be fine with me.

That would let us drop the whole second `ruff format` invocation, which would be nice. In the meantime, a comment here and below as to why the second invocation is necessary (in addition to the one in `pyproject.toml`) would be helpful, unless you plan to do a quick follow-up to standardise.

STYLE.md
```diff
@@ -135,16 +139,17 @@ Function and method signatures should include a type annotation for the returned
 **Do:**
 
 ```python
-def method1(arg1: type1, arg2: type2) -> None:
-    ...
+def method1(arg1: type1, arg2: type2) -> None: ...
```
james-garner-canonical: This is idiomatic for stubs in code, but I don't like it for examples. How about:

```suggestion
def method1(arg1: type1, arg2: type2) -> None:
    return
```

STYLE.md
```diff
@@ -135,16 +139,17 @@ Function and method signatures should include a type annotation for the returned
...
 
 def method2() -> str:
     ...
     return 'Hello world!'
 
+
 class C:
-    def __init__(self, x: type1):
-        ...
+    def __init__(self, x: type1): ...
```
james-garner-canonical: And here:
```suggestion
    def __init__(self, x: type1):
        pass
```

pyproject.toml
```diff
@@ -206,6 +210,12 @@ ignore = [
     # Manual dict comprehension.
     "PERF403",
 
+    # `__init__` module should only contain docstrings and re-exports.
+    # ops/__init__.py wraps main() and imports ops_tracing at the bottom;
+    # ops/_private/__init__.py exposes a module-level tracer; ops/lib/__init__.py
+    # is the legacy library loader. All intentional.
+    "RUF067",
+
```
james-garner-canonical: Should we scope this ignore to the specific init files in question?

AGENTS.md
```diff
@@ -59,10 +59,12 @@ import ops
 import subprocess
 from typing import Generator  # typing is an exception
 
+
```
james-garner-canonical, line 62: Big AI increasing our token spend 😆

testing/tests/test_charm_spec_autoload.py
```diff
@@ -26,14 +26,16 @@ class MyCharm(CharmBase): pass
 
 
 @contextmanager
-def import_name(name: str, source: Path) -> Iterator[type[CharmBase]]:
+def import_name(name: str, source: Path) -> Generator[type[CharmBase]]:
```
james-garner-canonical, line 29: Nice, I'm somehow surprised that single argument `Generator` is accepted on our target Python versions.

ops/framework.py
```diff
@@ -950,10 +949,11 @@ def _event_context(self, event_name: str):
...
         backend._hook_is_running = event_name
-        yield
-        backend._hook_is_running = old_hook_is_running
-
-        self._event_name = old_event_name
+        try:
+            yield
+        finally:
+            backend._hook_is_running = old_hook_is_running
+            self._event_name = old_event_name
```
james-garner-canonical, line 956: Nice! Maybe promotes this from a chore to a fix?

docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -519,22 +520,24 @@ To handle these events, add the following lines to the `__init__` method of the
...
+    tinyproxy.uninstall()
+
+
+def wait_for_not_running(self) -> None:
+    """Wait for tinyproxy to not be running."""
+    for _ in range(3):
+        if not tinyproxy.is_running():
+            return
+        time.sleep(1)
+    raise RuntimeError("tinyproxy was still running after the expected time")
```
james-garner-canonical: These were intentionally indented so the reader could copy and paste them into a class body. How about this (with the real class name)? Less convenient since you can't copy the whole code block, but satisfies the linter ... likewise the other example above.

```suggestion
class Foo:
    ...

    def _on_stop(self, event: ops.StopEvent) -> None:
        """Handle stop event."""
        tinyproxy.stop()
        self.wait_for_not_running()
    
    
    def _on_remove(self, event: ops.RemoveEvent) -> None:
        """Handle remove event."""
        tinyproxy.uninstall()
    
    
    def wait_for_not_running(self) -> None:
        """Wait for tinyproxy to not be running."""
        for _ in range(3):
            if not tinyproxy.is_running():
                return
            time.sleep(1)
        raise RuntimeError("tinyproxy was still running after the expected time")
```

docs/howto/run-workloads-with-a-charm-machines.md
```diff
@@ -304,7 +305,8 @@ def test_install_calls_apt(monkeypatch: pytest.MonkeyPatch):
 
 
 def test_reload_config_sends_sigusr1(
-    monkeypatch: pytest.MonkeyPatch, tmp_path,
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path,
```
james-garner-canonical:
```suggestion
    tmp_path: pathlib.Path,
```

docs/howto/manage-secrets.md
```diff
@@ -326,13 +318,13 @@ Note that [`Secret.id`](ops.Secret.id), despite the name, is not really a unique
 Sometimes, before reconfiguring to use a new credential revision, the observer charm may want to peek at its contents (for example, to ensure that they are valid). Use `peek_content` for that:
 
 ```python
-    def _on_secret_changed(self, event: ops.SecretChangedEvent):
```
james-garner-canonical, line 340: See comment on other indentation change.

docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -410,32 +410,33 @@ Your charm now needs a way to load the value of the `slug` configuration option,
 In `src/charm.py`, add the following methods to the charm class:
 
 ```python
-    def configure_and_run(self) -> None:
```
james-garner-canonical, line 413: See comment on indentation change below.

docs/howto/debug-your-charm.md
```diff
@@ -272,7 +272,9 @@ In your charm code, call [](ops.Framework.breakpoint) to define breakpoints that
 ```python
 class MyCharm(ops.CharmBase):
     def _on_config_changed(self, event: ops.ConfigChangedEvent):
-        self.framework.breakpoint('config-start')  # 'config-start' is an arbitrary string you use with `--at`
+        self.framework.breakpoint(
+            'config-start'
+        )  # 'config-start' is an arbitrary string you use with `--at`
```
james-garner-canonical:
```suggestion
        # 'config-start' is an arbitrary string you use with `--at`
        self.framework.breakpoint('config-start')  
```

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
```diff
@@ -257,11 +255,12 @@ Similarly all that string parsing we were doing to get a hold of the snap versio
...
-    def _on_start(self, _event):  # noqa
-        systemd.service_start("snap.microsample.microsample.service")
+def _on_start(self, _event):  # noqa
+    systemd.service_start("snap.microsample.microsample.service")
+
 
-    def _on_stop(self, _event):  # noqa
-        systemd.service_stop("snap.microsample.microsample.service")
+def _on_stop(self, _event):  # noqa
+    systemd.service_stop("snap.microsample.microsample.service")
```
james-garner-canonical: See indentation change comments.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
```diff
@@ -235,15 +234,14 @@ In the `_on_install` method we had translated one-to-one the calls to `snap info
 
 Then we can replace all that `Popen` piping with simpler calls into the lib's API; `_on_install `becomes:
 ```python
-    def _on_install(self, _event):
```
james-garner-canonical, line 238: Indentation change.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
```diff
@@ -195,12 +197,9 @@ We are going to add a helper method:
 
 That allows us to fetch the Relation wherever we need it and access its contents or mutate them in a natural way:
 ```python
-    def _on_website_relation_joined(self, _event):
```
james-garner-canonical, line 198: Indentation change.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
```diff
@@ -155,28 +158,27 @@ Let's begin with `install`.
...
+        self.unit.status = ops.MaintenanceStatus("installing microsample")
+        out = check_call("snap install microsample --edge")
+
+    self.unit.status = ops.ActiveStatus()
 ```
 
 For `on-start` and `on-stop`, which are simple instructions to `systemctl` to start/stop the `microsample` service, we can copy over the commands as they are:
 
 ```python
-    def _on_start(self, _event):  # noqa
```
james-garner-canonical, line 175: Indentation change.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
```diff
@@ -155,28 +158,27 @@ Let's begin with `install`.
 The `/hooks/install` script checks if a snap package is installed; if not, it installs it. We need to still reach out to a shell to grab the `snap` package info and install the package, but we can have the logic and the status management in Python, which is nice. We use `subprocessing.check_call` to reach out to the OS. And yes, there is a better way to do this, we'll get to that later.
 
 ```python
-    def _on_install(self, _event):
```
james-garner-canonical, line 158: Indentation change.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
```diff
@@ -100,18 +100,21 @@ The minimal-effort solution in this case could be to create a file `/src/charm.p
...
-    self.framework.observe(self.on.stop, lambda _: os.popen('../hooks/stop'))
-    # etc...
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.framework.observe(
+            self.on.config_changed, lambda _: os.popen('../hooks/config-changed')
+        )
+        self.framework.observe(self.on.install, lambda _: os.popen('../hooks/install'))
+        self.framework.observe(self.on.start, lambda _: os.popen('../hooks/start'))
+        self.framework.observe(self.on.stop, lambda _: os.popen('../hooks/stop'))
```
james-garner-canonical: Does this even work? I thought there was some restriction that the callback actually had to be a method on a charm object?

james-garner-canonical (commented):
Thanks for this. There are a couple of cases where we should override the formatter's decisions: most universally, some of our docs examples are intentionally indented so you can copy and paste them to a class body (which perhaps can just be framed with a `class <actual class name>:` in the example. Also some examples I've suggested manually reformatting to better preserve their intent.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
tonyandrewmeyer: Yes, because the formatter cannot tell that this is continuing a previous block. I feel it's ok, the less indentation is simpler to read, and people shouldn't be copying the exact text and pasting it somewhere.

If we aren't ok with this change, it probably blocks using the formatter, unless we have some ugly "if true" type thing or adjust the doc to have one large block of code.

Curious what @dwilding thinks here.

ops/framework.py
tonyandrewmeyer: ruff flagged it. I'm not sure you could really trigger this outside of tests since we have short-lived processes. If it's better as a fix probably it should be its own tiny PR since the rest of this is definitely not fix.

tox.ini
tonyandrewmeyer: Is the existing comment not clear enough? I could repeat it below if needed.

I'm not sure we can standardise. I believe we are firmly one the single-quote path in our own code, but not pushing everyone else to it. So the profiles are quote quotes for example, which means the tutorials are, etc.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
james-garner-canonical: My suggestion elsewhere in the PR is to wrap it with the class header to preserve the correct indentation.

```py
class WhateverTheCharmIsCalled:
    ...

    def _on_install(self, _event):
```

tox.ini
james-garner-canonical: > Is the existing comment not clear enough? I could repeat it below if needed.

I didn't understand what was going on with the second run without referring back to the `pyproject.toml` change and the PR description.

> I'm not sure we can standardise. I believe we are firmly one the single-quote path in our own code, but not pushing everyone else to it. So the profiles are quote quotes for example, which means the tutorials are, etc.

That split seemed reasonable to me, but now that it's introducing tooling funkiness it seems worth reconsidering (separately from this PR).

docs/tutorial/write-your-first-machine-charm.md
tonyandrewmeyer: What do you think of [this approach](https://github.com/canonical/operator/pull/2544) instead?

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
tonyandrewmeyer: I feel like repeating the class when it's just a few lines above would look worse. Not so bad down the page, I guess, but still not ideal.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
tonyandrewmeyer: Huh, good point. No it would not. This has been there [since the import from Discourse](https://github.com/canonical/operator/pull/1481).

tonyandrewmeyer: Oh, right after in an "important" block it says:

> In practice we cannot bind lambdas to `observe`, we need to write dedicated _methods_ for that.

But also:

> Relying on `popen` is _not_ how Ops is supposed to be used. However, this code will work,

Hilarious!

tonyandrewmeyer: I doubt anyone is using this, but it's simple enough to fix, so I threw it in as a drive-by.

docs/howto/migrate/migrate-from-a-hooks-based-charm.md
dwilding: I need to think about this more, but don't have time to review the PR today, sorry!

Another couple of options to throw into the mix:

- For places where we really care about exact formatting (in the tutorial maybe?) we could include code snippets from the full source file instead of having the code in the doc source.
- It looks possible to tell ruff to skip particular code blocks, with `<!-- fmt:off -->` and `<!-- fmt:on -->`. See https://docs.astral.sh/ruff/formatter/#markdown-code-formatting

tonyandrewmeyer: > I need to think about this more, but don't have time to review the PR today, sorry!

No urgency at all!
 
> * For places where we really care about exact formatting (in the tutorial maybe?) we could include code snippets from the full source file instead of having the code in the doc source.

Yes, I've suggested this for the tutorial. I don't think we care about these other places.

> * It looks possible to tell ruff to skip particular code blocks, with `<!-- fmt:off -->` and `<!-- fmt:on -->`. See https://docs.astral.sh/ruff/formatter/#markdown-code-formatting

Ah, thanks I missed that. So we could do this for the tutorial and not the rest. I do think if we are turning it off for the majority then we would be better just not linting at all.

james-garner-canonical: > I feel like repeating the class when it's just a few lines above would look worse. Not so bad down the page, I guess, but still not ideal.

Good point!

> > * For places where we really care about exact formatting (in the tutorial maybe?) we could include code snippets from the full source file instead of having the code in the doc source.
> 
> Yes, I've suggested this for the tutorial. I don't think we care about these other places.

Agreed that this only really matters for the tutorial. Dedenting everywhere else seems reasonable. I really like the proposed code snippets approach.

docs/tutorial/write-your-first-machine-charm.md
james-garner-canonical: I really like it. How would you prefer to handle landing that? Land this PR with the indentation change in the tutorial, and replace with code snippets in follow up? Land the code snippets PR first?

ops/framework.py
james-garner-canonical: Makes sense, your call on keeping it in the chore PR or separating it so it gets a separate changelog entry. I'm fine with either.

tox.ini
dwilding: We already exclude the example charms from repo ruff. Presumably we could exclude `docs/tutorial` too. Then let ruff change the remaining docs to single quotes. Maybe odd for the tutorial to use a different style, but overall things would be more consistent that what we have now. And the tooling setup would be nice enough.

pyproject.toml
dwilding: I'd update this PR to exclude `docs/tutorial`. We can handle that separately by literalincluding code from the example charms, which already have their own linting rules. For the (small amount of) code in the tutorial that would remain in the doc source, I think it's fine to have that unchecked by ruff.

docs/explanation/state-transition-testing.md
```diff
@@ -263,11 +263,7 @@ class MyCharmType(ops.CharmBase):
 
 
 td = tempfile.TemporaryDirectory()
-ctx = testing.Context(
-    charm_type=MyCharmType,
-    meta={'name': 'my-charm-name'},
-    charm_root=td.name
-)
+ctx = testing.Context(charm_type=MyCharmType, meta={'name': 'my-charm-name'}, charm_root=td.name)
```
dwilding: I'd prefer that we force this to stack:

```suggestion
ctx = testing.Context(
    charm_type=MyCharmType,
    meta={'name': 'my-charm-name'},
    charm_root=td.name,
)
```

Because of the way the docs render. Lines longer than ~80 chars require horizontal scrolling, which I find quite annoying.

General comment - Do you think we want to apply a shorter line length to `docs/`? Would that require two separate runs (which we're trying to avoid)? If it's complex tooling-wise, I think the best solution is to keep this in mind when writing and reviewing code snippets for the docs.

tonyandrewmeyer: I've put a ruff.toml in `docs/`, with a line length of 80. Is that the number you'd pick?

tox.ini
tonyandrewmeyer: I've excluded docs/tutorial and removed the double run. It's a bit of a shame to lose the checks against the tutorial blocks, but the majority of them are duplicated from code that is listed elsewhere, so not too bad, and does simplify the setup.

This does increase the number of changes quite a bit (both the quote change and the line length one). Do you think it's too noisy for later blame usage? I could do a separate PR with *only* those changes and add it to the blame ignore list after merging, and then this one would be much smaller.

docs/tutorial/write-your-first-machine-charm.md
tonyandrewmeyer: David wanted to exclude the tutorials, so I guess I can just close that PR, since it doesn't need to solve a problem here any more.

james-garner-canonical: I guess it's not needed for this PR since David suggested excluding the tutorial:

> I'd update this PR to exclude docs/tutorial.

But worth landing anyway to achieve the goal of this PR since David's rationale for excluding the tutorial was:

> We can handle that separately by literalincluding code from the example charms, which already have their own linting rules. For the (small amount of) code in the tutorial that would remain in the doc source, I think it's fine to have that unchecked by ruff.

dwilding: Sorry, I wasn't totally clear: I support https://github.com/canonical/operator/pull/2544. Then given that we'll be pulling in the majority of the tutorial code, I don't think we'll lose much (in the end) by excluding `docs/tutorials` in this PR.

tox.ini
dwilding: If James has a strong opinion I'm happy to follow that. But I don't mind reviewing the quote change + line length change in this PR.

To answer a different thread - yes, 80 looks good to me for the docs line length. Thanks!

I was assuming that we'd go ahead with using literalinclude in the tutorial, but we hadn't necessary decided that, had we. (My bad.) Anyway, I think excluding the tutorial for now is good. We can always unexclude it later if we end up taking a different direction from literalinclude.

james-garner-canonical (approved):
Single quotes in the docs, how exciting! Should we document that decision somewhere, or just let CI speak for itself?

tonyandrewmeyer:
> Single quotes in the docs, how exciting! Should we document that decision somewhere, or just let CI speak for itself?

I feel like the tooling config is the documented decision.

james-garner-canonical:
> > Single quotes in the docs, how exciting! Should we document that decision somewhere, or just let CI speak for itself?
> 
> I feel like the tooling config is the documented decision.

Works for me!

dwilding (approved):

PR #2541, opened 2026-06-04
test: treat warnings as errors when running unit tests

tonyandrewmeyer:
The key line here is the `-W error` in `tox.ini`. That raises all warnings (the most common are deprecation and resource) to errors, when running the unit tests. The bulk of the rest of the PR is cleaning up tests (and one docstring, because we test those too, via doctest) to properly clean up resources.

The changes not in the tests (or docs) need more careful examination:
* in `ops.storage` we wrap setup so that if opening the database fails, we properly clean up. Should never happen in practice (no other process should be touching the db) and even if it did, we would close on the crash, but better to tidily manage it, and this also fixes test cases.
* in `ops._main` when instantiating the charm if something goes wrong (like the charm `__init__` raising) cleanly close the framework. Definitely could happen in practice, but we would close everyone on the crash. Doing it explicitly is cleaner in general, and helps when running tests.
* in `ops.pebble` when handling garage coming back from Pebble, the `urllib.error.HTTPError` is also a response object and carries the open HTTP response body in e.fp (we use it already with `.read()` and so on). Under Python 3.14 there's a tempfile (https://github.com/python/cpython/pull/15002, which looks unrelated but look at the code) so we get a warning on exit for failing to clean up. This could happen in production with a broken Pebble, although again  no real consequence, but nice to clean up for the tests and correctness.

### DIFF
--- a/ops/_main.py
+++ b/ops/_main.py
@@ -315,8 +315,15 @@ class _Manager:
         self.dispatcher.run_any_legacy_hook()
 
         self.framework = self._make_framework(self.dispatcher)
-        with self.framework._event_context('__init__'):
-            self.charm = self._charm_class(self.framework)
+        try:
+            with self.framework._event_context('__init__'):
+                self.charm = self._charm_class(self.framework)
+        except BaseException:
+            # If charm construction fails, close the framework so that the
+            # underlying storage (SQLite) is released; main() only reaches
+            # the run()/destroy() path when __init__ completes successfully.
+            self.framework.close()
+            raise
 
     def _load_charm_meta(self):
         return _charm.CharmMeta.from_charm_root(self._charm_root)
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -2237,15 +2237,16 @@ class Client:
         try:
             response = self.opener.open(request, timeout=self.timeout)
         except urllib.error.HTTPError as e:
-            code = e.code
-            status = e.reason
-            try:
-                body: dict[str, Any] = json.loads(e.read())
-                message: str = body['result']['message']
-            except (OSError, ValueError, KeyError) as e2:
-                # Will only happen on read error or if Pebble sends invalid JSON.
-                body: dict[str, Any] = {}
-                message = f'{type(e2).__name__} - {e2}'
+            with e:  # close the underlying tempfile so it doesn't leak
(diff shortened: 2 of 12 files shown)


ops/pebble.py
```diff
@@ -2222,15 +2222,16 @@ def _request_raw(
...
-            code = e.code
-            status = e.reason
-            try:
-                body: dict[str, Any] = json.loads(e.read())
-                message: str = body['result']['message']
-            except (OSError, ValueError, KeyError) as e2:
-                # Will only happen on read error or if Pebble sends invalid JSON.
-                body: dict[str, Any] = {}
-                message = f'{type(e2).__name__} - {e2}'
+            with e:  # close the underlying tempfile so it doesn't leak
```
tonyandrewmeyer, line 2240: TIL that an HTTPError can also function as a non-exceptional file-like return value. I'm uncertain this is required, though.

tonyandrewmeyer: It is, I checked.

ops/pebble.py
```diff
@@ -2237,15 +2237,16 @@ def _request_raw(
...
```
james-garner-canonical, line 2240: This certainly feels unusual, it's a TIL for me as well for sure.

ops/_main.py
```diff
@@ -315,8 +315,15 @@ def __init__(
...
         self.framework = self._make_framework(self.dispatcher)
-        with self.framework._event_context('__init__'):
-            self.charm = self._charm_class(self.framework)
+        try:
+            with self.framework._event_context('__init__'):
+                self.charm = self._charm_class(self.framework)
+        except BaseException:
+            # If charm construction fails, close the framework so that the
+            # underlying storage (SQLite) is released; main() only reaches
+            # the run()/destroy() path when __init__ completes successfully.
```
james-garner-canonical, line 324: This makes sense, and I guess you considered if there would be a clean way to unify the init path cleanup with the even handling path cleanup -- trusting your judgement here.

james-garner-canonical (approved):

ops/pebble.py
tromai: Same for me 😄

tromai (approved):
Thank you!

PR #2542, opened 2026-06-04
fix: restore sys.breakpointhook on _Manager teardown

tonyandrewmeyer:
`Framework.set_breakpointhook()` has always returned the old hook so the caller can restore it, but `_Manager._make_framework()` discarded the return value. In a real Juju hook execution this doesn't matter, since the process is short-lived, but tests that drive `ops.main()` directly leave a polluted `sys.breakpointhook` for every subsequent test in the same process. The leaked hook is a bound `framework.breakpoint` method whose `_juju_debug_at` came from whichever test set `JUJU_DEBUG_AT`, so a later test calling `breakpoint()` drops into `pdb` regardless of its own `PYTHONBREAKPOINT`.

The PR fixes this by saving the return value in `_make_framework()` and restoring it in `destroy()`. The attribute is a class-level `None` so that `destroy()` is safe even if `_make_framework()` never ran.

Also fixes a test that relied on the bug.


test/test_main.py
```diff
@@ -89,27 +89,40 @@ class EventSpec:
...
 
     def test_no_debug_breakpoint(self):
+        pdb_set_trace_calls: list[int] = []
+
         class MyCharm(ops.CharmBase):
-            pass
+            def __init__(self, framework: ops.Framework):
+                super().__init__(framework)
+                # See note in test_breakpoint about why this runs inside the
+                # charm.
```
james-garner-canonical, line 118: Deduplicating like this makes sense -- we can't easily lift the classes to the module level as they need a reference to a local variable to store their calls in, and adopting a different approach (e.g. storing calls in an instance variable) would require more of a refactor.

ops/_main.py
```diff
@@ -407,7 +408,12 @@ def _make_framework(self, dispatcher: _Dispatcher):
             event_name=dispatcher.event_name,
             juju_debug_at=self._juju_context.debug_at,
         )
-        framework.set_breakpointhook()
+        # set_breakpointhook() returns the old sys.breakpointhook; capture it
+        # so destroy() can put the global state back rather than leaving the
+        # framework's bound method dangling. In production this is moot (the
+        # process is exiting anyway), but tests and other long-running
+        # ops.main() callers inherit the polluted hook otherwise.
+        self._saved_breakpointhook = framework.set_breakpointhook()
```
james-garner-canonical, line 416: My only hesitation reading this is that it's not re-entry safe, so we're relying on `_make_framework` being called exactly once (from any instance) before `destroy` is called. I assume that matches the current behaviour at runtime and how scenario behaves, but maybe there are some tortuous cases where it would matter -- or maybe they're not accessible via the public API? Either way this seems like a strict improvement over not restoring `sys.breakpointhook` at all.

james-garner-canonical (approved):

ops/_main.py
tonyandrewmeyer: Yes, it's only called in `__init__`. It's a separate method because that lets Scenario easily override the behaviour but it is really meant to be done once for the `_Manager` object. None of this is accessible publicly, so I think we are safe enough.

test/test_main.py
```diff
@@ -89,27 +89,40 @@ class EventSpec:
 class TestCharmInit:
     @patch('sys.stderr', new_callable=io.StringIO)
     def test_breakpoint(self, fake_stderr: io.StringIO):
+        pdb_set_trace_calls: list[int] = []
```
tromai, line 92: I assume we use a list here purely to overcome the scope limitation and modify it within `MyCharm`'s constructor.

tromai (approved):
The fix is clean to me. 

I agree that it is not a major problem because `_make_framework` is called once per `_Manager` instance.

I also have one question to further understand the fix.

test/test_main.py
tonyandrewmeyer: Yes, a simple way to adjust the scope and also fail if somehow it's duplicsted.

Issue #2441, opened 2026-04-21
In the tutorials, advise how to handle "Too many levels of symbolic links" when packing


dwilding:
We've observed some people hit an error when packing the tutorial charms. See https://github.com/canonical/charmcraft/issues/2661.

If it won't be possible to fix the issue in the near future, we should advise people how to avoid the error. For example, should they delete `.venv` and `.tox` before trying to pack again? Would it be helpful to ask for people to report the issue, to help the Charmcraft team identify the specific cause & fix?

gcomneno:
Opened a docs update here: https://github.com/canonical/operator/pull/2452

This adds a warning to the tutorial pack step with a workaround for `Too many levels of symbolic links` failures during `charmcraft pack`, and suggests reporting persistent failures upstream while the root cause is still being investigated.

tonyandrewmeyer:
Fixed upstream.

PR #2452, opened 2026-04-25
docs: add workaround for symlink loop during packing

gcomneno:
## Summary

Add a warning to the Kubernetes tutorial pack step documenting a workaround for `Too many levels of symbolic links` failures during `charmcraft pack`.

The note suggests removing local virtual environment directories such as `.venv` and `.tox`, then retrying. It also suggests reporting persistent failures upstream, since the root cause is still under investigation.

## Testing

- `make -C docs html`
- `make -C docs vale TARGET='docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md'`
- `make -C docs spelling TARGET='docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md'`

PR #2543, opened 2026-06-04
docs: add warning for symbolic links error during charmcraft pack

chamoflag:
When running `charmcraft pack`, some users encounter a `Too many levels 
of symbolic links` error caused by the presence of `.venv` or `.tox` 
directories in the project directory.

This adds a note to the tutorial with the workaround — removing those 
directories before packing — and a link to the upstream Charmcraft issue 
for users who continue to see the error.

Relates to #2441.


tonyandrewmeyer:
Thanks, but #2452 (linked from that issue) is already working on this.

PR #2544, opened 2026-06-04
docs: Pull tutorial code snippets from examples/

tonyandrewmeyer:
Replace hardcoded code fences in the tutorial pages with Sphinx `literalinclude` directives that pull from the matching example charms in `examples/`. This keeps the tutorial automatically in sync with the canonical source, and removes ~750 lines of duplicated code.

Strategy per snippet type:
- Full method or class: `:pyobject: Class.method`
- Single line or small fragment:` :start-at:`/`:end-at:` patterns on the line content (no marker comments needed in sources)
- Multi-line block: `:start-at:`/`:end-before:` patterns anchored on the next section
- Full file (skipping copyright header): `:start-at:` with the first real line

Intermediate states that don't exist in `examples/` (for example, the empty starting charm class, lines shown only to be removed, contrastive "don't do this" snippets, and a troubleshooting timeout hint) are left inline.

[Preview](https://canonical-ubuntu-documentation-library--2544.com.readthedocs.build/ops/2544/)


james-garner-canonical:
This is super cool! I didn't know Sphinx could do this.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -54,33 +54,20 @@ These files currently contain placeholder code and configuration.
...
-```yaml
-title: Web Server Demo
-summary: A demo charm that operates a small Python FastAPI server.
-description: |
-  This charm demonstrates how to write a Kubernetes charm with Ops.
+```{literalinclude} ../../../examples/k8s-1-minimal/charmcraft.yaml
+:language: yaml
+:start-at: 'title: Web Server Demo'
+:end-at: how to write a Kubernetes charm with Ops.
 ```
```
james-garner-canonical, line 57: We could use `start-after` and `end-after` with comment anchors in cases where we worry that this won't be robust. Thinking about this case specifically, I guess Sphinx would yell at us if we changed the end line of the description so there was no match, so this seems great as-is.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -153,8 +126,11 @@ As you can see, a charm is a pure Python class that inherits from the [`CharmBas
 
 In the `__init__` function of your charm class, we'll tell Ops which method of your charm class to run for each event. Let's start with when the Juju controller tells us that the workload container's Pebble is up and running.
 
-```python
-framework.observe(self.on["demo-server"].pebble_ready, self._on_demo_server_pebble_ready)
+```{literalinclude} ../../../examples/k8s-1-minimal/src/charm.py
+:language: python
+:start-at: framework.observe(self.on["demo-server"]
+:end-at: framework.observe(self.on["demo-server"]
```
james-garner-canonical, line 132: This line is too long and requires scrolling in the docs. I wonder if we should adopt a lower line length for the example charms now that we're using them this way.

james-garner-canonical (approved):
I haven't checked every example, but I really like this direction.

I assume that if a matching anchor isn't found, the docs build complains loudly, so this should be far less prone to drift that before.

I have one non-blocking concern that may be worth addressing at this point: the maximum line length we use in the charm code can lead to long lines that require horizontal scrolling in the docs -- previously we would have reformatted them manually in the docs to avoid this. Should we use a shorter maximum line length for the examples now that we're using them this way? Does this have implication for the profiles -- I think it would be fine if this was one of our divergences from the profiles, but maybe it would make the tutorial harder to follow, in which case maybe we just live with the horizontal scrolling ...

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -484,26 +466,18 @@ First, in `tests/integration/test_charm.py`, import `json`, `time`, `urllib.erro
...
 Next, still in `tests/integration/test_charm.py`, define the new fixture:
 
-```python
-@pytest.fixture(scope="module")
-def cos(juju_factory: pytest_jubilant.JujuFactory):
-    yield juju_factory.get_juju(suffix="cos")
+```{literalinclude} ../../../examples/k8s-5-observe/tests/integration/test_charm.py
+:language: python
+:start-at: '@pytest.fixture(scope="module")'
+:end-at: yield juju_factory.get_juju
```
dwilding, line 480: This will work even though the function has a decorator. I verified by building locally.

```suggestion
:pyobject: cos
```

docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -506,30 +372,23 @@ If Juju wants to remove a unit of the application, your charm (on that unit) rec
...
-        self.wait_for_not_running()
-
-    def _on_remove(self, event: ops.RemoveEvent) -> None:
-        """Handle remove event."""
-        tinyproxy.uninstall()
+```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
+:language: python
+:start-at: def _on_stop
+:end-at: tinyproxy.uninstall()
+```
```
dwilding, line 387: I think we should switch to `pyobject` for both methods, matching what we do for "Deploy COS Lite and integrate with Loki" in the machine charm tutorial.

````suggestion
```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
:language: python
:pyobject: TinyproxyCharm._on_stop
```

```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
:language: python
:pyobject: TinyproxyCharm._on_remove
```
````

File: docs/tutorial/write-your-first-machine-charm.md [hunk 5]
```diff
@@ -404,65 +297,48 @@ Your charm now needs a way to load the value of the `slug` configuration option,
...
+:language: python
+:start-at: import time
+:end-at: import time
+```
 
-PORT = 8000
+```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
+:language: python
+:start-at: PORT = 8000
+:end-at: PORT = 8000
```
dwilding, line 317: Having two code blocks looks strange in the rendered doc. I think this would be better as:

````suggestion
```python
import time

PORT = 8000
````

And we accept that we lose validation that the example charm also contains these lines. I think in this case (only) it's an acceptable trade-off.

**Review by dwilding — APPROVED** [review]
Thank you for implementing this.

I checked that the Sphinx build errors if a `literalinclude` doesn't match the charm source. That means that even the almost trivial includes (e.g., the one-liners) have value, as we're kind of asserting the charm source hasn't changed since the tutorial was last updated.

There is one exception where I feel `literalinclude` is making the doc look worse, and where I think it's worth losing the assertion. Noted in a suggestion.

Re line lengths: I would prefer if the docs had enough space to comfortably show all lines. Next cycle we need to switch to the [Ulwazi](https://github.com/canonical/ulwazi) theme, which should help _a bit_, but not that much.

I'm not keen on setting a lower line width in the example charms - deviating from the Charmcraft profiles. I'm not keen on modifying the profiles either. For now, I'm happy to accept that some lines will scroll.

Actually, related to this, I noticed that some `literalinclude`s set `:dedent:` and some don't. I think this matches preexisting differences between the machine and K8s tutorials. My observation has been that people like being able to copy/paste from the tutorial snippets. Does dedent make that harder?

I think we should make a decision on dedent (for the tutorials) as part of this PR. As I write this, I lean towards removing all dedents from the tutorial snippets. This would make the line-length problem worse, though.

File: docs/tutorial/write-your-first-machine-charm.md (hunk 5 above)
Comment (james-garner-canonical) [reply docs/tutorial/write-your-first-machine-charm.md:317]: What if we put text between the blocks? "Add the following import to `src/charm.py`:", "Define the `PORT` constant like this:".

Unrelated, I was briefly confused that these weren't referred to anywhere below, until I realised that they were the requirements for the code block above them to run. I assume this was an intentional choice, but if reordering them would make the separate blocks flow better, that could be an option too.

**Comment by james-garner-canonical** [comment]
Re: to `:dedent:` or not to `:dedent:`, I suspect there's not a perfect solution for fragments, because of how editors will automatically indent the first line anyway. For example, say the reader has this in their file already:

```py
def foo():
    x = []
    x.append(1)
    # We want to insert some text below this line.
    # But above this line.
    return x
```

And we want to insert the lines:
```py
    x.append(2)
    x.append(3)
```

I'm not sure what the UX would be that would reliably let you copy exactly the lines snippet, and paste it into the existing file, because you'll first need to insert a linebreak after `below this line.`, and your editor will almost definitely indent the new line. So if you paste the lines above, you'll get this:

```py
def foo():
    x = []
    x.append(1)
    # We want to insert some text below this line.
        x.append(2)
    x.append(3)
    # But above this line.
    return x
```

On the other hand if the lines were dedented, you'd get this:
```py
def foo():
    x = []
    x.append(1)
    # We want to insert some text below this line.
    x.append(2)
x.append(3)
    # But above this line.
    return x
```

I guess indented snippets could start with a linebreak, but would users actually realise to paste at the end of the preceding line?

So maybe we should `:dedent:` everything? Or maybe there are some cases where it's useful and some where it's not?

I don't mind if the rule we go with is "never `:dedent:`", this just got me thinking and I thought I'd share in case you agree or in case your experience is different (and so I'm missing something).

docs/tutorial/write-your-first-machine-charm.md
```diff
@@ -404,65 +297,48 @@ Your charm now needs a way to load the value of the `slug` configuration option,
...
 ```
```
dwilding, line 318: How about replacing all of this with:

````
First, at the beginning of `src/charm.py`, define the port we'll write to the configuration file:

```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
:language: python
:start-at: PORT = 8000
:end-at: PORT = 8000
```

Then add the following import to `src/charm.py`:

```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
:language: python
:start-at: import time
:end-at: import time
```

We're now ready to the implement the logic to load `slug` and configure & start tinyproxy. Add the following methods to the charm class:

```{literalinclude} ../../examples/machine-tinyproxy/src/charm.py
:language: python
:start-at: def configure_and_run
:end-at: "# The Juju logs will show the error message, to help you debug the error."
```
````

docs/tutorial/write-your-first-machine-charm.md
dwilding: > if reordering them would make the separate blocks flow better, that could be an option too

I think this would be the best approach. I've just added a suggestion for it: https://github.com/canonical/operator/pull/2544/changes#r3976828595

PR #2545, opened 2026-06-05
docs: move docs to canonical.com/juju/docs/ops

dwilding:
This PR updates our doc config so that all URLs use `canonical.com/juju/docs/ops` instead of the backend URL from Read the Docs. Our docs are still hosted on Read the Docs, but we'll be proxying them under canonical.com/juju/docs.

<mark>**[Preview docs](https://canonical.com/juju/docs/ops/latest/)** built from this branch. After this PR merges, I'll switch those docs to build from `main` and redirect the existing docs.</mark>

**Main changes**

- Updated Ops doc URLs in `docs/conf.py`.
- Updated hardcoded URLs for Ops, excluding `CHANGES.md`.
- Added a script `docs/_static/overwrite_links.js`. This script makes sure that the version switcher from Read the Docs has correct URLs. It's a standard part of the new docs setup at Canonical.

**Extra changes**

- Updated the intersphinx URL and hardcoded URLs for Jubilant. To account for [jubilant#337](https://github.com/canonical/jubilant/pull/337).
- Updated the intersphinx URL and hardcoded URLs for Charmlibs. To account for [charmlibs#515](https://github.com/canonical/charmlibs/pull/515)
- Updated the intersphinx URL and hardcoded URLs for Pebble. To account for [pebble#882](https://github.com/canonical/pebble/pull/882).

**Future work (not in this PR)**

- Update URLs for other docs - Juju, Charmcraft, ...
- Update URLs in comments in the Charmcraft profiles. This PR updates the affected comments in our example charms, bringing them ahead of the latest Charmcraft profiles. I think that's acceptable to reduce busywork later.


docs/_static/overwrite_links.js
```diff
@@ -0,0 +1,28 @@
+// Our docs are proxied through to canonical.com/juju/docs/ops.
```
tromai, line 1: Should we have a copyright notice for this file?

ops/_private/harness.py
```diff
@@ -318,7 +318,7 @@ def __init__(
 
         warnings.warn(
             'Harness is deprecated. For the recommended approach, see: '
-            'https://documentation.ubuntu.com/ops/latest/howto/write-unit-tests-for-a-charm/',
+            'https://canonical.com/juju/docs/opslatest/howto/write-unit-tests-for-a-charm/',
```
tromai:
```suggestion
            'https://canonical.com/juju/docs/ops/latest/howto/write-unit-tests-for-a-charm/',
```

tromai (approved):
Thank you for this change.

~I saw in the PR description the link to the Preview docs is `https://canonical.com/juju/docs/ops/latest/`. Does this mean we have publish the docs built from this branch to the new home?~ Ah nevermind, I saw you mentioned switching back to `main` after this PR is merged.

docs/_static/overwrite_links.js
tromai: As this script is a new Canonical docs standard, I assume that it has been battle-tested upstream. But please let me know if I need to have a closer look 😄. 

I had a scan through and it looks good to me.

ops/_private/harness.py
dwilding: Yikes, thanks! Fixed a few instances.

docs/_static/overwrite_links.js
dwilding: Answering several comments in one :slightly_smiling_face: 

The script has been tested, but we can always update it here (and update upstream) if anything needs changing significantly. I think we're fine for now.

Copyright - that's a good question. We don't seem to add copyright notices to our doc source files and assets. Perhaps because the docs render with a copyright notice at the bottom of each page? I'm not against adding one though, if you think it's important.

On https://github.com/canonical/jubilant/pull/337, Tony says:

> It's a bit awful we have to do this, although I understand why. I guess if you don't have JS enabled, you would just end up at a page that redirects you anyway?

Without `overwrite_links.js`, the version switcher links to the RTD backend URLs, which are private, so you end up at a page that requires login. However, if you don't have JS enabled, the version switcher doesn't load in the first place... So I suppose we aren't making things materially worse, at least.

dwilding:
Doc builds are currently failing because https://ubuntu.com/docs/pebble returns a 502. That's an infra issue and is being tracked internally. I'll retry the builds next week.

ops/_private/harness.py
benhoyt: I'm curious -- was this an old fashioned fat-finger mistake, or an AI mistake?

docs/_static/overwrite_links.js
tromai: Thanks for the reply. Regarding the copyright - I think it's alright to **not** add it here, and keep it consistent with other source files and assets.

I somehow assume we always have copyright notices in docs scripts :laughing: 

>  I think we're fine for now.

Great! Thanks for double checking.

ops/_private/harness.py
dwilding: That was entirely me. I accidentally added a double https:// to some URLs, then messed up the fix...

james-garner-canonical (approved):
Looks good!

PR #2550, opened 2026-06-07
fix: chdir to charm root during event in testing.Context

### LINKED ISSUE #2045: os.getcwd() is not the charm root in ops[testing]
When running in Juju, the current directory is the charm root at the start of a hook. When running in Scenario, the current directory is typically the root directory of the repository.

[This hockeypuck-k8s branch](https://github.com/canonical/hockeypuck-k8s-operator/pull/150) has examples of where this fails in a paas-charm.

* test_scenario fails because we [don't support charmcraft extensions in autoload](https://github.com/canonical/operator/issues/1699)
* ~test_scenario_one needs to have the context created as `Context(HockeypuckK8SCharm, meta=METADATA, actions=METADATA['actions'])` because [you can't pass all of charmcraft.yaml in the 'meta' argument](https://github.com/canonical/operator/issues/1424)~
* test_scenario_three needs to have all of the actions provided in the metadata, because otherwise `charm.on` won't have them all.
* test_scenario_two (and test_scenario_one and test_scenario_three once the above issues are fixed) fail because they cannot load the config.

paas_charm [loads the config.yaml file directly](https://github.com/canonical/paas-charm/blob/d101fb07594b9ae4d8740f747c1b9611abb8e074/src/paas_charm/utils.py#L100) rather than using `.config` (or the newer `.load_config`, which would provide the "load a secret as a secret" functionality that they're also implementing). This fails under Scenario because [it looks for config.yaml in the current directory](https://github.com/canonical/paas-charm/blob/d101fb07594b9ae4d8740f747c1b9611abb8e074/src/paas_charm/utils.py#L137), but Scenario carefully creates a temporary directory for executing the tests. For most charms, this would still work (if it fell back to `charmcraft.yaml`), because the YAML is just copied, but in this case the YAML is being explicitly provided because of the lack of extension support.

(I think there is some other issue with the tests, because some of them end up with `UnknownStatus` rather than the expected `BlockedStatus`, but I don't think that's an issue with Scenario).

A simple `os.chdir()` (and restoring the value after the yield) in `_runtime.py` fixed this for me locally, and should probably be a simple and safe fix, I think. It would be nicer if the charms didn't directly access the metadata files but it probably is reasonable that they do. The paas_charm code would be better if it used `Framework`'s `charm_dir` rather than relying on `getcwd()`, but I think it is also reasonable to expect the initial working directory to be the charm dir.

tonyandrewmeyer:
Under Juju, the current working directory at the start of a hook is the charm root; under Scenario it is wherever pytest ran from. Charms (and libraries) that read files relative to the cwd therefore break only under Scenario.

This PR mirrors the existing `JUJU_CHARM_DIR` handling: save the cwd, chdir to `temporary_charm_root` after updating the env, and restore in finally before the virtual charm root tempdir is cleaned up. However, to avoid breaking existing charms, it only does this if an environment variable is set, and otherwise warns for incorrect behaviour. Ideally, in the next major version we would flip the default for the env variable.

Fixes #2045.


tonyandrewmeyer:
The obvious concern here is backwards compatibility, and validly so.

I ran `hyrum unit --framework scenario --workers 2` over the curated 233-charm cache. Three runs over the same 50-charm filter set: **A** `--no-patch` (88% pass), **B** swap to `canonical/operator:main` (34/50 pass: the 6-charm degradation vs A is unrelated dependency-swap noise, same in both controls), **C** swap to the patched branch (24/50 pass).

**B → C diff: 10 charms flipped passed → failed, every one cwd-attributable.** Of those, 7 raise `FileNotFoundError` on a path that's relative-to-cwd (charm code reads `templates/X.j2` / `src/Y.json` / `src/prometheus_alert_rules/...` directly instead of from `self.framework.charm_dir`); 2 surface as Jinja `TemplateNotFound` (`vault-k8s-operator/{k8s,machine}`, the same root cause via the template loader's search path); and 2 (`prometheus-k8s-operator`, `grafana-agent-operator`) shell out to `rm ./Certum_Trusted_Root_CA.pem` for test cleanup. Pre-patch that hits the test's own working copy; post-patch the relative path resolves into the system CA store and bounces off permissions, which is the patch incidentally surfacing a *test-side* safety bug (test would have deleted a system file if it ran with elevated perms).

Some options:
* We accept that, although wrong, this is the behaviour of Scenario.
* As above, but add a DeprecationWarning and change it in a future major version bump.
* Add a flag (environment variable maybe, like the others we have) to switch the behaviour, plan to swap the defaults in a future major version bump.
* Accept that this is a fix and break tests. We can open PRs against the known 10 charms in advance.

*Correction on the Certum_Trusted_Root_CA.pem characterisation above*: re-checking the source, neither charm has any rm ./Certum_Trusted_Root_CA.pem in charm or test code. Both shell out to subprocess.run(["update-ca-certificates", "--fresh"]) on the host (prometheus-k8s-operator/src/charm.py:624, grafana-agent-operator/src/charm.py:446). That command rewrites /etc/ssl/certs/ and prints Adding debian:Certum_Trusted_Root_CA.pem to stdout, which is where the path came from in the failure log. Since update-ca-certificates --fresh uses absolute paths, these two are likely not cwd-attributable and the real count is closer to 8.

testing/src/scenario/_runtime.py
```diff
@@ -321,6 +321,13 @@ def exec(
             previous_env = os.environ.copy()
             os.environ.update(env)
 
+            # In Juju, the current working directory is the charm root when a
+            # hook starts. Charms (and libraries) rely on this, e.g. to read
+            # files relative to it. Match that behaviour, restoring the
+            # original working directory once the event has been handled.
+            previous_cwd = os.getcwd()
+            os.chdir(temporary_charm_root)
```
james-garner-canonical: I'm a little confused about the current behaviour -- if we didn't previously expose the `temporary_charm_root` to the charm under test by making it the current working directory, how were its contents actually exposed to and (potentially) used by the charm during testing?

james-garner-canonical (commented):
> The obvious concern here is backwards compatibility, and validly so.
> 
> I ran `hyrum unit --framework scenario --workers 2` over the curated 233-charm cache. Three runs over the same 50-charm filter set: **A** `--no-patch` (88% pass), **B** swap to `canonical/operator:main` (34/50 pass: the 6-charm degradation vs A is unrelated dependency-swap noise, same in both controls), **C** swap to the patched branch (24/50 pass).

Why the 50 vs 233 charm count discrepancy here?

> **B → C diff: 10 charms flipped passed → failed, every one cwd-attributable.** Of those, 7 raise `FileNotFoundError` on a path that's relative-to-cwd (charm code reads `templates/X.j2` / `src/Y.json` / `src/prometheus_alert_rules/...` directly instead of from `self.framework.charm_dir`); 2 surface as Jinja `TemplateNotFound` (`vault-k8s-operator/{k8s,machine}`, the same root cause via the template loader's search path); and 2 (`prometheus-k8s-operator`, `grafana-agent-operator`) shell out to `rm ./Certum_Trusted_Root_CA.pem` for test cleanup. Pre-patch that hits the test's own working copy; post-patch the relative path resolves into the system CA store and bounces off permissions, which is the patch incidentally surfacing a _test-side_ safety bug (test would have deleted a system file if it ran with elevated perms).
> 
> Some options:
> 
> * We accept that, although wrong, this is the behaviour of Scenario.
> * As above, but add a DeprecationWarning and change it in a future major version bump.
> * Add a flag (environment variable maybe, like the others we have) to switch the behaviour, plan to swap the defaults in a future major version bump.
> * Accept that this is a fix and break tests. We can open PRs against the known 10 charms in advance.

Modulo improving my understanding of our current behaviour with the temporary charm root, I lean towards the deprecation warning + change -- our concurrent push to have charms treat warnings as errors should hopefully move people in the direction we want. I'm open to the environment variable gating with default swapping in the future release as well.

That said, contingent on us actually finding all the users this would break, and again modulo my understanding of our current behaviour, I'm not opposed to the fix approach.

tonyandrewmeyer:
> Why the 50 vs 233 charm count discrepancy here?

The scenario filter. There are lots of charms that are not ops, and lots that don't use Scenario so wouldn't be impacted.

tromai:
The fix looks good to me.

> Pre-patch that hits the test's own working copy; post-patch the relative path resolves into the system CA store and bounces off permissions

😨. I am curious how it happened. I couldn't find an example searching through their source code.

Overall, I am also okay with these two options:
  
> 1. As above, but add a DeprecationWarning and change it in a future major version bump.
> 2. Add a flag (environment variable maybe, like the others we have) to switch the behaviour, plan to swap the defaults in a future major version bump.

I lean toward the first one more. To me, the DeprecationWarning can communicate to authors of 10 charms. We can add the flag if they request keeping the behavior in https://github.com/canonical/operator/issues/2045.

testing/src/scenario/_runtime.py
tonyandrewmeyer: The virtual charm root is exposed via the JUJU_CHARM_DIR env var, which ops reads into framework.charm_dir. Charms that use `self.framework.charm_dir / "templates/foo.j2` were fine.

The problem is code that uses relative paths, assuming the current directory, like `open("templates/foo.j2")`. Those rely on the real location you get in Juju being the charm root. Under Scenario (main), that's is wherever pytest ran, so those paths resolve into the test directory.

When no charm_root is passed, the tempdir contents are just the metadata/config/actions YAMLs that ops loads through charm_dir anyway (although there's a whole conversation there about maybe it should be whatever would get packed). When a custom charm_root is passed, its files were reachable through charm_dir but not via relative paths.

tonyandrewmeyer:
> > Pre-patch that hits the test's own working copy; post-patch the relative path resolves into the system CA store and bounces off permissions
> 
> 😨. I am curious how it happened. I couldn't find an example searching through their source code.

You're right to be suspicious. I went back to the source and Certum_Trusted_Root_CA.pem doesn't appear anywhere in those two charms' code. What actually happens is that prometheus-k8s-operator (src/charm.py:624) and grafana-agent-operator (src/charm.py:446) call `subprocess.run(["update-ca-certificates", "--fresh"])` on the host. That command rewrites `/etc/ssl/certs/`, printing lines like `Adding debian:Certum_Trusted_Root_CA.pem` to stdout, which is where the path came from in the failure log.

I conflated stdout with file access in my original write-up; apologies for the noise.

Since `update-ca-certificates --fresh` uses absolute paths and not cwd-relative ones, those two probably aren't cwd-attributable at all, likely run-to-run noise on a command that's flaky under tests anyway. Real count is closer to 8 cwd-attributable charms, not 10.

tonyandrewmeyer:
I've changed this to go the deprecation warning and env var direction. Not simple to warn at the right time - I've used an audit hook (added in 3.8, so fine for our 3.10+), which is a little wild, and my first use of these for real, but doesn't seem outside the bounds of what they are meant for.

james-garner-canonical:
> **B → C diff: 10 charms flipped passed → failed, every one cwd-attributable.** Of those, 7 raise `FileNotFoundError` on a path that's relative-to-cwd (charm code reads `templates/X.j2` / `src/Y.json` / `src/prometheus_alert_rules/...` directly instead of from `self.framework.charm_dir`); 2 surface as Jinja `TemplateNotFound` (`vault-k8s-operator/{k8s,machine}`, the same root cause via the template loader's search path); and 2 (`prometheus-k8s-operator`, `grafana-agent-operator`) shell out to `rm ./Certum_Trusted_Root_CA.pem` for test cleanup. Pre-patch that hits the test's own working copy; post-patch the relative path resolves into the system CA store and bounces off permissions, which is the patch incidentally surfacing a _test-side_ safety bug (test would have deleted a system file if it ran with elevated perms).

I'm particularly concerned about `FileNotFoundError` on a path that's relative to the actual charm root (`src/Y.json`, `src/prometheus...`) etc. Arguably they should use `self.framework.charm_dir`, sure, but aren't these particular breakages the opposite of what we're trying to address with this PR? Here the charm is presumably running `pytest` from the charm root (matching Juju's behaviour of running the charm from there), and we're introducing an error for them with this change due to our selective population of the virtual charm root.

testing/src/scenario/_runtime.py
```diff
@@ -40,6 +43,74 @@
...
+    state = getattr(_CWD_AUDIT_STATE, 'current', None)
+    if state is None:
+        return
+    if event not in ('open', 'os.listdir', 'os.scandir'):
+        return
+    if state.get('reentrant'):
+        return
+    path = args[0] if args else None
+    if path is None or isinstance(path, int):
+        return
```
tromai, line 67: I'm curious in which case can `path` be an int.
I am thinking of the  argument `buffering` to `open`, is that correct?

testing/src/scenario/_runtime.py
```diff
@@ -40,6 +43,74 @@
...
+    Under Juju, the working directory at hook start is the charm root. Scenario
+    historically left it as wherever pytest ran, so relative reads silently
+    resolve against the test directory. This hook flags any relative path
+    access that would resolve differently between the two, so charms can be
+    updated before the default flips.
+    """
+    state = getattr(_CWD_AUDIT_STATE, 'current', None)
+    if state is None:
+        return
+    if event not in ('open', 'os.listdir', 'os.scandir'):
```
tromai, line 61: I'm curious about the criteria for selecting this set of events. I assume it came from the 8 charms which depends on Scenario's 'unexpected' dir behavior ? If that's true, it's all good for me, as I don't think we should include more events.

testing/src/scenario/_runtime.py
```diff
@@ -40,6 +43,74 @@
...
+    if isinstance(path_str, bytes):
+        try:
+            path_str = path_str.decode()
+        except UnicodeDecodeError:
+            return
+    if not path_str or os.path.isabs(path_str):
+        return
+    if path_str in state['warned']:
+        return
+    state['reentrant'] = True
```
tromai, line 81: Just to confirm, `reentrant` is used to prevent infinite recursion from this function because we do file operations included in the warning event `('open', 'os.listdir', 'os.scandir')`?

If that is true, may I suggest having a small comment describing the purpose here?

tromai (commented):
Thank you for adding the deprecation warning. Audit hook looks really powerful to me. 

I made a minor request, and some questions to help me understand the fix better.

tonyandrewmeyer:
I think we should perhaps solve #1673 before continuing with this, or at least decide what we will do there.

james-garner-canonical (commented):
Comment review so you can re-request review after moving out of draft.

benhoyt:
Per discussion, closing till we decide on direction in #1673 -- and could always reopen.

PR #2552, opened 2026-06-08
fix: use the correct value for JUJU_HOOK_NAME when objects contain dashes

### LINKED ISSUE #2511: Does Scenario produces incorrect JUJU_HOOK_NAME?

tonyandrewmeyer:
Scenario over eagerly normalises the Juju hook name. Juju keeps the entity name exactly as provided, but Scenario normalises dashes to underscores unconditionally. This PR fixes that to preserve the correct name as Juju would provide it. 

This replaces #2515, which was closed by the author. Essentially all of that PR had been redone by me, and this is pushing from my local copy of that work. I'll redo the change that James requested as well.

Fixes #2511


james-garner-canonical (approved):

testing/src/scenario/state.py
```diff
@@ -2140,10 +2140,12 @@ class _EventPath(str):
         owner_path: list[str]
         suffix: str
         prefix: str
+        original_prefix: str
         is_custom: bool
         type: _EventType
 
     def __new__(cls, string: str):
+        original_string = string
```
hpidcock, line 2148: Where is this used?

hpidcock (approved):
LGTM.

testing/src/scenario/state.py
tonyandrewmeyer: Down in line 2160.

PR #2554, opened 2026-06-09
docs: clarify storage-detaching applies to both machine and K8s charms

### LINKED ISSUE #2257: Improve guidance about storage-detaching event
In [How to manage storage](https://documentation.ubuntu.com/ops/latest/howto/manage-storage/#handle-storage-detaching) we have a section called "Handle storage detaching".

We should clarify whether this section applies to machine charms, K8s charms, or both.

Juju only supports manually detaching storage for machine charms. If you try `juju detach-storage` for a K8s charm, you'll get an error:

> ERROR Juju command "detach-storage" not supported on container models

However, I don't think this means it's pointless to observe the storage-detaching event in a K8s charm. You might still want to observe the event so that you can clean up storage during the teardown phase.

tonyandrewmeyer:
I double-checked the items in the issue against Juju 3.6.23. A minimal `ops` charm observed `storage-attached`, `storage-detaching`, `stop`, and `remove` and logged each event.

| Claim | Verified | Evidence |
|---|---|---|
| Machine: `juju detach-storage` emits `storage-detaching` | yes | `STORAGE-TEST: storage-detaching event for /var/lib/storage-test/data/3` |
| Machine: `juju remove-storage` emits `storage-detaching` | yes | `STORAGE-TEST: storage-detaching event for /var/lib/storage-test/data/4` (requires `--force` while attached) |
| Machine: emitted during unit teardown | yes | `data/5` storage-detaching fired before `stop`/`remove` on `juju remove-application --destroy-storage` |
| K8s: `juju detach-storage` returns `ERROR Juju command "detach-storage" not supported on container models` (verbatim) | yes | Returned exactly that string |
| K8s: `storage-detaching` fires during unit teardown | yes | `data/0` storage-detaching fired before `stop`/`remove` on `juju remove-application --destroy-storage` |

[Preview](https://canonical-ubuntu-documentation-library--2554.com.readthedocs.build/ops/2554/howto/manage-storage/#handle-storage-detaching)

Fixes #2257


tonyandrewmeyer:
3.10 failures are unrelated (as one would hope for a docs PR). Will be fixed by another PR and rebasing will fix here.

docs/howto/manage-storage.md
```diff
@@ -166,6 +166,12 @@ Alternatively, use {external+charmlibs:class}`pathops.ContainerPath` to access `
 
 ## Handle storage detaching
 
+This section applies to both machine and Kubernetes charms.
+
+For a machine charm, Juju emits a storage-detaching event whenever attached storage is being released: when a user runs `juju detach-storage` (or `juju remove-storage --force`), and during unit teardown. Observing the event lets the charm release resources cleanly before the storage goes away.
+
+For a Kubernetes charm, `juju detach-storage` isn't supported (it returns `ERROR Juju command "detach-storage" not supported on container models`), so storage is only detached during unit teardown, and cleanup can be done in the `stop` or `remove` events.
```
james-garner-canonical: WDYT about frontloading the common info + recommendation, and then mentioning the difference?
```suggestion
Juju emits a `storage-detaching` event whenever attached storage is being released. This is emitted during unit teardown for both machine and Kubernetes charms. Observe this event to cleanly release resources before the storage goes away.

On machine charms only, `storage-detaching` is also emitted when a user runs `juju detach-storage` (this command isn't currently supported for Kubernetes).
```

I'm not clear on what we're aiming to convey beyond this with the 'cleanup can be done in the `stop` or `remove` events` bit for Kubernetes. I've dropped it in the suggestion, but maybe it should instead be expanded on?

tonyandrewmeyer: >  I'm not clear on what we're aiming to convey beyond this with the 'cleanup can be done in the `stop` or `remove` events` bit for Kubernetes. I've dropped it in the suggestion, but maybe it should instead be expanded on?

It's a modified version of a suggestion in the issue. The issue suggests saying that storage-detaching can be used for cleanup in K8s, so is still useful there even though you only get it in teardown, since storage lifetime is tied to the life of the container. David doesn't remember the source of that suggestion.

I don't agree with that suggestion. I don't see any benefit in doing storage tidying in storage-detaching on K8s rather than in stop or remove. So I added the version in the PR.

Another way of handling this would be to focus the events specifically on machines, and a small comment mentioning that they happen on K8s but are not generally useful, and why.

dwilding: I think it would be easier to explain if we lead with this being about storage cleanup, rather than storage detachment.

The first part of the doc is organised into a machine section and a K8s section. How about adding a subsection in both sections called "Clean up storage".

The machine version could talk about `storage-detaching` and how it's emitted on demand and on teardown.

The K8s version could essentially be a recommendation to handle storage cleanup in `stop` or `remove`, as there's no on-demand way to request storage detachment. I think it's fine to mention `storage-detaching` here, but emphasise that we don't recommend observing it in K8s charms.

james-garner-canonical: Why don't we recommend observing `storage-detaching` in K8s charms? Because you should put all your cleanup logic in `stop` and/or `remove` and assume storage will be detached afterwards (and hasn't been already)?

james-garner-canonical: David's suggestion sounds reasonable to me. My suggestion should be ignored if we don't recommend `storage-detaching` for K8s charms.

tonyandrewmeyer: > Why don't we recommend observing `storage-detaching` in K8s charms? Because you should put all your cleanup logic in `stop` and/or `remove` and assume storage will be detached afterwards (and hasn't been already)?

Well, the issue said we *should* recommend it. I talked to David about it last week (because he opened the issue) arguing that since you get the storage detached at the same time (more or less) as stop/remove, why would you bother doing it separately. Like I would do all my setup in pebble-ready, not storage-attached.

If you feel observing storage-detached is better, I'm happy to hear the argument.

james-garner-canonical: I suppose an argument in favour of observing it would be to reduce differences between substrates. I don't know how substrate specific the non-observing code would be, which might change things, though the charm could always do K8s/machine specific helpers.

That said, maybe we don't really know what the right recommendation is here? At least I'm not convinced we should be recommending *avoiding* observing this event in K8s charms.

tonyandrewmeyer: I'm ok being more neutral. My concern was more recommending *for* using storage-detached. I still don't see any benefit, though. I'll do the change David suggested too.

james-garner-canonical (approved):
Looks good!

docs/howto/manage-storage.md
```diff
@@ -86,6 +86,30 @@ self.model.storages.request('cache', 2)  # Request two more instances.
...
+
+In the `src/charm.py` file, in the `__init__` function of your charm, set up an observer for the detaching event associated with your storage and pair that with an event handler. For example:
+
+```python
+framework.observe(
+    self.on['cache'].storage_detaching, self._on_storage_detaching
+)
+```
+
+Now, in the body of the charm definition, define the event handler, or adjust an existing holistic one. For example, to warn users that data won't be cached:
```
dwilding:
```suggestion
Now, in the body of the charm class, define the event handler, or adjust an existing holistic one. For example, to warn users that data won't be cached:
```

docs/howto/manage-storage.md
```diff
@@ -164,29 +188,11 @@ charm_cache_root = pathops.LocalPath(charm_cache_path)
...
-Now, in the body of the charm definition, define the event handler, or adjust an existing holistic one. For example, to warn users that data won't be cached:
-
-```python
-def _on_storage_detaching(self, event: ops.StorageDetachingEvent):
-    """Handle the storage being detached."""
-    self.unit.status = ops.ActiveStatus(
-        'Caching disabled; provide storage to boost performance'
-    )
-```
+On Kubernetes, `juju detach-storage` isn't supported, so storage is only detached during unit teardown. Juju emits a [storage-detaching](ops.StorageDetachingEvent) event during teardown, along with the other teardown events, like `stop` and `remove`.
```
dwilding:
```suggestion
On Kubernetes, `juju detach-storage` isn't supported, so storage is only detached during unit teardown. Juju emits a [storage-detaching](ops.StorageDetachingEvent) event during teardown, along with `stop`, `remove`, and other teardown events.
```

dwilding (approved):

PR #2555, opened 2026-06-10
feat: note the socket path in Pebble tracing spans

### LINKED ISSUE #1884: Tracing: note the container name in Pebble spans
Today, a charm would call:

```py
self._amf_container = self.unit.get_container(self._amf_container_name)
...
if not self._amf_container.can_connect(): ...
```

Which is:

```py
    def can_connect(self) -> bool:
        try:
            self._pebble.get_system_info()
        ...
```

Which is:

```py
    def get_system_info(self) -> SystemInfo:
        """Get system info."""
        with tracer.start_as_current_span('pebble get_system_info'):
            resp = self._request('GET', '/v1/system-info')
            return SystemInfo.from_dict(resp['result'])
```

Where neither `._request` nor `._request_raw` are instrumented.

This generates a tracing span.

Except we lost the container name in the process.

Which is fine for ~90% charms that have only one container, but must be clunky for those charms that have at least two.

tonyandrewmeyer:
The PR adds a `pebble.socket_path` attribute when creating tracing spans, so that each span includes which container it relates to (for example, executing a command in this specific container). If Pebble is not being used in a container, the path is still likely a useful data point if there is more than one Pebble involved.

Fixes #1884


tonyandrewmeyer:
This was the obvious way to do it, but it feels heavy for just getting the name into spans. We could possibly do it automatically (either let pebble ask for it, or make assumptions about the socket name, or just use the socket name and not the container). Or we could have an attribute on the client without changing the get_pebble method, or get/set methods on the client.

I'm going to think/experiment a bit more and talk this over with Ben next week.

tonyandrewmeyer:
Ben and I discussed this today over a call.

> This was the obvious way to do it, but it feels heavy for just getting the name into spans.

Ben didn't feel it was *too* heavy.

> We could possibly do it automatically (either let pebble ask for it, or make assumptions about the socket name, or just use the socket name and not the container).

Let's indeed see whether the socket name would work instead of passing a name through. Is this likely to be too long for an attribute in tracing systems?

If we don't use the socket name, then consider whether "name" is the right name, particularly outside of the Juju context (maybe even outside of container use).

ops/pebble.py
```diff
@@ -2241,6 +2242,10 @@ class Client:
     connection attempt to Pebble; used by ``urllib.request.OpenerDirector.open``. It's not for
     methods like :meth:`start_services` and :meth:`replan_services` mentioned above, and it's not
     for the command execution timeout defined in method :meth:`Client.exec`.
+
+    The ``socket_path`` is attached to every tracing span emitted by the client as the
+    ``pebble.socket_path`` attribute, which is useful for distinguishing spans in charms
```
tonyandrewmeyer:
```suggestion
    ``pebble.socket_path`` attribute, which is useful for distinguishing spans where multiple
```

ops/pebble.py
```diff
@@ -2241,6 +2242,10 @@ class Client:
...
+    with more than one container.
```
tonyandrewmeyer:
```suggestion
    pebble daemons are used.
```

tonyandrewmeyer:
I like this much more, and I think it provides what's needed. In the Juju context, the socket paths should be as consistent as the container names are, so you get the same information, and can manipulate strings if you want just the name.

I couldn't find any documented limits for the length of the attributes.

I'll try to test this against a real trace receiver.

tonyandrewmeyer:
Tested with Tempo/Grafana and seems fine:

IMG_5432

ops/pebble.py
```diff
@@ -2241,6 +2242,10 @@ class Client:
     connection attempt to Pebble; used by ``urllib.request.OpenerDirector.open``. It's not for
     methods like :meth:`start_services` and :meth:`replan_services` mentioned above, and it's not
     for the command execution timeout defined in method :meth:`Client.exec`.
+
+    The ``socket_path`` is attached to every tracing span emitted by the client as the
+    ``pebble.socket_path`` attribute, which is useful for distinguishing spans where multiple
+    pebble daemons are used.
```
dwilding: I had to read this a couple of times to fully get the point. How about this. Also pebble --> Pebble.

```suggestion
    ``socket_path`` is available as the ``pebble.socket_path`` attribute in every tracing span
    emitted by the client. This is useful for distinguishing spans where multiple Pebble daemons
    are used.
```

dwilding (approved):
Thanks 🙂 

This part of the PR description is a bit odd:

> The PR adds a `pebble.socket_path` attribute to the Pebble `Client`

It seems to be mixing up the `Client.socket_path` parameter, which already exists, and `pebble.socket_path` in tracing spans, which is new.

tonyandrewmeyer:
> This part of the PR description is a bit odd:
> 
> > The PR adds a `pebble.socket_path` attribute to the Pebble `Client`

Ah, yes, sorry. The first version of this added a `name` attribute to the client and piped that through. Careless editing on my part when I changed it to just use the socket instead. I'll fix.

tromai (approved):
Thank you. The change looks great to me.

PR #2557, opened 2026-06-10
refactor: de-vendor the certificate transfer and tracing interface handling

tonyandrewmeyer:
This PR replaces the two charm libraries vendored under `tracing/ops_tracing/vendor/charms/` with small requirer-side dataclasses driven through `ops.Relation.load` / `ops.Relation.save`, removing pydantic (and its `pydantic-core`, `annotated-types`, `typing-inspection` transitive deps) from `ops_tracing`'s dependency tree. I believe this also removes Rust as a requirement for building charms that use `ops[tracing]`, although that's not as much of a problem as it was before the uv plugin (bug? feature?) skipping building pydantic from source.

The vendored copies had exactly one consumer (`ops_tracing` itself) and were already a fork, so this also drops the upstream API surface `ops_tracing` never exercises: the provider-side classes, the `TracingEndpointRequirer` / `CertificateTransferRequires` wrappers, the charmhub publish metadata, the redundant relation-direction validation, the `charm_tracing_config` convenience wrapper, and the pydantic-v1 compatibility branches.

This is the biggest difference compared to moving to the new charmlibs versions of the packages (the #2739 approach). In that case, we're dropping the hundreds of lines of unused code from operator (which is great) but it's still bundled into every charm, just in a different place in the venv.

To minimise the risk of getting out of sync with the official schema for the two interfaces, new tests are added that check that our implementation matches the upstream one.


james-garner-canonical:
Driveby comment -- there's a larger discussion to be had here as to how we want to handle this long term. With the tracing charm library migrating to charmlibs this cycle, we could in principle depend on it a as a regular Python dependency instead of vendoring. On the other hand, if our needs really do diverge, then maybe we'd prefer to switch to vendoring our own tracing logic in a more first-class way (with this PR being one step on that road).

tonyandrewmeyer:
> Driveby comment -- there's a larger discussion to be had here as to how we want to handle this long term.

Yes, that's one of the reasons this is in draft. I'll address it before moving it out.

> With the tracing charm library migrating to charmlibs this cycle, we could in principle depend on it a as a regular Python dependency instead of vendoring.

We obviously should do that when it happens, if we're still vendoring. But that's not the point of the issue. The point is to get rid of large dependencies that are not really needed, and slim down the code to only what is actually required.

james-garner-canonical:
> > Driveby comment -- there's a larger discussion to be had here as to how we want to handle this long term.
> 
> Yes, that's one of the reasons this is in draft. I'll address it before moving it out.

:)

> > With the tracing charm library migrating to charmlibs this cycle, we could in principle depend on it a as a regular Python dependency instead of vendoring.
> 
> We obviously should do that when it happens, if we're still vendoring.

It's not entirely clear to me that we should, given the desire to keep Ops as dependency free as possible, and given the trickiness introduced by the mutual dependency between Ops and the tracing library. If we can feasibly maintain our own slim subset of the functionality the library provides instead, I think that's worth considering.

> But that's not the point of the issue. The point is to get rid of large dependencies that are not really needed, and slim down the code to only what is actually required.

The more our vendored copy drifts (e.g. dropping Pydantic), the more involved migrating to the `charmlibs` version becomes. If we definitely intend to use the `charmlibs` version, which we have roadmap time allocated to migrating this cycle, then IMO we shouldn't be aiming to merge this PR -- and I figured it was worth raising that point while it's still in draft.

But I fully support the idea of exploring how slim we can make this, that's definitely an important point in the dependency vs own-our-own discussion.

tonyandrewmeyer:
> The more our vendored copy drifts (e.g. dropping Pydantic), 

What's currently here is not a vendored copy that drops Pydantic, it's a library built for just ops-tracing. It's not in a vendor folder, and it doesn't have anything in it that isn't used by ops-tracing. It might be too similar to the old version and need reworking - it's a draft and I haven't looked very closely at the code yet, just given instructions to an agent.

> If we definitely intend to use the `charmlibs` version,

My understanding was that we definitely intend to *not* use the regular library, wherever it is hosted. I thought when we'd discussed this previously everyone was in favour of doing that, just not of allocating roadmap time to it.

> which we have roadmap time allocated to migrating this cycle

My understanding was we are doing that because it's the library charms should use for workload tracing, and we want to accelerate getting people away from Charmhub hosted libraries, rather than anything to do with our own use. Did I miss something?

> But I fully support the idea of exploring how slim we can make this, that's definitely an important point in the dependency vs own-our-own discussion.

If this is more of an open question than I understood, I don't mind that. The direction here will be more-or-less the same either way. I am aware of the love for keeping Pydantic in charms and wouldn't be surprised if, even though my understanding is we already agreed to it, this gets rejected.

james-garner-canonical:
I'm definitely in favour of the 'maintain our own slim, Pydantic-free tracing stuff' direction, so that's music to my ears. I must have forgotten the previous discussion where we agreed to that, and gotten confused by the PR description (I haven't looked at the code).

dimaqq:
Happy to see this (whichever way it ends up) 🚀

.github/workflows/tracing-upstream-schemas.yaml
```diff
@@ -0,0 +1,33 @@
+name: Tracing upstream schema drift
+
+# Opt-in drift check: compares ``ops_tracing``'s de-pydantic'd dataclasses
+# against the canonical/charmlibs schemas at HEAD. Needs network access and
+# pydantic, so it's gated to PRs/pushes that touch the tracing tree.
```
tonyandrewmeyer:
```suggestion
# pydantic, so it's limited to PRs/pushes that touch the tracing tree.
```

tracing/tox.ini
```diff
@@ -21,7 +21,25 @@ base = unit
...
+[testenv:upstream-schemas]
+description =
+    Drift check: compare our de-pydantic'd dataclasses against the upstream
+    canonical/charmlibs schemas. Opt-in (requires network + pydantic);
+    not part of the default ``unit`` run.
+# Deliberately NOT on the uv-venv-lock-runner the other envs inherit: this env
+# needs pydantic, and the lock runner installs only what `uv.lock` resolves.
+# Adding pydantic to a dependency group would put it back in `uv.lock` for
+# ops-tracing — exactly what this branch removes — so the drift check keeps its
+# own virtualenv and its own `deps`.
```
tonyandrewmeyer, line 35: I believe this is wrong, because the ops tests *should* have pydantic *only for testing*, so that the typed loading/saving tests get exercised. I think they have a skip if pydantic isn't installed (maybe that's not really needed any more, since uv handles everything), which would be why the problem doesn't show up.

tracing/test/test_upstream_schemas.py
```diff
@@ -0,0 +1,240 @@
...
+
+This is an "opt-in" test: it requires network access to fetch the upstream
+schemas at canonical/charmlibs HEAD, and it requires pydantic to instantiate
+them. It is not part of the default ``unit`` tox env; run it via::
+
+    tox -e upstream-schemas
+
+The intent is to detect drift: if the canonical schemas under
+``interfaces/{tracing,certificate_transfer}/`` change shape, this test should
+fail and force a conscious decision about whether to follow upstream.
```
tonyandrewmeyer:
```suggestion
fail and force an explicit decision about whether to follow upstream.
```

tracing/test/test_upstream_schemas.py
```diff
@@ -0,0 +1,240 @@
...
+    with urllib.request.urlopen(url, timeout=30) as resp:  # ruff: ignore[suspicious-url-open-usage]
+        return resp.read().decode('utf-8')
+
+
+def _load_upstream(url: str) -> dict[str, typing.Any]:
+    """Exec an upstream schema.py in an isolated namespace.
+
+    The upstream files import ``interface_tester.schema_base.DataBagSchema``,
+    which is a tiny pydantic ``BaseModel`` subclass. Stub it so we don't have
+    to install ``pytest-interface-tester``.
```
tonyandrewmeyer, line 66: I think @james-garner-canonical is in the process of changing this, so it might need a refresh.

james-garner-canonical: https://github.com/canonical/charmlibs/pull/664

PR #2558, opened 2026-06-10
fix: don't leak exec I/O threads when waiting on the change fails

### LINKED ISSUE #2556: `ExecProcess.wait()`/`wait_output()` leak non-daemon I/O threads when `wait_change` raises (e.g. exec timeout)
I first noticed the issue in the OpenSerach k8s charm where the charm just hangs indefinitely on a hook.

Disclaimer: I generated the issue description with AI

## Summary

When an `ExecProcess`'s underlying change wait fails — most commonly when an exec
timeout elapses, but also on any error from `wait_change()` — `ExecProcess._wait()`
propagates the exception **before** it joins its I/O pump threads and shuts down the
exec websockets. Because those pump threads are created with the default
`daemon=False`, they remain blocked in `websocket.recv()` on websockets that are
never closed.

This has two consequences:

1. **Interpreter shutdown hangs.** `threading._shutdown()` blocks forever joining the
   still-running non-daemon threads (observed as a hang in `lock.acquire()` at process
   exit).
2. **Noisy unraisable tracebacks.** If/when the threads eventually unblock (peer closes
   the connection), they raise
   `websocket._exceptions.WebSocketConnectionClosedException: socket is already closed.`
   to `threading.excepthook`, printing full tracebacks from
   `shutil.copyfileobj` → `ops.pebble._WebsocketReader.read` → `ws.recv()`.

## Root cause

In `ExecProcess._wait()` (`ops/pebble.py` ~L1784):

```python
def _wait(self) -> int:
    self._waited = True
    timeout = self._timeout
    if timeout is not None:
        timeout += 1
    change = self._client.wait_change(self._change_id, timeout=timeout)   # L1790  <-- can raise

    # --- cleanup below only runs if wait_change() returned normally ---
    if self._cancel_stdin is not None:        # L1793
        self._cancel_stdin()
    for thread in self._threads:              # L1797
        thread.join()
    if self._cancel_reader is not None:       # L1802
        os.close(self._cancel_reader)
    self._control_ws.shutdown()               # L1806
    self._stdio_ws.shutdown()
    if self._stderr_ws is not None:
        self._stderr_ws.shutdown()
    ...
```

`self._client.wait_change(...)` at L1790 raises on the client-side websocket timeout
(`timeout + 1`) when the server doesn't complete the change in time. The exception
propagates straight out of `_wait()`, so the cleanup at L1793–L1809 (cancel stdin,
join threads, shutdown websockets) is skipped.

The leaked threads are:

- the `shutil.copyfileobj` pumps started in `wait_output()` (L1845–L1850), and
- the forwarder threads started in `Client.exec()`
  (`_reader_to_websocket` / `_websocket_to_writer`, ~L3172–L3197).

They are started via `_start_thread()` (L412–L416):

```python
def _start_thread(target, *args, **kwargs):
    thread = threading.Thread(target=target, args=args, kwargs=kwargs)  # daemon=False
    thread.start()
    return thread
```

Both `wait()` (L1769) and `wait_output()` (L1819) are affected, since both delegate to
`_wait()`.

## Reproduction

```python
container.exec(["sleep", "60"], timeout=5, combine_stderr=True, encoding="utf-8").wait_output()
```

The exec server-side timeout / client-side `wait_change` timeout fires, `wait_output()`
raises, and the stdout `copyfileobj` thread is left blocked on `self._stdio_ws`. The
process then either hangs at interpreter exit (threads still blocked) or prints, when
they later unblock:

```
Exception in thread Thread-14 (copyfileobj):
...
File ".../ops/pebble.py", line 1982, in read
    chunk = self.ws.recv()
File ".../websocket/_core.py", line 395, in recv
...
File ".../websocket/_socket.py", line 95, in recv
    raise WebSocketConnectionClosedException("socket is already closed.")
websocket._exceptions.WebSocketConnectionClosedException: socket is already closed.
```

## Impact

- Any consumer that runs `exec(..., timeout=...)` and handles the resulting exception
  can hang at process exit, or emit alarming-looking tracebacks from background threads
  that are actually benign teardown noise.
- This is particularly painful in Juju charm hooks: a single timed-out exec can wedge
  the hook process (Juju then has to kill it) or flood `juju debug-log` with thread
  tracebacks.

## Suggested fix

Guarantee the teardown runs whether or not `wait_change()` succeeds — i.e. move the
cancel-stdin / join-threads / shutdown-websockets steps into a `finally` block in
`_wait()`:

```python
def _wait(self) -> int:
    self._waited = True
    timeout = self._timeout
    if timeout is not None:
        timeout += 1
    try:
        change = self._client.wait_change(self._change_id, timeout=timeout)
    finally:
        if self._cancel_stdin is not None:
            self._cancel_stdin()
        for thread in self._threads:
            thread.join()
        if self._cancel_reader is not None:
            os.close(self._cancel_reader)
        self._control_ws.shutdown()
        self._stdio_ws.shutdown()
        if self._stderr_ws is not None:
            self._stderr_ws.shutdown()
    ...
```

> **Note:** `thread.join()` in the cleanup path is itself only safe to call **after**
> the websockets are shut down (otherwise the join can block on a still-blocked
> `recv()`); the team may want to shut the websockets down first, then join, in the
> failure path. A simpler/more robust belt-and-braces option is to also create the I/O
> pump threads as `daemon=True` so a missed teardown can never hang interpreter
> shutdown.

## Environment

- `ops` 3.7.1 (also present in 3.7.0; the relevant code is unchanged across recent 3.x)
- `websocket-client` 1.9.0
- Python 3.12 / 3.13

tonyandrewmeyer:
If the change for an exec doesn't finish before the client-side timeout, or waiting on it fails in any other way, `ExecProcess._wait()` propagates the exception without any teardown. The I/O pump threads are left blocked in `recv()` on the still open websockets, and being non-daemon threads they then block interpreter shutdown indefinitely, wedging the Juju hook. See #2556 for the reproduction details (note that the leak needs the change to outlive the client-side wait, for example an orphan holding the exec's stdio pipe open; a plain exec timeout completes the change server-side and already cleans up correctly).

This tears the exec down when the wait fails:

* Shut the websocket connections down at the OS level (`SHUT_RDWR`) rather than only closing the file descriptors. Closing a socket doesn't wake other threads blocked in `recv()` on it, so the `finally` approach suggested in the issue (or joining the threads first, as the success path does) would hang the hook immediately instead of at exit.
* Treat a connection closed underneath an I/O pump thread (`WebSocketException` or `OSError`) as end-of-stream rather than letting the exception escape the thread, which printed noisy `WebSocketConnectionClosedException` tracebacks via `threading.excepthook`.
* Create the exec I/O threads as daemon threads, so any teardown path that's still missed (including never calling `wait()` at all) can't block interpreter shutdown.
* Close any already-connected websockets when connecting the exec websockets fails partway through, rather than leaking them (found while auditing for related leaks).

The behaviour of the success path is unchanged: the threads are still joined before the websockets are shut down, so output is never truncated.

Verified against a real Pebble (v1.31.0) on Python 3.10 through 3.14: before the fix the reproduction from #2556 leaks two `copyfileobj` threads and hangs at exit; with the fix it raises `ops.pebble.TimeoutError` as before, leaves no threads behind, and exits cleanly.

Fixes #2556

### DIFF
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -431,6 +431,10 @@ def _force_close_websocket(ws: _WebSocket):
     closing a socket doesn't wake up other threads that are blocked in
     ``recv()`` on the same socket; an OS-level shutdown of the connection
     does.
+
+    This reaches into websocket-client's ``WebSocket.sock`` (the underlying
+    ``socket.socket``) because the library exposes no public way to shut the
+    connection down without first closing the fd.
     """
     sock = ws.sock
     if sock is not None:
@@ -1828,15 +1832,19 @@ class ExecProcess(Generic[AnyStr]):
         # If stdin reader thread is running, stop it
         if self._cancel_stdin is not None:
             self._cancel_stdin()
+            self._cancel_stdin = None
 
         # Wait for all threads to finish (e.g., message barrier sent)
         for thread in self._threads:
             thread.join()
 
         # If we opened a cancel_reader pipe, close the read side now (write
-        # side was already closed by _cancel_stdin().
+        # side was already closed by _cancel_stdin()). Clear it afterwards so
+        # that a second call to _wait() (e.g. wait() then wait_output()) can't
+        # close the same fd again, matching _teardown_after_error().
         if self._cancel_reader is not None:
             os.close(self._cancel_reader)
+            self._cancel_reader = None
 
         # Close websockets (shutdown doesn't send CLOSE message or wait for response).
         self._control_ws.shutdown()
--- a/ops/pebble.py
+++ b/ops/pebble.py
@@ -340,6 +340,8 @@ if TYPE_CHECKING:
 
 
 class _WebSocket(Protocol):
+    sock: socket.socket | None
+
     def connect(self, url: str, socket: socket.socket): ...
 
     def shutdown(self): ...
@@ -410,10 +412,33 @@ def _format_timeout(timeout: float) -> str:
 
 
 def _start_thread(target: Callable[..., Any], *args: Any, **kwargs: Any) -> threading.Thread:
-    """Helper to simplify starting a thread."""
-    thread = threading.Thread(target=target, args=args, kwargs=kwargs)
+    """Helper to simplify starting a thread.
+
+    The thread is marked as daemon so that exec I/O threads can never keep
+    the interpreter alive at exit, even when teardown is missed (for
+    example, if wait() is never called and the server holds the websockets
+    open).
+    """
+    thread = threading.Thread(target=target, args=args, kwargs=kwargs, daemon=True)
     thread.start()
     return thread
+
+
+def _force_close_websocket(ws: _WebSocket):
+    """Close a websocket so that reads and writes blocked on it unblock.
+
+    ``WebSocket.shutdown()`` only closes the socket's file descriptor, and
+    closing a socket doesn't wake up other threads that are blocked in
+    ``recv()`` on the same socket; an OS-level shutdown of the connection
+    does.
+    """
+    sock = ws.sock
+    if sock is not None:
+        try:
+            sock.shutdown(socket.SHUT_RDWR)
+        except OSError:
+            pass  # Already disconnected or closed.
+    ws.shutdown()
 
 
 class Error(Exception):
@@ -1787,7 +1812,18 @@ class ExecProcess(Generic[AnyStr]):
         if timeout is not None:
             # A bit more than the command timeout to ensure that happens first
             timeout += 1
-        change = self._client.wait_change(self._change_id, timeout=timeout)
+        try:
+            change = self._client.wait_change(self._change_id, timeout=timeout)
+        except BaseException:
+            # The wait failed or was interrupted, for example because the
+            # change didn't finish before the client-side timeout above. The
+            # I/O threads may still be blocked on their websockets, so tear
+            # the connections down to unblock them, and reap the threads
+            # before propagating the error, otherwise they're left running
+            # and at interpreter shutdown block the join of non-daemon
+            # threads (#2556).
+            self._teardown_after_error()
+            raise
 
         # If stdin reader thread is running, stop it
         if self._cancel_stdin is not None:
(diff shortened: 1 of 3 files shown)


hpidcock (approved):
Seems logical, not a python expert so had to read up on threading/daemon threads.

ops/pebble.py
```diff
@@ -410,12 +412,39 @@ def _format_timeout(timeout: float) -> str:
...
+    """
+    thread = threading.Thread(target=target, args=args, kwargs=kwargs, daemon=True)
     thread.start()
     return thread
 
 
+def _force_close_websocket(ws: _WebSocket):
+    """Close a websocket so that reads and writes blocked on it unblock.
+
+    ``WebSocket.shutdown()`` only closes the socket's file descriptor, and
```
tromai, line 430: Just to double check, `WebSocket` here refers to the `WebSocket` class provided in the `websocket-client` library ?

ops/pebble.py
```diff
@@ -3155,14 +3251,21 @@ def exec(
...
+                connected.append(stdio_ws)
                 if not combine_stderr:
                     stderr_ws = self._connect_websocket(task_id, 'stderr')
+                    connected.append(stderr_ws)
             except websocket.WebSocketException as e:
                 # Error connecting to websockets, probably due to the exec/change
-                # finishing early with an error. Call wait_change to pick that up.
+                # finishing early with an error. Close any websockets that did
+                # connect, then call wait_change to pick up the change error.
+                for ws in connected:
```
tromai, line 3267: Is it because we haven't start any operation on the sockets, so no teardown is needed here ?

test/test_real_pebble.py
```diff
@@ -268,6 +268,20 @@ def test_exec_timeout(self, client: pebble.Client):
             process.wait()
         assert 'timed out' in excinfo.value.err
 
+    def test_exec_wait_timeout_reaps_io_threads(self, client: pebble.Client):
+        # Regression test for canonical/operator#2556. The orphaned grandchild
+        # keeps the exec's stdio pipe open, so the change can't finish even
+        # after Pebble kills the command at the 0.1s timeout, and the wait
+        # then times out client-side. The I/O threads were left blocked on the
+        # websockets, and being non-daemon they then hung interpreter
+        # shutdown.
```
tromai: I'm not quite clear on this part.

Is the doc string describing the original buggy behavior ? Or does it describe the correct behavior, as captured in the test.

tromai (approved):
Thank you for the detailed PR description and the fix.

ops/pebble.py
tonyandrewmeyer: Yes, websocket-client's WebSocket. The _WebSocket Protocol at line 342 just mirrors its interface for typing.

ops/pebble.py
tonyandrewmeyer: Yeah, at this point no I/O threads have been started yet (we haven't reached the _start_thread calls below), so nothing is blocked in recv() waiting to be woken. We just need to release the file descriptors of any websockets that did connect, which is what WebSocket.shutdown() does on its own. _force_close_websocket() is only needed once readers are  running on the socket.

test/test_real_pebble.py
tonyandrewmeyer: Good point, the comment is ambiguous. It's describing the original buggy scenario as motivation for the test; the assertions below then verify the fix (the I/O threads exit). I'll restructure it so the "what this reproduces" and "what we now assert" are clearly separated.

PR #2559, opened 2026-06-10
docs: grab workload version in k8s tutorial

### LINKED ISSUE #2282: Grab workload version in K8s tutorial
As follow-up work to #1749, I recommend that we make use of the workload module that comes with the Charmcraft profile. I think it would be great to modify the workload module to fetch the version from the API server, then the charm can set the workload version accordingly.

I don't think this would need to be a separate chapter - we should be able to add it to [Create a minimal Kubernetes charm](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm/).

tromai:
Closes #2282.

This PR modify the `From zero to hero: Write your first Kubernetes charm` tutorial:
- Added a section in Chapter 1: creating a helper module for the charm to interact with the workload application. This is inspired by what we have in the machine charm tutorial. The helper module contains the function to get the version of the FastAPI web server.
- Added another section in Chapter 1: exposing this version as the workload version to Juju (following [this guide](https://documentation.ubuntu.com/ops/latest/howto/manage-the-workload-version/)).
- Added unit tests and integration tests to cover the above logic.
- Propagate the changes to the other chapters.

The PR reviewer can focus on the Markdown docs. We can leave the code changes in `examples/` at the end, when we have finalized the tutorial, as the changes in `examples/` are plenty and mostly reflect the tutorial.


docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -306,6 +307,86 @@ kubectl -n testing describe pod fastapi-demo-0
 
 In the output you should see the definition for both containers. You'll be able to verify that the default command and arguments for our application container (`demo-server`) have been displaced by the Pebble service. You should be able to verify the same for the charm container (`charm`).
 
+### Set your workload version
```
tromai: I put this section here because it builds on top of the previous sections:
- Pack and deploy the charm
- Using `curl` to fetch the version from the fastapi server
- Viewing `juju status`. With the workload version set, you can see the difference in the `Version` column, which I highlight in this section.

But please let me know if it's better to merge into the previous sections instead.

tromai: @dwilding Sorry for tagging you into a Draft PR. Just want to have your opinion on:
- Is is better to keep "grab workload version" in its own section, or merge it to the previous steps. I am leaning toward "merging", to keep it consistent with the machine charm tutorial.
- The overall structure of this PR: I want to finalize it first before updating the remaining chapters.

Thank you very much :smile:

dwilding: Don't worry about the ping, very happy to discuss 🙂 

By merging, do you mean adding a subsection within "Write your charm", so that the version grabbing behaviour is in the charm from the very first `charmcraft pack`? I would lean towards that too. I think a single pack in this chapter is better.

I think putting a subsection before "Define the charm class" would make sense, similar to the machine charm tutorial.

For the explanation of how a charm communicates with its workload - "...the charm container is in the same pod as the workload..." - I'd consider shifting that to the chapter introduction that talks about how K8s charms work. That introduction currently says "...managing the workload application in various ways via Pebble", which could get confusing if we suddenly introduce a non-Pebble way to talk to the workload. It would be best to explain both aspects of workload management/communication in the same place.

tromai: Thanks David! 

I have added a subsection within `Write your charm` to discuss about the helper module, following the machine charm tutorial. 

I also brought the discussion about charm container reaching workload container to the very beginning of the tutorial. I hope it is much better now.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -11,6 +11,7 @@ When you deploy a Kubernetes charm, the following things happen:
 1. The same Juju controller injects Pebble -- a lightweight, API-driven process supervisor -- into each workload container and overrides the container entrypoint so that Pebble starts when the container is ready.
 1. When the Kubernetes API reports that a workload container is ready, the Juju controller informs the charm that the instance of Pebble in that container is ready. At that point, the charm knows that it can start communicating with Pebble.
 1. Typically, at this point the charm will make calls to Pebble so that Pebble can configure and start the workload and begin operations.
+1. After that, our workload, the `api_demo_server` app, serves on the deployed IP on port 8000. The charm can reach that endpoint because our charm container and the workload container are deployed in the same pod.
```
tromai: I shouldn't mention about the `api_demo_server` app here, and keep it general. I will rewrite it.

examples/k8s-1-minimal/tests/unit/test_charm.py
```diff
@@ -15,11 +15,22 @@
...
 
 from charm import FastAPIDemoCharm
 
 
+def mock_get_version(port: int):
+    """Get a mock version string without executing the workload code."""
+    return "1.0.4"
+
+
+@pytest.fixture(autouse=True)
```
benhoyt: Let's avoid autouse fixtures unless absolutely necessary, as they can be quite confusing action-at-a-distance. Something like:

```python
@pytest.fixture
def fake_version(monkeypath):
    monkeypath.setattr(...)

def test_pebble_layer(fake_version):
    ...
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +233,31 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
...
+    # Add initial Pebble config layer using the Pebble API
+    container.add_layer("fastapi_demo", self._get_pebble_layer(), combine=True)
+    # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
+    container.replan()
+    # Set the workload version of this charm.
+    version = fastapi_demo.get_version(port=8000)
+    self.unit.set_workload_version(version)
+    # Learn more about statuses at
+    # https://documentation.ubuntu.com/juju/3.6/reference/status/
+    self.unit.status = ops.ActiveStatus()
```
benhoyt: It's hard to tell what's being added here. Could we consider something like this?

```suggestion
def _on_demo_server_pebble_ready(self, event: ops.PebbleReadyEvent) -> None:
    # ... previous lines ...
    # Set the workload version of this charm.
    version = fastapi_demo.get_version(port=8000)
    self.unit.set_workload_version(version)
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -440,6 +523,16 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
...
+
+
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
+    version = juju.status().apps["fastapi-demo"].version
+    # We'll need to update this version every time we upgrade to a new workload
+    # version. If the workload has an API or some other way of getting the
+    # version, the test should get it from there and use that to compare to the
+    # unit setting.
+    assert version == "1.0.4"
```
benhoyt: Consider using something a bit more general, to avoid updates if the demo-server is updated with a minor or patch release:

```suggestion
    assert version.startswith("1.")
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,49 @@ resources:
...
+```python
+import json
+import logging
+import urllib.request
+
+logger = logging.getLogger(__name__)
+
+
+def get_version(port: int) -> str:
+    """Get the version of fastapi_demo installed.
```
tromai:
```suggestion
    """Get the version of api_demo_server installed.
```
I should double check other places too.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -245,22 +306,22 @@ Monitor your deployment:
 juju status --watch 1s
 ```
 
-When all units are settled down, you should see the output below, where `10.152.183.215` is the IP of the K8s Service and `10.1.157.73` is the IP of the pod.
+When all units are settled down, you should see the output below, where `10.152.183.215` is the IP of the K8s Service and `10.1.157.73` is the IP of the pod. The workload version is `1.0.4`, as seen in the `Version` column.
```
tromai:
```suggestion
When all units are settled down, you should see the output below, where `10.152.183.215` is the IP of the K8s Service and `10.1.157.73` is the IP of the pod. The workload version is located in the `Version` column.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -245,22 +306,22 @@ Monitor your deployment:
...
 
 ```text
 Model    Controller     Cloud/Region  Version  SLA          Timestamp
 testing  concierge-k8s  k8s           3.6.13   unsupported  13:38:19+01:00
 
 App           Version  Status  Scale  Charm         Channel  Rev  Address         Exposed  Message
-fastapi-demo           active      1  fastapi-demo             0  10.152.183.215  no
+fastapi-demo  1.0.4    active      1  fastapi-demo             0  10.152.183.215  no
```
tromai, line 308: Should I go with a general version here, instead of hard-coding? 
```suggestion
fastapi-demo  1.*.*    active      1  fastapi-demo             0  10.152.183.215  no
```
Alternatively, I can add a small note saying that this workload version can be different on the readers machine.

I prefer going with a general version. But I'm not against the second approach.

tromai:
While working on this task, I noticed that in [Chapter 5](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite/#create-a-model-for-cos-lite), we introduced `requests` for the integration tests. However, we didn't mention adding this dependency using `uv add` or modifying `pyproject.toml`. Please let me know if this is a real gap in the doc, and I can fix it in this PR too.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: On second thought, we should use `fastapi_demo` to keep it consistent with the module name `fastapi_demo.py`.

dwilding:
> While working on this task, I noticed that in [Chapter 5](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite/#create-a-model-for-cos-lite), we introduced `requests` for the integration tests. However, we didn't mention adding this dependency using `uv add` or modifying `pyproject.toml`. Please let me know if this is a real gap in the doc, and I can fix it in this PR too.

Oops, you're quite right. Well spotted! In the tutorial we say "Then import `pytest`, `pytest_jubilant`, and `requests`" - but only the first two are predefined dependencies in the Charmcraft profile.

tromai:
> > While working on this task, I noticed that in [Chapter 5](https://documentation.ubuntu.com/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite/#create-a-model-for-cos-lite), we introduced `requests` for the integration tests. However, we didn't mention adding this dependency using `uv add` or modifying `pyproject.toml`. Please let me know if this is a real gap in the doc, and I can fix it in this PR too.
> 
> Oops, you're quite right. Well spotted! In the tutorial we say "Then import `pytest`, `pytest_jubilant`, and `requests`" - but only the first two are predefined dependencies in the Charmcraft profile.

Thanks for the confirmation David. I added the step to run `uv add requests` in the tutorial https://github.com/canonical/operator/pull/2559/commits/80bbc8316209d27306d1156b60d294a37573d7a8

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -11,6 +11,7 @@ When you deploy a Kubernetes charm, the following things happen:
 1. The same Juju controller injects Pebble -- a lightweight, API-driven process supervisor -- into each workload container and overrides the container entrypoint so that Pebble starts when the container is ready.
 1. When the Kubernetes API reports that a workload container is ready, the Juju controller informs the charm that the instance of Pebble in that container is ready. At that point, the charm knows that it can start communicating with Pebble.
 1. Typically, at this point the charm will make calls to Pebble so that Pebble can configure and start the workload and begin operations.
+1. Besides Pebble exec, the charm can communicate to the workload using HTTP requests. This is possible because they share the same pod (Kubernetes charms are deployed following sidecar pattern).
```
dwilding: I think it's good to make this feel more sequential, following on from "begin operations" in the previous item. Also I don't think we need to mention the sidecar pattern, as item 1 essentially explains it.

```suggestion
1. During operations, the charm may need to directly communicate with the workload using HTTP requests. This is possible because the charm container and workload container share the same pod.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,49 @@ resources:
     upstream-source: ghcr.io/canonical/api_demo_server:1.0.4
 ```
 
+### Write a helper module
+
+Your charm will interact with our workload application `api_demo_server`. It’s a good idea to write a helper module that wraps `api_demo_server`. Charmcraft created `src/fastapi_demo.py` as a placeholder helper module.
```
dwilding:
```suggestion
Your charm will interact with our workload application `api_demo_server`. It's a good idea to write a helper module that wraps `api_demo_server`. Charmcraft created `src/fastapi_demo.py` as a placeholder helper module.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,49 @@ resources:
...
+
+The helper module will be independent of the main logic of your charm. This will make it easier to test your charm. In this tutorial, the helper module only contains the logic to get the version of `api_demo_server`. The server has an endpoint at `/version` that returns a JSON payload containing the version number (see {ref}`study-your-application`).
+
+This is called the workload version. To make things easier for Juju admins, our charm should expose the workload version to Juju. It will be visible in `juju status`. For more information, see {ref}`how-to-set-the-workload-version`.
```
dwilding: I'm OK to drop the reference link because people typically follow in sequence and there's not much be gained by jumping back at this point.

A couple of other suggestions: It's best to use the term "Juju users" instead of admins (although we still say "admins" in a bunch of places for historical reasons). I recommend changing `juju status` because we haven't introduced this command yet.

```suggestion
The helper module will be independent of the main logic of your charm. This will make it easier to test your charm. In this tutorial, the helper module only contains the logic to get the version of `api_demo_server`. The server has an endpoint at `/version` that returns a JSON payload containing the version number. This is called the workload version.

To make things easier for Juju users, your charm should expose the workload version to Juju. It will be visible in Juju's status output. For more information, see {ref}`how-to-set-the-workload-version`.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,49 @@ resources:
...
+
+We first define a function to get the workload version. Replace the content of `src/fastapi_demo.py` with:
```
dwilding: I think the first sentence can be dropped because we've already said what what we're going to do (and the code is short and clear enough)

```suggestion
Replace the content of `src/fastapi_demo.py` with:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: How about this instead?

```suggestion
    """Get the version of fastapi_demo that is running.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,49 @@ resources:
...
+
+    Args:
+        port: The port where fastapi_demo web server is listening.
+    """
+    response = urllib.request.urlopen(f"http://0.0.0.0:{port}/version")
+    data = json.loads(response.read())
+    return data["version"]
+```
+
+The `get_version` function sends a HTTP GET to `api_demo_server` in the workload container, decodes the result JSON payload, and extracts the version string.
```
dwilding: Personally, I would remove this sentence, as I think the code shows it clearly enough. But I could be too familiar with what's going on. Up to you.

If you decide to remove the sentence, maybe expand the docstring to "Get the version of fastapi_demo that is running in the workload container".

And I'd switch "a HTTP" to "an HTTP" for consistency with later in the doc (avoiding a big linguistic debate 😆)

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,49 @@ resources:
...
+
+Notice that the helper module is stateless. In fact, your charm as a whole will be stateless. The main logic of your charm will:
+
+1. Receive an event from Juju.
+2. Use the functions in the helper module to manage `api_demo_server` and check its status.
```
dwilding:
```suggestion
2. Use Pebble calls and the helper module to manage `api_demo_server` and check its status.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +233,22 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
     return ops.pebble.Layer(pebble_layer)
 ```
 
+### Set the workload version
+
+The workload version is available after the workload starts, which is after Pebble reevaluates its plan. We will use the `src/fastapi_demo.py` helper module for this step.
```
dwilding: Very minor suggestions

```suggestion
The workload version is available after the workload starts, which happens after Pebble reevaluates its plan. We'll use the `src/fastapi_demo.py` helper module for this step.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +233,22 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
...
+
+In `src/charm.py`, add `import fastapi_demo` in the imports at the top of the file. Then append the following lines to the `_on_demo_server_pebble_ready` function:
```
dwilding: Normally there's no need to worry about the import, as it should already be there from the Charmcraft profile. See https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/src/charm.py.j2#L13

But we've lost it thanks to the "Deploy your charm" step. I'm not sure what's best, but I lean towards updating the code in "Deploy your charm" so that it keeps the import. What do you think? Then making an edit here:

```suggestion
In `src/charm.py`, append the following lines to the `_on_demo_server_pebble_ready` function:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +233,22 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
...
+
+```python
+def _on_demo_server_pebble_ready(self, event: ops.PebbleReadyEvent) -> None:
+    # ... previous lines ...
+    # Set the workload version of this charm.
+    version = fastapi_demo.get_version(port=8000)
+    self.unit.set_workload_version(version)
```
dwilding: We're considering updating the tutorial to pull in code from the charm source files, rather than having the code in the doc source, so I think it's best to reduce this to just the relevant lines (which I feel should be clear enough):


```suggestion
    # Set the workload version of this charm.
    version = fastapi_demo.get_version(port=8000)
    self.unit.set_workload_version(version)
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +233,22 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
...
+```
+
+We invoke `fastapi_demo.get_version` to get the workload version, and expose it to Juju with `self.unit.set_workload_version`. We use port 8000 as the app is deployed on this port, as seen in `_get_pebble_layer`.
```
dwilding: I'd drop the explanation of `fastapi_demo.get_version`, but keep `self.unit.set_workload_version` because that will be a novel function for people. Also how about this restructuring?

```suggestion
We get the workload version over port 8000 because `_get_pebble_layer` deploys the app on this port. Then `self.unit.set_workload_version` exposes the workload version to Juju.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -245,22 +306,22 @@ Monitor your deployment:
 juju status --watch 1s
 ```
 
-When all units are settled down, you should see the output below, where `10.152.183.215` is the IP of the K8s Service and `10.1.157.73` is the IP of the pod.
+When all units are settled down, you should see the output below, where `10.152.183.215` is the IP of the K8s Service and `10.1.157.73` is the IP of the pod. The workload version is located in the `Version` column.
```
dwilding: Suggest clarifying as there are two Version columns in the output:

```suggestion
When all units are settled down, you should see the output below, where `10.152.183.215` is the IP of the K8s Service and `10.1.157.73` is the IP of the pod. The workload version is located in the app's `Version` column.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: I think showing 1.0.4 is fine as you have it. No further explanation needed. We control `api_demo_server`, and when we release a new version we go through and update the tutorial.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -325,12 +386,23 @@ Replace the contents of `tests/unit/test_charm.py` with:
...
+import pytest
 from ops import testing
 
 from charm import FastAPIDemoCharm
 
 
-def test_pebble_layer():
+def mock_get_version(port: int):
+    """Get a mock version string without executing the workload code."""
+    return "1.0.4"
```
dwilding: Let's make this an obviously mock version number. Something like:

```suggestion
    return "0.0.1"
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -361,10 +433,14 @@ def test_pebble_layer():
         state_out.get_container(container.name).service_statuses["fastapi-service"]
         == ops.pebble.ServiceStatus.ACTIVE
     )
+    # Check the workload version is set
+    assert state_out.workload_version is not None
 ```
 
 This test checks the behaviour of the `_on_demo_server_pebble_ready` function that you set up earlier. The test simulates your charm receiving the pebble-ready event, then checks that the unit and workload container have the correct state.
 
+In unit tests, we avoid any interaction with the outside world to keep the unit tests deterministic. The `get_version` method performs a HTTP call, which must be patched. We use the fixture `mock_version` to achieve this. From now on, any unit test on the charm needs to use fixture.
```
dwilding: Also keeps them fast 😁 I think it's fine to reduce.

This statement isn't _necessarily_ true: "From now on, any unit test on the charm needs to use fixture". It depends on which event is being testing, and also whether we've defined a collect-status handler gets the version (which would then run after any other event). I think it's best to avoid mentioning it in this chapter - we can address the topic later when it comes up.

```suggestion
In unit tests, we avoid any interaction with the outside world. The `get_version` method performs an HTTP call, which must be patched. We use the `mock_version` fixture to achieve this.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -440,6 +516,16 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
...
+
+
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
+    version = juju.status().apps["fastapi-demo"].version
+    # Ideally, the test should get the version directly from the workload application
+    # (for example, through an API call) and use that in this assertion.
+    # For simplicity, we hardcode the major version here to avoid updates
+    # if api_demo_server is updated with a minor or patch release.
+    assert version.startswith("1.")
```
dwilding: I know `1.` came from Ben's suggestion, but I disagree 😉 

The next version of `api_demo_server` will likely be `2.x`. And we're already committed to finding-and-replacing version numbers when we release a new version of the server. So it would actually be easier if this stayed as `version == "1.0.4"`.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,29 @@ def _replan_workload(self) -> None:
...
-
-        self.unit.status = ops.ActiveStatus()
     except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
         logger.info("Unable to connect to Pebble: %s", e)
         self.unit.status = ops.MaintenanceStatus("Waiting for Pebble in workload container")
+        return
+
+    try:
+        version = fastapi_demo.get_version(config.server_port)
+    except (urllib.error.URLError, json.JSONDecodeError) as version_e:
```
dwilding: Would it be more practical to put the error handling inside the function in the workload module, perhaps returning a runtime error on failure? Then we wouldn't need to import `json` and `urllib.error` in the charm.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,29 @@ def _replan_workload(self) -> None:
...
+        logger.error(
+            "Failed to get version from the server: %s. Please double check your port config",
+            version_e,
+        )
+        self.unit.status = ops.BlockedStatus(str(version_e))
```
dwilding: I think it's best to use a hardcoded status string here. Blocked status implies that the Juju user should be able to unblock the charm by themself, and ideally the message should give some indication of how.

As I write this, now I'm wondering whether blocked is really right. I'd like to hear other opinions - but here's what I'm thinking:

- If getting the version fails because we can't connect, it might be that the workload container is temporarily unavailable. The charm should go into Maintenance status with an appropriate message. The user needs to wait.
- If we got the version, but we couldn't decode it, there's nothing the user can do. We shouldn't catch the error; we should let the charm go into Error state. (The user needs to bug the charm developer about it.)

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,29 @@ def _replan_workload(self) -> None:
...
+        return
+
+    self.unit.status = ops.ActiveStatus()
+    self.unit.set_workload_version(version)
 ```
 
 When the config is loaded as part of creating the Pebble layer, if the config is invalid (in our case, if the port is set to 22), then a `ValueError` will be raised. The `_replan_workload` method handles that by logging the error and setting the status of the unit to blocked, letting the Juju user know that they need to take action.
 
+We also add error handling for getting the workload version. If the charm fails to obtain a valid version number, we set the unit status to blocked. This lets the Juju user know that either the charm code or the workload application behaved unexpectedly.
```
dwilding: This will need updating if the error handling strategy changes (which I think it should)

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -242,7 +256,7 @@ Since we added a new feature to configure `server-port` and use it in the Pebble
 First, we'll add a test that sets the port in the input state and asserts that the port is used in the service's command in the container layer:
 
 ```python
-def test_config_changed():
+def test_config_changed(mock_version):
```
dwilding, line 247: [Continuing discussion from another comment]

I agree that we need `mock_version` here. Because `_on_config_changed` calls `_replan_workload`, which gets the workload version.

I think it would be worth adding a brief sentence or two immediately after the code block, to explain why `mock_version` is needed in this instance. Then don't mention it again for the rest of the tutorial.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding, line 301: Arguably we should get the workload version at the end of this method, as a way of confirming that the workload has started correctly. I'd class that as icing on the cake, not a must-have for this PR.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -156,7 +156,7 @@ Congratulations, you now know how to expose operational tasks via actions!
 Let's add a test to check the behaviour of the `get_db_info` action that we just set up. Our test sets up the context, defines the input state with a relation, then runs the action and checks whether the results match the expected values:
 
 ```python
-def test_get_db_info_action():
+def test_get_db_info_action(mock_version):
```
dwilding: This test shouldn't need `mock_version`, because `_on_get_db_info_action` doesn't end up getting the workload version. But if we change `_on_collect_status` so that it gets the workload version, then we will need `mock_version` here.

dwilding (commented):
This is a serious chunk of work - thanks for working through it so thoroughly! It's looking great. I'm sending a first batch of comments and will hold off reviewing further until you've had a chance to think about them.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
benhoyt: That's fine, thanks David -- let's keep it `version == '1.0.4'`

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
benhoyt: I discussed this a bit with Nhan, and a couple of things:

- Agreed it's better to use a hard-coded message, more like the log message, for example: "Failed to get version from server: check port config"
- I don't think Maintenance status would be good here. That means we the charm is doing some maintenance and we *know* it's going to come right in a minute. But here it might never come right until they fix the port. So I'd recommend keeping BlockedStatus
- Agreed re decoding being a hard error -- I think we should just omit the JSONDecodeError and let that one bubble up

james-garner-canonical (commented):
This is looking good. I agree with David's suggestions.

tromai:
Thanks everyone for the reviews. I really appreciate it ! 

I will address the feedback

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: > And I'd switch "a HTTP" to "an HTTP" for consistency with later in the doc (avoiding a big linguistic debate 😆)

Oh, thanks for pointing that out :smile: . I really wasn't aware of it hahaha.

I agree with you to remove the this sentence, as it doesn't add much benefit. We can also avoid maintaining this this description on top of the code.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: I will go with your suggestion. I also lean towards changing the code snippet, and then we can remove the redundant `add import fastapi_demo` part.  

> updating the code in "Deploy your charm" 

I assume you are referring to "Define the charm class" where we first introduce `src/charm.py`?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: Sounds good to me! I will reduce it. I will also remove the indentation as it is not necessary anymore. The final version will be 

````
In `src/charm.py`, append the following lines to the `_on_demo_server_pebble_ready` function:

```python
# Set the workload version of this charm.
version = fastapi_demo.get_version(port=8000)
```
````

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md (hunk 16 above)
Comment (dwilding) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md]: > I assume you are referring to "Define the charm class" where we first introduce `src/charm.py`?

Yes, I had "Define the charm class" in mind but must have copied a different heading that looked close enough!

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md (hunk 18 above)
Comment (tromai) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md]: That works really well! Thank you. I will go with this version. I agree with removing `fastapi_demo.get_version`

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md (hunk 19 above)
Comment (tromai) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md]: Will do. Thanks for the catch.

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md (hunk 26 above)
Comment (tromai) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md:247]: Thanks. I have added a small paragraph for this. See https://github.com/canonical/operator/pull/2559/commits/54d186ea5aea4e1eacb92b0dd56faf74c95a4a72

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
Comment (tromai) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md:301]: Interesting, I have never thought of getting workload version this way.

I assumed that `_replan_workload` sets `ActiveStatus` if the server has started properly. This `ActiveStatus` is read by `status.is_running()` here, so we don't need to get the workload version again. 

I have an impression that `ops.CollectStatusEvent` is used for ongoing 'status monitoring', so we can definitely get the workload version here, as a "check if the server is still running".

Comment (dwilding) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md:301]: The thing is, `status.is_running()` only tells us that the Pebble service is running, not whether the service is actually serving over HTTP.

Another complication: although we've put all the status reporting logic in `_on_collect_status` (good), we'd still need to call `get_version` in two places:
- In `_replan_workload`, to tell Juju the workload version - it's not good practice to do that in the collect-status handler
- In `_on_collect_status` to check that the service is serving

We've discussed these questions several times in the past and I think I still don't find any solution completely satisfying...

For the K8s tutorial and this PR, I think it's best to leave `get_version` out of the collect-status handler. I think that Active meaning "the service is running" is good enough.

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md [hunk 28]
```diff
@@ -260,10 +273,12 @@ def test_config_changed():
     assert "--port=8080" in command
 ```
 
+We needs the `mock_version` fixture because `_on_config_changed` calls `_replan_workload`, which gets the workload version by `get_version`. The fixture patches `get_version` to avoid making a real HTTP call and make this unit test deterministic.
```
dwilding: A few minor suggestions for readability

```suggestion
We need the `mock_version` fixture because `_on_config_changed` calls `_replan_workload`, which gets the workload version using `fastapi_demo.get_version`. The fixture patches `get_version` to avoid making a real HTTP call, so that the this unit test deterministic.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,28 @@ def _replan_workload(self) -> None:
         # service if required.
         self.container.replan()
         logger.info(f"Replanned with '{self.pebble_service_name}' service")
-
-        self.unit.status = ops.ActiveStatus()
     except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
         logger.info("Unable to connect to Pebble: %s", e)
         self.unit.status = ops.MaintenanceStatus("Waiting for Pebble in workload container")
+        return
+
```
dwilding: I feel we can drop the blank line, and the one after try/except, to match the rest of the function

```suggestion
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,28 @@ def _replan_workload(self) -> None:
...
+        )
+        return
+
+    self.unit.status = ops.ActiveStatus()
+    self.unit.set_workload_version(version)
 ```
 
 When the config is loaded as part of creating the Pebble layer, if the config is invalid (in our case, if the port is set to 22), then a `ValueError` will be raised. The `_replan_workload` method handles that by logging the error and setting the status of the unit to blocked, letting the Juju user know that they need to take action.
 
+We also add error handling for getting the workload version. If we cannot access the workload to get its version, a `RuntimeError` will be raised. The `_replan_workload` method will log the error, and set the unit status to `blocked` with a message. It lets the Juju user know that they need to check the port configuration.
```
dwilding:
```suggestion
We also add error handling for getting the workload version. If we cannot access the workload to get its version, a `RuntimeError` will be raised. The `_replan_workload` method will log the error, and set the unit status to `blocked` with a message. This lets the Juju user know that they need to check the port configuration.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -331,6 +340,12 @@ self.unit.status = ops.MaintenanceStatus("Waiting for Pebble in workload contain
 self.unit.status = ops.BlockedStatus(str(e))
 ```
 
+```python
+self.unit.status = ops.BlockedStatus(
+    "Failed to get version from server: check port config"
+)
```
dwilding: On balance, I think it's best to keep this in the charm code. (So only remove three `self.unit.status` lines.) The text above here says:

> other than anywhere we're wanting a status to show up *during* the event execution

I think failure to get the version is a reasonable candidate for setting status during event execution.

If we remove this `BlockedStatus` from this part of the charm code, we should recreate it in the collect-status handler. But on reflection, I don't like that idea so much. So I think the most practical approach is to keep it in `_replan_workload`.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
dwilding: Elsewhere I'm recommending we don't get the workload version in the collect-status handler, so I think `mock_version` can be dropped here.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
```diff
@@ -186,7 +186,7 @@ def test_get_db_info_action():
 Since the `get_db_info` action has a parameter `show-password`, let's also add a test to cover the case where the user wants to show the password:
 
 ```python
-def test_get_db_info_action_show_password():
+def test_get_db_info_action_show_password(mock_version):
```
dwilding: We should be able to drop `mock_version` here too.

dwilding (commented):
I'm holding off approving until the collect-status question is settled, but the latest changes look good 🙂 

I haven't cross-checked the doc pages against the charm source. I'd be very happy to do that as a final review step.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -480,7 +480,13 @@ Let's write some integration tests to check that our application works correctly
 
 The existing integration tests use a model that is created by the `juju` fixture. We'll define a similar fixture that creates a separate model for COS Lite.
 
-First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest`, `pytest_jubilant`, and `requests`. Your imports should now look like this:
+First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest` and `pytest_jubilant`, which are already dependencies of your charm. We also need a new dependency `requests`. Run the following command to add it:
```
dwilding: I think there's a (small) chance that the reader will miss that they need to import `requests`. How about this adjusted wording:

```suggestion
First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest` and `pytest_jubilant`, which are already dependencies of your charm. Also import `requests`, which is a new dependency. Run the following command to add `requests` as a dependency:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tromai: Ah I see. Thank you so much for the explanation. 

> For the K8s tutorial and this PR, I think it's best to leave get_version out of the collect-status handler.

I agree. I will not add `get_version` to this `_on_collect_status` method.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tromai: Thanks for the catch. That's my habit of adding newlines after try/catch and such :smile:, I should have kept the original style. 

I have fixed it in https://github.com/canonical/operator/pull/2559/commits/49e6a120fd682e8fa0a8c1bae5113c402727201a.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/expose-operational-tasks-via-actions.md
tromai: Thank you. I have removed `mock_version` for `test_get_db_info_action` and `test_get_db_info_action_show_password` in our tutorials.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tromai: Thanks for pointing that out. Yes, I also think it's best to keep this in `_replan_workload` (https://github.com/canonical/operator/pull/2559/commits/272592ca173e09bc9653f97f88446a9ea2b71be1).

In the beginning, I made this change following the removal of `self.unit.status = ops.BlockedStatus(str(e))`, which was for exceptions from `self.load_config(FastAPIConfig)`. The failure of `get_version` can happen not only from wrong port config, but also from a broken workload.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -236,9 +236,19 @@ def _replan_workload(self) -> None:
...
+        logger.error("Failed to get workload version: %s", version_e)
+        self.unit.status = ops.BlockedStatus(
+            "Failed to get version from server: check port config"
+        )
+        return
+    self.unit.set_workload_version(version)
 ```
 
-We removed three `self.unit.status = ` lines from this version of the method. We'll handle replacing those shortly.
+We removed four `self.unit.status = ` lines from this version of the method. We'll handle replacing those shortly.
```
dwilding: This should switch back, now that we're keeping the "check port config" status

```suggestion
We removed three `self.unit.status = ` lines from this version of the method. We'll handle replacing those shortly.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding, line 93: (Unrelated to your changes) The dependency specifier in `pyproject.toml` in the example charm is missing `,<2`. Would you mind adding it while we're here?

examples/k8s-5-observe/tests/unit/test_charm.py
```diff
@@ -53,7 +64,7 @@ def test_pebble_layer():
     )
```
dwilding, line 67:
```suggestion
    # Check the workload version is set
    assert state_out.workload_version is not None

```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: (Not related to your changes) Fixing a mismatch between the tutorial and the example charm:

```suggestion
def test_deploy_cos(charm: pathlib.Path, cos: jubilant.Juju):
```

dwilding:
```suggestion
def test_integrate_loki(charm: pathlib.Path, juju: jubilant.Juju, cos: jubilant.Juju):
```

dwilding:
```suggestion
def test_loki_data(charm: pathlib.Path, cos: jubilant.Juju):
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -480,7 +480,13 @@ Let's write some integration tests to check that our application works correctly
 
 The existing integration tests use a model that is created by the `juju` fixture. We'll define a similar fixture that creates a separate model for COS Lite.
 
-First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest`, `pytest_jubilant`, and `requests`. Your imports should now look like this:
+First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest` and `pytest_jubilant`, which are already dependencies of your charm. Also import `requests`, which is a new dependency. Run the following command to add `requests` as a dependency:
+
+```text
+uv add requests
```
dwilding: We need to specify the group too (sorry, I didn't realise this earlier)

```suggestion
uv add --group integration requests
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -480,7 +480,13 @@ Let's write some integration tests to check that our application works correctly
 
 The existing integration tests use a model that is created by the `juju` fixture. We'll define a similar fixture that creates a separate model for COS Lite.
 
-First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest`, `pytest_jubilant`, and `requests`. Your imports should now look like this:
+First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest` and `pytest_jubilant`, which are already dependencies of your charm. Also import `requests`, which is a new dependency. Run the following command to add `requests` as a dependency:
```
dwilding:
```suggestion
First, in `tests/integration/test_charm.py`, import `json` and `time` from the standard library. Then import `pytest` and `pytest_jubilant`, which are already dependencies of the integration tests. Also import `requests`, which is a new dependency. Run the following command to add `requests` as a dependency:
```

dwilding (approved):
Very nice indeed! I cross-checked the details and found a couple of small adjustments needed.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
tromai: I'm more than happy to. 

It has been added.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
tromai: No problem at all. I completely missed this detail too :smile:

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -11,6 +11,7 @@ When you deploy a Kubernetes charm, the following things happen:
 1. The same Juju controller injects Pebble -- a lightweight, API-driven process supervisor -- into each workload container and overrides the container entrypoint so that Pebble starts when the container is ready.
 1. When the Kubernetes API reports that a workload container is ready, the Juju controller informs the charm that the instance of Pebble in that container is ready. At that point, the charm knows that it can start communicating with Pebble.
 1. Typically, at this point the charm will make calls to Pebble so that Pebble can configure and start the workload and begin operations.
+1. During operations, the charm may need to directly communicate with the workload using HTTP requests. This is possible because the charm container and workload container share the same pod.
```
james-garner-canonical: Not quite accurate since you can make HTTP requests between pods with the correct ceremony. Perhaps we mean to say that we can communicate using `localhost` ([docs](https://kubernetes.io/docs/concepts/workloads/pods/#pod-networking))?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -440,6 +521,16 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     }
     juju.deploy(charm, app=APP_NAME, resources=resources)
     juju.wait(jubilant.all_active)
+
+
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
+    version = juju.status().apps["fastapi-demo"].version
```
dwilding: One more thing I just thought of. `"fastapi-demo"` can be `APP_NAME` instead. Here and in all the example charm test files.

dwilding: One more thing I just thought of. `"fastapi-demo"` can be `APP_NAME` instead. Here and in all the example charm test files.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -83,6 +85,54 @@ resources:
     upstream-source: ghcr.io/canonical/api_demo_server:1.0.4
 ```
 
+### Write a helper module
+
+Your charm will interact with our workload application `api_demo_server`. It's a good idea to write a helper module that wraps `api_demo_server`. Charmcraft created `src/fastapi_demo.py` as a placeholder helper module.
```
james-garner-canonical: Why the name discrepancy? If the workload is `api_demo_server` then shouldn't that be the name we pass to `charmcraft init`?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +240,20 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
...
+
+In `src/charm.py`, append the following lines to the `_on_demo_server_pebble_ready` function:
+
+```python
+# Set the workload version of this charm.
+version = fastapi_demo.get_version(port=8000)
+self.unit.set_workload_version(version)
+```
+
+We get the workload version over port 8000 because `_get_pebble_layer` deploys the app on this port. Then `self.unit.set_workload_version` exposes the workload version to Juju.
```
james-garner-canonical: Minor: Is it worth factoring out the port number into a constant? e.g. global `_WORKLOAD_PORT`, referred to in both the Pebble layer and here. Perhaps not since the duplication goes away once we start pulling the workload from config.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -440,6 +521,16 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
...
     juju.wait(jubilant.all_active)
+
+
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
+    version = juju.status().apps["fastapi-demo"].version
+    # Ideally, the test should get the version directly from the workload application
+    # (for example, through an API call) and use that in this assertion.
+    # For simplicity, we hardcode the version here. We update the tutorial whenever we
+    # release a new version of api_demo_server.
```
james-garner-canonical: I don't know that this whole comment is needed here. Perhaps:
```suggestion
    # Hardcoded version for simplicity. Ideally we'd get the version directly from the workload.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -236,6 +236,16 @@ def _replan_workload(self) -> None:
...
+    except RuntimeError as version_e:
+        logger.error("Failed to get workload version: %s", version_e)
+        self.unit.status = ops.BlockedStatus(
+            "Failed to get version from server: check port config"
+        )
+        return
+    self.unit.set_workload_version(version)
 ```
 
 We removed three `self.unit.status = ` lines from this version of the method. We'll handle replacing those shortly.
```
james-garner-canonical, line 244: The following text for this code block is now kind of contradictory -- it's explaining that we removed `self.unit.status =` lines (presumably switching to the collect status pattern), but we're conspicuously leaving a `self.unit.state = ops.BlockedStatus` without further explanation.

If this is the right pattern, then I think it could use some explanation as to why it's not replaced.

But is it the right pattern? It looks like our collect status handler will happily clobber this blocked status with active status.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,26 @@ def _replan_workload(self) -> None:
...
+    try:
+        version = fastapi_demo.get_version(config.server_port)
+    except RuntimeError as version_e:
+        logger.error("Failed to get workload version: %s", version_e)
+        self.unit.status = ops.BlockedStatus(
+            "Failed to get version from server: check port config"
+        )
+        return
+    self.unit.status = ops.ActiveStatus()
+    self.unit.set_workload_version(version)
```
james-garner-canonical: Very minor, but perhaps `self.unit.status = ops.ActiveStatus()` should be the very last line?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
```diff
@@ -513,14 +519,14 @@ Add two test functions to `tests/integration/test_charm.py`:
 
 ```python
 @pytest.mark.juju_setup
-def test_deploy_cos(cos: jubilant.Juju):
+def test_deploy_cos(charm: pathlib.Path, cos: jubilant.Juju):
```
james-garner-canonical, line 522: Is this a driveby fix? Why do we need `charm` here (and below)?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -188,6 +240,20 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
...
+
+The workload version is available after the workload starts, which happens after Pebble reevaluates its plan. We'll use the `src/fastapi_demo.py` helper module for this step.
+
+In `src/charm.py`, append the following lines to the `_on_demo_server_pebble_ready` function:
+
+```python
+# Set the workload version of this charm.
+version = fastapi_demo.get_version(port=8000)
+self.unit.set_workload_version(version)
+```
```
james-garner-canonical: Minor: I guess we don't introduce error handling here, because we introduce setting statuses later. That makes sense, but I wonder if it's worth mentioning. Perhaps not, but WDYT about something like "For simplicitly, we'll assume there's no `URLError` for now. In the next chapter we'll introduce Juju statuses and error handling."

james-garner-canonical (commented):
This is looking good, it's nice to add this feature to the K8s tutorial (and more complicated to do so than in the machine tutorial both due to the container architecture and the tutorial structure).

I have a few questions and suggestions. @dwilding could you please take a look at my suggestions too in case your opinion differs.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
tromai: Yep, it's a driveby fix. See the original [discussion](https://github.com/canonical/operator/pull/2559#discussion_r3464830727)

james-garner-canonical: I'm OK with aligning with the examples as a driveby in this PR. But it seems to me that the tutorial is correct not to request the `charm` fixture in tests that don't need it, and we should update the examples instead. WDYT @dwilding?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -127,15 +127,26 @@ def _replan_workload(self) -> None:
...
         self.unit.status = ops.MaintenanceStatus("Waiting for Pebble in workload container")
+        return
+    try:
+        version = fastapi_demo.get_version(config.server_port)
+    except RuntimeError as version_e:
+        logger.error("Failed to get workload version: %s", version_e)
+        self.unit.status = ops.BlockedStatus(
+            "Failed to get version from server: check port config"
+        )
+        return
```
james-garner-canonical: If this isn't the first time we've run replan workload successfully (e.g. the container restarted), couldn't the workload version already be set at this point? The blocked status message in combination with the version field being populated in status could be confusing UX. Maybe we should unset the version in a case like this, e.g. `self.unit.set_workload_version("")`. I'm not sure though. @tonyandrewmeyer do you have a sense for what the best practice would be here?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
dwilding: We previously standardised the profiles, examples, and docs (e.g., [How to write integration tests for a charm](https://canonical.com/juju/docs/ops/latest/howto/write-integration-tests-for-a-charm/#test-a-relation)) to depend on the `charm` fixture even when it's not needed. That's so that all tests fail fast if the charm hasn't been packed. When I added integration tests to the COS chapter, I took that approach in the example charm, but I forgot to take the same approach in the chapter source.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: > But is it the right pattern? It looks like our collect status handler will happily clobber this blocked status with active status.

Is that true? I had assumed that the collect-status handler wouldn't override higher priority status with a lower priority one - but I actually don't know.

In tutorial text we say:

> We also want to clean up the code to remove the places where we're setting the status outside of this method, other than anywhere we're wanting a status to show up _during_ the event execution (such as `MaintenanceStatus`).

Which suggested to me that a maintenance status would survive collect-status. And so presumably blocked status would too.

I think it would be good to test this to confirm.

dwilding: Oh but I suppose the point is that maintenance status gets overridden after the work in the event execution has completed. Hmm

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite.md
james-garner-canonical: Ah I see, that makes sense, thanks for explaining.

examples/k8s-1-minimal/src/fastapi_demo.py
```diff
@@ -0,0 +1,44 @@
...
+    """Get the version of fastapi_demo that is running.
+
+    Args:
+        port: The port where fastapi_demo web server is listening.
+
+    Raises:
+        RuntimeError: If the server can't be reached, for example because of an invalid port.
+    """
+    try:
+        response = urllib.request.urlopen(f"http://0.0.0.0:{port}/version")
```
tromai:
```suggestion
        response = urllib.request.urlopen(f"http://localhost:{port}/version")
```

examples/k8s-1-minimal/src/fastapi_demo.py
```diff
@@ -0,0 +1,44 @@
...
+    except urllib.error.URLError as e:
```
tromai: Drop this try except

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
james-garner-canonical: Per discussion, there's nothing the user or the charm can do to to resolve this error, so we should not try/except, and just let the error bubble up instead. Ops will immediately put the charm in error status immediately, bypassing collect status entirely.

james-garner-canonical: In principle a more holistic charm might check the workload health like this whenever it runs. For simplicity for now it seems reasonable for us only to check in this event.

examples/k8s-2-configurable/src/charm.py
```diff
@@ -85,11 +87,20 @@ def _replan_workload(self) -> None:
             # service if required.
             self.container.replan()
             logger.info(f"Replanned with '{self.pebble_service_name}' service")
-
-            self.unit.status = ops.ActiveStatus()
         except (ops.pebble.APIError, ops.pebble.ConnectionError) as e:
             logger.info("Unable to connect to Pebble: %s", e)
             self.unit.status = ops.MaintenanceStatus("Waiting for Pebble in workload container")
+            return
```
tromai, line 93: From our daily sync. We decided to remove any error catching for `version = fastapi_demo.get_version(config.server_port)` and let all error bubble up. 

See rationale: https://github.com/canonical/operator/pull/2559#discussion_r3471410546

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tromai: Thank you for putting the summary here James. I will remove the try/except for `get_version` from this tutorial.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: You are right. We should simplify it in the tutorial to prevent confusion. I will update it.

I initially made this comment following the example here - https://canonical.com/juju/docs/ops/latest/howto/manage-the-workload-version/#write-integration-tests.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: Yep, you are right.

I will switch to "localhost" here. This also aligns with our decision on using `localhost` in the `get_version` helper function.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: This is a historical mismatch between the workload and charm names. I agree it's slightly confusing. If we want to change the name of the charm, a dedicated find-and-replace PR would be best I think.

tromai: I was referring to the application we deployed in the workload container. I picked `api_demo_server` because it's the name of the OCI image, and the [git repo](https://github.com/canonical/api_demo_server/).

I thought about `fastapi-demo` as defined in `charmcraft.yaml`. But I view it as the charm's name, and not the workload application.

I also think this name discrepancy can be confusing for readers. May I suggests renaming `api_demo_server` to "workload application" in my changes here? It is still clear I think, and leave room for cleaning up later.

dwilding: "Workload application" sounds good to me for now - good idea.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tromai: Yep, initially `self.unit.status = ops.ActiveStatus()` is the last line.

However, we decided that putting `self.unit.set_workload_version(version)` last would simplify the tutorial, and went with it. 

In chapter 1, section `Set the workload version`, we can simply write

``````
In `src/charm.py`, append the following lines to the `_on_demo_server_pebble_ready` function:

```python
# Set the workload version of this charm.
version = fastapi_demo.get_version(port=8000)
```
``````

And this code propagated into Chapter 2 as you see here, and I kept it like that.

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md (hunk 46 above)
Comment (tromai) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md]: We decided to let the errors bubble up, so we won't have error handling for this `get_version` function anymore.

However, I can add a paragraph saying "For simplicitly, we will let any exception bubble up and handled by Juju.", or something similar ?

Comment (dwilding) [reply docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md]: In the machine charm tutorial we have this comment. Maybe a similar one can work here?

```py
raise RuntimeError("tinyproxy was not running within the expected time")
# Raising a runtime error will put the charm into error status.
# The Juju logs will show the error message, to help you debug the error.
```

_Edit:_ Also this one, which is maybe closer to what we want here:

```py
apt.add_package("tinyproxy-bin", "1.11.1-3")
# If this call fails, the charm will go into error status. The Juju logs will show the error:
# charmlibs.apt.PackageError: Failed to install packages: tinyproxy-bin
```

tromai: Ah thanks. Those would work really well.

I will draft something following your examples.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: I also think it doesn't bring a big benefit. I originally used `port=8000` to keep this snippet tight in scope, and easy to grasp for readers.

So I would prefer to keep it.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
james-garner-canonical: It's unlikely to happen in practice, but it would be odd to blip to `active` before potentially going to `error` if `get_version` failed. Even though we'll ultimately recommend using the collect status pattern instead of assigning statuses directly, I think it's good to model best practices for setting statuses directly. The tutorial could read: ``In `src/charm.py`, add the following lines to the `_on_demo_server_pebble_ready` function before the final `self.unit.status = ops.ActiveStatus()`:``

examples/k8s-1-minimal/tests/integration/test_charm.py
```diff
@@ -39,3 +39,10 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     }
     juju.deploy(charm, app=APP_NAME, resources=resources)
     juju.wait(jubilant.all_active)
+
+
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
+    version = juju.status().apps[APP_NAME].version
+    # Hardcoded version for simplicity. Ideally we'd get the version directly from the workload.
+    assert version == "1.0.4"
```
james-garner-canonical: TBH I don't know if it would be ideal to get the version directly from the workload. The charm has opinions on the workload it's deployed with (with `upstream-source` being the source of truth for our tests currently), so perhaps a charm would want to test that they're deployed with the correct version too. How about:

```suggestion
    assert version == "1.0.4"  # Hardcoded for simplicity.
```

examples/k8s-1-minimal/tests/unit/test_charm.py
```diff
@@ -51,3 +62,5 @@ def test_pebble_layer():
         state_out.get_container(container.name).service_statuses["fastapi-service"]
         == ops.pebble.ServiceStatus.ACTIVE
     )
+    # Check the workload version is set
+    assert state_out.workload_version is not None
```
james-garner-canonical: After going to the trouble to hardcode a mock version for testing, shouldn't we assert that it roundtrips? Either like this, or with a constant:

```suggestion
    # Check the workload version is set:
    assert state_out.workload_version == "0.0.1"
```

james-garner-canonical (approved):
Thanks, this is a very nice improvement to the tutorial and examples. I've left a few minor suggestions.

cc @dwilding for a second pair of eyes on my suggestions.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
dwilding: I can't remember whether we deliberately wanted `self.unit.status = ops.ActiveStatus()` before setting the version, but James's recommendation sounds reasonable to me.

examples/k8s-1-minimal/tests/unit/test_charm.py
tromai: Oh yep. Good point. I will use the hardcoded constant here.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tromai: I traced the changes back to the beginning. The decision to have `self.unit.status = ops.ActiveStatus()` before setting the version were suggested by Ben from our 1:1.

I agree that keeping the tutorial with best practices is the right way. And James's suggestion is great. I will change it.

PR #2561, opened 2026-06-11
fix: avoid double-closing the framework in scenario teardown

tonyandrewmeyer:
In #2541, I added a defensive `framework.close()` to `scenario._ops_main_mock.Ops.destroy()`, on the basis that `Framework.close()` is idempotent. It is, but what I didn't consider is that someone might subclass it and it would not be true then. Tempo's widely-vendored `charm_tracing` charm lib replaces the instance's `close` with a wrapper that ends the root span, detaches the otel context, and resets single-use `ContextVar` tokens:

```python
# charm_tracing.py v0 (LIBPATCH 12), _setup_root_span_initializer
original_close = framework.close

@functools.wraps(original_close)
def wrap_close():
    span.end()
    opentelemetry.context.detach(span_token)
    tracer.reset(_tracer_token)   # ContextVar tokens are single-use
    ...
framework.close = wrap_close
```

`_Manager.run()` already closes the framework in its `finally`, so the new close in `destroy()` invokes `wrap_close` a second time and the suite fails with `RuntimeError: <Token ...> has already been used once`. Every `Context.run()` on a `@trace_charm`-decorated charm fails. Nothing the charm or test does opts out, since the decorator instruments unconditionally at charm init. The charm-construction-failure path (`_main.py`'s close-then-raise, also from #2541) double-closes the same way.

Minimal reproduction (only needs `charms/tempo_coordinator_k8s/v0/{charm_tracing,tracing}.py` on `PYTHONPATH`):

```python
import ops
from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm

@trace_charm(tracing_endpoint='tracing_endpoint')
class MyCharm(ops.CharmBase):
    @property
    def tracing_endpoint(self):
        return None

def test_charm_tracing_under_scenario():
    ctx = Context(MyCharm, meta={'name': 'repro'})
```

The fix the PR makes is to make close-state explicit instead of assuming idempotence of a monkeypatchable method:

- `Framework.close()` records `self._closed = True`.
- `Ops.destroy()` skips its defensive close when the framework is already closed.

Real-world check: with ops + ops-scenario installed from this branch, the repro above passes and self-signed-certificates-operator's unit suite returns to its released-version state (3 consecutive runs; its 2 remaining failures are the intentional #2378 tz-aware secret-expiry change, present on released 3.7.1 too).


testing/src/scenario/_ops_main_mock.py
```diff
@@ -271,10 +271,13 @@ def destroy(self):
...
-        # `mgr.run()`. Close defensively so the SQLite storage is released;
-        # framework.close() is idempotent.
+        # `mgr.run()`. Close defensively so the SQLite storage is released.
+        # Only close if it hasn't already happened: Framework.close() itself
+        # is idempotent, but charm libraries (for example tempo's
+        # charm_tracing) replace the instance's close with a wrapper that is
+        # not safe to invoke twice.
         framework = getattr(self, 'framework', None)
-        if framework is not None:
+        if framework is not None and not getattr(framework, '_closed', False):
```
james-garner-canonical: Why do we need to get `framework` with `getattr`? This `Ops` class inherits from `_Manager` and calls `super().__init__`, and `_Manager.__init__` unconditionally sets `self.framework`.

What stands out in this PR is that we call `getattr(framework, '_closed', False)` when we already know that `framework` isn't `None`. From a type checking perspective this should already be strictly unnecessary, since `framework` via `getattr` is typed as `Any | None`, so here we have `Any`. Because we use lockstep versioning for Ops and Scenario, `Framework` should always have a `_closed` attribute.

Suggestions:
1. Since we're expecting a `Framework` or `None`, write `framework: Framework | None = getattr(...)`.
2. If that `getattr` isn't actually necessary, just access `self.framework`.

tonyandrewmeyer: > Why do we need to get `framework` with `getattr`? This `Ops` class inherits from `_Manager` and calls `super().__init__`, and `_Manager.__init__` unconditionally sets `self.framework`.

Yeah, it's probably too cautious.

> Because we use lockstep versioning for Ops and Scenario, `Framework` should always have a `_closed` attribute.

Yes, fair point here too. The "provide a different framework class" doesn't really allow changing that and it's not something people are meant to be doing themselves anyway.

james-garner-canonical (approved):

tromai (approved):

PR #2562, opened 2026-06-11
test: hookcmds integration tests for secret access and port management

tonyandrewmeyer:
Follow-up to #2520. This PR adds three integration tests to `test/integration/test_hookcmds.py`:

- `test_secret_grant_revoke`. Exercises `secret_grant` and `secret_revoke` over the real cross-app relation.
- `test_ports_endpoint_scoped`. Opens TCP port 7766 scoped to the `peer` endpoint, asserts `opened_ports(endpoints=True)` returns `endpoints=['peer']`, then closes it.
- `test_relation_model_get`. This really exists for CMR, but doing that seems too much for these tests, so it just validates it's the right shape and consistent across use.

I also (had Claude) rewrite `test_juju_reboot_queues_reboot`. I wasn't paying enough last time - you can't reboot in K8s, and you can't reboot in an action (I definitely knew both of these already). I did the same kind of workaround as I've done previously: trigger it via config.

Also a few extra fixes for problems that running against all of 3/stable, 4.0/stable, 4.0/edge, 4.1/edge × machine/k8s) showed up. Some of this was that I failed to run the whole matrix after changes from refiew: I'll make sure to do that this time, sorry!

- `test-file` resource declaration needed `filename:` for juju 3/stable and newer to accept the deploy.
- `data` storage needed `multiple: { range: 1-2 }` so `test_storage_add` could attach a second instance.
- `test_setup` deploys with `trust=True` so `credential_get` works.
- `test_network_get_returns_addresses` xfails the Juju 4.0/k8s `network-get` placeholder issue (`juju/juju#22616`) via the new shape; the `interface-name` non-empty assertion is gated on machine substrates.
- `_xfail_juju4_commit_bug` helper centralises the xfail-on-error pattern for the Juju 4 uniter commit-phase bugs (`juju/juju#22523`, `juju/juju#22524`), widened to catch `TimeoutError` alongside `TaskError` since the symptom on 4.0/k8s is sometimes a driver timeout.

Run on my fork: [`tonyandrewmeyer/operator` run 27414403142](https://github.com/tonyandrewmeyer/operator/actions/runs/27414403142). All eight hookcmds cells green:

| Cell | Result |
|---|---|
| `hookcmds (machine, 3/stable)` | ✅ 8m37s |
| `hookcmds (k8s, 3/stable)` | ✅ 6m13s |
| `hookcmds (machine, 4.0/stable)` | ✅ 7m38s |
| `hookcmds (k8s, 4.0/stable)` | ✅ 10m03s |
| `hookcmds (machine, 4.0/edge)` | ✅ 7m59s |
| `hookcmds (k8s, 4.0/edge)` | ✅ 9m14s |
| `hookcmds (machine, 4.1/edge)` | ✅ 7m31s |
| `hookcmds (k8s, 4.1/edge)` | ✅ 9m28s |

The `secrets (4.0/stable)` failure is the pre-existing flaky `test_secrets` cell from main, not anything this PR touches.


test/integration/test_hookcmds.py
```diff
@@ -473,29 +506,39 @@ def test_resource_get(juju: jubilant.Juju, any_unit: str):
     assert pathlib.Path(path).name == 'test-file.txt'
 
 
-# Secret grant / revoke (secret_grant / secret_revoke)
+# Secrets (secret_grant / secret_revoke) — Follow-up 3
```
tonyandrewmeyer:
```suggestion
# Secrets (secret_grant / secret_revoke)
```

test/integration/test_hookcmds.py
```diff
@@ -516,6 +559,30 @@ def test_storage_add(juju: jubilant.Juju, any_unit: str):
     assert count_after == count_before + 1
 
 
+# Ports — endpoint-scoped variant (Follow-up 4)
```
tonyandrewmeyer:
```suggestion
# Ports — endpoint-scoped variant
```

tonyandrewmeyer:
@dwilding and @tromai sorry to hit you up again immediately following the other one, but I assume bringing over that context is the most efficient thing to do. This one should be the end of this work!

test/charms/test_hookcmds/.gitignore
```diff
@@ -0,0 +1,3 @@
+uv.lock
+*.tar.gz
```
dwilding, line 2: What produces .tar.gz files?

test/integration/test_hookcmds.py
```diff
@@ -52,39 +57,76 @@ def _juju_major(juju: jubilant.Juju) -> int:
...
+    `jubilant.TaskError` (action returned non-zero) or a `TimeoutError`
+    (driver gave up waiting) depending on the failure path.
+
+    Note: a previous version of this helper attempted strict-style
+    semantics (pytest.fail on Juju 4 success, mimicking `xfail(strict=True)`)
+    so we'd notice when to drop the guard. That doesn't work in a multi-
+    channel matrix — when the upstream fix lands in 4.0/edge or 4.1/edge
+    but 4.0/stable still ships the bug, the guard must stay. The cells
+    where the bug is fixed are the signal; nothing to do at test-author
+    time.
```
dwilding: I think this comment can be reduced. Something like:

```suggestion
    Note: We can't strictly require failure; our CI matrix might run
    against a fixed 4.0/edge and a broken 4.0/stable (for example).
```

test/integration/test_hookcmds.py
```diff
@@ -516,16 +586,73 @@ def test_storage_add(juju: jubilant.Juju, any_unit: str):
...
+    # recover after the reboot.
+    juju.config('test-hookcmds', {'reboot-trigger': 'reboot-please'})
     juju.wait(jubilant.all_active)
 
+    task = juju.run(any_unit, 'test-reboot-marker')
+    assert task.success
+    assert task.results['marker-exists'] == 'true'
+    uptime_after = float(task.results['uptime-seconds'])
+    # If a reboot actually happened the container's uptime resets — the new
+    # value must be smaller than the value we captured before the trigger.
```
dwilding: `test_juju_reboot_queues_reboot` strategically placed at the end of the file. So, yes, I'm willing to believe this 😀

test/integration/test_hookcmds.py
```diff
@@ -516,16 +586,73 @@ def test_storage_add(juju: jubilant.Juju, any_unit: str):
...
+    The charm's config-changed handler watches for
+    `reboot-trigger=reboot-please` and, on the first such config-changed,
+    writes a marker file and calls hookcmds.juju_reboot(). The marker
+    prevents the post-reboot config-changed re-run from looping. The
+    test then verifies the marker exists via a read-only action.
+    """
+    # k8s pods can't reboot themselves and the hookcmd surfaces as a
+    # NotImplementedError; only exercise on machine substrates.
+    if _is_caas(juju):
+        pytest.skip('juju-reboot is not supported on Kubernetes / CAAS substrates')
```
dwilding: I don't think there's any need to mention "CAAS" in the test output, as we hardly use that term anywhere. I'm wondering about even renaming `_is_caas` to `_is_k8s`, but I don't think it matters too much.

```suggestion
        pytest.skip('juju-reboot is not supported on Kubernetes substrates')
```

test/integration/test_hookcmds.py
```diff
@@ -516,16 +586,73 @@ def test_storage_add(juju: jubilant.Juju, any_unit: str):
...
 
 
+# Ports — endpoint-scoped variant
+
+
+def test_ports_endpoint_scoped(juju: jubilant.Juju, any_unit: str):
+    """open_port with endpoints='peer' appears scoped in opened_ports(endpoints=True)."""
+    # On Juju 4.0/k8s this action times out on the driver side, matching the
+    # uniter commit-phase regression pattern. May be a separate upstream bug
+    # from the secret ones; xfail until pinned down.
```
dwilding, line 603: Is this (potential) timeout bug tracked somewhere for further investigation?

test/integration/test_hookcmds.py
```diff
@@ -452,14 +505,15 @@ def test_credential_get(juju: jubilant.Juju, any_unit: str):
...
     assert task.success
-    uuid = task.results['uuid']
-    assert uuid  # non-empty
-    # For a peer relation the remote model is the same model.
-    model_info = juju.show_model()
-    assert uuid == model_info.model_uuid
+    # For a peer relation the relation isn't cross-model, so the server-side
+    # implementation (apiserver/facades/agent/uniter/uniter.go) returns the
+    # local model UUID — same as JUJU_MODEL_UUID inside the hook. Validated
+    # live against Juju 3.6.23 and 4.0.5 LXD.
```
dwilding: Is this commentary from Claude? I think we can simplify it:

```suggestion
    # Peer relations aren't cross-model, so the server-side implementation
    # returns the local model UUID — same as JUJU_MODEL_UUID inside the hook.
```

test/charms/test_hookcmds/src/charm.py
```diff
@@ -343,23 +354,46 @@ def _on_test_credential_get(self, event: ops.ActionEvent):
         cloud = hookcmds.credential_get()
         event.set_results({'cloud-type': cloud.type, 'cloud-name': cloud.name})
 
-    # Reboot
+    # Reboot — driven by config-changed because juju forbids juju-reboot
+    # from an action context (see jujuc/reboot.go).
```
dwilding: I don't think we need to reference the Juju code:

```suggestion
    # Reboot — driven by config-changed because juju forbids juju-reboot
    # from an action context.
```

test/integration/test_hookcmds.py
```diff
@@ -516,16 +586,73 @@ def test_storage_add(juju: jubilant.Juju, any_unit: str):
...
 # Reboot (juju_reboot) — placed last because it reboots the unit
 
 
 def test_juju_reboot_queues_reboot(juju: jubilant.Juju, any_unit: str):
-    """juju_reboot(now=False) queues a reboot; the unit recovers to active."""
-    task = juju.run(any_unit, 'test-juju-reboot')
+    """juju_reboot(now=False) queues a reboot; the unit recovers to active.
+
+    juju explicitly forbids juju-reboot inside an action context
+    (jujuc/reboot.go), so the trigger has to come from a regular hook.
```
dwilding:
```suggestion
    juju explicitly forbids juju-reboot inside an action context,
    so the trigger has to come from a regular hook.
```

dwilding (approved):

test/charms/test_hookcmds/.gitignore
tonyandrewmeyer: The integration conftest runs `uv build --sdist` into the charm directory so charmcraft picks up local ops (and ops_tracing) as ops.tar.gz / ops_tracing.tar.gz. Only an issue if you're running this locally but seems cleanest to ignore it.

test/integration/test_hookcmds.py
tonyandrewmeyer: Half Claude, half me.

test/integration/test_hookcmds.py
tonyandrewmeyer: I *think* it has the same underlying cause and will get fixed at the same time as the other one. To be honest, partly I got tired digging into Juju bugs deep enough to write a decent issue, even with agent help, and didn't want to do another one when I thought it was likely the same.

When the other one is fixed, if this doesn't start passing, I'll do the investigation then. The tracking is this comment plus my subscriptions to the Juju issues.

test/integration/test_hookcmds.py
tonyandrewmeyer: It was indeed strategically placed, but on reflection, I should just be using boot time, not uptime.

test/charms/test_hookcmds/src/charm.py
```diff
@@ -343,23 +355,49 @@ def _on_test_credential_get(self, event: ops.ActionEvent):
         cloud = hookcmds.credential_get()
         event.set_results({'cloud-type': cloud.type, 'cloud-name': cloud.name})
 
-    # Reboot
+    # Reboot — driven by config-changed because juju forbids juju-reboot
```
tromai, line 358: > juju forbids juju-reboot

That's interesting. Could you point me to the place where I can read more about this Juju behavior?

test/charms/test_hookcmds/src/charm.py
```diff
@@ -343,23 +355,49 @@ def _on_test_credential_get(self, event: ops.ActionEvent):
...
+        if self.config.get('reboot-trigger') != 'reboot-please':
+            return
+        if _REBOOT_MARKER.exists():
+            # Already triggered this deployment; don't reboot again.
+            return
+        _REBOOT_MARKER.write_text('triggered')
         hookcmds.juju_reboot(now=False)
 
+    def _on_test_reboot_marker(self, event: ops.ActionEvent):
+        # Compute the boot epoch as `now - uptime`. /proc/stat's `btime` is
```
tromai, line 371: Just to double check: This test assumes that it will be run within a LXD container (I'm not sure if it need to / will be run elsewhere in our test suite) ?

tromai (approved):
Thanks for the fix.

I think investigating Juju issues mentioned in this PR is a lot of work. So I really appreciate it.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: There's [our docs](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.Unit.reboot) and [the juju docs](https://canonical.com/juju/docs/juju-cli/latest/reference/hook-command/list-of-hook-commands/juju-reboot/), probably not much else.

test/charms/test_hookcmds/src/charm.py
tonyandrewmeyer: Yes, [the actual test skips if it's k8s](https://github.com/tonyandrewmeyer/operator/blob/d5b4512bbb2be96e79a55b3aa81bd6b8cbcc0c7f/test/integration/test_hookcmds.py#L637). If this code in the charm is called on k8s it won't do any harm, just write a pretty meaningless marker.

tromai: Ah, cool. I missed the part that it's skipped for k8s. Thanks!

test/charms/test_hookcmds/src/charm.py
tromai: Ah great. Thank you !

PR #2564, opened 2026-06-11
docs: replace CoC with link to Ubuntu Code of Conduct

tonyandrewmeyer:
The current `CODE_OF_CONDUCT.md` is an old Contributor Covenant 1.4 template that points reports at `charmcrafters@lists.launchpad.net`, which I think doesn't exist or will soon not exist, and also does not go to anyone in Charm Tech as far as I am aware.

This PR replaces it with a short pointer to the [Ubuntu Code of Conduct](https://ubuntu.com/community/ethos/code-of-conduct), which:

- is used by two other Charm Tech projects (`jubilant`, `pebble`), and
- has its own reporting and enforcement path through the Community Council, so we don't have to invent a contact for this project.

There is no behaviour change for contributors. It's just a clearer, project-agnostic pointer.

It's now an extremely small file. We could put this into CONTRIBUTING.md instead, but I think there's value in having it prominent and visible. We could tuck it away in `.github` or `docs/` and GitHub would still show it in the main page tabs. It doesn't feel like a GitHub file, so that doesn't seem 100% correct to me, and I'm not sure it should be in width our published docs. So I've left it in the same place.


james-garner-canonical (approved):
I like the separate file so it shows up nicely in the Github UI for visitors.

dwilding (approved):

PR #2567, opened 2026-06-12
fix: take Pebble defaults into consideration when consistency checking Checks

### LINKED ISSUE #2566: `InconsistentScenarioError` when Pebble check omits `threshold` (relies on default)
When a Pebble layer defines a pebble notice without an explicit `threshold`, relying on Pebble's default value of `3`, ops-scenario raises `InconsistentScenarioError` after the layer is applied, that is, a container 'my-container' has a check 'my-check' with a different 'threshold' (3) than the plan (None).

This forces us to specify `"threshold": 3` explicitly in their plan, even when the intent is to use the default.

## To reproduce

```python
import ops
from ops import testing

# charm.py: add a Pebble layer with no explicit threshold (relying on Pebble default of 3)
pebble_layer = {
    "services": {
        "my-service": {
            "command": "/bin/my-service",
            "startup": "enabled",
            "override": "replace",
        }
    },
    "checks": {
        "my-check": {
            "override": "replace",
            "exec": {"command": "<some command>"},
        }
    },
}

# test: running any subsequent event after the layer is applied raises:
# InconsistentScenarioError: container 'my-container' has a check 'my-check'
# with a different 'threshold' (3) than the plan (None).
context = testing.Context(ops.CharmBase, {"name": "foo", "containers": {"c": {}}})
container = testing.Container("c", can_connect=True, layers={"base": ops.pebble.Layer(pebble_layer)})
state = testing.State(containers={container})
with context(context.on.pebble_ready(container), state) as mgr:
    mgr.charm.unit.get_container("c").replan()
    state_out = mgr.run()
context.run(context.on.update_status(), state_out)  # <-- raises here
```

```
$ uvx --with ops[testing] python 2566.py
  File "/tmp/2566.py", line 30, in <module>
    context.run(context.on.update_status(), state_out)  # <-- raises here
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [...]
  File "/home/tameyer/.cache/uv/archive-v0/n9RiYsXEkVxCjJUocLiBt/lib/python3.14/site-packages/scenario/_consistency_checker.py", line 117, in check_consistency
    raise InconsistentScenarioError(
        f'Inconsistent scenario. The following errors were found: {err_fmt}',
    )
scenario.errors.InconsistentScenarioError: Inconsistent scenario. The following errors were found: container 'c' has a check 'my-check' with a different 'threshold' (3) than the plan (None).
```

## Workaround

Explicitly specify "threshold": 3 in the Pebble layer even though it is the default value.

tonyandrewmeyer:
When Pebble returns information about a check, it copies the level, startup, and threshold into the response from the computed plan. If none of the layers have provided values for these, the plan has defaults (unset, enabled, and 3, respectively).

When we check a `CheckInfo` for consistency in Scenario (meaning that the values should be ones Pebble could sensibly return, given the rest of the state), we need to take that default value into account, as Pebble would. We don't actually model checks going up and down currently, so it's really only the consistency that we are concerned with.

This PR fixes the consistency check so that it will fall back to the Pebble defaults when needed. It repeats the defaults, because the existing copy is buried in the mock Pebble client in Harness, and the source of truth is in the Pebble code.

Fixes #2566


testing/tests/test_consistency_checker.py
```diff
@@ -239,6 +239,27 @@ def test_checkinfo_matches_layer(check: CheckInfo, consistent: bool):
     )
 
 
+def test_checkinfo_matches_layer_with_default_threshold():
```
dwilding: Should this be renamed? I don't see that it's specific to threshold

```suggestion
def test_checkinfo_matches_layer_with_defaults():
```

dwilding (approved):

testing/src/scenario/_consistency_checker.py
```diff
@@ -708,8 +710,23 @@ def check_containers_consistency(
...
             plan_check = plan.checks[check.name]
-            for attr in ('level', 'startup', 'threshold'):
-                if getattr(check, attr) != getattr(plan_check, attr):
+            # Pebble fills in defaults for attributes the plan does not set,
+            # so treat the plan's unset value (None or an UNSET enum member) as
+            # equivalent to the Pebble default for that attribute, and treat
+            # an absent value on the check info as also matching the default.
+            unset_defaults = {
+                'level': ({None, pebble.CheckLevel.UNSET}, pebble.CheckLevel.UNSET),
+                'startup': ({None, pebble.CheckStartup.UNSET}, pebble.CheckStartup.ENABLED),
```
james-garner-canonical: Pebble seems to technically default to `UNSET` -- in `Check.__init__`: `self.startup = CheckStartup(dct.get('startup', ''))`. Though apparently they're functionally equivalent ...
```py
class CheckStartup(enum.Enum):
    """Enum of check startup options."""

    UNSET = ''  # Note that this is treated as ENABLED.
    ENABLED = 'enabled'
    DISABLED = 'disabled'
```

Interestingly for this change though, if we use `UNSET` as the default, then our `_normalise` logic becomes `is not None`, and the need for the helper and the set of 'unset' values seems to go away.

testing/src/scenario/_consistency_checker.py
```diff
@@ -708,8 +710,23 @@ def check_containers_consistency(
...
+                'threshold': ({None}, 3),
+            }
+
+            def _normalise(value: Any, unset: set[Any], default: Any) -> Any:
+                return default if value in unset else value
+
+            for attr, (unset, default) in unset_defaults.items():
+                check_value = _normalise(getattr(check, attr), unset, default)
+                plan_value = _normalise(getattr(plan_check, attr), unset, default)
```
james-garner-canonical: Do we need to explicitly normalise the plan value, or is this actually a no-op (or even counterproductive) since `State.containers.plan` ultimately is populated by calling the real `pebble.Plan`?

I'm a little unsure here because of the private-looking `Container._base_plan` attribute -- if users never pass this, then our defaults all come from the real `ops.pebble.Plan` directly. If they do pass it, then that's funky, but maybe whatever they pass in should be the ground truth for comparison rather than attempting to normalise it.

testing/src/scenario/_consistency_checker.py
```diff
@@ -708,8 +710,23 @@ def check_containers_consistency(
                 )
                 continue
             plan_check = plan.checks[check.name]
-            for attr in ('level', 'startup', 'threshold'):
-                if getattr(check, attr) != getattr(plan_check, attr):
```
james-garner-canonical, line 712: Maybe we should take this opportunity to unroll this loop and get type checking coverage for this code.

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: For reference:

Pebble:
  - Combined plan defaults: Threshold → 3 (in plan.go). Level and Startup are not defaulted in the combined plan, they stay as UnsetLevel/CheckStartupUnknown ("", in the Go way of doing things).
  - /v1/checks CheckInfo response (checkstate/manager.go): Startup is normalised server-side: Unknown → Enabled. Level passes through unchanged. Threshold passes through (so always 3 from the combined plan).

(I wish this worked differently, so startup ended up as enabled in the combined plan, even though it would be unset in the layers so combining works. Then the plan would reflect better what happens, and it'd be more consistent with threshold. But it can't be changed now.)

ops.pebble:
  - Check: startup = CheckStartup(dct.get('startup', '')) → UNSET. threshold = dct.get('threshold') → None when not in layer. level → UNSET.
  - CheckInfo.from_dict: startup = CheckStartup(d.get('startup', 'enabled')). Defaults to ENABLED if the key is missing (so it can never be UNSET from a real Pebble response, since Pebble sends enabled/disabled; and even if the key were missing, ops defaults to ENABLED). level → UNSET if missing. threshold is a required key.

scenario:

 - CheckInfo has different defaults than real Pebble:
   - level default: None (not CheckLevel.UNSET)
   - startup default: CheckStartup.ENABLED (not UNSET)
   - threshold default: 3 (int, never None)

(I think I did a poor job when I originally added this. It would probably make more sense to mirror Pebble more closely. But also not really changeable now.)

Harness:

 - passes check.level/check.startup/check.threshold (None→3) straight through to pebble.CheckInfo.

tonyandrewmeyer: I'm good unrolling the loop as per another thread. Putting the logic there probably makes it clear enough?

testing/src/scenario/_consistency_checker.py
tonyandrewmeyer: The thing is that Container.plan does build a real pebble.Plan from _base_plan, but _render_checks then writes the layer's pebble.Check objects straight into plan.checks[name], bypassing Pebble's combine/defaulting logic for checks. So plan.checks[name].threshold stays None, and .startup/.level stay UNSET, whenever the layer didn't set them. That's what motivated the threshold default. So the plan-side normalisation is needed.

Possibly the way Container.plan works could be changed, the various render methods are mostly taken from pebble at the moment, but I'm not sure I want to do that big a change.

test/test_testing.py
```diff
@@ -7136,6 +7137,21 @@ def test_get_cloud_spec_without_set_error(self, request: pytest.FixtureRequest):
...
+        Harness does not call setup_root_logging, so the first call to
+        Container.stop_checks in a worker emits a RuntimeWarning. Because
+        _get_juju_log_and_app_id is functools.cached, subsequent calls in
+        the same xdist worker would otherwise silently reuse whatever the
+        first test produced, leading to race-y behaviour where pytest.warns
+        sometimes passes and sometimes fails.
+        """
+        _get_juju_log_and_app_id.cache_clear()
+        yield
+        _get_juju_log_and_app_id.cache_clear()
```
james-garner-canonical: Clearing again afterwards seems like it might just hide other tests forgetting to request this fixture when they really should.

testing/src/scenario/_consistency_checker.py
```diff
@@ -708,13 +710,41 @@ def check_containers_consistency(
...
-                        f'different {attr!r} ({getattr(check, attr)}) '
-                        f'than the plan ({getattr(plan_check, attr)}).',
-                    )
+            # Pebble fills in defaults for attributes the plan does not set, so
+            # for each attribute treat the "unset" representation on either
+            # side as equivalent to whatever Pebble would effectively report.
+            # The scenario CheckInfo uses None for an absent level, while the
+            # plan uses CheckLevel.UNSET, so those two also need to compare
+            # equal.
+            check_level = check.level if check.level is not None else pebble.CheckLevel.UNSET
```
james-garner-canonical: Thanks, I like the unrolled form. WDYT about localising comments per attribute?
```suggestion
            # Scenario allows None for unset, Pebble defaults to UNSET.
            check_level = check.level if check.level is not None else pebble.CheckLevel.UNSET
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -708,13 +710,41 @@ def check_containers_consistency(
...
+            if check_level != plan_check.level:
+                errors.append(
+                    f'container {container.name!r} has a check {check.name!r} with a '
+                    f"different 'level' ({check.level}) than the plan ({plan_check.level}).",
+                )
+            check_startup = (
```
james-garner-canonical, line 721:
```suggestion
            # Scenario defaults to None, Pebble defaults to UNSET and collapses to ENABLED.
            check_startup = (
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -708,13 +710,41 @@ def check_containers_consistency(
...
+                pebble.CheckStartup.ENABLED
+                if plan_check.startup == pebble.CheckStartup.UNSET
+                else plan_check.startup
+            )
+            if check_startup != plan_startup:
+                errors.append(
+                    f'container {container.name!r} has a check {check.name!r} with a '
+                    f"different 'startup' ({check.startup}) than the plan ({plan_check.startup}).",
+                )
+            check_threshold = 3 if check.threshold is None else check.threshold
```
james-garner-canonical, line 737:
```suggestion
            # Scenario and Pebble allow None, collapsing to 3.
            check_threshold = 3 if check.threshold is None else check.threshold
```

james-garner-canonical (approved):
Looks good, thanks!

test/test_testing.py
tonyandrewmeyer: Yeah, fair, I think I was too aggressively copying from a similar fix I did twice elsewhere. One of those should change too, but the other (the first one I think) is a more general cleanup, so is right.

PR #2569, opened 2026-06-12
docs: do not imply that action names can contain underscores

tonyandrewmeyer:
Juju action names cannot contain underscores. The [regular expression for names](https://github.com/juju/charm/blob/v12/actions.go#L20) allows letters, digits, and dashes only.

This can also be trivially verified by creating a charm with an action containing an underscore, packing it (which does not fail) and attempting to deploy:

```
$ juju deploy ./actionname_amd64.charm
ERROR attempting to deploy "./actionname_amd64.charm": bad action name run_me
```

This PR changes two examples in our docs that implies otherwise, removes a no-op normalisation in Scenario that implies otherwise, and adds a validation check for Scenario.


testing/src/scenario/state.py
```diff
@@ -2530,3 +2502,12 @@ def test_backup_action():
 
     Every action invocation is automatically assigned a new one. Override in
     the rare cases where a specific ID is required."""
+
+    def __post_init__(self):
+        # Juju action names cannot contain underscores; they use dashes,
+        # matching the name as declared in actions.yaml.
+        if '_' in self.name:
```
tonyandrewmeyer, line 2537: We could have the full regex here. Maybe that would be better?

dwilding: I don't have a strong opinion. Defer to Harry

dwilding (approved):

tonyandrewmeyer:
Sorry for the noise here. I'm going to take this opportunity to do the bigger Scenario underscore/dash cleanup I have wanted for quite a while. #2570 will have this but more.

PR #26, opened 2019-10-17
Implement comparison operators for stored types

dshcherb:
* dict comparison works based on the abstract implementation that
  compares items;
* for sets, `__eq__` has an abstract implementation based on `__le__`
  and https://docs.python.org/3/library/collections.abc.html documents
  that `__le__` and `__ge__` both need to be implemented to make other
  operations work. Lib/_collections_abc.py in cpython reveals that they
  `__le__` and `__ge__` do not depend on each other and so need to be
  reimplemented separately;
* for lists, besides `__eq__`, there are also other comparison
  operations that need to be implemented for lexicographical comparison.
  https://docs.python.org/3/tutorial/datastructures.html#comparing-sequences-and-other-types

PR #2570, opened 2026-06-12
fix: align Juju naming rules with testing class rules

### LINKED ISSUE #2565: pebble_check_failed/pebble_check_recovered events always fail consistency check for containers with dashes in their names?
When testing `pebble_check_failed` or `pebble_check_recovered` events for a container whose name contains dashes (e.g. `temporal-worker`), the consistency checker unconditionally raises `InconsistentScenarioError` even when the `check_infos` are correctly set on the container in the state.

I believe that n `_consistency_checker.py`, `all_checks` is built using raw container names `all_checks = {(c.name, check.name) for c in state.containers for check in c.check_infos}`, but `evt_container_name` (a bit below that) is derived from event.name, which has been normalised.

# To reproduce

In a scenario test case:

```python
check_info = ops.testing.CheckInfo("my-check", failures=3, status=ops.pebble.CheckStatus.DOWN)
container = dataclasses.replace(
      state.get_container("temporal-worker"),
      check_infos={check_info},
)
state = dataclasses.replace(state, containers=[container])
context.run(context.on.pebble_check_failed(container, check_info), state) # <-- will raise InconsistentScenarioError
```

tonyandrewmeyer:
I have long been irritated by the messy way Scenario handled names, particularly with regards to underscores and dashes. #2565 has provided me with the excuse to do a full clean-up of how this is handled, making sure that everything matches real Juju properly, and avoiding inefficient (and sometimes wrong) round-tripping.

This PR aligns Scenario's simulated dispatch with how real Juju invokes hooks, and rejects entity names that Juju itself would reject, so tests can no longer pass against a charm that Juju would refuse to deploy or run.

I verified all of this against the juju/names and juju/charm sources for 3.x and 4.x, and against a live Juju 3.6.23 k8s model.

### Dispatch environment and event names

#2552 made `JUJU_HOOK_NAME` preserve the entity-name spelling by carrying two forms of the event path (a normalised string plus an original_prefix). That is now collapsed to one representation: `_EventPath.name` is the Python-attribute event kind (what ops calls the event, and what its handle paths use), and `_EventPath.prefix` is the verbatim entity name from which the Juju hook name is built.

Other places where the simulated environment diverged from real Juju:

* Actions are dispatched as `actions/<name>` with no `hooks/` prefix and no `-action` suffix; `JUJU_ACTION_NAME` keeps the verbatim name, and `JUJU_ACTION_TAG` is now set.
* `JUJU_RELATION_ID` is `<endpoint>:<id>`, not just the ID.
* `State.get_relations` matches the endpoint verbatim rather than treating `foo-bar` and `foo_bar` as the same endpoint (Juju allows both spellings, as distinct endpoints).

The consistency checker now compares the event's entity-name prefix to the entity name exactly, rather than a normalised `startswith` check (which also wrongly accepted, for example, a `foobar` event for a `foo` container).

### Name validation

Scenario accepted any string as a container, storage, relation endpoint, or action name. Names are now validated at construction time, raising `StateValidationError`, using the same rules as Juju:

* relation endpoints: lowercase alphanumeric runs separated by single hyphens or underscores, starting with a letter (`RelationSnippet`). Underscores are valid here, and Juju produces hook names such as `peer_under-relation-created` for them.
* storage: hyphen-separated lowercase alphanumeric, starting with a letter, each part containing at least one letter (`StorageNameSnippet`). `juju deploy` fails for an underscored storage name.
* actions: lowercase alphanumeric and hyphens, starting and ending alphanumeric (`actionNameRule`). `juju deploy` fails with "bad action name" for an underscored action name.
* containers: Juju passes the name verbatim to Kubernetes, so the RFC 1123 DNS-label rule applies (no underscores, max 63 characters); the pod is never created for an underscored container name.

### Documentation fixes

* The `manage-actions` how-to used `ctx.on.action('do_backup')`, a name Juju would reject.
* `ops.CharmEvents` now documents how dashed metadata names map to the prefixed event attributes (dashes become underscores) versus the indexed `self.on[<name>]` form (verbatim).
* The `_Action` docstring example used `do_backup`, which Juju would reject; corrected.

### Drive-bys

Checks that were previously unreachable now run: the workload-event "container is in the state" error and the cannot-connect warning were nested inside the name-mismatch branch, so they only fired when the event name was already wrong; they now run unconditionally, like the equivalent relation and storage checks.

The "no action instance" error message used to say "workload event" and "container", presumably a copy and paste error; it now refers to actions.

Fixes #2565.


testing/tests/test_e2e/test_deferred.py
```diff
@@ -164,6 +164,53 @@ def test_deferred_workload_event(mycharm: type[ops.CharmBase]):
     assert isinstance(start, ops.StartEvent)
 
 
+def test_deferred_workload_event_dashed_container_name():
+    """A deferred event for a dashed container name must be re-emitted.
+
+    Regression test: the deferred handle path must use the Python-attribute
+    form of the event name (``workload_a_pebble_ready``), not the verbatim
+    container name, or ops silently drops the notice (gh-2565).
```
tonyandrewmeyer:
```suggestion
    container name, or ops silently drops the notice.
```

testing/tests/test_e2e/test_state.py
```diff
@@ -317,6 +318,71 @@ def test_container_positional_arguments():
...
+    assert Container(name).name == name
+
+
+@pytest.mark.parametrize(
+    'name',
+    ('workload_a', 'Workload', '-workload', 'workload-', '', 'wörkload', 'a' * 64),
+)
+def test_container_name_invalid(name: str):
+    # Juju passes container names verbatim to Kubernetes, which requires
+    # RFC 1123 DNS labels.
```
tonyandrewmeyer:
```suggestion
```

testing/tests/test_e2e/test_state.py
```diff
@@ -317,6 +318,71 @@ def test_container_positional_arguments():
...
+
+
+@pytest.mark.parametrize('name', ('store', 'store-a', 's', 'store-a1', 's0'))
+def test_storage_name_valid(name: str):
+    assert Storage(name).name == name
+
+
+@pytest.mark.parametrize(
+    'name',
+    ('store_a', 'Store', '0store', '-store', 'store-', 'store-1', ''),
```
tonyandrewmeyer, line 342:
```suggestion
    'name', ('store_a', 'Store', '0store', '-store', 'store-', 'store-1', ''),
```

testing/tests/test_e2e/test_state.py
```diff
@@ -317,6 +318,71 @@ def test_container_positional_arguments():
...
+)
+def test_storage_name_invalid(name: str):
+    # Juju storage names cannot contain underscores, must start with a letter,
+    # and each hyphen-separated part must contain at least one letter.
```
tonyandrewmeyer:
```suggestion
```

testing/tests/test_e2e/test_state.py
```diff
@@ -317,6 +318,71 @@ def test_container_positional_arguments():
...
+        (PeerRelation, {}),
+        (SubordinateRelation, {}),
+    ],
+)
+@pytest.mark.parametrize('endpoint', ('0db', 'DB', '-db', 'db-', 'db--x', 'db__x', ''))
+def test_relation_endpoint_name_invalid(
+    klass: type[RelationBase], kwargs: dict[str, Any], endpoint: str
+):
+    # Juju relation endpoint names may contain underscores or hyphens (singly,
+    # between alphanumeric runs), and must start with a letter.
```
tonyandrewmeyer:
```suggestion
```

testing/tests/test_consistency_checker.py
```diff
@@ -104,23 +106,34 @@ def test_check_info_container_matches_event():
...
         # The check is in the wrong container.
+        container = Container('bar')
         assert_inconsistent(
-            State(containers={Container('foo', check_infos={check}), Container('bar')}),
-            _Event(event, container=Container('bar'), check_info=check),
+            State(containers={Container('foo', check_infos={check}), container}),
+            _Event(event, container=container, check_info=check),
             _CharmSpec(MyCharm, {'containers': {'foo': {}, 'bar': {}}}),
         )
+        # The container name contains a dash (gh #2565).
```
tonyandrewmeyer:
```suggestion
        # The container name contains a dash.
```

testing/tests/test_context_on.py
```diff
@@ -672,6 +672,79 @@ def __init__(self, framework: ops.Framework):
...
+def test_workload_event_juju_name_preserves_container_name(
+    container_name: str,
+    event_attr: str,
+    expected_suffix: str,
+):
+    """Juju preserves the container name verbatim in JUJU_HOOK_NAME.
+
+    Regression test: scenario was emitting hook names with the container
+    name's dashes mangled to underscores (see gh-2511, gh-2565).
+    """
```
tonyandrewmeyer:
```suggestion
    """Juju preserves the container name verbatim in JUJU_HOOK_NAME."""
```

testing/tests/test_context_on.py
```diff
@@ -672,6 +672,79 @@ def __init__(self, framework: ops.Framework):
...
+        ('receive-ca-cert', 'relation_changed', '-relation-changed'),
+        ('receive_ca_cert', 'relation_changed', '-relation-changed'),
+    ],
+)
+def test_relation_event_juju_name_preserves_endpoint(
+    endpoint: str,
+    event_attr: str,
+    expected_suffix: str,
+):
+    """Juju preserves the relation endpoint verbatim in JUJU_HOOK_NAME (gh-2511)."""
```
tonyandrewmeyer:
```suggestion
    """Juju preserves the relation endpoint verbatim in JUJU_HOOK_NAME."""
```

testing/tests/test_runtime.py
```diff
@@ -4,14 +4,15 @@
...
+from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any
 
 import pytest
 from scenario import ActiveStatus, Context
 from scenario._runtime import Runtime
 from scenario.errors import UncaughtCharmError
-from scenario.state import Relation, State, _CharmSpec, _Event
+from scenario.state import Relation, State, Storage, _Action, _CharmSpec, _Event
```
tonyandrewmeyer:
```suggestion
from scenario.state import Container, Relation, Secret, State, Storage, _Action, _CharmSpec, _Event
```

testing/tests/test_runtime.py
```diff
@@ -4,14 +4,15 @@
 from __future__ import annotations
 
 import os
+from pathlib import Path
```
tonyandrewmeyer:
```suggestion
import pathlib
```

testing/tests/test_runtime.py
```diff
@@ -271,3 +272,84 @@ def test_bare_charm_errors_not_set(monkeypatch: pytest.MonkeyPatch):
...
+def _env_for(charm_meta: dict[str, Any], event: _Event, state: State) -> dict[str, str]:
+    """Build the dispatch env that Runtime would set, for direct assertions."""
+    with TemporaryDirectory() as tmp:
+        rt: Runtime[ops.CharmBase] = Runtime(
+            app_name='dashchk',
+            charm_spec=_CharmSpec(charm_type(), charm_meta),
+            charm_root=tmp,
+            unit_id=0,
+        )
+        return rt._get_event_env(state, event, Path(tmp))
```
tonyandrewmeyer:
```suggestion
        return rt._get_event_env(state, event, pathlib.Path(tmp))
```

testing/tests/test_runtime.py
```diff
@@ -271,3 +272,84 @@ def test_bare_charm_errors_not_set(monkeypatch: pytest.MonkeyPatch):
...
+
+
+def test_workload_event_dispatch_path_preserves_container_name():
+    """JUJU_HOOK_NAME/JUJU_DISPATCH_PATH carry the container name verbatim (gh-2565)."""
+    from scenario.state import Container
```
tonyandrewmeyer:
```suggestion
```

testing/tests/test_runtime.py
```diff
@@ -271,3 +272,84 @@ def test_bare_charm_errors_not_set(monkeypatch: pytest.MonkeyPatch):
...
+            app_name='dashchk',
+            charm_spec=_CharmSpec(charm_type(), charm_meta),
+            charm_root=tmp,
+            unit_id=0,
+        )
+        return rt._get_event_env(state, event, Path(tmp))
+
+
+def test_workload_event_dispatch_path_preserves_container_name():
+    """JUJU_HOOK_NAME/JUJU_DISPATCH_PATH carry the container name verbatim (gh-2565)."""
```
tonyandrewmeyer:
```suggestion
    """JUJU_HOOK_NAME/JUJU_DISPATCH_PATH carry the container name verbatim."""
```

testing/tests/test_runtime.py
```diff
@@ -271,3 +272,84 @@ def test_bare_charm_errors_not_set(monkeypatch: pytest.MonkeyPatch):
...
+        ('config_changed', 'config-changed'),
+        ('upgrade_charm', 'upgrade-charm'),
+        ('leader_elected', 'leader-elected'),
+        ('start', 'start'),
+        ('secret_changed', 'secret-changed'),
+    ],
+)
+def test_builtin_event_dispatch_path_is_hyphenated(path: str, hook_name: str):
+    """Juju hook names are hyphenated; scenario used to leak underscores."""
+    from scenario.state import Secret
```
tonyandrewmeyer:
```suggestion
```

testing/tests/test_runtime.py
```diff
@@ -271,3 +272,84 @@ def test_bare_charm_errors_not_set(monkeypatch: pytest.MonkeyPatch):
...
+
```
tonyandrewmeyer:
```suggestion
```

testing/tests/test_runtime.py
```diff
@@ -271,3 +272,84 @@ def test_bare_charm_errors_not_set(monkeypatch: pytest.MonkeyPatch):
...
+            charm_root=tmp,
+            unit_id=0,
+        )
+        return rt._get_event_env(state, event, Path(tmp))
+
+
+def test_workload_event_dispatch_path_preserves_container_name():
+    """JUJU_HOOK_NAME/JUJU_DISPATCH_PATH carry the container name verbatim (gh-2565)."""
+    from scenario.state import Container
+
```
tonyandrewmeyer:
```suggestion
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -205,10 +204,11 @@ def _check_relation_event(
...
         )
     else:
-        if not event.name.startswith(_normalise_name(event.relation.endpoint)):
+        if event._path.prefix != event.relation.endpoint:
             errors.append(
-                f'relation event should start with relation endpoint name. {event.name} does '
-                f'not start with {event.relation.endpoint}.',
+                f'relation event prefix {event._path.prefix!r} does not match the relation '
+                f'endpoint name {event.relation.endpoint!r} (the endpoint name must appear '
+                f'verbatim, as declared in the charm metadata).',
```
tonyandrewmeyer:
```suggestion
                exactly as declared in the charm metadata).',
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -230,22 +230,23 @@ def _check_workload_event(
             'cannot construct a workload event without the container instance. Please pass one.',
         )
     else:
-        if not event.name.startswith(_normalise_name(event.container.name)):
+        if event._path.prefix != event.container.name:
             errors.append(
-                f'workload event should start with container name. {event.name} does '
-                f'not start with {event.container.name}.',
+                f'workload event prefix {event._path.prefix!r} does not match the container '
+                f'name {event.container.name!r} (the container name must appear verbatim, '
```
tonyandrewmeyer:
```suggestion
                f'name {event.container.name!r} (the container name must appear exactly '
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -230,22 +230,23 @@ def _check_workload_event(
...
+                f'as declared in the charm metadata).',
+            )
+        if event.container not in state.containers:
+            errors.append(
+                f'cannot emit {event.name} because container {event.container.name} '
+                f'is not in the state (a container with the same name is not '
+                f'sufficient - you must pass the object in the state to the event).',
```
tonyandrewmeyer:
```suggestion
                f'sufficient. You must pass the object in the state to the event).',
```

testing/src/scenario/_consistency_checker.py
```diff
@@ -296,10 +297,11 @@ def _check_storage_event(
         errors.append(
             'cannot construct a storage event without the Storage instance. Please pass one.',
         )
-    elif not event.name.startswith(_normalise_name(storage.name)):
+    elif event._path.prefix != storage.name:
         errors.append(
-            f'storage event should start with storage name. {event.name} does '
-            f'not start with {storage.name}.',
+            f'storage event prefix {event._path.prefix!r} does not match the storage '
+            f'name {storage.name!r} (the storage name must appear verbatim, as declared '
```
tonyandrewmeyer:
```suggestion
            f'name {storage.name!r} (the storage name must appear exactly as declared '
```

testing/src/scenario/_runtime.py
```diff
@@ -67,13 +67,20 @@ def __init__(
         self._availability_zone = availability_zone
         self._principal_unit = principal_unit
 
-    def _get_event_env(self, state: State, event: _Event, charm_root: Path):
+    def _get_event_env(self, state: State, event: _Event, charm_root: Path) -> dict[str, str]:
         """Build the simulated environment the operator framework expects."""
+        if event._is_action_event and (action := event.action):
+            # Real Juju dispatches actions as `actions/<action-name>` (no
```
tonyandrewmeyer:
```suggestion
            # Juju dispatches actions as `actions/<action-name>` (no
```

testing/src/scenario/_runtime.py
```diff
@@ -67,13 +67,20 @@ def __init__(
...
+            # `_action` suffix, no `hooks/` prefix); JUJU_HOOK_NAME is empty.
+            # action.name cannot contain underscores (Juju rejects them).
```
tonyandrewmeyer:
```suggestion
```

testing/src/scenario/state.py
```diff
@@ -366,11 +366,43 @@ def _update_metadata(
             object.__setattr__(self, 'rotate', rotate)
 
 
-def _normalise_name(s: str):
-    """Event names, in Scenario, uniformly use underscores instead of dashes."""
+def _to_python_attr(s: str):
+    """Translate a Juju metadata name (which may contain dashes) into Python-attr form.
```
```suggestion
    """Translate a Juju metadata name (which may contain dashes) into attribute form.
```

testing/src/scenario/state.py
```diff
@@ -366,11 +366,43 @@ def _update_metadata(
...
+
+    This is **lossy** — calling it on an already-dash-free name discards no
+    information, but a verbatim metadata name like ``foo-bar`` and the
+    Python-attr form ``foo_bar`` are not distinguishable afterwards. Use it
```
```suggestion
    Python attribute form ``foo_bar`` are not distinguishable afterwards. Use it
```

testing/src/scenario/state.py
```diff
@@ -366,11 +366,43 @@ def _update_metadata(
...
+    only for matching/dispatch; never round-trip through it to build outgoing
+    Juju env vars, since Juju preserves the metadata name verbatim.
+    """
     return s.replace('-', '_')
 
 
+# Real Juju's naming rules for charm metadata entities. Scenario enforces
```
```suggestion
# Juju's naming rules for charm metadata entities. Scenario enforces these
```

testing/src/scenario/state.py
```diff
@@ -366,11 +366,43 @@ def _update_metadata(
...
+# these so that tests cannot pass with names that a real deployment would
+# reject.
```
```suggestion
```

testing/src/scenario/state.py
```diff
@@ -366,11 +366,43 @@ def _update_metadata(
...
+    information, but a verbatim metadata name like ``foo-bar`` and the
+    Python-attr form ``foo_bar`` are not distinguishable afterwards. Use it
+    only for matching/dispatch; never round-trip through it to build outgoing
+    Juju env vars, since Juju preserves the metadata name verbatim.
+    """
     return s.replace('-', '_')
 
 
+# Real Juju's naming rules for charm metadata entities. Scenario enforces
+# these so that tests cannot pass with names that a real deployment would
```
```suggestion
# these so that tests cannot pass with names that a real deployment would reject.
```

james-garner-canonical (commented):
I'm a big fan of this, aligning Scenario's behaviour with Juju's here is clearly the right thing to do, and should make everyone's lives easier in the long run.

This does seem like it has the potential to be very breaky though, doesn't it? And there are two distinct types of breakage it might cause: catching real charm bugs that slipped through due to Scenario previously being incorrect; and catching testing-only reliance on Scenario's previous (incorrect) behaviour.

I haven't reviewed in depth at this point -- I'd really like to see a super-tox/Hyrum comparison for this PR before going further.

tonyandrewmeyer:
> This does seem like it has the potential to be very breaky though, doesn't it?

My hope, perhaps naïvely, is that everyone is using names in their tests that match the names in their charm, and so if their charm works, their tests will still work.

> And there are two distinct types of breakage it might cause: catching real charm bugs that slipped through due to Scenario previously being incorrect;

These ones I am ok with, although if there are a small number the practice of opening a few downstream PRs is probably good here.

> and catching testing-only reliance on Scenario's previous (incorrect) behaviour.

This would be worse, although I would still be tempted to say that this is a bug in the charm tests.

> I haven't reviewed in depth at this point -- I'd really like to see a super-tox/Hyrum comparison for this PR before going further.

Sure, running now.

tonyandrewmeyer:
Main: 208 passed / 58 failed
This branch: 205 passed / 61 failed

1. airflow-coordinator-k8s-operator, `state.get_relations("airflow_coordinator")[0]` → IndexError.
2. hook-service-operator: `InconsistentScenarioError: cannot emit hook_service_pebble_ready because container hook-service is not in the state`
3. hydra-operator: same pattern as hook-service (hydra_pebble_ready / container hydra).

I'll open downstream PRs.

james-garner-canonical:
> Main: 208 passed / 58 failed This branch: 205 passed / 61 failed
> 
> 1. airflow-coordinator-k8s-operator, `state.get_relations("airflow_coordinator")[0]` → IndexError.
> 2. hook-service-operator: `InconsistentScenarioError: cannot emit hook_service_pebble_ready because container hook-service is not in the state`
> 3. hydra-operator: same pattern as hook-service (hydra_pebble_ready / container hydra).
> 
> I'll open downstream PRs.

I assume those three are the only movement between passed and failed (that is, nothing new *passes* because of this PR). But I wonder if we need more information on any changes within the 58 that failed in both cases (e.g. any new individual tests failing). Why they're already failing is also important -- for example, if it's because we're not setting them up properly yet, then perhaps some additional manual or AI analysis of those cases would be valuable to understand if that isn't masking a possible new failure here.

Since we're currently working on fleshing out super-tox (into hyrum), this PR seems like a great testing ground for this kind of analysis.

tonyandrewmeyer:
> I assume those three are the only movement between passed and failed (that is, nothing new _passes_ because of this PR). But I wonder if we need more information on any changes within the 58 that failed in both cases (e.g. any new individual tests failing). Why they're already failing is also important -- for example, if it's because we're not setting them up properly yet, then perhaps some additional manual or AI analysis of those cases would be valuable to understand if that isn't masking a possible new failure here.

I can dig into that. Hyrum is already better than super-tox at this, in that logging is better.

> Since we're currently working on fleshing out super-tox (into hyrum), this PR seems like a great testing ground for this kind of analysis.

Fair, although the comparison feature has not landed yet. But I will be moving on to that now (well, after the fairly trivial private names and de-dependency changes).

tonyandrewmeyer:
Baseline pass rate: **73%** (78/107). Current pass rate: **73%** (78/107). 0 new failure(s), 0 resolved, 0 new error(s).

| Charm | Baseline | Current |
| --- | --- | --- |
| charm-ubuntu | failed: 3 errors; PendingDeprecationWarning: Harness is deprecated. For the recommended approach, see: https://documentation.ubuntu.com/ops/latest/howto/write-unit-tes… | same |
| charmed-etcd-operator | failed: 13 failed, 74 passed; InconsistentScenarioError x8 | same |
| exim-operator | failed: Exception: Can not find valid pkg-config name. | same |
| indico-operator | failed: uv resolve: no solution | same |
| kafka-k8s-operator | failed: collection error in tests/unit/test_tls_manager.py: subprocess.CalledProcessError: Command 'keytool' returned non-zero exit status 127. | same |
| kafka-operator | failed: collection error in tests/unit/test_tls_manager.py: subprocess.CalledProcessError: Command 'keytool' returned non-zero exit status 127. | same |
| ks-charmed | failed: FutureWarning: The 'wheel' package is no longer the canonical location | same |
| kubeflow-roles-operator | failed: 1 error; ModuleNotFoundError: No module named 'ops' | same |
| kyuubi-k8s-operator | failed: psycopg2 build: pg_config not found | same |
| mattermost-k8s-operator | failed: uv resolve: no solution | same |
| netbox-k8s-operator | failed: uv resolve: URL dep must be a direct requirement | same |
| opensearch-operator | failed: pytest: no tests ran | same |
| parca-agent-operator | failed: 5 failed, 17 passed; UncaughtCharmError x5 | same |
| pgbouncer-k8s-operator | failed: psycopg2 build: pg_config not found | same |
| pgbouncer-operator | failed: psycopg2 build: pg_config not found | same |
| postgresql-k8s-operator | failed: psycopg2 build: pg_config not found | same |
| postgresql-operator | failed: psycopg2 build: pg_config not found | same |
| powerflex-cinder-charm | failed: C extension build: Python.h not found | same |
| powerflex-nova-charm | failed: C extension build: Python.h not found | same |
| redis-operator | failed: 1 failed, 3 passed; ConnectionError x1 | same |
| self-signed-certificates-operator | failed: 5 errors; ModuleNotFoundError: No module named 'charms' | same |
| snappass-test | failed: 6 failed; TypeError x4 (+2 other) | same |
| synapse-operator | failed: uv resolve: URL dep must be a direct requirement | same |
| temporal-k8s-operator | failed: 13 failed, 40 passed; InconsistentScenarioError x13 | same |
| temporal-ui-k8s-operator | failed: 9 failed, 4 passed; InconsistentScenarioError x9 | same |
| timescaledb-charm | failed: 3 failed, 3 passed; InvalidSpecError x3 | same |
| traefik-k8s-operator | failed: uv resolve: no solution | same |
| zookeeper-k8s-operator | failed: 8 failed, 151 passed; InconsistentScenarioError x6 (+2 other) | same |
| zookeeper-operator | failed: 113 failed, 47 passed; UncaughtCharmError x106 (+7 other) | failed: 67 failed, 93 passed; UncaughtCharmError x57 (+10 other) |

james-garner-canonical:
Thanks for the tabulated results @tonyandrewmeyer. A few questions:
- Why are those 29 specifically tabulated?
- Why does `zookeeper-operator` have less failures now? The implication seems to be that they have 100+ failing tests in their CI currently, and this PR would resolve 50 or so of them.
- I notice that the three failures called out in [your earlier comment](https://github.com/canonical/operator/pull/2570#issuecomment-4705832382) aren't in the table (and also the table is reporting on ~100 charms while your earlier comment mentioned ~200 charms).

tonyandrewmeyer:
> * Why are those 29 specifically tabulated?

They are the ones that changed from run to run.

> * Why does `zookeeper-operator` have less failures now? The implication seems to be that they have 100+ failing tests in their CI currently, and this PR would resolve 50 or so of them.

I will double-check this, but I think it's because the patching fixes things.

> * I notice that the three failures called out in [your earlier comment](https://github.com/canonical/operator/pull/2570#issuecomment-4705832382) aren't in the table

Yes, all three PRs I opened downstream have been merged 😄.

> (and also the table is reporting on ~100 charms while your earlier comment mentioned ~200 charms).

I may have been using the big list from #26, I think.

tonyandrewmeyer:
> > * Why are those 29 specifically tabulated?
> 
> They are the ones that changed from run to run.

Sorry, that's obviously not right, otherwise there would not be the "same" rows. I was thinking of that, but decided against it. They are the charms that failed.

tonyandrewmeyer:
Something is broken with Zookeeper (and hyrum, not ops).

Without ops-source:

```
=== meta ===
repo: /home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator
runner: tox
target: unit
status: failed
returncode: 4
duration_s: 1.16
=== stdout ===
unit: commands[0]> poetry install --with unit
Installing dependencies from lock file

No dependencies to install or update
unit: commands[1]> poetry run coverage run --source=/home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator/src -m pytest -v --tb native -s /home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator/tests/unit
unit: exit 4 (0.53 seconds) /home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator> poetry run coverage run --source=/home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator/src -m pytest -v --tb native -s /home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator/tests/unit pid=1334864
  unit: FAIL code 4 (1.03=setup[0.00]+cmd[0.49,0.53] seconds)
  evaluation failed :( (1.05 seconds)
=== stderr ===
ImportError while loading conftest '/home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator/tests/unit/conftest.py'.
tests/unit/conftest.py:9: in <module>
    from literals import SUBSTRATE
E   ModuleNotFoundError: No module named 'literals'
/home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator/.tox/unit/lib/python3.12/site-packages/coverage/control.py:915: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")
```

With this branch:

```
=== meta ===
repo: /home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator
runner: tox
target: unit
status: failed
returncode: 1
duration_s: 7.14
=== stdout ===
unit: commands[0]> poetry install --with unit
Installing dependencies from lock file

Package operations: 0 installs, 2 updates, 0 removals

  - Updating ops (3.7.1 9a31478 -> 3.7.1 316ee32)

  TypeError

  Link.__init__() got an unexpected keyword argument 'size'

  at .tox/unit/lib/python3.12/site-packages/poetry/repositories/link_sources/json.py:50 in _link_cache
       46│                     else:
       47│                         metadata = bool(metadata_value)
       48│                     break
       49│ 
    →  50│             link = Link(
       51│                 url,
       52│                 requires_python=requires_python,
       53│                 hashes=hashes,
       54│                 yanked=yanked,

Cannot install ops.

unit: exit 1 (7.00 seconds) /home/tony.meyer@canonical.com/.cache/hyrum/charms/canonical/zookeeper-operator> poetry install --with unit pid=1335309
  unit: FAIL code 1 (7.00=setup[0.00]+cmd[7.00] seconds)
  evaluation failed :( (7.02 seconds)
=== stderr ===
```

Also, the comparison table output in hyrum seems wrong in terms of numbers. I guess this is the risk of using an unfinished branch for the comparison.

I'll dig into it more in hyrum. But I don't believe it is related here.

testing/src/scenario/_runtime.py
```diff
@@ -67,13 +67,19 @@ def __init__(
         self._availability_zone = availability_zone
         self._principal_unit = principal_unit
 
-    def _get_event_env(self, state: State, event: _Event, charm_root: Path):
+    def _get_event_env(self, state: State, event: _Event, charm_root: Path) -> dict[str, str]:
         """Build the simulated environment the operator framework expects."""
+        if event._is_action_event and (action := event.action):
```
tromai: This is not related to this PR, it's something I noticed.

`event._is_action_event` 's doc string says `Whether the event name indicates that this is a relation event.` Should it be "... an action event" instead 🤔 ?

testing/src/scenario/_runtime.py
```diff
@@ -67,13 +67,19 @@ def __init__(
...
+            # Juju dispatches actions as `actions/<action-name>` (no
+            # `_action` suffix, no `hooks/` prefix); JUJU_HOOK_NAME is empty.
```
tromai: I wonder if `_action suffix` be `-action suffix`, following the PR description ?

tromai (approved):
Thank you.

I really like this change. It's good to have a validation mechanism for naming and avoid diverting from Juju's behavior.

testing/src/scenario/_runtime.py
tonyandrewmeyer: Almost certainly. Feel free to open a tiny PR :)

testing/src/scenario/_runtime.py
tonyandrewmeyer:
```suggestion
            # suffix, no `hooks/` prefix); JUJU_HOOK_NAME is empty.
```

tonyandrewmeyer:
FYI in terms of the Hyrum runs, as far as I can tell, running the main charm set (not the extended one still in a PR) there are a few charm unit tests that are flakey with their regular ops version (missing mocks for the most part), but aside from that there is no difference between ops main and this branch. The zookeeper-operator case is the most extreme, where their lack of mocking intersects with a Scenario bug unrelated to this branch (#2581). With the fix from that branch, (patched) main and this (patched) branch fail in exactly the same way.

testing/src/scenario/_runtime.py
james-garner-canonical: Lazy question: why would an action event not have an action field set? If this is actually just satisfying the type checker, than perhaps an `assert` would be more appropriate?

james-garner-canonical: I see there's a pre-existing instance of the same check a few lines below.

I wonder if it would be a readability improvement to move this if/else logic there so we only check this condition once. Not a strong suggestion, just wondering.

testing/src/scenario/state.py
```diff
@@ -366,11 +366,62 @@ def _update_metadata(
...
+    only for matching/dispatch; never round-trip through it to build outgoing
+    Juju env vars, since Juju preserves the metadata name verbatim.
+    """
     return s.replace('-', '_')
 
 
+# Juju's naming rules for charm metadata entities. Scenario enforces these
+# these so that tests cannot pass with names that a real deployment would reject.
+#
+# Sources (juju/names and juju/charm, identical in Juju 3.x and 4.x):
```
james-garner-canonical: > `# Sources (juju/names and juju/charm, identical in Juju 3.x and 4.x):`

This comment will age. Maybe worth mentioning the actual versions checked?

testing/src/scenario/state.py
```diff
@@ -366,11 +366,62 @@ def _update_metadata(
...
+# - relation endpoints: RelationSnippet — underscores ARE allowed.
```
james-garner-canonical, line 450: Question: do we do anything with interface names?

testing/src/scenario/state.py
```diff
@@ -2140,25 +2201,24 @@ class _EventPath(str):
...
-        instance.prefix = string.removesuffix(instance.suffix)
-        # The original (un-normalised) prefix, preserving the exact spelling
-        # of the entity name as declared in the charm metadata. _normalise_name
-        # only swaps dashes for underscores, so lengths match.
-        instance.original_prefix = original_string[: len(instance.prefix)]
-        instance._is_custom = instance.suffix == ''
+        # `name` is the Python-attribute form: what the event is called on
+        # `charm.on`, and the event kind ops uses in handle paths (for
+        # example ``foo_bar_pebble_ready`` for container ``foo-bar``).
+        instance.name = _to_python_attr(raw_name)
```
james-garner-canonical: Since this is private, maybe we could call it `python_name` or `underscored_name` for clarity?

testing/src/scenario/state.py
```diff
@@ -2140,25 +2201,24 @@ class _EventPath(str):
...
+        instance.suffix, instance.type = _EventPath._get_suffix_and_type(instance.name)
+        # The prefix is sliced from the raw string by length so the entity
+        # name is preserved verbatim (e.g. ``foo-bar`` in
+        # ``foo-bar_pebble_ready``): Juju hook names keep the metadata name
+        # exactly as declared, and only the suffix is known to be hyphenated.
+        instance.prefix = raw_name[: len(raw_name) - len(instance.suffix)]
```
james-garner-canonical: How about using `removesuffix`? Or can we not, because the `suffix` isn't *actually* the suffix of the raw name (since it's underscore normalised)?
```suggestion
        instance.prefix = raw_name.removesuffix(instance.suffix)
```

testing/src/scenario/state.py
```diff
@@ -2140,25 +2201,24 @@ class _EventPath(str):
...
-        # The original (un-normalised) prefix, preserving the exact spelling
-        # of the entity name as declared in the charm metadata. _normalise_name
-        # only swaps dashes for underscores, so lengths match.
-        instance.original_prefix = original_string[: len(instance.prefix)]
-        instance._is_custom = instance.suffix == ''
+        # `name` is the Python-attribute form: what the event is called on
+        # `charm.on`, and the event kind ops uses in handle paths (for
+        # example ``foo_bar_pebble_ready`` for container ``foo-bar``).
+        instance.name = _to_python_attr(raw_name)
+        instance.suffix, instance.type = _EventPath._get_suffix_and_type(instance.name)
```
james-garner-canonical: If we stick with the current suffix semantics, I wonder if we should call this `python_suffix` or `underscored_suffix` for clarity too.

Though I also wonder if it would be cleaner to have the suffixes defined the other way around. Or be a mapping of some kind like `_suffixes.pebble_ready_event.underscored`. Maybe over engineering, but the current pattern still feels confusing. Maybe it's fine if you know they're all underscored by default.

testing/src/scenario/state.py
```diff
@@ -2287,14 +2334,18 @@ def _path(self) -> _EventPath:
...
     @property
     def name(self) -> str:
-        """Full event name.
+        """Full event name, in Python-attribute form (as ops names the event).
 
         Consists of a 'prefix' and a 'suffix'. The suffix denotes the type of the event, the
         prefix the name of the entity the event is about.
 
-        "foo-relation-changed":
+        "foo_relation_changed":
```
james-garner-canonical: Maybe `foo_bar_relation_changed` to make it clearer it's fully Python-normalised?

I see this example is added as a follow-up paragraph already, maybe just fold it into the single example?

testing/tests/test_e2e/test_deferred.py
```diff
@@ -164,6 +164,53 @@ def test_deferred_workload_event(mycharm: type[ops.CharmBase]):
...
+def test_deferred_workload_event_dashed_container_name():
+    """A deferred event for a dashed container name must be re-emitted.
+
+    Regression test: the deferred handle path must use the Python-attribute
+    form of the event name (``workload_a_pebble_ready``), not the verbatim
+    container name, or ops silently drops the notice.
+    """
+    captured: list[ops.EventBase] = []
+
+    class DashCharm(ops.CharmBase):
```
james-garner-canonical: I was briefly confused by `DashCharm` since I've been thinking of hyphens vs underscores. (`-craft`? 😆)

Not a big deal, but reckon it's worth being consistent with the terminology throughout? Now that I look, we're swapping back and forward between dash and hyphen in different places. I think typographically these are hyphens.

james-garner-canonical (commented):
Still digesting these dark corners of Scenario, thanks for your patience. I've read through the code changes now, and they look reasonable to me. I've left a few questions before final approval.

testing/src/scenario/_runtime.py
tonyandrewmeyer: I can get rid of a walrus with this change, I think, so a win from my point of view 😉.

testing/src/scenario/state.py
tonyandrewmeyer: I don't think the exact version is needed, since it should remain true for a minor version, and realistically for all Juju 3 at this point. I'll change to 3.6 and 4.0.

testing/src/scenario/state.py
tonyandrewmeyer: No. Scenario doesn't validate the `interface` field on Relation. We could add some, I guess? Would you want that, and in this PR?

testing/src/scenario/state.py
tonyandrewmeyer: Yeah, we can't use removesuffix here. `instance.suffix` is derived from the Python name, but the suffix in `raw_name` may use hyphens for the static part. The two strings only have the same length, not the same separators.

testing/src/scenario/state.py
tonyandrewmeyer: Happy to do the smaller change. I'm not sure it's worth doing the bigger mapping one. Convince me if you think we should 😄.

testing/tests/test_e2e/test_deferred.py
tonyandrewmeyer: Fair enough, dash is really em and en dashes, I think, and we're using a hyphen indeed. I'll cleanup the comments and class and whatever from this branch's changes but leave all other "dash" alone unless you think I should bundle that into the PR too.

testing/src/scenario/state.py
james-garner-canonical: Not in this PR, just wondering if it was a gap.

testing/src/scenario/_runtime.py
```diff
@@ -90,10 +90,14 @@ def _get_event_env(self, state: State, event: _Event, charm_root: Path):
             env['JUJU_PRINCIPAL_UNIT'] = self._principal_unit
 
         if event._is_action_event and (action := event.action):
```
james-garner-canonical: Can we ever have `event._is_action_event` but not `event.action`?

---

If so, we'll have a bad `JUJU_DISPATCH_PATH` set (we won't enter this branch, and so will retain the `hooks/<name>` path from the generic setup). We'd need to consider what it means for there to be an action event with no action, and what the environment should be.

---

If not, perhaps we should do this:

```suggestion
        if event._is_action_event:
            action = event.action
            assert action
```

And also set `JUJU_DISPATCH_PATH` in this branch (so all the action specific environment variable setup is here).

testing/src/scenario/state.py
```diff
@@ -2136,29 +2197,30 @@ class _EventType(str, Enum):
 
 class _EventPath(str):
     if TYPE_CHECKING:  # pragma: no cover
-        name: str
+        python_name: str
         owner_path: list[str]
-        suffix: str
+        python_suffix: str
         prefix: str
```
james-garner-canonical: Minor: Do you think it's worth making `prefix` be `juju_prefix` or similar so it's explicit, like the `python_name` and `python_suffix`?

james-garner-canonical (commented):
Thanks for addressing my feedback. I have one big open question about the action event handling.

testing/src/scenario/_runtime.py
tonyandrewmeyer: The consistency checker makes sure this doesn't happen, but I'll make it clearer in the code.

testing/src/scenario/_runtime.py
james-garner-canonical: What do you think about setting `JUJU_HOOK_NAME` for action events in the `event._is_action_event` branch instead of here?

james-garner-canonical (approved):

PR #2574, opened 2026-06-17
docs: fix first unit tests command in machine charm tutorial

dwilding:
In [Write tests for the helper module](https://documentation.ubuntu.com/ops/latest/tutorial/write-your-first-machine-charm/#write-tests-for-the-helper-module) in the machine charm tutorial, we add a test for the workload module, then say that the test can be run with:

```
tox -e unit -- tests/unit/test_tinyproxy.py
```

But this command fails with an AssertionError on the unit status. Cause:

- The `tox` command isn't restricting to a single file; it's running the state-transition tests (`test_charm.py`) as well as the workload module tests. Because in the profile's `tox.ini`, args are inserted after the test path. [See here](https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-machine/tox.ini.j2#L56-L57)

- At this point in the tutorial, the charm code has been updated to install and configure tinyproxy, but we haven't yet mocked tinyproxy in the state-transition tests. The placeholder test (from the profile) triggers the start event, which sets the wrong status because tinyproxy isn't installed on the host.

I figure the simplest fix is to use `-k` to select the only test we want to run.

### DIFF
--- a/docs/tutorial/write-your-first-machine-charm.md
+++ b/docs/tutorial/write-your-first-machine-charm.md
@@ -707,7 +707,7 @@ def test_version(monkeypatch: pytest.MonkeyPatch):
 We'll run all the tests later in the tutorial. But if you'd like to see whether this test passes, run:
 
 ```text
-tox -e unit -- tests/unit/test_tinyproxy.py
+tox -e unit -- -k test_version
 ```
 
 ### Write state-transition tests


tromai (approved):

tonyandrewmeyer (approved):
Discussed in person, makes sense to do this here. I'll see if I can figure out the larger problem that specifying a file doesn't really work.

tonyandrewmeyer:
I think the most correct fix is to do this:

```diff
@@ -64,8 +64,7 @@ commands =
         -v \
         -s \
         --tb native \
-        {[vars]tests_path}/unit \
-        {posargs}
+        {posargs:{[vars]tests_path}/unit}
     coverage report
```

So the `test/unit` is the default value, but skipped if there's one passed (after the `--`). So:

* `tox -e unit -- test/unit/test_foo.py` expands out to `pytest ... tests/unit/test_foo.py` as expected
* `tox -e unit` expands out to `pytest ... test/unit` like before
* *but* `tox -e unit -- -k my_test` expands out to `pytest ... -k my_test` so it'll catch anything in the integration tests as well, so you'd be expected to do `tox -e unit -- test/unit -k my_test` explicitly, expanding out to `pytest ... tests/unit -k my_test`

Probably we want to leave things alone until we move to `make`. It'll have a similar issue, just with a different way of handling defaults, but maybe we can figure out something there?

james-garner-canonical:
> I think the most correct fix is to do this:
> 
> ```diff
> @@ -64,8 +64,7 @@ commands =
>          -v \
>          -s \
>          --tb native \
> -        {[vars]tests_path}/unit \
> -        {posargs}
> +        {posargs:{[vars]tests_path}/unit}
>      coverage report
> ```
> 
> So the `test/unit` is the default value, but skipped if there's one passed (after the `--`). So:
> 
> * `tox -e unit -- test/unit/test_foo.py` expands out to `pytest ... tests/unit/test_foo.py` as expected
> * `tox -e unit` expands out to `pytest ... test/unit` like before
> * _but_ `tox -e unit -- -k my_test` expands out to `pytest ... -k my_test` so it'll catch anything in the integration tests as well, so you'd be expected to do `tox -e unit -- test/unit -k my_test` explicitly, expanding out to `pytest ... tests/unit -k my_test`
> 
> Probably we want to leave things alone until we move to `make`. It'll have a similar issue, just with a different way of handling defaults, but maybe we can figure out something there?

TBH I don't think that's quite right either, because it stops you from passing other flags without also passing the path. For example, `tox -e unit -- -k foo` would match `foo` in `tests/integration` as well -- to preserve the current behaviour there you'd need `tox -e unit -- tests/unit -k foo`.

I think with command runners like this, you pretty much always end up wanting a proper CLI interface -- so you could have `paths` be (optional) positional arguments, and pass everything else through unmodified. Easy with the approach we recently adopted in `charmlibs` (thin `justfile` to dispatch recipes from anywhere to a custom Python script that uses `argparse` to extract known arguments and forward the rest on https://github.com/canonical/charmlibs/pull/518), but I'm not sure the world is ready for that yet ;)

tonyandrewmeyer:
> TBH I don't think that's quite right either, because it stops you from passing other flags without also passing the path. For example, `tox -e unit -- -k foo` would match `foo` in `tests/integration` as well -- to preserve the current behaviour there you'd need `tox -e unit -- tests/unit -k foo`.

I feel like this is exactly what the third bullet point says?

> I think with command runners like this, you pretty much always end up wanting a proper CLI interface

I feel like we can maybe get something along these lines in the upcoming Makefile, which is why I suggested doing nothing until then.

james-garner-canonical:
> > TBH I don't think that's quite right either, because it stops you from passing other flags without also passing the path. For example, `tox -e unit -- -k foo` would match `foo` in `tests/integration` as well -- to preserve the current behaviour there you'd need `tox -e unit -- tests/unit -k foo`.
> 
> I feel like this is exactly what the third bullet point says?

Oops, you're right.

> > I think with command runners like this, you pretty much always end up wanting a proper CLI interface
> 
> I feel like we can maybe get something along these lines in the upcoming Makefile, which is why I suggested doing nothing until then.

:heavy_plus_sign:

PR #2546, opened 2026-06-05
docs: move 2.x docs to canonical.com/juju/docs/ops

dwilding:
This PR matches https://github.com/canonical/operator/pull/2545 for the 2.x docs.

<mark>**[Preview docs](https://canonical.com/juju/docs/ops/2.x/)** built from this branch. After this PR merges, I'll switch those docs to build from `2.23-maintenance` and redirect the existing docs.</mark>

**Main changes**

- Updated Ops doc URLs in `docs/custom_conf.py`.
- Updated hardcoded URLs for Ops, excluding `CHANGES.md`.
- Added a script `docs/.sphinx/_static/overwrite_links.js`. This script makes sure that the version switcher from Read the Docs has correct URLs. It's a standard part of the new docs setup at Canonical.

**Extra changes**

- Updated the intersphinx URL and hardcoded URLs for Pebble. To account for [pebble#882](https://github.com/canonical/pebble/pull/882).

PR #2091, opened 2025-10-07
chore: support Python 3.14

dimaqq:
Screenshot 2025-10-08 at 8 41 14

Pi-thon is here!
All rejoice Π🐍

PR #2576, opened 2026-06-22
ci: test Ops 2.23 with Python 3.14

james-garner-canonical:
#2546 is blocked because we don't run the tests on this branch with Python 3.14, but these are configured as required checks. Since Ops 2.23 only specifies Python `>=3.8`, we'd expect it to run fine with Python 3.14.

image

In addition to adding 3.14 to the list of versions in CI, we also need to add it to the list of environments in `tox.ini`, and follow #2091 in patching some of the tests: we now need to wrap our `functools.partial` objects in `staticmethod` due to `partial` implementing the decriptor protocol in Python 3.14 (https://github.com/python/cpython/issues/121027, https://github.com/python/cpython/pull/121089). However, because we're on an older `pyright` version in this maintenance branch, this leads to type checking errors which we ignore (bumping `pyright` resolves this but introduces 20+ other errors which would add noise to this PR).

### DIFF
--- a/.github/workflows/framework-tests.yaml
+++ b/.github/workflows/framework-tests.yaml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.10", "3.12"]
+        python-version: ["3.8", "3.10", "3.12", "3.14"]
         exclude:
         - {python-version: "3.8", os: "macos-latest"}  # macos-14 is arm64, and there's no Python 3.8 build for arm64
 
--- a/test/test_model_relation_data_class.py
+++ b/test/test_model_relation_data_class.py
@@ -129,8 +129,8 @@ nested_decode = functools.partial(json.loads, object_hook=json_nested_hook)
 
 
 class MyCharm(BaseTestCharm):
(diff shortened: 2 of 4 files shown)


tonyandrewmeyer (commented):
If we do this, we should do a follow-up PR to adjust the tools/versions page. I left off 26.04 support for 2.23 there specifically because we weren't doing this testing.

I'm not sure if we want to always do this (3.16 in 28.04, 3.18 in 30.04, and so on), because we'd rather that people have new branches for newer bases and move to 3.x, but then maybe being LTS means that we ought to. As long as it's almost no work, I guess it doesn't hurt.

(Submitting this now because I'd already written it, even though you just said leave this until it's out of draft.)

james-garner-canonical:
Thanks for the comment, Tony. I don't have strong feelings about whether we want to support Python 3.14 on Ops 2.23. We don't require any code changes to support it, this PR is just enabling testing on Python 3.14. But maybe it's fair for us to say our LTS support is limited to specific Python versions.

tonyandrewmeyer (approved):
Nice, thanks for doing this. I think we maybe should have a discussion about whether 3.14 (or really 26.04) is support for 2.23, but we can do that later.

dwilding (approved):
Thanks!

PR #2578, opened 2026-06-23
refactor: replace jsonpatch with an inline dict-diff

tonyandrewmeyer:
`testing/tests/helpers.py:jsonpatch_delta` is the only consumer of the `jsonpatch` package.

This PR replaces it with a small recursive `_dict_diff`: recurse into dicts, treat key-presence asymmetries as `add` / `remove`, treat any mismatched non-dict (lists, scalars) as a whole-value `replace`. Output is still sorted by `path + op` as before.

`jsonpatch_delta` is renamed to `state_delta` now that the library is no longer involved. The new name describes what the function does without baking in an implementation choice that's gone.

The advantage is that `jsonpatch` and its `jsonpointer` transitive are removed from the `unit` test extras, in exchange for ~17 lines of code.


dwilding (approved):
Nice! My agent tells me that performance is better too (although insignificant because `dataclasses.asdict()` dominates).

testing/tests/helpers.py
```diff
@@ -58,7 +57,24 @@ def trigger(
...
 
 
-def jsonpatch_delta(self: State, other: State) -> list[dict[str, Any]]:
+def _dict_diff(a: Any, b: Any, path: str, out: list[dict[str, Any]]) -> None:
+    # Emit RFC 6902-shaped patch ops describing how to transform `a` into `b`.
+    # Recurses into dicts; treats list / scalar differences as whole-value replaces.
+    if isinstance(a, dict) and isinstance(b, dict):
+        ad = cast('dict[str, Any]', a)
+        bd = cast('dict[str, Any]', b)
+        for key in ad.keys() - bd.keys():
```
tromai, line 71: In `jsonpatch`, it handles keys that contains `/` or `~` (see [here](https://github.com/stefankoegl/python-json-patch/blob/d8e1a6e244728c04229d601bc9a384d9b034c603/jsonpatch.py#L927-L931)). I don't think we would run into issues related to this, but still want to double check with you.

tromai (approved):
Thanks for the PR!

Just one minor change request: I can see that in `testing/UPGRADING.md`, `jsonpatch` is still mentioned. Should we remove it there too ?

tonyandrewmeyer:
> Just one minor change request: I can see that in `testing/UPGRADING.md`, `jsonpatch` is still mentioned. Should we remove it there too ?

No, because that is telling users that *they* can use jsonpatch, not referring to our use (in fact, it's saying to use it directly because we don't exposing functionality of it). If charmers want to use the library themselves, that's fine, we're just not using it internally any more.

A bigger question is how long the 6.0->7.0 upgrading guide should be kept. I think it's probably a little early to drop it now.

testing/tests/helpers.py
tonyandrewmeyer: Good point. I added `~`/`/` escaping. Dataclass field names were safe, but nested dicts can hold arbitrary strings, so better to be RFC 6902-correct.

tromai:
> No, because that is telling users that they can use jsonpatch, not referring to our use

Ah thanks. I understand it incorrectly the first time. Let's keep it.

PR #2581, opened 2026-06-24
fix: tear down `Runtime.exec()` when the charm raises

tonyandrewmeyer:
`Manager.run()` and `Manager.__exit__` both called `self._wrapped_ctx.__exit__` only on the happy path, so any exception escaping the charm during dispatch bypassed `Runtime.exec()`'s cleanup. That left the process environment polluted (most visibly with `OPERATOR_DISPATCH=1`), causing every subsequent test in the same process to hit `raise _Abort(0)` in `_Dispatcher.__init__` and cascade-fail with `UncaughtCharmError`.

The leak was most visible in suites where a test enters the context manager but doesn't call `mgr.run()` explicitly, relying on `__exit__` to do it implicitly: if that implicit dispatch raised, the wrapped context was never closed in test scope and only finalised at interpreter shutdown.

Specifically, I found this when investigating inconsistent failures running the zookeeper-operator tests. The charm tests are faulty, in that they assume a file path that may not exist, and races (in my case most easily triggered by running with and without coverage, to change timing) mean that tests fail at different points and then from that point onwards the dreaded `_Abort` shows up, as the environment causes all subsequent tests to fail.

The PR wraps `ops.run()` in `try/except` in `Manager.run()` so the wrapped context closes either way, and puts `Manager.__exit__`'s cleanup in `finally`. Both paths now route through `_exit_wrapped_ctx`, which is idempotent so the double-close from the happy path is harmless.


tonyandrewmeyer:
This may seem familiar to you @james-garner-canonical , from digging into other context cleanups a while back. Turns out, I missed one 😞.

I wondered whether I should just use this as an excuse to try to get rid of some of the layers of context managers of different types, and particularly the manual `__enter__` and `__exit__` calls, which still seem like a bad smell to me. But I'm not sure this is the right time to do that, rather than this simpler fix.

testing/src/scenario/context.py
```diff
@@ -121,23 +124,45 @@ def run(self) -> State:
...
+        try:
+            self.ops.run()
+        except BaseException:
+            # The charm raised: still tear down Runtime.exec() so the env
+            # snapshot is restored and resources are released, then propagate.
+            self._exit_wrapped_ctx(*sys.exc_info())
+            raise
+        else:
+            # wrap up Runtime.exec() so that we can gather the output state
+            self._exit_wrapped_ctx(None, None, None)
```
james-garner-canonical: Nit: I'd drop the `else`, since we `raise` unconditionally in the `except` -- it's not clear to me that this line belongs conceptually in the `else` when the following few lines don't.

testing/src/scenario/context.py
```diff
@@ -121,23 +124,45 @@ def run(self) -> State:
...
-
-        # wrap up Runtime.exec() so that we can gather the output state
-        self._wrapped_ctx.__exit__(None, None, None)
+        try:
+            self.ops.run()
+        except BaseException:
+            # The charm raised: still tear down Runtime.exec() so the env
+            # snapshot is restored and resources are released, then propagate.
+            self._exit_wrapped_ctx(*sys.exc_info())
+            raise
```
james-garner-canonical, line 133: Is `except BaseException ... raise` preferrable over `finally`? e.g.

```suggestion
        finally:
            # The charm raised: still tear down Runtime.exec() so the env
            # snapshot is restored and resources are released, then propagate.
            self._exit_wrapped_ctx(*sys.exc_info())
```

If there's an important reason we don't use `finally` (ordering?) then perhaps that's worth mentioning in a comment.

testing/src/scenario/context.py
```diff
@@ -121,23 +124,45 @@ def run(self) -> State:
...
+            if not self._emitted:
+                logger.debug(
+                    "user didn't emit the event within the context manager scope. "
+                    'Doing so implicitly upon exit...',
+                )
+                self.run()
+        finally:
+            # Always tear down Runtime.exec(), even if the implicit run() above
+            # raised: leaving it open leaks the env snapshot (including
+            # OPERATOR_DISPATCH) into subsequent tests in the same process.
```
james-garner-canonical: I was briefly confused why it's an "implicit run()" since it's called explicitly two lines above. I understand now that the comment is echoing the debug message above. How about instead:
```suggestion
            # Always tear down Runtime.exec(), even if run() raised. Leaving it open leaks the env
            # snapshot (including OPERATOR_DISPATCH) into subsequent tests in the same process.
```

testing/src/scenario/context.py
```diff
@@ -121,23 +124,45 @@ def run(self) -> State:
...
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        """Close the Runtime.exec() context exactly once."""
+        if self._wrapped_ctx_exited:
+            return
+        self._wrapped_ctx_exited = True
+        self._wrapped_ctx.__exit__(exc_type, exc_val, exc_tb)
```
james-garner-canonical, line 155: Do you think it's worth adding an assertion in a unit test that `__exit__` is only called once on the happy path? Could be helpful for eventual refactoring.

testing/tests/test_context.py
```diff
@@ -106,6 +106,42 @@ def test_context_manager():
...
+
+
+def test_context_manager_does_not_leak_env_when_implicit_run_raises():
+    """Manager.__exit__ must tear down Runtime.exec() even if the implicit run() raises.
+
+    Otherwise OPERATOR_DISPATCH (set during _Dispatcher.__init__) leaks into the
+    process environment and every subsequent test in the same process hits
+    `raise _Abort(0)`.
+    """
+    os.environ.pop('OPERATOR_DISPATCH', None)
```
james-garner-canonical: Should we use a monkeypatched environment instead?

james-garner-canonical (approved):
Thanks for this fix, I do remember the pain of these context managers well. I'd be in favour of a refactor at some point, as this part of the codebase is quite complicated to touch. For now this fix seems clean enough though. I have a few comments that I think are worth addressing, but approving optimistically.

testing/src/scenario/context.py
tonyandrewmeyer: Yeah, it's implicit because the *user* did not explicitly call it, Scenario did it for them. I'll adjust as suggested.

tromai (approved):
Thanks! The fix looks great.

PR #2583, opened 2026-06-25
chore: pin pydantic to 2-3

tonyandrewmeyer:
We only really supported Pydantic 2 currently, and can't know whether Pydantic 3 would work, so this keeps us honest.

This also helps with tooling that figures out which dependencies match when mixing ops[...] with other dependencies.

### DIFF
--- a/tracing/pyproject.toml
+++ b/tracing/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 dependencies = [
     "opentelemetry-sdk~=1.30",
     "ops==2.23.3.dev0",
-    "pydantic",
+    "pydantic>=2.0,<3",
 ]
 
 [project.urls]
--- a/tracing/pyproject.toml
+++ b/tracing/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 dependencies = [
     "opentelemetry-sdk~=1.30",
     "ops==2.23.3.dev0",
-    "pydantic>=2.0,<3",
+    "pydantic>=2,<3",
 ]
 
 [project.urls]
--- a/uv.lock
+++ b/uv.lock
@@ -2113,7 +2113,7 @@ testing = [
 requires-dist = [
     { name = "opentelemetry-sdk", specifier = "~=1.30" },
     { name = "ops", editable = "." },
-    { name = "pydantic" },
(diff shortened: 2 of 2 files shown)


tracing/pyproject.toml
```diff
@@ -33,7 +33,7 @@ classifiers = [
 dependencies = [
     "opentelemetry-sdk~=1.30",
     "ops==2.23.3.dev0",
-    "pydantic",
+    "pydantic>=2.0,<3",
```
dwilding: Would `>=2,<3` mean the same?

dwilding (approved):

tracing/pyproject.toml
tonyandrewmeyer: Yes. Or `~=2.0`. "There should be one-- and preferably only one --obvious way to do it." 😞

I can drop the `.0` if you prefer.

dwilding: I feel like `>=2,<3` is the way to go, because why not `2.0.0` rather than `2.0`

tonyandrewmeyer:
```suggestion
    "pydantic>=2,<3",
```

uv.lock
```diff
@@ -2113,7 +2113,7 @@ testing = [
 requires-dist = [
     { name = "opentelemetry-sdk", specifier = "~=1.30" },
     { name = "ops", editable = "." },
-    { name = "pydantic" },
+    { name = "pydantic", specifier = ">=2.0,<3" },
```
james-garner-canonical, line 2116: Do we need a relock, or does `uv` turn `>=2,<3` into `>=2.0,<3` in the lockfile?

james-garner-canonical (approved):
Good call, well spotted.

uv.lock
tonyandrewmeyer: It doesn't generate a change, so seems fine:

```
tony.meyer@tam-canonical-3:/w/operator$ git pull
Enter passphrase for key '/home/tony.meyer@canonical.com/.ssh/id_ed25519': 
Already up to date.
tony.meyer@tam-canonical-3:/w/operator$ uv lock
Resolved 189 packages in 0.83ms
tony.meyer@tam-canonical-3:/w/operator$ git diff
tony.meyer@tam-canonical-3:/w/operator$ 
```

PR #2584, opened 2026-06-25
test: fix integration test flakes on Juju 3 k8s and Juju 4 secrets

tonyandrewmeyer:
Small clean-ups for the integration tests:

- broaden the allowed goal-state unit-status set in `test_goal_state_reports_units` so a unit still in `maintenance` at action time doesn't fail the assertion (caught on k8s + Juju 3/stable)
- skip the whole `test_secrets` module on Juju >= 4 via an autouse fixture, citing juju/juju#22523 / juju/juju#22524; the suite was burning the full 20-minute job timeout because every secret-touching action hangs at the uniter commit phase

### DIFF
--- a/test/integration/test_hookcmds.py
+++ b/test/integration/test_hookcmds.py
@@ -422,9 +422,7 @@ def test_goal_state_reports_units(juju: jubilant.Juju, any_unit: str):
         'maintenance',
         'waiting',
         'error',
-        'unknown',
         'alive',
-        'dying',
     )
     for status in statuses:
         assert status in allowed, f'Unexpected goal status: {status}'
--- a/test/integration/test_hookcmds.py
+++ b/test/integration/test_hookcmds.py
@@ -412,12 +412,22 @@ def test_goal_state_reports_units(juju: jubilant.Juju, any_unit: str):
     assert unit_count >= 1
     unit_names = r['unit-names'].split(',')
     assert any(u.startswith('test-hookcmds/') for u in unit_names)
-    # Every unit should be in an 'alive' or 'waiting' goal state.
+    # Every unit should report a known goal status. This is the unit's
+    # workload status, plus the lifecycle values that goal-state can
+    # surface before the workload status is set.
     statuses = r['unit-statuses'].split(',')
+    allowed = (
+        'active',
+        'blocked',
+        'maintenance',
+        'waiting',
+        'error',
+        'unknown',
+        'alive',
+        'dying',
+    )
     for status in statuses:
-        assert status in ('alive', 'waiting', 'dying', 'active'), (
-            f'Unexpected goal status: {status}'
-        )
+        assert status in allowed, f'Unexpected goal status: {status}'
     assert len(unit_names) == unit_count
     assert len(statuses) == unit_count
     # The peer relation is always present in this deployment.
--- a/test/integration/test_secrets.py
+++ b/test/integration/test_secrets.py
@@ -27,6 +27,10 @@ _JUJU4_COMMIT_BUG = (
     'Juju 4.0 uniter commit-phase regression breaks secret-add (juju/juju#22523, juju/juju#22524)'
 )
 
+# Flip one test to xfail so CI fails loudly once Juju fixes the bug; the
+# rest stay as skip to keep the suite fast.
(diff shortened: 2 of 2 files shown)


test/integration/test_secrets.py
```diff
@@ -23,6 +23,23 @@
...
+def _skip_on_juju4(juju: jubilant.Juju):
+    """Skip the whole module on Juju >= 4 until the upstream bug is fixed.
+
+    Every `juju run` against an action that touches a secret returns aborted
+    at Juju's default 60-second action-wait, because the uniter never
+    commits the hook. Without this gate the suite spends 20 minutes timing
+    out every scheduled run.
+    """
+    if int(juju.status().model.version.split('.', 1)[0]) >= 4:
+        pytest.skip(_JUJU4_COMMIT_BUG)
```
james-garner-canonical: I'm OK with disabling these tests temporarily, but what's our mechanism for knowing when to turn them back on?

Maybe we can `xfail` one of them and skip the rest?

---

Here's one approach:

```suggestion
_JUJU4_MARKED_ONE_XFAIL = False  # Facilitate marking one test xfail, the rest are skipped.

@pytest.fixture(autouse=True)
def _skip_on_juju4(juju: jubilant.Juju):
    """Skip the whole module on Juju >= 4 until the upstream bug is fixed.

    Every `juju run` against an action that touches a secret returns aborted
    at Juju's default 60-second action-wait, because the uniter never
    commits the hook. Without this gate the suite spends 20 minutes timing
    out every scheduled run.
    """
    global _JUJU4_MARKED_ONE_XFAIL
    if int(juju.status().model.version.split('.', 1)[0]) >= 4:
        if _JUJU4_MARKED_ONE_XFAIL:
            pytest.skip(_JUJU4_COMMIT_BUG)
        else:
            pytest.xfail(_JUJU4_COMMIT_BUG)
            _JUJU4_MARKED_ONE_XFAIL = True
```

---

Alternatively, maybe that's complicated enough that we should just add `@pytest.mark.skipif` and `@pytest.mark.xfail` to the tests directly. Both accept a boolean first argument. We could use a `_JUJU4 = int(...)` global for that, and put the comment about the 60-second waits there.

james-garner-canonical (approved):
Thanks for taking care of this.

test/integration/test_secrets.py
tonyandrewmeyer: Yeah, I had trouble with this when I did all the other existing skips/xfails for Juju bugs I opened. To some extent, the answer is that I will get an email when each bug is closed (like https://github.com/juju/juju/issues/22639 yesterday) and follow-up on it. I know that's not as automated and multi-person as we would like, though.

The complexity mostly comes from fixes making their way through different versions and risks. I'll try to figure out something cleaner.

test/integration/test_hookcmds.py
```diff
@@ -412,12 +412,22 @@ def test_goal_state_reports_units(juju: jubilant.Juju, any_unit: str):
...
+    # workload status, plus the lifecycle values that goal-state can
+    # surface before the workload status is set.
     statuses = r['unit-statuses'].split(',')
+    allowed = (
+        'active',
+        'blocked',
+        'maintenance',
+        'waiting',
+        'error',
+        'unknown',
```
dwilding: I have limited knowledge here... But `unknown` would suggest a goal state that isn't known - contradicting the comment above.

dwilding (approved):

test/integration/test_hookcmds.py
tonyandrewmeyer: Fair point. I hit a case where the status was `maintenance` and solved it by adding all the valid states, since we don't really care what state it is, just that goal-state gives us one, but `unknown` probably is a problem.

test/integration/test_hookcmds.py
```diff
@@ -412,12 +412,22 @@ def test_goal_state_reports_units(juju: jubilant.Juju, any_unit: str):
...
+        'alive',
+        'dying',
```
tonyandrewmeyer:
```suggestion
```

test/integration/test_hookcmds.py
tonyandrewmeyer:
```suggestion
```

PR #2585, opened 2026-06-26
refactor!: move the otlp-json package to be a regular ops-tracing module

tonyandrewmeyer:
This PR moves the `vendor/otlp_json` package to be a single-file private module in ops-tracing.

There is an archived repository with the otlp-json package, and that fed the PyPI package, but we didn't use those and had our own copy, but in a vendor directory. The package was developed by Canonical. This cleans up the situation, and also avoids a vendor package, which we are trying to stop having in general.

This is a breaking change:

```terminal
$ git checkout de-vendor-otlp-json
Switched to branch 'de-vendor-otlp-json'
Your branch is up to date with 'origin/de-vendor-otlp-json'.
$ uv run --with '.[tracing]' python -c 'import ops_tracing.vendor.otlp_json'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'ops_tracing.vendor.otlp_json'
$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
$ uv run --with '.[tracing]' python -c 'import ops_tracing.vendor.otlp_json'
$ 
```

(If you do this, make sure you don't have a `__pycache__` lying around.)

However, I think the name "vendor" should have been a clear enough signal that no-one should have been doing that.

Possibly easiest to review commit-by-commit, as two are simple moves. The rest is automated ruff cleanup and import changes.


james-garner-canonical (approved):
Looks good, modulo the type checking failure.

The breaking change seems fine to me, I don't find any hits in our org other than `megademo-pebble-everywhere` which just has a copy of Ops.

james-garner-canonical:
WDYT about moving `ops_tracing/vendor` to `ops_tracing/_vendor` in a separate PR, while we're breaking things?

tonyandrewmeyer:
> Looks good, modulo the type checking failure.

Odd, it passes locally. Python version maybe, I will check.

tonyandrewmeyer:
> > Looks good, modulo the type checking failure.
> 
> Odd, it passes locally. Python version maybe, I will check.

Ah, you need to lint separately not from the top. We need to clean that up at some point.

tromai (approved):
Thank you. It looks good.

tonyandrewmeyer:
> WDYT about moving `ops_tracing/vendor` to `ops_tracing/_vendor` in a separate PR, while we're breaking things?

My preference would be to find time to get #2557 in a good shape and get that approved and merged. But I'm not opposed to the simple small change now if you're keen.

PR #2586, opened 2026-06-27
ci: re-enable the tracing integration tests

tonyandrewmeyer:
The upstream tracing issue is closed, so we can run the tracing integration tests again.

* Enable the disabled tracing integration tests
* Consolidate all the integration tests into a single job, which increases the coverage to match hookcmds (we lose MicroK8s, which should be fine, but we could add it in if needed)
* In the tracing conftest, install stable tempo rather than edge, and use the provided resource, not a custom one
* Pull the latest patch version of the vendored library
* Pin minio to a sidecar charm on Juju 4 (the unpinned channel still works on Juju 3) and grant `--trust`, since the default channel resolves to a podspec charm and Juju 4 dropped `pod-spec-set`
* Route host→in-cluster traffic (minio bucket setup, tempo span queries) through `kubectl port-forward`, since GitHub Actions runners return EPERM on direct ClusterIP connects and Juju 4 stopped exposing the ClusterIP on `apps[X].address` for Kubernetes
* Remove the uv.lock file on cleanup
* Document the remaining Juju 4 / k8s xfails against the upstream bugs they cover:
  * `test_with_tls` and the `test_secrets` action variants xfail on Juju 4 + k8s with a link to juju/juju#22485 (the `juju-secret-consumer-<id>` Role cannot patch `secrets`, so any charm that owns a secret hits 403 on every secret-set)
  * `test_hookcmds` retains its existing links to juju/juju#22523, juju/juju#22524, and juju/juju#22639
* Replace the inline concierge `-p k8s` / `-p machine` preset flags with checked-in `.github/integration/concierge-*.yaml` files that mirror the upstream presets and set `juju.extra-bootstrap-args: --config bootstrap-timeout=1800`. The k8s/4.0/stable slot was consistently failing on `unable to contact api server after 60 attempts` because the default 1200s bootstrap timeout was too short for the controller pod to expose its API on slow runners

The full matrix is green: https://github.com/tonyandrewmeyer/operator/actions/runs/28365210161. Run-to-run flakes are still possible on heavily-loaded runners (notably `snap install rockcraft` exceeding snapd's 5-minute hook runtime, and intermittent charmhub download failures), but they pass cleanly on rerun.


.github/workflows/integration.yaml
```diff
@@ -74,17 +20,30 @@ jobs:
...
+            test: test_direct_connection
+          - preset: machine
+            test: test_with_tls
 
     steps:
       - run: sudo snap install --classic concierge
       - run: >
           sudo concierge prepare
           --juju-channel=${{ matrix.juju-channel }}
+          --charmcraft-channel=3.x/stable
```
tonyandrewmeyer:
```suggestion
```

test/integration/conftest.py
```diff
@@ -34,12 +37,119 @@ def juju() -> Generator[jubilant.Juju]:
         print(juju.debug_log())
 
 
+def _kubectl_cluster_ip(namespace: str, service: str) -> str:
+    """Return the ClusterIP of `service` in `namespace` via kubectl.
+
+    Juju 4 no longer exposes the Service ClusterIP on apps[X].address for caas
```
tonyandrewmeyer:
```suggestion
    Juju 4 no longer exposes the Service ClusterIP on apps[X].address for machine
```

test/integration/conftest.py
```diff
@@ -34,12 +37,119 @@ def juju() -> Generator[jubilant.Juju]:
         print(juju.debug_log())
 
 
+def _kubectl_cluster_ip(namespace: str, service: str) -> str:
+    """Return the ClusterIP of `service` in `namespace` via kubectl.
+
+    Juju 4 no longer exposes the Service ClusterIP on apps[X].address for machine
```
tonyandrewmeyer:
```suggestion
    Juju 4 no longer exposes the Service ClusterIP on apps[X].address for Kubernetes
```

test/integration/conftest.py
```diff
@@ -51,19 +161,24 @@ def tracing_juju(juju: jubilant.Juju) -> Generator[jubilant.Juju]:
...
-        access_key='accesskey',
-        secret_key='mysoverysecretkey',
-        secure=False,
-    )
-
-    found = mc_client.bucket_exists('tempo')
-    if not found:
-        mc_client.make_bucket('tempo')
+    # Juju 4 stopped exposing the Service ClusterIP on apps[X].address for
+    # caas; resolve it from k8s directly. The ClusterIP is what in-cluster
```
tonyandrewmeyer:
```suggestion
    # Kubernetes; resolve it from k8s directly. The ClusterIP is what in-cluster
```

test/integration/conftest.py
```diff
@@ -34,12 +37,119 @@ def juju() -> Generator[jubilant.Juju]:
...
+resource "secrets"` warnings.
+"""
+
+
+def _xfail_on_k8s_juju4(juju: jubilant.Juju, reason: str) -> None:
+    """xfail the current test on Kubernetes substrates with Juju 4.x.
+
+    Caller must pass a reason that names the specific upstream Juju issue.
+    """
+    status = juju.status()
```
james-garner-canonical: Cleaner to use `juju.show_model()`?

test/integration/conftest.py
```diff
@@ -34,12 +37,119 @@ def juju() -> Generator[jubilant.Juju]:
...
+    if status.model.type == 'caas' and status.model.version.startswith('4.'):
```
james-garner-canonical: Interesting that the model type is surfaced as 'containers as a service'. Is that how Juju surfaces K8s in status, or is it a quirk of the environment? Maybe `model.model_type` has a more intuitive value than `model.type`?

test/integration/conftest.py
```diff
@@ -34,12 +37,119 @@ def juju() -> Generator[jubilant.Juju]:
...
+    # crashes on install (pod-spec-set was removed in Juju 4); latest/edge is
+    # the only sidecar variant. On Juju 3 the latest/edge charm doesn't reach
+    # active either, but the unpinned default does.
+    minio_deploy: dict[str, object] = {
+        'config': {'access-key': 'accesskey', 'secret-key': 'mysoverysecretkey'},
+    }
+    if juju.status().model.version.startswith('4.'):
+        minio_deploy['channel'] = 'latest/edge'
+        minio_deploy['trust'] = True
+    juju.deploy('minio', **minio_deploy)  # type: ignore[arg-type]
```
james-garner-canonical: Suggestion to simplify typing and localise comments.
```suggestion
    minio_config = {'access-key': 'accesskey', 'secret-key': 'mysoverysecretkey'},
    if juju.status().model.version.startswith('4.'):
        # On Juju 4, the default minio channel resolves to a podspec charm that crashes on install
        # (pod-spec-set was removed in Juju 4); latest/edge is the only sidecar variant.
        juju.deploy('minio', channel='latest/edge', trust=True, config=minio_config)
    else:
        # On Juju 3 the latest/edge charm doesn't reach active, but the unpinned default does.
        juju.deploy('minio', config=minio_config)
```

test/integration/conftest.py
```diff
@@ -51,19 +161,24 @@ def tracing_juju(juju: jubilant.Juju) -> Generator[jubilant.Juju]:
...
-        secret_key='mysoverysecretkey',
-        secure=False,
-    )
-
-    found = mc_client.bucket_exists('tempo')
-    if not found:
-        mc_client.make_bucket('tempo')
+    # Juju 4 stopped exposing the Service ClusterIP on apps[X].address for
+    # Kubernetes; resolve it from k8s directly. The ClusterIP is what in-cluster
+    # consumers (the Tempo pods, via s3-integrator) need.
```
james-garner-canonical: Nit: redundant with the docstring of the function. I notice the function is only used here, so maybe in-line?

test/integration/conftest.py
```diff
@@ -181,8 +296,16 @@ def cleanup():
...
             sdist.rename(charm_dir / 'ops_tracing.tar.gz')
 
+        # uv lock does not refresh the recorded hash for `path = ` sources
+        # when the package version is unchanged, so the committed uv.lock
+        # keeps its stale hash and charmcraft pack later fails with a hash
+        # mismatch inside the build container. --refresh-package forces a
+        # recompute.
+        refresh_packages = ['--refresh-package', 'ops']
+        if build_tracing:
+            refresh_packages += ['--refresh-package', 'ops-tracing']
```
james-garner-canonical: We don't commit the `uv.lock` file for the charms afaict, so maybe it would be cleaner to add it to `cleanup` instead.

test/integration/conftest.py
```diff
@@ -51,19 +161,24 @@ def tracing_juju(juju: jubilant.Juju) -> Generator[jubilant.Juju]:
...
+    # consumers (the Tempo pods, via s3-integrator) need.
+    cluster_ip = _kubectl_cluster_ip(juju.model, 'minio')
+
+    # The host can't necessarily reach the ClusterIP (on GitHub Actions a
+    # direct connect returns EPERM), so create the bucket via a port-forward.
+    with kubectl_port_forward(juju.model, 'svc/minio', 9000) as (host, port):
+        mc_client = minio.Minio(
+            f'{host}:{port}',
+            access_key='accesskey',
+            secret_key='mysoverysecretkey',
```
james-garner-canonical, line 162: Might we get dinged by linting for these hard-coded keys? 😆

tracing/ops_tracing/vendor/charms/certificate_transfer_interface/v1/certificate_transfer.py
```diff
@@ -4,7 +4,15 @@
 """Library for the certificate_transfer relation.
```
james-garner-canonical, line 4: What about switch to [charmlibs-interfaces-certificate-transfer](https://pypi.org/project/charmlibs-interfaces-certificate-transfer/) rather than updating the vendored lib?

.github/workflows/integration.yaml
```diff
@@ -74,17 +20,33 @@ jobs:
           - 4.0/stable
           - 4.0/edge
           - 4.1/edge
+        test:
+          - test_direct_connection
+          - test_with_tls
+          - test_relation_units
+          - test_secrets
+          - test_hookcmds
```
james-garner-canonical: I know the hard-coded list approach was used in the previous version, but WDYT about switching to generating this list with `pytest --collect-only` now that we're unifying to a single list? It always bothered me that new integration tests would be silently skipped unless the workflow was updated correctly.

`pytest-jubilant` uses this approach in its [integration test workflow](https://github.com/canonical/pytest-jubilant/blob/main/.github/workflows/integration.yaml).

.github/integration/concierge-k8s.yaml
```diff
@@ -0,0 +1,37 @@
+# Mirror of the upstream concierge k8s preset
+# (https://github.com/canonical/concierge/blob/main/presets/k8s.yaml) with
+# juju.extra-bootstrap-args set: `juju bootstrap` regularly fails on slow GHA
+# runners with `unable to contact api server after 60 attempts` because the
+# default bootstrap-timeout is too short for the controller pod to expose its
+# API. Pin a longer timeout here (1800s = 30min, vs the 1200s default).
+juju:
+  extra-bootstrap-args: --config bootstrap-timeout=1800
```
james-garner-canonical, line 8: I wonder if we should consider raising the default in Concierge itself somehow, since the primary use of  Concierge is likely Github CI. But I like the pattern of committing our own config files either way.

.github/workflows/integration.yaml
james-garner-canonical: Feel free to reject for now as out of scope for this PR, but while we're resurrecting these, should we increase the cadence from monthly to weekly and open issues on failure, following [Jubilant's approach](https://github.com/canonical/jubilant/blob/main/.github/workflows/ci.yaml)?

james-garner-canonical (approved):
Thanks for taking care of this, great to have these healthy again. I've left a few comments and suggestions.

test/integration/conftest.py
tonyandrewmeyer: It's what the `type` is in status, although you don't see that in the regular tabular format, only the cloud.

Not relevant with the switch to show_model.

test/integration/conftest.py
tonyandrewmeyer: Maybe, although it seemed like TIOBE were specifically looking for "password" (and probably similar keys). We had these before, just in a different place in the code, so should be ok.

tracing/ops_tracing/vendor/charms/certificate_transfer_interface/v1/certificate_transfer.py
tonyandrewmeyer: I've done this, because it does seem more correct than bumping the patch version, which is needed to get things to work. It does seem like a large change, though. It provides a more interesting comparison for my de-vendoring PR, I guess.

.github/workflows/integration.yaml
tonyandrewmeyer: I've switched to this. I don't think we want every test to be done individually (that would be a *lot* of setup and teardown for the hook commands). Per file seems ok, although it collapses the two tracing ones. We could split them into two files if we don't want that.

.github/integration/concierge-k8s.yaml
tonyandrewmeyer: Yeah, I already opened a PR for that, https://github.com/canonical/concierge/pull/221, but didn't want to wait for Concierge 1.7 in a month's time to get the tests to reliably pass here.

tracing/ops_tracing/vendor/charms/certificate_transfer_interface/v1/certificate_transfer.py
tonyandrewmeyer: Oh, except this is complicated, and I had forgotten. We shove in an "internal" prefix to avoid clashing with other users. So we can't use plain upstream. We could probably monkeypatch or work around this, but I think this tips me over into the "too much for this PR" bucket. I think it could reasonably be done as an alternative to my de-vendoring PR, though, especially when tracing could be done too.

(I typed this in the wrong thread previously. Deleted there and copied here.)

tromai (approved):
Thanks for your patience. It took me some time to wrap my head around the purpose of `kubectl_port_forward`.

.github/workflows/integration.yaml
tonyandrewmeyer: > Feel free to reject for now as out of scope for this PR,

Doing this :)

> but while we're resurrecting these, should we increase the cadence from monthly to weekly and open issues on failure, following [Jubilant's approach](https://github.com/canonical/jubilant/blob/main/.github/workflows/ci.yaml)?

I'm ok leaving this for the rainy day item for now. Moving to weekly is worth considering, for sure.

PR #2548, opened 2026-06-05
fix: avoid errors in finalisation due to already-closed websocket

tonyandrewmeyer:
This PR ensures that when closing the Pebble websocket:
* If the socket is already closed, we do nothing.
* If we get an error because it's already closed while closing (presumably a race), we ignore that.

#2547 has more details on why we have issues in Python 3.13+ when finalising at the moment.

An alternative would be to change `_wait` to close, but that is riskier, since it would be a behaviour change and could in theory trigger issues if someone is doing something odd with a wrapped websocket. Quite unlikely in our case, but the change in this PR solves the issue and seems clean enough, and doesn't have the risk.

Fixes #2547

PR #2588, opened 2026-06-28
fix: cherry-pick fixes from main into the 2.23-maintenance branch

tonyandrewmeyer:
A follow-up to the collection that was done a couple of months ago:

* #2548
* #2507 
* #2542 
* #2567 
* #2506 
* #2558 
* #2499

### DIFF
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -19,7 +19,7 @@ jobs:
           - charm-repo: canonical/postgresql-operator
             commit: 33ca7f308fa4289cdf45ceefcb335f373713431d  # 2025-06-21T22:22:22Z
           - charm-repo: canonical/postgresql-k8s-operator
-            commit: a6753b27440a17ed5f37d4c2c5c6f53b1d3a1f7f  # 2025-06-22T11:27:17Z
+            commit: c434df80953409f290e7ea6e76f78a3ec46cecc2  # 2026-05-30T03:35:40Z
           - charm-repo: canonical/mysql-operator
             commit: 411fb45ecfe200d22b46af9cbacaf72e8eb09da7  # 2025-06-23T14:51:47Z
           - charm-repo: canonical/mysql-k8s-operator
--- a/.github/workflows/db-charm-tests.yaml
+++ b/.github/workflows/db-charm-tests.yaml
@@ -6,12 +6,17 @@ on:
       - main
   pull_request:
   workflow_call:
+  workflow_dispatch:
 
 permissions: {}
 
 jobs:
+  build-wheels:
+    uses: ./.github/workflows/_build-wheels.yaml
+
   db-charm-tests:
     runs-on: ubuntu-latest
+    needs: build-wheels
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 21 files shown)


.github/workflows/db-charm-tests.yaml
```diff
@@ -26,29 +31,45 @@ jobs:
...
+          name: ${{ needs.build-wheels.outputs.artifact-name }}
+          path: operator-wheels
 
       - name: Install patch dependencies
         run: pip install poetry~=2.0
 
       - name: Update 'ops' dependency in test charm to latest
         run: |
+          ops_wheel="$(ls operator-wheels/ops-[0-9]*.whl)"
+          ops_tracing_wheel="$(ls operator-wheels/ops_tracing-*.whl)"
```
tromai, line 52: I wonder why the pattern for `ops_tracing` wheel is more relaxed compared to `ops_wheel` here?

.github/workflows/db-charm-tests.yaml
```diff
@@ -26,29 +31,45 @@ jobs:
...
+            echo -e "\n${ops_wheel}" >> requirements.txt
           else
-            sed -i -e "s/^ops[ ><=].*/ops = {path = \"myops\"}/" pyproject.toml
-            poetry lock
+            # The charm may declare ops with the 'tracing' extra, which pulls in
+            # ops-tracing from PyPI. Replacing ops with a local wheel drops the
+            # extra, so install the matching ops-tracing wheel explicitly when
+            # the charm asked for it. Check before `poetry remove`, which
+            # strips the ops line from pyproject.toml.
+            if grep -qE 'ops[^a-z]+=.*tracing|ops\[[^]]*tracing' pyproject.toml; then
```
tromai, line 62: Just to double check my understanding, the first sub pattern matches something like

```
ops = {extras = ["tracing"], version = ">=2.0"}
```

while the second pattern matches 

```
ops[testing,tracing]
```

tromai (approved):
Thank you.

I reviewed the commits after the cherry picked one. I understand why we need those fixes, and the fixes looks good to me.

tracing/pyproject.toml
```diff
@@ -33,7 +33,7 @@ classifiers = [
 dependencies = [
     "opentelemetry-sdk~=1.30",
     "ops==2.23.3.dev0",
-    "pydantic>=2,<3",
+    "pydantic>=1.10,<3",
```
james-garner-canonical, line 36: Why did we leave this unconstrained on `main`? I don't mind a bandaid fix in this PR, but this seems like something we should iron out more consistently. Will we be able to declare `>=2,<3` in future and rely on `hyrum` to force the library to work with the charm? Or do we need a more specific solution to relax our constraints in testing?

tonyandrewmeyer: I changed this to be constrained on the 2.23 branch because it was needed for some branch resolution that might be important later on (DM if that is too vague).

I would like Pydantic gone in 3.x, so I didn't bother making the change there, but it probably should be too. I think it may also need to be 1.x and above to handle downstream.

james-garner-canonical: OK, leaving this as-is here for now seems reasonable in that context.

.github/workflows/db-charm-tests.yaml
tonyandrewmeyer: The `[0-9]` in `ops-[0-9]*.whl` is defensive: without it, the glob would also match anything starting with `ops-` (for example a future `ops-tools-*.whl`). Wheel filenames normalise `-` to `_` in the package portion, so `ops_tracing-*.whl` already can't be matched by `ops-*` — there's no similarly-named package to disambiguate against, so the looser pattern is fine.

.github/workflows/db-charm-tests.yaml
tonyandrewmeyer: Yes, that's right. The first alternative (`ops[^a-z]+=.*tracing`) catches the Poetry table form like `ops = {extras = ["tracing"], ...}`, and the second (`ops\[[^]]*tracing`) catches the PEP 508 extras form like `ops[testing,tracing]` or `ops[tracing]`.

(I wish all the poetry usage would go away, but some people really like it.)

james-garner-canonical (approved):

Issue #2224, opened 2025-12-09
Decouple local and remote operations in release script and add tests


james-garner-canonical:
Per discussion in #2199 ([comment](https://github.com/canonical/operator/pull/2199#issuecomment-3634444335)), doing the following would make testing, updating, and maintaining the release script easier:
- decouple the flow of the local-only operations (choosing the new versions and applying them both pre and post release) from operations on the Github remote (creating a draft release, creating PRs)
- add unit tests for the cleanly decoupled parts
- add flag or prompts to run local operations only for easier almost-end-to-end testing

tonyandrewmeyer:
At the Madrid sprint we decided that we might to a more complete re-implementation of the release script.

dwilding:
We ran into an annoyance with the release script today. `uv lock` downgraded the revision of `uv.lock` from 3 to 2, even though I have the latest version of uv installed on my system. See https://github.com/canonical/operator/pull/2589#discussion_r3495537510.

What I think is happening: tox-uv bundles uv at `~/.local/share/uv/tools/tox/bin/uv`. That version of uv hadn't been updated since I originally installed tox-uv. Because the release script runs via tox, it ran `uv lock` with that old bundled copy of uv.

We should try to fix this.

tonyandrewmeyer:
Let's also see if we can drop rich as a dependency.

benhoyt:
@tonyandrewmeyer or @james-garner-canonical to spend a couple of days on this before they next do a release.

james-garner-canonical:
Intended UX (per discussion with @tonyandrewmeyer):
- [Workflow 1] Human kicks things off by running a workflow manually (with optional version input, optional branch argument (default main)).
  - If version not specified, figure out minor/patch from conventional commits.
  - Workflow creates a version bump PR with changelog and agent drafted release notes in the PR description.
  - Human approves and merges PR.
- [Workflow 2] Merging the PR results in a (draft?) release being created, populated from the changelog + release notes (from PR description).
  - Watch merges to all branches, create release from appropriate branch.
  - Watches for non-dev version bumps?
  - Human publishes the draft release.
  - In future we could go straight to this workflow publishing the release rather than creating a draft.
- [Workflow 3] Publishing release results in:
  - Triggering publish workflow (existing -- but merge with testing workflow, which can be manually run).
  - Creating the post-release PR (version bumps to dev versions).
  - Human approves and merges PR.

PR #2589, opened 2026-06-30
chore: update changelog and versions for 3.8.0 release

dwilding:
This PR prepares the release of version 3.8.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,8 +36,7 @@
 
 * Treat warnings as errors when running unit tests (#2541)
 * Fix mock_temp_dir fixture's open() patch (#2553)
-* Add integration tests for ops.hookcmds (#2520)
-* Hookcmds integration tests for secret access and port management (#2562)
+* Add integration tests for ops.hookcmds (#2520, #2562)
 * Fix integration test flakes on Juju 3 k8s and Juju 4 secrets (#2584)
 
 ## Refactoring
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@
 * Restore sys.breakpointhook on _Manager teardown (#2542)
 * Avoid errors in finalisation due to already-closed websocket (#2548)
 * Use the correct value for JUJU_HOOK_NAME when objects contain dashes (#2552)
-* Avoid double-closing the framework in scenario teardown (#2561)
+* Avoid double-closing the framework in testing teardown (#2561)
 * Don't leak exec I/O threads when waiting on the change fails (#2558)
 * Take Pebble defaults into consideration when consistency checking Checks (#2567)
 * Tear down `Runtime.exec()` when the charm raises (#2581)
@@ -30,7 +30,7 @@
 * Fix first unit tests command in machine charm tutorial (#2574)
 * Extend security explanation page to cover SEC0030 V1.3 gaps (#2571)
 * Move docs to canonical.com/juju/docs/ops (#2545)
-* Fix doc string of _Event._is_action_event (#2580)
+* Fix docstring of _Event._is_action_event (#2580)
 
 ## Tests
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,5 @@
 # 3.8.0 - 30 June 2026
 
-## Breaking Changes
-
-* Refactor: Move the otlp-json package to be a regular ops-tracing module (#2585)
-
 ## Features
 
 * Note the socket path in Pebble tracing spans (#2555)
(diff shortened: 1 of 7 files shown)


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
tonyandrewmeyer: I guess this doesn't matter? It's interesting that the revision dropped but there are no changes, other than the ones we are explicitly making.

I guess we don't pin our `uv` version, which is how this happens.

CHANGES.md
```diff
@@ -1,3 +1,57 @@
...
+* Extend security explanation page to cover SEC0030 V1.3 gaps (#2571)
+* Move docs to canonical.com/juju/docs/ops (#2545)
+* Fix doc string of _Event._is_action_event (#2580)
+
+## Tests
+
+* Treat warnings as errors when running unit tests (#2541)
+* Fix mock_temp_dir fixture's open() patch (#2553)
+* Add integration tests for ops.hookcmds (#2520)
+* Hookcmds integration tests for secret access and port management (#2562)
```
tonyandrewmeyer: If you'd rather just leave these separate that's fine, but I would do:


```suggestion
* Add integration tests for ops.hookcmds (#2520, #2562)
```

CHANGES.md
```diff
@@ -1,3 +1,57 @@
...
+* Replace CoC with link to Ubuntu Code of Conduct (#2564)
+* Add a small note about -W error in the unit test how-to (#2560)
+* Add doc about subordinate charms (#2455)
+* Add 26.04 to the versions page and mark 2.23 as LTS (#2517)
+* Clarify storage-detaching applies to both machine and K8s charms (#2554)
+* Add guidance about names of workload-less charms (#2496)
+* Fix first unit tests command in machine charm tutorial (#2574)
+* Extend security explanation page to cover SEC0030 V1.3 gaps (#2571)
+* Move docs to canonical.com/juju/docs/ops (#2545)
+* Fix doc string of _Event._is_action_event (#2580)
```
tonyandrewmeyer:
```suggestion
* Fix docstring of _Event._is_action_event (#2580)
```

CHANGES.md
```diff
@@ -1,3 +1,57 @@
...
+
+## Fixes
+
+* Treat remote unit zero as explicit (#2454)
+* Ensure resources are cleaned up in testing.Context (#2506)
+* Close SQLite storage in Harness.cleanup() (#2507)
+* Restore sys.breakpointhook on _Manager teardown (#2542)
+* Avoid errors in finalisation due to already-closed websocket (#2548)
+* Use the correct value for JUJU_HOOK_NAME when objects contain dashes (#2552)
+* Avoid double-closing the framework in scenario teardown (#2561)
```
tonyandrewmeyer:
```suggestion
* Avoid double-closing the framework in testing teardown (#2561)
```

tonyandrewmeyer (approved):

CHANGES.md
```diff
@@ -1,3 +1,57 @@
...
+* Don't leak exec I/O threads when waiting on the change fails (#2558)
+* Take Pebble defaults into consideration when consistency checking Checks (#2567)
+* Tear down `Runtime.exec()` when the charm raises (#2581)
```
tromai, line 18: We use back ticks here, but seems like back ticks are not used elsewhere (for code methods) in the Change log. I think we should remove the back ticks.

tromai (approved):

uv.lock
dwilding: I don't understand why this happens, as I have the latest version of uv installed. I just ran `uv lock --refresh` to refresh cached data and it reverted to revision 3. Also one other change - see https://github.com/canonical/operator/pull/2589/changes/03117fc70e2ae42ff183aa0eb5321dcffcb3ea20.

Maybe the release script should lock that way too?

CHANGES.md
tonyandrewmeyer: For what it's worth, that's because it's the one time I remembered to think about putting the backticks in because of how it would land in the changelog when I opened the PR. We should be using them consistently in the PR titles so that the changelog (and release notes) render better.

uv.lock
tonyandrewmeyer: Huh, interesting. Yeah, I guess the release script doing this would be better?

CHANGES.md
dwilding: I don't like the inconsistency. But we previously decided that we'd prefer to minimise manual editing of the changelog - we'd prefer that people use formatting in the original PR titles. (I'm sure I don't always remember though!)

CHANGES.md
dwilding: Ahh, you can have this one because I'm fixing the scenario one; otherwise I would leave it as-is :stuck_out_tongue:

uv.lock
dwilding: Added a comment about this on https://github.com/canonical/operator/issues/2224

PR #2609, opened 2026-07-01
ci: adopt new dependabot conventions

tonyandrewmeyer:
This PR adjusts the dependabot config to match the new Charm Tech standard ([spec: OP0xx-dependabot-config-conventions](https://github.com/canonical/charm-tech/blob/main/specs/OP0xx-dependabot-config-conventions.md)). Operator was the reference repo the canonical template was designed against.

The file is renamed `dependabot.yml` → `dependabot.yaml` and its contents replaced. Net effect:

**GitHub Actions lane**: keeps monthly cadence; gains:
- `open-pull-requests-limit: 100`.
- `cooldown.semver-major-days: 14` alongside the existing 7-day default.
- A single `actions: ["*"]` group so the monthly bump is one PR.

**Python (uv), root**: keeps monthly cadence; gains:
- Same `open-pull-requests-limit: 100` and `semver-major-days: 14`.
- Four named groups along the spec's seams:
  - `charm-tech`: `jubilant`, `pytest-jubilant`.
  - `dev-tooling`: canonical list (`ruff`, `pyright`, `ty`, `codespell`, `coverage`, `pre-commit`, `types-*`, `zizmor`) plus this repo's `mypy`, `flake8`, `pylint`.
  - `test-deps`: `pytest`, `pytest-*` (covers `pytest`, `pytest-xdist`, `pytest-benchmark`).
  - `runtime`: catch-all, **minor + patch only** so a runtime major never silently rides a patch bundle.

**Python (uv), `examples/httpbin-demo`**: converted to the canonical routine-lane shape (groups, cooldown, no per-directory `ignore:` list).

**Dropped `examples/k8s-5-observe` and `examples/machine-tinyproxy` entries** per spec: their committed `uv.lock` files are being removed and will be generated in the test runs instead, so Dependabot has nothing to track in those directories. The hand-rolled `ignore:` lists go away with them.


.github/dependabot.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - "dependencies"
+    open-pull-requests-limit: 100
+    cooldown:
+      default-days: 7
+      semver-major-days: 14
+    groups:
+      # Charm Tech's own releases (lockstep-versioned; we trust the release).
+      charm-tech:
+        patterns:
+          - "ops"
```
tonyandrewmeyer:
```suggestion
```

.github/dependabot.yaml
```diff
@@ -0,0 +1,117 @@
...
+          - "ops-scenario"
```
tonyandrewmeyer:
```suggestion
```

.github/dependabot.yaml
```diff
@@ -0,0 +1,117 @@
...
+          - "ops-tracing"
```
tonyandrewmeyer:
```suggestion
```

.github/dependabot.yaml
```diff
@@ -0,0 +1,114 @@
...
+    schedule:
+      interval: "monthly"
+    labels:
+      - "dependencies"
+    open-pull-requests-limit: 100
+    cooldown:
+      default-days: 7
+      semver-major-days: 14
+    groups:
+      # Charm Tech's own releases (lockstep-versioned; we trust the release).
```
tonyandrewmeyer:
```suggestion
      # Charm Tech's own releases (we trust the release).
```

james-garner-canonical (approved):
Looks good to me. I guess we need a note somewhere (where? `HACKING.md`?) about adding our/dev/testing deps to the right group in `dependabot.yaml` when adding new dependencies.

tromai (approved):
Thank you. It looks great to me.

tonyandrewmeyer:
> Looks good to me. I guess we need a note somewhere (where? `HACKING.md`?) about adding our/dev/testing deps to the right group in `dependabot.yaml` when adding new dependencies.

Good call; I've done that. I didn't do it on all the rest, I think we can figure that out when we unify things a bit more rather than repeating it now.

PR #2611, opened 2026-07-02
docs: wait longer for Loki data in K8s tutorial integration tests

dwilding:
Integration tests sometimes fail for the COS-integrated charm from our K8s tutorial. For example, https://github.com/canonical/operator/actions/runs/28460231844/job/84345818442

After asking an agent to run experiments to investigate the cause, it seems there is nothing fundamentally wrong with the tests - we just need to wait longer.

We need to account for delays in Pebble forwarding request data from the workload, Loki indexing the data, and the Loki endpoint propagating across the cross-model relation.

Waiting for 180 seconds (instead of 60) ought to be enough time.


james-garner-canonical (approved):
Nice solution, thanks. Making these less flaky is great.

tromai (approved):
Thanks!

PR #2612, opened 2026-07-02
ci: hash-pin actions and drop zizmor config

tonyandrewmeyer:
## Summary

- Pin every third-party GitHub Actions reference to a full commit SHA, with the original tag kept as a trailing comment for readability.
- Drop `.github/zizmor.yaml`. It only existed to relax `unpinned-uses` to `ref-pin` for `actions/*`, `github/*`, and `pypa/*`; with hash pins in place the default policy is satisfied.

## Test plan

- [x] `zizmor .github/` locally: no findings.
- [x] CI zizmor job stays green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/actions/update-charm-pins/action.yaml
```diff
@@ -17,7 +17,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5
```
dwilding: Elsewhere we pin v6. Are we going to let dependabot bump this?

dwilding (approved):

.github/actions/update-charm-pins/action.yaml
tonyandrewmeyer: Odd that this one is different to the rest in the repo. Like dependabot missed it somehow?

james-garner-canonical (approved):

PR #2614, opened 2026-07-04
fix: do not attempt to mutate secret data in grant/revoke during tests

### LINKED ISSUE #2613: ops.testing: revoking a previously-granted secret fails on 3.8.0 (worked on 3.7.1)
While evaluating the upgrade to ops 3.8.0, one of my charm unit tests started failing. 
The test is a charm revoking a secret grant when a relation goes away. 
The same test passes on ops 3.7.1 and fails on ops 3.8.0
(which pulls in ops-scenario 8.8.0 via the testing extra).

I've asked Claude to reduce it to the minimal self-contained reproduction below. 

Let me know what you think.

This is the output I get `AttributeError: 'frozenset' object has no attribute 'remove'`

```
# repro_revoke.py
import ops
from ops import testing


class RevokeCharm(ops.CharmBase):
        super().__init__(framework)
        framework.observe(self.on["db"].relation_broken, self._on_broken)

    def _on_broken(self, event):
        # Revoke the secret's grant to the relation that is going away.
        secret = self.model.get_secret(label="my-secret")
        secret.revoke(event.relation)


def test_revoke_on_relation_broken():
    ctx = testing.Context(
        RevokeCharm,
        meta={"name": "revoke-charm", "provides": {"db": {"interface": "x"}}},
    )
    secret = testing.Secret(
        {"token": "s3cret"},
        owner="app",
        label="my-secret",
        remote_grants={1: {"remote-app"}},  # relation 1 was previously granted
    )
```

tonyandrewmeyer:
Audit of the rest of `mocking.py` for other places affected by the state-type refactor in #2331:

- `open_port`/`close_port`, `secret_add`/`secret_remove`: already copy the frozenset to a mutable set, mutate, then reassign via the `_update_*` helpers.
- `_update_state_check_infos`: reassigns `check_infos` via `object.__setattr__` with a fresh `frozenset`.
- `secret_set`: goes through `Secret._update_metadata`, which uses `object.__setattr__`.
- Relation data (`local_app_data`, etc.) and Container mappings (`layers`, `service_statuses`, `mounts`): annotations were tightened to `Mapping`, but the state classes still construct plain `dict`s in `__init__`, so in-place mutation is still valid at runtime.

`remote_grants` was the only field that actually flipped from `set` to `frozenset` at runtime, and `secret_grant`/`secret_revoke` were the only callers assuming mutability.

Pyright would have caught this — running it directly on the pre-fix code flags `Cannot access attribute "add"/"remove" for class "frozenset[str]"` — but `tool.pyright.include` in `pyproject.toml` uses `testing/src/*.py`, which does not match `testing/src/scenario/*.py`, so the scenario source is silently unchecked. I'll fix that in a separate PR.

Fixes #2613


testing/src/scenario/mocking.py
```diff
@@ -583,10 +583,10 @@ def secret_grant(self, id: str, relation_id: int, *, unit: str | None = None):
...
         )
 
-        if not secret.remote_grants.get(relation_id):
-            secret.remote_grants[relation_id] = set()
-
-        secret.remote_grants[relation_id].add(cast('str', grantee))
+        secret.remote_grants[relation_id] = secret.remote_grants.get(
+            relation_id,
+            frozenset(),
+        ) | {cast('str', grantee)}
```
james-garner-canonical: This feels odd, because the `frozenset()` default argument to `get` is explicitly never used. We also don't typically use `|` like this in Ops, or even `or`, we tend to prefer an explicit `if`. As a very minor note also, the styling choice here is quite odd -- why split the `get` arguments across multiple lines like this?

It also looks like we're introducing a behaviour change here. Previously we'd always add `grantee` to `remote_grants[relation_id]`. Now that only triggers if `remote_grants[relation_id]` is empty.

james-garner-canonical (changes requested):
I only looked at the first diff hunk, but there seem to be some issues worth highlighting there, which may or may not be relevant to the rest of the PR.

testing/src/scenario/mocking.py
tonyandrewmeyer: I've adjusted the style. You're right: I like the operators with sets but we use the named methods in ops.

I don't see a behaviour change, though. There's a union of either the grantees in the old object or the empty set with the grantee. Either way, the union result will include the grantee. There's also a test covering the behaviour that I would expect to fail if it was changed.

james-garner-canonical: Ah, yeah I think this points to one downside of operators: I completely missed that this was a set union operation and interpreted it as `or` (which of course `|` is not an alias for in Python). You're right that the union operation preserves the previous semantics.

testing/src/scenario/mocking.py
```diff
@@ -583,10 +583,11 @@ def secret_grant(self, id: str, relation_id: int, *, unit: str | None = None):
...
 
-        if not secret.remote_grants.get(relation_id):
-            secret.remote_grants[relation_id] = set()
-
-        secret.remote_grants[relation_id].add(cast('str', grantee))
+        grants = secret.remote_grants
+        if relation_id in grants:
+            grants[relation_id] = grants[relation_id].union({cast('str', grantee)})
+        else:
+            grants[relation_id] = frozenset({cast('str', grantee)})
```
james-garner-canonical: With the knowledge that this is a `union` operation, and if we use an `assert` (or `overload`) to eliminate the `cast`, I actually like the `get` approach you implemented earlier -- my comment about preferring `if` was motivated by my misreading of `|` as `or`.
```suggestion
        assert grantee is not None  # guaranteed by _raise_on_error=True
        grants = secret.remote_grants
        grants[relation_id] = grants.get(relation_id, frozenset()).union({grantee})
```

testing/src/scenario/mocking.py
```diff
@@ -596,9 +597,10 @@ def secret_revoke(self, id: str, relation_id: int, *, unit: str | None = None):
             relation_id,
             _raise_on_error=True,
         )
-        secret.remote_grants[relation_id].remove(cast('str', grantee))
-        if not secret.remote_grants[relation_id]:
-            del secret.remote_grants[relation_id]
+        grants = secret.remote_grants
+        grants[relation_id] = grants[relation_id].difference({cast('str', grantee)})
+        if not grants[relation_id]:
+            del grants[relation_id]
```
james-garner-canonical, line 602: Dropping the (now double) `cast` would be nice here too:
```suggestion
        assert grantee is not None  # guaranteed by _raise_on_error=True
        grants = secret.remote_grants
        grants[relation_id] = grants[relation_id].difference({grantee})
```

testing/src/scenario/mocking.py
james-garner-canonical, line 581: Unrelated to this PR, but it seems we treat `unit=''` and `unit=None` synonomously in these methods. I suppose that's harmless.

james-garner-canonical (approved):
Thanks for clarifying, the original PR was semantically correct. I appreciate the effort to tidy up the code -- the `frozenset` makes the operations trickier to express clearly, and I actually like your original approach a lot.

The code in `main` has an annoying `cast`, but just one, so it's not that bad, but here we end up needing to `cast` an extra time (and I somewhat blame the `cast` for my confusion when reading the original PR 😆). We can clean that up one of two ways: a local `assert` (see suggestions), or adding an `overload` to `relation_remote_app_name`. I'd be happy with either approach in this PR (I haven't investigated what's public vs private here and whether an `overload` would be appropriate in this case).

tromai (approved):
Thanks for the fix. 

I understand why the bug exists. And I'm okay with the direction of the fix. Big fan of using `get(..).union` instead of `|` here.

testing/src/scenario/mocking.py
tonyandrewmeyer: If you're granting in the framework, you pass an `ops.Unit` so it'll never be the empty string unless you've made the `Unit` yourself rather than getting it from the framework. In `hookcmds` we do treat them differently and would do a `--unit` by itself for the empty string and I'm pretty sure Juju errors in that case. So we could copy that behaviour here but I think it would be better in a different PR.

PR #2615, opened 2026-07-04
test: type-check testing/src/scenario and fix uncovered errors

tonyandrewmeyer:
The pyright `include` glob `testing/src/*.py` did not match the subdirectory where the scenario source actually lives (`testing/src/scenario/*.py`), so that tree was silently unchecked 😞 😢. Add the correct glob and fix the 122 (!) errors that surfaced.

Highlights of the fixes:

- **mocking.py** — use `dict(x)` instead of `.copy()` on `Mapping`-typed state fields; `cast` where the runtime is a `dict` but the annotation is intentionally `Mapping` (relation databags, `remote_grants`).
- **context.py** — drop dead `type: ignore` comments and default the derived `app_name` to `''` so the existing bad-metadata tests still exercise the downstream failure path.
- **state.py** — widen internal validator params to `Mapping`, split `from_context()`'s `**kwargs` TypedDict so it no longer overlaps with named parameters, annotate `default_factory=dict` fields with their element types, and replace the bare `''` assigned to `CheckInfo.change_id` (`ChangeID | None`) with `pebble.ChangeID('')`.
- **_runtime.py** — replace deprecated `Iterator` return type with `Generator`, parameterise `TemporaryDirectory`, normalise Optional `unit_id`.
- **_consistency_checker.py** — cast action-spec lookups through `Any` so pyright can resolve `get()`/`items()` types.
- **_ops_main_mock.py** — initialise the loop variable before the `try` so the `except` branch never sees an unbound name.

Refs #2613


testing/src/scenario/_ops_main_mock.py
```diff
@@ -34,6 +34,9 @@
     StoredState,
 )
 
+if TYPE_CHECKING:
+    from opentelemetry.sdk.trace import ReadableSpan
+
 if TYPE_CHECKING:  # pragma: no cover
     from .state import State, _CharmSpec, _Event
```
dwilding: Can these be merged? Why does one have `pragma: no cover` but the other not?

testing/src/scenario/context.py
```diff
@@ -753,8 +756,12 @@ def __init__(
                 'Juju 2.x is closed and unsupported. You may encounter inconsistencies.',
             )
 
-        self.app_name = app_name or self._charm_spec.meta.get('name')
-        self.unit_id = unit_id
+        # If neither ``app_name`` nor a ``name`` in the charm metadata is
+        # provided, fall back to an empty string and let the charm code fail
+        # later. Raising here would prevent tests from asserting on the
+        # downstream failure path (see ``test_init_and_run_with_bad_meta``).
```
dwilding: By "let the charm code fail", do we mean that the charm code is guaranteed to fail, because of something that Ops tries to do? That's how I interpret the downstream failure path. Just want to check I got the right understanding.

testing/src/scenario/state.py
```diff
@@ -48,20 +48,17 @@
 
     from . import Context
 
-    class _StateKwargs(TypedDict, total=False):
-        config: Mapping[str, str | int | float | bool]
-        relations: Iterable[RelationBase]
+    # Fields that from_context() forwards through **kwargs — the ones it
+    # doesn't already accept explicitly to merge with metadata-derived defaults.
```
dwilding: The comment wording is a bit confusing. I'm not sure what "to merge with metadata-derived defaults" is meant to apply to. Is that talking about the fields that `from_context` accepts or the ones that `from_context` doesn't accept?

testing/src/scenario/mocking.py
```diff
@@ -941,12 +944,12 @@ def _container(self) -> ContainerSpec:
...
-    def _layers(self) -> dict[str, pebble.Layer]:
-        return self._container.layers
+    def _layers(self) -> dict[str, pebble.Layer]:  # pyright: ignore[reportIncompatibleVariableOverride]
+        return cast('dict[str, pebble.Layer]', self._container.layers)
 
     @property
-    def _service_status(self) -> dict[str, pebble.ServiceStatus]:
-        return self._container.service_statuses
+    def _service_status(self) -> dict[str, pebble.ServiceStatus]:  # pyright: ignore[reportIncompatibleVariableOverride]
+        return cast('dict[str, pebble.ServiceStatus]', self._container.service_statuses)
```
dwilding: I think I get the point of these. Is this a good-enough workaround for something that really should be changed in the design in the future? If no design change is needed, do you think it's worth adding comments to discourage people from trying to remove the ignores in the future?

dwilding (approved):
I bet finding this was a bit of a shock. Thanks for fixing!

testing/src/scenario/_ops_main_mock.py
tonyandrewmeyer: Oops, yes. The pragma probably always ought to be there since coverage won't ever hit this, but it's not really important either way, particularly since w eons't pay a lot of attention to coverage.

testing/src/scenario/context.py
tonyandrewmeyer: The comment is badly put. I'll fix.

testing/src/scenario/state.py
tonyandrewmeyer: I've reworded, is it clearer now?

testing/src/scenario/mocking.py
tonyandrewmeyer: I'll add a comment. I think the best fix will be when we move Harness so it's not in ops by default, and Scenario presumably at that point stops building on top of it.

testing/src/scenario/state.py
dwilding: It's clearer now - thanks

testing/src/scenario/state.py
```diff
@@ -1164,7 +1164,7 @@ def __init__(
         object.__setattr__(self, 'threshold', threshold)
         if change_id is None:
             if self.status == pebble.CheckStatus.INACTIVE:
-                change_id = ''
```
tromai, line 1167: This is a good catch 😨.

testing/tests/test_e2e/test_secrets.py
```diff
@@ -501,6 +501,34 @@ class GrantingCharm(ops.CharmBase):
         output.get_secret(label='mylabel')
 
 
+def test_revoke_preexisting_remote_grant():
```
tromai, line 504: I'm curious. Does running the type-checker surface an issue with `remote_grant`, and lead to this test?

Or is this test a drive-by thingy?

tromai (approved):
Thanks for the PR!

testing/tests/test_e2e/test_secrets.py
tromai: Ah sorry, it seems like this test is for https://github.com/canonical/operator/issues/2613, which I haven't looked at it yet :sweat_smile:

testing/src/scenario/_ops_main_mock.py
```diff
@@ -244,6 +246,7 @@ def _get_event_args(self, bound_event: ops.BoundEvent) -> tuple[list[Any], dict[
     def _get_owner(root: Any, path: Sequence[str]) -> ops.ObjectEvents:
         """Walk path on root to an ObjectEvents instance."""
         obj = root
+        step = ''
```
james-garner-canonical: This is a false positive, so maybe an ignore is more appropriate -- we'll get alerted that the ignore is unnecessary if the check every improves.

testing/src/scenario/_consistency_checker.py
```diff
@@ -792,7 +794,7 @@ def check_storedstate_consistency(
         # We don't need the marshalled state, just to know that it can be.
         # This is the same "only simple types" check that ops does.
         try:
-            marshal.dumps(ss.content)
+            marshal.dumps(dict(ss.content))
```
james-garner-canonical: We type `StoredState.content` as `Mapping` but I think it's probably always actually a `dict`. Actually constructing a new `dict` here just to satisfy the type checker seems wasteful. If a non-`dict` `Mapping` would actually fail at runtime, then maybe an `isinstance` check would be more appropriate here. Otherwise an `ignore` that goes away if the `marshal.dumps` type stubs improve might be best.

testing/src/scenario/_ops_main_mock.py
```diff
@@ -279,7 +282,7 @@ def destroy(self):
             self.framework.close()
         if self._tracing_mock:
             assert self._tracing_exporter
-            self.trace_data = self._tracing_exporter.get_finished_spans()
+            self.trace_data = list(self._tracing_exporter.get_finished_spans())
```
james-garner-canonical: `get_finished_spans` returns a `tuple`, so our current runtime behaviour is to instantiate `trace_data` as an empty list, and then set it to a tuple of spans. Since this is the live behaviour, presumably having a tuple instead of a list isn't breaking anyone, so I'd be hesitant to introduce a list here.

The correct fix might be one of:
- Type `self.trace_data` as `Sequence[...]` explicitly in `__init__`.
- Instantiate `trace_data` as `()` instead of `[]`.

testing/src/scenario/_runtime.py
```diff
@@ -62,7 +62,7 @@ def __init__(
         self._charm_root = charm_root
 
         self._app_name = app_name
-        self._unit_id = unit_id
+        self._unit_id = 0 if unit_id is None else unit_id
```
james-garner-canonical, line 65: This seems reasonable since `_unit_id` is private, and is a nice normalization.

testing/src/scenario/context.py
```diff
@@ -101,7 +101,7 @@ def charm(self) -> CharmType:
             raise RuntimeError(
                 'you should __enter__ this context manager before accessing this',
             )
-        return self.ops.charm
+        return cast('CharmType', self.ops.charm)
```
james-garner-canonical: Rather than introducing a `cast` here to hide this issue, perhaps it would be better to ensure that `CharmType` is properly piped through the `_Manager` class the same way it is through the `Context`. Alternatively if that's out of scope, I'd prefer an `ignore` here, since that will get detected as unnecessary and be removed if we ever do make `_Manager.charm` type aware, whereas the `cast` may linger forever and hide real typing issues in future.

testing/src/scenario/context.py
```diff
@@ -722,7 +722,10 @@ def __init__(
         if not any((meta, actions, config)):
             logger.debug('Autoloading charmspec...')
             try:
-                spec: _CharmSpec[CharmType] = _CharmSpec.autoload(charm_type)
+                spec = cast(
+                    '_CharmSpec[CharmType]',
+                    _CharmSpec.autoload(charm_type),
+                )
```
james-garner-canonical: It looks like `_CharmSpec` and `autoload` are already generic, so we shouldn't get an error here -- if we are, we presumably have a bug. The `cast` will hide the bug indefinitely, so I'd prefer to either address it in this PR or use an `ignore` comment and open an issue to investigate it.

testing/src/scenario/context.py
```diff
@@ -753,8 +756,13 @@ def __init__(
...
             )
 
-        self.app_name = app_name or self._charm_spec.meta.get('name')
-        self.unit_id = unit_id
+        # If neither ``app_name`` nor a ``name`` in the charm metadata is
+        # provided, fall back to an empty string. Raising here would prevent
+        # tests from asserting on the downstream failure that Ops raises when
+        # it later looks up the missing metadata (see
+        # ``test_init_and_run_with_bad_meta``).
+        self.app_name = app_name or self._charm_spec.meta.get('name') or ''
```
james-garner-canonical: Any reason why not?
```suggestion
        self.app_name = app_name or self._charm_spec.meta.get('name', '')
```

testing/src/scenario/context.py
```diff
@@ -753,8 +756,13 @@ def __init__(
...
+        self.unit_id = 0 if unit_id is None else unit_id
```
james-garner-canonical, line 762: Behaviour change but reflects the current attribute typing, so a good one I think.

testing/src/scenario/mocking.py
```diff
@@ -181,7 +181,7 @@ def get_pebble(self, socket_path: str) -> Client:
         ]  # /charm/containers/<container_name>/pebble.socket
         container_root = self._context._get_container_root(container_name)
         try:
-            mounts = self._state.get_container(container_name).mounts
+            mounts = dict(self._state.get_container(container_name).mounts)
```
james-garner-canonical: Looks like this is necessary because `_MockPebbleClient` expects a `dict`, but the only thing it does with `mounts` is iterate over the values. It would be cleaner to change the argument typing to `Mapping` instead of doing a runtime `dict` call here.

testing/src/scenario/mocking.py
```diff
@@ -268,7 +268,7 @@ def relation_get(
         data = self._relation_get(
             relation_id, member_name=member_name, is_app=is_app, relation_name=relation_name
         )
-        return data.copy()
+        return dict(data)
```
james-garner-canonical, line 272: AFAICT by following the call chain, this is already a `dict`, so I don't understand the motivation for this change.

testing/src/scenario/mocking.py
```diff
@@ -334,8 +334,8 @@ def relation_list(
         remote_name = self.relation_remote_app_name(relation_id, relation_name=relation_name)
         return tuple(f'{remote_name}/{unit_id}' for unit_id in relation._remote_unit_ids)
 
-    def config_get(self):
-        state_config = self._state.config.copy()  # dedup or we'll mutate the state!
+    def config_get(self) -> dict[str, str | int | float | bool]:
+        state_config = dict(self._state.config)  # dedup or we'll mutate the state!
```
james-garner-canonical, line 339: This also seems to be a dictionary already. I haven't looked at the other `...copy()` -> `dict(...)` translations below.

testing/src/scenario/mocking.py
```diff
@@ -583,10 +583,11 @@ def secret_grant(self, id: str, relation_id: int, *, unit: str | None = None):
...
 
-        if not secret.remote_grants.get(relation_id):
-            secret.remote_grants[relation_id] = set()
-
-        secret.remote_grants[relation_id].add(cast('str', grantee))
+        remote_grants = cast('dict[int, frozenset[str]]', secret.remote_grants)
+        remote_grants[relation_id] = remote_grants.get(
+            relation_id,
+            frozenset(),
+        ) | {cast('str', grantee)}
```
james-garner-canonical: Overlaps with #2614, take care when merging.

testing/src/scenario/mocking.py
```diff
@@ -596,9 +597,10 @@ def secret_revoke(self, id: str, relation_id: int, *, unit: str | None = None):
             relation_id,
             _raise_on_error=True,
         )
-        secret.remote_grants[relation_id].remove(cast('str', grantee))
-        if not secret.remote_grants[relation_id]:
-            del secret.remote_grants[relation_id]
+        remote_grants = cast('dict[int, frozenset[str]]', secret.remote_grants)
+        remote_grants[relation_id] = remote_grants[relation_id] - {cast('str', grantee)}
+        if not remote_grants[relation_id]:
+            del remote_grants[relation_id]
```
james-garner-canonical: Overlaps with #2614, take care when merging.

testing/src/scenario/mocking.py
```diff
@@ -940,13 +943,20 @@ def _container(self) -> ContainerSpec:
...
+    # state on the Container dataclass rather than on the client. Pyright
+    # rightly flags overriding a variable with a differently-typed property,
+    # but resolving it properly would require reworking the parent class to
+    # let subclasses plug in a state source. Don't remove the ignores without
+    # that refactor.
     @property
-    def _layers(self) -> dict[str, pebble.Layer]:
-        return self._container.layers
+    def _layers(self) -> dict[str, pebble.Layer]:  # pyright: ignore[reportIncompatibleVariableOverride]
+        return cast('dict[str, pebble.Layer]', self._container.layers)
```
james-garner-canonical: Since this is private, can we adjust the return type to `Mapping` to avoid the `cast`? (I understand the `ignore` is unrelated and remains).

testing/src/scenario/mocking.py
```diff
@@ -940,13 +943,20 @@ def _container(self) -> ContainerSpec:
...
 
     @property
-    def _service_status(self) -> dict[str, pebble.ServiceStatus]:
-        return self._container.service_statuses
+    def _service_status(self) -> dict[str, pebble.ServiceStatus]:  # pyright: ignore[reportIncompatibleVariableOverride]
+        return cast('dict[str, pebble.ServiceStatus]', self._container.service_statuses)
```
james-garner-canonical: Since this is private, can we adjust the return type to `Mapping` to avoid the `cast`? (I understand the `ignore` is unrelated and remains).

testing/src/scenario/state.py
```diff
@@ -188,8 +188,8 @@ def __init__(
     def _to_ops(self) -> CloudCredential_Ops:
         return CloudCredential_Ops(
             auth_type=self.auth_type,
-            attributes=self.attributes,
-            redacted=self.redacted,
+            attributes=dict(self.attributes),
+            redacted=list(self.redacted),
```
james-garner-canonical, line 194: I suppose that we should really be updating the `ops.model.CloudCredential` (`frozen`) `dataclass` to use `Mapping` and `Sequence` typing, but that would be out of scope for this PR and has backwards compatibility concerns. I'm not sure if an `ignore` comment rather than a runtime conversion would be better in this case. Either way, a comment explaining the context would be a good idea IMO.

testing/src/scenario/state.py
```diff
@@ -263,7 +263,7 @@ def _to_ops(self) -> CloudSpec_Ops:
             identity_endpoint=self.identity_endpoint,
             storage_endpoint=self.storage_endpoint,
             credential=self.credential._to_ops() if self.credential else None,
-            ca_certificates=self.ca_certificates,
+            ca_certificates=list(self.ca_certificates),
```
james-garner-canonical, line 270: [Ditto](https://github.com/canonical/operator/pull/2615/changes#r3540535576).

testing/src/scenario/state.py
```diff
@@ -698,12 +696,12 @@ def __post_init__(self):
         # Deepcopy mutable fields to disassociate from the caller's objects.
         for attr, value in self.__dict__.items():
             if isinstance(value, (dict, list)):
-                object.__setattr__(self, attr, copy.deepcopy(value))
+                object.__setattr__(self, attr, copy.deepcopy(cast('Any', value)))
```
james-garner-canonical: I think a type ignore comment is a better fit here.

testing/src/scenario/state.py
```diff
@@ -1089,7 +1089,7 @@ def _to_ops(self) -> pebble.Notice:
             last_occurred=self.last_occurred,
             last_repeated=self.last_repeated,
             occurrences=self.occurrences,
-            last_data=self.last_data,
+            last_data=dict(self.last_data),
```
james-garner-canonical, line 1096: This is presumably good since it also prevents the same mutable `dict` instance being shared, though it is a behaviour change -- presumably not one that matters, in which case I'm all for it.

testing/src/scenario/state.py
```diff
@@ -2173,20 +2173,24 @@ def _apply_extensions(meta: dict[str, Any], extensions: list[str]) -> None:
             if key not in meta:
                 meta[key] = copy.deepcopy(ext_value)
             elif isinstance(ext_value, dict) and isinstance(meta[key], dict):
-                overlap = set(ext_value) & set(meta[key])
+                ext_dict = cast('dict[str, Any]', ext_value)
+                meta_dict = cast('dict[str, Any]', meta[key])
+                overlap = set(ext_dict) & set(meta_dict)
```
james-garner-canonical: It would be nice if this could also be an ignore, since it's an annoying limitation of the current intersection of `isinstance`, Pyright, and generics, where `isinstance` doesn't permit generics at runtime, so Pyright interprets the contents as `[Unknown]`, and there's no nice way to make it not do that. But unfortunately even an ignore doesn't prevent `Unknown` from tainting the `overlap` variable. But this change shows the danger of `cast` -- the `meta_dict` `cast` is completely unnecessary.

```suggestion
                ext_dict = cast('dict[str, Any]', ext_value)
                overlap = set(ext_dict) & set(meta[key])
```

testing/src/scenario/state.py
```diff
@@ -2173,20 +2173,24 @@ def _apply_extensions(meta: dict[str, Any], extensions: list[str]) -> None:
...
-                meta[key] = merged
+                merged_dict: dict[str, Any] = copy.deepcopy(ext_dict)
+                merged_dict.update(meta_dict)
+                meta[key] = merged_dict
             elif isinstance(ext_value, list) and isinstance(meta[key], list):
-                merged = copy.deepcopy(ext_value)
-                merged.extend(i for i in meta[key] if i not in merged)
-                meta[key] = merged
+                ext_list = cast('list[Any]', ext_value)
+                meta_list = cast('list[Any]', meta[key])
```
james-garner-canonical: Unnecessary `cast` like [above](https://github.com/canonical/operator/pull/2615/changes#r3540567209)?

james-garner-canonical (commented):
I was going to just do a drive-by comment when I spotted a couple of things, but I ended up going through all the Scenario changes (not the tests).

testing/src/scenario/mocking.py
tonyandrewmeyer: It's a `_RawDatabagContents`, which is a `Mapping` not a dict. So we need a cast or ignore or convert like this. We do know it will be a dict in practice (until eventually it is a frozendict, I assume). Would you rather a cast?

testing/src/scenario/mocking.py
tonyandrewmeyer: Yeah, this PR stacks on that one, so this should disappear. I needed this fix in place to get everything to pass, but (maybe wrongly) thought that one with the small scope would land quickly.

testing/src/scenario/mocking.py
tonyandrewmeyer: As above.

testing/src/scenario/mocking.py
james-garner-canonical: Ah I may have been looking at the wrong version of the code when I wrote this comment. The `dict` call seems reasonable.

testing/src/scenario/mocking.py
tonyandrewmeyer: Unfortunately I don't think we can. `_TestingPebbleClient` does `self._layers[label] = layer_obj` (and similar assignments to `_service_status`) in a handful of methods that the scenario mock inherits, so narrowing the return type to `Mapping` breaks those callers at type-check time. That's what the "resolving it properly would require reworking the parent class" bit in the comment above is meant to say.

I think when we get Harness out of ops we'll want to have Scenario not build on it and can perhaps clean some of this up then.

testing/src/scenario/mocking.py
tonyandrewmeyer: Same as above.

testing/src/scenario/mocking.py
james-garner-canonical: In that case I'd suggest an `ignore` and explanation so we get alerted when that's no longer necessary.

james-garner-canonical (approved):
Thanks for doing this, and for addressing my feedback. One minor comment where it wasn't possible to cleanly eliminate a `cast` by changing our internal typing, where we should probably use an `ignore` instead so we notice when it's no longer necessary.

Issue #2350, opened 2026-02-27
Changes to consider for next major release (4.x)


benhoyt:
Please add to this list as we go:

- [ ] Remove the "compatibility imports" from `ops.testing`, `ops.model`, etc (for example see https://github.com/canonical/operator/pull/2339)
- [ ] Things that have a deprecation warning/docstring (see comment below)
- [ ] Drop Python 3.10
- [ ] Drop Juju 2.9 / Juju < 3.6
- [ ] https://github.com/canonical/operator/pull/2361
- [ ] `SCENARIO_BARE_CHARM_ERRORS` default
- [ ] Bump min TLS to 1.3 in tracing
   - We'd want logs sinks to use Mozilla's generator
   - Case in point: https://github.com/canonical/traefik-k8s-operator/pull/535 
- [ ] Use TLS VERIFY_STRICT in tracing [matrix message](https://matrix.to/#/!yAkGlrYcBFYzYRvOlQ:ubuntu.com/$ibCVWod_F_dRl7afdHzfgDI5lGTbFhZI3jKtaQVbX50?via=ubuntu.com&via=matrix.org&via=laquadrature.net)
- [ ] Move Harness out to separate package? It could be a required dependency still for one or more versions. Pending deprecation warning for not being in extras if possible.
- [ ] Stop using pickle for storage serialisation? It's not good practice from a security point of view, theoretically you could shove something bad in the storage and get ops to execute it (quite theoretical, to be fair). Maybe a mid point where we can load existing pickled data but dump to something else (YAML? JSON? with some custom serialisers?) and drop pickle some time in the future?
- [x] #2333
- [ ] maybe we could change how Scenario handles the default keys in the databags, so the Juju 3 vs. Juju 4 awkwardness is handled better

dimaqq:
### Deprecations to consider
  - [ ] `ops.main.main(...)` legacy entrypoint is deprecated in favour of `ops.main()` ([ops/main.py](ops/main.py#L30-L39)).
  - [ ] Legacy `ops.main` import path is maintained for compatibility and re-exports internal symbols from `ops._main` (legacy/deprecated path context) ([ops/main.py](ops/main.py#L15-L27)).
  - [ ] Entire `ops.lib` functionality is deprecated ([ops/lib/\_\_init\_\_.py](ops/lib/__init__.py#L15-L21)).
[...]

PR #2618, opened 2026-07-05
fix: don't include private-address in default testing.Relation databags on Juju 4

### LINKED ISSUE #2185: Don't include private-address in the default Scenario database when mocking Juju 4
The `private-address` key is not included in the default relation settings (databag) as of Juju 4.0, so we should not include it in the mock one either, if the mock Juju version is 4.0+.

The default contents are currently set when a `Relation` (etc) object is created. We don't have access to the relevant `Context` object at that time (which is where the Juju version is), so we'll have to figure out what the best way of handling this is.

tonyandrewmeyer:
At exec time, remove the `private-address` key from the databags, if it's set to our default value and the mock Juju version is 4+.

I don't love this solution, since we're editing the State after the user has finished with it, but it's only when we have the Context that the version is available, so there doesn't seem to be a backwards-compatible alternative. It would probably be better if the runtime injected the automated keys and they weren't there when the user had the State, or maybe if you had to provide a version when getting a fresh Relation, but we can't do either of those in 8.x. This change at least gives the Juju behaviour, even if it's a bit quirky.

Regarding [the comment in the issue](https://github.com/canonical/operator/issues/2185#issuecomment-3781929516), what I am proposing is that we do this workaround so that the Scenario 8.x behaviour matches Juju 3 and 4, and we do something more elaborate and backwards incompatible in 9.0 (I added to the list in https://github.com/canonical/operator/issues/2350). We could instead reject this change and opt in to new behaviour with a env flag or similar, but I feel even if we do that we would want this change so that the default behaviour is the Juju one.

Fixes #2185


testing/tests/test_e2e/test_relations.py
```diff
@@ -487,6 +487,70 @@ def test_relation_default_unit_data_peer():
...
+    peer_rel = state_out.get_relation(peer.id)
+    sub_rel = state_out.get_relation(sub.id)
+    assert isinstance(rel, Relation)
+    assert isinstance(peer_rel, PeerRelation)
+    assert isinstance(sub_rel, SubordinateRelation)
+    assert ('private-address' in rel.local_unit_data) is expect_private_address
+    assert ('private-address' in rel.remote_units_data[0]) is expect_private_address
+    assert ('private-address' in peer_rel.local_unit_data) is expect_private_address
+    assert ('private-address' in sub_rel.local_unit_data) is expect_private_address
+    assert ('private-address' in sub_rel.remote_unit_data) is expect_private_address
```
dwilding, line 529: Do we prefer `is` or `==` for boolean comparison?

testing/src/scenario/state.py
```diff
@@ -716,12 +716,40 @@ def _validate_databag(self, databag: dict[str, str]):
...
     'egress-subnets': _DEFAULT_IP,
     'ingress-address': _DEFAULT_IP,
     'private-address': _DEFAULT_IP,
 }
 
+# Keys in `_DEFAULT_JUJU_DATABAG` that Juju 4 no longer auto-populates.
+_JUJU_4_REMOVED_DATABAG_KEYS = ('private-address',)
+
+
+def _strip_juju4_defaults(state: State, juju_version: str) -> None:
```
dwilding, line 736: Couldn't we use `JujuVersion` instead of custom logic to parse the version? (As in `mocking.py`)

testing/src/scenario/state.py
```diff
@@ -716,12 +716,40 @@ def _validate_databag(self, databag: dict[str, str]):
...
+    """
+    parts = juju_version.split('.', 1)
+    try:
+        major = int(parts[0])
+    except ValueError:
+        return
+    if major < 4:
+        return
+    for relation in state.relations:
+        for databag in relation._databags:
```
dwilding, line 745: My review agent tells me that this would remove `private-address` from `local_app_data` and `remote_app_data` - if someone has set `private-address` to the default value in those databags. I feel like we're safe to assume nobody would ever do that... What do you think?

dwilding (commented):
Main comment is about (lack of) `JujuVersion`

testing/tests/test_e2e/test_relations.py
tonyandrewmeyer: True and False are singletons (from 2.3 onwards), so `is` is correct. You don't see it a lot because normally just `assert b` or `assert not b` (or similar with `if`) is all that's needed.

dwilding: TIL, thanks

testing/src/scenario/state.py
tonyandrewmeyer: I did wonder about this. There's some chance a charm could do this to make their Juju 4 databag look like a Juju 3 one. I'd argue that it's not a good idea, but it's not impossible.

If someone does that their tests would break and they wouldn't be able to get around that.

The only ideas I have for avoiding that are ugly (like a special object we set instead, which we can recognise later, but pretends to be the normal one when people work with their state).

I think this comes down to a decision about whether we want to have a fix in Scenario 8 so the Juju 4 behaviour is surfaced, only have a fix with something opt-in (where we could do this more cleanly), or only fix it in 9.0 (definitely could do it clean, but people's tests may need updates). Or someone comes up with an idea I haven't thought of, of course.

I lean towards what's in the PR, but I can definitely be convinced we should go another way.

testing/src/scenario/state.py
tonyandrewmeyer: I feel like the value of `JujuVersion` is the properties, not parsing a very simple string. But sure, I can switch to that.

testing/src/scenario/state.py
dwilding: I'm OK with what's in this PR. My knowledge is a bit limited here, so no objection if you and James decide to go in a different direction.

dwilding (approved):

james-garner-canonical:
- How breaky is this PR?
- I'd be curious to see just how breaky changing the behaviour in the desired way instead would be: changing the default factories to provide empty dicts and populating the correct Juju-inserted keys at exec time (+/- having the Juju-inserted keys flow through to the output state).

I think it would be great to quantify the difference in breaky-ness so we can make an informed choice about whether this is the right fix for now.

PS: An idea for making this PR's approach safer -- track if the default factory was called with a private marker on the relation object, and only replace in that case. But maybe that falls apart as soon as we hit `dataclasses.replace` ...

tonyandrewmeyer:
> * How breaky is this PR?

In terms of running Hyrum? Or in my opinion given what code changes?

> * I'd be curious to see just how breaky changing the behaviour in the desired way instead would be: changing the default factories to provide empty dicts and populating the correct Juju-inserted keys at exec time (+/- having the Juju-inserted keys flow through to the output state).

This feels objectively more break-y to me, because it impacts Juju 3 as well as Juju 4, and also the keys that are still injected in both. But I could implement it and run it over the fleet tests to get actual numbers.

> PS: An idea for making this PR's approach safer -- track if the default factory was called with a private marker on the relation object, and only replace in that case. But maybe that falls apart as soon as we hit `dataclasses.replace` ...

Yeah, that sort of thing (as well as using a dict-subclass or something) is the kind of thing I was thinking of when I was saying there are "ugly" solutions. My preference would be to avoid those, and go with a clean solution (that I can't think of), the one in this PR, or an opt-in fix that becomes the default in 9.0.

james-garner-canonical:
> > * How breaky is this PR?
> 
> In terms of running Hyrum? Or in my opinion given what code changes?
> 
> > * I'd be curious to see just how breaky changing the behaviour in the desired way instead would be: changing the default factories to provide empty dicts and populating the correct Juju-inserted keys at exec time (+/- having the Juju-inserted keys flow through to the output state).
> 
> This feels objectively more break-y to me, because it impacts Juju 3 as well as Juju 4, and also the keys that are still injected in both. But I could implement it and run it over the fleet tests to get actual numbers.

You're probably right, but with how cheap implementation is I'd love to see the Hyrum numbers for the different approaches.

james-garner-canonical (commented):
Comment review to allow explicit re-review request after we have the Hyrum numbers.

tonyandrewmeyer:
Alternative implementation in https://github.com/canonical/operator/pull/2637 - the "inject at runtime" solution.

| Run | ops source | Passed | Pass rate |
| --- | --- | --- | --- |
| baseline | `canonical/operator@main` | 175/267 | 66% |
| databag | `tonyandrewmeyer/operator@rainy/2185-juju-default-databag-at-exec` | 172/267 | 64% |
| no-private | `tonyandrewmeyer/operator@rainy/2185-no-private-address-juju-4` | 178/267 | 67% |

`hyrum compare` against the main baseline reported 6 new failures for databag and 0 for no-private.

## Verdict

### `rainy/2185-no-private-address-juju-4` — clean

Zero new failures across 267 charms. The small change is because this branch is behind main.

### `rainy/2185-juju-default-databag-at-exec` — 5 real regressions

Each passes at the branch's own merge-base and fails on the branch.

| Charm | Failing test | Error |
| --- | --- | --- |
| alertmanager-k8s-operator | `test_log_forwarding.py::test_charm_active_on_logging_departed` | `InconsistentScenarioError` |
| cos-proxy-operator | `test_alerts.py::test_relation[1,5,10]` | `InconsistentScenarioError` ×3 |
| prometheus-k8s-operator | `test_receive_ca_cert.py::test_ca_forwarded_over_rel_data` | `InconsistentScenarioError` |
| trino-k8s-operator | `test_charm.py::test_trino_worker_relation_broken` | `InconsistentScenarioError` |
| microcluster-token-distributor-operator | `test_provides_lib.py::test_handle_mirror_mirror_hostnames` | `AssertionError` |

Two distinct consumer assumptions are being broken.

**1. Relation identity (4 charms).** Scenario's consistency checker rejects the event:

```
cannot emit monitors_relation_changed because relation 1 is not in the state
(a relation with the same ID is not sufficient - you must pass the object in
the state to the event)
```

These tests build a `Relation` for the event that is equal by ID to the one in the state but is not the same object.

**2. Default databag keys (1 charm).** `microcluster-token-distributor` asserts the exact key list of its own unit databag:

```python
assert list(cluster_relation.data[manager.charm.unit].keys()) == [...]
# Left contains 3 more items, first extra item: 'egress-subnets'
```

The branch adds Juju's defaults (`egress-subnets`, `ingress-address`, `private-address`), so the list grows. This is the branch working as intended; the charm relied on the databag holding only what it wrote.

tonyandrewmeyer:
So this method is definitely the more conservative and safer approach. The question is whether there is enough breakage that we wouldn't want to do the more correct change, probably after submitting PRs downstream.

james-garner-canonical:
Thanks for running the numbers! There are few enough failures that I'm definitely interested in if we can get away with doing the 'more correct' change after pre-emptively fixing up the charm tests.

> Run	ops source	Passed	Pass rate
> baseline	`canonical/operator@main`	175/267	66%
> databag	`tonyandrewmeyer/operator@rainy/2185-juju-default-databag-at-exec`	172/267	64%
> no-private	`tonyandrewmeyer/operator@rainy/2185-no-private-address-juju-4`	178/267	67%
> `hyrum compare` against the main baseline reported 6 new failures for databag and 0 for no-private.

The apparently 3 new passes in this branch jumped out at me. Do you know why they are?
 
> **1. Relation identity (4 charms).** Scenario's consistency checker rejects the event:
> 
> ```
> cannot emit monitors_relation_changed because relation 1 is not in the state
> (a relation with the same ID is not sufficient - you must pass the object in
> the state to the event)
> ```
> 
> These tests build a `Relation` for the event that is equal by ID to the one in the state but is not the same object.

I had a look through #2637, and I'm struggling to understand how this category of test failure is a consequence of it. Could you explain, please?
 
> **2. Default databag keys (1 charm).** `microcluster-token-distributor` asserts the exact key list of its own unit databag:
> 
> ```python
> assert list(cluster_relation.data[manager.charm.unit].keys()) == [...]
> # Left contains 3 more items, first extra item: 'egress-subnets'
> ```
> 
> The branch adds Juju's defaults (`egress-subnets`, `ingress-address`, `private-address`), so the list grows. This is the branch working as intended; the charm relied on the databag holding only what it wrote.

This does seem like the kind of breakage we'd expect, and seems easy enough to provide a fix for.

james-garner-canonical (commented):
Comment review so you can re-request review after addressing the comment above.

tonyandrewmeyer:
> I had a look through #2637, and I'm struggling to understand how this category of test failure is a consequence of it. Could you explain, please?

It's the Relation equality, rather than anything about the databag contents. `RelationBase.__hash__` is `hash(self.id)`, but `__eq__` is the dataclass-generated one, and the consistency checker does `event.relation not in state.relations`, so the set membership matches on the ID and then falls through to `__eq__`.

On main, a freshly constructed Relation already has the Juju default keys in `local_unit_data`, so it still compares equal to the same relation in an output state that the charm hasn't written to. On this branch, construction gives you `{}` and the keys only appear in the output state, so it doesn't.

With a charm that does nothing at all (like CharmBase):

```python
rel = Relation('foo')
s1 = ctx.run(ctx.on.relation_created(rel), State(relations={rel}))
ctx.run(ctx.on.relation_joined(rel), s1)  # InconsistentScenarioError here
```

That passes on main and raises on this branch. It's what alertmanager's add_relation_sequence does (tests/unit/helpers.py), and that helper already works around the same problem on the next line, passing state_after_relation_joined.get_relation(relation.id) to relation_changed rather than the original object.

So it isn't really "the test depends on the default databag contents" - it's that reusing a Relation you constructed across successive ctx.run calls stops working as soon as the output state grows keys that your object doesn't have. We could fix that by matching on id in the checker, by keeping the injected keys out of the returned state, or downstream with get_relation.

tonyandrewmeyer:
Discussed this with @james-garner-canonical in a call just now. We both prefer to go with the #2637 approach - it's not so much more breaking than this way, and it's definitely the "more correct" way to do things, so let's try to get there rather than going halfway.

One of the remaining issues (the one mentioned in the comment from earlier today) is that we have Scenario add in the keys but don't remove them, so the output state's different in a way that some charm tests don't expect. We agree that we probably should remove them, which leads to the question about what if they were already there, and we feel that probably should be an error or warning. That then leads to the issue that some of the time that they are there, it's because they should have a specific address not the one fixed address that Scenario adds: really that should be Scenario using the Network from the State (or something similar, I need to refresh my knowledge of Juju spaces) rather than always using the fixed one.

So we'll reject this, and continue in the other PR, and go as far towards "correct" as possible, and see what the breaking situation is. If it's small, we can open some upstream PRs and go ahead, particularly now we're much more confident that we know about most existing charms, and if it's large we'll consider which pieces we can have and which we cannot.

PR #2619, opened 2026-07-06
docs: add howto configure jubilant logs

tromai:
This PR is a follow up from:
- https://github.com/canonical/jubilant/pull/351
- https://github.com/canonical/pytest-jubilant/pull/106

This PR add the following content into `howto/write-integration-tests-for-a-charm.md`:
- How to configure Jubilant log modes from `pyproject.toml`
- Update the `Generate crash dumps` section to mention the new `pytest-jubilant` logger object.

There is one small change in `howto/set-up-continuous-integration-for-a-charm.md` because it uses the GitHub Action workflow example in `howto/write-integration-tests-for-a-charm.md`.


docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,159 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+```diff
+  # In your integration test job
+-  - run: tox -e integration
++  - run: tox -e integration -- --juju-dump-logs logs
   - name: Upload logs
     if: ${{ !cancelled() }}
     uses: actions/upload-artifact@v7
     with:
-      name: juju-dump-logs
+      name: integration-test-logs
```
tromai, line 632: I rename to keep it general, as this artifact now contain both the log file generated by pytest, and the files generated by `pytest-jubilant`.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,159 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
 > - [`pytest | How to invoke pytest`](https://docs.pytest.org/en/7.1.x/how-to/usage.html)
 > - [](#validate-your-charm-with-every-change)
 
+(write-integration-tests-for-a-charm-configuring-jubilant-logs)=
+## Configuring Jubilant logs
```
dwilding: Recommend changing for consistency with the rest of the doc (which is our usual style for headings)

```suggestion
(write-integration-tests-for-a-charm-configure-jubilant-logs)=
## Configure Jubilant logs
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,159 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+...
+```
+
+You can freely configure it to your needs.
+
+> See more:
+> - [`pytest | How to manage loggings`](https://docs.pytest.org/en/stable/reference/reference.html#logging)
+> - [`pytest` | pyproject.toml](https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml)
+
+The sections below cover the configuration that we suggest.
```
dwilding: I'd reduce the intro. I usually assume that people are going to skip over section intros and jump straight to a subsection, so I try to avoid putting any crucial details in the intro. In this case, I don't think there are any crucial details, as the structure of `pyproject.toml` is shown well in the examples in the subsections.

How about something like this:


```suggestion
Jubilant emits logs during the integration tests. If you're using `pytest-jubilant`, these logs are surfaced by pytest. You can configure the logging behaviour in the `[tool.pytest.ini_options]` table in `pyproject.toml`.
```

I recommend changing "handled by" to "surfaced by", as handled may come across as a bit vague. Maybe surfaced is not quite right either - feel free to suggest a better word!

I recommend removing the links because I couldn't see how they immediately apply. I think your examples are more helpful, so the links could be distracting. If a link would help at a specific point, I think it would be better to add it there.

Also changed "behavior" to "behaviour" as Ops is still predominantly NZ/AU/UK English even though Canonical's preference is US English.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,159 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+### Brief mode
```
dwilding: No problem having a subsection for each mode. Since the doc already has a long TOC, I might consider collapsing them into a single subsection called "Logging modes" (or similar). But I haven't tried combining them myself, so fair enough if you feel that separate sections works better.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,159 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+  - run: tox -e integration
+  - name: Upload logs
+    if: ${{ !cancelled() }}
+    uses: actions/upload-artifact@v7
+    with:
+      name: integration-test-logs
+      path: logs
+```
+
+### Behavior when no pytest's logging config is set
```
dwilding: Could this be simplified to "Default behaviour"? (Note the non-US spelling 😁)

docs/howto/write-integration-tests-for-a-charm.md
dwilding: If you have a single section for logging modes, I think it would be reasonable to mention that this can also be controlled by CLI arguments, and provide a link. Similarly in the "Log to a file" section. I wouldn't mind the potential duplication of links, because it's likely that people will skip directly to the subsection they're interested in without necessarily reading the other subsections.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,159 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+This section covers the behavior when no logging configuration is set, either in config files, or from CLI arguments. In this case, pytest captures all log messages at `WARNING` level or above, and there is no logging to a file. You can still see messages from:
```
dwilding: I generally wouldn't say what the section covers - I'd get straight into things. Something like this:


```suggestion
If no logging configuration is set by `tool.pytest.ini_options` or pytest CLI arguments, pytest captures all log messages at `WARNING` level or above and doesn't log to a file. You can still see messages from:
```

dwilding (commented):
I haven't finished reviewing, but need to stop for today. Sending my initial comments in case they are helpful.

docs/howto/write-integration-tests-for-a-charm.md
tromai: > I recommend changing "handled by" to "surfaced by", as handled may come across as a bit vague. Maybe surfaced is not quite right either - feel free to suggest a better word!

I agree, `handled` doesn't indicate that the log is shown to the terminal. How about "display" as used in the [pytest documentation](https://docs.pytest.org/en/stable/how-to/logging.html#how-to-manage-logging) ?

I also made a mistake in:
```
These logs are handled by pytest if you are using `pytest-jubilant`
```

Jubilant logs are captured and displayed by `pytest`, even if `pytest-jubilant` is not used. I know we suggest `pytest-jubilant`, but I can't shake off the feeling that this sentence can give a wrong impression: "the log is *only* shown when pytest-jubilant is used".

May I use this version, based on your suggestion ?

```
Jubilant emits logs during the integration tests. 
These logs are captured and displayed by pytest.
You can configure the logging behaviour in the `[tool.pytest.ini_options]` table in `pyproject.toml`.
```

I love your short introduction. I think it's better to focus on what we suggest, and not give extra information that are slightly out of scope (for example, the `You can freely configure it to your needs.` part and its links).

docs/howto/write-integration-tests-for-a-charm.md
tromai: Yep! `Default behaviour` is a good title.

I will also go through and fix all US spellings too :smiley:

docs/howto/write-integration-tests-for-a-charm.md
tromai: That works way better. Thank you.

And I noticed something: the word pytest in this PR is not surrounded by backticks. I will use backticks for them.

dwilding: Personally I'm OK without backticks around pytest, unless we're referring to the `pytest` command or package, as it's the name of the whole framework. Similar to Ops vs `ops`.

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Your version looks good!

docs/howto/write-integration-tests-for-a-charm.md
tromai: Ah I see.

In this PR, pytest is mentioned as the name of the whole framework. I will not add the backticks in.

docs/howto/write-integration-tests-for-a-charm.md
tromai: I think collapsing the modes is a good idea. I will give that a try.

Each log mode subsection is not too big, so keeping them together is okay. I will put each mode in bold to separate them

```
### Logging modes
**Brief**
**Verbose**
**Error**
### Log to a file
```

tromai:
Commit https://github.com/canonical/operator/pull/2619/commits/d69899c493b39f46c0ab5ff36396d828c5295cc8 assumes the changes from https://github.com/canonical/pytest-jubilant/pull/106.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+      path: logs
+```
+
+### Default behaviour
+
+If no logging configuration is set by `tool.pytest.ini_options` or pytest CLI arguments, pytest captures all log messages at `WARNING` level or above and doesn't log to a file. You can still see messages from:
+
+- pytest. For example: `tests/integration/test_charm.py::test_deploy PASSED`.
+- other modules if they are at `WARNING` or above.
+- `pytest-jubilant`. For example: `Models were torn down...`.
```
tonyandrewmeyer: @dwilding might override me on style here.

```suggestion
- pytest, for example: `tests/integration/test_charm.py::test_deploy PASSED`;
- other modules if they are at `WARNING` or above;
- and `pytest-jubilant`, for example: `Models were torn down...`.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+    if: ${{ !cancelled() }}
+    uses: actions/upload-artifact@v7
+    with:
+      name: integration-test-logs
+      path: logs
+```
+
+### Default behaviour
+
+If no logging configuration is set by `tool.pytest.ini_options` or pytest CLI arguments, pytest captures all log messages at `WARNING` level or above and doesn't log to a file. You can still see messages from:
```
tonyandrewmeyer:
```suggestion
If no logging configuration is set by `tool.pytest.ini_options` or pytest CLI arguments, pytest captures all log messages at `WARNING` level or above and doesn't log to a file. You will still see messages from:
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+- pytest. For example: `tests/integration/test_charm.py::test_deploy PASSED`.
+- other modules if they are at `WARNING` or above.
+- `pytest-jubilant`. For example: `Models were torn down...`.
+
+(write-integration-tests-for-a-charm-view-juju-logs)=
+## View `juju debug-log` logs
+
+Save the complete `juju debug-log` logs to disk with the `--juju-dump-logs` option from `pytest-jubilant`. For example:
```
tonyandrewmeyer: I think we should link out to pytest-jubilant's docs here, probably with an inline link, but otherwise with a "see more".

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+```text
+tox -e integration -- --juju-dump-logs logs
+```
+
+Use `--juju-dump-logs` in CI with `actions/upload-artifact` to make `juju debug-log` files available as build artifacts. We can modify the GitHub Actions workflow from [](write-integration-tests-log-to-a-file):
```
tonyandrewmeyer:
```suggestion
Use `--juju-dump-logs` in CI with `actions/upload-artifact` to make `juju debug-log` files available as build artifacts. Modify the GitHub Actions workflow from [](write-integration-tests-log-to-a-file):
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+```diff
```
tonyandrewmeyer: I don't mind the diff here, but it seems like it might be simpler to not show it at all and just tell people to add the option as it is above.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+```toml
+[tool.pytest.ini_options]
+log-file-mode = "a"
+...
+```
+
+These options can also be configured from the command line.
+
+> See more:
```
tonyandrewmeyer: I think we're moving away from using quotes for these blocks. That means we're a bit inconsistent at the moment, but we should avoid adding new ones.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+`log_level = "INFO"` is intentional to keep the `Captured log call` consistent with Brief mode. If you want Verbose mode in `Captured log call`, you can set `log_level` to `DEBUG`.
+
+The local file where Verbose mode logs land is `logs/verbose.log`. It contains logs from one pytest session. If you want to run multiple test suites, and store Verbose logs in separated files. You can override `log_file` for each pytest invocation:
+
+```
+pytest --log-file "run1.log" ...
+pytest --log-file "run2.log" ...
+```
+
+Alternatively, you can set `log-file-mode` to `append`. Pytest will put verbose logs from all invocations into one file:
```
tonyandrewmeyer: I believe pytest styles its name as always lowercase.


```suggestion
Alternatively, you can set `log-file-mode` to `append`. pytest will put verbose logs from all invocations into one file:
```

tonyandrewmeyer (commented):
I didn't follow the PRs to adjust the logging earlier, so I might have missed some discussion here, sorry. Do we want to describe these "Brief" and "Verbose" styles here, and do we refer to them by those names anywhere else? It seems like we are just describing how we'd recommend configuring stream and file output, and the names just make it confusing. But if we are using them elsewhere or they have wider meaning, then that would be different.

I think it would be good to have some example output for each of the cases, so people know what kind of content they should get, particularly with different log levels (the format is just standard logging and should be fairly obvious).

In Jubilant or pytest-jubilant logs do we have somewhere information about what gets logged to each level? If so, it would be good to link to that. If not, maybe this is the place for that, although it seems like it belongs in Jubilant.

tromai:
@tonyandrewmeyer . Thanks for the questions.

> Do we want to describe these "Brief" and "Verbose" styles here, and do we refer to them by those names anywhere else? 

`Brief` and `Verbose` are only defined as names for the modes, and they are not mentioned anywhere else. They originated from the Spec, and followed the CLI standard. I was hoping the names can easily communicate what each mode is. But thinking again, it doesn't give any benefit in this context, and can be confusing. Instead of a name, may be a short description sentence would be good enough.

> I think it would be good to have some example output for each of the cases, so people know what kind of content they should get, particularly with different log levels (the format is just standard logging and should be fairly obvious).

Good point. I will add them in. I can reuse the examples in the Spec.

> In Jubilant or pytest-jubilant logs do we have somewhere information about what gets logged to each level? If so, it would be good to link to that. If not, maybe this is the place for that, although it seems like it belongs in Jubilant.

The logs don't contain this information.

Or, I assume you are referring to the docs :sweat_smile: . We don't have this information anywhere in the docs. ~Although I'm not sure if we should mention it, since this is not part of our contract, and can change later on ?~ Ah nevermind, my mistake. I think we definitely want to mention what type of information to expect at each level in Jubilant, and link it here.

docs/howto/write-integration-tests-for-a-charm.md
tromai: Yep, that's a good point. I will change it.

tonyandrewmeyer:
> Or, I assume you are referring to the docs 😅. 
 
Yes, sorry!

> I think we definitely want to mention what type of information to expect at each level in Jubilant, and link it here.

Agreed. I don't think this makes it an exact promise, and certainly the format is not. But as a user I need some way of deciding what makes it into each level so I can make a sensible choice of level. I think you're right that it should be the type of information rather than specific details.

dwilding:
Calling Brief, Verbose, and Error "logging modes" could be confusing. As you say, those names aren't used anywhere else. How about trying this structure:

> ## Configure Jubilant logs
> 
> Intro as currently
> 
> ### Log levels
> 
> Briefly explain the differences between INFO, DEBUG, and ERROR. E.g., how does Jubilant decide what to emit at what level?
> 
> ### Examples
> 
> Two examples with minimal discussion.
> 
> Here's our recommended config for brief logging:
> 
> ...
> 
> And here's some sample output:
> 
> ...
> 
> Here's a config for verbose logging:
> 
> ...
> 
> And here's some sample output:
> 
> ...
> 
> (I'd drop the ERROR example because I think you'll cover it sufficiently in the "Log levels" section)
> 
> ### Default behaviour
> 
> Info as currently

docs/howto/write-integration-tests-for-a-charm.md
dwilding: That's right. We ought to do a bulk replace one day. I've created https://github.com/canonical/operator/issues/2643 for that.

I'm OK if this PR adds them as body text, even if it's inconsistent. There's also no need to use bullets unless there are 3+ links.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+These options can also be configured from the command line.
+
+> See more:
+> - [`pytest | How to manage logging`](https://docs.pytest.org/en/stable/how-to/logging.html)
+
+(write-integration-tests-log-to-a-file)=
+### Log to a file
+
+This is an ideal configuration for long-running integration tests (for example, in CI). It uses Brief mode for logging to the console, and Verbose mode for logging to a local file.
```
dwilding: I'd change this slightly, per earlier discussion today:

```suggestion
This is an ideal configuration for long-running integration tests (for example, in CI). It outputs brief logs to the console and verbose logs to a local file.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+log_cli_level = "INFO"
+log_cli_format = "%(levelname)s %(name)s %(message)s"
+
+log_file = "logs/verbose.log"
+log_file_level = "DEBUG"
+log_file_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
+log_file_date_format = "%Y-%m-%dT%H:%M:%SZ"
+```
+
+When a test item fails, pytest prints all logs captured during the test to the terminal:
```
dwilding:
```suggestion
If a test item fails, pytest outputs all logs captured during the test to the console:
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+```
+------- Captured log call -------
+...
```
dwilding: I think it would be helpful to show at least one log line here.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+```
+
+`log_level = "INFO"` is intentional to keep the `Captured log call` consistent with Brief mode. If you want Verbose mode in `Captured log call`, you can set `log_level` to `DEBUG`.
```
dwilding: If I understand correctly, this should be clear from the example and the preceding description of log levels. I'd drop the paragraph.

```suggestion
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+The local file where Verbose mode logs land is `logs/verbose.log`. It contains logs from one pytest session. If you want to run multiple test suites, and store Verbose logs in separated files. You can override `log_file` for each pytest invocation:
```
dwilding: I'd adjust this to avoid repeating the example and emphasise the point about one pytest session:

```suggestion
If you run the integration tests multiple times, `logs/verbose.log` only contains logs from the last pytest session. To use a separate file for each session, override `log_file` for each pytest invocation:
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,22 +460,153 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+log-file-mode = "a"
+...
+```
+
+These options can also be configured from the command line.
+
+> See more:
+> - [`pytest | How to manage logging`](https://docs.pytest.org/en/stable/how-to/logging.html)
+
+Use this mode in CI with `actions/upload-artifact` to make `logs/verbose.log` available as build artifacts:
```
dwilding: Do we need "this mode"? How about:

```suggestion
In CI, you can use `actions/upload-artifact` to make `logs/verbose.log` available as a build artifact:
```

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Yeah my preference would be the version with full stops. But with a couple of tweaks:


```suggestion
- pytest. For example: `tests/integration/test_charm.py::test_deploy PASSED`.
- Other modules if the messages are at `WARNING` or above.
- `pytest-jubilant`. For example: `Models were torn down...`.
```

docs/howto/write-integration-tests-for-a-charm.md
tromai: I agree. 
May be I should remove the `Captured log call` too (from [here](https://github.com/canonical/operator/pull/2619#discussion_r3570821600)).

`Captured log call` and this paragraph explain why we need `log_level = "INFO"`. But since we explain the log levels and show the configs as is, I don't think we need to explain anymore. 

Let me know what you think.

docs/howto/write-integration-tests-for-a-charm.md
tromai: I think an inline link would be good. I had added it.

docs/howto/write-integration-tests-for-a-charm.md
tromai: I ended up not showing the diff, but still show the block in its final state. 

I made this decision in favor of the copy-and-paste usability. But please let me know if it's better otherwise.

tromai:
Thank you for the review. 

@dwilding  I have revised the logging level discussion in the docs following your suggested structure. And I think it works extremely well.

@tonyandrewmeyer I also added a short description on what is shown in each log level by Jubilant. It's under the new `Log levels` section. Although being minimal, it covers all things that Jubilant is logging. I can work on a more detailed description on the Jubilant side, but I can see that it won't be much different (if we don't go into specific details) :thinking: .

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+## Configure Jubilant logs
+
+Jubilant emits logs during the integration tests. These logs are captured and displayed by pytest. You can configure the logging behaviour in the `[tool.pytest.ini_options]` table in `pyproject.toml`.
+
+### Logging levels
+
+Jubilant logs across three levels:
+- `INFO`: brief progress updates, including run Juju CLI commands, the test model's bootstrap status, and any application or unit status changes during `juju.wait()`.
+- `DEBUG`: verbose model status changes during `juju.wait()`.
+- `ERROR`: application or unit status changes during `juju.wait()` whenever they transition into error.
```
dwilding: Very minor suggestions. All are style preferences.

```suggestion
Jubilant logs at three levels:
- `INFO` — Brief progress updates, including Juju commands, the test model's bootstrap status, and any application or unit status changes during `juju.wait()`.
- `DEBUG` — Verbose model status changes during `juju.wait()`.
- `ERROR` — Application or unit status changes during `juju.wait()` whenever the state becomes "error".
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
```
dwilding, line 474: ~~I think it would be good to add a sentence or two (max) about how the different levels interact. E.g., if I set DEBUG I'm still going to get everything that INFO logs.~~

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+### Examples
+
+**Config for brief logging**
```
dwilding: We could make this a `#`x4 heading, which would then be linkable and show in the TOC. But I _really_ try to avoid headings at that level of depth

Now that I can see what it looks like in the rendered doc, I'd consider two `#`x3 headings:

```
### Example of brief logging

...

### Example of verbose logging

...
```

For more than two examples this pattern would be annoying, but for two examples I think it's OK. Or I'm happy with what you have now - you choose 🙂

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+```toml
+[tool.pytest.ini_options]
+...
+log_cli = true
+log_cli_level = "INFO"
+log_cli_format = "%(levelname)s %(name)s %(message)s"
+```
+
+A sample output:
```
dwilding:
```suggestion
Sample output:
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+```toml
+[tool.pytest.ini_options]
+...
+log_cli = true
+log_cli_level = "DEBUG"
+log_cli_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
+log_cli_date_format = "%Y-%m-%dT%H:%M:%SZ"
+```
+
+We define the timestamps format with the `log_cli_date_format` key, following ISO 8601.
```
dwilding: I think this would be fine.

```suggestion
The timestamp format follows ISO 8601.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+See more: [`datetime | strftime() and strptime() format codes`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)
```
dwilding: This can be an intersphinx link, which will render as an external link in the docs.

```suggestion
See more: {external+python:ref}`strftime() and strptime() behavior <strftime-strptime-behavior>`
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+A sample output (timestamps removed for clarity):
```
dwilding: I'm not sure if this makes things clearer - but I know what you're getting at!

As an alternative approach, how about putting `<timestamp>` on the fist line, then `...` in the same location on all subsequent lines.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+INFO jubilant.wait [fastapi-demo/0] status changed: waiting (installing agent) -> waiting (agent initialising)
+DEBUG jubilant.wait wait: status changed:
+- .apps['fastapi-demo'].app_status.message = 'installing agent'
++ .apps['fastapi-demo'].app_status.message = 'agent initialising'
+- .apps['fastapi-demo'].units['fastapi-demo/0'].workload_status.message = 'installing agent'
++ .apps['fastapi-demo'].units['fastapi-demo/0'].workload_status.message = 'agent initialising'
+...
+```
+
+The pytest logging options can also be configured from the command line.
```
dwilding: Very minor: I'd make this more active. Something like:

```suggestion
You can also configure pytest logging options on the command line.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+See more: [`pytest | How to manage logging`](https://docs.pytest.org/en/stable/how-to/logging.html)
```
dwilding: No need for the code font (I know that's the style used elsewhere in the doc, but it's a historical leftover)

```suggestion
See more: [pytest | How to manage logging](https://docs.pytest.org/en/stable/how-to/logging.html)
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+### Default behaviour
+
+If no logging configuration is set by `tool.pytest.ini_options` or pytest CLI arguments, pytest captures all log messages at `WARNING` level or above. You will still see messages from:
+
+- pytest. For example: `tests/integration/test_charm.py::test_deploy PASSED`.
+- Other modules if the messages are at `WARNING` or above.
+- `pytest-jubilant`. For example: `Models were torn down...`.
```
dwilding: Super minor style preference

```suggestion
- pytest. For example, `tests/integration/test_charm.py::test_deploy PASSED`.
- Other modules if the messages are at `WARNING` or above.
- `pytest-jubilant`. For example, `Models were torn down...`.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+(write-integration-tests-log-to-a-file)=
+### Log to a file
+
+This is an ideal configuration for long-running integration tests (for example, in CI). It outputs brief logs to the console and verbose logs to a local file.
+
+```toml
+[tool.pytest.ini_options]
+...
+
+log_level = "INFO"
```
dwilding, line 570: Is this left over? I don't see `log_level` mentioned anywhere else.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+```
+pytest --log-file "run1.log" ...
+pytest --log-file "run2.log" ...
+```
+
+Alternatively, you can set `log-file-mode` to `append`. pytest will put verbose logs from all invocations into one file:
+
+```toml
+[tool.pytest.ini_options]
+log-file-mode = "a"
```
dwilding: Should this be underscored for the config option version?

```suggestion
log_file_mode = "a"
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+...
+```
+
+These options can also be configured from the command line.
```
dwilding:
```suggestion
You can also configure these options on the command line.
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+
+See more: [`pytest | How to manage logging`](https://docs.pytest.org/en/stable/how-to/logging.html)
```
dwilding:
```suggestion
See more: [pytest | How to manage logging](https://docs.pytest.org/en/stable/how-to/logging.html)
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+    with:
+      name: integration-test-logs
+      path: logs
+```
 
 (write-integration-tests-for-a-charm-view-juju-logs)=
-## View Juju logs
+## View `juju debug-log` logs
+
+Save the complete `juju debug-log` logs to disk with the `--juju-dump-logs` option from [pytest-jubilant](https://github.com/canonical/pytest-jubilant#--juju-dump-logs). For example:
```
dwilding: I'd switch this around for slightly smoother reading. Also suggest dropping "for example" because the example already seems pretty generic.

```
Use the `--juju-dump-logs` option from [pytest-jubilant](https://github.com/canonical/pytest-jubilant#--juju-dump-logs) to save the complete `juju debug-log` logs to disk:
```

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
 
-If any tests fail, `pytest-jubilant` automatically prints the last 1000 lines of `juju debug-log` to stderr. You can also save the complete logs to disk with the `--juju-dump-logs` option.
+```text
+tox -e integration -- --juju-dump-logs logs
+```
 
-Use `--juju-dump-logs` in CI with `actions/upload-artifact` to make debug logs available as build artifacts:
+Use `--juju-dump-logs` in CI with `actions/upload-artifact` to make `juju debug-log` files available as build artifacts:
```
dwilding: I'd use similar wording to the previous section:

```suggestion
In CI, you can use `--juju-dump-logs` with `actions/upload-artifact` to make `juju debug-log` files available as build artifacts:
```

dwilding (approved):
I added several comments, but all minor. Nice work on this!

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Looks like it's removed in the latest version, which looks good to me :+1:

docs/howto/write-integration-tests-for-a-charm.md
tonyandrewmeyer: I'm less sure about this. Isn't it reasonable to assume a basic knowledge of Python logging?

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -460,13 +460,152 @@ tox -e integration -- tests/integration/test_charm.py -k "not test_one"
...
+See more: [`datetime | strftime() and strptime() format codes`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)
+
+A sample output (timestamps removed for clarity):
+
+```text
+INFO jubilant cli: juju deploy --model jubilant-6ffd8f91-test-charm ...
+INFO jubilant.wait [fastapi-demo] status changed: waiting (installing agent)
+INFO jubilant.wait [fastapi-demo/0] status changed: waiting (installing agent)
+DEBUG jubilant.wait wait: status changed:
++ .apps['fastapi-demo'].app_status.current = 'waiting'
```
tonyandrewmeyer, line 525: I think it would be good if the debug example included something that isn't in the info level. Otherwise, it looks like it's all duplicate information. But actually you get more at debug level.

tonyandrewmeyer (approved):
Thanks for this! Nothing much other than agreeing with David's comments.

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Yeah, that's true. I've changed my mind after reading Tony's comment. The examples do show it too - in case someone didn't know already.

docs/howto/write-integration-tests-for-a-charm.md
tromai: I'm in favor of making them linkable. So I will pick the two `###` headings approach. Thank you for the suggestion.

docs/howto/write-integration-tests-for-a-charm.md
tromai: This is not a left over. This option is intentional to keep the `Captured log call` section, which is printed to the terminal, consistent with the brief Jubilant logs. We only need it in this specific config.

The explanation for this option was removed (as discussed in https://github.com/canonical/operator/pull/2619#discussion_r3574743631), because I thought it is less about "howto", and more about "explanation".

tonyandrewmeyer: What about:

```suggestion
log_level = "INFO"  # Retain the "Captured log call" section when run interactively.
```

dwilding: Tony's suggestion looks good to me. I still don't fully understand why `log_level = "INFO"` would be needed here and not in the config for brief logging.

tromai: I love the idea of adding a comment there. May I modify it a little bit, since `Captured log call` is still there whether you set `log_level = "INFO"` or not. `log_level = "INFO"` makes sure that only INFO level messages are retained in `Captured log call`

```
log_level = "INFO"  # Retain INFO logs in the "Captured log call" section when run interactively.
```

tromai: > I still don't fully understand why log_level = "INFO" would be needed here and not in the config for brief logging.

It's one of pytest's surprised behaviours. The capture log handler (that handles `Captured log call` section) is different from live cli handler (configured with `log_cli_level`) and file handler (configured with `log_file_level`). However, capture log handler level depends on the minimum level between `log_cli_level` and `log_file_level`. If we set `log_cli_level` to INFO, and `log_file_level` to DEBUG, capture log handler will have DEBUG level -> `log_level = "INFO"` is the cleanest way to enforce this back to INFO.

To me, it's also not ideal to have `log_level = "INFO"` here but not in brief logging. If it's too confusing, I can definitely remove it, and leave the verbose DEBUG logs in `Captured log call`.

docs/howto/write-integration-tests-for-a-charm.md
tromai: Oops. Yes, thanks for the catch.

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Thanks for the explanation! I understand better now and think `log_level` is worth keeping.

How about this for the comment. Too much?

```
# Retain INFO logs in the "Captured log call" section when run interactively.
# Otherwise, that section will have DEBUG logs (coming from log_file_level).
log_level = "INFO"
```

tromai: I like your suggestion. I will use it.

dwilding (approved):
I looked over the latest version. Great work!

Issue #2424, opened 2026-04-13
Add explicit failure notification to scheduled workflows


tonyandrewmeyer:
As discovered in https://github.com/canonical/jubilant/pull/301, if a scheduled workflow fails, the person who last modified the schedule (or originally created it, if no modifications) is notified. This is not a very resilient or team-focused solution.

Let's update our scheduled workflows to have an explicit failure notification mechanism (creating an issue with `gh` is likely simplest):

* example charm integration tests
* integration tests
* published charms tests - **maybe not this one, because it regularly fails, although we have made progress, or maybe we can have a set of expected failures or something similar?**
* smoke test - **currently failing on 2.9, but let's get that fixed**
* TIOBE reporting
* updating the best practice doc
* updating the charm test pinning

Please also check there aren't new scheduled workflows when doing this issue. Also, check if there are any in other Charm Tech repos that would also benefit from similar changes.

tonyandrewmeyer:
Harry suggested an alternative: have @prints-charming-bot edit all the schedules, and have it's email address not be Ben, but a new Google group for Charm Tech^wCollections.

tonyandrewmeyer:
Other repos that need similar treatment:

* charm-ubuntu: TIOBE
* pebble: TIOBE
* concierge: TIOBE
* Jubilant: TIOBE
* pytest-jubilant: TIOBE
* Hyrum: update-charm-list

@james-garner-canonical would you say that we've experience this enough with the Jubilant `ci` one that we could roll this out across all of those already? Or do you think it's better to get some experience with it in this larger operator case first?

benhoyt:
We've done the operator part, we'll get to the other repos when we get to them (low pri)

PR #2627, opened 2026-07-06
ci: open an issue if a scheduled workflow fails

tonyandrewmeyer:
For each of our workflows that runs on a schedule, if the scheduled run fails, open an issue. This ensures everyone is aware and that there is tracking in GitHub, rather than just an email to the person that most recently set the schedule.

This uses the approach from Jubilant, where it's a new issue every time, rather than updating an existing one. The title and body are quite bare-boned, like Jubilant, but we could adjust that as needed. In practice, I think what we need is the discipline to promptly investigate opened issues (which were not doing a great job of in Jubilant) and convert them into detailed reports of the specific failure, and then close as duplicates new ones if needed. Alternatively, we could throw in some AI to investigate them and convert them into a detailed investigation.

I haven't tested this, but it's very strongly based on [the Jubilant one](https://github.com/canonical/jubilant/blob/aafdeea1b70cc4f7db8ade29e6fcc808a8f044a9/.github/workflows/ci.yaml#L172), which has had several executions already. Testing would need to be done on my fork, but I can if desired.

Refs #2424


.github/workflows/example-charm-charmcraft-test.yaml
```diff
@@ -65,3 +65,10 @@ jobs:
         # allocator (runs against the runner itself) instead of launching
         # a nested LXD VM for the spread target.
         run: charmcraft test "craft:ubuntu-24.04:spread/integration/${{ matrix.task }}"
+
+  open-issue-on-failure-if-scheduled:
+    if: ${{ failure() && github.event_name == 'schedule' }}
+    needs: [collect-spread-jobs, integration]
```
james-garner-canonical:
```suggestion
    needs: [integration]
```

.github/workflows/integration.yaml
```diff
@@ -71,3 +71,10 @@ jobs:
         # tracing tests each deploy the tempo stack into their own fresh
         # model, so that file needs longer than the others.
         timeout-minutes: ${{ matrix.test == 'test/integration/test_tracing.py' && 40 || 20 }}
+
+  open-issue-on-failure-if-scheduled:
+    if: ${{ failure() && github.event_name == 'schedule' }}
+    needs: [collect, integration]
```
james-garner-canonical:
```suggestion
    needs: [integration]
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  open-issue:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Create issue on failure
+        env:
+          GH_TOKEN: ${{ github.token }}
+          WORKFLOW_NAME: ${{ github.workflow }}
```
james-garner-canonical:
```suggestion
          WORKFLOW_NAME: ${{ github.workflow }}  # The workflow that called this one.
```

james-garner-canonical (approved):
Thanks for doing this, it'll be interesting to see how this works out with a lot more separate workflows.

tromai (approved):
Everything looks good to me. Thank you.

I agree with James on adding a small comment to `WORKFLOW_NAME: ${{ github.workflow }}`.

PR #2628, opened 2026-07-06
fix: allow `IdentityDict` to be assigned `IdentityAccess`

### LINKED ISSUE #1977: Improve type hints for Pebble identities
[test/test_pebble.py](https://github.com/canonical/operator/pull/1967/files/7c4bd420c772a1af8c3aa5262b6a8948e9df7e52#diff-4cce6d8f9579f921ed13f437c91fcffb7a763b7930c8d6b15c3d6613f4a8ce8d)

```diff
@@ -4030,7 +4030,7 @@ def test_local_identity_from_dict(self):

    def test_local_identity_from_dict_with_access_enum(self):
        identity = pebble.Identity.from_dict({
            'access': pebble.IdentityAccess.ADMIN,
            'access': pebble.IdentityAccess.ADMIN,  # type: ignore
```

Member
Author
@[dimaqq](https://github.com/dimaqq) dimaqq [4 days ago](https://github.com/canonical/operator/pull/1967#discussion_r2262146198)
This is of potential concern to our users.

If we want them to call Identity.from_dict both with strings and enum members and have that type check strictly, we may have to do some extra work.

For now, I've decided not to adjust our public types, not within the scope of this PR anyway.

Member
@[tonyandrewmeyer](https://github.com/tonyandrewmeyer) tonyandrewmeyer [4 days ago](https://github.com/canonical/operator/pull/1967#discussion_r2262359048)
Is this because the type checker isn't ok with a str subclass when only a small set of literals is acceptable? If the typed dict accepted a string or the enum would that fix it? That seems a safe change, and more correct. I think we should fix it now, even if not in this PR.

Member
@[james-garner-canonical](https://github.com/james-garner-canonical) james-garner-canonical [yesterday](https://github.com/canonical/operator/pull/1967#discussion_r2265474269)
I wonder if adding a type annotation to IdentityAccess.ADMIN (Literal['admin']) etc would resolve this.

Member
Author
@[dimaqq](https://github.com/dimaqq) dimaqq [1 hour ago](https://github.com/canonical/operator/pull/1967#discussion_r2268226909)
It's not possible to add type annotations to enum members.

+1 on making the user-facing change in a separate PR, with pros and cons discussion on that separate PR.

tonyandrewmeyer:
Allow the `access` field of `IdentityDict` to be the enum or the string literals.

Fixes #1977


tonyandrewmeyer:
The issue has a bunch of discussion copied from somewhere else that makes me wonder if this is too simple a fix, but it seems to solve the `type: ignore` at the heart of that, and the code still works since we handle the enum, and this should be backwards compatible since we're expanding not contracting. But maybe I am missing something?

james-garner-canonical:
I think this technically runs into backwards compatibility issues with `Identity.to_dict`. We still convert `access` to a string, but the type checker doesn't know that. Previously a user could know at type checking time that `identity.to_dict('access')` was a `str`, but now the type checker will think it might be an enum member instead.

james-garner-canonical (commented):
Comment review to allow explicit re-review request once the backwards compatibility side is addressed.

tonyandrewmeyer:
> I think this technically runs into backwards compatibility issues with `Identity.to_dict`. We still convert `access` to a string, but the type checker doesn't know that. Previously a user could know at type checking time that `identity.to_dict('access')` was a `str`, but now the type checker will think it might be an enum member instead.

Good point. What about: type `access` as plain `str` instead of `IdentityAccess | Literal[...]`. Since `IdentityAccess` subclasses `str`, users can still pass the enum in without a `type: ignore`, and `to_dict()['access']` stays `str` for existing readers.

The cost is that callers using `d['access']` with an `IdentityDict` lose the `Literal[...]` narrowing. Perhaps more true, since `from_dict` already accepts unknown values but realistically we know Pebble will be providing one of those and would know if Pebble changed and would update ops.

I'm not sure it's better than what we have. We could also just accept that the status quo is ok, and look to do something nice in general in a breaking release?

james-garner-canonical:
> > I think this technically runs into backwards compatibility issues with `Identity.to_dict`. We still convert `access` to a string, but the type checker doesn't know that. Previously a user could know at type checking time that `identity.to_dict('access')` was a `str`, but now the type checker will think it might be an enum member instead.
> 
> Good point. What about: type `access` as plain `str` instead of `IdentityAccess | Literal[...]`. Since `IdentityAccess` subclasses `str`, users can still pass the enum in without a `type: ignore`, and `to_dict()['access']` stays `str` for existing readers.

This seems like an elegant solution on the `from_dict` and `to_dict` paths. I had missed that `IdentityAccess` was a `str` subclass -- I think that means that the PR is actually OK as-is, since the previous (`Literal` only) string values didn't support any operations the `IdentityAccess` enum members don't -- they're both `str`.

> The cost is that callers using `d['access']` with an `IdentityDict` lose the `Literal[...]` narrowing. Perhaps more true, since `from_dict` already accepts unknown values but realistically we know Pebble will be providing one of those and would know if Pebble changed and would update ops.

I think the unknown Pebble values side is a bit of a wash overall, but where this would bite us is with the `Identity` class. We'd want `Identity(d['access'])` to stay legal, so we'd need to broaden the permitted argument type to accept `str` as well (and because it's a `dataclass` :grimacing:  (a non-`frozen` one too :scream_cat:) we'd have to deal with what that means for reading (and writing) the attribute.

> I'm not sure it's better than what we have. We could also just accept that the status quo is ok, and look to do something nice in general in a breaking release?

I think the `str` idea was a good one, but perhaps unnecessary -- after some thought, it seems that the initial union might be fine since both values are ultimately `str`. The `Literal` values only matter for passing to functions or assigning to attributes that expect only those values -- if that's just `Identity.access`, then the change is fine since it already permits `IdentityAccess` as well.

To move forward with the current PR, I'd say we just need a scan of our code to ensure we don't have anything accepting the `Literal` values only, and a Hyrum run of charm linting just in case it picks up a breakage we haven't thought of with this change.

tonyandrewmeyer:
I did a scan (grep and eyeballing). The only `Literal['untrusted', 'metrics', 'read', 'admin']` annotations in ops/ (and testing/src/) are the two the PR already widens. Nothing else parameterises on the bare literal set, so no other call sites need broadening. Kicking off a Hyrum run now.

james-garner-canonical (approved):
Thanks, approving optimistically.

tromai (approved):
Thanks for the fix, and checking for backward compatibility.

tonyandrewmeyer:
Hyrum results: no real regressions. Both runs patched (main vs this branch):

- **lint**: 0 new failures, 0 resolved. Identical pass rate.
- **unit**: 1 apparent new failure, `canonical/openstack-exporter-operator`, but the summary is "tests passed (155 passed); post-test step failed (exit 1)". Every actual unit test passed; a post-test step in the tox recipe (likely coverage or a lint hook) exited non-zero. Not attributable to widening `IdentityDict.access`.

Combined with the earlier scan, I think we're good to go.

PR #2631, opened 2026-07-07
docs: fix tempo test_deploy url

tromai:
This PR fixes a broken link to tempo-worker's `test_deploy.py`.

In https://github.com/canonical/tempo-operators/pull/270 this file was relocated from `worker/tests/integration/test_deploy.py` to `tests/integration/test_worker_deploy.py`.

In the future, should we consider using a permanent link in such cases ?


tonyandrewmeyer (approved):
Thanks!

Good question about how to avoid this. Shouldn't our link checker already find these issue when they happen and we at least know to fix them?

I'm not sure about linking to a specific revision. When linking to examples it is good to link to the newest version of the, to get the latest best practice. If the line disappears entirely then we probably need to think about whether it's still an example we want to use, or even if we are suggesting something outdated.

dwilding (approved):
The link checker ought to have found this, if we used the link checker. But we don't use it in this repo because it can be quite unreliable. We (Canonical doc practice) are working on improving the reliability this cycle.

tromai:
Thanks Tony and David. 

I found out about this broken link while viewing the docs. I just finished running the link checker and it could find this broken link :+1: too.

> I'm not sure about linking to a specific revision. 

Agreed. Pinning the revision is a bad idea here. And we should double check if the example is relevant, each time we update it.

PR #2635, opened 2026-07-07
docs: revert to installed charmcraft for initing charms

dwilding:
Now that Charmcraft 4.3 is available on `latest/stable`, we can remove the `uvx` methods of running `charmcraft init`. I did a diff locally to check that the profiles from snap Charmcraft are identical to what the `uvx` method produced.


james-garner-canonical (approved):
🥳

tromai (approved):
Wonderful!

PR #2636, opened 2026-07-09
fix: only decode fields used by the data class in Relation.load()

### LINKED ISSUE #2591: `ops.Relation.load()` on unit data results in `JSONDecodeError: Extra data`
### Description

When `ops.Relation.load()` is used to load from unit data into a data class, a `json.decoder.JSONDecodeError: Extra data` is raised.

I believe this is due to the default `json.loads()` decoder failing to interpret the standard values in the unit data, e.g. `{'egress-subnets': '10.200.245.189/32', 'ingress-address': '10.200.245.189', 'private-address': '10.200.245.189'}`.

### To reproduce

The error can be reproduced by this charm code, which `save()`s then `load()`s unit data in a peer relation:

```python
class PeerData(pydantic.BaseModel):
    myflag: bool = False

class ReloadIssueCharm(ops.CharmBase):
    """Charm the application."""

        super().__init__(framework)

    def _on_start(self, _: ops.StartEvent):
        if peers := self.model.get_relation("peers"):
            save_data = PeerData(myflag=True)
            peers.save(save_data, self.unit)
            logger.info("Saved peer data: %s", save_data)

            load_data = peers.load(PeerData, self.unit)
            logger.info("Loaded peer data: %s", load_data)
```

The `_on_start` handler fails at `peers.load(PeerData, self.unit)`:

<details>

<Summary>`json.decoder.JSONDecodeError: Extra data` traceback</summary>

```
2026-07-01 09:58:38 INFO unit.reload-issue-charm/0.juju-log server.go:405 Saved peer data: myflag=True
2026-07-01 09:58:38 ERROR unit.reload-issue-charm/0.juju-log server.go:405 Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/src/charm.py", line 39, in <module>
    ops.main(ReloadIssueCharm)
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/__init__.py", line 360, in __call__
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 507, in main
    manager.run()
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 491, in run
    self._emit()
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 426, in _emit
    self._emit_charm_event(self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 470, in _emit_charm_event
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/framework.py", line 337, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/framework.py", line 910, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/framework.py", line 1028, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/src/charm.py", line 32, in _on_start
    load_data = peers.load(PeerData, self.unit)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-reload-issue-charm-0/charm/venv/lib/python3.12/site-packages/ops/model.py", line 1829, in load
    value = decoder(value)
            ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 340, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1 column 7 (char 6)
```

</details>

Or, more immediately:

```python
>>> json.loads('10.200.245.189')
  File "/usr/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 340, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1 column 7 (char 6)
```

### Workaround

We are working around this with a custom decoder that interprets the values as strings:

```python
def _unit_data_decoder(value: str) -> str:
    if not (value.startswith('"') and value.endswith('"')):
        value = f'"{value}"'
    return json.loads(value)
```

E.g. equivalent to `json.loads('"10.200.245.189"')`.

Then `load_data = peers.load(PeerData, self.unit, decoder=_unit_data_decoder)` runs as expected:

```
2026-07-01 10:01:29 INFO unit.reload-issue-charm-workaround/0.juju-log server.go:405 Saved peer data: myflag=True
2026-07-01 10:01:29 INFO unit.reload-issue-charm-workaround/0.juju-log server.go:405 Loaded peer data: myflag=True
```

### Expectation

`ops.Relation.load()` should be able to read unit data without without raising a `JSONDecodeError` and without need for a custom decoder.

joelnguemeta:
## Description

`Relation.load()` decodes every value in the databag before filtering by the class fields. Unit databags always contain keys automatically set by Juju (`egress-subnets`, `ingress-address`, `private-address`) whose values are plain strings, not JSON, so loading unit data into a dataclass or Pydantic model raised `json.decoder.JSONDecodeError: Extra data` — even when the class doesn't declare those fields.

This change filters by the class fields first and only decodes matching values, so keys that are not part of the data class are ignored without ever reaching the decoder. Round-trip semantics are unchanged: declared fields are still decoded with `json.loads` (or the custom `decoder`), so types survive `save()`/`load()` round trips.

Out of scope: for classes that are neither dataclasses nor Pydantic models, `_juju_fields()` cannot determine the fields, so the whole databag is still decoded and passed to `__init__` (documented behaviour — such classes were already incompatible with unit databags via the unexpected keyword arguments).

## Testing

- New regression test `test_relation_load_unit_data_ignores_juju_keys`, parametrized over dataclass, Pydantic dataclass and `BaseModel`: saves to and loads from the unit databag with the Juju-populated keys present (ops-scenario's defaults). It fails with `JSONDecodeError: Extra data` before the fix and passes after.
- Also reproduced the exact charm from the issue (peer relation, `save`/`load` in `_on_start`) via ops-scenario: same traceback as reported before the fix, loads correctly after.
- `test_model_relation_data_class.py`  and `test_model.py`  all pass; `tox -e lint` clean.

Fixes #2591

### DIFF
--- a/test/test_model_relation_data_class.py
+++ b/test/test_model_relation_data_class.py
@@ -502,6 +502,25 @@ def test_relation_load_then_save(charm_class: type[BaseTestCharm]):
     }
 
 
+@pytest.mark.parametrize('charm_class', [c for c in _test_classes if c is not MyCharm])
+def test_relation_load_unit_data_ignores_juju_keys(charm_class: type[BaseTestCharm]):
+    class Charm(charm_class):
+        def _on_relation_changed(self, event: ops.RelationChangedEvent):
+            saved = self.databag_class(foo='value', bar=1, baz=['a', 'b'])
+            event.relation.save(saved, self.unit, encoder=self.encoder)
+            self.data = event.relation.load(self.databag_class, self.unit, decoder=self.decoder)
+
+    ctx = testing.Context(Charm, meta={'name': 'foo', 'requires': {'db': {'interface': 'db-int'}}})
+    rel = testing.Relation('db')
+    state_in = testing.State(leader=True, relations={rel})
+    with ctx(ctx.on.relation_changed(rel), state_in) as mgr:
+        mgr.run()
+        obj = mgr.charm.data
+    assert obj.foo == 'value'
+    assert obj.bar == 1
+    assert obj.baz == ['a', 'b']
+
+
 @pytest.mark.parametrize('charm_class', _test_classes)
 def test_relation_save_invalid(charm_class: type[BaseTestCharm], monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setenv('SCENARIO_BARE_CHARM_ERRORS', 'true')
--- a/test/test_model_relation_data_class.py
+++ b/test/test_model_relation_data_class.py
@@ -508,10 +508,19 @@ def test_relation_load_unit_data_ignores_juju_keys(charm_class: type[BaseTestCha
         def _on_relation_changed(self, event: ops.RelationChangedEvent):
             saved = self.databag_class(foo='value', bar=1, baz=['a', 'b'])
             event.relation.save(saved, self.unit, encoder=self.encoder)
+            assert 'ingress-address' in event.relation.data[self.unit]
+            assert 'private-address' in event.relation.data[self.unit]
             self.data = event.relation.load(self.databag_class, self.unit, decoder=self.decoder)
 
     ctx = testing.Context(Charm, meta={'name': 'foo', 'requires': {'db': {'interface': 'db-int'}}})
-    rel = testing.Relation('db')
+    rel = testing.Relation(
+        'db',
+        local_unit_data={
+            'egress-subnets': '192.0.2.0',
+            'ingress-address': '192.0.2.0',
+            'private-address': '192.0.2.0',
+        },
+    )
     state_in = testing.State(leader=True, relations={rel})
     with ctx(ctx.on.relation_changed(rel), state_in) as mgr:
         mgr.run()
--- a/ops/model.py
+++ b/ops/model.py
@@ -1826,11 +1826,10 @@ class Relation:
         if decoder is None:
             decoder = json.loads
         for key, value in sorted(self.data[src].items()):
-            value = decoder(value)
             if fields is None:
-                data[key] = value
+                data[key] = decoder(value)
             elif key in fields:
-                data[fields[key]] = value
+                data[fields[key]] = decoder(value)
         return cls(*args, **data)
 
     def save(


test/test_model_relation_data_class.py
```diff
@@ -502,6 +502,25 @@ def _on_relation_changed(self, event: ops.RelationChangedEvent):
...
+@pytest.mark.parametrize('charm_class', [c for c in _test_classes if c is not MyCharm])
+def test_relation_load_unit_data_ignores_juju_keys(charm_class: type[BaseTestCharm]):
+    class Charm(charm_class):
+        def _on_relation_changed(self, event: ops.RelationChangedEvent):
+            saved = self.databag_class(foo='value', bar=1, baz=['a', 'b'])
+            event.relation.save(saved, self.unit, encoder=self.encoder)
+            self.data = event.relation.load(self.databag_class, self.unit, decoder=self.decoder)
+
+    ctx = testing.Context(Charm, meta={'name': 'foo', 'requires': {'db': {'interface': 'db-int'}}})
+    rel = testing.Relation('db')
```
tonyandrewmeyer: Let's add an assert after this that the keys are present (at least the two still present in Juju 4), or explicitly have them in the unit data, so it's (a) clearer what the test is doing, and (b) protected against Scenario not auto injecting the values in case that changes.

tonyandrewmeyer (approved):
Thanks @joelnguemeta ! The fix is the correct one, I think. One small suggestion for the test, but otherwise seems good to go to me.

test/test_model_relation_data_class.py
joelnguemeta: Done.  You can check it

test/test_model_relation_data_class.py
```diff
@@ -502,6 +502,34 @@ def _on_relation_changed(self, event: ops.RelationChangedEvent):
     }
 
 
+@pytest.mark.parametrize('charm_class', [c for c in _test_classes if c is not MyCharm])
```
tromai, line 505: I understand why we need this list comprehension. 

But I'm a bit concern that this introduce non-determinism to this test, because `_test_classes` can depend on whether `pydantic` is available.

Would it be better if we use one concrete class for this test, to make it isolated and independent from the environment (whether pydantic is available or not) ?

@tonyandrewmeyer I would like to know what you think.

tromai (commented):
Thank you for the fix!

It's good not to decode the value eagerly.

I have one concern on the test.

test/test_model_relation_data_class.py
tromai: There is one downside, we cannot test the pydantic version :thinking:, so I'm not opposed to keep the test like what you already have.

tonyandrewmeyer: We already have quite a lot of tests for the .load/.save functionality (and load_config and load_params) that are skipped if Pydantic is not available, and we know that it is when we run the tests ourselves in CI (ops-tracing currently pulls it in).

So I don't see any issue with how it is now, it seems to follow the existing pattern. I do think we should be testing that we get the expected behaviour with Pydantic classes as well as standard library ones, and it seems fine to me to have that via this parametrisation rather than having two test methods.

tromai (approved):

joelnguemeta:
Thank you @tonyandrewmeyer and @tromai for your constructive feedback. I hope to improve next time.

PR #2638, opened 2026-07-09
docs: separate test_workload_version_is_set in k8s tutorial chapter 3

tromai:
This PR brought `test_workload_version_is_set` back to life in chapter 3, and propagate that change to chapter 4 and 5. It also updated the example `examples/k8s-*/tests/integration/test_charm.py` files.

[Doc preview](https://canonical-juju-ops--2638.com.readthedocs.build/2638/)

It was raised by David:

> In the k8s tutorial Chapter, section [write an integration test](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#write-an-integration-test), we replace the contents of test_charm.py. That loses `test_workload_version_is_set` and we never recreate it. In the corresponding file in the example charm, we keep the version check as part of test_database_integration, but we don't have that in the tutorial.

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -478,11 +478,9 @@ For example, it should be able to pack, deploy, and integrate without throwing e
 
 You can ensure this by writing integration tests for your charm. In the charming world, these are usually written with {external+jubilant:doc}`Jubilant <reference/jubilant>` and [`pytest-jubilant`](https://github.com/canonical/pytest-jubilant).
 
-In this section we'll write a small integration test to check that the charm packs and deploys correctly.
-
 ### Write integration tests
 
-Let's write some simplest possible integration tests, [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)). These tests will deploy the charm, verify that the installation event is handled without errors and the workload version is set correctly.
+Let's write some integration tests as [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)). These tests will deploy the charm, verify that the installation event is handled without errors and the workload version is set correctly.
 
 Replace the contents of `tests/integration/test_charm.py` with:
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -510,16 +510,18 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_active)
 
 
-def test_workload_version_is_set(juju: jubilant.Juju):
-    # Verify that the workload version has been set.
+def test_workload_version_is_set(charm: pathlib.Path, juju: jubilant.Juju):
+    """Verify that the workload version has been set."""
     version = juju.status().apps[APP_NAME].version
     assert version == "1.0.4"  # Hardcoded for simplicity.
 ```
 
-This test depends on two fixtures:
+These tests depend on two fixtures:
 
 - `charm` - The `.charm` file to deploy. This fixture is defined in `tests/integration/conftest.py`.
 - `juju` - A Jubilant object for interacting with a temporary Juju model. This fixture is provided by the `pytest-jubilant` plugin.
+
+Both tests depend on the `charm` fixture even though only `test_deploy` uses the fixture. This ensures that the tests fail immediately if a `.charm` file isn't available.
 
 ### Run the test
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -480,9 +480,9 @@ You can ensure this by writing integration tests for your charm. In the charming
 
 In this section we'll write a small integration test to check that the charm packs and deploys correctly.
 
-### Write a test
+### Write integration tests
 
-Let's write the simplest possible integration test, a [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)). This test will deploy the charm, then verify that the installation event is handled without errors.
+Let's write some simplest possible integration tests, [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)). These tests will deploy the charm, verify that the installation event is handled without errors and the workload version is set correctly.
 
 Replace the contents of `tests/integration/test_charm.py` with:
 
@@ -523,7 +523,7 @@ These tests depend on two fixtures:
 
 Both tests depend on the `charm` fixture even though only `test_deploy` uses the fixture. This ensures that the tests fail immediately if a `.charm` file isn't available.
 
-### Run the test
+### Run tests
 
 Run the following command from anywhere in the `~/fastapi-demo` directory:
 
@@ -531,14 +531,14 @@ Run the following command from anywhere in the `~/fastapi-demo` directory:
 tox -e integration
 ```
 
-The test takes some time to run as a new Juju model is created and your charm is deployed. If successful, it'll verify that your packed charm can be deployed as expected.
+These tests take some time to run as a new Juju model is created and your charm is deployed. If successful, they'll verify that your packed charm can be deployed as expected.
 
 The result should be similar to the following output:
 
 ```text
 ...
 
-============================= 1 passed in 55.43s =============================
+============================= 2 passed in 55.43s =============================
   integration: OK (57.79=setup[0.23]+cmd[57.57] seconds)
   congratulations :) (57.84 seconds)
 ```
@@ -547,7 +547,7 @@ The result should be similar to the following output:
 `tox -e integration` doesn't pack your charm. If you modify the charm code and want to run the integration tests again, run `charmcraft pack` before `tox -e integration`.
 ```
 
-The Juju model is destroyed at the end of the test. If you want to run the test and keep the model for further exploration, see the example commands in [](#write-integration-tests-for-a-charm-run-your-tests). The `@pytest.mark.juju_setup` marker on `test_deploy` gives you the option of skipping this test on subsequent runs, for iterative testing on a deployed application.
+The Juju model is destroyed at the end of the tests. If you want to run the tests and keep the model for further exploration, see the example commands in [](#write-integration-tests-for-a-charm-run-your-tests). The `@pytest.mark.juju_setup` marker on `test_deploy` gives you the option of skipping this test on subsequent runs, for iterative testing on a deployed application.
 
 ### Run tests with `charmcraft test`
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -525,7 +525,7 @@ Both tests depend on the `charm` fixture even though only `test_deploy` uses the
 
 Run the following command from anywhere in the `~/fastapi-demo` directory:
 
-```text
+```shell
 tox -e integration
 ```
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -480,7 +480,7 @@ Now run `tox -e unit` to make sure all test cases pass.
 
 ## Write an integration test
 
-Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. In your `tests/integration/test_charm.py` file, update `test_deploy` to expect blocked status:
+Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. In `tests/integration/test_charm.py`, update `test_deploy` to expect blocked status:
 
 ```diff
 -    juju.wait(jubilant.all_active)
@@ -489,7 +489,7 @@ Now that our charm integrates with the database, if there's not a database relat
 
 Then, let's add another test case to check the integration is successful. For that, we need to deploy a database to the test cluster and integrate both applications. If everything works as intended, the charm should report an active status.
 
-In your `tests/integration/test_charm.py` file add the following test case after `test_workload_version_is_set`:
+In `tests/integration/test_charm.py`, add the following test case after `test_workload_version_is_set`:
 
 ```python
 @pytest.mark.juju_setup
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -503,10 +503,10 @@ def test_database_integration(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_active)
 ```
 
-In your Multipass Ubuntu VM, run the tests again:
+Run the following command from anywhere in the `~/fastapi-demo` directory:
 
 ```text
-ubuntu@juju-sandbox-k8s:~/fastapi-demo$ tox -e integration
+tox -e integration
 ```
 
 The tests may take some time to run, depending on your computer and network.
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -482,9 +482,20 @@ Now run `tox -e unit` to make sure all test cases pass.
 
 Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. In `tests/integration/test_charm.py`, update `test_deploy` to expect blocked status:
 
-```diff
--    juju.wait(jubilant.all_active)
-+    juju.wait(jubilant.all_blocked)
+```python
+@pytest.mark.juju_setup
+def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
+    """Deploy the charm under test.
+
+    Assert on the unit status before any relations/configurations take place.
+    """
+    resources = {
+        "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
+    }
+
+    # Deploy the charm and wait for it to report blocked, as it needs Postgres.
+    juju.deploy(charm, app=APP_NAME, resources=resources)
+    juju.wait(jubilant.all_blocked)
 ```
 
 Then, let's add another test case to check the integration is successful. For that, we need to deploy a database to the test cluster and integrate both applications. If everything works as intended, the charm should report an active status.
@@ -505,7 +516,7 @@ def test_database_integration(charm: pathlib.Path, juju: jubilant.Juju):
 
 Run the following command from anywhere in the `~/fastapi-demo` directory:
 
-```text
+```shell
 tox -e integration
 ```
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
@@ -480,46 +480,16 @@ Now run `tox -e unit` to make sure all test cases pass.
 
 ## Write an integration test
 
-Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. Let's tweak our existing integration test `test_deploy` accordingly, to expect blocked status in `juju.wait`. Replace the contents of `tests/integration/test_charm.py` with:
+Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. In your `tests/integration/test_charm.py` file, update `test_deploy` to expect blocked status:
 
-```python
-import logging
-import pathlib
-
-import jubilant
-import pytest
-import yaml
-
-logger = logging.getLogger(__name__)
-
-METADATA = yaml.safe_load(pathlib.Path("charmcraft.yaml").read_text())
-APP_NAME = METADATA["name"]
-
-
-@pytest.mark.juju_setup
-def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
-    """Deploy the charm under test.
-
-    Assert on the unit status before any relations/configurations take place.
-    """
-    resources = {
-        "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
-    }
-
-    # Deploy the charm and wait for it to report blocked, as it needs Postgres.
-    juju.deploy(charm, app=APP_NAME, resources=resources)
-    juju.wait(jubilant.all_blocked)
-
-
-def test_workload_version_is_set(juju: jubilant.Juju):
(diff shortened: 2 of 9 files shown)


docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -492,7 +492,7 @@ import yaml
 
 logger = logging.getLogger(__name__)
 
-METADATA = yaml.safe_load(pathlib.Path("./charmcraft.yaml").read_text())
+METADATA = yaml.safe_load(pathlib.Path("charmcraft.yaml").read_text())
```
tromai: I made this change to following what we have in [Chapter 1](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm/#id4). Please let me know if the `./` is intentional here.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tromai: I'm not sure if it's in the scope.

The content we suggest here is different from Chapter 2 in 2 places:
- From `juju.wait(jubilant.all_active)` to `juju.wait(jubilant.all_blocked)`
- The doc string `Assert on the unit status before any relations/configurations take place.`

We can simplify by describing a diff, instead of a whole file replacement:

```diff
- """Deploy the charm under test."""
+ """Deploy the charm under test.
+
+ Assert on the unit status before any relations/configurations take place.
+ """

...
- juju.wait(jubilant.all_active)
+ juju.wait(jubilant.all_blocked)
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: Good catch, thanks. The version without `./` matches the Charmcraft profile (see https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/tests/integration/test_charm.py.j2)

examples/k8s-3-postgresql/tests/integration/test_charm.py
```diff
@@ -55,6 +61,3 @@ def test_database_integration(charm: pathlib.Path, juju: jubilant.Juju):
     juju.deploy("postgresql-k8s", channel="14/stable", trust=True)
     juju.integrate(APP_NAME, "postgresql-k8s")
     juju.wait(jubilant.all_active, timeout=10 * 60)
```
dwilding: This must be left over from previous work by one of us. In the tutorial we mention `timeout` as an optional addition only.

```suggestion
    juju.wait(jubilant.all_active)
```

examples/k8s-3-postgresql/tests/integration/test_charm.py
```diff
@@ -46,6 +46,12 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_blocked)
 
 
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
+    version = juju.status().apps[APP_NAME].version
+    assert version == "1.0.4"  # Hardcoded for simplicity.
```
dwilding: Interesting, I was expecting we'd put this test after `test_database_integration` (which unblocks the charm). But I think you're right; it doesn't matter that the charm is blocked, as that's coming from the collect-status handler.

examples/k8s-5-observe/tests/integration/test_charm.py
```diff
@@ -50,6 +50,12 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_blocked)
 
 
+def test_workload_version_is_set(juju: jubilant.Juju):
+    # Verify that the workload version has been set.
```
dwilding: I suppose the initial comment would be better as a docstring - in all the `k8s-*` charms, and in the tutorial source.

examples/k8s-5-observe/tests/integration/test_charm.py
```diff
@@ -50,6 +50,12 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_blocked)
```
dwilding, line 50:
```suggestion
    """Deploy the charm under test."""
    resources = {
        "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
    }
    juju.deploy(charm, app=APP_NAME, resources=resources)
    juju.wait(jubilant.all_blocked)
```

examples/k8s-4-action/tests/integration/test_charm.py
```diff
@@ -46,6 +46,12 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_blocked)
```
dwilding, line 46:
```suggestion
    """Deploy the charm under test."""
    resources = {
        "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
    }
    juju.deploy(charm, app=APP_NAME, resources=resources)
    juju.wait(jubilant.all_blocked)
```

examples/k8s-3-postgresql/tests/integration/test_charm.py
```diff
@@ -46,6 +46,12 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
     juju.wait(jubilant.all_blocked)
```
dwilding, line 46:
```suggestion
    """Deploy the charm under test."""
    resources = {
        "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
    }
    juju.deploy(charm, app=APP_NAME, resources=resources)
    juju.wait(jubilant.all_blocked)
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: I agree with you. I don't see why we need to rewrite the whole file. I think it will be clearer for the reader if we tell them exactly what to change.

With that in mind, I suggest making the format of `test_deploy` match closer with chapter 2. Keep the same docstring and remove the comment before `juju.deploy` (the tutorial already explains things sufficiently).

This means the only change needed is `all_active` --> `all_blocked`. The tutorial can effectively just say that 😁 

I've added suggestions on the testing code.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -575,6 +578,12 @@ INFO     jubilant.wait:_juju.py:1164 wait: status changed:
 PASSED
 ```
 
+```text
+tests/integration/test_charm.py::test_workload_version_is_set
+...
+PASSED
+```
+
```
dwilding: The testing code runs this before `test_database_integration`, so the tutorial should use the same order.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: Maybe we should clarify where to add it, just in case anyone is stuck wondering where to put it.


```suggestion
In your `tests/integration/test_charm.py` file add the following test case after `test_workload_version_is_set`:
```

examples/k8s-5-observe/tests/integration/test_charm.py
```diff
@@ -50,6 +50,12 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
...
 
 
+def test_workload_version_is_set(juju: jubilant.Juju):
```
dwilding: I'd make this test depend on `charm: pathlib.Path` too - in all the `k8s-*` charms, and in the tutorial source. Matching the other integration tests.

In chapter 3 we have this text:

> This test depends on the `charm` fixture so that the test fails immediately if a `.charm` file isn't available.

I'd remove that from chapter 3 and move it to chapter 1, after the "This test depends on two fixtures" bullets. Something like:

> These tests depend on two fixtures:
>
> - `charm` - ...
> - `juju` - ...
>
> Both tests depend on the `charm` fixture even though only `test_deploy` uses the fixture. This ensures that the tests fail immediately if a `.charm` file isn't available.

examples/k8s-3-postgresql/tests/integration/test_charm.py
tromai: Oh, I only moved this function because it follows the [tutorial](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#write-an-integration-test) where we said "add the test_database_integration" test. 
I thought it's more natural for `test_database_integration` to be after `test_workload_version_is_set`. It works well as you confirmed, but I'm happy to revert it back this change doesn't make sense :thinking:

examples/k8s-5-observe/tests/integration/test_charm.py
tromai: I agree. I will update them. Thanks for the catch.

examples/k8s-5-observe/tests/integration/test_charm.py
tromai: Thanks for the catch. I will change them.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -480,40 +480,16 @@ Now run `tox -e unit` to make sure all test cases pass.
 
 ## Write an integration test
 
-Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. Let's tweak our existing integration test `test_deploy` accordingly, to expect blocked status in `juju.wait`. Replace the contents of `tests/integration/test_charm.py` with:
+Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. In your `tests/integration/test_charm.py` file, update `test_deploy` to expect blocked status:
```
tromai: This is my attempt to simplify the description. I am happy to keep the "Let's tweak ..." if that is preferred

```
Let's tweak our existing integration test `test_deploy` accordingly, to expect blocked status in `juju.wait`. 
Update `test_deploy` in your `tests/integration/test_charm.py` with:
```

dwilding: It looks good. I might make one tiny simplification:

```suggestion
Now that our charm integrates with the database, if there's not a database relation, the app will be in `blocked` status instead of `active`. In `tests/integration/test_charm.py`, update `test_deploy` to expect blocked status:
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -480,40 +480,16 @@ Now run `tox -e unit` to make sure all test cases pass.
...
-        "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
-    }
-
-    # Deploy the charm and wait for it to report blocked, as it needs Postgres.
-    juju.deploy(charm, app=APP_NAME, resources=resources)
-    juju.wait(jubilant.all_blocked)
+```diff
+-    juju.wait(jubilant.all_active)
++    juju.wait(jubilant.all_blocked)
 ```
```
dwilding: I hadn't thought of using a diff - we rarely use them (I see only one other instance in the Ops docs). Definitely not against it! I'd like to see @tonyandrewmeyer's opinion on whether the intent is clear enough.

examples/k8s-3-postgresql/tests/integration/test_charm.py
dwilding: I'm good with the order as you have it now - unless @tonyandrewmeyer has any objections based on something I didn't think of. I haven't actually run the integration tests, but I did run them quickly with jjx (postgres support coming soon :dart:) and everything passed.

dwilding (approved):
Nice work!

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: I just remembered one other thing, while we're here. This says "a test" and "this test", but now there are two test functions. We also say "run the test" below, so there are a bunch of places that would need updating if we want to change "test" -> "tests". I think it's worth changing, as later in the tutorial we refer to the integration tests (collectively), so we ought to do the same here.

tromai: I agree that it's worth changing. I have updated them in https://github.com/canonical/operator/pull/2638/commits/27dfd0591392535f00635566de56b2381f15f248.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -480,9 +480,9 @@ You can ensure this by writing integration tests for your charm. In the charming
 
 In this section we'll write a small integration test to check that the charm packs and deploys correctly.
 
-### Write a test
+### Write integration tests
 
-Let's write the simplest possible integration test, a [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)). This test will deploy the charm, then verify that the installation event is handled without errors.
+Let's write some simplest possible integration tests, [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)). These tests will deploy the charm, verify that the installation event is handled without errors and the workload version is set correctly.
```
dwilding: I question whether they are the simplest _possible_ 😆

"simple" is also not an ideal word to have in the docs - we think they are simple, but someone with less experience might not. How about adjusting like this:

```suggestion
Let's write some integration tests as [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)). These tests will deploy the charm, verify that the installation event is handled without errors and the workload version is set correctly.
```

tromai:
Another observation I have, we instructed the readers to run integration tests in 2 ways:
``````
In your Multipass Ubuntu VM, run the tests again:

```text
ubuntu@juju-sandbox-k8s:~/fastapi-demo$ tox -e integration
```
``````
[here](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#write-an-integration-test), and
``````
Run the following command from anywhere in the `~/fastapi-demo` directory:

```text
tox -e integration
```
``````
[here](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm/#id5) and [here](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/observe-your-charm-with-cos-lite/#run-the-tests). 

In the [`setup your development environment`](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/set-up-your-development-environment/#set-up-your-development-environment) chapter, we instructed the reader to switch to the VM, so I assume we don't need the "In your Multipass Ubuntu VM" part anymore ?

File: docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md [hunk 13]
```diff
@@ -480,9 +480,9 @@ You can ensure this by writing integration tests for your charm. In the charming
 
 In this section we'll write a small integration test to check that the charm packs and deploys correctly.
```
dwilding: I think we can remove this line. We don't need it because the subsection is right after. Also, it talks about checking that the charm packs, which the integration tests don't do.

dwilding (approved):

dwilding:
> I assume we don't need the "In your Multipass Ubuntu VM" part anymore ?

Yeah, I think it's fine to remove this - good thinking. Earlier in the tutorial we established that commands are run in the VM, even if the reader is editing files using their regular editor.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -510,33 +508,35 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
...
 ```
 
-This test depends on two fixtures:
+These tests depend on two fixtures:
 
 - `charm` - The `.charm` file to deploy. This fixture is defined in `tests/integration/conftest.py`.
 - `juju` - A Jubilant object for interacting with a temporary Juju model. This fixture is provided by the `pytest-jubilant` plugin.
 
-### Run the test
+Both tests depend on the `charm` fixture even though only `test_deploy` uses the fixture. This ensures that the tests fail immediately if a `.charm` file isn't available.
```
tonyandrewmeyer, line 522: I guess this is because we don't have `abort_on_fail` yet. I'm not a huge fan of this approach, but perhaps it's ok for a tutorial, where you don't want to have to build something more complex.

It also seems a bit strange if you skip setup, where you presumably have the deployed app already, and don't actually need the charm, although probably it is findable and the cost is low, since it needs to already have been packed.

Is this a pattern we are creating or something we see in real charms? I see we already do this with the database test, is this just documenting something we already decided to do?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tonyandrewmeyer: I'm not keen on the diff, because it seems harder to both maintain and to use, although perhaps easier to read. For maintenance, we want to load this in from the actual files, and I don't think that can be done as a diff, so we would have to manually re-create the diff every time the code changes. For use, I think it's much simpler to hit the copy button and then paste than to apply a patch.

However, if this is the only line that's changing, then I think we are showing too much of the file. Just the test would be enough.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -525,17 +501,12 @@ def test_database_integration(charm: pathlib.Path, juju: jubilant.Juju):
...
-    version = juju.status().apps[APP_NAME].version
-    assert version == "1.0.4"  # Hardcoded for simplicity.
 ```
 
-This test depends on the `charm` fixture so that the test fails immediately if a `.charm` file isn't available.
-
-In your Multipass Ubuntu VM, run the tests again:
+Run the following command from anywhere in the `~/fastapi-demo` directory:
 
 ```text
```
tonyandrewmeyer: If we're dropping the leading text, shouldn't this be shell or terminal now?

examples/k8s-3-postgresql/tests/integration/test_charm.py
```diff
@@ -27,25 +27,26 @@
...
-    """Deploy the charm under test.
-
-    Assert on the unit status before any relations/configurations take place.
-    """
+    """Deploy the charm under test."""
     resources = {
         "demo-server-image": METADATA["resources"]["demo-server-image"]["upstream-source"]
     }
-
-    # Deploy the charm and wait for it to report blocked, as it needs Postgres.
```
tonyandrewmeyer, line 44: Why drop this comment? It seems useful to explain why we are waiting for blocked rather than active.

examples/k8s-3-postgresql/tests/integration/test_charm.py
```diff
@@ -54,7 +55,4 @@ def test_database_integration(charm: pathlib.Path, juju: jubilant.Juju):
     """
     juju.deploy("postgresql-k8s", channel="14/stable", trust=True)
     juju.integrate(APP_NAME, "postgresql-k8s")
-    juju.wait(jubilant.all_active, timeout=10 * 60)
-
-    version = juju.status().apps[APP_NAME].version
-    assert version == "1.0.4"  # Hardcoded for simplicity.
+    juju.wait(jubilant.all_active)
```
tonyandrewmeyer, line 63: Why drop the timeout? Do we expect this to finish in the default 3m? Although I see it was already dropped in the next two chapters. Is this a drive-by cleanup?

examples/k8s-3-postgresql/tests/integration/test_charm.py
tonyandrewmeyer: Either order seems fine to me.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: It's the approach we previously standardised on in our docs and example charms.

examples/k8s-3-postgresql/tests/integration/test_charm.py
dwilding: I felt it was OK to drop this in service of the diff approach, as the tutorial already explains what's happening. In your other comment, I think you make a good case against the diff. I'll reply there too.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
dwilding: This is a convincing reason not to do a diff. I'm good if we show the new version of the `test_deploy` function, which can retain the comment about why we're waiting for blocked status.

examples/k8s-3-postgresql/tests/integration/test_charm.py
tonyandrewmeyer: I don't think the code should rely on someone also reading the tutorial to be easily understood.

examples/k8s-3-postgresql/tests/integration/test_charm.py
tromai: This is a drive-by cleanup as mentioned [here](https://github.com/canonical/operator/pull/2638#discussion_r3555626787). 
We mentioned increasing the timeout duration as an optional step (see [here](https://canonical.com/juju/docs/ops/latest/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/#write-an-integration-test)):
```
If the tests fail with a timeout error,
override the default timeout in test_database_integration:
```
so I thought it's okay to remove the timeout here.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tromai: I used `text` here to be consistent with other chapters where we also instruct readers to run `tox -e integration`. I agree that `shell` is better, I will change them.

examples/k8s-3-postgresql/tests/integration/test_charm.py
tromai: Thanks for the question. It's a good point to keep the code understandable by itself. I have added this comment back in https://github.com/canonical/operator/pull/2638/commits/8b27c59377c57927668b9528c30f0aa45ddb8bf9

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tromai: Thanks for the feedback. I added this function back in https://github.com/canonical/operator/pull/2638/commits/8b27c59377c57927668b9528c30f0aa45ddb8bf9.
It's a good point on copy-and-paste experience for readers.

tonyandrewmeyer (approved):

PR #2645, opened 2026-07-14
ci: standardise and consolidate pre-commit and non-hook linting

tonyandrewmeyer:
Adjust the pre-commit and tox-based linting:

* Ensure dependabot tracks updates to pre-commit hooks, and that the pre-commit version is tracked in pyproject.toml.
* Run pyright as part of pre-commit (annoyingly slow, but eventually we'll move to a faster type check tool and this does really belong there).
* Define the linting commands only once by having tox call the precommit command runner. This also means `tox` and CI run the extra precommit checks.

Also has some small fixes via running precommit (whitespace normalisation, executable bit).


.github/dependabot.yaml
```diff
@@ -5,9 +5,7 @@
 version: 2
 
 updates:
-  # ===================================================================
-  # GitHub Actions: routine lane (monthly, single grouped PR)
-  # ===================================================================
+  # GitHub Actions: routine lane (monthly, single grouped PR).
```
tonyandrewmeyer:
```suggestion
  # GitHub Actions: monthly, single grouped PR.
```

.github/dependabot.yaml
```diff
@@ -24,9 +22,24 @@ updates:
         patterns:
           - "*"
 
-  # ===================================================================
-  # Python (uv), root: routine lane (monthly, grouped along three seams)
-  # ===================================================================
+  # pre-commit hooks: routine lane (monthly, single grouped PR).
```
tonyandrewmeyer:
```suggestion
  # pre-commit hooks: monthly, single grouped PR.
```

.github/dependabot.yaml
```diff
@@ -24,9 +22,24 @@ updates:
...
+    commit-message:
+      prefix: "chore"
+    cooldown:
+      default-days: 7
+    groups:
+      pre-commit:
+        patterns:
+          - "*"
+
+  # Python (uv), root: routine lane (monthly, grouped along three seams).
```
tonyandrewmeyer:
```suggestion
  # Python (uv), root: monthly, grouped into seams.
```

.github/dependabot.yaml
```diff
@@ -74,9 +87,7 @@ updates:
           - "minor"
           - "patch"
 
-  # ===================================================================
-  # Python (uv), examples/httpbin-demo: routine lane (same shape as root)
-  # ===================================================================
+  # Python (uv), examples/httpbin-demo: routine lane (same shape as root).
```
tonyandrewmeyer:
```suggestion
  # Python (uv), examples/httpbin-demo: same shape as root.
```

.pre-commit-config.yaml
```diff
@@ -37,6 +37,12 @@ repos:
         entry: uv run --group lint -- codespell
         language: system
         types: [text]
+      - id: pyright
+        name: pyright (uv)
+        entry: uv run --group lint -- pyright
+        language: system
+        types: [python]
+        pass_filenames: false
```
james-garner-canonical, line 48: What is the default behaviour of `pass_filenames`? I'm wondering if it's `true` by default, meaning that our `ruff check` and `ruff format` calls will have different semantics when run via `pre-commit` to previously. For example, an edit made in one file might result in `ruff` wanting changes in another (non-edited) file, most notably when editing `pyproject.toml`.

james-garner-canonical (changes requested):
> Run pyright as part of pre-commit (annoyingly slow, but eventually we'll move to a faster type check tool and this does really belong there).

I don't think we should do this. `pre-commit`'s messaging is very clear on not putting slow checks there. Perhaps worth a team discussion if you feel strongly that some slow checks belong in `pre-commit`. I would just keep the `pyright` call in `tox.ini`'s `lint` environment.

benhoyt (commented):
Hmm, on playing with this a bit, I'm also not sure it's a good idea to change the local `tox -e lint`. I think I'd prefer to leave that alone, and only folks who opt in to pre-commit get the new checks. Can we add this to pre-commit and CI only (auto), but not to tox (manual)?

In addition, when I `pre-commit install` and try to `git commit -m test` locally, I get 3049 PyRight errors -- is that just me?

As far as the speed goes, I'm on the fence. 5s is slow for a git commit, but doesn't feel crazy slow to me. I usually run unit tests using `tox -e unit` before committing anyway, which takes about 30s.

tonyandrewmeyer:
> `pre-commit`'s messaging is very clear on not putting slow checks there.

Slow is subjective, of course. On my corporate laptop, it's about 4s for pyright. The unit tests are about 4x slower. ty is about 0.3s, pyrefly about 0.5s. flake8 takes about 8s, pylint about 15s. Where does the "slow" line sit?

> Perhaps worth a team discussion if you feel strongly that some slow checks belong in `pre-commit`. I would just keep the `pyright` call in `tox.ini`'s `lint` environment.

My understanding was that I had to propose keeping the "lint" recipe in one place, which that would violate. I would personally rather we didn't add pyright to precommit until we move to a sub 1s check. But I would rather we had precommit with a slow check than no precommit.

james-garner-canonical:
@tonyandrewmeyer said:

> My understanding was that I had to propose keeping the "lint" recipe in one place, which that would violate. I would personally rather we didn't add pyright to precommit until we move to a sub 1s check. But I would rather we had precommit with a slow check than no precommit.

To me, a tox recipe that calls `pre-commit` for the fast checks, and then also runs some slow checks still seems single source of truth -- the fast checks are defined in just one place, and `tox` only *adds* extra slow tests (pyright, unit tests, integration tests), there aren't any checks that only live somewhere else. If do we want the `pre-commit` to correspond 1:1 to a `tox` environment, then we could add a `fast-lint` environment, and have `lint` be `pyright` + `fast-lint`.

---

@benhoyt said:

> Hmm, on playing with this a bit, I'm also not sure it's a good idea to change the local `tox -e lint`. I think I'd prefer to leave that alone, and only folks who opt in to pre-commit get the new checks. Can we add this to pre-commit and CI only (auto), but not to tox (manual)?

I'm curious why this is.

I don't like the idea of CI only checks that could trivially run locally (like trailing whitespace, or malformed YAML, for example). I suppose CI + pre-commit checking of things like that is better than only checking in pre-commit though.

But I also don't love how (at least as currently configured), `pre-commit` swallows all the feedback/output from checks that pass (and doesn't show it live), and I'm also concerned about the semantics of running the checks via `pre-commit`. For example:
- Does `pre-commit run --all-files` include untracked files (previously covered by linting)?
- Can `lint` now make changes, whereas previously it read only? After changing some linting rules in `pyproject.toml` and running `tox -e lint`, I see the message `pre-commit hook(s) made changes.` I guess this could be (e.g.) whitespace trimming.
- Why do I see the message "All changes made by hooks:" followed by the uncommitted changes that I made to test `tox -e lint`? `pre-commit` thinks any uncommitted changes are changes made by hooks?

.pre-commit-config.yaml
tonyandrewmeyer: > What is the default behaviour of `pass_filenames`? 

[True](https://pre-commit.com/#creating-new-hooks).

> I'm wondering if it's `true` by default, meaning that our `ruff check` and `ruff format` calls will have different semantics when run via `pre-commit` to previously. For example, an edit made in one file might result in `ruff` wanting changes in another (non-edited) file, most notably when editing `pyproject.toml`.

I think *only* with `pyproject.toml`, no? But I can add `pass_filenames: false` to the others. I think ruff is fast enough that the advantage of only running on specific files is minimal, and when run as part of the commit hook everything unmodified is stashed anyway.

tonyandrewmeyer:
\[Ben\]
> Hmm, on playing with this a bit, I'm also not sure it's a good idea to change the local `tox -e lint`. I think I'd prefer to leave that alone, and only folks who opt in to pre-commit get the new checks. Can we add this to pre-commit and CI only (auto), but not to tox (manual)?

This is easily done, except that it will violate the "define what lint means in one place" rule that I thought I was meant to follow.

I think it's an inferior experience, since you have to manually fix whatever the pre-commit checks have found (broken JSON, trailing whitespace, and so on), rather than it being fixed for you in most cases. But perhaps that's what you'd prefer?

\[James\]
> To me, a tox recipe that calls `pre-commit` for the fast checks, and then also runs some slow checks still seems single source of truth

I still have trouble with "slow" and "fast" as ways of delineating tests, when we don't define what "slow" means.

> -- the fast checks are defined in just one place, and `tox` only _adds_ extra slow tests (pyright, unit tests, integration tests),

We already decided that type checking should live in `lint`. That was not based on speed, but based on "linting" logically including static type checking as well as other static checks ("static" might have been a better name, but we were going with the existing names to make transition easier, I think). I would rather stick with the logical naming than bundling by a time constraint.

I think it's reasonably clear that you don't want integration tests running on every commit, because you don't always have the infrastructure available to run integration tests. Maybe something like workshop changes this, but for this project at least, not until it's able to have a Juju running in the workshop.

I think it's less obvious whether or not you'd want to run other types of tests. Functional tests probably have similar issues in terms of what's installed, but there probably workshop can already solve those issues. Unit tests (if done properly) should be able to be run. I think probably it's taste more than anything else as to whether you run unit tests on every commit or not. I feel like in an agentic world running them is likely best, but I understand that other people might not want to.

What I would like is something like this, but with the actual commands:

```yaml
repos:
  - repo: local
    hooks:
      - id: lint
        name: Lint
        entry: ruff check
        language: system
        pass_filenames: false
        stages: [pre-commit]

      - id: unit
        name: Unit tests
        entry: pytest
        language: system
        pass_filenames: false
        stages: [pre-push]
```

But this felt like a step too far, considering that the team is dubious about pre-commit at all. I imagine that most of us are doing something similar to this (running unit tests before pushing) anyway.

\[James\]
> But I also don't love how (at least as currently configured), `pre-commit` swallows all the feedback/output from checks that pass (and doesn't show it live),

Ah, interesting. I like the "only show me when there's a problem" conciseness. Adding `-v` is obviously simple to do and if that's the majority choice, or simply the preference because it's more similar to what we have at the moment, I'm fine adding that.

> and I'm also concerned about the semantics of running the checks via `pre-commit`. For example:
> 
> * Does `pre-commit run --all-files` include untracked files (previously covered by linting)?

Yes, a plain `run` doesn't do the stashing that running via the hook does.

> * Can `lint` now make changes, whereas previously it read only? After changing some linting rules in `pyproject.toml` and running `tox -e lint`, I see the message `pre-commit hook(s) made changes.` I guess this could be (e.g.) whitespace trimming.

Yes. This would probably be a good discussion. Other linting jobs in the charming world also do this (have ruff fix things, for example). I have mixed feelings about this - on the one hand it's very convenience, but on the other it is good to verify things.

> * Why do I see the message "All changes made by hooks:" followed by the uncommitted changes that I made to test `tox -e lint`? `pre-commit` thinks any uncommitted changes are changes made by hooks?

Where do you see this? For example, I added some trailing whitespace to README.md and I see:

```
❯ tox -e lint
lint: uv-sync> uv sync --locked --python-preference system --no-default-groups --group lint --group unit -p /home/tameyer/.local/share/uv/tools/tox/bin/python
lint: commands[0]> pre-commit run --all-files --show-diff-on-failure
check for added large files..............................................Passed
check python ast.........................................................Passed
check for case conflicts.................................................Passed
check that executables have shebangs.....................................Passed
check that scripts with shebangs are executable..........................Passed
check for merge conflicts................................................Passed
check for broken symlinks................................................Passed
check json...............................................................Passed
check yaml...............................................................Passed
check toml...............................................................Passed
mixed line ending........................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing README.md

detect private key.......................................................Passed
ruff (uv)................................................................Passed
ruff-format (uv).........................................................Passed
codespell (uv)...........................................................Passed
pyright (uv).............................................................Passed
zizmor (uv)..............................................................Passed
lint: exit 1 (8.44 seconds) /tmp/test-precom> pre-commit run --all-files --show-diff-on-failure pid=196462
  lint: FAIL code 1 (8.47=setup[0.02]+cmd[8.44] seconds)
  evaluation failed :( (8.48 seconds)
```

I'm guessing you're seeing something because of `--show-diff-on-failure`, which we can remove if that's preferred. But can you explain how to reproduce?

\[Ben\]
> In addition, when I pre-commit install and try to git commit -m test locally, I get 3049 PyRight errors -- is that just me?

Was this just literally those two commands? I did a fresh clone, checked out this branch, `pre-commit install` and `git commit -m test` and I get "skipped" for all the checks (including pyright), which is what I'd expect, because there's nothing changed. A `tox -e lint` gives "passed" for everything.

james-garner-canonical:
> We already decided that type checking should live in lint. That was not based on speed, but based on "linting" logically including static type checking as well as other static checks ("static" might have been a better name, but we were going with the existing names to make transition easier, I think). I would rather stick with the logical naming than bundling by a time constraint.

Yes, `tox -e lint` should still run `ruff`, `pyright` and any other linters. This doesn't preclude these being split into a fast group, suitable to run in `pre-commit`, and a slow group. `pyright` seems too slow to me, but if the `pre-commit` users in the team are happy with the speed of `pyright`, then I don't object to it being in `pre-commit`.

> What I would like is something like this, but with the actual commands:

I have no objection to `pre-commit` being purely a consumer of `tox` and running `tox -e lint` (and presumably `tox -e format`) and `tox -e unit` if both commands are deemed fast enough (though this would mean that whitespace checking and so on would need to be somehow defined somewhere `tox` can reach them).

I also don't mind `pre-commit` being consumed by `tox` (like in this PR) if we want to define our linting and unit recipes there and call them from the appropriate `tox` commands (looks like that would be possible by passing a hook ID to `pre-commit run`). But see also the next comment below.

> > But I also don't love how (at least as currently configured), pre-commit swallows all the feedback/output from checks that pass (and doesn't show it live),

> Ah, interesting. I like the "only show me when there's a problem" conciseness. Adding -v is obviously simple to do and if that's the majority choice, or simply the preference because it's more similar to what we have at the moment, I'm fine adding that.

This raises an interesting issue that may be a blocker: can we still pass extra arguments through to underlying commands when running via `pre-commit`? I think you mean adding `-v` to our `tox` call to `pre-commit`, but it reminded me that `tox -e something -- some extra args` is a common pattern.

> I'm guessing you're seeing something because of --show-diff-on-failure, which we can remove if that's preferred. But can you explain how to reproduce?

I deleted some linting ignores from `pyproject.toml` and ran `tox -e lint`. The uncommitted deletions in `pyproject.toml` are displayed as "changes made by hooks".

---

I raised one issue that I don't think has been addressed: are we happy with `tox -e lint` now modifying files, overlapping with `tox -e format` (whitespace trimming)?

tonyandrewmeyer:
> This doesn't preclude these being split into a fast group, suitable to run in `pre-commit`, and a slow group.

I can't get on board with any sort of slow/fast split until it comes with a definition of slow.

> > What I would like is something like this, but with the actual commands:
> 
> I have no objection to `pre-commit` being purely a consumer of `tox`

Sorry, I think my example was not presented well enough. The point I was trying to make was having stages be pre-commit for linting and pre-push for unit tests. I don't like pre-commit running tox/make.

> This raises an interesting issue that may be a blocker: can we still pass extra arguments through to underlying commands when running via `pre-commit`? I think you mean adding `-v` to our `tox` call to `pre-commit`, but it reminded me that `tox -e something -- some extra args` is a common pattern.

For unit tests it is, but this is already somewhat broken with linting (this also came up in a recent PR with @tromai). If we are running multiple commands, we don't have any way with our tox or make structure to say which command gets the extra arguments. So either they need to make sense to all the commands (ruff format, ruff check, pyright, codespell, for example), or be ignored by the ones not intended to have them, or we have special names (more feasible with make than tox I think), or we don't support it.

> I raised one issue that I don't think has been addressed: are we happy with `tox -e lint` now modifying files, overlapping with `tox -e format` (whitespace trimming)?

I "addressed" it by saying it would be a good discussion :)

james-garner-canonical:
> > This doesn't preclude these being split into a fast group, suitable to run in `pre-commit`, and a slow group.
> 
> I can't get on board with any sort of slow/fast split until it comes with a definition of slow.

Anything over the order of 1 second feels slow when you're waiting for it to give you your terminal back. How about a generous threshold of 2 seconds? That said, I don't really have a horse in this race, since I don't run `pre-commit` hooks.

> > > What I would like is something like this, but with the actual commands:
> > 
> > I have no objection to `pre-commit` being purely a consumer of `tox`
> 
> Sorry, I think my example was not presented well enough. The point I was trying to make was having stages be pre-commit for linting and pre-push for unit tests. I don't like pre-commit running tox/make.

I missed that the important part of the example wasn't defining `lint` and `unit` in `pre-commit`, but having both `pre-commit` and `pre-push` checks. Again I'm not opposed in principle to the checks being defined in the `pre-commit` yaml file and callable from `tox` manually. As I said, it looks like this is possible by passing hook IDs to `pre-commit run`.

> > This raises an interesting issue that may be a blocker: can we still pass extra arguments through to underlying commands when running via `pre-commit`? I think you mean adding `-v` to our `tox` call to `pre-commit`, but it reminded me that `tox -e something -- some extra args` is a common pattern.
> 
> For unit tests it is, but this is already somewhat broken with linting (this also came up in a recent PR with @tromai). If we are running multiple commands, we don't have any way with our tox or make structure to say which command gets the extra arguments. So either they need to make sense to all the commands (ruff format, ruff check, pyright, codespell, for example), or be ignored by the ones not intended to have them, or we have special names (more feasible with make than tox I think), or we don't support it.

Yes, it's tricky with multiple checks. For linting I've typically forwarded extra args to `ruff check --fix` since it's often useful to pass `--unsafe-fixes` and I don't often need to forward anything to the other commands. However, that's orthogonal to the `pre-commit` discussion.

What I wanted to know is whether `pre-commit` supports passing extra arguments to a specific check. It seems that it doesn't. I guess your argument would be that we live without that for linting already? Though that seems like a mark against the unit test command being defined in `pre-commit`.

> > I raised one issue that I don't think has been addressed: are we happy with `tox -e lint` now modifying files, overlapping with `tox -e format` (whitespace trimming)?
> 
> I "addressed" it by saying it would be a good discussion :)

Thanks, I missed that comment.

tonyandrewmeyer:
> > I can't get on board with any sort of slow/fast split until it comes with a definition of slow.
> 
> Anything over the order of 1 second feels slow when you're waiting for it to give you your terminal back. How about a generous threshold of 2 seconds?

This is a start, but is that 2 seconds total or 2 seconds per check? On whose box? Based on the numbers in this discussion, the unit tests take about twice as long for Ben as for me, both of us on corporate laptops, for example.

I can be convinced to split by time if we define it well enough, but I still don't like it. Note that any sort of "2s" rule would have disallowed all the linters we used before ruff, and probably would allow type checking when/if we move to one of the modern implementations. That means the definition moves based on the speed of tools rather than what tools *do*, which is an inferior split in my opinion.

> For linting I've typically forwarded extra args to `ruff check --fix` since it's often useful to pass `--unsafe-fixes` and I don't often need to forward anything to the other commands.

That's fine if you're running it manually, but this causes most of the lint commands to error. I feel passing extra arguments only works reliably if all the env/target commands share the same args (not even true for ruff, sadly) or if there is a single command. So this seems equally broken if we are tox/make lint'ing or if we are precommit linting.

> However, that's orthogonal to the `pre-commit` discussion.

Somewhat, but you presented not being able to pass through arguments as an issue, and my counter is that it's broken either way, so not an issue.

Our commands (other than the ones from pre-commit-hooks) are "local" so we can tweak them. For example, we could have `entry: bash -c 'ruff check $RUFF_ARGS "$@"' --` and then pass in arguments with that env var. It's not the nice `-- extra stuff here` that tox offers, but we already decided to lose that by moving to Make.

> Though that seems like a mark against the unit test command being defined in `pre-commit`.

I am not suggesting that we have pre-commit run or define the unit command. This is purely about lint. Perhaps my "ideal world" comment muddied the waters -- if so, please just ignore that whole bit.

tromai:
Thanks everyone for the discussion! 

It took me a bit of time to digest the PR. This is a draft of my opinions so far:

- One source of truth: 
    - I'm happy with the direction to keep the configuration at one place (from what I see, we are going for `.pre-commit-config.yaml`).
    - I also see that we use CI as the gate keeper. As in, no matter what hooks I ran/skip locally, CI's `precommit run --all-files` will be the final judge. I'm okay with this. For me, there are 2 sides: 1. when prototyping, I can rapidly commit + push (using `git commit --no-verify` to skip `pre-commit` hooks). This is okay for prototyping, where I want to iterate on my ideas quickly; 2. When I'm ready to open a PR, I will split the changes and commit them again with pre-commit hooks enabled. I prefer a setup where I can either run `tox -e lint/unit` locally or leave it to CI (**no pre-push hook would be good**). I expect `tox -e lint` command to be **consistent** running locally vs CI (as in, they should have the same result).
- About `pyright`: I gave this hook a try, and felt that it's visibly slower compared to hooks like ruff or trailing white spaces. But it's not a deal breaker for me. My git workflow would run it at the very end before opening a PR, so it's not a big problem. I can see the benefit for running with agents from Tony's comment.
- I agree that unit tests should not be added as a pre-commit hook, and I would prefer running it separately, where I have more control (e.g. run unit against one test module, run unit tests with extra parameters).
- For me, it's more natural to have ruff in pre-commit hook fix issues automatically. This aligns with other hooks like trailing white spaces. I don't think this would be an issue running in CI, because it still exits on non-zero status code.

They are not final so I'm happy to discuss them further. 

There were some other topics in this PR that I didn't mention here, as I don't have a strong opinion about them (e.g. `--show-diff-on-failure`, pre-commit swallowing tools output).

james-garner-canonical:
> > > I can't get on board with any sort of slow/fast split until it comes with a definition of slow.
> > 
> > 
> > Anything over the order of 1 second feels slow when you're waiting for it to give you your terminal back. How about a generous threshold of 2 seconds?
> 
> This is a start, but is that 2 seconds total or 2 seconds per check? On whose box? Based on the numbers in this discussion, the unit tests take about twice as long for Ben as for me, both of us on corporate laptops, for example.
> 
> I can be convinced to split by time if we define it well enough, but I still don't like it. Note that any sort of "2s" rule would have disallowed all the linters we used before ruff, and probably would allow type checking when/if we move to one of the modern implementations. That means the definition moves based on the speed of tools rather than what tools _do_, which is an inferior split in my opinion.

Sounds like you fundamentally don't consider speed to be a relevant factor for what gets run in `pre-commit` hooks, which is fine, but good context to have.

---

Regarding passing extra arguments, it sounds like your position is indeed that we live without that for linting already since it's broken right now (tries to pass arguments to both `pyright` and `codespell`).

I don't understand the syntax of your example, but it seems like we could do something with environment variables to pass extra arguments, which is a fine workaround, and arguably better than `{posargs}` since you can route to different commands, so perhaps this is a non-issue either way.

---

Happy to leave unit tests out of the discussion, though I don't fully understand the principle by which they're excluded -- is it just that they're not part of this PR?

tonyandrewmeyer:
> Sounds like you fundamentally don't consider speed to be a relevant factor for what gets run in `pre-commit` hooks, which is fine, but good context to have.

I personally feel like speed isn't the choice, and if speed is a problem the solution is making things faster rather than redefining what should be run. But this does seem somewhat a personal choice.

> Happy to leave unit tests out of the discussion, though I don't fully understand the principle by which they're excluded -- is it just that they're not part of this PR?

I don't think there was any suggestion that we should move unit tests. The problems that triggered the team discussion were all about adding extra lint checks. I mentioned unit tests above because I feel that an ideal situation would be that there was automatic running of unit tests too, but I don't think anyone has proposed making changes to unit (or format or integration).

james-garner-canonical (approved):
Per team discussion, happy to try it this way.

tromai (approved):

PR #2648, opened 2026-07-16
docs: use autofunction for `layer_from_rockcraft`

tonyandrewmeyer:
`layer_from_rockcraft` is a module-level function, but was documented with the `automethod` directive. Sphinx's method documenter assumes the first positional parameter is `self` and strips it from the rendered signature, which hid the path argument. Use `autofunction` so the full signature is shown.

[Current](https://canonical.com/juju/docs/ops/latest/reference/ops-testing/#ops.testing.layer_from_rockcraft), [preview](https://canonical-juju-ops--2648.com.readthedocs.build/2648/reference/ops-testing/#ops.testing.layer_from_rockcraft)

### DIFF
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1413,7 +1413,7 @@ def layer_from_rockcraft(path: pathlib.Path | str) -> pebble.Layer:
     """Create a layer from a `rockcraft.yaml` file.
 
     This is a convenience function to create a Pebble layer from a
-    rockcraft.yaml file, that can then be passed to :class:`testing.Container`,
+    rockcraft.yaml file, that can then be passed to :class:`Container`,
     rather than duplicating the layer content in the test code. For example::
 
         container = Container(
--- a/testing/src/scenario/state.py
+++ b/testing/src/scenario/state.py
@@ -1416,7 +1416,7 @@ def layer_from_rockcraft(path: pathlib.Path | str) -> pebble.Layer:
     rockcraft.yaml file, that can then be passed to :class:`Container`,
     rather than duplicating the layer content in the test code. For example::
 
-        container = Container(
+        container = testing.Container(
             name='my-container',
             layers={'rock': layer_from_rockcraft(pathlib.Path('rockcraft.yaml'))},
         )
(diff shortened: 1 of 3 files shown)


dwilding (approved):
Great, thanks!

testing/src/scenario/state.py
```diff
@@ -1413,7 +1413,7 @@ def layer_from_rockcraft(path: pathlib.Path | str) -> pebble.Layer:
     """Create a layer from a `rockcraft.yaml` file.
 
     This is a convenience function to create a Pebble layer from a
-    rockcraft.yaml file, that can then be passed to :class:`testing.Container`,
+    rockcraft.yaml file, that can then be passed to :class:`Container`,
     rather than duplicating the layer content in the test code. For example::
 
         container = Container(
```
james-garner-canonical: I wonder if we should go the other way with the change to this file, and use `testing.Container` in our example, since that's how we recommend people use it.

```suggestion
        container = testing.Container(
```

james-garner-canonical (approved):
Nice fix, wondering if we really do prefer the short `Container` link to the old `testing.Container`.

testing/src/scenario/state.py
dwilding: My feeling when reading the preview was that `Container` in the text is good, because we're already in the `ops.testing` reference. I think you make a good point about having the sample code match real usage.

tonyandrewmeyer: Sure, I'm fine bundling this in.

testing/src/scenario/state.py
```diff
@@ -1413,10 +1413,10 @@ def layer_from_rockcraft(path: pathlib.Path | str) -> pebble.Layer:
     """Create a layer from a `rockcraft.yaml` file.
 
     This is a convenience function to create a Pebble layer from a
-    rockcraft.yaml file, that can then be passed to :class:`testing.Container`,
+    rockcraft.yaml file, that can then be passed to :class:`Container`,
```
james-garner-canonical, line 1416: Sorry if this part of my comment wasn't clear, I was thinking this should also be `testing.Container`.
```suggestion
    rockcraft.yaml file, that can then be passed to :class:`testing.Container`,
```

tonyandrewmeyer: Ah, but that fails, which is why it's changed. Sphinx complains about not being able to find it for some reason. Also, in the rendered docs, it'll show as a link to the proper Container class, so it's only in `help()` or similar that you'd see the whole thing.

I'd prefer to not spend the time figuring out how to make Sphinx happy with it as `testing.Container` and just go with the simple fix.

james-garner-canonical: Sounds reasonable, agreed that it's not worth fighting Sphinx on this.

PR #2649, opened 2026-07-17
docs: switch K8s tutorial charms to rock version of demo server

dwilding:
This PR updates the K8s tutorial and example charms to use version 2 of the demo server image, which is built as a rock. The rock is defined here: https://github.com/canonical/api_demo_server/blob/master/rockcraft.yaml

Switching to a rock lets us take advantage of the baked-in Pebble layer.

- `k8s-1-minimal` doesn't need to add a layer. The unit tests reference the rock's layer and verify that Pebble's plan exactly matches that layer. Notable changes to **[Create a minimal Kubernetes charm](https://canonical-juju-ops--2649.com.readthedocs.build/2649/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm/)**:
    - Added an explanation of the rock and its layer. See the text after `_on_demo_server_pebble_ready` is defined.
    - Explained when you'd be able to use `testing.layer_from_rockcraft`. See the text after the unit tests are written.

- `k8s-2-configurable` is now the first charm that adds a layer to the workload container, so that we can redefine `command`. Notable changes to **[Make your charm configurable](https://canonical-juju-ops--2649.com.readthedocs.build/2649/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable/)**:
    - Define `_get_pebble_layer` before `_replan_workload`.
    - Explained about `override: merge`. See the text after `_get_pebble_layer` is defined.
    - Update the `test_pebble_layer` test so that it expects `override: merge` in Pebble's plan.  See the beginning of the "Write unit tests" section.

- `k8s-3-configurable` extends the layer construction, so that we can add to `environment`. Notable change to **[Integrate your charm with PostgreSQL](https://canonical-juju-ops--2649.com.readthedocs.build/2649/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql/)**:
    - Explained what `override: merge` does to `environment`. See the note after `_get_pebble_layer` is updated.

    <mark>To review the changes to `k8s-3-configurable` in isolation, look at [these commits](https://github.com/canonical/operator/pull/2649/changes/BASE..74f664a580fef32ed15e53c61508ea487b3dd82d).</mark>

As I was working on the changes, I did some cleanup in **[Study your application](https://canonical-juju-ops--2649.com.readthedocs.build/2649/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/study-your-application/)** -- from "Structure" onward.

### DIFF
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -76,7 +76,7 @@ resources:
     # used by the 'canonical/charming-actions' GitHub action for automated releases.
     # The test_deploy function in tests/integration/test_charm.py reads upstream-source
     # to determine which OCI image to use when running the charm's integration tests.
-    upstream-source: ghcr.io/canonical/api_demo_server:1.0.4
+    upstream-source: ghcr.io/canonical/api_demo_server/api-demo-server:2.1.0
 ```
 
 ### Write a helper module
@@ -193,10 +193,10 @@ def _on_demo_server_pebble_ready(self, event: ops.PebbleReadyEvent) -> None:
 
 The custom Pebble layer that you just added is defined in the  `self._get_pebble_layer()` method. We'll now add this method.
 
-In the `__init__` method of your charm class, name your service to `fastapi-service` and add it as a class attribute:
+In the `__init__` method of your charm class, name your service to `fastapi` and add it as a class attribute:
 
 ```python
-self.pebble_service_name = "fastapi-service"
+self.pebble_service_name = "fastapi"
 ```
 
 Finally, define  the `_get_pebble_layer` function as below. The `command` variable represents a command line that should be executed in order to start our application.
@@ -206,10 +206,10 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
     """Pebble layer for the FastAPI demo services."""
     command = " ".join(
         [
-            "uvicorn",
+            "/bin/uvicorn",
             "api_demo_server.app:app",
-            "--host=0.0.0.0",
-            "--port=8000",
+            "--host 0.0.0.0",
+            "--port 8000",
         ]
     )
     pebble_layer: ops.pebble.LayerDict = {
@@ -217,10 +217,8 @@ def _get_pebble_layer(self) -> ops.pebble.Layer:
         "description": "pebble config layer for FastAPI demo server",
         "services": {
             self.pebble_service_name: {
-                "override": "replace",
-                "summary": "fastapi demo",
+                "override": "merge",
                 "command": command,
-                "startup": "enabled",
             }
         },
     }
@@ -281,7 +279,7 @@ Deploy the `.charm` file, as below. Juju will create a Kubernetes `StatefulSet` 
 
 ```text
 juju deploy ./fastapi-demo_amd64.charm --resource \
-     demo-server-image=ghcr.io/canonical/api_demo_server:1.0.4
+     demo-server-image=ghcr.io/canonical/api_demo_server/api-demo-server:2.1.0
 ```
 
 
@@ -305,7 +303,7 @@ Model    Controller     Cloud/Region  Version  SLA          Timestamp
 testing  concierge-k8s  k8s           3.6.13   unsupported  13:38:19+01:00
 
 App           Version  Status  Scale  Charm         Channel  Rev  Address         Exposed  Message
-fastapi-demo  1.0.4    active      1  fastapi-demo             0  10.152.183.215  no
+fastapi-demo  2.1.0    active      1  fastapi-demo             0  10.152.183.215  no
 
 Unit             Workload  Agent  Address      Ports  Message
 fastapi-demo/0*  active    idle   10.1.157.73
@@ -322,7 +320,7 @@ curl 10.1.157.73:8000/version
 You should see a JSON string with the version of the application:
 
 ```
-{"version":"1.0.4"}
+{"version":"2.1.0"}
 ```
 
 Congratulations, you've successfully created a minimal Kubernetes charm!
@@ -383,6 +381,24 @@ from ops import testing
 
 from charm import FastAPIDemoCharm
 
+# The default Pebble layer in the application image.
+# Defined in https://github.com/canonical/api_demo_server/blob/master/rockcraft.yaml
+ROCK_LAYER = ops.pebble.Layer(
+    {
+        "services": {
+            "fastapi": {
+                "override": "replace",
+                "summary": "FastAPI demo server",
+                "command": "/bin/uvicorn api_demo_server.app:app --host 0.0.0.0 --port 8000",
+                "startup": "enabled",
+                "environment": {"DEMO_SERVER_LOGFILE": "/tmp/demo_server.log"},
+                "on-success": "shutdown",
+                "on-failure": "shutdown",
+            }
+        },
+    }
+)
+
 
 def mock_get_version(port: int):
     """Get a mock version string without executing the workload code."""
@@ -396,25 +412,17 @@ def mock_version(monkeypatch: pytest.MonkeyPatch) -> None:
 
 def test_pebble_layer(mock_version):
     ctx = testing.Context(FastAPIDemoCharm)
-    container = testing.Container(name="demo-server", can_connect=True)
+    container = testing.Container(
+        name="demo-server", can_connect=True, layers={"rock": ROCK_LAYER}
+    )
     state_in = testing.State(
         containers={container},
         leader=True,
     )
     state_out = ctx.run(ctx.on.pebble_ready(container), state_in)
     # Expected plan after Pebble ready with default config
-    expected_plan = {
-        "services": {
-            "fastapi-service": {
-                "override": "replace",
-                "summary": "fastapi demo",
-                "command": "uvicorn api_demo_server.app:app --host=0.0.0.0 --port=8000",
-                "startup": "enabled",
-                # Since the environment is empty, Layer.to_dict() will not
-                # include it.
-            }
-        }
-    }
+    expected_plan = ops.pebble.Plan(ROCK_LAYER.to_dict())
+    expected_plan.services["fastapi"].override = "merge"
 
     # Check that we have the plan we expected:
     assert state_out.get_container(container.name).plan == expected_plan
@@ -422,7 +430,7 @@ def test_pebble_layer(mock_version):
     assert state_out.unit_status == testing.ActiveStatus()
     # Check the service was started:
     assert (
-        state_out.get_container(container.name).service_statuses["fastapi-service"]
+        state_out.get_container(container.name).service_statuses["fastapi"]
         == ops.pebble.ServiceStatus.ACTIVE
     )
     # Check the workload version is set:
@@ -511,7 +519,7 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
 def test_workload_version_is_set(charm: pathlib.Path, juju: jubilant.Juju):
     """Verify that the workload version has been set."""
     version = juju.status().apps[APP_NAME].version
-    assert version == "1.0.4"  # Hardcoded for simplicity.
+    assert version == "2.1.0"  # Hardcoded for simplicity.
 ```
 
 These tests depend on two fixtures:
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -186,7 +186,7 @@ def _on_demo_server_pebble_ready(self, event: ops.PebbleReadyEvent) -> None:
 
 We're using the `ActiveStatus` class to set the charm status to active. Note that almost everything you need to define your charm is in the `ops` package that you imported earlier - there's no need to add additional imports.
 
-The `api_demo_server` application is packaged as a "rock" using Canonical's OCI-compliant format for container images. The image has a Pebble layer that defines how to run the app. Since our minimal charm doesn't need to change how the app runs, our charm just needs to call `replan()` to start the service defined in the Pebble layer.
+The `api_demo_server` application is packaged as a [rock](https://documentation.ubuntu.com/rockcraft/1.19/explanation/rocks/), which is an OCI-compliant container image. The image has a Pebble layer that defines how to run the app. Since our minimal charm doesn't need to change how the app runs, our charm just needs to call `replan()` to start the service defined in the Pebble layer.
 
 To inspect the Pebble layer, see [rockcraft.yaml](https://github.com/canonical/api_demo_server/blob/master/rockcraft.yaml). Look for a service called `fastapi`:
 
--- a/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
+++ b/docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
@@ -170,64 +170,44 @@ framework.observe(self.on["demo-server"].pebble_ready, self._on_demo_server_pebb
 
 ```
 
-Next, define the event handler, as follows:
-
-We'll use the `ActiveStatus` class to set the charm status to active. Note that almost everything you need to define your charm is in the `ops` package that you imported earlier - there's no need to add additional imports.
-
-Use `ActiveStatus` as well as further Ops constructs to define the event handler, as below. As you can see, what is happening is that, from the `event` argument, you extract the workload container object in which you add a custom layer. Once the layer is set you replan your service and set the charm status to active.
-
+Next, define the event handler:
 
 ```python
 def _on_demo_server_pebble_ready(self, event: ops.PebbleReadyEvent) -> None:
     """Define and start a workload using the Pebble API."""
     # Get a reference the container attribute on the PebbleReadyEvent
     container = event.workload
-    # Add initial Pebble config layer using the Pebble API
-    container.add_layer("fastapi_demo", self._get_pebble_layer(), combine=True)
-    # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
+    # Start the service defined by the Pebble layer in the application image.
     container.replan()
     # Learn more about statuses at
     # https://documentation.ubuntu.com/juju/3.6/reference/status/
     self.unit.status = ops.ActiveStatus()
 ```
 
-The custom Pebble layer that you just added is defined in the  `self._get_pebble_layer()` method. We'll now add this method.
+We're using the `ActiveStatus` class to set the charm status to active. Note that almost everything you need to define your charm is in the `ops` package that you imported earlier - there's no need to add additional imports.
 
-In the `__init__` method of your charm class, name your service to `fastapi` and add it as a class attribute:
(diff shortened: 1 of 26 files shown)


examples/k8s-3-postgresql/src/charm.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, framework: ops.Framework) -> None:
         super().__init__(framework)
         # See 'containers' in charmcraft.yaml.
         self.container = self.unit.get_container("demo-server")
-        self.pebble_service_name = "fastapi-service"
+        self.pebble_service_name = "fastapi"
```
dwilding, line 59: The rock's Pebble layer defines a service called `fastapi`. We need to use the same service name, otherwise the container would end up with two services competing to bind to the same port.

examples/k8s-3-postgresql/src/charm.py
```diff
@@ -143,21 +143,19 @@ def _get_pebble_layer(self, port: int, environment: dict[str, str]) -> ops.pebbl
...
             [
-                "uvicorn",
+                "/bin/uvicorn",
                 "api_demo_server.app:app",
-                "--host=0.0.0.0",
-                f"--port={port}",
+                "--host 0.0.0.0",
+                f"--port {port}",
             ]
         )
```
dwilding: Match exactly how the command is defined in the rock's Pebble layer. Not strictly needed for the charm to work, but keeps things neater in the unit tests.

examples/k8s-3-postgresql/charmcraft.yaml
```diff
@@ -56,4 +56,4 @@ resources:
     # used by the 'canonical/charming-actions' GitHub action for automated releases.
     # The test_deploy function in tests/integration/test_charm.py reads upstream-source
     # to determine which OCI image to use when running the charm's integration tests.
-    upstream-source: ghcr.io/canonical/api_demo_server:1.0.4
+    upstream-source: ghcr.io/canonical/api_demo_server/api-demo-server:2.1.0
```
dwilding, line 59: The image URL is kind of ugly. I plan to rename the `api_demo_server` repo at some point (after this PR lands).

examples/k8s-3-postgresql/tests/unit/test_charm.py
```diff
@@ -68,25 +78,24 @@ def test_pebble_layer(mock_version):
...
     state_out = ctx.run(ctx.on.config_changed(), state_in)
-    command = (
-        state_out.get_container(container.name)
-        .layers["fastapi_demo"]
-        .services["fastapi-service"]
-        .command
-    )
-    assert "--port=8080" in command
+    command = state_out.get_container(container.name).plan.services["fastapi"].command
+    assert "--port 8080" in command
```
dwilding, line 91: Includes a drive-by improvement: Assert on the ultimate plan rather than the specific layer that the charm added.

examples/k8s-3-postgresql/tests/unit/test_charm.py
```diff
@@ -119,9 +130,8 @@ def test_relation_data(mock_version):
...
     state_out = ctx.run(ctx.on.relation_changed(relation), state_in)
 
-    assert state_out.get_container(container.name).layers["fastapi_demo"].services[
-        "fastapi-service"
-    ].environment == {
+    assert state_out.get_container(container.name).plan.services["fastapi"].environment == {
+        **ROCK_LAYER.services["fastapi"].environment,
         "DEMO_SERVER_DB_HOST": "example.com",
         "DEMO_SERVER_DB_PORT": "5432",
         "DEMO_SERVER_DB_USER": "foo",
```
dwilding, line 137: Includes a drive-by improvement: Assert on the ultimate plan rather than the specific layer that the charm added.

james-garner-canonical (approved):
This looks clean and reasonable to me.

examples/k8s-3-postgresql/src/charm.py
tromai: Sounds good to me. It's good to keep them consistent.

Out of curiousity: were there a reason we used this command in the rock's Pebble layer, instead of the existing uvicorn commmand (for example: why we switch to an absolute path `/bin/uvicorn`) ?

tromai (approved):
It looks great. Thanks!

The integration tests run well for me (except for that `test_workload_version_is_set` issue).

examples/k8s-3-postgresql/src/charm.py
dwilding: I don't think `/bin` was needed in the end. I had to try several approaches to get the Rock working properly with a uv-managed application. I think `/bin` was needed during one of the iterations and it stuck.

dwilding:
@james-garner-canonical & @tromai, thanks for your reviews! I have finished updating the other charms and the tutorial source. Please could you review again? See the PR description for a summary of what I changed.

The biggest thing to be aware of is that chapter 1 doesn't need to add a layer to the workload container. That's now done in chapter 2, where we set `command` in the layer. Chapter 3 sets `command` and `environment` (as you reviewed). Chapters 4 and 5 do the same as chapter 3.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -87,14 +87,45 @@ We'll define `_replan_workload` shortly.
...
+        "summary": "FastAPI demo service",
+        "description": "pebble config layer for FastAPI demo server",
+        "services": {
+            self.pebble_service_name: {
+                "override": "merge",
+                "command": command,
+            }
+        },
+    }
+    return ops.pebble.Layer(pebble_layer)
```
james-garner-canonical: How about expressing this less verbosely?
```py
    command = f"/bin/uvicorn api_demo_server.app:app --host 0.0.0.0 --port {port}"
    layer: ops.pebble.LayerDict = {
        "summary": "FastAPI demo service",
        "description": "pebble config layer for FastAPI demo server",
        "services": {self.pebble_service_name: {"override": "merge", "command": command}},
    }
    return ops.pebble.Layer(layer)
```
The longest line here is 90 characters, which is fine for code, but longer than the 80 visible character limit we get in the docs. So I guess we actually want:
```py
    cmd = f"/bin/uvicorn api_demo_server.app:app --host 0.0.0.0 --port {port}"
    services = {self.pebble_service_name: {"override": "merge", "command": cmd}}
    layer: ops.pebble.LayerDict = {
        "summary": "FastAPI demo service",
        "description": "pebble config layer for FastAPI demo server",
        "services": services,
    }
    return ops.pebble.Layer(layer)
```
Though maybe this makes the type checker angry since we don't define the whole dict in-line? I haven't checked, sorry. We could tidy that up by using a shorter attribute name, for example:
```py
    cmd = f"/bin/uvicorn api_demo_server.app:app --host 0.0.0.0 --port {port}"
    layer: ops.pebble.LayerDict = {
        "summary": "FastAPI demo service",
        "description": "pebble config layer for FastAPI demo server",
        "services": {self.service: {"override": "merge", "command": cmd}},
    }
    return ops.pebble.Layer(layer)
```

dwilding: I think it's best to apply this approach to the k8s-3 version that also sets the service environment, then backport to the k8s-2 charm. But I get what you're going for - I'll try this, thanks

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/study-your-application.md
```diff
@@ -23,40 +23,35 @@ Finally, every time a user interacts with the database, our app writes logging i
 
 To summarize, our demo application is a minimal but real-life-like application that has external API endpoints, performs database read and write operations, and collects real-time metrics and logs for observability purposes.
```
tromai: I can see we start using the short form `app` from `## Features`. Should we make this change here too:

```suggestion
To summarize, our demo app is a minimal but real-life-like application that has external API endpoints, performs database read and write operations, and collects real-time metrics and logs for observability purposes.
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: Not relevant to this PR: I think it's called `GitHub Container Registry` ([reference](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)). I assume we are referring to `ghcr` where we publish the OCI image.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -170,66 +170,44 @@ framework.observe(self.on["demo-server"].pebble_ready, self._on_demo_server_pebb
...
 
-The custom Pebble layer that you just added is defined in the  `self._get_pebble_layer()` method. We'll now add this method.
-
-In the `__init__` method of your charm class, name your service to `fastapi-service` and add it as a class attribute:
+We're using the `ActiveStatus` class to set the charm status to active. Note that almost everything you need to define your charm is in the `ops` package that you imported earlier - there's no need to add additional imports.
 
-```python
-self.pebble_service_name = "fastapi-service"
-```
+The `api_demo_server` application is packaged as a "rock" using Canonical's OCI-compliant format for container images. The image has a Pebble layer that defines how to run the app. Since our minimal charm doesn't need to change how the app runs, our charm just needs to call `replan()` to start the service defined in the Pebble layer.
```
tromai: From my understanding, a rock is an OCI-compliant container image. And `OCI` is the image format specification. I have a feeling that `Canonical's OCI-compliant format` may be ambiguous: we doesn't have our own format specification, but we build images that comply to this format specification.

Here is what I have in mind:

```
The `api_demo_server` application is packaged as a "rock", an OCI-compliant container image, using Canonical's [rockcraft](https://documentation.ubuntu.com/rockcraft/1.19/explanation/rocks/#explanation-rocks).
```

tromai (commented):
Thank you for this PR.

Some initial comments first (on `study your application` and `create a minimal k8s charm`). I will continue my review tomorrow 😄.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: Great point, thanks. I have changed to this:

> The `api_demo_server` application is packaged as a [rock](https://documentation.ubuntu.com/rockcraft/1.19/explanation/rocks/), which is an OCI-compliant container image.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: Well spotted - changed

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/study-your-application.md
dwilding: Ah, I started changing some instances as I was rewriting things, and wasn't sure how far to go in this PR. Elsewhere, I kept "application" the first time I used the word (in a while) and then shortened to "app" after that.

I've updated the chapter to use "application" when talking about general applications, then "app" when we start talking about our specific application.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
dwilding: I've switched to this, which I find nicely readable:

```python
    cmd = f"/bin/uvicorn api_demo_server.app:app --host 0.0.0.0 --port {port}"
    service: ops.pebble.ServiceDict = {
        "override": "merge",
        "command": cmd,
        "environment": env,
    }
    layer: ops.pebble.LayerDict = {
        "summary": "FastAPI demo service",
        "description": "pebble config layer for FastAPI demo server",
        "services": {self.pebble_service_name: service},
    }
    return ops.pebble.Layer(layer)
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -87,14 +87,37 @@ We'll define `_replan_workload` shortly.
...
+    }
+    layer: ops.pebble.LayerDict = {
+        "summary": "FastAPI demo service",
+        "description": "pebble config layer for FastAPI demo server",
+        "services": {self.pebble_service_name: service},
+    }
+    return ops.pebble.Layer(layer)
+```
+
+This method defines a service with `override` set to `merge`, which means that unspecified properties should match the existing service definition.
```
tromai: Just to confirm I understand this correctly: does `unspecified properties` mean properties from `ops.pebble.ServiceDict` that we don't define in `service` ?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -136,37 +159,9 @@ def _replan_workload(self) -> None:
...
-                "override": "replace",
-                "summary": "fastapi demo",
-                "command": command,
-                "startup": "enabled",
-            }
-        },
-    }
-    return ops.pebble.Layer(pebble_layer)
-```
+When the config is loaded as part of constructing the Pebble layer, if the config is invalid (in our case, if the port is set to 22), then a `ValueError` will be raised. The `_replan_workload` method handles that by logging the error and setting the status of the unit to blocked, letting the Juju user know that they need to take action.
```
tromai: I can see that the config is loaded before we construct the Pebble layer.

```
config = self.load_config(FastAPIConfig)

...

self.container.add_layer(
    "fastapi_demo", self._get_pebble_layer(config.server_port), combine=True
)
```
Should we remove the `as part of constructing the Pebble layer` in the first sentence ?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
tromai:
```suggestion
+ .apps['fastapi-demo'].version = '2.1.0'
```

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
```diff
@@ -241,25 +236,32 @@ juju status
...
+
+```python
+# Expected plan after Pebble ready with default config.
+expected_plan = ops.pebble.Plan(ROCK_LAYER.to_dict())
+expected_plan.services["fastapi"].override = "merge"
+```
+
+This is needed because the charm's `_get_pebble_layer` method sets `override` to `merge` in the layer that it constructs.
+
+Next, we'll add a test that sets the port in the input state and asserts that the port is used in the service's command in the container layer:
```
tromai: I know `container layer` is from the previous version of the tutorial. I assume `container layer` refers to `testing.Container`. I wonder if we should use `Pebble layer` instead, because that is what we want to test. Or would that be too confusing 🤔

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/integrate-your-charm-with-postgresql.md
```diff
@@ -246,38 +246,32 @@ We removed three `self.unit.status = ` lines from this version of the method. We
...
 With these changes, we've made sure that our application knows how to access the database.
 
 When Pebble starts or restarts the service:
 
 * If there's a database relation and database authentication data is available from the relation, our application can get the database authentication data from environment variables.
-* Otherwise, the service environment is empty, so our application can't get database authentication data. In this case, we'd like the unit to show `blocked` or `maintenance` status, depending on whether the Juju user needs to take action.
+* If there's no database relation or database authentication data, our application can't connect to the database. In this case, we'd like the unit to show `blocked` or `maintenance` status, depending on whether the Juju user needs to take action.
+
+```{note}
+The service defined in [rockcraft.yaml](https://github.com/canonical/api_demo_server/blob/master/rockcraft.yaml) has an environment variable related to logging. When Pebble combines the rock's layer with our constructed layer, the final service environment has variables from both layers (because our constructed layer sets `override` to `merge`).
```
tromai: > the final service environment has variables from both layers (because our constructed layer sets `override` to `merge`).

I think we are referring to the environment variables here 

```
"environment": env
```

`variables` alone could be mistaken as a Python variables. May I suggest we change this sentence to something like

```suggestion
The service defined in [rockcraft.yaml](https://github.com/canonical/api_demo_server/blob/master/rockcraft.yaml) has an environment variable related to logging. When Pebble combines the rock's layer with our constructed layer, the final service has environment variables from both layers (because our constructed layer sets `override` to `merge`).
```

tromai (commented):
This tutorial rocks !

Thank you. I think it's in great shape. I had some comments on the wordings.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
dwilding: That's what I mean, yeah. Do you think there's a better way to say it?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
dwilding: Great point, thanks. I think that paragraph is very redundant with the code. I'd like to simplify it to:

> If the loaded config is invalid (in our case, if the port is set to 22), we set the status of the unit to blocked. This lets the Juju user know that they need to take action.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
dwilding: I totally agree. This is the only instance of the term "container layer" in our whole docs. I'll change to "Pebble layer".

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tromai: Thanks for confirming. I think the current way is okay, I'm happy with it.

This is what I had in mind, but didn't think it is better:

```suggestion
This method defines a service with `override` set to `merge`, which means that unspecified properties should inherit from the existing service definition.
```

Since `match` (to me) would indicate that the properties are defined, and contradict `unspecified properties`.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
tromai: The new paragraph looks great to me. Thanks!

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/make-your-charm-configurable.md
dwilding: Recording internal discussion: We decided to go with "unspecified properties should be copied from the existing service definition".

tromai:
Thanks! The new updates look great.

PR #2651, opened 2026-07-20
ci: use uv-venv-lock-runner in tracing tox, align deps

james-garner-canonical:
This PR updates the `tracing/` `tox` config to use `uv-venv-lock-runner` like `ops` itself does. This allows us to specify the tool versions in `tracing/pyproject.toml` intsead of using ad-hoc `deps` config in `tracing/tox.ini`.

This PR was prompted by @tromai's report that  CI is [failing](https://github.com/canonical/operator/actions/runs/29561974117/job/87826210342) due to `tracing` not pinning `ruff`, so its failing on new rules from the latest release.

We should get the `ruff` version bump from `dependabot` next month and can address the new linting rules in a single PR at that time.

I think ideally we wouldn't duplicate the `ruff` (etc) versions in `tracing/pyproject.toml` and the top-level `pyproject.toml`, but this is a step in the right direction -- and maybe it's fine if `dependabot` is bumping both sets so we stay in sync.


tracing/tox.ini
```diff
@@ -2,39 +2,38 @@
...
     pyright {posargs}
 
 [testenv:format]
-deps =
-    ruff==0.7.0
-commands = ruff format --preview {toxinidir}
+dependency_groups = lint
+commands =
+    ruff format --preview {toxinidir}
+    ruff check --preview --fix {toxinidir}
```
james-garner-canonical: Drive-by alignment with the top-level `tox -e format`.

tracing/tox.ini
```diff
@@ -2,39 +2,38 @@
...
+[testenv]
+runner = uv-venv-lock-runner
+setenv =
+  PY_COLORS=1
+passenv =
+  HOME
+  PATH
+  MODEL_SETTINGS
+  SCENARIO_SKIP_CONSISTENCY_CHECKS
+  SCENARIO_BARE_CHARM_ERRORS
```
james-garner-canonical, line 14: Copied from the top level, perhaps we don't need them at all in this file. Thoughts?


```suggestion
```

tracing/pyproject.toml
```diff
@@ -43,11 +43,13 @@ dependencies = [
...
-    "ruff ~= 0.9",
+lint = [
+    "ruff==0.15.20",
+    "codespell==2.4.2",
+    "pyright~=1.1",
 ]
-testing = [
+unit = [
+    "ops[testing]",
     "pytest ~= 9.0",
```
dwilding: Matching the format of the others

```suggestion
    "pytest~=9.0",
```

tracing/tox.ini
```diff
@@ -2,39 +2,38 @@
...
-    # Pyright doesn't support editable installs, uses extraPaths in pyproject.toml.
-    # Tox weirdly tries to build this package regardless of flags like skipdist.
-    # The above line is there so that ops_tracing's ops dependency can be met.
-    pyright
-    pytest
+dependency_groups = lint, unit
 commands =
     ruff check --preview
     ruff format --preview --check
     codespell
```
dwilding, line 32: Possibly out of scope for this PR - but the top-level `tox.ini` has:

```
codespell {posargs}
```

tracing/tox.ini
```diff
@@ -2,39 +2,38 @@
...
+passenv =
+  HOME
+  PATH
+  MODEL_SETTINGS
+  SCENARIO_SKIP_CONSISTENCY_CHECKS
+  SCENARIO_BARE_CHARM_ERRORS
+
 
 [testenv:py{3.10,3.12,3.14}-unit]
+runner = uv-venv-lock-runner
```
dwilding, line 19: I don't think we need this, as we've already set the runner on `[testenv]`. (I see that the top-level `tox.ini` has the same redundancy)

tracing/tox.ini
dwilding: Would `check` then `format` be a more robust order? These [ruff docs](https://docs.astral.sh/ruff/formatter/#sorting-imports) suggest yes. But even so, I guess it's OK to keep the same pattern used elsewhere in the repo.

tracing/tox.ini
dwilding: What actually uses `MODEL_SETTINGS`? It's in the example charms and Charmcraft profiles too. It doesn't seem to be read by Jubilant, pytest-jubilant, Juju, etc. According to my agent, it's "ecosystem-wide cargo-cult".

Since we don't use `MODEL_SETTINGS` in Ops own integration tests, I think we should remove it from all `tox.ini` files except the example charms.

For the example charms and Charmcraft profiles, maybe it's a useful convention for integration tests to support if they choose to? If we don't think it's useful as a placeholder, let's just drop it. I've created https://github.com/canonical/operator/issues/2652.

---

I think it's worth keeping `HOME`, `PATH`, and the Scenario variables here. ops-tracing does use Scenario in its tests, so those variables don't seem irrelevant.

dwilding (approved):

tracing/tox.ini
james-garner-canonical: Yeah, that's presumably an error since it means that any args passed have to be compatible with both `pyright` and `codespell`.

tracing/tox.ini
tromai: I'm also happy with keeping the same pattern used in the root `tox.ini`.

I'm still not sure about the intention behind `ruff check` after `ruff format`. I can see in [Jubilant](https://github.com/canonical/jubilant/blob/2704ac02b0f3b5a4a3698aba88d61ee57bbdb187/Makefile#L24) and [charmhub-listing](https://github.com/canonical/charmhub-listing-review/blob/ad0cbb9156e01ba8164bc13752e67d5b04926a12/Makefile#L18) the format step only contains `ruff format`. But this is out-of-scope for this PR.

tromai (approved):
Thank you for the fix!

It's nice to use dependency group to get the pinned version from `pyproject.toml`.

tracing/tox.ini
james-garner-canonical: I've dropped `MODEL_SETTINGS` from this PR. Cleanup elsewhere can be a separate PR.

tracing/tox.ini
james-garner-canonical: I've dropped the redundant runner from this PR. Tidying up the top-level `tox.ini` can be a separate PR.

tracing/tox.ini
james-garner-canonical: I think you're right. I've dropped adding `ruff check` in this PR. A consistent `format` command can be a separate PR.

tracing/tox.ini
james-garner-canonical: Turns out it wasn't redundant, and MacOS tests fail when this is removed. Presumably this is due to the curse of multiple inheritance (here we're inheriting both `unit` and `py3.10` (etc) I believe). I've added a comment for now. This will go away when we drop `tox`.

dwilding: Interesting... thanks for figuring this out

PR #2654, opened 2026-07-20
docs: remove log_cli and log_file ini options; keep them as cli arguments for integration tests

tromai:
In https://github.com/canonical/operator/pull/2619, I added the `Configure Jubilant logs` howto section. 

In this guide, I set `log_cli` and `log_file` in the `[tool.pytest.ini_options]` table in `pyproject.toml`.

- `log_cli = true` enables live logging for all pytest invocation.
- `log_file = <path>` enables file logging for all pytest invocation.

This works well for `tox -e integration`. However, as a side effect, `tox -e unit` emitted live logs from the charm's code. This was mentioned by @dwilding.

```
tests/unit/test_charm.py::test_one PASSED
tests/unit/test_charm.py::test_two 
------------------------- live log call -------------------------
ERROR <error_level_message>
PASSED
```

We don't want this behaviour in unit tests.

This PR fixes the howto guide:
- Remove `log_cli = true`, `log_cli_level = <level>` and `log_file = <path>` from all suggested configs.
- Add instructions to enable them from the CLI:
    - `tox -e integration` provides `--log-cli-level=INFO` to the pytest command, which enable live logging.
    - `tox -e integration -- --log-file logs/verbose.log` will emit the verbose logs to `logs/verbose.log`. This requires changes to the GitHub Action workflow example.
    - For `juju debug-log` from `pytest-jubilant` the command will be `tox -e integration -- --log-file logs/verbose.log --juju-dump-logs logs`. See a test run [here](https://github.com/tromai/api-demo-server/actions/runs/29784841172/job/88494128435#step:6:1)
    
[Preview](https://canonical-juju-ops--2654.com.readthedocs.build/2654/howto/write-integration-tests-for-a-charm/#configure-jubilant-logs)

### DIFF
--- a/docs/howto/write-integration-tests-for-a-charm.md
+++ b/docs/howto/write-integration-tests-for-a-charm.md
@@ -477,7 +477,6 @@ Jubilant logs at three levels:
 ```toml
 [tool.pytest.ini_options]
 ...
-log_cli = true
 log_cli_level = "INFO"
 log_cli_format = "%(levelname)s %(name)s %(message)s"
 ```
@@ -497,7 +496,6 @@ INFO jubilant.wait [fastapi-demo/0] status changed: waiting (installing agent) -
 ```toml
 [tool.pytest.ini_options]
 ...
-log_cli = true
 log_cli_level = "DEBUG"
 log_cli_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
 log_cli_date_format = "%Y-%m-%dT%H:%M:%SZ"
@@ -544,7 +542,7 @@ Sample output:
 + .apps['fastapi-demo'].units['fastapi-demo/0'].address = '10.1.0.108'
 ```
 
-You can also configure pytest logging options on the command line.
+You can also configure these options on the command line. If you don't see the live Jubilant logs, make sure the pytest command (for example, in `tox.ini`) has the `--log-cli-level=INFO` argument.
 
 See more: [pytest | How to manage logging](https://docs.pytest.org/en/stable/how-to/logging.html)
 
@@ -569,14 +567,18 @@ This is an ideal configuration for long-running integration tests (for example, 
 # Otherwise, that section will have DEBUG logs (coming from log_file_level).
 log_level = "INFO"
 
-log_cli = true
 log_cli_level = "INFO"
 log_cli_format = "%(levelname)s %(name)s %(message)s"
 
-log_file = "logs/verbose.log"
 log_file_level = "DEBUG"
 log_file_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
 log_file_date_format = "%Y-%m-%dT%H:%M:%SZ"
+```
+
+Use the `--log-file` option from pytest to enable file logging:
+
+```text
+tox -e integration -- --log-file logs/verbose.log
 ```
 
 If you run the integration tests multiple times, `logs/verbose.log` only contains logs from the last pytest session. To use a separate file for each session, override `log_file` for each pytest invocation:
@@ -602,7 +604,7 @@ In CI, you can use `actions/upload-artifact` to make `logs/verbose.log` availabl
 
 ```yaml
   # In your integration test job
-  - run: tox -e integration
+  - run: tox -e integration -- --log-file logs/verbose.log
   - name: Upload logs
     if: ${{ !cancelled() }}
     uses: actions/upload-artifact@v7
@@ -624,7 +626,7 @@ In CI, you can use `--juju-dump-logs` with `actions/upload-artifact` to make `ju
 
 ```yaml
   # In your integration test job
-  - run: tox -e integration -- --juju-dump-logs logs
+  - run: tox -e integration -- --log-file logs/verbose.log --juju-dump-logs logs
   - name: Upload logs
     if: ${{ !cancelled() }}
     uses: actions/upload-artifact@v7
--- a/docs/howto/write-integration-tests-for-a-charm.md
+++ b/docs/howto/write-integration-tests-for-a-charm.md
@@ -474,14 +474,17 @@ Jubilant logs at three levels:
 
 ### Example of brief logging
 
+In `pyproject.toml`:
+
 ```toml
 [tool.pytest.ini_options]
 ...
-log_cli_level = "INFO"
 log_cli_format = "%(levelname)s %(name)s %(message)s"
 ```
 
-Sample output:
+Also check the `[testenv:integration]` environment in `tox.ini` to make sure the pytest command has `--log-cli-level=INFO`.
+
+Sample output of `tox -e integration`:
 
 ```text
 INFO jubilant cli: juju deploy --model jubilant-b3578475-test-charm ...
@@ -493,10 +496,11 @@ INFO jubilant.wait [fastapi-demo/0] status changed: waiting (installing agent) -
 
 ### Example of verbose logging
 
+In `pyproject.toml`:
+
 ```toml
 [tool.pytest.ini_options]
 ...
-log_cli_level = "DEBUG"
 log_cli_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
 log_cli_date_format = "%Y-%m-%dT%H:%M:%SZ"
 ```
@@ -505,7 +509,9 @@ The timestamp format follows ISO 8601.
 
 See more: {external+python:ref}`strftime() and strptime() behavior <strftime-strptime-behavior>`
 
-Sample output:
+Also modify the `[testenv:integration]` environment in `tox.ini` so that the pytest command has `--log-cli-level=DEBUG`.
+
+Sample output of `tox -e integration`:
 
 ```text
 2026-07-15T09:42:15Z INFO jubilant cli: juju deploy --model jubilant-6966ceb1-test-charm ...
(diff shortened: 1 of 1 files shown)


docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -544,7 +542,7 @@ Sample output:
 + .apps['fastapi-demo'].units['fastapi-demo/0'].address = '10.1.0.108'
 ```
 
-You can also configure pytest logging options on the command line.
+You can also configure these options on the command line. If you don't see the live Jubilant logs, make sure the pytest command (for example, in `tox.ini`) has the `--log-cli-level=INFO` argument.
```
tromai: I changed `pytest logging` to `these` to keep it consistent with the same sentence in `Log to a file` section.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -569,16 +567,20 @@ This is an ideal configuration for long-running integration tests (for example,
...
-log_file = "logs/verbose.log"
 log_file_level = "DEBUG"
 log_file_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
 log_file_date_format = "%Y-%m-%dT%H:%M:%SZ"
 ```
 
+Use the `--log-file` option from pytest to enable file logging:
+
+```text
+tox -e integration -- --log-file logs/verbose.log
```
tromai, line 588: From our daily, my impression is that we are okay with leaving `--log-file logs/verbose.log` here.

When we update the Charmcraft profiles with GitHub Action configuration, `--log-file logs/verbose.log` can go into `tox.ini` or `Makefile`. We want the behaviour to be consistent running the integration tests command both in CI and local.

dwilding: What would be the downside of adding `--log-file logs/verbose.log` to `tox.ini` now rather than later?

docs/howto/write-integration-tests-for-a-charm.md
dwilding: I think "pytest logging options" works better in this case because there's a sample output directly above, and it might be a little unclear what "these" refers to.

For the second sentence, would you mind explaining this to me again? If `log_cli_level = "DEBUG"` is already set in `pyproject.toml`, why wouldn't I see the live logs? And why are we suggesting to set it to INFO on the CLI instead of DEBUG?

tromai: > I think "pytest logging options" works better in this case because there's a sample output directly above, and it might be a little unclear what "these" refers to.

Oh, sorry. Somehow I thought this was a typo made by me in the last PR. I will revert it. I agree that `pytest logging` works better.

> log-cli-level = "DEBUG" is already set in pyproject.toml, why wouldn't I see the live logs?

If you set `log_cli_level = <level>` and not set `log_cli = true` in `pyproject.toml`, live logging is disabled. To enable live logging, without using `log_cli = true`, we need an equivalent CLI argument. However, there is no `--log-cli` argument in pytest :disappointed:  ([here](https://docs.pytest.org/en/stable/reference/reference.html#logging)). The only way we could enable live logging from CLI is to use `--log-cli-level=<level>`, which implicitly enable live logging.

> And why are we suggesting to set it to INFO on the CLI instead of DEBUG?

Ah thanks for pointing this out. In this section, we have 1 example for brief terminal logging and 1 for verbose terminal logging. Because we need `--log-cli-level=<level>` to enable live logging from CLI, `level` should be the same level defined by `log_cli_level` in `pyproject.toml`.

The sentence I have here is confusing. How about:

```
If you don't see the verbose Jubilant logs, 
make sure the pytest command (for example, in `tox.ini`)
has the `--log-cli-level=DEBUG` argument.
```

We don't need to care about brief logs here because the default `tox.ini` already set that up.

```
[testenv:integration]
...
commands =
    pytest \
        -v \
        -s \
        --tb native \
        --log-cli-level=INFO \
        {[vars]tests_path}/integration \
        {posargs}
```

As an additional improvement, we can even remove `log_cli_level` from `pyproject.toml` completely, and rely on its CLI version `--log-cli-level` to prevent confusion.

docs/howto/write-integration-tests-for-a-charm.md
tromai: Hmmm, there is no practical downside to usability as far as I know of. I think the decision was more about: implementing log to file together with the GitHub Action workflow config and maintaining consistency.

docs/howto/write-integration-tests-for-a-charm.md
dwilding: OK, I think I understand, thanks!

So this is how I understand it:

`pyproject.toml`:

```
log_cli_format = "%(levelname)s %(name)s %(message)s"
```

`tox.ini`:

- **`[testenv:unit]`** - No logging options on the `pytest` command. No live logging. In effect, `log_cli_format` in `pyproject.toml` is ignored.
- **`[testenv:integration]`** - `--log-cli-level=INFO` on the `pytest` command. Produces brief logs.

Then for verbose logs during integration tests, we recommend the developer changes two places: Switch to `--log-cli-level=DEBUG` in `tox.ini`, and adjust `pyproject.toml` to have:

```
log_cli_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
log_cli_date_format = "%Y-%m-%dT%H:%M:%SZ"
```

Did I get that right?

tromai: Yes that is correct!

dwilding: Great 🙂 I would suggest a couple of changes to the doc in that case. Instead of having an "If you don't see..." sentence here, I'd tell people how we expect them to use the pytest command. Something like this (which includes a few other adjustments):

> ### Example of brief logging
>
> In `pyproject.toml`:
>
> ```toml
> [tool.pytest.ini_options]
> ...
> log_cli_format = "%(levelname)s %(name)s %(message)s"
> ```
>
> Also check the `[testenv:integration]` environment in `tox.ini` to make sure the pytest command has `--log-cli-level=INFO`.
>
> Sample output of `tox -e integration`:
>
> [...]
>
> ### Example of verbose logging
>
> In `pyproject.toml`:
>
> ```toml
> [tool.pytest.ini_options]
> ...
> log_cli_format = "%(asctime)s %(levelname)s %(name)s %(message)s"
> log_cli_date_format = "%Y-%m-%dT%H:%M:%SZ"
> ```
>
> The timestamp format follows ISO 8601.
>
> See more: {external+python:ref}`strftime() and strptime() behavior <strftime-strptime-behavior>`
>
> Also modify the `[testenv:integration]` environment in `tox.ini` so that the pytest command has `--log-cli-level=DEBUG`.
>
> Sample output of `tox -e integration`:
>
> [...]

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -569,16 +567,20 @@ This is an ideal configuration for long-running integration tests (for example,
 # Otherwise, that section will have DEBUG logs (coming from log_file_level).
 log_level = "INFO"
 
-log_cli = true
 log_cli_level = "INFO"
```
dwilding: This should be removed too, right? After we show `pyproject.toml` I think we should confirm that the pytest command in `tox.ini` should have `--log-cli-level=INFO`.

```suggestion
```

tonyandrewmeyer (approved):
Looks good to me, thanks! I like David's suggestion for the configuration, no other comments from me.

docs/howto/write-integration-tests-for-a-charm.md
tromai: This is great. Thank you so much. 

I updated the doc to use this version.

docs/howto/write-integration-tests-for-a-charm.md
tromai: Yep. I removed this line. And your suggestion in https://github.com/canonical/operator/pull/2654#discussion_r3626661798 contains the confirmation on `--log-cli-level=INFO` so it works well.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -569,16 +573,19 @@ This is an ideal configuration for long-running integration tests (for example,
 # Otherwise, that section will have DEBUG logs (coming from log_file_level).
 log_level = "INFO"
```
dwilding, line 574: I just had one thought. Does `log_level = "INFO"` mean it doesn't matter what `--log-cli-level` is set to in `tox.ini`? If `tox.ini` has `--log-cli-level=DEBUG`, will pytest output brief or verbose logs to the console? Will `log_level = "INFO"` take precedence?

If `log_level = "INFO"` doesn't take precedence, I think we need to include the same sentence as earlier:

> Also check the [testenv:integration] environment in tox.ini to make sure the pytest command has --log-cli-level=INFO

We could put this immediately after the `pyproject.toml` sample.

dwilding (approved):

docs/howto/write-integration-tests-for-a-charm.md
tromai: Oh good point. Thank you. `--log-cli-level` will take precedence. 

If `tox.ini` has `--log-cli-level=DEBUG`, pytest will output verbose logs to the console.

I will add the sentence in.

dwilding: Looks good, thanks!

PR #2660, opened 2026-07-22
feat: add a native logs API to ops.pebble

tonyandrewmeyer:
Pebble has a `/v1/logs` endpoint, but `ops.pebble.Client` doesn't expose it. This PR changes adds support, including to Scenario.

The PR adds `LogEntry` (a frozen dataclass with the parsed RFC 3339 time, service name, and message) and two `Client` methods mirroring the Go client's Logs/FollowLogs split: `get_logs()` eagerly returns the buffered entries (n=-1 for the whole buffer), and `follow_logs()` returns a generator that streams new entries. `Container` gains `get_logs()` only, because you probably don't want to block in a hook.

In Scenario, the buffer is modelled as `Container.service_logs`, holding `state.LogEntry` entries (time defaults to now, so tests can write `LogEntry('redis', 'ready')`). The mock client sorts them by time (stably, keeping the state's order for equal times), filters by service, and applies Pebble's *n* semantics.

Verified by hand against a real Pebble 1.32 daemon (default n, filtering, n=-1, n=0, APIError on bad n, and live streaming via follow_logs).


ops/model.py
```diff
@@ -2697,6 +2697,23 @@ def get_service(self, service_name: str) -> pebble.ServiceInfo:
...
+        """Fetch the most recent log entries from services in the workload container.
+
+        To follow the logs as they are written, use
+        :meth:`ops.pebble.Client.follow_logs` on :attr:`pebble` instead.
+
+        Args:
+            service_names: Optional service names to get logs for. If no
+                service names are specified, get logs for all services.
+            n: Maximum number of log entries to return, across all requested
+                services combined. Use -1 to get everything in the buffer.
```
tonyandrewmeyer, line 2710: I think this probably should be `None`, and probably `entries` or `count` or similar rather than `n`.

ops/pebble.py
```diff
@@ -2717,6 +2759,81 @@ def get_services(self, names: Iterable[str] | None = None) -> list[ServiceInfo]:
...
             return [ServiceInfo.from_dict(info) for info in resp['result']]
 
+    def get_logs(self, services: Iterable[str] | None = None, *, n: int = 30) -> list[LogEntry]:
+        """Get log entries from the buffer Pebble keeps for each service.
+
+        Args:
+            services: Names of the services to get logs for. If not specified,
+                get logs for all services.
+            n: Maximum number of log entries to return, across all requested
+                services combined. Use -1 to get everything in the buffer.
```
tonyandrewmeyer, line 2769: Also here.

ops/pebble.py
```diff
@@ -2717,6 +2759,81 @@ def get_services(self, names: Iterable[str] | None = None) -> list[ServiceInfo]:
...
+
+            for entry in client.follow_logs(['database']):
+                if 'ready to accept connections' in entry.message:
+                    break
+
+        Args:
+            services: Names of the services to get logs for. If not specified,
+                get logs for all services.
+            n: Number of buffered log entries to yield before following. This
+                defaults to 0 (only new entries); use -1 to first get
```
tonyandrewmeyer, line 2803: And here.

tonyandrewmeyer:
We'd prefer to keep ops.pebble charm focused.

PR #2662, opened 2026-07-24
docs: extract sections in how to write integration tests to their own howto guide

tromai:
# Description

This PR made changes to the [How to write integration tests for a charm](https://canonical.com/juju/docs/ops/latest/howto/write-integration-tests-for-a-charm/#how-to-write-integration-tests-for-a-charm) page. 

The purpose is keeping this page generic, and move specific topic to their own how-to guides.

- Actions
- Relations
- Configuration
- Deploying with resources

The PR also added some real world charm examples for those topics.

# TODO 

- [x] Look for real world examples of these topics and add link (for example, at the end of this [section](https://canonical.com/juju/docs/ops/latest/howto/manage-the-workload-version/#implement-the-feature)).
- [x] Try the examples in those pages in a sample charm.
- [x] Keep the style consistent with https://github.com/canonical/operator/pull/2666


docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -230,96 +230,7 @@ def test_integrate(charm: pathlib.Path, juju: jubilant.Juju):
 
 ### Deploy your charm with resources
```
tromai: I am not sure if I should remove this heading.

The instructions for this section has been moved to the `Manage resources` page. Keeping this section here with one line

```
See: {ref}`Manage resources | Write integration tests <manage-resources-integration-tests>`
```

is confusing, I think.

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
...
 
 ```python
-def test_logger(juju: jubilant.Juju):
+def test_snapshot_action(charm: pathlib.Path, juju: jubilant.Juju):
     action = juju.run(
         'your-app/0', 'snapshot', {'filename': 'db-snapshot.tar.gz'}
     )
     assert action.status == 'completed'
-    assert action.results['snapshot-size'].isdigit()
+    assert action.results['result'].startswith('Stored snapshot in')
```
tromai: This test is modeled after the instructions in this page 

```python
def _on_snapshot_action(self, event: ops.ActionEvent):
    ...
    # Set the results of the action.
    msg = f'Stored snapshot in {params.filename}.'
    # These will be displayed in the `juju run` output.
    event.set_results({'result': msg})
```

which I think is okay because this is at the end of the guide.

I'm also a bit confused on why the original doc asserted 
```
assert action.results['snapshot-size'].isdigit()
```

`_on_snapshot_action` doesn't explicitly set `snapshot-size` anywhere. Am I missing something 🤔 ?

I also see `snapshot-size` mentioned in the unit tests guide
```
def test_backup_action():
    ctx.run(
        ctx.on.action('snapshot', params={'filename': 'db-snapshot.tar.gz'}),
        testing.State(),
    )
    assert ctx.action_logs == [
        'Starting snapshot',
        'Table1 complete',
        'Table2 complete',
    ]
    assert 'snapshot-size' in ctx.action_results
```

docs/howto/manage-configuration.md
```diff
@@ -110,6 +110,28 @@ def test_short_wiki_name():
     assert isinstance(state_out.unit_status, testing.BlockedStatus)
 ```
 
+### Write integration tests
+
+> See first: {ref}`write-integration-tests-for-a-charm`
+
+You can set a configuration option in your application and check its results.
+
+```python
```
tromai, line 125: I modified this test compared to the current [version](https://canonical.com/juju/docs/ops/latest/howto/write-integration-tests-for-a-charm/#test-a-configuration):
- Used a name with spaces for the invalid case 
- Added a valid name test case

docs/howto/manage-relations.md
```diff
@@ -364,20 +364,26 @@ relation.remote_unit_name  # 'zookeeper/42'
...
 ```python
 # This assumes that your integration tests already include the standard
 # build and deploy test.
 
 
-def test_active_with_another_app(juju: jubilant.Juju):
+def test_active_with_another_app(charm: pathlib.Path, juju: jubilant.Juju):
     juju.deploy('another-app')
     juju.integrate('your-app:endpoint', 'another-app:endpoint')
-
```
tromai, line 381: This is a style change to keep it consistent with other locations, where there is no new line in between the `juju.*` invocations.

docs/howto/manage-relations.md
```diff
@@ -364,20 +364,26 @@ relation.remote_unit_name  # 'zookeeper/42'
...
     juju.wait(jubilant.all_active)
 ```
 
+See more: [](jubilant.Juju.integrate)
+
+<!-- valkey-operator example is probably too complicated -->
+Examples for testing relations with Jubilant: [`valkey-operator`](https://github.com/canonical/valkey-operator/blob/9/edge/tests/integration/clients/test_client_relation.py), [`kafka-k8s-operator`](https://github.com/canonical/kafka-k8s-operator/blob/main/tests/integration/test_provider_v1.py).
+
+This test (and subsequent tests) don't need to depend on the `charm` fixture. However, it's helpful for each test to depend on `charm`, so that each test fails immediately if a `.charm` file isn't available.
```
tromai: This paragraph was originally from https://canonical.com/juju/docs/ops/latest/howto/write-integration-tests-for-a-charm/#test-a-relation. As we moved those tests into their own Howto guides, I wonder if we should either:
1. Duplicate this paragraph in those Howto guides (`Manage relation`, `Manage resources`, and `Manage configuration`) OR
1. Remove this paragraph and the `charm` fixture completely from the relevant tests.

Since we encourage using the `charm` fixture to fail early, I am okay with the first solution (and remove the `and subsequent tests` phrase).

docs/howto/manage-relations.md
```diff
@@ -14,11 +14,11 @@ To integrate with another charm, or with itself (to communicate with other units
 
 **If you're using an existing interface:**
 
-Make sure to consult [the `charm-relations-interfaces` repository](https://github.com/canonical/charm-relation-interfaces) for guidance about how to implement them correctly.
+Make sure to consult {external+charmlibs:doc}`how-to/design-relation-interfaces` for guidance about how to implement them correctly.
```
tromai: Drive by fix. https://github.com/canonical/charm-relation-interfaces is now archived.

docs/howto/write-integration-tests-for-a-charm.md
dwilding: I recommend removing the "Deploy your charm with resources" section from this page. I think it's worth mentioning resources in the "Deploy your charm" section. After the paragraph "Jubilant provides several other helpers [...]" we could say that for a Kubernetes charm you'll deploy with a resource, and show an example. Then provide the link to `manage-resources-integration-tests` for more general info about managing resources in integration tests.

docs/howto/write-integration-tests-for-a-charm.md
dwilding: I think it would be helpful to add some links to the feature-specific advice about integration tests. At minimum, for relations and configuration.

docs/howto/manage-resources.md
dwilding: I don't think this comment is needed. We now have a how-to guide "How to publish your charm on Charmhub" and Charmcraft has guidance about publishing resources.

dwilding: I'd remove this sentence. It duplicates the sentence after the example. I'd replace it by something like:

> A charm can require `file` or `oci-image` resources, defined in `charmcraft.yaml`. When you publish your charm, you'll upload the resources to Charmhub. Then when a user deploys your charm from Charmhub, the resources will be available to your charm code.

docs/howto/manage-resources.md
```diff
@@ -89,3 +91,37 @@ with ctx(ctx.on.start(), testing.State(resources={resource})) as mgr:
...
 ```
+
+(manage-resources-integration-tests)=
+### Write integration tests
+
+> See first: {ref}`write-integration-tests-for-a-charm`
+
+A charm can require `file` or `oci-image` resources to work, which have revision numbers on Charmhub. OCI images can be referenced directly, while file resources are typically built during packing.
+
+In `charmcraft.yaml`'s `resources` section, the `upstream-source` field is, by convention, a usable default image or file that can be used in testing:
```
dwilding:
```suggestion
During development and testing, it's useful to specify resource locations when deploying the charm.

The conventional place to specify resource locations for testing is the `upstream-source` field in `charmcraft.yaml`'s `resources` section:
```

dwilding (commented):
Adding some initial comments. I'm still reviewing.

docs/howto/manage-actions.md
dwilding:
```suggestion
### Record the ID of an action task
```

dwilding, line 144:
```suggestion
## Test the feature

### Write unit tests
```

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
 To verify that an action works correctly against a real Juju instance, write an integration test with `jubilant`. For example:
```
dwilding: I think we can drop this line. We already assume people have read the how-to guide, so the line doesn't add any further information.

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
...
 
 ```python
-def test_logger(juju: jubilant.Juju):
+def test_snapshot_action(charm: pathlib.Path, juju: jubilant.Juju):
     action = juju.run(
         'your-app/0', 'snapshot', {'filename': 'db-snapshot.tar.gz'}
     )
     assert action.status == 'completed'
```
dwilding: I don't think we need this line. The [`run`](https://canonical.com/juju/docs/jubilant/reference/jubilant/#jubilant.Juju.run) method already raises `TaskError` if the action failed.

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
 To verify that an action works correctly against a real Juju instance, write an integration test with `jubilant`. For example:
 
 ```python
-def test_logger(juju: jubilant.Juju):
+def test_snapshot_action(charm: pathlib.Path, juju: jubilant.Juju):
     action = juju.run(
```
dwilding: Not critical, but while we're here we could change `action` to `task`, which would better match Jubilant's naming.

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
...
         'your-app/0', 'snapshot', {'filename': 'db-snapshot.tar.gz'}
     )
     assert action.status == 'completed'
-    assert action.results['snapshot-size'].isdigit()
+    assert action.results['result'].startswith('Stored snapshot in')
 ```
+
+See also: {external+juju:ref}`Action <action>`, [](jubilant.Juju.run)
```
dwilding: I think this is fine without the Juju doc link. We have a link to "Juju | Manage actions" at the top of the page.

```suggestion
See also: [](jubilant.Juju.run)
```

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
...
+
+Examples: [`discourse-k8s` runs the create-user action and asserts on the result](https://github.com/canonical/discourse-k8s-operator/blob/main/tests/integration/test_users.py)
```
dwilding: I think it's fine without the extra text, as on the "Manage resources" page.

```suggestion
Examples: [`discourse-k8s`](https://github.com/canonical/discourse-k8s-operator/blob/main/tests/integration/test_users.py)
```

docs/howto/manage-actions.md
dwilding: It looks like the unit testing section has become out of sync with the rest of the page - I don't think you're missing anything. I recommend a separate PR to clean up any inconsistencies in the examples throughout the page.

docs/howto/manage-actions.md
```diff
@@ -184,10 +184,14 @@ def test_backup_action_failed():
 To verify that an action works correctly against a real Juju instance, write an integration test with `jubilant`. For example:
 
 ```python
```
dwilding, line 200: I think we should include the imports. We do that on the "Manage resources" page and we do that for the unit testing example above.

Same comment for the configuration and relations how-to guide.

docs/howto/manage-configuration.md
```diff
@@ -110,6 +110,28 @@ def test_short_wiki_name():
     assert isinstance(state_out.unit_status, testing.BlockedStatus)
 ```
 
+### Write integration tests
```
dwilding, line 121:
```suggestion
## Test the feature

### Write integration tests
```

docs/howto/manage-resources.md
```diff
@@ -89,3 +91,37 @@ with ctx(ctx.on.start(), testing.State(resources={resource})) as mgr:
...
+        name: res['upstream-source']
+        for name, res in METADATA['resources'].items()
+    }
+    juju.deploy(charm, resources=resources)
+    juju.wait(jubilant.all_active)
+```
+
+See also: [](jubilant.Juju.deploy)
+
+Examples of deploying with resources using Jubilant: [`valkey-operator`](https://github.com/canonical/valkey-operator/blob/9/edge/tests/integration/test_charm.py), [`kafka-k8s-operator`](https://github.com/canonical/kafka-k8s-operator/blob/main/tests/integration/test_balancer.py)
```
dwilding: I think the context should be clear enough

```suggestion
Examples: [`valkey-operator`](https://github.com/canonical/valkey-operator/blob/9/edge/tests/integration/test_charm.py), [`kafka-k8s-operator`](https://github.com/canonical/kafka-k8s-operator/blob/main/tests/integration/test_balancer.py)
```

docs/howto/manage-configuration.md
```diff
@@ -110,6 +110,28 @@ def test_short_wiki_name():
...
+
+See first: {ref}`write-integration-tests-for-a-charm`
+
+You can set a configuration option in your application and check its results.
```
dwilding: I'd probably drop this sentence. But am OK to keep it if you think it's helpful.

docs/howto/manage-configuration.md
```diff
@@ -110,6 +110,28 @@ def test_short_wiki_name():
...
+
+```python
+def test_config_invalid_name(charm: pathlib.Path, juju: jubilant.Juju):
+    juju.config('your-app', {'name': 'invalid name has spaces'})
+    # A name with spaces should put the charm into blocked status.
```
dwilding: I think it would be more helpful for the comment to explain why the timeout is overridden. My interpretation: Setting a valid config is likely to require a replan/restart, which might take some time (so we use the default timeout). But setting an invalid config should be caught by the charm and rejected immediately (so there's no need for a long timeout). If that's the case I might even reduce the timeout, to something like 10 seconds.

docs/howto/manage-configuration.md
```diff
@@ -110,6 +110,28 @@ def test_short_wiki_name():
...
+    juju.wait(jubilant.all_blocked, timeout=60)
```
dwilding: Arguably we should reset the config at the end of the test, to get back into a known-good configuration for subsequent tests. But since the next test is also a config test, it doesn't make a difference in this example. Maybe it's worth mentioning after the example?

docs/howto/manage-configuration.md
```diff
@@ -110,6 +110,28 @@ def test_short_wiki_name():
...
+
+
+def test_config_valid_name(juju: jubilant.Juju):
+    juju.config('your-app', {'name': 'charming-wiki'})
+    juju.wait(jubilant.all_active)
+```
+
+See also: [](jubilant.Juju.config)
+
+Examples of changing configuration with Jubilant: [`kafka-k8s-operator` changes the log-evel configuration and checks the logs from all units](https://github.com/canonical/kafka-k8s-operator/blob/main/tests/integration/test_charm.py).
```
dwilding: Since there's only one test function that changes config, and many other test functions, how about this:

```suggestion
Example: [`kafka-k8s-operator` (`test_log_level_change`)](https://github.com/canonical/kafka-k8s-operator/blob/main/tests/integration/test_charm.py).
```

docs/howto/manage-configuration.md
dwilding: It looks like this is the only place we have such a section. I don't think it adds much value, so I'd probably drop it.

dwilding (commented):
Stopping for today - I still need to review the relations how-to guide.

docs/howto/manage-resources.md
tromai: I agree this sentence is duplicated. I will remove it.

I like your suggestion, but I might modify it a bit. From what I understand, publishing resources to Charmhub is an explicit step after you publish a charm, and we don't want users to misunderstand that they are published together automatically.

> A charm can require `file` or `oci-image` resources, defined in `charmcraft.yaml`. After you publish your charm, you must explicitly upload the resources to Charmhub. Then when a user deploys your charm from Charmhub, the resources will be available to your charm code.

If `must explicitly` is too strong, perhaps I can use `need to` instead :thinking: .

docs/howto/manage-configuration.md
tromai: I will drop this sentence. I think the examples are clear enough without this description.

docs/howto/write-integration-tests-for-a-charm.md
tromai: > we could say that for a Kubernetes charm you'll deploy with a resource, and show an example.

I have been thinking of an example for this, and came to a conclusion. `manage-resources-integration-tests` defines a conventional way to deploy with a resource integration test:

```python
import pathlib

import jubilant
import pytest
import yaml


METADATA = yaml.safe_load(pathlib.Path('./charmcraft.yaml').read_text())


@pytest.mark.juju_setup
def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
    resources = {
        name: res['upstream-source']
        for name, res in METADATA['resources'].items()
    }
    juju.deploy(charm, resources=resources)
```

I think it's best to use the same example. `manage-resources-integration-tests` contains this example, and 2 short paragraph describing it.

I'm leaning toward duplicating this example here, and provide a link to `manage-resources-integration-tests`. Please let me know if this is a good idea.

Another way is to add this paragraph after `Jubilant provides several other helpers [...]`

```
Jubilant provides several other helpers [...]

For a Kubernetes charm, you will deploy it with a resource. See: {ref}`Manage resources | Write integration tests <manage-resources-integration-tests>` for more examples.
```

docs/howto/manage-actions.md
tromai: Thanks for confirming. I created https://github.com/canonical/operator/issues/2688 to track this.

docs/howto/manage-actions.md
tromai: Thanks for pointing this out. Yep, adding the imports is great. I added them in the integration test section of those pages.

docs/howto/manage-configuration.md
tromai: I assume you suggested putting `## Test the feature` before both `### Write unit tests` and `### Write integration tests` :smile:

docs/howto/manage-configuration.md
tromai: Yep, your interpretation is spot on. I want to test the fail-fast behavior. I picked 60 seconds as a safe middle-ground (especially if someone is running this on CI). I gave 10 seconds a try, and it worked well on my simple example (I tested it locally).

I hope the comment:

```
    # Setting an invalid name should be caught by the charm and rejected
    # immediately. The timeout is overridden to test this fail-fast behavior.
```

would be clear enough for readers to adjust this timeout for their own setup.

docs/howto/manage-configuration.md
tromai: This is a very good point that I missed. Thank you. I added a paragraph for your suggestion.

---

I think there are different ways to achieve this reset:

1. Directly at the end of this test function (similar to [`test_log_level_change`](https://github.com/canonical/kafka-k8s-operator/blob/cc8a5ba0ab9339b4a338eb9e2c062def03a6bb9d/tests/integration/test_charm.py#L191)
1. Reset in another test function (in the example)
1. Use a fixture for `test_config_invalid_name`. 

I wonder what we would typically recommend to Charmers? I guess either the first or second, depending on their test setup. They are more straight-forward to me than using a fixture.

On that note, the [`test_log_level_change`](https://github.com/canonical/kafka-k8s-operator/blob/cc8a5ba0ab9339b4a338eb9e2c062def03a6bb9d/tests/integration/test_charm.py#L191) example is marked with `pytest.mark.skip`, so I might need another example :sweat_smile: .

docs/howto/write-integration-tests-for-a-charm.md
dwilding: Hmm, I think this might be one time that duplication is the best approach. I can imagine someone searching or scanning the page looking for `upstream-source` or the general shape of this example. A link to `manage-resources-integration-tests` wouldn't make this possible.

You could consider putting the duplicated content in a separate file and reusing it. In general, I try to avoid that solution, as it makes the doc build more complicated. But I'd be OK with it here.

We already have a `reuse` directory: https://github.com/canonical/operator/tree/main/docs/reuse. You could put a `.md` file in there. See [how it's done for Pebble](https://github.com/canonical/pebble/blob/master/docs/how-to/install-pebble.md?plain=1#L46-L49). Although the `start-after` and `end-before` parts seem over complicated - I assume you wouldn't need them.

If you take this approach, it's best to exclude the `reuse` directory from the generated sitemap ([as for Pebble](https://github.com/canonical/pebble/blob/master/docs/conf.py#L136)).

docs/howto/manage-resources.md
dwilding: How about something like:

> As one of the charm publishing steps, you'll upload the resources to Charmhub.

*Edit:* Or maybe it reads better the other way round:

> You'll upload the resources to Charmhub as one of the charm publishing steps.

docs/howto/manage-configuration.md
dwilding: Looks good. I wonder whether "reduced" might be clearer then "overridden", but that would assume the reader is familiar with Jubilant's default timeout. Up to you :slightly_smiling_face:

docs/howto/manage-configuration.md
dwilding: > On that note, the [`test_log_level_change`](https://github.com/canonical/kafka-k8s-operator/blob/cc8a5ba0ab9339b4a338eb9e2c062def03a6bb9d/tests/integration/test_charm.py#L191) example is marked with `pytest.mark.skip`, so I might need another example 😅 .

Ha, I hadn't spotted that!

For the different ways to achieve the reset, I'm not so sure either. You'll need a second reviewer on this PR so I'm curious to see what opinion they have.

docs/howto/manage-resources.md
tromai: > You'll upload the resources to Charmhub as one of the charm publishing steps.

Your suggestion is much better. I will go with it. Thank you.

docs/howto/write-integration-tests-for-a-charm.md
tromai: Thanks for the suggestion and guide. I thought about it and went with the `reuse` solution (in https://github.com/canonical/operator/pull/2662/commits/fab9460a13c06273539b9fe02d96b2b2e15a56f1)

tromai:
@dwilding Thank you so much for your review. I have addressed the PR feedback. I also made another attempt to review `manage-relations.md` (on other sections that I didn't touch in this PR). There are some inconsistencies in the examples, which I am planning to address in another PR. Let me know if that is okay.

Some examples:

```python
def _on_db_relation_created(self, event: ops.RelationCreatedEvent):
    if not self.unit.is_leader():
        return
    credentials = self.create_database(event.app.name)
    data = DatabaseProviderAppData(credentials=credentials)
    relation.save(data, event.app)
```

`relation` is undefined. I assume we need `event.relation.save(data, event.app)`. 
 
```python
def _update_configuration(self, _: ops.Eventbase):
    ...
    secret_contents = self.model.get_secret(id=secret_id).get_contents(
        refresh=True
    )
    self.push_configuration(
        username=secret['username'],
        password=secret['password'],
    )
```

I assume `secret` should be `secret_contents`.

docs/howto/manage-configuration.md
tromai: I found another example to replace [test_log_level_change](https://github.com/canonical/kafka-k8s-operator/blob/cc8a5ba0ab9339b4a338eb9e2c062def03a6bb9d/tests/integration/test_charm.py#L191). It is [`cassandra-operator`](https://github.com/canonical/cassandra-operator/blob/main/tests/integration/test_config.py).

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -192,6 +192,11 @@ This test deploys your charm and waits for all applications and units to be acti
 
 Jubilant provides several other helpers, in case you need to change the `wait` condition. See {external+jubilant:ref}`Use a custom wait condition <use_a_custom_wait_condition>`.
 
+For a Kubernetes charm, you will deploy it with a resource. See the modified example (from {ref}`Manage resources | Write integration tests <manage-resources-integration-tests>`):
```
dwilding: A few small suggestions to make this more concise. I think we can keep the link pointing to the integration testing section without mentioning that in the link text.

```suggestion
For a Kubernetes charm, deploy with a resource. See this modified example (from {ref}`Manage resources <manage-resources-integration-tests>`):
```

docs/howto/manage-relations.md
dwilding: The linked doc ("How to design relation interfaces") is for people developing an interface. But the first part of the note look like it's for people _using_ an interface. I think we should link to [https://canonical.com/juju/docs/ops/latest/howto/manage-libraries/](https://canonical.com/juju/docs/ops/latest/howto/manage-libraries/) and [https://canonical.com/juju/docs/charmlibs/how-to/manage-libraries/](https://canonical.com/juju/docs/charmlibs/how-to/manage-libraries/) instead.

The link to "How to design relation interfaces" could go in the second part of the note, before "How to distribute charm libraries".

While we're here, I recommend deescalating this from a Caution note to a normal note.

docs/howto/manage-relations.md
```diff
@@ -360,24 +360,35 @@ relation = testing.SubordinateRelation(
 relation.remote_unit_name  # 'zookeeper/42'
 ```
 
+(manage-relations-integration-tests)=
 ### Write integration tests
 
 > See first: {ref}`write-integration-tests-for-a-charm`
 
-To verify that charm behaves correctly when integrated with another in a real Juju environment, write an integration test with `jubilant` that deploys another application and relates your charm to it.
+To verify that a charm behaves correctly when integrated with another in a real Juju environment, write an integration test with `jubilant` that deploys another application and relates your charm to it.
```
dwilding: I recommend dropping this or shortening it. Something like:

```suggestion
Use Jubilant to verify that your charm behaves correctly when integrated with another charm:
```

docs/howto/manage-relations.md
dwilding: I don't like having to explain this everywhere, but I think that including `charm` in every test is the most practical/uncomplicated way to ensure fast failure. So we ought to explain it, since it's not intuitive.

I think it's good to keep the explanation as concise as we can. Maybe something like:

> We recommend including the `charm` fixture (even though it's not used) so that the test fails immediately if a `.charm` file isn't available.

I agree that we might as well mention this in the other how-to guides too.

docs/howto/manage-resources.md
```diff
@@ -89,3 +87,19 @@ with ctx(ctx.on.start(), testing.State(resources={resource})) as mgr:
...
 ```
+
+(manage-resources-integration-tests)=
+### Write integration tests
+
+> See first: {ref}`write-integration-tests-for-a-charm`
+
+During development and testing, it's useful to specify resource locations when deploying the charm.
+
+The conventional place to specify resource locations for testing is the `upstream-source` field in `charmcraft.yaml`'s `resources` section:
```
dwilding, line 102: Reading this again, it sounds like we're about to show `charmcraft.yaml`. If you think there's potential for confusion, we could add an extra sentence:

```suggestion
The conventional place to specify resource locations for testing is the `upstream-source` field in `charmcraft.yaml`'s `resources` section. To use `upstream-source` in an integration test:
```

I'm not convinced this addition is needed - up to you.

dwilding:
> which I am planning to address in another PR. Let me know if that is okay.

I agree on a separate PR. Otherwise it will be hard to draw a line, and I think the more you check, the more inconsistencies you'll find!

dwilding (approved):
Looks great!

docs/howto/manage-resources.md
tromai: Interesting. I didn't recall having this sentence in the latest commit. 

I checked and the latest state of this branch didn't have it, so it's okay. I don't think we need it too.

docs/howto/manage-relations.md
tromai: Thanks for this. 

I think I confused myself over `how to implement them correctly`. I read it as "implementing the library itself", while it should be understand as "implementing the charm code to use the library" :thinking: 

I will follow your suggestion.

docs/howto/manage-actions.md
```diff
@@ -177,17 +179,26 @@ def test_backup_action_failed():
...
+
+import jubilant
+
+
+def test_snapshot_action(charm: pathlib.Path, juju: jubilant.Juju):
+    task = juju.run(
         'your-app/0', 'snapshot', {'filename': 'db-snapshot.tar.gz'}
     )
-    assert action.status == 'completed'
-    assert action.results['snapshot-size'].isdigit()
```
tonyandrewmeyer, line 206: Do we want to remove this? If this was a real charm, I would be more interested in whether the snapshot had a size than some arbitrary result text.

docs/howto/write-integration-tests-for-a-charm.md
```diff
@@ -192,6 +192,11 @@ This test deploys your charm and waits for all applications and units to be acti
 
 Jubilant provides several other helpers, in case you need to change the `wait` condition. See {external+jubilant:ref}`Use a custom wait condition <use_a_custom_wait_condition>`.
 
+For a Kubernetes charm, deploy with a resource. See this modified example (from {ref}`Manage resources <manage-resources-integration-tests>`):
```
tonyandrewmeyer: Should this say "modified" when it's literally the same file?

docs/howto/manage-configuration.md
```diff
@@ -111,10 +113,34 @@ def test_short_wiki_name():
...
 
-```text
-juju config <name of application deployed by your charm> name=charming-wiki
+
+def test_config_valid_name(juju: jubilant.Juju):
+    juju.config('your-app', {'name': 'charming-wiki'})
+    juju.wait(jubilant.all_active)
 ```
+
+`test_config_invalid_name` leaves the config in an invalid state. Reset the config to its previous known-good state so that subsequent tests start from a valid configuration. In this example, `test_config_valid_name` performs that reset.
```
tonyandrewmeyer: I think if I was writing these tests for a real charm, I wouldn't like to have that sort of ordering requirement in the tests.

I'd suggest having a finally in the test_config_invalid_name test restore the config (probably reset rather than setting to known good).

And then this sentence could be removed.

docs/howto/manage-relations.md
```diff
@@ -10,15 +10,15 @@ To add relation capabilities to a charm, you’ll have to define the relation in
...
 
 **If you're using an existing interface:**
 
-Make sure to consult [the `charm-relations-interfaces` repository](https://github.com/canonical/charm-relation-interfaces) for guidance about how to implement them correctly.
+Make sure to consult {ref}`manage-interfaces` and {external+charmlibs:doc}`how-to/manage-libraries` for guidance about how to implement them correctly.
 
 **If you're defining a new interface:**
 
-Make sure to add your interface to [the `charm-relations-interfaces` repository](https://github.com/canonical/charm-relation-interfaces).
+Make sure to add your interface to [the `charmlibs` repository](https://github.com/canonical/charmlibs). See {external+charmlibs:doc}`how-to/design-relation-interfaces`, {external+charmlibs:doc}`how-to/python-package`.
```
tonyandrewmeyer: Maybe we should add some sort of "If this is an interface other charms should use" caveat here? I'm not sure we are 100% sorted on exactly which interfaces should/shouldn't go into charmlibs.

docs/howto/manage-relations.md
```diff
@@ -127,7 +127,7 @@ In `src/charm.py`, create a class that defines the schema for the relation data.
 For example:
 
 ```python
-class SMTPProviderUnitData(pydantic.BaseMode):
+class SMTPProviderUnitData(pydantic.BaseModel):
```
tonyandrewmeyer, line 134: Nice catch :)

docs/howto/manage-relations.md
```diff
@@ -343,7 +343,7 @@ that, `SubordinateRelation`, compared to `Relation`, always talks in terms of
 `remote`:
 
 - `Relation.remote_units_data` becomes `SubordinateRelation.remote_unit_data`
-  taking a single `dict[str:str]`. The remote unit ID can be provided as a
+  taking a single `dict[str, str]`. The remote unit ID can be provided as a
```
tonyandrewmeyer, line 350: Here too :)

tonyandrewmeyer (approved):
Looks good, thanks!

A few minor comments, nothing blocking.

docs/howto/manage-actions.md
tromai: I agree that asserting on the snapshot-size is much more useful that the `result` text. I removed this line to keep it consistent with the `_on_snapshot_action` implementation at the beginning of the page.

I can change `_on_snapshot_action`'s implementation to put the size into `event.results` instead:

```diff
def _on_snapshot_action(self, event: ops.ActionEvent):
    """Handle the snapshot action."""
    ...
-    success = self.do_snapshot(    
+    size = self.do_snapshot(
        filename=params.filename,
        kind=params.compression.kind,
        quality=params.compression.quality,
    )
-    if not success:
+    if size is None:
        ...
        return
    # Set the results of the action. These will be displayed in the
    # `juju run` output.
-    msg = f'Stored snapshot in {params.filename}.'
-    event.set_results({'result': msg})
+    event.set_results({'snapshot-size': str(size)})
```

Then I this assertion line can be reverted

docs/howto/write-integration-tests-for-a-charm.md
tromai: No it shouldn't. I have removed `modified` from this sentence. Thanks for the catch!

docs/howto/manage-configuration.md
tromai: I think resetting to its original value is a good idea. I modified this example to do that. I also added the `charm` fixture into the valid config name test.

tonyandrewmeyer (approved):

docs/howto/manage-actions.md
tromai: I modified this example after Tony's comment later in this PR. So this issue was fixed :smile:

PR #2663, opened 2026-07-24
ci: enrich scheduled-failure issues with an LLM triage pass

tonyandrewmeyer:
Today, when a scheduled workflow fails, `notify-scheduled-failure.yaml` opens an issue whose title and body are a generic one-liner plus a link to the job. A human then has to read the logs, check whether it duplicates an existing issue, and rewrite the title and body before it is useful to anyone.

This adds a second stage that does that first pass automatically: extract a deterministic failure signature from the failing jobs' logs, dedupe against real candidate issues, ask an LLM to draft a descriptive issue or comment, check its output, and apply it, falling back to today's plain notice if anything goes wrong.

Stage 1 (`notify-scheduled-failure.yaml`) keeps its current guarantee: no secrets, no LLM, no network beyond `gh`, always produces a notification. The only change is a rough de-duplication, searching open issues for the workflow name and commenting on a match instead of opening a duplicate.

Stage 2 (`ai-failure-enrich.yaml`, new) does the enrichment. It is an `enrich` job in the notifier that `uses:` a separate file, so that stage 1's guarantee stays legible: everything that needs an API key, an environment or a checkout is on the other side of that line.

The enrichment itself is not in this repository. Stage 2's only real step runs the `ai-failure-notifier` package out of `canonical/charm-tech-code`, pinned to a commit, so the signature extractor, the dedup ladder, the schema and the applier all live there, and what is left here is workflow YAML.

Because stage 2 runs inside the caller's run, the seven callers have to hand down what it needs: `actions: read` for the log fetch, and `OPENROUTER_API_KEY` passed by name.

`notify` passes the issue number and what it did (`new` or `comment`) down to `enrich` as inputs, rather than letting stage 2 search for the marker stage 1 has just stamped. GitHub's issue search index doesn't necessarily return something written seconds ago, and a miss there opens a second issue for a run that already has one. Stage 2 can still look the issue up itself, which is what a repo that adopts the enricher before it passes those inputs will get, and when the notification is what failed there is nothing to enrich and the job doesn't run at all.


tonyandrewmeyer:
@benhoyt adding you as a reviewer specifically around the Zizmor/workflow question (feel free to review anything else as well, of course).

benhoyt (commented):
I've only reviewed the Zizmor change - the rationale for that looks reasonable to me.

benhoyt:
Interesting idea, though! Let's see how it works.

test/test_ai_failure_notifier.py
```diff
@@ -0,0 +1,1136 @@
+#!/usr/bin/env python3
```
james-garner-canonical: We don't normally make our tests executable like this, do we? I see we also have an `if __name__ == '__main__'` at the bottom.

test/test_ai_failure_notifier.py
```diff
@@ -0,0 +1,1136 @@
...
+# /// script
+# requires-python = ">=3.11"
+# ///
```
james-garner-canonical: Won't we run this test with Python 3.10 in CI? Why does it need Python 3.11? Can we drop this?

test/test_ai_failure_notifier.py
```diff
@@ -0,0 +1,1136 @@
...
+import os
+import pathlib
+import sys
+import tempfile
+import unittest
+import urllib.error
+from typing import Any
+from unittest import mock
+
+sys.path.insert(0, str(pathlib.Path(__file__).parent.parent / '.github'))
```
james-garner-canonical: I'm not super thrilled by this approach: 1) including a test of unrelated infra in the unit tests for `ops` itself, and 2) requiring path manipulation to make it work (though that bit might be unavoidable). I'm not opposed to landing it this way, but I think we should at least have a rainy day item for running workflow/script specific tests in CI (and ideally locally too -- maybe easier once we land the `Makefile` move).

scripts/test/test_ai_failure_notifier.py
```diff
@@ -0,0 +1,1136 @@
...
+            "errors. canonical/traefik-k8s-operator: ImportError: cannot import name 'Event' from "
+            "'scenario'."
+        ),
+        closed_at=None,
+    ),
+]
+
+FIXTURE_ENVELOPE: dict[str, Any] = {
+    'action': 'comment',
+    'target_issue': 9010,
```
james-garner-canonical: The typing here makes me wonder if `target_issue` should be a string, but I suppose the integer typing pays for itself in serialization time errors?

test/test_ai_failure_notifier.py
```diff
@@ -0,0 +1,1136 @@
...
+FIXTURE_ENVELOPE: dict[str, Any] = {
+    'action': 'comment',
+    'target_issue': 9010,
+    'body': 'Another occurrence: '
+    'https://github.com/canonical/operator/actions/runs/28141163589\n'
+    '\n'
+    '4 of the 5 failing charms match #9010 unchanged. New this run: '
+    'canonical/charm-ubuntu is now also failing its unit tests.',
+    'dedup_reason': "4 of 5 failing charms match #9010's signature; charm-ubuntu is a "
+    'new failure not present in #9010',
```
james-garner-canonical: It's a bit nitpicky, but this implicit string concatenation style feels a bit odd. Do we use this elsewhere?

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,151 @@
...
+on:  # zizmor: ignore[dangerous-triggers]
+  workflow_run:
+    workflows:
+      - "Example Charm charmcraft test"
+      - "Example Charm Integration Tests"
+      - "ops Integration Tests"
+      - "ops Smoke Tests"
+      - "TIOBE Quality Checks"
+      - "Update Best Practices Doc"
+      - "Update Charm Pins"
```
james-garner-canonical: Given the amount of explanatory text that this design ends up producing, I wonder if it would be better to make this workflow run on `workflow_call`, in one of the following ways:
1. Have `notify-scheduled-failure` call this workflow at the end, and wire any required data through from the original call site.
2. Just add a block to the workflows that use `notify-scheduled-failure` to call this workflow, like:
    ```yaml
      enrich-failure-issue:
        needs: [open-issue-on-failure-if-scheduled]
        permissions:
          issues: write
        uses: ./.github/workflows/ai-failure-enrich.yaml
    ```

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,151 @@
...
+  # `always()` plus checking `needs.enrich.outputs.handled` (rather than
+  # `needs.enrich`'s job status) means this fires both when `enrich` goes red
+  # AND when it somehow succeeds without having handled anything.
+  #
+  # By the time this can run, the notifier has already produced a
+  # notification: `workflow_run: completed` only fires once the caller's run,
+  # including its `open-issue` job, has finished. So creating an issue
+  # unconditionally means two issues for one failure. The only case that
+  # genuinely needs a new issue is the notifier itself having failed, and
+  # that is exactly what the absence of its marker says.
```
james-garner-canonical: Do we really need this? We're defending here against basic infra failing (`uv run`, random errors in our own script) -- I think we can just rely on the fallback in `ai_failure_notifier.py` itself, and live with only having the failing workflow if the job inexplicably fails without reaching the internal fallback.

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -13,14 +22,34 @@ jobs:
...
+
+          if [[ -n "$match" ]]; then
+            comment_body="Scheduled workflow '$WORKFLOW_NAME' failed again: $RUN_URL"$'\n\n'"${marker}comment -->"
+            gh issue comment "$match" --repo "$REPO" --body "$comment_body"
+          else
+            issue_body="Scheduled workflow '$WORKFLOW_NAME' failed: $RUN_URL"$'\n\n'"${marker}new -->"
+            gh issue create --repo "$REPO" \
+              --title "Scheduled workflow '$WORKFLOW_NAME' failed" \
+              --body "$issue_body"
+          fi
```
james-garner-canonical, line 66: It seems like it would be simpler to fold the dumb match and comment logic into the fallback branch of the enrich script and just call the enrich workflow here ... or even just fold the workflow in here and call the script directly. If the concern is making the AI parts easy to tear out or disable later, I think we could do that with clearly commented sections of input parameters and in the script itself; or perhaps factor the script into two separate modules.

james-garner-canonical (commented):
I like the high-level idea. I have a number of concerns with the way this PR proposes integrating the workflow+script and its tests into the existing CI -- mostly I wonder if we could simplify it by calling the workflow/script directly and trusting the script to handle the fallback branch.

I've read parts of the script itself -- in particular the system + user prompt seem very reasonable to me. I've held of on reviewing the script further because I imagine it might churn a fair bit if we make the surrounding infra changes I'm suggesting.

test/test_ai_failure_notifier.py
tonyandrewmeyer: Ah, yes, that's because I originally had the test alongside the script, where it didn't get run with the rest, but then changed my mind and moved it to go with the others. I'll remove these.

test/test_ai_failure_notifier.py
tonyandrewmeyer: Same reasoning as above, will adjust this too.

test/test_ai_failure_notifier.py
tonyandrewmeyer: Yeah, this is why the test didn't start out there. We do have test_infra already, which has a cross ops/ops-scenario documentation test, and previously has had stuff like "are there copyright headers".

I could move it elsewhere, like back into `.github` or maybe a `.github/tests` folder or a new top-level folder (or tests subfolder?) for more infra type stuff, and things like release script tests could end up there too.

tonyandrewmeyer: Thinking about it a bit more, I like the idea of a proper tested home for this, release.py,and the other github scripts. Like a top-level `scripts` folder with a `test` folder int here, which pytest would pick up automatically, and we could get rid of the path manipulation. Not being a dot directory would mean pytest would just work both with tox and make.

A PR for that first? Intro in this one and then move the others in a follow-up?

james-garner-canonical: `.github/tests` sounds reasonable (and I could see us ending up in `.scripts/tests` or similar instead/as well if we move non-Github-CI specific stuff out of `.github` eventually). Happy for that to be a follow-up though, as long as we track it with an issue.

james-garner-canonical: I sent my reply before seeing your second reply. I'd be on board with that approach too. Happy for either PR ordering.

scripts/test/test_ai_failure_notifier.py
tonyandrewmeyer: Yes, that's the reason. It's a field the model produces (or is meant to!), so the integer typing is what rejects "#9010", "9010 " and a full URL at validation time rather than at `gh issue comment`, where we'd be failing partway through writing to a real issue. The JSON schema declares `{"type": "integer", "minimum": 1}` and the hand-validator checks the same thing, so a string would widen the accepted set to exactly the shapes we don't want. It's only stringified at the single point it's used, so the int costs nothing downstream. My inclination is to leas as-is.

james-garner-canonical: Sounds good to me.

test/test_ai_failure_notifier.py
tonyandrewmeyer: We do, which doesn't make it less ugly, though 😆. I'll try to reformat it a bit.

james-garner-canonical: A slight improvement in my mind (though still ugly) is to use parentheses to at least separate the values clearly from the keys.
```py
{
    'key': (
        'my'
        'really'
        'long'
        'value'
    ),
}
```
A multi-line string would probably be nicest, but probably runs afoul of our old friend E501 (line too long) with no ability to tack on individual line ignores.

.github/workflows/notify-scheduled-failure.yaml
tonyandrewmeyer: The enricher is now `on: workflow_call` and this workflow calls it after `open-issue`, so the `workflow_run` trigger, the hardcoded list of seven caller workflow names, and the `zizmor: ignore[dangerous-triggers]` suppression are all gone. I had tried something along those lines when experimenting with this on my fork but hadn't found this path, which doesn't seem better, thanks! (I'm double-checking it on the fork at the moment.)

I'd like to keep the coarse match as bash in its own job rather than folding it into the script:
* `open-issue` currently runs with nothing but `gh` and `github.token`. Folding it in adds a checkout, setup-uv, and the environment as prerequisites for a notification happening at all.
* `enrich` needs `environment: ai-failure-triage` for the OpenRouter key, and environments can carry protection rules including required reviewers. One job means the notification inherits that, so a misconfigured environment or an approval gate doesn't degrade the notification, it blocks it pending a human for a workflow whose whole purpose is telling people something broke overnight. Admittedly, we would presumably configure that in canonical-repo-automation so hopefully notice it, but it could happen.

The split is down to about 30 lines of bash now that it isn't also carrying the workflow_run argument, which feels ok for a notification path with no checkout, no uv and no environment on it. Happy to revisit if you still think it's not worth it, particularly once all the other changes are verified to be working.

james-garner-canonical: The current split and permissions rationale makes sense to me, thanks.

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -13,14 +23,74 @@ jobs:
...
+
+  # Stage 2. Called from here, rather than subscribing to this workflow's
+  # callers with `workflow_run`, so that there is no list of caller workflow
+  # names to keep in sync and no `dangerous-triggers` suppression to justify.
+  # `needs: [open-issue]` is what orders it after the placeholder exists;
+  # `always()` runs it even if stage 1 failed, since the enricher's own
+  # fallback handles a missing marker by creating the issue itself.
```
james-garner-canonical: I'm not sure about this -- stage 1 doesn't look like it should fail on a missing marker, instead it creates the issue. If `gh issue create` fails I don't think we want to fall through to another attempt at it, do we?

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -13,14 +23,74 @@ jobs:
...
+            issue_body="Scheduled workflow '$WORKFLOW_NAME' failed: $RUN_URL"$'\n\n'"${marker}new -->"
+            gh issue create --repo "$REPO" \
+              --title "Scheduled workflow '$WORKFLOW_NAME' failed" \
+              --body "$issue_body"
+          fi
+
+  # Stage 2. Called from here, rather than subscribing to this workflow's
+  # callers with `workflow_run`, so that there is no list of caller workflow
+  # names to keep in sync and no `dangerous-triggers` suppression to justify.
+  # `needs: [open-issue]` is what orders it after the placeholder exists;
```
james-garner-canonical:
```suggestion
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -13,14 +23,74 @@ jobs:
...
+  enrich:
+    needs: [open-issue]
+    if: ${{ always() }}
+    permissions:
+      issues: write
+    uses: ./.github/workflows/ai-failure-enrich.yaml
+
+  # If stage 1 failed *and* the enricher never got far enough to use its own
+  # fallback, nothing has notified anyone -- the one case here that needs an
+  # issue created. An un-enriched placeholder is a red job, not a lost
```
james-garner-canonical: I see you've just marked this as a draft, but since I've already typed it: I'm likewise unconvinced by this -- why will `gh issue create` succeed here if it fails in `open-issue`? (Which probably needs a rename now.)

tonyandrewmeyer:
FYI, I tried the alternative approaches in my fork and couldn't get them to work. The logs don't seem to be available until the (original) workflow finishes, so workflow_call can't get hold of them, and that means that it doesn't have the context to be able to do the enrichment. I'll see if I can find other alternatives but I suspect that moving back to the original approach may be needed.

james-garner-canonical:
> FYI, I tried the alternative approaches in my fork and couldn't get them to work. The logs don't seem to be available until the (original) workflow finishes, so workflow_call can't get hold of them, and that means that it doesn't have the context to be able to do the enrichment. I'll see if I can find other alternatives but I suspect that moving back to the original approach may be needed.

Thanks, I saw the explanation in the revert commit message too. I think the inability to get the logs until the workflow completes is all the justification we need for the `workflow_run` approach.

.github/workflows/ai-failure-enrich.yaml
tonyandrewmeyer: Claude and I were wrong about the logs not being available, there was a different problem with my testing on my fork.

So we can do the workflow call approach. The catch is that it requires a `secrets: inherit` to get the OpenRouter API key passed through (one per workflow). That's also a Zizmor warning, so we would suppress 7 (in this repo) rather than one - but it's a lesser risk I think (the workflows that have a separate secret, like to update the charm pins, expose that to the LLM workflow).

I'm going to put the PR back to that approach.

james-garner-canonical: Interesting, my brief reading of docs and stuff indicated that the logs were indeed unavailable from within a run.

Can we pass just the OpenRouter API key explicitly instead (like [this](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-inputs-and-secrets-in-a-reusable-workflow)) instead of using `secrets: inherit`?

tonyandrewmeyer: Verified on my fork, and you're right, we can pass it explicitly. All seven callers now pass OPENROUTER_API_KEY by name, the enricher declares it, and the repository is back to zero zizmor suppressions.

One weird thing: what each level passes is an empty string: no job outside the environment can read an environment secret, so the caller genuinely has nothing to hand over. The real value comes from environment: on the enrich job. But if a call site does not name the secret, the environment does not fill it in either, and the key arrives empty. Removing the "redundant" pass-through therefore breaks enrichment silently. Fork runs, if you want them: 32681905277 names it and the key arrives, 32682872399 does not and it is empty, and 32682645113 shows the caller holding length 0 while the enrich job holds the real length in the same run. I can't find this documented anywhere, so I'm not certain a it is meant to work like this, but it seems like it does...

.github/workflows/notify-scheduled-failure.yaml
tonyandrewmeyer: Fair. It had already narrowed to firing only when open-issue and enrich had both failed. open-issue creates the issue; if that failed, the enricher finds no marker and creates it itself. Both failing leaves a red workflow, which is where we were before any of this existed.

.github/workflows/ai-failure-enrich.yaml
tonyandrewmeyer: Covered by [this reply](https://github.com/canonical/operator/pull/2663#discussion_r3840406024) below.

.github/workflows/notify-scheduled-failure.yaml
tonyandrewmeyer: Both fair. The fallback job is gone (see the other threads), and open-issue is now notify, since commenting on a match is the branch that matters most and the old name only described the other one. The callers' open-issue-on-failure-if-scheduled has the same problem, but it predates this PR and is in all seven of them, so I have left it rather than widening the diff.

tonyandrewmeyer:
The infra concerns should now be addressed, and in the direction you (James) suggested: the enricher is called via workflow_call from the notifier, the workflow-level fallback is gone so the script's own fallback is the only one, and the script and its tests moved to a top-level scripts/ with scripts/test/, which the normal unit run collects with no path manipulation. The repository is also back to zero zizmor suppressions after the secrets: change on the other thread.

That should be the churn done, so the script itself is ready whenever you have time for it. I'm just going to do one more review pass myself and then will move out of draft.

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+  # `actions: read` and the OpenRouter key, where `workflow_run` would have
+  # given it an independent run needing nothing from them.
+  #
+  # `needs: [notify]` is what orders it after the placeholder exists;
+  # `always()` runs it even if stage 1 failed, since the enricher's own
+  # fallback creates the issue itself when it finds no marker. That is the
+  # last line of defence: there is deliberately no workflow-level fallback
+  # job behind it, because reaching one would mean `gh issue create` had
+  # already failed in `notify`, and a third attempt in a third job is
+  # not a reason to believe the fourth would work.
```
james-garner-canonical: I don't think we need to document the tradeoffs with `workflow_run` here. I think it's obvious that `needs: [notify]` means this runs after `notify`, and also doesn't need a comment. The `always()` comment could live on its own line, but see comment below.
```suggestion
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+  enrich:
+    needs: [notify]
+    if: ${{ always() }}
```
james-garner-canonical: I don't think this should run if `notify` fails. `notify` already just looks for a match, updates if it finds it, or creates a new issue if it doesn't. I don't see anything there that we'd expect to fail or be able to do something particularly interesting if it did (e.g. Github authentication failure, infra is down).

```suggestion
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+    permissions:
+      issues: write
+      # A called workflow cannot hold a scope its caller did not grant, so
+      # `actions: read` has to be repeated at every level of the chain for
+      # the enricher's log fetch to have it.
+      actions: read
```
james-garner-canonical, line 72: I don't think we need to document the Github workflow permission model here -- that a called workflow must have the required permissions granted by the caller. Also, the enricher workflow only has a single job (with that job requiring `actions: read`) so I'm not convinced that we need to document here that it's for log fetching.
```suggestion
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+    # This passes an empty string, and is still required. Nothing in this
+    # chain can read the OpenRouter key: it lives on stage 2's
+    # `ai-failure-triage` environment, and only the job declaring that
+    # environment resolves it. But unless the secret is named at each call
+    # site, stage 2's own environment does not fill it in either, and
+    # enrichment degrades to a plain notice with nothing failing loudly.
```
james-garner-canonical: They key piece of information here is that you require both the `secret` the `environment` to read this specific secret. But does that need to be documented here?

```suggestion
```

.github/workflows/smoke.yaml
```diff
@@ -50,4 +50,15 @@ jobs:
     needs: [test]
     permissions:
       issues: write
+      # Stage 2 reads this run's job logs, and a called workflow cannot hold
+      # a scope its caller withheld, so `actions: read` has to be granted
+      # here even though nothing in this file uses it.
+      actions: read
```
james-garner-canonical:
```suggestion
      actions: read  # for ai-failure-enrich.yaml
```
or

```suggestion
      actions: read
```

.github/workflows/smoke.yaml
```diff
@@ -50,4 +50,15 @@ jobs:
...
+    # This passes an empty string, and is still required. The OpenRouter key
+    # lives on stage 2's `ai-failure-triage` environment, which no job here
+    # can read, but unless each call site names the secret, stage 2's own
+    # environment does not fill it in either and enrichment quietly degrades
+    # to a plain notice.
```
james-garner-canonical: Again seems over-documented here. But perhaps a small comment is worthwhile, see below.
```suggestion
```

.github/workflows/smoke.yaml
```diff
@@ -50,4 +50,15 @@ jobs:
...
+    secrets:
+      OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
```
james-garner-canonical, line 56:
```suggestion
      # For ai-failure-enrich.yaml -- can only be read in the ai-failure-triage environment.
      OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
```

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+# rewrite whatever it produced into a triaged, deduplicated artefact.
+#
+# Both stages run inside the caller's run, so the run this reads logs from is
+# `github.run_id` and the workflow whose name it reports is `github.workflow`
+# -- inside a reusable workflow both refer to the original caller, which is
+# the scheduled workflow that failed.
+#
+# Keeping this in its own file, rather than as another job in the notifier,
+# keeps stage 1's guarantee legible: `notify` needs no secrets, no LLM
+# and no network beyond `gh`, and everything that does need them lives here.
```
james-garner-canonical:
```suggestion
# Called by notify-scheduled-failure.yaml, which is called by scheduled workflows on failure.
# - notify-scheduled-failure.yaml creates or comments on an issue before calling this workflow.
# - github.workflow and github.run_id refer to the scheduled workflow that failed. 
```

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+on:
+  workflow_call:
+    secrets:
+      # Declared, and passed by every level of the chain, even though what
+      # each level passes is empty. See the OPENROUTER_API_KEY comment below.
```
james-garner-canonical:
```suggestion
```

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+      OPENROUTER_API_KEY:
+        required: false
+
+permissions: {}
+
+jobs:
+  enrich:
+    runs-on: ubuntu-latest
+    environment: ai-failure-triage
+    permissions:
```
james-garner-canonical, line 31: WDYT about having the actual LLM part run in the least privileged job possible?

For example, we could drop `actions: read` from this job by having a `get-logs` job run first and save them as output. Then this job runs with `needs: get-logs` and accesses `needs.get-logs.output.logs`. But because logs are large (and you need `actions: read` to download an artifact, so that doesn't tighten our permissions any), this half probably isn't really practical.

More interesting perhaps would be dropping write permissions from the LLM job. This job could output a JSON blob that a subsequent (simple, deterministic) job with `issues: write` permissions consumes. Looking at the Python script, I think this would just mean having it output JSON after running `validate_envelope` instead of posting to Github directly. But since we only make a single request to open router and don't execute any tool calls on our system on the LLM's behalf, this is probably overkill.

So I'd say the current approach is good and we don't need to refactor anything here.

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+      issues: write
+      # Reading the failing run's job logs is an `actions: read` call, and a
+      # `permissions:` block sets every scope it does not name to `none`.
+      actions: read
```
james-garner-canonical, line 33:
```suggestion
      issues: write  # Write to the workflow failure issue.
      actions: read  # Read the failing workflow's logs.
```
or
```suggestion
      issues: write
      actions: read
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+# .github/workflows/ai-failure-enrich.yaml, is called afterwards from here
+# and rewrites what this one produced into a triaged artefact. It needs an
+# API key and an environment, and if it is unprovisioned or broken the
+# notification has already happened regardless.
 
 on:
   workflow_call:
+    secrets:
+      # Passed straight through to stage 2, which is the only thing that
+      # uses it. See the comment at the call below.
```
james-garner-canonical:
```suggestion
      # For ai-failure-enrich.yaml -- can only be read in the ai-failure-triage environment.
```

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+      # each level passes is empty. See the OPENROUTER_API_KEY comment below.
+      OPENROUTER_API_KEY:
+        required: false
+
+permissions: {}
+
+jobs:
+  enrich:
+    runs-on: ubuntu-latest
+    environment: ai-failure-triage
```
james-garner-canonical:
```suggestion
    environment: ai-failure-triage  # Required to access secrets.OPENROUTER_API_KEY
```

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+# withheld, and cannot use a secret no call site named.
+#
+# This is the first of two stages: a cheap, deterministic, always-works
+# dedup pre-check, with no LLM dependency and no secrets. The `notify`
+# job below is the whole of that guarantee -- once it has run, a
+# notification exists. The second stage,
+# .github/workflows/ai-failure-enrich.yaml, is called afterwards from here
+# and rewrites what this one produced into a triaged artefact. It needs an
+# API key and an environment, and if it is unprovisioned or broken the
+# notification has already happened regardless.
```
james-garner-canonical:
```suggestion
# `permissions: issues: write` plus `actions: read`, and grant access to
# `secrets.OPENROUTER_API_KEY`.
```

It `notify` and `enrich` need descriptions, a terse numbered list here or a short comment on each job could be added.

.github/workflows/notify-scheduled-failure.yaml
```diff
@@ -1,26 +1,100 @@
...
+          set -euo pipefail
+          marker="<!-- ai-failure-notifications:run=$RUN_ID:origin="
+
+          # Match on the workflow name alone, not the full placeholder
+          # sentence: the enricher rewrites title and body, but always
+          # preserves a "Workflow: <name>" footer line specifically so this
+          # search keeps working after enrichment. Deliberately not an
+          # `in:title` search either -- on a corpus of past scheduled
+          # failures, issues that a human had retitled kept the workflow name
+          # in the body but had lost it from the title.
```
james-garner-canonical:
```suggestion
          # Search for the workflow name in issue bodies or titles.
```

.github/workflows/notify-scheduled-failure.yaml
james-garner-canonical: WDYT about `notify` outputing the issue number for us (commented on or created), which could simplify the `ai_failure_notifier.py` script since we wouldn't need to duplicate the issue lookup logic and can just treat it as an input?

If we do want `enrich` to run even if `notify` fails, this also lets us cleanly distinguish "we were passed a specific issue" from "we have to track down an issue on our own", so I think outputting the issue number could be helpful even if we end up needing to retain issue lookup logic in the script.

pyproject.toml
```diff
@@ -108,6 +108,11 @@ include = ["ops", "ops._private", "ops.hookcmds", "ops.lib"]
 version = {attr = "ops.version.version"}
 
 # Testing tools configuration
+[tool.pytest.ini_options]
+# The workflow scripts in scripts/ are standalone files, not a package, so
+# their tests in scripts/test/ import them by bare module name.
+pythonpath = ["scripts"]
+
```
james-garner-canonical: I wonder if this would be more cleanly isolated in `scripts/test/conftest.py`, WDYT?

scripts/ai_failure_notifier.py
```diff
@@ -0,0 +1,1461 @@
+#!/usr/bin/env python3
+#
```
james-garner-canonical: We use `uv run` so we could drop the shebang here (and this wouldn't need to be executable). Alternatively, we could use a `uv` shebang here and then execute this script directly in ci (`run: scripts/ai_failure_notifier.py`).

scripts/ai_failure_notifier.py
```diff
@@ -0,0 +1,1461 @@
...
+    api_key = os.environ.get('OPENROUTER_API_KEY', '')
+    model = os.environ.get('OPENROUTER_MODEL') or DEFAULT_MODEL
+
+    try:
+        enriched_issue, origin_kind, origin_issue = locate_run_markers(repo, run_id)
+    except Exception as exc:  # search API rejection, rate limit, transient 5xx.
+        # The marker lookup is the first thing main() does, so an uncaught
+        # failure here takes out the whole enrich job and hands every run to
+        # the workflow-level plain-fallback -- losing enrichment silently
+        # rather than degrading through the script's own fallback path.
```
james-garner-canonical: This comment seems outdated.

scripts/ai_failure_notifier.py
```diff
@@ -0,0 +1,1461 @@
...
+        )
+        write_step_summary(
+            f'Rung zero: run {run_id} already enriched on #{enriched_issue}; '
+            'commented re-run note.'
+        )
+        return 0
+
+    if origin_issue is None:
+        # Shouldn't happen -- the notifier always stamps a marker -- but
+        # don't lose the notification if it does.
```
james-garner-canonical: I guess this is where we'd branch on explicit input if we had `notify` output the issue number.

james-garner-canonical (commented):
Thanks for all the work and investigation on this. I like the shape of the workflows now. I've read through them and left comments -- mostly cutting down excessive LLM comments. The only scheduled workflow I commented on is `smoke.yaml`, but my comments there apply equally to all the other scheduled workflows.

---

I have two suggestions for hand-off from `notify` to `enrich`:
1. `notify` could output the issue number, and `enrich` could take it as explicit input, so it doesn't need to fish through recent issues looking for a match -- if it doesn't get an issue number as input, then `notify` definitely failed.
2. If `notify` fails, I still think that bailing out is fine because we wouldn't expect `gh issue comment/create` to suddenly pass in the next job.

If you're not convinced by those suggestions, then we can stick with the current approach.

---

I haven't read much of `ai_failure_notifier.py` yet, just because it's 1500 LOC of code that's tricky to follow on Github.

I'm not sure what the best move here is:
- The various chunks of code do seem to be semantically grouped, but maybe it would be easier if collections of functions were grouped onto classes (even as static/class methods), or split into separate modules?
- Maybe just arranging things in a review-friendly ordering would be enough (main, then what main calls, and so on, with related functions grouped into sections ordered similarly)?
- Maybe a thorough read-through of the code isn't necessary here -- this isn't library code, so I'm not intending to review it to that standard, but do we still want someone to have read it all?

If you'd like to keep the script as-is, I'll try to read through it in an IDE instead of just on Github.

tonyandrewmeyer:
I've moved most of this to https://github.com/canonical/charm-tech-code/pull/1/ (I still need to update this PR to reflect that), and tried to address all the comments here in that PR since I can't move the comments across.

tonyandrewmeyer:
@tromai and @james-garner-canonical this is now much smaller since it uses the code in canonical/charm-tech-code. Would you prefer a fresh PR that doesn't have all the baggage here from when it had the code as well or to keep things here with all the historic context (in which case I'll update the PR body)?

james-garner-canonical:
I'd say update this one.

tromai:
@tonyandrewmeyer Thanks for the update. I'm happy with keeping it here, and update the PR.

tonyandrewmeyer:
@james-garner-canonical @tromai I've verified again that this works on my fork, so I think this is ready for review again, thanks!

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+      # means the script looks the issue up itself, which is what happens when
+      # `notify` fails before it can create one.
+      issue:
+        description: The issue the notifier created or commented on.
+        required: false
+        type: string
+      origin:
+        description: Either "new" or "comment", matching what the notifier did.
+        required: false
+        type: string
```
james-garner-canonical, line 17: We always pass these, so how about `required: true`. I'm guessing the comment is outdated since `enrich: needs: [notify]` and doesn't run on failure.

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+          #     declared above. That only ever passes the empty string, which
+          #     is why it looks like it does nothing.
+          #
+          # Take the second one out and the key arrives empty, and nothing
+          # complains: we just get a plain notice instead of an enriched one.
+          # If enrichment stops happening for no obvious reason, look here.
+          #
+          # Each repo that adopts this provisions its own key. Without one,
+          # the script writes the plain fallback body and never calls
+          # OpenRouter (see the package's cli.main()).
```
james-garner-canonical: I don't think this is helpful here.
```suggestion
```

.github/workflows/ai-failure-enrich.yaml
```diff
@@ -0,0 +1,70 @@
...
+        description: The issue the notifier created or commented on.
+        required: false
+        type: string
+      origin:
+        description: Either "new" or "comment", matching what the notifier did.
+        required: false
+        type: string
+    secrets:
+      OPENROUTER_API_KEY:
+        required: false
```
james-garner-canonical: This should probably be `required: true` as well -- we always pass the secret.

If we want a note that callers must explicitly pass the secret to grant access to it, this is where it should go.

james-garner-canonical (approved):

tromai (approved):
This is great.

Thank you for working on this feature. The setup here is very clean and nice. I like the direction of keeping the enrichment code in `charm-tech-code`.

Just one question to clarify my understanding: if `uses: ./.github/workflows/ai-failure-enrich.yaml` fails (for example, the agent has a hiccup, then the status of `notify-schedule-failure.yaml` will be FAILED too. This shouldn't be a problem because a GitHub issue/comment will still be created regardless. 
 
Very excited to see it in actions.

tonyandrewmeyer:
> Just one question to clarify my understanding: if `uses: ./.github/workflows/ai-failure-enrich.yaml` fails (for example, the agent has a hiccup, then the status of `notify-schedule-failure.yaml` will be FAILED too. This shouldn't be a problem because a GitHub issue/comment will still be created regardless.

Your understanding is right, with one wrinkle: an LLM hiccup specifically won't fail the job. The notifier catches everything around the OpenRouter call (network errors, non-2xx, unparseable JSON, and output that fails schema validation) and falls back to the plain body, same as when there's no API key configured, exiting 0 either way. It notes what happened in the step summary.

What would fail `enrich` is infrastructure: the checkout, setup-uv, the uvx fetch of the notifier, or a `gh` call. In that case, yes, the reusable workflow's conclusion is failure and the caller's job shows as failed. That's not a problem for the reason you give: `notify` runs first and has already created or commented on the issue, and the caller only reaches this at all because the scheduled run had already failed.

PR #2664, opened 2026-07-27
docs: make the custom-endpoint-name sample test actually test something

### LINKED ISSUE #2657: Correct the docs about testing custom endpoint names
[How to manage libraries > Test custom endpoint names](https://canonical.com/juju/docs/ops/latest/howto/manage-libraries/#test-custom-endpoint-names) has two problems:

- **Main problem** - The sample test can't distinguish between a correct and incorrect implementation of support for custom endpoint names. Running `start` is not enough; it needs to run a relation endpoint.

   See PR [dwilding/basic-charms#8](https://github.com/dwilding/basic-charms/pull/8), which constructs two libraries and proves that the sample test can't distinguish between them. The PR also constructs a test that _can_ distinguish between them. We should be able to base our doc updates on that test.

- **Minor problem** - The library in [How to manage libraries > Write a library](https://canonical.com/juju/docs/ops/latest/howto/manage-libraries/#write-a-library) doesn't attempt to support custom endpoint names. The first mention of custom endpoint names is in the section about testing, which is confusing.

tonyandrewmeyer:
Fixes two issues with the library docs. 

The "Test custom endpoint names" test ran `start` and asserted a `saw_start` flag, so it never touched the relation endpoint and would pass just as happily against a `DatabaseRequirer` that ignored its `endpoint` argument and observed a hard-coded relation. It now fires `relation_changed` on the custom-named relation and asserts a `DatabaseReadyEvent` is emitted.

The `DatabaseRequirer` in "Write a library" hard-coded `charm.on['database']`, so custom endpoint names were first mentioned in the testing section. It now takes `endpoint: str = 'database'` and observes `charm.on[endpoint]`, with a lead-in sentence explaining why the parameter exists.

That run also turned up why the sample `META` declares `database` alongside the custom endpoint. If only the custom endpoint is declared, a hard-coded requirer fails at charm construction with `AttributeError("'CharmEvents' object has no attribute 'database_relation_changed'")` — the test fails, but for the wrong reason. Declaring both means it fails with no `DatabaseReadyEvent` emitted, which is the defect the section is teaching about.

Fixes #2657


docs/howto/manage-libraries.md
```diff
@@ -62,7 +62,7 @@ Custom events must inherit from `EventBase`, but not from an Ops subclass of
...
 to the custom event object.
 
-For example, suppose you have a charm library wrapping a relation endpoint. The wrapper might want to check that the remote end has sent valid data and, if that is the case, communicate it to the charm. In this example, you have a `DatabaseRequirer` object, and the charm using it is interested in knowing when the database is ready. In your `lib/charms/my_charm/v0/my_lib.py` file, the `DatabaseRequirer` then will be:
+For example, suppose you have a charm library wrapping a relation endpoint. The wrapper might want to check that the remote end has sent valid data and, if that is the case, communicate it to the charm. In this example, you have a `DatabaseRequirer` object, and the charm using it is interested in knowing when the database is ready. Take an `endpoint` argument so that the charm using the library can name the relation whatever they like in `charmcraft.yaml` and pass the name through to the requirer -- the library then observes the caller-supplied endpoint rather than a hard-coded one. In your `lib/charms/my_charm/v0/my_lib.py` file, the `DatabaseRequirer` then will be:
```
dwilding: A couple of suggestions for readability:

```suggestion
For example, suppose you have a charm library wrapping a relation endpoint. The wrapper might want to check that the remote end has sent valid data and, if that is the case, communicate it to the charm.

In this example, you have a `DatabaseRequirer` object, and the charm using it is interested in knowing when the database is ready. Take an `endpoint` argument in `DatabaseRequirer` so that the charm using the library can name the relation whatever they like in `charmcraft.yaml` and pass the name through to the requirer. Then have `DatabaseRequirer` observe the caller-supplied endpoint rather than a hard-coded one. Your `lib/charms/my_charm/v0/my_lib.py` file will be:
```

docs/howto/manage-libraries.md
```diff
@@ -161,13 +165,16 @@ allow customising the name of the endpoint that the object is wrapping.
 > Examples: Traefik's [`ingress-per-unit`](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v1/ingress_per_unit.py) lib
 
 In your `tests/unit/test_my_lib.py` file, add a test that validates that custom
-names are supported:
+names are supported. Trigger a `relation_changed` event on the custom-named
+relation and assert that the library's custom event fires -- if the library
+had hard-coded the endpoint name and ignored the `endpoint` argument, no
+`DatabaseReadyEvent` would be emitted and the test would fail:
```
dwilding:
```suggestion
names are supported. Trigger a `relation_changed` event on the custom-named
relation and assert that the library's custom event fires. If the library
had hard-coded the endpoint name and ignored the `endpoint` argument, no
`DatabaseReadyEvent` would be emitted and the test would fail.
```

docs/howto/manage-libraries.md
dwilding: I think this `META` needs a `requires` block, otherwise the test will fail when `DatabaseRequirer` calls `charm.on[endpoint].relation_changed`. The library code won't be able to find an attribute called `my_relation_relation_changed`. Yet another preexisting issue with the doc.

docs/howto/manage-libraries.md
```diff
@@ -94,10 +94,11 @@ class DatabaseRequirerEvents(ops.ObjectEvents):
 class DatabaseRequirer(ops.Object):
     on = DatabaseRequirerEvents()
 
-    def __init__(self, charm: ops.CharmBase, relation_name: str):
-        super().__init__(charm, relation_name)
+    def __init__(self, charm: ops.CharmBase, endpoint: str = 'database'):
+        super().__init__(charm, endpoint)
+        self._endpoint = endpoint
```
dwilding: We don't use `self._endpoint` anywhere. But I don't object to including it to demo good practice.

docs/howto/manage-libraries.md
dwilding: I think we should mention `endpoint` here, for consistency across tests.

```suggestion
        self.db = DatabaseRequirer(self, endpoint='my-relation')
```

dwilding (approved):
Thanks!

docs/howto/manage-libraries.md
tromai: I'm not against this. 

I might miss something obvious here. Is the purpose of `self._endpoint = endpoint` here: to persist the provided `endpoint` for use later ? Is there another reason behind this practice, or any consequence of not doing this.

tromai (approved):
Thank you for the fix.

docs/howto/manage-libraries.md
tonyandrewmeyer: You're probably going to need it later, when you do whatever you're doing with the relation. Maybe not, if the library is quite delta and so you're getting the relation via event handlers. I think I'll remove it since it's not clear without expanding into more code, and at that point people presumably figure out they need to store it anyway.

PR #2684, opened 2026-08-07
fix: compare full event paths when skipping duplicate notices

### LINKED ISSUE #2683: Deferred event dedup can silently drop different events
Since the duplicate-notice skipping added in #1372 (`skip_duplicate_events`, on by
default since ops 2.17), `Framework._event_is_in_storage()` builds its
"same event, ignoring the event ID" key like this
([ops/framework.py line 852 on current main](https://github.com/canonical/operator/blob/1fdae24075724aa99176bc4b809bdb71b3b107ca/ops/framework.py#L846-L853)):

```python
or existing_event_path.split('[')[0] != event_path.split('[')[0]
```

The comment above it says the intent is to strip the trailing `[id]`:

```python
# The notices all have paths that include [id] at the end.
```

But `split('[')[0]` truncates at the first `[` in the path, not the last.
Handle paths for events emitted by keyed child objects (the standard
charm-library pattern, `super().__init__(charm, key)`) contain earlier brackets:

```
MyCharm/MyRequirer[db1]/on/ready[41]   -> split('[')[0] == 'MyCharm/MyRequirer'
MyCharm/MyRequirer[db2]/on/ready[42]   -> split('[')[0] == 'MyCharm/MyRequirer'
MyCharm/MyRequirer[db1]/on/gone[43]    -> split('[')[0] == 'MyCharm/MyRequirer'
```

All three collapse to the same key: the instance key and the event kind are both cut off. So while one such event sits deferred, any event with the same observer method and an equal snapshot  is judged a duplicate, even from a different instance or with a different event kind. It is dropped silently,

The existing tests (`test_repeated_defer`, for example) don't catch this even though their emitter is keyed: every cross-kind pair they emit happens to differ in snapshot data or in observer method, so one of the other dedup guards masks the broken path comparison.

## To reproduce

```python
import os
import ops
from ops.storage import SQLiteStorage


class ReadyEvent(ops.EventBase):
    pass


class GoneEvent(ops.EventBase):
    pass


class MyObjEvents(ops.ObjectEvents):
    ready = ops.EventSource(ReadyEvent)
    gone = ops.EventSource(GoneEvent)


class MyObj(ops.Object):
    on = MyObjEvents()


class Observer(ops.Object):
    def __init__(self, parent, key):
        super().__init__(parent, key)
        self.seen = []
        self.defer_next = 0

    def on_any(self, event):
        self.seen.append(event.handle.path)
        if self.defer_next:
            self.defer_next -= 1
            event.defer()


os.environ.setdefault('JUJU_VERSION', '0.0.0')
framework = ops.Framework(
    SQLiteStorage(':memory:'), '.', meta=ops.CharmMeta(), model=None,
)

obj_a = MyObj(framework, 'db1')
obj_b = MyObj(framework, 'db2')
observer = Observer(framework, 'observer')
framework.observe(obj_a.on.ready, observer.on_any)
framework.observe(obj_b.on.ready, observer.on_any)
framework.observe(obj_a.on.gone, observer.on_any)

observer.defer_next = 1
obj_a.on.ready.emit()   # delivered, deferred -> notice stays queued

obj_b.on.ready.emit()   # different emitter -- silently dropped
obj_a.on.gone.emit()    # different event kind -- silently dropped

print(observer.seen)
```

Output: `['MyObj[db1]/on/ready[1]']`. Expected: all three events delivered,
since neither dropped event is a duplicate of the deferred one.

Applies to ops main (1fdae2407, 3.8.1) and every release since 2.17.1.

## Suggested fix

Strip only the trailing event ID:

```python
or existing_event_path.rsplit('[', 1)[0] != event_path.rsplit('[', 1)[0]
```

I verified that with this one-line change the reproduction above delivers all
three events, and that the full `test/test_framework.py` suite still passes
(99/99). Happy to send a PR with the fix plus a regression test covering keyed
emitters and distinct event kinds.

Ali-932:
The dedup check in `Framework._event_is_in_storage()` compared notice paths truncated at the first `[`, which erases the emitter's key and the event kind for events emitted by keyed objects. While one such event sat deferred, distinct events (a different instance, or a different event kind) with equal snapshots and the same observer method were skipped as duplicates and lost.

This changes the comparison to strip only the trailing `[id]`, matching the intent described in the comment above it.

Added a regression test with two keyed emitters and two event kinds sharing one observer: it fails before the change (only 1 of 3 events delivered, `assert 1 == 3`) and passes after. The full `test/test_framework.py` suite passes (100/100).

Fixes https://github.com/canonical/operator/issues/2683

### DIFF
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -849,7 +849,7 @@ class Framework(Object):
                 # The notices all have paths that include [id] at the end. If one
                 # was somehow missing, then the split would be the empty string and
                 # match anyway.
-                or existing_event_path.split('[')[0] != event_path.split('[')[0]
+                or existing_event_path.rsplit('[', 1)[0] != event_path.rsplit('[', 1)[0]
             ):
                 continue
             # Check if the snapshot for this notice is the same.
--- a/ops/framework.py
+++ b/ops/framework.py
@@ -846,9 +846,7 @@ class Framework(Object):
             if (
                 existing_observer_path != observer_path
                 or existing_method_name != method_name
-                # The notices all have paths that include [id] at the end. If one
-                # was somehow missing, then the split would be the empty string and
-                # match anyway.
+                # The notices all have paths that include [id] at the end.
                 or existing_event_path.rsplit('[', 1)[0] != event_path.rsplit('[', 1)[0]
             ):
                 continue
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -574,12 +574,11 @@ class TestFramework:
             def __init__(self, parent: ops.Object, key: str):
                 super().__init__(parent, key)
                 self.seen: list[str] = []
-                self.defer_next = 0
 
             def on_any(self, event: ops.EventBase):
+                first = len(self.seen) == 0
                 self.seen.append(event.handle.path)
-                if self.defer_next:
-                    self.defer_next -= 1
+                if first:
                     event.defer()
 
         pub1 = MyNotifier(framework, 'db1')
@@ -590,7 +589,6 @@ class TestFramework:
         framework.observe(pub1.gone, obs.on_any)
 
         # Defer the first event so that its notice stays in the queue.
-        obs.defer_next = 1
         pub1.ready.emit()
 
         # An event from a different (keyed) emitter instance and an event of a
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -576,9 +576,8 @@ class TestFramework:
                 self.seen: list[str] = []
 
             def on_any(self, event: ops.EventBase):
-                first = len(self.seen) == 0
                 self.seen.append(event.handle.path)
-                if first:
+                if len(self.seen) == 1:
                     event.defer()
 
         pub1 = MyNotifier(framework, 'db1')
@@ -597,7 +596,11 @@ class TestFramework:
         pub2.ready.emit()
         pub1.gone.emit()
 
-        assert len(obs.seen) == 3, obs.seen
+        assert obs.seen == [
+            'MyNotifier[db1]/ready[1]',
+            'MyNotifier[db2]/ready[2]',
+            'MyNotifier[db1]/gone[3]',
+        ]
 
     def test_two_observers_one_deferring(self, request: pytest.FixtureRequest):
         framework = create_framework(request)
--- a/test/test_framework.py
+++ b/test/test_framework.py
@@ -557,6 +557,50 @@ class TestFramework:
         assert len(notices_for_observer(1)) == 0
         assert len(notices_for_observer(2)) == 4
 
+    def test_repeated_defer_distinct_events_not_skipped(self, request: pytest.FixtureRequest):
+        framework = create_framework(request)
+
+        class ReadyEvent(ops.EventBase):
+            pass
+
+        class GoneEvent(ops.EventBase):
+            pass
+
+        class MyNotifier(ops.Object):
+            ready = ops.EventSource(ReadyEvent)
+            gone = ops.EventSource(GoneEvent)
+
+        class MyObserver(ops.Object):
+            def __init__(self, parent: ops.Object, key: str):
+                super().__init__(parent, key)
+                self.seen: list[str] = []
+                self.defer_next = 0
+
+            def on_any(self, event: ops.EventBase):
+                self.seen.append(event.handle.path)
+                if self.defer_next:
+                    self.defer_next -= 1
+                    event.defer()
+
+        pub1 = MyNotifier(framework, 'db1')
+        pub2 = MyNotifier(framework, 'db2')
+        obs = MyObserver(framework, 'obs')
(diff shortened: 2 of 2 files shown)


benhoyt (approved):
Rationale and logic makes sense to me, thanks for this.

tromai (approved):
Thank you for the fix. 

I really appreciate the detailed issue write up.

test/test_framework.py
```diff
@@ -557,6 +557,50 @@ def notices_for_observer(n: int):
...
+            def __init__(self, parent: ops.Object, key: str):
+                super().__init__(parent, key)
+                self.seen: list[str] = []
+                self.defer_next = 0
+
+            def on_any(self, event: ops.EventBase):
+                self.seen.append(event.handle.path)
+                if self.defer_next:
+                    self.defer_next -= 1
+                    event.defer()
```
james-garner-canonical, line 581: It looks like the intention of this logic is to defer only the first event, since we set `obs.defer_next = 1` before the first `emit` and never touch it again. If that's correct, maybe something like this would be easier to follow:
```py
first = len(self.seen) == 0
self.seen.append(event.handle.path)
if first:
    self.defer()
```

Ali-932: Good point, i have updated the tests.

test/test_framework.py
```diff
@@ -557,6 +557,48 @@ def notices_for_observer(n: int):
...
+        class MyObserver(ops.Object):
+            def __init__(self, parent: ops.Object, key: str):
+                super().__init__(parent, key)
+                self.seen: list[str] = []
+
+            def on_any(self, event: ops.EventBase):
+                first = len(self.seen) == 0
+                self.seen.append(event.handle.path)
+                if first:
+                    event.defer()
```
tonyandrewmeyer: A nit, but it would seem simpler to be:

```suggestion
                self.seen.append(event.handle.path)
                if len(self.seen) == 1:
                    event.defer()
```

Or if you really want the local, I'd say this is more Pythonic:

```suggestion
                first = not self.seen
                self.seen.append(event.handle.path)
                if first:
                    event.defer()
```

test/test_framework.py
```diff
@@ -557,6 +557,48 @@ def notices_for_observer(n: int):
...
+        # Defer the first event so that its notice stays in the queue.
+        pub1.ready.emit()
+
+        # An event from a different (keyed) emitter instance and an event of a
+        # different kind share the observer method and an empty snapshot, but
+        # they are not duplicates of the queued notice and must be delivered.
+        pub2.ready.emit()
+        pub1.gone.emit()
+
+        assert len(obs.seen) == 3, obs.seen
```
tonyandrewmeyer: I think it would be nicer to check we have exactly the events we expect and in the expected order, rather than just that there were three, even though that's enough to show the bug.

```suggestion
        assert obs.seen == ['MyNotifier[db1]/ready[1]', 'MyNotifier[db2]/ready[2]', 'MyNotifier[db1]/gone[3]']
```

(Ruff might want to wrap that.)

ops/framework.py
```diff
@@ -849,7 +849,7 @@ def _event_is_in_storage(
                 # The notices all have paths that include [id] at the end. If one
```
tonyandrewmeyer: This comment isn't true any more. We should probably just keep the first sentence. I don't believe there is any way to have the id missing. If we want to be extra careful we should just skip de-duping if there is no ID, because we don't know what we are dealing with in that situation.

tonyandrewmeyer (approved):
Thanks for fixing this!

test/test_framework.py
james-garner-canonical: I suggested the current form because I felt that `len` made it clearer that `seen` is a collection, and went with the `first` local because testing for emptiness felt somehow cleaner than testing for a single entry (that we just appended). I think your first version is nice though, since the reader knows `seen` is a collection from the `append`.

Ali-932: done. i went with your first version since James preferred it too.

test/test_framework.py
Ali-932: done

ops/framework.py
Ali-932: done, i kept the first sentence only. For the dedup,  when there is no ID, I went and checked, in the runtime it is not reachable, `storage.save_notice` has one call site in ops, `framework._emit`, and the path it saves is `event.handle.path` where the key comes from `framework._next_event_key()` in `BoundEvent.emit`,  so the last segment is always `kind[N]`.

tonyandrewmeyer: It is now, yes. What I hadn't done is look to see if that was always the case. These are in persistent databases, so theoretically a charm could have deferred events from very old versions of ops that get processed by a new one, and that shouldn't crash.

I dug back, and while I didn't verify every possible in-between state, it seems like since 2019-09-26 it has been true (not before then, but going back pre "ops" naming seems old enough to ignore). So it should be safe to assume this.

PR #2689, opened 2026-08-13
docs: add retry loop to version check in K8s tutorial charm

### LINKED ISSUE #2703: Scheduled workflow 'Example Charm Integration Tests' failed
Scheduled workflow 'Example Charm Integration Tests' failed: https://github.com/canonical/operator/actions/runs/32869662108

dwilding:
This part of the K8s tutorial charm has a potential race:

```python
container.replan()
version = fastapi_demo.get_version(port=8000)
```

~In practice we're OK because our workload starts serving with Pebble's 1 second early-exit check. But for a production workload, it would be better to wrap `get_version()` in a retry loop.~

I observed the race in an [integration test run](https://github.com/dwilding/operator/actions/runs/32230089789/job/96259628692), so we ought to make the charm code more robust.

This PR:

- Updates the charm code to include a retry loop (in all five charms)
- In "Create a minimal Kubernetes charm":
    - Adds an explanation of why the loop is needed.
    - Adds an instruction to import `time` and `urllib.error`.
    - Moves the "Add logger functionality" section earlier so that we've introduced `logger`.

**[Preview doc](https://canonical-juju-ops--2689.com.readthedocs.build/2689/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm/#set-the-workload-version)**


docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -209,16 +209,19 @@ The `fastapi` service has `startup: enabled`, so why does our charm need to star
 
 The workload version is available after the workload starts, which happens after Pebble starts the `fastapi` service. We'll use the `src/fastapi_demo.py` helper module for this step.
 
-In `src/charm.py`, add the following lines to the `_on_demo_server_pebble_ready` function before the final `self.unit.status = ops.ActiveStatus()`:
+In `src/charm.py`, add the following lines to the `_on_demo_server_pebble_ready` function after the `container.replan()` line:
```
dwilding, line 223: I'm changing this part so that talking about `replan()` later makes more sense.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -209,16 +209,19 @@ The `fastapi` service has `startup: enabled`, so why does our charm need to star
...
 ```python
 # Set the workload version of this charm.
+# For a production workload, wrap get_version() in a retry loop.
 version = fastapi_demo.get_version(port=8000)
 ```
 
 We get the workload version over port 8000 because the `fastapi` service runs the app on this port. Then `self.unit.set_workload_version` exposes the workload version to Juju. If the `get_version` call fails (for example, an `URLError` exception is raised), the charm will go into error status. The Juju logs will show the error message, to help you debug the error.
 
+For a general workload, the `get_version` call could fail because the workload's service hasn't fully started. `container.replan()` tells Pebble to start the service, but doesn't wait to confirm that the service has finished starting up. Pebble actually waits one second to check that the service didn't crash on startup, which in our case is enough time for the `fastapi` service to fully start.
```
dwilding: I'm not sure whether "general workload" or "production workload" would be better here.

tromai: I think `For a general workload, the get_version ...` might give the impression that any workload would have the same `get_version` function. For me, `get_version` is specific to our tutorial and `api_demo_server` application. `production workload` seems to be better here, and consistent with the code comment `For a production workload ...` above.

Thinking more about it. I wonder if the difference is more about the environment where the application is deployed (in production, `fastapi` can take longer to start), or more about the type of workload (`api_demo_server` or another application). I assume it is the first, how about:

```
# In production, wrap get_version() in a retry loop.
```

And

```
In production, the `get_version` call could fail ...
```

dwilding: I was thinking the distinction is more about the type of workload. But my "in production" wording is muddling that, isn't it. I think our tutorial code should be production-ready - it just happens that our workload has a small footprint.

Maybe the best answer is to actually just include the retry loop in the charm code. Or build it into the `get_version()` function in the first place. What do you you think?

tromai: Thanks for the clarification. I understood it wrong, sorry. 

My impression from the internal discussion: we aimed for production ready code in the tutorial, but we wanted to keep the tutorial code simple in this specific case. I still think adding a comment is a sensible option.

> it just happens that our workload has a small footprint.

In that case, `For a general workload` would work well. In our tutorial, this is not a major risk, but Charmers should be careful when working with their own workload. Not sure if this is okay: can we give an example in the tutorial, without having it in the example code ?

What I have in mind:

> For a general workload, the `get_version` call could fail because the workload's service hasn't fully started. `container.replan()` tells Pebble to start the service, but doesn't wait to confirm that the service has finished starting up. Pebble actually waits one second to check that the service didn't crash on startup, which in our case is enough time for the `fastapi` service to fully start.
> However, have a retry loop when working with another workload, since it can take longer to fully start:
> Example: a small snippet of retry code in `charm.py`

dwilding: Thanks for your help with this. I just made some changes based on your suggestions:

- Changed "For a production workload" to "For a general workload" in the comment.
- Added a generic example of a retry loop after replan, inspired by how the Charmcraft profile does it.

See the [preview doc](https://canonical-juju-ops--2689.com.readthedocs.build/2689/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm/#set-the-workload-version). How does it look now?

tromai: It looks great, Thank you!

tromai (approved):

dwilding:
Moving this back into draft status. An [integration test run](https://github.com/dwilding/operator/actions/runs/32230089789/job/96259628692) hit the race condition, so we ought to update the charm code. I'll propose an update.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -184,7 +195,7 @@ def _on_demo_server_pebble_ready(self, event: ops.PebbleReadyEvent) -> None:
     self.unit.status = ops.ActiveStatus()
 ```
 
-We're using the `ActiveStatus` class to set the charm status to active. Note that almost everything you need to define your charm is in the `ops` package that you imported earlier - there's no need to add additional imports.
+We're using the `ActiveStatus` class to set the charm status to active. Note that almost everything you need to define your charm is in the `ops` package that you imported earlier.
```
dwilding, line 198: I'm removing the last part because we actually do need additional imports now.

dwilding:
@tromai I updated the charm code and tutorial to include a retry loop - see the updated PR description for details. Would you mind reviewing again? Thanks!

I also thought how we might test what the charm does when the race occurs. I think we'd have to change `mock_version` in the unit tests so that it fails a certain number of times, or until sufficient time has passed. It seems fiddly for the tutorial, so I didn't include it. What do you think?

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -209,26 +220,32 @@ The `fastapi` service has `startup: enabled`, so why does our charm need to star
...
-In the imports section of `src/charm.py`, import the Python `logging` module and define a logger object, as below. This will allow you to read log data in `juju`.
+The `get_version` call could fail because the workload's service hasn't fully started. `container.replan()` tells Pebble to start the service, but doesn't wait to confirm that the service has finished starting up. Pebble actually waits one second to check that the service didn't crash on startup, but this isn't necessarily enough time for the `fastapi` service to fully start.
 
-```python
-import logging
-
-# Log messages can be retrieved using juju debug-log
-logger = logging.getLogger(__name__)
-```
+If `get_version` hasn't succeeded after three attempts, the charm will go into error status. The Juju logs will show the error message, to help you debug the error. The error message is for you, the charm developer, not for the user of the charm. It tells you there's a bug in the charm.
```
tromai: Could we add a description on how to get `Juju logs` ? For example:

```
The Juju logs (from `juju debug-log`) will show you the error message, to help you debug the error. 
```

I don't have a strong opinion on this, since we mentioned `juju debug-log` in the `Add logger functionality` at the beginning. I'm happy to go with your decision.

tromai (approved):
Thanks for the PR. It looks good to me.

> I think we'd have to change mock_version in the unit tests so that it fails a certain number of times, or until sufficient time has passed.

Ah, that makes sense. 

I originally thought about an "un-happy" path test, where I just mock `get_version` to always trigger `urllib.error.URLError`. The test would run `ctx.run(ctx.on.pebble_ready(container), state_in)` and expects RuntimeError. An example:

```python
    mock_sleep = Mock()
    monkeypatch.setattr("time.sleep", mock_sleep)
    mock_get = Mock(side_effect=urllib.error.URLError("not ready"))
    monkeypatch.setattr("fastapi_demo.get_version", mock_get)

    ctx = testing.Context(FastAPIDemoCharm)
        name="demo-server", can_connect=True, layers={"rock": ROCK_LAYER}
    )
    state_in = testing.State(
        containers={container},
        leader=True,
    )

    with pytest.raises(RuntimeError, match="workload is not available"):
        ctx.run(ctx.on.pebble_ready(container), state_in)
```

I agree that the setup can be fiddly. I am happy to exclude this test.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: Added your suggestion - thanks :slightly_smiling_face:

dwilding:
@hpidcock this is ready for review now after rework - thanks!

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
tromai: Thank you. I reviewed this change. The PR looks good.

My green tick is still there so I didn't press the approve button another time.

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
```diff
@@ -209,26 +220,32 @@ The `fastapi` service has `startup: enabled`, so why does our charm need to star
...
+    if attempt:
+        time.sleep(1)  # If not the first attempt, wait before retrying.
+    try:
+        version = fastapi_demo.get_version(port=8000)
+        break
+    except urllib.error.URLError:
+        continue
+else:
+    logger.error("The workload was not available within the expected time")
+    raise RuntimeError("workload is not available")
```
hpidcock, line 238: Showing a more durable retry strategy  in tutorials is a good idea.
```suggestion
# The workload may not be ready immediately after replan(), so try get_version() in a loop.
for attempt in range(6):
    if attempt:
        time.sleep(2 ** attempt)  # If not the first attempt, an exponential back-off before retrying.
    try:
        version = fastapi_demo.get_version(port=8000)
        break
    except urllib.error.URLError:
        continue
else:
    logger.error("The workload was not available within the expected time")
    raise RuntimeError("workload is not available")
```

hpidcock (approved):
LGTM

docs/tutorial/from-zero-to-hero-write-your-first-kubernetes-charm/create-a-minimal-kubernetes-charm.md
dwilding: It's overkill for this particular charm, because the workload really does just need a couple more seconds to get ready. If it's going to fail for someone working through the tutorial, I would want it to fail in less than ~1 minute.

But I get your point that we want to teach a robust strategy. And now I remember that we use exponential back-off in the httpbin demo charm too ([here](https://github.com/canonical/operator/blob/main/examples/httpbin-demo/src/charm.py#L113-L124)).

How about this:

```python
for attempt in range(3):  # In general, allow more attempts for a complex workload.
```

